PR #3987, opened 2021-05-23
Added the amphibious promotion and the marine unit

xlenstra:
This introduces the Amphibious promotion, and adds the marine units from the Gods&Kings expansion.

Due to my inability to correctly handle github, this pull request contains 6 commits, with only the last one having the actual changes, so feel free to squash this into a single commit before merging.


SomeTroglodyte:
I feel ya... My approach when a PR is ready but _not_ yet published (leads to chaos when tried for an update to an existing PR - which is why I so often close PR's and reopen replacements - there must be an easier way even then but I don't understand that situation well enough): Simply delete the _online_ branch, bring your master up to date, then checkout-and-rebase-onto-master the branch you're working on, make absolutely sure local git has noticed the online copy is gone and re-push... But wait, you worked this on your master - in that case unsure, may even work the same with the rebase happening in the pull... Did I describe this well enough in my [notes](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Unciv-notes/Working%20in%20the%20IDE%20and%20Contributing.md)?

SomeTroglodyte:
...But other than Marines "Introduced in Gods & Kings", seems to conform to how Civ5 did it.

android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
```diff
@@ -135,6 +135,13 @@
 		"effect": "Double movement rate through Forest and Jungle",
 		"unitTypes": ["Melee"]
 	},
+	{
+		"name": "Amphibious",
+		"prerequisites": ["Shock I", "Drill I"],
+		"effect": "Eliminates combat penalty for attacking over a river",
```
yairm210, line 141: Remove 'effect' and put this also in "uniques", so it has 2 strings

yairm210:
Since we're adding both a unit and a promotion, we'll need a new promotion icon in /android/images/unitpromotionicons and a new unit icon in android/imagestopackseparately/uniticons
Both white on transparent, unit icon 100*100px, promotion 50*50px

xlenstra:
They should be added now, though my commit history has not become cleaner

PR #3757, opened 2021-04-02
Redraw TechPickerScreen and PickerScreen, Make map symmetrical if it's not wrapped

lishaoxia1985:

Issue #3892, opened 2021-05-07
Load game screen buttons are part of saved games list


jahodovykolac:
**Describe the bug**
The buttons like Delete save, Load from clipboard etc. are moving together with save files.

**To Reproduce**
1. Save some games or have a long list
2. Go to 'Load game'
3. The buttons are not on the top, they're in the middle of the list, moving together with the list.

If game data is required, please add from menu, save game, copy to clipboard, paste here

**Expected behavior**
The list is always on the same place.

**Screenshots**
![Screenshot_20210507-111411_UnCiv](https://user-images.githubusercontent.com/79529669/117427543-6f3f4400-af25-11eb-8ed1-0fad268d5c22.jpg)
![Screenshot_20210507-111417_UnCiv](https://user-images.githubusercontent.com/79529669/117427565-76fee880-af25-11eb-9478-59ff5b0eea1b.jpg)

**Additional context**
I saw it on 3.14.6-patch2, any Android

SomeTroglodyte:
I've been itching to fix the missing padding between three of the widgets - so when i get around to that I might have a look. 

Of course, ModManager would have had the same problem and has patched its way around the single ScrollPane built into PickerScreen, so some system to allow all PickerScreens to have more control over their central area would be in order...

SomeTroglodyte:
Other screens plural? I remember `ModManagementScreen` - what else? Point is, Save/Load need columns scrolling independently, not no scrolling at all - like `ModManagementScreen` does for  2 of its 3 columns. But if several `PickerScreen` subclasses need to disable features of their superclass, then the superclass should offer better features... Or we could do another layer of subclasses - PickerScreenBase, PickerScreenSingleColumn, PickerScreenTwoColumn, PickerScreenThreeColumn...

yairm210:
If anything, all it needs is a function of disableScroll(), or maybe a boolean isScrollEnabled=true that can be set to false. An entire new class is overkill :)

SomeTroglodyte:
Should have looked at the sources earlier - I ass-umed this screen didn't have a separate scrollpane for the existing saves, but it does -> you're absolutely right. And a helper in the parent class to 'set up' a multicolumn topTable for its clients is overkill.

PR #3991, opened 2021-05-24
Revert "Revert "Redraw TechPickerScreen and PickerScreen, Make map sy…

lishaoxia1985:
…mmetrical if it's not wrapped (#3757)""

This reverts commit 6d5cb002715aeb41afec0a79f6b0e2c2e9519547.

In class pickscreen, the var "disableScroll" have been never used.

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
@@ -6,13 +6,12 @@ import com.unciv.UncivGame
 import com.unciv.ui.utils.*
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
-open class PickerScreen(val disableScroll: Boolean = false) : CameraStageBaseScreen() {
+open class PickerScreen : CameraStageBaseScreen() {
 
     internal var closeButton: TextButton = Constants.close.toTextButton()
     protected var descriptionLabel: Label
-    protected var rightSideGroup = VerticalGroup()
+    private var rightSideGroup = VerticalGroup()
     protected var rightSideButton: TextButton
-    private var screenSplit = 0.85f
 
     /**
      * The table displaying the choices from which to pick (usually).
@@ -20,15 +19,15 @@ open class PickerScreen(val disableScroll: Boolean = false) : CameraStageBaseScr
      */
     protected var topTable: Table
     var bottomTable:Table = Table()
-    internal var splitPane: SplitPane
     protected var scrollPane: ScrollPane
 
     init {
+        val bottomTableHeight = 150f
         bottomTable.add(closeButton).pad(10f)
 
         descriptionLabel = "".toLabel()
         descriptionLabel.wrap = true
-        val labelScroll = ScrollPane(descriptionLabel,skin)
+        val labelScroll = ScrollPane(descriptionLabel)
         bottomTable.add(labelScroll).pad(5f).fill().expand()
 
         rightSideButton = "".toTextButton()
@@ -36,18 +35,16 @@ open class PickerScreen(val disableScroll: Boolean = false) : CameraStageBaseScr
         rightSideGroup.addActor(rightSideButton)
 
         bottomTable.add(rightSideGroup).pad(10f).right()
-        bottomTable.height = stage.height * (1 - screenSplit)
 
         topTable = Table()
         scrollPane = ScrollPane(topTable)
 
-        scrollPane.setScrollingDisabled(disableScroll, disableScroll)
-        scrollPane.setSize(stage.width, stage.height * screenSplit)
-
-        splitPane = SplitPane(scrollPane, bottomTable, true, skin)
-        splitPane.splitAmount = screenSplit
-        splitPane.setFillParent(true)
-        stage.addActor(splitPane)
+        val pickerScreenTable = Table()
+        pickerScreenTable.add(scrollPane).height(stage.height - bottomTableHeight - 2f).row()
+        pickerScreenTable.addSeparator()
+        pickerScreenTable.add(bottomTable).height(bottomTableHeight).fillX().expandX().row()
+        pickerScreenTable.setFillParent(true)
+        stage.addActor(pickerScreenTable)
     }
 
     fun setDefaultCloseAction(previousScreen: CameraStageBaseScreen?=null) {
--- a/core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
@@ -15,6 +15,7 @@ import com.unciv.models.translations.tr
 import com.unciv.ui.utils.*
 import java.util.*
 import kotlin.collections.ArrayList
+import kotlin.math.max
 
 
 class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Technology? = null) : PickerScreen() {
@@ -60,7 +61,7 @@ class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Tec
 
         createTechTable()
         setButtonsInfo()
-        topTable.add(techTable)
+        topTable.add(techTable).fill().expand()
 
         rightSideButton.setText("Pick a tech".tr())
         rightSideButton.onClick(UncivSound.Paper) {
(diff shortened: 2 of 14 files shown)


core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
```diff
@@ -95,30 +96,30 @@ class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Tec
...
             if (!erasNamesToColumns.containsKey(era)) erasNamesToColumns[era] = ArrayList()
-            val columnNumber = tech.column!!.columnNumber
-            if (!erasNamesToColumns[era]!!.contains(columnNumber)) erasNamesToColumns[era]!!.add(columnNumber)
+            val columnNumber = technology.column!!.columnNumber
+            if (!erasNamesToColumns[era]!!.contains(columnNumber))
+                erasNamesToColumns[era]!!.add(columnNumber)
         }
         var i = 0
-        for ((era, columns) in erasNamesToColumns) {
+        erasNamesToColumns.forEach { (era, columns) ->
```
SomeTroglodyte, line 113: No minimum API level warning? There was one in one of those collections...

core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
```diff
@@ -6,48 +6,45 @@ import com.unciv.UncivGame
 import com.unciv.ui.utils.*
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
-open class PickerScreen(val disableScroll: Boolean = false) : CameraStageBaseScreen() {
```
SomeTroglodyte, line 9: That was there for a reason

SomeTroglodyte: Tested and yes, this breaks at least load game and mod manager screens.
I admit there's other possible solutions - maybe those PickerScreen's actually using that ScrollPane are in the minority so the creation could be removed and refactored to those specifically - or another layer of subclass or or or...

core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
```diff
@@ -6,48 +6,45 @@ import com.unciv.UncivGame
...
-        scrollPane.setScrollingDisabled(disableScroll, disableScroll)
-        scrollPane.setSize(stage.width, stage.height * screenSplit)
-
-        splitPane = SplitPane(scrollPane, bottomTable, true, skin)
-        splitPane.splitAmount = screenSplit
-        splitPane.setFillParent(true)
-        stage.addActor(splitPane)
+        val pickerScreenTable = Table()
+        pickerScreenTable.add(scrollPane).height(stage.height - bottomTableHeight - 2f).row()
+        pickerScreenTable.addSeparator()
```
SomeTroglodyte, line 44: The `SplitPane` handle is a different colour than the separator - I liked the version with a little less contrast better. Alternatively, one could argue that maybe _all_ separators should use the SplitPane handle colour from the skin... Like ... like [here](https://github.com/SomeTroglodyte/Unciv/blob/b531404cb18db401cd0e1bdab7b38c110706d86a/core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt#L120).

core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
```diff
@@ -285,4 +286,9 @@ class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Tec
         }
     }
 
+    override fun resize(width: Int, height: Int) {
+        if (stage.viewport.screenWidth != width || stage.viewport.screenHeight != height) {
+            game.setScreen(TechPickerScreen(civInfo))
+        }
+    }
```
SomeTroglodyte, line 293: Not the last place where resizing throws you unexpectedly back, but ++:+1:

SomeTroglodyte (changes requested):
So, please test with the mod manager and load screen (needs lots of saves to show in the latter) so you see the reason behind the PickerScreen.ScrollPane master switch.

Also - you touched the ZoomableScrollPane - making the tech tree scroll horizontally instead of vertically for a few pixels would take the same approach, but simpler - just switch X and Y. Actually, disabling vertical scroll will also switch the wheel to scroll horizontally, but, well, we need that too... Think that worthwhile?

core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
yairm210: Even without the minimum API, for() is better than forEach() for debug and readabillity (although readability here is debatable)

core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
lishaoxia1985: SplitPane is not very good if you want to assign an exact width or height for widget

SomeTroglodyte: I believe you.Also using SplitPane when you know it's never gonna be dragged is some unnecessary overhead.  I was talking about visual appearance only.

lishaoxia1985:
@yairm210 about this, how do you think🌝🌝

yairm210:
Screens that need to be checked:
- tech picker screen
- mod management screen
- improvement picker screen
- policy picker screen
- promotion picker screen
- great person picker screen
- save game screen
- load game screen
- save/load map screen
- new game screen
- new map screen
- victory screen

Most of these should be super easy to check - just enter the screen and check that visually everything is ok
Only the great person picker requires any setup

SomeTroglodyte:
I don't want to nitpick, but you're reverting issue #3892:
image
image

yairm210:
Did you check all screens?

lishaoxia1985:
If you need, I will shortcut some pictures 🌝🌝

yairm210:
As long as you checked, I'm fine with that :)

SomeTroglodyte:
I just pulled the branch and it was unchanged, so I _guess_ my screenshots of the regression still apply. I'll test the new master then.

SomeTroglodyte:
https://user-images.githubusercontent.com/63000004/121412045-8ded8c00-c964-11eb-8d5f-354212431c40.mp4

yairm210:
Oh
Back to the drawing board I guess
Sorry @lishaoxia1985

lishaoxia1985:
@SomeTroglodyte I DON'T UNDERSTAND.
I cannot find the bug.
It seams to appear in mod screen.

SomeTroglodyte:
Oh - I _know_ you're capable, must be one of those mental blocks we all have. Sorry then I wasn't clearer back in [this comment](https://github.com/yairm210/Unciv/pull/3991#discussion_r638153668).

PickerScreen as a base class is an 'offer' with a structure, and by default that has a top part including a scrollable container for whatever the 'clients' want to put in - `topTable`. Once placed, removing that is - well - a bit ugly - you cannot remove e.g. Table cells but you can replace their actors - anyway, not all 'clients' of PickerScreen can use that as is, some need several  _independently_ scrolling columns instead. 

Instead of searching for the ScrollPane actor (it's in the SplitPane) and using splitPane.removeActor on it then adding a replacement in the splitPane they decided to lock that ScrollPane in place instead, effectively killing its function; and then to add a new Table for the columns and into this new independent ScrollPane's... That was done separately first, later I moved the disabling - and only that - into PickerScreen, controlled by a primary constructor parameter. I had the 'client's use the fully named parameter notation because that is color-coded nicely, and immediately catches the eye, so you know this is one of those special PickerScreen's - and can easily do a global search to find all of them.

An aside: The actual reason the 'wrong' scrollpane 'takes' those events surprised me just a few pushes ago when looking at keyboard events. To explain - a bit more background: That mod manager has five columns: 1 empty, 1 installed-mods - scrolling, 1 online-mods - scrolling, 1 information - static, 1 empty. Ignore the empty - they just help auto-width and centering. You put your pointer over the center column and drag - why doesn't the center ScrollPane (you didn't remove it after all, and all its config is there) react first as we would expect it to be on top in z-order? Well, the stage manages the listeners, both listeners are in its listeners array - and Gdx fires events in ascending array index order. The _bottom_ ScrollPane receives the event first and marks it as handled right away, so the top one never hears about it... (I hope I got that right from memory)

So, you 'just' need to remove or disable that bottom ScrollPane's "listening to events" - find and remove the actual listener from the stage or set touchable to 'nope' or use setScrollingDisabled() on it... Which is exactly what's there. Sorry, _was_ there.

The original PickerScreen 'clients' using this were:
image

I would suggest:
- You stash that branch locally - e.g. file copy or using the 'create patch feature' or using git features I don't know enough about. I feel comfortable with Meld or WinMerge (external tools) so I do selective folder tree copies.
- You close the PR and delete the online branch
- Leave the local branch - as backup only
- Checkout master and make sure it's level with upstream master - various methods, look in my user->Scratchpad->notes, somewhere I have text to help myself relearn that stuff next time
- Create a new branch (please with a more descriptive name) and check it out of course
- Selectively and step-by-step merge the copy from step 1 into it, looking for the ScrollPane issue and leaving that alone, and looking for whatever you originally intended to improve and merge that in.
- Stick to the minimum set of changes that achieves the first goal and offer a PR. Don't mix goals.
- Several goals achieved in your backup? Wait for merge and then sync your master, new branch, continue merging from that backup.

OK, and now for some guessing: (a Monty Python reference - they originally used "And now for something completely different" as running gag). Sorry I didn't read all the discussion or the referenced original PR entirely, so I'm surely missing information.
- From "Map symmetrical" I could guess that you noticed rectangular maps were once limited to multiples of 2, but they are no longer. custom rectangular sizes are followed exactly now, meaning a map with both coords even has 1 symmetry - rotation by 180° - one axis even one odd you get one mirror plane - both odd you get two mirror planes and some rotational symmetry - both odd and equal - maximum symmetry. I don't think limiting that to maximize symmetry is good - but that's debatable / matter of taste.
- "TechPickerScreen" - I never looked too closely on that one so I trust you - both that if you want to touch it then there's room for improvement and you to be able to achieve it. Now you know what's up with that ScrollPane you'll handle it just fine - or come up with an even better, this time elegant, solution for the usecase "I need several variants of PickerScreen with varying counts of independently scrolling columns".

Have fun!

lishaoxia1985:
You said Too complex:
1. I think your mean is that not all the pickscreens need scollpane.
the better way is:
in class pickscreen, only create toptable and bottomtable, the code of adding scrollpane is written in pickscreen child class.
The much better way is:
in class pickscreen, create toptable centertable and bottomtable, the class pickscreen can be even used in overview screen (use this construction pickscreen(showtoptable = true, showcentertable = true, showbottomtable = true)). Do you agree?
2. About Map symmetrical, it is about this:
in mainmenuscreen, there is a map as the background.
If your setting map is wrapped in recent game, so the background map is, and then you will find the bug: the backgroundmap is not in the center of screen (or rather about 1 tilegroup width), so we should fix it.
3. I find you upload a video which shows a bug that bottomtable disappears, but i cannot repeat this bug, can you give me some message.

PR #3992, opened 2021-05-24
Improved location notifications

SomeTroglodyte:
Went through all uses of the `addNotification(String,Vector2...` overload and checked whether more locations might be useful. I think they might be... e.g. Forest cut -> show where the forest was and the recipient city, show both attacker and defender in battles and so on.
Found a todo close by and obeyed (Citadel notification for attacking civ if victim destroyed).
Also 'linted' all files touched until 'Problems' window empty.
Had my doubts about `CivInfo.getLeaderDisplayName` - shouldn't it reduce to undecorated `Nation.getLeaderDisplayName` in purely single-player games? Well, left logic as it was just followed the IDE's 'lift out and fold' moanings.

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -638,15 +638,16 @@ class CivilizationInfo {
     }
 
     fun giftMilitaryUnitTo(otherCiv: CivilizationInfo) {
-        val city = NextTurnAutomation.getClosestCities(this, otherCiv).city1
+        val cities = NextTurnAutomation.getClosestCities(this, otherCiv)
+        val city = cities.city1
         val militaryUnit = city.cityConstructions.getConstructableUnits()
                 .filter { !it.unitType.isCivilian() && it.unitType.isLandUnit() }
                 .toList().random()
         // placing the unit may fail - in that case stay quiet
         val placedUnit = placeUnitNearTile(city.location, militaryUnit.name) ?: return
-        // Point to the places mentioned in the message in that order
+        // Point to the places mentioned in the message _in that order_ (debatable)
         val placedLocation = placedUnit.getTile().position
-        val locations = LocationAction(listOf(getCapital().location, placedLocation, city.location))
+        val locations = LocationAction(listOf(placedLocation, cities.city1.location, city.location))
         addNotification("[${otherCiv.civName}] gave us a [${militaryUnit.name}] as gift near [${city.name}]!", locations, otherCiv.civName, militaryUnit.name)
     }
 
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -647,7 +647,7 @@ class CivilizationInfo {
         val placedUnit = placeUnitNearTile(city.location, militaryUnit.name) ?: return
         // Point to the places mentioned in the message _in that order_ (debatable)
         val placedLocation = placedUnit.getTile().position
-        val locations = LocationAction(listOf(placedLocation, cities.city1.location, city.location))
+        val locations = LocationAction(listOf(placedLocation, cities.city2.location, city.location))
         addNotification("[${otherCiv.civName}] gave us a [${militaryUnit.name}] as gift near [${city.name}]!", locations, otherCiv.civName, militaryUnit.name)
     }
 
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -302,7 +302,8 @@ class CivilizationInfo {
 
     fun getEquivalentUnit(baseUnitName: String): BaseUnit {
         val baseUnit = gameInfo.ruleSet.units[baseUnitName]
-            ?: throw UncivShowableException("Unit $baseUnitName doesn't seem to exist!")
+        @Suppress("FoldInitializerAndIfToElvis")
+        if (baseUnit == null) throw UncivShowableException("Unit $baseUnitName doesn't seem to exist!")
         if (baseUnit.replaces != null) return getEquivalentUnit(baseUnit.replaces!!) // Equivalent of unique unit is the equivalent of the replaced unit
 
         for (unit in gameInfo.ruleSet.units.values)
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -442,16 +443,16 @@ class MapUnit {
     private fun tryProvideProductionToClosestCity(removedTerrainFeature: String) {
         val tile = getTile()
         val closestCity = civInfo.cities.minByOrNull { it.getCenterTile().aerialDistanceTo(tile) }
-        if (closestCity == null) return
+            ?: return
```
yairm210, line 446: Disagree

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -636,8 +642,12 @@ class CivilizationInfo {
         val militaryUnit = city.cityConstructions.getConstructableUnits()
                 .filter { !it.unitType.isCivilian() && it.unitType.isLandUnit() }
                 .toList().random()
-        placeUnitNearTile(city.location, militaryUnit.name)
-        addNotification("[${otherCiv.civName}] gave us a [${militaryUnit.name}] as gift near [${city.name}]!", city.location, otherCiv.civName, militaryUnit.name)
+        // placing the unit may fail - in that case stay quiet
+        val placedUnit = placeUnitNearTile(city.location, militaryUnit.name) ?: return
+        // Point to the places mentioned in the message in that order
+        val placedLocation = placedUnit.getTile().position
+        val locations = LocationAction(listOf(getCapital().location, placedLocation, city.location))
```
yairm210, line 649: Why the capital? I think just the placed location is needed

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -573,9 +576,12 @@ class CivilizationInfo {
         val cityToAddTo = city ?: cities.random()
         if (!gameInfo.ruleSet.units.containsKey(unitName)) return
         val unit = getEquivalentUnit(unitName)
-        placeUnitNearTile(cityToAddTo.location, unit.name)
-        if (unit.isGreatPerson())
-            addNotification("A [${unit.name}] has been born in [${cityToAddTo.name}]!", cityToAddTo.location, unit.name)
+        // silently bail if no tile to place the unit is found
+        val placedUnit = placeUnitNearTile(cityToAddTo.location, unit.name) ?: return
+        if (unit.isGreatPerson()) {
+            val locations = LocationAction(listOf(placedUnit.getTile().position,cityToAddTo.location))
```
yairm210, line 582: No need for the city location, I feel.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -301,7 +302,7 @@ class CivilizationInfo {
 
     fun getEquivalentUnit(baseUnitName: String): BaseUnit {
         val baseUnit = gameInfo.ruleSet.units[baseUnitName]
-        if (baseUnit == null) throw UncivShowableException("Unit $baseUnitName doesn't seem to exist!")
+            ?: throw UncivShowableException("Unit $baseUnitName doesn't seem to exist!")
```
yairm210, line 305: Same here. I feel the existing was more immediately obvious

core/src/com/unciv/logic/map/MapUnit.kt
SomeTroglodyte: See below

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SomeTroglodyte: The capital of the gifting civ - the city state - you don't always remember where they are, even if the gift appears at your closest city. Scouts explore _much_ quicker than your empire expands, right?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SomeTroglodyte: OK but the null pointer fix stays - subject to out Elvis policy below?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SomeTroglodyte: Doesn't your Android Studio complain? Alternatives would be a `@Suppress("FoldInitializerAndIfToElvis")` or disabling the inspection project-wide - but that would be local I think and not part of what we're syncing as source, so every dev would need their own and we should have documentation?

I'm not overly fond of every inspection myself, but they're there and express opinions of the Kotlin team, right? Ignoring them is like swimming against some social trend - often goes well for a while then you're suddenly an outsider. `?: return` is nice enough, plus the val is actually the simpler type right away - so `?: throw` is no different really... Like the `minByOrNull{}` - ugly as hell but hey we're being forced?

All this of course subject to - _do_ we want Android Studio's 'Problems' window `Alt-6` to be as useful as possible? If it's empty when we start a change then any stuff cropping up here will be immediately obvious....

SomeTroglodyte:
Push is a bit tongue-in-cheek, but hey, same file:
```kotlin
    fun setNationTransient() {
        nation = gameInfo.ruleSet.nations[civName]
                ?: throw java.lang.Exception("Nation $civName is not found!")
    }
```
Not my code, and  _Evil Elvis_! But I'd sooner criticize why a _java_ Exception not a Kotlin one?

So, you're right, consistent policy project-wide trumps (or should).

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
yairm210: I see
This is a misunderstanding based on lies.
This function *actually* just gives YOU a unit, and says that the unit came from the city state (this is also an open issue)
Therefore the capital in this case will also be your capital, until we fix the function
So we can change it to otherCiv.getCapital()

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
yairm210: A. There's definitely a bug difference between 'deprecated, do not use' to 'we think this is better but the other is ok too'
B. I think I'll need to rethink my approach to elvis operators. I still don't think it's add transmit, but I'll meditate on this.

SomeTroglodyte: Meditating is good. Sorry I discovered the 'Problems' window as new toy - but perhaps you agree to start making it useful...

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SomeTroglodyte: Ah -
```
                if (!hasFlag(DiplomacyFlags.ProvideMilitaryUnit)) setFlag(DiplomacyFlags.ProvideMilitaryUnit, 20)
```
runs on _my_ civ, so the citystate is the `otherCiv` - and the function should be called `getGiftFrom`, right?

And #3967 is correct because civ1 in `getClosestCities` is me, and it takes city1 -> to fix the issue we would simply take city2 as source and gate through a filter allowing unique differences but not lack tech - would need some changes in that rejection reason function or even more code duplication, so I left it out. We should get the city states their unique units as told by the wiki first.......

Anyway, that was the one I didn't test, of course - found no save with militaristic CS alive. So, changing my decision to present stuff in the order of the message -> push imminent (message says who what near, I think jumps may be better what who near). I'll recheck my saves for an example...

SomeTroglodyte:
Last commit fixes a :facepalm:  facepalm-level oversight, and it now shows what I intended: unit, CS, receiving city. Noticed through actual testing (which is why it took a while + some savegame edits)...

PR #3994, opened 2021-05-25
Declare & Revoke protection for city-state.

ninjatao:


yairm210:
Looks good in general!
All UI texts need to be added to template.properties

yairm210:
There should be a minimum number of turns

ninjatao:
> Looks good in general!
> All UI texts need to be added to template.properties

Busy weekend. Will do on Manday.

ninjatao:
Hope I got the right file. LOL

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -114,6 +115,15 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
             diplomacyTable.add(allyString.toLabel()).row()
         }
 
+        val protectors = otherCiv.getProtectorCivs()
+        if (protectors.size > 0) {
+            var protectorString = "{Protected by}: ".tr()
+            for (protectorName in protectors.map{it.civName}) {
+                protectorString = protectorString + "{$protectorName} ".tr()
```
yairm210, line 122: This would look like: "Protected by: America Germany Japan"
Maybe add a "," so it looks like "Protected by: America, Germany, Japan"
Something like 
```
var protectorString = "{Protected by}: "+ protectors.map { it.civName }.joinToString(", ")
```
You don't even need to .tr() anything since the .toLabel() below does that :)

ninjatao: Done. Thanks for the suggestion O_-

ninjatao:
Done. Thanks for the suggestion O_-

PR #4008, opened 2021-05-27
Enable BaseUnit.attackSound

SomeTroglodyte:
Actually, inspired by that 3998 issue, this hack just works. Really. My aliens finally fire 'phasers'. :smile:

The ugly hack is that a var property of an enum entry is modified - thus essentially abused as global variable with all the horrors that may bring (comod). I could also envision a variant where the same syntax UncivSound.getCustom() actually produces independent instances - no enum and all previous enum symbols now a kind factory method / lazy val in the companion object / something like that.

Interested? Pursue and polish or ditch and forget?

Oh btw the credits.md does not mention where the current sound mp3's come from, or am I blind? For new sounds - `select * from soundbible.com where license in ('Public Domain','Attribution 3.0')`? :grin:

### DIFF
--- a/core/src/com/unciv/models/UncivSound.kt
+++ b/core/src/com/unciv/models/UncivSound.kt
@@ -1,6 +1,9 @@
 package com.unciv.models
 
-enum class UncivSound(val value: String) {
+enum class UncivSound(
+    val value: String,
+    var custom: String? = null
+) {
     Click("click"),
     Fortify("fortify"),
     Promote("promote"),
@@ -12,5 +15,14 @@ enum class UncivSound(val value: String) {
     Policy("policy"),
     Paper("paper"),
     Whoosh("whoosh"),
-    Silent("")
+    Silent(""),
+    Custom("");
+    companion object {
+        fun getCustom(name: String?): UncivSound {
+            if (name == null || name.isEmpty()) return Click
+            val item = Custom
+            item.custom = name
+            return item
+        }
+    }
 }
--- a/core/src/com/unciv/ui/utils/Sounds.kt
+++ b/core/src/com/unciv/ui/utils/Sounds.kt
@@ -2,22 +2,40 @@ package com.unciv.ui.utils
 
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.audio.Sound
+import com.badlogic.gdx.files.FileHandle
 import com.unciv.UncivGame
 import com.unciv.models.UncivSound
 
 object Sounds {
-    private val soundMap = HashMap<UncivSound, Sound>()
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/models/UncivSound.kt
```diff
@@ -12,5 +15,14 @@ enum class UncivSound(val value: String) {
     Policy("policy"),
     Paper("paper"),
     Whoosh("whoosh"),
-    Silent("")
+    Silent(""),
+    Custom("");
+    companion object {
+        fun getCustom(name: String?): UncivSound {
+            if (name == null || name.isEmpty()) return Click
```
yairm210, line 22: name.isNullOrEmpty() ? 
I mean, I think they're exactly the same, but since there's a prebuilt function for this...

core/src/com/unciv/models/UncivSound.kt
```diff
@@ -12,5 +15,14 @@ enum class UncivSound(val value: String) {
...
+            val item = Custom
```
yairm210, line 23: I'm like 80% sure that this means every getCustom will return THE SAME INSTANCE of Custom, but just set one of its parameters differently. This means that if you have multiple custom sounds, the last initialized one overrides the rest. This seems an unsafe way of doing things,

core/src/com/unciv/ui/utils/Sounds.kt
```diff
@@ -2,22 +2,40 @@ package com.unciv.ui.utils
...
 
-    fun get(sound: UncivSound): Sound {
-        if (!soundMap.containsKey(sound)) {
-            soundMap[sound] = Gdx.audio.newSound(Gdx.files.internal("sounds/${sound.value}.mp3"))
+    fun get(sound: UncivSound): Sound? {
+        val fileName = if (sound == UncivSound.Custom) sound.custom!! else sound.value
+        if (!soundMap.containsKey(fileName)) {
+            val mods = mutableSetOf<String>()
+            if (UncivGame.Current.isInitialized) {
+                mods.addAll(UncivGame.Current.settings.visualMods)
```
yairm210, line 17: All things considered, it seems to me that A. the soundMap change is correct, but B. we should make a get(soundName:String) function, and have get(sound:UncivSound) call it. When we want a custom sound, we call get(soundName:String).

core/src/com/unciv/models/UncivSound.kt
SomeTroglodyte: Make that 100% - I said so it's essentially a single global var...

SomeTroglodyte:
As I said - this was just a _very_ uncouth demo that it's not _that_ far off. I'll see what I can come up with... Better make that a fresh branch anyway.

PR #4011, opened 2021-05-28
Spruced up Civilopedia - phase 3 - Interface, flavour text, new Tutorial

SomeTroglodyte:
I debated whether to go for internal linking next while keeping the next phase small and decided against it. This is more preparation with little visual effect, but it gets stuff out of the way so that the rest of the project has just 2 kinds of change - no, three:

- The rest of the visual candy capabilities of `FormattedLine`
- Attaching the interface to the rest of the ruleset objects - this alone will enable json-provided text, formattable/linkable, which will then be mixed with the old, unchanged content
- Using the interface's capabilities to provide a replacement for the old descriptions that can use formatting/linking - this was designed so one could do that in very small steps, but as I have everything done and tested I'd rather push that in one big block - your call.

What this _does_ visibly:
- A help page for the world screen done as new tutorial, under the hood pulling in four of the remaining `FormattedLine` attributes.
- Flavour text for a few Improvements - check out Roads, Barbarian camp - you see the json diff. Minimal demo, under the hood the interface is attached to TileImprovement but it hasn't gotten its overrides yet.

Notes:
- CivilopediaScreen received all its remaining changes - it looks for the interface in all source objects (but of course finds it only on two of the types - actually the compiler already knows that, it's not a runtime test - probably), and has the container to display the new stuff - might get simpler in the future, but for now a lot of the lines exist to make mixing old and new pretty enough, and I haven't yet had the inspiration on how to convert that fat block of `categoryToEntries` initialization to become a loop over `CivilopediaCategories.values()`.
- `xcf` is the Gimp format - provided in case someone wants to change that graphic.
- ~~I'm really not happy about the miniparser in that otherwise almost pointless `SimpleCivilopediaText` helper - but the alternative would have been to rewrite quite some Tutorials stuff and make the Tutorials json incompatible??? Can mods add to tutorials??? Or could/should we build a class that understands both formats? I'm sure you'll have some ideas.~~


SomeTroglodyte:
Forgot to mention: Those WorldScreen help lines _will_ create problematic automatic translation lines.
`#The World Screen` will be added to templates but actually `The World Screen` should get the translations, and `✯WorldScreenHelp` will be auto-added as such but should be ignored...

Maybe rewriting tutorials to be Collection<FormattedLine> instead of Array<String> _would_ be better... If I had an idea to make the new format optional - but I haven't. Array<String> seems to offer no tricks to get json-level alternatives in... Simply wrap each one, container "v1" contains old json as-is, container "v2" gets civilopediaText, both optional per tutorial entry???

SomeTroglodyte:
Just discovered TutorialRender.showDialog ... 'ExtraImages'. Completely unused, but I suppose that opens another option not requiring translator trouble or a miniparser... Wouldn't look the same without a large-font header, but close enough.

Thing is I'm still trying to find a clean solution to a fundamental problem with KeypressDispatcher and Popup - Gdx keeps listeners in an Array and fires an event in sequential order, that is, it's not the firing order you come to expect from other frameworks. A 'library'-style approach coded from the dispatcher itself fails as screens are sometimes instantiated without being used right away - so enforcing a last-come-first-serve policy is quite tricky.

So - please postpone for now. I will either patch soon or close and replace.

SomeTroglodyte:
Even if I found - of course - `<facepalm/>` that line in `TutorialRender` isn't used in the Civilopedia, doing something similar wasn't too complicated, and I do think this improves matters. The new page now looks like this:
image
... hope you like the idea, and the choice of words in the text. Should avoid translation issues completely, and leaves us freedom to rebuild tutorials on top of `FormattedLine`s in the future - or not - at our leisure.

yairm210:
Sorry for not getting around to this... I can't find a time when I have both the time and headspace to go over this thoroughly...

SomeTroglodyte:
Hey! No worries! I hope that means you assign the correct priorities (RL waaaaaaay over Unciv) :wink:  I know I do _not_ :sob:  ...

android/assets/jsons/Civ V - Vanilla/TileImprovements.json
```diff
@@ -101,14 +101,20 @@
 		"turnsToBuild": 4,
 		"techRequired": "The Wheel",
 		"uniques": ["Can be built outside your borders", "Costs [1] gold per turn when in your territory"],
-		"shortcutKey": "R"
+		"shortcutKey": "R",
+        "civilopediaText": [
+            {text:"Reduces movement cost to ½ if the other tile also has a Road or Railroad"},
+            {text:"Reduces movement cost to ⅓ with Machinery",link:"Technology/Machinery"},
```
yairm210, line 107: These are dependent on tech uniques, and so shouldn't be part of the civilopedia text, methinks

android/assets/jsons/Civ V - Vanilla/TileImprovements.json
```diff
@@ -186,7 +193,8 @@
 	},
 	{
 		"name": "Citadel",
-		"uniques": ["Gives a defensive bonus of [100]%", "Deal 30 damage to adjacent enemy units", "Great Improvement", "Can be built just outside your borders"]
+		"uniques": ["Gives a defensive bonus of [100]%", "Deal 30 damage to adjacent enemy units", "Great Improvement", "Can be built just outside your borders"],
+        "civilopediaText": [{text:"Constructing it will take over the tiles around it and assign them to your closest city"}]
```
yairm210, line 197: Same - this should be a unique

yairm210:
Okay, this seems pretty consistent. And 'civilopedia text' that tells me about a mechanical function means that there's a hidden unique here that's currently hardcoded

SomeTroglodyte:
Precisely. Some of those c'pedia entries were completely empty not long ago - that's one of the advantages of the civilopediaText json - you can use it for flowery romance novels or to explain a mechanic that's not obvious until we get around to cleaning up such things... These here were just samples, in no way did I set out to make every 'pedia page consistent and pretty - that's for much later.

SomeTroglodyte:
Over a month - needs redoing.

PR #4013, opened 2021-05-29
Sound upgrade - enable BaseUnit.attackSound, a few extras

SomeTroglodyte:
Well here's the cleaner version of #4008...

Architecture:
- UncivSound as Class instead of Enum, instantiation only allowed through factories in the companion object, with the overrides needed for use as HashMap keys. The changed file alone - without the rest, specifically with an unchanged Sounds.kt - already works (tested). The standard sounds exist as single instance reused throughout, more or less same as before. Custom sounds are fetched as individual instances from a factory method.
- Sounds.kt would probably also work standalone with the old Enum (untested - well they're tested together of course) and manages moddable sounds, while for safety also allowing to silently ignore missing files (maybe a console message?)

Extras:
- Couldn't resist - Having the Click sound also serve for two of the sliders on the Options screen has already annoyed me before, so I made a new one.. -6dB softer than the others. Also inexplicably makes it easier to actually slide them without being interrupted, though I think the interruption is due to losing focus to the ScrollPane...
- Cities as attacker needed their own sound! I hope this fits all eras.
- And a nuclear missile without a sound was unthinkable, too... Sorry... :rocket: :bomb: :boom:


yairm210:
This looks good!
The only part I have a problem with is the nuke mp3, which is 500kb! Could easily be a tenth of that!
Not a reason to deny this pr, but I do hope you'll find (or compile) a smaller file

SomeTroglodyte:
OK, I didn't really pay attention to size there - exported from Audacity with mp3 144 CBR. VBR and AAC not allowed due to compatibility with old hardware - or?

SomeTroglodyte:
- Existing sounds are already VBR, at bitrates around 80kbps, min 58 max 140. So I should redo slider and bombard, too. Thing with 'nuke' is - it's 40 seconds and IMHO needs to be. Sirens, suspense, and the moment the listener doubts anything is going to happen at all, boom. And the aftermath is needed, too... At absolute minimum, mono 44kbps it's 180k and sounds flat and tinny with a loud high-pitched whine due to compression harmotic distortion. Stereo preserves the eerieness, and at 86kbps the whine is much better... makes 352k.

Question: Pack a horrible/boring smaller sound (maybe even without the sirens) into main distribution and put up the better sounding one as mod? Or go for the 352k?

SomeTroglodyte:
P.S.: You could check my work files/variants [on my scratchpad](https://github.com/SomeTroglodyte/Scratchpad/tree/master/Media) if you wish.

PR #4019, opened 2021-05-30
Smaller mp3's for the new files from #4013

SomeTroglodyte:
This would be the 'not-as-hires-mod' variant. I can't get nuke smaller without horrible harmonic distortion _and_ loss of eerieness. The alternative - go for a 44k built-in (just an excerpt of the explosion itself at 40kbps mono) for main distribution and offer the nice stuff as mod (which would then include the other sounds at 144kbps quality too)  - we could still decide for later.


SomeTroglodyte:
Oh, the 'as mod' idea - that would raise the question:
- Replacing built-in sounds from mods _is_ handled by 4013, but only if you have that mod explicitly tied to your GameInfo.
- Having it work for already 'running' games - same as 'Permanent visual mods' - would require 1 extra line of code or so, using the same switch.
- But - in that case the 'visual' mod label would be slightly off and we would need a name encompassing all media... 'audiovisual'? clunky.

PR #4024, opened 2021-05-30
More power to improvement uniques (resourceTerrainAllow-be-gone)

SomeTroglodyte:
#### Motivation
My Aliens mod... I noticed the current code cannot limit an improvement to a base terrain irrespective of its features. Either my improvement goes on snow w/o hills only, or it goes on Snow + all 'climates' of hills. So - allow all in `terrainsCanBeBuiltOn` then cut down using uniques - their capabilities were already close, but not quite.

As always I went overboard and made it flexible enough to replace two existing tools:
-  `resourceTerrainAllow` - I apologize profusely for inventing the horrible name. I hope I picked a better one this time - now the circumstance that it's only needed when terrainsCanBeBuiltOn does not already allow the improvement needs to go into the documentation, not code. json change for this one included, as is a tentative deprecation annotation.
- `"Can only be built on Coastal tiles"` - "Can only be built on [Coastal] tiles" would also work, as would a modder idea like "Cannot be built on [Coastal] tiles". Didn't change json with this.

#### Problems to resolve
- Translateability of the `any` and `all` variants of tileFilter
- Updating your [wiki docs](https://github.com/yairm210/Unciv/wiki/Uniques)
- TileImprovementConstructionTests not inspected for consequences

I'll compile a suggestion for wiki text snippets and check for existing mods including resourceTerrainAllow shortly and update this accordingly.


yairm210:
You need to remember that there are existing mods using these features, that will break if you change the code from only A to only B. The way we do this is we allow both A and B, mark A as deprecated, and in a month or so notify modders in the Discord that A is being deprecated and delete it.

SomeTroglodyte:
> allow both A and B

I think I did? But I was too optimistic, the version needs a bump:
```kotlin
    @Deprecated("As of 3.14.13")
    var resourceTerrainAllow: Collection<String> = ArrayList()
```

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -376,22 +379,35 @@ open class TileInfo {
...
 
+    /**
+     * Implementation of _`tileFilter`_ 
+     * @see <a href="https://github.com/yairm210/Unciv/wiki/uniques#user-content-tilefilter">tileFilter</a>
+     */
     fun matchesUniqueFilter(filter: String, civInfo: CivilizationInfo? = null): Boolean {
-        return filter == baseTerrain
+        return filter == "All"
+                || '+' in filter && filter.split('+').all { matchesUniqueFilter(it, civInfo) }
+                || ',' in filter && filter.split(',').any { matchesUniqueFilter(it, civInfo) }
```
yairm210, line 400: Such uniques won't be translated

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -68,5 +69,24 @@ class TileImprovement : NamedStats() {
...
+     * Background: This not used for e.g. a lumbermill - it derives the right to be placed on forest
+     * from [terrainsCanBeBuiltOn]. Other improvements may be candidates without fulfilling the
+     * [terrainsCanBeBuiltOn] check - e.g. they are listed by a resource as 'their' improvement.
+     * I such cases, the 'unbuildable' property of the Terrain feature might prevent the improvement,
+     * so this check is done in conjunction - for the user, success means he does not need to remove
+     * a terrain feature, thus the unique name.
+     */
+    fun isAllowedOnFeature(name: String): Boolean {
+        if (name in resourceTerrainAllow) return true
+        return uniqueObjects.filter { it.placeholderText == "Does not need removal of []"
```
yairm210, line 87: Instead of .fliter{}.any() you can just .any{}

yairm210:
> > allow both A and B
> 
> I think I did? But I was too optimistic, the version needs a bump:
> 
> ```kotlin
>     @Deprecated("As of 3.14.13")
>     var resourceTerrainAllow: Collection<String> = ArrayList()
> ```

My bad, I was looking at the "cannot be built on" and saw that you were using the unique directly

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: I mentioned that in "Problems to resolve" - That was asking for ideas: teach the translator or find another way to express such filters?

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
SomeTroglodyte: Thanks - shows I didn't yet learn everything about the uniques system...

SomeTroglodyte:
> "cannot be built on" and saw that you were using the unique directly

I'm fiddling in another branch, but I think for that one too I left both old and new open, too but there wasn't a clear place to put a deprecation - there's none to deprecate specific uniques, is there?

But it seems you're not opposed to the general idea, so there's progress. I just started from the point "I want to express this is allowed on snow and snow+hill but not on other hills" - and the PR is successful in this, but not necessarily the best way to go about it. Open to ideas.

yairm210:
The 'cannot be built on' is good, AND the implementation is good! No fault with that at all

yairm210:
I don't want to include Boolean logic to the tile matches unique, but that's a different story

SomeTroglodyte:
> I don't want to include Boolean logic to the tile matches unique, but that's a different story

Exactly - just placing several instances of the same unique is an option, but then somehow you do not feel quite sure whether that will be a logical _and_ or an _or_ - and the '+' notation feels natural as we're seeing it for the layering while ',' feels natural anyway... And easily done... And then I had no really elegant idea on how to handle i10l for that. Go patch tr() '[]' handler _and_ the template generator after all? Not too 'tasty' either (sorry for the obscure lit ref nobody is likely to get - but you'll get the sound of it).

SomeTroglodyte:
In the meantime I found 2 relevant factors:
- `TranslationFileWriter.generateTranslationFiles` at line 95 - the text within [] is ignored. Therefore the auto-generation of templates relies on all placeholder components to be covered elsewhere as standalone strings - or to not need translation at all. That would also cover my new approach.
- Creation of comma-joined pre-translated lists that are then fed into placeholders where the `tr` will a bit later try to re-translate the list (unsuccessfully) is all over the place - e.g. `Building.getUniquesStrings()`. So, CSV as specialized use is already there, while I tried to introduce one filter to  support CSV in a generic way for all uniques using the filter. Sure, this discrepancy is ugly, but it _can_ coexist.
- As an aside: `DesktopLauncher.tryActivateDiscord` prevents debugging translations. Needs thinking as well - why is this even active for singleplayer games...

Please take a short look at the recent push - that works fine locally (and includes the "gerfinishen" lang from testing)...
Then we would go with max (as pushed) or min (I'll just remove '+' and ',' specials so my mod would have three 'Cannot' uniques which should work, too).

Note: Most if not all of these existing CSV uses are in the description generators, and in my rewrites of these I've completely ditched the CSV aspect - I wanted each entry to have link capability so the comma went out the door anyway, and the markup-generating description functions would also not pre-translate but leave translation of the `[]` parameters to the final `tr()`...

SomeTroglodyte:
`Coastal = ends without a space at the end`
I edit these in an external editor to preserve those and Studio _still_ bungles them while pushing... That's insubordination :laughing: ...

yairm210:
Please **remove** the boolean logic stuff from this PR. 
There are uniques that contain "+" and "," symbols already.
Let's commit what works - namely the new unique as a replacement for the olf parameter - and deal with the other stuff in a different PR.

SomeTroglodyte:
The latest push removes the '+' and ',' options and reverts the additions to `tr()`. I'm tempted to squash those 4 commits into 1 - but I only know how to safely do that within a local branch not tracking an online PR. Just say the word if you'd prefer the shorter history, too.

PR #4025, opened 2021-05-30
Added the paratrooper unit

xlenstra:
This adds the paratrooper unit with the ability to paradrop.
Currently it is implemented that the 'paradrop' action switches between normal movement and paradrop movement.
In accordance with the base game, the paratrooper can only paratroop when in friendly territory, and then only up to 5 visible tiles away. This should be increased to 9 (?) for at least one of the expansions


xlenstra:
I have no idea why it won't build, could someone else look at it?

SomeTroglodyte:
Well the fail message says "This test will only pass when there is a translation for all unit actions" - but I assume you know where to click to read that... So from the diff alone I'd say what the test sounds like is testing seems like it's there... If this were mine, I'd backup the branch locally, Drop the PR, delete the online branch, fix your local branch until it runs as you intend it to (might do right away, if not diff with your backup...)... Squash commits into one and read its diff closely, there should be nothing you don't understand... (*) and _then_ refresh your master and rebase the working branch onto it. Then re-push...

(*) at this point you could debug the tests? I vaguely remember you could indeed set a breakpoint in there and run - not 'debug desktop' or Shift-F9 but something else - right click on 'tests' -> debug???...  ...that would actually hit a breakpoint in the test code. Then look what it actually is that fails.

SomeTroglodyte:
P.S.: The diff looks also like you permanently removed the 'build road' action - true, the worker doesn't use this - but the roman legionary?

SomeTroglodyte:
Well, the check that ran the tests explicitly says "Paradrop" is the unit action missing a translation... Debugging tests isn't that hard, but I just found there's configuration missing in a fresh fork that the tests need to find our assets, so forget that... And the "Paradrop = " line is there and has the necessary blanks, so... Are those blanks by chance not ASCII 0x20 but some obscure Unicode blank? Or github confused stuff in which case my recommendation to reassemble the branch locally would help. Remember when you have deleted the online copy and git fetch has finally noticed that - _then_ you can re-push, and on the push (or just before from the git window) double-check the diff if it meets expectations. That's why I would squash so it's easier to get just one diff.

SomeTroglodyte:
Ha! The test doesn't look for `Paradrop = ` in `template.properties` _or_ `English.properties`, only the others. So put a `Paradrop = neerstorting` in `Dutch.properties` and it passes. Probably.

SomeTroglodyte:
Oh and the secret to debugging tests is a) Just open test sources and right-click -> debug the line with the test function's declaration and b) set the assets folder in its configuration manually...

xlenstra:
> P.S.: The diff looks also like you permanently removed the 'build road' action - true, the worker doesn't use this - but the roman legionary?

It is still there. I reordered them, so it is now together with other unit-specific actions such as 'set up' and 'paradrop' actions.


> Ha! The test doesn't look for `Paradrop = ` in `template.properties` _or_ `English.properties`, only the others. So put a `Paradrop = parachutelanding` in `Dutch.properties` and it passes. Probably.

Thanks! This fixed it

SomeTroglodyte:
> It is still there. I reordered them,...

No worries - that was from a superficial look.

yairm210:
I don't see anything limiting you from paradropping into the sea, or maybe even impassable tiles. I see that those types aren't coloured, but I don't think that would stop the actual action.
What happens if you try?

SomeTroglodyte:
Well, with xander's branch you can paradrop into water (and continue moving) or across an isthmus, but not into a mountain. You can also paradrop onto unobserved tiles (can't test unexplored right now) - and there's a hint in the wiki for the xcom unit it can't so perhaps its predecessor shouldn't as well? Also, I can't see it listed _explicitly_ in the wiki dropping into water isn't allowed.

xlenstra:
> I don't see anything limiting you from paradropping into the sea, or maybe even impassable tiles. I see that those types aren't coloured, but I don't think that would stop the actual action.
> What happens if you try?

Good point! I somehow forgot to do this.
It should be implemented now

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -343,7 +343,14 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
         }
         return false
     }
-
+    
+    // Can a paratrooper land at this tile?
+    fun canParadropOn(destination: TileInfo): Boolean {
+        // Can only move to land tiles within range that are visible and not impassible
+        // Based on some testing done in the base game
+        if (!destination.isLand || destination.isImpassible() || !unit.civInfo.viewableTiles.contains(destination)) return false
```
SomeTroglodyte, line 351: `return destination.isLand && !destination.isImpassible() && unit.civInfo.viewableTiles.contains(destination)` seems more readable to me :eyes:  - shouldn't the automatic inspections whine about such things?

core/src/com/unciv/ui/worldscreen/unit/UnitActionsTable.kt
```diff
@@ -41,6 +41,7 @@ class UnitActionsTable(val worldScreen: WorldScreen) : Table() {
                 "Move unit" -> return UnitIconAndKey(ImageGetter.getStatIcon("Movement"))
                 "Stop movement" -> return UnitIconAndKey(ImageGetter.getStatIcon("Movement").apply { color = Color.RED }, '.')
                 "Promote" -> return UnitIconAndKey(ImageGetter.getImage("OtherIcons/Star").apply { color = Color.GOLD }, 'o')
+                "Upgrade" -> return UnitIconAndKey(ImageGetter.getImage("OtherIcons/Star").apply { color = Color.GOLD })
```
yairm210, line 44: You will never have an Upgrade action, always an Upgrade to, which is caught on line 26

yairm210:
Looks good!
Just remove the upgrade action and we'll merge :)

xlenstra:
Fixed!

SomeTroglodyte:
I like it, and thanks XL for teaching me something about the unit tests. I hope you don't mind I credited you on my [notes](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Unciv-notes/Working%20in%20the%20IDE%20and%20Contributing.md#run-unciv-unit-tests) for the new chapter on them.

ravignir:
If i can add something; 
Paratroopers should have 65 Strength, as 40 is vanilla civ5 value (and unciv unit strengths are all from g&k):
[wikia](https://civilization.fandom.com/wiki/Paratrooper_(Civ5))
Also they should not be able to attack after paradroping, unless they have blitz promotion.

PR #4028, opened 2021-05-31
Patch ModManager exit to allow deactivating a selected tileset

### LINKED ISSUE #4016: Slight problems with 5Hex as permanent visual mod
I believe these belong here even though I repro with this specific mod - or is there any other supplying a new tileset?

a) Glitch after deactivating from main screen
- Make sure 5Hex is installed and set as PVM and set as tileset in Options
- From main (not world) screen go to mod manager and deactivate PVM for 5Hex, close MM
I get a white background map with a few resources.
image
(probably just a missing reset to FantasyHex - simple restart fixes)

b) Map editor using 5Hex widget sizing
- Make sure 5Hex is installed and set as PVM and set as tileset in Options
- Enter map editor
Terrain and features are OK, resources improvements are squashed - still work though.
image
(likely scaling the terrain to the fixed widget size forces the same scaling to be applied to other images but they didn't grow due to 5Hex - something like that)

I'll probably solve these myself in time, but any help is welcome...

SomeTroglodyte:
... and make TileInfo debug display a little nicer.

Resolves #4016 part (a) - though I'm not sure the extra code is worth fixing such a borderline non-crashing situation. You throw the dice.

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -450,18 +450,17 @@ open class TileInfo {
     /** Shows important properties of this tile for debugging _only_, it helps to see what you're doing */
     override fun toString(): String {
         val lineList = arrayListOf("TileInfo @$position")
-        if (this::baseTerrain.isInitialized) {
-            if (isCityCenter()) lineList += getCity()!!.name
-            lineList += baseTerrain
-            for (terrainFeature in terrainFeatures) lineList += terrainFeature
-            if (resource != null) lineList += resource!!
-            if (naturalWonder != null) lineList += naturalWonder!!
-            if (roadStatus !== RoadStatus.None && !isCityCenter()) lineList += roadStatus.name
-            if (improvement != null) lineList += improvement!!
-            if (civilianUnit != null) lineList += civilianUnit!!.name + " - " + civilianUnit!!.civInfo.civName
-            if (militaryUnit != null) lineList += militaryUnit!!.name + " - " + militaryUnit!!.civInfo.civName
-            if (this::baseTerrainObject.isInitialized && isImpassible()) lineList += Constants.impassable
-        } else lineList += "uninitialized"
+        if (!this::baseTerrain.isInitialized) return lineList[0] + ", uninitialized"
+        if (isCityCenter()) lineList += getCity()!!.name
+        lineList += baseTerrain
+        for (terrainFeature in terrainFeatures) lineList += terrainFeature
+        if (resource != null) lineList += resource!!
+        if (naturalWonder != null) lineList += naturalWonder!!
+        if (roadStatus !== RoadStatus.None && !isCityCenter()) lineList += roadStatus.name
+        if (improvement != null) lineList += improvement!!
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -449,17 +449,19 @@ open class TileInfo {
...
+            if (isCityCenter()) lineList += getCity()!!.name
+            lineList += baseTerrain
+            for (terrainFeature in terrainFeatures) lineList += terrainFeature
+            if (resource != null) lineList += resource!!
+            if (naturalWonder != null) lineList += naturalWonder!!
+            if (roadStatus !== RoadStatus.None && !isCityCenter()) lineList += roadStatus.name
+            if (improvement != null) lineList += improvement!!
+            if (civilianUnit != null) lineList += civilianUnit!!.name + " - " + civilianUnit!!.civInfo.civName
+            if (militaryUnit != null) lineList += militaryUnit!!.name + " - " + militaryUnit!!.civInfo.civName
+            if (this::baseTerrainObject.isInitialized && isImpassible()) lineList += Constants.impassable
```
yairm210, line 463: Can remove this check

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -449,17 +449,19 @@ open class TileInfo {
...
-        for (terrainFeature in terrainFeatures) lineList += terrainFeature
-        if (resource != null ) lineList += resource!!
-        if (naturalWonder != null) lineList += naturalWonder!!
-        if (roadStatus !== RoadStatus.None && !isCityCenter()) lineList += roadStatus.name
-        if (improvement != null) lineList += improvement!!
-        if (civilianUnit != null) lineList += civilianUnit!!.name + " - " + civilianUnit!!.civInfo.civName
-        if (militaryUnit != null) lineList += militaryUnit!!.name + " - " + militaryUnit!!.civInfo.civName
-        if (isImpassible()) lineList += Constants.impassable
+        val lineList = arrayListOf("TileInfo @$position")
+        if (this::baseTerrain.isInitialized) {
```
yairm210, line 453: Just return here if uninitialised, to exit early and minimize indentation
(I'm a big believer in 'exit early' if you hadn't noticed)

SomeTroglodyte: True, and of course I did consider it and weighed having more indentation against an extra `joinToString` call (or `lineList[0] + " not yet ripe"`) the compiler might optimize away or may not... I draw the line at 4-5 levels of indentation in python or visual basic (yes, even that - in SSIS transforms) and maybe one more in curly braces sporting languages...  :partying_face:  - Thinking about compiler optimizations in code that only ever gets run by the debugger is a good one.

image
.. seems to work. Except it wants a merge... OK, time for experiments (aka 'does chromium have profiles'?).

SomeTroglodyte:
No, a fresh empty no-extensions profile doesn't make the online conflict manager work for me. Still, since my bigger modmanager thingy is the cause it's better to merge locally, anyway.

PR #4038, opened 2021-06-02
Add Downloads section & Add info about Flatpak

TuxKruger:
I thought it would be better if there's a Downloads section and since there's a Flatpak for this game, it deserves to be mentioned in this repo's README.md


TuxKruger:
> Please put the download badges at the top of the page, since usually the first thing a potential user wants to know it 'how do I get this' :)

Done!

PR #4039, opened 2021-06-02
Add global alerts for certain constructions

avdstaaij:
In the original game, certain constructions alert all players upon their completion (see for example https://civilization.fandom.com/wiki/Apollo_Program_(Civ5)). This PR introduces a new unique for this behavior and adds this unique to the relevant constructions. The global alert is implemented as a notification.

To my knowledge, the relevant constructions are Manhattan Project, Apollo Project and all space parts. If there are any more, please let me know!

I was unsure whether to place the code that implements the unique in Building.postBuildEvent or in CityConstructions.constructionComplete. I choose the former, since the unique is currently only used by buildings. If other kinds of constructions require a global alert as well, the code could be moved to CityConstructions.constructionComplete.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -41,6 +41,7 @@ Consumes 1 [resource] =
 Consumes [amount] [resource] = 
 Required tech: [requiredTech] = 
 Requires [PolicyOrNationalWonder] = 	
+Triggers a global alert upon completion = 
 Cannot be purchased = 
 
 Current construction = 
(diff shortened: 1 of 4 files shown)


android/assets/jsons/translations/template.properties
```diff
@@ -41,6 +41,7 @@ Consumes 1 [resource] =
 Consumes [amount] [resource] = 
 Required tech: [requiredTech] = 
 Requires [PolicyOrNationalWonder] = 	
+Triggers a global alert upon completion =
```
yairm210, line 44: Since the unique is defined in one of the Json files, we don't need to define it in the template file - only non-json, UI elements

yairm210:
Looks good!
If there wasn't an existing wonder notification, this would be good to go
As it stands, though, both notifications should be in the same place - I think the wonder notification is in cityConstructiond, and you can move it here

avdstaaij:
I have applied the changes you requested!

PR #4040, opened 2021-06-02
Key dispatcher precedence with several Popups open

SomeTroglodyte:
The 'keyboard project' still has some way to go, but I pulled this forward.

Motivation: Popups stacked onto each other should be closed one by one with back/esc, and some skipped to the bottom-most non-Popup screen.

Key point I learned:
- Gdx listeners get called in the order they were defined - it's an Array deep down and it is enumerated 'normally'. This means the bottom-most/oldest widget/screen gets the events first and decides whether the stacked widgets get the right to be informed, too. I fully expected otherwise, so I had to think for a while how to handle this.

Dismissed ideas: Fork Gdx and reverse - ouch. KeyPressDispatcher tracking its instances and auto-disabling some - died when I found some popups get created in advance and I had to admit that even if I 'fixed' that one I can't guarantee the last-created instance should always be the only active one.

Pragmatic solution: Similar to hasOpenPopups() - but counting them. Rearranged some code so a popup can count itself. I left all changes to purely debugging help stuff in - markedly the move of KeyPressDispatcher's naming. I propose if we want we can remove that at the end of the keyboard 'project'?

Future roadmap:
- Unify the two Exit game prompts and base them on YesNoPopup.
- Main menu - while there I explored the option to show supported keys as Gdx ToolTip instead of the "(X)" way we did so far. Surprise - they do work and are nicely easy. Not so nice - two differences to what we're used elsewhere (long initial delay, after that all tips on screen react immediately, they show initially where the pointer is but don't follow it) actually make them feel klunky, plus touchscreens need test -> postponed as 'need more investigation'. Please comment?
- Map editor some obvious nice-to-have keys and fixing some translation issues I found while there.
- Scan _all_ Popup uses and adapt them - a few (text + 1-button ones) probably using a new subclass InfoPopup.

### DIFF
--- a/core/src/com/unciv/ui/utils/CameraStageBaseScreen.kt
+++ b/core/src/com/unciv/ui/utils/CameraStageBaseScreen.kt
@@ -28,7 +28,7 @@ open class CameraStageBaseScreen : Screen {
 
     protected val tutorialController by lazy { TutorialController(this) }
 
-    val keyPressDispatcher = KeyPressDispatcher()
+    val keyPressDispatcher = KeyPressDispatcher(this.javaClass.simpleName)
 
     init {
         val resolutions: List<Float> = game.settings.resolution.split("x").map { it.toInt().toFloat() }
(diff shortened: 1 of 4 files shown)


yairm210:
No gdx tooltip use if it feels worse as a user.

If you're finished with debug lines, please remove instead of uncommenting, especially if it's just text with no logic

Issue #4015, opened 2021-05-29
Domination Victory according to Civ V rules 


whatsupboy:
Unciv's domination victory requires you to wipe out every civ in the game or atleast for you to be the only one left with cities. In a game with alot of civs and a large map, that is very tedious. 

Vanilla Civ V and the expansions focus on the capitols instead. In vanilla you have to be the last one left with your original capitol. In the expansions you have to own every original capital. 

I think going with either of the two would give clearer ways to go about winning and be more fun.

And if Unciv's domination victory is the more preferred would having both Civ V's and Unciv's victory types in the game be an option?

ravignir:
I think all victory types need some overhaul:
[Spaceship](https://civilization.fandom.com/wiki/Spaceship_(Civ5)) parts should be units, than need to be transeferred to capital to conenct them to spaceship.
Completing 5 Policy branches should allow [Utopia Project](https://civilization.fandom.com/wiki/Utopia_Project_(Civ5))

jahodovykolac:
Maybe you should put these to #663.

logicminimal:
> I think all victory types need some overhaul:
> [Spaceship](https://civilization.fandom.com/wiki/Spaceship_(Civ5)) parts should be units, than need to be transeferred to capital to conenct them to spaceship.
> Completing 5 Policy branches should allow [Utopia Project](https://civilization.fandom.com/wiki/Utopia_Project_(Civ5))

agree

yairm210:
All according to #663

PR #4041, opened 2021-06-02
Updated domination and culture victory to more closely resemble the original game

xlenstra:
- Updated the domination victory so you only have to capture all capitals.
- Added the Utopia building, which is unlocked after 5 policy trees have been completed.

Some part of it is still depends on #4039 to implement global alerts for starting the building of the Utopia project, but I will implement behavior related to this global alert when that PR is fixed, possibly in another PR.
Partly solves #4015.


yairm210:
The main reason this change hadn't been implemented so far is the change to the AI that needs to be made to prevent them from launching and continuing 'useless' wars when aiming for domination, and focus on capital captuing and defending

xlenstra:
I see.
During a few test games I ran with only domination victory enabled (deity difficulty) and myself a spectator, the AI was almost never able to win anyway. So while I agree the AI should be changed to be able to win with this new victory condition, I don't feel like this change makes the AI much worse. If anything, it lowers the amount of work the AI's need to do which might make it more likely for them to win a domination victory.
I might look into the AI in the future, but I wasn't planning on doing it for this PR. If this means that it won't be merged, I'll close it and make a new PR for just the changes to the culture victory.
For the Culture Victory, later this week I'll make it so that the AI actually builds the Utopia Project when able, as well as fix the merge conflicts, so I'll mark this PR as draft for now.

yairm210:
The AI currently does pose a challenge to the player aiming for domination.
If all he needs to do is blitzkrieg towards the capital of each Civ, capture it and then make peace, it really becomes Way Too Easy (tm)

SomeTroglodyte:
Might be pretty easy to make that constant "5" moddable? (otherwise looks like thorough work :+1: )

xlenstra:
> The AI currently does pose a challenge to the player aiming for domination.
> If all he needs to do is blitzkrieg towards the capital of each Civ, capture it and then make peace, it really becomes Way Too Easy (tm)

This is a fair point. I'll close the PR later and open a new one for only the culture victory changes.

> Might be pretty easy to make that constant "5" moddable? (otherwise looks like thorough work +1 )

Currently, the 5 is already inside the entire [5 social policy branches have been completed], but I could rework it to be "Hidden until [5] [social policy branches have been completed]". I'll do that as well when I open the new PR.

Issue #3522, opened 2021-01-12
Crash at load on Mac OSX with OpenJDK11 after changing resolution


MRigal:
**Describe the bug**
It crashes when starting the game via `java -jar unciv.jar`. After one resizes the window and/or changes the resolution to something other (I tested only higher) than the default 900*600 It is only starting the smiley shortly and then crashing. "DesktopLauncher" is crashing too.

**To Reproduce**
Modify the resolution and close the game. Next time it will crash, I always need to delete the GameSettings.json (or remove the windowState key from the json file to get it started again

**Expected behavior**
I guess it should not happen

**Additional context**
I have version 3.12.7-patch1. Here is the output
```
Error in LSRegisterURL: -10811
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.lwjgl.LWJGLUtil$3 (file:/Users/foobar/Documents/Unciv.jar) to method java.lang.ClassLoader.findLibrary(java.lang.String)
WARNING: Please consider reporting this to the maintainers of org.lwjgl.LWJGLUtil$3
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
AL lib: (EE) ca_open_playback: AudioUnitInitialize failed
Loading ruleset - 235ms
Loading percent complete of languages - 1ms
2021-01-12 11:40:37.903 java[70531:1012115] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow drag regions should only be invalidated on the Main Thread!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff204fc6af __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007fff202343c9 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff20524a9a -[NSException raise] + 9
	3   AppKit                              0x00007fff22cc88a0 -[NSWindow(NSWindow_Theme) _postWindowNeedsToResetDragMarginsUnlessPostingDisabled] + 321
	4   AppKit                              0x00007fff22cf0c1b -[NSThemeFrame setStyleMask:] + 142
	5   AppKit                              0x00007fff22cf09b1 __25-[NSWindow setStyleMask:]_block_invoke + 2057
	6   AppKit                              0x00007fff22cf014e NSPerformVisuallyAtomicChange + 132
[...]

PR #4042, opened 2021-06-03
Fix the app window changing may lead to that the app can't open in macOS

lishaoxia1985:


yairm210:
Nice!!!!
Tested on MacOS, I assume?
Resolves https://github.com/yairm210/Unciv/issues/3522 ?

lishaoxia1985:
I think it runs very wellᥬ😂᭄

PR #4046, opened 2021-06-03
Re-show redundant tech prereq mod errors without impeding playability

SomeTroglodyte:
Combination of my #4035 and your [commit a70bb43d](https://github.com/yairm210/Unciv/commit/a70bb43daefa7e05aa94aa4f796dcda8b6d3c9f5). Screenshot over there still applies.
DeCiv and the likes are nagged about in both options-modcheck and newgame, but can be played.
Would make adding other non-fatal checks in the future easier.

Not really important anymore, but if you like...

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -227,9 +227,37 @@ class Ruleset {
         return stringList.joinToString()
     }
 
-
-    fun checkModLinks(): String {
+    /** Severity level of Mod RuleSet check */
+    enum class CheckModLinksStatus {OK, Warning, Error}
+    /** Result of a Mod RuleSet check */
+    // essentially a named Pair with a few shortcuts
+    class CheckModLinksResult(val status: CheckModLinksStatus, val message: String) {
+        // Empty constructor just makes the Complex Mod Check on the new game screen shorter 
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -227,9 +227,37 @@ class Ruleset {
...
+                    when {
+                        lines.isEmpty() -> CheckModLinksStatus.OK
+                        lines.size == warningCount -> CheckModLinksStatus.Warning
+                        else -> CheckModLinksStatus.Error
+                    },
+                    lines)
+        // Allows $this in format strings 
+        override fun toString() = message
+        // Readability shortcuts
+        val isError
```
yairm210, line 253: fun isError() = status == CheckModLinksStatus.Error

That way it's more obvious to the casual reader that this isn't a field saved in the class
In general I think that readonly param

SomeTroglodyte: Chuckle. Yea, properties without backing fields aren't to be preferred 'just because we can'. And here the compiled result is identical anyway. Is this mayhaps related to me running the 'Darcula' theme with its garish colours? Sorry.
Their value is marginal anyway, I just wanted to have the call site as short as feasible...

PR #4048, opened 2021-06-04
Unified exit game popups, removed obsolete closing handler

SomeTroglodyte:
Next step in the 'keyboard project'.
Only one exit game prompt - as YesNoPopup subclass. Side effect: arrangement of yes and no same as everywhere else - no, I think there's still maybe a popup asking yes and no but not using yesno out there... Those will be assimilated later.
Both analysis and testing shows the normal popup 'close' handling is sufficient for the tutorial logic - though I may have missed something? The comment was outdated anyway.

### DIFF
--- a/core/src/com/unciv/MainMenuScreen.kt
+++ b/core/src/com/unciv/MainMenuScreen.kt
@@ -14,7 +14,6 @@ import com.unciv.logic.map.MapSize
 import com.unciv.logic.map.MapSizeNew
 import com.unciv.logic.map.MapType
 import com.unciv.models.ruleset.RulesetCache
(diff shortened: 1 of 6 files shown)


yairm210:
Have you tested this with closing tutorials? A simple way is to rename your gamesettings file to GameSettings-Old, and then the game autogenerates a new one, and you 'start from scratch'.

SomeTroglodyte:
> tested this with closing tutorials

Of course, several times. I usually prefer editing the config in a notepad - that's open anyway so I can distinguish single-line from multi-line ones. But you're welcome do test yourself, in fact, I'd feel better if you did.

Issue #3937, opened 2021-05-14
Swap unit position and link units feature.


ArchDuque-Pancake:
I don't know if this is in civilization. But, it basically allows you to swap positions of the unit A, to the tile position of unit B.
I had thought this could be done in the game by thouching unit A, then unit B, this would make appear a little pop up, like the one that appears when Capturing a unit, giving you the option to swap their positions. When swaping positions, the movement tile cost will be applyed.

The other feature is to be able to link up a military unit, with a civilian unit. The process could be: You select the civilian unit, and then the military unit, then the pop up would appear, enabling you to link them.
When you have linked a unit, both units move to the same tile selected when moved. If, the military unit has more movement points than the civilian one, the unit will get unlinked, but the civil unit will move aproximatedly to the wished position, but still, lagging behind.
If the civilian unit has more movement points than the other unit, the civilian unit will not consume it all.

SomeTroglodyte:
The swap idea would be a rule change, the "escort" idea would not. Changing rules - nah, unless you can show that the Original offered it.

For the other idea - with units of identical speed you can get that anyway, so I'd first declare it to always match speed to the lower common denominator. Still, I would envision quite a lot of code, and more properties to be persisted in saves - not worth it IMHO.

ArchDuque-Pancake:
Yeah, I understand it would need a lot of work, but, I really think the swap feature is valuable, because when using a big army, or there is to little space, manoeuvering with units is very complicated and inefficient. I'll try to search for footage.

But I get that it may take too long, and if it doesn't get implemented, I'll understand.

ArchDuque-Pancake:
Here!!! https://youtu.be/MXezgEWjglo
Minute: 0:25

SomeTroglodyte:
There's a PR for swapping (#4049), and what you mean by linking is close to what multi-select might do when it's finished - and I'd say any other approach would be too complicated, so I think this could be closed?

yairm210:
Closed in #4049

PR #4049, opened 2021-06-04
Unit swapping

avdstaaij:
As proven in #3937, the original Civilization V has the feature to swap units.
If you try to move a unit to a tile with another of your units (of the same type), and both units have enough movement points left to reach the other's tile, they will swap places. They will consume only the movement points needed to reach the other's tile when doing so.

This change implements unit swapping for Unciv.

To prevent all kinds of problems from arising with automatic unit movement, unit swapping can only be done explicitly. This also means that it can only be done if the unit-swap movement is possible in a single turn (for both of the swapping units). Swapping is however not limited to adjacent tiles: such a limitation would prevent you from swapping two units while keeping a unit in between them in place.

Swapping is implemented for both civilian and military units. Each type can only swap with units of the same type.

The UI for unit swapping looks as follows:

All units now have an additional togglable "Swap Units" unit action.

If the unit action is toggled on, the unit enters unit-swap mode. In this mode, the units with which he could swap are highlighted.

Clicking on a highlighted tile creates a unit-swap button, similar to the movement button, or instantly performs the swap if single-tap-move is enabled.

If the button is clicked, the swap is performed.

Swapping with a unit in a city center is also possible, and looks as follows:

I chose to use a special unit-swapping mode because Unciv does not distinguish between unit-selection clicks and unit-movement clicks in general (right click is a thing, but only for desktop). Without a special mode, unit swap would have to be performed when you click on an eligible unit while you have another one selected, an action that would currently just select the other unit. I did not want to make it more difficult to select other units, and swapping units is a fairly specific action that you probably only want to do explicitly, so a separate mode seemed to be the best approach. An option to always swap-move when an eligible tile is clicked instead of requiring the unit-swapping mode, similar to the existing single-click-movement option, could perhaps be added later.

I realize that this PR touches quite a lot of parts of the codebase, but this is the minimum amount of changes required to properly implement the feature.

Many thanks to @xlenstra for helping to find various edge-cases and bugs! The code is now so robust, it can even handle swaps that involve a paradrop!

Of course, feel free to ask me about any specifics of how unit swap is implemented.


ravignir:
now we need zone of control. Also for those wondering why in the video linked in #3937  swordsman can't swap with composite and composite can with swordsman- composite was fortified.

yairm210:
Question: Why do we need the Swap units action? Why not just open, along with the regular 'move unit here' circle, a 'swap units' circle?
I'll tell you why I ask - the current implementation adds an extra button for every unit at all times, taking valuable screen space, even though this action is only relevant a very small percentage of the game.
If we can have this appear only when context-relevant, that's much better (and also less visual and therefore cognitive overload for the user who already has a gazillion things on screen)

avdstaaij:
> now we need [zone of control](https://civilization.fandom.com/wiki/Zone_of_control_(Civ5)).

Agreed. This would probably be best implemented as additional movement cost, for example in `getMovementCostBetweenAdjacentTiles`. Because the unit swap eligibility checking uses `getDistanceToTilesWithinTurn`, which returns the tiles a unit can reach in the current turn, the unit swapping system would automatically take such a change into account. In fact, an implementation of ZOC would probably not even have any merge conficts with this PR.

> Question: Why do we need the Swap units action? Why not just open, along with the regular 'move unit here' circle, a 'swap units' circle?
> I'll tell you why I ask - the current implementation adds an extra button for every unit at all times, taking valuable screen space, even though this action is only relevant a very small percentage of the game.
> If we can have this appear only when context-relevant, that's much better (and also less visual and therefore cognitive overload for the user who already has a gazillion things on screen)

If I understand your question correctly, this is what I explain in the large paragraph below the UI screenshots.

In general, Unciv does not distinguish between "unit selection"-clicks and "unit movement" clicks: on mobile devices, both are screen presses. Currently, if you have a unit selected and then click on another unit, you select that other unit. If there was no separate unit-swapping mode, then this behavior would change: clicking on another unit that is eligible for swapping would instead create the swap button. If you have a lot of units bunched up, like in a war, I expect that this would make it more difficult to select a different unit: you would have to deselect your unit first and then select a different one. It may also be confusing for players who are used to the current behavior. The original game does not have this problem, since it distinguishes selection clicks (left-clicks) from movement clicks (right-clicks).

I do agree that the new unit action clutters the list somewhat, but given the above, it seemed like the best approach to me. Like I explained, I can imagine adding an option later (similar to the single-click-movement option) to make the system behave like you suggest: with this option enabled, you would always swap-move when clicking on another unit that is eligible instead of selecting that unit. However, as I explained, I feel like this should not be the default. Such an option could be added in a followup PR.

Another change that would reduce the clutter is not showing the button if there is no possible unit to swap with. This is easy to implement, but I chose not to, because it would require doing the single-turn movement calculations every time a unit is merely selected, for every unit in its movement range. I imagined this might be a performance issue. However, if you want, I could definitely enable it.

xlenstra:
> If you have a lot of units bunched up, like in a war, I expect that this would make it more difficult to select a different unit: you would have to deselect your unit first and then select a different one.

This would be even worse with singe-tap movement enabled, as then you would have to manually deselect each unit every time you wanted to select another unit within swapping range. And even if you don't have it enabled, getting a viewblocking swap-button every time you tap another unit close by also takes up valuable screen space :)

Another option would be to have an one of the unit actions be a 'show more actions'-button, containing the more rarely used actions This is used in the original game to hide options like disband unit, automate, explore (?), fortify when damaged as fortify until healed is the more logical option (?), etc.). This would reduce the amount of actions on screen without reducing possibilities. If 'swap' is a very situational option, this could also be put in that menu.

SomeTroglodyte:
Isn't is somewhat established on 'droid that a long-press usually fills the role of right-clicks or close to it? This is not the first time I've asked myself why we do not use them. `override fun GestureListener.longPress` or something? Must try some time...

avdstaaij:
> Isn't is somewhat established on 'droid that a long-press usually fills the role of right-clicks or close to it? This is not the first time I've asked myself why we do not use them. `override fun GestureListener.longPress` or something? Must try some time...

This is indeed another option, but we would then have to somehow explain to the user how to swap units, since it would be entirely hidden. I am however not entirely sure if hiding unit swapping -- a feature that can provide a significant combat advantage -- behind a long press, which currently isn't used anywhere else, is a good idea. I also imagine that it could cause a delay when selecting units, since the game would have to wait to see if user is long-pressing. Still, this is an option.

avdstaaij:
I have thought of another possible UI, in between the unit swapping mode I have implemented and unit swapping whenever you click an eligible unit. It would work as follows:

If you have a unit selected and you click on a unit with which you could swap, the game shows both a "select this unit"-button and a "swap with this unit button", just like what already happens when you try to swap with a unit in city center, as shown in the last screenshot. This still has the disadvantage of requiring additional clicks to select another nearby unit, which may be very annoying during a war, but it is not as bad as requiring you to deselect the unit and it removes the need for a unit-swapping unit action and mode.

If this option is preferred, I would like to know what should happen when single-click movement is enabled. Should it cause the unit-swap move to always be chosen without showing the buttons, or should it be ignored and should the buttons always be shown?

I kind of regret suggesting this approach, because it would certainly be more work to implement than the others. :P

SomeTroglodyte:
> explain to the user
I hope we can do videos in the civilopedia soon... No, but images vertically interspersed with text is in a working stash. So we'd at least have a place to communicate this, _if_ we marginalize the feature that badly. I think I'll have to try your code to see for myself... Oh, and a longpress handler switching to a new option after the selection for the first touch has already done its thing isn't impossible either.

SomeTroglodyte:
With 'move units with a single tap' on, everything seems to work, with the option in its default state I can neither move nor swap any unit.

avdstaaij:
> With 'move units with a single tap' on, everything seems to work, with the option in its default state I can neither move nor swap any unit.

That's strange. I cannot reproduce this issue. Do you still see the circle overlays that show to which tiles you could move? Note that you cannot perform a normal move in unit-swapping mode, that is intentional. Perhaps you have a unit in swap mode but there are no units with which you could swap. In that case, you should still be able to toggle the swap mode off (and swap mode should be off whenever you select a new unit).

SomeTroglodyte:
No, the possible destinations are still marked upon first click, the second is the problem, where it should display the overlay move button. I can debug right down to `addOverlayOnTileGroup` and everything seems to be there, but later nothing gets displayed... Maybe I should reboot first. This isn't windoze requiring the 'bill gates car trouble method' all the time, but still... Anyway, got an appointment. Later. Forget it should you not hear from me.

avdstaaij:
I still cannot reproduce what you are describing, so do let me now later if rebooting helped!

SomeTroglodyte:
Of course. But reboot means update which means kernel 5.8.0.55 and chromium 91 - uh, oh, 91 breaks any corporate environment brutally, lets hope github will work... Nah, a pause is in order.

SomeTroglodyte:
Getting closer, still stumped (no, new kernel, new AS plugins or AS invalidate caches / clean project's didn't change anything).
- I start any size and shape map, different number of civs/cs, no mods, no world wrap: everything works.
- I start any size and shape map, different number of civs/cs, no mods, _with_ world wrap: the move or swap overlay goes missing.
Display of where one could go works, display of possible swaps works, but no actual move/swap button appears. Debug as I said proves the widget gets created, a breakpoint at `WorldMapHolder.addTileOverlays()+29`, the `table.moveBy(0f, 60f)` line, at a cursory glance shows identical properties (especially coords) as for the same game when loaded into current master branch. Reaffirm: switching branches and loading the same fresh turn 0 save switches the problem on/off, at least on this box here. And I _still_ can't find the cause... I'll switch off for a day, maybe that helps - or gives one of you time to repro and then teach me how I should have found the issue in no time...

avdstaaij:
>     * I start any size and shape map, different number of civs/cs, no mods, no world wrap: everything works.
> 
>     * I start any size and shape map, different number of civs/cs, no mods, _with_ world wrap: the move or swap overlay goes missing.

With this additional piece of information, I was able to reproduce the issue! And I believe I have found it's cause as well. It's interesting enough to warrant an explanation.

The `addTileOverlays` function used to take an optional argument containing the data required to draw the move button. Because I added a different kind of button, the swap button, I changed `addTileOverlays` to instead take an already created button, allowing me to call it with both the move button and the new swap button. The problem is that, when world wrap is enabled, there are essentially three "copies" of every tile, and the button is drawn for each of them. Since I changed `addTileOverlays` to take a pre-made button instead of the data to create such a button, the function places the _same_ button at every copy of the selected tile. The button can only be in one place, so it ends up at the last position it is placed, which happens to be the left copy of the map.

amazing-bug

I will commit a fix soon.

SomeTroglodyte:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: 
But I need to facepalm too - I _could_ have guessed because I faced the same thing with the modmanager decoration icons at one point...

avdstaaij:
The world-wrap issue has been fixed!

With that out of the way: @yairm210, could you tell me if anything still needs to be changed about the UI?
Multiple UI options have been suggested in this thread. To summarize:

---

**Option 1**

The behavior as it stands right now: All units have a "swap units" unit action, which puts the unit in a specific unit-swapping mode. A user setting could be added to switch to the behavior of option 2 or option 3.

As for when to show the button, there are two sub-options:
1. The behavior as it stands right now: the button always shows.
2. The button only shows when a swap movement is possible. This may impact performance.


_Advantages:_
It remains easy to select other units that are close to the currently selected unit. All existing selection behavior remains unchanged.

_Disadvantages:_
An additional unit action is added for all units, increasing the cognitive load for users.

---

**Option 2**

If a unit is selected and the user clicks on another unit with which the selected unit could swap, the swap button will show. There is no special unit-swapping mode. A user setting could be added to switch to the behavior of option 1.

For the behavior of the single-click-movement user setting, there are two sub-options:
1. The setting acts as advertised: no swap button is shown, and the swap movement instantly happens on click. This may cause many accidental swap movements where users simply wanted to select another unit.
2. The setting has no effect on unit swaps: the button always shows.

_Advantages:_
No additional unit action action is required.

_Disadvantages:_
It becomes harder to select a different close-by unit: simply clicking on him will create a swap button. To select a close-by unit, the user now has to deselect his current unit, requiring an additional click.

---

**Option 3**

If a unit is selected and the user clicks on another unit with which the selected unit could swap, both a swap button and a "select this unit" button will show, similar to the behavior when clicking on a city center with a unit, as shown in the last screenshot of the initial post. A user setting could be added to switch to the behavior of option 1. I believe this option to be a superior version of option 2.

For the behavior of the single-click-movement user setting, there are two sub-options:
1. The setting causes swap-movement to have priority: no buttons will be shown. This may be annoying.
2. The setting has no effect on this situation: the buttons always show.

_Advantages:_
No additional unit action is required.

_Disadvantages:_
It becomes slightly harder to select a different close-by unit: users now have to click on them and then click on the "select this unit" button, requiring an additional click. The additional effort is however not as bad as in option 2.

---

**Option 4**

Long-presses are interpreted as right clicks, and a unit swap button appears (only) when the target unit is right-clicked.

_Advantages:_
No additional unit action is required, and selecting close-by units remains as easy as before. All existing behavior remains unchanged.

_Disadvantages:_
A new form of input, the long press, is introduced, and will probably need to be explained to the user.

---

**Option 5**

The behavior as it stands right now (option 1), except that the "swap units" unit action, along with some other rarely used actions, is hidden behind a "more actions" unit action button.

_Advantages:_
No additional unit action is required, and selecting close-by units remains as easy as before.

_Disadvantages:_
Still requires an additional unit action, and is far more work to implement than the other options. The addition of a "more actions" button probably deserves its own PR, meaning that if this option is preferred, I believe we should go with option 1 to begin with and add the "more actions" button later.

---

Of course, do tell me if you have another alternative!

yairm210:
Great summary!
I think, considering the different disadvantages, that currently option 1, suboption 2 is the most immediate and minimal.
It's possible that in the future we'll implement option 5, or perhaps 3, but that feels like something for another PR. 
The performance loss in 1.2. should be negligible - only happens for a specific unit, most surrounding tiles can be immediately disregarded, and using sequences should make the whole thing add maybe as millisecond if we're pessimistic.

avdstaaij:
I have changed the unit action button to only appear when there is at least one possible swap movement, so option 1.2 is now fully implemented!

yairm210:
Crashing bug for some players:

  at com.unciv.logic.map.MapUnit.putInTile (MapUnit.kt:611)
  at com.unciv.logic.map.UnitMovementAlgorithms.canUnitSwapToReachableTile (UnitMovementAlgorithms.kt:281)
  at com.unciv.logic.map.UnitMovementAlgorithms.access$canUnitSwapToReachableTile (UnitMovementAlgorithms.kt:8)
  at com.unciv.logic.map.UnitMovementAlgorithms$getUnitSwappableTiles$1.invoke (UnitMovementAlgorithms.kt:251)
  at com.unciv.logic.map.UnitMovementAlgorithms$getUnitSwappableTiles$1.invoke (UnitMovementAlgorithms.kt:8)
  at kotlin.sequences.FilteringSequence$iterator$1.calcNext (Sequences.kt:171)
  at kotlin.sequences.FilteringSequence$iterator$1.hasNext (Sequences.kt:194)
  at kotlin.sequences.SequencesKt___SequencesKt.none (_Sequences.kt:1921)
  at com.unciv.ui.worldscreen.unit.UnitActions.addSwapAction (UnitActions.kt:79)
  at com.unciv.ui.worldscreen.unit.UnitActions.getUnitActions (UnitActions.kt:50)
  at com.unciv.ui.worldscreen.unit.UnitActionsTable.update (UnitActionsTable.kt:69)

avdstaaij:
I will take a look at it!

avdstaaij:
According to the stack trace, this is what happens:
1. A unit is selected.
2. To determine whether to show the "swap units" button, the tiles to which the unit could swap are calculated.
3. During this calculation, `canUnitSwapTo` gets called for some tile.
4. All check pass, up to the final check whether the unit on the target tile can move to our tile.
5. The code below gets executed (this is the `unit.putInTile` from the stack trace). Essentially, it removes the selected unit from its tile, then checks if the other unit can move there and then puts the selected unit back (.
```kotlin
val ourPosition = unit.getTile()
// (...)
// Check if they could enter our tile if we wouldn't be here
unit.removeFromTile()
val theyCanEnterOurTile = otherUnit.movement.canMoveTo(ourPosition)
unit.putInTile(ourPosition)
if (!theyCanEnterOurTile) return false
```

6. When attempting to put the unit back, in `putInTile`, the crash happens at the following line:
```kotlin
!movement.canMoveTo(tile) -> throw Exception("I can't go there!")
```

I can only assume the exception gets thrown. This means that, according to `canMoveTo`, the selected unit can't move to the tile they are standing on!

---

I have no idea what could cause this, but I could probably figure it out if with a save file in which the crash occurs. It would also help to know which unit is being selected before the crash.

Adding the following line before the code from step 5 might prevent the crash:
```kotlin
if (!unit.canMoveTo(ourPosition)) return
```
That would just a band-aid though, and I'm not completely sure if it will work. I'd need a save file to know for sure. @yairm210, could you provide a save file?

yairm210:
Unortunately not, I only get the exception stack trace from Google Play, but Google's privacy policy (which I'm completely in favor of, obviously) means that not only do I not get the data, I don't even know what user got this exception

PR #4055, opened 2021-06-04
Map editor fix translations, a few keys

SomeTroglodyte:
- Currently the map editor ignores translations in its loaded mods
- The checks when changing mods lack translation support
- Back/ESC does nothing, some may expect a reaction

These changes allow e.g.:
image
...the modded penguin hatchery _is_ now xlated.

German already included _for testing_.

@Mape6 FYI: Those two lines of yours I changed - this fits context better IMHO.


Mape6:
You are right. In that context fits your translation better. Good job.

PR #4056, opened 2021-06-05
Map editor save / load / download keys

SomeTroglodyte:
Small stuff like saving an already named map re-offers the name, keys, the label that should disappear actually disappears...
Question: Where is the upload for those downloadable maps? I feel like I should know, but it's not there in the editor? They would definitely need some profanity / spam and maybe age filtering.


yairm210:
There is no more upload, because we encountered all the problems that came with the mod as well (versioning, management, authorship, obscenities).
Maps can now be shared only by adding them to mods.

yairm210:
In fact, I think it's high time we remove the download map option altogether. But that's for another day...

SomeTroglodyte:
> remove the download map option altogether

Easy and would touch only a limited set of source... But - the actual dropbox folder - I'm not too familiar - does dropbox have the UI to remove it in one go or would we need to script the deletes somehow?

Issue #663, opened 2019-04-19
Missing features from Civ V


Smashfanful:
As a GRAND POLICY CHANGE, we're changing our goal from Vanilla Civ V to G&K Civ V.

The items in this issue are still relevant,

- [x] Aircraft Carriers
- [x] Civilization Rankings (Richest, Strongest, Most Fertile, Largest, Happiest, Most Productive)
- [x] No Ancient Ruins starting option
- [x] Owned Cities need a growth bar, to see in how many turns the City will grow
- [x] Random Civ choice in the Civ select window
- [ ] Diplomatic Victory
- [ ] Peace request, denunciation and friendship request quotes
- [ ] Civ timeline for momentous events
- [x] City build queue
- [ ] Unique denunciation quotes (?)
- [ ] When about to promote an unity, you can change their name
- [ ] Missing wonders - Petra, United Nations
- [x] Civilizations - Aztecs, Mongols, Incas, Danish, etc.
- [x] Forts and Citadels
- [x] Research agreements
- [ ] Civ Teams on new game
- [ ] Cities requiring a resource and consequent "Hail the King" celebrations
- [ ] A Civ can ask another Civ to make peace with a third Civ
- [x] Policy Branch: Order
- [x] Policy Branch: Patronage
- [ ] "Build road to" action for workers
- [ ] Achievements (Google Play exclusive)

**City states**

 - [x] Cities owned by City-States should have an Influence bar to see what are their relations with you.
 - [x] Should give gold when first met
 - [x] City-states with good relations with you may give you resources
 - [x] Personalities
 - [x] Quests

Am-per-Sand:
what is _Allow Policy accumulation_?

BTW - on immortal and deity on small maps never can find any ruins

Smashfanful:
@Am-per-Sand _Allow Policy accumulation_ allows to accumulate Culture points over the minimum limit for a new policy, without forcing you to choose one.

Am-per-Sand:
That's funny - what can a player use such overload of points for?

yairm210:
This is a mix of a few things.
First is #559 
Second is civ rankings
Third is 'policy accumulation' which from your description already exists so I'm not sure what you want
Please move all game starting options to 559
And call this "add civ rankings"

amunizp:
So Spain is only missing natural wonder? According to #703 ?

r3versi:
All vanilla NW are in game and so is Spain

amunizp:
[...]

PR #4058, opened 2021-06-05
Refund wasted production as gold

### LINKED ISSUE #4047: The game should compensate for the production wasted in certain situations
Situation 1: A wonder providing a building you are halfway building it. for instance being at 30 production on the library just when you built great library

Situation 2: a wonder being built elsewhere when you have already invested in production to it

those should not be complex to implement in game right?

Thyrum:
Any wasted production, from for instance a partially completed wonder or obsolete units, will now be refunded as gold (in a 1:1 ratio). When obsolete units get replaced by their new version, this code should probably be updated.

I first added the gold refund code in `validateInProgressConstructions`, but this function is only called upon `endTurn`. Hence a refund for a wonder build elsewhere would occur after the relevant turn instead of before the turn. Hence the implementation is now put in `validateConstructionQueue`.

The current implementation does result in code duplication between `validateInProgressConstructions` and `validateConstructionQueue`. Any ideas on how to neatly remove this duplication are appreciated.

Another thing to consider would be a notification/popup explaining why the gold was added, but this probably clutters the interface too much.

Fixes #4047

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -311,8 +311,9 @@ class CityConstructions {
         // Should at least be called before each turn - if another civ completes a wonder after our previous turn, we should get the refund this turn
         val inProgressSnapshot = inProgressConstructions.keys.filter { it != currentConstructionFromQueue }
         for (constructionName in inProgressSnapshot) {
+            val construction = getConstruction(constructionName)
             val rejectionReason: String =
-                    when (val construction = getConstruction(constructionName)) {
+                    when (construction) {
                         is Building -> construction.getRejectionReason(this)
                         is BaseUnit -> construction.getRejectionReason(this)
                         else -> ""
@@ -323,8 +324,24 @@ class CityConstructions {
                     || rejectionReason.startsWith("Don't need to build any more")
                     || rejectionReason.startsWith("Obsolete")
             ) {
-              cityInfo.civInfo.gold += getWorkDone(constructionName)
-              inProgressConstructions.remove(constructionName)
+                if (construction is Building) {
+                    // Production put into wonders gets refunded
+                    if (construction.isWonder && getWorkDone(constructionName) != 0) {
+                        cityInfo.civInfo.gold += getWorkDone(constructionName)
+                        val buildingIcon = "BuildingIcons/${constructionName}"
+                        cityInfo.civInfo.addNotification("Could not continue work on [${constructionName}], excess production converted to [${getWorkDone(constructionName)}] gold", NotificationIcon.Gold, buildingIcon)
+                    }
+                } else if (construction is BaseUnit) {
+                    // Production put into upgradable units gets put into upgraded version
+                    if (rejectionReason.startsWith("Obsolete") && construction.upgradesTo != null) {
+                        if (!inProgressConstructions.contains(construction.upgradesTo)) {
+                            inProgressConstructions[construction.upgradesTo!!] = getWorkDone(constructionName)
+                        } else {
+                            inProgressConstructions[construction.upgradesTo!!] = inProgressConstructions[construction.upgradesTo!!]!! + getWorkDone(constructionName)
+                        }
+                    }
+                }
+                inProgressConstructions.remove(constructionName)
             }
         }
     }
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -329,7 +329,7 @@ class CityConstructions {
                     if (construction.isWonder && getWorkDone(constructionName) != 0) {
                         cityInfo.civInfo.gold += getWorkDone(constructionName)
                         val buildingIcon = "BuildingIcons/${constructionName}"
-                        cityInfo.civInfo.addNotification("Could not continue work on [${constructionName}], excess production converted to [${getWorkDone(constructionName)}] gold", NotificationIcon.Gold, buildingIcon)
+                        cityInfo.civInfo.addNotification("Excess production for [$constructionName] converted to [${getWorkDone(constructionName)}] gold", NotificationIcon.Gold, buildingIcon)
                     }
                 } else if (construction is BaseUnit) {
                     // Production put into upgradable units gets put into upgraded version
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -334,6 +334,7 @@ class CityConstructions {
                 } else if (construction is BaseUnit) {
                     // Production put into upgradable units gets put into upgraded version
                     if (rejectionReason.startsWith("Obsolete") && construction.upgradesTo != null) {
+                        // I'd love to use the '+=' operator but since 'inProgressConstructions[...]' can be null, kotlin doesn't allow me to
                         if (!inProgressConstructions.contains(construction.upgradesTo)) {
                             inProgressConstructions[construction.upgradesTo!!] = getWorkDone(constructionName)
                         } else {
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -272,6 +272,9 @@ class CityConstructions {
     fun constructIfEnough() {
         validateConstructionQueue()
 
+        // Update InProgressConstructions for any available refunds
+        validateInProgressConstructions()
+
         val construction = getConstruction(currentConstructionFromQueue)
         if (construction is PerpetualConstruction) chooseNextConstruction() // check every turn if we could be doing something better, because this doesn't end by itself
         else {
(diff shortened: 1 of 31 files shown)


Thyrum:
Just noticed a quite big big. This only accounts for constructions that were in your queue when they became obsolete, not for ones that you had already removed from your queue. I will fix this shortly, until then this will be a draft.

Thyrum:
The code has now been updated. This did however require calling `validateInProgressConstructions` from `constructIfEnough` since `constructIfEnough` is called at the start of each turn. It might therefore be better to add a `startTurn` function which calls both `constructIfEnough` and `validateInProgressConstructions`.

ravignir:
in civ5 all of the production invested into obsoleted units is transferred to progress towards their respective upgrades.

Thyrum:
> 
> 
> in civ5 all of the production invested into obsoleted units is transferred to progress towards their respective upgrades.

I am aware of this but decided not to implement this yet because the issue did not include this (and I still need to get a little more familiar with the codebase 🙃). It can always be added in a later pull request.

SomeTroglodyte:
I'm not convinced the original game refunded you for a lost wonder race at all. I would be fully in favour of this if the refund ratio were moddable - then we could easily make it zero in the vanilla ruleset and thereby postpone the final decision while mods could do what they want... 

Production for obsoleted units being preserved and applied to the successor units is IMHO a completely separate matter, as is production overflow - which Civ4 (four not five) definitely had, and Civ5 99.9% _not_.

xlenstra:
> I'm not convinced the original game refunded you for a lost wonder race at all.

By googling it I found multiple sources saying you were refunded with some amount of gold. Next week I can do some tests to determine how much exactly.

> [...], as is production overflow - which Civ4 (four not five) definitely had, and Civ5 99.9% _not_.

_Actually_, according to this page of the civ 5 forums, this is an extensively documented mechanic:
https://forums.civfanatics.com/threads/hammer-overflow.419352/

ravignir:
I've recorded all of the cases in vanilla civ5:

https://user-images.githubusercontent.com/30041793/120906850-23370a80-c65d-11eb-960c-1f46a3798697.mp4


https://user-images.githubusercontent.com/30041793/120906869-46fa5080-c65d-11eb-850f-1a98ce207d89.mp4


https://user-images.githubusercontent.com/30041793/120906878-5a0d2080-c65d-11eb-91e2-4c9664a67c84.mp4

anyway, the topic on civfamatics linked by @xlenstra explains production overflow really well.

SomeTroglodyte:
... and ravignir's input of course is beyond doubt. :+1:

ravignir:
Here is what happens when an obsoleted unit has no upgrade:
https://user-images.githubusercontent.com/30041793/120907153-8c1f8200-c65f-11eb-9340-4ca25a31626b.mp4

You get a regular production overflow and nothing more.

Thyrum:
So if I understand it correctly, we have observed that in Civ 5:
 * production spent on a wonder built elsewhere is converted into gold (with a 1:1 ratio, notified to the user via notification).
 * production spent on a unit which has become obsolete is put into the upgraded version of the unit.
 * all other 'lost' production is stored separately (as overflow) and added to the production of the next turn. The overflow stored can be at most the current production income or cost of the last item, whichever is greater.

I suggest these points get added to #663. I can implement the first to points quite easily and then have this branch merged. The third point can then be added in a later pull request.

ravignir:
> 
> 
> So if I understand it correctly, we have observed that in Civ 5:
> 
>     * production spent on a wonder built elsewhere is converted into gold (with a 1:1 ratio, notified to the user via notification).
> 
>     * production spent on a unit which has become obsolete is put into the upgraded version of the unit.
> 
>     * all other 'lost' production is stored separately (as overflow) and added to the production of the next turn. The overflow stored can be at most the current production income or cost of the last item (the one that caused overflow), whichever is greater.
> 
> 
> I suggest these points get added to #663. I can implement the first to points quite easily and then have this branch merged. The third point can then be added in a later pull request.

yes, notice the 12 production invested into scout in the last video is lost forever.

yairm210:
This looks good - I have no comments at all on the existing code!
I do have another addition, though - if the user is getting refunded, the user notification (could not continue work on []) should also contain "Excess Production converted to [] Gold", in both the notification itself and the translation line in template.properties

SomeTroglodyte:
> The overflow stored can be at most...

Didn't I read a -1 somewhere? 

Otherwise :+1:. I could only add the reminder - If it's not too much hassle then allowing mods to change the behaviours implemented for this might be nice.

Thyrum:
I have implemented both gold refund for wonders and changing obsolete units into their upgraded form (if available). The relevant notifications can be seen in the screenshots below. I currently only alert users of the unit production transfer if the relevant unit was queued. I have chosen to do it this way since you probably don't care for the production swap if the unit wasn't queued.

The civlopedia should probably also be updated to reflect the changes made.
image
image
(note the last screenshot has a comma in the notification, not a dot)

SomeTroglodyte:
Time to optimize the wording (not nitpicking, if this weren't so good I woudn't bother)? I feel the notifications are a bit long (I can already _see_ the issues from people with 640x360 @ 72dpi screens) and 'has been obsolete' doesn't sound quite perfect... So,  just thinking & feel free to ignore:

- `Could not continue work on [wonder], excess production converted to [goldAmount] gold = ` hmm - if it's guaranteed you'll be getting the "[wonder] has been built by those other meanies" note right alongside, might "Excess production for [wonder] converted to [goldAmount] gold = " suffice?
- `[oldUnit] has been obsolete and production will be changed to [newUnit] in [cityName]! = ` ... ... "[cityName] changed production from [oldUnit] to [newUnit]"? Maybe people are clever enought to guess - or learn - that the reason is obsoletion? Also: If N cities have these - maybe a single message "[count] cities..." with a LocationAction cycling through all of them? Just to save screen 'real estate'?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -319,7 +323,26 @@ class CityConstructions {
...
+                        val buildingIcon = "BuildingIcons/${constructionName}"
+                        cityInfo.civInfo.addNotification("Could not continue work on [${constructionName}], excess production converted to [${getWorkDone(constructionName)}] gold", NotificationIcon.Gold, buildingIcon)
+                    }
+                } else if (construction is BaseUnit) {
+                    // Production put into upgradable units gets put into upgraded version
+                    if (rejectionReason.startsWith("Obsolete") && construction.upgradesTo != null) {
+                        if (!inProgressConstructions.contains(construction.upgradesTo)) {
+                            inProgressConstructions[construction.upgradesTo!!] = getWorkDone(constructionName)
+                        } else {
+                            inProgressConstructions[construction.upgradesTo!!] = inProgressConstructions[construction.upgradesTo!!]!! + getWorkDone(constructionName)
```
SomeTroglodyte, line 340: Kotlin does have `+=` (feel free to ignore if that's just your style)

Thyrum: Previously I used the `+=` but got an error because `inProgressConstructions[construction.upgradesTo!!]` can return `null` and in that case the `+=` operator is not defined. What I have now is my workaround for that. Is there a good/standard way to work around this?

avdstaaij:
> So if I understand it correctly, we have observed that in Civ 5:
> (...)
> * all other 'lost' production is stored separately (as overflow) and added to the production of the next turn. The overflow stored can be at most the current production income or cost of the last item (the one that caused overflow), whichever is greater.

Minor note: I'm fairly sure that production put into a building that you then rush-buy is not refunded in any way.

Thyrum:
> I feel the notifications are a bit long

I totally agree but couldn't think of a shorter wording at that time. I will shorten the wording to your suggestions.


> [...] and 'has been obsolete' doesn't sound quite perfect

I agree but the existing notification for removing obsolete constructions was
```
[construction] has been obsolete and will be removed from construction queue in [cityName]!
```
thus I used this wording in order to be more uniform. I would personally prefer to change the old and new notification to
```
[construction] has become obsolete and will be removed from construction queue in [cityName]!
[cityName] changed production from [oldUnit] to [newUnit]
```

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: (why does it say 'reviewed'? Did I hit the wrong button?)
> `+=`

Right. My bad. :crying_cat_face:   ...Then the Elvis<sup>®</sup> is the most 'Kotlin'-elegant short form... or `map[key] = delta + (if (key in map) map[key]!! else 0)` or... All versions including yours probably give the same bytecode or close.

Tempts one to overload HashMap to allow adding to nonexisting entries directly - oh, wait, that's [already](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/models/Counter.kt) been done in Unciv... But changing the `inProgressConstructions` type could break serialization.

SomeTroglodyte (approved):
Sorry for accidentally hitting the green pill instead of the white one.

Thyrum:
The notifications have now been shortened to what's below. Should I also just change the `has been obsolete` to `has become obsolete` in all `.properties` language files? Or do we maybe want to reword it entirely (because it is quite a long notification)?
```
[construction] has been obsolete and will be removed from construction queue in [cityName]! = 
[construction] has been obsolete and will be removed from construction queue in [amount] cities! = 
[cityName] changed production from [oldUnit] to [newUnit] = 
[amount] cities changed production from [oldUnit] to [newUnit] = 
Excess production for [wonder] converted to [goldAmount] gold = 
```

yairm210:
"has become obsolete"
Also if you want to make it shorter, 

[construction] has become obsolete and was removed from the queue in [cityName] =

ravignir:
> 
> 
> > So if I understand it correctly, we have observed that in Civ 5:
> > (...)
> > 
> > * all other 'lost' production is stored separately (as overflow) and added to the production of the next turn. The overflow stored can be at most the current production income or cost of the last item (the one that caused overflow), whichever is greater.
> 
> Minor note: I'm fairly sure that production put into a building that you then rush-buy is not refunded in any way.

yup, it is not.

Thyrum:
Rewording of `been obsolete -> become obsolete` has been added. I think we are ready to merge.

logicminimal:
happy to see this resolved

Thyrum:
For completeness sake, this is yet to be implemented and will require a different pull request:
> all other 'overflow' production is stored separately (as overflow) and added to the production of the next turn. The overflow stored can be at most the current production income or cost of the last item (the one that caused overflow), whichever is greater.

PR #4060, opened 2021-06-05
Updated the culture victory so it now requires the Utopia Project to be built

xlenstra:
The utopia project is unlocked after 5 policy trees have been completed.
The AI will try to build it as fast as possible when it has it unlocked.
When a nation starts to build it, this city (if it is known) is named in a global alert send to all nations.


yairm210:
Question: Was this is Vanilla Civ? Or only in expansions?

android/assets/jsons/Civ V - Vanilla/Buildings.json
```diff
@@ -937,7 +937,7 @@
 	{
 		"name": "Manhattan Project",
 		"isNationalWonder": true,
-		"uniques": ["Enables nuclear weapon", "Triggers a global alert upon completion"],
+		"uniques": ["Enables nuclear weapon", "Triggers a global alert upon [completion]"],
```
yairm210, line 940: I like the idea of templating this, but in practice it means an additional translation for "completion" - which will arrive without context, a separate notification, which means that anyway you have only the preprogrammed options...
In short, in this specific instance, templating adds complexity instead of removing it.

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
```diff
@@ -187,6 +186,11 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
         if (preferredVictoryType == VictoryType.Cultural
                 && wonder.name in listOf("Sistine Chapel", "Eiffel Tower", "Cristo Redentor", "Neuschwanstein", "Sydney Opera House"))
             return 3f
+        // Only start building if we are the city that would complete it the soonest
+        if (wonder.name == "Utopia Project" && cityInfo == civInfo.cities.minByOrNull {
```
yairm210, line 190: I know that there are other buildings referenced by name, but that's an antipattern - buildings should be referenced by unique, to enable modding

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
```diff
@@ -187,6 +186,11 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
         if (preferredVictoryType == VictoryType.Cultural
                 && wonder.name in listOf("Sistine Chapel", "Eiffel Tower", "Cristo Redentor", "Neuschwanstein", "Sydney Opera House"))
             return 3f
+        // Only start building if we are the city that would complete it the soonest
+        if (wonder.name == "Utopia Project" && cityInfo == civInfo.cities.minByOrNull { 
+                it.cityConstructions.getRemainingWork(it.cityConstructions.getCurrentConstruction().name) / cityInfo.cityStats.currentCityStats.production.roundToInt() + wonder.cost / it.cityStats.currentCityStats.production
```
yairm210, line 191: I don't understand the calculation here, something seems confused

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -225,7 +225,7 @@ class CityConstructions {
     fun turnsToConstruction(constructionName: String, useStoredProduction: Boolean = true): Int {
         val workLeft = getRemainingWork(constructionName, useStoredProduction)
         if (workLeft < 0) // we've done more work than actually necessary - possible if circumstances cause buildings to be cheaper later
-            return 1 // we'll finish this next turn
+            return 0 // we'll finish this next turn
```
yairm210, line 229: You can't finish a building mid-turn, even with boosts

xlenstra: Suppose during a multiplayer game you are building something that has only 20 production remaining, and you have 10 production/turn in this specific city. Then it will of course say you have 2 turns remaining to build it. After pressing 'next turn', the production will be added, and while you are waiting for other players to do their turn, the city label notes that you have only 1 turn remaining. During your next turn you still have 1 turn remaining, but after pressing 'next turn' again, you have provided exactly enough work into the building for it to finish. In this the label would read 0 turns remaining. _However_, if you had 11 production/turn, at this point 'workLeft == -2', so instead (before this change), it would say you would still have 1 turn remaining for it to finish building. This change is only so that this minor inconsistency is fixed.

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
xlenstra: What is does is basically:

For each city:
    Calculate the amount of turns remaining for building the current construction
    Add to this the amount of turns required to build the Utopia Project
If the current city has the lowest amount of turns among all cities:
    Start building the utopia project here

android/assets/jsons/Civ V - Vanilla/Buildings.json
xlenstra: Good point, I'll revert this change

xlenstra:
> Question: Was this is Vanilla Civ? Or only in expansions?

The wiki page for this existed long before G&K came out, so I believe this is also in Vanilla Civ

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
xlenstra: I'll change this

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
xlenstra: Looking at it further I did use the wrong variable once, should be fixed in the next commit

core/src/com/unciv/logic/city/CityConstructions.kt
xlenstra: Or, in other words: during multiplayer games it was sometimes possible for a construction to show '0 turns remaining' or '1 turn remaining' during another players turn, even though in both cases the construction would finish at the start of the next turn. This fixes this, but does introduce the more rare bug where if production requirements change for a building during your turn, it might say '0 turns remaining'.

yairm210: I see, thanks for the explanation!

ravignir:
img

android/assets/jsons/Civ V - Vanilla/Buildings.json
```diff
@@ -1058,4 +1058,13 @@
...
 		"uniques": ["Spaceship part", "Triggers a global alert upon completion", "Cannot be purchased"]
 	}
+	
+	// All Era's
+	
+	{
+		"name": "Utopia Project",
+		"cost": 1500,
+		"isNationalWonder": true,
+		"uniques": ["Hidden until [5] [social policy branches have been completed]", "Triggers a global alert upon build start", "Wins the game when built"]
```
yairm210, line 1068: Here too the templating of the text doesn't help, since it doesn't increase the options for modders. The number template is good

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -347,7 +368,7 @@ class CityConstructions {
             cityInfo.civInfo.addNotification("[${construction.name}] has been built in [" + cityInfo.name + "]",
                     cityInfo.location, NotificationIcon.Construction, icon)
         }
-        if (construction is Building && "Triggers a global alert upon completion" in construction.uniques) {
+        if (construction is Building && construction.uniqueObjects.any { it.placeholderText == "Triggers a global alert upon completion]" } ) {
```
yairm210, line 371: Additional ] at the end of the string

android/assets/jsons/Civ V - Vanilla/Buildings.json
yairm210: Also "Wins the game when built" -> "Triggers Cultural victory upon completion"

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
```diff
@@ -187,6 +186,11 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
         if (preferredVictoryType == VictoryType.Cultural
                 && wonder.name in listOf("Sistine Chapel", "Eiffel Tower", "Cristo Redentor", "Neuschwanstein", "Sydney Opera House"))
             return 3f
+        // Only start building if we are the city that would complete it the soonest
+        if (wonder.uniques.contains("Wins the game when built") && cityInfo == civInfo.cities.minByOrNull { 
+                it.cityConstructions.getRemainingWork(it.cityConstructions.getCurrentConstruction().name) / it.cityStats.currentCityStats.production.roundToInt() + wonder.cost / it.cityStats.currentCityStats.production
```
yairm210, line 191: CityConstructions contains a function of getTurnsToCompletion or something of that sort

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -323,6 +327,23 @@ class CityConstructions {
         }
     }
 
+    private fun constructionBegun(construction: IConstruction) {
+        if (construction is Building) {
```
yairm210, line 331: "Exit quickly" is a Good Practice (tm) since it both reduces doubt and indentation and makes functions easier to parse

if (construction !is Building) return
if (construction.uniqueObjects.none { it.placeholderText == "Triggers a global alert upon build start" }) return

core/src/com/unciv/logic/civilization/VictoryManager.kt
```diff
@@ -36,7 +36,7 @@ class VictoryManager {
     fun hasWonScientificVictory() = hasVictoryType(VictoryType.Scientific) && spaceshipPartsRemaining() == 0
 
     fun hasWonCulturalVictory() = hasVictoryType(VictoryType.Cultural)
-            && civInfo.policies.adoptedPolicies.count { it.endsWith("Complete") } > 4
+            && civInfo.cities.any{ it.cityConstructions.builtBuildingUniqueMap.getUniques("Wins the game when built").isNotEmpty() }
```
yairm210, line 39: civInfo.containsUnique("Triggers Cultural Victory when built")

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -198,15 +198,15 @@ class BaseUnit : INamed, IConstruction {
         var XP = cityConstructions.getBuiltBuildings().sumBy { it.xpForNewUnits }
 
 
-        for (unique in cityConstructions.cityInfo.cityConstructions.builtBuildingUniqueMap
+        for (unique in cityConstructions.builtBuildingUniqueMap
```
yairm210, line 201: Nice catch :) Wonder what the history behind this was

SomeTroglodyte: 2020-08-05: `for (unique in construction.cityInfo.civInfo.getMatchingUniques2("New military units start with [] Experience"))`
2020-08-06: `for (unique in construction.cityInfo.civInfo.getMatchingUniques("New military units start with [] Experience"))`
2020-08-14: `for (unique in civInfo.getMatchingUniques("New military units start with [] Experience"))`
2020-08-18: `for (unique in construction.cityInfo.cityConstructions.builtBuildingUniqueMap.getUniques("All newly-trained [] units in this city receive the [] promotion")) {`
... @HadeanLake dunnit

SomeTroglodyte: Corrected history excerpt:
2020-08-14: `if (unit.type in listOf(UnitType.Melee,UnitType.Mounted,UnitType.Armor) && construction.cityInfo.containsBuildingUnique("All newly-trained melee, mounted, and armored units in this city receive the Drill I promotion")`
2020-08-18: +`for (unique in construction.cityInfo.cityConstructions.builtBuildingUniqueMap.getUniques("All newly-trained [] units in this city receive the [] promotion")) {`
    ... replacing `cityInfo.containsBuildingUnique` with `cityInfo.cityConstructions.builtBuildingUniqueMap.getUniques` without realizing that `cityInfo` in that case came from `cityConstructions` ...
2020-10-14: +`for (unique in construction.cityInfo.cityConstructions.builtBuildingUniqueMap.getUniques("New [] units start with [] Experience in this city")`
    ... simply copied ...

xlenstra:
All the requested changes should have been made

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
```diff
@@ -187,6 +186,11 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
         if (preferredVictoryType == VictoryType.Cultural
                 && wonder.name in listOf("Sistine Chapel", "Eiffel Tower", "Cristo Redentor", "Neuschwanstein", "Sydney Opera House"))
             return 3f
+        // Only start building if we are the city that would complete it the soonest
+        if (wonder.uniques.contains("Triggers a Cultural Victory upon completion") && cityInfo == civInfo.cities.minByOrNull {
+                cityInfo.cityConstructions.turnsToConstruction(it.cityConstructions.getCurrentConstruction().name) + cityInfo.cityConstructions.turnsToConstruction(wonder.name)
```
yairm210, line 191: Surely we don't care about the current construction, since we're attempting to replace it?

xlenstra: Ah, I see what happened here, I messed up the last update, I'll fix it this night

android/assets/jsons/translations/template.properties
```diff
@@ -403,6 +403,9 @@ Cannot provide unit upkeep for [unitName] - unit has been disbanded! =
 [wonder] has been built in a faraway land = 
 [civName] has completed [construction]! = 
 An unknown civilization has completed [construction]! = 
+[cityname] has started constructing [construction]! = 
+[civilization] has started construction [construction]! =
```
yairm210, line 407: Spelling error - 'constructing'

core/src/com/unciv/logic/civilization/VictoryManager.kt
```diff
@@ -36,7 +36,7 @@ class VictoryManager {
     fun hasWonScientificVictory() = hasVictoryType(VictoryType.Scientific) && spaceshipPartsRemaining() == 0
 
     fun hasWonCulturalVictory() = hasVictoryType(VictoryType.Cultural)
-            && civInfo.policies.adoptedPolicies.count { it.endsWith("Complete") } > 4
+            && civInfo.cities.any{ it.cityConstructions.builtBuildingUniqueMap.getUniques("Triggers a Cultural Victory upon completion").isNotEmpty() }
```
yairm210, line 39: civInfo.hasUnique()

yairm210:
It seems like the Utopia project is buildable even when cultural victory is disabled

xlenstra:
I suppose I should test my code better before opening PR's 😖
I'll work on it tonight, and come back when I believe it to actually work well.

android/assets/jsons/Civ V - Vanilla/Buildings.json
```diff
@@ -1046,4 +1046,14 @@
...
 	}
+	
+	// All Era's
+	
+	{
+		"name": "Utopia Project",
+		"cost": 1500,
+		"isNationalWonder": true,
+		"uniques": ["Hidden until [5] social policy branches have been completed", "Triggers a global alert upon build start", 
+			"Triggers a Cultural Victory upon completion", "Hidden when cultural victory is disabled"]
```
yairm210, line 1057: I think that "Triggers a Cultural Victory upon completion" is enough - when you see that, hide when disabled - and we don't need a separate unique for "hidden when victory disabled".

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
```diff
@@ -187,6 +186,11 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
         if (preferredVictoryType == VictoryType.Cultural
                 && wonder.name in listOf("Sistine Chapel", "Eiffel Tower", "Cristo Redentor", "Neuschwanstein", "Sydney Opera House"))
             return 3f
+        // Only start building if we are the city that would complete it the soonest
+        if (wonder.uniques.contains("Triggers a Cultural Victory upon completion") && cityInfo == civInfo.cities.minByOrNull {
+                it.cityConstructions.turnsToConstruction(it.cityConstructions.getCurrentConstruction().name) + it.cityConstructions.turnsToConstruction(wonder.name)
```
yairm210, line 191: Again, I think we only need `it.cityConstructions.turnsToConstruction(wonder.name) ` and we can ignore the current construction.

xlenstra: This calculation also includes the time it takes before we can start constructing the wonder. If our capital can build it in 10 turns, and we can build it in 15 turns, but our capital is still taking 8 turns building the manhatten project, we are faster to build it here than there: 15 turns instead of 18. If we were to replace the current building, this would of course be invalidated.
This is based on the assumption this function is only called whenever the old production is finished, is that assumption incorrect?

android/assets/jsons/Civ V - Vanilla/Buildings.json
xlenstra: Isn't that actually an assumption that may not always be true in different mods? Maybe you want your mod to have an extensive path of multiple buildings that should be build to win a cultural victory, and disabling the first improves the AI so that it won't go that route? This way it is more modular, although I agree that it is cluttering space a bit more.

Another reason to do this, would be to template it to "Hidden when [] victory is disabled", so that we can hide all spaceship parts & the apollo program for the science victory in the same way.

yairm210: Now THAT is a nice idea! I agree that a templated version could solve other problems

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
yairm210: This function is called every turn, for the AI, it can decide to switch productions in the middle
A building that basically done the game is definitely important enough to stop all other construction

xlenstra: Alright, than you are completely correct, and I'll change it.

yairm210:
It looks like there's  a conflict in the PNG
I think the best way to resolve this would be for you to either A. pull the unciv/master into your branch or B. revert the changes to the png (it'll be rebuilt upon game run anyway)

xlenstra:
I'll fix this shortly

xlenstra:
Should be fixed

yairm210:
Right!
This has been a busy PR and no mistake, but it is a major gameplay change, so that's to be expected

PR #4065, opened 2021-06-06
Fix gold able to over- and underflow

### LINKED ISSUE #4064: My gold vanishes!
I have exactly 2147483646 gold. But Unciv lies:
image

I earn gold - pass a turn. Unciv has _stolen all **my** gold and then some!!!!!!_:
image

## ~~Unacceptable! That outrage has to stop!~~

OK, sorry for the bad joke. Hope someone gets a chuckle out of this...

SomeTroglodyte:
Resolves #4064  :astonished:

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -122,7 +122,7 @@ object Battle {
         }
 
         civUnit.getCivInfo().policies.addCulture(cultureReward)
-        civUnit.getCivInfo().gold += goldReward
+        civUnit.getCivInfo().addGold(goldReward)
     }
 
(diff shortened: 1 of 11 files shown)


yairm210:
And the proud winner of "most niche PR" is...

SomeTroglodyte:
image O U C H - you actually merged the joke? Wow. _Respect_.

You'll probably even punish me by publishing the attribution in big fat bold on the play store...

yairm210:
I merged this because it makes it easy to see where we're setting the value of the civ gold, to make sure that most interactions only add or decrease, and don't set the value directly.

will-ca:
Oh, so *that's* why this one particular resource has oddly strict rules for setting!

PR #4073, opened 2021-06-07
Trade UI - Leader portraits, keys, layout

SomeTroglodyte:
This is closer to the screenshot in #4067 but I left the "gold" font symbol out - because I couldn't resist reworking TradeOffer / TradeType and other Diplomacy entities would merit the same changes (e.g. the when block in `getDiplomacyModifiersTable` could be replaced by enum properties), and it somehow no longer seemed really worthwhile. Will pack as separate PR as an excercise.

### Changes, from nice to marginal:
- Leadername, nation icon, portrait and greeting wrapped in a new Table, reused where applicable (I hope I didn't miss a place) → consistent looks where a virtual conversation takes place.
    - Places: Diplomacy, WarDeclaration, Defeated, FirstContact, BorderConflict, DemandToStopSettlingCitiesNear, CitySettledNearOtherCivDespiteOurPromise, DeclarationOfFriendship, TradePopup
- `getLeaderDisplayName` adds AI/Human only in actual multiplayer games. As single player I don't need to know all the others are AI's? Was I right in thinking that?
- AlertPopups can be closed with Back/ESC and with space - if they have at least one button without side effect.
- AlertPopups DemandToStopSettlingCitiesNear, DeclarationOfFriendship: Can be answered with Y or N keys.
- AlertPopup CityConquered: Can be answered with keys (D, A, P, R)
- AlertPopups BorderConflict, WarDeclaration: Separated the glued-together buttons (when do we make the choice actually matter?)
- TradePopup can be answered with keys Y, N or E(lse).
- TradePopup: Extra empty column in their offers (e.g. they sue for peace) gone - bad colspan
- TradePopup: Extra empty column between the two offer tables - to better visually separate them.
- TradePopup: Extra separator below trade 'content' to better set if off optically
- DiplomacyScreen: Centered close button with the nation symbols (imprecise, would need to subtract at least part of SplitPane's handle width - who's going to notice)
- DiplomacyScreen: The "Very well" reactions to denounce and declare war can be dismised with the space key (and the whole screen already accepts ESC to close).

### Issues
- A glaring white separator where I wanted one was too intrusive so I coded a separate one. I now know of three separator implementations - these would need rethink & merging sometime. But: one uses an actual png pixel to get its colour, one pulls the SplitPane handle colour from the skin and one sets it directly on a Table.background - I'm unsure how to unify these. Use `Table.background = ImageGetter.getBackground` method for all cases, hardcoding the one skin use away?
- No no other DiplomacyScreen keyboard support yet, needs some thought.
- The "has accepted your trade request" and "has denied your trade request" choice of words is used for all trade types, at least peace negotiations could use a more specific text.
- Left a fat TODO with these...

### Screenshots
Note: I doubled the space Name - Nation icon after taking some of these, and my leader portraits have around 40-50 transparent pixels around actual content which makes the spacing look a bit too wide on that side - it's the png not the padding.

#### FirstContact no portrait - no surprises
image

#### FirstContact with portrait
image

#### On DiplomacyScreen
image

#### TradePopup ***before***
image

#### TradePopup _after_
image

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -365,9 +365,16 @@ class CivilizationInfo {
 
     fun isAtWar() = diplomacy.values.any { it.diplomaticStatus == DiplomaticStatus.War && !it.otherCiv().isDefeated() }
 
+    /**
+     * Returns a civilization caption suitable for greetings including player type info:
+     * Like "Milan" if the nation is a city state, "Caesar of Rome" otherwise, with an added
+     * " (AI)", " (Human - Hotseat)", or " (Human - Multiplayer)" if the game is multiplayer.
+     */ 
     fun getLeaderDisplayName(): String {
         return nation.getLeaderDisplayName().tr() +
             when {
+                !gameInfo.gameParameters.isOnlineMultiplayer ->
+                    ""
                 playerType == PlayerType.AI ->
                     " (" + "AI".tr() + ")"
                 gameInfo.civilizations.count { it.playerType == PlayerType.Human } > 1 ->
--- a/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
+++ b/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
@@ -49,7 +49,7 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
         closeButton.labelCell.pad(10f)
         closeButton.pack()
         closeButton.y = stage.height - closeButton.height - 10
-        closeButton.x = 10f
+        closeButton.x = (stage.width * 0.2f - closeButton.width) / 2   // center, leftSideTable.width not known yet
         stage.addActor(closeButton) // This must come after the split pane so it will be above, that the button will be clickable
     }
 
@@ -99,11 +99,7 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
         val diplomacyTable = Table()
         diplomacyTable.defaults().pad(10f)
 
-        val displayNameTable = Table()
-        displayNameTable.add(ImageGetter.getNationIndicator(otherCiv.nation, 24f))
-                .pad(0f, 0f, 5f, 10f)
-        displayNameTable.add(otherCiv.getLeaderDisplayName().toLabel(fontSize = 24))
-        diplomacyTable.add(displayNameTable).row()
+        diplomacyTable.add(LeaderIntroTable(otherCiv)).row()
 
         diplomacyTable.add("{Type}:  {${otherCiv.cityStateType}}".toLabel()).row()
         diplomacyTable.add("{Personality}:  {${otherCiv.cityStatePersonality}}".toLabel()).row()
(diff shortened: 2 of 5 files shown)


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -365,9 +365,16 @@ class CivilizationInfo {
...
 
+    /**
+     * Returns a civilization caption suitable for greetings including player type info:
+     * Like "Milan" if the nation is a city state, "Caesar of Rome" otherwise, with an added
+     * " (AI)", " (Human - Hotseat)", or " (Human - Multiplayer)" if the game is multiplayer.
+     */ 
     fun getLeaderDisplayName(): String {
         return nation.getLeaderDisplayName().tr() +
             when {
+                !gameInfo.gameParameters.isOnlineMultiplayer ->
```
yairm210, line 376: ???

yairm210:
Looks good! Apart from the commented line, looks mergable!

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SomeTroglodyte: Quote: "getLeaderDisplayName adds AI/Human only in actual multiplayer games. As single player I don't need to know all the others are AI's? Was I right in thinking that?" - so I was already unsure - and rightfully so...

<details><summary>Warning: Explanation as pictorial!</summary>

(original)
image
(this branch)
image
(original)
image
</details>

<details><summary>So I came up with:</summary>

```kotlin
    fun getLeaderDisplayName(): String {
        /* Logic:
            We don't need to know AI/Human if:
                - It is a city state, recognizable via its nation icon?
                - We're single player: !isOnlineMultiplayer && Human.count==1
                - We're displaying the spectator
            We want to know AI/Human if:
                - We're online multiplayer
                - We're hotseat multiplayer
                - We're spectator????? -> those are counted as human / this doesn't know who's looking
         */
        val severalHumans = gameInfo.civilizations.count { it.playerType == PlayerType.Human } > 1
        val online = gameInfo.gameParameters.isOnlineMultiplayer
        return nation.getLeaderDisplayName().tr() +
            when {
                nation.name == Constants.spectator ->
                    ""                      // everybody knows the spectator is always human
                !online && !severalHumans ->
                    ""                      // offline single player will know everybody else is AI
                isCityState() ->
                    ""                      // City states are known to always be AI and are recognizable
                playerType == PlayerType.AI ->
                    " (" + "AI".tr() + ")"
                online ->
                    " (" + "Human".tr() + " - " + "Multiplayer".tr() + ")"
                else ->
                    " (" + "Human".tr() + " - " + "Hotseat".tr() + ")"
            }
    }
```
</details>

... But I think I'll leave the spectator and citystate cases out. Who wants <sup><sub>(can)</sub></sup> meet a ghost anyway?

SomeTroglodyte:
OK, all 3 meet-major-civ scenarios tested. Bless tiny maps! AI/Human now is only omitted if your're really alone, and online MP no longer incorrectly displays hotseat.

SomeTroglodyte:
Oh, worth mentioning: Nation leader display in civilopedia will one day be part of the 'markup' miniproject, and new game screen is a to-do I'm not sure I'll start on any time soon. As is the welcoming message (startIntroPart[12]). ...

SomeTroglodyte:
Sorry for the extra commit - that wasn't planned. (((bad boy - use squash not amend)))

SomeTroglodyte:
Actually - `startIntroPart1`:
image

...want it?

Issue #3568, opened 2021-02-01
Higher res buttons


skyace65:
The buttons look a bit low res on desktop. Would it be possible to add higher res images for them? I can make some in inkscape if that's necessary.

EDIT: Wanted to make sure I'm being clear. The images for the buttons themselves, not the icons in the buttons

lishaoxia1985:
@skyace65 it don't need to redraw the picture. we can set filter for picture. because we don't want to set filter for all picture, now we close this feature.

yairm210:
@lishaoxia1985 No, this is a different problem entirely.
We can totally set the filter for the buttons specifically in the ImageGetter.getRoundedEdgeTableBackground() function:

```kotlin
    fun getRoundedEdgeTableBackground(tintColor: Color? = null): NinePatchDrawable {
        val region = getDrawable("OtherIcons/buttonBackground").region
        region.texture.setFilter(Texture.TextureFilter.MipMapLinearLinear, Texture.TextureFilter.MipMapLinearLinear)
        val drawable = NinePatchDrawable(NinePatch(region, 25, 25, 0, 0)).apply {
            setPadding(5f, 15f, 5f, 15f)
        }
        if (tintColor == null) return drawable
        return drawable.tint(tintColor)
    }
```

The problem is that the current ninepatch takes from a 50x50 circle - namely, OtherIcons/buttonBackground. No matter what filter you put on it, in screen-size resolution, this will look bad.
The problem is that I can't find a way to create a ninepatch from a larger circle and resize (e.g. OtherIcons/Circle). It llooks llike we can choose between either 'scaling the image' or 'having a ninepatch'.

lishaoxia1985:
@yairm210 you can try Texture.TextureFilter.Linear, sometimes Texture​.​TextureFilter​.​MipMapLinearLinear has no effect. some days ago I find this problem.

yairm210:
I tried these, all the results are still grainier than I would like. When you start from a 50x50 circle, there's a limit to how nice that circle will look no matter what filters you put...

Texture.TextureFilter.MipMapLinearLinear, Texture.TextureFilter.MipMapLinearLinear:
![image](https://user-images.githubusercontent.com/8366208/112854049-1f639680-90b6-11eb-820a-2fddab17c261.png)

Texture.TextureFilter.Nearest, Texture.TextureFilter.Nearest:
[...]

PR #4080, opened 2021-06-08
Redraw skin Images, edit skin load method

lishaoxia1985:


yairm210:
O_o all the packed images are now nearest nearest? That sounds bad

What are we trying to achieve here?

lishaoxia1985:
in texturepack, setting filter to MipMap makes no sense. You can test that (change setting filter with linear)
You can view my answer in the issue #3568
By the way, now skin uses the filter linear.
if you want that some pictures use the filter linear, please remove them into folder "Skin"

SomeTroglodyte:
Well this looks better even to the untrained eye. :eyeglasses: 
### master:
master
### lishaoxia1985:
liaoxia

Just FYI - I'm running off a custom workdir, so I had to manually link ../Skin (chose this over the atlas pair, should make no diff in a debug session) and Skin.json into it - would that make a difference for jar packaging? apk builds fine, and runs fine on an AVM.

PR #4081, opened 2021-06-08
Implement production overflow

Thyrum:
As discussed on yairm210#4058, production (or hammer) overflow in Civ 5 is a thing and goes according to rules described on https://forums.civfanatics.com/threads/hammer-overflow.419352/ . This has been added to the game.

Currently there are no visual indicators that there is an overflow present. I don't know if and where these should be added. I think (but don't know) Civ 5 doesn't communicate this in any way either so maybe it is good as is. Suggestions are welcome.

I also don't know how an older version would react to a new game with the field `productionOverflow`. In general, how do older versions react to units/data fields they don't think should exist?

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -49,6 +49,7 @@ class CityConstructions {
         }
     var currentConstructionIsUserSet = false
     var constructionQueue = mutableListOf<String>()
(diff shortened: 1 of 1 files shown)


yairm210:
Clean, simple, and elegant
Excellent work!

yairm210:
Fields not present are given the default value - in this case, 0.
Fields present but not in class are ignored.

PR #4082, opened 2021-06-08
Main menu key bindings

SomeTroglodyte:
No key indicators or instructions elsewhere yet. Watch this please:
https://user-images.githubusercontent.com/63000004/121243217-c1b1ae80-c89d-11eb-9c52-3c003f654f58.mp4

I thought Gdx Tooltip's hopeless but like that they're not bad. Worth it - or should I add some instructions to the 'Keyboard' tutorial?


yairm210:
Cute!!! I like the tooltips!
Not everyone reads the tutorials, I think the tooltips will reach a much higher audience percentage ;)

yairm210:
This is still mergable as-is, though
The tooltips can come later

Issue #4068, opened 2021-06-07
Some of social policies are different than in vanilla civ5


ravignir:
https://civilization.fandom.com/wiki/Social_policies_(Civ5)

yairm210:
I can't try and guess from the lists
Can you give specific examples?

ravignir:
Tradition:
Oligarchy | Garrisoned units cost no maintenance and cities with a garrison gain **+100%** Ranged Combat Strength. 

Honor:
Warrior Code | +15%  Production when training Melee units **and a Great General appears outside the  Capital.**
Discipline | **+10%**  combat strength for melee units which have another military unit in an adjacent tile.
Professional Army | Gold cost of upgrading Military Units reduced by 33% **and+1 Happiness from every defensive building (Walls, Castle, Arsenal, Military Base).**

Commerce - this tree is almost completely different:
Trade Unions | Maintenance paid on Roads and Railroads reduced by 33%. **Harbors and Seaports gain +1  Gold.**
Naval Tradition | +1  Movement for Naval units and +1 sight range for naval combat units.
Mercantilism | Requires Trade Unions.
Merchant Navy | +3  Production in all coastal Cities. Requires Naval Tradition
Commerce Complete | +1 Gold Gold from every Specialist 

Rationalism: | Science gained from research agreements by 50% 
Scientific Revolution | Gives 2 free Technologies.

Freedom:
Universal Suffrage | Combat Strength of Cities increased by 33%.
Free Speech | 8 units are maintenance free.

xlenstra:
Which version of the game should the policies follow? The vanilla or the Gods&Kings version?

xlenstra:
> Which version of the game should the policies follow? The vanilla or the Gods&Kings version?

If someone (@yairm210?) can answer this question I'll look into it later this week

yairm210:
I think this might be a good time to differentiate the Vanilla and G&K 'base rulesets'.
I'm wondering what the best way to do this would be, but in general I'm thinking a mod.
That would clear up all confusing conflicts between versions, AND give us a good starting point to evaluate future mods.

I think 'vanilla Unciv' should be G&K, since it's much more mature that Vanilla Civ V, but also without all the very weird stuff they added in BNW.
We don't have religion yet, but I think that G&K is still the goalpost.

yairm210:
I can't remember wat was Faith-based in G&K policies, but that's obviously not going to be relevant now. nything else can be changed to suit G&K.

[...]

PR #4084, opened 2021-06-09
Updated policies to G&K

xlenstra:
WHEN COMPLETE, this PR will have updated all the policies to G&K except for the piety branch, as that requires the implementation of faith. Also, it will have implemented the patronage and order branches. I'll check all existing policies to make them more modular, which will probably break a lot of translations.

Currently, only tradition, liberty and honour have been updated.

Question: Is it better to already create this draft PR so it is visible that I'm working on this, or should I just create a PR when I'm done?

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Policies.json
+++ b/android/assets/jsons/Civ V - Vanilla/Policies.json
@@ -12,13 +12,13 @@
 			},
 			{
 				"name": "Legalism",
-				"uniques":["Immediately creates a cheapest available cultural building in each of your first 4 cities for free"],
+				"uniques":["Immediately creates the cheapest available cultural building in each of your first [4] cities for free"],
 				"row": 1,
 				"column": 3
 			},
 			{
 				"name": "Oligarchy",
-				"uniques": ["Units in cities cost no Maintenance", "+50% attacking strength for cities with garrisoned units"],
+				"uniques": ["Units in cities cost no Maintenance", "+[50]% attacking strength for cities with garrisoned units"],
 				"row": 1,
 				"column": 5
 			},
@@ -38,7 +38,7 @@
 			},
 			{
 				"name": "Tradition Complete",
-				"uniques": ["+[15]% growth [in all cities]","[+2 Food] [in all cities]"]
+				"uniques": ["+[15]% growth [in all cities]","Immediately creates a free [aqueduct] in each of your first [4] cities for free"]
 			}
 		]
 	},
@@ -48,21 +48,21 @@
 		"uniques": ["[+1 Culture] [in all cities]"],
 		"policies": [
 			{
-				"name": "Collective Rule",
-				"uniques": ["+[50]% Production when constructing [Settler] units [in capital]", "Free [Settler] appears"],
+				"name": "Republic",
+				"uniques": ["[+1 Production] [in all cities]", "+[5]% Production when constructing [Buildings]"],
 				"row": 1,
 				"column": 1
 			},
 			{
 				"name": "Citizenship",
-				"uniques": ["Tile improvement speed +25%", "Free [Worker] appears"],
+				"uniques": ["Tile improvement speed +[25]%", "Free [Worker] appears"],
 				"row": 1,
 				"column": 4
 			},
 			{
-				"name": "Republic",
-				"uniques": ["[+1 Production] [in all cities]", "+[5]% Production when constructing [Buildings]"],
-				"requires": ["Collective Rule"],
+				"name": "Collective Rule",
+				"uniques": ["+[50]% Production when constructing [Settler] units [in capital]", "Free [Settler] appears"],
+				"requires": ["Republic"],
 				"row": 2,
 				"column": 1
 			},
@@ -76,6 +76,7 @@
 			{
 				"name": "Meritocracy",
 				"uniques": ["[+1 Happiness] [in all cities connected to capital]", "Unhappiness from population decreased by [5]%"],
+				// Should only be unhpainness from population in non-occupied cities
 				"requires": ["Citizenship"],
 				"row": 2,
 				"column": 5
@@ -89,19 +90,19 @@
 	{
 		"name": "Honor",
 		"era": "Ancient era",
-		"uniques": ["+25% bonus vs Barbarians", "Earn [100]% of killed [Barbarian] unit's [Strength] as [Culture]",
+		"uniques": ["+[33]% bonus vs Barbarians", "Earn [100]% of killed [Barbarian] unit's [Strength] as [Culture]",
 			"Notified of new Barbarian encampments"],
 		"policies": [
 			{
 				"name": "Warrior Code",
-				"uniques":["+[20]% Production when constructing [Melee] units [in all cities]"],
+				"uniques":["+[15]% Production when constructing [Melee] units [in all cities]", "Free Great General appears"],
 				"row": 1,
 				"column": 2
 			},
 			{
 				"name": "Discipline",
-				"uniques":["+15% combat strength for melee units which have another military unit in an adjacent tile"],
-				"row": 1,
+				"uniques":["+[15]% combat strength for [melee] units which have another military unit in an adjacent tile"],
+				"row": 1, 
 				"column": 4
 			},
 			{
@@ -120,7 +121,11 @@
 			},
 			{
 				"name": "Professional Army",
-				"uniques": ["Gold cost of upgrading military units reduced by 33%"],
+				"uniques": ["Gold cost of upgrading military units reduced by 33%", "[+1 Happiness] from every [Walls]", 
+					"[+1 Happiness] from every [Castle]", 
+//					"[+1 Happiness] from every [Arsenal]", "[+1 Happiness] from every [Military Base]"
+// 					Should be added when these buildings are implemented					
+				],
 				"requires": ["Military Caste"],
 				"row": 3,
 				"column": 4
@@ -131,6 +136,7 @@
 			}
 		]
 	},{
+		// I have not updated these to Gods & Kings as faith was not implemented yet when I updated the rest.
 		"name": "Piety",
 		"era": "Classical era",
 		"uniques": ["+[15]% Production when constructing [Culture] buildings"],
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -75,12 +75,16 @@ object BattleDamage {
                 }
             }
 
-            if (civInfo.hasUnique("+15% combat strength for melee units which have another military unit in an adjacent tile")
-                && combatant.isMelee()
-                && combatant.getTile().neighbors.flatMap { it.getUnits() }
+            if (combatant.getTile().neighbors.flatMap { it.getUnits() }
                     .any { it.civInfo == civInfo && !it.type.isCivilian() && !it.type.isAirUnit() }
-            )
-                modifiers["Discipline"] = 15
+            ) {
+                var adjacentUnitBonus = 0;
+                for (unique in civInfo.getMatchingUniques("+[] combat strength for [] units which have another military unit in an adjacent tile")) {
+                    if ((combatant.isMelee() && unique.params[1] == "melee") || (combatant.isRanged() && unique.params[1] == "ranged"))
+                        adjacentUnitBonus += unique.params[0].toInt()
+                }
+                modifiers["Adjacent unit"] = adjacentUnitBonus
+            }
 
             val civResources = civInfo.getCivResourcesByName()
             for (resource in combatant.unit.baseUnit.getResourceRequirements().keys)
@@ -109,8 +113,11 @@ object BattleDamage {
         if (enemy.getCivInfo().isBarbarian()) {
             modifiers["Difficulty"] =
                 (civInfo.gameInfo.getDifficulty().barbarianBonus * 100).toInt()
-            if (civInfo.hasUnique("+25% bonus vs Barbarians"))
-                modifiers["vs Barbarians"] = 25
+            var barbarianBonus = 0;
+            for (unique in civInfo.getMatchingUniques("+[]% bonus vs Barbarians")) {
+                barbarianBonus += unique.params[0].toInt()
+            }
+            modifiers["vs Barbarians"] = barbarianBonus 
         }
 
         return modifiers
@@ -166,11 +173,15 @@ object BattleDamage {
             )
                 modifiers["Statue of Zeus"] = 15
         } else if (attacker is CityCombatant) {
-            if (attacker.getCivInfo()
-                    .hasUnique("+50% attacking strength for cities with garrisoned units")
-                && attacker.city.getCenterTile().militaryUnit != null
-            )
-                modifiers["Oligarchy"] = 50
+            if (attacker.city.getCenterTile().militaryUnit != null) {
+                var OlichargyBonus = 0;
+                for (unique in attacker.getCivInfo()
+                    .getMatchingUniques("+[]% attacking strength for cities with garrisoned units")) {
+                    OlichargyBonus += unique.params[0].toInt()
+                }
+                if (OlichargyBonus != 0)
+                    modifiers["Garrisoned unit"] = OlichargyBonus
+            }
         }
 
         return modifiers
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -92,14 +92,17 @@ class CityConstructions {
      * @return Maintenance cost of all built buildings
      */
     fun getMaintenanceCosts(): Int {
-        var maintenanceCost = getBuiltBuildings().sumBy { it.maintenance }
-        val policyManager = cityInfo.civInfo.policies
-        if (cityInfo.id in policyManager.legalismState) {
-            val buildingName = policyManager.legalismState[cityInfo.id]
-            maintenanceCost -= cityInfo.getRuleset().buildings[buildingName]!!.maintenance
+        return getBuiltBuildings().sumBy { it.maintenance } - getMaintenanceReductionForFreeBuildings()
+    }
+    
+    /** Calculates the amount of maintenance reduced by buildings granted by policies */
+    fun getMaintenanceReductionForFreeBuildings(): Int {
+        var maintenanceReduction = 0
+        var freeBuildings = cityInfo.civInfo.policies.getListOfFreeBuildings(cityInfo.id)
+        for (building in freeBuildings) {
+            maintenanceReduction += cityInfo.getRuleset().buildings[building]!!.maintenance
         }
-
-        return maintenanceCost
+        return maintenanceReduction
     }
 
     /**
--- a/core/src/com/unciv/logic/civilization/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/PolicyManager.kt
@@ -25,8 +25,8 @@ class PolicyManager {
     var shouldOpenPolicyPicker = false
         get() = field && canAdoptPolicy()
     
-    var cultureBuildingsAdded = HashMap<String, String>()
-    var specificBuildingsAdded = HashMap<String, MutableSet<String>>()
+    var cultureBuildingsAdded = HashMap<String, String>() // Maps cities to buildings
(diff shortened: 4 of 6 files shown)


android/assets/jsons/Civ V - Vanilla/Policies.json
```diff
@@ -38,7 +38,7 @@
 			},
 			{
 				"name": "Tradition Complete",
-				"uniques": ["+[15]% growth [in all cities]","[+2 Food] [in all cities]"]
+				"uniques": ["+[15]% growth [in all cities]","Immediately creates a free [aqueduct] in each of your first [4] cities for free"]
```
SomeTroglodyte, line 41: double 'free'

ravignir: and "aqueduct" should start with a capital letter.

SomeTroglodyte (approved):
I think I'll need to adblock that review button if I keep hitting it unintentionally, sorry

android/assets/jsons/Civ V - Vanilla/Policies.json
```diff
@@ -76,6 +76,7 @@
 			{
 				"name": "Meritocracy",
 				"uniques": ["[+1 Happiness] [in all cities connected to capital]", "Unhappiness from population decreased by [5]%"],
+				// Should only be unhpainness from population in non-occupied cities
```
SomeTroglodyte, line 79: µBlock filter `##form.js-inline-comment-form  button.btn-primary` works... Anyway, 'unhappiness', no pain. But - the comments in json are illegal json issue is still kind of hanging and Android Studio will complain, I'd not create new json comments unless painfully necessary.

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -166,11 +173,15 @@ object BattleDamage {
             )
                 modifiers["Statue of Zeus"] = 15
         } else if (attacker is CityCombatant) {
-            if (attacker.getCivInfo()
-                    .hasUnique("+50% attacking strength for cities with garrisoned units")
-                && attacker.city.getCenterTile().militaryUnit != null
-            )
-                modifiers["Oligarchy"] = 50
+            if (attacker.city.getCenterTile().militaryUnit != null) {
+                var OlichargyBonus = 0;
```
SomeTroglodyte, line 177: AS will complain and say: var names please start with lower case - AS has a 'Problems' view and I tend to keep it empty if possible - makes the tool more useful when it's suddenly no longer empty.

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -24,7 +24,9 @@ class PolicyManager {
     var numberOfAdoptedPolicies = 0
     var shouldOpenPolicyPicker = false
         get() = field && canAdoptPolicy()
-    var legalismState = HashMap<String, String>()
```
SomeTroglodyte, line 27: This is serialized in existing save games - Typically we have conversion code somewhere in game load or settransients for a few versions - Just mentioning in case you didn't think of this

xlenstra: Good point, didn't think about that

SomeTroglodyte:
General Impression: This boat is sailing a good course...

> Question: Is it better to already create ...
As you can see you get peer input earlier. I seem to be often doing it the other way round, which has its own advantages: Try to follow up an idea until I'm pleased with the result, then stashing it and doing a full diff with the (by now way ahead) master, merging it in - advantages: partitioning into small PR's is easier, you often find overlooked issues with the merge itself, and an implicit rebase making things easier for YM? 

Depends. 

For this, going public early is good for a simple reason - G&K and policy. It's clear the intention was there to produce a _pure_ Unciv following vanilla civ5 rules first, then maybe go for the expansion features. It's also clear that intention has already been muddled - some non-vanilla features were maybe simply too much fun to play. I still think in the long run we should have a vanilla rules folder moved back to a pure form and new folders for the G&K/BNW rules. Let there be Rule Purity discussion... 😸

SomeTroglodyte:
> break a lot of translations

Avoid if you can, but _do_ if it's a long term advantage? <sup><sub>Reminds me I wanted to offer a patch replacing all xlations depending on leading or trailing blanks...</sub></sup>

ravignir:
You would have to write a separate logic for vanilla because expansions change many mechanics for the better.

android/assets/jsons/Civ V - Vanilla/Policies.json
xlenstra: While I understand your reasoning, I don't particularly mind adding comments myself. In the cases where I add it, I do it as an indication that there is still work to be done, and if completed, the comment should of course be removed. This way, in the 'finished product', the comments would be gone, but while developing they are useful in this sense and since our parser is able to handle them, I don't feel like there is a big downside to them.

xlenstra:
> For this, going public early is good for a simple reason - G&K and policy. It's clear the intention was there to produce a _pure_ Unciv following vanilla civ5 rules first, then maybe go for the expansion features. It's also clear that intention has already been muddled - some non-vanilla features were maybe simply too much fun to play. I still think in the long run we should have a vanilla rules folder moved back to a pure form and new folders for the G&K/BNW rules. Let there be Rule Purity discussion... smile_cat

Before starting this branch, I already had a discussion in #4068 whether the policies should be the vanilla policies or the G&K policies. The conclusion seemed to be that the 'vanilla' version of Unciv should by G&K, with it (somewhen?) being possible to switch to either vanilla civ or BNW, either through a mod (changing techs/policies/units/etc. shouldn't be to difficult, although changes in mechanics might prove more difficult) or an option

core/src/com/unciv/logic/battle/BattleDamage.kt
yairm210: In this case I agree with the tool, since this doesn't affect readability it's best to stick with the official kotlin style

SomeTroglodyte:
OK, 4068 is good news, I had missed that.

Idea: You're dealing a bit with [cityFilter](https://github.com/yairm210/Unciv/wiki/uniques#cityfilter) uses - now the tileFilter implementation can be found by Ctrl-Shitf-F "tileFilter", and links to its documentation. Maybe while you're at it make sure the other filters are consistent? (🤦 I should have renamed `TileInfo.matchesUniqueFilter` when I had the chance to make at least the names consistent - `matchesFilter` already has majority...)

yairm210:
Please split this into prs for every like 5 policies.
The way it is this will be a 200-comment pr,
I don't have the required headRAM to deal with that.
Also, if we're changing uniques, let's include unit types as well.

xlenstra:
Sounds reasonable, I'll open a separate PR for every policy branch  I've updated so far later today.
My plan was to first update the policy uniques, and later look at the other ones (buildings/units/promotions/etc.), but that was going to be a separate PR for each JSON file.

SomeTroglodyte:
Don't forget to include new png's and atlases when done... And that those three branches are supposed to be mutually exclusive.  (me saying this proves I like the direction this is going)

The UI suffers from slight geometry issues due to the break being hardcoded and separator getting its colspan from current column count.
<details><summary>Here's an offering of code as patch for fiddled-with UI - fleece whatever you wish from it</summary>

```patch
Index: core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
IDEA additional info:
<+>UTF-8
--- core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt	(revision c63cbd715b5c282584ddf678e6d089362eacdcb4)
+++ core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt	(date 1623344692399)
@@ -46,13 +46,20 @@
         if (!UncivGame.Current.worldScreen.canChangeState)
             rightSideButton.disable()
 
-        topTable.row().pad(30f)
+        topTable.row().pad(10f)
 
-        for (branch in viewingCiv.gameInfo.ruleSet.policyBranches.values) {
-            if (branch.name == "Commerce") topTable.addSeparator()
+        val branches = viewingCiv.gameInfo.ruleSet.policyBranches
+        val rowChangeIndex = (branches.size + 1) / 2
+        var wrapper = Table()
+        for ( (index, branch) in branches.values.withIndex()) {
+            if (index == rowChangeIndex) {
+                topTable.add(wrapper)
+                topTable.addSeparator()
+                wrapper = Table()
+            }
             val branchGroup = Table()
             branchGroup.row().pad(20f)
-            branchGroup.add(getPolicyButton(branch, false)).row()
+            branchGroup.add(getPolicyButton(branch, false)).minWidth(160f).row()
 
             var currentRow = 1
             var currentColumn = 1
@@ -60,7 +67,7 @@
             for (policy in branch.policies) {
                 if (policy.name.endsWith("Complete")) continue
                 if (policy.row > currentRow) {
-                    branchTable.row()
+                    branchTable.row().pad(2.5f)
                     currentRow++
                     currentColumn = 1
                 }
@@ -73,10 +80,11 @@
             branchTable.pack()
             branchGroup.add(branchTable).height(150f).row()
 
-            branchGroup.add(getPolicyButton(branch.policies.last(), false)) // finisher
+            branchGroup.add(getPolicyButton(branch.policies.last(), false)).pad(15f,0f) // finisher
 
-            topTable.add(branchGroup)
+            wrapper.add(branchGroup).pad(0f, 10f)
         }
+        topTable.add(wrapper)
         topTable.pack()
     }
 
@@ -129,4 +137,4 @@
         return policyButton
     }
 
-}
\ No newline at end of file
+}
```
</details>

(Branches no longer align vertically which saves space, the individual buttons no longer glued together in places, break depends on branch count not branch name, some redistribution, and the narrowest branch start buttons equalized a bit. Should be kinder to total conversion mods, too - or maybe coding the break trigger into json?)

<details><summary>And here's before/after screenshots</summary>

##### Before
image
##### After
image
</details>
... regrettably this would still need scrolling below "1200x700"

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -75,12 +75,16 @@ object BattleDamage {
...
-                && combatant.isMelee()
-                && combatant.getTile().neighbors.flatMap { it.getUnits() }
+            if (combatant.getTile().neighbors.flatMap { it.getUnits() }
                     .any { it.civInfo == civInfo && !it.type.isCivilian() && !it.type.isAirUnit() }
-            )
-                modifiers["Discipline"] = 15
+            ) {
+                var adjacentUnitBonus = 0;
+                for (unique in civInfo.getMatchingUniques("+[] combat strength for [] units which have another military unit in an adjacent tile")) {
+                    if ((combatant.isMelee() && unique.params[1] == "melee") || (combatant.isRanged() && unique.params[1] == "ranged"))
```
SomeTroglodyte, line 83: `combatant.unit.matchesFilter(unique.params[1])`? a) the [] notation suggests the standard unitFilter is supported, so maybe we should support it; b) shorter to write    ...?

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -103,14 +107,23 @@ object BattleDamage {
                     .isCityState() && civInfo.hasUnique("+30% Strength when fighting City-State units and cities")
             )
                 modifiers["vs [City-States]"] = 30
+            
+            if (combatant.getTile().isFriendlyTerritory(combatant.getCivInfo())) {
+                val friendlyTerritoryModifier = civInfo.getMatchingUniques("+[]% Strength while fighting in friendly territory")
+                    .fold(0) { sum, it -> sum + it.params[0].toInt() }
```
SomeTroglodyte, line 113: Isn't that exactly the same as `.sumBy { it.params[0].toInt() }`? And if it's not I want to learn the difference!

ravignir:
Rationalism and Piety are also mutually exclusive

SomeTroglodyte:
True, as your nice collage shows... Maybe we can use that to get UI space usage down if necessary...

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -71,6 +71,8 @@ class CityConstructions {
 
     fun getBasicCultureBuildings() = cityInfo.getRuleset().buildings.values
             .asSequence().filter { it.culture > 0f && !it.isWonder && !it.isNationalWonder && it.replaces == null }
+    
+    fun getCultureBuildings() = getBasicCultureBuildings().map { cityInfo.civInfo.getEquivalentBuilding(it.name) }
```
SomeTroglodyte, line 75: This looks suspicious - `getBasicCultureBuildings` loops all buildings and picks non-uniques, then we go and find the corresponding unique - which again loops all buildings (deferred and early exit nonwithstanding). Also, we translate an object to its String name then back to its Object representation. Lastly, the knowledge only base buildings get into the loop is lost and little later another `getBaseBuilding` is done... Probably doesn't matter much in the big picture. Would matter none if we'd cache the Culture buildings in some lazy on CivilizationInfo or Nation?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -92,14 +94,17 @@ class CityConstructions {
      * @return Maintenance cost of all built buildings
      */
     fun getMaintenanceCosts(): Int {
-        var maintenanceCost = getBuiltBuildings().sumBy { it.maintenance }
```
SomeTroglodyte, line 95: Clear approach, but my nose tingles... getListOfFreeBuildings(cityInfo.id) returns a Set<String> - MutableSet interface, acutally the LinkedHashSet implementation if I'm not mistaken. Lookups should be _fast_. So - cache the set reference, loop buildings, sum if !in set should be a few percent faster.

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -178,15 +178,17 @@ class CityStats {
...
         if (civInfo.hasUnique("Unhappiness from number of Cities doubled"))
-            unhappinessFromCity *= 2f//doubled for the Indian
+            unhappinessFromCity *= 2f //doubled for the Indian
 
         newHappinessList["Cities"] = unhappinessFromCity * unhappinessModifier
 
         var unhappinessFromCitizens = cityInfo.population.population.toFloat()
-        if (civInfo.hasUnique("Specialists produce half normal unhappiness"))
-            unhappinessFromCitizens -= cityInfo.population.getNumberOfSpecialists() * 0.5f
+        var SpecialistUnhappinessReductionModifier = civInfo.getMatchingUniques("Specialists only produce []% of normal unhappiness")
```
SomeTroglodyte, line 188: Also triggers the "LocalVariableName" inspection. And we get 1% reduction for free (the fold initializer - I wouldn't be surprised there's a reason but that would merit a comment). And - I dislike implicit casts (on SQL Server it's definitely worth it to pay attention to those) - anyway, the `1` two lines down (and in lines you didn't touch) is implicitly cast to Float and I prefer having those in the open - just a personal quirk, I won't mention it again.

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -31,8 +31,9 @@ class TileImprovement : NamedStats() {
         // todo UNIFY THESE
```
SomeTroglodyte, line 31: While you're at it, why not take care of this tutu? It's only used by the Pyramids. Or maybe not - you're not yet touching Buildings.json.

core/src/com/unciv/models/stats/Stats.kt
```diff
@@ -39,11 +39,20 @@ open class Stats() {
 
 
     fun add(stat: Stat, value: Float): Stats {
+        // This looks super inefficient to me, for changing a single stat we assign EVERY stat TWICE
```
SomeTroglodyte, line 42: What  bugs me more is that Stats is an open class with subclasses adding _tons_ of stuff so in those the `isEmpty` or `toString` implementations from Stats are at times really surprising. And surprises lead to bugs.

Anyway, too late, finding an alternative might be a bit tricky - if you want to keep the way these classes are serialized and deserialized intact - and I'd _really_ want to keep _that_ as it is.

What we could _maybe_ do: Ditch the hashMap detour. Leave data members as they are. Add a property to the Stat Enum so you can do a Stat.addTo(target: Stats, value: Float) -> at this point there's actually seven or eight different delegates being called so the 'mapping' comes free...  

Actually,
```kotlin
enum class Stat (val addTo: (target: Stats, value: Float) -> Unit){
    Production ({ target: Stats, value: Float -> target.production += value }),
    Food ({ target: Stats, value: Float -> target.food += value }),
    Gold ({ target: Stats, value: Float -> target.gold += value }),
    Science ({ target: Stats, value: Float -> target.science += value }),
    Culture ({ target: Stats, value: Float -> target.culture += value }),
    Happiness ({ target: Stats, value: Float -> target.happiness += value }),
    Faith  ({ target: Stats, value: Float -> target.faith += value }),
}

// and elsewhere in class Stats
    fun add(stat: Stat, value: Float): Stats {
        stat.addTo(this, value)
        return this
    }
```
compiles just fine, but _testing_ this thoroughly? **S C A R Y** !!!.

SomeTroglodyte:
OK, last comment: That barrage is in no way meant to discourage, it only proves I find this intriguing.

The biggest problem I actually see is that replacing uniques or other json structures/conventions/whatever should (afaik) usually go -> add new way, leave in but deprecate old way -> let modders adapt for a few versions -> remove deprecated stuff...
Of course adding the compatibility code at the end after testing is one valid approach - seeing all candidates in a diff isn't too hard - I'm mentioning that just to make sure you're aware. 

So, keep up the good work, and don't feel obligated to answer any of those.

core/src/com/unciv/models/stats/Stats.kt
yairm210: ...no.
If we're doing ANYTHING in that direction, specifying each addition in Stats.add() would be much more readable and much simpler.

yairm210:
The uniques replacement is an ongoing project, and yes, we're getting rid of old outdated uniques all the time. Mods that don't keep up, well, that's life. Not a problem.

core/src/com/unciv/logic/city/CityConstructions.kt
xlenstra: This could be changed, but I do not feel like the speedup would be massive. Might look into it nonetheless. For now, the new version over at #4106 has cached the map of maintenance-free buildings for each cities, which should also improve performance.

xlenstra: I've changed it after all

SomeTroglodyte:
>...no.

Actually, I've welcomed that comment at first but then explored the idea a bit further. I tend to disagree, but will shelve my (working!) branch for now. I might pull it out and do some probing later, but here's my reasoning:

- Current code uses a fresh hashmap built from a list of pairs each time it wants indexed(by the Stat Enum) access to a specific 'stat' - or to enumerate over them. The pairs, the list and the map are left to the GC quickly. The construction of the map may not cost too much time, but the memory allocation and GC load may be another matter - perhaps.
- If you say 'specifying each addition in Stats.add()' - to me that sounds like implementing indexed access using a when(stat) {} construct... No needless memory allocation, yes, but - honestly - clunkier than my idea.
- Stat would need more of those delegates to really work, but that would just be - admittedly a lot - of boilerplate text. Under the hood those would just be a bunch of very lightweight delegates - all of them static and their references known, all of them just passing the work on to one or two known members of the Stats class (my draft has get, set and add, doing the rarer mul via get and set, so get has 1 field accessor call, set too, and add has two). Zero memory allocation, and everything array-indexed, no conditionals...
- toHashSet is used something like 15 times outside Stats.kt - and after adding a simple sequence generator as replacement for the looping usecase of toHashMap every single one of them gets simpler and easier to read, and most get more efficient... And Stats.kt got about 20 lines shorter.

Let's leave this to stew for a while.

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
xlenstra: Fixed in #4115.

core/src/com/unciv/logic/battle/BattleDamage.kt
xlenstra: It was probably me trying to do something simple in a complicated way :)

core/src/com/unciv/logic/battle/BattleDamage.kt
xlenstra: Fixed in #4118

PR #4086, opened 2021-06-09
Fix turnsLeft returns 0 if 0 < workLeft <= overflow

Thyrum:


SomeTroglodyte:
Good, then I can drop the attempt to get a screenshot showing negative 'turns left'... :+1:

PR #4090, opened 2021-06-09
For the F-Droid missing libgdx.so problem

### LINKED ISSUE #3946: Crash on arm64 (Samsung Tab S2) from F-droid - missing libgdx.so
Version: Fresh from F-Droid moments ago = 3.14.8
Symptom: None - doesn't launch, screen doesn't even flash
Platform: Samsung Tab S2 (SM-T813 = gts210vewifi) running Lineage 16.0 (should correspond to Pie, I keep forgetting), arm64
Logcat: java.lang.UnsatisfiedLinkError yada yada zip file "/data/app/com.unciv.app-garble/base.apk" yada yada ... /lib/arm64, /system/lib64, /system/vendor/lib64 ... couldn't find "libgdx.so"

Sounds like [this here](https://stackoverflow.com/questions/54414066/updating-android-app-with-libgdx-for-64-bit-devices)...

SomeTroglodyte:
... does the trick according to the procedure by @relan, all kudos to them.
Might resolve #3946

_Please_ evaluate carefully, I can't claim I really _understand_ gradle. The line after `tasks.whenTaskAdded` is the only really relevant one.

And relan suggested modifying the fdroid recipe instead... While this improves pure local builds in marginal cases... Can't decide what's best.

Note:- The regex thing is an early change I left in - output difference is only that this no longer puts an empty armabi folder inside the apk, and it might be flexible enough to handle architectures added in the future transparently?

### DIFF
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -77,21 +77,11 @@ task("copyAndroidNatives") {
     val natives: Configuration by configurations
 
     doFirst {
-        file("libs/armeabi/").mkdirs()
-        file("libs/armeabi-v7a/").mkdirs()
-        file("libs/arm64-v8a/").mkdirs()
-        file("libs/x86_64/").mkdirs()
-        file("libs/x86/").mkdirs()
+        val rx = Regex(""".*natives-([^.]+)\.jar$""")
         natives.forEach { jar ->
-            val outputDir: File? = when {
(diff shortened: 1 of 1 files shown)


yairm210:
Let's try it!

SomeTroglodyte:
For good measure I added some blahblah to their issue over [here](https://gitlab.com/fdroid/fdroiddata/-/issues/2385#note_594038193)...

PR #4091, opened 2021-06-09
Implement dormant Nation.startIntroPart1+2

SomeTroglodyte:
Mentioned in #4073 - as tentative offer. We could just as well go wihtout, but then maybe we should delete those lines from the jsons.

If welcome: How to go about errors in the existing texts, e.g. Greece: "ragedy"? Fix English in Nations.json and wait for our translators or fix all of them together with the "key"?

image


yairm210:
Spelling corrections welcome :)

yairm210:
Add translation for "let's begin"

SomeTroglodyte:
I can remember looking for the "Farewell" to place the template near the first time around - but then moving this from box to box as a patch wasn't a good idea anyway.

PR #4094, opened 2021-06-10
Nations spellcheck - mainly for start-intro popup

SomeTroglodyte:
OK - this is the result :astonished: 

### Noteable translation notes:

All startIntros quickly proofread, not only spellchecked - but I did not pay much attention to grammar or punctuation - as these are copies from a certain wiki except for the typos. No need to check translations - none present.
Whole file spellchecked - those must-fix places also needed touching all translation files, so I tried to ignore what could be somehow ignored.
e.g. the spanish city names have many errors in the json - especially "Eiche" is annoyingly wrong (it's Elche), and with some names sporting accents, one even wrong, there should be no reason to omit so many - but I refrained from touching it. Sorry.
Adjectives not checked - Inconsistent but unused and untranslated.

### The german city names
There were two charset errors in there (Mצnchengladbach,Gצttingen) otherwise I wouldn't have felt _forced_ to touch that...

(my complete log - ignore if you want)
- Munich, Cologne stayed in these forms - clearly established 'nicknames' outside germany.
- English 'Dusseldorf' stayed as is even though the city is named Düsseldorf and always was. Two reasons: There *are* those "New York, Tokyo, Paris, Dusseldorf" fashion shops (duckduck suggests more have it right), and as a joke: Dussel means moron - as someone with Friends there I'm allowed to perpetrate that.
- All other city names are now 'real' - I'm reasonably sure that said nickname reasons do not apply.

- Language files where german city names were completely untranslated -> I (mostly) deleted the block to ensure the generator fills in the new key as 'requires translation' and there remains no trace of the obsolete keys.

- Japanese, Chinese, etc, even Polish and French:
I adapted only the left side faithfully to the new json. Polish was a close decision - I trusted the original xlator that Poles want to see no Umlauts, so Dusseldorf stayed Dusseldorf etc.

- Spanish, Swedish, even Indonesian:
These got the german block verbatim - Sp/Swe originally followed actual german names better than the english template, cultural exchange close. Indonesian actually fixed so many of the original errors in the english that I thought a copy appropriate, too.

- Portuguese, Dutch:
These got the german block verbatim even though they originally had no translations, but cultural exchange is close so they might know most real names.

- Turkish - the original xlator obviously tried to follow actual german names more closely than the template, so Turkish got nearly all 'German' german city names verbatim, too. The following lines got replaced that weren't clearly 100% better:
Mצnchengladbach = Münchengladbach
Kiel = Kirl
Exception: Munich = Münih stayed, confirmed by g00 translate.

- Italian - close call, I did my best. Looked liked the xlator wanted Umlaut-free, so Rostock and Ludwigshafen got fixed, Mצnchengladbach fixed to Umlaut-free, most stayed...
- Hungarian - closer to real names than original english was, so I left obvious deviations (Bréma,Lipcse,Dresda) and fixed the rest.
- Suomi - closer to real names than original english was, so I left obvious deviations (Berliini,Hampuri) and fixed the rest.
- Czech - had many bad english keys fixed, also many nicknames - I fixed the right side only where obvious.
- Bulgarian - no translations, went into the 'delete' box. I know a few german speaking Bulgarians but am not sure about the average closeness, so I didn't apply the same decision as for Portuguese.
- Brazilian - tough decision. Cultural closeness, give them same as Portuguese? There's Blumenau, but... Was Umlaut-free so I left it that way, fixed the rest, left strange entries like 'Leverkussen' or 'Wurzuburg' - may be established nicknames, who am I to know, haven't been there in decades.


android/assets/jsons/translations/English.properties
```diff
@@ -3636,116 +3636,6 @@ Furor Teutonicus =
...
- # Requires translation!
-Rostok = 
- # Requires translation!
-Chemnitz = 
- # Requires translation!
-Braunschweig = 
- # Requires translation!
-Halle = 
- # Requires translation!
-Mצnchengladbach =
```
Mape6, line 3690: Still with strange Umlaut?

android/assets/jsons/translations/English.properties
```diff
@@ -3636,116 +3636,6 @@ Furor Teutonicus =
...
- # Requires translation!
-Darmstadt = 
- # Requires translation!
-Herne = 
- # Requires translation!
-Wurzburg = 
- # Requires translation!
-Recklinghausen = 
- # Requires translation!
-Gצttingen =
```
Mape6, line 3740: Still with strange Umlaut?

android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -340,10 +340,10 @@
 		"uniques": ["67% chance to earn 25 Gold and recruit a Barbarian unit from a conquered encampment", "-[25]% [Land] unit maintenance costs"],
 		"cities": ["Berlin","Hamburg","Munich","Cologne","Frankfurt","Essen","Dortmund","Stuttgart","Dusseldorf","Bremen",
 			"Hannover","Duisburg","Leipzig","Dresden","Bonn","Bochum","Bielefeld","Karlsruhe","Gelsenkirchen","Wiesbaden",
-			"Munster","Rostok","Chemnitz","Braunschweig","Halle","Mצnchengladbach","Kiel","Wuppertal","Freiburg","Hagen",
+			"Münster","Rostock","Chemnitz","Braunschweig","Halle","Mönchengladbach","Kiel","Wuppertal","Freiburg","Hagen",
```
Mape6, line 343: As far as I know you should use the english names/translations here. Right?
These name will then be written by the TranslationWriter to the langugage files and everyone can translate it to their languages.

SomeTroglodyte: Precisely. So "Mönchenglatzbach = " should appear automatically in English.properties instead of what I removed - a while later - and promptly be ignored. When the key is already there - like in most of the files changed herein - I hope the TranslationWriter will simply let it be.

android/assets/jsons/translations/English.properties
SomeTroglodyte: Those are "-" lines, meaning: removed by this

Mape6: You're totally right 😅

android/assets/jsons/Civ V - Vanilla/Nations.json
Mape6: I'm confused 🤯
I thought that only english words will be in the JSON files and the TranslationWriter will copy them as "to translate" strings into the translation files.
So from my point of view, "Munster" should be in the Nations.json and after TRWriter did its job, we can change "Munster = " to "Munster = Münster"

SomeTroglodyte: Only if english natives are already used to the "Munster" form somehow - at least that was my reasoning. When 99% of them never heard of Mülheim they should get Mülheim... So Munich was clearly the prejudiced category, Dusseldoof borderline, and Munster - no the horror movie doesn't count?

Mape6: Ok. Do it like that.

PR #4103, opened 2021-06-10
Possible fix for rare unit swapping crash

avdstaaij:
This is a possible fix for the crash described in #4049 (look there for details).

To determine whether a unit swap is possible, `canUnitSwapToReachableTile` at some point removes the selected unit, then checks if the other unit can move there, then puts the selected unit back. The crash appears to be caused by the selected unit sometimes not being able to move to the tile it was standing on, for some mysterious reason. This change makes the function put the unit back anyway, ignoring whether or not he can move there. This is really just a band-aid, but it will probably prevent the crash from occuring.

### DIFF
--- a/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
+++ b/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
@@ -258,8 +258,10 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
     private fun canUnitSwapToReachableTile(reachableTile: TileInfo): Boolean {
         // Air units cannot swap
         if (unit.type.isAirUnit()) return false
+
         // We can't swap with ourself
         if (reachableTile == unit.getTile()) return false
+
         // Check whether the tile contains a unit of the same type as us that we own and that can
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -270,16 +272,31 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
         ) ?: return false
         val ourPosition = unit.getTile()
         if (otherUnit.owner != unit.owner || !otherUnit.movement.canReachInCurrentTurn(ourPosition)) return false
+
+        // This is a variant of unit.putInTile that doesn't check whether the unit can actually move
```
SomeTroglodyte, line 276: Candidate for `/**` snytax? Non-private, helpful text?

yairm210:
I see
If you can't find a problem in the existing code (and neither can I) I prefer to remain with the buggy version, since the crash only happens if the user actively chooses to do something, and hope for a user to provide a buggy save game to reproduce

Your code is fine, this is purely a 'business' decision :)

PR #4106, opened 2021-06-10
Updated Tradition branch to G&K

### LINKED ISSUE #4109: Aristocracy dosen't prove happiness boost
Aristocracy broken again? Have multiple cities with pop > 10. After enacting policy, I get the wonder production boost but no happiness boost.

xlenstra:
First part of the split of #4084 into multiple PR's for each branch.
This PR updates the Tradition branch to have G&K effects instead of those currently present.
This also fixes #4109 and the bugs noted in the conversation there.

Added uniques:
- "Immediately creates a [building] in each of your first [int] cities for free"
- "Immediately creates the cheapest available cultural building in each of your first [int] cities for free"
- "+[int]% attacking strength for cities with garrisoned units"
- "[stats] Per [int] Population [cityFilter]"

Removed uniques:
- "Immediately creates a cheapest available cultural building in each of your first 4 cities for free"
- "+50% attacking strength for cities with garrisoned units"
- "[stats] per [int] population in all cities"
- "[stats] per [int] population in this city"

This version has backwards compatibility with older versions.
One of the variables previously used in PolicyManager() has been renamed to reflect the changes to uniques, and the old version has been deprecated because of this. For now they point to the same Hashmap, making it backwards compatible.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -52,7 +52,7 @@
 		"name": "Library",
 		"hurryCostModifier": 25,
 		"maintenance": 1,
-		"uniques": ["[+1 Science] Per [2] Population in this city"],
+		"uniques": ["[+1 Science] Per [2] Population [in this city]"],
 		"requiredTech": "Writing"
 	},
 	{
@@ -72,7 +72,7 @@
 		"uniqueTo": "China",
 		"hurryCostModifier": 25,
 		"gold": 2,
-		"uniques": ["[+1 Science] Per [2] Population in this city"],
+		"uniques": ["[+1 Science] Per [2] Population [in this city]"],
 		"requiredTech": "Writing"
 	},
 	{
@@ -752,7 +752,7 @@
 		"requiredBuilding": "University",
 		"maintenance": 3,
 		"hurryCostModifier": 0,
-		"uniques": ["[+1 Science] Per [2] Population in this city"],
+		"uniques": ["[+1 Science] Per [2] Population [in this city]"],
 		"requiredTech": "Scientific Theory"
 	},
 	{
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -167,9 +167,9 @@ object BattleDamage {
                 modifiers["Statue of Zeus"] = 15
         } else if (attacker is CityCombatant) {
             if (attacker.city.getCenterTile().militaryUnit != null) {
-                val olichargyBonus = attacker.getCivInfo().getMatchingUniques("+[]% attacking strength for cities with garrisoned units")
-                    .fold(0) { sum, it -> sum + it.params[0].toInt() }
-                if (olichargyBonus != 0)
+                val garrisonBonus = attacker.getCivInfo().getMatchingUniques("+[]% attacking strength for cities with garrisoned units")
+                    .map {it.params[0].toInt()}.sum()
+                if (garrisonBonus != 0)
                     modifiers["Garrisoned unit"] = olichargyBonus
             }
         }
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -166,11 +166,12 @@ object BattleDamage {
             )
                 modifiers["Statue of Zeus"] = 15
         } else if (attacker is CityCombatant) {
-            if (attacker.getCivInfo()
-                    .hasUnique("+50% attacking strength for cities with garrisoned units")
-                && attacker.city.getCenterTile().militaryUnit != null
-            )
-                modifiers["Oligarchy"] = 50
+            if (attacker.city.getCenterTile().militaryUnit != null) {
+                val olichargyBonus = attacker.getCivInfo().getMatchingUniques("+[]% attacking strength for cities with garrisoned units")
+                    .fold(0) { sum, it -> sum + it.params[0].toInt() }
+                if (olichargyBonus != 0)
+                    modifiers["Garrisoned unit"] = olichargyBonus
+            }
         }
 
         return modifiers
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -170,7 +170,7 @@ object BattleDamage {
                 val garrisonBonus = attacker.getCivInfo().getMatchingUniques("+[]% attacking strength for cities with garrisoned units")
                     .map {it.params[0].toInt()}.sum()
                 if (garrisonBonus != 0)
-                    modifiers["Garrisoned unit"] = olichargyBonus
+                    modifiers["Garrisoned unit"] = garrisonBonus
             }
         }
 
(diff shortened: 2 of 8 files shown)


core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -166,11 +166,12 @@ object BattleDamage {
...
                 modifiers["Statue of Zeus"] = 15
         } else if (attacker is CityCombatant) {
-            if (attacker.getCivInfo()
-                    .hasUnique("+50% attacking strength for cities with garrisoned units")
-                && attacker.city.getCenterTile().militaryUnit != null
-            )
-                modifiers["Oligarchy"] = 50
+            if (attacker.city.getCenterTile().militaryUnit != null) {
+                val olichargyBonus = attacker.getCivInfo().getMatchingUniques("+[]% attacking strength for cities with garrisoned units")
+                    .fold(0) { sum, it -> sum + it.params[0].toInt() }
```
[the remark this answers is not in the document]
xlenstra: Fixed

android/assets/jsons/Civ V - Vanilla/Buildings.json
```diff
@@ -52,7 +52,7 @@
 		"name": "Library",
 		"hurryCostModifier": 25,
 		"maintenance": 1,
-		"uniques": ["[+1 Science] Per [2] Population in this city"],
+		"uniques": ["[+1 Science] Per [2] Population [in this city]"],
```
yairm210, line 55: Did you check this works? 'in this city' uniques need to be checked

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -81,7 +81,8 @@ class CityConstructions {
             stats.add(building.getStats(cityInfo.civInfo))
 
         for (unique in builtBuildingUniqueMap.getAllUniques()) when (unique.placeholderText) {
-            "[] Per [] Population in this city" -> stats.add(unique.stats.times(cityInfo.population.population / unique.params[1].toFloat()))
+            "[] Per [] Population []" -> if (cityInfo.matchesFilter(unique.params[2]))
```
yairm210, line 84: lowercase - see CityStats:258

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -158,13 +158,9 @@ class CityStats {
...
     fun getGrowthBonusFromPoliciesAndWonders(): Float {
         var bonus = 0f
-        // This requires more... complex navigation of the local uniques to merge into "+[amount]% growth [cityFilter]"
-        for (unique in cityInfo.civInfo.getMatchingUniques("+[]% growth in all cities"))
-            bonus += unique.params[0].toFloat()
-
         // "+[amount]% growth [cityFilter]"
         for (unique in cityInfo.civInfo.getMatchingUniques("+[]% growth []"))
-            if (cityInfo.matchesFilter(unique.params[0]))
+            if (cityInfo.matchesFilter(unique.params[1]))
```
yairm210, line 163: D'oh! Good catch!

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -24,8 +24,15 @@ class PolicyManager {
...
-    var legalismState = HashMap<String, String>()
+
+    private var cultureBuildingsAdded = HashMap<String, String>() // Maps cities to buildings
+    private var specificBuildingsAdded = HashMap<String, MutableSet<String>>() // Maps buildings to cities
+    private var maintenanceFreeBuildings = HashMap<String, MutableSet<String>>() // Maps cities to buildings
     var autocracyCompletedTurns = 0
+    
+    @Deprecated("Deprecated since 3.14.16") // Replaced with cultureBuildingsAdded
+    var legalismState = cultureBuildingsAdded // Maps cities to buildings
+    // We make it a reference copy of the original variable. This way, it can still works in older versions
```
yairm210, line 35: The problem is that if I have an old saved game, and I upgrade, I lose my older data. The way we've dealt with these kinds of things so far is to have a 'copying' function in GameInfo.setTransients from old to new variables.

android/assets/jsons/Civ V - Vanilla/Buildings.json
xlenstra: For all of the branch PR's, I checked every single policy unique again, just to be sure they all worked.
This one worked in the capital after building a library there, so I assumed it worked in all cities with a library

core/src/com/unciv/logic/city/CityConstructions.kt
yairm210: Also, please keep the older version for backwards compatibility, but mark it with a "Deprecated at 3.15.17" comment

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -441,7 +441,7 @@ class Building : NamedStats(), IConstruction {
     fun isStatRelated(stat: Stat): Boolean {
         if (get(stat) > 0) return true
         if (getStatPercentageBonuses(null).get(stat) > 0) return true
-        if (uniqueObjects.any { it.placeholderText == "[] Per [] Population in this city" && it.stats.get(stat) > 0 }) return true
+        if (uniqueObjects.any { it.placeholderText == "[] Per [] Population []]" && it.stats.get(stat) > 0 }) return true
```
yairm210, line 444: Same as above - keep older version, mark with 'deprecated' comment.
Also, you have an extra ] at the end here

yairm210: And lowercase

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -24,8 +24,15 @@ class PolicyManager {
     var numberOfAdoptedPolicies = 0
     var shouldOpenPolicyPicker = false
         get() = field && canAdoptPolicy()
-    var legalismState = HashMap<String, String>()
+
+    private var cultureBuildingsAdded = HashMap<String, String>() // Maps cities to buildings
+    private var specificBuildingsAdded = HashMap<String, MutableSet<String>>() // Maps buildings to cities
+    private var maintenanceFreeBuildings = HashMap<String, MutableSet<String>>() // Maps cities to buildings
```
yairm210, line 30: I fail to see why you need this - isn't the information you need already present in the above 2 new variables? We don't need to save this as part of the state

core/src/com/unciv/logic/city/CityConstructions.kt
xlenstra: I'm confused. In [this](https://github.com/yairm210/Unciv/pull/4084#issuecomment-859038672) comment in the old thread you seemed to imply that removing uniques wasn't a problem in general. So is deprecating to support backwards compatibility something that should be done in general, or just something for this specific case as this was a (probably) widely used unique?

core/src/com/unciv/logic/civilization/PolicyManager.kt
xlenstra: Was mostly for caching for easy access, but you could indeed derive this information from the above maps. I'll remove it.

SomeTroglodyte:
P.S. (sorry, I'm reading this in a completely unrelated diff):
```Kotlin
        if (cultureBuildingsAdded.isEmpty() && legalismState.isNotEmpty()) {
            cultureBuildingsAdded.putAll(legalismState)
        }
```
I would have cleared legalismState here, but since cultureBuildingsAdded is a hashmap and the whole thing short-lived it's not going to hurt.

PR #4108, opened 2021-06-11
Fix for 4094 - one city name in Nations.json

SomeTroglodyte:
Mistake in #4094 - I had intended this one city to have its 'real' name in english and templates. As #4105 now shows, I messed that one up.
** Question **: As I adapted the individual language files that had an entry for this city to the "Saarbrücken" spelling in 4094, is it now easier to fix the json or to jus tlet it be... The more important goals of #4094 are fine after all...?


yairm210:
I have no problem either way, but if you're changing this, change it in all translation files as well

SomeTroglodyte:
No, I see the bot was quicker, those lines were in 4094 but are already gone. So I'll do it properly - sometime, maybe, not now.

PR #4114, opened 2021-06-11
Display Movement Paths

### LINKED ISSUE #3434: path of movement
let the trajectory of the movement of units be displayed without taking into account obstacles in the fog of war, for example, using a line with marks of the number of moves. this would be useful not so much for visualization as for determining the shortest path and throughput in units in mountainous terrain

r3versi:
This implements and close #3434 

It was a low-hanging fruit: we were already calculating shortest paths towards a tile. I only added very-basic visualization logic.
I believe this is could also be a useful tool to debug and improve current BFS that is the guy draining our batteries :)

Further steps might include:
- [ ] Showing a dotted line between tiles
- [ ] Showing a number, representing the turn the tile will be occupied by the unit (potential issues in case of multiple selected units). How does Civ implements this?

Screenshot from 2021-06-11 18-37-33

### DIFF
--- a/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
+++ b/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
@@ -113,7 +113,7 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
     }
 
     /**
-     * Does not consider if the destination tile can actually be entered, use canMoveTo for that.
+     * Does not consider if the [destination] tile can actually be entered, use [canMoveTo] for that.
      * Returns an empty list if there's no way to get to the destination.
      */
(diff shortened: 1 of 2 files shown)


ravignir:
So the way civ5 does that thing: it shows the path to a tile with dotted arrow lines and displays the turn count on each of the tiles on that path. If the path is unknown it assumes 1 tile =1 turn.

The unit stops movement and asks you for another action instead if there are obstacles that prevent it from reaching the destination point - happens when it can't move there because the final tile is already occupied by another unit - for example when you order two units to move into same tile, the second unit will stop on its final turn.

yairm210:
Nice!
I think this is already mergable as-is, future improvements always welcome :)

PR #4115, opened 2021-06-11
Updated Liberty branch to G&K

xlenstra:
Second part of the split of #4084 into multiple PR's for each branch.
This PR updates the Liberty branch to have G&K effects instead of those currently present.
This PR also unifies the citizenship and pyramids uniques, which both impact the build speed of improvements.

Added uniques:
- "Tile improvement build time -[int]%"
- "Unhappiness from population decreased by [int]% [cityFilter]"

Removed uniques:
- "Worker construction increased 25%"
- "Tile improvement speed +25%"

This version mostly has backwards compatibility with older versions.
Previously, the republic policy was locked behind the collective rule policy, but it now is the opposite.
Due to this, some players may suddenly find themselves having a policy without having its required policies.
This is minor, and should not impact the game that much.

### DIFF
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -467,14 +467,15 @@ class CityInfo {
     }
 
     fun matchesFilter(filter: String): Boolean {
-        return when {
-            filter == "in this city" -> true
-            filter == "in all cities" -> true
-            filter == "in all coastal cities" && getCenterTile().isCoastalTile() -> true
-            filter == "in capital" && isCapital() -> true
-            filter == "in all cities with a world wonder" && cityConstructions.getBuiltBuildings().any { it.isWonder } -> true
-            filter == "in all cities connected to capital" -> isConnectedToCapital()
-            filter == "in all cities with a garrison" && getCenterTile().militaryUnit != null -> true
+        return when (filter) {
+            "in this city" -> true
+            "in all cities" -> true
+            "in all coastal cities" -> getCenterTile().isCoastalTile()
+            "in capital" -> isCapital()
+            "in all non-occupied cities" -> !cityStats.hasExtraAnnexUnhappiness()
+            "in all cities with a world wonder" -> cityConstructions.getBuiltBuildings().any { it.isWonder }
+            "in all cities connected to capital" -> isConnectedToCapital()
+            "in all cities with a garrison" -> getCenterTile().militaryUnit != null
             else -> false
         }
     }
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -472,7 +472,7 @@ class CityInfo {
             "in all cities" -> true
             "in all coastal cities" -> getCenterTile().isCoastalTile()
             "in capital" -> isCapital()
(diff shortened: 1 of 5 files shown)


yairm210:
English suggestion: "Tile improvement build time -[25]%" -> "-[25]% tile improvement construction time"

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -467,14 +467,15 @@ class CityInfo {
...
-            filter == "in capital" && isCapital() -> true
-            filter == "in all cities with a world wonder" && cityConstructions.getBuiltBuildings().any { it.isWonder } -> true
-            filter == "in all cities connected to capital" -> isConnectedToCapital()
-            filter == "in all cities with a garrison" && getCenterTile().militaryUnit != null -> true
+        return when (filter) {
+            "in this city" -> true
+            "in all cities" -> true
+            "in all coastal cities" -> getCenterTile().isCoastalTile()
+            "in capital" -> isCapital()
+            "in all non-occupied cities" -> !cityStats.hasExtraAnnexUnhappiness()
```
yairm210, line 475: Do you mean puppeted? "in all non-puppeted cities"?

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -28,11 +28,10 @@ class TileImprovement : NamedStats() {
 
     fun getTurnsToBuild(civInfo: CivilizationInfo): Int {
         var realTurnsToBuild = turnsToBuild.toFloat() * civInfo.gameInfo.gameParameters.gameSpeed.modifier
-        // todo UNIFY THESE
-        if (civInfo.hasUnique("Worker construction increased 25%"))
-            realTurnsToBuild *= 0.75f
-        if (civInfo.hasUnique("Tile improvement speed +25%"))
-            realTurnsToBuild *= 0.75f
```
yairm210, line 35: Keep old uniques for compatibility, mark as deprecated with comment

core/src/com/unciv/logic/city/CityInfo.kt
xlenstra: This is the official wording in the game, but it wasn't clear to me what exactly it meant either. On internet I found a [reddit](https://www.reddit.com/r/civ/comments/5oq3c1/civ5_meritocracy_unhappiness_bonus_do_annexed/) post saying that this effect applies to all cities that are annexed but don't have a courthouse yet. I don't know whether this applies to puppetted cities as well.

xlenstra: For now I'll assume it to mean 'annexed but without courthouse || puppetted'.

yairm210: Ok
We can always change it later

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -28,11 +28,17 @@ class TileImprovement : NamedStats() {
...
+        for (unique in civInfo.getMatchingUniques("-[]% tile improvement construction time")) {
```
SomeTroglodyte, line 31: Sorry again for being late - thus just food for thought: I wouldn't have hardcoded this to be a bonus - allowing weird mods to slow down workers wouldn't have cost us anything here, just moving the sign to json...

PR #4117, opened 2021-06-11
CityScreen resize

SomeTroglodyte:
CityScreen adapts better to both desktop resizes and Android orientation change - just a copy&paste (but tested).


yairm210:
I wish I could figure out a good way to make this 'generic' for all screens, but since it involves creating a new screen I don't see how. I really don't want to go Reflection, and adding a generic 'clone' function to screens is exactly the same - and probably worse for debug - as doing this for all screens.

SomeTroglodyte:
> generic

Well, if you can stomach a solution that uses reflection and requires standardizing constructors over all subclasses...
`this::class.constructors.first().call()` from the superclass actually gives you a new instance of the subclass calling it. :vomiting_face:  ..brain malfunction..

PR #4120, opened 2021-06-12
Deleted the World Wrap part from README.md

jahodovykolac:
World Wrap has been implemented, so it shouldn't already be as a thing-that-people-can-try


yairm210:
Correct

PR #4124, opened 2021-06-13
Proposal for New borders 

r3versi:
Screenshot from 2021-06-12 11-14-51


core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -792,9 +779,12 @@ open class TileGroup(var tileInfo: TileInfo, var tileSetStrings:TileSetStrings,
         }
     }
 
-    fun showCircle(color: Color, alpha: Float = 0.3f) {
+    fun showCircle(color: Color, alpha: Float = 0.3f, number: Int = -1) {
         circleImage.isVisible = true
         circleImage.color = color.cpy().apply { a = alpha }
+
+        if (number >= 0)
+            miscLayerGroup.addActor(number.toLabel())
```
yairm210, line 787: This looks like part of a different PR.

PR #4128, opened 2021-06-13
Tooltips - all key indicators

SomeTroglodyte:
This is the cleaned version of the code behind that clip in #4082. Differences:
- Sizing - had to learn about that `resizeActor` parameter first. Letters no longer square, a little higher and narrower.
- _All_ existing indicators replaced (yes this was a single-file diff not long ago, now it touches six).
- Special-cased "i" - looks IMHO better as minuscule
- As so often emptied "Problems" panes except EmpireOverviewScreen - my bad, I read "DiplomacyGroup" and thought there's a bug in there anyway so postpone - the bug is actually in the real Diplomacy screen...


SomeTroglodyte:
... And I still overlooked the ones in improvement picker. But the 'keyboard project' isn't completely done, yet, so there's hope..

PR #4129, opened 2021-06-13
Updated Commerce branch to G&K

xlenstra:
Fourth part of the split of #4084 into multiple PR's for each branch.
This PR updates the Commerce branch to have G&K effects instead of those currently present.

Added uniques:
- "+[int] happiness from each type of luxury resource"

Deprecated uniques:
- "+1 happiness from each type of luxury resource"
- "All military naval units receive +1 movement and +1 sight"

After merging this PR some problems might arise when updating an existing game. This is, as the commerce tree contains multiply differently named policies, and shuffles those with the same name around quite a bit. I have fixed this by automatically converting removed policies to the new policies, but this might result in some policy trees looking weird. It is, for example, possible to only have both the second tier policies of the branch, without unlocking the first tier.


xlenstra:
I'll fix the merge conflicts once master builds again

yairm210:
Master builds

xlenstra:
Fixed merge conflicts

android/assets/jsons/Civ V - Vanilla/Buildings.json
```diff
@@ -203,7 +203,7 @@
 		"greatPersonPoints": {"gold": 1},
 		"isWonder": true,
 		"providesFreeBuilding":  "Lighthouse",
-		"uniques": ["Must be next to [Coast]", "All military naval units receive +1 movement and +1 sight"],
+		"uniques": ["Must be next to [Coast]", "+[1] movement and +[1] sight for all [military water] units"],
```
yairm210, line 206: Split movement and sight into 2 separate uniques

yairm210: I think that they should be capitalized as well, I can't recall what we usually use

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -141,14 +141,25 @@ class MapUnit {
...
+        // Deprecated since 3.14.17
+            if (type.isMilitary() && type.isWaterUnit() && civInfo.hasUnique("All military naval units receive +1 movement and +1 sight")) {
+                movement += 1
+            }
+        //
+        
         for (unique in civInfo.getMatchingUniques("+[] Movement for all [] units"))
             if (matchesFilter(unique.params[1]))
                 movement += unique.params[0].toInt()
+        for (unique in civInfo.getMatchingUniques("+[] movement for every []")) {
```
yairm210, line 159: "+[] Movement for all [] units" - that way we can have it match unitFiltet

PR #4131, opened 2021-06-13
Fix silent crash when ancient ruins cascade gifted units

SomeTroglodyte:
Probability on generated maps near zero, but on edited maps it's just quite small. Scenario: An ancient ruin pops a unit of the same civilian/military class as the unit that stepped on the ruins, and that new unit is placed onto another ruin, and that ruin chooses the unit gift, too. The second will eventually call `civInfo.updateStatsForNextTurn` _before_ the first unit is fully initialized, namely before it has a valid `currentTile`. :bomb: 

Took me a few minutes to debug due to being buried under a few thread switches and a try catch{}. (actually - debugging on hardware your process would die _while_ you were single stepping something else or analyzing watches - and desktop died with zero messages, not even 'gradle build failed' - but one step _after_ the problem). Therefore I left all the purely 'defensive' changes _in_ this PR.

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -607,6 +607,11 @@ class CivilizationInfo {
         }
     }
 
+    /** Tries to place the a [unitName] unit into the [TileInfo] closest to the given the [position]
+     * @param location where to try to place the unit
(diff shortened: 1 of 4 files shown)


SomeTroglodyte:
`Unresolved reference: BONUS_VS_UNIT_TYPE` - not my baby. Seems I'll have to test where that comes from...

PR #4133, opened 2021-06-13
Fix crash caused by cascaded ruins unit gifts

SomeTroglodyte:
#4131 retry, now based on #4132 - I'm super curious whether the github builds will be OK...

Original comment:
Probability on generated maps near zero, but on edited maps it's just quite small. Scenario: An ancient ruin pops a unit of the same civilian/military class as the unit that stepped on the ruins, and that new unit is placed onto another ruin, and that ruin chooses the unit gift, too. The second will eventually call civInfo.updateStatsForNextTurn before the first unit is fully initialized, namely before it has a valid currentTile. bomb

Took me a few minutes to debug due to being buried under a few thread switches and a try catch{}. (actually - debugging on hardware your process would die while you were single stepping something else or analyzing watches - and desktop died with zero messages, not even 'gradle build failed' - but one step after the problem). Therefore I left all the purely 'defensive' changes in this PR.

### DIFF
--- a/android/assets/jsons/translations/Czech.properties
+++ b/android/assets/jsons/translations/Czech.properties
@@ -2882,7 +2882,7 @@ Professional Army = Profesionální armáda
 Gold cost of upgrading [unitType] units reduced by [amount]% = 
 Honor Complete = Kompletní Čest
 Honor = Čest
-+[amount]% Strength vs [param] = +[amount]% síly proti [unitType]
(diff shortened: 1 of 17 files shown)


yairm210:
My bad...

SomeTroglodyte:
The occasional slip just shows how bafflingly good you are at this. Coding skills _and_ management skills combined isn't that common. I had fun with this... 👍

yairm210:
Aw shucks 😍

PR #4136, opened 2021-06-13
Updated Rationalism branch to G&K

xlenstra:
Fifth part of the split of #4084 into multiple PR's for each branch.
This PR updates the Rationalism branch to have G&K effects instead of those currently present.

Added uniques:
- "Science gained from research agreements +[int]%"
- "+[int]% [stat] while the empire is happy"

Deprecated uniques:
- "Science gained from research agreements +50%"
- "+15% science while empire is happy"

This PR is fully backwards compatible.


core/src/com/unciv/models/ruleset/Unique.kt
```diff
@@ -81,7 +81,8 @@ object UniqueTriggerActivation {
                     city.population.population += 1
                     city.population.autoAssignPopulation()
                 }
-            "Free Technology" -> civInfo.tech.freeTechs += 1
+            "Free Technology" -> if (!civInfo.isSpectator()) civInfo.tech.freeTechs += 1
+            "[] Free Technologies" -> if (!civInfo.isSpectator()) civInfo.tech.freeTechs += unique.params[0].toInt()
```
avdstaaij, line 85: Since this adds the "[] Free Technologies" unique, perhaps "Free technology" should be deprecated and replaced by "[1] Free Technologies" to limit duplication.

yairm210: The single/multiple phrasing doesn't really accommodate both in one sentence, this is good

PR #4142, opened 2021-06-14
Updated Freedom branch to G&K

xlenstra:
Sixth part of the split of #4084 into multiple PR's for each branch.
This PR updates the Freedom branch to have G&K effects instead of those currently present.

Added uniques:
- "-[int]% food consumption by specialists"
- "[int] units cost no maintenance"
- "Specialists only produce [int]% of normal unhappiness"
- "Tile yield from [tileFilter] +[int]%"
- "-[int]% unit upkeep costs"

Deprecated uniques:
- "-50% food consumption by specialists"
- "Specialists produce half normal unhappiness"
- "Tile yield from Great Improvements +100%"
- "-33% unit upkeep costs"

This PR is fully backwards compatible.

### DIFF
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -192,50 +192,57 @@ object BattleDamage {
         return modifiers
     }
 
-    fun getDefenceModifiers(attacker: ICombatant, defender: MapUnitCombatant): Counter<String> {
+    fun getDefenceModifiers(attacker: ICombatant, defender: ICombatant): Counter<String> {
         val modifiers = getGeneralModifiers(defender, attacker)
         val tile = defender.getTile()
+    
+        if (defender is MapUnitCombatant) {
 
-        if (defender.unit.isEmbarked()) {
-            // embarked units get no defensive modifiers apart from this unique
-            if (defender.unit.hasUnique("Defense bonus when embarked") ||
-                defender.getCivInfo().hasUnique("Embarked units can defend themselves")
+            if (defender.unit.isEmbarked()) {
+                // embarked units get no defensive modifiers apart from this unique
+                if (defender.unit.hasUnique("Defense bonus when embarked") ||
+                    defender.getCivInfo().hasUnique("Embarked units can defend themselves")
+                )
+                    modifiers["Embarked"] = 100
+
+                return modifiers
+            }
+
+            modifiers.putAll(getTileSpecificModifiers(defender, tile))
+
+            val tileDefenceBonus = tile.getDefensiveBonus()
+            if (!defender.unit.hasUnique("No defensive terrain bonus") && tileDefenceBonus > 0
+                || !defender.unit.hasUnique("No defensive terrain penalty") && tileDefenceBonus < 0
             )
-                modifiers["Embarked"] = 100
+                modifiers["Tile"] = (tileDefenceBonus * 100).toInt()
 
-            return modifiers
+            if (attacker.isRanged()) {
+                val defenceVsRanged = 25 * defender.unit.getUniques()
+                    .count { it.text == "+25% Defence against ranged attacks" }
+                if (defenceVsRanged > 0) modifiers["defence vs ranged"] = defenceVsRanged
+            }
+
+            for (unique in defender.unit.getMatchingUniques("+[]% Strength when defending")) {
+                modifiers.add("Defender Bonus", unique.params[0].toInt())
+            }
+
+            for (unique in defender.unit.getMatchingUniques("+[]% defence in [] tiles")) {
+                if (tile.matchesUniqueFilter(unique.params[1]))
+                    modifiers["[${unique.params[1]}] defence"] = unique.params[0].toInt()
+            }
+
+            if (defender.unit.isFortified())
+                modifiers["Fortification"] = 20 * defender.unit.getFortificationTurns()
+        } else if (defender is CityCombatant) {
+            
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -192,50 +192,57 @@ object BattleDamage {
...
+            }
 
-        if (defender.unit.isFortified())
-            modifiers["Fortification"] = 20 * defender.unit.getFortificationTurns()
+            if (defender.unit.isFortified())
+                modifiers["Fortification"] = 20 * defender.unit.getFortificationTurns()
+        } else if (defender is CityCombatant) {
+            
+            modifiers["Defensive Bonus"] = defender.city.civInfo.getMatchingUniques("+[]% Defensive strength for cities")
+                .fold(0f) { sum, it -> sum + it.params[0].toFloat() / 100f }.toInt()
```
yairm210, line 239: Again, .map{it.params[0].toFloat()/100}.sum().toInt()

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -499,8 +510,17 @@ class CityStats {
...
-        if (cityInfo.civInfo.hasUnique("-50% food consumption by specialists"))
-            foodEaten -= cityInfo.population.getNumberOfSpecialists()
+        var foodEatenBySpecialists = 2f * cityInfo.population.getNumberOfSpecialists()
+        
+        foodEaten -= foodEatenBySpecialists
+        for (unique in cityInfo.civInfo.getMatchingUniques("-[]% food consumption by specialists")) 
+            foodEatenBySpecialists *= 1f - unique.params[0].toFloat() / 100f
+        
+        // Deprecated since 3.15
+            if (cityInfo.civInfo.hasUnique("Specialists produce half normal unhappiness"))
```
yairm210, line 520: Wrong unique name

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -16,7 +16,11 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
 
     private fun getUnitMaintenance(): Int {
         val baseUnitCost = 0.5f
-        val freeUnits = 3
+        var freeUnits = 3
+        for (unique in civInfo.getMatchingUniques("[] units cost no maintenance")) {
```
yairm210, line 20: if unit matches unique, surely?

yairm210: Ah that's an int
Sorry, you're right

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -315,7 +311,16 @@ open class TileInfo {
                 }
                 stats.add(unique.stats.times(numberOfBonuses.toFloat()))
             }
-
+        
+        for (unique in observingCiv.getMatchingUniques("Tile yield from [] +[]%"))
```
yairm210, line 315: If we're rewriting this, I would go with "+[]% yield from []", but even that's problematic.
You see, you're checking the unique filter, but the stats are for the improvement only, yes? So if it says like "+[50]% Tile yield from [Ocean]" it will still give the yield bonus to the improvement...

xlenstra: I have implemented this, but by doing so also cleaned up some other code which _might_ result in incompatible mods, but I don't expect it. If you do expect it, please tell me and I'll revert the change. Change is mostly the split of the three uniques inside the ```if (city != null)``` block in this same function.

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -174,15 +174,27 @@ class CityStats {
...
+        
+        for (unique in civInfo.getMatchingUniques("Specialists only produce []% of normal unhappiness")) {
+            unhappinessFromSpecialists *= (1f - unique.params[0].toFloat() / 100f)
+        }
+        // Deprecated since 3.15
+            if (civInfo.hasUnique("Specialists produce half normal unhappiness"))
+                unhappinessFromSpecialists *= 0.5f
+        //
+        
+        unhappinessFromCitizens += unhappinessFromSpecialists
```
yairm210, line 197: Surely it would make more sense to just remove the unhappiness at the end instead of adding and then subtracting?

xlenstra: It would, but then I would have to do maths and this was easier :p
I'll update it in a bit.

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -499,8 +510,17 @@ class CityStats {
...
+        foodEaten -= foodEatenBySpecialists
+        for (unique in cityInfo.civInfo.getMatchingUniques("-[]% food consumption by specialists")) 
+            foodEatenBySpecialists *= 1f - unique.params[0].toFloat() / 100f
+        
+        // Deprecated since 3.15
+            if (cityInfo.civInfo.hasUnique("-50% food consumption by specialists"))
+                foodEatenBySpecialists *= 0.5f
+        //
+        
+        foodEaten += foodEatenBySpecialists
```
yairm210, line 524: Same here. Maybe it's just me, I can see what you're doing but that's only because I know how the system works, to someone new this would look very strange.

xlenstra:
Sorry, as `isAdjacentToFreshWater` looked like a variable, I assumed it would just be set to some value instead of calling a function to determine its value. My bad!

PR #4146, opened 2021-06-14
Request for comments: Prepare resources for better mod-ability

SomeTroglodyte:
#### Inspired by #3242: Concept on how to begin making resources more mod-able.

- Centralize logic (e.g. terrainsCanBeFoundOn would still work as private)
- Allow more than 1 unique with compatibility
- Parameterize the one existing unique
- Two new uniques as samples

#### Questions:
- Is this the correct approach?
- Why does TileInfo.normalizeToRuleset use a different logic? Would making it use `! isAllowedOnTile` hurt? How? (and if it stays it would need to respect those uniques too)
- That hardcoded Snow Hill exclusion - how to best 'soften' that? I gave it lower precedence so mods could override, but this still looks a tad inelegant
- +15 production could use the stats mechanic, but percent? Smells like it needs an idea, which is why i postponed parameterizing the wonder part

### DIFF
--- a/core/src/com/unciv/logic/city/CityStats.kt
+++ b/core/src/com/unciv/logic/city/CityStats.kt
@@ -36,7 +36,7 @@ class CityStats {
     @Transient
     lateinit var cityInfo: CityInfo
 
-    //region pure fuctions
+    //region pure functions
     private fun getStatsFromTiles(): Stats {
         val stats = Stats()
         for (cell in cityInfo.tilesInRange
@@ -53,7 +53,7 @@ class CityStats {
             stats.gold = civInfo.getCapital().population.population * 0.15f + cityInfo.population.population * 1.1f - 1 // Calculated by http://civilization.wikia.com/wiki/Trade_route_(Civ5)
             for (unique in civInfo.getMatchingUniques("[] from each Trade Route"))
                 stats.add(unique.stats)
-            if (civInfo.hasUnique("Gold from all trade routes +25%")) stats.gold *= 1.25f // Machu Pichu speciality
+            if (civInfo.hasUnique("Gold from all trade routes +25%")) stats.gold *= 1.25f // Machu Picchu speciality
         }
         return stats
     }
@@ -81,7 +81,7 @@ class CityStats {
         if (railroadImprovement == null) return stats // for mods
         val techEnablingRailroad = railroadImprovement.techRequired!!
         // If we conquered enemy cities connected by railroad, but we don't yet have that tech,
-        // we shouldn't get bonuses, it's as if the tracks aare layed out but we can't operate them.
+        // we shouldn't get bonuses, it's as if the tracks are laid out but we can't operate them.
         if (cityInfo.civInfo.tech.isResearched(techEnablingRailroad)
                 && (cityInfo.isCapital() || isConnectedToCapital(RoadStatus.Railroad)))
             stats.production += 25f
@@ -91,11 +91,13 @@ class CityStats {
     private fun getStatPercentBonusesFromResources(construction: IConstruction): Stats {
         val stats = Stats()
 
-        if (construction is Building
-                && construction.isWonder
-                && cityInfo.civInfo.getCivResources()
-                        .any { it.amount > 0 && it.resource.unique == "+15% production towards Wonder construction" })
-            stats.production += 15f
+        if (construction is Building && construction.isWonder) {
+            val bonus = cityInfo.civInfo.getCivResources()
+                .filter { it.amount > 0 }
+                .maxOfOrNull { it.resource.getWonderProductionBonus() }
+            if (bonus != null)
+                stats.production += bonus
+        }
 
         return stats
     }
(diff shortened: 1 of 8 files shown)


SomeTroglodyte:
P.S.: This would cover [@GoldenBronze's "River" post](https://github.com/yairm210/Unciv/issues/3242#issuecomment-704946704) cheaply - but as I said, that part is just a demo of the possibilities

yairm210:
If anything, resource uniques should be merged into global uniques, so they won't need any special attention. This is actually much simpler that other solutions as well :)

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -433,7 +435,7 @@ class CityStats {
         if (cityInfo.getRuleset().modOptions.uniques.contains("Can convert gold to science with sliders")) {
             val amountConverted = (newFinalStatList.values.sumByDouble { it.gold.toDouble() }
                     * cityInfo.civInfo.tech.goldPercentConvertedToScience).toInt().toFloat()
-            if (amountConverted > 0) // Don't want you converting negative gold to negative science yaknow
```
yairm210, line 436: :(

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -661,8 +661,7 @@ open class TileInfo {
 
         if (resource != null && !ruleset.tileResources.containsKey(resource)) resource = null
         if (resource != null) {
-            val resourceObject = ruleset.tileResources[resource]!!
-            if (resourceObject.terrainsCanBeFoundOn.none { it == baseTerrain || terrainFeatures.contains(it) })
+            if (ruleset.tileResources[resource]!!.mustRemoveFromTile(this))
```
yairm210, line 664: Why not just !isAllowedOnTile(this)? Why make an extra function? I think the performance difference is negligible

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -320,13 +320,7 @@ class Ruleset {
...
 
         for (resource in tileResources.values) {
-            if (resource.revealedBy != null && !technologies.containsKey(resource.revealedBy!!))
-                lines += "${resource.name} revealed by tech ${resource.revealedBy} which does not exist!"
-            if (resource.improvement != null && !tileImprovements.containsKey(resource.improvement!!))
-                lines += "${resource.name} improved by improvement ${resource.improvement} which does not exist!"
-            for (terrain in resource.terrainsCanBeFoundOn)
-                if (!terrains.containsKey(terrain))
-                    lines += "${resource.name} can be found on terrain $terrain which does not exist!"
+            warningCount += resource.checkModLinks(this, lines)
```
yairm210, line 323: I really don't think this is good. Mod links should be checked in the ruleset, not in the resource...

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -1,29 +1,125 @@
...
+    fun isAllowedOnTerrain(terrain: String): Boolean {
+        // todo: Called from Terrain.getDescription - how to support tileFilter=terrain uniques?
+        return terrain in terrainsCanBeFoundOn
+    }
+
+    /** Get a terrain the map editor can use as background
+     * @param ruleset The ruleset to search for viable terrains
+     * @return Terrain name or null if impossible due to [ruleset] inconsistency 
+     */
+    fun getSampleTerrain(ruleset: Ruleset): String? {
```
yairm210, line 83: For one usage? I don't think this is needed.

SomeTroglodyte:
> If anything, resource uniques should be merged into global uniques, so they won't need any special attention. This is actually much simpler that other solutions as well :)

Limit them to civ-global effects? Marble isn't global, or is it?

core/src/com/unciv/models/ruleset/tile/TileResource.kt
SomeTroglodyte: I know. That's got to do with "general structure" - I generalized as much as possible first, planning to fold such stuff later if rarely (or not at all) used...

core/src/com/unciv/models/ruleset/Ruleset.kt
SomeTroglodyte: Marginal decision either way. Makes property access a bit shorter on average was the idea methinks.

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: There's a comment to that (is there)? The original code did already check differently, so I left the discrepancy but wrapped it.

core/src/com/unciv/logic/city/CityStats.kt
SomeTroglodyte: Didn't I just remove the "yaknow" which triggers the spellchecker? ... yup. Such language may be nice to lighten a mood - but aren't we in the worst possible mood beyond redemption when coding anyway? :wink:

core/src/com/unciv/logic/map/TileInfo.kt
yairm210: I think this should be the same check, if we're unifying things

yairm210:
Marble is global. Marble anywhere = marble in civ = wonder bonus.

SomeTroglodyte:
You're right, I just checked - CityStats line 95 decides that.  

Then we should maybe change that to per city: ["in the city where it is worked"](https://civilization.fandom.com/wiki/Marble_(Civ5))? Would this kind of change be an argument in favour of the general architecture I suggested here? Seems not to me, I mean, woudn't make any big difference.

PR #4147, opened 2021-06-14
Update Spanish.properties

ArchDuque-Pancake:
I have no  motivation for the leader start intros... I'll do it later.


SomeTroglodyte:
No te apures. Do we need an _option_ to turn them off?
(1046: m<i><b>e</b></i>joras?)

PR #4149, opened 2021-06-15
Updated Autocracy branch to G&K

xlenstra:
Seventh part of the split of #4084 into multiple PR's for each branch.
This PR updates the Autocracy branch to have G&K effects instead of those currently present.

Added uniques:
- "-[int]% unit upkeep costs"
- "Upon capturing a city, receive [int] times its [Stat] production as [Stat] immediately"
- "Gold cost of purchasing units -[int]%"
- "Quantity of strategic resources produced by the empire +[int]%"
- "+[int]% attack bonus to all Military Units for [int] turns"

Deprecated uniques:
- "-33% unit upkeep costs"
- "Quantity of strategic resources produced by the empire increased by 100%"
- "+20% attack bonus to all Military Units for 30 turns"

This PR is fully backwards compatible, but it might have merge conflicts with the Freedom update


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -566,6 +573,18 @@ class CivilizationInfo {
             else -> gold + delta
         }
     }
+    
+    fun addStat(stat: Stat, amount: Int) {
```
yairm210, line 577: Nice! Been thinking of adding something like this for a while now 🙂

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -112,10 +112,17 @@ class BaseUnit : INamed, IConstruction {
 
     override fun getGoldCost(civInfo: CivilizationInfo): Int {
         var cost = getBaseGoldCost(civInfo)
-        if (civInfo.hasUnique("Gold cost of purchasing units -33%")) cost *= 0.66f
+        for (unique in civInfo.getMatchingUniques("Gold cost of purchasing units -[]%")) {
```
yairm210, line 115: I'm debating whether it's worth adding a unit filter for this unique

android/assets/jsons/Civ V - Vanilla/Policies.json
```diff
@@ -403,34 +403,35 @@
...
-				"uniques": ["+[15]% Production when constructing [Military] units [in all cities]", "New [Military] units start with [15] Experience"],
+				"uniques": ["+[25]% Production when constructing [Military] units [in all cities]", "New [Military] units start with [15] Experience"],
 				"requires": ["Police State","Fascism"],
 				"row": 3,
 				"column": 4
 			},
 			{
 				"name": "Autocracy Complete",
-				"uniques": ["+20% attack bonus to all Military Units for 30 turns"]
+				"uniques": ["+[25]% attack bonus to all Military Units for [50] turns"]
```
yairm210, line 434: Unitfilter here, while you're at it ;)
I hope this isn't too annoying for you, but if we already need to change uniques in both mods and translations, let's not have to do it again in the future...

xlenstra: No worries, I'll update it this evening

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -27,11 +27,14 @@ class PolicyManager {
 
     private var cultureBuildingsAdded = HashMap<String, String>() // Maps cities to buildings
     private var specificBuildingsAdded = HashMap<String, MutableSet<String>>() // Maps buildings to cities
+
+    @Deprecated("Deprecated since 3.15")
     var autocracyCompletedTurns = 0
-    
+    var activeCombatBonuses = mutableSetOf<Pair<Float, Int>>()
```
yairm210, line 33: I think maybe change this to a countdown timer for all uniques, String to Int - maybe even in CivInfo
That way we can allow any variance in future countdown uniques
This may be way out of scope for this PR so probably just ignore this policy for now and we'll do it properly in another PR?

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -112,10 +112,18 @@ class BaseUnit : INamed, IConstruction {
 
     override fun getGoldCost(civInfo: CivilizationInfo): Int {
         var cost = getBaseGoldCost(civInfo)
-        if (civInfo.hasUnique("Gold cost of purchasing units -33%")) cost *= 0.66f
+        for (unique in civInfo.getMatchingUniques("Gold cost of purchasing [] units -[]%")) {
+            if (matchesFilter(unique.params[0]))
```
yairm210, line 116: Maybe "[] units cost []% less Gold to purchase"

Your thoughts?

core/src/com/unciv/logic/civilization/PolicyManager.kt
xlenstra: If we change it, I'd rather do it in this PR, as then we don't have to support two different deprecated systems for the same feature. I'll think about the best approach some more

xlenstra: I now implemented it so that each civilization has a list of 'temporary uniques'. These uniques are added together with an integer representing how many turns they are active. If this ends, it expires. Not sure if this is the best way to do it, but it works.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
xlenstra: I don't have a huge preference either way, so I kept it as is.

PR #4148, opened 2021-06-14
Show promises not to settle, ask on found city action

SomeTroglodyte:
Might resolve #3976. Old code, and instead of ditching it I decided to test and offer.

https://user-images.githubusercontent.com/63000004/121965462-59673f00-cd6d-11eb-83ed-79060b241142.mp4

PR #4156, opened 2021-06-15
Fix question on found city action breaking promise - no AI

SomeTroglodyte:
I fear #4148 will break AI city founding, so this reverts the getFoundCityAction to its former behaviour - ***if*** the unit owner is an AI. This hit my pain threshold for complex nesting so I pulled it apart a little, too.

Please double-check my reasoning - I've just tested past _one_ city founding. I deemed this necessary because Alt-F7 found a usage in `SpecificUnitAutomation`. I have not found the code where the AI considers breaking such a promise or not - and thus have not checked possible correlations.


yairm210:
OK, seems legit. I'm not sure about the founding AI but I'd rather just revert this now and we can deal with implications later :)

SomeTroglodyte:
Meanwhile I playtested this through 1 AI-founded city, much relieved. Since the question "really break promise" still works, I'm content to await other player's complaints...

PR #4158, opened 2021-06-15
Created Order branch with G&K policies

xlenstra:
Eighth and final part of the split of #4084 into multiple PR's for each branch.
This PR creates the order branch to have G&K effects instead of those currently present.
It also updates some uniques to more modular or better worded versions.

Added uniques:
- "+[int]% Strength for units fighting in [tileFilter]"
- "+[int]% Production when constructing a [buildingFilter]"
- "-[int]% maintenance costs for buildings [cityFilter]"
- "Militaristic City-States grant units [int] times as fast when you are at war with a common nation"
- "[Stats] from every [improvementFilter] improvement"

Deprecated uniques:
- "+[int]% combat bonus for units fighting in [tileFilter]"
- "-[int]% building maintenance costs [cityFilter]"

Apart from adding the order branch itself, the following have also been done:
- The policypicker screen has been updated with the code provided by @SomeTroglodyte in #4084, so that it looks decent
- The CityStats class has been slightly refactored
- The Freedom, Order and Autocracy branches have been made incompatable

Something else I noticed: The `matchesUniqueFilter` of the `TileInfoClass` returns true if (among other things) either of these two is true: ```filter == resource``` and ```civInfo != null && hasViewableResource(civInfo) && resource == filter```
This seemed weird to me, so I removed the first one. Most of the time you don't want to activate effects for tiles the observing civ can't see anyway. I have, however, not thoroughly tested this, so this might lead to weird behaviour.

This PR is fully backwards compatible, but it might have merge conflicts with the Autocracy and Freedom branches.


SomeTroglodyte:
> code provided by

Did I? Already forgot... :older_man: :thought_balloon: :stop_sign: 

> been made incompatible

Nicely done. Is it time to close #4068 yet?

tileFilter for a resource without visibility check might work for ruleset/ map generator placement stuff - no idea if there is any - but you're right there was a contradiction there. `civInfo == null && resource == filter || civInfo != null && hasViewableResource(civInfo) && resource == filter` - that I would understand...
OK, one possible effect of your change would be if mad modder sets a strategic resource as startBias - my version would cover that.
OR - mad modder uses ""+[100]% Strength if within [99] tiles of a [Coal]" -> call omits civInfo. Same for "+[]% defence in [] tiles" - should probably patch the call to include civ
OR - (improvement) "Can only / Cannot be built on [] tiles" -> those must be called without civInfo or else possible contradiction... (as is - yes - if mad modder uses those testing for a strategic resource you get premature information disclosure, but better than allowing an improvement first and then disallowing after some research...)

I could of course be typing utter nonsense. In that case - Sorry.

ravignir:
What about Patronage branch?

xlenstra:
> What about Patronage branch?

Comming up, that's going to be the next thing I'm working on

ravignir:
I think only the Great Person policy needs to be implemented, afaik all others are already functioning.

yairm210:
Remember to resize all icons to 50*50 px

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -312,31 +317,33 @@ class CityStats {
...
+    private fun constructionMatchesFilter(construction: IConstruction, filter: String): Boolean {
+        if (construction is Building) return construction.matchesFilter(filter)
+        if (construction is BaseUnit) return construction.matchesFilter(filter)
+        return false
     }
 
+    // WHY is this function in this class? It doesn't have to do anything with a city's stats,
+    // only whether or not it is connected to the capital? Shouldn't this just be in cityInfo????
+    // Yes, there are properties depending on whether or not this is the case, but it is in itself
+    // not a stat, so those properties should just call cityInfo.isConnectedToCapital() instead.
```
yairm210, line 346: Not sure, tbh. Could go either way in my opinion.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -276,13 +276,19 @@ open class TileInfo {
...
-        val stats = improvement.clone()
+        val stats = improvement.clone() // clones the stats of the improvement, not the improvement itself
         if (hasViewableResource(observingCiv) && getTileResource().improvement == improvement.name)
             stats.add(getTileResource().improvementStats!!.clone()) // resource-specific improvement
 
         for (unique in improvement.uniqueObjects)
             if (unique.placeholderText == "[] once [] is discovered" && observingCiv.tech.isResearched(unique.params[1]))
                 stats.add(unique.stats)
+        
+        for (unique in observingCiv.getMatchingUniques("[] from every [] improvement")) {
```
yairm210, line 287: I think we can stay with [] from every [],
We just need to compare only to the improvement name.

yairm210:
Please update the wiki uniques page to accommodate all the changes in uniques you've made so far :)

yairm210:
In all PRs

xlenstra:
> In all PRs

Might If I do this later? After finishing all the policy branches, I was planning on going through all the ```matchesFilter()``` functions of all classes to convert them to the same style I've been changing them during the policy PR's, and during that I could also update the wiki will all of the new filter values & uniques

xlenstra:
> Please update the wiki uniques page to accommodate all the changes in uniques you've made so far :)

Should be done

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -482,47 +486,65 @@ class CityStats {
...
-        val maintenanceUniques = citySpecificUniques.filter { it.placeholderText == maintenanceUniqueTemplate } +
-                cityInfo.civInfo.getMatchingUniques(maintenanceUniqueTemplate).filter { cityInfo.matchesFilter(it.params[1]) }
-        for (unique in maintenanceUniques) buildingsMaintenance *= (1f - unique.params[0].toFloat() / 100)
+        // e.g. "-[50]% maintenance costs for buildings [in this city]"
+        for (unique in getUniquesForThisCity("-[]% maintenance cost for buildings []", citySpecificUniques)) {
+            buildingsMaintenance *= (1f - unique.params[0].toFloat() / 100)
+        }
+
+        // Deprecated since 3.15
+            for (unique in getUniquesForThisCity("-[]% building maintenance costs []", citySpecificUniques)) {
```
yairm210, line 528: I'm pretty sure you created this unique, no? Then you can just delete it

xlenstra: According to git blame, this was here before

yairm210:
Okay, looks mergable!

xlenstra:
Feel free to merge it

PR #4160, opened 2021-06-16
Fixed honor policy not adding bonus vs barbarians

xlenstra:


yairm210:
I think we can just add the uniques together in line 23

PR #4167, opened 2021-06-16
French translation: fix typos and translate some

MayeulC:
I don't have time to finish translating RN, you can either merge, or wait for further improvements to come.

### DIFF
--- a/android/assets/jsons/translations/French.properties
+++ b/android/assets/jsons/translations/French.properties
@@ -601,7 +601,7 @@ Build [building] = Construire [building]
 Train [unit] = Former [unit]
 Produce [thingToProduce] = Produire [thingToProduce]
 Nothing = Rien
-Annex city = Cité annexée
+Annex city = Annexer la ville
 Specialist Buildings = Bâtiments spécialisés
 Specialist Allocation = Affectation de spécialiste
 Specialists = Spécialiste
@@ -1126,7 +1126,7 @@ National Epic = Épopée nationale
 
 Market = Marché
 
-National Treasury = Trésorerie national
+National Treasury = Trésorerie nationale
 
 Provides 1 extra copy of each improved luxury resource near this City = Produit un exemplaire supplémentaire de chaque ressources de luxe à côté de cette ville
 Bazaar = Bazar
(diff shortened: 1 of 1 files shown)


android/assets/jsons/translations/French.properties
```diff
@@ -713,23 +692,20 @@ mounted units = unités montées
...
- # Requires translation!
-Bonus resource = 
- # Requires translation!
-Strategic resource = 
- # Requires translation!
-Luxury resource = 
+Fresh water = Eau douce
+non-fresh water = Eau salée
+# That's an adjective. What is it applied to? I need to find out whether it's masculine.
+Coastal = Côtier(e)
```
MayeulC, line 705: I would really like to know the gender. Is that a Coastal tile? Resource? Or should I keep it generic?

xlenstra: Based on the context (but I don't know this for sure) I would say that it would appear in contexts like: ```+2 food from coastal tiles```, where coastal means land tiles adjacent to the coast.

MayeulC:
https://github.com/yairm210/Unciv/pull/4167/files#diff-6e923159c026fd23144f896837e0f75f0d2ab4dc76c1a70bb539dbdee9512527L1138

https://github.com/yairm210/Unciv/blob/4d5ccf30b8618f12fdccad7f993807f254237894/android/assets/jsons/translations/French.properties#L1091

Somehow "…'" appears as a target icon in-game.

yairm210:
I used that as a 'nobody actually  uses this I can use it for my own purposes' char, I'm genuinely surprised to see it used. Is this common in French?

MayeulC:
> I used that as a 'nobody actually uses this I can use it for my own purposes' char, I'm genuinely surprised to see it used. Is this common in French?

Well, the [ellipsis](https://en.wikipedia.org/wiki/Ellipsis#In_French) is more common in French than in English. Using the proper unicode char looks better than just writting three dots, most word processors will do it automatically, and I try to do it manually when translating.

However, it looks like a mistake here, as it shouldn't appear after a final dot, but replace it.

> I can use it for my own purposes' char

Isn't that what [private unicode areas](https://en.wikipedia.org/wiki/Private_Use_Areas) are for? You could even install a unciv font on your system :)

PR #4174, opened 2021-06-16
Change improvement picker key indicators to tooltip

SomeTroglodyte:
One existing set of "(K)" key indicators I overlooked now use tooltips.

Simple enough - buuuuut - not affecting _this_ PR - I want to allow longer tooltip texts, and need advice. In the following clip I have two tooltips showing "W" - the bottom one is the type merged in master now. The middle one (and top) use a different approach.

https://user-images.githubusercontent.com/63000004/122274248-a4ef2980-cee2-11eb-8ae8-83933428f8a2.mp4

Note how the bottom tip is perfectly round and centered but pixelated - and the middle one is not quite perfectly round but antialiased. The bottom tip can scale (not shown) linearly to more or less any size, while the new variant won't scale smaller cleanly while larger scales would make it a rounded rectangle. You guessed it, no longer an Image.surroundWithCircle but a Label on a getRoundedEdgeRectangle - but - the original of that wouldn't do it nicely at all so I had to create a dupe right down to the backing png. Also, for the new one the ratio of space reserved for descenders becomes relevant for visually pleasing centering, and is now empirically hacked, so scaling up may also affect that.

Questions:
- Is it even worth the effort to allow longer text in tooltips?
- The discrepancy smooth/pixelated is more eye-catching elsewhere - there are surroundWithCircle's inside a ninepatch button, and you _do_ see it. Ideas what to do about that? _Can_ a scaled circle be forced to antialias in Gdx at all?
- If "yes we want all that" - the half-radius getRoundedEdgeRectangle should also go in ImageGetter, yes? Separate fun or by parameter?

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/ImprovementPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/ImprovementPickerScreen.kt
@@ -14,6 +14,7 @@ import com.unciv.models.ruleset.tile.TileImprovement
 import com.unciv.models.stats.Stats
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.*
+import com.unciv.ui.utils.StaticTooltip.Companion.addStaticTip
 import kotlin.math.round
 
 class ImprovementPickerScreen(val tileInfo: TileInfo, val onAccept: ()->Unit) : PickerScreen() {
@@ -81,7 +82,6 @@ class ImprovementPickerScreen(val tileInfo: TileInfo, val onAccept: ()->Unit) : 
             }
 
             var labelText = improvement.name.tr()
-            if (shortcutKey != null) labelText += " ($shortcutKey)"
             val turnsToBuild = if (tileInfo.improvementInProgress == improvement.name) tileInfo.turnsToImprovement
             else improvement.getTurnsToBuild(currentPlayerCiv)
             if (turnsToBuild > 0) labelText += " - $turnsToBuild${Fonts.turn}"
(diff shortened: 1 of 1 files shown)


yairm210:
I think the background for the longer text could be exactly like the background for the label, or maybe the notifications
We probably don't need to make another ninepatch

yairm210:
And honestly, if it's just small visual discrepancies, we don't need to fix every tiny thing, we can leave some as 'wontfix'

SomeTroglodyte:
Okay. Sorry for delaying this.

- I originally tried to force the Drawable getRoundedEdgeRectangle delivers down to fit the smaller tooltips - with the label scaled via fontsize. Possible, but gives ugly elongated ovals at best.
- I tried setScale on a wrapper - insubordination, size unaffected - and gave up on that line of thought and went back to approach 1 and made it work - which as you saw required a smaller PNG & new ninepatch.
- The secret to the notifications is that ScrollPane actually _obeys_ setScale while Table does _not_. That can be made to work for tooltips but of course runs into a complication - the coordinate system origin of the scaled ScrollPane is no longer on any of its corners, which seems to require an un-intuitive positioning calculation... Additionally, while Notifications also look not perfectly aligned vertically (descenders being given more room than ascenders) it's barely noticeable, but for these tootips it is imho a little too visible, especially as the text would often have no descenders. I'll have to mull all that over.

Thanks for providing the nudge in a better direction.

Meanwhile, the code in this PR is completely independent as it doesn't touch the Tooltip implementation at all.

SomeTroglodyte:
Purely FYI:
 I'm hitting some hard to understand roadblocks with the label-based tooltips.
- "ScrollPane actually obeys setScale while Table does not" was of course silly, Table just has a different default for `isTransform`.
- The most lightweight structure would be a scaled, rounded-background Gdx `Container` containing the label. Works except for positioning.
- And that's my roadblock: Trying a tamed ScrollPane, or a Table, or a Group, or a Container as Tooltip widget all work but place the tip at _varying_ offsets from the position it should go and I can't find any consistent correction function. Yes, changing containers moves the tips, and I haven't figured out why - the only possible answer is all those containers use differing transformations for their `localToStageCoordinates` - but why should they?

Next thing I might try is rewriting the system instead of subclassing it, but I'll need some courage for that I'm not willing to muster today.

PR #4186, opened 2021-06-18
Created Patronage policy branch

### LINKED ISSUE #4068: Some of social policies are different than in vanilla civ5

xlenstra:
Finally implements the last policy branch: fixes #4068 (except piety, but we need faith for that).
This PR implements the patronage branch and all the policies in it.
The Order branch should be merged before this, as it implements visual changes required to make this branch look good.

Added uniques:
- "[relationFilter] City-States provide [stat] equal to [int]% of what they produce for themselves"
- "Happiness from Luxury Resources gifted by City-States increased by [int]%"
- "Triggers a global alert" *
- "Influence of all other civilizations with all city-states degrades [int]% faster"
- "Allied City-States will occasionally gift Great People"

`*` Sends a global alert containing the name of the policy to all other players upon acquiring the policy. 

Deprecated uniques:
- "Allied City-States provide Science equal to []% of what they produce for themselves"
 
It is backwards compatible with respect to the recently removed 'Patronage' policy, but does this by introducing a
'Patronage ' (yes, with an extra space) as the name of the new policy.
This is ugly, but it works.
Other then that, there should be no problems upgrading save files to after this PR is merged.

I'm not sure whether the relationFilter actually adds anything, especially as it only checks for a specific state (e.g.: Allied), and not something like 'at least Friends' or 'at most Enemies', which would probably be more useful.
If it doesn't make sense, I can remove it again.


SomeTroglodyte:
> good solution for this too

No, but a bad one: Leave the "key" in code aching, and use English.properties to give it a display string...

ravignir:
```
			"name": "Patronage Complete",
			"uniques": ["Influence of all other civilizations with all city-states degrades [33]% faster", 
			"Triggers a global alert"],
			// This triggers a global alert in the G&K game also, based on my memory of playing it
```
y

ravignir:
You can base GP timers for city states on how game speed change other values;
Epic is 1.5x of a Standard and Marathon is 3x of a Standard.

xlenstra:
> You can base GP timers for city states on how game speed change other values;
> Epic is 1.5x of a Standard and Marathon is 3x of a Standard.

Good idea, I have implemented this

android/assets/jsons/Civ V - Vanilla/Policies.json
```diff
@@ -196,29 +203,50 @@
...
+				// You will then receive a random great person from a random city-state you are allied to
+				// The very first time after acquiring this policy, the timer is set to half of its normal value
+				// This is the basics, and apart from this, there is some randomness in the exact turn count, but I don't know how much
+				
+				// There is suprisingly little information findable online about this policy, and the civ 5 source files are
+				// also quite though to search through, so this might all be incorrect.
+				// For now this mechanic seems decent enough that this is fine.
+				
+				// Note that the way this is implemented now, this unique does NOT stack
+				// I could paramatrize the [Allied], but eh.
```
yairm210, line 237: This is a good comment, but should go in the code, not the json

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -225,4 +229,21 @@ class PolicyManager {
...
+        var extraNotificationText = ""
+        // Not only is this an anti-pattern, it is the wrong anti-pattern :D
+        if (policy.name == "Patronage Branch Complete") {
+            extraNotificationText = "\nThis causes our influence with City-States to drop faster!"
+            // Preferably this should go in the JSON file, but I don't feel at ease editing such basic data types as the Policy class yet.
+        }
+        for (civ in civInfo.gameInfo.civilizations) {
+                if (civ == civInfo) continue
+            if (civ.getKnownCivs().contains(civInfo)) {
+                civ.addNotification("${civInfo.civName} has adopted the ${policy.name} policy$extraNotificationText", NotificationIcon.Culture)
```
yairm210, line 243: This is tricky. You're adding two strings together for the notification, you can't .tr() the two strings before but also you need to ensure that the final text is translatable.
I would probably go with wrapping both texts in {}, like b = "{b...}", finalText= "{a...}$b"
Also I would go with var notificationText = if() "..." else "..."; civ.addNotification("$notificationText}$extraNotificationText")

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -225,4 +230,26 @@ class PolicyManager {
...
+            extraNotificationText = "\n${extraNotificationText}"
+        }
+        for (civ in civInfo.gameInfo.civilizations) {
+            if (civ == civInfo) continue
+            val defaultNotificationText = 
+                if (civ.getKnownCivs().contains(civInfo)) {
+                    "[${civInfo.civName}] has adopted the [${policy.name}] policy"
+                } else {
+                    "An unknown civilization has adopted the [${policy.name}] policy"
+                }.tr()
```
yairm210, line 264: No, don't .tr() this because then it will be 'stuck' in the language of the notification generator instead of being translated for each player individually

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -225,4 +230,26 @@ class PolicyManager {
         }
         return freeBuildings
     }
+
+    private fun triggerGlobalAlerts(policy: Policy) {
+        // Preferably this should go in the JSON file, but I don't feel at ease editing such basic data types as the Policy class yet.
+        var extraNotificationText = when (policy.name) {
```
yairm210, line 249: This should be by unique name, not by policy name

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -713,6 +742,22 @@ class CivilizationInfo {
         addNotification("[${otherCiv.civName}] gave us a [${militaryUnit.name}] as gift near [${city.name}]!", locations, otherCiv.civName, militaryUnit.name)
     }
 
+    /** Gain a random great person from a random city state */
+    private fun gainGreatPersonFromCityState() {
+        val givingCityState = getKnownCivs().filter { it.isCityState() && it.getAllyCiv() == civName}.random()
+        val giftedUnit = getGreatPeople().random()
```
yairm210, line 748: This should be givingCityState.getGreatPeople() so you don't get your own unique GP from a city state

core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -238,4 +243,26 @@ class PolicyManager {
...
+            // Yes, this extra space is intentional, see the policies.json file.
+            "Patronage  Complete" -> "This causes our influence with City-States to drop faster!"
+            else -> ""
+        }.tr()
```
yairm210, line 253: No .tr() here either

yairm210 (changes requested):

ravignir:
on a topic of city states -  here is african forest elephant from Almaty:
obraz
there is no Carthage in game and i play as Greece. And another unique unit next turn:
obraz
obraz

but most importantly here is a great person gift from a city state:
obraz
Khan. **City States can give you any Great Person including the unique ones** which is often omitted on wiki/ in th internet sources.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -713,6 +742,22 @@ class CivilizationInfo {
...
+        val placedUnit = placeUnitNearTile(cities.city1.location, giftedUnit.name)
+        if (placedUnit == null) return
+        val locations = LocationAction(listOf(placedUnit.getTile().position, cities.city2.location))
+        addNotification( "[${givingCityState.civName}] gave us a [${giftedUnit.name}] as a gift!", locations, givingCityState.civName, giftedUnit.name)
+    }
+
+    fun turnsForGreatPersonFromCityState(): Int = (40 + -2 + Random().nextInt(5)) * gameInfo.gameParameters.gameSpeed.modifier.toInt()
+        // There seems to be some randomness in the amount of turns between receiving each great person,
+        // but I have no idea what the actual lower and upper bound are, so this is just an approximation
+
```
ravignir, line 759: This can be found in civ5 files along with the entire function, but i forgot what the name of that thing was. Search through file contents and you should find it. 

 https://github.com/Gedemon/Civ5-DLL/blob/master/CvGameCoreDLL_Expansion1

I believe line 11540 in cvplayer.cpp file, but it is mentioned and modified in other files.

Issue #4185, opened 2021-06-17
Option to switch to old UI scaling


avdstaaij:
A while ago, the UI scaling logic was changed to be much sharper and less anti-aliased. I'm not sure when exactly this was done and whether it was intentional. Personally, I think the new scaling looks significantly worse, especially on the larger resolutions you'd use on desktop. I can see people preferring the new scaling though, so I would like to have an option to use the old scaling.

Here is a comparison of the old and new scaling (which may be of course be altered by whatever compression GitHub uses):
![unciv-scaling-comparison](https://user-images.githubusercontent.com/32672293/122485492-58d1e100-cfd7-11eb-98d8-c36bce1eb6c1.png)

xlenstra:
If I'm not mistaken, this was changed in #4080

avdstaaij:
Seems like this was unintentional, then.

ArchDuque-Pancake:
I would prefer a middlepoint instead

avdstaaij:
I don't think a middle point is possible, since the difference is essentially nearest neighbor filtering versus linear mipmap filtering. There is not really a continuous scale between the two.

PR #4187, opened 2021-06-18
Make atlas textures use mipmap filtering again

avdstaaij:
In #4080, the texture atlases were changed to use nearest neighbor filtering instead of mipmap filtering. This resulted in the visual change shown in #4185. Using nearest neighbor filtering for the texture atlases makes the game look jaggier, and in my opinion, significantly worse. This PR makes the atlases use mipmap filtering again (the "Old" version from #4185).

@lishaoxia1985, the author of #4080, stated that mipmap filtering makes no sense for texture packs. I believe this is not correct, as can be seen in #4185. However, they are right in that mipmap filtering can cause a bug when used for texture packs (as they stated in #3568): it can cause the edges of the individually packed images to bleed into each other. This effect was indeed present before #4080, and can be observed by running an older version of the game. The image below shows this (there is a line above the granary icon):

Screenshot from 2021-06-18 04-15-50

To mitigate this effect, I added some additional padding to the atlas images and set the padding to duplicate the edges of the images. This seems to fully eliminate the bleeding effect. Note that although the edge duplication makes some of the atlas images look very strange, the in-game result is improved.

Finally, I realize that the filtering method remains a matter of taste. One user has already stated in #4185 that they would prefer a midway between the old and new filtering (which I believe to be impossible). I certainly advocate for mipmap filtering, though.


avdstaaij:
Note that this doesn't undo the button improvements from #4080

yairm210:
I see.
I hope we won't have any further back and forth on this subject, nice save with the padding!

PR #4190, opened 2021-06-18
Russian Translation

sockybob:
Translated new texts, fixed some incorrect translations and made small improvements.


sockybob:
The error says i have two [stat] parameters in my translation but it's same for original english text. Probably a bug?

sockybob:
It's on 2887 line

yairm210:
I fixed the broken unique line

Issue #2788, opened 2020-07-02
no diplomatic penalty from cutting trades short (easily abused to take the AI's money)


refola:
**Describe the bug**
When a trade is cut short (e.g., from losing access to a luxury), the AI accepts it without protest. Intentionally doing this repeatedly allows efficient farming of the AI's money / tech / luxuries.

**To Reproduce**
1. Trade with an AI for all their gold and tech, offering enough gold-per-turn that it would be worth it.
2. Also put all of one of your luxuries into the trade. (They don't have to want it. Just any luxury that you can pillage.)
3. Wait for the AI to accept the trade.
4. Pillage one instance of that luxury in your land.
5. Wait one turn for the trade to be cut short.
6. Repeat for the other AIs (or with the same AI if you didn't have enough gold-per-turn to get everything in one go).
7. Trade gold (not gold-per-turn) with AIs for any luxuries and open borders you want.
8. Repeat 1-5 to get your gold back.

**Expected behavior**
If a trade is cut short, I expect some sort of penalty. I definitely shouldn't be able to get out of all my obligations after the AI has already fulfilled their end of the bargain upfront.

**Additional context**
I remember Civ5 having a diplomacy bonus for something like "our trade relations have been just and forthright". I expect that respective diplomatic penalties and bonuses for cancelled and completed trades would be the best fix.

ravignir:
I don't remember this in Civ5. I think you were unable to pillage your own resources.

logicminimal:
> I don't remember this in Civ5. I think you were unable to pillage your own resources.

i saw this exact thing in a civ 5 bug video, only difference being that the player builds a fort on the luxury instead of pillaging it

logicminimal:
technically this is the case in civ 5 but i think it is a bad mechanic and there should absolutely be penalties for missing a trade short, aside from precautions like only being able to trade gpt when not friends etc

yairm210:
This has been done a while back, forgot to close it

PR #4204, opened 2021-06-19
AI will never pillage their own tiles anymore

### LINKED ISSUE #4203: AI pillages their own tiles.
AI, and mostly city-states, pillage and destroy their own tiles very often to heal, or even just because. They pillage even their own luxuries and resources without much logic, as seen in the save files.

Save files (Latin american civs.0.9):

https://1drv.ms/u/s!AorYSegPgGoLgl9IpCDo2IZMd45K

xlenstra:
In some cases it might be better to pillage your own tiles, but more often than not it is better to not do this.
This PR makes it so AI never pillage their own tiles.
Fixes #4203.

Feel free to discuss if this change makes sense, and in which cases this makes the AI worse, I'll update the PR if need be.

### DIFF
--- a/core/src/com/unciv/logic/automation/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/UnitAutomation.kt
@@ -223,8 +223,9 @@ object UnitAutomation {
         if (unit.type.isCivilian()) return false
         val unitDistanceToTiles = unit.movement.getDistanceToTiles()
         val tilesThatCanWalkToAndThenPillage = unitDistanceToTiles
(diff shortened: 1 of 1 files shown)


logicminimal:
what if they are in a war and they need some lifesaving hp? what about putting an exception for wars also if the enemy is near?

xlenstra:
This PR only makes it so AI won't do it, not that it is impossible, I could do that as well though

avdstaaij:
Making self-pillaging impossible would also make #2788 harder to abuse.

PR #4206, opened 2021-06-20
Fix potential corrency problem

### LINKED ISSUE #4198: Gold wasted while upgrading mid-turn end.
**Describe the bug**
Upgrading a unit while waiting for a turn to end results in the gold being consumed, but the upgrades not being carried over.

**To Reproduce**
1. End Turn on a slow game (turn 600, large map)
2. Click on Warrior
3. Upgrade to Swordsmen
4. Turn ends, swordsmen convert back to the warrior, and the gold is still consumed.

**Expected behavior**
I would expect the upgraded units to remain upgraded, or at the least not being able to be upgraded when waiting for your turn to end.

**Additional context**
This also applies when purchasing a building in a city or purchasing a unit; the gold remains consumed once the turn is over and the unit does not spawn.

xlenstra:
This probably fixes #4198, by making the copy of the complete gameInfo on the main thread, so it cannot be changed midcopy, resulting to the behaviour described in this problem.
NOTE: Merging this may result in a small window in time in which the game is unresponsive, especially for larger longer games. 
That might be enough reason to close this PR already.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -578,8 +578,8 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Cam
         shouldUpdate = true
 
 
+        val gameInfoClone = gameInfo.clone()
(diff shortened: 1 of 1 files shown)


yairm210:
Nope, this can cause Application Not Responding crashes on Android.
The solution is to remove the option to conduct changes when turn is... turning

PR #4207, opened 2021-06-20
Added Nuclear Fusion tech and the Giant Death Robot

avdstaaij:
The game had a distinct lack of giant death robots, which obviously needed rectification.

I have also reordered all units in the json in accordance with the tech tree order.


logicminimal:
wow great work man

PR #4211, opened 2021-06-20
Added Advanced Ballistics Tech, Atomic Bomb Unit, Updated how nukes work

xlenstra:
This PR implements the Advanced Ballistics Tech, the Atomic Bomb Unit and reworks how nukes work.
Nukes now have strength level indicated in the unique, which determines how much damage nukes do exactly.
Also, the exact amount of damage they do to units, how much fallout is placed, removal of forests/jungles, etc. is updated according to:
- https://forums.civfanatics.com/resources/unit-guide-modern-future-units-g-k.25628/
- https://forums.civfanatics.com/threads/unit-guide-modern-future-units-g-k.429987/#2
- https://www.carlsguides.com/strategy/civilization5/units/aircraft-nukes.php
- Testing done by Ravignir

Lastly, missiles are no longer considered air units, mostly because they cannot be stationed on aircraft carriers but on nuclear submarines, and because they cannot be intercepted. Instead, missiles are now their own type of unit. Implementing this change required a lot of small changes in many files, which increased the files changed by a lot, even though the changes themselves are relatively few.

Added uniques:
- "Nuclear weapon of strength [amount]" - amount is 1 or 2
- "Blast radius [amount]" 
- "Resistant to nukes"
- "Can be destroyed by nukes"
- "Cannot be intercepted"

Deprecated uniques:
- "Nuclear weapon"


ravignir:
while we are at it: what about the way Fallout is generated (should permamently remove forests/jungles, has a random chance to appear)? Also fallout does not reduce yields in civ5, it just makes the tile unworkable.

ravignir:
If i can add one thing: atomic bomb is actually a bit stronger: the 60% chance to destroy all civilian units in blast radius is probably innacurate. I've run some tests in G&K and civilians were all destroyed except for the 3 on that edge (not to mention archers did not take any damage), so i assume it is a bug  and it should destroy all of them: 
obraz

Also the damage dealt to units seems to be 75+/- 30, that is a rough approximation tho.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -383,60 +393,179 @@ object Battle {
...
+        if (city != null && city.location == tile.position) {
+            var populationLoss = 1f;
+            for (unique in city.civInfo.getMatchingUniques("Population loss from nuclear attacks -[]%")) {
+                populationLoss *= 1 - unique.params[0].toFloat() / 100f
+            }
+            if (city.population.population < 5) {
+                city.destroyCity()
+            } else {
+                city.population.population = (city.population.population * (0.6 + Random().nextFloat() * 0.2) * populationLoss).toInt()
+                if (city.population.population < 5 && populationLoss == 1f) city.population.population = 5
```
yairm210, line 546: This seems very odd. If I had 5 population, that was then decreased to 3, then you increase it back to 5? Why?

xlenstra: Because on 
https://www.carlsguides.com/strategy/civilization5/units/aircraft-nukes.php
I found the following sentence:
> Bomb Shelters will not stop a Nuclear Missile's massive power from leveling a City unless it has 5 or higher Population. In that case, the population will not fall below 5. If the Population is 4, the City is gone.

So it seems to me that if you have a bomb shelter, your city should not go below 5 population. I did implement this wrong (it should be `populationLoss < 1f`), so there is something to change here, I do agree.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -383,60 +393,179 @@ object Battle {
...
+            unit.destroy()
+            postBattleNotifications(attacker, MapUnitCombatant(unit), unit.currentTile)
+            destroyIfDefeated(unit.civInfo, attacker.getCivInfo())
+        }
+        // Remove improvements
+        tile.improvement = null
+        tile.improvementInProgress = null
+        tile.turnsToImprovement = 0
+        tile.roadStatus = RoadStatus.None
+        if (tile.isLand && !tile.isImpassible() && !tile.terrainFeatures.contains("Fallout") &&
```
yairm210, line 566: It looks like unlike strength 1, we don't remove the features?

xlenstra: My bad, must have missed this, I'll fix it later today

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -160,8 +161,14 @@ class BaseUnit : INamed, IConstruction {
...
         if (civInfo.gameInfo.ruleSet.units.values.any { it.uniqueTo == civInfo.civName && it.replaces == name })
             return "Our unique unit replaces this"
-        if (!civInfo.gameInfo.gameParameters.nuclearWeaponsEnabled
-                && uniques.contains("Nuclear weapon")) return "Disabled by setting"
+        if (!civInfo.gameInfo.gameParameters.nuclearWeaponsEnabled && 
+            (
+                uniques.map{ it.getPlaceholderText()}.contains("Nuclear weapon of strength []") ||
+                // Deprecated since 3.15.3
+                    uniques.map{ it.getPlaceholderText()}.contains("Nuclear weapon")
+                //
```
yairm210, line 169: You're using this combo a lot, at least 4 times so far, I think it's OK to have BaseUnit.isNuclearWeapon()

core/src/com/unciv/logic/battle/Battle.kt
yairm210: K

PR #4217, opened 2021-06-20
Fixes bug where in some cases building improvements would increase the base yield of the tile

xlenstra:
Take the unique "[+1 Science] from every [Jungle] tile in this city".
This unique should only ever be checked when we compute the base yield of the tile.
However, due to a small oversight in #4142, it is now also checked when an improvement is on the tile.
Because of this, the bonus was applied twice.
This would lead to jungles giving 2 science when a university was build, and 4 science when any improvement was build.
This PR rectifies this.


yairm210:
I see, but removing the improvement entirely from the tileinfo filter precludes many unique options (e.g. +10% strength when adjacent to moai, improvement yield adjacency bonuses, etc).
It should stay in the filter as well.
If this means we'll have to change some of the checks around, to be only on / only not on improvements, that's doable.

xlenstra:
I believe I have fixed the problem you described, but did it in a different way then you requested (as per usual at this point), do look over if you agree with these changes. Basically, I split the `matchesUniqueFilter()` function in two new functions: `matchesTerrainFilter()` which only checks if the terrain matches (what was previously `matchesUniqueFilter()`), and `matchesFilter()`, which checks if either the terrain matches, or the improvement (if it exists) matches

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -205,7 +205,7 @@ object BattleDamage {
             }
 
             for (unique in defender.unit.getMatchingUniques("+[]% defence in [] tiles")) {
-                if (tile.matchesUniqueFilter(unique.params[1]))
+                if (tile.matchesTerrainFilter(unique.params[1]))
```
yairm210, line 208: Disagree. I want '+ defense in improvement' as well :)

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -228,22 +228,22 @@ object BattleDamage {
                 + unit.getCivInfo()
             .getMatchingUniques("+[]% Strength for units fighting in []")) {
             val filter = unique.params[1]
-            if (tile.matchesUniqueFilter(filter, unit.getCivInfo()))
+            if (tile.matchesTerrainFilter(filter, unit.getCivInfo()))
```
yairm210, line 231: Same here, and in all combat modifiers

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -235,7 +235,7 @@ open class TileInfo {
             for (unique in cityWideUniques + civWideUniques) {
                 val tileType = unique.params[1]
                 if (tileType == improvement) continue // This is added to the calculation in getImprovementStats. we don't want to add it twice
-                if (matchesUniqueFilter(tileType, observingCiv)
+                if (matchesTerrainFilter(tileType, observingCiv)
```
yairm210, line 238: [+1] Production from every [Mine] should be a valid civInfo unique, there's someone that already has something like that but I forget who

xlenstra: That would be calculated in the `getImprovementStats()` function right? This function is only for getting the yield of the base tile

PR #4219, opened 2021-06-21
Fixed open terrain bonus working in rough terrain

### LINKED ISSUE #4139: "Open" terrain bonus ignores terrain features
The current implementation looks at the underlying terrain (plains/grassland/desert) and ignores features (hills/forests/jungles). Therefore, every passable land tile fits this definition, making shock and accuracy overpowered.

avdstaaij:
Fixed #4139 - The open terrain combat bonus no longer works in rough terrain.

Deprecated "Open terrain" and "Rough terrain" terrain uniques and undeprecated terrain property "rough", effectively reverting b312d24d4a70092e340b6f3c9a468832c5c846cf (but not https://github.com/yairm210/Unciv/commit/d825352bf42181c4e49c34ec9096de1b54272b1a).

Also made the tileFilter implementation consider uniques of natural wonders, which it didn't before.

Motivation:

Tiles are either open or rough, and cannot be both at the same time. This change makes tiles open until proven otherwise - open unless they have a base terrain, natural wonder or terrain feature that is rough. This is already how it worked for other purposes such as movement costs, though natural wonders were not always considered.

An alternative would be to allow terrain features to "override" the roughness of the base terrain, thereby allowing mods with rough base terrains and features that make those terrains open again. However, this would probably be needlessly complicated, and terrains can have multiple features, so you'd have to consider what to do if a tile has both an open and a rough terrain feature.

Because of the new approach, the "Open terrain" unique wasn't used anymore (except for the terrain description in the civlopedia). Because of this, and the fact that terrain roughness is a very basic concept, I deprecated the uniques in favor of the old "rough" property. Of course, we could also keep using the uniques, or only the "Rough terrain"  unique instead, if that is preferred.

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -408,12 +408,12 @@ open class TileInfo {
             "River" -> isAdjacentToRiver()
             improvement -> true
             naturalWonder -> true
+            "Open terrain" -> !isRoughTerrain()
+            "Rough terrain" -> isRoughTerrain()
             "Foreign Land" -> civInfo != null && !isFriendlyTerritory(civInfo)
             "Friendly Land" -> civInfo != null && isFriendlyTerritory(civInfo)
             else -> {
                 if (terrainFeatures.contains(filter)) return true
-                if (filter == "Open terrain" && !isRoughTerrain()) return true
-                if (filter == "Rough terrain" && isRoughTerrain()) return true
                 if (hasUnique(filter)) return true
                 if (resource != null && getTileResource().resourceType.name + " resource" == filter) return true
                 if (civInfo != null && hasViewableResource(civInfo) && resource == filter) return true
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -201,6 +201,8 @@ open class TileInfo {
         yieldAll(terrainFeatures.asSequence().mapNotNull { ruleset.terrains[it] })
     }
 
+    fun isRoughTerrain() = getAllTerrains().any{ it.isRough() }
+
     fun hasUnique(unique: String) = getAllTerrains().any { it.uniques.contains(unique) }
 
     fun getWorkingCity(): CityInfo? {
@@ -410,8 +412,9 @@ open class TileInfo {
             "Friendly Land" -> civInfo != null && isFriendlyTerritory(civInfo)
             else -> {
                 if (terrainFeatures.contains(filter)) return true
-                if (baseTerrainObject.uniques.contains(filter)) return true
-                if (terrainFeatures.isNotEmpty() && getTerrainFeatures().last().uniques.contains(filter)) return true
+                if (filter == "Open terrain" && !isRoughTerrain()) return true
+                if (filter == "Rough terrain" && isRoughTerrain()) return true
+                if (hasUnique(filter)) return true
                 if (resource != null && getTileResource().resourceType.name + " resource" == filter) return true
                 if (civInfo != null && hasViewableResource(civInfo) && resource == filter) return true
                 return false
--- a/core/src/com/unciv/models/ruleset/tile/Terrain.kt
+++ b/core/src/com/unciv/models/ruleset/tile/Terrain.kt
@@ -36,22 +36,19 @@ class Terrain : NamedStats() {
     var defenceBonus:Float = 0f
     var impassable = false
 
-    /**
-     * Do not use this directly! Use isRough() instead.
-     * Modded tiles may use the deprecated "Rough terrain" unique instead of this value, which is handled by isRough().
-     */
+    /** Use isRough() instead */
+    @Deprecated("As of 3.14.1")
     var rough = false
 
     fun isRough(): Boolean {
-        // "Rough terrain" unique deprecated since 3.15.4
-        return rough || uniques.contains("Rough terrain")
+        // "rough" property deprecated since 3.14.1
+        return uniques.contains("Rough terrain") || rough
     }
 
     fun getColor(): Color { // Can't be a lazy initialize, because we play around with the resulting color with lerp()s and the like
         if (RGB == null) return Color.GOLD
         return colorFromRGB(RGB!!)
     }
-
 
     fun getDescription(ruleset: Ruleset): String {
         val sb = StringBuilder()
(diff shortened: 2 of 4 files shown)


freddyhayward:
Is water considered open in civ 5?

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -410,8 +412,9 @@ open class TileInfo {
             "Friendly Land" -> civInfo != null && isFriendlyTerritory(civInfo)
             else -> {
                 if (terrainFeatures.contains(filter)) return true
-                if (baseTerrainObject.uniques.contains(filter)) return true
-                if (terrainFeatures.isNotEmpty() && getTerrainFeatures().last().uniques.contains(filter)) return true
+                if (filter == "Open terrain" && !isRoughTerrain()) return true
+                if (filter == "Rough terrain" && isRoughTerrain()) return true
```
xlenstra, line 416: Please just put this in the list above as:
"Open terrain" -> !isRoughTerrain()
"Rough Terrain" -> isRoughTerrain()

core/src/com/unciv/models/ruleset/tile/Terrain.kt
```diff
@@ -35,10 +35,17 @@ class Terrain : NamedStats() {
...
     var defenceBonus:Float = 0f
     var impassable = false
-    
-    @Deprecated("As of 3.14.1")
+
+    /**
+     * Do not use this directly! Use isRough() instead.
+     * Modded tiles may use the deprecated "Rough terrain" unique instead of this value, which is handled by isRough().
+     */
     var rough = false
```
xlenstra, line 43: It seems to me that your implementation removes the uniques "Rough Terrain" and "Open Terrain", so this variable would be the only way to tell them apart. But it isn't set anywhere? How would that work?

yairm210:
There's no logical difference between 'rough:true' and "Rough Terrain" unique, the field was deprecated for a reason.
This is a tough one: rough is 'one of the layers is rough', and open is 'none of the layers are rough'. So basically, fikterMatches("Open terrain")= not filterMatches("Rough terrain")

ravignir:
> 
> 
> Is water considered open in civ 5?

obraz
yes

xlenstra:
Can't this entire problem just be solved by putting:
```  if (filter.endsWith(" terrain")) return getLastTerrain().uniques.contains(filter) ```
at the top of the else clause in `TileInfo.matchesUniqueFilter()`?

File: core/src/com/unciv/logic/map/TileInfo.kt (hunk 1 above)
Comment (avdstaaij) [reply core/src/com/unciv/logic/map/TileInfo.kt:416]: I am ashamed of myself. This will be fixed!

File: core/src/com/unciv/models/ruleset/tile/Terrain.kt (hunk 2 above)
Comment (avdstaaij) [reply core/src/com/unciv/models/ruleset/tile/Terrain.kt:43]: It is automagically set by the json parser.

**Comment by avdstaaij** [comment]
> There's no logical difference between 'rough:true' and "Rough Terrain" unique, the field was deprecated for a reason.

There is indeed no difference. The reasons why I went for "rough:true" property instead of the "Rough terrain" unique are that terrain roughness is a very basic terrain property, not something very special, and that only terrains can have it. Essentially, the same reasons for using "movementCost:2" and "defenceBonus:0.25" instead of uniques like "Costs [2] movement points to traverse" and "Provides [25]% defense".

From the commit history, I gathered that "rough:true" was deprecated in favor of "Rough terrain" to to implement tileFilter more cleanly, but #4139 has shown that that implementation was incorrect. Since we now have to explicitly check "filter == "Open terrain" (as it means "not rough"), the implementation advantage of the unique is lost anyway. Was there any other reason the property was deprecated?

If you insist on the uniques, should there be both a "Open terrain" (which will never be used) and a "Rough terrain" unique, or only a "Rough terrain" unique?

> This is a tough one: rough is 'one of the layers is rough', and open is 'none of the layers are rough'. So basically, filterMatches("Open terrain")= not filterMatches("Rough terrain")

Indeed, which is exactly what the following achieves:
```kotlin
fun isRoughTerrain() = getAllTerrains().any{ it.isRough() }
```

> Can't this entire problem just be solved by putting:
> ` if (filter.endsWith(" terrain")) return getLastTerrain().uniques.contains(filter)`
> at the top of the else clause in `TileInfo.matchesUniqueFilter()`?

Almost, but not quite, for the following reasons:
1. I am decidedly against checks like `filter.endsWith(" terrain")` which make assumptions about future uniques. What if we later change "Friendly land" to "Friendly terrain", for example? It would match this check, leading to unintended behavior. We could however parametrize the terrain uniques and check if the filter matches "[] terrain".
2. `return getLastTerrain().uniques.contains(filter)` will give the incorrect result if the terrain contains multiple features / natural wonders, of which one is rough, and the rough one is not the last. It will also give the incorrect result if the base terrain is rough (a mod could add one) and the last feature happens not to be. We need to check all terrain layers, which `isRoughTerrain` does.
3. `uniques.contains(filter)` is in itself also not enough: even if we keep using the uniques, we should still support the previously deprecated "rough:true" property (and if we deprecate the uniques in favor of undeprecating "rough:true", as I propose, we also still need to check for both "rough:true" and "Rough terrain". I added `Terrain.isRough()` for this purpose.

yairm210:
I'm in favour of keeping the unique for multiple purposes: User visibility, translation, and the highly technical 'we already have modders change things quite often, let's limit that if possible'.
The upside of the field, is that it allows for faster checks when calculating unit movement for units that have a penalty. Which is not insignificant, but still doesn't outweigh the other considerations in my eyes.

avdstaaij:
> I'm in favour of keeping the unique for multiple purposes: User visibility, translation, and the highly technical 'we already have modders change things quite often, let's limit that if possible'.

I don't want to be too insistent, but do want to note that there should be no effect on user visibility or translation: this change should be completely transparent in those two respects. I can't refute the last point, though.

Anyway, should I restore both "Open terrain" and "Rough terrain", or only "Rough terrain"? With the "open until proven rough"-approach, the "Open terrain" unique wouldn't be used at all; there would be no difference between a terrain having it or not.

yairm210:
Only rough terrain, I think that 'open until proven otherwise' is the only way to solve this in either case...

xlenstra:
> Only rough terrain, I think that 'open until proven otherwise' is the only way to solve this in either case...

From some quick tests this does seem to be the way the main game did it

avdstaaij:
I redeprecated the "rough" property and switched back to the "Rough terrain" unique, so this is now ready.
You may also want to consider the alternative solution #4226.

avdstaaij:
> * What went wrong:
A problem occurred configuring root project 'Unciv'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not download gwt-gradle-plugin-0.6.jar (de.richsource.gradle.plugins:gwt-gradle-plugin:0.6)
      > Could not get resource 'https://plugins.gradle.org/m2/de/richsource/gradle/plugins/gwt-gradle-plugin/0.6/gwt-gradle-plugin-0.6.jar'.
         > Could not GET 'https://plugins.gradle.org/m2/de/richsource/gradle/plugins/gwt-gradle-plugin/0.6/gwt-gradle-plugin-0.6.jar'. Received status code 500 from server: Internal Server Error
         
Looks like this isn't caused by me, I'll try again in a bit.

PR #4226, opened 2021-06-21
Alternative fix for the open-rough terrain problem

### LINKED ISSUE #4139: "Open" terrain bonus ignores terrain features

xlenstra:
This is a counter-PR to #4219, with the intent of solving the same problem in a more modular way.
As it was quickly thrown together, some variables names still need to change (TerrainType2, for example), but it should work. The current version is for showing an alternative fix, that, if deemed better than the previous proposed, can be developed further into a quality PR.
Fixes #4139.

What this PR does, is create a new JSON file named 'terrainType.json', in which the different types of terrain - open, rough - are written, along with a priority. When a test is run if a certain tile is of a certain terrain type, only the terrain with the highest priority will be used. This way, there will be no ambiguity about which terrain type a certain tile has, yet still provides modders with all the tools they need to change everything they want.

Note; I spoke with @avdstaaij before making this PR, and he was fine with me proposing this alternate fix.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/TerrainTypes.json
+++ b/android/assets/jsons/Civ V - Vanilla/TerrainTypes.json
@@ -1,0 +1,10 @@
+[
+    {
+        "name": "Rough terrain",
+        "priority": 1
+    },
+    {
+        "name": "Open terrain",
+        "priority": 0
+    }
+]
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -6,14 +6,16 @@
 		"food": 1,
 		"gold": 1,
 		"movementCost": 1,
-		"RGB": [45,108,145]
+		"RGB": [45,108,145],
+		"uniques": ["Open terrain"]
 	},
 	{
 		"name": "Coast",
 		"type": "Water",
 		"food": 1,
 		"movementCost": 1,
-		"RGB": [107,167,193]
+		"RGB": [107,167,193],
+		"uniques": ["Open terrain"]
 	},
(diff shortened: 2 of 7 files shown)


avdstaaij:
As we touched upon in our conversation, the main tradeoff of this more modular approach is that it allows for terrains to be neither open nor rough. I imagine that this may cause unforeseen problems in the future. For example, one immediate concern is the implementation of the "Rough terrain penalty" unique, which the [chariot archer](https://civilization.fandom.com/wiki/Chariot_Archer_(Civ5)) has. Does it mean the chariot archer loses all movement points when entering rough terrain? Or does it mean that it does so when entering non-open terrain? We can't know what the original game intended, since it has only two terrain types. Then again, these kinds of problems may be minor.

xlenstra:
> it allows for terrains to be neither open nor rough

> We can't know what the original game intended

While the proposed change would in theory allow this, this would never happen in unmodded unciv, as there all base terrains all have the 'open terrain' terrainType, so there would be no tiles that are neither open nor rough. In contrast, in modded unciv this would be possible. But implementing uniques "[+1] Movement in all [Rough terrain] tiles" and "[+3] Movement in all non-[Open terrain] tiles" is not very difficult, and gives modders full controll over this mechanic that is otherwise inaccessible.

yairm210:
Merged the other solution.
If possible, I prefer not to add new files, because that means we need to go through the whole ruleset loading, error checking, and translating yadda, and if we can replace that with a different unique and some lines of code then that's definitely preferable.

PR #4227, opened 2021-06-21
Make permanent visual mods into perm. audiovisual mods

SomeTroglodyte:
I've been mulling for a while on this - since I got no better inspiration I'm offering it in this form.

Overriding sounds with a mod currently works, if the mods in question are active in the running game, that is, selected on the new game screen.

This puts mods selected as "permanent visual mods" at the front of Sounds.kt search order, even from the main menu, so such a mod can override sounds even after starting a game without it.

Correspondingly, I've renamed the text in the buttons to "En/Disable as permanent audiovisual mod" and since that was getting a bit long I made the button wrap it.

The UI change is maybe not the best choice - input welcome. Perhaps "permanent media mods"?

Note: Will collide with #4033 - but that one is old enough now to redo anyway.


yairm210:
I think we should change the button info a checkbox. Just "Permanent audiovisual mod", with checkbox.

SomeTroglodyte:
Ha - saving space by replacing "En/dis-able", hadn't thought of that. Say, all the current "checkboxes" have their text pratically touching the "box" circle - to get a little separation here is a matter of the skin definition, not of adding a space in front of all the checkbox labels, right? I'll look at the code in a few hours.

PR #3460, opened 2020-12-27
genericise + population unique

9kgsofrice:
genericise + population in all cities unique
add + population in this city unique

PR #4246, opened 2021-06-23
Implements the Telecommunications tech

xlenstra:
This PR implements the telecommunciations tech, and two of it's buildings:
The Bomb Shelter and the CN Tower
WIP: Implement nuclear submarine & Missile Cruiser.
This PR depends on #4211 for the implementation of the unique of the Bomb Shelter
This PR replaces #3460 as it implements the unique in the requested way for use in the CN tower.
It also creates a function `Populationmanager.addpopulation` which adds population and then automatically assignes extra population/unassignes overflow population.

Uniques added:
- +[amount] population [cityFilter]


xlenstra:
This PR is starting to be a mess, I'm closing it and will reopen it once I've fixed everything.

PR #4247, opened 2021-06-23
Promotion picker keeps vertical scroll pos on promote or resize

### LINKED ISSUE #4189: Make the Promotion screen not go back to the top of the page when you promote an unit
when you promote an unit the screen goes all the way up which is frusturating when you rapidly try to upgrade the units. this issue, obviously is not the most important thing and the people directly working on the game shouldnt tske their time to fix this but it would be cooled if someone just fixed this in their free time and merged it

SomeTroglodyte:
Resolve #4189


yairm210:
Disagree, this doesn't need to be a parameter
We can create the screen as before, and then set the value of the scrollpane of the new screen after creation and before setting it as the current screen

SomeTroglodyte:
You mean because both objects exist at the same time? OK, will consider this later.

PR #4249, opened 2021-06-23
Implements the Telecommunications tech (v2)

xlenstra:
This PR implements the telecommunciations tech, and two of it's buildings:
The Bomb Shelter and the CN Tower
This PR depends on #4211 for the implementation of the unique of the Bomb Shelter
This PR replaces #3460 as it implements the unique in the requested way for use in the CN tower.
It also creates a function `Populationmanager.addPopulation()` and a function `Populationmanager.setPopulatoin()` which add and set population respetively and then automatically assign extra population/unassigne overflow population.

Uniques added:
- [amount] population [cityFilter]

Uniques depcreated:
-  "+1 population in each city"


yairm210:
Please change to [+1] for unique, so it can also handle negative numbers ;)

avdstaaij:
This has already been merged, but CN tower should provide a free Broadcast tower in every city, not just its own city.

Furthermore, the wiki states the following:
> Note that the Broadcast Towers will go to whichever civilization owns the CN Tower. That means that the original builder will lose the free Broadcast Towers if another player captures the city with the CN Tower (and the new owner will gain them), although the increases in Population and Happiness will only be rewarded to the builder. The additional Broadcast Towers awarded when this wonder is completed go to all cities controlled by that civilization, even if they have just been captured and are still in resistance. The free Broadcast Tower also goes to any city founded or acquired after the CN Tower has been completed, but, as when the city with the wonder switches hands, there will be no increase in population or happiness. 

Those details sound really painful to implement though, so you may want to postpone them.

PR #4252, opened 2021-06-24
Generalized building of improvements

xlenstra:
This PR generalizes the concept of building improvements. This is mostly done to combine the 'build road' and 'build fort' unit actions into the 'build improvements' unit action, but is made extendible enough that mods can implement units that only create certain types of improvements.
Another advantage of this implementation, is that the Legion unit now has an 'automate' button, which will automatically build roads between cities. While I have not tested this, it should also be possible for modded units that can only build some improvements to be automated without problems.

I was thinking about also including fishing boats & great improvements into this same system, but decided against it, as they are built immediately and therefore need a different system.

Added uniques:
- "Can build [tileFilter] improvements on tiles"

Deprecated uniques:
- "Can build improvements on tiles"
- "Can construct roads"

### DIFF
--- a/core/src/com/unciv/logic/automation/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/UnitAutomation.kt
@@ -89,7 +89,7 @@ object UnitAutomation {
             if (unit.hasUnique(Constants.settlerUnique))
                 return SpecificUnitAutomation.automateSettlerActions(unit)
 
-            if (unit.hasUnique(Constants.workerUnique))
+            if (unit.hasUnique(Constants.canBuildImprovements))
                 return WorkerAutomation(unit).automateWorkerAction()
 
             if (unit.name == "Work Boats")
--- a/core/src/com/unciv/logic/automation/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/UnitAutomation.kt
@@ -89,10 +89,11 @@ object UnitAutomation {
             if (unit.hasUnique(Constants.settlerUnique))
                 return SpecificUnitAutomation.automateSettlerActions(unit)
 
-            if (unit.hasUnique(Constants.canBuildImprovements))
+            // Constants.workerUnique deprecated since 3.15.5
+            if (unit.hasUnique(Constants.canBuildImprovements) || unit.hasUnique(Constants.workerUnique))
                 return WorkerAutomation(unit).automateWorkerAction()
 
-            if (unit.name == "Work Boats")
+            if (unit.name == "Work Boats") // This is really not modular
                 return SpecificUnitAutomation.automateWorkBoats(unit)
 
             if (unit.hasUnique("Bonus for units in 2 tile radius 15%"))
--- a/core/src/com/unciv/logic/automation/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/WorkerAutomation.kt
@@ -187,7 +187,6 @@ class WorkerAutomation(val unit: MapUnit) {
         }
 
         // turnsToBuild is what defines them as buildable
-        // ^ that sounds quite ugly
         val tileImprovements = civInfo.gameInfo.ruleSet.tileImprovements.filter { it.value.turnsToBuild != 0 }
         val uniqueImprovement = tileImprovements.values
                 .firstOrNull { it.uniqueTo == civInfo.civName }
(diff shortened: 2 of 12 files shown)


ravignir:
> 

> I was thinking about also including fishing boats & great improvements into this same system, but decided against it, as they are built immediately and therefore need a different system.
> 

What about "Can construct []" and "Can immediately construct []"?

Or just have "construct" and "build" with one being instant and the other one taking turns to complete?

core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -189,6 +187,7 @@ class WorkerAutomation(val unit: MapUnit) {
         }
 
         // turnsToBuild is what defines them as buildable
+        // ^ that sounds quite ugly
```
yairm210, line 190: Well, if every buildable improvement has to have a "turns to build" and every non-buildable improvement doesn't, that's basically a boolean determinator

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -540,12 +547,17 @@ class MapUnit {
     fun endTurn() {
         doAction()
 
-        if (currentMovement > 0 && hasUnique(Constants.workerUnique)
+        if (currentMovement > 0 && 
+            // Constants.workerUnique deprecated since 3.15.5
+            (hasUnique(Constants.canBuildImprovements) || hasUnique(Constants.workerUnique))
```
yairm210, line 552: Aren't these already checked in the canBuildImprovement below?

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -257,8 +258,14 @@ class MapUnit {
 
     fun isIdle(): Boolean {
         if (currentMovement == 0f) return false
-        if (hasUnique(Constants.workerUnique) && getTile().improvementInProgress != null) return false
-        if (hasUnique("Can construct roads") && currentTile.improvementInProgress == "Road") return false
+        // Constants.workerUnique deprecated since 3.15.5
+        if ((hasUnique(Constants.canBuildImprovements) || hasUnique(Constants.workerUnique))
```
yairm210, line 262: Aren't these already checked in the canBuildImprovement below?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -32,8 +32,10 @@ object UnitActions {
 
         if (unit.isMoving()) actionList += UnitAction(UnitActionType.StopMovement) { unit.action = null }
 
-        val workingOnImprovement = unit.hasUnique("Can build improvements on tiles")
+        // Constants.workerUnique deprecated since 3.15.5
+        val workingOnImprovement = (unit.hasUnique(Constants.canBuildImprovements) || unit.hasUnique(Constants.workerUnique))
```
yairm210, line 36: Aren't these already checked in the canBuildImprovement below?

yairm210:
This looks good! I like the distinction between "Create" and "Build" for instant vs long term :)

core/src/com/unciv/logic/automation/WorkerAutomation.kt
xlenstra: True, my first thought was to question whether this assumption will always hold in the future, but I can't think of a reasonable situation in which it shouldn't apply (building improvements in 0 turns is impossible and would be weird, could you build multiple improvements in a single turn?) so I guess its fine.

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -89,10 +89,11 @@ object UnitAutomation {
             if (unit.hasUnique(Constants.settlerUnique))
                 return SpecificUnitAutomation.automateSettlerActions(unit)
 
-            if (unit.hasUnique(Constants.workerUnique))
+            // Constants.workerUnique deprecated since 3.15.5
+            if (unit.hasUnique(Constants.canBuildImprovements) || unit.hasUnique(Constants.workerUnique))
                 return WorkerAutomation(unit).automateWorkerAction()
 
-            if (unit.name == "Work Boats")
+            if (unit.name == "Work Boats") // This is really not modular
```
yairm210, line 96: Agree, and that's a job for another time :)

PR #4254, opened 2021-06-24
Added Drama and Poetry tech and replaced Temples with Amphitheaters

avdstaaij:
This PR adds the missing Drama and Poetry technology below Philosophy. It includes the Amphitheater, which takes over the purpose Temple had in Vanilla. This means that the Temple won't be needed anymore until religion is implemented. All Temple-replacing unique buildings have been changed to temporarily replace Amphitheater instead, and provisions have been made to convert Temples in existing saves to Amphitheaters. It's probably best to phase out Temples as early as possible if we want to turn them into religious buildings later.

I have also made `GameInfo.changeBuildingName` a bit more robust.


core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -346,6 +346,12 @@ class GameInfo {
         }
 
         for (city in civilizations.asSequence().flatMap { it.cities.asSequence() }) {
+            // Temple was replaced by Amphitheater in 3.15.5. For backwards compatibility, we
+            // replace existing temples with amphitheaters. This replacement should be removed
+            // when temples are reintroduced as faith buildings.
+            if (!ruleSet.buildings.containsKey("Temple"))
```
yairm210, line 352: I think this 'if' should be part of the 'rename building' function

avdstaaij: Agreed. I didn't change it at first because it wasn't reflected by the name of the function, but I've changed the name to match its new semantics.

PR #4258, opened 2021-06-24
Translations update

yairm210:


yairm210:
summary

SomeTroglodyte:
This moved and renamed the complete turkish file, thereby deactivating it - in case that is unintentional?

PR #4260, opened 2021-06-24
Added Nuclear Submarines & Missile Cruisers, capable of transporting missiles

xlenstra:
This PR adds the Nuclear Submarine & Missile Cruiser units. 
Both these units are capable of carrying missile units, which was also implemented (rather easily, actually).

Added uniques:
- "Can carry [amount] [unitFilter] units"

- "Can carry [amount] extra [unitFilter] units"

Both of these filters can only have "Air" or "Missile" as its value

Deprecated uniques:
- "Can carry 2 aircraft"
- "Can carry 1 extra aircraft"


yairm210:
How about "Can carry [amount] [unitType] units", ""Can carry [amount] extra [unitType] units"?
The "genericifying the unit types" is a long-standing goal of mine, so when possible, I try ad make things that aren't hardcoded to a specific unit type :)

xlenstra:
> How about "Can carry [amount] [unitType] units", ""Can carry [amount] extra [unitType] units"?
> The "genericifying the unit types" is a long-standing goal of mine, so when possible, I try ad make things that aren't hardcoded to a specific unit type :)

I thought about it, but decided against it as it is quite complicated behaviour that will probably create a lot of unforseen bugs, and might also be weird. But the more I think about it, the more useful and less weird it seems. For example. allowing land units to 'embark' on a destroyer for traveling over the ocean with more protection, or infantry hopping on a low strength but fast truck to travel large distances could be cool features in mods. I might make another PR for this, but I'm not planning on implementing it in this PR. It would probably still require reworking how exactly carried units are stored, whether or not they can attack (a truck with 5 rocket artillery on the same space might be _slightly_ op), and more things to consider, which would all be beyond the scope of this PR.

yairm210:
As-is, it won't work for land or sea units, because they're saved in a different 'slot' so to speak. But I do hope eventually (?) to make a different way of storing units, where they're all stored in the same big array. This will be HELL on performance because of all the optimizations that have been made so far, and also will require figuring out how we show all these units on the map, but I hope for it to be optional in the modOptions.

PR #4261, opened 2021-06-24
add edits

rafaeldfigueiredo:
This was just some additions. There are random apostrophes that end up quoting large chunks of text until a new apostrophe is found. I don't know if this can be a problem for the technical part but for now, is just a minor pain when translating because of lack of auto-complete.


yairm210:
I'm not sure why the tests weren't running properly for your PR, but there were some errors that were caught when running on the 'all translations together' that originate here:


    Placeholder `[cityname]` not found in `Brazilian_Portuguese` for entry `The city of [cityname] has started constructing [construction]!`
    Placeholder `[cityFilter]` not found in `Brazilian_Portuguese` for entry `[stats] per [amount] population [cityFilter]`

rafaeldfigueiredo:
Hello! It was my mistake on the first one (Casing on cityname where it should be cityName)

But I don't know what happened in the other one, is identical with the placeholders and camelCase.

PR #4265, opened 2021-06-25
Added Great Admiral

xlenstra:
This PR adds the great admiral unit, including all its abilities.
It also disables great general points being earned for water combat, instead generating great admiral points for those. If your mod has neither of those, but does have other units with "Great Person - [War - Water]" or "Great Person - [War - Land]" as uniques, points are generated for those instead.
Lastly it generalizes the way combat bonuses from great admirals and great generals are calculated.

Added uniques:
- "+[amount]% Strength for [unitFilter] units within [amount] tiles" - Amount is normally at most 2, but this can be changed in modOptions
- "+[amount]% Strength vs [unitFilter] for [unitFilter] units within [amount] tiles" - see above
- "[unitFilter] units provide [amount]% their normal Strength bonus" - Only applies to the modifier above
- "Combat bonus for nearby units does not stack with bonus granted by other instances of this unit" - I really dislike the wording of this unique. The problems: 2 great generals should not both apply their bonus to the same unit in the base game, but in mods this may actually be desired. It is therefore logical to split off this ability into a separate unique, yet wording it is difficult.
- "Instantly fully repair [unitFilter] units within [amount] tiles" - Great admiral ability
- "Great Person - [War - Water]"
- "Great Person - [War - Land]"

Deprecated uniques:
- "Bonus for units in 2 tile radius 15%"
- "Great Person - [War]"
- "Great General provides double combat bonus"

This PR is ready to be reviewed, and should be backwards & mod compatible.


avdstaaij:
> * "Combat bonus for nearby units does not stack with bonus granted by other instances of this unit" - I really dislike the wording of this unique. The problems: 2 great generals should not both apply their bonus to the same unit in the base game, but in mods this may actually be desired. It is therefore logical to split off this ability into a seperate unique, yet wording it is difficult.

How about:\
"(Strength bonus for nearby units / Strength aura) does not stack with (bonus provided by) [unitFilter] units"

I think this is sufficiently descriptive, and it allows modders to freely determine which bonus-providing units stack together. It may be somewhat confusing to users if Great Generals state that their bonus does not stack with Great Generals, but that is still semantically correct, and I think the intention is clear enough.

To support this unique for tileFilters as well (A tile improvement with a strength aura does not seem far-fetched), it could also be reworded as:\
"Strength aura does not stack with other strength auras"\
Or for more granularity, the following could be added:\
"(Strength bonus for nearby units / Strength aura) does not stack with (bonus provided by) [tileFilter]"

Furthermore, it may be possible to parametrize the number in "+[amount]% Strength for [unitFilter] units within 2 tiles" by first scanning the ruleset for the largest number in such a unique. It's a number for which parametrization would certainly be beneficial.

xlenstra:
> > * "Combat bonus for nearby units does not stack with bonus granted by other instances of this unit" - I really dislike the wording of this unique. The problems: 2 great generals should not both apply their bonus to the same unit in the base game, but in mods this may actually be desired. It is therefore logical to split off this ability into a seperate unique, yet wording it is difficult.
> 
> How about:\
> "(Strength bonus for nearby units / Strength aura) does not stack with (bonus provided by) [unitFilter] units"
> 
> I think this is sufficiently descriptive, and it allows modders to freely determine which bonus-providing units stack together. It may be somewhat confusing to users if Great Generals state that their bonus does not stack with Great Generals, but that is still semantically correct, and I think the intention is clear enough.

Implementing this with unit filters is hell, I tried it for a bit before giving up (the screenshot I send you was a non-working version and already overly complicated). It is also unclear how it should work, if unit A does not stack with B, but B has no limitations, which aura should be implemented? The highest? What if there are units A and C both blocking B, and B blocking A and C, with A and C having each a lower bonus than B, but combined a higher one? Which one should get precedence? I believe it to be NP-hard to determine the exact optimal combination of units that provide the highest bonus (knapzakprobleem oid).

If you just give precedence to the first unit, this might look weird also, "why is my +20% great general buff not applied, while my +10% smaller general buff is?"

The only easy solution is to only not have it stack with itself, e.g. 2 great generals should not stack, but a great general and smaller general should.

xlenstra:
> > * "+[amount]% Strength for [unitFilter] units within 2 tiles"
> 
> It may be possible to parametrize the number in this unique by first scanning the ruleset for the largest number that appears in it. It's a number for which parametrization would certainly be beneficial.

This could work, as long as I can find a logical place to save it so it doesn't have to be recalculated. But even then, getting all the units would start costing time for very large values. I'll look into it after the weekend.

ravignir:
if we are at it, i would really appreciate such unique:
"[+%] Strength vs [unit filter/city] for [unitFilter] units within 2 tiles"
[2] maybe also parametrize tile radius?

xlenstra:
> if we are at it, i would really appreciate such unique:
> "[+%] Strength vs [unit filter/city] for [unitFilter] units within 2 tiles"

Sure thing, I'll look at it after the weekend

> [2] maybe also parametrize tile radius?

Problem is that we determine active combat bonuses from the point of view of the combatant, not the unit that has this unique. Unless we change that, for each battle we first have to look in the closest [amount] tiles to see if there is any unit with this unique, which quickly becomes time consuming if we have to check large values of [amount] without even knowing whether we will actually find something. I can probably implement this to [3] or [4] without it slowing down too much, I'll look into it after the weekend.

avdstaaij:
> (...) for each battle we first have to look in the closest [amount] tiles to see if there is any unit with this unique, which quickly becomes time consuming if we have to check large values of [amount] without even knowing whether we will actually find something. I can probably implement this to [3] or [4] without it slowing down too much, I'll look into it after the weekend.

Well, it would be the responsibility of the modder to not use a value that is too large. But you could always cap it just in case.

ravignir:
I have one question: can you have a naval unit, a great admiral and an embarked unit on the same tile?

ravignir:
z
Workboat/Great admiral (civilian naval)
Embarked unit
naval military

xlenstra:
> I have one question: can you have a naval unit, a great admiral and an embarked unit on the same tile?

Nope, but implementing that is a whole different system out of the scope of this PR. Currently, embarked units are still split between military & civilian, and thus you can't have a military naval unit and military embarked unit on the same tile.

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -70,15 +67,51 @@ object BattleDamage {
...
+            //
+            
+            // Calculate strength bonus from nearby great generals/great admirals or mod equivalents
+            val nearbyBonusUnits = combatant.unit.getTile().getTilesInDistance(combatant.getCivInfo().gameInfo.ruleSet.modOptions.maxGreatGeneralSearchDistance)
+                .flatMap { it.getUnits() }
+                .filter { it.civInfo == combatant.unit.civInfo && it.hasUnique("+[]% Strength for [] units within [] tiles")}
+            
+            for (unit in nearbyBonusUnits) {
+                var totalApplicableBonus = 0
+                for (unique in unit.getMatchingUniques("+[]% Strength for [] units within [] tiles"))
```
yairm210, line 87: "[]% Strength" - + or - should be within brackets

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -70,15 +67,51 @@ object BattleDamage {
...
+            
+            // Calculate strength bonus from nearby great generals/great admirals or mod equivalents
+            val nearbyBonusUnits = combatant.unit.getTile().getTilesInDistance(combatant.getCivInfo().gameInfo.ruleSet.modOptions.maxGreatGeneralSearchDistance)
+                .flatMap { it.getUnits() }
+                .filter { it.civInfo == combatant.unit.civInfo && it.hasUnique("+[]% Strength for [] units within [] tiles")}
+            
+            for (unit in nearbyBonusUnits) {
+                var totalApplicableBonus = 0
+                for (unique in unit.getMatchingUniques("+[]% Strength for [] units within [] tiles")) 
+                    if (combatant.unit.matchesFilter(unique.params[1]) && HexMath.getDistance(combatant.unit.getTile().position, unit.getTile().position) <= unique.params[2].toInt())
```
yairm210, line 88: combatant.unit.getTile().aerialDistanceTo(unit.getTile())
HexMath.getDistance should be deprecated since it doesn't consider world-wrapped coordinates

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -70,15 +67,51 @@ object BattleDamage {
...
+                        totalApplicableBonus += unique.params[0].toInt()
+                for (unique in unit.getMatchingUniques("+[]% Strength vs [] for [] units within [] tiles"))
```
yairm210, line 90: If this isn't used yet by any unit, don't add it.

yairm210 (commented):
I think we're going to have to put off the Great Admiral until we finish remorming the Great People. I'll write up about that.

logicminimal:
> I think we're going to have to put off the Great Admiral until we finish remorming the Great People. I'll write up about that.

if admirals won't be implemented soon then i suggest changing the great admiral bonus from commerce into great general for now

yairm210:
Great People accumulation needs to be completely reformed.

Currently, GPP from buildings, specialists etc. are assigned *by the stat name*.
The end goal is to have them assigned *by GP name*.
This is similar to the reform we went though for specialists, albeit lesser.


-- NON game-changing steps --

- GreatPersonManager.getGreatPersonPointsStored which returns Counter<String>(), converts GreatPersonManager.greatPersonPoints to GreatPersonManager.greatPersonPointsStored with (ugh) statToGreatPersonMapping
- CityInfo.getGreatPersonPointsForNextTurn() copy (called ..._new or something), which returns... a list of objects with 3 params, it'll have to be. {greatPerson:String, points:int, source:String}
- Support for the above in CityInfoTable, CityInfo.getGPP(), CivInfo.getGPP(). Convert BACK to the current format in GreatPersonManager.addGreatPersonPoints.

All changes until now basically support the current config - jsons remain the same, stored info remains the same.
These PRs can be separate.

AFTER ALL THIS, comes the final stage of changing the actual saved info.

- Convert Building.greatPersonPoints from Stats to Counter<String>, and add support for conversion from stat to GP name in CityInfo.getGreatPersonPointsForNextTurn - this allows us to start changing `"greatPersonPoints": {"culture": 2},` to `"greatPersonPoints": {"Great Artist": 2},`

And then, the final stage:
- Change saved info in GreatPersonManager from greatPersonPoints:Stats to greatPersonPointsNew:counter<String>, to allow modding for ANY Great People you feel like!

Only after ALL OF THIS can we start supporting Great Admirals.
- Unique that determines starting cost, and incremental cost, of GP (e.g. "Initially requires [amount] points, rising by [amount] for each unit created") - to be applied to Great General. This will obvious change how we store this informmation in saved games as well.
- Unique that determines how we gain said points - "XP gained from battles by [All] units is also added as [great General] points". Here we'll have to deal with replacement units as well (Khan).
- After ALL THAT IS DONE, the Great Admiral is just an extension of the existing rules.


NOW, WHY SHOULD WE GO TO ALL THIS TROUBLE FOR ONE MEASLY UNIT?!?!?!

A. It's not just this unit. This system is hardcoded as hell, and doesn't support modding at all. Yukky yuk, who wrote this crap (hint: me)
B. By moving up to G&K rules, we should still ALWAYS be able to support Vanilla rules. If I now created a mod for Vanilla, all previous abilities are replicable.

yairm210:
That's a big infodump and checklist.
This is non-critical, and can definitely be put off for a different time, or different person.
But that's what we really need,

ravignir:
Add to the list:
Great Person generated by specialist/great person points should be generated per city and not per empire like in civ5. 
Example:
City 1:
98/100 progress towards great scientist. +7 GS points per turn
City 2:
18/100 progress towards great scientist. +3 GS points per turn
** Next turn **
City 1:
Great Scientist is born in City 1. 0/200 progress towards great scientist. +7 GS points per turn.
City 2:
21/200 progress towards great scientist. +3 GS points per turn

logicminimal:
> Add to the list:
> Great Person generated by specialist/great person points should be generated per city and not per empire like in civ5. 
> Example:
> City 1:
> 98/100 progress towards great scientist. +7 GS points per turn
> City 2:
> 18/100 progress towards great scientist. +3 GS points per turn
> ** Next turn **
> City 1:
> Great Scientist is born in City 1. 0/200 progress towards great scientist. +7 GS points per turn.
> City 2:
> 21/200 progress towards great scientist. +3 GS points per turn

@yairm210 this is a good idea but this is not how it is implemented in civ 5, what do you think? do you think there is a degree of freedom when implementing such things?

ravignir:
>
>this is not how it is implemented in civ 5
>
This IS how it is implemented in civ5.

ravignir:
https://civilization.fandom.com/wiki/Great_People_(Civ5)
>Great Person Points for Great Artists, Engineers, Merchants, and Scientists accumulate in cities, and when a city accumulates enough points towards a particular Great Person, (s)he is "born" in that city. These points are earned by Specialists, as well as several Wonders and Social Policies. There are also enhancers that will make cities generate GPPs faster, as seen below.
>

ravignir:
obz
z

logicminimal:
> >
> >this is not how it is implemented in civ 5
> >
> This IS how it is implemented in civ5.

it isnt pooled globally?

PR #4270, opened 2021-06-25
Fixed the game crashing when any main menu option is picked

avdstaaij:
Commit 823e6ffc2a4d14c04a661e2e0f8c2253c0d0844a causes the game to crash when you pick any option in the main menu. This PR fixes this critical bug.

The problem was that the first character of string was accessed without checking whether the string was empty. It's interesting how such a tiny mistake can cause the entire game to be unusable.


SomeTroglodyte:
Thanks for the catch - my draft didn't have the problem because I had an empty-short-out _at the very beginning_ of tr() specifically to avoid this, but without adding a comment to that effect. And when Yair 'liked' the stats short-out I completely forgot to mention. Shows how valuable comments could be. And how good teamwork is :+1:!

PR #4271, opened 2021-06-26
Added Stealth tech and Stealth Bombers

avdstaaij:
This PR implements the Stealth technology, Stealth Bombers and all uniques required for them.

Added uniques:
- "Damage taken from interception reduced by []%"
- "Cannot be carried by [unitFilter] units"

Deprecated uniques:
- "Reduces damage taken from interception by 50%"

Reworded uniques (added and deprecated):
- "Can not be intercepted" -> "Cannot be intercepted"
  This really irked me, and the new version is also how it was already written in the wiki.
  
I parametrized "Reduces damage taken from interception by 50%" in order to give the Stealth Bomber 100% damage reduction, but after doing so, I was not actually sure if this was accurate. There is a difference between not being interceptible at all and reducing interception damage by 100%: the former does not consume the interceptions of defenders. For the Stealth Bomber, I assumed its "Evasion 100" ability meant the former, but I'd appreciate it if someone could test this.

I made "Cannot be carried by [unitFilter] units" generic in expectation of the changes to the unit carrying uniques from #4260. I also worded it to match the uniques from that PR, though "Cannot be transported by [unitFilter] units" might be better.

The interception damage bonus seemed to be incorrect before: it divided an integer percentage by 100, which probably led to the bonus being either 0% with Interception I or 100% with Interception II or interception III. This has also been fixed.


ravignir:
It cannot be intercepted, just like missiles. For that very reason, before that unit is added, I think air combat needs a rework to finally work like in civ5, i.e Air units like bombers/fighters should take retaliation damage from the units they are attacking.

ravignir:
Go with the 100% damage reduction from interception i think.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -847,6 +849,13 @@ class MapUnit {
             .sumBy { it.params[0].toInt() }
     }
 
+    fun receivedInterceptDamageFactor(): Float {
+        return getMatchingUniques("Damage taken from interception reduced by []%")
+            .fold(1f) { product, it -> product * 1f - it.params[0].toFloat() / 100f }
```
xlenstra, line 854: I believe you missed some brackets here.
I personally prefer doing this with a for loop instead of `.fold()`, but that is a matter of opinion.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -578,14 +578,19 @@ object Battle {
     }
 
     private fun tryInterceptAirAttack(attacker: MapUnitCombatant, defender: ICombatant) {
+        if (attacker.unit.hasUnique("Cannot be intercepted")) return
+        // Deprecated in 3.15.6
         if (attacker.unit.hasUnique("Can not be intercepted")) return
```
xlenstra, line 583: I personally like to indent the deprecated part, so it is clear what is deprecated and what isn't, but this is of course personal choice
```
        // Deprecated in 3.15.6
            if (attacker.unit.hasUnique("Can not be intercepted")) return
        //
```

core/src/com/unciv/logic/map/MapUnit.kt
yairm210: Please do this with a for loop, as it's more immediately readable.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -847,6 +849,13 @@ class MapUnit {
...
+            // This unique is deprecated since 3.15.6
+            .times(0.5f.pow(getUniques().filter{it.text == "Reduces damage taken from interception by 50%"}.count()))
```
yairm210, line 856: getUniques.count{...}

core/src/com/unciv/logic/map/MapUnit.kt
avdstaaij: It's too bad Kotlin has no `productBy`. I have replaced this with a for loop.

core/src/com/unciv/logic/battle/Battle.kt
avdstaaij: That does indeed improve clarity! I have changed this.

avdstaaij:
I have applied the various suggested style changes, and gave Stealth Bombers 100% interception damage reduction instead of making them non-interceptable.

PR #4282, opened 2021-06-26
Update Swedish.properties

SimonCeder:


SimonCeder:
what is the stargazers field?

PR #4284, opened 2021-06-26
Fix "National Wonder is being built elsewhere" not displayed

SomeTroglodyte:
Self-explaining?

Anyway, I don't think this is good design, so this is the 'kludge' variant. I'd probably declare a `data class ConstructionRejectionReason(val text: String, val isRejected: Boolean = text.isNotEmpty(), val showRejected: Boolean = false)` and have it filled out by getRejectionReason, but that would take more effort. You think the idea worth it?


yairm210:
No, this is fine

PR #4285, opened 2021-06-26
Nicer distribution of policy picker branches

### LINKED ISSUE #4172: make the policy screen scroll vertically instead of horizontally for portait mode

SomeTroglodyte:
My attempt to make the Policy Picker a little prettier on cramped screen space or portrait orientations. Should resolve #4172.

Notes:
- Includes alternate approach to displaying "All policies adopted" which is currently impossible - as of now untested, test underway.
- Not 100% happy with two mechanisms to decide when a break is needed, but out of ideas on how to make one of the approaches also fit the other's usecase.
- Not 100% happy with the padding above and below the separator line. On high resolutions it's a bit small, under cramped screen space it's better this way... Make it dependent on resolution?

- Alternate approach: Save space by reducing number of "branch boxes". e.g. Piety might be completely absent once Rationalism is chosen and vice versa, and while neither is chosen we display one box containing only the starting choices, same for the evil/neutral/good trio -> down to 7 branches... A bit more complex, but doable even in a mod-flexible way. Not in this PR though.

Might do screenies later, and for now I'll leave this a normal PR because I'm pretty sure it improves things despite the notes above :) Critiques of course welcome.

### DIFF
--- a/core/src/com/unciv/logic/civilization/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/PolicyManager.kt
@@ -133,14 +133,14 @@ class PolicyManager {
      * Test whether a policy is adoptable according to the RuleSet (ignoring cost).
      * Note: branch completion policies are automatic and therefore not adoptable in this test.
      * @param policy The Policy to check
-     * @param testEra Include era test (with false the function returns whether the policy is adoptable now or in the future)
+     * @param checkEra Include era test (with false the function returns whether the policy is adoptable now or in the future)
      * @return `true` if the policy can be adopted, `false` if some rule prevents it (including when it's already adopted) 
      */
-    fun isAdoptable(policy: Policy, testEra: Boolean = true): Boolean {
+    fun isAdoptable(policy: Policy, checkEra: Boolean = true): Boolean {
         if (isAdopted(policy.name)) return false
         if (policy.name.endsWith("Complete")) return false
         if (!getAdoptedPolicies().containsAll(policy.requires!!)) return false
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -129,11 +129,18 @@ class PolicyManager {
...
 
-    fun isAdoptable(policy: Policy): Boolean {
+    /**
+     * Test whether a policy is adoptable according to the RuleSet (ignoring cost).
+     * Note: branch completion policies are automatic and therefore not adoptable in this test.
+     * @param policy The Policy to check
+     * @param testEra Include era test (with false the function returns whether the policy is adoptable now or in the future)
+     * @return `true` if the policy can be adopted, `false` if some rule prevents it (including when it's already adopted) 
+     */
+    fun isAdoptable(policy: Policy, testEra: Boolean = true): Boolean {
```
yairm210, line 139: 'test' has a very specific meaning - it means out-of-game tests of the code. Such a variable would indicate that the era is 
a 'fake' era or something similar. This should be renamed to 'checkEra'.

SomeTroglodyte:
Here's a visual:

https://user-images.githubusercontent.com/63000004/123525274-b2b96180-d6cf-11eb-8445-240a34630781.mp4

Oh, and I forgot a detail: Omitting the updateVisualScroll looks good in fewer cases than it looks ugly, so... Shortly.

yairm210:
Looks good!

SomeTroglodyte:
image
So the button works - question is, does the wording still match the reason? Would "All possible policies adopted" be too long or unduly stress out our translator staff?

.. but the last push is only what's already been decided.

yairm210:
I think this is good as is

PR #4287, opened 2021-06-27
Translation template fixes

### LINKED ISSUE #4275: Missing translations
Some strings are missing from translation files. Those are:
"Show experimental world wrap for maps" - from settings menu
"Enable portrait orientation" - from settings menu
"Author:" - from mods menu
"Size:" - from mods menu
"Load from custom location" - from saves menu
"Open terrain" - from civilopedia

Also there's some strings which is present in translation file but still not translated by some reason. Those are:
"Penalty vs City 33%"
"Our proposed trade is no longer relevant!" which is present as slightly different "Our proposed trade **request** is no longer relevant!"

I'll post more if i find.

SomeTroglodyte:
Resolves #4275 (most of, plus a few similar others)
- Touches _all_ translation files because: Preserve existing "Our proposed trade request is no longer relevant" translations. Other changes are template only and rely on a translation writer run to propagate them. German has translations for the issue cases just for testing, but if they stay, they'd be fine.
- Turkish - since renaming and moving it to root is an obvious mistake I reversed that - just so I could include above minifix.
- Penalty vs [city] left in from testing, but is moot unless the translation writer learns about translateBonusOrPenalty
- Code changes are marginal: Unified custom location load/save error messages to both have the same number of exclamation marks and made the world wrap option template easier to read - my perception of easy.


SomeTroglodyte:
I _thought_ "this needs a local unit tests run" and then forgot a few seconds later...

logicminimal:
> I _thought_ "this needs a local unit tests run" and then forgot a few seconds later...
> Was just a lowercase/uppercase mismatch in that Turkish file - see #4256 for where it was disabled.

i couldnt see where the disabled line is, could you please describe further?

logicminimal:
> I _thought_ "this needs a local unit tests run" and then forgot a few seconds later...
> Was just a lowercase/uppercase mismatch in that Turkish file - see #4256 for where it was disabled.

oops.

Issue #4290, opened 2021-06-27
Implementation of Religion - Roadmap


xlenstra:
This issue is meant as a collection of features that religion should include once implemented. I have started working on this very large new feature, and am planning to gradually implement more parts of it over the following weeks/months. While waiting for previous PR to be merged, I'll make smaller updates for other features in the meantime. The issues below are in the approximate order at which I'll start implementing them and includes features already implemented.

Roadmap:
- [x] Button for the user to access this HIGHLY EXPERIMENTAL feature - #4291
- [x] Shrines, a civReligionManager for civs - #4291
- [x] Pantheons, a cityRelgionManager for cities (basic implementation, only a single religion per city) - #4296 
- [x] Great Prophets, used for spreading religion to different cities (immediately replaces previous religion), can be earned with faith - #4356 
- [x] Great Prophet can create religions - #4505 
- [x] Follower beliefs - #4530
- [x] Founder beliefs
- [x] A UI for viewing the beliefs you and other players have chosen
- [x] Purchase missionaries with faith (groundwork for buying other units also) - #4568
- [x] Cities can have multiple different religions competing inside
- [x] Religion spreads to nearby cities with trade routes or something, I have yet to read up on how exactly this works
- [x] A UI for viewing the religions inside cities
- [x] Inquisitors
- [x] Great Prophets can reform religions with Enhancer and extra Follower believes
- [x] Great prophets may be used before founding a pantheon to also choose a pantheon belief
- [x] AI will found & enhance religions (randomly chosen beliefs for now)
- [x] Buy Great People with faith when completing certain policy trees
- [x] Somewhere add to a UI what cities are holy cities
- [x] Starting from the industrial era, lots of things change (no more great prophet generation, etc.)
- [x] Upon founding a religion, change all existing great prophets to be of that religion
- [x] May choose a different name for your religion
- [x] Spreading religion with a great prophet also removes other religions somewhat or something
- [x] Renaming a religion changes notifications
- [x] Religious city-states
- [x] Sharing a religion with a city-state decreases rate of influence decrease or something
- [x] Religious wonders (Taj Mahal, Grand Temple, etc.)
- [x] AI improvement so that they will choose beliefs that are directly benefiting them
- [x] AI improvement so they can actually buy and use religious units
- [x] AI buys religious buildings
- [ ] AI for buying great persons with Faith
- [x] AI chooses when to spend faith and when to save up
- [x] AI for great prophets when they don't have a religion
[...]

PR #4291, opened 2021-06-27
Added Shrine, option for enabling religion

xlenstra:
This is the first of many PR's which will eventually include all of the religion mechanic: See #4290 . This PR adds two things: a button that enables a button in the new game screen to enable religion (mostly for myself so I can more easily test things), and the shrine, a building for generating faith.


yairm210:
The game options look good!
The shrine should be hidden (both in tech buttons and as buildable in cities) for non-religion games though. You could either do this by catching every building that provides faith, or by adding a special unique - I think the first is probably better but I'm not sure.

xlenstra:
Good point, I'll look into it

xlenstra:
Shrine should now be hidden, both in the tech and buildings, but is still visible in the civilopedia

logicminimal:
btw the game freezes when i select found pantheon, 3.15.6

xlenstra:
> btw the game freezes when i select found pantheon, 3.15.6

Known bug, fixed in 3.15.7

PR #4293, opened 2021-06-27
Cultural victory display - new template, Utopia Project

SomeTroglodyte:
Relatively minor:

- I noticed the new cultural victory didn't update its victory message template - fixed
- Then I thought "If science lists Apollo completion why not list Utopia completion for cultural"
- Then I thought "It's getting a bit cramped - why not drop the locked-out branches"

I think a game where one maxes policies but refuses to build Utopia, then triggers another victory looks better and has good chances to not need scrolling. On the other hand, one starts out with all 10 branches listed as un-completed on the victory screen and that list gets shorter during the game which may confuse some. You decide what's over the top, I'll revert. ;)

### DIFF
--- a/android/assets/jsons/translations/German.properties
+++ b/android/assets/jsons/translations/German.properties
@@ -737,7 +737,7 @@ Science victory = Wissenschaftssieg
 Cultural victory = Kultursieg
 Conquest victory = Dominanzsieg
 Complete all the spaceship parts\n to win! = Um zu gewinnen, baut\n alle Raumschiffteile!
-Complete 5 policy branches\n to win! = Um zu gewinnen vervollständige\n 5 Politik-Zweige!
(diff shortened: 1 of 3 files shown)


yairm210:
Okay, 2 points here.
A. There are built-in assumptions in the current notifications that this is a good time to revisit.
-  The number of policy branches needed to build the wonder
- The name of the wonder

These should be abstracted away with templates

B. I agree that once you've finished all X policy branches you need, the only thing displayed should be "Have you constructed the [cultural victory building]"
But that also sounds like a "different PR" problem

SomeTroglodyte:
Agreed. I shouldn't have touched @xlenstra 's work in the first place. Hi Xander! I think 'abstracting away' that literal '5' might turn out better / more consistent if you do it. Feel free to pilfer any code/idea here without further mention.

xlenstra:
Most of what you did seemed fine, and I don't personally have a problem with other people improving my code, so feel free to reopen this and improve it further :)
I'll put this on my todolist regardless, if you don't do anything with this PR anymore, I'll probably open a PR sometime about these points

PR #4295, opened 2021-06-28
Respect visualMods for Sound - CheckBox, formats, modchange detect

SomeTroglodyte:
Replacement for #4227 - still haven't learned how to properly rebase a branch that is already a PR

Changes:
- CheckBox as suggested. Much better.
- Supports all Gdx-supported formats (triggered by an issue in the wrong place which shows people will get the idea to use ogg files)
- Cache clear detection now runs often enough to actually work

All formats tested, cache rebuild tested through logging, on desktop only. Mobile should be OK since format support is in the Gdx documentation.


android/assets/jsons/translations/German.properties
```diff
@@ -990,8 +990,7 @@ Current mods = Installierte Modifikationen
 Downloadable mods = Verfügbare Modifikationen
 Next page = Nächste Seite
 Open Github page = Auf Github öffnen
-Enable as permanent visual mod = Als permanente visuelle Modifikation aktivieren
-Disable as permanent visual mod = Als permanente visuelle Modifikation deaktivieren
+Permanent visual mod = Permanente audiovisuelle Mod
```
Mape6, line 993: In template.properties heißt der String "Permanent **audio**visual mod"

SomeTroglodyte: Aua, danke. Und ich habe es nur abgekürzt weil da wenig Platz ist - ist in Portrait eh eine Katastrophe und bräuchte eine gute Idee...

SomeTroglodyte:
OK, finally found out why Android hardware so often doesn't play our sounds at all (ligGdx #1175) - so I'm rethinking the approach of our Soundcache again. Way I see it, there's no nice way to get a clean Logcat with libGdx Sound, but perhaps I'll have some intuition later. Loading _and decoding to PCM_ all sounds into Byte arrays (not Sound objects) then instantiating Sound objects from that on demand would do it - I think - and that would be ghastly.

SomeTroglodyte:
### Question:
If I do something about the android not playing quirks, I'll need to Thread.Sleep until Sound.play() returns a valid stream ID. So - threads. Everything that goes through `onClickEvent` and the keyboard unit actions have their own "Sound" thread already, but other calls do not (e.g. movement woosh or option screen sliders). Suggested approach?

core/src/com/unciv/ui/utils/Sounds.kt
```diff
@@ -16,42 +16,66 @@ import java.io.File
...
 
     fun get(sound: UncivSound): Sound? {
+        checkCache()
         if (sound in soundMap) return soundMap[sound]
         val fileName = sound.value
         var file: FileHandle? = null
-        for (modFolder in getFolders()) {
-            val path = "${modFolder}sounds$separator$fileName.mp3"
+        for ( (modFolder, extension) in getFolders().flatMap { 
+            folder -> SupportedExtensions.values().asSequence().map { folder to it }
```
yairm210, line 69: This is neat, but confusing. I can't think of a way to make this simpler, so I'm not considering this a 'tofix', but I am pointing it out - this is non-trivial.

yairm210:
I think that fixing 'playing sounds' is definitely a "different PR" problem.
I suggest 'ignore entirely and wait for LibGDX-level fix' if you don't find this to be a significant problem

core/src/com/unciv/ui/utils/Sounds.kt
SomeTroglodyte: I decided to include the Android-not-playing-first-sound fix here, so the whole file is already thoroughly worked over. I think I added a comment for this one - WAAAAIT - OK, next PR. :racehorse:

PR #4296, opened 2021-06-28
Added a few pantheons; religion uniques affect cities instead of civs; updated pantheon picker screen

xlenstra:
Second part of implementation of Religion - see #4290.
This PR has two following features implemented:
- A few basic pantheons were implemented, I haven't tested every single one specifically, but they should work
- The picker screen for pantheons can now be closed without choosing one, so you can check your game again to make a better choice. If you like UI design _please_ improve my design, I suck at UI and just basically copy+pasted the GreatPersonPicker
- Uniques for religions can now differ between cities. Each city only receives the uniques of the largest religion it contains
- When founding a pantheon, the capital city is automatically converted to this pantheon. No religion can be spread apart from this as of yet, so this update basically boils down to a small bonus for capitals if you produce enough faith


yairm210:
Before you go all out on Religion, I'm reminding you to go over all changed uniques and document the change in the wiki.
In fact, there were so many changes that I think we should add a section of deprecated uniques, where we say "unique A had been replaced with B"

ravignir:
Do you have a reliable source on how spreading religion/pantheons works?

xlenstra:
> Before you go all out on Religion, I'm reminding you to go over all changed uniques and document the change in the wiki.
> In fact, there were so many changes that I think we should add a section of deprecated uniques, where we say "unique A had been replaced with B"

I did update all the uniques I changed in the policy updates, and I think also most of the PR's I created after that. Just to be sure, I'll go over them again and fill in any I missed.

A section of deprecated uniques might be a good Idea, I'll add it to the bottom for now, feel free to move it somewhere else where it fits better

xlenstra:
> Do you have a reliable source on how spreading religion/pantheons works?

This one seems quite reliable:
https://www.carlsguides.com/strategy/civilization5/religion/convertingandreligiouspressure.php

yairm210:
You're adding the Pantheons to the BASE GAME.
If I recall your previous 'does this game include religion' boolean change, it was 'either pantheons exist OR the user chose to include religion'.
That means that ALL new games would have religion by default...

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -307,6 +309,13 @@ open class TileInfo {
                     stats.add(unique.stats)
                 }
             }
+            // This entire function should really be refactored, all of these almost similar unique checks are really ugly.
+            // But for that, we would probably first have to generalize _all_ uniques to use cityFilters, and that might take a while
+            for (unique in city.religion.getMatchingUniques("[] from every []")) {
```
yairm210, line 314: Just add it to line 307. 
for (unique in city.civInfo.getMatchingUniques("[] from every []")
     + city.civInfo.getMatchingUniques("[] from every []")) ...

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -175,13 +177,22 @@ class Building : NamedStats(), IConstruction {
             else
                 for (unique in civInfo.getMatchingUniques("[] from every Wonder"))
                     stats.add(unique.stats)
-
+            
+            if (city != null) {
+                for (unique in city.religion.getMatchingUniques("[] from every []")) {
```
yairm210, line 182: This too, add to line 162. There's an 'if' in the way so it's a little longer, like:

var fromEveryUniques = civInfo.getMatchingUniques("[] from every []")
if(city != null) fromEveryUniques += city.religion.getMatchingUniques("[] from every []")

for (unique in fromEveryUniques)
   if (matchesFilter(unique.params[1]))
      stats.add(unique.stats)

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -694,19 +691,20 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Cam
...
+//                        if (belief.type != "Pantheon" || gameInfo.civilizations.any { it.religionManager.pantheonBelief == belief.name }) continue
+//                        val beliefTable = Table().apply { touchable = Touchable.enabled; background = ImageGetter.getBackground(ImageGetter.getBlue()) }
+//                        beliefTable.pad(10f)
+//                        beliefTable.add(belief.name.toLabel(fontSize = 24)).row()
+//                        beliefTable.add(belief.uniques.joinToString().toLabel())
+//                        beliefTable.onClick { viewingCiv.religionManager.choosePantheonBelief(belief); pantheonPopup.close(); shouldUpdate = true }
+//                        beliefsTable.add(beliefTable).fillX().row()
+//                    }
+//                    pantheonPopup.add(ScrollPane(beliefsTable)).maxHeight(stage.height * .8f)
+//                    pantheonPopup.open()
```
yairm210, line 706: There should not be code comments in code - if we'll want this code later, it'll be in older git versions :)

yairm210:
Ah
I see I inadvertently created conflicts when adding the isPickablePantheonBelief() function

xlenstra:
No worries, I'll fix them right away

yairm210:
Instead of 'city.civinfo.getmatchinguniques + city.religion.getmatchinguniques' everywhere, we should probably make a 'city.getmatchinguniques' that will return both

xlenstra:
> Instead of 'city.civinfo.getmatchinguniques + city.religion.getmatchinguniques' everywhere, we should probably make a 'city.getmatchinguniques' that will return both

I agree, my plan was to make a function getUniquesApplicableForThisCity() which also queried the civInfo for the uniques with a matching cityFilter attached, but for that to work, all uniques that apply to cities should use cityFilters instead of 'in this city' being suffixed. As this is a lot of work, I decided it would be better in a different PR and reverted the small amount of progress I made.

yairm210:
Legit

PR #4297, opened 2021-06-29
Missing single quotation mark in quote

Mape6:
Inserted ' in quote of Advanced Ballistics

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Techs.json
+++ b/android/assets/jsons/Civ V - Vanilla/Techs.json
@@ -575,7 +575,7 @@
 				"name": "Advanced Ballistics",
 				"row": 5,
 				"prerequisites": ["Nuclear Fission","Rocketry"],
-				"quote": "Our scientific power has outrun our spiritual power, we have guided missiles and misguided men. – Martin Luther King Jr. "
+				"quote": "'Our scientific power has outrun our spiritual power, we have guided missiles and misguided men.' – Martin Luther King Jr. "
 			},
 			{
 				"name": "Satellites",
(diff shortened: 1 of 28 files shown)


SomeTroglodyte:
I would have propagated the change to all translations where a translation exists - after all, the words stay valid, and otherwise the json - to - tr generator would just throw them out, or am I mistaken?

Mape6:
> 
> 
> I would have propagated the change to all translations where a translation exists - after all, the words stay valid, and otherwise the json - to - tr generator would just throw them out, or am I mistaken?

Good question. I tried to figure out how the TranslationFileWriter works in detail but my programming knowledge is not good enough to understand everything. That's why I'm not really sure what happens. But I agree with your suggestion and will change all languages files.

SomeTroglodyte:
:+1: I know that's tedious, especially with needing to ensure some 'intelligent' editor does not change the number of leading or trailing spaces...

Mape6:
The translation file of Thai is completely strange. The quote is missing here.
In the files for Turkish and Vietnamese the tech "Advanced Ballistics" was also missing. I added that.

Probably someone should check all translation files and compare them to each other to find and correct the strings and formats.

SomeTroglodyte:
Isn't thai somehow deactivated? Turkish may lag a little due to its translator renaming it away a while ago.
```kotlin
    // This function is too strange for me, however, let's keep it "as is" for now. - JackRainy
    private fun getLanguagesWithTranslationFile(): List<String> {
(...)
        languages.remove("Thai") // Until we figure out what to do with it
        languages.remove("Vietnamese") // Until we figure out what to do with it
```

Mape6:
Got to know that. Thanks for that hint. I don't need to care about both languages then.

PR #4301, opened 2021-06-29
Added Privateer unit; updated Coastal Raider promotion

xlenstra:
This PR adds the Privateer unit, and updates the Coastal Raider promotion to also include the gain gold from attacking cities unique.

Added uniques:
- "May capture killed [unitFilter] units"
- "Earn [amount]% of the damage done to [unitFilter] units as [stat]" - stat = Gold, Culture, Science or Faith


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -413,7 +465,7 @@ object Battle {
         
         // Declare war on the owners of all hit tiles
         for (hitCiv in hitTiles.map { it.getOwner() }.distinct()) {
-            hitCiv!!.addNotification("A(n) [${attacker.getName()}] exploded in our territory!".tr(), targetTile.position, NotificationIcon.War)
```
yairm210, line 416: Good catch!

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -127,33 +129,83 @@ object Battle {
...
+        newUnit.health = 50
     }
 
     private fun takeDamage(attacker: ICombatant, defender: ICombatant) {
-        var damageToDefender = BattleDamage.calculateDamageToDefender(attacker, attacker.getTile(), defender)
-        var damageToAttacker = BattleDamage.calculateDamageToAttacker(attacker, attacker.getTile(), defender)
+        var potentialDamageToDefender = BattleDamage.calculateDamageToDefender(attacker, attacker.getTile(), defender)
+        var potentialDamageToAttacker = BattleDamage.calculateDamageToAttacker(attacker, attacker.getTile(), defender)
+        
+        var damageDoneToAttacker = 0
```
yairm210, line 162: Instead of creating 2 new variables, just 'normalize' the value of the existing variables after combat to adjust for overkill.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -127,33 +129,83 @@ object Battle {
...
+    private fun tryCaptureUnit(attacker: ICombatant, defender: ICombatant) {
+        // https://forums.civfanatics.com/threads/prize-ships-for-land-units.650196/
+        // https://civilization.fandom.com/wiki/Module:Data/Civ5/GK/Defines
+        if (!defender.isDefeated()) return
+        if (attacker !is MapUnitCombatant) return
+        if (defender is MapUnitCombatant && !defender.getUnitType().isMilitary()) return
+        if (attacker.unit.getMatchingUniques("May capture killed [] units").none { defender.matchesCategory(it.params[0]) }) return
+        
+        var defendingStrength = 0
+        if (defender is CityCombatant) defendingStrength = defender.getCityStrength()
```
yairm210, line 143: This is what ICombatant.getStrength() is for

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -127,33 +128,79 @@ object Battle {
...
-                    damageToAttacker--
+                    potentialDamageToAttacker--
                     attacker.takeDamage(1)
                     if (attacker.isDefeated()) break
                 }
             }
         }
+        
+        damageToAttacker -= potentialDamageToAttacker // We first set this to the maximal damage to be done, and now we
+        damageToDefender -= potentialDamageToDefender // subtract the amount of damage remaining. This yields the damage done
```
yairm210, line 182: U wot m8t
You realize this means that for ranged attacks, both values will be 0, yes?

```

        var damageToAttacker = potentialDamageToAttacker
        var damageToDefender = potentialDamageToDefender


            defender.takeDamage(potentialDamageToDefender) // straight up


        damageToAttacker -= potentialDamageToAttacker // We first set this to the maximal damage to be done, and now we
        damageToDefender -= potentialDamageToDefender // subtract the amount of damage remaining. This yields the damage done
```

I'm not sure what you tried to do here, but this is probably not the way.

core/src/com/unciv/logic/battle/Battle.kt
xlenstra: You said that before, but I don't understand what you mean. The variables get changed during the while loop, so to me it looks like the only way to determine what the total damage done is, is either to save their starting values and subtract the end value from it, or keep track of them during the while loop. Either that, or saving the starting health and subtract fhe ending health of both units. All of these options require extra variables

PR #4303, opened 2021-06-29
Research Tech Button shows progress; Small bug fix

xlenstra:
This PR makes some small changes to the TechPicker:
- The "Research [techName]"-button in the bottom right now shows progress in the researched tech. If a path of multiple techs is selected, the sum of the progress of all these techs is shown.
- Previously, whenever you opened the TechPicker for selecting a free tech, it would first still say "Research [techName]", and only after clicking on any tech it would become "Pick [techName] as free tech" (should this be "as _a_ free tech"?). This was due to init blocks being called before constructors, making a variable set late. This has been solved.

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
@@ -17,10 +17,9 @@ import java.util.*
 import kotlin.collections.ArrayList
 
 
-class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Technology? = null, freeTechPick: Boolean = false) : PickerScreen() {
+class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Technology? = null, private val freeTechPick: Boolean = false) : PickerScreen() {
 
     private var techNameToButton = HashMap<String, TechButton>()
-    private var isFreeTechPick: Boolean = false
     private var selectedTech: Technology? = null
     private var civTech: TechManager = civInfo.tech
     private var tempTechsToResearch: ArrayList<String>
@@ -48,7 +47,6 @@ class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Tec
     private val turnsToTech = civInfo.gameInfo.ruleSet.technologies.values.associateBy({ it.name }, { civTech.turnsToTech(it.name) })
     
     init {
-        isFreeTechPick = freeTechPick
         setDefaultCloseAction()
         onBackButtonClicked { UncivGame.Current.setWorldScreen() }
         scrollPane.setOverscroll(false, false)
@@ -61,7 +59,7 @@ class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Tec
         rightSideButton.setText("Pick a tech".tr())
         rightSideButton.onClick(UncivSound.Paper) {
             game.settings.addCompletedTutorialTask("Pick technology")
-            if (isFreeTechPick) civTech.getFreeTechnology(selectedTech!!.name)
+            if (freeTechPick) civTech.getFreeTechnology(selectedTech!!.name)
             else civTech.techsToResearch = tempTechsToResearch
 
             game.setWorldScreen()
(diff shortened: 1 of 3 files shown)


logicminimal:
> This PR makes some small changes to the TechPicker:
> - The "Research [techName]"-button in the bottom right now shows progress in the researched tech. If a path of multiple techs is selected, the sum of the progress of all these techs is shown.
> - Previously, whenever you opened the TechPicker for selecting a free tech, it would first still say "Research [techName]", and only after clicking on any tech it would become "Pick [techName] as free tech" (should this be "as _a_ free tech"?). This was due to init blocks being called before constructors, making a variable set late. This has been solved.

what will happen to part science spent on techs that have had a free ech bonus used on them?

also im striking out this one in my a lot of features and issues thread

xlenstra:
> > This PR makes some small changes to the TechPicker:
> > 
> > * The "Research [techName]"-button in the bottom right now shows progress in the researched tech. If a path of multiple techs is selected, the sum of the progress of all these techs is shown.
> > * Previously, whenever you opened the TechPicker for selecting a free tech, it would first still say "Research [techName]", and only after clicking on any tech it would become "Pick [techName] as free tech" (should this be "as _a_ free tech"?). This was due to init blocks being called before constructors, making a variable set late. This has been solved.
> 
> what will happen to part science spent on techs that have had a free ech bonus used on them?

Will be lost, that hasn't changed in this PR.

logicminimal:
are you thinking of implementing it to the top bar as well?

xlenstra:
Probably not, don't feel like it belongs there and it would clutter the space there. You already see your approximate progress on the techButton itself, as well as how many turns are left. Additional information is not really required, and if you do want it, tapping the techButton will reveal it instantly.

logicminimal:
> Probably not, don't feel like it belongs there and it would clutter the space there. You already see your approximate progress on the techButton itself, as well as how many turns are left. Additional information is not really required, and if you do want it, tapping the techButton will reveal it instantly.

👍 :thumbsup: great job man

yairm210:
As with all UI-centric changes, pics please :)

core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
```diff
@@ -17,7 +17,7 @@ import java.util.*
 import kotlin.collections.ArrayList
 
 
-class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Technology? = null) : PickerScreen() {
+class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Technology? = null, freeTechPick: Boolean = false) : PickerScreen() {
 
     private var techNameToButton = HashMap<String, TechButton>()
     private var isFreeTechPick: Boolean = false
```
yairm210, line 23: This is never changed... so why even have this var, if its only value is that of freeTechPick?

core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
```diff
@@ -262,9 +258,18 @@ class TechPickerScreen(internal val civInfo: CivilizationInfo, centerOnTech: Tec
...
 
-        pick("Research [${tempTechsToResearch[0]}]".tr())
+        val label = "Research [${tempTechsToResearch[0]}]".tr()
+        val techProgression = getTechProgressLabel(tempTechsToResearch)
+        
+        pick("${label}\n${techProgression}")
         setButtonsInfo()
     }
+    
+    private fun getTechProgressLabel(techs: List<String>): String {
```
yairm210, line 268: Nice!!

xlenstra:
> As with all UI-centric changes, pics please :)

Regarding the change of the tech label:

For a single Tech:

Before:
image

After:
image

For multiple techs:
Before:
image

After:
image


The other change is difficult to capture in a screenshot, and should be selfexplanratory enough imo

core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
xlenstra: How should it be implemented? As far as my limited knowledge of Kotlin goes, this is the only way to pass a parameter to the init {} block? If I make it a `private val`, Android Studio starts acting up that it is never used outside of the init {} block.

SomeTroglodyte: If I prefix `private val` on `freeTechPick`, drop isFreeTechPick and replace its uses with the former -> no compiler or inspection complaints. Just the way old name shadowed and until ones... Runs fine, too. AS/Plugin version diff maybe.

So that would be a class-level private val initialized automatically with the same-named constructor parameter. Accessibility in the anonymous function created in init{} not a problem. However, a simple constructor parameter without a backing field should work just as well, as the anon would get it via closure? Nope, `setButtonsInfo` and `selectTechnology` want it - so - first alternative.

xlenstra: Ah thanks, that solved it. Cleaning my branches is something I maybe should do sometime

Issue #4277, opened 2021-06-26
Research agreements in mods with non-default era names are broken


logicminimal:
![Screenshot_20210626-144410_UnCiv.jpg](https://user-images.githubusercontent.com/56484801/123511926-140c1100-d68d-11eb-9b8a-31c8524473cc.jpg)

amogus

SomeTroglodyte:
What's "amogus"? [this one](https://unanything.fandom.com/wiki/Amogus_(disease))?
Also, not enough info to repro. I know there's a mod using water as currency, but which one?

logicminimal:
> What's "amogus"? [this one](https://unanything.fandom.com/wiki/Amogus_(disease))?
> Also, not enough info to repro. I know there's a mod using water as currency, but which one?

Well i definitely didn't mean it as a disease name, it is just a sarcastic version of game among us, i just put it there, there is no reason behind it

it's the deciv redux mod, sorry for lack of information, i assumed you would guess from the screenshot

jahodovykolac:
- Probably not caused by different era names, Water (Gold) is the broken thing here
- This was fixed in a version earlier than Deciv, so it may be a mod issue (outdated mod)

yairm210:
Resolved by #4336

PR #4307, opened 2021-06-29
Fixed bug where research agreements in the Atomic Era cost 0 gold

xlenstra:
Due to the cost of research agreements being hardcoded per era, this wasn't updated after in #4252 the Era names where updated.


logicminimal:
TIL the numbers in research agreement stood for gold, i have one suggestion, add a gold unicode character next to the gold that is required when doing a research agreement

yairm210:
Good find, but wrong fix.
This should be a techColumn parameter, to make it moddable and resolve #4277

PR #4311, opened 2021-06-29
Adds the nation of The Huns

xlenstra:
This PR implements the nation of the Huns, complete with all unique units and nation uniques it has.
All uniques in this PR have been tested and seem to work.

Added uniques:
- "Cities are razed [amount] times as fast"
- "\"Borrows\" city names from other civilizations" - Once all the provided city names have been fully used, start using citynames of other civilizations.
- "Can only attack [unitFilter] units"

Deprecated uniques:
- "Can only attack water"


ravignir:
Why are you introducing "MountedRanged" unit type? It does not exist in civ5.

xlenstra:
> Why are you introducing "MountedRanged" unit type? It does not exist in civ5.

Because Chariot Archer are mounted units according to the civilopedia text, so they can be hit harder by spearman, etc., but they are also ranged units, according to the civilopedia. So to do justice to both of these, I introduced a new unit type.

ravignir:
Civilopedia is wrong. Mounted Ranged units exist only in some Civ5 mods and Civ6 in form of Ranged Cavalry. Spearmen/Pikemen do not have any bonus against them.
z

xlenstra:
Huh, interesting. I'll change it then

ravignir:
z

xlenstra:
My confusion started when reading the 'Strategy' text block, as it remarks: "As a mounted unit, the Chariot Archer is vulnerable to Spearmen." But I guess the developers just forgot to implement this then. Question is: should we? For now I've removed the unitType, but re-adding isn't difficult, and it might be interesting for mods to have access to this type

ravignir:
Maybe poll this.

logicminimal:
> Maybe poll this.

my vote is on giving it ranged mounted type

SomeTroglodyte:
I do like. But since I also like to find [holes](https://elderscrolls.fandom.com/wiki/Category:Skyrim:_Caves): ["or all possible city names have been used, Hunnic cities will take their names from the city lists of civilizations that are **not** in the current game"](https://civilization.fandom.com/wiki/Hunnic_(Civ5)) - that's not in your code, ha! Since this points out there's not enough fallback options - does this mean that it could with a little bad luck leave name `null` and later crash??? `Sequence<T>.last{false}` _is_ `null` isn't it?

xlenstra:
> I do like. But since I also like to find [holes](https://elderscrolls.fandom.com/wiki/Category:Skyrim:_Caves): ["or all possible city names have been used, Hunnic cities will take their names from the city lists of civilizations that are **not** in the current game"](https://civilization.fandom.com/wiki/Hunnic_(Civ5)) - that's not in your code, ha! Since this points out there's not enough fallback options - does this mean that it could with a little bad luck leave name `null` and later crash??? `Sequence<T>.last{false}` _is_ `null` isn't it?

They way I found it on the interwebs, is that they use names from other civilizations in the game. I'll check it in game later to see what exactly is true. 
Theoretically, yes, if you have a very large map with a small number of civs that all create lots and lots of cities, this implementation might result in crashes. But the amount of city names most civilizations have is large enough that that would rarely happen.

android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -254,6 +270,22 @@
...
+		"replaces": "Spearman",
+		"uniqueTo": "Huns",
+		"unitType": "Melee",
+		"movement": 2,
+		"strength": 10,
+		"cost": 75,
+		"requiredTech": "Bronze Working",
+		"obsoleteTech": "Physics",
+		"upgradesTo": "Trebuchet",
+		"uniques": ["+[300]% Strength vs [City]", "No defensive terrain bonus", "+[-33]% Strength when defending",
```
yairm210, line 284: See, "+[-33]% Strength when defending" is why we need uniques that have the + inside the square brackets ;)

logicminimal:
> > I do like. But since I also like to find [holes](https://elderscrolls.fandom.com/wiki/Category:Skyrim:_Caves): ["or all possible city names have been used, Hunnic cities will take their names from the city lists of civilizations that are **not** in the current game"](https://civilization.fandom.com/wiki/Hunnic_(Civ5)) - that's not in your code, ha! Since this points out there's not enough fallback options - does this mean that it could with a little bad luck leave name `null` and later crash??? `Sequence<T>.last{false}` _is_ `null` isn't it?
> 
> They way I found it on the interwebs, is that they use names from other civilizations in the game. I'll check it in game later to see what exactly is true. 
> Theoretically, yes, if you have a very large map with a small number of civs that all create lots and lots of cities, this implementation might result in crashes. But the amount of city names most civilizations have is large enough that that would rarely happen.

what about just naming them city1 city2 after you've run of city names?

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -123,6 +123,19 @@ class CityInfo {
...
         val cityNameRounds = civInfo.citiesCreated / nationCities.size
+        if (cityNameRounds > 0 && civInfo.hasUnique("\"Borrows\" city names from other civilizations in the game")) {
+            val usedCityNames =
+                civInfo.gameInfo.civilizations.flatMap { it.cities.map { city -> city.name } }
+            // We select a random other civ in this game, and choose the last unused name in their list of city names.
+            name = civInfo.gameInfo.civilizations
+                .filter { it.isMajorCiv() && it != civInfo }
+                .random()
+                .nation.cities
+                .last { city -> usedCityNames
```
yairm210, line 134: .last { it !in usedCityNames }

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -361,8 +374,9 @@ class CityInfo {
         cityConstructions.endTurn(stats)
         expansion.nextTurn(stats.culture)
         if (isBeingRazed) {
-            population.addPopulation(-1)
-            if (population.population <= 0) { // there are strange cases where we get to -1
+            val removedPopulation = 1 * civInfo.getMatchingUniques("Cities are razed [] times as fast").sumBy { it.params[0].toInt() }
```
yairm210, line 377: Um, wouldn't this mean that if you have no such uniques, removedPopulation would be 0...?

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -245,13 +245,18 @@ class BaseUnit : INamed, IConstruction {
             unitType.name -> true
             name -> true
             "All" -> true
+            
+            "Melee" -> unitType.isMelee()
+            "Ranged" -> unitType.isRanged()
```
yairm210, line 250: There is a Problem.
The Problem is, that both "Ranged" and "Melee" are both *attack types* and *unit types*. Meaning we need to decide, if something says that it affects Melee units, do we mean 'attack type melee' or 'unit type melee'?
Beforehand, it was 'unit type'. But I'm fine with attack type, as implemented here.
The real goal is to have unit types defined in a separate json, and then we can call them whatever we want, but that's a story for another time...
I'm thinking Archery and Gunpowder to replace Ranged, and Infantry to replace Melee, which differentiates them from Mounted units.

yairm210:
Apparently you have self-conflicts :)

SomeTroglodyte:
> Apparently you have self-conflicts :)

If you mean the checks failing due to "Placeholder `[param]` not found in `Czech` for entry `+[amount]% [stat] from every [param]`" etc, then - No, it's just not yet rebased to include #4309 ?

core/src/com/unciv/logic/city/CityInfo.kt
SomeTroglodyte: Good catch. And do we really want a "2 times" and a "3 times" to stack to "5 times"? Also, a textbook case for `.fold(1)`?

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
ravignir: All land ranged units in civ5 are archery (including bazooka and machine guns), and all land melee units starting with musketman are gunpowder (but tercio is still a melee).

yairm210:
image
No, I meant that he changed both the 'can only attack water units' and the '+1 visibility range' uniques in different PRs, because he's a good boy

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
yairm210: I have no problem dividing land melee into Infantry and Gunpowder, but Bazookas aren't Archery. That's silly.
This does mean we'll need to find a different name for them though.
Or perhaps not. Perhaps Type can *usually* determine Attack Type but can be overridden at the unit level. Food for thought.

android/assets/jsons/Civ V - Vanilla/Units.json
xlenstra: Jup, I've been convinced. Changing all the uniques again is going to make a lot of people angry though.... I'll starting doing this in the future anyway

SomeTroglodyte: ... or from another json: `"+[-33]% Production when constructing [Military] units",` ...

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
ravignir: What about "Shooter"? Would that fit both Archers and Machine Guns?

SomeTroglodyte:
>  I'll fix it soon

Sorry - or, no, I mean it as a kind of show of respect if I dissect your code - there's a few things fishy in there, I see a name shadowed, a `MapEntry in List<String>` and the very first random() is already not Exception-proof. Lemme see if I can come up with a counter-offer of code... You know, the "Heart of Gold" kind of probabilities means "virtually impossible" is equivalent to "bound to happen sooner than you thunk", so better safe than sorry.

SomeTroglodyte:
It's [Attila](https://en.wikipedia.org/wiki/Attila), not Atilla...

SomeTroglodyte:
image

SomeTroglodyte:
:+1: Come to think of it, why not leave "Atilla's Court" and see how long it takes for the first issue marking that as a bug? I like the "with no name". Reference to the 'America' song or the first book in the 'Bourbon kid' trilogy by Anonymous? Again, thanks for all the f~~ish~~un, good work.

xlenstra:
> +1 Come to think of it, why not leave "Atilla's Court" and see how long it takes for the first issue marking that as a bug? I like the "with no name". Reference to the 'America' song or the first book in the 'Bourbon kid' trilogy by Anonymous? Again, thanks for all the f~ish~un, good work.

Thanks for your input, I took inspiration off of it. Two pairs of eyes think about more than 1 one after all. As for the reference, I just though this sounded funny, wasn't really trying to reference anything

SomeTroglodyte:
Who said anything about ___pairs___ of eyes? 
image image

ravignir:
> 
> 
> I do like. But since I also like to find [holes](https://elderscrolls.fandom.com/wiki/Category:Skyrim:_Caves): ["or all possible city names have been used, Hunnic cities will take their names from the city lists of civilizations that are **not** in the current game"](https://civilization.fandom.com/wiki/Hunnic_(Civ5)) - that's not in your code, ha! Since this points out there's not enough fallback options - does this mean that it could with a little bad luck leave name `null` and later crash??? `Sequence<T>.last{false}` _is_ `null` isn't it?

I've tested it and that is only true for Venice. If there is no Venice in the game and they run out of city names they just have empty names. I guess Firaxis was lazy.
z

android/assets/jsons/Civ V - Vanilla/Units.json
yairm210: Since technically none of the old uniques have been deprecated, I think it's legitimate to make the changes to the 'even newer' and then deprecate the older ones.

SomeTroglodyte:
image

Looks nice. The missing symbol is due to my symlinked workdir setup, not a problem with the main atlas.

PR #4317, opened 2021-06-30
The GDR needs an impressive attack sound.

SomeTroglodyte:
Of course, the problem is, it's 12s long and by the time the sound is "firing" the victim is long dead. But it's a cool sound. A shorter one might be [this one](https://freesound.org/people/vultraz168/sounds/334665/)?


yairm210:
You can cut off bits of it with Audacity or other such tools. That's what I did to the sounds I took ;)

SomeTroglodyte:
This already went through 4 or 5 Audacity Effects. Amplify(-3dB), Change Tempo, cut, a little Fade in and Fade out to reduce slight click artifacts, resample... The original is 26 seconds, not 12! Maybe I should prep 1-2 alternates and you choose?

SomeTroglodyte:
So, the sound included in the PR is sourced as listed in credits.md, shortened to 12 seconds as described, encoded to VBR ~ 100kbit/s, giving a file size of 150k.
Same using Change Tempo down to 8s (6s sounds surprisingly boring):
https://github.com/SomeTroglodyte/Scratchpad/blob/master/Media/gdrAttack-8s.mp3
A completely different sound, ([source](https://freesound.org/people/vultraz168/sounds/334660/)) brought down to 2.2s / 30k size:
https://github.com/SomeTroglodyte/Scratchpad/blob/master/Media/gdrAttack-alt.mp3

I almost re-uploaded them to freesound just so there'd be an interactive player, but a short download to hear them isn't too bad.

Shall we use the 30k version for built-in and put the other one in a mod?

yairm210:
Sounds good to me ;)

yairm210:
It seems like your more of a audio pro than I am

In fact
you could say that you edited that file
*with Audacity*

SomeTroglodyte:
Is that.... A PUN??? Now I'll somehow need to get you to include an _actual easter egg_ (or slip it in unnoticed)!

OK - that push just switched the sound and credit. @24kHz and ~66 - no wait that wasn't the _final_ file!

SomeTroglodyte:
OK, that one. I do try to match the majority of the existing sounds in sample and bit rate - and thus it shrank to only 18k.

SomeTroglodyte:
How do cultures that have a non-redneck-western concept of 'easter' do software easter eggs anyway?

SomeTroglodyte:
Oh, and thanks @avdstaaij for the opportunity, I hope you see this as a kind of praise and like it.

yairm210:
> How do cultures that have a non-redneck-western concept of 'easter' do software easter eggs anyway?

I don't work in an English-speaking country, but like almost all software terms, everyone just uses the English word for this ;)

Any time that someone uses a translated, or more culture-specific term, it's either because they're trying to be really 'official' or it's a tongue-in-cheek poke at the fact that everyone uses English more than the country's language...

SomeTroglodyte:
Tongue-in-cheek is nice, especially if it's because you can't stop harassing that place where you accidentally bit yourself 🧊

avdstaaij:
> Oh, and thanks @avdstaaij for the opportunity, I hope you see this as a kind of praise and like it.

You're welcome! The fun thing about adding units like the GDR is that it seems like a big deal even though it is actually one of the easiest things to add :)

PR #4319, opened 2021-06-30
Translate nested placeholders for english too

### LINKED ISSUE #4259: Nested brackets in translatable strings are sometimes partially displayed
Nested brackets in translatable strings like the uniques containing [{Military} {Water}] are partially displayed to the user.

Or: If your language does not have a translation (which is the case for english) for the outer placeholder, e.g. "+[] Movement for all [] units", the square brackets are removed and no further translation is attempted, even if there would be translations available for the parameter / inner parts - or parameter translation would be required in order to process nested placeholders.

Possible approaches:
- Add a `.tr()` in Translations.kt at the end of line 257 (just below the `// Translation placeholder doesn't exist for this language, default to English` comment): Quick and dirty - works in superficial testing but somehow slightly beside the point.
- Instead of shorting out, set `languageSpecificPlaceholder` to `this` and continue, skipping line 267 a bit later. Idea is to let placeholders be processed normally.
- I keep forgetting the third idea.

Separate observation:
- Line 278 Stats... smells like a major performance guzzler - since when is that in there? Since [2020-07-24](https://github.com/yairm210/Unciv/commit/066dd615b4fc76e031ef2ad71703acd18d9ed636#diff-9692dcc94e23f6c10ef9e118f721a33ae286f08c0680f737fc8eaa2463bcc5cc) - OK, don't know why my mind so clearly remembered the previous contains - contains - hashmap lookup version. Someone willing to measure how much faster a short-out test like `if (this[0] in "+-" && `... could be?

I open this as a reminder to myself, to allow and encourage comments, and maybe hints on how that Profiler works (how do I tell it to only instrument parts of the code for profiling? I'm too used to M$ tools here) - while I pause for a day or two.

<details><summary>Current 'fixing the issue' variant of the tr() function:</summary>

```kotlin
fun String.tr(): String {
    if (length == 0) return this

    val activeMods = with(UncivGame.Current) {
        if (isGameInfoInitialized()) gameInfo.gameParameters.mods else translations.translationActiveMods
    }

    // There might still be optimization potential here!
    if (contains('[')) { // Placeholders!
        /**
         * I'm SURE there's an easier way to do this but I can't think of it =\
         * So what's all this then?
         * Well, not all languages are like English. So say I want to say "work on Library has completed in Akkad",
         * but in a completely different language like Japanese or German,
         * It could come out "Akkad hast die worken onner Library gerfinishen" or whatever,
         * basically, the order of the words in the sentence is not guaranteed.
         * So to translate this, I give a sentence like "work on [building] has completed in [city]"
         * and the german can put those placeholders where he wants, so  "[city] hast die worken onner [building] gerfinishen"
         * The string on which we call tr() will look like "work on [library] has completed in [Akkad]"
         * We will find the german placeholder text, and replace the placeholders with what was filled in the text we got!
         */

        // Convert "work on [building] has completed in [city]" to "work on [] has completed in []"
        val translationStringWithSquareBracketsOnly = this.getPlaceholderText()
        val language = UncivGame.Current.settings.language
        // That is now the key into the translation HashMap!
        val translationEntry = UncivGame.Current.translations
                .get(translationStringWithSquareBracketsOnly, language, activeMods)

        var (languageSpecificPlaceholder, originalEntry) = 
            if (translationEntry == null || !translationEntry.containsKey(language))
                // Translation placeholder doesn't exist for this language, default to English
                this to this
            else
                translationEntry[language]!! to translationEntry.entry

        // Take the terms in the message, WITHOUT square brackets
        val termsInMessage = this.getPlaceholderParameters()
        // Take the term from the placeholder, INCLUDING the square brackets
        val termsInTranslationPlaceholder = squareBraceRegex.findAll(originalEntry).map { it.value }.toList()
        if (termsInMessage.size != termsInTranslationPlaceholder.size)
            throw Exception("Message $this has a different number of terms than the placeholder $translationEntry!")

        for (i in termsInMessage.indices) {
            languageSpecificPlaceholder = languageSpecificPlaceholder.replace(termsInTranslationPlaceholder[i], termsInMessage[i].tr())
        }
        return languageSpecificPlaceholder      // every component is already translated
    }

    if (contains('{')) { // sentence
        return curlyBraceRegex.replace(this) { it.groups[1]!!.value.tr() }
    }

    if (this[0] in "+-" && Stats.isStats(this)) return Stats.parse(this).toString()

    return UncivGame.Current.translations.getText(this, UncivGame.Current.settings.language, activeMods)
}
```
</details>

Oh - if some wants to use these changes in any way, go ahead, no attribution necessary.

SomeTroglodyte:
Time to resolve #4259 and get it out of the way. Enough pondering, if I haven't had a better idea by now this must be it.

image

### DIFF
--- a/core/src/com/unciv/models/translations/Translations.kt
+++ b/core/src/com/unciv/models/translations/Translations.kt
@@ -231,7 +231,7 @@ fun String.tr(): String {
     }
 
     // There might still be optimization potential here!
-    if (contains("[")) { // Placeholders!
+    if (contains('[')) { // Placeholders!
         /**
          * I'm SURE there's an easier way to do this but I can't think of it =\
          * So what's all this then?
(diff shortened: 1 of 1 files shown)


yairm210:
Nice work!!!

PR #4323, opened 2021-07-01
Fixed effect string of "Flight Deck II"

Mape6:
Deleted the space in the beginning of the string and deleted all unnecessary translation lines.


yairm210:
Done as part of translation merge to avoid conflicts

Issue #4318, opened 2021-06-30
Some issues with translations


sockybob:
Translation for `'Our scientific power has outrun our spiritual power, we have guided missiles and misguided men.' – Martin Luther King Jr.` which was previosly here are now gone.

Edit 1:
~~Seems to be true only for russian and i already fixed it so not a issue.~~

Edit 2:
Ok it also seems to be the case for italian. Sorry for confuse, it's hard for me to check all that stuff on mobile. Still not sure about other languages.

There are two `Can carry [amount] extra [param] units` strings which is differ only by space at the beggining.

Quotes have inconsistent dash characters (between the quote and author), in some cases there's long variant of it and in others it's short.

Mape6:
The quote from Martin Luther King Jr. was changed with #4297 by me
The effect string of "Flight Deck II" will be changed with #4323

Do you have examples for the different dashes?

sockybob:
> Do you have examples for the different dashes?

`'Nothing travels faster than light with the possible exception of bad news, which obeys its own special rules.' – Douglas Adams` with long dash

`'The wonder is, not that the field of stars is so vast, but that man has measured it.'  - Anatole France` with short dash

`'The more we elaborate our means of communication, the less we communicate.' – J.B. Priestly` with long dash

`'All men can see these tactics whereby I conquer, but what none can see is the strategy out of which victory is evolved.' - Sun Tzu` with short dash

xlenstra:
Ah, I see what happened here. Those two quotes with long dashes I copy+pasted straight from the fandom, while the others were probably manually typed in, leading to different dashes.

SomeTroglodyte:
...And the fandom user probably supplied a short dash, and the fandom _server_ ""thought"" a long one might look better...

Technically, we can leave the json and code as is so no existing translation is lost, and fix that in translations only - including English.properties... I always thought those loooong text walls shouldn't be used as lookup keys (which they are if they appear on the left side of the equals in a translation file), the key should be some shorter mnemonic, and the long novels should exist _only_ in the translation files.

yairm210:
Already done :)

PR #4329, opened 2021-07-01
Fixed some small mistakes with building & tech quotes

### LINKED ISSUE #4326: Quote of Sun Tzu for Stealth
@xlenstra Added Stealth with the quote:
Be extremely subtle, even to the point of formlessness, be extremely mysterious, even to the point of soundlessness. Thereby you can be the director of the opponent' state.

The original quote from [here](https://www.goodreads.com/work/quotes/3200649) is:
Be extremely subtle even to the point of formlessness. Be extremely mysterious even to the point of soundlessness. Thereby you can be the director of the opponent'_s_ fate.

Should we change that to the "original" quote?

xlenstra:
Fixes part of #4318 
Fixes #4326


yairm210:
Replaced as part of translation merge to keep existing translations

PR #4292, opened 2021-06-27
Updated promotions - make more generalizable, update to G&K

xlenstra:
This PR updates all promotions to their G&K versions. Each commit of this PR will update a single promotion, so the changes can more easily be reviewed. If this PR turns out too big, I'll split it up into multiple PR's.

New uniques:
- "Heal this Unit by [amount] HP"
- "Doing so will consume this opportunity to choose a Promotion" - This also implies the promotion is not permanent and can be awarded next time the unit levels.
- "[amount] Range"
- "[amount] additional attacks per turn"
- "[amount] HP when healing"
- "All adjacent units heal [amount] extra HP when healing"
- "[amount] HP when healing in [tileFilter] tiles" 
- "[amount] Visibility Range"
- "[amount] Movement"

Deprecated uniques:
- "Heal this Unit by 50 HP; Doing so will consume this opportunity to choose a Promotion"
- "+1 Range"
- "+2 Range"
- "1 additional attack per turn"
- "This unit and all others in adjacent tiles heal 5 additional HP. This unit heals 5 additional HP outside of friendly territory."
- "+1 Visibility Range"
- "+2 Visibility Range"
- "+1 Movement"

PR #4331, opened 2021-07-01
Updated and generalized more promotions

xlenstra:
This PR is the follow-up for #4292, generalizing and updating basically all other promotions.
Again, each commit only changes a single promotion to allow for easier reviewing.
The following promotion was added:
- Supply: Water units can heal outside of friendly territory (15 hp/turn). Note that in the base game this was locked behind Tier II promotions. I choose to lock it behind Tier III promotions instead as it is very powerful in my opinion.

Added uniques:
- "[amount] extra interceptions may be made per turn"
- "[signedAmount]% Strength when defending vs [unitFilter]"
- "[signedAmount]% Strength for enemy [unitFilter] units in adjacent tiles"
- "May heal outside of friendly territory" - for water units
- [signedAmount]% Damage when intercepting

Deprecated uniques:
- "1 extra Interception may be made per turn"
- "+25% Defence against ranged attacks"
- "-10% combat strength for adjacent enemy units"
- "Bonus when intercepting [amount]%"

### DIFF
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -51,17 +51,19 @@ object BattleDamage {
                 }
             }
 
-            var adjacentUnitBonus = 0
+            val adjacentUnits = combatant.getTile().neighbors.flatMap { it.getUnits() }
+            
             for (unique in civInfo.getMatchingUniques("+[]% Strength for [] units which have another [] unit in an adjacent tile")) {
                 if (combatant.matchesCategory(unique.params[1])
-                    && combatant.getTile().neighbors.flatMap { it.getUnits() }
-                    .any { it.civInfo == civInfo && it.matchesFilter(unique.params[2]) } 
+                    && adjacentUnits.any { it.civInfo == civInfo && it.matchesFilter(unique.params[2]) } 
                 ) {
-                    adjacentUnitBonus += unique.params[0].toInt()
+                    modifiers.add("Adjacent units", unique.params[0].toInt())
                 }
             }
-            if (adjacentUnitBonus != 0)
-                modifiers["Adjacent unit"] = adjacentUnitBonus
+            
+            for (unique in adjacentUnits.flatMap { it.getMatchingUniques("[]% Strength for [] units in adjacent tiles you are at war with") })
+                if (combatant.matchesCategory(unique.params[1]))
+                    modifiers.add("Adjacent enemy units", unique.params[0].toInt())
 
             val civResources = civInfo.getCivResourcesByName()
             for (resource in combatant.unit.baseUnit.getResourceRequirements().keys)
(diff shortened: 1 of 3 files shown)


SomeTroglodyte:
> Cover+33, Ambush+33

You sure? [This](https://civilization.fandom.com/wiki/Promotions_(Civ5)) doesn't mention somesuch

> I choose to lock it behind Tier III

I'd leave such intentional "deviations" from original rules to mods.
Also, I would interpret the wiki as 15 _total_ per turn. Wouldn't these uniques add up two healing sources? (just a doubt, I didn't check)

xlenstra:
> > Cover+33, Ambush+33
> 
> You sure? [This](https://civilization.fandom.com/wiki/Promotions_(Civ5)) doesn't mention somesuch
> 

I checked the ingame civilopedia with G&K enabled and BNW disabled, I don't think it can get more accurate than that

> > I choose to lock it behind Tier III
> 
> I'd leave such intentional "deviations" from original rules to mods.

I'll leave this to Yairm to decide. It is really strong and basically removes one of the major downsides of a navy.

> Also, I would interpret the wiki as 15 _total_ per turn. Wouldn't these uniques add up two healing sources? (just a doubt, I didn't check)

The way it is implemented now, is that water units still heal 0 per turn when outside friendly territory, but with this unique they are allowed to have other uniques providing them bonuses. This is so that mod makers can choose themselves how strong the Supply bonus is. Relveant code is in in the `rankTileForHealing()` in the `mapUnit` class.

SomeTroglodyte:
:+1: Oooooooooooh so you have access to      _t h e   O r i g i n a l_     ...
OK, you're the official new primary source of accurate information :wink: ... Besides - thorough work - kudos.

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -51,17 +51,19 @@ object BattleDamage {
...
+                    && adjacentUnits.any { it.civInfo == civInfo && it.matchesFilter(unique.params[2]) } 
                 ) {
-                    adjacentUnitBonus += unique.params[0].toInt()
+                    modifiers.add("Adjacent units", unique.params[0].toInt())
                 }
             }
-            if (adjacentUnitBonus != 0)
-                modifiers["Adjacent unit"] = adjacentUnitBonus
+            
+            for (unique in adjacentUnits.flatMap { it.getMatchingUniques("[]% Strength for [] units in adjacent tiles you are at war with") })
```
yairm210, line 64: Do you think 'you are at war with' is more clear-cut than 'enemy'?

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -842,9 +845,13 @@ class MapUnit {
     }
 
     fun canIntercept(attackedTile: TileInfo): Boolean {
-        if (attacksThisTurn > 1) return false
         if (interceptChance() == 0) return false
-        if (attacksThisTurn > 0 && !hasUnique("1 extra Interception may be made per turn")) return false
+        val maxAttacksPerTurn = 1 + 
+            getMatchingUniques("[] extra interceptions may be made per turn").sumBy { it.params[0].toInt() } + 
+            // Deprecated since 3.15.7
+                getMatchingUniques("1 extra interception may be made per turn").sumBy { 1 }
```
yairm210, line 852: sumBy { 1 } -> .count()

core/src/com/unciv/logic/battle/BattleDamage.kt
xlenstra: The main problem I found when doing these kind of things, is that the order of adjectives doesn't line up.
For example when using "[]% Strength for enemy [] units" you also have "[]% Strength for enemy All units" which doesn't sound right, but if you use "[]% Strength for [] enemy units" you also have "[]% Strength for land enemy units" which also doesn't sound right. Moving the enemy to a different spot solves this, but does make it longer.

For now I've changed it to "[]% Strength for all enemy [] units in adjacent [tileFilter] tiles", which has the same problem but with the tileFilter, but that is the way it is.

yairm210: Sounds thought-out :)

PR #4334, opened 2021-07-01
Spruced up Civilopedia - phase 3 - Interface, flavour text, new Tutorial

SomeTroglodyte:
This code from PR #4011 was over a month old, so I pulled the PR and reworked it into the current codebase, up to The Huns. See that PR for explanations.

Reminder what this PR cannot yet do but what the completed project can do:
image
image

The code I did these with is [in this branch](https://github.com/SomeTroglodyte/Unciv/tree/Civilopedia-complete-20210701) or [in this apk](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Debug%20apk's/Unciv-debug-Civilopedia-Flavour-005.apk?raw=true) just in case.

... and I still forgot to include a cosmetic fix to the left ScrollPane positioning I've got shelved somewhere ...


core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
```diff
@@ -159,31 +179,35 @@ class CivilopediaScreen(
                     CivilopediaEntry(
                         it.name,
                         it.getDescription(false, null, ruleset),
-                        CivilopediaCategories.Wonder.getImage?.invoke(it.name, imageSize)
+                        CivilopediaCategories.Wonder.getImage?.invoke(it.name, imageSize),
+                        (it as? ICivilopediaText).takeUnless { ct -> ct==null || ct.isCivilopediaTextEmpty() }
```
yairm210, line 183: This line repeats itself, couldn't the CivilopediaEntry constructor handle this?

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -111,3 +177,106 @@ object MarkupRenderer {
...
+class SimpleCivilopediaText(lines: List<FormattedLine>, val isComplete: Boolean = false) : CivilopediaText() {
+    init {
+        civilopediaText = lines
+    }
+    override fun hasCivilopediaTextLines() = true
+    override fun replacesCivilopediaDescription() = isComplete
+    constructor(strings: Sequence<String>, isComplete: Boolean = false) : this(
+        strings.map { FormattedLine(it) }.toList(), isComplete)
+    constructor(first: Sequence<FormattedLine>, strings: Sequence<String>, isComplete: Boolean = false) : this(
+        (first + strings.map { FormattedLine(it) }).toList(), isComplete)
```
yairm210, line 195: why .toList() after the .map{}? You're getting a sequence, which is already good?

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -111,3 +177,106 @@ object MarkupRenderer {
...
+    /** Assemble json-supplied lines with automatically generated ones.
+     *
+     * The default implementation will insert [getCivilopediaTextLines] before the first [linked][FormattedLine.link] [civilopediaText] line and [getCivilopediaTextHeader] on top.
+     *
+     * @param ruleset The current ruleset for the Civilopedia viewer
+     * @return A new CivilopediaText instance containing original [civilopediaText] lines merged with those from [getCivilopediaTextHeader] and [getCivilopediaTextLines] calls.
+     */
+    fun assembleCivilopediaText(ruleset: Ruleset): CivilopediaText {
+        val outerLines = civilopediaText.iterator()
+        val newLines = sequence {
```
yairm210, line 254: Oh, nice! Clever use of yield!

logicminimal:
haha copyright from artwork goes brrrrr

yairm210:
This looks really good. I think the CivilopediaText class is a bit overzealous, but I can't argue with the results.

core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
SomeTroglodyte: It's nearing the point where we would expect a `CivilopediaCategories.values().forEach{}`, right? No accident that, but - we'd need a common interface to all those classes, and there isn't one - yet. Right now the `it` in all those 'factories' varies over incompatible types. Studio even notices some of those as? will never succeed and the compiler probably optimizes away the whole line...

And there you have it. My mental model is wishing for that loop, and the code would exist only once, and then moving parts into the helper class (at the moment pure data with just one additional partial-copy kind of constructor) would tend to reduce, not improve readability. So - the other way didn't even occur to me. Hmmm... I haven't yet solved all the impediments on the way to that loop, so... And then my personal dislike of code relying on Any... Can we get away with a constructor signature `CivilopediaEntry (name: String, description: String, category: CivilopediaCategories, object: INamed)` for now?

Suggestion: Let's agree that part isn't our final word, and revisit later. Discuss interface hierarchy elsewhere? At the moment the new abstraction tool is named `ICivilopediaText` because its goal is to enable text handling - but it could be a subclass to INamed making it.name easier to access, and it could abstract the common getImage instead of that CivilopediaCategories delegates kludge, and it could abstract the ruleset object source enumerators. Then it would be a `ICivilopediaObject`, and remain as only necessary parameter to the CivilopediaEntry constructor. But where in the 'inheritance tree' would it be in relation to NamedStats (which isn't an interface)?

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
SomeTroglodyte: We have `constructor(List<FormattedLine>, Boolean)` and  `constructor(Sequence<String>, Boolean)` up to that point, not a `constructor(Sequence<FormattedLine>, Boolean)` - trivial to write but would be single-use and it would toList() anyway. The final storage is a List, int's not abstract, because elsewhere the superclass wants to be able to have that easily deserialized from json.

I think that's a 'no that kind of sequence is not good yet'

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
SomeTroglodyte: Still one of the parts I wasn't too happy with - but I wanted a specific ability, namely mixing json lines with generated lines in a grouped kind of order, so it came out much longer than I'd have preferred.

SomeTroglodyte:
> haha copyright from artwork goes brrrrr

Which is why I won't be including these assets in the final form. Modders can then do that. I just modded myself a little to build an appetizing teaser without needing to gimp it.

SomeTroglodyte:
> overzealous

_Quite_ true. I got `bold` and `italic` defined in the final version without a viable implementation, just in case, for future expandability. It's meant to be as universally useable as feasible.

(Actually, bolding using MSDF fonts -using com.maltaisn.msdfgdx- was very promising, and all text in small sizes was much more readable, but that had baseline issues my tiny brain didn't mange to break open and I shelved that about six weeks ago)

SomeTroglodyte:
Big question: can I partition the rest into separate PR's at all? ... Let's go decorations+1 class, then class by class perhaps...

Issue #4152, opened 2021-06-15
from what i have seen on wiki you get different bonuses for starting on different eras


logicminimal:
but this does not seem to be the case on unciv

yairm210:
#663

PR #4336, opened 2021-07-02
Split off properties of era's to another JSON file

xlenstra:
This PR adds another JSON file, eras.json, which contains data that applies to the entire era. Examples of this are:
- Cost of research agreements
- Color of the icons in the techPickerScreen
- Which and how many units should spawn for each nation at the beginning of the game
- Which buildings should be build automatically when you use a settler

This PR then also implements the following features:
- When starting in a non-ancient era, settled cities will already have some population and buildings in them
- When starting in a non-ancient era, the units you receive differ, from just a warrior in ancient era to 5 marines in the information era.
- When starting in a non-ancient era, some culture and gold is given to each civilization
- When starting in a non-ancient non-classical era, old wonders are no longer buildable (>1 era's since game start)
- Cost of Research agreements can be modded for each era (implements #4307 better)
- All of the above can also be modded easily

NOTE: THIS PR VERY LIKELY BREAKS MOD COMPATIBILITY!
Previously, which units each civilization had at the start of the game was only part of the Difficulty.json file. Due to it now being a combination of the difficulty and starting era, I have removed the default starting units from the Difficulty.json file, and moved them to the new eras.json file. If the eras.json file is missing, it is assumed that 1 warrior and 1 settler should be added. This results in all mods with also having these units in the difficulty.json file adding these units twice, resulting in 2 warriors and 2 settlers spawning at the beginning. There are workarounds to fix this, but it is currently 3AM and I'm not willing to look into those right now, maybe later


core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -96,6 +96,18 @@ class CityInfo {
...
+        val ruleset = civInfo.gameInfo.ruleSet
+        val startingEra = civInfo.gameInfo.gameParameters.startingEra
+        if (startingEra in ruleset.eras) {
+            population.setPopulation(ruleset.eras[startingEra]!!.settlerPopulation)
+            for (building in ruleset.eras[startingEra]!!.settlerBuildings) {
+                if (ruleset.buildings[building]!!.isBuildable(cityConstructions)) {
+                    cityConstructions.addBuilding(building)
+                }
+            }
+        }
```
yairm210, line 110: This looks like it applies to ALL cities that we'll found in the game. Is that how it works in Civ V? I really don't know, I don't think I ever started on a later era

core/src/com/unciv/models/ruleset/Difficulty.kt
```diff
@@ -14,7 +14,7 @@ class Difficulty: INamed {
     var policyCostModifier:Float = 1f
     var unhappinessModifier:Float = 1f
     var barbarianBonus:Float = 0f
-    var startingUnits = ArrayList<String>()
+    var startingUnits = ArrayList<String>() // I would like to rename these to 'bonus starting units' now that they also exist in Eras.json, but something something mod compatibility
```
yairm210, line 17: Au Contraire Snapeydoodle
Creating a new bonusStartingUnits would actually help differentiate the 'new' functionality from the 'old' functionality and help with the gradual deprecation!

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -371,8 +375,27 @@ class Ruleset {
...
+            for (building in era.value.settlerBuildings)
+                if (building !in buildings)
+                    lines += "Nonexistent building ${building} built by settlers when starting in ${era.key}"
+            if (era.value.startingMilitaryUnit !in units)
+                lines += "Nonexistent unit ${era.value.startingMilitaryUnit} marked as starting unit when starting in ${era.key}"
+            if (era.value.researchAgreementCost < 0 || era.value.startingSettlerCount < 0 || era.value.startingWorkerCount < 0 || era.value.startingMilitaryUnitCount < 0 || era.value.startingGold < 0 || era.value.startingCulture < 0)
+                lines += "Unexpected negative number found while parsing era ${era.key}"
+            if (era.value.settlerPopulation <= 0)
+                lines += "Population in cities from settlers must be strictly positive! Found value ${era.value.settlerPopulation} for era ${era.key}"
+        }
```
yairm210, line 396: 👍🏿

yairm210:
This is very well done!
Apart from 'add bonusStartingUnits instead of existing startingUnits' I really have nothing to add!

logicminimal:
> This PR adds another JSON file, eras.json, which contains data that applies to the entire era. Examples of this are:
> - Cost of research agreements
> - Color of the icons in the techPickerScreen
> - Which and how many units should spawn for each nation at the beginning of the game
> - Which buildings should be build automatically when you use a settler
> 
> This PR then also implements the following features:
> - When starting in a non-ancient era, settled cities will already have some population and buildings in them
> - When starting in a non-ancient era, the units you receive differ, from just a warrior in ancient era to 5 marines in the information era.
> - When starting in a non-ancient era, some culture and gold is given to each civilization
> - When starting in a non-ancient non-classical era, old wonders are no longer buildable (>1 era's since game start)
> - Cost of Research agreements can be modded for each era (implements #4307 better)
> - All of the above can also be modded easily
> 
> 
> NOTE: THIS PR VERY LIKELY BREAKS MOD COMPATIBILITY!
> Previously, which units each civilization had at the start of the game was only part of the Difficulty.json file. Due to it now being a combination of the difficulty and starting era, I have removed the default starting units from the Difficulty.json file, and moved them to the new eras.json file. If the eras.json file is missing, it is assumed that 1 warrior and 1 settler should be added. This results in all mods with also having these units in the difficulty.json file adding these units twice, resulting in 2 warriors and 2 settlers spawning at the beginning. There are workarounds to fix this, but it is currently 3AM and I'm not willing to look into those right now, maybe later

Resolves #4152, great work man!

core/src/com/unciv/logic/city/CityInfo.kt
ravignir: yup it applies to all cities

xlenstra: I haven't tested this in the original, but I remember reading somewhere (probably the fandom or some random forum post?) that it applies to all cities, even those not founded by the originally spawned settlers

PR #4341, opened 2021-07-02
Added missing notification when a privateer is attacked by a unit and plunders goods as a result

xlenstra:


yairm210:
If the same thing is true both ways, make a plunder(plunderer, plundered) function to nry

PR #4349, opened 2021-07-02
Unified separators, CheckBox helper

SomeTroglodyte:
Another 'my bad' found - turning the "Permanent audiovisual" button into a CheckBox if forgot to actually translate. Apart from that, this PR changes just one barely visible detail, and is mostly preparation for other UI work (which I cherry picked this out from).

- A String Extension mirroring `.toTextButton` and `.toLabel` (could have done a Subclass too, threw some dice, this won), shortens `CheckBox` instantiation by a line or two, auto-`tr()`
- Visual tweak: Tiny padding between the round "Checkmark" icon and the actual text - current ones often look glued together. 1f seems like nothing but looks _much_ better IMHO. Actually implemented for just the one `CheckBox` I needed the missing `tr()` fixed, the rest we can treat when we come across them. Those on NewGameScreen are queued for a later PR.
- Treated my own to-do "unify separators". The modified horizontal and vertical separator factories are much more flexible. Simulating a `SplitPane` handle exactly (which I did on the ModManager) is moved here, colours and width can be asked for, and the automatic colspan can be instead given explicitly (example: called on an empty Table() the old one simply disappears. You could add null Actor cells until you get enough columns _or_ -simpler- use the new optional parameter)
- There was only one case of the separator being stored for later use, setting visibility. That needed changing, as the separator is now a `Cell<Container>` instead of a `Cell<Image>`. Using Actor is OK and will probably survive the next redesign (it survived the move from a Table to Container which is more light-weight).


yairm210:
Why the Container though? Why not just a normal Image?

SomeTroglodyte:
Container is the lightweight single-Cell Table, and has a background property. I trust Gdx to handle that more efficiently than _scaling_ a uniformly colored pixel array in some way to fill the required area. SplitPane does it's handle like that under the hood if I remember correctly.

SomeTroglodyte:
But you're right, that's unproven, and comparing Profiling runs should be in order. I'm just too lazy, and ass-ume what my guts say is most efficient...

yairm210:
Both do the exact same thing - scaling a 1px by 1px image. I would assume the opposite, that they would both do the exact same thing...

SomeTroglodyte:
OK, on that point the approaches are equivalent. What about color being passed in a simpler way, you don't need to worry about loading a png, and you don't need to worry whether tinting a black or a white pixel is the correct way? Either way - you put the hammer down I'll obey. With delay. Taming the ScrollPane - murders - Slider thing is _so_ challenging, and then outdoor time (no sleep).

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -177,13 +206,37 @@ fun String.toLabel(fontColor: Color = Color.WHITE, fontSize:Int=18): Label {
     return Label(this.tr(), labelStyle).apply { setFontScale(fontSize/Fonts.ORIGINAL_FONT_SIZE) }
 }
 
-fun Label.setFontColor(color: Color): Label { style= Label.LabelStyle(style).apply { fontColor=color }; return this }
+/**
+ * Translate a [String] and make a [CheckBox] widget from it.
+ * @receiver The text to use as [CheckBox] label
+ * @param initialState The initial checked state
+ * @param changeAction A callback to call on change with a boolean lambda parameter containing the current [isChecked][CheckBox.isChecked]
+ * @return the initialized [CheckBox]
```
yairm210, line 214: Again, TMI. String.toCheckBox returns THE CHECKBOX? initialState is "The initial checked state"? Golly goshness, Batman!
The changeAction is the only non-trivial information here.

SomeTroglodyte: True.

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -93,19 +93,48 @@ fun Actor.addBorder(size:Float, color: Color, expandCell:Boolean=false): Table {
...
-    val image = ImageGetter.getWhiteDot()
-    return add(image).width(2f).fillY()
+/**
+ * Create a vertical separator as an empty Container with a colored background.
+ * 
+ * Note: Unlike the horizontal [addSeparator] this cannot automatically span several rows. Repeat the separator if needed.
+ * @receiver A Gdx [Table] to add this separator to
+ * @param color The color to use
+ * @param width Width of the separator
+ * @return The Cell that has been added to the receiver Table
```
yairm210, line 131: I know I'm being a bit pedantic, but if it's either removing these now or being stuck with unneeded lines in these files for probably forever, then the first it is.

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -93,19 +93,48 @@ fun Actor.addBorder(size:Float, color: Color, expandCell:Boolean=false): Table {
...
+    }
+}
+
+/**
+ * Create a horizontal separator as an empty Container with a colored background.
+ * @receiver A Gdx [Table] to add this separator to
+ * @param color The color to use
+ * @param colSpan Number of columns to span (omit to use the current column count)
+ * @param height Height of the separator
+ * @return The Cell that has been added to the receiver Table
```
yairm210, line 112: And here

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
SomeTroglodyte: > pedantic

Really? I'll need a big can of coffee before judging that... :judge: :hammer:

> now or forever

Precisely my motivation too - this is groundwork so others will build on it so it should be crystal clear to coders using it. IMHO the whole _file_ could do with more precise Kdoc's, I see myself revisiting it all the time to see how things tick or what other overloads might be there or... And those Studio tooltips could help with that (even if they're the main source of hard crashes). It tempts me again and again to add them to every single element in the file, and then I see the sheer number of them and get discouraged...

(Don't forget to expressly _say_ the final word on getImage vs getBackground)

SomeTroglodyte: Yes - the function signature already offered by Studio in those tooltips is definitely enough here.

SomeTroglodyte:
That's the doc comments trimmed.

Now - getBackground or
```kotlin
/** get background _image_ for a new separator */
private fun getSeparatorImage(color: Color) = ImageGetter.getDot(
    if (color.a != 0f) color else CameraStageBaseScreen.skin.get("color", Color::class.java) //0x334d80
)
```
... And how to get the skin's 'SplitPane' handle color: As shown, or `CameraStageBaseScreen.skin.get("splitpane-vertical-c", Skin.TintedDrawable::class.java).color` - theoretically cleanest, or - hardcoded. And - want to cache it?

You could also tell me off for using `Color.CLEAR` as shortcut-by-convention to mean that certain skin element. An alternative would be to declare some accessor somewhere, maybe as Color extension or, if it's to be cached, in `CameraStageBaseScreen.Companion`?

yairm210:
I think image beats background for quick understanding of what's going on.

yairm210:
I think all separators should be assumed white (like the text) unless otherwise stated

SomeTroglodyte:
> image

:+1: Decided, then. Finishing something else, then this.

> white

Stays the default. Other colors can be useful as grid guides and the like.

SomeTroglodyte:
That 'patch3' is the move back to Image, and for the Color thing I left that draft as it was, because Skin accesses _should_ be fast.

PR #4350, opened 2021-07-02
Spruced up Civilopedia - phase 4 - Visual candy, Units

SomeTroglodyte:
So, I chose to include in this next step both the _entire rest_ of the renderer abilities and one category: Units.

- Smaller issue: was dependent on newer addSeparator() so I nerfed that feature a bit - has a todo comment
- Fix: The selection of a specific entry on first open operated on unfinished layout and so some entries at the end of a list were imperfect - in view but ScrollPane scrolled too far - pack()ed
- I slipped in an unrelated fix caused by the 'Huns' being renamed to 'The Huns' - github will probably mess up the history of the file this way, you might prefer to do it differently. I just wanted to have it pretty while testing... 
- Big issue this exemplifies: The existing multiline-String-producing getDescription or equivalent functions will get obsolete or partially so. Approaches: ignore now and trim later, or optimize - I did the latter in this case, which led to a half-rewrite, a parameter less, so some files got touched which approach number one would have left alone (Nation and ConstructionInfoTable). Also, one could leave the call in the c'pedia category preparation alone or kill it - here I killed it (which I defined as being linked to replacesCivilopediaDescription in ICivilopediaText - if and only if that returns true the old-style text in the c'pedia preparation can be dropped), and completed the 'kill' for the already implemented tutorials. Input welcome.

- This also shows a visibly mixed Civilopedia can work just fine, and publishing a public version while the project is still underway wouldn't break things (after all some elements of the architecture exist only to allow this) - at worst we'd provoke some touchy fellows to complain you can follow a link from Slinger to Incas but not back (yet). Your call.

- Food for thought: At some point we should redesign the actual pedia screen - pre-'rendering' _all_ content while it is being created is a bit expensive. Some lazy and caching might be in order - not only for the CPU spent but for the GC stress - just how exactly to achieve that best is still extremely hazy in my skull.


core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -8,28 +8,66 @@ import com.badlogic.gdx.utils.Align
...
+ *                      1 = aligned to a little more than 3 icons, each step above that adds 30f.
+ * @param padding       Defines vertical padding between rows, defaults to 5f.
+ * @param color         Sets text color, accepts Java names or 6/3-digit web colors (e.g. #FFA040).
+ * @param separator     Renders a separator line instead of text. Can be combined only with color and size (the latter being line width, defaulting to 2)
+ * @param italic        Renders text in italic font style (not implemented)
+ * @param bold          Renders text in bold font style (not implemented)
+ * @param strike        Renders text in strikethrough font style (not implemented)
+ * @param shadow        Renders text with a dark shadow (not implemented)
+ * @param starred       Decorates text with a star icon - if set, it receives the [color] instead of the text.
+ * @param centered      Centers the line (and turns off wrap)
```
yairm210, line 51: I feel that much of this text is unnecessary. Where relevant, this should be on each field, and not on the class header.

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -53,31 +91,93 @@ class FormattedLine (
         }
     }
 
+    /** Translates [centered] into [libGdx][Gdx] [Align] value */
+    val align
```
yairm210, line 95: If it's computed on-the-fly it should be a function to indicate that. Also this saves us a line.

fun align() = ...

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -53,31 +91,93 @@ class FormattedLine (
...
+        get() = if (centered) Align.center else Align.left
+
+    private val iconToDisplay
```
yairm210, line 98: Same

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -8,28 +8,66 @@ import com.badlogic.gdx.utils.Align
...
+    val size: Int = Int.MIN_VALUE,
     val header: Int = 0,
+    val indent: Int = 0,
+    val padding: Float = Float.NaN,
+    val color: String = "",
     val separator: Boolean = false,
+    val italic: Boolean = false,        // Not implemented - would need separate font
+    val bold: Boolean = false,          // Not implemented - see 'MSDF'?
+    val strike: Boolean = false,        // Not implemented - could do a simple line
+    val shadow: Boolean = false,        // Not implemented - see 'MSDF'.
```
yairm210, line 68: Anything that's not implemented should not be added to the class yet. This is just more unnecessary code for 'stuff that should be' and 'whoops, you used something that isn't implemented'. This doesn't help at all.

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -101,17 +201,61 @@ class FormattedLine (
...
+        if (linkType != LinkType.None && !noLinkImages) {
+            table.add( ImageGetter.getImage(linkImage) ).size(iconSize).padRight(iconPad)
+            iconCount++
+        }
+        if (iconToDisplay.isNotEmpty() && !noLinkImages) {
+            val parts = iconToDisplay.split('/', limit = 2)
+            if (parts.size == 2) {
+                val category = CivilopediaCategories.fromLink(parts[0])
+                if (category != null) {
+                    if (category.getImage != null) {
```
yairm210, line 221: Unify

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -101,17 +201,61 @@ class FormattedLine (
...
+                        @Suppress("UNNECESSARY_NOT_NULL_ASSERTION") 
+                        val image = category.getImage!!(parts[1], iconSize)
+                        if (image != null) {
+                            table.add(image).size(iconSize).padRight(iconPad)
+                            iconCount++
+                        }
+                    }
+                }
+            }
+        }
```
yairm210, line 234: In general there is way too much indentation here. My eyes

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
SomeTroglodyte: :+1: Understood. In that case _preparing_ (but not implementing) mod checker to tell us about incorrect uses goes shelved, too, for now.

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
SomeTroglodyte: :+1: I wanted a complete documentation _somewhere_ and didn't think enough about the usefulness in terms of _how it looks in Studio_, and accessibility without actually opening CivilopediaText.kt. Tricky - changing `@param` to `@property` seems fine at first, then just `[mention]` the properties in the description shortly, and the class tooltip looks _much_ easier. But the individual properties - ugh:
   image
I'd say that's a bug in the IDE - but how do we deal with that? Gimme some time.

SomeTroglodyte: FYI: "`@property name﻿
Documents the property of a class which has the specified name. This tag can be used for documenting properties declared in the primary constructor, where putting a doc comment directly before the property definition would be awkward.`"
... But the "awkward" version is the one rendered properly in tooltips. So -  use `@property` and open another can of worms, create jetbrains account, start the war to get that ugliness fixed. ***or*** - use ugly source and get prettier studio tooltips now. I'll go for the latter.

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
SomeTroglodyte: I think that was meant as "...let's see if caching is worth it..." -> do implementation -> forget while testing. Now calls _are_ repeated, so I'd prefer to complete the lazy. Even if the thing is cheap. I tend to see this whole system as potentially performance-critical, the worst offenders are still glaringly untouched, so cheap ones get a +1.

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
SomeTroglodyte: .. are hopefully just fine. Never forget this project is meant to be `fun`, not ruin our organs image. (Who am I telling that?)

Maybe a leftover from trying to get that warning gone the peaceful way. Maybe an inline fun with early `return`'s?

SomeTroglodyte:
I think I've addressed everything and my local branch is a better candidate. However, merging to get that atlas combined is still necessary and _I still haven't learned how to do that correctly without duplicating [someone else's commits](https://en.wikipedia.org/wiki/Somebody_else%27s_problem)_. So don't be surprised if I close this one and resubmit as new - but I'll try.

SomeTroglodyte:
Hey wow this workflow seems to have been the right one... List of commits and file diff was way wrong in Studio but seems quite perfect here... :partying_face:  So yes this atlas now contains both "The Huns" under the new name and my link deco - hopefully - if I didn't manage my symlinked workdir wrong... :smile_cat:

SomeTroglodyte:
P.S.: Should you merge #4349 first then a to-do in this could be patched before merging.

yairm210:
And you have image conflicts 🙃

SomeTroglodyte:
> And you have image conflicts

Yes seems like every single commit you merge throws my 'link' icon out of the atlas so I have to rebuild it. Third atlas-only commit coming up shortly.

SomeTroglodyte:
Not every single one at all, but one like 'The Netherlands' will of course come with an atlas...
High time that nation appeared if you ask me, with all the excellent dutch contributors currently active.

SomeTroglodyte:
Almost forgot - that todo now resolved. BaseUnit.kt still has 1 compile warning and 8 inspection warnings - not this time.

Though the `The corresponding parameter in the supertype 'IConstruction' is named 'cityConstructions'. This may cause problems when calling this function with named arguments.` one - similar to the `IPreviousScreen` / `PickerScreen` collision - wouldn't even know how to solve that properly.

yairm210:
I'm accepting this, but I do think that some of the names could be a bit clearer (padIndent and padimage for example - each padding is on some specific side)

PR #4351, opened 2021-07-02
Add a very simple display for the religion in a city

xlenstra:
Later this will probably split off into its own class, but for now something as small as this will work

image
image

The double separator is due to there being one for specialists and another for religion.

If this city has specialist slots, it looks like this:

image


xlenstra:
Closed, replaced by #4356.

PR #4355, opened 2021-07-03
Fix multiple bugs resulting from updating promotions and implementing missile cruisers

xlenstra:
Fixes the following bugs:
- Aircraft in mods can't be on carriers
- Survivalism promotions had no prerequisites or image
- Scouting promotions prerequisites were wrong
- Logistics promotion in mods doesn't work
- Supply wasn't enabled
- (WIP) on higher starting era's, city states spawn with multiple settlers


yairm210:
Even as of now this is all good and mergable, obviously if there's extra work you can create another PR

PR #4356, opened 2021-07-03
Added a simplified version of great prophets, implemented a basic city religion UI

xlenstra:
This PR continues the implementation of religion - see #4290.
It implements a basic version of great prophets, which can be used to build holy sites or spread your pantheon to up to four of your own cities.
Holy sites are tile improvements which provide +6 faith.
It implements a basic UI for viewing the religion of a city.
This PR also replaces #4351, which previously implemented this feature - for images, see there.


SomeTroglodyte:
For a new unit action you need a template entry ***and*** at least one actual translation for the tests to pass.

xlenstra:
> For a new unit action you need a template entry _**and**_ at least one actual translation for the tests to pass.

Although I should know this by now, I keep forgetting it every time I add one, thanks for reminding :)

SomeTroglodyte:
> reminding

Nah, was only meant to save you some time since I had already navigated into the travis build logs :smiley_cat:

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -11,6 +12,30 @@ class ReligionManager {
...
+
+    fun clone(): ReligionManager {
+        val clone = ReligionManager()
+        clone.pantheonBelief = pantheonBelief
+        clone.storedFaith = storedFaith
+        return clone
+    }
+
+    fun startTurn() {
+        if (canGenerateProphet() && storedFaith > faithForNextGreatProphet()) {
```
yairm210, line 25: Surely the storedFaith > ... should go *inside* canGenerateProphet? Or call it 'shoudGenerateProphet'

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -11,6 +12,30 @@ class ReligionManager {
...
+            val prophetSpawnChange = (5f + storedFaith - faithForNextGreatProphet()) / 100f
+            if (Random.nextFloat() < prophetSpawnChange) {
+                civInfo.addUnit("Great Prophet")
+                storedFaith = 0 
+                // No, we don't just subtract the amount of faith it requires, you lose ALL of it. Why would we be nice?
```
yairm210, line 30: Is this REALLY how the base game works? That's disgusting. If so, I vote we change this.

xlenstra: Sadly, this is the case (at least according to the source I found online). I completely agree on changing it though

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -11,6 +12,30 @@ class ReligionManager {
...
+        val clone = ReligionManager()
+        clone.pantheonBelief = pantheonBelief
+        clone.storedFaith = storedFaith
+        return clone
+    }
+
+    fun startTurn() {
+        if (canGenerateProphet() && storedFaith > faithForNextGreatProphet()) {
+            val prophetSpawnChange = (5f + storedFaith - faithForNextGreatProphet()) / 100f
+            if (Random.nextFloat() < prophetSpawnChange) {
```
yairm210, line 27: Random.nextFloat should be seeded by the turn number, to minimize 'retry and hope RNGesus smiles'

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -105,6 +105,9 @@ class MapUnit {
     var promotions = UnitPromotions()
     var due: Boolean = true
     var isTransported: Boolean = false
+    
+    var religionSpreadsUsed: Int = 0
```
yairm210, line 109: Generalization., think towards the future. 'UsesLeft', or something of the sort.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -105,6 +105,9 @@ class MapUnit {
...
+    var religion: String? = null
```
yairm210, line 110: HMMM I wonder about this. Not sure if a sort of hashmap<string,string> of param to value is better. Because I feel we will have similar cases in the future.
Maybe we can start using templates IN THE UNIT NAME? Like "Great Prophet ([religionName])"

xlenstra: > Maybe we can start using templates IN THE UNIT NAME?
I have tried to get this working, and eventually gave up. The problem is that the name is used as a key in _many_ places, including the filename of the icon it should use, and there are bugs everywhere. For now I'll keep using this field, and change the displayname instead.

yairm210: I see

core/src/com/unciv/logic/civilization/ReligionManager.kt
logicminimal: > Is this REALLY how the base game works? That's disgusting. If so, I vote we change this.

so you say the game is open to change disliked things in civ 5 🧐

yairm210: Things that *I* dislike.

yairm210: That's the prerogative of the creator.

SomeTroglodyte: One could also argue that 'losing faith' isn't necessarily disgusting, it can be a _very healthy_ thing, depending on which cult you ascribe to... I'd vote for moddable, but - your call 😇

yairm210:
I'll be honest, there's a lot I don't like about this. But since we're only getting started with Religion, and getting the feel of things, I'm okay with commiting this now and we'll improve later.

PR #4357, opened 2021-07-03
ExpanderTab revisited

SomeTroglodyte:
#### Motivation:
`ExpanderTab` is underused, there's independent code duplicating the feature, and I have a new usecase. The existing felt too unclear - it's not obvious enough to the user this is a control and what it's good for. So I tried to 'spruce it up'.
This works and is an improvement over the current behaviour, but I'll only 'draft' it due to the open ***questions*** below. 

The following clip is done with stashed code not part of this PR, the visuals are just a bit clearer here than in a trade offer:

https://user-images.githubusercontent.com/63000004/124360455-32908000-dc2a-11eb-8b8f-411b8d83cf43.mp4

#### Questions:
- I have tried and failed to animate the 'content' part. Gdx just seems to ignore everything I tried - the best result ***was*** a gradual swipe-type showing and hiding, but it kept the total height reserved - ugly. Any idea how to clip _and_ limit the height of a larger widget? Typing this I get 'ScrollPane' - maybe I'll try later.
- With the bottom part not animating I've chosen to show/hide it at the _end_ of the arrow animation. Would beginning or middle be nicer?
- Applied to city constructions as in the video it shows a non-related problem there: While the container is nicely glued to the top left, its height is most of the available screen and the sub-contents will center within that if smaller... This, too, resists any .top() or .align(Align.top) instructions I threw at it. Hmm... maybe an empty row with expandY() at the bottom will do it? Anyway - would the shown reaction to all construction categories minimized be a showstopper?
- Another non-related thing: "Queue empty" is in the code but never shown, as queue.size is always at least 1. The video shows if(size>1) - but I'd actually prefer it gone _along_ with the header if the current construction is the only one... Your preference?

#### Plans
- Move to ui.utils once more than 1 use
- The `CityInfoTable` can be converted easily to use the ExpanderTab
- `CityConstructionsTable` might profit - per se not too useful, but if we persist open/closed state across cities then it could make thinks like adding a bank to all cities _much_ easier.
- I would use it to make NewGameScreen portrait-compatible.

### DIFF
--- a/core/src/com/unciv/ui/trade/ExpanderTab.kt
+++ b/core/src/com/unciv/ui/trade/ExpanderTab.kt
@@ -1,45 +1,100 @@
 package com.unciv.ui.trade
 
+import com.badlogic.gdx.graphics.Color
+import com.badlogic.gdx.math.Interpolation
 import com.badlogic.gdx.scenes.scene2d.Touchable
-import com.badlogic.gdx.scenes.scene2d.ui.Skin
+import com.badlogic.gdx.scenes.scene2d.actions.FloatAction
 import com.badlogic.gdx.scenes.scene2d.ui.Table
+import com.badlogic.gdx.utils.Align
+import com.unciv.UncivGame
+import com.unciv.ui.utils.CameraStageBaseScreen
 import com.unciv.ui.utils.ImageGetter
 import com.unciv.ui.utils.onClick
 import com.unciv.ui.utils.toLabel
 
-class ExpanderTab(private val title:String,skin: Skin): Table(skin){
-    private val toggle = Table(skin) // the show/hide toggler
-    private val tab = Table() // what holds the information to be shown/hidden
-    val innerTable= Table() // the information itself
-    var isOpen=true
+/**
+ * A widget with a header that when clicked shows/hides a sub-Table
+ * 
+ * Add content to be shown/hidden to [innerTable]!
+ * 
+ * @param title The header text, automatically translated
+ * @param initialOpen Should it start out opened? Defaults to `true`.
+ */
+class ExpanderTab(
+    title: String,
+    initialOpen: Boolean = true,
+    defaultPad: Float = 10f,
+    initContent: ((Table) -> Unit)? = null
+): Table(CameraStageBaseScreen.skin) {
+    private companion object {
+        const val fontSize = 24
+        const val arrowSize = 18f
+        const val arrowImage = "OtherIcons/BackArrow"
+        val arrowColor = Color(1f,0.96f,0.75f,1f)
+        const val animationDuration = 0.2f
+    }
 
-    init{
-        toggle.defaults().pad(10f)
-        toggle.touchable= Touchable.enabled
-        toggle.background(ImageGetter.getBackground(ImageGetter.getBlue()))
-        toggle.add("+ $title".toLabel(fontSize = 24))
-        toggle.onClick {
-            if(isOpen) close()
-            else open()
+    private val header = Table(skin)  // Header with label and icon, touchable to show/hide
+    private val headerLabel = title.toLabel(fontSize = fontSize)
+    private val headerIcon = ImageGetter.getImage(arrowImage)
+    private val contentWrapper = Table()  // Wrapper for innerTable, this is what will be shown/hidden
+
+    /** The container where the client should add the content to toggle */
+    val innerTable = Table()
+
+    /** Indicates whether the contents are currently shown */
+    var isOpen = initialOpen
+        private set(value) {
+            if (value == field) return
+            field = value
+            update()
         }
-        add(toggle).fill().row()
-        tab.add(innerTable).pad(10f)
-        add(tab)
+
+    init {
+        header.defaults().pad(10f)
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/trade/ExpanderTab.kt
```diff
@@ -1,45 +1,100 @@
...
+ * A widget with a header that when clicked shows/hides a sub-Table
+ * 
+ * Add content to be shown/hidden to [innerTable]!
+ * 
+ * @param title The header text, automatically translated
+ * @param initialOpen Should it start out opened? Defaults to `true`.
+ */
+class ExpanderTab(
+    title: String,
+    initialOpen: Boolean = true,
```
yairm210, line 25: maybe rename this to "startsOutOpened"

core/src/com/unciv/ui/trade/ExpanderTab.kt
```diff
@@ -1,45 +1,100 @@
...
-        toggle.onClick {
-            if(isOpen) close()
-            else open()
+/**
+ * A widget with a header that when clicked shows/hides a sub-Table
+ * 
+ * Add content to be shown/hidden to [innerTable]!
+ * 
+ * @param title The header text, automatically translated
+ * @param initialOpen Should it start out opened? Defaults to `true`.
```
yairm210, line 21: I'm a believer in 'if a param needs an explanation, it either has hidden functionality or it's badly named'

core/src/com/unciv/ui/trade/ExpanderTab.kt
```diff
@@ -1,45 +1,100 @@
...
-        isOpen=true
+    fun close() {
+        isOpen = false
+    }
+    fun open() {
+        isOpen = true
+    }
+    fun toggle() {
+        isOpen = !isOpen
     }
```
yairm210, line 99: Don't pre-create unused functions. If we need them, we can make them later.

core/src/com/unciv/ui/trade/ExpanderTab.kt
SomeTroglodyte: My guideline is 'do I find the tooltip useful' (and don't invest more than a few seconds for docs):
image

Methods often need less because Studio shows the signature anyway, so all parameter names and types are already there. Classes lack that part, so I try to mention them all - and in this case I failed, forgot the two I added later.

core/src/com/unciv/ui/trade/ExpanderTab.kt
SomeTroglodyte: Happily - but two were already there before I started and `toggle` is used. Will patch later.

core/src/com/unciv/ui/trade/ExpanderTab.kt
SomeTroglodyte: Not bad, maybe apply to String.toCheckBox as well.

core/src/com/unciv/ui/trade/ExpanderTab.kt
```diff
@@ -1,45 +1,100 @@
...
+        val action = object: FloatAction ( 90f, 180f, animationDuration, Interpolation.linear) {
+            override fun update(percent: Float) {
+                super.update(percent)
+                headerIcon.rotation = this.value
+                if (this.isComplete) {
+                    contentWrapper.clear()
+                    if (isOpen) contentWrapper.add(innerTable)
+                }
+            }
+        }.apply { isReverse = isOpen }
```
yairm210, line 87: This is clever, I didn't even know this existed!

SomeTroglodyte: To learn how to use any Gdx object I always look into the sources. Helps. Ctrl-B, in this case twice to get into TemporalAction.java, two pages to skim, half of that quickly skipped boilerplate, and there it is:
```java
	/** When true, the action's progress will go from 100% to 0%. */
	public void setReverse (boolean reverse) { this.reverse = reverse; }
```

SomeTroglodyte: Oh, by the way, that saved you from a too-clever-by-half:
```kotlin
        val (start, end) = if (isOpen) 180f to 90f else 90f to 180f
```

SomeTroglodyte:
OK, I learned something: Studio tooltips for a class are different whether you look from the declaration or from an instantiation. I always checked the former, but the latter has more info (namely a constructor listing with full signatures), so I was indeed way too verbose. The last commit gives this tooltip which I would deem just helpful enough:
image
It still lacks the info what other public properties/methods exist. `@see` could do that without supporting explanations - don't like it much, and `@property` is something else entirely.

SomeTroglodyte:
I'd still like input on those questions, but I wouldn't mind this merged, so - 'Draft' removed.

yairm210:
1. No idea how to clip and limit, it's all guesswork I'm afraid
2. Show/hide at the end is good
3. ???
4. Gone with the header is good

SomeTroglodyte:
Thanks. #3 was just an alignment thing I'll look into more closely in one of the next steps.

PR #4360, opened 2021-07-03
Fixed more bugs

xlenstra:
This PR fixes two bugs:
- In eras after the classical era, city-states will only spawn with a single setter-like unit. Previously, this could be any number
- The icon for the huns was renamed so it functions properly


core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -176,10 +176,18 @@ object GameStarter {
...
+        var startingUnits: MutableList<String>
+        var eraUnitReplacement: String
+        
         val startingLocations = getStartingLocations(
                 gameInfo.civilizations.filter { !it.isBarbarian() },
                 gameInfo.tileMap)
-        
+
+        val settlerLikeUnits = ruleSet.units.filter {
+            it.value.uniqueObjects.any { it.placeholderText == Constants.settlerUnique }
```
yairm210, line 189: ruleSet.units.values.filter{ it.hasUnique (Constants.settlerUnique) } perhaps? Don't recall if hasUnique is also on baseunit or only on mapunit

xlenstra: baseUnit has no hasUnique or getMatchingUniques, sadly. I can add them if you'd like though

yairm210: No need

PR #4367, opened 2021-07-04
Fixed bug where heal instantly would not heal

### LINKED ISSUE #4366: Instant health doesn't work
version is 3.15.7, android

it either doesn't heal it by 50 hp or doesnt heal at all, i couldnt understand

xlenstra:
Fixes #4366


logicminimal:
so this was caused by just one upper/lowercase mismatch? why doesn't the code just turn all key strings into lowercase then check according to that?

yairm210:
Because things need to mean things to be accurate.
This is also part of the player experience, not just a computer perception issue.

PR #4368, opened 2021-07-04
Fixed more bugs I introduced myself

xlenstra:
Fixed bug where mobility promotion doesn't work
Fixed bug where supply promotion wouldn't work


ravignir:
There is a problem with the huns, icon is not displaying, and the battering ram is unique to "Huns" instead of "The Huns".

xlenstra:
The icon should be fixed already (both by me and someTroglodyte in different PR's if I remember correctly), I'll look into the battering ram problem

PR #4370, opened 2021-07-04
More bug fixes

xlenstra:
At some point there should be no more bugs right?
- Forgot to add an icon for Air Targeting
- Forgot to rename the 'huns' to 'the huns' for battering ram uniqueTo
- Plundering with multiple levels of coastal raider now only gives a single notification for each stat


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -186,19 +187,27 @@ object Battle {
...
-                            "Your [${plunderingUnit.getName()}] plundered [${resourcesPlundered}] [${unique.params[2]}] from [${plunderedUnit.getName()}]",
+                        unique.params[0].toFloat() / 100f * damageDealt
+                    plunderedGoods.add(Stat.valueOf(unique.params[2]), resourcesPlundered)
+                }
+            }
+        }
+        val plunderableStats = listOf("Gold", "Science", "Culture", "Faith").map { Stat.valueOf(it) }
+        for (stat in plunderableStats) {
+            val resourcesPlundered = plunderedGoods.get(stat)
+            if (resourcesPlundered != 0f) {
```
yairm210, line 203: Reverse if for lower indentation. If(==0) continue

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -186,19 +187,27 @@ object Battle {
     }
 
     private fun plunderFromDamage(plunderingUnit: ICombatant, plunderedUnit: ICombatant, damageDealt: Int) {
+        val plunderedGoods = Stats()
         if (plunderingUnit is MapUnitCombatant) {
             for (unique in plunderingUnit.unit.getMatchingUniques("Earn []% of the damage done to [] units as []")) {
                 if (plunderedUnit.matchesCategory(unique.params[1])) {
```
yairm210, line 193: Here too now that I mention it

yairm210:
> At some point there should be no more bugs

If only you knew how bad things really are.

PR #4372, opened 2021-07-04
Added the nation of the Netherlands

xlenstra:
This PR adds the nation of the Netherlands.
This includes:
- The Sea Beggar (unique unit)
- Polder (unique improvement, can be built on marsh & flood plain)
- National ability: the following unique:

Added uniques:
- "Retain [amount]% of the happiness from a luxury after the last copy has been traded away"


core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -169,10 +172,17 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
         
         val luxuriesProvidedByCityStates = 
             civInfo.getKnownCivs().filter { it.isCityState() && it.getAllyCiv() == civInfo.civName }
-                .map { it.getCivResources().map { res -> res.resource } }.flatten().count { it.resourceType === ResourceType.Luxury }
+                .map { it.getCivResources().map { res -> res.resource } }.flatten().distinct().count { it.resourceType === ResourceType.Luxury }
```
yairm210, line 175: .map.flatten should probably be replaced by .flatmap

yairm210:
Seems pretty straightforward. You do have conflicts due to the image files though.

xlenstra:
I'll fix all conflicts in a few minutes, first I'm opening yet another PR :)

xlenstra:
Oh, also another point: if I'm not mistaken the changelog for the last version did not include the addition of pantheons or the huns, even though they were added in this version.

yairm210:
I often limit the changlelog to stay within the 500-character limit of the Google Play 'what's new'
And it's easier to have a single changelog everywhere than have to deal with different ones in different places...

PR #4377, opened 2021-07-04
Updated template.properties

xlenstra:
This adds a few translations that were previously missed (most notably updates unitFilter, tileFitler and cityFilter to contain all the values currently allowed), and shuffles the rest of the file around so that strings about similar things are closer together. Also removed a few strings that should really be included in the automatic generation (mass media, +1 population in each city).

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -25,7 +25,6 @@ Choose a free great person =
 Get [unitName] = 
 
 Hydro Plant = 
-+1 population in each city = 
 [buildingName] obsoleted = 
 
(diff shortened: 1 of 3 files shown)


android/assets/jsons/translations/template.properties
```diff
@@ -245,10 +244,7 @@ Max Coast extension =
 Biome areas extension = 
 Water level = 
 Reset to default = 
-
-Show experimental world wrap for maps = 
-HIGHLY EXPERIMENTAL - YOU HAVE BEEN WARNED! =
```
yairm210, line 250: Keep this, it's always useful

PR #4381, opened 2021-07-05
Fixed bug where city-states would not share their science income even if the player had the right policy

xlenstra:


core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -105,15 +105,13 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
...
             if (otherCiv.isCityState())
                 for (unique in civInfo.getMatchingUniques("Allied City-States provide [] equal to []% of what they produce for themselves")) {
-                    if (otherCiv.diplomacy[civInfo.civName]!!.matchesCityStateRelationshipFilter(unique.params[0]) && otherCiv.cities.isNotEmpty()) {
-                        statMap.add(
-                            "City-States", 
-                            Stats().add(
-                                Stat.valueOf(unique.params[1]), 
-                                otherCiv.statsForNextTurn.get(Stat.valueOf(unique.params[1])) * unique.params[2].toFloat() / 100f
-                            )
+                    statMap.add(
```
yairm210, line 108: It doesn't seem we're filtering for *allied* city states

logicminimal:
so the science boost i thought i was getting was just a placebo

PR #4391, opened 2021-07-05
Spruced up Civilopedia - phase 5 - buildings

SomeTroglodyte:
... and the _other_ buildings, too. Builds on #4350.

#### Marginal features:
- The Civilopedia version of the description is more mod friendly (nicer when replaces==null or cost==0) and allows linking from a building to its nation-unique variants so you can go back and forth.
- The "old" description is trimmed from its now obsolete parts and the parameter toggling them.
- This time no sample "civilopedia text" added - I'll leave those for other contributors. Start PR for that somewhere? (not pull request, public relations)
- Religion buildings need Religion Option to be displayed in Civilopedia, and Utopia needs its Victoria type.
    - Noticed later that similarly gating spaceship parts might be in order - or we decide not to suppress such entries at all?
- Found a few missing templates, translatability fully tested with included german, @Mape6 - also some improvements, I hope, feel free to revert next in iteration

#### Notes:
- This shows that where ever things were listed comma separated in-line which I want to link, I need to format separate lines now, and that will most of the time provoke new translatables. I still treated {} and [] as 100% mutually exclusive, which they aren't anymore, but here I don't see opportunities to simplify using that. Better eyes may see some.
- I will not tolerate templates that _depend_ on leading or trailing spaces in new code by myself, and I think somewhere in the still shelved portions are some such cases - not here. I hope.
- It's plain that I began when there were far less parameterized uniques - with one exception (hardcoded in here) those will still come out undecorated and un-linked. A generalized function that simply scans all parameters for any of the ruleset object names is feasible, though potentially a performance killer. Postponed.

Example - from a mod to show it no longer displays the word "null":
image
The uneven line distances are intentional and meant to group similar things together. These take a few extra code lines and maybe we won't like them that much in the end - can be tuned later.


android/assets/jsons/translations/template.properties
```diff
@@ -33,6 +37,9 @@ Requires [buildingName] to be built in the city =
 Requires [buildingName] to be built in all cities = 
 Provides a free [buildingName] in the city = 
 Requires worked [resource] near city = 
+Requires worked [resource] = 
+or [resource] = 
+or [resource] near city =
```
yairm210, line 42: When does this appear?

android/assets/jsons/translations/template.properties
```diff
@@ -597,6 +605,7 @@ Maintenance cost =
 Pick construction = 
 Pick improvement = 
 Provides [resource] = 
+Provides [amount] [resource] =
```
yairm210, line 608: I'm pretty sure this is already translated, no? It should be an atom bomb requirement

android/assets/jsons/translations/template.properties
```diff
@@ -1034,6 +1045,7 @@ Must be on [terrain] =
 Can only be built on [tileFilter] tiles = 
 Cannot be built on [tileFilter] tiles = 
 Does not need removal of [feature] = 
+Gain a free [building] [cityFilter] =
```
yairm210, line 1048: This too already exists.

android/assets/jsons/translations/template.properties
yairm210: Oh no that was the 'requires'
nvm

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -200,6 +197,147 @@ class Building : NamedStats(), IConstruction {
...
+            else
+                for (unique in getUniquesStrings()) textList += FormattedLine(unique)
+        }
+
+        if (!stats.isEmpty()) {
+            textList += FormattedLine(stats.toString())
+        }
+
+        if (!percentStats.isEmpty()) {
+            if (percentStats.production != 0f)
```
yairm210, line 274: Just percentStats.toHashMap().forEach((key, value)) this.

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -200,6 +197,147 @@ class Building : NamedStats(), IConstruction {
...
+                textList += FormattedLine(percentStats.science.formatSignedInt() + "% {Science}")
+            if (percentStats.food != 0f)
+                textList += FormattedLine(percentStats.food.formatSignedInt() + "% {Food}")
+            if (percentStats.culture != 0f)
+                textList += FormattedLine(percentStats.culture.formatSignedInt() + "% {Culture}")
+        }
+
+        if (this.greatPersonPoints != null) {
+            val gpp = this.greatPersonPoints!!
+            if (gpp.production != 0f) textList +=
```
yairm210, line 288: Here too, you can use the 'stat to great person' mapping although I hate it

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -200,6 +197,147 @@ class Building : NamedStats(), IConstruction {
...
+        if (specialists.isNotEmpty()) {
+            for ((specialistName, amount) in specialists)
+                textList += FormattedLine("+$amount " + "[$specialistName] slots".tr())
+        }
+
+        if (requiredNearbyImprovedResources != null) {
+            textList += FormattedLine()
+            requiredNearbyImprovedResources!!.withIndex().forEach {
+                textList += FormattedLine(
+                        (if (it.index == 0) "Requires worked" else "or") +
```
yairm210, line 311: U wot m8
There's no way this will be translated well.

android/assets/jsons/translations/template.properties
SomeTroglodyte: Well, let's see - the one building tied to three resources... Monastery? No. Mint? No. Stable!
image
old: 'csv' style new: separate lines to allow individual links. That way to tie them together with words might not be the best idea, but - well, I _tried_ to focus more on 'works', maybe even 'reliably', less on perfect looks. Giving each line the same original template would have been easier, sure.

android/assets/jsons/translations/template.properties
SomeTroglodyte: In code yes, but try template.properties Ctrl-F "Gain a free" -> not found. That was my reason to include, not because I introduced the text.

core/src/com/unciv/models/ruleset/Building.kt
SomeTroglodyte: Both are more or less copies of the 'old' getDescription implementation - had I optimized everything, well, the diff engine might not have found anything left to sync on...

core/src/com/unciv/models/ruleset/Building.kt
SomeTroglodyte: image
The 'von' is @Mape6 's work methinks, and the only point this sounds ugly to us krauts. "Benötigt bewirtschaftete ressource [x]" would read nicer - so, yes, these are a challenge because they limit the structural redistribution liberties of a translator severely. But a) I had no better idea at the moment, b) I prioritized nice english a little and c) Unciv in german is horrible anyway because you can't provide for grammar, noun gender, sometimes plural/singular is not fixed, etc. It's a compromise anyway.

Any better idea welcome, but pushes will have to wait a little - I need to drive a few thousand miles after work.

SomeTroglodyte:
By the way, next phase is not the next category but uniques. Works fast & takes less than a MB ram to cache every ruleset name and map it to the most likely category. Ignores those 'logic and/or within filter' ones, though. That first screenshot could have every single line nicely linked - if the mod weren't faulty.

android/assets/jsons/translations/template.properties
yairm210: This is EXTREMELY language-specific. I highly doubt that all languages can fit this in.
Instead, I suggest the following. A non-link line, followed by the resource names and links.

- Requires a worked resource near the city:
- Horses
- Sheep
- Cattle

SomeTroglodyte: I agree (Had more or less the same idea over the long weekend). That it's an 'or' not an 'and' is clear enough in english from the "a" worked resource, but that may not be immediately clear to translators - comments in template.properties do propagate keeping their place?

logicminimal:
by the way there is a huge space in "replaces" part in unit description

SomeTroglodyte:
> by the way there is a huge space in "replaces" part in unit description

So what? That is one intentional indent level, because it's a continuation of the line before.

core/src/com/unciv/models/ruleset/Building.kt
SomeTroglodyte: So happiness and faith were not omitted intentionally...

core/src/com/unciv/models/ruleset/Building.kt
SomeTroglodyte: I just couldn't use it as is. (#civs + #overview uses + #city screens opened) instances in a run when 1 is enough? See next commit...

core/src/com/unciv/models/ruleset/Building.kt
SomeTroglodyte: Also, toHashMap is just as ugly a kludge. I counted 55k instantiations over 10 turns in a very average game - a large majority of these are quickly abandoned to the GC, and IMHO an implementation without any of these is possible.

SomeTroglodyte:
Could someone please show me a screenshot of where actual conflict resolution controls are in the github conflict resolution screen? I see the conflict marked (in this case a silly one, why should adding some lines be a conflict but other added lines aren't?), I see previous/next buttons, a control for indent mode/size and line wrap, and an inactive "mark as resolved", that's it.

android/assets/jsons/translations/template.properties
```diff
@@ -33,6 +37,8 @@ Requires [buildingName] to be built in the city =
 Requires [buildingName] to be built in all cities = 
 Provides a free [buildingName] in the city = 
 Requires worked [resource] near city = 
+# Translator note: Meaning "Requires at least one of the following resources worked near the city":
```
yairm210, line 40: Well then, just replace this with Requires at least one of the following resources worked near the city
It may be longer, but explicit is better, o?

android/assets/jsons/translations/German.properties
```diff
@@ -647,6 +650,7 @@ Pick [technology] as free tech = [technology] als kostenlose Technologie auswäh
 Units enabled = Freigeschaltete Einheiten
 Buildings enabled = Freigeschaltete Bauwerke
 Wonder = Wunder
+Wonder of the World = Weltwunder
```
yairm210, line 653: Not needed - this is just "Wonder".

android/assets/jsons/translations/template.properties
```diff
@@ -644,6 +652,7 @@ Pick [technology] as free tech =
 Units enabled = 
 Buildings enabled = 
 Wonder = 
+Wonder of the World =
```
yairm210, line 655: I'm serious about this. this should be removed, the game already has a clear "Wonder / National Wonder" phrasing, we shouldn't be adding new indicators to things that already exist.

android/assets/jsons/translations/template.properties
yairm210: This too.

android/assets/jsons/translations/template.properties
SomeTroglodyte: OK - was just an attempt to make the entries look cooler:
image

Not that anybody would have noticed...

SomeTroglodyte:
Sorry it took me a while to react. I hope I didn't miss one of your suggestions...

SomeTroglodyte:
Some of my other PR's may depend on a changed function signature in here. Let me rebase them, should save you time.

SomeTroglodyte:
Nah, I was mistaken. I didn't remove the unused parameter in `getShortDescription` and `getDescription` after all, I thought I did. All my other PR's are merged forward or unaffected.

PR #4393, opened 2021-07-06
Refactored the way cities determine what uniques should apply when

xlenstra:
Previously, whenever we needed to check what uniques were provided by a city, we could simply query `cityConstructions.builtBuildingUniqueMap.getUniques()` or `cityConstructions.builtBuildingMap.getAllUniques()`.
However, due to the implementation of religion, this was no longer the only source of uniques in cities.
Adding calls to `religion.getAllUniques()` everywhere was space consuming, unmodular and a source of potential bugs if you forgot one.
This PR fixes this problem by creating access functions on `CityInfo` for all uniques. To be specific, it creates the following functions:
- `getLocalMatchingUniques`: this function returns uniques from this city (buildings or religion) with the matching unique
- `getAllLocalUniques`: returns all uniques originating from this city
- `getMatchingUniquesWithNonLocalEffects`: returns all matching uniques that don't have 'in this city' as a unique parameter. This is a step up from the previous design, in which only the last parameter was checked, but still might lead to bugs in the future. My testing was unable to find any for now.
- `getAllUniquesWithNonLocalEffects`: Same as above, but returns all uniques instead of only the matching ones.
- `getMatchingUniques`: This queries `getAllLocalUniques` AND `civilizationInfo.getMatchingUniques`, so that also uniques from outside of city might influence it. For it not to doublecount uniques also provided by this city, a paramter `cityToIgnore` was added to `civilizationInfo.getMatchingUniques`.

By calling either `getMatchingUniques` or `getLocalMatchingUniques` the correct uniques can be found in all cases in the current codebase.

It also updates a lot of uniques to add the 'in this city' modifier to them, so they no longer apply to other cities under this new system:
- "[stats] from [tileFilter] tiles in this city" -> "[stats] from [tileFilter] tiles [cityFilter]"
- "+[amount]% great person generation in this city" -> "[amount]% great person generation [cityFilter]"
- "+[amount]% great person generation in this city" -> see above
- "All newly-trained [unitFilter] units in this city receive the [promotionName] promotion" -> "All newly-trained [unitFilter] units [cityFilter] receive the [promotionName] promotion"
- "New [unitFilter] units start with [amount] Experience in this city" -> "New [unitFilter] units start with [amount] Experience [cityFilter]"

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -26,9 +26,9 @@
 		"food": 2,
 		"maintenance": 1,
 		"hurryCostModifier": 25,
-		"uniques": ["[+1 Food] from [Deer] tiles in this city",
-			"[+1 Food] from [Bananas] tiles in this city",
-			"[+1 Food] from [Wheat] tiles in this city"],
+		"uniques": ["[+1 Food] from [Deer] tiles [in this city]",
+			"[+1 Food] from [Bananas] tiles [in this city]",
+			"[+1 Food] from [Wheat] tiles [in this city]"],
 		"requiredTech": "Pottery"
 	},
 	{
@@ -57,8 +57,8 @@
 		"maintenance": 1,
 		"hurryCostModifier": 0,
 		"uniques": ["Must not be on [Plains]",
-			"[+1 Production] from [Stone] tiles in this city",
-			"[+1 Production] from [Marble] tiles in this city"],
+			"[+1 Production] from [Stone] tiles [in this city]",
+			"[+1 Production] from [Marble] tiles [in this city]"],
 		"requiredTech": "Calendar"
 	},
 	{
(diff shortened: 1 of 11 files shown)


yairm210:
Despite the length, this is remarkably clean!
I will be putting this off until the next version, but that's just so we can space out the amount of changes over time

SomeTroglodyte:
:+1: couldn't resist checking some of it. I even found it way down the branch manager - luckily my screen is 32".

Random thoughts in no particular order:
- builtBuildingUniqueMap could it now be package-private to protect a little from new code bypassing your new API? All current refrences _are_ in `com.unciv.logic.city`? Is that even worth the hassle?
- Five functions for structurally similar usecases? My API would have one and a 'scope' parameter, possibly a bitmask enum - if they even exist in kotlin, as the java approach was such a strange afterthought... But when I look at the function bodies - OK maybe not.
- Checkin .asSequence calls... Maybe the whole thing could look cleaner making `UniqueMap.getUniques(placeholderText: String)` return a Sequence, too. As it is `getLocalMatchingUniques` calls `+` for different types forcing the more generic `Collection.plus(Sequence)` - which luckily exists, I'd have expected it to be a `Collection.plus(Collection)` - and thus precluding a potentially more optimized `Sequence.plus(Sequence)`  ...  ... and then you'd need an `asSequence` less in `getLocalMatchingUniques`and another in `getMatchingUniquesWithNonLocalEffects`. You know Studio can show you which overload of an operator it infers? Select the character and wait extra long for the tooltip.  ... ... Yes, compiles and runs with UniqueMap: `fun getUniques(placeholderText: String) = this[placeholderText]?.asSequence() ?: sequenceOf()` and those two asSequence removed, though I didn't test too far in.

- +3 for avoiding new 'it shadows it'
- +5 for +[amount] -> [amount]
- +13 for "I'm so sorry for the size of this PR"

Issue #4392, opened 2021-07-06
Weird Bug I encountered


NoviceCoder2000:
For some weird reason, whenever I launch a mod that I made, the game won't launch (PC Version). But works perfectly fine on the mobile app, it's only a specific mod, other mods works perfectly fine in desktop. Because of it, I can't develop a new atlas and game.png files for my mod.

**To Reproduce**
- You just attach my mod in the mod files and just launch the desktop app. I'll edit it and try to attach the files here, probably in a reply

**Expected behavior**
The game won't simply launch

**Additional context**
I'm using a windows 7 os.

NoviceCoder2000:
Repository of the mod that I'm currently working on.

https://github.com/NoviceCoder2000/Wars-Mod

SomeTroglodyte:
Strange. Browser opens that link, Unciv says 404 Not found...
Alternate query to see what a normal listing with the topic would return (third one): https://api.github.com/search/repositories?q=Unciv%20Wars%20World
Default branch is not 'master' but 'main' which the "direct url input" button can't handle. We could maybe test for a given URL including the '.zip' extension and if so take it as is and not guess the branch.

So that's a marginal issue with downloading that from Unciv directly, now understood. Next test...

SomeTroglodyte:
Next marginal issue now understood: Exceptions from checkModLinks in ModCheckBoxTable are 'flattened' to that "The mod you selected is incompatible" toast, even if it contains useful information.

In your mod's case that is **"Building (Black Cannon) must either have an explicit cost or reference an existing tech"**

However, at that point I have already a valid atlas generated. Those LeaderIcons - too large for that much unsharpness ;)

Still open: Why does Options-Check Mods not show that error?

SomeTroglodyte:
Conclusion: I can _not_ reproduce "The game won't simply launch" with the current codebase, but thank you por pointing us to those other details concerning mod download and error display.

I'll assume your jar is way outdated. If you're _sure_ it's up-to-date please provide console output. Launch the jar directly from a cmd window using `java -jar Unciv.jar`.

NoviceCoder2000:
Yes, I think the jar is outdated considering I can't update my java.

considering inputting the command in the prompt only results to "unable to access jarfile"

NoviceCoder2000:
Thanks for the help @SomeTroglodyte but after adding an atlas and jpeg file the game worked for me once again. I'm still unsure over how and why that happened, but I think this issue is resolved now.

PR #4396, opened 2021-07-06
Better mod download and error display

SomeTroglodyte:
inspired by #4392

#### Features
- Mod download attempts from well-formed url but resulting in a 404 will now display the "couldn't" message
- Allows translation of "incompatible mod" message (the one on new game screen), but not all individual modcheck errors
- Modchecker throwing exception will now display (on new game screen)
- Manual mod URL entry allows providing not a repo but a zip link - which allows default branches not called master - with prompt

#### Please:
- Check wording of "enter A or B" label. Open to suggestions. Mention fact that default branch <> "master" _needs_ the zip variant? Screen space?
- Check Regexes - you understand those stoneage-android-java issues, I don't

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1000,7 +1000,8 @@ Download [modName] =
 Update [modName] = 
 Could not download mod list = 
 Download mod from URL = 
-Download = 
+Please enter the mod repository -or- archive zip url: = 
+Download =
 Done! = 
 Delete [modName] = 
 Are you SURE you want to delete this mod? = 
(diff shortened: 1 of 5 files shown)


SomeTroglodyte:
D* Android Studio so-called "intelligence" ruining lines I didn't even touch... Never use it when Notepad++ does the job, too! Will need an hour or three to start a fix, though.

android/assets/jsons/translations/template.properties
```diff
@@ -1019,7 +1020,8 @@ No description provided =
 [stargazers]✯ = 
 Author: [author] = 
 Size: [size] kB = 
- 
+The mod you selected is incompatible with the defined ruleset! =
```
yairm210, line 1023: I'm sure this already exists

SomeTroglodyte: So was I. Ctrl-F, F3...

PR #4401, opened 2021-07-06
ExpanderTab revisited - move and apply to city screen

SomeTroglodyte:
Sequel to #4357 

- ExpanderTab moved to package utils
- Added features to support usecases from city screen (building icon, width)
- A little extra effort to guarantee the construction type expanders have the same width
- The buildings under CityInfoTable actually look different because I did not add even more modifiers to ExpanderTab (remove background and animated arrow), otherwise they _would_ look almost identical to before. I think it's a pleasant change:
image
(left = half way, only the categories rewritten, right = this PR)

### Question:
- I didn't move the separator under an 'opened' building (see screenie above), but I think it's ~~wrong~~ suboptimal. Shouldn't be between a building and its specialists. We decide and change that later, agreed? Any input on desired outcome? A case for an inconspicuous (gray/1f) separator, permanent instead of toggled with the info?
- I'd like to persist startsOutOpened at least between screen invocations -> OffersListScroll expanders are almost pointless because each added offer makes the screen forget every expander state, sequentially going through all cities to add a researched building could profit....
Architecture? One non-serialized subproperty of `GameSettings` - `expanderStates` - each usecase a HashSet<String> under that, listing the closed ones by title?

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
@@ -18,6 +18,7 @@ import com.unciv.models.stats.Stat
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.*
 import kotlin.concurrent.thread
+import kotlin.math.max
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
 class CityConstructionsTable(val cityScreen: CityScreen) : Table(CameraStageBaseScreen.skin) {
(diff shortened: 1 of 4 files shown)


SomeTroglodyte:
P.S.: Includes the fix for "Queue empty is never shown", as discussed earlier, in the 'remove including header' variant. _I left the template in place to keep existing translations for possible later reuse_ - open to discussion - but the string is 100% unused after this PR.

yairm210:
A. You can remove the separator
B. I think that a hashset could work well, but not in the global settings, you could easily get conflicts. One per screen, methinks.

PR #4402, opened 2021-07-06
Updated TranslationFileWriter to include the new values that filters can have

xlenstra:
In the past time, many new options for filters were added. These were, however, not added to the TranslationFileWriter.
This resulted in multiple occurrences of `param` being written, while the type of the parameter was decidable.
This PR fixes this, by adding all those new filter options to the TranslationFileWriter.
It also fixes piety policies that affected temple's to affect amphitheaters instead, as temples currently don't exist.


core/src/com/unciv/models/translations/TranslationFileWriter.kt
```diff
@@ -197,21 +197,52 @@ object TranslationFileWriter {
...
+                            UnitType.values().any { it.name == parameter } 
+                                    || parameter == "Military" 
+                                    || parameter == "Civilian"
+                                    || parameter == "non-air"
+                                    || parameter == "relevant"
+                                    || parameter == "Nuclear Weapon"
+                                    || parameter == "Submarine"
+                                // These are up for debate
+                                    || parameter == "Land" // This one especially, is it can also apply to improvements
+                                    || parameter == "Water"
```
yairm210, line 231: Yeah, both of these are both unit and tile filters, so no

yairm210:
This needs to be done very carefully. Changing the existing params to filters can break the existing translations.
I'll merge all translations and deal with this now, I agree this is necessary but that doesn't mean it's simple

yairm210:
image
Gotta love regex replace

PR #4404, opened 2021-07-06
Update Spanish Properties - ¡read my description!

ArchDuque-Pancake:
In the tutorials part, I am not sure if all those emojis are different or not, I can not tell. So please, check that and if you can correct it if you can.

### DIFF
--- a/android/assets/jsons/translations/Spanish.properties
+++ b/android/assets/jsons/translations/Spanish.properties
@@ -445,10 +445,8 @@ A(n) [nukeType] exploded in our territory! = ¡¡Un(as) [nukeType] detonaron en 
 After being hit by our [nukeType], [civName] has declared war on us!
 The civilization of [civName] has been destroyed! = ¡La civilización de [civName] ha sido destruida!
 The City-State of [name] has been destroyed! = ¡La ciudad-estado [name] ha sido destruida!
- # Requires translation!
-Your [ourUnit] captured an enemy [theirUnit]! = 
- # Requires translation!
-Your [ourUnit] plundered [amount] [Stat] from [theirUnit] = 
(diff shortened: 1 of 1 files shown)


Kensvin28:
Some emojis are still incorrect, but I can't edit it, sorry

SomeTroglodyte:
The commits of *your* PR 'live' in *your* repository, so only you can update them. But this is a respectable amount of work, :+1:

The "bad news":
- Using those circled number/letter unicode characters is my fault. Por favor disculpe. I'm less happy now with them than when I did that, mainly due to the inexplicable choice of many fonts to give them so widely varying sizes. In the character map tool I used they were all pretty much the same circle size.
- I updated four of those in an outstanding PR - two typos, two omissions. So there'll be some conflict, and I hope I'll have the opportunity to resolve that myself and preserve your work. But maybe not - keep a copy in case those reappear als 'needs translation'.
- Seems like you tried to re-enter them instead of copy/pasting the character from the template. I see several (b) where it should be (b) (c) (d) - one per line! Feel free to actually use three characters like that instead of those unicode symbols. Also, to be better able to recognize them - did you try to zoom in? If you did that in github, the browser's zoom should do it. git clone and notepad++? Ctrl-Mousewheel works, too. Android Studio? Look for Settings - Editor - General - first option...
- "Y este es el circulo rojo de apunte que lleva al panel para atacar abajo" - I think you misunderstood. 'under' is meant to just reference an earlier line, so - "en (13)" o algo parecido?
- `"Toma prestado" nombres de ciudades` - not sure, but may also be a misunderstanding. Lo que pasa es que... "The Huns" have 1 own city name, the second and any other city they found will "steal" the name from another civilizations's list. Just 'Toman los nombres de sus ciudades...' without the `"` quotes perhaps?

yairm210:
Merging this, changes can be made later

PR #4418, opened 2021-07-07
Update Portuguese.properties

batatasmor:
Adicionei mais traduções.
I added more translations.


batatasmor:
Some translations that were posted after mine were already approved (or rejected) and mine was not. Do you have a problem with mine?

PR #4419, opened 2021-07-07
Add missing pantheons, fix minor bugs

xlenstra:
This PR adds all the pantheons from the base game that were not included in the previous PR for pantheons.
It also makes AI pick random pantheons when they are available, and fixes AI trying to go for mutually exclusive policy branches.
In implementing the pantheons, the following uniques were added and deprecated:

Uniques added:
- "Earn [amount]% of [unitFilter] unit's [param] as [stat] when killed within 4 tiles of a city following this religion"
- "[amount]% attacking Strength for cities"
- "[amount]% cost of natural border growth"
- "[stats] in cities with [amount] or more population"
- "[stats] from cities on [tileFilter] tiles"
- "[unitFilter] Units adjacent to this city heal [amount] HP per turn when healing"
- "[stats] from [tileFilter] tiles without [tileFilter] [cityFilter]"

Uniques deprecated:
- "Increased rate of border expansion" - Was unused anyway

### DIFF
--- a/core/src/com/unciv/logic/city/CityExpansionManager.kt
+++ b/core/src/com/unciv/logic/city/CityExpansionManager.kt
@@ -35,12 +35,17 @@ class CityExpansionManager {
     // The second seems to be more based, so I'll go with that
     fun getCultureToNextTile(): Int {
         var cultureToNextTile = 6 * (max(0, tilesClaimed()) + 1.4813).pow(1.3)
-        for (unique in cityInfo.civInfo.getMatchingUniques("-[]% Culture cost of acquiring tiles []")) {
+        for (unique in cityInfo.civInfo.getMatchingUniques("-[]% cost of acquiring tiles []")) {
             if (cityInfo.matchesFilter(unique.params[1]))
                 cultureToNextTile *= (100 - unique.params[0].toFloat()) / 100
         }
-
-        if (cityInfo.civInfo.hasUnique("Increased rate of border expansion")) cultureToNextTile *= 0.75
+        
+        for (unique in cityInfo.getMatchingUniques("[]% cost of natural border growth")) 
+            cultureToNextTile *= 1 + unique.params[0].toFloat() / 100f
+        
+        // Unique deprecated since 3.15.10 (seems unused, and should be replaced by the unique above)
+            if (cityInfo.civInfo.hasUnique("Increased rate of border expansion")) cultureToNextTile *= 0.75
+        //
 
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/logic/city/CityExpansionManager.kt
```diff
@@ -39,8 +39,13 @@ class CityExpansionManager {
...
                 cultureToNextTile *= (100 - unique.params[0].toFloat()) / 100
         }
-
-        if (cityInfo.civInfo.hasUnique("Increased rate of border expansion")) cultureToNextTile *= 0.75
+        
+        for (unique in cityInfo.getMatchingUniques("[]% cost of natural border growth")) 
+            cultureToNextTile *= 1 + unique.params[0].toFloat() / 100f
+        
+        // Unique deprecated since 3.15.10 (seems unused, and should be replaced by the unique above)
+            if (cityInfo.civInfo.hasUnique("Increased rate of border expansion")) cultureToNextTile *= 0.75
```
yairm210, line 47: I recall this being an American Civ unique

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -268,6 +270,14 @@ class CityStats {
                 val amountOfEffects = (cityInfo.population.population / unique.params[1].toInt()).toFloat()
                 stats.add(unique.stats.times(amountOfEffects))
             }
+            
+            // "[stats] in cities with [amount] or more population
+            if (unique.placeholderText == "[] in cities with [] or more population" && cityInfo.population.population >= unique.params[1].toInt())
+                stats.add(unique.stats)
+            
+            // "[stats] from cities on [tileFilter] tiles"
+            if (unique.placeholderText == "[] from cities on [] tiles" && cityInfo.getCenterTile().matchesTerrainFilter(unique.params[1]))
```
yairm210, line 279: "In cities" sound better that "from cities" to me, you added 2 unique with different phrasings

core/src/com/unciv/logic/city/CityExpansionManager.kt
xlenstra: American civ has the following uniques:
```"uniques": ["+[1] Sight for all [{Military} {Land}] units", "-[50]% Gold cost of acquiring tiles [in all cities]"],```
I also `grep`'ed the entire project, and wasn't able to find this unique anywhere else (except for discontinued translations), which is why I wrote that comment

Issue #4411, opened 2021-07-07
a (probably) major bug and a really minor bug


logicminimal:
1: when you nuke a city, (dont know if you specifically have to aim for the city center but) the city center gets destroyed, and i am almost certain that it will cause some problems, this probably only applies to capitals (and unrazeable cities like soon to be implemented holy cities) while a normal city would get destroyed capitals dont but they still lose their city center improvement

2 (the minor bug): when you nuke a city the city gets destroyed, that is fine but its civilization icon on the minimap does not disappear from the map until you progress onto the next turn

yairm210:
Good catch!

SomeTroglodyte:
@yairm210  - independent of fixing the nuke code:
- Idea - Would we want a safety net 'just in case' in the load game code, like: build two Set<Vector2>, one all cities, the other all tiles with the improvement, then do A-B and add the improvement and B-A and remove?
- Would we want to tie the special rules for city center yields to the improvement, complete with moddability and uniques?
- If yes to the second - could be used to easily fix the fountain of youth to original rules. "Provides yield without a worker" "Cannot be assigned a worker" (those maybe just 1 unique, or explicit stats, or..)...

logicminimal:
also i think the second one is caused due to unciv not updating stats (or atleast stats in the UI) when something happens mid turn, for instance, getting a luxury resource just when you are at -1 happiness (due to another civ proposing so, as trade offers from you require proceeding into the next turn) and therefore so the growth until next turn stays the same until the next turn which shouldn't be the case

also city production doesnt re assign when yields change or it expands manually (and probably automatically)

logicminimal:
@yairm210 is it time to close it?

PR #4423, opened 2021-07-08
City Centers can no longer be removed by nukes

xlenstra:
Fixes part of #4411

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -558,7 +558,7 @@ object Battle {
         }
 
         // Remove improvements, add fallout
-        if (tile.improvement != null && !attacker.getCivInfo().gameInfo.ruleSet.tileImprovements[tile.improvement]!!.hasUnique("Indestructible")) {
+        if (tile.improvement != null && !attacker.getTile().getTileImprovement()!!.hasUnique("Indestructible")) {
             tile.improvement = null
         }
         tile.improvementInProgress = null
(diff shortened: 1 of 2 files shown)


logicminimal:
:pepehappy:

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -618,7 +620,9 @@ object Battle {
         }
 
         // Remove improvements
-        tile.improvement = null
+        if (tile.improvement != null && !attacker.getCivInfo().gameInfo.ruleSet.tileImprovements[tile.improvement]!!.hasUnique("Indestructible")) {
```
yairm210, line 623: I'm prettu sure there's a tile.getImprovement() function or whatnot
tileInfo should have a ruleSet reference as well iirc

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -558,7 +558,9 @@ object Battle {
         }
 
         // Remove improvements, add fallout
-        tile.improvement = null
+        if (tile.improvement != null && !attacker.getCivInfo().gameInfo.ruleSet.tileImprovements[tile.improvement]!!.hasUnique("Indestructible")) {
```
yairm210, line 561: Here too, same as below, !attacker.getTile().getTileImprovement()!!.hasUnique("Indestructible")

xlenstra: I probably should have thought of that, should be fixed now.

logicminimal:
this pr doesn't make nukes unable to destroy razeable cities does it?

xlenstra:
No, because if a city would be destroyed, the function city.destroy() is called. This PR only makes it so the nuke won't remove City Center improvements

PR #4432, opened 2021-07-09
Update Units.json

saud2410:
Battering Ram's unique changed "+[-amount]" to "-[amount]"


yairm210:
This unique isn't currently genericified, so this will just not work.

PR #4435, opened 2021-07-09
City states give gold when met; updates to city state gold gifts

xlenstra:
This PR makes some small changes to city states:
- When a city state is first met, they give gold to the major civ. The first major civ receives 30 gold, all others receive 15.
- It is now possible to give different amounts of gold to city states for influence
- The influence gained from giving gold is dependent on many factors in the base game. These have now all been added.

As this also changes the UI for city states, here are some pictures of the new layout.
Main screen:
image

After pressing 'Give a Gift':
image


android/assets/jsons/translations/template.properties
```diff
@@ -407,6 +407,8 @@ You have entered a Golden Age! =
 [n] sources of [resourceName] revealed, e.g. near [cityName] = 
 A [greatPerson] has been born in [cityName]! = 
 We have encountered [civName]! = 
+[cityStateName] has given us 15 gold as a token of goodwill for meeting us = 
+[cityStateName] has given us 30 gold as we are the first major civ to meet them =
```
yairm210, line 411: I have a feeling that gold given changes with game speed, so these should probably be parameters

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -340,6 +341,26 @@ class CivilizationInfo {
 
         if (isCurrentPlayer() || otherCiv.isCurrentPlayer())
             UncivGame.Current.settings.addCompletedTutorialTask("Meet another civilization")
+        
+        // If exactly one of the civilizations is a city state, we gain gold for meeting them 
+        if (!(isCityState() || otherCiv.isCityState()) || isCityState() && otherCiv.isCityState()) return
+        
+        val cityState = if (isCityState()) this else otherCiv
+        val majorCiv = if (isCityState()) otherCiv else this
+
```
yairm210, line 350: A lot of things are done both ways and checked for each way.
We should make a "one-sided meeting", like "private CivInfo.activateMeetLogic(otherCiv)" and meetCivilization(otherCiv){ otherCiv.activateMeetLogic(this); activateMeetLogic(otherCiv) }

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -144,21 +143,26 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
...
 
         diplomacyTable.addSeparator()
 
-        val giftAmount = 250
-        val influenceAmount = viewingCiv.influenceGainedByGift(otherCiv, giftAmount)
-        val giftButton = "Gift [$giftAmount] gold (+[$influenceAmount] influence)".toTextButton()
-        giftButton.onClick {
-            viewingCiv.giveGoldGift(otherCiv, giftAmount)
-            updateRightSide(otherCiv)
+        val giveGoldButton = "Give a Gift".toTextButton()
```
yairm210, line 159: This text is not added to template.properties

yairm210 (changes requested):

android/assets/jsons/translations/template.properties
xlenstra: Just tested this, and turns out they don't. However, when religious city states are added they provide a small amount of faith when meeting them, so I'll parametrize this as a preparation for that

logicminimal:
also i think you were able to gift units from cstates  or threaten them if they dont give a worker etc

xlenstra:
> also i think you were able to gift units from cstates or threaten them if they dont give a worker etc

True, but not in this PR.

xlenstra:
There appear to be _many_ problems with this PR, I'm closing it for now. I'll test more and fix everything soon.

PR #4442, opened 2021-07-09
Fixed a small bug where unhappiness stopped food multipliers instead of growth multipliers

xlenstra:
When an empire was unhappy, this would remove all percentage food multipliers.
However, the source provided for this decision only said it should remove all percentage _growth_ multipliers.
This PR rectifies this.

### DIFF
--- a/core/src/com/unciv/logic/city/CityStats.kt
+++ b/core/src/com/unciv/logic/city/CityStats.kt
@@ -441,7 +441,7 @@ class CityStats {
         for (entry in newFinalStatList.values) {
             entry.gold *= 1 + statPercentBonusesSum.gold / 100
             entry.culture *= 1 + statPercentBonusesSum.culture / 100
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -479,9 +479,9 @@ class CityStats {
         totalFood = newFinalStatList.values.map { it.food }.sum() // recalculate because of previous change
 
         // Since growth bonuses are special, (applied afterwards) they will be displayed separately in the user interface as well.
-        if (totalFood > 0) {
+        if (totalFood > 0 && !isUnhappy) { // Percentage Growth bonus revoked when unhappy per https://forums.civfanatics.com/resources/complete-guide-to-happiness-vanilla.25584/
             val foodFromGrowthBonuses = getGrowthBonusFromPoliciesAndWonders() * totalFood
-            newFinalStatList["Policies"]!!.food += foodFromGrowthBonuses
+            newFinalStatList["Policies"]!!.food += foodFromGrowthBonuses // Why Policies? Wonders can also provide this?
```
yairm210, line 484: Sure, next pr we can change this

PR #4444, opened 2021-07-09
Laid some groundwork for implementing further religions

xlenstra:
This PR lays groundwork for expanding religions beyond pantheons. 
I would work more on this before publishing, but the next step would be implementing the founding of religions, and that is large enough that I'd rather create a special PR for only that.

### DIFF
--- a/core/src/com/unciv/logic/civilization/ReligionManager.kt
+++ b/core/src/com/unciv/logic/civilization/ReligionManager.kt
@@ -1,5 +1,7 @@
 package com.unciv.logic.civilization
 
+import com.unciv.logic.map.MapUnit
+import com.unciv.models.Religion
 import com.unciv.models.ruleset.Belief
 import kotlin.random.Random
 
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -9,13 +11,13 @@ class ReligionManager {
 
     var storedFaith = 0
 
-    var pantheonBelief: String? = null
+    var religion: Religion? = null
```
yairm210, line 14: Religions shouldn't be stored both in civs and in game info.
There should be only one source of truth for each religion

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -50,14 +52,16 @@ class ReligionManager {
...
         return true
     }
 
-    fun choosePantheonBelief(belief: Belief){
+    fun choosePantheonBelief(belief: Belief) {
         storedFaith -= faithForPantheon()
-        pantheonBelief = belief.name
+        religion = Religion(belief.name, civInfo)
+        religion!!.pantheonBeliefs.add(belief.name)
+        civInfo.gameInfo.religions[belief.name] = religion!!
```
yairm210, line 64: What will happen here, is that the first turn that this happens it will be passed by reference.
However, of you load from a saved game, the religion will be serialised into the save twice, and therefore will be deserislized twice

xlenstra: Ah, that makes sense, hadn't thought of that

PR #4457, opened 2021-07-10
Update English.properties

Kensvin28:
Corrected some misspelling and grammatical errors in the new contents


yairm210:
Good changes, but these need to be done in all files at once to not ruin existing translations - that's on me

PR #4459, opened 2021-07-10
Fixed bug where submarines could not attack embarked units

### LINKED ISSUE #4450: Civ V embarked unit behaviour
Can anyone with a copy of Civ V answer the following:

1. Can submarines attack embarked units? (Unciv: no)
2. Do 'vs land units' bonuses affect embarked units? (Unvic: yes)
3. Do 'vs water units' bonuses affect embarked units? (Unciv: no)

xlenstra:
Fixes #4450

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Units.json
+++ b/android/assets/jsons/Civ V - Vanilla/Units.json
@@ -1004,8 +1004,8 @@
 		"cost": 325,
 		"requiredTech": "Refrigeration",
 		"upgradesTo": "Nuclear Submarine",
-		"uniques": ["+[75]% Strength when attacking", "Invisible to others", "Can only attack [Water] units", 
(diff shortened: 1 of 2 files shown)


yairm210:
I think that instead of two "can only attack" uniques, which looks of, I think a single "can only attack [tilefilter] tiles" should be better

android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -1004,7 +1004,8 @@
 		"cost": 325,
 		"requiredTech": "Refrigeration",
 		"upgradesTo": "Nuclear Submarine",
-		"uniques": ["+[75]% Strength when attacking", "Invisible to others", "Can only attack [Water] units", "Can attack submarines", "Can enter ice tiles"],
+		"uniques": ["+[75]% Strength when attacking", "Invisible to others", "Can only [Water] tiles",
```
yairm210, line 1007: Typo - can only water

PR #4461, opened 2021-07-10
Make map symmetrical if it's not wrapped

lishaoxia1985:


core/src/com/unciv/ui/utils/ZoomableScrollPane.kt
```diff
@@ -11,10 +11,6 @@ open class ZoomableScrollPane: ScrollPane(null) {
     var continuousScrollingX = false
 
     init{
-        // Remove the existing inputListener
-        // which defines that mouse scroll = vertical movement
-        val zoomListener = listeners.last { it is InputListener && it !in captureListeners }
```
yairm210, line 16: I remember there's a reason this is here. Don't remember what it is, though

core/src/com/unciv/ui/map/TileGroupMap.kt
```diff
@@ -120,9 +120,11 @@ class TileGroupMap<T: TileGroup>(tileGroups: Collection<T>, private val leftAndR
 
         // there are tiles "below the zero",
         // so we zero out the starting position of the whole board so they will be displayed as well
-        // The width has to be lowered by groupSize because wrapped maps are missing a tile on the right.
+        // Map's width is reduced by groupSize if it is wrapped, because wrapped map will miss a tile on the right.
         // This ensures that wrapped maps have a smooth transition.
-        setSize(topX - bottomX + leftAndRightPadding * 2 - groupSize, topY - bottomY + topAndBottomPadding * 2)
+        // If map is not wrapped, Map's width doesn't need to be reduce by groupSize
+        if (worldWrap) setSize(topX - bottomX + leftAndRightPadding * 2 - groupSize, topY - bottomY + topAndBottomPadding * 2)
```
yairm210, line 126: I'm pretty sure this already existed, when did this disappear?

core/src/com/unciv/ui/utils/ZoomableScrollPane.kt
lishaoxia1985: compile give a warning "**Calling non-final function removeListener in constructor**"

core/src/com/unciv/ui/map/TileGroupMap.kt
lishaoxia1985: This edit is in the commit #3991, but you revert it, so I split it.

logicminimal:
what exactly does make map symetrical mean? does world wrap even need ot to function?

lishaoxia1985:
@logicminimal Like this, if you don't use the wrap map and this commit, you will find the background map is not in the center of the screen.
YueR

logicminimal:
> @logicminimal Like this, if you don't use the wrap map and this commit, you will find the background map is not in the center of the screen.
> YueR
> 

ok, does this change ingame world wrap though?

lishaoxia1985:
@logicminimal No, When you play the game it is meaningless because in the game the map is not in the center of screen usually.

PR #4462, opened 2021-07-10
Fixed bug where gold from rivers wasn't increased by one during golden ages

### LINKED ISSUE #4454: Golden age tile bonuses dont work properly
The river tiles are supposed to yield one extra gold if they provide at least one gold but it does not despite being in a golden age

xlenstra:
Fixes #4454

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -260,10 +260,10 @@ open class TileInfo {
             if (stats.production < 1) stats.production = 1f
         }
 
+        if (isAdjacentToRiver()) stats.gold++
+
         if (stats.gold != 0f && observingCiv.goldenAges.isGoldenAge())
             stats.gold++
-
-        if (isAdjacentToRiver()) stats.gold++
 
         if (stats.production < 0) stats.production = 0f
 


logicminimal:
also rivers are only one example, there are probably lot many. also they aren't meant to be doubled, this will probably cause bigger problems If not patched, tiles are meant to provide +1 gold if they yield atleast 1 gold during a golden age

xlenstra:
Basically what happens in the code was:
- First calculate the exact yields that a tile has
- If any of them is gold, AND the player is in a golden age, increase the gold by one
- ONLY AFTER THIS calculate the gold you get from tiles next to rivers.

All I did was swap these two. This did not increase the amount of gold you get from river tiles, read the code. This was also the only example, as this line was simply misplaced.

logicminimal:
> Basically what happens in the code was:
> - First calculate the exact yields that a tile has
> - If any of them is gold, AND the player is in a golden age, increase the gold by one
> - ONLY AFTER THIS calculate the gold you get from tiles next to rivers.
> 
> All I did was swap these last two. This did not increase the amount of gold you get from river tiles, read the code. This was also the only example, as this line was simply misplaced. 

why would the code spend an extra step adding 1 gold to the yields of a river tile? doesnt it already have its base yields pre-set?

xlenstra:
Rivers aren't tiles, they modify other tiles, as they can be placed on the edge of basically any tile. There needs to be code _somewhere_ for this modification to happen, and that is in this exact function that is changed.

logicminimal:
> Rivers aren't tiles, they modify other tiles, as they can be placed on the edge of basically any tile. There needs to be code _somewhere_ for this modification to happen, and that is in this exact function that is changed.

oh ok, great work 👍

PR #4464, opened 2021-07-10
Fixed spurious notifications of revealed resources too far away or in foreign territory

freddyhayward:


logicminimal:
i'd suggest including in allied city states' borders as well, woth a max distance of 4

freddyhayward:
> i'd suggest including in allied city states' borders as well, with a max distance of 4

You'd have to decide whether the player really wants to know this. Personally it's disappointing when I see '6 sources of coal revealed!' only to find 1 actually accessible by me. You'd also have to include the city states in the distance calculations. I'm sure that's simple enough but since this is my first PR I'll leave that to someone else.

SomeTroglodyte:
You know, my motivation was 'Well, if one of my units can see it, why not notify about it'? But I'm OK with this, too. But why didn't I use `sortedBy`?

PR #4473, opened 2021-07-11
Fix worldSizeModifier in TechManager

lishaoxia1985:


SomeTroglodyte:
:+1: good catch. My bad - facepalm level.

PR #4475, opened 2021-07-11
Fixed bug where tile construction time was increased instead of decreased

### LINKED ISSUE #4471: Improvement time is increased
**Describe the bug**
Improvement time is not reduced, and it is increased when i choose Liberty-Citizenship

**Expected Behavior**
Building Farm takes 9 turns instead of 7 turns, and road takes 5 turns instead of 7 turns.

**Additional context**
3.15.10

xlenstra:
Fixes #4471


logicminimal:
is it possible that this will be released in a aptch instead of a new version? this seems game breaking

PR #4476, opened 2021-07-11
Fix Hun city names

freddyhayward:


ravignir:
My favorite type of bug in the code, just a missing return

PR #4483, opened 2021-07-12
Move fun updateTransientBooleans() into class CivilizationInfo

lishaoxia1985:


core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -286,15 +286,11 @@ class GameInfo {
...
-
-
-        // This doesn't HAVE to go here, but why not.
-
-        for (civInfo in civilizations) civInfo.setNationTransient()
-        for (civInfo in civilizations) civInfo.setTransients()
-        for (civInfo in civilizations) civInfo.updateSightAndResources()
-
         for (civInfo in civilizations) {
+            civInfo.setNationTransient()
```
lishaoxia1985, line 290: Decrease the loop times, so is the below!😄

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -70,6 +70,25 @@ class CivilizationInfo {
...
 
+    // UnitMovementAlgorithms.canPassThrough is the most called function in the game,
+    // so defining these extremely specific booleans which are usually used in this function
+    // will improve the time cost
+    @Transient
+    var wayfinding = false
+
+    @Transient
+    var unitsCanEmbark = false
+
```
lishaoxia1985, line 81: I think use the method hasUnique directly is easy to read. It's up to you!

core/src/com/unciv/logic/civilization/TechManager.kt
```diff
@@ -25,20 +25,6 @@ class TechManager {
     @Transient
     private var researchedTechUniques = ArrayList<Unique>()
 
-    // MapUnit.canPassThrough is the most called function in the game, and having these extremely specific booleans is or way of improving the time cost
```
lishaoxia1985, line 28: I think we should split policy and tech efforts by use Uniques from policymanager and techmanager, like you do now.

core/src/com/unciv/logic/GameInfo.kt
xlenstra: I don't know if this is possible. Maybe `updateSightAndResources()` requires _other_ civilizations to already have their transients set, and if so, this could lead to crashes when loading saves

lishaoxia1985: @xlenstra Sorry, you are right!😄
I remove this modification.

yairm210:
The reason that all these transients were in the tech class to begin with is because the CivilizationInfo class is way too packed.
The reason that these exist in the first plaace, and we don't use .hasUnique() every time to check them, is because they're part of the decision-making process in finding movement points between adjacent tiles, which is a huge chunk of all turn automation, and this was a massive time saver.
All in all, I don't think that this PR actually improves the state of the game.

If you would create a separate class, called CivilizationInfoTransients or something, and put them all *there*, that could be a valid change.

PR #4487, opened 2021-07-12
Fixed bug where effects of all aquaducts nationwide stacked in each city

xlenstra:
Added uniques:
- "[amount]% of food is carried over after population [cityFilter] increases"

Deprecated uniques:
- "[amount]% of food is carried over after population increases"


yairm210:
The cityFilter should be at the end of the sentence, because "[40]% of food is carried over after population [in all cities] increases" sounds like the *population increased* in all cities, whereas "[40]% of food is carried over after population increases [in all cities]" sounds more like the effect is relevant to all cities.

PR #4490, opened 2021-07-12
Wonder Built Screen

SimonCeder:
This PR adds functionality for small wonder splash screens.
The wonder built popup is slightly reformatted and I made a graphic for The Pyramids as a start.
wondersplash

The basic icons are used as fallback if there is no graphic yet.
fallbacksplash

I'd appreciate any feedback since I'm very new at this. In particular I'm not sure if it makes sense for the wonder graphics to be in the game atlas since they'll probably only be displayed once per game.


SomeTroglodyte:
Tiny question: To me it seems you are forcing aspect ratio relative to screen, so you might want to test android portrait orientation or drag the desktop window to silly sizes. Might also be a problem if someone decided to supply a bunch of wonder images with inconsistent aspect ratios.

Idea: Prepare generalization to allow the same on text-researched popups?

Just FYI Notes:
Placing them under ExtraImages would avoid the automatic atlas packing, though in that case imageExists and getImage wouldn't work. You might check out CivilopediaText.kt lines 176+ for reference.
Possible synergy with #4391 - if accepted we might want those pix in the civilopedia too. Easy since FormattedLine already has the ability to pull such images. Lastly, I already thought about making this popup and at least the Tech one use a linkable update of getShortDescription.
APK size is a factor, so it's good you managed to keep the included example to an incredible 2.6k. It would still add up, and the low size limits the potential visual impact - so mods containing nothing but larger versions of these should be possible. Right now I see no reason why this shouldn't be fully moddable - but using ExtraImages would be a blocker. We might want to make a mod-scanning ImageGetter.getExternalImage in that case.

yairm210:
They shouldn't be in the base game for space consideration purposes, but as modded images they're great!
This is a good addition, just remove the pyramids image to the PR, and add documentation in the wiki saying what the image name should be for such images

SomeTroglodyte:
Closed by yourself? If it's to fine-tune this in peace, then a "P.S.": to "make a mod-scanning ImageGetter.getExternalImage" the Sound class has reference code. Or holler and I could prepare the way.

If you're unsure how to bring this up to date with the new stuff in master, I struggled with that too and took notes - [in here](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Unciv-notes/Working%20in%20the%20IDE%20and%20Contributing.md).

SimonCeder:
Hi, i closed this and opened [https://github.com/yairm210/Unciv/pull/4509](url) mainly because I got unsure how to revert the added image. In retrospect there probably was a more elegant way. Thanks for the link!

SomeTroglodyte:
Ah, okay - well done btw. You _can_ push renames, file moves and deletions. Maybe I should add a paragraph for that sometime. Renames and moves require that you use the "refactor" menu, that's more or less all. Your link would work with just `#4509` as markdown - or `[here is a link to #4509](https://github.com/yairm210/Unciv/pull/4509)` - probably unnecessary to tell you. Some markdown half-secrets are in that same repo I linked above somewhere.

Issue #4347, opened 2021-07-02
Counts for map generation options 


Mape6:
The sliders are really small on an Android device. So it would be cool to have count next to each slider where you could see the percentage of the individual option.

ravignir:
I would also like to have plus and minus arrows to modify these values instead of a slider. The whole thing has thresholds so it makes no sense to have sliders imo.

logicminimal:
also while we're at it the generation should be like in civ 5

Mape6:
> also while we're at it the generation should be like in civ 5

That's not the topic here. Please open a separate request for that. Especially with a more detailed description of what you want.

SomeTroglodyte:
@ravignir  what thresholds exactly? How could it look without a slider? Dropdown select using labels like "none-few-some-average-many-gadzooks"? Label flanked by -/+ buttons? The only slider I know how the underlying values work without checking the source would be coast extension...

Mape6:
I like the idea from @ravignir, too. Imo the sliders does not have so good usability on my Android device as the following:
Just a field where I can see a number between 1 to 9, or whatever the threshold for a slider is, and a minus left and a plus button right of it. If the thresholds are much bigger like 1 to 99, the field should be writable, so that you don't need to tap 98 times on the plus button to get the maximum.
Did I describe my wish clearly enough?

ravignir:
> 
> 
> @ravignir what thresholds exactly? How could it look without a slider? Dropdown select using labels like "none-few-some-average-many-gadzooks"? Label flanked by -/+ buttons? The only slider I know how the underlying values work without checking the source would be coast extension...

I meant that as of now sliders only stop at the predefined values (thresholds). And yes, a label with +/- buttons and current value (whether it is numerical or a text like small-normal-huge) in the middle is what I meant.

SomeTroglodyte:
The problem with the sliders is mostly size and far-too-easy focus loss. I already tried a few times to force them to follow as long as I don't lift the finger, and never got perfection (testing with the volume sliders, not these). Make no sound, disallow scroll, and add 20f padding, and they're markedly easier to 'slide', but still lose you sometimes. And yes I'm speaking of a fast-ish hardware Android device, no Mouse involved. Of course, those changes are already too much of a throwback. And a textbox with +/- buttons would IMHO look ugly - this is a game not a technical tool.

Options:
[...]

PR #4506, opened 2021-07-13
Tweaked Slider, use in MapParametersTable

SomeTroglodyte:
Improved slider and application to one single usecase: Advanced map parameters. Inspired by #4347.
- +/- buttons for easier single steps or stylus or just wasting resources and looking cool
- Shows the actual value as a temporary tooltip. Not implemented using Gdx ToolTip because it denies too many internals, so triggering a 'show' from the +/- buttons only worked after the mouseover tip fired once.
- Loses focus to any ScrollPane it's nested in less often. Should be never, but I can't find the cause of the remaining imperfection. Better than before anyway, you can actually drag them with mouse or finger.

The [video in #4347](https://github.com/yairm210/Unciv/issues/4347#issuecomment-873486707) still applies, and the description of the imperfection below it as well.


yairm210:
Damn

SomeTroglodyte:
An expletive? Is that one even in `templates.properties`? Or did you mean a beaver construction?

PR #4507, opened 2021-07-13
Map size will modify policy cost, techUniques is initialized by UniqueMap(), Replace "effect" with "uniques" in UnitPromotions.json

lishaoxia1985:


core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -181,11 +189,8 @@ class PolicyManager {
         }
 
         for (unique in policy.uniques) {
-            if (unique == "Triggers a global alert") {
```
yairm210, line 184: Why did you delete this?

core/src/com/unciv/models/ruleset/Nation.kt
```diff
@@ -168,7 +168,7 @@ class Nation : INamed {
                 for (unique in originalUnit.uniques.filterNot { it in unit.uniques })
                     textList += "  " + "Lost ability".tr() + "(" + "vs [${originalUnit.name}]".tr() + "): " + unique.tr()
                 for (promotion in unit.promotions.filter { it !in originalUnit.promotions })
-                    textList += "  " + promotion.tr() + " (" + ruleset.unitPromotions[promotion]!!.effect.tr() + ")"
+                    textList += "  " + promotion.tr() + " (" + ruleset.unitPromotions[promotion]!!.uniques.joinToString(",") { it.tr() } + ")"
```
yairm210, line 171: This isn't the first place that the difference between Chinese and other languages (in joinToString(",") vs joingToString(", ")  ) appears.
We should probably have a dedicated function like .joinToStringLanguageDependant()

core/src/com/unciv/models/ruleset/unit/Promotion.kt
```diff
@@ -8,17 +8,16 @@ import com.unciv.models.translations.tr
 class Promotion : INamed{
     override lateinit var name: String
     var prerequisites = listOf<String>()
-    var effect = ""
```
yairm210, line 11: Don't deprecate this. Deprecation should be preceded by a deprecation warning.

core/src/com/unciv/logic/civilization/PolicyManager.kt
lishaoxia1985: you can search this word in policy, it have not been used.🌝🌝

core/src/com/unciv/models/ruleset/unit/Promotion.kt
lishaoxia1985: @yairm210 👌Now I don't edit this.
By the way, about **.joinToStringLanguageDependant()**, this fun should be written in the class Translation.

core/src/com/unciv/logic/civilization/PolicyManager.kt
xlenstra: I added it as an alternative to the other unique below, when you don't want some extra text to show up, so I would keep it for moddability

lishaoxia1985: @xlenstra These can be replaced with "Triggers the following global alert: [param]" if we give a special "[param]"
by the way, don't mix the mod needed code in the original branch😄

PR #4509, opened 2021-07-13
Wonder build screens redux

SimonCeder:
According to feedback from previous PR. Now without added images.
Works in all resolutions. Added a check for portrait mode to use more screen space in that case.
Wonder images can be any size but will get resized to 2:1 ratio.

portraitsplash
wondersplash


yairm210:
Looks good!
Since this isn't in the base game, remember to document this in wiki/mods

core/src/com/unciv/ui/worldscreen/AlertPopup.kt
```diff
@@ -184,11 +184,27 @@ class AlertPopup(val worldScreen: WorldScreen, val popupAlert: PopupAlert): Popu
                 val wonder = worldScreen.gameInfo.ruleSet.buildings[popupAlert.value]!!
                 addGoodSizedLabel(wonder.name)
                 addSeparator()
+                if(ImageGetter.wonderImageExists(wonder.name)) {    // Wonder Graphic exists
+                    if(worldScreen.stage.height * 3 > worldScreen.stage.width * 4) {    // Portrait
```
SomeTroglodyte, line 188: For next time: Consider whether `CameraStageBaseScreen.isPortrait()` might be sufficient.

PR #4510, opened 2021-07-14
Updated the tile choosing algorithm for city expansion

xlenstra:
It is now (mostly) the same as was used in the base game.
I've also looked into the culture thresholds for expanding the border, but neither the formula on the wiki nor the formula currently in use in unciv match my testing in the original, so that'll be fixed at some later point.

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -148,10 +148,9 @@ object Automation {
 
         // Resources are good: less points
         if (tile.hasViewableResource(cityInfo.civInfo)) {
-            val tileResource = tile.getTileResource()
-            if (tileResource.resourceType == ResourceType.Bonus) {
-                if (distance <= 3) score += -105 + 1
-            } else score += -105
+            if (tile.getTileResource().resourceType != ResourceType.Bonus) score -= 105
+            else if (distance <= 3) score -= 104
+            
         } else {
             // Water tiles without resources aren't great
             if (tile.isWater) score += 25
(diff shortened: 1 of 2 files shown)


yairm210:
Please replace all += -, like "+= -1", with -=

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -130,6 +132,71 @@ object Automation {
...
+
+        // Improvements are good: less points
+        score += when (tile.improvement) {
+            // The moddibility of this is questionable
+            null, "Barbarian encampment", "Ancient ruins", "City ruins" -> 0
+            else -> -5
+        }
+
+        // This is in the source code, no can do
+        if (tile.roadStatus != RoadStatus.None) score += 0
```
yairm210, line 170: This is just a meaningless line.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -130,6 +132,71 @@ object Automation {
...
+            // Water tiles without resources aren't great
+            if (tile.isWater) score += 25
+            // Can't work it anyways
+            if (distance > 3) score += 100
+        }
+
+        // Improvements are good: less points
+        score += when (tile.improvement) {
+            // The moddibility of this is questionable
+            null, "Barbarian encampment", "Ancient ruins", "City ruins" -> 0
```
yairm210, line 165: Maybe instead we can check improvements with 0 yield

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -130,6 +132,71 @@ object Automation {
...
+        // Implementing these will require an additional variable for each terrainType
+        val distance = tile.aerialDistanceTo(cityInfo.getCenterTile())
+
+        // Higher score means tile is less likely to be picked
+        var score = distance * 100
+
+        // Resources are good: less points
+        if (tile.hasViewableResource(cityInfo.civInfo)) {
+            val tileResource = tile.getTileResource()
+            if (tileResource.resourceType == ResourceType.Bonus) {
```
yairm210, line 152: Sounds like if resourcetype != Bonus || distance<3

xlenstra: Note the +1 for bonus resources within range, otherwise this could be simplified indeed. Merging it now would result in everything having to be checked twice:
```
if (resourceType != Bonus || distance < 3) score += -105
if (resourceType == Bonus && distance < 3) score += 1
```

PR #4519, opened 2021-07-14
Fixed bug which leads to crashes for almost all users

xlenstra:
So apparently I messed up a double negative, and there should be a ! there.
Without this fix, the diplomacy screen will show all civs the player has _not_ met.
Clicking on any of them will crash your game.

Although it is your choice, I'd advice on bringing out a patch for this.


logicminimal:
but i didnt see other civs in this diplomacy window that i haven't met, why does this happen?

xlenstra:
3.15.12

PR #4530, opened 2021-07-15
Implemented Follower beliefs for religions

xlenstra:
This PR implements follower beliefs for religions - see #4290.
One of these can be chosen when founding the new religion, and they apply to each city following this religion.
Currently, only 6 are implemented, the rest require either purchasing with faith or followers, which have not been fully implemented yet.

Added uniques:
- "[stats] if this city has at least [amount] specialists"
- "[signedAmount]% growth [cityFilter] when not at war"

<details><summary>Pictures of new UI:</summary>

UI when opening the found religion screen:
image

UI after clicking on "choose a follower belief":
image

UI after clicking on "Religious art":
image

</details>

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Beliefs.json
+++ b/android/assets/jsons/Civ V - Vanilla/Beliefs.json
@@ -117,6 +117,27 @@
         "name": "Stone Circles",
         "type": "Pantheon",
         "uniques": ["[+2 Faith] from every [Quarry]"]
+    },
+///////////////////////////////////////// Follower beliefs /////////////////////////////////////////
+    
+    {
+        "name": "Divine inspiration",
+        "type": "Follower",
+        "uniques": ["[+2 Faith] from every wonder"]
+    },
+    {
+        "name": "Feed the World",
+        "type": "Follower",
+        "uniques": ["[+1 Food] from every [Shrine]", "[+1 Food] from every [Temple]"]
+    },
+    {
+        "name": "Peace Gardens",
+        "type": "Follower",
+        "uniques": ["[+2 Happiness] from every [Garden]"]
+    },
+    {
+        "name": "Religious Art",
+        "type": "Follower",
+        "uniques": ["[+8 Culture] from every [Hermitage]"]
     }
-
 ]
--- a/android/assets/jsons/Civ V - Vanilla/Beliefs.json
+++ b/android/assets/jsons/Civ V - Vanilla/Beliefs.json
@@ -126,7 +126,7 @@
     {
         "name": "Divine inspiration",
         "type": "Follower",
-        "uniques": ["[+2 Faith] from every wonder"]
+        "uniques": ["[+2 Faith] from every Wonder"]
     },
     {
         "name": "Feed the World",
--- a/android/assets/jsons/Civ V - Vanilla/Beliefs.json
+++ b/android/assets/jsons/Civ V - Vanilla/Beliefs.json
@@ -120,6 +120,9 @@
     },
 ///////////////////////////////////////// Follower beliefs /////////////////////////////////////////
     
+    // Missing: asceticism (requires followers)
+    // Missing: cathedrals (requires purchasing with faith)
+    // Missing: choral music (requires followers)
     {
         "name": "Divine inspiration",
         "type": "Follower",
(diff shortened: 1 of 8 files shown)


android/assets/jsons/Civ V - Vanilla/Beliefs.json
```diff
@@ -117,6 +117,47 @@
...
+    },
+///////////////////////////////////////// Follower beliefs /////////////////////////////////////////
+    
+    // Missing: asceticism (requires followers)
+    // Missing: cathedrals (requires purchasing with faith)
+    // Missing: choral music (requires followers)
+    {
+        "name": "Divine inspiration",
+        "type": "Follower",
+        "uniques": ["[+2 Faith] from every wonder"]
```
yairm210, line 129: Capital W for Wonder

core/src/com/unciv/models/Religion.kt
```diff
@@ -40,6 +40,24 @@ class Religion() : INamed {
         this.gameInfo = gameInfo
     }
 
+    fun getPantheonBeliefs(): Sequence<Belief> {
+        val rulesetBeleifs = gameInfo.ruleSet.beliefs
```
yairm210, line 44: typo

core/src/com/unciv/models/Religion.kt
```diff
@@ -40,6 +40,24 @@ class Religion() : INamed {
...
+            else rulesetBeleifs[it]!!
+        }.filter { it.type == "Pantheon" }
+            .asSequence()
+    }
+    
+    fun getFollowerBeliefs(): Sequence<Belief> {
+        val rulesetBeliefs = gameInfo.ruleSet.beliefs
+        return followerBeliefs.mapNotNull {
+            if (it !in rulesetBeliefs) null
+            else rulesetBeliefs[it]!!
```
yairm210, line 56: Instead of doing this twice, create a getFollowerBeliefsInRuleset():Sequence<Belief> function

core/src/com/unciv/ui/pickerscreens/FoundReligionPickerScreen.kt
```diff
@@ -89,11 +92,84 @@ class FoundReligionPickerScreen (
...
+    }
+
+    private fun updateLeftTable() {
+        leftChosenBeliefs.clear()
+        val currentReligion = choosingCiv.religionManager.religion ?: Religion("Unknown", gameInfo, choosingCiv.civName)
+        
+        for (pantheon in currentReligion.getPantheonBeliefs()) {
+            val beliefButton = convertBeliefToButton(pantheon)
+            leftChosenBeliefs.add(beliefButton).pad(10f).row()
+            beliefButton.disable()
```
yairm210, line 110: Looks like youre doing the same thing twice. So maybe just for (belief in currentReligion.getFollowerBeliefs() + currentReligion.getPantheonBeliefs())

core/src/com/unciv/ui/pickerscreens/FoundReligionPickerScreen.kt
```diff
@@ -89,11 +92,84 @@ class FoundReligionPickerScreen (
...
+    }
+
+//    private fun setupRightTable() {
+//        val currentReligion = choosingCiv.religionManager.religion ?: Religion("Unknown", gameInfo, choosingCiv.civName)
+//
+//        for (pantheon in currentReligion.getPantheonBeliefs()) {
+//            val beliefButton = convertBeliefToButton(pantheon)
+//            rightBeliefsToChoose.add(beliefButton).pad(10f).row()
+//        }
+//    }
```
yairm210, line 138: Remove

SomeTroglodyte:
Small side observation - A Great Prophet gotten from Liberty complete early will have no religion and an empty abilityUsedCount, so none of its actions will be offered... Would mayUseGreatProphetAtAll+3: `if ((prophet.abilityUsedCount["Religion Spread"] ?: 0) != 0) return false` be the correct approach to fix this? Or exclude Prophets from get free GP? Or both?

xlenstra:
Good catch! I probably will run into this same problem when implementing Missionaries and solve it there. Probably adding it to some postConstruction function() seems like the best approach

logicminimal:
sorru i couldnt find the exact pr but i see in one of the commits merged to the master you disabled great prophets being able to be earned from free great people but in the wiki it says that it could be

	 	

Automatic Faith purchase

Hagia Sophia

Liberty finisher

Leaning Tower of Pisa

Piety finisher

xlenstra:
I disabled it in games that did not have Religion. In games with religion
it is very much enabled.

yairm210:
I'm fine with accepting this as is, but I do think that all the beliefTypes should be an enum of beliefType. Having them as a string when in all honesty they only have 3 acceptable values and no way of even future-modding doesn't sit right.

PR #4531, opened 2021-07-15
TileGroup linting and 0.001% optimization

SomeTroglodyte:
Not quite sure this is worth it, I just came across the file while trying to understand the disappearing farms.

- Linting here means removing dead code, no more public members never used from outside, following Studio's suggestions (not blindly), and later emptying the 'problems' pane (adding words to the dictionary can't be included here, but I left for example 'overlayed' and didn't replace it with 'overlaid' because it's kind of a portmanteau describing a technical concept... and sounds cool).
- Optimization means list reversal, array stuff and a few less (2 to 5) operators in a frequently called function:
- I counted 150000 calls of `setHexagonImageSize` for the first 50 turns of a small map game, which isn't really that much, but still - average 3000 per turn. So premultiply since groupSize is a val and can't change during the instance lifetime... Oh, I should have compared that with the count of instantiations and didn't... Maybe I should take that out.
- List reversal using .reversed() incurs memory operations of twice the size of the List - a copy and afterwards in-place swapping. asReversed() should be close to having a ListIterator starting at the end which would be far less readable.
- addActor; toBack is equivalent to addActorAt(0) but incurs more copying and some handfuls of extra checks - remember Gdx children collections are Arrays under the hood and altering ZIndex means a lot of copying around because ZIndex _is_ the Array index.
- Similarly the remove loop: Starting the remove at low indexes copies also soon-to-be-removed entries, starting at the highest index avoids that. And the first list element _is_ at low indices in its parent's children array...

That said the performance gain is probably not measurable...

### DIFF
--- a/core/src/com/unciv/ui/tilegroups/TileGroup.kt
+++ b/core/src/com/unciv/ui/tilegroups/TileGroup.kt
@@ -18,9 +18,9 @@ import com.unciv.ui.cityscreen.YieldGroup
 import com.unciv.ui.utils.ImageGetter
 import com.unciv.ui.utils.center
 import com.unciv.ui.utils.centerX
-import com.unciv.ui.utils.toLabel
 import kotlin.math.PI
 import kotlin.math.atan
+import kotlin.math.atan2
 import kotlin.random.Random
 
 /** A lot of the render time was spent on snapshot arrays of the TileGroupMap's groups, in the act() function.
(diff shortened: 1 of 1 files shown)


logicminimal:
this is a great enough optimisation, now lets implement ai that can think 50 turns ahead with a neural network that requires a specialized computer

PR #4548, opened 2021-07-16
Small fixes

SomeTroglodyte:
Single commits - their descriptions says it all, except:
"create water improvements outside borders" - can't find evidence, but I think allowing them everywhere is silly. If the rules are correct as they are, this commit would still be OK - the offshore platform and fishing boats in this case should get the allowed outside borders unique. This way means better moddability, too. I think.

### DIFF
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -599,7 +599,7 @@ class CityInfo {
     }
     
     fun canBeDestroyed(): Boolean {
-        return !isCapital() && !isHolyCity()
+        return !isOriginalCapital && !isCapital() && !isHolyCity()
     }
 
     //endregion
--- a/core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
+++ b/core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
@@ -490,10 +490,8 @@ object UnitActions {
                         city.cityStats.update()
                         city.civInfo.updateDetailedCivResources()
                     }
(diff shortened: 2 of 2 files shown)


logicminimal:
what about creating sea improvements in city state borders and creating archeological sites in other civs? are those disabled in this pr too?

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -599,7 +599,7 @@ class CityInfo {
     }
     
     fun canBeDestroyed(): Boolean {
-        return !isCapital() && !isHolyCity()
+        return !isOriginalCapital && !isCapital() && !isHolyCity()
```
yairm210, line 602: Hm, this should probably replace the isCapital actually

SomeTroglodyte: Maybe. You know the codebase better than me. I'm actually sure the behaviour is a regression but haven't yet browsed the histories to look for the break...

yairm210: This is part of an attempt to enable 'capture all original capitals' as a victory condition in the future, so yes, it should replace.
This was written before isOriginalCapital even existed, hence the slight inaccuracy

SomeTroglodyte: I seem to remember not being able to raze conquered city states no matter what and now I can? `isOriginalCapital` must have already existed in that 'before' picture. Probably some other code changed. But `canBeDestroyed` sounds logical, unequivocal, and if it's to hold to the promise of its name it should respect `isOriginalCapital` methinks...

PR #4549, opened 2021-07-16
ExpanderTab persists open/closed state

SomeTroglodyte:
I expect you to criticize harshly, and that's fine - I'm kind of stuck on the decision whether basic decisions are good or utter BS, so I present it. However, it works and does everything I set out to do.

Specifically, choices in doubt:
- Differentiate Expanders by title and "client" - with the latter being done by pulling the method they're created in from a call stack. Which wasn't enough for the trade offers...
- The method chosen to avoid ugly json
- dispose() on the CameraStageBaseScreen as saving hook - Gdx openly admits that method of the Screen interface is never called automatically - so it's a bit pointless. We dont' do so with perfect reliability ourselves, and saving the states at that point in time is often too late - since screens are so often ditched and recreated, and the new screen should read the states saved by the old screen, the old screen would need to save states before we're done with it???


yairm210:
I think the correct phrase to employ here is "just...no."

SomeTroglodyte:
Try playing with and without and you would see how many screen re-instantiations reset those expanders and how impractical that feels. IMHO when you just have researched an important building and want to add it to most construction queues - going through all cities with the left/right arrows - the units expander _staying_ closed is worth gold. Or trades - each gold per turn increase reopening all expanders you may have just closed...

I get this may be an ugly way to achieve that. So - I'll keep an ear/eye open for alternate approaches?

PR #4553, opened 2021-07-17
Fix for missing Farms (all mod TileSetConfigs were always active)

### LINKED ISSUE #4266: Farms pixel improvement not visible for specific Android user
3.15.4 dont think i need to provide a save file

### LINKED ISSUE #4362: Forest on Tundra Hills not displayed
Screenshot_20210704_094714

In short, everything else about forest on tundra hills is normal (as in I can cut it down and build lumber mills as usual), except that it's not displayed on the map. This can be very misleading, especially when I'm playing as Iroquois.

SomeTroglodyte:
Result of this comment: https://github.com/yairm210/Unciv/issues/4266#issuecomment-881034942_
Resolves #4266 for me, may also resolve #4362 - untested.

I definitely feel like I haven't poured enough brain power and/or testing into this, so please, review carefully, find testers, whatever.

Oops I left thge asserts in, these were meant to go before push - too tired now.

### DIFF
--- a/core/src/com/unciv/models/tilesets/TileSetCache.kt
+++ b/core/src/com/unciv/models/tilesets/TileSetCache.kt
@@ -10,14 +10,22 @@ object TileSetCache : HashMap<String, TileSetConfig>() {
     private data class TileSetAndMod(val tileSet: String, val mod: String)
     private val allConfigs = HashMap<TileSetAndMod, TileSetConfig>()
 
-    fun assembleTileSetConfigs() {
-        val mods = mutableSetOf("")     // Vanilla always active, even with base ruleset mods selected
+    /** Combine [TileSetConfig]s for chosen mods.
+     * Vanilla always active, even with a base ruleset mod active.
+     * A mod with a mod name matching the currently chosen tileset is used if in permanent visual mods.
+     *    Note the mod name must exactly match the [TileSet] name with `" Tileset"` appended.
+     * Other active mods can be passed in parameter [ruleSetMods], if that is `null` and a game is in
+     * progress, that game's mods are used instead.
+     */
+    fun assembleTileSetConfigs(ruleSetMods: HashSet<String>? = null) {
+        val mods = mutableSetOf("")
         if (UncivGame.isCurrentInitialized()) {
             val settings = UncivGame.Current.settings
             if (settings.tileSet + " Tileset" in settings.visualMods) mods.add(settings.tileSet + " Tileset")
-            if (UncivGame.Current.isGameInfoInitialized()) {
+            if (ruleSetMods != null)
+                mods.addAll(ruleSetMods)
+            else if (UncivGame.Current.isGameInfoInitialized())
                 mods.addAll(UncivGame.Current.gameInfo.ruleSet.mods)
-            }
         }
         clear()
         allConfigs.filter { it.key.mod in mods }.forEach {
--- a/core/src/com/unciv/models/tilesets/TileSetCache.kt
+++ b/core/src/com/unciv/models/tilesets/TileSetCache.kt
@@ -3,24 +3,44 @@ package com.unciv.models.tilesets
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
 import com.unciv.JsonParser
+import com.unciv.UncivGame
 import com.unciv.ui.utils.ImageGetter
 
-object TileSetCache : HashMap<String, TileSetConfig>(){
-    fun loadTileSetConfigs(consoleMode: Boolean = false, printOutput: Boolean = false){
+object TileSetCache : HashMap<String, TileSetConfig>() {
+    private data class TileSetAndMod(val tileSet: String, val mod: String)
+    private val allConfigs = HashMap<TileSetAndMod, TileSetConfig>()
+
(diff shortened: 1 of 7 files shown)


SomeTroglodyte:
... and I forgot map editor doesn't set gameInfo but manages its mods itself.

SomeTroglodyte:
OK I like this better. Map editor works, that "TundraJungle" mystery is understood - that came from inactive DeCiv - and #4362 tested too.

core/src/com/unciv/models/tilesets/TileSetCache.kt
```diff
@@ -3,24 +3,52 @@ package com.unciv.models.tilesets
...
+     * A mod with a mod name matching the currently chosen tileset is used if in permanent visual mods.
+     *    Note the mod name must exactly match the [TileSet] name with `" Tileset"` appended.
+     * Other active mods can be passed in parameter [ruleSetMods], if that is `null` and a game is in
+     * progress, that game's mods are used instead.
+     */
+    fun assembleTileSetConfigs(ruleSetMods: HashSet<String>? = null) {
+        val mods = mutableSetOf("")
+        if (UncivGame.isCurrentInitialized()) {
+            val settings = UncivGame.Current.settings
+            if (settings.tileSet + " Tileset" in settings.visualMods) mods.add(settings.tileSet + " Tileset")
```
yairm210, line 24: Not sure why this is needed.
If you want a tileset to be available, make the mod permanently visible and that's it. We shouldn't have to determine what miss the tiles come from.

yairm210:
It seems to me that active tileconfigs come from the same set of miss add the active images.
Mods active in current game plus permanent visual mods.

core/src/com/unciv/models/tilesets/TileSetCache.kt
SomeTroglodyte: That's the idea - we no longer mix all TileSetConfigs together because then we got rules from DeCiv while the atlas from DeCiv isn't loaded which causes the observed render errors. When a mod is requested in GameInfo - merge that TileSetConfig, because the atlas will be active too. When a mod is permavisual we still need a way to merge the TileSetConfig when it's needed...

This may be a knot in my brain... What does marking 5Hex as permavisual but then _not_ selecting it in options actually do? Is merging the ruleVariants in that case good or bad? As I wrote that I thought 'bad' - but since the atlas is in and the merging is per TileSet - the actual answer is 'good'?

Also, my pinglish fails at your uses of "miss" - twice - may be some simple typo I don't get. If it was something important, please choose different words.

yairm210: "What do you mean, 'miss'?"
"I'm sorry, I have a cold"

That is to say, typo

SomeTroglodyte: So the miss didn't miss a missing m'iss'eaning - OK

yairm210: I still disagree with the implementation of this line, though.
If I want to have a new tileset, I shouldn't be limited to calling it something ending with " Tileset".

yairm210: Let me rephrase - I think this line shouldn't exist.
Permanent visual mods should come through the rulesetMods parameter, above.
Game-specific mods should come through the `mods.addAll(UncivGame.Current.gameInfo.ruleSet.mods)` below.

SomeTroglodyte: And you're right. Sorry for not marking this back to draft earlier - the longer I think about it the more I tend to think 'I need to _test_ the case a TileSet mod is marked visual but _not_ selected in options'. I'm stalling because I'm not sure how to test properly - that is, which ruleVariants I should look out for...

SomeTroglodyte:
OK - I feel better about this but haven't tested much.

Another idea: Ditch this and instead forbid mods from messing with vanilla tilesets? Currently it's only DeCiv Redux that comes with a FantasyHex.json, and DeCiv's basic premise would be much better served by actually renaming the tileset because it's depicting a different world...

logicminimal:
> OK - I feel better about this but haven't tested much.
> 
> Another idea: Ditch this and instead forbid mods from messing with vanilla tilesets? Currently it's only DeCiv Redux that comes with a FantasyHex.json, and DeCiv's basic premise would be much better served by actually renaming the tileset because it's depicting a different world...

what if here are mods with the sole reason of changing fantasyhex? (more tile layer combinations retextures etc)

SomeTroglodyte:
> mods with sole

Then they need to fork instead of interfering. Was just an idea, not saying the best one, but avoids a lot of complications...

yairm210:
Looks good! Ready for merge as far as I'm concerned. What say you?

SomeTroglodyte:
... OK. My mediocre gut feeling (I would have felt better if we had more capable coders review and/or observant testers test) may be pacified with - Can't get worse, improves one rare situation markedly, and we'll fix whatever hydra rears their heads up when they do...

yairm210:
Precisely.
We'll burn that bridge when we get to it ;)

SomeTroglodyte:
"Ignus"!

PR #4555, opened 2021-07-18
Let translators provide strings for MultiplayerTurnCheckWorker

SomeTroglodyte:
I just stumbled upon android/res/values/strings.xml and noticed a certain lack of translations - or the reverse.
- Option 1: Ignore
- Option 2: Delete german and french translations to restore all-humans-are-created-equal
- Option 3: Delegate to our 'normal' translators.

For Option 3, this would be one way. Wait for the `# requires translation` to propagate and the diligent worker bees to provide. Harvest by manual copy and paste. Keys used so format conversion could be done with a simple notepad regex replace.

### DIFF
--- a/android/assets/jsons/translations/French.properties
+++ b/android/assets/jsons/translations/French.properties
@@ -4464,3 +4464,10 @@ This is where you spend most of your time playing Unciv. See the world, control 
  # Requires translation!
 What you don't see: The phone/tablet's back button will pop the question whether you wish to leave Unciv and go back to Real Life. On desktop versions, you can use the ESC key. = 
 
+
+################### From XML ###################
+Notify_YourTurn_Short = Unciv - C'est à vous !
+Notify_YourTurn_Long = Vos amis attendent votre action.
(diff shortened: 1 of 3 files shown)


yairm210:
I think the strings xml should not be used.
If we can use the regular .tr() translations, then great.
If not, no translations are better than obscure unknown ones.

SomeTroglodyte:
I think this is the stuff displayed in Android's notification drawer...

> If we can use the regular .tr() translations, then great.

Not impossible but significant overhead - The uses of these Strings live in the android-specific code and so to access the tr() function would require another one of those interface+pointer constructs in UncivGameParameters, marshalling data in both directions... Intimidating.

> no translations are better than obscure unknown ones

Someone must have done them originally... 'wrov' early 2020 - you know what, this was mainly to make you aware (shortly - to be forgotten soon) that here is another kind of translation in the codebase without all the nice automatics. So I leave it to you whether to ignore or delete those one-and-a-half translations. This change here I agree is a "NO".

PR #4557, opened 2021-07-18
Unit gifting

SimonCeder:
Adds the ability to gift units to city states or major civs. The button appears when you are in a non-hostile territory. Gifting a unit to a city-state gives 5 influence, gifting to a major civ gives 5 relationship bonus that decays within 20 turns.

Currently mostly useful for helping city-states that are attacked, but will be needed to implement certain civs.

gift


xlenstra:
To solve the failing checks, you should add the text for the unit action to `template.properties`, and (iirc) to a file for any non-english language. I don't think you need to translate it, only add it to the file

SimonCeder:
thanks!

logicminimal:
were there any way to gift units outside the other civs territory in the original game?

SimonCeder:
When the game was first released there was an option to "gift a unit" from the city state interaction screen. You could then pick any unit on the map and it would start a countdown for travel time for 3 turns, then it would arrive at the city state. This was later removed, probably because it was confusing to players.

I elected to not implement this because:
1. it was removed from the later versions of the game anyway
2. it's a lot trickier to implement

So I'm not planning to do that at this time.

yairm210:
Is gifting units to major civs in the base game?

ravignir:
i will provide a broader answer in 30 minutes. but answer is yes generally

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -658,6 +658,17 @@ class MapUnit {
             .forEach { unit -> unit.destroy() }
     }
 
+    fun gift(recipient: CivilizationInfo) {
+        civInfo.removeUnit(this)
+        civInfo.updateViewableTiles()
+        // all transported units should be destroyed as well
+        currentTile.getUnits().filter { it.isTransported && isTransportTypeOf(it) }
```
yairm210, line 665: IIRC unit.destroy() does this internally, no?

yairm210: I definitely recall this being somewhere. Also that air units try to move to closest city if they still have movement points so that they aren't destroyed.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -611,4 +612,45 @@ object UnitActions {
...
+            else recipient.getDiplomacyManager(unit.civInfo).addModifier(DiplomaticModifiers.GaveUsUnits, 5f)
+
+            unit.gift(recipient)
+            // UncivGame.Current.worldScreen.shouldUpdate = true
+        }
+
+        if (unit.civInfo.playerType == PlayerType.AI)
+            return UnitAction(UnitActionType.GiftUnit,  uncivSound = UncivSound.Silent, action = giftAction)
+
+        return UnitAction(UnitActionType.GiftUnit, uncivSound = UncivSound.Silent, action = giftAction)  // sound?
```
yairm210, line 654: Both AI and player actions should return the same thing. The sound is only activated when a player clicks the button, so won't activate for AI actions.

core/src/com/unciv/logic/map/MapUnit.kt
SimonCeder: It does, but also removeFromTile(), which we don't want

SimonCeder:
> Is gifting units to major civs in the base game?

It definitely is as of G&K

ravignir:
Ok, so here is a more detailed answer.

Gifting units to city states: You can give a military unit (any except nukes) to a city state from ANY place on the map (even if unit can't travel, has no movement or is surrounded by an enemy units inside enemy land) and it will provide +5 Infuence (but only after unit reaches city state). The unit will appear under City state control in 3 turns (In BNW it is limited to gifting one unit at a time - until unit reaches city state, in G&K any number of units can be gifted at once)

Gifting units to major civs (or city state): you can give any unit, it just has to be within their borders and has to have some movement left. It will instantenously switch owership once gifted. Provides no diplomacy change(unless it is not displayed). If gifted to a city state, it will provide 5 influence (it has instant effect, so it is not limited to one at a time).

SimonCeder:
You're right, gifting distant units is still in, just moved to a different part of the menu.
Do we want this in?
In that case I can see this done in one of two ways:
1) From the diplomacy menu, choose "gift a unit", then go back to the world map to select a unit. This is how Civ V does it.
2) From the unit orders list, choos "gift unit", create a popup asking which civ to gift to. Slightly easier to implement and might be easier on a touchscreen.

ravignir:
also in vanilla/g&k gifting units to CS (whichever way) provides at max 3 units times 2 influence (almost nothing).
In BNW it is always 5 Influence per unit.
oz

logicminimal:
sorry for late feedback but is there any way you could put a confirmation pop up while gifting units? (i just lost my scout to it)

PR #4560, opened 2021-07-18
Sort Civilopedia entries using locale

### LINKED ISSUE #4353: Civilopedia should sort using alphabet with non-english letters
Currently, Civilopedia lists by English alphabet, and then it puts other letters, which is annoying. 'Č' goes after C, but it's after Z.
On the other side, I don't know if all computer alphabets don't sort like that.

SomeTroglodyte:
Before: Sort goes by 'natural order' (from kotlin doc), which can be read as 'binary collation'. After: uses system default locale. Tests for german show Umlauts are now properly sorted with their nearest vowel, not at the end of the lists. Should resolve #4353.

Note: It may be better to use a locale matching the user-selected language, but we'd need a locale code mapping first.

### DIFF
--- a/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
+++ b/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
@@ -10,6 +10,7 @@ import com.unciv.models.ruleset.Ruleset
 import com.unciv.models.stats.INamed
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.*
+import java.text.Collator
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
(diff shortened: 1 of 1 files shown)


yairm210:
Nice! First time I hear of such a thing!

SomeTroglodyte:
So, start thinking how / where we could "get" locales matching the user-selected Unciv language... Right in the translation files as `_sort_locale_ = en` or somesuch?

`Collator.getInstance()` can also take an argument, and that's what we would need to feed to get culture-specific sorting. I expect system default would mostly work well, but if someone plays Thai on a box set to `en` it's [likely to not give expected results](https://medium.com/digital-thai-valley/kotlin-sort-locale-with-inline-function-example-thailand-provinces-3d6ce80b188f).

PR #4566, opened 2021-07-19
Added Sweden Civ

SimonCeder:
Adds the civilization Sweden, with UUs Carolean and Hakkapeliitta, and the UP Nobel Prize.
Added uniques:

- "Transfer Movement to General"
- "+[15]% Combat Bonus when stacked with Great General"
- "Gain [] [] with a [] gift to a City-State"
- "When declaring [], both parties gain a +[]% boost to []"

I took the liberty of redoing the city name list. The new list includes only cities in modern Sweden (so no Finland), and no Viking Age settlements that are not currently inhabited.


android/assets/jsons/translations/English.properties
```diff
@@ -5292,6 +5292,123 @@ Coevorden =
...
+Gränna = 
+ # Requires translation!
+Kiruna = 
+ # Requires translation!
+Borgholm = 
+ # Requires translation!
+Strängnäs = 
+ # Requires translation!
+Sveg = 
+
```
xlenstra, line 5411: It looks like you did this manually, this is not necessary. Before each update, Yairm regenerates all these files automatically, so you don't have to.

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -74,8 +74,15 @@ object BattleDamage {
             val nearbyCivUnits = combatant.unit.getTile().getTilesInDistance(2)
                 .flatMap { it.getUnits() }.filter { it.civInfo == combatant.unit.civInfo }
             if (nearbyCivUnits.any { it.hasUnique("Bonus for units in 2 tile radius 15%") }) {
-                val greatGeneralModifier =
+                var greatGeneralModifier =
                     if (combatant.unit.civInfo.hasUnique("Great General provides double combat bonus")) 30 else 15
+
+                // Hakkapeliitta unique
+                for (unique in combatant.unit.getMatchingUniques("+[]% Combat Bonus when stacked with Great General")) {
+                    if (combatant.unit.getTile().getUnits().any { it.hasUnique("Bonus for units in 2 tile radius 15%") })
```
xlenstra, line 82: Maybe `it.hasUnique("Great Person - [War]")`?

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -344,7 +344,20 @@ class CityInfo {
                 if (!matchesFilter(unique.params[1])) continue
                 stats[entry.key]!!.timesInPlace(1 + unique.params[0].toFloat() / 100f)
             }
-            
+
+            // Sweden UP
+            var friendshipMultiplier = 0f
+            for (otherciv in civInfo.getKnownCivs()) {
+                if (civInfo.getDiplomacyManager(otherciv).hasFlag(DiplomacyFlags.DeclarationOfFriendship)) {
+                    for(ourunique in civInfo.getMatchingUniques("When declaring [], both parties gain a +[]% boost to []"))
```
xlenstra, line 352: - `friendship` does not need to be parametrized, as it cannot be filled with any other value. 
- The `+` sign should be inside the parameterization, to allow a `-` sign in some situations
- Do you keep track of the different amounts for the different stats? From what you have written it looks like you multiply all stats equally?

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -634,6 +634,16 @@ class MapUnit {
         attacksThisTurn = 0
         due = true
 
+        // Hakkapeliitta movement boost
+        if(hasUnique("Bonus for units in 2 tile radius 15%")) {
```
xlenstra, line 638: I would use `getMatchingUniques("Great Person - []").any { it.params[0] == "War" }` here, as that is the more general form of great generals.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -636,16 +644,25 @@ object UnitActions {
...
         val giftAction = {
             if (recipient.isCityState()) {
-                if (unit.isGreatPerson())
-                    recipient.getDiplomacyManager(unit.civInfo).influence += 90
-                else
-                    recipient.getDiplomacyManager(unit.civInfo).influence += 5
+                for (unique in unit.civInfo.getMatchingUniques("Gain [] [] with a [] gift to a City-State")) {
+                    if((unit.isGreatPerson() && unique.params[2] == "Great Person")
+                        || unit.matchesFilter(unique.params[2])) {
+                        // Slightly hardcoded as we assume that influence is what should be gained
```
xlenstra, line 650: If you do that, just replace the parameter with 'influence' directly

xlenstra:
A tip: currently this PR also contains all commits you made for other PR's. If you do `git fetch upstream` followed by `git rebase upstream/master` on your master branch before creating a new branch instead of merging, these extra commits will not show in new PR's.

android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -882,6 +882,23 @@
...
+		"unitType": "Mounted",
+		"replaces": "Lancer",
+		"uniqueTo": "Sweden",
+		"movement": 4,
+		"strength": 25,
+		"cost": 185,
+		"requiredTech": "Metallurgy",
+		"requiredResource": "Horses",
+		"uniques": ["Can move after attacking","No defensive terrain bonus","-[33]% Strength vs [City]",
+			"Transfer Movement to General", "+[15]% Combat Bonus when stacked with Great General"],
```
xlenstra, line 896: As far as I understand it, we're trying to move away from `combat bonus` and towards `strength`. Also, you can parametrize the unit names here, so imo it's better to use `[+15]% Strength when stacked with [unitFilter] units` and `Transfer Movement to [unitFilter] units`

core/src/com/unciv/logic/city/CityInfo.kt
SimonCeder: I'm unsure what you mean by the last point? In this loop the only entries should be GP points from buildings and GP points from specialists, correct? In that case they should be multiplied equally.
Also probably [great people generation] makes little sense as a parameter as I didn't code an implementation for science or whatever

xlenstra: I misunderstood your code, for some reason I thought it would be used for `[amount]% boost to [Science]` or something instead of [great people generation]. It does make sense now that I look at it again.

android/assets/jsons/Civ V - Vanilla/Units.json
SimonCeder: I changed the + Strength when stacked part, now it makes less assumptions.
I _could_ parametrize the transfer movement one as well, but the problem is that it is checked during the great generals start of turn. So if it was made generic every unit in the game would have to check every turn if they are stacked with a unit that can give them extra movement, which seems unneccesary?

core/src/com/unciv/logic/map/MapUnit.kt
SimonCeder: ok, done

core/src/com/unciv/logic/city/CityInfo.kt
SimonCeder: I mean, it could, but I didn't implement that

android/assets/jsons/Civ V - Vanilla/Units.json
xlenstra: Good point, I'll leave that for @yairm210 to decide.

core/src/com/unciv/logic/battle/BattleDamage.kt
SomeTroglodyte: I see the `+[]` went away, too - :+1: - in code but the old unique is still in templates? Maybe even "fix" the original PR comment so when Yair reads it no wrong assumptions?

android/assets/jsons/Civ V - Vanilla/Units.json
yairm210: I think that's a fair check. Most units are the only unit on their tile anyway, and "once per turn" isn't a heavy cost.

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -76,7 +76,20 @@ object BattleDamage {
             if (nearbyCivUnits.any { it.hasUnique("Bonus for units in 2 tile radius 15%") }) {
                 val greatGeneralModifier =
                     if (combatant.unit.civInfo.hasUnique("Great General provides double combat bonus")) 30 else 15
+
                 modifiers["Great General"] = greatGeneralModifier
+
+                // Hakkapeliitta unique
+                for (unique in combatant.unit.getMatchingUniques("[]% Strength when stacked with []")) {
+                    var stackedUnitsBonus = 0
+                    if (unique.params[1] == "Great General" && combatant.unit.getTile().getUnits().any { it.getMatchingUniques("Great Person - []").any { it.params[0] == "War" } })
```
yairm210, line 85: If the recognition is by name, then the unit should be by name as well.
`combatant.unit.getTile().getUnits().any{it!=combatant.unit && it.name==unique.params[1]}

yairm210: Actually matchesFilter as used below is even better.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -634,6 +634,16 @@ class MapUnit {
         attacksThisTurn = 0
         due = true
 
+        // Hakkapeliitta movement boost
+        if(getMatchingUniques("Great Person - []").any { it.params[0] == "War" }) {
```
yairm210, line 638: This you already know should be by template name ;)

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -625,7 +625,15 @@ object UnitActions {
...
         if (recipient.isCityState()) {
-            if (unit.isGreatPerson()) return null    // Unless Sweden
+            if (unit.isGreatPerson()) {
+                // Do we have a unique ability to gift GPs?
+                var canGiftGPs = false
+                for (unique in unit.civInfo.getMatchingUniques("Gain [] Influence with a [] gift to a City-State")) {
+                    if(unique.params[1] == "Great Person")
+                        canGiftGPs = true
+                }
+                if (!canGiftGPs) return null
```
yairm210, line 635: Can be simplified to 
if(unit.civInfo.getMatchingUniques("Gain [] Influence with a [] gift to a City-State").none{it.params[1]=="Great Person"} return null

core/src/com/unciv/logic/city/CityInfo.kt
SimonCeder: Having thought about it I simply removed the parameter for [great people generation] so it reflects the implementation

SomeTroglodyte:
> A tip: currently this PR also contains all commits you made for other PR's. If you do `git fetch upstream` followed by `git rebase upstream/master` on your master branch before creating a new branch instead of merging, these extra commits will not show in new PR's.

If you're curious, I seem to have a different approach, mostly using Android Studio's GUI only. Look [in my notes](https://github.com/SomeTroglodyte/Scratchpad/tree/master/Unciv-notes). Observable difference: Xander's online master is at the point where he forked or so, mine is where I have followed up to - meaning you can see what my fresh code would be based on. Also, I see mostly force-pushes in his stuff, which I normally don't need. I mention because I trust you will come up with the perfect mix/synthesis and tell us how and why, so we can improve...

xlenstra:
> > A tip: currently this PR also contains all commits you made for other PR's. If you do `git fetch upstream` followed by `git rebase upstream/master` on your master branch before creating a new branch instead of merging, these extra commits will not show in new PR's.
> 
> If you're curious, I seem to have a different approach, mostly using Android Studio's GUI only. Look [in my notes](https://github.com/SomeTroglodyte/Scratchpad/tree/master/Unciv-notes). Observable difference: Xander's online master is at the point where he forked or so, mine is where I have followed up to - meaning you can see what my fresh code would be based on. Also, I see mostly force-pushes in his stuff, which I normally don't need. I mention because I trust you will come up with the perfect mix/synthesis and tell us how and why, so we can improve...

To further make clear what I am doing: I work with GitHub from the command line and haven't really used Android Studio's GitHub interface, so I don't know how that works. I have a local master which I reguarly update (fetch + rebase) to load in the new versions, but I don't regularly push this as it doesn't really need to be pushed (should just reflect a recent master version). From there I branch away into new branch (git checkout -B "blabla") to implement different features. I don't believe I have ever had to force-push stuff using this approach, except for the one time where I accidentally pushed some files in the wrong branch, so I don't really know where you have seen all these phantom force-pushes :).

This is of course only my approach, you can always come up with your own better one :)

SomeTroglodyte:
Good to hear. Command line and UI are of course different worlds, so a direct comparison is more or less moot. On CL you need more brain storage and -power to keep yourself aware of what relates to what else. Since I don't have that covered well, I prefer the UI kludge which _can_ help (but also distract or mislead). 
 Keeping online master on par with Yair's master is also just a tool to keep up mentally, too - like a marker "I still need to evaluate any progress beyond this" - which one can also keep elsewhere. I only needed the CL git once, documented the how and why in those notes and promptly had to refer to them shortly after to repro on another rig...
 And force-pushes - I can still find one easily : "xlenstra force-pushed the xlenstra:borderGrowth branch from 1fc4616 to 20fc5a7 6 days ago". But you're right they're rare and definitely not cause for criticism - I probably just stumbled across two of them and got curious as in "is there a secret to learn here?".

Have Phun! :+1:  :partying_face:

PR #4567, opened 2021-07-19
Rename Railroad tech to Railroads

### LINKED ISSUE #4390: Allow different translation for "Railroad tech" and "Railroad tracks"
In german, "Railroad" is currently translated as "Schienen" meaning the pair of iron rails on the ground a train might move on. For the technology, this is not really appropriate on the level of "slightly grating my nerves", it should be "Eisenbahn".

Easiest would be to leave the tracks (which are hardcoded everywhere) as they are and rename the tech to "Railroading", "Railroad tech", "_railroad_tech_key" or "a8a98dc2-69ab-4394-8ea3-3c28cff6cc14", and translate that to "Railroad" in English.properties.

Anyone with a better idea? Or pro/con votes?

SomeTroglodyte:
#### Motivation:
- Resolve #4390.

#### To be considered / please comment:
- String "Railroad" has an almost independent meaning as png filename in the tileset root - I left that hardcoded. It has a `val` in `TileSetStrings` which is only ever used once - in `TileGroup.updateRoadImages`.
- I noticed "Road.png" in the tileset root exists but is not used (`getDot(Color.BROWN)` is unsed instead). Wouldn't that frustrate modders?
- I moved some temporary compatibility code into `TechManager`, otherwise I would have been forced to make public more than I cared for.
- While there I noticed `scienceSpentOnTech` and `researchOfTech` did the same thing, so I consolidated. The older name won.
- A city stats bonus list named "Railroad" exists, and I chose to adapt it to the tech name. Could just as well be called "Railroad network". Yeah, nicer this way in german - I can imagine it may be the other way round in other languages. If so - see before.

#### Features:
- As suggested renames the tech, and not the improvement, from "Railroad" to "Railroads". Special cases see above.
- Eliminates _all_ hardcoding of Road and Railroad (the tile improvements). Everything now controlled by the RoadStatus enum, including effect. Some preparation to make adding entries easier. This _is_ a step in the direction of full moddability, even though the enum might disappear along the way.
- Makes german translation "Eisenbahn" (for tech) and "Schienen" (for the tracks).
- Translates older save games to the new tech name (unless playing mods that bring the old tech or remove the new).

image


SomeTroglodyte:
"Connection timed out downloading gradle" - so how the `<size=Int.MAX_VALUE><expletive/></size>` do I get this ~~moron~~ intelligent machine to retry?

core/src/com/unciv/logic/civilization/TechManager.kt
```diff
@@ -358,14 +373,15 @@ class TechManager {
         researchedTechnologies.addAll(techsResearched.map { getRuleset().technologies[it]!! })
         researchedTechUniques.addAll(researchedTechnologies.asSequence().flatMap { it.uniqueObjects.asSequence() })
         updateTransientBooleans()
+        movementSpeedOnRoads = if (civInfo.hasUnique("Improves movement speed on roads"))
```
yairm210, line 376: This should be in updateTransientBooleans, not in setTransients

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -490,8 +491,9 @@ class MapUnit {
                     tile.terrainFeatures.remove(removedFeatureName)
                 }
             }
-            tile.improvementInProgress == "Road" -> tile.roadStatus = RoadStatus.Road
-            tile.improvementInProgress == "Railroad" -> tile.roadStatus = RoadStatus.Railroad
+            //todo: is it possible worldscreen-visible stats change by connecting stuff?
```
yairm210, line 494: Yes, connecting cities by railroad for example adds a whopping +25% to production, which affects construction times

SomeTroglodyte: :+1: - of course, I've even seen the code lately. But not this PR, so a glaring yellow todo is just fine methinks

core/src/com/unciv/logic/civilization/TechManager.kt
SomeTroglodyte: ... where I had it initially until I thought "wait a minute, it's no longer a Boolean, so..." I think I checked they're 1:1 at the moment anyway. Will look more closely later.

SomeTroglodyte: No I was wrong, my placement would have omitted an update on addTechnology() where it _is_ needed.

core/src/com/unciv/logic/map/MapUnit.kt
SomeTroglodyte: No, I removed the comment as this is in next-turn code, not interactive? Not quite scope of this PR anyway.

SomeTroglodyte:
***Important***: Will certainly conflict with #4607 - decide just one and let me merge if in doubt

Last commit due to forgotten atlas (renamed that tech png), and so some linting slipped in. Should be ready now.

```
    // Deprecated since 3.15.4
        ConstructRoad("Construct road"),
```
I think the rest of the deprecation is already removed - except in UnitActionType and UnitActionsTable - To be checked later, after this and #4607 are decided.

Found a major regression while testing (ImprovementPicker allows everything), but master shows that too, so I'll treat that separately.

core/src/com/unciv/logic/civilization/TechManager.kt
```diff
@@ -109,8 +108,23 @@ class TechManager {
...
+        }
+        val index = techsToResearch.indexOf(oldTechName)
+        if (index >= 0) {
+            techsToResearch[index] = newTechName
+        }
+        if (oldTechName in techsInProgress) {
+            techsInProgress[newTechName] = researchOfTech(oldTechName)
+            techsInProgress.remove(oldTechName)
+        }
     }
```
yairm210, line 128: Please put ALL state-changing-as-a-result-of-update code in GameInfo.kt (or alternatively we could create a new file that will deal just with that)

SomeTroglodyte: > Dedicated update compatibility module

***Excellent*** idea!

`<mumbling>`... and move all existing ... would be a branch of its own ...`</mumbling>`

yairm210: You're right, that's for a different PR :)
For now just move this to GameInfo

SomeTroglodyte:
Hope you don't mind I made it into an Extension function within class scope - which I'd normally dislike very much, but - quick & dirty and we don't want to keep it there for long.

PR #4568, opened 2021-07-19
Added missionairy units, which can spread religion and bought with faith

xlenstra:
This PR adds functionality for buying units with faith, and uses this to implement missionary units.
Missionary units are religious units that can spread religion 2 times, and can only be bought with faith in cities in which the majority religion is a major religion.

Added uniques:
- "Religious Unit" - required to set the `religion` parameter of new units
- Can be purchased with Faith [cityFilter]

New cityFilter:
- in all cities in which the majority religion is a major religion

<details><summary>Pictures for UI change:</summary>

When selecting a unit that can _not_ be bought with Faith:
image

When selecting a unit that can be bought with Faith:
image
</details>

I do agree the UI needs some work, showing the 0 production doesn't look that great.


logicminimal:
i suggest adding a faith icon for that missionary in the place of that + which puts the units in the queque, and when it is pressed it would function the same as just attempting to buy it from the faith buy menu

xlenstra:
Problem is that at some later point it might be possible for some units to be both constructed _and_ bought with faith, so replacing the + with the "buy with faith" button is not a working solution. Futhermore, only a handful of things can be bought with faith anyway (at least until I implement the 'may buy land units with faith' follower belief), so removing the button when it is not applicable feels like a way to remove unnecessary clutter.

logicminimal:
> Problem is that at some later point it might be possible for some units to be both constructed _and_ bought with faith, so replacing the + with the "buy with faith" button is not a working solution. Futhermore, only a handful of things can be bought with faith anyway (at least until I implement the 'may buy land units with faith' follower belief), so removing the button when it is not applicable feels like a way to remove unnecessary clutter.

what about implementing my second idea?

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -175,6 +176,9 @@ class BaseUnit : INamed, IConstruction, CivilopediaText() {
     }
 
     override fun canBePurchased() = "Cannot be purchased" !in uniques
+    override fun canBePurchasedWithFaith(cityInfo: CityInfo) =
```
yairm210, line 179: Hm. Maybe IConstruction should be an abstract class instead. I don't like how so many of the functions are identical.

core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
```diff
@@ -432,6 +435,43 @@ class CityConstructionsTable(val cityScreen: CityScreen) : Table(CameraStageBase
 
         return button
     }
+    
+    private fun getBuyWithFaithButton(construction: IConstruction?): TextButton? {
```
yairm210, line 439: So much here is common with the getButButton that I think it can be changed to 'getBuyButton(construction, withFaith:Boolean=false)

SomeTroglodyte:
The button arrangement is already in the back of my mind anyway because in some translations they reach way beyond the edge of the lists... But just switching them on and off should be fine for an initial go. 

And `getBuyButton(,Boolean)`- why not `getBuyButton(,Stat)`? Stat is a fine simple enum. One should just avoid enabling buying with Production or Happiness - all others can be imagined...

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
xlenstra: So I spend some time trying this, and it turns out this is not a trivial task. This is as `Building()` inherits both from `NamedStats()` and `IConstruction()`, which is not allowed as "only one class may appear in a supertype list". `BaseUnit` has a similar problem as it also inherits from `CivilopediaText`. I'll try to find a workaround tomorrow, otherwise we will just have to deal with the large amount of code duplication between these two classes

SomeTroglodyte: `CivilopediaText` is just convenience, you can always replace it with `ICivilopediaText` and add the one val manually...

SomeTroglodyte: Besides, isn't `IConstruction` already `abstract`? And if you want functions common to Building and BaseUnit, functions with implementations are allowed in interfaces, and can be left alone, reused or replaced in subclasses. You want `fun canBePurchased() = "Cannot be purchased" !in uniques` in both classes but not in `PerpetualConstruction`? No problem - intermediate interface.

<details><summary>...</summary>

```kotlin
interface IConstruction : INamed {
    fun getProductionCost(civInfo: CivilizationInfo): Int
    fun getGoldCost(civInfo: CivilizationInfo): Int
    fun isBuildable(cityConstructions: CityConstructions): Boolean
    fun shouldBeDisplayed(cityConstructions: CityConstructions): Boolean
    fun postBuildEvent(cityConstructions: CityConstructions, wasBought: Boolean = false): Boolean  // Yes I'm hilarious.
    fun getResourceRequirements(): HashMap<String,Int>
    fun canBePurchased(): Boolean
}

interface INonPerpetualConstruction: IConstruction {
    override fun canBePurchased(): Boolean = "Cannot be purchased" !in uniques

    var uniques: HashSet<String>
}

open class PerpetualConstruction(override var name: String, val description: String) : IConstruction {
//(...)

class BaseUnit : INamed, INonPerpetualConstruction, CivilopediaText() {
//(... drop canBePurchased in this class and make uniques an override ...)

//etc
```
</details>
... actually runs.

SomeTroglodyte:
Holy maloney, you really went for it 😃😎 wow...

xlenstra:
Last commit was quite large, so I'll write up a new description of this PR:
- Enabled buying things with almost all stats imaginable: Gold, Faith, Culture (from stored culture for next policy), Science (from stored science for next tech) and Food (from stored food for next population).
- The Gold button always shows (except for perpetual constructions), the other only if you can actually buy with it.
- Split the class `IConstruction` into `IConstruction` and `INonPerpetualConstruction`. The second class also contains a lot of functions that were virtually identical between `Building` and `BaseUnit`, to avoid code duplication.
- Removed functions from `PerpetualConstruction` that should never be called, and instead cast things to `INonPerpetualConstruction` when these functions are necessary.

To aid in the new 'buy with whatever resource you want with', the following uniques were added, half of which are currently unused:
- "Can be bought with [Stat]"
- "Can be bought with [Stat] [cityFilter]"
- "Can be bought for [amount] [Stat]"
- "Can be bought for [amount] [Stat] [cityFilter]"
- "[Stat] cost of purchasing items in cities [signedAmount]%" --  I'm not sure if `items` is the right term, and an open for improvements.
- "[Stat] cost of purchasing [buildingFilter] buildings [signedAmount]%"
- "[Stat] cost of purchasing [unitFilter] units -[signedAmount]%"

It deprecates the following uniques:
- "Cost of purchasing items in cities reduced by [amount]%"
- "Gold cost of purchasing [unitFilter] units -[amount]%"
- "Cost of purchasing [buildingFilter] buildings reduced by [amount]%"

In other news, I'll be on a week-long holiday starting tomorrow, so you'll have to do without my constant barrage of large PR's :)

SomeTroglodyte:
I love the trailing spaces... Thanks to them
- learned how to run the unit tests locally and to do it every time translations are touched (I wish I did)
- tons of fun for those entries breaking unless their leading blanks or double trailing blanks are religiously preserved
- hours of searching for the setting in Studio to turn off its insubordination only to find it doesn't help reliably

> "[Stat] cost of purchasing [unitFilter] units -[signedAmount]%"

I guess that's a mistype here, not both "-" and "signedAmount" in the codebase... Oops, it _is_ a hardcoded sign.

And try to forget Unciv during that week! You've earned it.

SomeTroglodyte:
Thanks unnecessary - "_teamwork_" - and that I don't really pay enough attention is proven by #4623

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -298,7 +303,7 @@ class CityConstructions {
 
     fun addProductionPoints(productionToAdd: Int) {
         if (!inProgressConstructions.containsKey(currentConstructionFromQueue))
-            inProgressConstructions[currentConstructionFromQueue] = 0
+            inProgressConstructions[currentConstructionFromQueue] = 0 // Should this be `productionToAdd`?
```
yairm210, line 306: No, because we add that in the last line. I'd rather have 1 line for that, rather than if/else

core/src/com/unciv/logic/city/IConstruction.kt
```diff
@@ -1,23 +1,99 @@
...
+                .any()
+        ) return true
+        if (getMatchingUniques("Can be purchased for [] []")
+                .filter { it.params[1] == stat.name}
+                .any()
+        ) return true
+        if (getMatchingUniques("Can be purchased for [] [] []")
+                .filter { it.params[1] == stat.name && cityInfo.matchesFilter(it.params[2]) }
+                .any()
+        ) return true
```
yairm210, line 49: instead of .filter{...}.any() you can just .any{...}
Also, it's hard to tell at a glance what each parameter is, so please give examples / decorate with e.g.
// "Can be purchased for [amount] [statName] [cityFilter]"

core/src/com/unciv/logic/city/IConstruction.kt
```diff
@@ -1,23 +1,99 @@
...
+        return (30.0 * getProductionCost(civInfo)).pow(0.75) * (1 + hurryCostModifier / 100f)
+    }
+    
+    // I can't make this function protected or private :(
+    fun getBaseBuyCost(cityInfo: CityInfo, stat: Stat): Double? {
+        if (stat == Stat.Gold) return getBaseGoldCost(cityInfo.civInfo)
+
+        val cost = getMatchingUniques("Can be purchased for [] []")
+            .filter { it.params[1] == stat.name }
+            .minByOrNull { it.params[0].toInt() }
```
yairm210, line 69: Just !! this, to make it Int instead of Int?, and your problems will disappear :)

core/src/com/unciv/logic/city/IConstruction.kt
```diff
@@ -1,23 +1,99 @@
...
+    }
+    
+    // I can't make this function protected or private :(
+    fun getBaseGoldCost(civInfo: CivilizationInfo): Double {
+        // https://forums.civfanatics.com/threads/rush-buying-formula.393892/
+        return (30.0 * getProductionCost(civInfo)).pow(0.75) * (1 + hurryCostModifier / 100f)
+    }
+    
+    // I can't make this function protected or private :(
+    fun getBaseBuyCost(cityInfo: CityInfo, stat: Stat): Double? {
```
yairm210, line 64: Why double, though? I cannot see a single instance in which this won't be an Int

core/src/com/unciv/logic/city/IConstruction.kt
yairm210: ALSO
we can just reduce these to 2 uniques - "Can be purchased for [amount] [statName] [cityFilter]", "Can be purchased with [statName] [cityFilter]"
If it's in all cities then cityFilter will be "in all cities"
Much easier than maintaining 4 uniques for this

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -345,22 +343,34 @@ class Building : NamedStats(), IConstruction, ICivilopediaText {
...
+            if (stat.name == unique.params[0])
+                cost *= 1 + (unique.params[1].toFloat() / 100)
+
+        for (unique in cityInfo.getMatchingUniques("[] cost of purchasing [] buildings []%")) {
+            if (stat.name == unique.params[0] && matchesFilter(unique.params[1]))
+                cost *= 1 + (unique.params[2].toFloat() / 100)
         }
 
-        return (cost / 10).toInt() * 10
+        return (cost / 10f).toInt() * 10
```
yairm210, line 371: No need, cost is already Float

core/src/com/unciv/logic/city/IConstruction.kt
xlenstra: I don't think I can do that, as `emptySequence().minByOrNull { [...] } == null`, so that would result in a nullptrexception when it construction has no such uniques, right?

core/src/com/unciv/logic/city/IConstruction.kt
xlenstra: Mostly because we will later multiply it with floats (0.65 when 35% discount), and should only round at the end

core/src/com/unciv/models/ruleset/Building.kt
xlenstra: I have a slight preference of doing float arithmetic with float constants, even though the generated bytecode will probably be the same. If you want I'll reverse this

SomeTroglodyte:
By the way - working on TranslationFileWriter (accelerated by ~5x so far, but that's not the goal) - BeliefTypes `Follower =` and `Founder =` are no longer picked up due to changing the type to an enum - my bad.

I can have the code pick the enum names up, those two would then be back, but a bunch of other enum instances that are not user-visible would pop up - ideas so far: Bite the apple and officially demand enum instances needing translation to always need a template.properties entry - or we use the 'pick up enums' code and invent an annotation they would need to have to be picked up - that would be once per field using and enum type... definitely doable... undecided.

core/src/com/unciv/models/ruleset/Building.kt
SomeTroglodyte: me too - due to too much SQL, where knowing where implicit conversions will tip your query plan over some edge by not allowing some statistic to be applied and thus provoking wrong cardinality assumptions is a very useful skill. This would here translate to "if you know that will be `Float.div` then feed it nothing but `Float`'s"...

Matter of taste really, I'd tend to "think SQL" in new code but leave old implicit conversions alone if it shortens the diff...

yairm210:
Hm
I thought I could solve the conflicts via the UI but apparently I made a wrong turn somewhere

xlenstra:
I'll see whether I can fix them

xlenstra:
Should be fixed now

SomeTroglodyte:
Missionary visible in Civilopedia when Relidschion switch is off: Extra filter for "Religious Unit" unique in pedia code or extra unique in json?

logicminimal:
> Missionary visible in Civilopedia when Relidschion switch is off: Extra filter for "Religious Unit" unique in pedia code or extra unique in json?

id suggest adding religious civilian and religious military uniques

freddyhayward:
Are they supposed to be able to spread religion yet? I can't see a button for it.

xlenstra:
> Are they supposed to be able to spread religion yet? I can't see a button for it.

Which version are you on? They should be able to spread their religion to either your own cities (if you are on 3.16.2 or lower) or all cities (if you are on 3.16.3 or higher). The button should show up when you are in the territory of the city you want to spread religion to

freddyhayward:
I'm on 3.16.3.
image
image

freddyhayward:
This check fails, and I can't find anything in the code that might initialise that "Religion Spread" key for missionaries. 
image

xlenstra:
Jup, that is indeed the problem. I pushed a fix in #4851 which should be implemented next version

PR #4573, opened 2021-07-19
City construction Civilopedia-linked

SomeTroglodyte:
This makes the box on the lower right of the city screen which shows a selected construction link to the corresponding Civilopedia entry. May be useful to get more info.
On the minus side, you can't get back from Civilopedia directly to where you were before, you'd have to go back to the city screen manually. When we redesign `UncivGame.Current.setScreen` to keep a stack for back navigation this would feel nicer... :)

Also fixes an oversight in Units Civilopedia which made the uniqueTo lines untranslated, and applies the CityInfo.getRuleset() shortcut where I should have seen it before. If it's there it should be used (or rethought).


SomeTroglodyte:
Pix are not that useful here - I didn't want a visual hint for that link. _**If we wanted**_, something like this would now be a one-liner, but it's not part of this PR:

https://user-images.githubusercontent.com/63000004/126359324-9e9d2926-be1d-4d7b-84c6-bd2c36b32ff8.mp4

The small fix could be shown:
image

:smiley_cat:

PR #4575, opened 2021-07-19
Added more translations

DuffLagerleff:
Not gonna add civilopedia texts yet, but here's what is already added:
* Promotions' effects
* Currently implemented Pantheon and Follower beliefs
* Religion-related actions' text
* Religions names (Would be nice if those were customisable as in civ5)
* Natural Wonder separately since One with Nature has it so as the unique

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -572,6 +572,7 @@ Happiness =
 Production = 
 Culture = 
 Food = 
+Faith = 
 
 Crop Yield = 
 Territory = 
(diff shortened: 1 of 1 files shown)


xlenstra:
Shouldn't these be added automatically when generating the translation files?

DuffLagerleff:
Actually it wasn't. And therefore I've had an experience of those exact keys (+civilopedia texts) in Ukrainian.properties being deleted.

xlenstra:
I'll look into why it isn't added, as those should be added automatically.

PR #4580, opened 2021-07-20
Hide lesser used action buttons to free up space

xlenstra:
Lately a lot of new unit actions have been added, and these have been taking up valuable screen space.
This PR fixes this, by moving rarely used unit actions to a submenu, opened as another unit action.
It also fixes a few minor bugs:
- The "Fortifying (20%)" text had no symbol
- The hotkey for "Fortify until healed" and "Sleep until healed" was the same as "Fortify" and "Sleep", respectively, and pressing it would always choose the action that stayed after healing, effectively making the hotkey useless.

The following actions are hidden by default:
- Automate (but not "stop automation")
- Explore
- Gift Unit
- Disband
- Fortify, when "Fortify until healed" is also an option
- Sleep, when "Sleep until healed" is also an option
- Swap units

<details><summary> Pictures: </summary>

Default action table:
image

After pressing "Show More":
image


yairm210:
"Back" should already be a translated word, you can use that

SomeTroglodyte:
:+1: I like, but there's a few things wrong - three actions mapped to "M" at the same time? are a lot of the icons black on dark blue for you, too or is my master branch botched? Empty tooltips are my mistake, as was "F"...
 
FYI - I got a branch here I'm too tired to finish right now. Enables all keys eligible for unit actions and makes more/back pageup/pagedown. Removes `UnitIconAndKey` entirely, bad idea, sorry. Not touching 'automate' and 'stop automation' present at same time - seems trivial, but there must have been some reason? Now investigating icon Color...

xlenstra:
> +1 I like, but there's a few things wrong - three actions mapped to "M" at the same time? are a lot of the icons black on dark blue for you, too or is my master branch botched? Empty tooltips are my mistake, as was "F"...

Tbh I forgot I added 'm' as a 'show more' button, feel free to remove that or rebind it to another button.
I did mess up the color of the arrows, I'll create a PR only touching those arrows as to avoid potential merge conflicts with your branch.

> FYI - I got a branch here I'm too tired to finish right now. Enables all keys eligible for unit actions and makes more/back pageup/pagedown. Removes `UnitIconAndKey` entirely, bad idea, sorry. Not touching 'automate' and 'stop automation' present at same time - seems trivial, but there must have been some reason? Now investigating icon Color...

The automate should disappear when 'stop automate' is shown, and it looks like I also put the 'automate' action in the wrong section - it should be in the extra actions. Feel free to change this in your branch, otherwise I'll create a PR for it when yours is merged.

SomeTroglodyte:
OK.. what I stopped at are the icon colours, too, and some consensus would be nice - black on blue is OK, black on the disabled dark blue ones recognizable on my phone but drowned on my desktop. There's others not black but white, not only the new more/less arrows... And we should try to standardize on a colour at least for new graphics, namely the one where Actor.color works directly, the other one needs tint or lerp or somesuch to recolour, I forget which was which.

Summary.. I think the unit actions need more contrast and more consistency, but no idea yet how I could imagine that. Not finished testing to compare all cases either. I think I need to cut the colour aspect from that WIP branch - and separate it in two. Later.

PR #4581, opened 2021-07-20
Deprecate "Can only be built on coastal tiles" unique

### LINKED ISSUE #4563: Unique "Can only be built on Coastal tiles" deprecation?
***Is*** that unique deprecated in favour of "Can only be built on [tileFilter] tiles"?

If yes:
- `TileImprovementConstructionTests.coastalImprovementsCanBeBuilt` - add `[]`
- `TileInfo.canImprovementBeBuiltHere` - remove line or comment as deprecated
- Moai definition - add `[]`
- template.properties - remove

If no: should it be?

SomeTroglodyte:
Would resolve #4563 - ***BUT***

#### Not so easy...
- "Can only be built on [tileFilter] tiles" was built as negative test, while "Can only be built on Coastal tiles" is a positive one, the "only" meaningful not because the test had some exclusivity but because Moai has no terrainsCanBeBuiltOn.
- "Cannot be built on bonus resource" is structurally _very similar_ and one should be able to replace "Cannot be built on [Bonus resource] tiles", except that tileFilter fails in this case because it depends on knowing the civ, and `canImprovementBeBuiltHere` doesn't pass one, because it's lost on the way from `canBuildImprovement`...
- Translations would in both cases very likely sound noticeably worse.

#### Questions
- The issue that canBuildImprovement knows the civ, doesn't pass it to canImprovementBeBuiltHere but reduces it to a resourceIsVisible flag, but matchesTerrainFilter would work better if it had the civ? Radical cure: Change canImprovementBeBuiltHere signature; or: Enable resources to be recognized by matchesTerrainFilter without a civ for all resources not requiring an enabling tech; - or both?
- Is my reworking "Can only be built on [tileFilter] tiles" good or bad? It is now an enabler(all) instead of a disabler(none), meaning if several of the same unique are set, they must all match. Because - `or` is already covered by `terrainsCanBeBuiltOn`, and you could do improvements that are only allowed on Tundra+Hill+Forest this way.
- Keep this work on "Cannot be built on bonus resource" or decide against?
- Ideas on the negative impact on translations?

#### Wiki
Is outdated a little anyway, I'd bring the specific paragraph up to date after we're done with this PR - one way or another.


yairm210:
Looks good to me. All in all, more generic uniques, and less specific ones, fully alignes with our goals.

SomeTroglodyte:
@Mape6  - any idea on how the negative impact on translation might be softened?

`Can only be built on Coastal tiles = Kann nur entlang der Küstenlinie gebaut werden`
would become
`Can only be built on [tileFilter] tiles = Kann nur auf [tileFilter]-Feldern gebaut werden`
with `Coastal = Küsten`, and
`Cannot be built on bonus resource = Darf nicht auf Bonus-Ressourcen gebaut werden`
would become
`Cannot be built on [tileFilter] tiles = Kann nicht auf [tileFilter]-Feldern gebaut werden`
with `Bonus resource = Bonus-ressource`

yairm210:
Looks mergable to me, are you ready to undraft this?

SomeTroglodyte:
Yea just giving Martin a few hours. But reacting to new inspiration later is OK, too.

Mape6:
In other translations I never used the dash (-)  before "Feld(er)". 
So the first one would be:
`Can only be built on [tileFilter] tiles = Kann nur auf [tileFilter] Feldern gebaut`

Unfortunately I never found a better way to resolve that in the german language.

The second one would be:
`Cannot be built on [tileFilter] tiles = Kann nicht auf [tileFilter] Feldern gebaut werden`
with `Bonus resource = Bonusressource`

It is the only appearance of `bonus resource` so I would follow the translation of `luxury resource = Luxusressource` then.

SomeTroglodyte:
Damn, I bungled this. I said...

> Is my reworking "Can only be built on [tileFilter] tiles" good or bad? It is now an enabler(all) instead of a disabler(none), meaning if several of the same unique are set, they must all match.

But.. `all(predicate)` is true for an empty collection. `.any() && .all(predicate)` is a lot of overhead, testing `.none(!predicate)` now.

PR #4582, opened 2021-07-20
Fixed bug where all great people suddenly were scientists

xlenstra:
Fixes bug where all great people were great scientists


logicminimal:
how did this even get into the code?

PR #4585, opened 2021-07-20
Implemented temples

xlenstra:
This PR implements temples, but does not implement the buildings that replace temples yet.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -265,8 +265,6 @@
 		"quote": "'Regard your soldiers as your children, and they will follow you into the deepest valleys; look on them as your own beloved sons, and they will stand by you even unto death.'  - Sun Tzu"
 	},
 	// Column 4
-	/*
-	// Temple and its replacements are religious buildings in G&K
 	{
 		"name": "Temple",
 		"faith": 2,
@@ -274,9 +272,8 @@
 		"maintenance": 2,
 		"hurryCostModifier": 25,
 		"requiredTech": "Philosophy",
-		"uniques": ""
+		"uniques": ["Hidden when religion is disabled"]
 	},
-	*/
 	{
 		"name": "National College",
 		"cost": 125,
@@ -324,6 +321,11 @@
 		// temporarily replaces Drama and Poetry and requires a Monument while having its Vanilla
 		// stats, so it's a mish-mash of vanilla and G&K. It should be changed when faith is
 		// implemented.
+
+		// Further edit: Temples are now implemented. However, due to Religion still being in 
+		// its alpha stages, these buildings are still the culture variant. When religion is finally
+		// fully released, this building should be replaced with amphitheatres in existing save files,
+		// and then later the G&K faith version can be added
 	},
 	{
 		"name": "Mud Pyramid Mosque",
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
My you're busy. And delivering quality :+1:

Say - in case you haven't seen - I experimented with a civilopedia page for beliefs & stuff. At first only to get a better idea on how and effort and usefulness. Would that be welcome? Interested in helping out/ supplying input? 

If yes - (in no particular order):
- Belief types _really_ would be better as an enum, and it would look nicer if they each had an icon. I could... but this is your turf.
- Religion as in the ruleset as just a HashSet<String> not backed by a class would make a pedia category for them harder to do since all other 'things' match LinkedHashMap<INamed> or so... But then, there wouldn't be anything to display really.
- Displaying the in-game Religion status as 'pedia page would be easy but would somehow violate the 'pedia pattern - undecided. Get the flavour stories in instead (like "Followed by more than 350 million people throughout the world today, Buddhism is based on the teachings of revered spiritual leader Siddhartha Gotama, known as the "Supreme Buddha," meaning "the enlightened one." ")? Or have a single entry at the top of a Belief category simply listing them on one page? (this I coded in my experiment) ...
- Or do an 'Empire Overview' page to look up religions, who's founded which, chosen beliefs and so on? Would the player need to be limited to some details he knows about, others he hasn't learned about? Or is all Religion info always visible to all in Civ5?

xlenstra:
A civilopedia page for each of the beliefs would be really nice to have, if you want to work on that: be my guest.
I am planning in the near future to add a "Religion Overview" screen, which would show the religions other players have chosen and what beliefs they follow in these, as well as some global information (how many cities follow it, holy city, founding civ, etc.). This will probably have some overlap with the civilopedia, although that could also be usefull: clicking the name of a belief opens the civilopedia page of that belief so all uniques can be viewed with links to the used resources/tile types/etc.

I'll probably spend my own time more on implementing the mechanics of religion, with possibly UI if necessary, and based on your previous great PR's I assume you are more than comfortable adding these civilopedia pages on your own :). 

> Belief types should be an Enum

Completely agree, this was something I was planning to add in a future update anyway. 

> [...] it would look nicer if they each had an icon

Do you mean the different beliefs or the different belief types? For the different beliefs there are just to many to supply an icon, for the different belief types that is something I hadn't thought of but sounds like it could make things clearer, especially in the religion founding picker screen. That would require finding more icons though... I'll keep this in mind for a future update

> Religion as just a HashSet not backed by a class would make pedia pages harder

I don't know whether the Religions themselves do really need their own civilopedia page. At some point I'm planning on making religion names customizable, so the only real part of religions still in use are their symbols (and their default names). While you could add a lot of flavour text for these symbols, it feels kinda unnecessary, and as we currently don't have that for other things either, so I don't think this is a problem. And if at some later point we decide to differentiate between the religions so that a civillopedia page for them would make sense, we will probably implement such a class and the civilopedia pages can be created base on those then.

> Displaying the in-game Religion status as civilopedia pages would be easy but violate the civilopedia pattern

In my opinion (feel free to form your own) I would only add a section for all the beliefs, as these are game-state-independent, and not the religions containing these beliefs, instead showing these in the to-be-added religion overview screen. 

> Would the player need to be limited to the details they know about

While this sounds like a fun idea and a reason to discover more of the world, in Civ5 basically all information is publicly known (even beliefs of religions founded by unmet civilizations are visible iirc), so I'm leaning to implementing it like that.

SomeTroglodyte:
I meant only Icons for belief _types_ - these would go on the left side buttons, not in the description. One exists. Icons for the individual beliefs - would be pure maquillage and a lot of work - not that it didn't occur to me, but I couldn't find evidence Civ5 had those, so I forgot about them. Hmmmmmmmmmm - leave to modders? Why not. Oh, forgot: the Religion ImageGetter is one of the few not sporting a size parameter, and on that civilopedia experiment it shows. I could always wrap and scale, but if you have an elegant idea...

_All_ the rest: agreed!

PR #4586, opened 2021-07-20
Update Español.propiedades

ArchDuque-Pancake:
I wanted to comment about if the language names, whether or not should be in the language name, like, Spanish -> Español, German -> Deutsch etc.


yairm210:
No, they should not, because then all kinds of things are liable to break

PR #4587, opened 2021-07-20
Spruced up Civilopedia - phase 6 - uniques

SomeTroglodyte:
#### Feature:
image
image

#### Notes:
- Recognizes any Civilopedia entry name in any unique ***parameter***
- ..If equal to the entire parameter, case-sensitive, meaning Lists are not matched, thus for e.g. Stone Works the _merged_ unique will not link
- Ambiguous names are prioritized
- Maybe I should have made the Civilopedia category per class part of that interface, idea came too late.
- I relented -again- for a few of the Studio inspections... Not the actual loud compiler warnings, though -> separate project


logicminimal:
i have one suggestion, and that suggestion is for the player to be able to directly return to the last page they visited on civilopedia, this is frusturating when you want to check something of a unit and when you press it you need to go through buttons and lists just to get back to where you started. i suppose it wouldnt be difficult to implement?

as an example: Knight>Cavalry>Horses there should be a back button like on web browsers

PR #4588, opened 2021-07-20
Map generation and start locations

SimonCeder:
Adds a couple extra passes to map generation and changes the algorithm for determining start locations.

Specifically adds cellular automata functions to move around mountains and hills, so that mountains appear in lines and hills are more clustered, so it's more similar to Civ V's maps.
The total amount of hills is not changed, the total amount of mountains is usually doubled from the input (and it's capped at double). I'd like some feedback on what is a reasonable amount of mountains, it should be tied in to the world generation options probably.

The spawns are now not random but are based on an algorithm that ranks all available spots and chooses the best. If a civ gets a very bad spawn anyway there is a "consolation prize" function that adds some bonuses to make up for it.

City states get generally worse spawns (since they pick last, they still pick the best available) but are guaranteed one luxury. City states can also spawn closer to each other than to other players.

Since everyone gets at least a decent spawn, the AI will now attempt to settle in place on turn 1.

Two maps with standard mapgen settings. The empty areas are where I was before I disbanded my settler to take the picture.
mapgen
mapgen2

One with height turned all the way up. The algorithm still ensures that the mountains aren't all clumped together.
mapgen3


yairm210:
Please add consideration for mods, where mountains/hills may not exist

yairm210:
Please add consideration for mods, where mountains/hills may not exist

SimonCeder:
> Please add consideration for mods, where mountains/hills may not exist
I added uniques "Occurs in chains at high elevations" and "Occurs in groups around high elevations" that trigger mountain-like and hill-like generation, respectively. Map generation now does not assume that they exist.

SomeTroglodyte:
:+1:! 

> AI will now attempt to settle in place on turn 1

That's bugged me before - while the wiki somewhere says city states always found on turn 0 so you should never see their settler, ours sometimes wandered for an entire era- if too hemmed in.

PR #4597, opened 2021-07-21
City states adjustments

SimonCeder:
Adds extra checks for city states to make them behave more like Civ V
- City states spawn only with one settler regardless of difficulty and era
- City states will not explore more than one hex outside their territory. They will still go after barbarians and enemy civs.
- EDIT: Added factors for slower production, pop growth, and border expansion for city states.
- Default number of city states in a new game is now 8 (the Civ V default; they are now less annoying than before)

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -290,11 +290,11 @@ object GameStarter {
 
             
             if (civ.isCityState()) {
+                // City states should spawn with one settler only irregardless of era and difficulty
                 val startingSettlers = startingUnits.filter { settlerLikeUnits.contains(it) }
-                if (startingSettlers.count() > 1) {
-                    startingUnits = startingUnits.filter { !settlerLikeUnits.contains(it) }.toMutableList()
-                    startingUnits.add(startingSettlers.random())
-                }
+
+                startingUnits.clear()
+                startingUnits.add( startingSettlers.random() )
             }
 
             for (unit in startingUnits) {
--- a/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
@@ -154,7 +154,8 @@ object SpecificUnitAutomation {
             }
         }
 
-        if (unit.getTile().militaryUnit == null) return // Don't move until you're accompanied by a military unit
+        if (unit.getTile().militaryUnit == null     // Don't move until you're accompanied by a military unit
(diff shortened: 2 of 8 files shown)


xlenstra:
Based on a quick skim of the source code for the base game, city states require 1.5 times as much culture for border expansion

logicminimal:
should i open a todo list issue that has missing features for city states?

xlenstra:
> should i open a todo list issue that has missing features for city states?

Sure! I would advise on putting it in #663, but that would save us work on figuring it all out

logicminimal:
> > should i open a todo list issue that has missing features for city states?
> 
> Sure! I would advise on putting it in #663, but that would save us work on figuring it all out

i think #663 is too cluttered, i will open the thread tomorrow

SimonCeder:
> should i open a todo list issue that has missing features for city states?

That would be very helpful!

> Based on a quick skim of the source code for the base game, city states require 1.5 times as much culture for border expansion

So it is! m_iMINOR_CIV_PLOT_CULTURE_COST_MULTIPLIER(150), CvGlobals.cpp. For some reason I never thought to look. Adjusted. Turns out there are also multipliers for growth, production, etc.

ravignir:
> * Default number of city states in a new game is now 8 (the Civ V default; they are now less annoying than before)
>
Yes, but civ5 map is bigger(56×36), unciv is (44x29).
That feels more like 3 civs/6 city states.

logicminimal:
@SimonCeder @xlenstra done! #4602

SimonCeder:
> > * Default number of city states in a new game is now 8 (the Civ V default; they are now less annoying than before)
> 
> Yes, but civ5 map is bigger(56×36), unciv is (44x29).
> That feels more like 3 civs/6 city states.

Actually it should be 5 city states for the same density! 6 sounds good for now.

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -290,11 +290,11 @@ object GameStarter {
 
             
             if (civ.isCityState()) {
+                // City states should spawn with one settler only irregardless of era and difficulty
```
yairm210, line 293: Please don't use the word irregardless

SomeTroglodyte: :rofl: I just had to check this discussion because I saw that creative word in Studio without the spellchecker whining - and - eagle eyes. I actually quite like it.

SomeTroglodyte:
:+1: I _like_ where this ~~is going~~ went. Just one thing comes to mind: It might help future coders to see the quality of research backing some code decisions so they won't go and re-research. So a ` *= 1.5f // from original source code` or similar could be useful someday. I get the current commenting is from before this discussion here - just - food for thought.

PR #4607, opened 2021-07-22
UnitActionType now knows keys, sounds and most icons

SomeTroglodyte:
~~This won't work as intended. It was locally based on #4606 and included that commit, but here the diff shows pre-4606 stuff on the 'before' side... I'll clean up after that PR is decided.~~

Mostly architectural overhaul as that horrible kludge with the after-the-fact approach to determine icons and keys was too much to bear (I seem to remember this was my bad idea, so I'm entitled to use harsh words). Knowledge about everything that is static for some UnitActionType is now contained right there with the type. UnitAction as class can still override some properties, but needs to do so far less.

Visual differences:
- There were two places getting the Shield for Fortify and I didn't quite understand them. I think the result was unsatisfactory, too, so I made it like this:
image
 That's actually linear, and looks much better 'live' - but might depend strongly on the physical screen, gamma settings and so on.
- Made the Show more/Back arrows black like (most of) the rest.


SomeTroglodyte:
Merged forward and retested - no ill effects.

SomeTroglodyte:
Small change: Prophet's 'Spread' action reverts to the pantheon icon if the Prophet does not have a full Religion, and Pantheon and Religion Icons coloured black for consistency. Also, merged forward.

yairm210:
I think the previous method was valid as well, as a more MVVM-type of approach.
IF it objectively lead to a problem - namely, that not all unit actions had keys or images, or something of the sort, that's one thing.
In this case, I think this implementation is minorly better in that when implementing a new action, you need to access less files, it's more self-contained.

SomeTroglodyte:
MVVM? Model-view-view-model? 1050-950? Multivendor virtual machines? :wink: 

> previous method was valid as well

If you're speaking of the architecture before I came along and wanted keyboard support, then yes, I concur, back then it was still readable. But my approach to mapping keys was, in hindsight, "shooting yourself through the back of your knee from the other side of the saloon". As in complicating things instead of simplifying.

I hope the next action addition will profit... And now I can easier analyze how Xander's decision to make "more/less" _one_ action that changes attributes compares with _two_ actions with static properties that have one-or-the-other activation filters... Or I hope understanding why automate/stop/other stop don't behave logically might be easier.

PR #4609, opened 2021-07-22
A Civilopedia category for Religion

SomeTroglodyte:
@xlenstra  - what do you think...

image
image
image

.. the `enum class BeliefType` thingy is quite rudimentary, just to the point it runs, I haven't actively looked for other uses.


xlenstra:
Looks good! I would change the category name to `religions and beliefs`, but that's a minor nitpick.
On the `BeliefType` class; I don't know what else it should be used for, but it being an enum instead of a string is already a good data hiding protection, maybe I come up with something later when implementing more parts of religion.

SomeTroglodyte:
Nitpicks are good! Ahem, in moderate doses... Yep, will definitely change that when I next can switch branches. Oh - which is now.
And the enum is minor future optimization potential plus immediate repository of the names in an as-good-as-const way?

core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
```diff
@@ -283,6 +287,26 @@ class CivilopediaScreen(
...
+        if (!hideReligionItems)
+            categoryToEntries[CivilopediaCategories.Belief] = (
+                ruleset.beliefs.values.asSequence()
+                .map {
+                    CivilopediaEntry(
+                        it.name,
+                        "",
+                        CivilopediaCategories.Belief.getImage?.invoke(it.type.name, imageSize),
+                        (it as? ICivilopediaText).takeUnless { ct -> ct==null || ct.isCivilopediaTextEmpty() },
+                        sortBy = it.type.ordinal
```
yairm210, line 299: Nice!

yairm210:
Looks good!
Obviously there's more to do with converting the beliefType strings to enums, but we'll get to that in due time

core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
SomeTroglodyte: Some day we'll be able to do away with most of the dead or repeated code in there... Someday... In a future far far away...

PR #4617, opened 2021-07-23
Redraw CityConstrunctionTable and ConstrunctionInfoTable

lishaoxia1985:

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -43,7 +43,7 @@ class CityConstructions {
     val inProgressConstructions = HashMap<String, Int>()
     var currentConstructionFromQueue: String
         get() {
-            if (constructionQueue.isEmpty()) return "" else return constructionQueue.first()
+            return if (constructionQueue.isEmpty()) "" else constructionQueue.first()
         }
         set(value) {
             if (constructionQueue.isEmpty()) constructionQueue.add(value) else constructionQueue[0] = value
@@ -374,14 +374,14 @@ class CityConstructions {
             val construction = getConstruction(constructionName)
             // Perpetual constructions should always still be valid (I hope)
             if (construction is PerpetualConstruction) continue
-            
-            val rejectionReason = 
-                (construction as INonPerpetualConstruction).getRejectionReason(this)
 
-            if (rejectionReason.endsWith("lready built")
-                    || rejectionReason.startsWith("Cannot be built with")
-                    || rejectionReason.startsWith("Don't need to build any more")
-                    || rejectionReason.startsWith("Obsolete")
+            val hideInConstructionListReason =
+                (construction as INonPerpetualConstruction).getHideInConstructionListReason(this)
+
+            if (hideInConstructionListReason.endsWith("lready built")
+                    || hideInConstructionListReason.startsWith("Cannot be built with")
+                    || hideInConstructionListReason.startsWith("Don't need to build any more")
+                    || hideInConstructionListReason.startsWith("Obsolete")
             ) {
                 if (construction is Building) {
                     // Production put into wonders gets refunded
@@ -392,7 +392,7 @@ class CityConstructions {
                     }
                 } else if (construction is BaseUnit) {
                     // Production put into upgradable units gets put into upgraded version
-                    if (rejectionReason.startsWith("Obsolete") && construction.upgradesTo != null) {
+                    if (hideInConstructionListReason.startsWith("Obsolete") && construction.upgradesTo != null) {
                         // I'd love to use the '+=' operator but since 'inProgressConstructions[...]' can be null, kotlin doesn't allow me to
                         if (!inProgressConstructions.contains(construction.upgradesTo)) {
                             inProgressConstructions[construction.upgradesTo!!] = getWorkDone(constructionName)
--- a/core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
+++ b/core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
@@ -40,6 +40,7 @@ class BaseUnit : INamed, INonPerpetualConstruction, ICivilopediaText {
     lateinit var unitType: String
     fun getType() = ruleset.unitTypes[unitType]!!
     var requiredTech: String? = null
+    var requiredPolicy: String? = null
     var requiredResource: String? = null
     override var uniques = ArrayList<String>() // Can not be a hashset as that would remove doubles
     override val uniqueObjects: List<Unique> by lazy { uniques.map { Unique(it) } }
@@ -78,12 +79,8 @@ class BaseUnit : INamed, INonPerpetualConstruction, ICivilopediaText {
     }
 
     /** Generate description as multi-line string for Nation description addUniqueUnitsText and CityScreen addSelectedConstructionTable */
-    fun getDescription(): String {
+    fun getDescription(isConstructionInfoTable: Boolean = false): String {
         val lines = mutableListOf<String>()
-        for ((resource, amount) in getResourceRequirements()) {
-            lines += if (amount == 1) "Consumes 1 [$resource]".tr()
-                     else "Consumes [$amount] [$resource]".tr()
-        }
         var strengthLine = ""
         if (strength != 0) {
             strengthLine += "$strength${Fonts.strength}, "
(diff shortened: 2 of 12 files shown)


yairm210:
Roll back the requiredBuildingInAllCities.
Uniques are better.

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -159,7 +159,7 @@ class CityConstructions {
         val turnsToConstruction = turnsToConstruction(constructionName, useStoredProduction)
         val currentProgress = if (useStoredProduction) getWorkDone(constructionName) else 0
         if (currentProgress == 0) return "\n$cost${Fonts.production} $turnsToConstruction${Fonts.turn}"
-        else return "\n$currentProgress/$cost${Fonts.production}\n$turnsToConstruction${Fonts.turn}"
+        else return "\n$currentProgress/$cost${Fonts.production} $turnsToConstruction${Fonts.turn}"
```
yairm210, line 162: This is deliberate, since larger progresses (say, 800/1200) don't fit well on-screen with the turns to construction and turn numbers

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -70,7 +72,8 @@ class BaseUnit : INamed, IConstruction, CivilopediaText() {
         lines += "$strengthLine$movement${Fonts.movement}"
 
         if (replacementTextForUniques != "") lines += replacementTextForUniques
-        else for (unique in uniques)
+        else for (unique in uniques.filterNot { isConstructionInfoTable
+                && (Unique(it).placeholderText == "Requires []" || Unique(it).placeholderText == "Requires at least [] population") })
```
yairm210, line 76: ????

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -30,6 +30,7 @@ class BaseUnit : INamed, IConstruction, CivilopediaText() {
     var interceptRange = 0
     lateinit var unitType: UnitType
     var requiredTech: String? = null
+    var requiredPolicy: String? = null
```
yairm210, line 33: No. This should be by unique, since very few buildings have this. Unlike required tech, which is almost every building.

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: Could `Label.wrap` help? Result would be a little less predictable, however...

SomeTroglodyte: Nah, looking at the screenies, lishaoxia moved the reason why something can't be queued to the bottom right pane so the numbers are all that's left after the name itself, and they should always fit in a line, at least that's the idea - and I do like what I see.

yairm210: The screenshot doesn't show how it looks when something is in progress, the problem was usually with XXX/XXXX progress.
Looks at the University in the queue, imagine it at 800/1200... and now imagine that it's in the middle of the queue and so there are up and down buttons...

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
lishaoxia1985: That means that the description  doesn't show construction require in constructioninfotable. If not, you will find the word like "consume 1 oil" will repeat in the table.

core/src/com/unciv/logic/city/CityConstructions.kt
lishaoxia1985: The line 161 of the code will Occupy 1 line, If we don't edit the line 162 and delete "\n", the original code will Occupy two lines.

lishaoxia1985:
About unique, you should concern these :
1. unique is often used in tech or policy effect calculate. Shorter is better.
2. Don't mix the effect and construction requirement.
That means we should split construction requirement (including terrain, resource and other require) from the unique.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
lishaoxia1985: If you want to make a mod, that will be good. I think splitting construction requirement and building effect is a good idea.

SomeTroglodyte:
"con-strunc-tion" - you know, phnonenticallly this reminds me of a not too common german word "Strunk" - meaning the harder core of some vegetables after you've eaten away all the finer parts.

Not meant as making fun of you or criticism!! Just trying to add a little fun as I think this idea is fine indeed and deserves praise - and fun.

Now back to business:
image

So - no problem with "larger progresses", fits alright (and if I remember correctly, if I had cheated another digit or two, then the lists - both - would just have grown in width). I repeat: cool idea! :heart: :+1: 

> Don't mix the effect and construction requirement

I'd say it like this: uniques are about rarer properties and properties with variable parameters, not specialized to effects - or requirements - but can cover both. And that's a good thing. Pity the name no longer fits well, but that's history.

yairm210:
Disagree completely about the uniques
Uniques are used BOTH for effects AND for requirements, and greatly simplify multiple problems in this regard.
User display, translation, templating, and combinability, are all points in which the current non-uniques still suffer, and indeed many 9for example requiredTech) have unique 'variants' to allow combinations.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
yairm210: I see
Then this should be 'else if(!isConstructionInfoTable) for(unique in uniqueObjects) ...

lishaoxia1985:
@yairm210 OK, I will do as you wish. But I have 2 questions:
1. I create 2 method getHideInQueueRejectionReason and getDisplayInQueueRejectionReason, do you have a better name?
2. The terrain requirement likes "Must be on []""Must not be on []""Must be next to []" should not show in constructioninfotable, but it's a hard work, do you have some advices? That's one of the reason why i want to split The terrain requirement from uniques.
3.when you built the wonder and use the method addunique, you should don't add the construction requirement to the civilization uniques.

lishaoxia1985:
@yairm210 Sorry, I find a problem,  the var **requiredBuildingInAllCities** is used in the more place and **the unique "Requires a [] in all cities"** is only used in the method getRejectReason (original branch).
If we use **the unique "Requires a [] in all cities"** in all the wonder,  we should give the unique "[]" value to the var **requiredBuildingInAllCities** at first.

yairm210:
A. I agree that we can add the "Requires a [] in all cities" unique check to the ruleset checks. This is a direction we've been aiming to go to, but haven't yet started - sounds like a different PR to me

B.  getRejectionReason() can return a class RejectionReason(text:String, buildingShouldBeDisplayedInQueue: Boolean, rejectionReasonShouldBeDisplayed:Boolean=true)? I think that can solve all of the problems at once, and we can add parameters to it later if we see that we need to

lishaoxia1985:
@yairm210 about getRejectionReason(),  it's no very good. because:
1. In sometimes, the building/unit may have more than 1 RejectionReasons
2. We don't need to get all the RejectionReasons word, beside of ConstructionRequirement and RejectionReasons which ends With "lready built" and starts With"Cannot be built with" "Don't need to build any more" "Obsolete" (these RejectionReasons will lead to delete the building in construction queue and civ gets some compensation, but we can get this aim by the other method), The only thing we need to know is the rejection reasons will make the building unbuildable.
So  I create 2 function: 
// These reasons will lead to that the building/unit hides in construction list
getHideInConstructionListReason(cityConstructions: CityConstructions): String
// If all Construction Requirements are true, the building will be buildable
getConstructionRequirement(construction: CityConstructions): HashMap<String, Boolean> = hashMapOf()

SomeTroglodyte:
> getRejectionReason(), it's no very good

I concur with both of you - return object which both indicates enable / show disabled / hide (enum or two or more Boolean), and allows for flexible display of reason(s). Could be the message part is (stays) a simple string and multiple reasons use '\n', or why not make it a List<FormattedLine> to allow the 'reason' function to control having one line red the other blue; or make the display an ExpanderTab (likely toned down in style) so one could even put a list of all cities missing a Library in there...

Postprocessing by any kind of parsing of user-readable text is an anti-pattern, especially so when the source of that text knows the answers we're trying to parse back out.

That return class should allow deconstruction (`val (status, message) = getRejectionReason()`) so make it a data class or implement the component interface ourselves as we do not need the other freebies of data class (copy and can be used as indexer key).

I'd probably give the function a 'usecase' parameter though, verbose: Boolean or similar, and fill out the text part only if it's on, but short-circuit out of the function if it's off and once the logical part of the result is final.

In any case let's separate refactoring that function from UI tuning?

lishaoxia1985:
@SomeTroglodyte @yairm210 
XueR
YueR

yairm210:
The requiredBuildingInAllCities is still not a good change
I think we can do this gradually, start with the "missing resource" check and then expand

yairm210:
Closing this since it's outdated, but if you PR the 'missing resource' part we can merge that pretty quickly :)

PR #4620, opened 2021-07-23
optimizing the spawn placement algorithm

SimonCeder:
So I thought the new spawn placement algorithm is pretty cool, but it's slow. It's only done once per game and premature optimization is the mother of all evils, but this is fairly easy.
Instead of calculating the spawn ratings for each tile for every attempt to place the civs, just calculate it once and store it in a HashMap.

**Does it help?**
It does, especially on large maps. Times are on my PC.
On standard settings, (Medium map, 4 civs, 6 CS):
Time to generate starting places went from 275 to 213 ms

Huge, 9 civs, 18 CS:
Went from 3314 to 1055 ms

Custom 60 radius, 9 civs, 28 CS:
Went from 10056 to 3193 ms

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -16,6 +16,7 @@ import com.unciv.ui.newgamescreen.GameSetupInfo
 import java.util.*
 import kotlin.NoSuchElementException
 import kotlin.collections.ArrayList
+import kotlin.collections.HashMap
 import kotlin.math.max
 
 object GameStarter {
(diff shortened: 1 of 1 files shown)


ravignir:
Anyway, I am 90% sure in civ5, the civ placing algorithm worked before placing resource.

SimonCeder:
It's possible. It does seem like all civs get at least 2 iron in civ V. I'll take a closer look at some point.

yairm210:
This looks like a good fix to the current algorithm, even if the algorithm itself may need tweaking.

PR #4626, opened 2021-07-24
Update template.properties

Smashfanful:
Added new line

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -784,6 +784,7 @@ Cultural victory =
 Conquest victory = 
 Complete all the spaceship parts\n to win! = 
 Complete 5 policy branches\n to win! = 
+Complete 5 policy branches and build\ the Utopia PRoject to win! = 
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
You sure upper/lower case is correct?
```kotlin
        if (culturalVictoryEnabled) myVictoryStatusTable.add("Complete 5 policy branches and build\n the Utopia Project to win!".toLabel())
```
`tr()` is case-sensitive, see golden ages. Also, `\n` is special-cased if I remember correctly... Best found out through testing.

yairm210:
No problem with the \n, but yes problem with the uppercase.

SomeTroglodyte:
But the proposal contains `"\ "` not `"\n"` - nevermind.
Idea: Take translation file writer out of core, make parallel component project like tests. Run via "Run/Debug config" like tests. Give it access to sources. Scan for toLabel & co...... 🤪

yairm210:
Ah, now I see... you're right, of course
I thought you meant 'can \n be part of a translation key'

PR #4522, opened 2021-07-15
Newgame screen overhaul for portrait mode

SomeTroglodyte:
#### Features
- Player nation picker now shows current selection, including scrolling the list to it (I kinda liked the animation so I didn't suppress it). Ever wanted to check the properties of the nation you just finished picking? Was two trips...
- Player nation picker will arrange its halves vertically in portrait.
- New game screen will rearrange its columns vertically in portrait.
- Fiddled map options until the generated/custom select layout mostly follows the rest of the map parameter stuff. Left and right properly aligned, grows with exaggeratedly wide advanced map parameter translations, the works, the dropdown just won't follow the width of the three other dropdowns as the current module structure makes them live in independent columns of separate tables.
- ExpanderTab used appropriately.
- Therefore Base and extension mod lists can finally be visually distinguished.
- Mods sorted alphabetically. For some the order in which they were downloaded may be helpful in finding a specfic one quickly, but not for me.
- Game will remember window size when you only use map editor or the other screens without actually loading or starting a game -> saved time testing!
- The city state count is an UncivSlider to resolve #4463 .
- One of the annoying warnings gone ("Names of the parameter # 1 conflict in the following members of supertypes: 'public abstract fun setRightSideButtonEnabled(boolean: Boolean): Unit defined in com.unciv.ui.newgamescreen.IPreviousScreen, public final fun setRightSideButtonEnabled(bool: Boolean): Unit defined in com.unciv.ui.pickerscreens.PickerScreen'. This may cause problems when calling this function with named arguments.") - by removing `fun setRightSideButtonEnabled` from `IPreviousScreen` - still left the inappropriate name.

#### Flaws
- You might not like the DARK_GRAY separators - I think they make the portrait mode prettier, but they cost a bunch of code, so - your call. Material design would do away with them - but who likes material design anyway?
- Map editor uses of the updated classes as yet untested
- Needs unmerged code 'ExpanderTab persistence' to really feel 'practical'
- It still annoys mightily by forgetting everything except in the rare case where you go there from a running game -> *where* should proper persistence of that be stored? GameSettings?
- Old landscape mode and new portrait mode "feel" markedly different. Okay, so I didn't want to redesign the basic model. I just added to the chaos...
- Some comments I set for myself may not be a help to others in the future

#### Gallery  
<details><summary>...</summary>

image
... not too different
image
Same except Polynesia was already the chosen nation for the player I clicked
image image
</details>

PR #4630, opened 2021-07-24
Alphabetized the order of nations in the json file and therefore new game screen

xlenstra:


SomeTroglodyte:
Result language-specific. Good idea to do it in code instead. Like in... [Newgame-Mods-UI - sort Nations](https://github.com/yairm210/Unciv/pull/4522/commits/ed0cf5818964d5e169779427767b9e1e1650d6a9)?

yairm210:
I agree with @SomeTroglodyte that this should be done in code, for multiple reasons:
- mods
- future-proofing
- translation-sensitive sort

xlenstra:
Even then, it brings order to an otherwise unordered json file, so I don't see it being a negative :)

SomeTroglodyte:
Wasting _your_ precious time not negative? Sorry for not preempting that - should have been obvious after the "sort civilopedia entries" issue and already fiddling with the new game screen - which was already underway as you saw.

PR #4631, opened 2021-07-24
TranslationFileWriter support for CivilopediaText

SomeTroglodyte:
Now that some "FormattedLine" instances created from json have begun to drop in - maybe you noticed I haven't (I hope!) included those in `templates.properties` - it's time they became available to our translators.

### Features
- TranslationFileWriter is a few times faster.
- `ICivilopediaText.civilopediaText` from json is properly processed, including field exclusions.
- Belief types used in Belief.json are back

### Details
- taught the writer to read and enumerate kotlin `List` as converting `ICivilopediaText.civilopediaText` to java ArrayList was too cumbersome
- static fields ignored w/o need for `untranslatableFieldSet`
- enum typed fields enabled if their type name is listed in a new set `translatableEnumsSet`
- class `CivilopediaText`  gone - had only two uses and was for convenience anyway, `ICivilopediaText` implemented directly instead.

- Tested by comparing german and ensuring all diffs are as intended.
- Binding the translation generator button to F12 was done to test quicker, I can take that out if you want.


SomeTroglodyte:
Note - the magnitude of the performance gain is not through cleverness on my part, but by inadvertently reducing the number of vanilla ruleset clone() calls. `getBaseRuleset` was done a bazillion times and that function clones. That's expensive and needs a thorough check, as of its 19 call sites I can only identify one so far that wouldn't be fine with a direct reference... Can't we make an entire RuleSet immutable? That would be somthing...

yairm210:
Looks good
There are some conflicts though

SomeTroglodyte:
Fighting with the AS "Arctic Fox" update, may take an hour or several before I can merge.

SomeTroglodyte:
By the way, Android Studio "Arctic Fox" has a bunch of new problems, but I'll keep it anyway. JUnit gone is easily fixed, and e.g. the formatter sometimes completely derailing just is incentive to split up too-complex source files (and can be cleared by switching branches)...

PR #4634, opened 2021-07-25
Minimap Slider UI

### LINKED ISSUE #4611: Mini-map Rework
I think we all know about this but this hasn´t benn adressed yet.

The current minimap is at "medium" size when the slidder is at the maximum reach. This, at first glance, isn't a problem, but, there is 2 main problems:
- 1°: To new players the mini-map is super mini, making those new players confused.
- 2°: When making screen shots on the mini-map, it is not big enough for a decent quality. I remerber that one of the purposes for an expandable minimap was for to be bigger in one of it stages.

So, I propose that in the last two "stages" of the slidder, those two would be a little bigger than usual. That way, the default size of the map would be bigger, this way, *possibly*, solving two problems in one.

SomeTroglodyte:
Would resolve #4611 if it pleases.. Visual demo over there, this version here just dropped the 'gargantuan' and so on joke labels.
Also replaces the sound sliders in the settings popup.

The UncivSlider rework looks big but isn't really. Some code movement because I made provisions to set up a slider that plays a sound on every change _without_ the sound playing for the initial setup. That and custom translate-ability of value to tooltip.

### DIFF
--- a/android/assets/jsons/translations/German.properties
+++ b/android/assets/jsons/translations/German.properties
@@ -388,6 +388,7 @@ Auto-assign city production = Automatische Zuordnung der Stadtproduktion
 Auto-build roads = Automatischer Straßenbau
 Automated workers replace improvements = Automatisierte Arbeiter ersetzen Verbesserungen
 Show minimap = Zeige Mini-Map
+off = aus
 Show pixel units = Zeige Pixel Einheiten
 Show pixel improvements = Zeige Pixel Verbesserungen
 Enable nuclear weapons = Aktiviere Atomwaffen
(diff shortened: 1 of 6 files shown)


yairm210:
I'm willing to believe you on the magic numbers :)

SomeTroglodyte:
Well, Belief _is_ now in Unciv! My belief in this PR has a similar basis - 'magic'

logicminimal:
@SomeTroglodyte there is a problem with city state numbers though, in the game creation menu the city state slider only shows amounts if the player has changed the slider value recently while it must be shown at all times

Issue #4602, opened 2021-07-22
City States Todo Thread


logicminimal:
in this thread i will be listing things done and not yet done about the implementation of city states in this game. Please tell me if i missed something or if i mismarked a todo.

- [ ] Religious City-states.
> Also the +%50 more food, faith and culture bonus of the Siamese should be expanded for faith as well.

- [ ] Different rate of City-State bonuses towards friends and allies.

- [ ] Culture from city states shouldn't be provided to any city and should be global, it provides 2 times the amount of culture for allies and normal amount for friends.

- [ ] The food from Maritime City-States, as it says on the wiki, provides food right on the City tile, which makes it subject to bonuses and disadvantages of the status-quo of that city, for instance Social Policies and We Love The King Day.
> Something about generalisation: The food bonus from Maritime City-states should be implemented differently if not already implemented this way. In the wiki it says that the Capital receives an additional +2 food and since every city (not every other city) also gets +1 food the capital gets +3 food, this should be generalised for modding like **[value] food for [capital]** and **[value] food in [all cities]**, the value could also be negative as in, hypothetically a modded City-state could provide some science somewhere and that could lower the linear food yield efficiency (+1, +2 instead of %10, %20) of that city due to population doing more science than farming.

- [ ] Military City-States should provide land units to the city center geographically the closest to it, the time it takes for them to gift units should gets
 lower in terms of turns at allies relationship level.
- [ ] Units gifted by Military City-States should arrive at the city with promotions depending on the xp bonuses a unit trained there receives upon construction (this is what it says on the wiki, i dont know if it is sent with promotions that the City-State decides on or of it is just sent with extra xp just as building an unit with Barracks, need someone with civ5 to test this)
- [ ] As of Gods & Kings expansion the Militaristic City-States has the ability provide unit**s** unique to a major civilization if you have researched the tech that unlocks that unique unit if you were that civilization AND if you are **Allied**. You should be able to hover on the militaristic trait on city state info tab and see the units you can receive. The unique units that a City-State can provide depends on a civilization that it will choose that is not in the game, **those unique units will always be land units** (I think this will cause bugs if the player chooses to have all civilizations in their game) also this unique unit gifting seems to have complex mechanics that i couldn't understand, could someone try to learn that and comment that on here? i didnt understand a word on what happens if a unique unit gets obsoleted.

- [ ] **Allied** Militaristic City-States could provide units from tecs you have not yet researched yet.

[...]

PR #4628, opened 2021-07-24
Civilopedia phase7

SomeTroglodyte:
I originally chose to do all terrain-related categories together, as that was a 4-file diff with only a few 10s of lines. But then I tuned and fixed most of what I saw that wasn't quite perfect, so this got big. Also, there's older PR's that will conflict. Therefore draft, let me do the merge work, otherwise this is stable and polished enough I think.

### Appetizers
<details><summary>Terrain</summary>

image

</details>
<details><summary>Terrain feature</summary>

image
</details>
<details><summary>Resource</summary>

image
</details>
<details><summary>Improvement, translated</summary>

image
</details>

### To-do's
In some cases I left todo's that would improve civilopedia display like:
- Ice would see-also the Submarines if their uniques were [terrainFilter] parameterized
- Fountain of Youth would link Rejuvenation is...
- Sorry for the todo in the Battle.kz nuke routines. The 'pedia looked puzzling so I had to look up in source what those uniques actually do - ahem. I can't imagine a Forest protecting other destructible features on the same tile is correct... 4+ copies of code where 1 would do...

### Translations
- Significant double or trailing blanks were something I could not tolerate, so they're gone from those strings I needed. About half of the existing translations were suffering anyway because the translator missed that extra blank - Spanish, Italian, Dutch, just to name a few. I fixed all translations so they don't get lost.
- Trailing colon - No worse than before, maybe not uniformly solved, might revisit later.

@Mape6:
<details><summary>Roh-Notizen von den Tests teils unbearbeitet:</summary>

```
"Moor trockenlegen / Sumpf" ?
"Fabrik / Fabrik / Manufaktur" ?
"nicht frisches Wasser / ohne..."?
+ "Fügt angrgenzenden"
"Mine -> Bergwerk"?
"Eisenbahn / Schienen entfernen" / "Bahnlinie" in Zukunft?
"Verseucht / Verseuchung / Fallout"?
```
... nur so zum Vormerken und/oder Ignorieren
</details>

### Roadmap
- Nations, Tech, Promotions - code already stashed
- Check TranslationFileWriter for ability to write Beliefs and CivilopediaText
- Leverage ability to link _to_ Tutorials (or Difficulties, though I guess none needed to those)
- **Question**: I think allowing the Barbarians a Nation entry would look nicer - there's at least one link somewhere already that tries and fails to show them - if so, implement and tune that
- Repackage Tutorials into actual CivilopediaText json, beautify a little, and start linking _from_ them where appropriate

PR #4641, opened 2021-07-25
Mercantile CS resources

SimonCeder:
Relates to #4602 
This PR adds Porcelain and Jewelry for Mercantile city states, and a resource overview for the city-state diplomacy screen. They are luxuries created out of thin air in the city state, and are extra incentives to ally them.

So how do you find out which is created in which city-state? With the new resource overview line:
diploresources

Added unique: "Can only be created by Mercantile City-States"


android/assets/jsons/translations/template.properties
```diff
@@ -847,6 +847,7 @@ Obsolete with [obsoleteTech] =
 Occurs on [listOfTerrains] = 
 Placed on [terrainType] = 
 Can be found on  = 
+Can only be created by Mercantile City-States =
```
SomeTroglodyte, line 850: No need, TranslationFileWriter will pick this up.

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -167,9 +168,23 @@ object GameStarter {
         availableCityStatesNames.addAll(ruleset.nations.filter { it.value.isCityState() }.keys
                 .shuffled().sortedByDescending { it in cityStatesWithStartingLocations })
 
+        val unusedMercantileResources = ruleset.tileResources.values.filter { it.unique == "Can only be created by Mercantile City-States" }.toMutableList()
+
         for (cityStateName in availableCityStatesNames.take(newGameParameters.numberOfCityStates)) {
             val civ = CivilizationInfo(cityStateName)
             civ.cityStatePersonality = CityStatePersonality.values().random()
+            if (ruleset.nations[cityStateName]?.cityStateType == CityStateType.Mercantile) {
+                if (!ruleset.tileResources.values.any { it.unique == "Can only be created by Mercantile City-States" }) {
```
SomeTroglodyte, line 177: Maybe cache `unusedMercantileResources.isEmpty()` outside the loop? Two uses inside the loop...

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -22,10 +22,14 @@ class TileResource : NamedStats() {
         stringBuilder.appendLine(this.clone().toString())
         val terrainsCanBeBuiltOnString: ArrayList<String> = arrayListOf()
         terrainsCanBeBuiltOnString.addAll(terrainsCanBeFoundOn.map { it.tr() })
-        stringBuilder.appendLine("Can be found on ".tr() + terrainsCanBeBuiltOnString.joinToString(", "))
-        stringBuilder.appendln()
-        stringBuilder.appendLine("Improved by [$improvement]".tr())
-        stringBuilder.appendLine("{Bonus stats for improvement}: ".tr() + "$improvementStats".tr())
+
+        if (terrainsCanBeBuiltOnString.isNotEmpty()) {
```
SomeTroglodyte, line 26: Conflicts with #4628 - purely FYI. The function is gone in that PR and replaced. I get the reason - I'll adapt #4628 in any case as such checks wouldn't hurt even if this doesn't make it. (actually I have a gut feeling I already had such tuning in before stashing and splitting the 'pedia project - because I tested the thing with some major mods long ago...? Did I miss something?)

SomeTroglodyte (approved):
:+1:

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -175,6 +176,7 @@ class CivilizationInfo {
     fun isCityState(): Boolean = nation.isCityState()
     val cityStateType: CityStateType get() = nation.cityStateType!!
     var cityStatePersonality: CityStatePersonality = CityStatePersonality.Neutral
+    var cityStateResource: TileResource? = null
```
yairm210, line 179: Since this is referencing a resource in the ruleset, the reference should be by name, as a string.

logicminimal:
great work, how are they implemented? are they a tile resource or judt appear out of thin air, they also shouldnt be buildings that happen to provide these as it shouldnt prpvide those luxuries when conquered

SomeTroglodyte:
> how are they implemented

See [files changed](https://github.com/yairm210/Unciv/pull/4641/files) link above. Even if you don't 'speak' kotlin, just looking at the file names and reading the json will often give you a clear idea how.

> "but I doubt great improvements like citadel work"

Great improvements don't "work" for _luxury_ resources anyway, and citadels don't "work" for any resource (at least it shouldn't)... right? That just _might_ make that source a little questionable. I'm not criticizing your research, just saying to take it with a grain of salt.

core/src/com/unciv/logic/GameStarter.kt
SimonCeder: Sorry, I'm not sure what you mean by this? Do you think I'm using an inappropriate data structure here? I wasn't bothered about optimizing this bit too heavily as it only gets called once.

SomeTroglodyte: I just saw like an obvious small optimization that would also improve readability.. But you're right nothing critical here. And I was wrong as well as the second reuse is not the bool 'existence' but the complete list.

android/assets/jsons/translations/template.properties
yairm210: Just this needs to be removed and it's good to merge :)

logicminimal:
> > how are they implemented
> 
> See [files changed](https://github.com/yairm210/Unciv/pull/4641/files) link above. Even if you don't 'speak' kotlin, just looking at the file names and reading the json will often give you a clear idea how.
> 
> > "but I doubt great improvements like citadel work"
> 
> Great improvements don't "work" for _luxury_ resources anyway, and citadels don't "work" for any resource (at least it shouldn't)... right? That just _might_ make that source a little questionable. I'm not criticizing your research, just saying to take it with a grain of salt.

`
        if (civInfo.isCityState() && isCapital() && civInfo.cityStateResource != null) {
            cityResources.add(getRuleset().tileResources[civInfo.cityStateResource]!!, 1, "Mercantile `
City-State")`
yep i think i understood it. now ill see whether the resource monitor shows strategic resources as swell

yairm210:
You seem to have conflicts with another PR that was merged

SomeTroglodyte:
That's likely the `getDescription()` change here and the `getDescription()` removal in Civilopedia phase 7 #4628 .

logicminimal:
@SimonCeder @SomeTroglodyte does anything wrong happen when a city state has negative resources? does the player receive -2 resource as well?

PR #4644, opened 2021-07-26
Update German translation

Mape6:


yairm210:
Conflicts with other German changes

Issue #4656, opened 2021-07-27
Civilopedia Unique parameter autolinking cannot react to ruleset change


SomeTroglodyte:
Another note to self, though anybody is welcome to patch...

As the title says. That auto-mapping is done in `FormattedLine.Companion` using a cache map initialized `by lazy` from the current ruleset. Consequences are:
- Will crash if the engine is used from main menu before any game was loaded (currently not possible to demonstrate)
- If I start with DeCiv then load a vanilla game, the cache will be primed for DeCiv and will miss vanilla names for the rest of the session. This is visible but easily overlooked.

Possible approach: Drop the nice lazy and implement caching more traditionally. Let the 'CivilopediaText' modules receive a notification when the ruleset changes like ImageGetter gets (Or, monitor a ruleset change by comparing hashes on every access - no thanks). Manage cache rebuild using that.

- [x] Find and document a path to demonstrate the problem, screenshot undesired result
- [x] Evaluate alternate approaches
    - [x] `fun invalidateCache` with enough hooks
    - [x] remember identity of active ruleset instance, compare with `===` ???
    - [x] remember some hash of GameParameters.mods and GameParameters.baseRuleset ???
- [x] implement - option 2 seems enough
- [ ] test and screenshot how reproducible symptom is fixed

AdityaMH:
If you want add civilopedia to main menu. My suggestion is just add ruleset mod menu above and refreshed every changing mod selection. Also make it hidden when game was loaded.

SomeTroglodyte:
> civilopedia to main menu

Nope, not what I said. "engine", not the screen.

AdityaMH:
>if the engine is used from main menu

What benefit using "that engine" from main menu?

SomeTroglodyte:
Way to reproduce:
- From main menu, load or start a game with vanilla. no mods, religion enabled
- Go into Civilopedia, look at Great Prophet -> Holy site _is_ linked
![image](https://user-images.githubusercontent.com/63000004/127752297-564250ea-83b8-4857-8f64-84b82818feb1.png)

- Quit Unciv and restart
- From main menu, load or start a game with DeCiv redux
- Go into Civilopedia, Promotions, Accuracy
- Load or start a game with vanilla. no mods, religion enabled
- Go into Civilopedia, look at Great Prophet -> Holy site is _not_ linked
![image](https://user-images.githubusercontent.com/63000004/127752300-6a905e33-9b4e-4280-982f-e679a2e135f9.png)

PR #4657, opened 2021-07-27
Civilopedia phase 8 - Nations and Promotions

SomeTroglodyte:
The usual...

<details><summary>Eye Candy</summary>

image

image

*** Not part of this PR ***
image
</details>

Remarks:
- Nation.kt code duplication is _not_ fine but I need to think about #4656 before resolving that. The last screenshot was done with an experiment that had `Nation.getUniqueString` and its 3 subroutines disabled, so it's not so far off.
- All languages touched only in order to "kill off" the next translateable depending on leading and trailing blanks - and this time the variation in actual blanks in the translations was no better.
- Preparation to move the "first tabstop" (indent=1) because someone keeps asking - will need separate tuning later.
- Displaying 1 movement for air units seemed like it needed to be hidden


yairm210:
The reason for the spaces of that in Chinese and similar languages, there are no spaces between the options.
A way around this could be to create a "[a] or [b]" template, and recursively use it to fold the list into a single string

logicminimal:
those trailing spaces are really confusing especially when doing translations and looking at some text in the game and seeing there are spaces there for no reason

yairm210:
I know, that's why I explained the problem and suggested a workaround :)

SomeTroglodyte:
With a single string for a list of promotions we'd be back to the old way without links? A link per line means a line per linkable item. But I get "OR [alternative] = " as template might be better. Only way to allow the translator to decide he wants no glue text at all, unless he uses [`\u200B`](https://en.wikipedia.org/wiki/Zero-width_space) which would be just as unmaintainable as significant leading or trailing blanks.

SomeTroglodyte:
... But in that case I'd best not presume and better _delete_ those existing translations I can't understand or second-guess. I mean if I fix " OR  = 或" to "OR = 或" without having an inkling of Chinese that's reasonable because if that symbol can mean OR-with-spaces then it must also be appropriate for the word OR alone. But with placeholder I'm implying a certain order and that I cannot predict.

SomeTroglodyte:
And `.tr()` supporting recursion for nested []'s or even nesting [] inside {} - no it doesn't at the moment, those regexes can't count.

SomeTroglodyte:
In that latest proofreading pass I was a bit surprised our `uniques` aren't all the same collection type... Value of interfaces... But that's for another day.

yairm210:
Even the new "or []" doesn't cut it. What do you do with the space before the "or"? That doesn't exist in Chinese either.
What I was suggesting isn't nesting, but rather like:


```kotlin
val list = listOf("first", "second", "third")
var result= list[0]
for(i in 1..list.length) result = "[$result] OR [${list[i]}]"
```

I'm sure there's a way to do this with .fold{} as well.

SomeTroglodyte:
Actually, as I understand it that would be `.joinToString` and be close to the previous Civilopedia - but I see where I explained badly and that there's a big to-do for phase 10 or 11: Make sure all cases structurally similar to this use the same appoach. I think that's on some list anyway.

Clarification:
image

(a) is how I understand your train of thought - remember, one link per line - not wishing to arbitrarily just link only the first one
(b) is the current PR
(c) is an alternative I probably used for some other of the objects, most compact but likely least practical for diverse translateability
(d) is what I _now_ would propose as best solution. Simplest code, too. For lists of size 1 we'd still do (a)/(c).

(And as you see there is no actual space character before the "OR" in (b) or (c), just padding, but there is in (a)... I probably went at it with an unflexible preconception because there is an existing translation for the standalone word "requires")

SomeTroglodyte:
... waaaaaaaaait, are you _not_ talking about the Civilopedia use but about the `getDescription()` for the picker screen? Ouch. I'd better revert that.

yairm210:
D does indeed seem like the best one :)

logicminimal:
another suggestion: to include multiple ranking of a promotion inside just one
like indtead of accuracy I, II, III there is only accuracy on civilopedia

PR #4658, opened 2021-07-28
Implements the Celtic people according to fandom wiki

SomeTroglodyte:
Early draft - seems complete and preliminary tests (turn 1 only) show the interesting details to be OK. Just so you know this is in the pipeline.

TODO
- [x] Test
- [x] Test unique unit's kill bonus
- [x] "An upgraded Pictish Warrior retains its foreign lands and pillage bonuses, but loses the ability to earn Faith"
- [x] Preferred Victory? Fits.
- [x] "Preferred Religion"? - Defer as Religion is far from Compleath
- [x] Edinburgh? Also Dublin. E replaced with Valletta, D removed.
- [ ] Achievement replacement for "Longest. Name. Ever."?
- [x] Game atlas exceeding bounds - Who did The Huns in 400x400???
- [x] Unit Icon not taking civ's inner colour

<details><summary>Screenshotz</summary>

image
image
image
image
image

</details>


logicminimal:
the bonus should be +2 faith bonus per 3 adjacent unimproved forest i think you have already implemented that as seen from the screenshot but it doesnt say so in the civilopedia.

> also i think pillage bonus wont be enough if it still just provides health instead of gold as well but this is my opinion

great work by the way

ravignir:
I did not expect Celts to be implemented before non-religious civs.

logicminimal:
> I did not expect Celts to be implemented before non-religious civs.

¯\_(ツ)_/¯

SomeTroglodyte:
> I did not expect Celts to be implemented before non-religious civs

So which _did_ you expect first?
And I assume the comment means you double-checked my interpretation of the not-always-reliable fandom wiki rules? :wink:

ravignir:
I expected Austria to be implemented next :)

Everything looks good so far.
For the preferred victory, Domination seems to be the most accurate. For the preferred religion,  Christianity.

SomeTroglodyte:
> Austria

I probably liked Boudicca's portrait better than Maria Theresia's  :cowboy_hat_face: 

> preferred Religion

I put that there because the attribute is still absent from the classes, so - defer / provide ignored json / whatever ... Just brainstorm chaff really.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -442,6 +442,8 @@ open class TileInfo {
      * @see <a href="https://github.com/yairm210/Unciv/wiki/uniques#user-content-tilefilter">tileFilter</a>
      */
     fun matchesFilter(filter: String, civInfo: CivilizationInfo? = null): Boolean {
+        if (filter == "unimproved Forest")
```
yairm210, line 445: That settles it... we're going to have to create sub-uniques so we can have "unimproved []" 
Hardcoding both the filter and the Forest value is too anti for me.

SomeTroglodyte:
No concept on how to create such a "sub-unique" system, so I'll wait for one.

logicminimal:
can we just merge this as plain forest tiles and implement unimproved forest part later?

logicminimal:
@SomeTroglodyte i think it will be better if you just release it now and implement unimproved part later

SomeTroglodyte:
No. Merging got too complicated. Discarded, including art.

SomeTroglodyte:
@xlenstra  - as master of `UniqueTarget.Conditional` - how would this Druidic Lore now fit into the system? UniqueType.StatsPerCity, then pack the whole ">= N unimproved Forest tiles in citycenter.neighbors" in one TileCount Conditional? Which has a sub-Conditional for the "and", with "unimproved" and "Forest" each being a simple tileFilter? Sounds way more complicated than having 1 tileFilter "unimproved Forest" which is already marked as undesirable....

xlenstra:
I'd say this still requires the and logic with the {} braces, but similar to CitySpecialists unique it could be written as:
"[+1 Faith] <in cities adjacent to at least [amount] [{unimproved} {forest}] tiles>"

Only question that remains is how we should implement translations for these {} parts

SomeTroglodyte:
I'm testing "[+1 Faith] [in all cities] <with [1] to [2] neighboring [unimproved] [Forest] tiles>"... minor kludge, but one we can replace and deprecate easily? (The N to M is to get the confusion that the second point is cumulative out of the way - that's +2.. 3 to 6 then)

{} and logic exists for units, not going to add another hardcoded one, lesson learned.

xlenstra:
> I'm testing "[+1 Faith] [in all cities] <with [1] to [2] neighboring [unimproved] [Forest] tiles>"... minor kludge, but one we can replace and deprecate easily? (The N to M is to get the confusion that the second point is cumulative out of the way - that's +2.. 3 to 6 then)
> 
> {} and logic exists for units, not going to add another hardcoded one, lesson learned.

The disadvantage of that approach is that it only flows nicely when used with uniques that already include a cityFilter, while having the <in cities> included in the conditional allows for it to be added to all uniques that can apply in/from a specific city and the unique still reading nicely

SomeTroglodyte:
- Whether the city filter goes on the base unique or on the conditional is - all the same?
- As I see it, the goal to allow any combination of base unique and conditional is impossible, and having invalid combinations detected will be challenging. Thus no fear of limited use conditionals in my brain..
- Anyhow, it's all pretty much equal overhead and efficiency to me. If you see a clear balance skew towards a specific solution, go ahead and request a change over on #5367.
:smile_cat:  (can you imagine I'm feeling slightly miffed because a certain exception _refuses_ to happen again? :lady_beetle: )

PR #4664, opened 2021-07-28
CS bonuses graded according to relationship level

SimonCeder:
Relates to #4602 

Edit:
After a couple reviews, now also adds the ability to configure city-state bonuses in eras.json. In case of missing definitions, the Civ V values are hardcoded fallbacks.

City states now give the proper bonuses depending on whether you are friends or allies, adjusted for era. Numbers are the same as in [this](https://www.carlsguides.com/strategy/civilization5/citystates.php), which is accurate for G&K as far as I was able to test.
Siam gets an explicit UP to give +10 XP for city state gifted units, which was a hidden bonus in the original.
There is a line in the diplomacy overview that shows exactly what you'll get. I changed "at 30 influence" to "when friends" since there is a nice string just above that tells you how to get there.
csbonuses

Added unique: "Military Units gifted from City-States start with [] XP"

Note: I'm still unsure in which cases it is necessary to adjust the translations file. I think the Siam unique will get autoadded, but not the new diploscreen strings? In which case I'll add another commit with those.


SimonCeder:
It's certainly possible to move all numbers out into json files. I left the numbers hardcoded as they were that way before. But the ability for getting units faster is already implemented and is not affected by this code.

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -137,11 +137,23 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
...
+        val bonusMultiplier = if (viewingCiv.hasUnique("Food and Culture from Friendly City-States are increased by 50%")) 1.5f else 1f
+
+
         val friendBonusText = when (otherCiv.cityStateType) {
-            CityStateType.Cultured -> ("Provides [" + (3 * (viewingCiv.getEraNumber() + 1)).toString() + "] culture at 30 Influence").tr()
-            CityStateType.Maritime -> "Provides 3 food in capital and 1 food in other cities at 30 Influence".tr()
-            CityStateType.Mercantile -> "Provides 3 happiness at 30 Influence".tr()
-            CityStateType.Militaristic -> "Provides land units every 20 turns at 30 Influence".tr()
+            CityStateType.Cultured -> ("Provides [" + (cultureBonus * bonusMultiplier).toString().removeSuffix(".0") + "] culture when Friends").tr()
+            CityStateType.Maritime -> ("Provides [" + (2f * bonusMultiplier).toInt().toString().removeSuffix(".0") + "] food in capital when Friends").tr()
```
yairm210, line 147: In both cases, if you .toInt() it then you don't need to removeSuffix, since it's a whole number

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -137,11 +137,23 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
             diplomacyTable.add(nextLevelString.toLabel()).row()
         }
 
+        val cultureBonus = if (viewingCiv.getEraNumber() in 0..1) 3f else if (viewingCiv.getEraNumber() in 2..3) 6f else 13f
+        val happinessBonus = if (viewingCiv.getEraNumber() in 0..1) "2" else "3"
```
yairm210, line 141: I really don't like that the amount is based on the number of the era.
This should probably be a unique for each era.

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
yairm210: Not if you .toInt() it. (3.5).toInt().toString() = "3", not "3.0"
What I'm saying, is EITHER .toInt(), OR .removeSuffix(). There's no case where you'll need both.

SimonCeder:
I'll be away for the coming week, so putting this as draft for now.

SomeTroglodyte:
- The idea with eras.json seems good, but that's quite a number of specialized values. Uniques? ... No, that would be too many instances of those, no gain. I'd say keep it as is.
- `.toString().removeSuffix(".0")` - nah. If it's a Float, never rely on being lucky with rounding, you might get ".00000001". Look at serialized MapParameters - the default 0.6 is nicely output as 0.59999996
- Maritime limits mods - of 4 possibilities 2 can be reached
- `val eraInfo = cityInfo.civInfo.gameInfo.ruleSet.eras[cityInfo.civInfo.getEra()]!!` sounds like a `CivInfo.getEraObject()` or so might be useful to shorten those
- Happiness - even if vanilla doesn't do it, maybe make an ally/friend difference moddable, while you're at it?
- `if (statMap.containsKey("City-States"))` - ah yes that `statMap` that isn't a `StatMap` instance irks again... I'd probably go with `statMap[key] = (statMap[key] ?: 0f) + eraInfo.mercantileHappiness` as most efficient of many solutions

All just brainstorming from code inspection, generally no objections, :+1:

yairm210:
Making the bonuses not be hardcoded is great, but these functions should be in uniques :)

Like:

```
        "mercantileHappiness": 2,
        "maritimeCapitalFood": 2,
        "maritimeAllCitiesFood": 1,
        "culturedFriendCulture": 3,
        "culturedAllyCulture": 6,
        "militaristicFriendDelay": 20,
        "militaristicAllyDelay": 17,
```
 ->

```json
  "friendBonus": {
    "Mercantile": ["[+3] Happiness"] // This is a unique that does not yet exist!
    "Maritime": ["[+2 Food] [in capital]"]
    "Militaristic": ["Gain a unit from this civilization every [20] turns"]
    "Cultured": "[+3 Culture] [in all cities]'"
 },
 "allyBonus": {
    "Mercantile": ["[+3] Happiness", "Provides access to a unique Luxury resource],
    "Maritime": ["[+2 Food] [in capital]", "[+1 Food] [in all cities]"]
    "Militaristic": ["Gain a unit from this civilization every [17] turns"]
    "Cultured": "[+6 Culture] [in all cities]'"
  }
```
About unique luxury for Mercantile city-states:
Adding this as a unique to the allyBonus is probably the way to go. Advantage: all Mercantile city-states are the same. Disadvantage - you need to search through bonus uniques to display the unique resource to the user(or to generate this upon settling the city, I don't recall which one it is, there was a discussion on the topic)
I'm mentioning this because there is an alternative way - putting the unique *on the Civ* and not on the bonus - but that could lead to some Mercantile civs providing resources and some not, which is not a great player experience.

The real moddability is the *moddable city-states types* we made along the way :D

yairm210:
I don't like the fact that most of these don't change over time and that there's a lot of copy-pasting, but given that some do,our options are either to stick them in the eras.json or to have era-specific uniques, and the second is probably much worse.

yairm210:
I'm not sure we need a cityStates.json file.
In either case, unique-ifying these is the way to go, no matter which file they end up in in the end :)

If we are going the "not by era but for all eras together", then the json format would still look as above (friendbonus, allybonus) but the friend/ally bonus for certain ones would be split by era, as you said.

FOR NOW let's focus on the initial "convert to uniques, put in eras.json" part, which is already quite a lot 😅

yairm210:
In my suggestion we can add those uniques to any city state type we want, even new ones that we invent.
I don't see how a separate json files helps us here.

As for era specific vs non-specific, this would be an option.

I feel that the sheer number of jsons is getting a bit out of hand, making it daunting for new modders. So if we can handle this without creating a new one, I think that's better :)

SimonCeder:
This is as far as I got from now.
All city state modifiers are read from eras.json and are now more moddable, for example one might change maritime city-states to give extra science as well.

A couple issues:
- It's still not possible to mod in entirely new city states, since the city state types are defined in an enum in kotlin.

- It would have been more elegant to read in the bonuses as `CityStateType: ArrayList<String>` rather than `String: ArrayList<String>` which is what I ended up with. I couldn't get reading in CityStateTypes to work for some reason, it would always scan as null or I would get type exceptions.

- I'm doing a lot of string parsing which is slow. Probably should create some class to hold the parameters analogous to Unique, but not sure how to implement it.

At least the diplomacy view is more elegant (code-wise) this way:
Screenshot_2021-08-12_14-49-29

SomeTroglodyte:
Gdx's json is a bit nasty at times. Tried declaring `var friendBonus = HashMap<CityStateType, ArrayList<String>>()`
and supplying `{class:com.unciv.logic.civilization.CityStateType,value:Mercantile}` instead of "Mercantile" etc in the json?

SimonCeder:
Tried a couple combinations with no luck. I could generate a separate `CityStateType, ArrayList<String>` from the `String, ArrayList<String>` by using `CityStateType.valueOf(key)`?

SomeTroglodyte:
... for moddable jsons, keep the class fields such that a direct deserialization works. Need a different API? Offer as some fun(), possibly cached ... Haven't tried personally, though, so that's cheap advice and not to be taken too seriously. That was just an idea born from the way it _serializes_ some stuff.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -495,11 +498,22 @@ class DiplomacyManager() {
...
+        if (otherCiv().isCityState()) {
+            val eraInfo = civInfo.getEraObject()
+
             if (relationshipLevel() < RelationshipLevel.Friend) {
                 if (hasFlag(DiplomacyFlags.ProvideMilitaryUnit)) removeFlag(DiplomacyFlags.ProvideMilitaryUnit)
             } else {
-                if (!hasFlag(DiplomacyFlags.ProvideMilitaryUnit)) setFlag(DiplomacyFlags.ProvideMilitaryUnit, 20)
+                val relevantBonuses = if (relationshipLevel() == RelationshipLevel.Friend) eraInfo.friendBonus[otherCiv().cityStateType.name]
+                                    else eraInfo.allyBonus[otherCiv().cityStateType.name]
+                if (relevantBonuses != null) {
```
xlenstra, line 509: In my opinion, indentation here goes to deep. Maybe something like
`if (relevantBonuses == null) return`
and later combining the two directly nested `if`-statements would solve that.

xlenstra (commented):
One major thing I'm missing in this PR, is that mods with their own `eras.json` file will not have the `friendBonus` and `allyBonus` uniques, making city-states in those mods give no bonuses whatsoever. This should if possible be avoided, with pieces of code marked as deprecated implementing it for those mods. Otherwise, looks great!

SimonCeder:
Added back in the hardcoded values as fallback when json definitions are missing.
Also, thank you all for your patient and helpful reviews.

yairm210:
We really don't need new "Provides" for values that already exist.
The idea is that getCivUniques() should also return all uniques from city states, and then they're be parsed like all civ-wide uniques, thus providing the moddability we've been talking about.

"Provides [amount] [stat]" IS a unique that we'll need to add, since unlike most uniques that are resolved on the *city* level, this one is resolved on the *civilization* level.

For instance,  "Provides [2] [Food] [in capital]" should be just "[+2 Food] [in capital]" .

BUT, I think we've driven you up a wall enough for one PR... Making CS uniques fully moddable can be round 2 ;)

For now, there are just a couple of things that I'll add since you have conflicts anyway ;)

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -122,13 +124,44 @@ class CityStats {
...
+                        val stattoadd = when (bonus.getPlaceholderParameters()[1]) {
+                            "Food" -> Stat.Food
+                            "Production" -> Stat.Production
+                            "Gold" -> Stat.Gold
+                            "Faith" -> Stat.Faith
+                            "Science" -> Stat.Science
+                            "Culture" -> Stat.Culture
+                            "Happiness" -> Stat.Happiness
+                            else -> null
+                        }
```
yairm210, line 154: Stat.valueOf(), and assume that it's fine. Better to crash if he has a misconfigured unique than to stay silent.

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -80,16 +82,44 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
...
+
+                if (relevantBonuses != null) {
+                    for (bonus in relevantBonuses) {
+                        if (bonus.getPlaceholderText() == "Provides [] [] per turn") {
+                            val stattoadd = when (bonus.getPlaceholderParameters()[1]) {
+                                "Gold" -> Stat.Gold
+                                "Faith" -> Stat.Faith
+                                "Science" -> Stat.Science
+                                "Culture" -> Stat.Culture
+                                else -> null // Can't add food or production since we are adding to empire-wide and not to a city. Happiness implemented elsewhere.
```
yairm210, line 103: Stat.valueOf() again

SimonCeder:
I'm unsure how to fix the conflict. It's whitespace replaced by 4 lines of code. Tried merging the current changes from the master for the same file but that didn't seem to do it.

core/src/com/unciv/logic/city/CityStats.kt
SomeTroglodyte: Um... `Stat.values().firstOrNull { it.name == whatever } ?: default` isn't really worse performance-wise than `try { Stat.valueOf(whatever) } catch (ex: Exception) { default }`, if you ever wonder.

... but shouldn't alarm bells ring when something involving Unique and Stat does not go through `Stats.parse()`?

Would make code shorter, unique more flexible, and get rid of 1 green warning - the spellcheck on the variable name :rofl:

SomeTroglodyte:
As you can see, I tried it locally. There's a whole bunch of other changes that need merging, but the same lines are the only ones marked red. Seems to work in the end though. You know I have my choice of workflow mostly documented, right?

logicminimal:
finally we got this in game

SomeTroglodyte:
@SimonCeder  - now that i playtest - observations:
- Mercantile CS as friend grant no resource, as ally all of them, in the quantity they receive themselves including the unique that's not on the map. Question - is that the correct rules, I haven't followed all discussions well enough, it's probably all there...
- Non-mercantile CS - they too now grant all their resources at ally (correct?)
- That fact (or whatever the facts should be) is not clearly evident on the trade screen. One could do some visual distinction up there where the resources are listed.

Umm - what would you like to improve and what part of that would you like to delegate? I could grey out the icons a little when only friends? Or some textual clarification "this is your incentive" vs "this is what you are already getting"?

logicminimal:
> @SimonCeder  - now that i playtest - observations:
> - Mercantile CS as friend grant no resource, as ally all of them, in the quantity they receive themselves including the unique that's not on the map. Question - is that the correct rules, I haven't followed all discussions well enough, it's probably all there...

i guess

@ravignir should correct me on this but afaik they give amount of any resource they produce without calculating in the usage from units (10 horse, 2 used the plauer receives 10 horse)

> - Non-mercantile CS - they too now grant all their resources at ally (correct?)

correct i guess

> - That fact (or whatever the facts should be) is not clearly evident on the trade screen. One could do some visual distinction up there where the resources are listed.

you cant trade away resources you dont directly produce. ravignir can u tell exactly what happens about %50 happiness preservation upon giving away a luxudy resource for Netherlands?

SimonCeder:
> @SimonCeder - now that i playtest - observations:
> 
>     * Mercantile CS as friend grant no resource, as ally all of them, in the quantity they receive themselves including the unique that's not on the map. Question - is that the correct rules, I haven't followed all discussions well enough, it's probably all there...
> 
>     * Non-mercantile CS - they too now grant all their resources at ally (correct?)
> 
>     * That fact (or whatever the facts should be) is not clearly evident on the trade screen. One could do some visual distinction up there where the resources are listed.
> 
> 
> Umm - what would you like to improve and what part of that would you like to delegate? I could grey out the icons a little when only friends? Or some textual clarification "this is your incentive" vs "this is what you are already getting"?

Unless I'm misunderstanding this is intended behavior, and in fact was not changed in this PR. Allied CS should give you their luxury and strategic resources.

PR #4669, opened 2021-07-28
Map climate overhaul

ravignir:
This should make the map look more like a civ5 map:
here is a spreadsheet i've created and used to make these changes: https://drive.google.com/file/d/18uacHvZlh4FoM8xEVH4J9jaAmtHomutr/view?usp=sharing
What it does:
- makes center of a map mostly Grassland with some bits of Plains
- next biome is Desert and Plains with a rare patches of Grasslands.
- next biome is again Grasslands and Plains (in about equal proportions)
- final biome is Tundra that turns into Snow towards poles.
- makes the whole map look less stripy.

It looks kind of messy, but i tried to make it as compact as i could.

1z

civ5 map (apparently oval map shape is closer to unciv pangaea)
2x

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -21,7 +21,14 @@
 		"food": 2,
 		"movementCost": 1,
 		"RGB": [97,171,58],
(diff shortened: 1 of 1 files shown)


yairm210:
Okay! Sounds like you've put a lot more thought and effort into this than I have, so I'll take this at face value!

PR #4682, opened 2021-07-30
Fix Hagia Sophia and CN Tower not giving civ 5 bonuses

logicminimal:
i got the information about Hagia Sophia from @ravignir and CN Tower from civ wiki. i will also add faith to buildings that give faith in G&K but will do it as a commit, that so that it would be easily revertable if faith doesnt just get ignored when religion is disabled

edit: two things to have in mind if this will merge into the game, i am using the vanilla stats for hagia sophia as we need religion properly implemented for its gnk values
also the cn tower provides a broadcast tower in all of cities of the civilization currently resides in, which means in case of a conquer of that cn tower those broadcast towers would disappear, not happiness and population bonus i think though. but i need to change the game logic to properly implement it which i cant


logicminimal:
I don't know why the checks are failing, i mean it says that it could be merged?  i checked and it is literally the same line as one with leaning tower of pisa (im talking about hagia sophia), what could be causing it? and in my patch 8 fork of unciv the buildings.json literally doesn't exist, so i cant even fix that i need help

ravignir:
>Note that the Broadcast Towers will go to whichever civilization owns the CN Tower. That means that the original builder will lose the free Broadcast Towers if another player captures the city with the CN Tower (and the new owner will gain them), although the increases in 20xPopulation5.png Population and 20xHappiness5.png Happiness will only be rewarded to the builder. The additional Broadcast Towers awarded when this wonder is completed go to all cities controlled by that civilization, even if they have just been captured and are still in resistance. The free Broadcast Tower also goes to any city founded or acquired after the CN Tower has been completed, but, as when the city with the wonder switches hands, there will be no increase in population or happiness.
>
What about this?
Anyway, the happiness bonus only applies as long as you own CN tower, so that's an error on the wiki.

logicminimal:
> >Note that the Broadcast Towers will go to whichever civilization owns the CN Tower. That means that the original builder will lose the free Broadcast Towers if another player captures the city with the CN Tower (and the new owner will gain them), although the increases in 20xPopulation5.png Population and 20xHappiness5.png Happiness will only be rewarded to the builder. The additional Broadcast Towers awarded when this wonder is completed go to all cities controlled by that civilization, even if they have just been captured and are still in resistance. The free Broadcast Tower also goes to any city founded or acquired after the CN Tower has been completed, but, as when the city with the wonder switches hands, there will be no increase in population or happiness.
> >
> What about this?
> Anyway, the happiness bonus only applies as long as you own CN tower, so that's an error on the wiki.
> 

i am unable to fix this as i cant do pull requests about game logic, should i remove the cn towers from this pr?

PR #4688, opened 2021-07-30
Bring `allUnitActionsHaveTranslation` test up to date

SomeTroglodyte:
More chaff...
The current allUnitActionsHaveTranslation does not actually test for translateability of the actual Create [improvement] or Spread [religionName] buttons, but for translateability of the UnitActionType.value, which for these is unused. This `<music="dramatic">` _rectifies_ that _horrible blunder_ `</music>` by making the actual template known to the test like was done for `Upgrade`, then moving all three strings into the enum.


yairm210:
Mmmm I don't like that you're putting values into the ENUM, just so you can make the TEST easier. Don't like it at all/

SomeTroglodyte:
I half considered (but did not yet try) actually building the labels from these properties using `String.fillPlaceholders` later? If that's a good idea, let's go that way next, if not, I'll revert and use some structure local to the tests, maybe back to the `when`.

The current advantage is only that coders of new actions with dynamic labels will immediately see the need to specify the placeholder, and can do so in a more familiar place?

SomeTroglodyte:
Made _much_ simpler. I think that first comment in the UnitAction file will help most.

PR #4689, opened 2021-07-30
Reduce atlas

SomeTroglodyte:
The Huns was 4x larger than the other nations (per axis), and a dirty 'Hexagon' dupe could be optimized, so we now have a few weeks buffer until the atlas spills again into a second png. Enough for 4 or 5 nations with their uniques I'd say...

The hexagons - one in OtherIcons, one in FantasyHex. CrosshatchHexagon also exists in both places, but as identical copies which the packer optimizes away. Why didn't it pack both Hexagon.png into one? Because the one in FantasyHax is clean and has straight edges, while the other had imperfections. So I copied the cleaner one over the dirty one, and this promptly freed up a lot of texture space.


AdityaMH:
When i'm trying to remove CrosshatchHexagon.png with add transparent CrosshatchHexagon.png and the result:

IMG_20210730_142217

PR #4691, opened 2021-07-31
Added modoptions unique for disabling city-state spawning with only a settler

xlenstra:
Also extracted the "may convert gold to science using sliders" unique to a constant.


yairm210:
Why? What's the background for this?

xlenstra:
A user on discord asked for this as he was unhappy that his mod wasn't able to give extra starting units to city-states

xlenstra:
Another point: the `difficulties.json` file contains the field `aiCityStateBonusStartingUnits`, which is completely ignored in the current version, so either such an option should be added, or this field needs to be removed.

PR #4693, opened 2021-07-31
Update Russian.properties - Sweden translation.

IUnderstandNothingInProgramming:
I just decided to translate Sweden, I can't call this a "good translation", but still better than nothing, right?


sockybob:
Already translated in #4684

PR #4694, opened 2021-07-31
Fixed bug where production from cutting down forests could apply to perpetual constructions

xlenstra:
It is now instead added to overflow production


logicminimal:
what does perpetual constructions mean in this case?

xlenstra:
convert production to gold/science

PR #4698, opened 2021-07-31
Privateer capture, plunder, and raze Notifications

### LINKED ISSUE #4374: Privateer capture while defending message
Resulting to-do list, moved from discussion below:
- [x] Icons on capture should show capturing unit - war - victim.
- [x] Any destroy message for the same battle should be suppressed.
- [x] Capture active only on attack (thanks @ravignir ).
- [x] `plunderFromDamage` - mention 'ingame civilopedia', and maybe this issue as sources
- [x] `plunderFromDamage` calls - make offensive only or defensive/offensive controlled by unique
- [x] toInt() earlier so plunder message won't show fractional Gold
- [x] Test thoroughly
- [ ] Update wiki in case #4698 gets merged

image

... that was one and the same Trireme.

Question: Does Civ5 allow Privateers to capture while defending?
To-do: Check that message. Defender wins - so far OK. But if it results in capture the word 'kill' isn't appropriate and we should consider special-casing this. Of course only if the answer above is 'yes'.

SomeTroglodyte:
Made to resolve #4374 - for some details see discussion there.

- Large portion of the diff is due to finding (`val plunderableStats`), that List<String> converted to List<enum> then used as loop filter - could have been a Map<enum> but then I thought why not refactor properly, thus the property on the Stat enum - that part should not change anything functionally.

Otherwise this is about notifications and Piracy gains while defending:
- Kill message corresponding to a Privateer capture is suppressed
- Disconnected from capital notification corresponding to razed to the ground suppressed
- Citadel notifications now do aggressor - what happened - victim in their notifications like normal attacks
- City icon in notifications centralized and remapped to city center, to free the citadel so the citadel notifications use that - decision can be switched in `object NotificationIcon` easily.
- Plunder while defending is now its own unique - @xlenstra 's idea - thanks!
- Original plunder unique now only active for aggressor
- Normal attack notifications use Death icon instead of War when a unit was destroyed
- Plunder notification no longer show decimal digits

#### Before
image

#### After
image

#### City destroys unit
image

#### Privateer capture
image

#### Plundering
image

</details>

Untouched ideas:
- "What happened" = capture could get its own icon 
- Capture while defending might have been its own unique too like plunder - later, if someone asks


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -244,15 +266,14 @@ object Battle {
                         else " [" + defender.getName() + "]"
                     else " our [" + defender.getName() + "]"
             val notificationString = attackerString + whatHappenedString + defenderString
-            val cityIcon = "ImprovementIcons/Citadel"
-            val attackerIcon = if (attacker is CityCombatant) cityIcon else attacker.getName()
-            val defenderIcon = if (defender is CityCombatant) cityIcon else defender.getName()
+            val attackerIcon = if (attacker is CityCombatant) NotificationIcon.City else attacker.getName()
+            val defenderIcon = if (defender is CityCombatant) NotificationIcon.City else defender.getName()
```
xlenstra, line 270: Would it be more clean if there was an `getIconName()` function on the `ICombatant` interface?

xlenstra (approved):

core/src/com/unciv/logic/battle/Battle.kt
SomeTroglodyte: The idea came to me, too, but hey, gotta stop somewhere.

AdityaMH:
Add notification for unit promotion

xlenstra:
> Add notification for unit promotion

Not relevant for this PR.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -188,35 +201,41 @@ object Battle {
         damageToAttacker -= attacker.getHealth() // ... but from here on they are accurate
         damageToDefender -= defender.getHealth()
 
-        plunderFromDamage(attacker, defender, damageToDefender)
-        plunderFromDamage(defender, attacker, damageToAttacker)
+        plunderFromDamage(attacker, defender, damageToDefender, "Earn []% of the damage done to [] units as []")
+        plunderFromDamage(defender, attacker, damageToAttacker, "Earn []% of the damage done while defending against [] units as []")
```
yairm210, line 205: I'm not sure why you added this. 
Does this appear in the base game?

core/src/com/unciv/models/stats/Stat.kt
```diff
@@ -1,11 +1,23 @@
...
+ */
+enum class Stat(val canBePlundered: Boolean = false){
     Production,
     Food,
-    Gold,
-    Science,
-    Culture,
+    Gold(true),
+    Science(true),
+    Culture(true),
```
yairm210, line 12: DEFINITELY NOT.
Don't add a super niche variable to the ENUM.

core/src/com/unciv/logic/battle/Battle.kt
SomeTroglodyte: Since the answer is no, this will go including side effects. Why? Somewhere I mentioned it - someone's idea to allow modding 'plunder while defending'.

PR #4699, opened 2021-07-31
Actually load and display the road texture found in FantasyHex

### LINKED ISSUE #4672: Road visual bug
Why i can't show my unique road from my own mod?
Road keep showing default road even my unique road was added to mod (game.png, game.atlas).

SomeTroglodyte:
This is meant to fix #4672 

- Don't special case Road to getDot(BROWN), but load using enum name
- Rename existing png's to match case
- The larger diff parts are slight optimizations I couldn't resist

image

Note: This is done with _changed_ textures! The current ones don't take into account that they will be squashed length-wise by 3-5x. _Not_ included as doing so would have lost the rename operation and made it delete+add thus breaking history.


logicminimal:
is it possible for them to be connected?

core/src/com/unciv/logic/map/TileMap.kt
```diff
@@ -357,6 +357,22 @@ class TileMap {
...
+        // Statically allocate the Vectors (in World coordinates)
+        // of the 6 clock directions for border and road drawing in TileGroup 
+        private val clockToHexVectors: Map<Int,Vector2> = mapOf(
+            2 to HexMath.hex2WorldCoords(Vector2(0f, -1f)),
+            4 to HexMath.hex2WorldCoords(Vector2(1f, 0f)),
+            6 to HexMath.hex2WorldCoords(Vector2(1f, 1f)),
+            8 to HexMath.hex2WorldCoords(Vector2(0f, 1f)),
+            10 to HexMath.hex2WorldCoords(Vector2(-1f, 0f)),
+            12 to HexMath.hex2WorldCoords(Vector2(-1f, -1f))
+        )
```
yairm210, line 370: This should go in HexMath

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -487,8 +476,7 @@ open class TileGroup(var tileInfo: TileInfo, var tileSetStrings:TileSetStrings,
     private fun updateRoadImages() {
         if (forMapEditorIcon) return
         for (neighbor in tileInfo.neighbors) {
-            if (!roadImages.containsKey(neighbor)) roadImages[neighbor] = RoadImage()
-            val roadImage = roadImages[neighbor]!!
+            val roadImage = roadImages[neighbor] ?: RoadImage().also { roadImages[neighbor] = it }
```
yairm210, line 479: We've been using .apply{} for the whole project, why .also{} here?

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -505,20 +493,10 @@ open class TileGroup(var tileInfo: TileInfo, var tileSetStrings:TileSetStrings,
             }
             if (roadStatus == RoadStatus.None) continue // no road image
 
-            val image = if (roadStatus == RoadStatus.Road) ImageGetter.getDot(Color.BROWN)
-            else ImageGetter.getImage(tileSetStrings.railroad)
+            val image = ImageGetter.getImage(tileSetStrings.tileSetLocation + roadStatus.name)
```
yairm210, line 496: This means we'll generate a NEW STRING for every single road or railroad image. Please use tileSetStrings.getString(tileSetStrings.tileSetLocation, roadStatus,name), which is built to avoid this problem.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
SomeTroglodyte: Just because `roadImages[neighbor] = it` reads 0.0001% nicer than `roadImages[neighbor] = this`

core/src/com/unciv/logic/map/TileMap.kt
SomeTroglodyte: Sounds plausible. Placement was steered by 'let's plonk this in right next to where my screen is currently looking at' - but then the whole routine might be misplaced as the only benefit it gets from `this` is `getNeighborTileClockPosition` having access to map dimensions... Needs a bit of thought.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
SomeTroglodyte: Oh, right, I think I misplaced the intention to cache the entire set as `tileSetStrings.railroad` is a bit orphaned, too. This looks more like proof-of-concept.

AdityaMH:
> is it possible for them to be connected?

For this time, absolutely no.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
SomeTroglodyte: No, getString isn't the best approach. Why are all those top-level string vals in TileSetStrings not passed through that function? Because that would gain nothing but overhead, and the roads can live there, too.

Besides, that hashmap makes me curious. Why doesn't it try to leverage the JVM's string interning? Why > 200 entries for era-specific images that don't exist?

SomeTroglodyte:
Latest push = the tileset path strings for road and railroad are again cached, extensible should RoadStatus change. The set of precalculated vectors now lives in HexMath.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
yairm210: You're more than welcome to improve it as well :)
I just had to find a way to lower the internal heap size and this was the 'quick' way that lead to the best savings

yairm210:
Looks OK.

core/src/com/unciv/ui/tilegroups/TileSetStrings.kt
```diff
@@ -13,7 +14,10 @@ class TileSetStrings {
     val hexagon = tileSetLocation + "Hexagon"
     val crosshatchHexagon = tileSetLocation + "CrosshatchHexagon"
     val cityOverlay = tileSetLocation + "CityOverlay"
-    val railroad = tileSetLocation + "Railroad"
+    val roadsMap = RoadStatus.values()
+        .filterNot { it == RoadStatus.None }
+        .map { it to tileSetLocation + it.name }
+        .toMap()
```
yairm210, line 20: Why .toMap? Isn't this already a map?

SomeTroglodyte: no, .map{} produces a List<Pair> similar to what you can feed to mapOf().

From a C++ background, you'd think "Array indexed directly by the enum", where the actual lookup could be compiled down to a single machine instruction (at least on 80xx as long as the elements are aligned at certain multiples of 4).  And indeed, `val roadsArray = RoadStatus.values().map { tileSetLocation + it.name }.toTypedArray()` and `val image = ImageGetter.getImage(tileSetStrings.roadsArray[roadStatus.ordinal])` runs fine, but - that's not a kotlin primitive Array under the hood but a java boxed Array, and the library code is full of unchecked casts, so I'm prejudiced against them...

SomeTroglodyte: I've timed it - `updateRoadImages` takes an average ~770ns with the Array, ~700ns with the Map, done over 3 realistic game turns on a large map netting ~2.6M calls.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
SomeTroglodyte: Yea, quickies and their documentation... But improving on _that_ one is daunting - and a bit futile, an in-depth solution should start with thinking outside the box and removing the need for those strings in the first place. Maybe think how ssas optimizes parent-child dimensions that properly use surrogate keys and try something similar........ Ouch my brain hurts.

PR #4703, opened 2021-08-01
Improve Road & Railroad for FantasyHex

AdityaMH:
Road recently have big fail curly and railroad still look like black line.
So, i re-art new one.


SomeTroglodyte:
I'm trusting you to fine-tune the looks - but I fear this is still too 'wysiwyg' - but you don't get what you see in your favourite Image editor, the texture will be severely squashed in-game. Also, two of them will be 'glued together' - one on each side of the hex border.

Tuning is best done when you can immediately look at the result of a change, and that's best done by running Unciv on desktop (and making sure you see it decided to re-pack the relevant atlas - see console output) - or from Android Studio is OK too of course.

Your new railroad looks still too 'detailed' to me - I'd think at most 5-6 'sleepers' on the image, symmetrical, and 'cut' exactly half-way between two sleepers might work best. You could try changing aspect ratio - maybe 60x28 will be squashed less and almost be wysiwyg? I haven't tried. Also, please make both images the same pixel dimensions.

Your call :+1:

AdityaMH:
Thanks for suggestion.

yairm210:
The road image change seems to conflict with an existing change

AdityaMH:
I can't resolve this conflict. I don't know why.

PR #4710, opened 2021-08-01
Added desktop resolutions

Alex25820:
Added 5 most common desktop resolutions according to [statcounter](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide).

When installing the game on my mac i noticed that it looked a bit weird, after having added the correct resolution (1440x900) to the game it now looks like it should. I decided to add the most common desktop resolutions to the game as there seemed to be no desktop resolutions in the game at the moment.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
@@ -31,7 +31,7 @@ class OptionsPopup(val previousScreen:CameraStageBaseScreen) : Popup(previousScr
     private var selectedLanguage: String = "English"
     private val settings = previousScreen.game.settings
     private val optionsTable = Table(CameraStageBaseScreen.skin)
-    private val resolutionArray = GdxArray(arrayOf("750x500", "900x600", "1050x700", "1200x800", "1500x1000"))
+    private val resolutionArray = GdxArray(arrayOf("750x500", "900x600", "1050x700", "1200x800", "1280x720", "1366x768", "1440x900", "1500x1000", "1536x864", "1920x1080"))
 
     init {
         settings.addCompletedTutorialTask("Open the options table")


SomeTroglodyte:
Those existing 'resolution' entries do not have a _direct_ relation to pixels, they control the virtual world coordinates used in an OpenGL viewport...

yairm210:
Not exactly relevant, I think.
Although, if you post a before and after screenshot, I'll be able to see the difference :)

Alex25820:
I thought that the text was much easier to read on the correct resolution, but now that I am seeing them side by side I think that it might just be because the text is bigger on 1440x900 than on 1500x1000. _(I chose 1500x1000 because it was the closest to my display resolution.)_

_Although I would still recommend adding at least a few common desktop resolutions since it feels really weird to boot up a game that only has these really weird resolutions to choose from. Another alternative could also be to add a AUTO option that automatically sets the resolution to the screen resolution of the device. :)_

**1440x900 (Correct)**
1440x900 (Correct)

**1500x1000(Incorrect)**
1500x1000(Incorrect)

yairm210:
I'm not seeing a big difference, nothing looks 'wrong' in the wrong resolution

SomeTroglodyte:
Maybe we should just relabel the option so the association to pixels isn't so immediate? By the way, a while ago I also experimented with additions to that list - just to see whether one could profit on a 2.5k or 4k monitor - and decided against.

Alex25820:
Relabelling sounds like a good idea that I think we should consider. I think I will be looking into if I can somehow make the text easier to read as it becomes a little blurry at higher resolutions.

Will be closing this pull-request now as it is no longer needed. :)

SomeTroglodyte:
> make the text easier to read

LibGdx Scene2D is still using OpenGL and a 3D surface, there's not much one can do. Aligning texture pixels and screen pixels is a little like fighting windmills, so it's easier to render text to texture with enough pixels so when the library does its thing it stays bearable - I'm exaggerating only a little.

I had a version in test that ran fine and used 'multi-channel signed distance field' fonts under the hood - that did look markedly better at small text sizes, but I couldn't reliably align character baselines so I ditched it. Also, that would reintroduce the character set completeness issues Unciv avoids by asking the system for a font, and generating MSDF on the fly from that is beyond me. I think you'll find the tools here under 'maltaisn' or similar - if you want to pursue that.

Issue #4709, opened 2021-08-01
Hill deprecation shows crash loophole in `getLastTerrain()`


SomeTroglodyte:
- Find an old map that has used an old Civ5ExpansionMod and that way has BaseTerrain Hill and TerrainFeature 'Desert Hill'
- Load it with current code and current mods
- The loader thread crashes, leaving the game in an unrecoverable state, stack trace fingers `isImpassible()`

The actual cause is that in `getLastTerrain()`, the assumption is made that `terrainFeatures.isNotEmpty()` means `getTerrainFeatures()` isn't empty either, which is not the case when some feature string is missing from the ruleset - that function uses `.mapNotNull` - uh-huh, obvious.

Proposed approach, completely untested:
- `val terrainFeatureObjects: List<Terrain> by lazy { terrainFeatures.mapNotNull { ruleset.terrains[it] } }` and use where appropriate, including returning it directly from `getTerrainFeatures()`

Comments, critique, lambasting, ridicule, the stocks, anything?

Note: Map loader UI needing more robustness in case its sub-thread crashes is a completely different story.

yairm210:
So is this resolved with latest PR?

Issue #4705, opened 2021-08-01
Maps in mods dont appear in Map Editor


ArchDuque-Pancake:
Currently when downloading maps via mods, they dont show up in the Map Editor automaticly.

Is this intended behaviour? I dont think so. By having this, editing for personal use or tweaks would be easier and shorter.

SomeTroglodyte:
Yes I think you should edit only your own maps. Modders can copy them as needed locally, and making them do so is good to promote 'switching you brain to _on_ mode'.

But thanks for bringing to our attention the map editor loader is buggy (doesn't handle bad maps gracefully).

ArchDuque-Pancake:
Ah ok, true. I'll close this then.

PR #4711, opened 2021-08-01
Harden map editor map loader against most bad maps

SomeTroglodyte:
Quick&dirty inspired by #4705
- Guards against exceptions in `MapSaver.loadMap` (e.g. something that isn't a gzipped json)
- Guards against exceptions in `MapEditorScreen(map)` (e.g.  #4709)
- Explicit message when the loaded map requires missing mods (hides the last one for the case outlined in 4709):
image

### DIFF
--- a/core/src/com/unciv/ui/mapeditor/SaveAndLoadMapScreen.kt
+++ b/core/src/com/unciv/ui/mapeditor/SaveAndLoadMapScreen.kt
@@ -89,9 +89,11 @@ class SaveAndLoadMapScreen(mapToSave: TileMap?, save:Boolean = false, previousSc
                         }
                     } catch (ex: Throwable) {
                         needPopup = false
-                        popup?.close()
-                        println("Error loading map \"$chosenMap\": ${ex.localizedMessage}")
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/mapeditor/SaveAndLoadMapScreen.kt
```diff
@@ -54,17 +55,43 @@ class SaveAndLoadMapScreen(mapToSave: TileMap?, save:Boolean = false, previousSc
...
+                                println("Error displaying map \"$chosenMap\": ${ex.localizedMessage}")
+                                Gdx.input.inputProcessor = stage
+                                ToastPopup("Error loading map!", this)
+                            }
+                        }
+                    } catch (ex: Throwable) {
+                        needPopup = false
+                        popup?.close()
+                        println("Error loading map \"$chosenMap\": ${ex.localizedMessage}")
+                        ToastPopup("Error loading map!", this)
```
yairm210, line 94: This should also have a postrunnable

PR #4712, opened 2021-08-01
Proposal on how to achieve "unimproved Forest"

SomeTroglodyte:
Request for comments... Thanks @xlenstra

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -439,11 +439,15 @@ open class TileInfo {
 
     /**
      * Implementation of _`tileFilter`_
-     * @see <a href="https://github.com/yairm210/Unciv/wiki/uniques#user-content-tilefilter">tileFilter</a>
+     * @see <a href="https://github.com/yairm210/Unciv/wiki/Uniques#tilefilter">tileFilter</a>
      */
     fun matchesFilter(filter: String, civInfo: CivilizationInfo? = null): Boolean {
+        if (filter.startsWith('{')) // multiple types at once - AND logic. Looks like:"{Forest} {Hill}"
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -439,11 +439,15 @@ open class TileInfo {
 
     /**
      * Implementation of _`tileFilter`_
-     * @see <a href="https://github.com/yairm210/Unciv/wiki/uniques#user-content-tilefilter">tileFilter</a>
+     * @see <a href="https://github.com/yairm210/Unciv/wiki/Uniques#tilefilter">tileFilter</a>
      */
     fun matchesFilter(filter: String, civInfo: CivilizationInfo? = null): Boolean {
+        if (filter.startsWith('{')) // multiple types at once - AND logic. Looks like:"{Forest} {Hill}"
+            return filter.removePrefix("{").removeSuffix("}").split("} {")
```
yairm210, line 446: Mmmm, we still get into the same problem we get into with the unit filters, which is that in Chinese they don't have spaces between them.
Maybe we should have a splitFilter(filter:String):Sequence<String> function that can split both with and without spaces? I think that's the simplest solution

core/src/com/unciv/models/translations/TranslationFileWriter.kt
```diff
@@ -206,6 +206,7 @@ object TranslationFileWriter {
             fun submitString(item: Any) {
                 val string = item.toString()
 
+                //todo This needs to learn about {}-using "AND" filters
```
yairm210, line 209: We could then use the same splitFilter function in the translationFileWriter, if the parameter is a multi-filter, split and take the first. Should be good enough, I think.

PR #4717, opened 2021-08-01
Split up unit types to its own JSON file

xlenstra:
Currently this only contains whether the unit is a(n) land/water/air unit, and some uniques that apply to all units of this type, but it lays the foundation for this to be extended to other properties as well.
It also renames the unit types to differentiate between pre-gunpowder and post-gunpowder melee/ranged units, and some other readability changes of this sort.
Fully compatible with mods, as it also includes all the old unitTypes whenever a mod without their own unitTypes.json is used.

If you had other plans for the direction this split should go in that are incompatible with this PR, feel free to close it.


core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -557,14 +556,17 @@ open class TileGroup(var tileInfo: TileInfo, var tileSetStrings:TileSetStrings,
...
+                militaryUnit.baseUnit.replaces != null &&
                         ImageGetter.imageExists(tileSetStrings.unitsLocation + militaryUnit.baseUnit.replaces) ->
                     tileSetStrings.unitsLocation + militaryUnit.baseUnit.replaces
-
-                unitType == UnitType.Mounted -> tileSetStrings.unitsLocation + "Horseman"
-                unitType == UnitType.Ranged -> tileSetStrings.unitsLocation + "Archer"
-                unitType == UnitType.Armor -> tileSetStrings.unitsLocation + "Tank"
-                unitType == UnitType.Siege -> tileSetStrings.unitsLocation + "Catapult"
+    
+                /** FIXME: these should probably not be hardcoded, but I don't know any other easy solution
```
yairm210, line 563: I think we can search the units for the first unit of this type that has an image, and take its sprite

yairm210:
AMAZING! I've been thinking about this change for about a year now, and every time baby steps towards it, but it's amazing to see it happen :D

One point, though - Instead of "domain", "movementType"

yairm210:
Once that's done, EVEN IF IT HAS PROBLEMS, I'm merging this so there are no takebacks ;)

SomeTroglodyte:
I would have named it domain, too :wink:

core/src/com/unciv/logic/battle/CityCombatant.kt
```diff
@@ -28,7 +28,7 @@ class CityCombatant(val city: CityInfo) : ICombatant {
         if (city.health < 1) city.health = 1  // min health is 1
     }
 
-    override fun getUnitType(): UnitType = UnitType.City
+    override fun getUnitType(): UnitType = UnitType("City")
```
SomeTroglodyte, line 31: You could make this drop-in compatible by delivering ".City" from the companion object - might also be potentially more efficient as the instance will be reused - and I see no reason why it shouldn't. The new UnitType is no longer immutable due to deserialization, but functionally they're stable or the old code wouldn't have worked... Or am I thinking wrong?

tests/src/com/unciv/logic/map/UnitMovementAlgorithmsTests.kt
```diff
@@ -235,7 +236,7 @@ class UnitMovementAlgorithmsTests {
         city.civInfo = otherCiv
         tile.owningCity = city
 
-        unit.baseUnit = BaseUnit().apply { unitType = UnitType.Melee }
+        unit.baseUnit = BaseUnit().apply { unitType = ruleSet.unitTypes.keys.random(); ruleset = ruleSet }
```
SomeTroglodyte, line 239: Unpredictability in a unit test? Since that's in `canNOTPassForeignTiles` I'd make that a `first()`, since all alternatives would serve equally well for the tests's goal

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -753,7 +751,7 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Cam
                     .flatMap { it.cities.asSequence() }.any { viewingCiv.exploredTiles.contains(it.location) }
         }
         displayTutorial(Tutorial.ApolloProgram) { viewingCiv.hasUnique("Enables construction of Spaceship parts") }
-        displayTutorial(Tutorial.SiegeUnits) { viewingCiv.getCivUnits().any { it.type == UnitType.Siege } }
+        displayTutorial(Tutorial.SiegeUnits) { viewingCiv.getCivUnits().any { it.type!!.name == "Siege" } }
```
SomeTroglodyte, line 754: Hardcoding - I'd have left the knowledge how to recognize a Siege unit in the UnitType class - `isSiege()` or a const or a factory in the companion, all would do it. With an instance comparison in the 'factory' choice, one should make sure the class satisfies the pattern for use as HashMap key - which is a good idea for this class anyway.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
SomeTroglodyte: These are _fallbacks_ in case a specific unit is missing a pixel unit image - I'd at least persist the choice outside of `TileGroup.updatePixelMilitaryUnit` - doesn't that run often? `imageExists` can be surprisingly expensive on Android. And - I'd put it in json, cheap, and if that's missing in e.g. a mod - simply no pixel unit.

SomeTroglodyte:
```
        "name": "AtomicBomber",
        "Domain": "Air"
```
yes - domain! Another is left in the secondary constructor

`getMovementType()` should be cached
I'd fold `if (test) true else false` to `test`
Promotion.unitTypes doesn't match so units won't be able to promote?

SomeTroglodyte:
PromotionPickerScreen line 50 has another `UnitType.toString()` -> use name instead...

SomeTroglodyte:
With those two toString's patched _no other surprises_ while playtesting - with this variant of 
<details><summary><code>UnitType.kt</code>:</summary>

```kotlin
package com.unciv.models.ruleset.unit

import com.unciv.models.ruleset.Unique
import com.unciv.models.stats.INamed


enum class UnitLayer { // The layer in which the unit moves
    Civilian,
    Military,
    Air
}

enum class UnitMovementType { // The types of tiles the unit can by default enter
    None,
    Land, // Only land tiles except when certain techs are researched
    Water, // Only water tiles
    Air // Only city tiles and carrying units
}

class UnitType(
    private val movementType: String? = null
) : INamed {
    override lateinit var name: String

    private val uniques: ArrayList<String> = ArrayList()
    
    val uniqueObjects: List<Unique> by lazy { uniques.map { Unique(it) } }
    
    constructor(name: String, movementType: String? = null) : this(movementType) {
        this.name = name
    }

    override fun toString(): String {
        throw Exception("UnitType.toString called!")
    }

    private val unitMovementType: UnitMovementType by lazy { 
        if (movementType == null) 
            UnitMovementType.None 
        else try {
            UnitMovementType.valueOf(movementType)
        } catch (ex: Throwable) {
            UnitMovementType.None
        }
    }
    fun getMovementType() = unitMovementType
    
    fun isLandUnit() = getMovementType() == UnitMovementType.Land
    fun isWaterUnit() = getMovementType() == UnitMovementType.Water
    fun isAirUnit() = getMovementType() == UnitMovementType.Air
    
    fun matchesFilter(filter: String): Boolean {
        return when (filter) {
            "Land" -> isLandUnit()
            "Water" -> isWaterUnit()
            "Air" -> isAirUnit()
            else -> uniques.contains(filter)
        }
    }
    
    companion object {
        val City = UnitType("City", "Land")
    }
}
```
</details>

No I don't mean you should use that - not at all - just food for thought. There's still the matter of constructor(String) _almost_ being ambiguous.

SomeTroglodyte:
Already merged? @yairm210  don't tag it then until promotions work, minimum patch PromotionPickerScreen line 50 and UnitPromotions line 49 toString() -> name

PR #4723, opened 2021-08-02
remove distribution from roadmap

logicminimal:


logicminimal:
i removed the "killer" phrase from the title as i added it accidentally due to google keyboards swipe write and pressed send instead of pressing the back button, sorry for that

PR #4728, opened 2021-08-02
Tweak UncivSlider and the one on NewGameScreen

SomeTroglodyte:
- `pack()` improves position calculation that was off by the "-" button's width in some cases
- Slider for Citystate count on NewGameScreen permanently shows the value tip

I can't find the issue that mentioned that last one, but I'm sure that was mentioned.


logicminimal:
i was the one to point out that city state one, i commented that in your pr that adds sliders

PR #4733, opened 2021-08-02
Update German translation

Mape6:


Mape6:
@yairm210 Did you merge this commit?

xlenstra:
Looking at it, it seems like something went wrong when merging this into the translation update for the 1.16.1-patch1 update. I don't see anything in your PR that would require its closure, so I'll try to reopen this for you.
This also looks like #4743 was closed, and if it works here I'll try the same there.

xlenstra:
Seems that after the `translations` branch was deleted, there is no longer an easy way for this PR to be saved, would it be very inconvenient for you to create this a new PR with the changes of this PR?

Mape6:
> Seems that after the `translations` branch was deleted, there is no longer an easy way for this PR to be saved, would it be very inconvenient for you to create this a new PR with the changes of this PR?

Thanks that you took a look into the problem. I will open another PR with my changes.

PR #4734, opened 2021-08-02
All stats available as text characters

SomeTroglodyte:
This is just preparation - I'd like to decorate costs in Civilopedia with their Stat icon, and - 2 of 3 are currently in our Font. Since we see amounts decorated in the fandom wiki all over the place, and Civ5 itself as far as I know likes to do that, too - why not bring the proportion of Stats that can be decorated in text from 29% to 100%?

Images edited from existing 200px ones: the not-quite-round ones made round first, scaled to 40x40 in a 50x50 canvas, then moved to baseline - no new credits necessary.

The comments - what they mean is that symbols needing two codepoints in utf-16 will not work at all. Trim & clarify?

image


logicminimal:
this implements my proposal on my ui thread, thanks!

PR #4736, opened 2021-08-02
Use stat characters in Civilopedia Unit/Building costs

SomeTroglodyte:
... and hide the Great Prophet if playing without Religion.
image
image

This displays _base_ cost, for rush buying, too, thus the patch to getBaseGoldCost() which wasn't quite _base_.

Somehow I doubt a little all units should be buyable...


logicminimal:
a few questions

1: why should great prophets be removed when religion is off in civilopedia? does this matter?

2: what happens if a player is playing in a different game pace? is this accounted for?

great pr btw

PR #4737, opened 2021-08-02
Civilopedia phase 9 - Technologies

SomeTroglodyte:
**Important**: Technology was the only "ruleset object" class _not_ inheriting INamed despite having a named property. I changed that - if there was a reason, please yell and we'll see. I found no negative effects in testing.

<details><summary>Screenshot:</summary>

image
</details>

This diff does not read well, sorry, seems the matching is getting confused by me moving the simple one-liners `toString`, `era` and `isContinuallyResearchable` to the beginning of the file where they somehow fit better, as the rest of the methods are all for display purposes.

Add to that the detail that `getEnabledBuildings`, `getObsoletedBuildings` and `getEnabledUnits` all had only callers that checked "Will not be displayed in Civilopedia" after calling them, so that test was better moved into them, and equalized their architecture to Sequence-based. Also, the new code should handle mods where a unique unit of a nation has a different tech prerequisite than its base unit a little better, as well as coping properly with unique units that do not replace anything.

Lastly, the filters when nuclearWeaponsEnabled is off should be cleaner using the correct current uniques instead of hardcoding the manhattan project.


core/src/com/unciv/models/ruleset/tech/Technology.kt
```diff
@@ -81,47 +90,161 @@ class Technology {
...
+        { it.uniqueObjects.any { unique -> unique.placeholderText == "Obsolete with []" && unique.params[0] == name } }
+    
+    // Helper: common filtering for both getEnabledBuildings and getObsoletedBuildings, difference via predicate parameter
+    private fun getFilteredBuildings(civInfo: CivilizationInfo, predicate: (Building)->Boolean): Sequence<Building> {
+        val nuclearWeaponsEnabled = civInfo.gameInfo.gameParameters.nuclearWeaponsEnabled
+        val religionEnabled = civInfo.gameInfo.hasReligionEnabled()
+
+        return civInfo.gameInfo.ruleSet.buildings.values.asSequence()
+            .filter {
+                predicate(it)   // expected to be the most selective, thus tested first
```
yairm210, line 116: I wonder, this looks like it applies to buildings in the Civilopedia on general.
Maybe building should have a "should be displayed in Civilopedia" function that accepts a civinfo?

logicminimal:
i have one thing to suggest: make the civilization icon and name on top left part of the screen direct you to the civilopedia page for that civilization instead of being a backup button for overview

SomeTroglodyte:
I actually had that idea too, but - marginal.

logicminimal:
> I actually had that idea too, but - marginal.

what do you mean by - marginal?

SomeTroglodyte:
https://en.wiktionary.org/wiki/marginal - 2.2 "barely productive"

PR #4742, opened 2021-08-03
Make screen orientation sensorable

AdityaMH:
Please don't merge it until tested by actual phone and worked.
Recent one can't do reserve-landscape except change it from phone setting.
Anyone, can you help me to compile this to apk then tested by someone (include me).

### DIFF
--- a/android/src/com/unciv/app/LimitOrientationsHelperAndroid.kt
+++ b/android/src/com/unciv/app/LimitOrientationsHelperAndroid.kt
@@ -29,7 +29,7 @@ class LimitOrientationsHelperAndroid(private val activity: Activity) : LimitOrie
     override fun allowPortrait(allow: Boolean) {
         val orientation = when {
             allow -> ActivityInfo.SCREEN_ORIENTATION_USER
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2 -> ActivityInfo.SCREEN_ORIENTATION_USER_LANDSCAPE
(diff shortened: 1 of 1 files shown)


AdityaMH:
> still play you'll need your brain to perform a 90° rotation on what you see.

I don't get it from this word.
But, i'm just want sensorable for landscape mode, and keep it constant for portait mode.

It's would be worked? If just change "that"?

yairm210:
If I understand correctly, if the user disables 'lock orientation', then the current code will act the way that this PR forces it to act.
Meaning, this PR actually reduces the use choice that he currently has.
@SomeTroglodyte did I understand correctly?

If this is the case, this PR will be denied - there's no reason to reduce user choice.

SomeTroglodyte:
I just noticed - this PR would override an OS-level user setting _only_ when our opt-in is OFF, and change nothing if either the opt-in is ON or the Android version is too old. Within that constraint, yes, it would reduce user choice by overriding something they can set in their system settings.

Readily duckduck'ed sources say the USER constants 'collaborate' with the OS level orientation lock while the SENSOR constants _override_ it... e.g.:
https://developer.android.com/reference/android/R.attr?hl=el#screenOrientation

What AdityaMH may have a problem with is certain Android vendor-specific adaptations make access to that orientation lock a tiny little bit harder than others, plus some globally forbid one of 4 possible orientations (through build.prop) - I think a Samsung S5 out of the box would refuse to follow if you turned it upside down, while the Tab S, almost same year, would follow all 4 possible orientations...

Summary: I _really_ believe these choices are best, and cases of the user wanting a certain orientation but not getting it are a matter of OS overrides or not properly operating the standard orientation lock (or our opt-in switch).

PR #4743, opened 2021-08-03
Update Spanish.properties

jgmy:
Avoid having two Sevillas and two Zaragozas.

In Spanish, CaesarAugusta is called Zaragoza, and Zaragoza is also called Zaragoza. In fact, they are the same city,
Same with Hispalis and Seville: both of them are the city of Seville, called Sevilla in Spanish.

But this can cause confusion if a player owns Zaragoza1 and is attacked from Zaragoza2.

So I opted for using Cesaraugusta and Híspalis for the Roman cities, and leaving the Spanish form for the other ones.

Please notice: in the past I detected some units have different names from the Civilopedia/help ones. I tried to solve it in a previous pull request. But I think I did something wrong and I never did the actual pull request.


jgmy:
I just looked my old unpublished translation and corrected some typos the current translation still had. Most of them deal with accented letters (tán => tan) and missing colons or dots. 

I also corrected the item badly translated. Court was translated as "delegación" in one message, and "Corte de Justicia" in the rest of them. I changed "Delegación" to "Corte de Justicia".

SomeTroglodyte:
With my limited español, I'd say thorough work. :+1:

jgmy:
Ok, as you wish.

PR #4744, opened 2021-08-03
Bugfixes from unitTypes so promotions work again

xlenstra:
As noted by someTroglodyte after the unitTypes split was merged, promotions didn't work anymore due to somewhere `unitType.toString()` being called instead fo `unitType.name`. In the old version this wouldn't have mattered, but due to `unitType` now being a class, it suddenly does. This PR fixes this problem and a few others noted by someTroglodyte


SomeTroglodyte:
Just saw that from Studio so I didn't search the other PR's, sorry. But it's nice to see we still cooperate somehow even if I can't get Discord to run. (I mean, absent from ubuntu repos - if that's not proof something is not 100% kosher then what is)

PR #4746, opened 2021-08-03
MCO - Randomize Plains/Grasslands around deserts

ravignir:
- Deserts are more common (but now have higher chance of having a stripy look)
- Slightly reduced range of Tundra.
- Slightly reduced amount of Snow.
- Grasslands can now appear next to desert tiles.
az

The question is whether Grasslands are now too common or not?


yairm210:
Looks good

PR #4747, opened 2021-08-03
Fix: Promotions broken by #4717

SomeTroglodyte:
UnitType.toString() is no longer what you expect...
 it's "com.unciv.models.ruleset.unit.UnitType:@1BAFF8" or something similar, therefore no unit, no matter how much XP it has, will be able to promote


xlenstra:
See also: #4744 :)

Issue #4707, opened 2021-08-01
Losing your starter settler shouldn't make you lose the game


logicminimal:
as ravignir says you need to lose all your units (all military+settler units im guessing) to lose the game if you havent settled a city. in the current version of the game losing your settler is enough which mustn't be, as you have a chance to upgrade your warrior get units from barbarian encampments and get a city

more information: say in a scenario you capture a city before settling your first settler and you lose that city, you dont lose as you have your special settler startersettler unique

lot more information: if you settle your first city normally then get a settler then get all your cities conquered you lose as you lost all your cities and your only settler isn't a special settler

yairm210:
This is inconsistent.
Either you lose by losing all cities, or you lose by losing all cities and settlers.
There's no such thing a special settler.

logicminimal:
> This is inconsistent.
> Either you lose by losing all cities, or you lose by losing all cities and settlers.
> There's no such thing a special settler.

in civ 5 tgere is, just ask @ravignir

yairm210:
This sounds like a "wontfix" to me.
I'm not going to make certain units 'special' versions of their kind.

ravignir:
This ain't a wontfix. There is no special settler.
Lose condition just changes from:
- lose all your units (until you control any original capital)
- lose all your cities (activates when you control any original capital).

ravignir:
I've made such tests to determine that:
- disband a settler right after the start (game still continues).
- capture another civs non-capital city while still having a settler, then let the AI retake it (game still continues)
- capture another civs capital city while still having a settler, then let the AI retake it (game is lost)

SomeTroglodyte:
> Lose condition just changes from...

Do I understand that correctly:
- If I've never founded my original capital myself and have never conquered any original capital, then it's irrelevant whether I own a city (which would be my capital automatically, including Palace, but it wouldn't be an _original_ capital) or not, if my unit count is zero, I lose. I have any unit, any type, even a worker, then I'm still in the game.
- Once I have ever controlled any original capital, my units are irrelevant for determining loss condition, only losing my last city counts, even if I have the map covered in settlers.

Therefore, what a Settler is (in a mod or Conquistador or...) or how many I have never matters at all.

ravignir:
> > Lose condition just changes from...
> 
> Do I understand that correctly:
> 
[...]

PR #4761, opened 2021-08-04
Change defeat conditions

SomeTroglodyte:
- Implements defeat conditions as per #4707, ravignir's clarification
- Fixes a typo - thus all language files touched
- Fixes 2^31-1 showing as turns-to-tech when science per turn is zero (possible when testing item 1) - why that division by zero doesn't crash is a mystery
- Fixes crash when player loses (regression from introducing victory type as unique parameter)

Holler if you'd rather not do the changed defeat logic, I would totally understand - I'll isolate the fixes.


SomeTroglodyte:
Wow, not AS stealing blanks but `lateinit property name has not been initialized` - each time I forget to run the UT locally it bites back. Gimme a quarter.

SomeTroglodyte:
Turns out the mock civInfos in the diplomacy tests were a bit fragile and stumbled over isBarbarian() crashing - I hope this approach is OK, adding a unit test just to make sure the prerequisites for the main unit tests are met may not be a good pattern. At least this way such dependencies are clearly documented...

core/src/com/unciv/logic/civilization/TechManager.kt
```diff
@@ -113,8 +113,10 @@ class TechManager {
     
     fun remainingScienceToTech(techName: String) = costOfTech(techName) - researchOfTech(techName)
 
-    fun turnsToTech(techName: String): String {
-        return if (civInfo.cities.isEmpty()) "∞" else max(1, ceil(remainingScienceToTech(techName).toDouble() / civInfo.statsForNextTurn.science).toInt()).toString()
+    fun turnsToTech(techName: String) = when {
+        civInfo.cities.isEmpty() -> "∞"
```
yairm210, line 117: I don't think we need this line anymore if we check the tech directly

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -76,15 +76,15 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
 
 
     private fun wonOrLost(description: String) {
-        
-        val endGameMessage = when (description.getPlaceholderParameters()[0]) {
+        // description will be empty when the player loses - don't `[0]`
+        val endGameMessage = when (description.getPlaceholderParameters().firstOrNull()) {
             // Taking the 0th element is a dirty hack to differentiate between the cases
```
yairm210, line 81: And this comment is deprecated too

SomeTroglodyte:
Quite right. The first one was "Double safety net" kind of thinking, and "maybe statsForNextTurn is an expensive lazy and I don't wanna check right now".
The interesting thing is how I got this to 'trigger', and I think I've understood now - only happened if the only city is a conquest and _puppeted_ therefore keeping its income from me - or so.

PR #4764, opened 2021-08-04
Replaced the last promotion effects with uniques

xlenstra:


SomeTroglodyte:
Roadmap? I'm reasonably sure there's no separate use of the field left. Deprecate, yes?

Some display usecases may need checking - 'pedia display is freshly done, picker is likely still OK for all existing promotions, but the abbreviated display for unique units sporting more free promotions than their base unit on the nation displays needs double-checking...

yairm210:
Deprecation is scheduled, but we'll need to notify the missed via Discord of the change and wait for them to change existing mods

PR #4771, opened 2021-08-05
Ruins now have their own file

xlenstra:
This PR updates ancient ruins.
It adds a ruins.json file, which contains all possible rewards, and one of these will randomly be chosen whenever ruins are entered.
It also adds a lot of uniques for the different rewards, and basically rewrites the `triggerCivwideUnique()` function to implement all these new ruins effects.


logicminimal:
id suggest adding chances to each one of it also making them usable with other uniques like tilrfilter or unitfilter to be more useful and moddable

ravignir:
Won't be percentage just 100%/amount of possible rewards?

logicminimal:
> Won't be percentage just 100%/amount of possible rewards?

iirc there are more rarer rewards

ravignir:
one thing to add:
add an option to specify how many turns after starting the game are some rewards available.

ravignir:
I spend some time on Faith ruins:
- Available after 20 turns into game
- Before Pantheon, they always provide enough faith to get one (slightly more actually, as received faith seem to be always a multiple of 10).
- After Pantheon, the amount of faith is 60 faith on standard speed and reward is only available if yopu have not found religion yet.

xlenstra:
> I spend some time on Faith ruins:
> 
>     * Available after 20 turns into game
> 
>     * Before Pantheon, they always provide enough faith to get one (slightly more actually, as received faith seem to be always a multiple of 10).
> 
>     * After Pantheon, the amount of faith is 60 faith on standard speed and reward is only available if yopu have not found religion yet.

Thanks! I've used these values in combination with the game files so it should be implemented correctly now.

core/src/com/unciv/logic/civilization/Notification.kt
```diff
@@ -18,6 +19,24 @@ object NotificationIcon {
...
     const val City = "ImprovementIcons/City center"
     const val Citadel = "ImprovementIcons/Citadel"
+    const val Happiness = "StatIcons/Happiness"
+    const val Population = "StatIcons/Population"
+    const val CityState = "NationIcons/CityState"
+    const val Production = "StatIcons/Production"
+    const val Food = "StatIcons/Food"
+    const val Faith = "StatIcons/Faith"
+    
+    fun statToIcon(stat: Stat): String {
```
SomeTroglodyte, line 29: I know Yair doesn't like it as much as I do but this is a prime example - for enum properties. The whole function completely unnecessary and replaced by direct access to the const & interned string in bytecode.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -346,16 +352,20 @@ class MapUnit {
         return unit
     }
 
-    fun canUpgrade(): Boolean {
+    /** @param preemptively: Ignore possible tech/culture requirements. Used for upgrading units
```
SomeTroglodyte, line 355: Isn't 'to preempt' something like 'to react to an event before it even happens', Like preemptively nuking Antarctica because the Nazis could be hiding there? Or preemptively banning @xlenstra from this repo because they _might_ deliver bullshit in some future PR :innocent: ?

Maybe 'unconditionally'? I mean preempt away or even 'xyzzy' would be OK as the Kdoc is fine, but...

core/src/com/unciv/models/ruleset/Unique.kt
```diff
@@ -126,10 +312,200 @@ object UniqueTriggerActivation {
...
+                if (notification != null)
+                    civInfo.addNotification(notification, NotificationIcon.Diplomacy)
+                return true
+            }
+        }
+        return false
+    }
+
+    /** @return boolean whether an action was successfully preformed */
+    fun triggerUnitwideUnique(
```
SomeTroglodyte, line 473: Another totally unimportant naming nitpick: Unitwide'? By belly girth?

android/assets/jsons/translations/template.properties
```diff
@@ -1127,3 +1127,7 @@ in all cities with a world wonder =
 in all cities connected to capital = 
 in all cities with a garrison = 
 
+# Uniques not found in JSON files
+
+Only available after [] turns = 
+This Unit upgrades for free =
```
SomeTroglodyte, line 1133: You sure the last translation template works without a linefeed at the end of the file?

SomeTroglodyte:
The civilopediaText in TileImprovements.json for the ancient ruins will have to go. I'll do a generator instead - in time. But it may help if the ruins were no longer made special by hardcoding their name. That first line could be a unique and control the trigger?

Or - whoever maintains the ruins json always mirrors into that text...

core/src/com/unciv/models/ruleset/Unique.kt
xlenstra: Was mostly so it sounded similar to `triggerCivwideUnique`, but I can change it of course

core/src/com/unciv/logic/map/MapUnit.kt
xlenstra: I used 'preemptively' as 'upgrade to a unit before we know that unit / are allowed to upgrade', but I suppose it doesn't quite match the definition of 'preemptively'. I'll see if I can find something better

logicminimal:
does this pr make the ruins account for when you dont have any cities yet? you can get s lot of ruins before settling your city i hope it ignores the 1pop bonus in this case

xlenstra:
> does this pr make the ruins account for when you dont have any cities yet? you can get s lot of ruins before settling your city i hope it ignores the 1pop bonus in this case

yes

logicminimal:
is it possible you can generalise the free tech with [era] instead of just ancient era for modding purposes?

xlenstra:
> is it possible you can generalise the free tech with [era1, era2, era3] instead of just ancient era for modding purposes?

It is already generalized.

android/assets/jsons/Civ V - Vanilla/Ruins.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    {
+        "name": "freeGold",
+        "notification": "We have found a stash of [goldAmount] Gold in the ruins!",
+        "uniques": ["Gain [50]-[100] [Gold]"]
+    },
+    {
+        "name": "freeAncientTech",
+        "notification": "We have discovered the lost technology of [techName] in the ruins!",
+        "uniques": ["[1] free random researchable Techs from the [Ancient era]"]
```
yairm210, line 37: 'Tech(s)' perhaps?

android/assets/jsons/Civ V - Vanilla/Ruins.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    {
+        "name": "unitUpgrade",
+        "notification": "Our unit finds advanced weaponry hidden in the ruins!",
+        "uniques": ["This Unit finds advanced weapons"]
```
yairm210, line 42: I think this should be phrased as "this unit upgrades", the notification can be flavor text but I think the unique should be mechanical effect

android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -27,6 +27,7 @@
 		"cost": 25,
 		"obsoleteTech": "Scientific Theory",
 		"uniques": ["Ignores terrain cost"],
+		"specialUpgradesTo": "Archer",
```
yairm210, line 30: This should be a unique - "Can upgrade to [unitType] via random through [Ancient Ruins]" (I tried thinking of a version that was ruins-replacement-agnostic but could not 😢 )

android/assets/jsons/Civ V - Vanilla/Ruins.json
xlenstra: Problem is that there is already an unique for "this unit upgrades", which follows the normal upgrade path. I'll probably change it to "this unit upgrades including special upgrades" or something to still clearly differentiate between the two

android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -38,7 +37,7 @@
 		"cost": 40,
 		"obsoleteTech": "Metal Casting",
 		"upgradesTo": "Swordsman",
-		"specialUpgradesTo": "Spearman",
+		"uniques" : ["May upgrade to [Spearman] through ruins-like effects"],
```
SomeTroglodyte, line 40: I would have named it 'special events', but, well, why not.

android/assets/jsons/Civ V - Vanilla/Ruins.json
```diff
@@ -0,0 +1,64 @@
+[
+    {
+        "name": "freeCulture",
+        "notification": "We have discovered cultural artifacts in the ruins! (+20 culture)",
+        "uniques": ["Gain [20] [Culture]"]
+    },
+    {
+        "name": "joinWorker",
+        "notification": "A worker has joined us!",
```
yairm210, line 9: These notifications should be changed to "A [Worker] has joined us!"
If the unit unique is configurable, the notification should be so too

yairm210: This and the Settler notification, obviously

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -678,7 +687,18 @@ class MapUnit {
...
-        if (tile.improvement == Constants.ancientRuins && civInfo.isMajorCiv())
+        if (civInfo.isMajorCiv() 
+            && tile.improvement != null
+            // I don't know who came up with the idea of making starting locations improvements, 
+            // but I don't like having to make this check here.
+            // Why not just cache these somewhere for easy access? Why do they have to be saved
+            // on the map instead of in some variable? I can't imagine it being easier to
+            // find where they are this way, and this kind of random check feels way out of place.
+            // This is just misuse of an existing variable for a completely different usecase at this point.
+            // Now we have to check this every time a unit moves over an improvement, which happens _quite often_.
```
yairm210, line 698: The answer is that this needs to be in the TileMap somewhere, since when you save a map that's the data object that gets serialized.
I agree that it could be anywhere in the TileMap, I think saving it within a specific tile is the correct thing. Whether this should be an improvement or not... not sure about that.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -298,35 +298,59 @@ object UnitActions {
...
+    
+    fun getAncientRuinsUpgradeAction(unit: MapUnit): UnitAction? {
+        val upgradedUnit = 
+            unit.civInfo.getEquivalentUnit(
+                unit.civInfo.gameInfo.ruleSet.units[
+                    when {
+                        unit.baseUnit.specialUpgradesTo != null -> unit.baseUnit.specialUpgradesTo
+                        unit.baseUnit.upgradesTo != null -> unit.baseUnit.upgradesTo
+                        else -> return null
+                    }
```
yairm210, line 338: This nessting is unnecessary and makes it less readable. Please put this in a val before.

android/assets/jsons/Civ V - Vanilla/Ruins.json
xlenstra: The reason I did it this way, is that the parameter in the unique is filled when the modmaker writes the json file, and so already knows what unit will be added. If it was unknown (for example, with the 50-100 gold), then a parameter would make sense, as the true value will only be determined at runtime and so the modmaker couldn't fill it in in the json file.

yairm210: I'm saying that from a translation standpoint, this would reduce 1 translation now, and would mean that modded effects are also translated
Otherwise you need to translate for each unit separately

xlenstra: That is a good point, I'll add it in

core/src/com/unciv/logic/map/MapUnit.kt
xlenstra: Downside of making it an improvement is that in random places you need a "!tile.improvement.startsWith("StartingLocation ")" check when this is only relevant when placing startingUnits. For now I'll leave it like this, but maybe sometime in the future I'll create a PR to fix this.

SomeTroglodyte: I say - this definitely needs to go away from being a pseudo-improvement. A saved map looks like this:
`{mapParameters:{name:My new map,type:Custom,..},tileList:[{position:{x:29,y:32},baseTerrain:Grassland},...,{position:{x:-7,y:20},baseTerrain:Grassland}]}`
... so there _is_ room parallel to the tileList. I say a `class StartingPosition(val nation: String, val location: Vector2)` and a `val startingPositions = listOf<StartingPosition>()` in TileMap - ignored everywhere except game start, cleared there and for safety in save game... 'In a specific tile' is ... a misunderstanding. No such kludges necessary!

PR #4755, opened 2021-08-04
City state resources

Interdice:
This is basically my old pull request but better!

PR #4782, opened 2021-08-06
Diplomacy: City State resource UI, improvement gift effect

SomeTroglodyte:
Looking at #4755 in action I made this - careful - commit 2 is effectively a rule change, and I'm not sure the resulting rules are 'correct':

<details><summary>Demo Clip</summary>

https://user-images.githubusercontent.com/63000004/128544594-004fddca-59e5-434b-aba6-ef67d5dc17dc.mp4
</details>

- The resources have a tooltip showing name in case you don't know all the icons, or clicking leads to Civilopedia if still unsure... Of course, we still can't return from there.
- The freshly improved resource shows up immediately. This raises the question...

Current effect of this '200 Gold for improvement' gift:
- You lose 200 gold
- The CS gains 200 gold
- You gain Influence same as normal gold gifts
- The improvement is placed on the tile, but the resource is not added to the CS's pool until endTurn

Change: CS available resources updated immediately

Questionable: Should the CS really keep that gold if they also gained an improvement?

### DIFF
--- a/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
+++ b/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
@@ -20,8 +20,10 @@ import com.unciv.models.ruleset.ModOptionsConstants
 import com.unciv.models.ruleset.Quest
 import com.unciv.models.ruleset.tile.ResourceType
 import com.unciv.models.translations.tr
+import com.unciv.ui.civilopedia.CivilopediaScreen
 import com.unciv.ui.tilegroups.CityButton
 import com.unciv.ui.utils.*
+import com.unciv.ui.utils.UncivTooltip.Companion.addTooltip
 import kotlin.math.floor
 import kotlin.math.roundToInt
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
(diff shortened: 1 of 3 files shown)


logicminimal:
@ravignir can you help about this?

ravignir:
-  You lose 200 gold - yup
  -  The CS gains 200 gold - nope - and as a side note - any gold given to CS vanishes into void.
  - You gain Influence same as normal gold gifts - nope!
  - The improvement is placed on the tile, but the resource is not added to the CS's pool until endTurn - it is instantly added

logicminimal:
> 
>   -  You lose 200 gold - yup
>   -  The CS gains 200 gold - nope - and as a side note - any gold given to CS vanishes into void.
>   - You gain Influence same as normal gold gifts - nope!
>   - The improvement is placed on the tile, but the resource is not added to the CS's pool until endTurn - it is instantly added
> 

kinda bad that the city state doesnt receive that gold

SomeTroglodyte:
I think I may be to blame for the CS actually receiving the gold subtracted from the player for the older normal gifts 'cause I couldn't imagine it vanishing being correct. I mean 'hey here's ten grand' you'd expect the recipient to go at least get a new bike, right? Personal expectations. With you, I trust it's not but hard facts...

But specifically for the gift improvement case, OK it should vanish and no influence changed. That I can code. The other gifts I'll leave untouched for now.

SomeTroglodyte:
OK, that should do it. Also found any work-in-progress on the receiving tile still needed cancelling. Which begs the question - if they're building a farm on incense because they want to and it's 1 turn from completion won't they be peeved when I throw away that work and 'magic' a plantation up there? Also - wouldn't that freed-up worker need new instructions?

ravignir:
I can see why it seemed like a correct behavior as it was impossible to check without using IGE.
Personally, I have nothing against City States keeping the gifted gold from regular gifts as it seems logical - but I can see why it vanishes in civ5: in late game CS tend to be gifted a lot of gold from many civilizations.

SomeTroglodyte:
Yes, true enough, current master can produce this: 
image

>  in late game CS tend to be gifted a lot of gold from many civilizations

Well then we should teach them that with too much gold they should start buying nukes and _use_ them. Say, at 100k? Over 40 N missiles. Won't some people be surprised... Another use for the 'personality'.

logicminimal:
> I can see why it seemed like a correct behavior as it was impossible to check without using IGE.
> Personally, I have nothing against City States keeping the gifted gold from regular gifts as it seems logical - but I can see why it vanishes in civ5: in late game CS tend to be gifted a lot of gold from many civilizations.

just make the amount it receives scale down with era

logicminimal:
> Yes, true enough, current master can produce this: 
> image
> 
> >  in late game CS tend to be gifted a lot of gold from many civilizations
> 
> Well then we should teach them that with too much gold they should start buying nukes and _use_ them. Say, at 100k? Over 40 N missiles. Won't some people be surprised... Another use for the 'personality'.

i would actually want this to be in the game, its bad that the game iw unable to get any updates outside civ 5 features

ravignir:
Well, some civs prefer diplomatic victory (which requirs donating to city states) over domination.

logicminimal:
i want to suggest one thing, from a peek to the code on the previous pr that this pr changes it checks whether the  influence between the cs and you is below 60, not if you are allies, this is a bug

SomeTroglodyte:
> checks whether the influence

Nope, that part hasn't been changed by this PR. This is UI and final effect only.

PR #4784, opened 2021-08-06
Cached getTerrainFeatures

### LINKED ISSUE #4709: Hill deprecation shows crash loophole in `getLastTerrain()`

SomeTroglodyte:
May resolve #4709 .

Please review thoroughly.

Rationale: We have `removeMissingModReferences` but in that issue I managed to run into `isImpassible()` being called on a tile with terrainFeatures missing from the ruleset. I can't remember how exactly, except running a save from one of the issues here was involved which has Civ5ExpansionMod where Hill still is base terrain.

So first question - do we do defense-in-depth and harden `getTerrainFeatures` or do we prefer to find the loophole where `removeMissingModReferences` did not run in time?

While some patching of `convertHillToTerrainFeature` or `isImpassible` or so might help, I thought that mapping operation (feature string to Terrain object) runs every time `getTerrainFeatures` runs which may be often, and the mapping may be more expensive than checking a cache for being outdated. If that is the case, then that could be leveraged to remove the marginal weakness in getLastTerrain...

But I'm not entirely convinced `terrainFeatures.hashCode()` is cheaper than `terrainFeatures.mapNotNull { ruleset.terrains[it] }` at least not necessarily by a worthwhile margin. Your opinion(s)?


SomeTroglodyte:
Brought up to date and simplified - I still think it's an improvement

yairm210:
Only case I can think of where terrain features change mid-turn is when 
- A-Bombing tiles
- Creating Great Improvements on tiles
In these situations we'll need to update the terrain feature objects

SomeTroglodyte:
> terrain features change mid-turn

True! The earlier design would have caught that, before commit [99e3a27](https://github.com/SomeTroglodyte/Unciv/commit/99e3a270fd84dbe161c1f95818af6609ed98102c) - not sure what to do with this, then. Revert that commit, ditch the idea, or make additional update transients calls from those places?

yairm210:
If you're going for transients, then updating those transients is the way to go.
I would personally turn the terrainFeatures into a non-editable List and add "addTerrainFeature" and "removeTerrainFeature" functions to both add/remove from list and to update the terrainFeature objects (this can be done by iterating on all of them, using the same updateTerrainFeatureObjects() function every time, it's quite rare anyway)

SomeTroglodyte:
OK, way forward unterstood. This might take some mulling over and maybe some analysis. Let's keep this on hold for a while.

SomeTroglodyte:
Measurements over three turns of a large-map game, in nanosecs:
countSetTransients=16227, timeSetTransients=4214803
countGetLastTerrain=2676391, timeGetLastTerrain=117164097, timeGetLastTerrainOld=254704670
countGetAllTerrains=1954280, timeGetAllTerrains=149024420, timeGetAllTerrainsOld=454436628

So, not worth the overhead.

PR #4788, opened 2021-08-06
Common interface for Unique-collections in ruleset objects

SomeTroglodyte:
Pure preparation, to keep the meaty PR's smaller.

- IHasUniques as way to declare which ruleset object have uniques and uniqueObjects
    - Minor problem: types must now match, so duplicate nation uniques are no longer silently merged
- Deprecating CivilopediaText in favour of ICivilopediaText - too little gain of a pure convenience thingy
- Repairs all those trailing blanks in templates
    - That was easier done by copying from a current local branch so I left a string in I'll use in the near future
- Tutorial Strings wrapping moved, so the other side in CivilopediaScreen is one step closer to using a loop (in a local branch the loop already works fine)


xlenstra (approved):

yairm210:
Conflicts

SomeTroglodyte:
> Conflicts

Yup, late repercussions of the extra blanks #4755 put into templates. Not even part of this PR. Won't be a Problem much longer, once all PR's started before the fix are gone...

PR #4789, opened 2021-08-06
Civilopedia - Difficulty

SomeTroglodyte:
Last one for today. Changes to the difficulties display in Civilopedia - tiny but I think worth it because they no longer stand out so much. A tad easier to understand, too, without kicking all translations into the bin - except the one that was too misleading.

May clash with #4788 - thus draft level for now.


core/src/com/unciv/models/ruleset/Difficulty.kt
```diff
@@ -31,36 +34,46 @@ class Difficulty: INamed {
...
-        lines += ""
-        lines += "{Turns until barbarians enter player tiles}: $turnBarbariansCanEnterPlayerTiles"
-        lines += "{Gold reward for clearing barbarian camps}: $clearBarbarianCampReward"
 
-        return lines.joinToString("\n") { it.tr() }
+    override var civilopediaText = listOf<FormattedLine>()
+
+//    fun getDescription(): String {
+//        return getCivilopediaTextLines(RulesetCache.getBaseRuleset()).joinToString("\n") { it.text.tr() }
+//    }
```
yairm210, line 42: Clean dead lines :)

core/src/com/unciv/models/ruleset/Difficulty.kt
```diff
@@ -31,36 +34,46 @@ class Difficulty: INamed {
...
+        lines += FormattedLine("AI settings", header = 3)
+        lines += FormattedLine("{AI city growth modifier}: ${aiCityGrowthModifier.toPercent()}% ${Fonts.food}", indent = 1)
+        lines += FormattedLine("{AI unit cost modifier}: ${aiUnitCostModifier.toPercent()}% ${Fonts.production}", indent = 1)
+        lines += FormattedLine("{AI building cost modifier}: ${aiBuildingCostModifier.toPercent()}% ${Fonts.production}", indent = 1)
+        lines += FormattedLine("{AI wonder cost modifier}: ${aiWonderCostModifier.toPercent()}% ${Fonts.production}", indent = 1)
+        lines += FormattedLine("{AI building maintenance modifier}: ${aiBuildingMaintenanceModifier.toPercent()}% ${Fonts.gold}", indent = 1)
+        lines += FormattedLine("{AI unit maintenance modifier}: ${aiUnitMaintenanceModifier.toPercent()}% ${Fonts.gold}", indent = 1)
+//        lines += FormattedLine("{AI free techs}: $aiFreeTechs", indent = 1)
+//        lines += FormattedLine("{AI major civilizations starting units}: $aiMajorCivStartingUnits", indent = 1)
+//        lines += FormattedLine("{AI city-state starting units}: $aiCityStateStartingUnits", indent = 1)
```
yairm210, line 71: Why not?

core/src/com/unciv/models/ruleset/Difficulty.kt
SomeTroglodyte: Ah, I abused this as test bed. Sorry. Out they go.

core/src/com/unciv/models/ruleset/Difficulty.kt
SomeTroglodyte: Was commented out before, I just moved the block and did a replace on them all so they look like I re-created them but actually - not. We can go look for the why and what additional info would be OK later.

SomeTroglodyte: Actually, I'm trying the unlisted properties now.

SomeTroglodyte:
Umm...
image
...that would be complete. Shows a basic difficulty the 'pedia text render has - lines with icons are higher than the same ones without. So far it hasn't been this eye-catching (was worse before I reordered the json), and I'm still waiting for the eureka moment telling me how to deal with it.

Do we want this? Hmmm.... OK, pushed. Should we do anything while we're at it to further clarify the modifiers? growth 60% = needs 60% of normal food surplus to grow etc? I could add a tooltip capability to FormattedLine...

logicminimal:
> Umm...
> image
> ...that would be complete. Shows a basic difficulty the 'pedia text render has - lines with icons are higher than the same ones without. So far it hasn't been this eye-catching (was worse before I reordered the json), and I'm still waiting for the eureka moment telling me how to deal with it.
> 
> Do we want this? Hmmm.... OK, pushed. Should we do anything while we're at it to further clarify the modifiers? growth 60% = needs 60% of normal food surplus to grow etc? I could add a tooltip capability to FormattedLine...

i think i have a few suggestions. on player settings on that civilopedia page add a big population icon to the left of those values (probably a bad idea so you can just add it to the left of player settings :population: Player settings
and for AI you can choose a random civ and choose that civs image to add to the left of the AI settings

another thing is i think you put something like 12 spaces (3 tabs) worth of spaces to it which i think looks unpleasant, could you please make it something like 4 spaces?

another thing is there any way to remove the space between tech and unique? like the 📎🏺Pottery

edit: on another thought these are already spaced 1 tabs wide, it's just that the default text is tabbed 3 tabs wide

SomeTroglodyte:
> you put something like 12 spaces

Must be the 5th time I explain this: That's indent = 1 or 'first tabstop'. It *must* stay to the right of all line icons so uniques can stay aligned. I provisioned for 3 icons and standard padding. May be changed to 2 icons but not in this PR.

PR #4793, opened 2021-08-07
Added follower beliefs for buying religious buildings

xlenstra:
Addes follower uniques for buying cathedrals, monasteries, mosques and pagodas.
Changes the current monastery into a building that can only be bought with faith. In save files where the building is built, this might be weird, but should work fine. In most cases, the output of the building only increases.
Also disables the check that a building must have either a cost or reference a tech to assign it a cost, as long as this building is "unbuildable" (i.e. must be purchased), as in those cases its cost does not matter and may well be 0.


core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -355,7 +382,7 @@ class Building : NamedStats(), INonPerpetualConstruction, ICivilopediaText {
 
     override fun getStatBuyCost(cityInfo: CityInfo, stat: Stat): Int? {
         var cost = getBaseBuyCost(cityInfo, stat)?.toDouble()
-        if (cost == null) return null
+        if (cost == null) { println(name); return null}
```
SomeTroglodyte, line 385: I have a little project here just cleaning up console output... I think we should try to:
- Have all debug output gated. Where and how? Uh, kotlin refuses to implement conditional compilation - argument if (const val Boolean) will optimize unused code away - that would mean the way I did it in Sounds is OK (except I left it on after it served its purpose).
- No println() - looks ugly in Logcat output
- No println(name) without some context hint...

Dumb idea?

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -302,7 +302,7 @@ class Ruleset {
         }
 
         for (building in buildings.values) {
-            if (building.requiredTech == null && building.cost == 0)
+            if (building.requiredTech == null && building.cost == 0 && !building.uniques.contains("Unbuildable"))
                 lines += "${building.name} must either have an explicit cost or reference an existing tech!"
```
SomeTroglodyte, line 306: This is in mod checker, right? No reason not to update the message, it's untranslated anyway if I remember correctly.

SomeTroglodyte (approved):
Diff inspection looks good. Adapting Civilopedia can come later.

PR #4797, opened 2021-08-07
Added Architecture as a prerequiste of Archaeology

xlenstra:
#4796 prompted me to check all other connections between techs, as maybe we had missed one somewhere.
This was the only one I could find, so I've added it in in this PR


logicminimal:
i too did find thid but i wasnt sure whether it was a bnw one or not, thanks for pr

PR #4804, opened 2021-08-07
Unit rename UI

SomeTroglodyte:

From:
image
To:
image

And from:
image
To:
image
</details>

Note the options pairs are same width now.

Mysterious how that `.apply { color = BLUE }` got in there - the function is my fault, but why did I place that? Inconsistent looks, should go, if someone wants a dark blue button they can apply afterwards. not in framework code...

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/RenameUnitPopup.kt
+++ b/core/src/com/unciv/ui/pickerscreens/RenameUnitPopup.kt
@@ -1,33 +1,32 @@
 package com.unciv.ui.pickerscreens
 
+import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.scenes.scene2d.ui.TextField
 import com.unciv.logic.map.MapUnit
-import com.unciv.ui.utils.CameraStageBaseScreen
-import com.unciv.ui.utils.Popup
-import com.unciv.ui.utils.onClick
-import com.unciv.ui.utils.toTextButton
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/ui/pickerscreens/RenameUnitPopup.kt
```diff
@@ -1,33 +1,32 @@
...
-        row()
-        add("OK".toTextButton().onClick {
-            if (nameField.text != "") {
+        // Disallowed special characters - used by tr() or simply precaution
+        val illegalChars = "[]{}\"\\<>"
+        nameField.textFieldFilter = TextField.TextFieldFilter { _, char -> char !in illegalChars}
+        nameField.maxLength = unit.civInfo.gameInfo.ruleSet.units.values.maxOf { it.name.length }
+        add(nameField).growX().colspan(2).row()
+        addOKButton {
+            if (nameField.text != "" && nameField.text != unit.name) {
```
xlenstra, line 23: So you can't name a unit back to it's original name if you have already changed it once?

SomeTroglodyte:
That detail I noticed (Idea to pre-fill and pre-select the input) - but I didn't change that. You're not allowed into the dialog if `instanceName` is set, so this `&&` actually increases your chances.

xlenstra (approved):

AdityaMH:
Apply "blue" color for hover text

PR #4808, opened 2021-08-08
Added an overview screen for religions

xlenstra:
This PR adds an overview screen for religions.
In this, you can see some basic stats for the religions currently founded, and what uniques they have active.
This screen is opened when clicking the Faith icon in the top bar of the world screen.
<details> <summary> Pictures: </summary>

Religion with only a pantheon use the nation Icon, as it cannot be spread outside of the nation anyway:
image

Real religions use the icon of the religion:
image

</details>

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -993,7 +993,6 @@ Religions and Beliefs =
 # Religion overview screen
 Religion Name: = 
 Founding Civ: = 
-??? = ???
 Cities following this religion: = 
 Click an icon to see the stats of this religion = 
 
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -827,7 +827,7 @@ Spaceship parts remaining =
 Branches completed = 
 Undefeated civs = 
  # The \n here means: put a newline (enter) here. If this is omitted, the sidebox in the diplomacy overview will become _really_ wide. 
- # Feel free to replace it with a space and put it somewhere else in your translation 
+ # Feel free to replace it with a space and put it between other words in your translation
 Turns until the next\ndiplomacy victory vote: [amount] = 
 Choose a civ to vote for = 
 Choose who should become the world leader and win a diplomatic victory! = 
@@ -990,6 +990,13 @@ Found Pantheon =
 Follow [belief] = 
 Religions and Beliefs = 
          
+# Religion overview screen
+Religion Name: = 
+Founding Civ: = 
+??? = ???
+Cities following this religion: = 
+Click an icon to see the stats of this religion = 
+
 # Terrains 
          
 Impassable = 
--- a/core/src/com/unciv/logic/civilization/ReligionManager.kt
+++ b/core/src/com/unciv/logic/civilization/ReligionManager.kt
@@ -137,7 +137,6 @@ class ReligionManager {
             return false
         // Mod maker did not provide enough religions for the amount of civs present
         
-        
         return true
     }
 
@@ -171,6 +170,13 @@ class ReligionManager {
         holyCity.religion[name] = 100
 
         foundingCityId = null
+    }
+    
+    fun citiesFollowingThisReligionCount(): Int {
+        if (religion == null) return 0
+        return civInfo.gameInfo.getCities()
+            .filter { it.religion.getMajorityReligion() == religion!!.name }
+            .count()
     }
 }
 
--- a/core/src/com/unciv/logic/civilization/ReligionManager.kt
+++ b/core/src/com/unciv/logic/civilization/ReligionManager.kt
@@ -172,11 +172,10 @@ class ReligionManager {
         foundingCityId = null
     }
     
-    fun citiesFollowingThisReligionCount(): Int {
+    fun numberOfCitiesFollowingThisReligion(): Int {
         if (religion == null) return 0
         return civInfo.gameInfo.getCities()
-            .filter { it.religion.getMajorityReligion() == religion!!.name }
-            .count()
+            .count { it.religion.getMajorityReligion() == religion!!.name }
     }
 }
 
--- a/core/src/com/unciv/models/Religion.kt
+++ b/core/src/com/unciv/models/Religion.kt
@@ -16,8 +16,7 @@ class Religion() : INamed {
 
     var founderBeliefs: HashSet<String> = hashSetOf()
     var followerBeliefs: HashSet<String> = hashSetOf()
-
-
+    
     @Transient
     lateinit var gameInfo: GameInfo
 
@@ -60,6 +59,12 @@ class Religion() : INamed {
             .asSequence()
     }
     
+    fun getFounderBeliefs(): Sequence<Belief> {
+        return mapToExistingBeliefs(founderBeliefs)
+            .filter { it.type == BeliefType.Founder }
+            .asSequence()
+    }
+    
     private fun getUniquesOfBeliefs(beliefs: HashSet<String>): Sequence<Unique> {
         return mapToExistingBeliefs(beliefs)
             .map { it.uniqueObjects }
@@ -80,7 +85,6 @@ class Religion() : INamed {
     }
 
     fun isMajorReligion(): Boolean {
-        if ("" in followerBeliefs) return true // Temporary as a result of follower beliefs not yet being implemented
         return founderBeliefs.isNotEmpty() && followerBeliefs
             .any { gameInfo.ruleSet.beliefs[it]!!.type == BeliefType.Follower}
     }
--- a/core/src/com/unciv/models/Religion.kt
+++ b/core/src/com/unciv/models/Religion.kt
@@ -49,26 +49,25 @@ class Religion() : INamed {
 
     fun getPantheonBeliefs(): Sequence<Belief> {
         return mapToExistingBeliefs(followerBeliefs)
-            .filter { it.type == BeliefType.Pantheon }
             .asSequence()
+            .filter { it.type == BeliefType.Pantheon }
     }
     
     fun getFollowerBeliefs(): Sequence<Belief> {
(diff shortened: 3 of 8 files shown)


logicminimal:
is it possible you can add their names on top of the religions? (and the pantheons if possible) for better ui

xlenstra:
> is it possible you can add their names on top of the religions? (and the pantheons if possible) for better ui

Names of the religion already shows when you select them, pantheons show up using the nation icon.

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -172,6 +171,13 @@ class ReligionManager {
 
         foundingCityId = null
     }
+    
+    fun citiesFollowingThisReligionCount(): Int {
+        if (religion == null) return 0
+        return civInfo.gameInfo.getCities()
+            .filter { it.religion.getMajorityReligion() == religion!!.name }
+            .count()
```
yairm210, line 179: .count {...}

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -172,6 +171,13 @@ class ReligionManager {
 
         foundingCityId = null
     }
+    
+    fun citiesFollowingThisReligionCount(): Int {
```
yairm210, line 175: numberOfCitiesFollowingThisReligion, perhaps?

core/src/com/unciv/models/Religion.kt
```diff
@@ -60,6 +59,12 @@ class Religion() : INamed {
             .asSequence()
     }
     
+    fun getFounderBeliefs(): Sequence<Belief> {
+        return mapToExistingBeliefs(founderBeliefs)
```
yairm210, line 63: .asSequence() before the .filter()

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
```diff
@@ -0,0 +1,117 @@
...
+import kotlin.math.min
+
+class ReligionOverviewTable(
+    private val viewingPlayer: CivilizationInfo,
+    private val overviewScreen: EmpireOverviewScreen
+): Table() {
+    
+    val gameInfo = viewingPlayer.gameInfo
+    private val topButtons = Table(CameraStageBaseScreen.skin)
+    private val topButtonLabel = "Click an icon to see the stats of this religion".tr().toLabel()
```
yairm210, line 22: .toLabel() includes .tr() internally

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
```diff
@@ -0,0 +1,117 @@
...
+        topButtonLabel.setText(religion.name.tr())
+        for (belief in religion.getPantheonBeliefs()) {
+            beliefsTable.add(createBeliefDescription(belief)).pad(10f).row()
+        }
+        for (belief in religion.getFollowerBeliefs()) {
+            beliefsTable.add(createBeliefDescription(belief)).pad(10f).row()
+        }
+        for (belief in religion.getFounderBeliefs()) {
+            beliefsTable.add(createBeliefDescription(belief)).pad(10f).row()
+        }
```
yairm210, line 87: Sounds like `for(belief in religion.getPantheonBeliefs() + ... + ...)`

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
```diff
@@ -0,0 +1,117 @@
...
+        beliefsTable.clear()
+        clear()
+
+        addReligionButtons()
+        if (religion != null) loadReligion(religion)
+        add(topButtons).pad(5f).row()
+        add(topButtonLabel).pad(5f)
+        addSeparator()
+        add(statsTable).pad(5f).row()
+        add(beliefsTable).pad(5f)
```
yairm210, line 44: Why are you resetting everything and not just resetting the parts in loadReligion?

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
```diff
@@ -0,0 +1,117 @@
...
+    private val topButtonLabel = "Click an icon to see the stats of this religion".tr().toLabel()
+    private val statsTable = Table(CameraStageBaseScreen.skin)
+    private val beliefsTable = Table(CameraStageBaseScreen.skin)
+    private var selectedReligion: String? = null
+    
+    init {
+        draw(null)
+    }
+    
+    private fun draw(religion: Religion?) {
```
yairm210, line 31: Give this function a better name :)
displayReligionInfo, perhaps?

SomeTroglodyte:
Good I didn't start on this, too. The first thing to criticize that I saw was the new file not ending in a LF... Goes to show. Also shows leftovers I thought I'd cleaned long ago. Like `Char.MIN_VALUE` is no longer the proper default for a 'no-op' key binding. A few lines out of the diff display somewhere...

android/assets/jsons/translations/template.properties
```diff
@@ -990,6 +990,13 @@ Found Pantheon =
 Follow [belief] = 
 Religions and Beliefs = 
          
+# Religion overview screen
+Religion Name: = 
+Founding Civ: = 
+??? = ???
```
yairm210, line 996: This really doesn't need translation. I've been using a single question mark for the random for years and translators have picked up on all kinds of really esoteric things, but the question mark appears to be universal.

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
SomeTroglodyte: Which is why I push Kdoc - says so right there in the tooltip.

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
SomeTroglodyte: update-something is already often used for similar usecases

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
SomeTroglodyte: + being efficient because they're sequences, some other Collection.plus() do copies.

android/assets/jsons/translations/template.properties
SomeTroglodyte, line 996: Love it. Tempts to slip in a nice translation for that somewhere... `??? = ❓❓❓` would probably show just fine.

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
```diff
@@ -0,0 +1,117 @@
...
+        
+        val minWidth = min(statsTable.minWidth, beliefsTable.minWidth) + 5f
+        
+        statsTable.width = minWidth
+        for (cell in beliefsTable.cells) {
+            cell.minWidth(minWidth)           
+        }
+    }
+    
+    private fun createBeliefDescription(belief: Belief): Table {
```
SomeTroglodyte, line 97: Other ruleset objects, no longer all, have their getDescription(). If one _really_ wants to have "describe me" logic in one place only, a `getCivilopediaTextLines(ruleset).joinToString { it.text.tr() }` would mostly work. Not the first time I thought about a bitmask 'detailLevel' parameter...

*here* however I would _include_ the civilopedia output as it is, using the original renderer, with or without linking is a matter of taste.

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
```diff
@@ -0,0 +1,106 @@
...
+import com.unciv.logic.civilization.CivilizationInfo
+import com.unciv.models.Religion
+import com.unciv.models.ruleset.Belief
+import com.unciv.models.translations.tr
+import com.unciv.ui.utils.*
+import kotlin.math.min
+
+class ReligionOverviewTable(
+    private val viewingPlayer: CivilizationInfo,
+    private val overviewScreen: EmpireOverviewScreen
```
SomeTroglodyte, line 15: `@Suppress("UNUSED_PARAMETER")       // Keep all OverviewScreen Pages compatible` if you want - 'Problems' page of Studio immediately clear. Copied from resource sibling.

yairm210:
Smash is definitely a better UI programmer than I am, so I'll merge this on his OK :)

core/src/com/unciv/ui/overviewscreen/ReligionOverviewTable.kt
SomeTroglodyte: Try 
```kotlin
    private fun createBeliefDescription(belief: Belief) =
        MarkupRenderer.render(
            belief.run { sequence {
                yield(FormattedLine(name, size = 24, link = makeLink()))
                yield(FormattedLine())
                yieldAll(getCivilopediaTextLines(gameInfo.ruleSet))
            } }.toList(),
            noLinkImages = true
        ) {
            UncivGame.Current.setScreen(CivilopediaScreen(gameInfo.ruleSet, link = it))
        }.apply {
            background = ImageGetter.getBackground(ImageGetter.getBlue())
        }
```
with noLinkImages as is or without. The whole thing could be simpler without linking, and maybe I should wrap what the CivilopediaScreen does anyway _somewhere_ - but that's for later. Looks similar enough on screen. Just food for thought.

SomeTroglodyte: <details><summary>omitting noLinkImages that looks like:</summary>

image
</details>

SomeTroglodyte:
> merge this

No objections. That civilopedia thing can come later. Maybe... Group the Beliefs. Once you can have more than so few of them.

PR #4809, opened 2021-08-08
Deprecate all mods without an eras.json file

xlenstra:
This PR enforces use of a well-defined eras.json file and removes the continuous checks for this in game logic.
Feel free to still wait some time before merging this PR, as it will break many mods, including some of the most popular ones.


SomeTroglodyte:
Shall we @ mention the >= 10 stars authors? ravignir, SpacedOutChicken, k4zoo, Subcher?

yairm210:
This will make many existing mods entirely unloadable.
The problem is, many of the top ranked mods are 'abandoned' mods that won't be changed.
Before we do this, we should figure out a way to remove such mods from the mod list. I have a couple of ideas.

A. On MAJOR BREAKING updates like this one, we can *switch the tag* that we're looking for. From "unciv-mod" to "unciv-mod-2", etc. Advantage - real simple. Disadvantage - all mod owners will have to make the tag switch.

B. Compile a "broken mods" file by loading each mod, and checking the ruleset errors. The disadvantages are obvious, but also has some advantages if we manage to do this well. This is a *cooler* idea but not probably not a *better* idea given the problem we're coming to solve.

ravignir:
are non-base ruleset mods affected by this?

xlenstra:
> are non-base ruleset mods affected by this?

Only if they depend on another mod that is a base ruleset mod without an eras.json, so most of them don't

xlenstra:
Option C:
Whenever a mod is downloaded, we only then check whether it is still compatible with the current game or not. On an error, the user is told this mod is outdated, we remove the mod and we cache this as an invalid mod in some file somewhere. Whenever we then later load the list of mods again, we remove mods in that file from the list.
Advantages:
- We don't have to check the entire list at once, yet still make sure that only valid mods are downloaded
- No extra work for us or for mod makers
Disadvantage:
- This might be weird for the user, first they download a mod and only then are told the mod is incompatible
- Mods depending on other mods (such as Ravignir's religion mod) are more difficult to test for compatibility, and might require another modOption or some other workaround
- If the mod is later updated, we need some way to reinclude it in the list

SomeTroglodyte:
I keep thinking we should offload some responsibility to the mod authors - a declarative "compatible with" to help with the "two Mexicoes" for starters. Could something similar help here? Problem is, the mod manager list-building *really* can't do a network roundtrip extra per mod - unless it's threaded, and does not use the github _API_ (I think raw content does not count). I've got the code stashed somewhere for a threaded get-splashscreen that would then pop in when it's ready (and get pushed up the queue when the user looks at the mod) - it _is_ a noticeable delay. But - doable. And then we would _have_ the modoptions file and plonk it in the cache folder for next time, discard if the query says some update happened (still can't differentiate between content update and the owner doing _anything_ else to the repo)...

yairm210:
This is a problem that's becoming bigger as time passes, so I think for now a "quick fix" is enough to start with, and then we can think of improving that later.
The quickest fix I can think of is adding a new configuration (aka "oh no not another json file") containing "blocked" mods (by their github url, perhaps, to make it explicit). These would simply not be shown together with the other mods, and therefore will not be downloaded.

There are 2 components to this:
A. This allows version-specific-ness if that becomes an issue
B (and more importantly). This allows *community* participation.

The more abstract problem we're trying to solve here is: how do I let *mod owners* determine that their content can be accessed initially, while allowing the *playerbase* to optionally block it when not relevant?

xlenstra:
For now I've added a separate file with a few of the most popular mods that are outdated to the point of making workers unable to build improvements. Later we can then look for a more general solution to this problem

yairm210:
Okay!
I have no doubt we'll encounter other era-related problems after this change, but at can dream with them gradually

PR #4810, opened 2021-08-08
Minor update to which unit actions are hidden

xlenstra:
This PR makes minor changes to unit actions, namely hiding the 'automate' action as was intended in the original PR, and unhiding the 'stop exploring' unit action.

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -268,8 +268,6 @@ class MapUnit {
 
     fun isMoving() = action?.startsWith("moveTo") == true
     
-    fun isAutomaticallyBuildingImprovements() = action != null && action == Constants.unitActionAutomation
-
     fun isCivilian() = baseUnit.isCivilian()
 
     fun getFortificationTurns(): Int {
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
I'll rework this soon anyway. I want the keyboard bindings of hidden actions to _work_ if they don't conflict - like they did before.
`isAutomaticallyBuildingImprovements` is staying, maybe renamed, all UnitAction-specific strings in Constants are dupes of the enum's values and those need merging*. And so on. To solve the key bindings, I think the actions will get a 'page number'...

No reason to reject this fine PR.

*: You think the compiler is clever enough to output the same byte code for `action == Constants.thingy` and for `action == UnitActionType.Thingy.value`? If not, the value needs to read the const; if yes I'll kill those const's. Already compiles & runs somewhere around here.

xlenstra:
> *: You think the compiler is clever enough to output the same byte code for `action == Constants.thingy` and for `action == UnitActionType.Thingy.value`? If not, the value needs to read the const; if yes I'll kill those const's. Already compiles & runs somewhere around here.

Tbh I have no idea how good the Kotlin compiler really is at stuff like this, but I feel like things like this are probably so minor that reducing data duplication between the enum and the const is the more important aspect here

SomeTroglodyte:
Must - pay - attention - not - to - revert - this - in - #4815 :)

PR #4815, opened 2021-08-09
Unit action constants and worker unique cleanup

SomeTroglodyte:
- Use existing enum properties instead of unit action constants
- Cache Worker unique detection (likely still worth it after workerUnique goes)
- Takes care of one "This is really not modular" comment
- Fixes defeat alert popping twice (captureCivilianUnit + puppetCity) in specific conditions

~~At least 1 merge artifact, so draft only for now. Make that two~~


core/src/com/unciv/ui/worldscreen/AlertPopup.kt
```diff
@@ -247,8 +247,8 @@ class AlertPopup(val worldScreen: WorldScreen, val popupAlert: PopupAlert): Popu
...
     }
 
-    override fun close() {
-        worldScreen.viewingCiv.popupAlerts.remove(popupAlert)
-        super.close()
-    }
+//    override fun close() {
+//        worldScreen.viewingCiv.popupAlerts.remove(popupAlert)
+//        super.close()
+//    }
```
xlenstra, line 253: Commented out lines?

xlenstra (approved):

core/src/com/unciv/ui/worldscreen/AlertPopup.kt
SomeTroglodyte: Oh sorry - that defeat alert popping up twice... Wasn't sure whether treating popupAlerts as a _true_ queue would help, but no, it was actually added twice after all. Will need to revert this and that `first/pop` one. Or not?

SomeTroglodyte:
Yes, I suppose it's safer to leave the alerts in the queue until actively dismissed, and - 'ncars'.

yairm210:
I'm ready to merge this, just resolve conflicts :)

SomeTroglodyte:
What a coincidence: I just noticed, too. Let's see if I can do that with this _other_ Android Schtuhdioh...

SomeTroglodyte:
Diff looks OK (and reveals _this_ A.Studio still auto-removes trailing spaces), a short playtest was OK too.

PR #4818, opened 2021-08-09
.

TheWrath666:
The remaining units in the capital of the country we destroyed automatically became our own unit


yairm210:
Um
A. It looks like you based off of a test branch that has very little to do with the base game
B. This is not how the mechanics in Civ V work, therefore not something where going to introduce to Unciv :)

PR #4819, opened 2021-08-09
Fixed crash when selecting certain buildings in the civilopedia

xlenstra:
Also hides the new religious buildings when religion isn't enabled

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -1125,14 +1125,14 @@
 		"culture": 3,
 		"happiness": 1,
 		"specialistSlots": {"Artist": 1},
-		"uniques": ["Unbuildable"]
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
... did ***I*** code that 'instantiate empty City just to be able to call `canBePurchasedWithStat`' in the Civilopedia provider? Now I think `canBePurchasedWithStat` should be `fun (CityInfo?, Stat)` instead of `fun (CityInfo, Stat, Boolean)`...

xlenstra:
> ... did _**I**_ code that 'instantiate empty City just to be able to call `canBePurchasedWithStat`' in the Civilopedia provider? Now I think `canBePurchasedWithStat` should be `fun (CityInfo?, Stat)` instead of `fun (CityInfo, Stat, Boolean)`...

Was probably me, but I did it to fix a merge with your code. Still, your idea is probably better, I'll implement it in future PR

PR #4822, opened 2021-08-09
City screen purchasable constructions prettier

SomeTroglodyte:
Started out with the intention to "only" hide cathedrals from pedia in non-religion games... 
image


logicminimal:
can you also change the color of can only be purchased with faith to blue-light blue while you are at it? (faith color)

SomeTroglodyte:
OK, now supports mods where a building can be built normally _and_ bought with something other than gold. A unique specifying 'can be bought with gold' for buildable stuff is overridden by rush cost anyway.

SomeTroglodyte:
Still begs the question whether "Religious Unit" isn't an internal-only thing as well, whether we need a _system_ to classify uniques globally into 'can be shown to user' or not - and whether we shouldn't refactor 'Rejection reason' string parsing first.

PR #4827, opened 2021-08-10
Added founder beliefs, updates to pantheon spreading

xlenstra:
This PR implements a few simple founder beliefs, with the rest requiring followers to be implemented
It also updates pantheons a bit to rectify a few implementation liberties I had taken earlier:
- When founding a new city while your main religion is a pantheon, the religion automatically takes the pantheon as its main religion
- You cannot use prophets or missionaries to spread pantheon religions
- 
Lastly, it changes the way the amount of followers in a city are saved. This means that `CityReligionManager` no longer inherits from `Counter`, but instead has a variable of type `Counter` containing these pressures. The main reason for this, is that for multiple reasons it is useful to know when a religion switches inside a city (some founder beliefs, adding notification, maybe more later), which is almost impossible when you have no control over the `add` function as it is inherited.
The downside of this change, is that it breaks all existing saves, which should be fine as people using religion have agreed to that by activating the option.


android/assets/jsons/Civ V - Vanilla/Beliefs.json
```diff
@@ -175,5 +175,33 @@
...
+    },
+    {
+        "name": "Church Property",
+        "type": "Founder",
+        "uniques": ["[+2 Gold] for each global city following this religion"]
+    },
+    {
+        "name": "Initiation Rites",
+        "type": "Founder",
+        "uniques": ["[+100 Gold] when a city adopts this religion for the first time (modified by game speed)"]
```
SomeTroglodyte, line 194: Makes you think how that (..) part could be avoided

android/assets/jsons/translations/template.properties
```diff
@@ -1151,4 +1155,4 @@ in all cities with a garrison =
 
 Only available after [] turns = 
 This Unit upgrades for free = 
-
+[stats] when a city adopts this religion for the first time =
```
SomeTroglodyte, line 1158: See, doesn't match - and aren't uniques picked up automaniacally?

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -57,7 +119,11 @@ class CityInfoReligionManager: Counter<String>() {
         for (city in allCitiesWithin10Tiles) {
             val majorityReligionOfCity = city.religion.getMajorityReligion()
             if (majorityReligionOfCity == null) continue
-            else add(majorityReligionOfCity, 6) // todo - when holy cities are implemented, *5
+            else addPressure(
+                majorityReligionOfCity, 
+                if (city.isHolyCity()) 11
```
SomeTroglodyte, line 124: So the comment was wrong about 6 **`*`** 5, as this is `+` - OK I don't know the sources

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -19,25 +37,69 @@ class CityInfoReligionManager: Counter<String>() {
...
+    
+    fun clearAllPressures() {
+        pressures.clear()
+    }
+    
+    fun addPressure(religionName: String, amount: Int) {
+        val currentMajorityReligion = getMajorityReligion()
+        pressures.add(religionName, amount)
+        val newMajorityReligion = getMajorityReligion()
+        if (currentMajorityReligion != newMajorityReligion && newMajorityReligion != null) {
```
SomeTroglodyte, line 68: At this point 'current' sounds... old.

android/assets/jsons/translations/template.properties
xlenstra: I implemented two uniques: one with the part between brackets and one without. The one with brackets is in a json so is automatically picked up, but the one without isn't so it is added here manually :)

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -19,25 +37,69 @@ class CityInfoReligionManager: Counter<String>() {
...
+        
+        
+        val religionOwningCiv = cityInfo.civInfo.gameInfo.getCivilization(cityInfo.civInfo.gameInfo.religions[newMajorityReligion]!!.foundingCivName)
+        
+        for (unique in cityInfo.civInfo.gameInfo.religions[newMajorityReligion]!!.getFounderUniques()) {
+            if (unique.placeholderText == "[] when a city adopts this religion for the first time (modified by game speed)") {
+                val statsGranted = unique.stats.times(cityInfo.civInfo.gameInfo.gameParameters.gameSpeed.modifier)
+                religionOwningCiv.addStats(statsGranted)
+                
+                if (cityInfo.location in religionOwningCiv.exploredTiles)
```
SomeTroglodyte, line 86: Better diplomacy manager? I _could_ imagine a civ you never met having founded a city where your scout passed by centuries earlier..

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -19,25 +37,69 @@ class CityInfoReligionManager: Counter<String>() {
     }
     
     fun getNumberOfFollowers(): Counter<String> {
-        val totalInfluence = values.sum()
+        val totalInfluence = pressures.values.sum()
         val population = cityInfo.population.population
         if (totalInfluence > 100 * population) {
```
SomeTroglodyte, line 42: Sounds really strange. And doesn't it make getNumberOfFollowers not adding up to cityInfo.population.population by a potentially wide margin? If that's significant at all? The other loop also does not guarantee it adds up, though it will come close...

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -19,25 +37,69 @@ class CityInfoReligionManager: Counter<String>() {
...
+                val statsGranted = unique.stats.times(cityInfo.civInfo.gameInfo.gameParameters.gameSpeed.modifier)
+                religionOwningCiv.addStats(statsGranted)
+                
+                if (cityInfo.location in religionOwningCiv.exploredTiles)
+                    religionOwningCiv.addNotification("You gained [$statsGranted] as your religion was spread to [${cityInfo.name}]", cityInfo.location, NotificationIcon.Faith)
+                else
+                    religionOwningCiv.addNotification("You gained [$statsGranted] as your religion was spread to an unknown city", NotificationIcon.Faith)
+            }
+            
+            if (unique.placeholderText == "[] when a city adopts this religion for the first time") {
```
SomeTroglodyte, line 92: ah - implement both but only one used in json... OK, but code duplication could be folded away.

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -108,6 +108,17 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
         statMap["Transportation upkeep"] = Stats().apply { gold = -getTransportationUpkeep().toFloat() }
         statMap["Unit upkeep"] = Stats().apply { gold = -getUnitMaintenance().toFloat() }
 
+        if (civInfo.religionManager.religion != null) {
+            for (unique in civInfo.religionManager.religion!!.getFounderBeliefs().flatMap { it.uniqueObjects }) {
+                if (unique.placeholderText == "[] for each global city following this religion") {
+                    statMap.add("Religion", unique.stats.apply {
+                        this.timesInPlace(civInfo.religionManager.numberOfCitiesFollowingThisReligion().toFloat())
+                    })
```
SomeTroglodyte, line 116: `timesInPlace`? side effects intended or guaranteed to not exist because unique is a deep copy already? OK, I'm only looking at the text diff not in Studio, might be easier to check there... But if side effect intended and that flatMap's receiver is a sequence, then wouldn't it go back into the ruleset because the iterator created for uniqueObjects is a view? And... No, my head starts hurting...

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -185,6 +196,19 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
...
+            statMap["Religion"] = 0f
+            for (unique in civInfo.religionManager.religion!!.getFounderBeliefs().flatMap { it.uniqueObjects }) {
+                if (unique.placeholderText == "[] for each global city following this religion") {
+                    statMap["Religion"] = statMap["Religion"]!! + unique.stats.apply {
+                        this.timesInPlace(civInfo.religionManager.numberOfCitiesFollowingThisReligion().toFloat())
+                    }.happiness
+                }
+            }
+            if (statMap["Religion"] == 0f) 
+                statMap.remove("Religion")
```
SomeTroglodyte, line 209: `add - if() remove` pattern instead of `if() add` -> -0.001 ❤️. Hmmm... We need a moddable 'currency' named 'Love' in Unciv? Sorry. A little more seriously: I need to go fetch my Studio, from that glance it shouldn't compile? Isn't StatMap[key] a `Stats` and why then can we assign/compare with a Float?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -712,6 +714,12 @@ class CivilizationInfo {
         }
     }
     
+    fun addStats(stats: Stats) {
+        for (stat in stats.toHashMap()) {
+            addStat(stat.key, stat.value.toInt())
```
SomeTroglodyte, line 719: clearly document it rounds _down_?

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -69,7 +69,7 @@ class ReligionManager {
     fun canFoundPantheon(): Boolean {
         if (!civInfo.gameInfo.hasReligionEnabled()) return false
         if (religionState != ReligionState.None) return false
-        if (!civInfo.isMajorCiv()) return false
+        if (civInfo.isCityState()) return false
```
SomeTroglodyte, line 72: Doesn't that mean a Barbarian Religion? Do they at least get a large Bonus?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -462,22 +462,24 @@ object UnitActions {
...
         if (!unit.abilityUsedCount.containsKey("Religion Spread")) return // This should be impossible anyways, but just in case
         if (maxReligionSpreads <= unit.abilityUsedCount["Religion Spread"]!!) return
         val city = tile.getCity()
+        if (city == null) return
         actionList += UnitAction(UnitActionType.SpreadReligion,
             title = "Spread [${unit.religion!!}]",
             action = {
                 unit.abilityUsedCount["Religion Spread"] = unit.abilityUsedCount["Religion Spread"]!! + 1
-                city!!.religion[unit.religion!!] = 100
+                city.religion.clearAllPressures()
```
SomeTroglodyte, line 475: Radical! I would have expected such a "Spread Religion" to just add some points... And less than the points a holy city gets initially... (prejudice by name, not Civ5 knowledge)

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -462,22 +462,24 @@ object UnitActions {
...
+                city.religion.addPressure(unit.religion!!, 100)
                 unit.currentMovement = 0f
                 if (unit.abilityUsedCount["Religion Spread"] == maxReligionSpreads) {
                     addGoldPerGreatPersonUsage(unit.civInfo)
                     unit.destroy()
                 }
-            }.takeIf { unit.currentMovement > 0 && city != null && city.civInfo == unit.civInfo } // For now you can only convert your own cities
```
SomeTroglodyte, line 480: Awww.... OK, if this is interim contradicting the intended final rules I'd make that a //todo one, but that's just a matter of taste. We'll have to do a todo week sometime... 😆

SomeTroglodyte (commented):
Just skimmed the text diffs and all my input is just food-for-thought, looks nice.

core/src/com/unciv/logic/city/CityReligion.kt
logicminimal: you can already see territories of nations you havent met before though

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
xlenstra: This is an interim, later on each city will have followers depending on how much the religion was spread to that city, and than spreading religion will only convert a few of these followers. That is yet to be implemented, so for now religions just replace each other

core/src/com/unciv/logic/city/CityReligion.kt
xlenstra: This function was already here before I started working on religion, I actually have no idea what it does and just changed the code based on my change of pressure from an inherited property to a member property. I'll look at it later when I implement cities having multiple religions competing inside them & followers

core/src/com/unciv/logic/civilization/CivInfoStats.kt
xlenstra: You would think that, but in this instance it is a map mapping 'origins' to 'happiness' values, as this function only computes happiness, which needs to be computed separately so that effects like "+15% science if the empire is happy" can be resolved correctly. And while I could test for it beforehand, that would require doing `getMatchingUniques()` twice, which would incur more work, and if in the end the values just happen to sum up to 0 it is still displayed which you also don't want. This seemed to be the best solution to me

core/src/com/unciv/logic/civilization/ReligionManager.kt
xlenstra: They shouldn't generate faith so it should be fine in the base game, but I'll implement a check for it nontheless, in some weird mods where barbarians might earn faith by attacking this could cause problems

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
xlenstra: The line change here actually rectifies a previous interim ruling, so the new rules should be correct. I'll add a ToDo to other places where there is an interim ruling as of yet, though

core/src/com/unciv/logic/civilization/CivInfoStats.kt
SomeTroglodyte: Riiight, with enough code in view it's plain. I just got confused as there is a class named StatMap.

core/src/com/unciv/logic/civilization/ReligionManager.kt
SomeTroglodyte: Yeh - ***I*** could only be a Barbarian if I had _tons_ of faith - in _myself_ :smile_cat:

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -200,9 +200,9 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
             statMap["Religion"] = 0f
             for (unique in civInfo.religionManager.religion!!.getFounderBeliefs().flatMap { it.uniqueObjects }) {
                 if (unique.placeholderText == "[] for each global city following this religion") {
-                    statMap["Religion"] = statMap["Religion"]!! + unique.stats.apply {
-                        this.timesInPlace(civInfo.religionManager.numberOfCitiesFollowingThisReligion().toFloat())
-                    }.happiness
+                    statMap["Religion"] = 
+                        statMap["Religion"]!! +
```
SomeTroglodyte, line 204: :+1: .  Really makes one think, though. There's a reason why so many CPU architectures have MulAdd opcodes. Plus I still think our Stats with its permanent toHashMap (meaning tons of small allocations and work for the GC) is - not nice. And - why does StatMap.add _not_ pull a clone when the addition is new, but _does_ pull a clone when it's already in? I'd have opted for the other way round.

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -57,7 +110,11 @@ class CityInfoReligionManager: Counter<String>() {
         for (city in allCitiesWithin10Tiles) {
             val majorityReligionOfCity = city.religion.getMajorityReligion()
             if (majorityReligionOfCity == null) continue
-            else add(majorityReligionOfCity, 6) // todo - when holy cities are implemented, *5
+            else addPressure(
+                majorityReligionOfCity, 
+                if (city.isHolyCity()) 11 // ToDo: check if this should be 6 + 5 or 6 * 5
```
yairm210, line 115: It should be 30

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -108,6 +108,17 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
         statMap["Transportation upkeep"] = Stats().apply { gold = -getTransportationUpkeep().toFloat() }
         statMap["Unit upkeep"] = Stats().apply { gold = -getUnitMaintenance().toFloat() }
 
+        if (civInfo.religionManager.religion != null) {
+            for (unique in civInfo.religionManager.religion!!.getFounderBeliefs().flatMap { it.uniqueObjects }) {
+                if (unique.placeholderText == "[] for each global city following this religion") {
+                    statMap.add("Religion", unique.stats.apply {
```
yairm210, line 114: unique.stats.times(...), you don *not* want to use timesInPlace because that will change the stats of the unique, which it will apply to the next city, so this will keep on growing...

core/src/com/unciv/logic/civilization/CivInfoStats.kt
yairm210: I'm pretty sure that stats.add goes over all values and adds them as ints for efficiency?
Also, statMap.add("Religion", stats) exists for this reason...

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -185,6 +196,19 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
 
         statMap["Natural Wonders"] = happinessPerNaturalWonder * civInfo.naturalWonders.size
 
+        if (civInfo.religionManager.religion != null) {
+            statMap["Religion"] = 0f
+            for (unique in civInfo.religionManager.religion!!.getFounderBeliefs().flatMap { it.uniqueObjects }) {
+                if (unique.placeholderText == "[] for each global city following this religion") {
+                    statMap["Religion"] = 
+                        statMap["Religion"]!! +
+                        unique.stats.times(civInfo.religionManager.numberOfCitiesFollowingThisReligion().toFloat()).happiness
```
yairm210, line 205: unique.stats.happiness * ...

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -193,8 +193,17 @@ class DiplomacyManager() {
...
     fun getCityStateInfluenceRestingPoint(): Float {
         var restingPoint = 0f
+        
         for (unique in otherCiv().getMatchingUniques("Resting point for Influence with City-States is increased by []"))
             restingPoint += unique.params[0].toInt()
+        
+        for (unique in otherCiv().getMatchingUniques("Resting point for Influence with City-States following this religion []"))
+            if (
+                otherCiv().religionManager.religion != null 
+                && civInfo.getCapital().religion.getMajorityReligion() == otherCiv().religionManager.religion!!.name
```
yairm210, line 203: if( otherCiv().religionManager.religion?.name == civInfo.getCapital().religion.getMajorityReligion()

core/src/com/unciv/logic/civilization/CivInfoStats.kt
xlenstra: Except that this `statMap` isn't a `statMap` but instead an `hashMap`, so it doesn't have that function sadly. Why it's still called `statMap` I don't know

PR #4828, opened 2021-08-10
Add Holy Site & Holy Place graphic

AdityaMH:


ravignir:
Holy Place is redundant. There is no such thing in civ5,

PR #4830, opened 2021-08-11
Stats optimization

SomeTroglodyte:
Some console output:
```
master:
statMathStressTest took 11938999µs and 259609kB
statMathStressTest took 10276942µs and 210343kB
statMathStressTest took 10153385µs and 285419kB

this:
statMathStressTest took 5295211µs and 175893kB
statMathStressTest took 4379497µs and 173967kB
statMathStressTest took 4319839µs and 235630kB
```

The approach here is:
- Avoid creating new instances where possible
- No use of `toHashMap` within the class - left in so the rest of Unciv needs no changes
- No loops over the seven stats (but this is likely a minor part of the gains)

All remaining uses of toHashMap are more or less as iterator. Two more little helpers and every one of the remaining 21 calls would be much shorter and efficient - e.g. `.toHashMap().values.sum()` could become `.values.sum()`, `.toHashMap().maxByOrNull { it.value }!!.key` could be `.maxByOrNull { it.value }!!.key`, and so on. `toHashMap` would be gone.

Yes the test code is horrible - because it emulates current use patterns of the class to a degree. Improving the usage patterns would bring memory stress down a little further.

So - big question: Do we want to go down this road for a measly 110% performance gain of a single class?


SomeTroglodyte:
I forget - there are several key components in Unciv already that depend on reflection, but which ones were those? I believe I even saw something within Gdx itself that does... I'll try to remember to holler when I next stumble upon one - or should I remember. Sad jokes aside, the all too ready availability of that `::` "gimme a boxed reference" operator speaks volumes I think. 

As I imagined it, the hashMap as iterator will be replaced by two sequences - one values only for aggregations and one as key/value pairs. They and all of the operations currently touching each field discretely could readily be rewritten into loops - but, squeeze every last cycle away...

The alternate approach would be to make the HashMap primary storage and to rewrite the discrete properties into using it as backing. Haven't measured that, I estimate it being on that 10sec to 4sec scale somewhere between 6 and 7, maybe 5. I'm lazy but holler loudly if you want that comparison. Oh, reflection-free. On the surface at least.

The actual implementation isn't the most important aspect here, but showing it can be done, that the potential gain is high, and the barely measurable gain due to relieving GC pressure (which in these runs most likely applied _after_ the run ended) is significant, too.

yairm210:
One thing I would do is make the map be lazy initialized.
Most Stats aren't going to use that, and making a map IS more memory and cpu...

SomeTroglodyte:
I had my doubts - lazy has _some_ overhead (I had the comparison on the road map anyway, I just didn't start out with it due to these prejudices), but that turned out to be an excellent idea. Final measurements, now including an implementation of the 'backing field _is_ already a map' idea:

current master: 10.4
pre-instantiated ::view: 4.6
lazy ::view: 3.8s :+1: :clap: :champagne: :partying_face: :birthday: 
map backing: 5.5s

I'll now clean up and document the thing - warning:
- toHashMap will stay but be reordered to the order of the enum. That changes the result of the `statMathRandomResultTest`, and potentially some toString() order. I want the order _consistent_. The function will be replaced anyway, and the replacement sequence _will_ have enum order.
- I'll reorder the methods a little more logically which will make the diff less readable. Functionally only the lazy bit and that ordering will change.

SomeTroglodyte:
Since this is so basic, I'd like every active coder to take a look. @yairm210  could you invite them? I might be prejudiced and forget the most important ones...

FYI @All:
I'm done with phase 1 - this PR-, which was made under the constraint to produce a compatible Stats class without changing a single file in the rest of the core/* sources. Phase 2 will add a few API elements and remove `toHashMap` and subsequently touch a lot of client code - `Stats(gold = 1f)` instead of `Stats().apply { gold = 1f }`, `for ( (stat,value) in <Stats>` instead of using `toHashMap` for that, or `<Stats>.values.sum()` instead of the current construct...

yairm210:
Looks good to me 😉

yairm210:
@xlenstra I would like to hear your thoughts on this

xlenstra (approved):
I only have two very minor nitpicks:
- On coding style I would prefer the closing bracket of `isEmpty()` to be on the following line
- We currently have the functions `add` (in place), `plus` (not in place) `times` (not in place) and `timesInPlace` (in place). Whether these are in place or not is not visible based on the name of the function, which is not very pretty, although the Kdoc does explain this.

Other than that, looks good to me

SomeTroglodyte:
> closing bracket

True. But doesn't change the outcome... And the follow-up is planned anyway.

>  not visible based on the name of the function

Very true. But - designed as drop-in, so no changing names. Yet. With name changes allowed - I'd say the non-inPlace would be the operator overrides only, so not times(), and either naming all inPlace with that suffix or some new naming idea. '`Mutant`'! (does that [no mutants allowed site](https://nma-fallout.com/) still exist?)

PR #4833, opened 2021-08-11
Update Lithuanian.properties

Pirdachu:
Added translate in the start and a bit in the middle closer to the bottom, changed some lithuanian translate found bad translate :/


SomeTroglodyte:
> Placeholder `[otherCiv]` not found in `Lithuanian` for entry `[otherCiv]'s items`

I should add a 'Rules and Tests' chapter to [the wiki](https://github.com/yairm210/Unciv/wiki/Translating) some time...

PR #4835, opened 2021-08-11
G&K resources, resource improvement yield changes to match civ5.

ravignir:
Side note:
Improvement yields + Improved resource yield do not stack in civ5, while they do in unciv. 
For example: Wheat + Farm provides just 1 additional food instead of 2(1+1). 

This commit also includes BNW resources sprites.


ravignir:
> And your side note - had to look at the diff to see this PR does _not_ change the farms or wheat. But - you really sure? Wouldn't that make wheat almost pointless? Grass=2, Grass+Wheat=3, Grass+Farm=3, Grass+Wheat+Farm=3?
> 
I think you messed your calculation here, Grass(2)+Wheat(1)+Farm(1) should be 4, not 3 :D
5 with Granary.
And atm (in unciv) it is Grassland=2 + Wheat=1 + Farm=2  +Granary=1 which is 6 total. Just ignore Wheat cannot spawn on Grass.

> And... Looks 👍
> 
> ... the Truffles look like the three little piggies because you can't see the truffles in the forest normally? But neither can you see gems lying around on the surface in RL.. Even agates which can be found as surface pebbles don't look like gems unless cut... I'm joking, fine choice.
>

That is also what civ5 does. No texture would also make sense xD

logicminimal:
> Say, I know that's offtopic, but that `"Gold Ore", // Not called "Gold" in order to not conflict with siege type units for translations` comment in the resource file is off. Fix in one go? "civ gold" instead of "siege type units" or so?
> 
> And your side note - had to look at the diff to see this PR does _not_ change the farms or wheat. But - you really sure? Wouldn't that make wheat almost pointless? Grass=2, Grass+Wheat=3, Grass+Farm=3, Grass+Wheat+Farm=3?
> 
> And... Looks :+1:
> 
> ... the Truffles look like the three little piggies because you can't see the truffles in the forest normally? But neither can you see gems lying around on the surface in RL.. Even agates which can be found as surface pebbles don't look like gems unless cut... I'm joking, fine choice.

so the wheat etc are broken sith this pr?

SomeTroglodyte:
"sith" as in "Lord of the Sith"? That belongs in a mod. 

> No texture

Hey! Truffles *do* have a taste! Like mouldy paper mostly...

> messed your calculation

_No_ calculation was my point, that's what I thought it _sounds_ like: Get one bonus or the other but not both. Yea, actually simplifies to delete `improvementStats` for Wheat, leaving its bonus static, and Farms add what they always do...  :stuck_out_tongue_winking_eye:

ravignir:
It is more of a overrideStats for improvement, than bonus yield when improved.

Issue #4697, opened 2021-07-31
Missing features from Civ V - G&K


xlenstra:
Issue #663 is quite old and contains lots of discussion on issues that have long been solved, cluttering up the issue. Furthermore, multiple new issues have been posted there, but have not been added to the comment at the top, as well as issues that have been solved that have not been removed. This issue aims to solve these problems, and stay up-to-date, at least for the foreseeable future.

Major things:
- [x] City States - see for examples #4602
- [x] Religion - #4290
- [ ] Espionage - #7610
- [x] Score and Time Victory

Units:
- [x] Ranged units may capture civilian units instead of destroying them by moving to their tile
- [x] Capturing civilian units should not cost an attack
- [x] Great people (except great prophets) & work-boats should be destroyed on capture
- [x] Supply cap
- [ ] Pillaging of roads when the improvement is already pillaged
- [x] You should not be able to pillage your own tiles
- [x] Pillaging should give gold, based on the improvement pillaged different amounts of gold are plundered
- [x] Barbarians can capture workers/settlers and escort them to encampments. From there they can be liberated, even to other civs. If not liberated, settlers turn into workers upon recapture by a different civ.
- [x] Spawning of barbarian encampments and units in them has other rates in the base game
- [ ] Embarked units should be on a different layer than military and civilian water units. Embarked military and civilian units should be on the same layer.
- [x] Embarked units (except for marines) should have -1 vision
- [x] Spaceship parts should be units, that are added to the spaceship by using an action in the capital
- [ ] 'Build road to' action for workers
- [ ] 'Remove [feature] and build improvement' action for workers
- [x] Helicopter Gunship has its own unit type: Helicopter (part of greater overhaul of unit types)
- [x] Scouts upgraded through ancient ruins to archers keep the 'no terrain cost' unique (turn it into a promotion)
- [x] Guided missiles should not cost maintenance
- [ ] Great Admiral

Combat:
- [x] Air Sweep action for fighters
- [x] Dogfighting promotion
- [x] Aircraft take damage when attacking
- [x] Submarines should no longer be invisible to enemy civs when there is one of their units next to it
- [x] The +Strength% vs City uniques of most siege units should only be offensive (see #3506)
- [x] Land melee units should not able to capture civilian or attack military units in water and vise versa.
- [ ] Cities should still inflict damage when attacked at 0 hp
- [x] Melee units should enter a city upon capture 
- [x] If you recapture your capital, it should become your capital again
[...]

PR #4839, opened 2021-08-11
Fix units not entering cities upon capture

avdstaaij:
This fixes the bug where units would not enter a city upon capturing it, as listed in #4697.

Only the changes in `Battle.attack` and `UnitMovementAlgorithms.canMoveTo` are necessary to fix the bug. The other changes are for different reasons.

I renamed `CivilizationInfo.canEnterTiles` to `canPassThroughTiles` and `TileInfo.canCivEnter` to `canCivPassThrough` to more accurately describe what they do. Almost everywhere in the code and the comments, "entering a tile" is used to refer to ending movement on that tile instead of merely passing through it. The renamed functions test whether certain tiles can be passed through, _not_ whether they can be "entered" in this sense, hence the renaming. I also added some documenting comments to both of these functions and to `UnitMovementAlgorithms.canPassThrough` to make it more clear what exactly they check.

Furthermore, the function that is now called `TileInfo.canCivPassThrough` used to contain an exception to always allow player civilization (but not AI) units to pass through city state tiles. I moved that exception to the slightly lower-level `CivilizationInfo.canPassThroughTiles`, which is called by `TileInfo.canCivPassThrough`. This makes the behavior of `CivilizationInfo.canPassThroughTiles` more accurate: player units _can_ indeed pass through tiles of city-states.

A side-effect of this change is that functions that used the lower-level `CivilizationInfo.canPassThroughTiles` function directly now consider this exception, while they previously didn't. For example, if a city-state expands its border, it should now no longer push out units of a player civilization. I do not know whether these side-effect changes are accurate the original Civ V, but the behavior of `CivilizationInfo.canPassThroughTiles` now certainly makes more sense. If we need a function that checks if a tile can be passed through without considering the city-state exception, I think that a separate function with a distinctive name should be added.

It may make even more sense to move the exception to the "can enter" logic, since AI units should probably be able to pass through city-state tiles even when they don't want to end on them. That might however affect the AI more significantly. I went for the safer change.

### DIFF
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -2,7 +2,6 @@ package com.unciv.logic.automation
 
 import com.unciv.Constants
 import com.unciv.logic.city.CityInfo
-import com.unciv.logic.city.IConstruction
 import com.unciv.logic.city.INonPerpetualConstruction
 import com.unciv.logic.city.PerpetualConstruction
 import com.unciv.logic.civilization.*
@@ -399,7 +398,7 @@ object NextTurnAutomation {
         fun isTileCanMoveThrough(tileInfo: TileInfo): Boolean {
             val owner = tileInfo.getOwner()
             return !tileInfo.isImpassible()
-                    && (owner == otherCiv || owner == null || civInfo.canEnterTiles(owner))
+                    && (owner == otherCiv || owner == null || civInfo.canPassThroughTiles(owner))
         }
 
         val reachableEnemyCitiesBfs = BFS(civInfo.getCapital().getCenterTile()) { isTileCanMoveThrough(it) }
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -69,11 +69,11 @@ object Battle {
 
         // check if unit is captured by the attacker (prize ships unique)
         // As ravignir clarified in issue #4374, this only works for aggressor
-        val captureSuccess = defender is MapUnitCombatant && attacker is MapUnitCombatant
+        val captureMilitaryUnitSuccess = defender is MapUnitCombatant && attacker is MapUnitCombatant
                 && defender.isDefeated() && !defender.unit.isCivilian()
                 && tryCaptureUnit(attacker, defender)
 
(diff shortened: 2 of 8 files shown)


yairm210:
I want through the code and it looks good, but since this deals with the fundamentals of unit movement, I want us to check this seriously.
In the UncivGame class there are flags to autoplay until a certain turn. Please run this to 400 turns on 2 games and ensure there are no errors.

avdstaaij:
Auto-running the game with a large `UncivGame.simulateUntilTurnForDebug` nearly always results in a "something went disastrously wrong" crash due to the following exception:

<details>
  <summary>Exception stack trace</summary>

```
Exception in thread "NextTurn" kotlin.UninitializedPropertyAccessException: lateinit property unit has not been initialized
	at com.unciv.logic.map.UnitPromotions.getAvailablePromotions(UnitPromotions.kt:51)
	at com.unciv.logic.map.UnitPromotions.canBePromoted(UnitPromotions.kt:19)
	at com.unciv.logic.automation.NextTurnAutomation.automateUnits(NextTurnAutomation.kt:507)
	at com.unciv.logic.automation.NextTurnAutomation.automateCivMoves(NextTurnAutomation.kt:53)
	at com.unciv.logic.GameInfo.nextTurn(GameInfo.kt:130)
	at com.unciv.ui.worldscreen.WorldScreen$nextTurn$1.invoke(WorldScreen.kt:592)
	at com.unciv.ui.worldscreen.WorldScreen$nextTurn$1.invoke(WorldScreen.kt:56)
	at kotlin.concurrent.ThreadsKt$thread$thread$1.run(Thread.kt:30)
```

</details>

Apparently, the auto-simulation sometimes results in the `unit` property from `UnitPromotions` to not be set, probably because `MapUnit.setTransients` is skipped somehow. I was not able to find the cause of this quickly.
Anyway, this crash is not related to this PR: it exists on master too. Perhaps an issue should be opened, but it is hard to call this a bug since it only occurs when using a debug variable.

Instead of auto-simulating, I manually clicked "Next turn" up to turn 400 for two games with three AI players and a spectator, and no errors occured. I did not notice any movement irregularities either, so I believe the changes of this PR are safe.

yairm210:
Nice!
I've seen the stacktrace of that bug and have been looking for a way to reproduce it, this is very helpful!

PR #4843, opened 2021-08-12
CityScreen Buy button rework

SomeTroglodyte:
- e.g. the empty "Buy" for Missionary gone
- Info pane no longer staying on screen even when bought building no longer offered
- Purchase Prompt a little prettier

Sibling to #4822 , will need merging.


core/src/com/unciv/models/stats/Stat.kt
```diff
@@ -2,13 +2,22 @@ package com.unciv.models.stats
...
-    Culture(NotificationIcon.Culture, UncivSound.Paper),
-    Happiness(NotificationIcon.Happiness, UncivSound.Click),
-    Faith(NotificationIcon.Faith, UncivSound.Choir);
-}
\ No newline at end of file
+enum class Stat(
+    val notificationIcon: String,
+    val purchaseSound: UncivSound,
+    val character: Char,
+    val isPotentialCurrency: Boolean = true
```
yairm210, line 11: Character, maybe. But this is definitely not a 'basic value' of the stat, this is totally gameplay-related. This should not be a field.
The question then becomes where to put it, which becomes - where is this actually used?

SomeTroglodyte: Oh, totally forgot about this. I had another approach coded first, quite straightforward - originally a listOf() was instantiated every use, once per Unciv run is enough so just move that to a singleton - Stat.Companion would be a good place IMHO. A companion closer to the use - city screen package only so far - would still be a source screen away at least, so I'd tend to Stat.kt...

SomeTroglodyte:
> Shouldn't we make the YesNo table prompt always be centered? We can solve this globally, not just locally

Good you picked up on that. Hacking it "locally" was just studiously avoiding influencing other people's work - or - `by lazy { "Don't wanna check all other uses right now if they all would look better, too" }`. Here that gets a `@Transient`, not a `@delegate:Transient` , the joke meaning I'll overcome that laziness shortly and look at the other usecases :wink:

SomeTroglodyte:
Okay, so checking the 17 YesNoPopups popped up just one that would be influenced - the one to buy a tile, same city screen. Made it look the same, centering now in class, all others are one-liners with text longer than the Yes/No buttons, so they look the same. Also, noticed the translation template had not been patched for the variable currencies - fixed incl. testing lang.

But to test in peace I had to merge in #4893, so this now depends on that.

PR #4849, opened 2021-08-13
Improve horse sound

AdityaMH:
I think horse sound is too worst, so i'm improve one.


ravignir:
I actually like already implemented sounds more.

AdityaMH:
I'm just concerning horse sound. Why sounded like bring ill horse to the war. So, i want to change it.

SomeTroglodyte:
While I'm all in favour of better sounds - before checking the actual sounds here: Attribution.

- New construction sound not credited
- New source site 'orangefreesounds'
    - pulls code from _six_ untrustworthy third parties (facebock, tweeter, googlesyndication, googletagmanager, pinterest, statcounter) while our established source only pulls two (fazeblock, twister) without even opening up scripting.
    - full of ads

Listening to the sources:
- Horse - yes I like the new one better, for the old one has some bumping noises that son't fit - didn't compare after Audacity
- But - plenty of alternatives.. How about a short excerpt from https://freesound.org/people/gurek/sounds/547967/ ?

AdityaMH:
How about this with some editing?
https://freesound.org/people/Kubuzz/sounds/347036/

AdityaMH:
Wait a minutes...
Why instanly rejected in a second? Github bug?

PR #4851, opened 2021-08-13
Fixed some bugs with spreading of religion

xlenstra:
Fixed bugs include:
- Great Prophets gained from other sources than faith generation do not have a religion
- Missionaries cannot spread religion
- Founder uniques were not included in `CivilizationInfo.getMatchingUniques()`


ravignir:
But what if you gain Prophet outside of Faith generation and you don't have religion (or even pantheon) yet?

xlenstra:
> But what if you gain Prophet outside of Faith generation and you don't have religion (or even pantheon) yet?

If you don't have a religion but do have a pantheon, this great prophet can be used to found a religion. If you don't have a pantheon yet, the prophet will not have a religion, but once you have founded a pantheon, you can use this prophet to found a religion.
This is different from the original, where if you somehow get a great prophet before founding a pantheon, you can choose a pantheon belief while creating your religion, but I'll implement that later.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -263,7 +263,12 @@ class CivilizationInfo {
                 } +
                 policies.policyUniques.getUniques(uniqueTemplate) +
                 tech.techUniques.getUniques(uniqueTemplate) +
-                temporaryUniques.filter { it.first.placeholderText == uniqueTemplate }.map { it.first }
+                temporaryUniques.filter { it.first.placeholderText == uniqueTemplate }.map { it.first } +
+                if (religionManager.religion != null) 
+                    religionManager.religion!!.getFounderUniques()
```
yairm210, line 268: . asSequence should be before .filter

SomeTroglodyte: Maybe. With expected average cardinality near or below 1 _could_ be marginally more efficient the other way round.

What I see here is repeated lines ending in an operator. Coding style and kotlin not needing `;` ... You know, if you _want_ the operators _leading_ the line for readability, that can be done - just have an unmatched bracket open... I think. Note to self.

yairm210:
Conflicts 🙃

xlenstra:
I'm currently not home, and the conflicts look to complicated to solve on mobile, so I'll fix these tomorrow sometime

PR #4854, opened 2021-08-13
Turkish translations again

logicminimal:


logicminimal:
what is wrong again

xlenstra:
> what is wrong again

All tests pass, so nothing?

logicminimal:
> > what is wrong again
> 
> All tests pass, so nothing?

it saus checks failed

xlenstra:
> > > what is wrong again
> > 
> > 
> > All tests pass, so nothing?
> 
> it says checks failed

Maybe for your commits, but that was due to a problem plaguing every PR. That problem has since been fixed, and your entire PR now has a green checkmark.

PR #4859, opened 2021-08-14
BFS cleanup

SomeTroglodyte:
- Kotlin ArrayDeque is no longer experimental
- Client code sidestepped class API even where API existed -> closed it down
- nextStep when already ended now no-op instead of crash, suits usecases
- predicate test very likely to be more expensive than tilesReached lookup
- getPathTo only used asSequence() - why not a Sequence in the first place
- Compleat documentation. This is very low level, so better help future coders to understand.
- Lint: One annoying hardcoded conditional breakpoint, one unhelpful exception

### DIFF
--- a/core/src/com/unciv/logic/map/BFS.kt
+++ b/core/src/com/unciv/logic/map/BFS.kt
@@ -1,59 +1,80 @@
 package com.unciv.logic.map
 
-// Kotlin's ArrayDeque is experimental
-import java.util.ArrayDeque
+import kotlin.collections.ArrayDeque
 
 /**
- * Defines intermediate steps of a breadth-first search, for use in either get shortest path or get onnected tiles.
+ * Defines intermediate steps of a breadth-first search, for use in either get shortest path or get connected tiles.
  */
-class BFS(val startingPoint: TileInfo, val predicate : (TileInfo) -> Boolean) {
-    var tilesToCheck = ArrayDeque<TileInfo>()
+class BFS(
+    val startingPoint: TileInfo,
+    private val predicate : (TileInfo) -> Boolean
+) {
+    /** remaining tiles to check */
+    private val tilesToCheck = ArrayDeque<TileInfo>(37)  // needs resize at distance 4
 
     /** each tile reached points to its parent tile, where we got to it from */
-    val tilesReached = HashMap<TileInfo, TileInfo>()
+    private val tilesReached = HashMap<TileInfo, TileInfo>()
 
     init {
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/logic/map/BFS.kt
```diff
@@ -1,59 +1,80 @@
...
 
+    /**
+     * Process one tile-to-search, fetching all neighbors not yet touched
+     * and adding those that fulfill the [predicate] to the reached set
+     * and to the yet-to-be-processed set.
+     * 
+     * Will do nothing when [hasEnded] returns `true`
+     */
     fun nextStep() {
-        val current = tilesToCheck.remove()
```
xlenstra, line 33: Am I missing something or is the current tile no longer removed from the set of tiles to check?

core/src/com/unciv/logic/map/TileMap.kt
```diff
@@ -254,10 +254,10 @@ class TileMap {
...
                         bNeighbor: TileInfo ->
                     val bNeighborHeight = bNeighbor.getHeight()
-                    if(cTile.resource=="Marble"
-                        && bNeighbor.terrainFeatures.contains("Forest")
-                    )
-                        println()
+//                    if(cTile.resource=="Marble"
+//                        && bNeighbor.terrainFeatures.contains("Forest")
+//                    )
+//                        println()
```
xlenstra, line 260: I vote for completely removing this code instead of just commenting it out

logicminimal:
i dont know if this is even possible but can we just alert x number of workers who happen to be logically sleeping (basically doesnt have any work to do) when x number of things workers can improve becomes unlocked? (if you have 5 hills and research mining and you have 6 workers it will only check for 5 of it, (or 6 of it if we add distance to the calculate in which case we will need to check distances of every one))

core/src/com/unciv/logic/map/BFS.kt
SomeTroglodyte: Damn. In my tested java variant that was a `pollFirst`, and that translates to `removeFirstOrNull`... The shame...

core/src/com/unciv/logic/map/TileMap.kt
SomeTroglodyte: Yup - another oops. I admit doing conditional breakpoints this way executes faster... But... Well, now you know where all those spurious empty lines on the console came from when you analyzed that masochist's save.

PR #4860, opened 2021-08-14
Update bug_report.md

SomeTroglodyte:
Some issues lately have been shockingly devoid of useful information...
- Reintroduce environment questions inspired by the standard template, albeit simpler
- That details trick to make large saved games manageable is already used by some, it *could* work in the hands of the masses - or too dangerous?


yairm210:
Looks good

Issue #4861, opened 2021-08-14
All current PR's failing due to Gradle 7 running under Java 8


SomeTroglodyte:
... or is it simply `Unciv/blob/master/.github/workflows/buildAndDeploy.yml` line 35? -> **Yes**. Use "11" instead of "8"...

... Upping all three sourceCompatibility = JavaVersion.VERSION_1_6 to 1_7 will allow the build to proceed further ...

... Fixing "Deprecated Gradle features were used in this build" is just 5 lines, 4 in the android build script, one in BuildConfig.kt because JFrog died ...

... I can keep lint from crashing (_yes, complete with 380M crash dump file_) by disabling _all_ warnings and setting abort on error to false ...

One _clean_ `./gradlew check --warning-mode all`.

SomeTroglodyte:
@yairm210  I think chances are OK I got a "green" branch. Testing.

PR #4862, opened 2021-08-14
Gradle upgrade fixes

SomeTroglodyte:
For #4861 

- build JVM from 8 to 11
- sourceCompatibility from 1_6 to 1_7
- updated deprecated gradle language
- Satisfies :android:lintDebug errors
- Tries to prevent it crashing all the time by matching kotlin-reflect to main kotlin version (1.5.20 soon please)
- Some linting

*** Still fails ***

Mix and match as you wish.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -3,5 +3,5 @@ plugins {
 }
 
 repositories {
-    jcenter()
+    mavenCentral()  // "JFrog announced JCenter's shutdown in February 2021"
 }
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -32,7 +32,7 @@ jobs:
 
       - uses: actions/setup-java@v1
         with:
-          java-version: '8'
+          java-version: '11'
 
       # Runs a set of commands using the runners shell
       - name: check
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
 }
 
 android {
-    compileSdkVersion(30)
+    compileSdk = 30
     sourceSets {
         getByName("main").apply {
             manifest.srcFile("AndroidManifest.xml")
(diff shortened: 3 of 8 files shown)


SomeTroglodyte:
Oh, you reverted. Well, some of the changes in here might still help in the long run and not hurt.

Primary lesson _I_ learned: Use terminal, `export JAVA_HOME = <some openjdk 11 or whatever matches github's VM>`, then ./gradlew clean --warning-mode all` and `./gradlew check --warning-mode all`... That, and Studio's inspections actually help you nicely in these build scripts.

buildSrc/build.gradle.kts
```diff
@@ -3,5 +3,5 @@ plugins {
 }
 
 repositories {
-    jcenter()
+    mavenCentral()  // "JFrog announced JCenter's shutdown in February 2021"
```
yairm210, line 6: Gosh

yairm210:
I think the changes here are good changes in any case.
It's better to make these changes before upgrading the gradle version to work incrementally :)

SomeTroglodyte:
Hours of trial and error, hoping to save _you_ a portion of that time. Adopting - would be part 'yes please why not', part 'would need to check whether that's not introduced until gradle 7', and part political - like what exactly would suffer if we dropped sourceCompatibility = JavaVersion.VERSION_1_6.

I haven't actually _tested_ the syntactical changes with gradle 6, but would guess good chances.

Note on lint - that Android lint: disable("all") _does_ what it sounds like, but in my tests I _still_ got crash dumps from the linting process. Since I fixed that "MissingTranslation" I went back to default - and that `isAbortOnError = false` is default, too. But, as said, the config in this PR runs fine locally but not as github action - at least not on my fork.

yairm210:
A. On upgrading to sourceCompatibility 1.7, that's an inevitability of time. If it does exclude some users, they're unsupportable almost by definition

B. We can start with small changes and see what works, not everything needs to be in the same pr

C. Any change that breaks the github actions breaks all tests and deployments, so that's obviously a no-go

yairm210:
All implemented and running :)

PR #4863, opened 2021-08-14
Implemented religious pressure

xlenstra:
This PR further implements parts of religion, see #4290.
In this PR, religious pressure is added. This mostly results in the following changes:
- Cities can have multiple religions inside them, with each population following any religion or none at all.
The amount of population following each religion has roughly the same proportions as the amount of pressure for that religion.
- When spreading a religion, you add pressure of your religion to the pool of pressure for that religion.
- Cities provide a small amount of pressure of their majority religion to cities close (<10 tiles) to them.

It also updates missionaries and great prophets, fully adding in the ability to spread faith in foreign territories. They are even able to enter territory of other civs without having an open borders agreement with that civ, but at a cost of 250 religious strength each turn. They start with 1000 religious strength, and die when they reach 0 religious strength. When spreading religion, their religious strength is added as pressure to the city they are spreading to.

Also made some behind-the-scene changes with how holy cities are saved.

Added uniques:
- "May enter foreign tiles without open borders, but loses [amount] religious strength each turn it ends there"


android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -1500,8 +1500,10 @@
 		"name": "Great Prophet",
 		"unitType": "Civilian",
 		"uniques": ["Can construct [Holy site] if it hasn't spread religion yet", "Can spread religion [4] times",
-			"May found a religion", "Great Person - [Faith]", "Unbuildable", "Religious Unit", "Hidden when religion is disabled"],
-		"movement": 2
+			"May found a religion", "Great Person - [Faith]", "Unbuildable", "Religious Unit", "Hidden when religion is disabled",
+			"May enter foreign tiles without open borders, but loses [250] religious strength each turn it ends there"],
```
SomeTroglodyte, line 1504: FYI: Unique order is preserved in civilopedia

core/src/com/unciv/Constants.kt
```diff
@@ -63,6 +63,9 @@ object Constants {
     const val unitActionParadrop = "Paradrop"
 
     const val futureTech = "Future Tech"
+    // Easter egg name. Hopefully is to hopefully avoid conflicts when later players can name their own religions.
+    // This religion name should never be displayed.
+    const val noReligionName = "The religion of TheLegend27"
```
SomeTroglodyte, line 68: From an _ad_???

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -619,6 +625,23 @@ class MapUnit {
...
+            && !civInfo.canPassThroughTiles(getTile().getOwner()!!)
+        ) {
+            val lostReligiousStrength =
+                getMatchingUniques("May enter foreign tiles without open borders, but loses [] religious strength each turn it ends there")
+                    .map { it.params[0].toInt() }
+                    .minOrNull()
+            if (lostReligiousStrength != null)
+                religiousStrengthLost += lostReligiousStrength
+            if (religiousStrengthLost >= baseUnit.religiousStrength) {
+                civInfo.addNotification("Your [${name}] was destroyed after spending too long inside enemy territory!", getTile().position, name)
```
SomeTroglodyte, line 641: Destroyed sounds harsh. Lost faith and ran away? :stuck_out_tongue_winking_eye:

core/src/com/unciv/Constants.kt
xlenstra: It was a huge meme at the time :p

SomeTroglodyte: Ouch. Lemme check my birth year - oh my, my car's number plate says 1940, no wonder...

core/src/com/unciv/logic/map/MapUnit.kt
SomeTroglodyte: ... ran away with the accountant's wife ... :poodle:

SomeTroglodyte:
Btw - didn't read your code for quality or pitfalls - just superficial skim, too distracted (rebooted that 4840 monster).

android/assets/jsons/translations/template.properties
```diff
@@ -533,6 +533,7 @@ Our influence with City-States has started dropping faster! =
 You gained [Stats] as your religion was spread to [cityName] = 
 You gained [Stats] as your religion was spread to an unknown city = 
 Your city [cityName] was converted to [religionName]! = 
+Your [unitName] lost its faith after spending too long inside enemy territory!" =
```
yairm210, line 536: Extra " at the end

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -16,18 +14,32 @@ class CityInfoReligionManager {
     val religionsAtSomePointAdopted: HashSet<String> = hashSetOf()
 
     private val pressures: Counter<String> = Counter()
+    // `getNumberOfFollowers()` was called a surprisingly large amount of time, so caching it feels useful
+    private val followers: Counter<String> = Counter()
```
yairm210, line 18: If this is a cached value, make it a @Transient

yairm210:
All in all looks good!

ravignir:
Prophets don't lose strength in civ5.

core/src/com/unciv/logic/city/CityReligion.kt
SomeTroglodyte: Sometimes just "reading" a save in notepadqq can give hints. We already got some stuff in there that could be avoided...

AdityaMH:
You missing follower type for each religion status on game screen

xlenstra:
> You missing follower type for each religion status on game screen

If you mean there is no UI to view what followers each city has: correct, that will be part of one of my next PR's. No need to do everything at once.

yairm210:
I feel that counting *down* the number of charges left instead of counting *up* the number of charges used is better, but I can't think of a practical use case where this makes a difference. In Civ VI individual units can gain extra charges, not sure if that's in V as well...

PR #4864, opened 2021-08-15
Make naval tradition provide a general instead of an admiral for now

logicminimal:
i didnt change the +2 movement for admirals to generals as it would make generals pretty op


logicminimal:
@yairm you can just close this if you dont want this

yairm210:
I'm not sure.
I think this is fine.

yairm:
How did I get CC'd on this :)?

I think you tried to reach yair morgenstern, according to the quoted mail
below..

Regards,

logicminimal:
i couldn't think that there could be someone nicknamed yairm sorry for the ping

PR #4869, opened 2021-08-15
City state bonuses

SomeTroglodyte:
Copy of #4664 with last master merged in - just to see whether it ends up OK


SomeTroglodyte:
OK, Simon's PR is "green", no need for this.

PR #4870, opened 2021-08-15
Fix DOS attack perpetrated by CityInfoReligionManager on Json Serializer

### LINKED ISSUE #4867: 3.16.4 will not Autosave
Massive stacktrace output on the console.

Investigating, more info soon.

Some `kotlin.SynchronizedLazyImpl` in the json serializer... Yup, that would do it.
pressureFromAdjacentCities$delegate

I'll have to come up with a permanent solution... That must be detectable in a unit test. The last time I've seen this is not too long ago, I think it was ... me ... whodunnit, but maybe locally only.

SomeTroglodyte:
Fixes #4867

### DIFF
--- a/core/src/com/unciv/logic/city/CityReligion.kt
+++ b/core/src/com/unciv/logic/city/CityReligion.kt
@@ -19,6 +19,7 @@ class CityInfoReligionManager {
     @Transient
     private val followers: Counter<String> = Counter()
     
+    @delegate:Transient
     private val pressureFromAdjacentCities: Int by lazy {
         when (cityInfo.civInfo.gameInfo.gameParameters.gameSpeed) {
             GameSpeed.Quick -> 9
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -62,7 +62,7 @@ class GameSettings {
     }
 
     fun addCompletedTutorialTask(tutorialTask: String) {
-        tutorialTasksCompleted.add(tutorialTask)
-        save()
+        if (tutorialTasksCompleted.add(tutorialTask))
+            save()
     }
 }


SomeTroglodyte:
As feared, a unit test _catching_ this is tricky, I'll have to teach some kotlin classes to that json beast first, so it can actually _throw_ instead of running into a stack overflow... But... not impossible I hope.

yairm210:
Does this also prevent regular saving? If so, I'll release a patch

SomeTroglodyte:
Yes it did (does). You get a broken save file, and a Save screen that doesn't display that its saving thread crashed...

It's the `lock` field in `class:kotlin.SynchronizedLazyImpl` that recurses on itself, because it's actually initialized as `private val lock = parameter ?: this` - and which they didn't annotate as nonserializable in a Serializable class. They thought nobody would be able to reflect in on that part, and I can't (I get the Lazy  interface instead), but Gdx somehow does.

xlenstra:
If you can wait three more hours before releasing the patch I'll also be able to solve a crash just opened on GitHub when a city is captured

SomeTroglodyte:
Asides: You will not get bitten with an empty game, but as soon as an instance of CityInfoReligionManager is visible to the Json serializer.

@xlenstra - I hope you got a chuckle out of the PR title. If I didn't have utmost respect for your skills I would never dare such a satire. Honestly, I'm that way - polite towards enemies or at least not entirely trusted people, and people I'm _positively_ prejudiced for, _they_ get the flak - because I trust them to be able to get it. Which not always works. :smiley_cat: !

And there's _super_ interesting lessons to be learned here.

xlenstra:
> @xlenstra - I hope you got a chuckle out of the PR title. If I didn't have utmost respect for your skills I would never dare such a satire. Honestly, I'm that way - polite towards enemies or at least not entirely trusted people, and people I'm _positively_ prejudiced for, _they_ get the flak - because I trust them to be able to get it. Which not always works. smiley_cat !


No worries, it certainly got a chuckle out of me :)

PR #4883, opened 2021-08-16
Removed the sight bonus from hills

avdstaaij:
This strikes off an item from #4697.


logicminimal:
shouldn't it provide extra sight though, as you technically can see more in real life also? @ravignir ?

ravignir:
No, but hills/mountains have other property: they are visible from further away than unit can normally see in civ5.

avdstaaij:
> No, but hills/mountains have other property: they are visible from further away than unit can normally see in civ5.

This is much more difficult to implement though, and deserves a PR of its own.

yairm210:
I districting recall that Civ V *does* provide extra sight for units in his, though.
Maybe this is only in Vanilla ruleset...

yairm210:
Ravignir is the resident rules expert, so I'll go by his judgement

ravignir:
nohill
archer is standing on a flat land.
hill
archer is standing on a hill.
this is vanilla civ5 (rules have not changed since).

PR #4886, opened 2021-08-16
Unit test catching un-annotated lazies

SomeTroglodyte:
After a whole lot of research (and blind poking around) I came up with this defense against future #4870 - like oversights.

_Both_ tests would have caught the recent lazy, but I've made them independent on purpose.
- One sees if our _default_ GameSaver Json can handle a small game. May be redundant.
- The other uses a hardened Json to look for un-annotated lazies, listing them by name (but without class, sorry).
(recursing the object model by reflection turned out too painful - I failed at seeing the transient annotations if present).
The minigame built may still be too incomplete to catch everything - Notifications come to mind, they would still be empty.

Hardening the GameSaver Json the same way as in the serializedLaziesTest isn't attractive as the saved games produced would still not be loadable due to the good old:
`Caused by: com.badlogic.gdx.utils.SerializationException: Class cannot be created (missing no-arg constructor): kotlin.SynchronizedLazyImpl`.

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -444,7 +444,7 @@ class Ruleset {
  * save all of the loaded rulesets somewhere for later use
  *  */
 object RulesetCache : HashMap<String,Ruleset>() {
-    fun loadRulesets(consoleMode: Boolean = false, printOutput: Boolean = false) {
+    fun loadRulesets(consoleMode: Boolean = false, printOutput: Boolean = false, noMods: Boolean = true) {
         clear()
         for (ruleset in BaseRuleset.values()) {
             val fileName = "jsons/${ruleset.fullName}"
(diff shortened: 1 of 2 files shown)


yairm210:
Interesting! Good work on both!

xlenstra (approved):

SomeTroglodyte:
I would -after having had time to mull it over- have opted to keep only one of these tests, with a slight preference for the second, as it does not need to wait for a stack overflow crash when it fails. But - they can be switched on an off anytime by just commenting out the @Test annotation.

For reference, for the really curious:
- The source of the problem is [Kotlin-LazyJVM.kt-Line 59](https://github.com/JetBrains/kotlin/blob/5e87d753b777427dbc10d6ce66bbccd14761037e/libraries/stdlib/jvm/src/kotlin/util/LazyJVM.kt#L59), where the infinite loop is created.
- And why we can't hook it is [here in Gdx Json.java line 515](https://github.com/libgdx/libgdx/blob/af899350378fc0908eb26867b8cc44fd3cfdce7c/gdx/src/com/badlogic/gdx/utils/Json.java#L515) - that array can be filled by `Json.setSerializer`, where one must provide a class type object. It's a generic, thus type must be given at compile time. The array is indexed by reflected `value.getClass()`, thus by `SafePublicationLazyImpl`, which is private and inaccessible to call a generic with - if they indexed by `knownType` instead, that would come from the field's reflection metadata and be the `Lazy` _interface_, which we _could_ provide to the setSerializer overriding hooks...
- Some discussion that helped me understand this is in [libgdx/#6629](https://github.com/libgdx/libgdx/issues/6629)

...Just for completeness.

Issue #4826, opened 2021-08-10
Improve modding documentation checklist


SomeTroglodyte:
## Improve modding documentation

- [x] [Document _all_ moddable json files](../../Unciv/wiki/JSON-files-for-mods)
- - [x] Terrains.json
- - [x] TileImprovements.json
- - [x] TileResources.json
- - [x] All others
- - [x] FantasyHex.json
- [x] TOC ordering
- [ ] Improve / tighten general json syntax chapter
- [ ] Mention relaxed json standard?
- [x] Split into sub-pages including fixing links
- [x] TOC to Sidebar
- [ ] [Uniques roadmap / what's missing?](../..//wiki/Uniques)
- [ ] Uniques improve structure and TOC
- [x] Check and improve information on game.atlas

Help or input welcome. If you plan on editing a page let's synchronize here. (Hi @xlenstra - thanks for the excellent work)

SomeTroglodyte:
@ravignir  - if I may be so bold - what I wrote for the TileSet jsons [(here)](https://github.com/yairm210/Unciv/wiki/JSON-files-for-mods#tileset-specific-json) isn't complete, Could you please take a look?

logicminimal:
y closed?

PR #4888, opened 2021-08-16
Removed the civ introduction trade option

### LINKED ISSUE #4272: Abuse of introduction to other civ
In the beggining of the game as soon as you meet first two civs you can offer them a trade to introduce one to another (if they didn't meet yet) for a good price and they will accept it. Also if you send each of them offer at the same turn they both can accept it and you will receive double reward. So the problem is AI doesn't take into account that it's just the very beggining of the game and it will probably meet other civilization soon enough.

The solution might be simply checking if amount of turns passed are too low, and if it is AI should reject the trade offer.

Another, more complex solution is to make reward for introduction dependent of turns passed, and maybe it can also take map size into account.

avdstaaij:
This strikes off an item from #4697 and resolves #4272.

I left in the minimum amount of code that ensures that saves with a pending introduction trade do not break. Any pending trades with an introduction offer are cancelled, and clicking on a pending introduction offer in the trade overview table removes it.
The left-in introduction trade code can be fully removed in a future version.

If we do not want to remove introduction trades after all, #4272 can probably also be fixed by cancelling pending introduction trades when the receiver already knows the introduced civ (in `TradeEvaluation.isOfferValid`).

### DIFF
--- a/core/src/com/unciv/logic/trade/TradeEvaluation.kt
+++ b/core/src/com/unciv/logic/trade/TradeEvaluation.kt
@@ -277,6 +277,8 @@ class TradeEvaluation {
     }
 
     private fun introductionValue(ruleSet: Ruleset): Int {
-        return ruleSet.modOptions.uniqueObjects.firstOrNull{ it.placeholderText == ModOptionsConstants.tradeCivIntroductions }?.let{ it.params[0].toInt() } ?: 0
+        val unique = ruleSet.modOptions.uniqueObjects.firstOrNull{ it.placeholderText == ModOptionsConstants.tradeCivIntroductions }
+        if (unique == null) return 0
+        return unique.params[0].toInt()
     }
 }
--- a/core/src/com/unciv/logic/trade/TradeEvaluation.kt
+++ b/core/src/com/unciv/logic/trade/TradeEvaluation.kt
@@ -6,6 +6,8 @@ import com.unciv.logic.automation.ThreatLevel
 import com.unciv.logic.city.CityInfo
 import com.unciv.logic.civilization.CivilizationInfo
 import com.unciv.logic.civilization.diplomacy.RelationshipLevel
+import com.unciv.models.ruleset.ModOptionsConstants
+import com.unciv.models.ruleset.Ruleset
 import com.unciv.models.ruleset.tile.ResourceType
 import kotlin.math.min
 import kotlin.math.sqrt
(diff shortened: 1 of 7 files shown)


yairm210:
Since other versions of Civ do have this feature, it should be enablable via a modOptions unique, bot removed entirely from the game.

The amount of requests I still receive for enabling tech trade... 🙃

core/src/com/unciv/logic/trade/TradeType.kt
```diff
@@ -15,7 +15,9 @@ enum class TradeType(val numberType: TradeTypeNumberType, val isImmediate: Boole
     Luxury_Resource     (TradeTypeNumberType.Simple, false),
     Strategic_Resource  (TradeTypeNumberType.Simple, false),
     Technology          (TradeTypeNumberType.None, true),
-    Introduction        (TradeTypeNumberType.None, true),
```
SomeTroglodyte, line 18: By the way, @Deprecated("message") works just fine on _individual_ enum Instances. Point moot if made moddable.

avdstaaij: I felt like "deprecated" was not the right term here, as the feature was essentially completely removed (Introductions could no longer be offered). The left-in code was just to convert old saves. But like you said, it does not really matter anyway since I will un-remove it and put it behind a modOption.

avdstaaij:
I added a modOption unique to enable introduction trades, and parametrized their value while I was at it:
"Can trade civilization introductions for [amount] Gold"

I also fixed #4272 in the alternative way that I described, so this still closes #4272.

core/src/com/unciv/logic/trade/TradeEvaluation.kt
```diff
@@ -274,4 +276,7 @@ class TradeEvaluation {
         }
     }
 
+    private fun introductionValue(ruleSet: Ruleset): Int {
+        return ruleSet.modOptions.uniqueObjects.firstOrNull{ it.placeholderText == ModOptionsConstants.tradeCivIntroductions }?.let{ it.params[0].toInt() } ?: 0
```
xlenstra, line 280: I don't particularly like how long this line is and how many times the 'null' is passed on to the next statement, but other than first checking existence I don't see a simple alternative

xlenstra (approved):

core/src/com/unciv/logic/trade/TradeEvaluation.kt
yairm210: Concur - split this into val unique = ...firstornull, then later checks

yairm210:
About documentation in the wiki, I can think of 2 places this could go:
A. The uniques file
B. The file that explains about modOptions, among others - https://github.com/yairm210/Unciv/wiki/Miscellaneous-JSON-files#modoptionsjson

@SomeTroglodyte has taken over the reorg of the wiki - what say you? The link there to modOptions uniques is a dead end

SomeTroglodyte:
Yes I'd say 3 options (number 3 being a copy in both places - nah). Prio one - link the two places, bidi -> fixed. Prio two - _all_ applying uniques in one place, or at least all applying _documented_ uniques in one place. OK, that is fine [in the Uniques page](/yairm210/Unciv/wiki/Uniques#modoptions-uniques). Then - just text away.

As for "Taken over" - that is more than I aimed for. I started out with "just" the goal to complete the field-by-field dox for all possible json's, which became the road map at #4826 .

~~"The link" - umm, can't find it.~~ OK - the one in the modOptions table. Same thing applies as in the following paragraph.

What I found was non-functional links using `../` notation - which worked in preview but not when saved (as it would then remove the `/wiki/`part of the URL). There's other places where `../` works but `./` does not - I replaced the ones I found with absolute (but without the protocol and host) ones - that should be robust. Thing is - in at least one place I didn't change the nonfuctional link, nor did I change the destination address. I'll take that as lesson: Don't rely on relative addressing within the wiki. Seems sometimes the anchor of a relative is relative to is the current page, sometimes it's the wiki home...

PR #4889, opened 2021-08-17
Made water oil wells require the Refrigeration tech

avdstaaij:
This strikes off an item from #4697.

Added unique: "Cannot be built on [tileFilter] until [techName] is discovered".

This is not the ideal solution: it would be preferable to support multiple improvements for a single resource, to add offshore platforms as a separate improvement for coastal oil that requires refrigeration and to disable building oil wells on coast. That way, we could also distinguish between oil wells and offshore platforms in other cases. Allowing multiple improvements for a single resource would however be a much more significant refactor. This works for now.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/TileImprovements.json
+++ b/android/assets/jsons/Civ V - Vanilla/TileImprovements.json
@@ -49,9 +49,9 @@
 	},
 	{
 		"name": "Oil well",
-		"terrainsCanBeBuiltOn": ["Coast"],
 		"turnsToBuild": 9,
 		"techRequired": "Biology",
+		"uniques": ["Cannot be built on [Coast] tiles until [Refrigeration] is discovered"],
 		"shortcutKey": "W"
 	},
 	{
@@ -78,7 +78,6 @@
 	},
 	{
 		"name": "Fishing Boats",
-		"terrainsCanBeBuiltOn": ["Coast"],
 		"food": 1,
 		"techRequired": "Sailing",
 		"uniques": ["[+1 Gold] once [Compass] is discovered"]
--- a/android/assets/jsons/Civ V - Vanilla/TileImprovements.json
+++ b/android/assets/jsons/Civ V - Vanilla/TileImprovements.json
@@ -49,6 +49,7 @@
 	},
 	{
 		"name": "Oil well",
+		"terrainsCanBeBuiltOn": ["Coast"],
 		"turnsToBuild": 9,
 		"techRequired": "Biology",
 		"uniques": ["Cannot be built on [Coast] tiles until [Refrigeration] is discovered"],
@@ -78,6 +79,7 @@
 	},
 	{
 		"name": "Fishing Boats",
+		"terrainsCanBeBuiltOn": ["Coast"],
 		"food": 1,
 		"techRequired": "Sailing",
 		"uniques": ["[+1 Gold] once [Compass] is discovered"]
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
Unit test improvementsCanNOTBeBuiltOnWrongResource FAILED for Fishing boats -> not _so_ surprising given the subject..
Actually, `TileInfo.canBuildImprovement()` returns true for fishing boats on a Plains+Cattle tile.

```kotlin
            // Tiles with no terrains, and no turns to build, are like great improvements - they're placeable
            improvement.terrainsCanBeBuiltOn.isEmpty() && improvement.turnsToBuild == 0 && isLand -> true
```
"Tiles" with no terrains? Starts with bad comment wording... And conceptually this is prone to... Produce exactly this failure. I never liked that `turnsToBuild==0` got assigned such side effects. I'll trust you to think this through. Just adding back that 'Coast' would be the cheap solution.

Side note: You know you can run these tests locally with breakpoints and all, right?

avdstaaij:
> Unit test improvementsCanNOTBeBuiltOnWrongResource FAILED for Fishing boats -> not _so_ surprising given the subject..
> Actually, `TileInfo.canBuildImprovement()` returns true for fishing boats on a Plains+Cattle tile.
> 
> ```kotlin
>             // Tiles with no terrains, and no turns to build, are like great improvements - they're placeable
>             improvement.terrainsCanBeBuiltOn.isEmpty() && improvement.turnsToBuild == 0 && isLand -> true
> ```
> 
> "Tiles" with no terrains? Starts with bad comment wording... And conceptually this is prone to... Produce exactly this failure. I never liked that `turnsToBuild==0` got assigned such side effects. I'll trust you to think this through. Just adding back that 'Coast' would be the cheap solution.
>
> Side note: You know you can run these tests locally with breakpoints and all, right?

I did notice the failing test after pushing and I had already investigated it as well, but I didn't have time to address it then. Good thing that the tests caught this bug! I came to about the same conclusions as you.

I must say that `TileInfo.canBuildImprovement` and `TileInfo.canImprovementBeBuiltHere` are some exquisite spaghetti. For water improvements, `terrainsCanBeBuiltOn` has to be non-empty, because they can otherwise be built everywhere on land due to the line you mentioned (and yet I can't build fishing boats on land when I try, so there must be some other check that I haven't found yet). Since `terrainsCanBeBuiltOn` can't be empty for water improvements, it makes sense to set it to `["Coast"]` (like on `master`), as water improvements can only be built on coast. _However_, that means that those improvements can be built on _all_ coast tiles according to `TileInfo.canImprovementBeBuiltHere`, even when they don't have a resource. To circumvent _this_ issue, there is an additional check in `UnitActions.getWaterImprovementAction` that prevents the building of water improvements on tiles without a resource (so mods can't add non-resource water improvements either). Actually, from looking at the code of `master`, it seems possible to build an oil well on oil you can't see yet: the check in `getWaterImprovementAction` does not consider whether the resource is visible. I haven't tested that, though.

This system probably needs a complete rewrite (including support for multiple improvements for a single resource). For now, I'll just add back `"terrainsCanBeBuiltOn": ["Coast"]`. I don't dare doing anything more.

yairm210:
I agree with the 'multiple improvements'

There are two ways of doing this:
A. "Provides [amount] resources when worked by [improvement]" on resource side
B.  "Provides [amount] [resource] when built on resource" on improvement side

I personally favour A because there are more improvements serving multiple resources than the other way around
This means moving the link from the resou

yairm210:
Also agree about the spaghetti... That's what you get from years of patch-on-patch.

The way I see it, terrainsCanBeBuiltOn should never be empty, but should for water tiles contain "Water" and for land tiles "Land", using tileFilter to merge at least 4 checks into 1

SomeTroglodyte:
I guess it's pure :ghost:  fear that prevents us from putting _some_ code thoroughly through the laundromat...

PR #4890, opened 2021-08-17
Implemented almost all missing founder & follower beliefs

xlenstra:
With the addition of religious pressure and followers in cities, most of the remaining follower & founder beliefs could be implemented easily.
This PR accomplishes this.
Only the follower belief 'Holy Warriors' remains, mostly because I need to invest the time to look up how much faith each unit should cost and how to best implement that.


core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -598,6 +600,11 @@ class CityInfo {
             "in all cities in which the majority religion is a major religion" -> 
                 religion.getMajorityReligion() != null
                 && civInfo.gameInfo.religions[religion.getMajorityReligion()]!!.isMajorReligion()
+            "in non-enemy foreign cities" ->
+                viewingCiv != civInfo
+                && civInfo.isAtWarWith(viewingCiv)
```
SomeTroglodyte, line 605: "Non-" ?

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -368,6 +369,15 @@ class CityStats {
             for (unique in uniques.filter { it.placeholderText == "[]% [] while the empire is happy"})
                 stats.add(Stat.valueOf(unique.params[1]), unique.params[0].toFloat())
         }
+        
+        for (unique in uniques.filter { it.placeholderText == "[]% [] from every follower, up to []%" })
+            stats.add(
+                Stat.valueOf(unique.params[1]),
```
SomeTroglodyte, line 375: We just hat the same in Simon's PR - I still feel the potential Exception when a modder writes garbage into param 1 needs addressing. Either harden and silently ignore - or have a check in mod checker - or at least document the prone uniques so such a check might be added later.   ---?

xlenstra: > We just had the same thing in Simon's PR - I still feel the potential Exception when a modder writes garbage into param 1 needs addressing. Either harden and silently ignore - or have a check in mod checker - or at least document the prone uniques so such a check might be added later.   ---?

I vote for checking this in the mod links, as we assume this for almost literally every unique, and there are a lot of them. Checking all of these at runtime feels like unnecessary performance loss. 
Doing it at mod loading time would require effectively a complete database of possible values for all uniques or something to determine what params should be what, but that would also solve potential problems with typo's in the source code.

SomeTroglodyte: Yup, thinking close. Maybe not a complete database but only those prone to lead to difficult to detect errors. And if the modchecker catches the error, this call site should silently ignore.

Do we want a generic `Enum.valueOfOrNull`? I don't think one is possible. One with a signature valueOfOrNull<Enum>(name): Enum? is not too hard (except you need `reified`), it could console log then return null and the client does Elvis to default it. Or it could silently return non-null values().first(). The extension-like signature could then be a very short call in each Enum's companion.

xlenstra: > Yup, thinking close. Maybe not a complete database but only those prone to lead to difficult to detect errors. And if the modchecker catches the error, this call site should silently ignore.
>
> Do we want a generic `Enum.valueOfOrNull`? I don't think one is possible. One with a signature valueOfOrNull(name): Enum? is not too hard (except you need `reified`), it could console log then return null and the client does Elvis to default it. Or it could silently return non-null values().first(). The extension-like signature could then be a very short call in each Enum's companion.

In my personal opinion detecting a faulty unique at runtime is more hassle than it's worth. Detect the faulty unique when loading in the mod, and give an error that it is faulty, refusing to further load the mod. The mod maker will have to find this and repair it when testing their own mod, and thus fix the unique before it can ever be used. Might be annoying for the mod maker, but at least this way the mod maker also knows for sure his uniques have correct parameters.

Having a database (in what shape or form it may turn out to be) of all uniques instead of only those prone to have errors has two more direct advantages:
- Firstly, we can check every unique against the entire database, and therefore can detect uniques with minor spelling mistakes easily, fully guaranteeing that all uniques in a mod actually work.
- Secondly, it will provide an additional source for all uniques, in case the wiki lags behind. I wouldn't be surprised if tens of uniques are currently missing from the wiki as they were never added to begin with, and such a database can easily add all these. If at some later point a unique is not added to the wiki, this can be rectified easier and earlier when we have a list of all possible valid uniques somewhere already.

Maybe at some later point we can even remove string comparisons to check for uniques and somehow interact with this database which might result in a small performance boost

SomeTroglodyte: Last paragraph -> Enum!

core/src/com/unciv/logic/city/CityReligion.kt
```diff
@@ -159,11 +159,21 @@ class CityInfoReligionManager {
...
+        return followers[religion]
+    }
+    
+    fun getFollowersOfMajorityReligion(): Int {
+        val majorityReligion = getMajorityReligion() ?: return 0
+        return followers[majorityReligion]!!
+    }
+    
+    fun getFollowersOfOtherReligionsThan(religion: String): Int {
+        var totalFollowers = 0
```
yairm210, line 172: Return followers.filterNot{...}.sum()

core/src/com/unciv/logic/city/CityStats.kt
yairm210: What's nice is that for uniques in the base game we already have such a database, called the translation files.
Since the placeholders get auto-named [stat], [stats] etc we can easily check this by taking the English translations, looking only at the keys

core/src/com/unciv/models/stats/Stats.kt
```diff
@@ -118,6 +118,16 @@ open class Stats(
         happiness *= number
         faith *= number
     }
+    
+    operator fun div(number: Float) = Stats(
```
yairm210, line 122: Nice!

yairm210: Hm I wonder, we could implement this as simply "times(1/number)", performance difference would be basically non existent

xlenstra: Probably better for future changability

SomeTroglodyte: We just optimized the hell out of Stats, and Xander just followed the pattern. I agree gain is negligible, simpler has precedence. We could just have gone with "[+0.5 Stat] for every global follower..." and avoided the div :wink:

PR #4897, opened 2021-08-18
Update French.properties

Marjolain:
Some traductions


SomeTroglodyte:
Mon français limité dirait: Bien fait. :+1:

Marjolain:
Thanks !
You can write french, so, bravo !
I use mostly deepl.com and I adapt.
I need context some time, but I will see in game.

Thanks for the good game !

Envoyé depuis ProtonMail mobile

-------- Message d'origine --------

PR #4898, opened 2021-08-18
Update Russian.properties

Grom1halo:
discord: Grom1halo#7692

PR #4899, opened 2021-08-18
Update Russian.properties

Grom1halo:
discord: Grom1halo#7692


xlenstra:
As this seems only a part of the translations added in #4898 and those added in both this PR and that are the same, I'm closing this.

Issue #4806, opened 2021-08-08
you should be able to raze non original capitals of cities


logicminimal:
if a civs capital is a and i take it. their current capital is b and when i take it too i cant raze it

SomeTroglodyte:
Annex then raze works for now.

logicminimal:
> Annex then raze works for now.

didnt know that BUT this is still a problem as you cant directly raze it

PR #4903, opened 2021-08-18
can now raze non-original capitals at capture

SimonCeder:
Tiny PR, relates to #4806 
Removes the check for `isCapital` at capture. I elected to create a new function for this so it can be more easily updated if we change the conditions for destroying cities in the future.


xlenstra (approved):

SomeTroglodyte:
Slight problem - the AI hasn't the same rights.

Source of the problem: When isCapital is queried, the city is still in possession of the defender. puppetCity and thus moveToCiv and thus the Palace removing code is run _after_ the decision to offer a Raze.

So big question - can the workflow change safely? Then no kludge would be needed.

If we go with the kludge - thorough documentation right there please? And look in Battle.conquerCity at the else of `if (attackerCiv.isPlayerCivilization())` please?

Alternate approach: parameter `ignorePalaceDueToConquestInProgress` to the original function - the name is not meant literally of course - please weigh options by readability and risk of future coders choosing the wrong test in some new code.

yairm210:
I'm not sure what you mean by kludge, but everything is changeable if you find a better way.

I'll go over the limitations.
When you capture a city in Civ V, your unit enters the city, and then you decide what to do with it.
Basically you have 2 main options - return or puppet.
Puppeted cities can be annexed, annexed cities can be razed, but you could do all this in later turns as well. The non-reversible choice is puppet vs return.
The capturing unit *enters the city capturing civilians* as part of the battle, before the choice is presented.

Currently, the city remains part of the defender's cities until the choice is made.
We could change this, to have the city automatically be puppeted by the conquering Civ, and presenting a choice to return the city to owner or do all the other stuff

SomeTroglodyte:
So removing the Palace from the city before the question 'what to do with it' is presented is not a good option because you could decide to return it... If we puppet first and then return we would need to make sure the Palace is rebuilt when that resurrects a Civ? Or leave as is and then we would indeed need two different ways to test for 'is razing allowed'. In the latter case let's just do it as best possible with an eye on potential new coders and help them choose the correct test.

yairm210:
There's no problem in pupetting first, we can remove the palace, it'll get re-added if necessary after the move to the defeated Civ anyway

yairm210:
My main question is, what is the problem you're encountering now, and what is your proposed solution
So much of the code has grown organically piece by piece that reorgs and rewrites are to be expected

SimonCeder:
Puppeting first seems like a good option. I looked into it and in Civ 5 you get gold, and reduce population even when liberating so that would even be an improvement. Only thing is that if you end up liberating you need to undo the warmonger penalties that are added when puppeting (or move the diplomatic repercussions call outside of `puppetCity`)

SomeTroglodyte:
I just wanted to sensitize a bit and broaden the horizon - precisely for long-term code quality. I'm not proposing "a" solution because I don't want to carry the responsibility to test. Yup, lazy bastahd.
Simon's first proposal has the advantage that it's easy to see that it would work perfectly as intended right now (except AI's would still be prevented from razing in those rare situations). I'm happy to trust him to evaluate and choose the best balance.

SimonCeder:
Here's my proposal. Still little is changed. Merged both flavors of `canBeDestroyed` into a single function with a default parameter for minimum future confusion.
Puppeting before choosing is more difficult than it seems at first glance because of the diplomatic penalties, and needing to remember the `oldCiv`

SomeTroglodyte (approved):
Seems fine - informed decision, and "conflicts" seems to be just the trailing blanks on that one line.

yairm210:
Tried to solve conflicts via UI since it looked simple enough. Failed terribly.
This is why I usually tell people to fix the conflicts themselves...

SimonCeder:
No worries, I would have done it myself yesterday but it took me a while to figure out these whitspace conflicts.

PR #4904, opened 2021-08-18
Added a UI for viewing the religions inside a city

xlenstra:
This PR adds a small UI for viewing the followers of religions inside a city.

<details> <summary> Picture: </summary>

image

</details>

Feedback is welcome.

Note that this PR will have merge conflicts with both #4902 and #4890.

PR #4905, opened 2021-08-18
Expander tab persist

### LINKED ISSUE #4807: UI suggestion: Don't re expand city build selections after first selection
When I go into a city screen, and I don't want to think about units or wonders, because I know I want to build a building / city improvement, I collapse those other categories with that little downwards triangle. This is great.

But then I try to click through the available options, to compare the details of different options, and suddenly those other categories have un-collapsed, and my scroll position doesn't recalculate. So I have to scroll far down, just as if I never collapsed those other categories in the first place. This defeats the entire benefit of using the little "collapse category" arrow in the first place.

I suggest not re-expanding categories just because the user clicked on something. Or at the very least, if you have to re-expand, recalculate the scroll position such that "the thing that was just clicked on" is still visible in the same spot it was a moment ago.

SomeTroglodyte:
This as usual got to be more than planned. Would resolve #4807
1. ExpanderTab persistence simpler, app lifetime scope.
2. City Screen rework so it doesn't break up visually as a consequence of the Expanders not resetting same way as before.
3. Tweaks to make the City Screen a bit nicer in portrait mode.

I can split it up if you want, but 1. looks horrible in some situations without 2.

Notes on 1
- ExpanderTab API is exploding, and the idea has just popped up to allow more control over header formatting (#4904 ) - but I don' like opening up internal data of such a "library" kind of object, and the needed parameterization is all known at instance creation, so - either keep adding params and use them by name or introduce an *Options class as some other libraries do. Not sure.
- Storage lives in the Companion for now - moving isn't too hard a refactor should voices get too loud about longer persistence.

Notes on 2
- Had to split the left side into two independent widgets as an empty table row with expandY in the middle didn't quite cut it. Two Tables each glued to a corner is easier to manage. However, to keep them talking to each other without hassle, data stayed in one class - which would mandate a rename to clarify it's no longer a widget, and that would mandate a file rename - which in turn makes overlapping merges _much_ harder IMHO, and a git rename seems fragile to me, too, I'm prejudiced they break up out of the blue into a delete+new without a discernible reason. So I didn't rename - yet?.
- Made the map part a Zoomable, as alternative to completely killing mouse wheel function - cuz that was more annoying than useful.

Notes on 3, in order of importance - individually open to discussion.
- pager widget, exit and raze/annex buttons never get covered by other widgets simply by Z-order - previously one-city-back was most of the time covered in portrait.
- Lower center widgets (pager, exit) -_if in portrait_- now center between the widget edges to left and right, no longer to screen, to minimize overlap
- Upper center (raze button) does the same
- The map part centers the city center between the permanent widget edges, no longer on screen center
- Problem: Due to the background thread the final width of the available constructions widget is not yet known at centering time, so it uses the construction queue width as imperfect estimate

<details><summary>Portrait screenshot</summary>

image
</details>

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
@@ -20,7 +20,13 @@ import kotlin.math.max
 import kotlin.math.min
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
-class CityConstructionsTable(val cityScreen: CityScreen) {
+/**
+ * Manager to hold and coordinate two widgets for the city screen left side:
+ * - Construction queue with switch to [ConstructionInfoTable] button and the enqueue / buy buttons.
+ *   The queue is scrollable, limited to one third of the stage height.
+ * - Available constructions display, scrolling, grouped with expanders and therefore of dynamic height.
+ */
+class CityConstructionsTable(private val cityScreen: CityScreen) {
     /* -1 = Nothing, >= 0 queue entry (0 = current construction) */
     private var selectedQueueEntry = -1 // None
     var improvementBuildingToConstruct: Building? = null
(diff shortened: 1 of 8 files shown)


yairm210:
Looks good!
I think the expander tab solution is better than it being in a permanent state, and since it's in the companion it'll remain between turns, which is nice
I'll take your word for all the UI stuff, as previously mentioned you're obviously better at that than I am :)

PR #4907, opened 2021-08-19
Ancient Ruins, Civilopedia and Translations

SomeTroglodyte:
Simple idea - display actual Ruins.json in Pedia - just shows need for discussion:
- Reward notifications currently untranslated? Or are they added manually to templates?
- First json with a name property unsuited for public display
- Rewards with multiple uniques would need a title to bind them visually together, but which? notification without context doesn't display well in all cases.
- Or stay with manually curating the loose description in the improvements json, but keep it up to date? Couldn't filter on religion switch in that case. But - translations exist. At least 3 sets not done by myself.

PR changes:
- TranslationFileWriter can process Ruins.json
- Hide Religion / Civilopedia uniques hardcoded String moved to Constants
- Civilopedia display of Ruins pulls actual json rewards in code
- Manually curated rewards description removed

<details><summary>Screens, old then new</summary>

image
image
</details>


logicminimal:
i think there should be icons right to the left to the each paragraph of text and i think there should be atleast 1 newlines inbetween each paragraph

also i think the "the possible rewards are" should stay

yairm210:
- Notifications should not be displayed
- Based on json is better than manual, especially considering mods

xlenstra (commented):
Diff looks fine.
On the pictures: I agree the notification should not be shown, especially as it has parameters in it ("the lost technology of [techName]", etc.). If it makes it clearer, feel free to change the names to something that could be shown to the user instead. Also, those stars did look pretty, any chance of them being revived for the new version :upside_down_face:?

For other things, I agree with logicMinimal that the 'possible rewards are' line would improve clarity on the lines below, and some more whitespace might also help. Just a thought, feel free to try out what works best

SomeTroglodyte:
> notification should not be shown, especially as it has parameters

I said so didn't I :partying_face: ?  OK - that idea to repurpose the name field of course is almost obvious, hearing it from you without prompting almost decides it. Didn't check for other uses though. The stars are no problem, their hues are (where to store them).

So now I'd tend to:
- drop what it's displaying now, use the name property. Re-text all of them close to what the current civilopediaText has, or the notification, but keep it short.
- No uniques listed, the name must be clear enough - in the sample most are just a reiteration in more technical terms
- Re-simplify TranslationFileWriter, untranslatable fields by class+field may be useful in the future, but no longer needed now
- Put the star's hue (it's a Stat hue btw) in Ruins.json

**OR**

Have RuinReward implement ICivilopediaText and simply flatMap those - much simpler code, markedly more json, more flexibility for modders. name would stay as is, as would the extension to exclude it in TranslationFileWriter...

SomeTroglodyte:
Question - if _this_ doesn't show the uniques from Ruins.json, nothing does, and they should be excluded from TranslationFileWriter, right?

SomeTroglodyte:
Update contains:
- Old rewards descriptions moved to RuinReward.name, color added
- Code now recreates old output closely
- civilopediaText allowed additionally, tested, but unused in vanilla Ruins.json
- TranslationFileWriter tweaked accordingly, tested to preserve existing translations
- A deeper BugFix to TranslationFileWriter is included: charset for reading was different from writing - system default vs. explicit utf8. This may have repeatedly removed the [stargazers] due to that star - if it was run on windows.

SomeTroglodyte:
<details><summary>New Screenshot</summary>

image
</details>

Note how it doesn't show the worker/settler rewards as that game is on deity.

logicminimal:
> <details><summary>New Screenshot</summary>
> 
> image
> </details>
> 
> Note how it doesn't show the worker/settler rewards as that game is on deity.

i think the specific stats or stat value ranges should be preserved

SomeTroglodyte:
Warning - this "breaks" the do-not-repeat-a-reward-too-soon function of RuinsManager with pre-PR saved games, as the "keys" change. IMHO nobody will notice. - But me. And it took me a few hours to see.

PR #4913, opened 2021-08-19
Implemented Unit Supply

r3versi:
Unit Supply (#4697) is a soft cap to the number of units. If the number of units of a civilization exceeds the total supply, a production malus is applied.

Total Supply is made of 
* **Base Supply**, dependant on difficulty
* **Number of cities** times a modifier dependant on difficulty
* **Total population**/2

Each unit consumes one supply. Each point of deficit yields a 10% production malus (capped at 7 points, hence 70%).

supply1
supply2


logicminimal:
will unit supply and its difficulty modifiers be included in the difficulties page of civilopedia?

yairm210:
This needs to be stressed somehow when choosing to construct units, because if someone is getting a -70% on production and he doesn't understand why, that's a really bad gameplay experience

yairm210:
Also, please make this option *opt in* from the base ruleset modOptions file - many mods would not suit this cap

logicminimal:
time to make an unciv no unit supply mod

SomeTroglodyte:
Welcome back, @r3versi  !

 I say: looks good :+1:  Civilopedia Info can come later - better decide moddability first. Switch or moddable factor/limit would do. What stuck _me_ was - silly I know - that the additional panel on the Unit overview might not fit too well in portrait orientation. Some `if (screen.isPortrait()) whateverTable.center().row()` or so might already help.

ravignir:
What about displaying it on the top bar like supply used/limit. Once player clicks on it, it would open the supply screen where all of current modifiers are listed and mechanic is explained. That is more work tho.

android/assets/jsons/translations/template.properties
```diff
@@ -789,6 +789,13 @@ Transportation upkeep =
 Unit upkeep = 
 Trades = 
 Units = 
+Unit Supply = 
+Base Supply = 
+Total Supply = 
+In Use = 
+Supply Deficit = 
+Production Penalty = 
+Increase your supply or reduce the amount of units to remove the production malus =
```
yairm210, line 798: use "penalty" instead of "malus", it's a nice word but not in line with our goal of being as simple to understand as possible :)

core/src/com/unciv/ui/overviewscreen/UnitOverviewTable.kt
```diff
@@ -14,59 +15,117 @@ import kotlin.math.abs
...
+            add(viewingPlayer.getCivUnitsSize().toLabel()).right().row()
+            addSeparator()
+            val deficit = viewingPlayer.stats().getUnitSupplyDeficit()
+            add("Supply Deficit".tr()).left()
+            add(deficit.toLabel()).right().row()
+            add("Production Penalty".tr()).left()
+            add((viewingPlayer.stats().getUnitSupplyProductionMalus()).toInt().toString()+"%").right().row()
+            if (deficit > 0) {
+                add("Increase your supply or reduce the amount of units to remove the production malus"
+                    .toLabel().apply { wrap = true; color = Color.FIREBRICK })
```
yairm210, line 53: .toLabel() accepts a color parameter :)
Not that this makes a big difference

yairm210:
All in all looks good!
There's game.png conflicts, so pull and rebuild :)

SomeTroglodyte:
@yairm210  seems @r3versi is taking a well-deserved sabbatical. I know I can merge this ([see here](https://github.com/SomeTroglodyte/Unciv/tree/new/unit-supply) - the atlas is still the only thing needing manual attention), but can we then keep 100% of the credit on r3versi's karma? Would be a pity to skip this.

yairm210:
Of course :)

yairm210:
If you're already at it, can you change the "production malus" to "production penalty"?

SomeTroglodyte:
#5234 is ***this***,
- Merged up to current master (up to 5230)
- Pennatly instead of darth mauls, both text and function name (actually, malus is pretty common in gaming circles I think) :stuck_out_tongue_winking_eye: 
- One sumBy and one semicolon linted (incidentally removing your second critique)
- I _refrained_ from touching functionality. This is r3versi's work. I _might_ change the unit overview addition later, or anybody else.

As discussed, please _full_ credit to Federico!

PR #4918, opened 2021-08-20
Modified Armored Unit into 2.5D from BUCKETEER

AdityaMH:


yairm210:
I'm not sure if this is actually better, but sure, why not

AdityaMH:
For me, actually better is just rocket artillery. But sure, why not?

PR #4924, opened 2021-08-20
Build test java11

yairm210:
Expanded metaspace size as per suggestion in https://github.com/gradle/gradle/issues/8354 to solve android:lint "out of metaspace" fail


yairm210:
As a bonus action, I improve build time from 8-9m to ~5-6m
Gradle version is next ;)

SomeTroglodyte:
4G isn't too much for the github actions docker VM's? Wow.

PR #4941, opened 2021-08-21
Fixed interpunction & typos in the tutorial

xlenstra:
Based on a note from a user on discord that a dot was missing in the worker tutorial, I quickly scanned the entire tutorial document for more missing punctuation. Also solved some android studio warnings while I was editing that file


yairm210:
These need to be ctrl-h'D across the entire project to not break existing translations

PR #4951, opened 2021-08-22
StartingLocation-Improvements-be-gone phase 1

SomeTroglodyte:
Goal: No more editor starting locations as non-ruleset """"improvements"""" (what's the inverse of improvement? deprovement? imguessment?)
Phase 2 would be to teach the editor, too, and remove the prefix completely. During that, the limitation to 1 per nation would be removed, too - several would then be alternates and one picked randomly.

**Question: Good idea or not worth the hassle?**

### DIFF
--- a/core/src/com/unciv/logic/MapSaver.kt
+++ b/core/src/com/unciv/logic/MapSaver.kt
@@ -10,20 +10,32 @@ object MapSaver {
     fun json() = GameSaver.json()
 
     private const val mapsFolder = "maps"
+    private const val saveZipped = false
 
     private fun getMap(mapName:String) = Gdx.files.local("$mapsFolder/$mapName")
 
+    fun mapFromSavedString(mapString: String): TileMap {
+        val unzippedJson = try {
+            Gzip.unzip(mapString)
+        } catch (ex: Exception) {
+            mapString
+        }
+        return mapFromJson(unzippedJson)
+    }
+    fun mapToSavedString(tileMap: TileMap): String {
+        val mapJson = json().toJson(tileMap)
+        return if (saveZipped) Gzip.zip(mapJson) else mapJson
+    }
+
     fun saveMap(mapName: String,tileMap: TileMap) {
-        getMap(mapName).writeString(Gzip.zip(json().toJson(tileMap)), false)
+        getMap(mapName).writeString(mapToSavedString(tileMap), false)
     }
 
     fun loadMap(mapFile:FileHandle):TileMap {
-        val gzippedString = mapFile.readString()
-        val unzippedJson = Gzip.unzip(gzippedString)
-        return json().fromJson(TileMap::class.java, unzippedJson)
+        return mapFromSavedString(mapFile.readString())
     }
 
-    fun getMaps() = Gdx.files.local(mapsFolder).list()
+    fun getMaps(): Array<FileHandle> = Gdx.files.local(mapsFolder).list()
 
-    fun mapFromJson(json:String): TileMap = json().fromJson(TileMap::class.java, json)
+    private fun mapFromJson(json:String): TileMap = json().fromJson(TileMap::class.java, json)
 }
--- a/core/src/com/unciv/logic/map/MapParameters.kt
+++ b/core/src/com/unciv/logic/map/MapParameters.kt
@@ -158,5 +158,6 @@ class MapParameters {
     }
 
     // For debugging and MapGenerator console output
-    override fun toString() = "($mapSize ${if (worldWrap)"wrapped " else ""}$shape $type, Seed $seed, $elevationExponent/$temperatureExtremeness/$resourceRichness/$vegetationRichness/$rareFeaturesRichness/$maxCoastExtension/$tilesPerBiomeArea/$waterThreshold)"
+    override fun toString() = if (name.isNotEmpty()) "\"$name\""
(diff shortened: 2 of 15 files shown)


core/src/com/unciv/logic/MapSaver.kt
```diff
@@ -10,20 +10,32 @@ object MapSaver {
...
     private const val mapsFolder = "maps"
+    private const val saveZipped = false
 
     private fun getMap(mapName:String) = Gdx.files.local("$mapsFolder/$mapName")
 
+    fun mapFromSavedString(mapString: String): TileMap {
+        val unzippedJson = try {
+            Gzip.unzip(mapString)
+        } catch (ex: Exception) {
+            mapString
```
yairm210, line 21: U wot m8

core/src/com/unciv/logic/map/MapParameters.kt
```diff
@@ -158,5 +158,6 @@ class MapParameters {
     }
 
     // For debugging and MapGenerator console output
-    override fun toString() = "($mapSize ${if (worldWrap)"wrapped " else ""}$shape $type, Seed $seed, $elevationExponent/$temperatureExtremeness/$resourceRichness/$vegetationRichness/$rareFeaturesRichness/$maxCoastExtension/$tilesPerBiomeArea/$waterThreshold)"
+    override fun toString() = if (name.isNotEmpty()) "\"$name\""
+        else "($mapSize ${if (worldWrap)"wrapped " else ""}$shape $type, Seed $seed, $elevationExponent/$temperatureExtremeness/$resourceRichness/$vegetationRichness/$rareFeaturesRichness/$maxCoastExtension/$tilesPerBiomeArea/$waterThreshold)"
```
yairm210, line 162: Please take the `if` out of here, I can barely read this

yairm210: I know it was there before but still

core/src/com/unciv/models/metadata/GameParameters.kt
```diff
@@ -40,11 +40,32 @@ class GameParameters { // Default values are the default new game
...
+                if (isOnlineMultiplayer) yield("Online Multiplayer")
+                if (noBarbarians) yield("No barbs")
+                if (oneCityChallenge) yield("OCC")
+                if (!nuclearWeaponsEnabled) yield("No nukes")
+                if (religionEnabled) yield("Religion")
+                if (godMode) yield("God mode")
+                if (VictoryType.Cultural !in victoryTypes) yield("No ${VictoryType.Cultural} Victory")
+                if (VictoryType.Diplomatic in victoryTypes) yield("${VictoryType.Diplomatic} Victory")
+                if (VictoryType.Domination !in victoryTypes) yield("No ${VictoryType.Domination} Victory")
+                if (VictoryType.Scientific !in victoryTypes) yield("No ${VictoryType.Scientific} Victory")
```
yairm210, line 67: Iterate over victory types

core/src/com/unciv/models/metadata/GameParameters.kt
```diff
@@ -40,11 +40,32 @@ class GameParameters { // Default values are the default new game
...
+            }.joinToString() +
+            (if (mods.isEmpty()) ", no mods" else mods.joinToString(",", ", mods=(", ")", 6) ) +
```
yairm210, line 69: Why not add this to the sequence? Would be much more readable

core/src/com/unciv/models/metadata/GameParameters.kt
SomeTroglodyte: Debug output only, so to be silent on the defaults I went the quick copy&paste way and inverted the default-off one.... OK.

core/src/com/unciv/logic/map/MapParameters.kt
SomeTroglodyte: :+1: - yea, a debug-only toString() also (unexpectedly) has the right to be readable

core/src/com/unciv/logic/MapSaver.kt
SomeTroglodyte: For tests it is easier to get readable json out of the clipboard save feature, but I left the `private const val saveZipped = false` switch on, not planned. Heck, once the Options screen is properly redesigned we could pack that into options, godmode, viewentiremap, whatnot - maybe a tab that requires a secret handshake.

That is the counterpart - you can feed the loader gzipped json or plain json, it will eat both.

core/src/com/unciv/models/metadata/GameParameters.kt
SomeTroglodyte: Same - quick&dirty debug-only output. But I can reformat - if your first reaction is not "we don't need that at all" then yes, prettify.

SomeTroglodyte:
Grrr. gh always starts reviews if I just comment normally - why would I want to review my own PR? Better answer down here then:

> U wot m8

For tests it is easier to get readable json out of the clipboard save feature, but I left the `private const val saveZipped = false` switch on, not planned. Heck, once the Options screen is properly redesigned we could pack that into options, godmode, viewentiremap, whatnot - maybe a tab that requires a secret handshake.

That is the counterpart - you can feed the loader gzipped json or plain json, it will eat both. Easier than checking first()=='{' and explaining which coincidence happens to make sure the zipped variant never begins with a curly (it _is_ guaranteed). Or at least I thought so at the time.

PR #4954, opened 2021-08-22
Scouts still ignore terrain costs after upgrades

xlenstra:
This PR makes it so that scouts upgraded to archers through ancient ruins still keep their ability to ignore terrain cost


SomeTroglodyte:
:+1: but - promotion would show white circle as icon? You can just copy the scout icon (if its black/white choice is OK), the texture packer will not need extra space for the copy in the atlas png.

xlenstra:
Good point and good solution!

xlenstra:
Though it will cost extra atlas space, as promotions go in a different atlas than units

logicminimal:
i think we should add a unique for promotions that make a promotion exist in a unit only if the unit it will upgrade into automatically has that promotion when freshly built

like: combosite bowmen that will upgrade into chonokus will have logistics promotion when they are chonokus but when they upgrade to something else say gatling gun that doesnt get logistics when freshly built will lose that logistics promotion

SomeTroglodyte:
> different atlas

Dang! But Moving the promotions into the same atlas as the units could help lessen texture swaps, and there's still free room in _that_ puny 1k x 2k texture... (in fact all "separates" could share one)  

_Something_ is already overdue, the main atlas spilling into a second file is a major no-go. Maybe unit test that... Didn't I have a stashed unit test already somewhere flagging outdated atlas? We got icon pushes without atlas updates all the time... Goes for mods too...

SomeTroglodyte:
OK: Deprecate "ImagesToPackSeparately", Do a scanner for "Images.<anything>"... and pack each _with all subfolders_ into <anything>.atlas/png. Document use somewhere. Document you have to delete atlas and run desktop whenever you _remove_ something or update without touching timestamps...

Short term - reduce Stat images to 100x100 and test. Who's got the original sources?

yairm210:
The first thing to be separated should be the *tech icons*. These are rendered ONCE on the world screen, which is the real fps-guzzler. They take up a huge part of the png.

yairm210:
Second are the promotions. They're only displayed in the unitTable, which is both not scaled and appears once. It IS a texture switch per promotion displayed, but if we see it's **really** laggy we can do tricks to make them render together instead of per promotion.

yairm210:
Actually promotions are third, techs are second.
Policies are first. They're not even displayed in worldScreen! Get outta here!

yairm210:
AS FOR THE STAT ICONS
THEY NEED TO BE REPLACED
By someone who knows what they're doing graphics-wise. I made them a bajillion years ago and they have haunted me since with the fact that I can read as much color theory as I want, it's not going to help me do any better,

SomeTroglodyte:
> color theory

Why? I'll go see if they're listed in cretids (typo too pretty to fix), then either fetch sources and gimp, or fetch current and gimp (isolate inner so one can redo the circle which isn't round, nohalo scale, layer merge, done). No colour theory...?

Another thing: This new promotion is a deviation in style, all our others so far have a "nicer" name. For this - is "Wayfinder" (or Ofelaš - remember that film?) taken? And how does Civ5 do it? Doesn't C5 even have temporary effects listed under "promotions"? What's the Scout's effect called? Well, no longer important.

xlenstra:
> > color theory
> 
> Why? I'll go see if they're listed in cretids (typo too pretty to fix), then either fetch sources and gimp, or fetch current and gimp (isolate inner so one can redo the circle which isn't round, nohalo scale, layer merge, done). No colour theory...?
> 
> Another thing: This new promotion is a deviation in style, all our others so far have a "nicer" name. For this - is "Wayfinder" (or Ofelaš - remember that film?) taken? And how does Civ5 do it? Doesn't C5 even have temporary effects listed under "promotions"? What's the Scout's effect called? Well, no longer important.

Scout effect is quite literaly called 'ignores terrain cost', at least according to the wiki, and as the unique was already literally that, it felt okay to call it this. I do agree a name like 'wayfinder' does sound nicer

SomeTroglodyte:
Doing the packer code reorg I mentioned... Grouping will be just a matter of moving and renaming folders.

logicminimal:
> And how does Civ5 do it? Doesn't C5 even have temporary effects listed under "promotions"?

would my proposal for a promotions unique that only makes the unit have x promotion if what it will upgrade inro gets that promotion out of thin air when freshly trained fix this problem?

PR #4959, opened 2021-08-22
Multi-Atlas / Image packer redo

SomeTroglodyte:
Triggered by discussion in #4954

New rules:
- You can have as many Images folders and atlases as needed
- Each Image folder is either named "Images" and maps to game.atlas or named "Images.something" and maps to something.atlas
- Subfolder structure under all such Image folder is identical, for each the output is 1 combined atlas
- Packing and loading is automatic as long as either an atlas is new or some png inside has a newer timestamp
- That means after moving a subfolder you have to delete the source and destination atlas files (only the .atlas) and regenerate
- This also means if a restructure obsoletes something the corresponding atlas/png must be removed manually

This combines the previous BuildingIcons and UnitIcons atlases (atlanti?) into Construction.atlas, the Flags rename is just laziness, and it separates the tech icons as hinted at by @yairm210 into Tech.atlas - just by moving the subfolder. The resulting game.atlas now has plenty room to spare - and as hinted tiral&error to optimize stuff is _easy_ locally. Getting git to recognize a move instead of a delete/add may be tricky, getting github to respect that even tricier...


yairm210:
This is one of those PRs that you'll need to resolve conflicts and then I'll have to try and accept this one before any other image-containing PR, since it's very liable to break.

SomeTroglodyte:
> very liable to break

Oh yes. And the risk of github losing track of moves and renames is also high (which is why there's 2 commits for the resulting atlassii - those in 1 and it confused what is renamed from which completely). I'd be perfectly fine with redoing it from scratch if need be.

But the idea is sound? I'll wait for #4961 anyway I guess...

SomeTroglodyte:
Oh that was quick... readysetgo...

SomeTroglodyte:
Wait, did I make a mistake? Where's that Inquisitor...?

PR #4961, opened 2021-08-23
Stat Icons Redone

SomeTroglodyte:
Here you go. Most match >99%, now with proper attribution, the ones I couldn't find the/an original for are still close. All 100x100 and Gimp source files included in case someone wants to do minor adjustments (in ImagesToNotAddToGame). See also #4951 #4954

Gimp thumbnails - left new right old:
image


yairm210:
Please merge and rebuild to resolve conflicts :)

PR #4966, opened 2021-08-23
Implemented a cap for the production boost of great engineers

### LINKED ISSUE #4823: Great Engineer Production Boost provides too much production
You can use a single Great Engineer to basically build 2 Wonders in the earlier stages of the game. In civ5, Great Engineer Production boost is capped so it provides at most (Remaining building cost - 1) Production. It can be used to hurry non-wonder buildings as well.

xlenstra:
Fixes #4823, in accordance to the current standings of the poll (20 against vs 37 for)

Added uniques:
- "Can speed up construction of a building"

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Units.json
+++ b/android/assets/jsons/Civ V - Vanilla/Units.json
@@ -1494,7 +1494,7 @@
 	{
 		"name": "Great Engineer",
 		"unitType": "Civilian",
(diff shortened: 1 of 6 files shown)


ravignir:
In original game it also displayed the exact amount of production provided - only in bnw tho. I think it might be a good idea to add in parentheses: `Hurry Construction (Number :productionIcon:)`

xlenstra:
As requested:
image

PR #4975, opened 2021-08-23
StartingLocation-Improvements-be-gone phase 2

SomeTroglodyte:
Sequel to #4951 - maybe a bit over the top.

## Features:
- All traces of "StartingLocation " prefix gone except for legacy map loader
- Tile renderer learned to render them. Might benefit from some global "I'm not in the map editor" switch, but I don't think this overhead is that much costlier than that string comparison.
- Editor now places and removes them using the new structure. A starting location and an improvement on the same tile now no problem.
- (The over the top bit) any number of starting locations per nation and any number of starting locations per tile allowed - and they work! The existing placement code with the small patch from phase 1 will both choose a random one from several on offer for a nation, and prevent following up on several nations per tile. The minimum distance between two nations may be ignored, but that was also the case before = responsibility of the map maker.

## Questions:
- Hide the multi-assignment capabilities from the users?
- Is having the nation icons transparent, too, OK? I think it's fitting that they look a little ghostly.
- There's a display limit of 3 icons, but not a placement limit - a "..." or hard limit or suppress feature entirely?
- What to do with the dead code for unit placement?
- Why does TileGroupIcons render improvements with transparency, but not resources?

## Screenshozz
<details><summary>open sesame</summary>

I placed the nation icons now in the center of the tile, shows coexistence start location-improvement, and multiple choice start location:
image

A start from that map - yup, it just chose one randomly:
image

And the mad bit (it works exactly as intended!):
image

</details>

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -21,8 +21,8 @@ import kotlin.math.max
 
 object GameStarter {
     // temporary instrumentation while tuning/debugging
-    private const val consoleOutput = true
-    private const val consoleTimings = true
+    private const val consoleOutput = false
+    private const val consoleTimings = false
 
     fun startNewGame(gameSetupInfo: GameSetupInfo): GameInfo {
         if (consoleOutput || consoleTimings)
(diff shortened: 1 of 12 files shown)


yairm210:
Damn that looks good

PR #4982, opened 2021-08-24
Multiple Start Locations per tile nicer display

SomeTroglodyte:


logicminimal:
can u provide a screenshot pls?

PR #5009, opened 2021-08-27
New Options Screen, tabbed, portrait-friendly

SomeTroglodyte:
### Why
The old one will often take two attempts to change the language since the UncivSliders are in <sup><sub>(Because the thing inexplicably  reacts with a re-layout to the tootip closing. Shouldn't as that actor floats. Anyway, width changes by a few pixel and SelectBox closes).</sub></sup> Bugged me, so I rewrote the thing and added a few bells & whistles. And fixed godMode crashing city screen. And made the mod checker in there more useful.

https://user-images.githubusercontent.com/63000004/131189963-c62f501f-a759-44a3-befe-08862f8215de.mp4
(a bit outdated)

### Decisions to discuss
- ~~Now we have more room for options. New ones to-do? Discord killswitch? Save games as gzip toggle?~~ -> later
- ~~Distribution of options? Move what? Cheat user ID to secret page or to Multiplayer? Translation generator button to secret page?~~ -> later
- Usability in portrait is good. The Android 'force landscape if resolution too low' needs removal.
- ~~Header button captions - shorten?~~ -> done
- ~~Icons - better without?~~ -> No
- ~~Include Unciv icon in atlas? More info for the About page?~~ -> No / later
- ~~General OK would need template additions before merge.~~ -> done
- ~~Multiplayer page untested - tomorrow.~~ -> done

<details><summary>And the password to that secret page is...</summary>

nah, trusted coders only, in PM, on demand
</details>


SomeTroglodyte:
Okay, changed another little bit. Opening it now shows:
<details><summary>(suspicious, en't it?)</summary>

image
</details>

ravignir:
> nah, trusted coders only, in PM
>

As if someone who understands how code works would need to ask 😆

SomeTroglodyte:
P R E C I S E L Y

... Though I trivially avoided actually writing the string into the source. *almost* looked up how to sha512 in kotlin...

logicminimal:
> nah, trusted coders only, in pm

..but why?

SomeTroglodyte:
Almost done.
Testers could if they want [try this as a jar](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Debug%20apk's/Unciv-OptionsScreen.jar) - the to-do is for Android (no need to force-landscape anymore).

AdityaMH:
> Testers could if they want [try this as a jar](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Debug%20apk's/Unciv-OptionsScreen.jar) - the to-do is for Android

Android support jar to run? I guess not.
So this tester for PC only?

core/src/com/unciv/ui/utils/TabbedPager.kt
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    /**
+     * Activate any [secret][addPage] pages by asking for the password.
+     *
+     * If the parent of this Widget is a Popup, then this needs to be called _after_ the parent
+     * is shown to ensure proper popup stacking.
+     */
+    @Suppress("MemberVisibilityCanBePrivate")  // This is part of our API
+    fun askForPassword(secretHashCode: Int = 0) {
```
yairm210, line 286: This doesn't seem to be called ever, I must be missing something?

yairm210:
image

A. Width should be less than 100%, it now looks like it doesn't fit
B. "About" should be the last tab, no one cares usually. Probably the order should be: Display, Gameplay, Language, Sound, Advanced, About
C. 
image
It looks like the dropdowns have a minimum width which is messing this up. Not *your* problem since this probably existed before, but still noticeable, and if we're "portrait-friendly"ing it...
D. The text in the language should be continuous, the split is a bad design choice I made really early in the process and it was a bad one.

SomeTroglodyte:
A - good point. At least a few pixels. Dig out a less than HD AVD...
B - Had wondered about that too :+1:
C - Yup. They had minWidth to align them when two are close; and i reduce that a bit when it gets cramped - but there's a limit to good-will ;)
D - OK, it's untranslated anyway. _should_ we put it into templates or will it start looking weirder having a language mix as language _names_ are untranslated by design? Actually, the flags could be enough to guide a westerner without language skills to change mandarin back to whatever they want... Could look sexy if the labels changed too...

> bad design choice

Nah, the way `Label` misbehaves when `wrap` is on is excuse enough. Didja see the kludge to tame that in there? The first _working_ draft of that had about 20 lines of reflection to manipulate the libgdx-private prefSize field! Only consistently working alternate I could find was to wrap each wrap label in its own Table... The doc of Label even _says_ it stops participating nicely in layout once wrap is on - on purpose!

SomeTroglodyte:
Oh, and the secret page _does_ ask a password after the secret handshake... So the routine _is_ called. It was inspired by the idea that there's stuff in options that a casual player running a release desktop has no use for, but I didn't move them yet - your call. Right now it only has checkboxes for supercharged, view entire map and god mode - so you don't need to set them (and accidentally distribute) in source, or in the debugger. What I didn't do - force the setting of such an option to have an immediate effect. Next turn does it for view map, or force a fresh world screen any other way...

SomeTroglodyte:
Your suggestions are underway... But since I originally designed that Widget to support minimum sizes, not maximum limits, that got to be a bit. Done, but needs new tuning of some previously guessed values. Could revert changes to Popup base class, already tested. Refactored classes a bit, all files involved now have the proper class icon in AS. Translations note is perfectly fine auto-wrapped. Since the Widget now has upped and lower bounds, width is controlled solely by the wordy checkbox I gave the mod checker (and it's still not clear enough that it only affects extension mods, not base ruleset ones), and I'll consider the effort to wrap that too tomorrow.

yairm210:
We could do this piecemeal, add long as it's better than the current situation it doesn't have to be perfect all in one go

SomeTroglodyte:
Tweaking 4 or 5 paddings (and on-demand-wrapping a checkbox) is less work than separating the components. I'm just too tired today. Got a new bike that needed testing. Result - I can't go slower than the limit where the motor stops helping. And those apps aren't LOS compatible.

SomeTroglodyte:
Phew. I think I'm done with this for now. Somewhere in the middle I mis-clicked rebase instead of merge and dealt myself some headaches, but the current diff here looks good.

Recap:
- New widget for tabbed pages
- Helper widget for a Label that doesn't stop participating in layout when wrap is on
- Reworked Options screen on that base
- Some rework in mod checking to allow that page to be more useful to modders
- "Secret page" accesses debug flags - pretty empty, but a chance to expand on, e.g. leave instrumentation in and expose the switch here, or a zipped savegame switch, experimental we don't want people to see yet...
- Language picker inherits "by accident" the nicer translation disclaimer, _tested_. Still not replaced it entirely.

Detail from your suggestions I did not adopt verbatim: About is still on the left, but the first page that opens is gameplay. Felt best.
In extreme resolution/aspect combos this still isn't perfect, as pages are built sequentially and no row can rely on knowing its final width, and there's no re-layout. But I think usability is still better.

yairm210:
I personally think that the tab class overexplains its parameters. I know what minWidth and maxWidth mean, if you called the parameter headerFontColor and headerFontSize I wouldn't need text there either...

yairm210:
And do you really need all those parameters? Did anyone use them with different values? I bet some of them yes, but I feel that some of them no, and if so, remove unnecessary parameters.

SomeTroglodyte:
No, I'm not fixing countless lines ending in 8 or more blanks _again_ to merge master into this.

PR #5010, opened 2021-08-28
Harden automateImprovementPlacer and detect Mod problems with these

### LINKED ISSUE #5007: null exception in improvement placement automation
**Platform**
Android, from GPC

**Version**
3.16.8

**Describe the bug**
```
java.lang.NullPointerException: 
  at com.unciv.logic.automation.SpecificUnitAutomation.automateImprovementPlacer (SpecificUnitAutomation.kt:442)
  at com.unciv.logic.automation.UnitAutomation.automateUnitMoves (UnitAutomation.kt:103)
  at com.unciv.logic.automation.NextTurnAutomation.automateUnits (NextTurnAutomation.kt:524)
  at com.unciv.logic.automation.NextTurnAutomation.automateCivMoves (NextTurnAutomation.kt:53)
  at com.unciv.logic.GameInfo.nextTurn (GameInfo.kt:139)
  at com.unciv.ui.worldscreen.WorldScreen$nextTurn$1.invoke (WorldScreen.kt:607)
  at com.unciv.ui.worldscreen.WorldScreen$nextTurn$1.invoke (WorldScreen.kt:57)
```

SomeTroglodyte:
Should resolve #5007 - but I haven't specifically tested with Civ6mod.

### DIFF
--- a/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
@@ -9,6 +9,7 @@ import com.unciv.logic.map.MapUnit
 import com.unciv.logic.map.TileInfo
 import com.unciv.models.ruleset.tile.ResourceType
 import com.unciv.models.ruleset.tile.TileResource
+import com.unciv.models.stats.Stat
 import com.unciv.models.stats.Stats
 import com.unciv.ui.worldscreen.unit.UnitActions
 
@@ -217,8 +218,9 @@ object SpecificUnitAutomation {
 
     fun automateImprovementPlacer(unit: MapUnit) {
         val improvementName = unit.getMatchingUniques("Can construct []").first().params[0]
-        val improvement = unit.civInfo.gameInfo.ruleSet.tileImprovements[improvementName]!!
-        val relatedStat = improvement.maxByOrNull { it.value }!!.key
+        val improvement = unit.civInfo.gameInfo.ruleSet.tileImprovements[improvementName]
+            ?: return
+        val relatedStat = improvement.maxByOrNull { it.value }?.key ?: Stat.Culture
 
         val citiesByStatBoost = unit.civInfo.cities.sortedByDescending {
             val stats = Stats()
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -299,7 +299,7 @@ class Ruleset {
         val lines = ArrayList<String>()
         var warningCount = 0
 
-        // Checks for all mods
+        // Checks for all mods - only those that can succeed without loading a base ruleset
         for (unit in units.values) {
             if (unit.upgradesTo == unit.name)
                 lines += "${unit.name} upgrades to itself!"
@@ -327,9 +327,10 @@ class Ruleset {
             }
         }
 
+        // Quit here when no base ruleset is loaded - references cannot be checked
         if (!modOptions.isBaseRuleset) return CheckModLinksResult(warningCount, lines)
 
-        val baseRuleset = RulesetCache.getBaseRuleset()
+        val baseRuleset = RulesetCache.getBaseRuleset()  // for UnitTypes fallback
 
         for (unit in units.values) {
             if (unit.requiredTech != null && !technologies.containsKey(unit.requiredTech!!))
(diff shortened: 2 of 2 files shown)


ravignir:
image
I will leave that here (RekMOD). Now tell me how do I fix it?

SomeTroglodyte:
To cater to what the AI _can_ do at the moment:
- If it's a Civilian, and has the 15% unique of the Great General, it will go into that routine, where the Citadel AI is - without that, no Citadel AI. Conversely, a unit with the 15% bonus unique plus a Can create [Academy] - the AI will not be able to use the latter as far as I can tell.
- If it's a Civilian, and has the Can create unique but not the 15% unique of the Great General, it will go into the improvement placing AI routine, where it will want to see a "primary" stat of the improvement that's placed, and it cannot do so from uniques alone, only the normal properties -> But with this commit, it will default to Culture (where it previously defaulted to whatever Stats.toHashSet would enumerate first). Then it seems to look for a city most in need of that Stat, and go there...

So a Citadel placer that's not a Great General too will sit idle. Changing that needs coding.
For the Archaeologist, isn't a config possible where the improvement it places _has_ a stat? Limit to tech another way? (didn't look at rekmod, judging by the definition of that other Archaeologist) Otherwise -> Coding more AI.

If it's not a civilian -> a) I _hope_ I limited these checks to Civilians only, already forgot. b) Haven't inspected the code, so I don't know whether the AI can deal with a Tank that can also sit down to make a Citadel, but I suspect it will not be able to consider the option, treating it as a normal Tank instead. That's probably what will happen with the AI and the Toa of the Civ6 mod.

Put another way - even non-kotlin-speakers should see that this code is either-or, pick just one AI routine:
```kotlin
        if (unit.isCivilian()) {
            if (unit.hasUnique(Constants.settlerUnique))
                return SpecificUnitAutomation.automateSettlerActions(unit)

            if (unit.hasUniqueToBuildImprovements)
                return WorkerAutomation.automateWorkerAction(unit)

            if (unit.hasUnique(Constants.workBoatsUnique))
                return SpecificUnitAutomation.automateWorkBoats(unit)

            if (unit.hasUnique("Bonus for units in 2 tile radius 15%"))
                return SpecificUnitAutomation.automateGreatGeneral(unit)

            if (unit.hasUnique("Can construct []"))
                return SpecificUnitAutomation.automateImprovementPlacer(unit) // includes great people plus moddable units

            return // The AI doesn't know how to handle unknown civilian units
        }
```

TBH, when I saw, I considered making that a bit fuzzier, allowing each specific to return 'sorry I couldn't decide' and give the next routine a chance - but that seemed like too much for my taste, just because _testing properly_ would be daunting.

SomeTroglodyte:
... but maybe we should put the distinction between test result "unplayable" and "warnings" a little clearer into that toast ... The "preventing unit automation" one is warning only, not preventing starting a game (I fervently hope not having bungled that aspect).

SomeTroglodyte:
@ravignir  - won't PR this right now, collides with option screen, but - what do you think of a nicer title like this?
```
The mod you selected is incorrectly defined! = 
The mod you selected has problems. = 
You can play it, but don't expect everything to work! = 
```
image

ravignir:
I honestly don't know. I don't pike either,because seeing such message is discouraging, especially when i can't easily fix it by just manipulating jsons :D 
However I can see how these issues can be fixed in the code, at least in case of RekMOD.

Background info:
Hetairoi is a Greek Great General replacement, that acts as a combat unit and upgrades to more advanced units, retaining all Great General abilities but ability to construct a Citadel. 
- So the possible fix here is to look if the unit has Great General combat unique hidden as a promotion.

Archaeologists are supposed to act like Workers that can build a single improvement and only on a specific resource >Antiquity Sites/Hidden Antiquity Sites<. 
- So the game needs an unique that allows that. Archaeologists exist in Civ5 BNW, so I guess it has to be implemented anyway.

ravignir:
Hmm, I think there might be a way to fix these without editing code.

SomeTroglodyte:
> Hetairoi is a Greek Great General replacement, that acts as a combat unit

So it is military, and I _did_ forget to not apply that check to them. Will fix I guess - the nagging. Your Hetairoi will be used by AI as combat only AFAIK - haven't inspected that code branch yet.

> Archaeologists are supposed to act like Workers

So it spends turns building? Different unique -> different AI. Would likely fail anyway as that AI too only goes by primary stats, not on-top-by-uniques. Is the unique condition negatable? Instead of +0, +5 once Radio is discovered -> +5, -5 until Radio is discovered? Probably not. Then -> code.

Honestly, debugging AI automation of exceptional modded units is hard, in the sense to _see_ what's going on. Maybe we could start leaving instrumentation in, capture output, and make it available...

ravignir:
At least I managed to get these errors to not display xD
- In case of Hetairoi I just added the +15% nearby units combat strength unique, as it would not stack anyway.
- I spend some time with Archaeologists, as I initially wanted them to build these improvements. But it was not really possible without enabling workers to build them as well. So I just made Archaeological Digs provide +1 Culture and granted Palace an unique that negates that tiny bonus.

PR #5012, opened 2021-08-28
units get promotions and xp bonuses from CS buildings

SimonCeder:
As described in #4602.
Newly gifted units from militaristic city states start with bonus XP and promotions as if they were built by the CS.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
+++ b/core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
@@ -334,16 +334,22 @@ class BaseUnit : INamed, INonPerpetualConstruction, ICivilopediaText {
 
         if (this.isCivilian()) return true // tiny optimization makes save files a few bytes smaller
 
+        addConstructionBonuses(unit, cityConstructions)
+
+        return true
+    }
+
+    fun addConstructionBonuses(unit: MapUnit, cityConstructions: CityConstructions) {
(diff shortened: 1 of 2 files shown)


logicminimal:
@ravignir if a city state has two cities then would only the cspitsl count?

SimonCeder:
Yes, this uses the buildings from the city state's capital

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -334,47 +334,51 @@ class BaseUnit : INamed, INonPerpetualConstruction, ICivilopediaText {
...
+                // Deprecated since 3.15.9
                 cityConstructions.cityInfo.getMatchingUniques("New [] units start with [] Experience") +
                 cityConstructions.cityInfo.getLocalMatchingUniques("New [] units start with [] Experience in this city")
-            //
+        //
         ) {
             if (unit.matchesFilter(unique.params[0]))
                 XP += unique.params[1].toInt()
         }
         unit.promotions.XP = XP
```
SomeTroglodyte, line 357: Sounds like an earlier `Military Units gifted from City-States start with [] XP` gets overwritten here

SimonCeder: It absolutely does

Issue #4993, opened 2021-08-26
Broken Transparency Button 


benyaminl:
**Platform**
Windows 10 20H2 19042.1164
```
openjdk version "11.0.12" 2021-07-20
IBM Semeru Runtime Open Edition 11.0.12.0 (build 11.0.12+7)
Eclipse OpenJ9 VM 11.0.12.0 (build openj9-0.27.0, JRE 11 Windows 10 amd64-64-Bit Compressed References 20210730_175 (JIT enabled, AOT enabled)
OpenJ9   - 1851b0074
OMR      - 9db1c870d
JCL      - 21849e2ca0 based on jdk-11.0.12+7)

javac 11.0.12
```

**Version**
This Github, version 3.16.7-uploadApk

**Describe the bug**
all button that has transparency are blank see the screenshoot

**To Reproduce**
Run Anything button is blank

**Expected behavior**
Normal button view

**Screenshots**
![image](https://user-images.githubusercontent.com/2817332/130939999-53b57db7-92ca-4cc8-90b0-c5187443cf1f.png)
![image](https://user-images.githubusercontent.com/2817332/130940354-8402c4a1-e917-424d-88ed-e3d513557513.png)
![image](https://user-images.githubusercontent.com/2817332/130940386-10db25d0-7362-48ae-9565-1e829daba711.png)
![image](https://user-images.githubusercontent.com/2817332/130940468-aec424bd-aa33-42d5-9d83-9def63b7afea.png)

PR #5014, opened 2021-08-28
Atlas reorg - packer to loader via json

SomeTroglodyte:
Follow-up to #4993 - instead of hardcoding or trying to enumerate files, the packer now leaves the list in a file, which is included in the jar, and ImageGetter uses it to find all additional atlases. This time, the resulting jar is tested.

Also refactored so the Skin atlas no longer needs special treatment, neither in ImagePacker nor ImageGetter. The "special" is done by allowing any Images folder to override default TexturePacker settings through a json. Side effect: Mods could change the settings too, for their images only, if they wanted. Which they have no reason to. That file can move with the Images subfolders if need be, and experimenting with the settings could also be made easier.

Disadvantage: More ballast that's packaged and never used from the actual distribution (the texturepacker settings file), but that pales in view of the ballast already in there - netty, ktor, thatched villagers are all AFAIK unnecessary and dead in there.


yairm210:
Ok fair enough

yairm210:
Rc

SomeTroglodyte:
"Rc" = "Ruddy Conflictz"? No, these were expected and easy enough.

PR #5016, opened 2021-08-28
Persistent new game setup

SomeTroglodyte:
- GameParameters and MapParameters are now saved and reused for a fresh New Game from main menu
- Save happens only when the start was successful, right before `GameStarter.startNewGame` is done
- Reuse happens from main menu New Game or Quickstart, _not_ map editor new map from current or worldscreen popup - new game from running one (or victory screen) -> **debatable**
- Moved `GameSetupInfo` to metadata - it was _very_ out of place in the ui packages
- Made cloning more consistent - defensive choice
- Removed `GameSetupInfo.gameId` : never written, only cloned. True gameId lives in GameInfo
- Tried hard not to touch MapParameters.reseed behaviour though I would liked to have made it `@ Transient`. Currently, only `Simulation` gets a verbatim clone without reseeding, and I kept it that way.

### DIFF
--- a/core/src/com/unciv/MainMenuScreen.kt
+++ b/core/src/com/unciv/MainMenuScreen.kt
@@ -16,7 +16,7 @@ import com.unciv.logic.map.mapgenerator.MapGenerator
 import com.unciv.models.ruleset.RulesetCache
 import com.unciv.ui.MultiplayerScreen
 import com.unciv.ui.mapeditor.*
-import com.unciv.ui.newgamescreen.GameSetupInfo
+import com.unciv.models.metadata.GameSetupInfo
 import com.unciv.ui.newgamescreen.NewGameScreen
 import com.unciv.ui.pickerscreens.ModManagementScreen
 import com.unciv.ui.saves.LoadGameScreen
@@ -236,7 +236,7 @@ class MainMenuScreen: CameraStageBaseScreen() {
     }
 
     private fun quickstartNewGame() {
-        val newGame = GameStarter.startNewGame(GameSetupInfo().apply { gameParameters.difficulty = "Chieftain" })
+        val newGame = GameStarter.startNewGame(GameSetupInfo.fromSettings("Chieftain"))
         game.loadGame(newGame)
     }
 
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -2,6 +2,7 @@ package com.unciv.logic
 
 import com.badlogic.gdx.math.Vector2
 import com.unciv.Constants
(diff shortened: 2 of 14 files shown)


ravignir:
I don't like that (i guess it will be now much quicker to just restart the game than to manually reset the settings).  I would rather have an option in the menu to disable/enable it.

SomeTroglodyte:
What do you mean? You prefer Unciv's arbitrary defaults to your own settings? We could do a reset to defaults button.
Or do you want to be able to "freeze" your own set of defaults? Could do that too.
Anyway, new options would better fit... on #5009 which is still quite WIP.

ravignir:
It will be a problem when somone plays a lot of multiplayer games, which often vary a lot in terms of game settings. Reset to defaults button would suffice i think.

By the way: what about saving multiplayer players id's when starting the game from the quick menu?

SomeTroglodyte:
> multiplayer players id's when starting the game from the quick menu

Not sure I understand either. Quickstart is single-player, no? Saving ID's from a Quickstart where there are no ID's???

But - that's a (:+1:) point, the saved setup can have multiplayer on, which I didn't consider. Quickstart should _stay_ single-player only. So, either save SP and MP setups separately and independently, or just don't persist MP, or make Quickstart revert to default if saved state is MP. Maybe even patch hotseat out of a Quickstart. When saving MP, maybe do so without the ID's... ?

ravignir:
> > multiplayer players id's when starting the game from the quick menu
> 
> Not sure I understand either. Quickstart is single-player, no? Saving ID's from a Quickstart where there are no ID's???
> 
> But - that's a (👍) point, the saved setup can have multiplayer on, which I didn't consider. Quickstart should _stay_ single-player only. So, either save SP and MP setups separately and independently, or just don't persist MP, or make Quickstart revert to default if saved state is MP. Maybe even patch hotseat out of a Quickstart. When saving MP, maybe do so without the ID's... ?

I meant from the menu opened during an ongoing game which is the current way to start a new game with the same settings (except map sed changes and i really wish there was an option to save it after starting a game).
if you want to quickly restart a multiplayer game that way, players id from ongoing game should be used .

image
image

SomeTroglodyte:
Well, if the behaviour going into new-game from a running one changed, then that's unintentional. At that point I thought I had expressly left the passing of the GameParameters alone. Will look into it later.

> map se(e)d changes and i really wish there was an option to save it after starting a game

It _is_ saved, in the save game at least, just not reused. If in dire need, open in text editor, go to the end of the file, copy and paste...

ravignir:
I really wish in the game set up screen there were two buttons: "save game settings" and "reset to defaults".

PR #5018, opened 2021-08-29
Anti-Armor, negative tile yield, LoadScreen

### LINKED ISSUE #5002: Anti-Armor icon missing
`UnitPromotionIcons\Anti-Armor.png` missing, introduced #4717

**To Reproduce**
Load any game, F1, Promotions, Anti-Armor I, Check Icon

Lazy fix option: Make it a copy of Ambush?

SomeTroglodyte:
Resolves #5002 

Sorry, this is three completely unrelated things together. Took me a while to figure out was was wrong with those promotions, and while testing I noticed stuff...
- A few tiny oversights with those Helicopter promotions. That extra colon caused them to have no unitType.
- Since the fandom wiki says 25% I changed Ambush to 25% - searched the issues and PR's here for the keyword, no relevant hits. After all, in Civ5 Ambush and Anti-Armor are not really distinct.
- My Stats icons repaint affected the LoadScreen. While at it, I made it show up earlier - by bypassing the atlas packaging. Cost: another loose png in ExtraImages. And on the fly that screenshot in the one Tutorial now scales nicer.
- Plus defense in depth against negative yields, for mods. (um, the Aliens mod gives bonuses for cold tiles and maluses everywhere else, which could end up with a -1 food on desert)

Atlas update not yet included - will conflict with #5014 so - delayed for next master-merge.


xlenstra:
On the +33 or +25 for ambush, we've had that discussion before: https://github.com/yairm210/Unciv/pull/4331 :)

I imagine the might have changed in BNW to make it less OP, but since we're aiming for G&K this should be the correct value

SomeTroglodyte:
> we've had that discussion before

And I had a vague recollection, searched for it, and couldn't find it. I even remember the search for "Ambush" without `is:issue is:open` gave only three hits... :blind:
Thanks for fix-PR-ing. So the wiki is wrong on them being actually the same promotion only different prerequisites, too.

xlenstra:
> > we've had that discussion before
> 
> And I had a vague recollection, searched for it, and couldn't find it. I even remember the search for "Ambush" without `is:issue is:open` gave only three hits... :blind:
> Thanks for fix-PR-ing. So the wiki is wrong on them being actually the same promotion only different prerequisites, too.

No problem, just doing my part in the team 🙃

PR #5026, opened 2021-08-29
Add Austrian civ

SimonCeder:
This PR adds the Austria civ, as seen in G&K.
Most of the code relates to Diplomatic Marriage mechanic, which is working well. The AI is able to use this mechanic as well.
Austria gets an extra button in the diplomacy view to marry city states. Clicking it gives you all cities and units from the city state and opens a popup asking you whether to annex or puppet the cities.

Added uniques:
- "Can spend Gold to annex or puppet a City-State that has been your ally for [] turns."
- "[]% to Flank Attack bonuses"

austria

On a coding note, I'm setting `foundingCiv = ""` for cities that are acquired by marriage, so the City-State can never be liberated (civ v behavior according to wiki). This means that I had to remove the `foundingCiv == ""` clauses for determining if a city is annexed from Building.kt and CityStats.kt. As far as I can tell, there was not an obvious reason for those clauses to be there in the first place, but something to bear in mind.


android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -743,6 +743,34 @@
 		"Södertälje","Mariestad","Östersund","Borås","Sundsvall","Vimmerby","Köping","Mora","Arboga","Växjö","Gränna",
 		"Kiruna","Borgholm","Strängnäs","Sveg"]
 	},
+	{
+		"name": "Austria",
+		"leaderName": "Maria Theresa",
+		"adjective": ["Austrian"],
+		"startBias": ["Hill"],
+		//"preferredVictoryType": "Diplomatic",
```
SomeTroglodyte, line 751: Shouldn't that be possible by now?

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -114,6 +114,15 @@ object NextTurnAutomation {
 
     /** allow AI to spend money to purchase city-state friendship, buildings & unit */
     private fun useGold(civInfo: CivilizationInfo) {
+        if (civInfo.getHappiness() > 0) {
+            // Marry all possible city-states
+            for (unique in civInfo.getMatchingUniques("Can spend Gold to annex or puppet a City-State that has been your ally for [] turns.")) {
```
SomeTroglodyte, line 119: No parameter check?

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -114,6 +114,15 @@ object NextTurnAutomation {
 
     /** allow AI to spend money to purchase city-state friendship, buildings & unit */
     private fun useGold(civInfo: CivilizationInfo) {
+        if (civInfo.getHappiness() > 0) {
```
SomeTroglodyte, line 117: Shouldn't a check per CS be better? AI knowns 12 CS and has tons of gold -> marries all 12 in 1 turn -> -36 Happy or whatever... Or limit to 1 per turn for the AI

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
SimonCeder: The parameter is read when first allying the city-state, and the check is baked into `canBeMarriedBy`

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
SimonCeder: 1 per turn max sounds good.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -977,6 +981,54 @@ class CivilizationInfo {
...
+            GameSpeed.Quick -> 335
+            GameSpeed.Standard -> 500
+            GameSpeed.Epic -> 750
+            GameSpeed.Marathon -> 1500
+        }
+        // Plus disband value of all units
+        for (unit in units) {
+            cost += unit.baseUnit.getDisbandGold(this)
+        }
+        // Round to nearest 5
```
SomeTroglodyte, line 996: This is not a nearest - it's a down to zero.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -977,6 +981,54 @@ class CivilizationInfo {
...
+        cost *= 5
+
+        return cost
+    }
+
+    fun canBeMarriedBy(otherCiv: CivilizationInfo): Boolean {
+        return (getDiplomacyManager(otherCiv).relationshipLevel() == RelationshipLevel.Ally
+                && !otherCiv.getDiplomacyManager(this).hasFlag(DiplomacyFlags.MarriageCooldown)
+                && otherCiv.gold >= getDiplomaticMarriageCost()
+                && !otherCiv.getMatchingUniques("Can spend Gold to annex or puppet a City-State that has been your ally for [] turns.").none()
```
SomeTroglodyte, line 1007: !none == any

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -977,6 +981,54 @@ class CivilizationInfo {
...
+        cost /= 5
+        cost *= 5
+
+        return cost
+    }
+
+    fun canBeMarriedBy(otherCiv: CivilizationInfo): Boolean {
+        return (getDiplomacyManager(otherCiv).relationshipLevel() == RelationshipLevel.Ally
+                && !otherCiv.getDiplomacyManager(this).hasFlag(DiplomacyFlags.MarriageCooldown)
+                && otherCiv.gold >= getDiplomaticMarriageCost()
```
SomeTroglodyte, line 1006: most expensive test last in a && chain - at least isCS is far cheaper

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -275,6 +275,20 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
             }
         }
 
+        for (unique in viewingCiv.getMatchingUniques("Can spend Gold to annex or puppet a City-State that has been your ally for [] turns.")) {
+            val diplomaticMarriageButton =
+                "Diplomatic Marriage ([${otherCiv.getDiplomaticMarriageCost()}] Gold)".toTextButton()
+            diplomaticMarriageButton.onClick {
+                val newCities = otherCiv.cities
+                otherCiv.diplomaticMarriage(viewingCiv)
+                UncivGame.Current.setWorldScreen() // The other civ will no longer exist
```
SomeTroglodyte, line 284: This or re-enter DiplomacyScreen - - - or rightSideTable.clear(); updateLeftSideTable() ?

core/src/com/unciv/ui/worldscreen/AlertPopup.kt
```diff
@@ -244,6 +245,46 @@ class AlertPopup(val worldScreen: WorldScreen, val popupAlert: PopupAlert): Popu
                 addGoodSizedLabel(civInfo.nation.startIntroPart2).row()
                 add(getCloseButton("Let's begin!"))
             }
+            AlertType.DiplomaticMarriage -> {
+                val city = worldScreen.gameInfo.getCities().first { it.id == popupAlert.value }
+                addGoodSizedLabel(city.name.tr() + ": " + "What would you like to do with the city?".tr(), 24) // Add name because there might be several cities
```
SomeTroglodyte, line 250: Doesn't addGoodSizedLabel translate, so you could do with "[${city.name}]: [What]"? Or the {} way.

core/src/com/unciv/ui/worldscreen/AlertPopup.kt
```diff
@@ -244,6 +245,46 @@ class AlertPopup(val worldScreen: WorldScreen, val popupAlert: PopupAlert): Popu
...
+                        worldScreen.shouldUpdate = true
+                        close()
+                    }
+                    add("Puppet".toTextButton().onClick(function = puppetAction) ).row()
+                    keyPressDispatcher['p'] = puppetAction
+                    addGoodSizedLabel("Puppeted cities do not increase your tech or policy cost, but their citizens generate 1.5x the regular unhappiness.").row()
+                    addGoodSizedLabel("You have no control over the the production of puppeted cities.").row()
+                    addGoodSizedLabel("Puppeted cities also generate 25% less Gold and Science.").row()
+                    addGoodSizedLabel("A puppeted city can be annexed at any time.").row()
+                }
```
SomeTroglodyte, line 286: Big coppypassta couldn't be refactored to a common usable helper?

core/src/com/unciv/ui/worldscreen/AlertPopup.kt
SimonCeder: It does, I just wanted to save on a string. What's the {} way?

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
SimonCeder: matter of taste imo, I like going back to the worldscreen and getting the popup at once.

core/src/com/unciv/ui/worldscreen/AlertPopup.kt
SomeTroglodyte: Read Translations  `fun String.tr()` line 263 `if (contains('{'))` or [wiki](https://github.com/yairm210/Unciv/wiki/Translating#rules-for-templates-added-manually) - and improve that if you can :wink:

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
SomeTroglodyte: OK yes that's a very good reason...

android/assets/jsons/Civ V - Vanilla/Units.json
```diff
@@ -995,6 +995,22 @@
...
+		"uniqueTo": "Austria",
+		"movement": 5,
+		"strength": 34,
+		"cost": 225,
+		"requiredTech": "Military Science",
+		"obsoleteTech": "Combustion",
+		"requiredResource": "Horses",
+		"upgradesTo": "Landship",
+		"uniques": ["Can move after attacking","No defensive terrain bonus","-[33]% Strength vs [City]" ,
+		"[+1] Visibility Range", "[+50]% Flank Attack modifier" ],
```
xlenstra, line 1011: I personally like the wording "[+50]% Strength when flanking and attacking" more, as it more closely follows the "[amount]% Strength <conditions>" template most uniques use

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -977,6 +981,55 @@ class CivilizationInfo {
         }
     }
 
+    fun getDiplomaticMarriageCost(): Int {
+        // https://github.com/Gedemon/Civ5-DLL/blob/master/CvGameCoreDLL_Expansion1/CvMinorCivAI.cpp, line 7812
+        var cost = when (gameInfo.gameParameters.gameSpeed) {
+            GameSpeed.Quick -> 335
+            GameSpeed.Standard -> 500
+            GameSpeed.Epic -> 750
+            GameSpeed.Marathon -> 1500
```
xlenstra, line 990: Isn't this just `500 * gameInfo.gameParameters.gameSpeed.modifier`?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -977,6 +981,55 @@ class CivilizationInfo {
...
+            getCapital().location, civName, NotificationIcon.Diplomacy, otherCiv.civName)
+        for (civ in gameInfo.civilizations.filter { it != otherCiv })
+            civ.addNotification("[${otherCiv.civName}] has married into the ruling family of [${civName}], bringing them under their control.",
+                getCapital().location, civName, NotificationIcon.Diplomacy, otherCiv.civName)
+        for (unit in units)
+            unit.gift(otherCiv)
+        for (city in cities) {
+            city.moveToCiv(otherCiv)
+            city.isPuppet = true // Human players get a popup that allows them to annex instead
+            city.foundingCiv = "" // This is no longer a city-state
```
xlenstra, line 1028: Wouldn't we then want to set it to the nation that it has been married into?

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -275,6 +275,20 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
             }
         }
 
+        for (unique in viewingCiv.getMatchingUniques("Can spend Gold to annex or puppet a City-State that has been your ally for [] turns.")) {
```
xlenstra, line 278: So if we have multiple uniques, multiple of the same buttons would appear? Should this just be an `if (viewingCiv.hasUnique("bla")) {`?

Especially since we don't seem to differentiate between the different uniques when setting the flags, so the multiple buttons would all turn enabled at exactly the same time, removing any possible difference between them

android/assets/jsons/Civ V - Vanilla/Units.json
SimonCeder: That's not what the bonus is. It's a bonus to the flanking modifier itself, ie instead of +10% per adjacent friendly unit you get +15%. Perhaps there is a clearer way to express that?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -977,6 +981,55 @@ class CivilizationInfo {
...
+        }
+        destroy()
+    }
+
```
yairm210, line 1032: More code in CivInfo that should probably be in its own class, or added to one of the existing side-classes

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -114,6 +114,16 @@ object NextTurnAutomation {
...
+            // Marry all possible city-states
+            for (unique in civInfo.getMatchingUniques("Can spend Gold to annex or puppet a City-State that has been your ally for [] turns.")) {
```
yairm210, line 119: This is a "for unique" check that should be replaced by a "if hasUnique" check (added to the civInfo.getHappiness if, above) - if indeed there's a max on 1 marriage per turn

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SimonCeder: No, Austria should need to build courthouses.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SimonCeder: I'll do it all at once in a separate PR if it's all the same

android/assets/jsons/Civ V - Vanilla/Units.json
xlenstra: Knowing what it is it is quite a clear way to describe it. Only things I could come up with was "Bonus from flanking [+50]%", or maybe "Strength bonus from flanking [+50]%", but if you don't like those and don't have a better idea feel free to leave it as is

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
xlenstra: On a related note, this function is getting quite large. Maybe it's time to split it up? Extracting this part would allow you to simply `if (!viewingCiv.hasUnique("bla")) return`, also reducing unnecessary indentation

yairm210:
K
Conflict resolution time :)

SimonCeder:
all set!

PR #5030, opened 2021-08-30
Update Korean.properties

cjddig:


android/assets/jsons/translations/Korean.properties
```diff
@@ -1239,18 +1222,15 @@ Sacred Waters = 신성한 물
...
 Follower = 신자
- # Requires translation!
-[stats] from every [building] in cities where this religion has at least [amount] followers = 
+[stats] from every [building] in cities where this religion has at least [amount] followers = 신자가 최소 [amount]명 있을 때 도시의 모든 [building]으로부터 [stats]
 
 Cathedrals = 대성당
- # Requires translation!
-May buy [building] buildings for [amount] [stat] [cityFilter] = 
+ # ↓ 검토해주세요/examine please
+May buy [building] buildings for [amount] [stat] [cityFilter] = [amount] [stat] [cityFilter]
```
xlenstra, line 1231: Comments like these will automatically be removed when the next version is released, so if you're not sure about a translation, it is better to leave it out entirely

xlenstra: Especially as each part between brackets is required in the translation, so all your other translations can't be merged because the `[building]` tag is missing.

xlenstra:
Are you still planning on fixing these problems?

xlenstra:
Closing as this is outdated and doesn't seem to be updated any more. If you disagree, comment, and I'll reopen.

PR #5031, opened 2021-08-30
AI will now found & enhance religions -- minor improvement to civilian AI

xlenstra:
This PR adds simple functionality for AI players to found and enhance religions.
It also gives an AI for prophets to use their 'found religion' and 'enhance religion' actions, though no other actions are supported as of yet, most notably 'spread religion'.
The beliefs for these are chosen randomly, I'll port the AI from the original in another PR.

I've also made some changes to automated civilian movements, mostly that if there is danger nearby they will always try to flee to either a city or a military unit, hopefully leading to less of them being destroyed and/or captured.

### DIFF
--- a/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
@@ -369,4 +369,27 @@ object SpecificUnitAutomation {
         }
         return false
     }
+
+    fun foundReligion(unit: MapUnit) {
+        val cityToFoundReligionAt = unit.civInfo.cities.first { !it.isHolyCity() }
+        if (unit.getTile() != cityToFoundReligionAt.getCenterTile()) {
+            unit.movement.headTowards(cityToFoundReligionAt.getCenterTile())
+            return   
+        }
+        unit.civInfo.religionManager.useGreatProphet(unit)
+        UnitActions.addStatsPerGreatPersonUsage(unit)
+    }
+    
+    fun enhanceReligion(unit: MapUnit) {
+        // Try go to a nearby city
+        if (!unit.getTile().isCityCenter())
+            UnitAutomation.tryEnterOwnClosestCity(unit)
+        
+        // If we were unable to go there this turn, unable to do anything else
+        if (!unit.getTile().isCityCenter())
+            return
+        
+        unit.civInfo.religionManager.useGreatProphet(unit)
+        UnitActions.addStatsPerGreatPersonUsage(unit)
+    }
 }
--- a/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
@@ -376,8 +376,9 @@ object SpecificUnitAutomation {
             unit.movement.headTowards(cityToFoundReligionAt.getCenterTile())
             return   
         }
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
```diff
@@ -369,4 +369,27 @@ object SpecificUnitAutomation {
...
         return false
     }
+
+    fun foundReligion(unit: MapUnit) {
+        val cityToFoundReligionAt = unit.civInfo.cities.first { !it.isHolyCity() }
+        if (unit.getTile() != cityToFoundReligionAt.getCenterTile()) {
+            unit.movement.headTowards(cityToFoundReligionAt.getCenterTile())
+            return   
+        }
+        unit.civInfo.religionManager.useGreatProphet(unit)
```
yairm210, line 379: This should be "use the unit action" instead, so the logic is only in one place

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -469,16 +497,43 @@ object UnitAutomation {
         unit.civInfo.addNotification("[${unit.displayName()}] finished exploring.", unit.currentTile.position, unit.name, "OtherIcons/Sleep")
         unit.action = null
     }
-
+    
+    /** Returns whether the civilian spends its turn hiding and not moving */
+    fun tryRunAwayIfNeccessary(unit: MapUnit): Boolean {        
+        // This is a little 'Bugblatter Beast of Traal': Run if we can attack an enemy
```
yairm210, line 503: "... A creature so mondbogglingly stupid..."

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -137,9 +137,11 @@ class ReligionManager {
             if (!mayFoundReligionNow(prophet)) return // How did you do this?
             religionState = ReligionState.FoundingReligion
```
yairm210, line 138: I think this whole function should be put into the unit actions class

core/src/com/unciv/logic/automation/UnitAutomation.kt
xlenstra: Not my comment - just copied and pasted it from somewhere :)

SomeTroglodyte: Finally someone noticed. Had to add those to the dictionary, too, so Marvin Studio stopped squiggying them!

SomeTroglodyte: But honestly, that should have been a to-do: The check _is_ the wrong way round, at least where I put the comment.

SomeTroglodyte:
Can't really check this in depth now, looks superficially like done with a concentrated mind. Only thing - I see a lot of code very specific to the Prophet _not_ be in SpecificUnitAutomation. But since you moved the 'run away' test from specific to the other file as it _is_ no longer specific I'll assume you chose function containers carefully enough...

A thought: It might be mod-friendlier to not 'switch' on the possible roles, but to try in turn, the specific subfunction returns success/nope, and if nope try another role - as per a discussion with @ravignir in... 5010? You've seen it, surely. Anyway, include and test that while you're at it? Or should we do a fuzzy evaluateWorthOfWhatICanDoInThisRole(): Float - sorted() - first() - action approach much later?

PR #5032, opened 2021-08-30
Diplomacy Screen Nation relation indicator

SomeTroglodyte:
A little idea on the side - not _necessarily_ to be taken for 'desirable':
- More differentiation on the little circle bottom left of the Diplomacy Nation indicators - all RelationLevels
- Allied CS get an extra decoration as the hue progression is very gradual
- First Civ doesn't disappear under the Close button
- Sorted by name (status quo: sorted by json order) - I'm not too sure about this
- Rest is just moderate linting

image
Afraid is not on the picture, too fresh, got no simulation. Would be bright purple.
</details>

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -1,5 +1,6 @@
 package com.unciv.logic.civilization.diplomacy
 
+import com.badlogic.gdx.graphics.Color
 import com.unciv.Constants
 import com.unciv.logic.civilization.*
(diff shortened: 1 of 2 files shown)


yairm210:
We should add a war icon to the "at war" as well, to separate it from "unforgivable but at peace"

SomeTroglodyte:
> separate war from unforgivable but at peace

That's FIREBRICK vs. RED - detectable. But yes, the existing thingy meaning war on the Notifications... Uh.. Mapped to OtherIcons/Pillage - could do. Not a priority for me though - I actually _only_ wanted to PR that padding at first...

logicminimal:
shouldn't those colors also apply on the diplomacy tab?

PR #5034, opened 2021-08-30
Update Swedish.properties

SimonCeder:


xlenstra:
Placeholder `[civName2]` not found in `Swedish` for entry `[civName] has married into the ruling family of [civName2], bringing them under their control.` :)

PR #5038, opened 2021-08-31
Finishing the later five policy trees now allows you to buy great people with faith

xlenstra:
Continuing the development of religion and faith, this PR adds the ability to buy great people with faith after completing the five bottom policy trees.
This cost increasing in a quadratic rate (1000 - 1500 - 2500 - 4000- 6000 - 8500 - etc.) and the units can only be bought when a certain number of conditions have been met.
Due to the large amount of prerequisites for this, the unique for this has become quite long:

Added uniques:
- "May buy [unitFilter] units for [amount] [Stat] [cityFilter] starting from the [eraName] at an increasing price ([amount])

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Policies.json
+++ b/android/assets/jsons/Civ V - Vanilla/Policies.json
@@ -275,7 +275,9 @@
 			},
 			{
 				"name": "Commerce Complete",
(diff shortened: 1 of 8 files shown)


SomeTroglodyte:
Could save one param if we allowed Stats.isStats and Stats.parse _not_ requiring the '+' sign... Or by writing the unique with the sign and accepting the minor ugliness.. And won't nitpickers complain that the actual price increase soon is not the amount shown in pedia?

PR #5039, opened 2021-08-31
Added UI to show what cities are holy cities to the player

xlenstra:
Added a line both in the religion overview screen and in the religion expander tab in the city screen

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1033,11 +1033,13 @@ Follow [belief] =
 Religions and Beliefs = 
 Majority Religion: [name] = 
 + [amount] pressure = 
+Holy city of: [religionName] = 
 Pressure = 
 
 # Religion overview screen
 Religion Name: = 
 Founding Civ: = 
+Holy City: = 
 Cities following this religion: = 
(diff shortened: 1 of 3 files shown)


SomeTroglodyte:
Now that I see it, I'd like it better centered.

xlenstra:
> Now that I see it, I'd like it better centered.

Completely agree, so feel free to PR it. My usual technique of spamming `.center()` everywhere didn't seem to yield results so I just gave up on it

xlenstra:
> ```
> Exception in thread "LWJGL Application" kotlin.UninitializedPropertyAccessException: lateinit property civInfo has not been initialized
> 	at com.unciv.logic.city.CityInfo.getCivInfo(CityInfo.kt:31)
> 	at com.unciv.models.ruleset.unit.BaseUnit.canBePurchasedWithStat(BaseUnit.kt:225)
> 	at com.unciv.models.ruleset.unit.BaseUnit.getCivilopediaTextLines(BaseUnit.kt:129)
> ```

Should be fixed in #5057 when I noticed this same crash myself

SomeTroglodyte:
Aw, not another puzzle for me then...

PR #5043, opened 2021-08-31
Rework City State functions into separate file

SimonCeder:
This PR moves the city-state related functions to a separate file.
All functions in this file are written from the point of view of the city state, so some functions were rewritten to reflect this - for example `giveGoldGift` becomes `receiveGoldGift`.
There are wrapper functions in `CivilizationInfo.kt` to call these.


SomeTroglodyte:
Looks well thought out, but as devil's advocate I'd say...
The moved routines are reached each and every one by a class instantiation. Sure, the class contains just one field and so is tiny, but it's still work for the GC. Alternate approaches: object and pass the CS civInfo each time; or keep the instance as transient in the parent civInfo.

SimonCeder:
I added a transient in the civinfo. It does seem to be a tiny speed increase after testing a couple functions with `measureNanoTime`.

SomeTroglodyte:
I think giving the garbage collector less work is worth more than nanoseconds. Especially since Unciv still runs on the antiquated JVM's of phones. :+1:

PR #5051, opened 2021-09-01
Refactored some code and uniques

xlenstra:
This PR doesn't add any features, but refactors a lot of small things in the codebase.
I started with the idea of allowing the `MapUnit.getMatchingUniques` access to the uniques of the main civ, like I had done previously with cities, but ended up just changing a lot of random things. Included are:
- `MapUnit.getMatchingUniques` now has access to the civ uniques
- Added `CivilizationInfo.hasTechOrPolicy`, to easier check for `with []` uniques
- Made the setter of the `influence` variable for city states private, only allowing access via `DiplomacyManager.addInfluence` and `.setInfluence` so the allying civ is always updated
- Fixed a bug where trading away luxuries provided by city states would still give additional bonuses from luxuries from city states
- Converted some things from `1 + thing.toFloat() / 100f` to `thing.toPercent()`
- Refactored the way embarked movement is determined
- Rewrote a few uniques to allow them to be used not only for techs/policies/founding beliefs but also buildings/follower beliefs:

Added uniques:
- "[signedAmount]% unhappiness from population [cityFilter]"
- "[Stats] from every specialist [cityFilter]"
- "Population loss from nuclear attacks [signedAmount]% [cityFilter]"
- "[signedAmount]% unhappiness from specialists [cityFilter]"
- "[signedAmount]% unhappiness from population [cityFilter]"

Deprecated uniques:
- "Unhappiness from population decreased by [amount]%"
- "[Stats] from every specialist"
- "Population loss from nuclear attacks -[amount]%"
- "+2 Culture per turn from cities before discovering Steam Power"
- "Specialists only produce [amount]% of normal unhappiness"
- "Unhappiness from population decreased by [amount]%"
- "Unhappiness from population decreased by [amount]% [cityFilter]"
- "Increases embarked movement +1"
- "+1 Movement for all embarked units" -- Use "+[1] movement for all [Embarked] units" instead

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -506,6 +506,9 @@ class CivilizationInfo {
         return if (!gameInfo.hasReligionEnabled()) greatPeople.filter { !it.uniques.contains("Great Person - [Faith]")}.toHashSet()
         else greatPeople.toHashSet()
     }
+    
+    fun hasTechOrPolicy(techOrPolicyName: String) =
+        tech.isResearched(techOrPolicyName) || policies.isAdopted(techOrPolicyName)
 
     //endregion
 
@@ -877,8 +880,7 @@ class CivilizationInfo {
         if (!cityState.isCityState()) throw Exception("You can only gain influence with City-States!")
         addGold(-giftAmount)
         cityState.addGold(giftAmount)
-        cityState.getDiplomacyManager(this).influence += influenceGainedByGift(giftAmount)
-        cityState.updateAllyCivForCityState()
+        cityState.getDiplomacyManager(this).addInfluence(influenceGainedByGift(giftAmount).toFloat())
         updateStatsForNextTurn()
     }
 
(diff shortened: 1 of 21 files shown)


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -936,20 +938,20 @@ class CivilizationInfo {
...
-        if(!knows(otherCiv) or isAtWarWith(otherCiv)) return //Exception
+        if (!isCityState() || !otherCiv.isMajorCiv() || otherCiv.isDefeated()) return
+        if (!knows(otherCiv) || isAtWarWith(otherCiv)) return //Exception
 
         val diplomacy = getDiplomacyManager(otherCiv.civName)
         diplomacy.diplomaticStatus = DiplomaticStatus.Protector
     }
 
     fun removeProtectorCiv(otherCiv: CivilizationInfo) {
-        if(!this.isCityState() or !otherCiv.isMajorCiv() or otherCiv.isDefeated()) return
```
SomeTroglodyte, line 947: Astounding. The [doc](https://kotlinlang.org/docs/basic-types.html#bitwise-operations) says bitwise operators cannot be applied to Boolean, so why did this even compile before?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -198,6 +196,15 @@ class DiplomacyManager() {
             else -> false
         }
     }
+    
+    fun addInfluence(amount: Float) {
+        setInfluence(max(influence + amount, MINIMUM_INFLUENCE))
+    }
+    
+    fun setInfluence(amount: Float) {
+        influence = amount
```
SomeTroglodyte, line 205: bounds check better here? there's coerceIn & co

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -769,15 +767,15 @@ object UnitActions {
         val giftAction = {
             if (recipient.isCityState()) {
                 for (unique in unit.civInfo.getMatchingUniques("Gain [] Influence with a [] gift to a City-State")) {
-                    if((unit.isGreatPerson() && unique.params[1] == "Great Person")
-                        || unit.matchesFilter(unique.params[1])) {
-                        recipient.getDiplomacyManager(unit.civInfo).influence += unique.params[0].toInt() - 5
+                    if ((unit.isGreatPerson() && unique.params[1] == "Great Person")
```
yairm210, line 770: Don't great people have a "Great Person" unique? If not, then they probably should, the stat-based marking is obsolete

PR #5052, opened 2021-09-01
Great Prophets now always have your religion as their religion

xlenstra:
This PR makes it so that great prophets always have the religion you founded as their religion, regardless of where and when they spawned.


ravignir:
Hmm. But that is not civ5 behavior.

xlenstra:
> Hmm. But that is not civ5 behavior (if i understand correctly and captured prophets also change allegiance)

Sorry, I might have overgeneralized. All this change does, is that newly-spawned prophets always have your religion, and when you found a religion, all prophets in your empire change to your religion, conform civ 5. Captured great prophets can still be of other religions

PR #5055, opened 2021-09-01
When spreading a religion with a great prophet, other religions are now removed

xlenstra:


logicminimal:
is this meant to affect holy cities of other religions? also do inquisitors remove holiness of a city?

ravignir:
Afaik, Holy cities can't be purged in civ5, they always emanate some pressure even if you get the followers count down to 0.

xlenstra:
> Afaik, Holy cities can't be purged in civ5, they always emanate some pressure even if you get the followers count down to 0.

According to the sources I've found online, holy cities can be purged by inquisiteurs. However, I've chosen to ignore those sources, as its unclear to me what needs to happen for the city to become a holy city again, among other things.

xlenstra:
> is this meant to affect holy cities of other religions? also do inquisitors remove holiness of a city?

No, it should only remove religious pressures in the city it is spreading to, regardless of whether it is holy or not

PR #5057, opened 2021-09-01
Implemented renaming of religions

xlenstra:
This PR continues the development of religion by introducing the options to rename your religion at the time of founding.
This rename is purely visual and has no effect whatsoever.
The rename is accomplished by introducing a new type of popup: `AskTextPopup`, which is a generalization of the unit rename at promotion prompt.

<details> <summary> Pictures: </summary>

Found religion screen:
image

Ask Text Popup:
image

</details>


core/src/com/unciv/ui/pickerscreens/PromotionPickerScreen.kt
```diff
@@ -47,21 +47,32 @@ class PromotionPickerScreen(val unit: MapUnit) : PickerScreen() {
...
             renameButton.isEnabled = true
             renameButton.onClick {
-                RenameUnitPopup(unit, this).open()
+                AskTextPopup(
+                    this,
+                    label = "Choose name for [${unit.baseUnit.name}]",
+                    icon = ImageGetter.getUnitIcon(unit.name).surroundWithCircle(80f),
+                    defaultText = unit.name,
+                    actionOnOk = { userInput -> 
+                        if (userInput != "" && userInput != unit.name) {
```
SomeTroglodyte, line 64: Cound go in AskTextPopup, all other usecases could have the same check I think?

core/src/com/unciv/ui/utils/AskTextPopup.kt
```diff
@@ -0,0 +1,44 @@
...
+        nameField.maxLength = maxLength
+        
+        add(nameField).growX().colspan(2).row()
+        
+        addOKButton { actionOnOk(nameField.text); close() }
+        addCloseButton()
+        equalizeLastTwoButtonWidths()
+        keyboardFocus = nameField
+    }
+}
```
SomeTroglodyte, line 44: Doesn't that red circle bother you? Only reason to have LF's at file endings, really, but still...

Anyway: There's a third use for ask text, which you could find via Alt-F7 on Popup.keyboardFocus - rename city... More or less the same probably. Could do with the max length and filter, and that its OK button is named "Save" is minor.

docs/Credits.md
```diff
@@ -632,6 +632,7 @@ Unless otherwise specified, all the following are from [the Noun Project](https:
 * [Hourglass](https://thenounproject.com/search/?q=hourglass&i=142268) by I Create Stuff for the 'Turn' icon
 * [Shield](https://thenounproject.com/search/?q=shield&i=813568) by Gregor Cresnar for Religious Strength
 * [skill sword flame](https://thenounproject.com/term/skill-sword-flame/2360212/) by Maxicons) for Remove Heresy
+* [Pencil](https://thenounproject.com/search/?q=pencil&i=4195852) by Muhamad Aldi Maulana for Enter Text Prompt Button / Pencil
```
SomeTroglodyte, line 635: Yeah, but somehow - just slightly - I feel it's a bit off the average Unciv style. We're klunkier. And this is... Office. Isn't there some quill in policies that would do? Just a though, feel free to ignore.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -219,7 +219,7 @@ class BaseUnit : INamed, INonPerpetualConstruction, ICivilopediaText {
         ignoreCityRequirements: Boolean
     ): Boolean {
         // May buy [unitFilter] units for [amount] [Stat] starting from the [eraName] at an increasing price ([amount])
-        if (cityInfo.civInfo.getMatchingUniques("May buy [] units for [] [] [] starting from the [] at an increasing price ([])")
+        if (!ignoreCityRequirements && cityInfo.civInfo.getMatchingUniques("May buy [] units for [] [] [] starting from the [] at an increasing price ([])")
```
yairm210, line 222: You seem to have 3 [] instead of 2?

xlenstra: ah, missed the cityFilter in the comment, I'll add it in

PR #5061, opened 2021-09-02
Create Greek

ASLST:


logicminimal:
..what?

yairm210:
Nothing seems to have actually been translated, week wait for translations before adding this :)

ASLST:
Should i try after a week or keep translating for a week? Also i had trasnlated the first part, i do not know why is it not shown, have i done something wrong?

PR #5062, opened 2021-09-02
Refactor more code, hopefully increasing maintainability

xlenstra:
This PR refactors more code, solving a few small bugs as well as refactoring the `getRejectionReason()` functions to return a `HashSet` of reasons why the building was rejected. The elements of these hashSets are Enum values. This removes dependencies of the ordering of statements in these functions, and makes it easier to modify existing behaviour, by changing the enums in stead of having to wade through 200 line long functions to find the right spot to put your new check.

The small bugs fixed are the following:
- Fixed randomness in great person gifts from city states to match up with the behaviour from the original
- Removed a boolean that really should have been a nullable value
- Made "Great person" a check in `BaseUnit.matchesFilter`, removing the check from elsewhere.

This PR might have conflicts with #4617

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -1064,7 +1064,8 @@
 		"name": "Apollo Program",
 		"cost": 750,
 		"isNationalWonder": true,
-		"uniques": ["Enables construction of Spaceship parts", "Triggers a global alert upon completion"],
+		"uniques": ["Enables construction of Spaceship parts", "Triggers a global alert upon completion", 
+			"Hidden when [Scientific] Victory is disabled"],
 		"requiredTech": "Rocketry"
 	},
 
@@ -1091,7 +1092,7 @@
 		"name": "SS Cockpit",
 		"requiredResource": "Aluminum",
 		"requiredTech": "Satellites",
-		"uniques": ["Spaceship part", "Triggers a global alert upon completion", "Cannot be purchased"]
+		"uniques": ["Spaceship part", "Triggers a global alert upon completion", "Cannot be purchased", "Hidden when [Scientific] Victory is disabled"]
 	},
 	{
 		"name": "Hubble Space Telescope",
@@ -1109,7 +1110,7 @@
 		"name": "SS Booster",
 		"requiredResource": "Aluminum",
 		"requiredTech": "Advanced Ballistics",
-		"uniques": ["Spaceship part", "Triggers a global alert upon completion", "Cannot be purchased"]
+		"uniques": ["Spaceship part", "Triggers a global alert upon completion", "Cannot be purchased", "Hidden when [Scientific] Victory is disabled"]
 	},
 	{
 		"name": "Spaceship Factory",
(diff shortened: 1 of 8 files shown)


yairm210:
I think that this is a good way of quickly changing the status quo, and if we have improvements we can make them on top of the new framework

SomeTroglodyte:
The three `private val ... hashSetOf()/listOf()` instantiate objects on the heap every time the reasons class is instantiated, but are semantically static. Companion - or property of RejectionReason. Enums are ordered intrinsically, and you can do ranges of them, also an option. Ranges and properties are free performance-wise, while looking up HashSets is not, even when static.

Otherwise - hell yeah, finally :+1: :balloon: :birthday: :partying_face:

SomeTroglodyte:
How could I miss this:
`RejectionReason` - an `enum` with a `var` that gets changed around at run time. Uh, every enum 'entry' is an instance that exists only once throughout the process. Enum properties are great, for ***static*** stuff. If there could ever be a case of two versions of the same RejectionReason with different errorText needed at the same time, what then?

See `BaseUnit.getRejectionReasons`: RejectionReason.RequiresTech could theoretically be added twice or more to the same result set. If that happens, you have the ***same*** instance of RejectionReason a few times in that collection, and only one errorText wins.

Bad bad pattern. var on enum - when you know you're going to initialize it once, but can't do so in a constructor, maybe. If the singleton character of the enum entry is a necessary property of the var - and you document the trick. Otherwise - think of enum values as const, and if you need something like a RejectionReason with a dynamic message, that's a normal class, maybe a data class, and maybe containing an enum val for its type...

So - can we ***guarantee*** no case of several RejectionReason's with different errorText will try to coexist at the same time? Otherwise we invite very hard to find bugs.

SomeTroglodyte:
OK, I think this can be solved with an interface, so the existing RejectionReason fulfills that contract mostly as it is, but there's a second implementation of the interface allowing dynamic creation with any errorMessage...

lishaoxia1985:
@xlenstra @yairm210 Although it is good, but it also have some problems:
1. By the method getRejectionReasons we get all the RejectionReasons,  but get all the RejectionReasons is not necessary, we only need to know these RejectionReasons:  
+ RejectionReasons makes the unit/ building hide in the construction list. we only need to know one of these, not all of them.
+ RejectionReasons that when we solve these, the unit/ building will be buildable. these type RejectionReason we should know all of them. These should show in info table.
2. Now using the method getRejectionReasons will lead to the construction list update very slowly, especially when you open City Screen.
3. The unique includes ConstructionRequire, effect, splitting them is an impossible task. So when you show the unit/building info and ConstructionRequire state perfect in the mean time is an impossible task, too.

xlenstra:
1. Agree, that should be changed
2. Agree, I've seen one or two complaints about this in the discord too
3. I'm not sure I quite understand what you mean, but I'll assume you mean that the 'effect' and 'requirements' of a building/unit are both parts of their unique, which makes it difficult to distinguish these two. If that's the case, I disagree that that is a problem; 'uniques' should be all unique abilities of a building/unit, regardless whether that are unique requirements to build them or unique effectes they provide. Splitting them is not impossible, especially as detecting them using either `[unique].placeholderText == "some string with [] parameters"` or `[unique].isOfType(Enum)` is possible. It would require more filtering when creating the text for displaying the abilities of the building, but not be impossible. If you don't merge this two, you still have to convert the requirements to a string when displaying it somewhere else, such as in the civilopedia

xlenstra:
> OK, I think this can be solved with an interface, so the existing RejectionReason fulfills that contract mostly as it is, but there's a second implementation of the interface allowing dynamic creation with any errorMessage...

Good catch, though I'd rather make `RejectionReason` a data class with two fields: the enum value of the error and a string of the error that could potentially be displayed to the user.

SomeTroglodyte:
Oops, I just deleted the PR with the case study to go with the "an enum with a var" comment and the interface segue - but the branch is still there:
https://github.com/SomeTroglodyte/Unciv/tree/Fix-RejectionReason
I still think writing into an enum instance is evil (maybe unless the enum exists explicitly for that purpose, to have an ordered/indexable set of singletons). But best the solution may be simpler - wrap the enum - I guess just as you said.

Also, using rejection reason enumeration as a boolean isn't necessarily a performance bottleneck - if the iterator is cheap (sequence) and the "ballast" is. Lugging around a string reference you discard later is cheap. Creating that is maybe not. In other words, if all messages are _interned_ string _constants_ somewhere, passing them is almost free. Formatting strings for the rejection reasons, or tr(), would be the other side. Personally, I suspect some other bottleneck in that constructions builder, just a hunch.

Lastly, separating effect and requirement uniques - many of our routines preserve json order, Civilopedia among them, so that should be the first place to look - because free & very little risk of anything breaking - ?.

PR #5072, opened 2021-09-02
You can now input distinct numbers when trading gold

xlenstra:
After creating the `AskTextPopup` in #5057, I felt like doing the exact same thing again, so I created an `AskNumberPopup`, which is very similar but only asks the user for a positive integer. Picture below. I then used this new widget to finally make it possible to send exact amounts of gold in trade requests, and also added the same features for adding 'gold per turn' to trade requests.
Also removed extraneous spaces from `template.properties`.

<details> <summary> Picture: </summary>

image

</details>


SomeTroglodyte:
> extraneous spaces

:rofl:  #5073  
.. And you didn't get the single-space ones.

core/src/com/unciv/ui/utils/AskNumberPopup.kt
```diff
@@ -0,0 +1,114 @@
...
+ * @param actionOnOk Lambda that will be executed after pressing 'OK'.
+ */
+
+class AskNumberPopup(
+    screen: CameraStageBaseScreen,
+    label: String = "Please enter a number",
+    icon: IconCircleGroup = ImageGetter.getImage("OtherIcons/Pencil").apply { this.color = Color.BLACK }.surroundWithCircle(80f),
+    defaultText: String = "",
+    amountButtons: List<Int> = listOf(),
+    bounds: Array<Int?> = arrayOf(null, null),
```
SomeTroglodyte, line 26: Range<Int>? - that does exist, right? No, it's called IntRange. I must not lie.

core/src/com/unciv/ui/utils/AskNumberPopup.kt
```diff
@@ -0,0 +1,114 @@
...
+     * new number after clamping now disallows overwriting the existing minus sign with a new minus
+     * sign, as there is already a minus sign in the number. Well, no problem, you can just remove
+     * the number before overwriting it with the clamped variant. But now you reset your cursor
+     * position every time you type a character. You might start trying to cache the cursor position
+     * as well, but at that point you're basically rewriting the setText() function, and when I
+     * reached this point I decided to stop. 
+     * 
+     * P.S., if you do decide to go on this quest of adding minus signs, don't forget that 
+     * `"-".toInt()` also crashes, so you need to exclude that before checking to clamp.
+     */
```
SomeTroglodyte, line 47: Should be possible to let'em spam as many minuses as they want in the Filter, and have a listener that monitors if it parses to int and sets OK.isEnabled... Ah you have validation further down anyway...

core/src/com/unciv/ui/utils/AskNumberPopup.kt
```diff
@@ -0,0 +1,114 @@
...
+        add(centerTable).colspan(2).row()
+
+        for (value in sortedAmountValues) {
+            addValueButton(value)
+        }
+        
+        val errorLabel = errorText.toLabel()
+        errorLabel.color = Color.RED
+
+        addOKButton(automaticallyCloseOnPress = false) {
```
SomeTroglodyte, line 105: I'd model the usecase more like other frameworks as open fun validate() = true or somesuch. If that returns false, block OK-action (now we need a sound - possibly quite flatulent-sounding :trumpet: )
_or_ the text entry box gets a kind of Type enum that controls filter, listener, and OK validation without exposing the intrinsics of it, and this class would only add the buttons.

core/src/com/unciv/ui/utils/AskNumberPopup.kt
```diff
@@ -0,0 +1,114 @@
...
+        } 
+        
+        nameField.onChange {
+            if (nameField.text.isNotEmpty()) {
+                nameField.text = clampInBounds(nameField.text.toInt()).toString()
+            }
+        }
+        
+        val centerTable = Table(skin)
+        val sortedAmountValues = amountButtons.sorted()
```
SomeTroglodyte, line 74: I would have left this the responsibility of the caller

core/src/com/unciv/ui/utils/AskNumberPopup.kt
```diff
@@ -0,0 +1,114 @@
...
+class AskNumberPopup(
+    screen: CameraStageBaseScreen,
+    label: String = "Please enter a number",
+    icon: IconCircleGroup = ImageGetter.getImage("OtherIcons/Pencil").apply { this.color = Color.BLACK }.surroundWithCircle(80f),
+    defaultText: String = "",
+    amountButtons: List<Int> = listOf(),
+    bounds: Array<Int?> = arrayOf(null, null), 
+    errorText: String = "Invalid input! Please enter a valid number.",
+    actionOnOk: (input: Int) -> Boolean = { true },
+): Popup(screen) {
```
SomeTroglodyte, line 29: Not building on top of AskTextThingy? interesting.

xlenstra: The code inside the `init` function differs on multiple points. The only way I could have built on top of the `AskTextPopup` was if I created a `AskInputPopup` which has a parameter asking for int/string input, which, although reducing code duplication, feels like overcomplication

core/src/com/unciv/ui/utils/AskNumberPopup.kt
xlenstra: That is probably better, as it also allows for the usecase were the caller wants them reordered for some strange reason

Issue #5071, opened 2021-09-02
Civilizations are incorrectly alphabetized in some languages


SimonCeder:
**Platform**
Linux 64 bit.

**Version**
Git, 0389945ab186774b2ec9ad2f600f22f52cf8f664 (3.16.11 more or less)

**Describe the bug**
The alphabetizer function puts 'ö'  right after 'o', whereas in Swedish (and some others??) 'ö' is the last letter of the alphabet.

**To Reproduce**
1. Set language to Swedish
2. Open new game screen
3. Open the civilization picker

**Expected behavior**
"Österrike" (Austria) should go at the end of the list of civilizations.

**Screenshots**
![Screenshot_2021-09-02_23-16-21](https://user-images.githubusercontent.com/63475501/131918137-bec0f44c-f6e8-4fbf-b484-b6c7ba8f3f8e.png)

**Additional information**
The civilopedia is likewise affected.

SomeTroglodyte:
Just search for `Collator.getInstance()` in the Project, and you'll find all instances. The javadoc for that is:
```
    /**
     * Gets the Collator for the current default locale.
     * The default locale is determined by java.util.Locale.getDefault.
     * @return the Collator for the default locale.(for example, en_US)
     * @see java.util.Locale#getDefault
     */
    public static synchronized Collator getInstance() {
        return getInstance(Locale.getDefault());
    }
```

... so you see, it's not dependent on Unciv's language choice but on the System default. I think I commented that fact in the original PR - blame says #4522 

For the collation to follow Unciv's language choice, we'd need to maintain a list of Java-compatible locales with the language. A nice challenge for a PR, but I didn't think it worthwhile. Also, the collation logic would still be the system's, and could still be "erroneous".

SimonCeder:
We could add a "LANGUAGE_CODE" string to template.properties? Which would then be translated into "sv" in Swedish for instance, there's a list of locales [here](https://www.oracle.com/java/technologies/javase/jdk11-suported-locales.html).
Trying `.sortedWith(compareBy(Collator.getInstance(Locale("sv")), { it.name.tr() }))` gives correct sorting for Swedish.
But perhaps there's some wrinkle I'm unaware of?

SomeTroglodyte:
Thanks for helping think this through, I haven't got much brain capacity free these days. 
[...]

PR #5082, opened 2021-09-03
implement locale for proper sorting on certain screens

SimonCeder:
As discussed in #5071 
This PR sets the appropriate java Locale upon choosing a language to allow for proper localized sorting in the civ picker, civilopedia, and religion picker.
Gives proper sorting in german and swedish. It should work for all supported languages but I can't really tell!


core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -13,6 +17,8 @@ class GameSettings {
     var checkForDueUnits: Boolean = true
     var singleTapMove: Boolean = false
     var language: String = "English"
+    @Transient
+    var locale: Locale? = null
```
SomeTroglodyte, line 21: Loading settings from disk will need to set this too, and I can't see it. Must be blind.

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -68,4 +74,64 @@ class GameSettings {
...
+    Indonesian("in", "ID"),
+    Italian("it", "IT"),
+    Japanese("ja", "JP"),
+    Korean("ko", "KR"),
+    Latvian("lv", "LV"),
+    Lithuanian("lt", "LT"),
+    Malay("ms", "MY"),
+    Norwegian("no", "NO"),
+    NorwegianNynorsk("nn", "NO"),
+    // Sorry Persian is not in the supported locales
```
SomeTroglodyte, line 122: Maybe called Iranian?

SomeTroglodyte:
I've linked the issue - you know that's [automatic](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) if you use (resolve, resolves, fixes) or a few others in the comment just before an issue reference, right?

core/src/com/unciv/models/metadata/GameSettings.kt
SomeTroglodyte: [Farsi](https://www.localeplanet.com/icu/fa-IR/index.html) - wonder if Collator.getInstance accepts "fa"

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -68,4 +74,64 @@ class GameSettings {
...
+    Hindi("hi", "IN"),
+    Hungarian("hu", "HU"),
+    Indonesian("in", "ID"),
+    Italian("it", "IT"),
+    Japanese("ja", "JP"),
+    Korean("ko", "KR"),
+    Latvian("lv", "LV"),
+    Lithuanian("lt", "LT"),
+    Malay("ms", "MY"),
+    Norwegian("no", "NO"),
```
SomeTroglodyte, line 120: I wonder if we ever are lucky enough to get a norwegian xlator if they wouldn't prefer their files to be named just Bokmål and Nynorsk - nah, no reason to worry now :laughing:

core/src/com/unciv/models/metadata/GameSettings.kt
SomeTroglodyte: Yup, I see it.

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -68,4 +74,64 @@ class GameSettings {
...
+        val languageName = language.filterNot { it in bannedCharacters }
+        val code = try {
+            LocaleCode.valueOf(languageName)
+        } catch (e: IllegalArgumentException) {
+            LocaleCode.English
+        }
+        locale = Locale(code.language, code.country)
+    }
+
+    fun getCurrentLocale(): Locale {
```
SomeTroglodyte, line 89: as most (if not all?) calls chain a get Collator instance - centralize that here too??

core/src/com/unciv/models/metadata/GameSettings.kt
SomeTroglodyte: OK, fa_IR doesn't crash on desktop or Oreo. Too lazy to fire up a Lollipop or thereabouts AVM. Of course, how it actually sorts - who knows. But moving the Locale(code.language, code.country) call inside the try should cover any stone age java having a shorter locale list. So a best-effort implementation for Persian is possible. As Persian sits at 20% I guess we have no tester to ask.

core/src/com/unciv/models/ruleset/Belief.kt
```diff
@@ -62,7 +62,7 @@ class Belief : INamed, ICivilopediaText, IHasUniques {
             name = "Religions"
             val lines = ArrayList<FormattedLine>()
             lines += FormattedLine(separator = true)
-            ruleset.religions.sortedWith(compareBy(Collator.getInstance(), { it.tr() })).forEach {
+            ruleset.religions.sortedWith(compareBy(Collator.getInstance(UncivGame.Current.settings.getCurrentLocale()), { it.tr() })).forEach {
```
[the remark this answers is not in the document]
SomeTroglodyte: No, pedia is not the place for specific game stuff, that's for overview and the like -> OK.

SomeTroglodyte (approved):
Maybe over the top perfectionism, but well done.

core/src/com/unciv/models/metadata/GameSettings.kt
SimonCeder: Literally nothing seems to crash the Locale constructor, it seems to fall back to some default value. English or system default? Iranian or Farsi is not in the list of supported languages over at Oracle, falling back to system locale is probably best option.

core/src/com/unciv/models/metadata/GameSettings.kt
SimonCeder: Mainly I wanted to avoid having a non-ASCII character in the name. If a Norwegian ever shows up we can ask their opinion.

yairm210:
The conflict here is very strange, needs your help.

PR #5083, opened 2021-09-04
A flag for Traditional_Chinese

SomeTroglodyte:
Working on the options screen it bothered me there was this optical "hole" in the list - the dark blue circle. From Simon's PR I learned that the language would be associated with Taiwan, so I fetched that flag from the already credited source. No, the Flags.atlas file needs no update as the position within the texture doesn't change.


yairm210:
There was a debate when including the traditional chinese on what flag to give it, whether to associate with China or not, in the end we "opted out" by having it a blank flag.
But this is probably fine, let's see if we get any comments about it, and we can always revert

PR #5095, opened 2021-09-05
From the indstrial era onwards, things change in religion

xlenstra:
Starting from the industrial era, the following changes occur:
- Great prophets are no longer automatically generated
- You may buy them with faith instead. I've assumed the prices to be the same as for other great people, though if someone could test that, that would be appreciated

Furthermore, starting a game in the industrial era or later completely disables religion, conform the base game. When religion will in the future be enabled by default, I'll make sure this is communicated clearly to the user


SimonCeder:
>     * You may buy them with faith instead. I've assumed the prices to be the same as for other great people, though if someone could test that, that would be appreciated

It's not.
Screenshot_2021-09-05_22-29-29

xlenstra:
> > ```
> > * You may buy them with faith instead. I've assumed the prices to be the same as for other great people, though if someone could test that, that would be appreciated
> > ```
> 
> It's not.
> Screenshot_2021-09-05_22-29-29

Welp, time to start spitting through the source code to find the formula they use for determining the price

SimonCeder:
CvReligionClasses.cpp, line 1557,
and the values:
```
<Row Name="RELIGION_MIN_FAITH_FIRST_PROPHET">
	<Value>200</Value>
</Row>
<Row Name="RELIGION_MIN_FAITH_FIRST_GREAT_PERSON">
	<Value>1000</Value>
</Row>
<Row Name="RELIGION_FAITH_DELTA_NEXT_PROPHET">
	<Value>100</Value>
</Row>
<Row Name="RELIGION_FAITH_DELTA_NEXT_GREAT_PERSON">
	<Value>500</Value>
</Row>
```

xlenstra:
> CvReligionClasses.cpp, line 1557,
> and the values:
> 
> ```
> <Row Name="RELIGION_MIN_FAITH_FIRST_PROPHET">
> 	<Value>200</Value>
> </Row>
> <Row Name="RELIGION_MIN_FAITH_FIRST_GREAT_PERSON">
> 	<Value>1000</Value>
> </Row>
> <Row Name="RELIGION_FAITH_DELTA_NEXT_PROPHET">
> 	<Value>100</Value>
> </Row>
> <Row Name="RELIGION_FAITH_DELTA_NEXT_GREAT_PERSON">
> 	<Value>500</Value>
> </Row>
> ```

So its just the same as the normal great prophet generation? That shouldn't be too difficult to implement

SimonCeder:
If I'm reading it right, the prophet cost is based on a global number of prophets (of this religion) created, and other GP's is based on a per player count of great people created.

xlenstra:
> If I'm reading it right, the prophet cost is based on a global number of prophets (of this religion) created, and other GP's is based on a per player count of great people created.
> Also see `CvCity::GetFaithPurchaseCost`, CvCity.cpp line 4413

Since all great prophets spawn with the religion of your civ, thoses numbers should coincide if I'm not mistaken, so I've implemented it that way here

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -718,9 +725,12 @@ class CityInfo {
                 if (!tilesList.contains(tile))
                     cityPositionList.add(tile)
 
-        return cityPositionList.asSequence()
-            .map { it.getOwner()?.civName }.filterNotNull()
-            .distinct().toList()
+        return cityPositionList
+            .asSequence()
+            .map { it.getOwner()?.civName }
```
yairm210, line 730: I think there's a mapNotNull

core/src/com/unciv/models/ruleset/Era.kt
```diff
@@ -40,6 +42,10 @@ class Era : INamed {
 
     fun getHexColor() = "#" + getColor().toString().substring(0, 6)
 
+    fun hasUnique(uniqueTemplate: String) = uniqueObjects.any { it.placeholderText == uniqueTemplate }
```
yairm210, line 45: Maybe we should have an interface for Inamed with uniques, this is getting repetitive

PR #5104, opened 2021-09-05
Update Greek.properties

ASLST:


xlenstra:
Placeholder `[ourUnit]` not found in `Greek` for entry `Your [ourUnit] captured an enemy [theirUnit]!`

PR #5107, opened 2021-09-06
GameStarter wrap and shape aware

SomeTroglodyte:
- Starting locations code aware of wrap and map shape
- RiverGenerator wrap aware
- RiverGenerator guard against endless loop

current master:
image

this PR:
image
(17 major civs + 18 CS on `mapParameters:{type:Archipelago,shape:Rectangular,mapSize:{radius:56,width:120,height:80,name:Custom},worldWrap:true,seed:1630921774011,tilesPerBiomeArea:9,maxCoastExtension:3,elevationExponent:0.75,temperatureExtremeness:0.68,vegetationRichness:0.5,rareFeaturesRichness:0.1,resourceRichness:0.15,waterThreshold:-0.049999997}}`)

a few less civs:
image
even fewer:
image

</details>

No screenies for River generation, I - uh - didn't test visually, I only tested the edge cases for the two RiverCoordinate.getAdjacent* functions in the debugger, and the result shouldn't be any worse :shrug: 

Actually, regarding the looping situation: I' now _certain_ was due to wrap, because get nearest water was already wrap-aware by virtue of using `getTilesAtDistance` but the 'routing' was only half so (`aerialDistanceTo` already knew wrap too), so this _is_ a bugfix for a rare situation. I've left the guard in anyway. Shall we up the limit? Worst case is some river just ending on land.

### DIFF
--- a/core/src/com/unciv/logic/HexMath.kt
+++ b/core/src/com/unciv/logic/HexMath.kt
@@ -2,6 +2,8 @@ package com.unciv.logic
 
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.math.Vector3
+import com.unciv.logic.map.MapParameters
+import com.unciv.logic.map.MapShape
 import kotlin.math.*
 
 @Suppress("MemberVisibilityCanBePrivate", "unused")  // this is a library offering optional services
@@ -20,7 +22,7 @@ object HexMath {
         if (size < 0) return 0
         return 1 + 6 * size * (size + 1) / 2
     }
-    
+
     /** Almost inverse of [getNumberOfTilesInHexagon] - get equivalent fractional Hexagon radius for an Area */
     fun getHexagonalRadiusForArea(numberOfTiles: Int) =
         if (numberOfTiles < 1) 0f else ((sqrt(12f * numberOfTiles - 3) - 3) / 6)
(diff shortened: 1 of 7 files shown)


yairm210:
I think this is fine

core/src/com/unciv/logic/HexMath.kt
```diff
@@ -194,4 +196,34 @@ object HexMath {
...
+        val y = vector.y.toInt()
+        if (mapParameters.shape == MapShape.rectangular) {
+            val height = mapParameters.mapSize.height
+            val width = mapParameters.mapSize.width
+            val left = if (mapParameters.worldWrap) Int.MAX_VALUE else width / 2 - (x - y)
+            val right = if (mapParameters.worldWrap) Int.MAX_VALUE else (width - 1) / 2 - (y - x)
+            val top = height / 2 -  (x + y) / 2
+            // kotlin's Int division rounds in different directions depending on sign! Thus 1 extra `-1`
+            val bottom = (x + y - 1) / 2 + (height - 1) / 2
+            return min(min(left, right), min(top, bottom))
```
yairm210, line 211: Can't min take parameters as needed? Do we really need nesting?

yairm210:
Nice job moving the adjacency function to the river coordinate! That's a much better place for them and it makes the whole thing more readable!

core/src/com/unciv/logic/HexMath.kt
SomeTroglodyte: Not that I know... There's surely some kotlin extension emulating SQL for such - but - it's just the one place and I doubt the bytecode could get better. At least I didn't nest simply left to right! A generic fun for that might be fun to do... Can you _slice_ varargs?

SomeTroglodyte:
Uh-oh, the merge shows a notification 'you have entered the com.unciv.models.ruleset.Era@address' which is not my fault but also a 'DISASTROUSLY' on next turn which I have to check whether it's mine...

SomeTroglodyte:
Note to self: `UnitAutomation.tryAccompanySettlerOrGreatPerson` is in the stack trace, and the Exception is the "I can't go there!" one in `MapUnit.putInTile`...

SomeTroglodyte:
Now it's coming from tryExplore - and it's failing because the target tile is already occupied. Cannot be due to this PR.

SomeTroglodyte:
Forgot an explanation: This also tweaks the locations slightly:
- Tiles that are one step away from the coast but aren't coast themselves are strongly discouraged. Effect easily visible with repeated starts as spectator, but such a location still can occur - which probably means I chose a good weight.
- Yield of the center tile now included in the score. It _can_ be better than the minimum 2 food 1 prod for a city center, and that should count.

core/src/com/unciv/logic/HexMath.kt
yairm210: Varargs is in the end an array, and you can inputvarargs with the .. operator, so yes
I think in tileSetStrings I played around with this a little

SomeTroglodyte: ... and the kotlin lib already has it. called `minOf`. next time.

PR #5108, opened 2021-09-06
Kotlin 1.5 new warnings - partial

SomeTroglodyte:
Since the new Kotlin level deprecates some features we've used prolifically, a clean rebuild now lists a few more warnings than usual. I've long had a branch named "Anti-warnings" aiming to at least get rid of the warnings visible in a github build (those that aren't our own deprecations), so now seems the time to start that fresh.

I've just picked a few more peripheral files - Question: Is this the right direction?

Oh, Simulation didn't even compile here without upping the min api level a few notches, so I refactored it to the kotlin Duration instead of the java one. Its [toString](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/-duration/to-string.html) may not give identical results to the dropped helper, but - it's standard and close.

### DIFF
--- a/core/src/com/unciv/logic/IdHelper.kt
+++ b/core/src/com/unciv/logic/IdHelper.kt
@@ -1,6 +1,7 @@
 package com.unciv.logic
 
 import java.util.*
+import kotlin.math.abs
 
 /**
  * This class checks whether a Game- or Player-ID matches the old or new format.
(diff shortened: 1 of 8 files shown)


SomeTroglodyte:
`@ExperimentalTime` is a bummer, though...

yairm210:
Adhering to the Kotlin standard is definitely the way to go

SomeTroglodyte:
> Adhering to the Kotlin standard is definitely the way to go

I concur, even if they're contradicting themselves. I mean, sumOf I knew was coming and sometimes I already coded accordingly. The Char changes - I even had that prepared. But the new uppercase/lowercase? _Why_??? :sob:

PR #5121, opened 2021-09-06
Kotlin 1.5 new warnings - partial - managers

SomeTroglodyte:
This warnings-be-gone has an interesting side **question**: What to do about the bad spelling of Denunciation in DiplomacyFlags? That's serialized into saves I suppose...

### DIFF
--- a/core/src/com/unciv/logic/city/PopulationManager.kt
+++ b/core/src/com/unciv/logic/city/PopulationManager.kt
@@ -64,7 +64,7 @@ class PopulationManager {
             var percentOfFoodCarriedOver = cityInfo
                 .getMatchingUniques("[]% of food is carried over [] after population increases")
                 .filter { cityInfo.matchesFilter(it.params[1]) }
-                .sumBy { it.params[0].toInt() }
+                .sumOf { it.params[0].toInt() }
             // Try to avoid runaway food gain in mods, just in case 
             if (percentOfFoodCarriedOver > 95) percentOfFoodCarriedOver = 95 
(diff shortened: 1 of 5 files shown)


xlenstra:
The way I've done these so far:
- Add an extra flag, so there is both the wrong and right flag
- Deprecate the wrong flag
- Whenever `setTransients` of the object is called, move the wrong flag to the right flag and then remove the wrong flag
- Eventually, remove the deprecation

SomeTroglodyte:
Yup - worth it or hide under carpet is the question. Never fix a working gadget and so on.

yairm210:
> The way I've done these so far:
> 
> * Add an extra flag, so there is both the wrong and right flag
> * Deprecate the wrong flag
> * Whenever `setTransients` of the object is called, move the wrong flag to the right flag and then remove the wrong flag
> * Eventually, remove the deprecation

Precisely!
Since I understand you're currently in college, I will say that this is good practice.
Almost all production systems I've worked on have had some kind of "data migration" at one point or another, often multiple at once, and soft depreciation is a good solution

SomeTroglodyte:
Yup again - that may even exist in my old "Anti-warnings-campaign" branch, I can't even remember. But as `Denunceation` is just an opcode and not displayed, I'll think of it as a 'hide under carpet' case for now.

SomeTroglodyte:
... Since I went pretty arbitrarily about choosing where to patch the new kotlin warnings first, that "issue" is not finished, though we should be more than halfway. I promise I'll go for the rest, including all the old friends, in time.

PR #5122, opened 2021-09-06
Fixes #5113

### LINKED ISSUE #5113: AI never improves strategic resources if automated improvement replacement is disabled
**Platform**
Android

**Version**
Play Store, 3.16.10

**Describe the bug**
AI players will never replace improvements on tile resources if "Automated workers replace improvement" is disabled in the game's settings.

City states will reach the endgame and beyond without ever giving you aluminium in their borders, if they've previously built a farm on it.

Competing empires will stay locked off from advanced units and nuclear deterrence, because their uranium and oil are covered in trading posts.

Enabling "Automated workers replace improvements" causes them all to immediately start building the right improvements on unimproved strategic resource tiles.

**To Reproduce**
1. Play to the endgame with "Automated workers replace improvements" disabled.
2. Check the strategic resource tiles in AI-controlled territory, and see that they're often covered with the wrong improvements.
3. Enable "Automated workers replace improvements", and play a few more turns.
4. Check for new improvements on the strategic resource sites.

**Expected behavior**
AI workers should probably always be treated as if "Automated workers replace improvements" is turned on.

xlenstra:
Fixes #5113

### DIFF
--- a/core/src/com/unciv/logic/automation/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/WorkerAutomation.kt
@@ -281,7 +281,7 @@ class WorkerAutomation(
         val city = tile.getCity()
         if (city == null || city.civInfo != civInfo)
             return false
-        if (tile.improvement != null && !UncivGame.Current.settings.automatedWorkersReplaceImprovements)
+        if (tile.improvement != null && !UncivGame.Current.settings.automatedWorkersReplaceImprovements && UncivGame.Current.gameInfo.currentPlayerCiv.isPlayerCivilization())
             return false
 
         if (tile.improvement == null) {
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -281,14 +281,15 @@ class WorkerAutomation(
         val city = tile.getCity()
         if (city == null || city.civInfo != civInfo)
             return false
-        if (tile.improvement != null && !UncivGame.Current.settings.automatedWorkersReplaceImprovements)
+        if (tile.improvement != null && !UncivGame.Current.settings.automatedWorkersReplaceImprovements && UncivGame.Current.gameInfo.currentPlayerCiv.isPlayerCivilization())
```
SomeTroglodyte, line 284: The class has the `civInfo` _owning_ the worker cached - wouldn't that be the safer test?

Issue #5045, opened 2021-09-01
Disable two-finger tap on mobile, conflicts with pinch-to-zoom?


will-ca:
**Platform**
Android, Play Store version

**Version**
Play Store
3.16.7

**Describe the bug**
It seems really easy to accidentally order a unit move when trying to pinch-to-zoom on mobile, without the confirmation bubble or anything.

I *think* this is because a quick "pinch" gesture (two-finger tap) gets registered as a right click event, which translates to an instant move for Desktop.

**To Reproduce**
* Select a unit with movement points.
* Do a quick tap with two fingers, like you were going to pinch to zoom but then thought better of it.

**Expected behavior**
The multitouch gesture should not immediately order your worker halfway into enemy territory.

logicminimal:
how to reprodude this problem exactly;
1:press and hold first finger in anywhere the game screen
2: press a tile that your selectwd unit can move onto

i knew this bug for months on end but didn't care about opening an issue about it

PR #5125, opened 2021-09-06
No right-click on Android

SomeTroglodyte:
Simplest way to solve #5045 ?
Testing ~~underway~~ successful.


logicminimal:
ooh, so it counts *this* action as a right click? TIL

PR #5129, opened 2021-09-06
Added "Consumes [amount] [resource]" for improvements

xlenstra:
Conform request in #3242, but only for improvements buildable by workers, not placeables such as great improvements


ravignir:
It is a pity it does not work with placeable improvements.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -585,11 +585,15 @@ object UnitActions {
...
         if (unit.religiousActionsUnitCanDo().all { unit.abilityUsesLeft[it] == unit.maxAbilityUses[it] })
             uniquesToCheck += unit.getMatchingUniques("Can construct [] if it hasn't used other actions yet")
+        val civResources = unit.civInfo.getCivResourcesByName()
         
         for (unique in uniquesToCheck) {
             val improvementName = unique.params[0]
             val improvement = tile.ruleset.tileImprovements[improvementName]
                 ?: continue
+            if (improvement.uniqueObjects.any { 
+                    it.placeholderText == "Consumes [] []" && civResources[unique.params[1]] ?: 0 < unique.params[0].toInt()
```
yairm210, line 595: If the resources don't exist, the action should be grayed out, not disappeared completely

PR #5130, opened 2021-09-06
Wrote an extensive tutorial documenting most of religion

xlenstra:


android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
 "ⓨ: By the way, here's how an empire border looks like - it's in the national colours of the nation owning the territory.",
 "ⓩ: And this is the red targeting circle that led to the attack pane back under ⑬.",
 "What you don't see: The phone/tablet's back button will pop the question whether you wish to leave Unciv and go back to Real Life. On desktop versions, you can use the ESC key."
-  ]
+  ],
+    "Faith": [
+        "After building a shrine, you're civilization will start generating faith.",
```
SomeTroglodyte, line 173: 'you are' vs. 'your'

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "When enough faith has been generated, you will be able to found a pantheon.",
+        "A pantheon will provide a small bonus for your civilization that will apply to all your cities.",
+        "Each civilization can only choose a single pantheon belief, and each pantheon can only be chosen once.",
+        "Generating more faith will allow you to found a religion."
+    ],
+    "Religion": [
+        "When you have gathered enough religion, a great prophet will be born in one of your cities.",
```
SomeTroglodyte, line 180: enough faith? By the way, the Fonts.faith symbol will work, even if you don't have access to the name from json

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "This great prophet can be used for multiple things: Constructing a holy city, founding a religion and spreading your religion.",
+        "When founding your religion, you may choose another two beliefs. The founder belief will only apply to you, while the follower belief will apply to all cities following your religion.",
+        "Additionally, the city where you used your great prophet will become the holy city of that religion.",
+        "If after founding your religion another great prophet is born in one of your cities, you can use him to enhance your religion.",
```
SomeTroglodyte, line 184: "If, after founding your religion," - commas or restructure entirely ... maybe "Once your Religion is founded, it can be enhanced by another GP" then maybe hint at "birthing" rate formula? Dunno.

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+    ],
+    "Beliefs": [
+        "There are four types of beliefs: Pantheon, Founder, Follower and Enhancer beliefs.",
+        "Pantheon and Follower beliefs apply to each city following your religion, while Founder and Enhancer beliefs only apply to the founder of a religion."
+    ],
+    "Spreading_Religion": [
+        "When founding a city, it won't follow a religion immediately.",
+        "The religion a city follows depends on the total pressure each religion has within the city.",
+        "Followers are allocated in the same proportions as these pressures, and these followers can be viewed in the city screen.",
+        "Based on this, you can get a feel for which religions have a lot of pressure build up in the city, and which have almost none.",
```
SomeTroglodyte, line 195: "build" - if that's meant as past tense, "built up". Present - "building up"  ... right?

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "The city follows a religion if a majority of its population follows the city, and will only than be effected by Follower and Pantheon beliefs of that religion.",
```
SomeTroglodyte, line 196: "follows the city"? "follows that religion" I'd guess. "only than" -> "only then"

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "",
+        "Spreading religion happens naturally, but can be sped up using missionaries or great prophets.",
+        "Missionaries can be bought in cities following a major religion, and take the religion of that city.",
+        "So do take care where you are buying them! If another civilization has converted one of your cities to their religion, missionaries bought their will follow their religion.",
+        "Great prophets always have your religion, regardless if they are bought or spawned naturally, but captured great prophets do retain their original religion.",
```
SomeTroglodyte, line 201: Somehow that regardless doesn't quite fit together ... "no matter whether they are bribed or hellspawn" perhaps? _Can_ one still capture Great Prophets?

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "Both great prophets and missionaries are able to spread religion to cities when they are inside their borders, even cities of other civilizations.",
+        "You can even enter tiles of civilizations with whom you don't have an open borders agreement!",
+        "But do take care, missionaries will lose 250 religious strength each turn they inside foreign lands.",
```
SomeTroglodyte, line 204: they... "remain" inside? stay? linger? dwell? procrastinate? tour? travel? run around?   ;)

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "This diminishes their effectiveness when spreading religion, and if their religious strength ever reaches 0, they have lost their faith and disappear.",
+        "When you do spread your religion, the religious strength of the unit is added as pressure for that religion.",
```
SomeTroglodyte, line 206: I always wondered: Does distance to next city factor in?

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "",
+        "Cities also passively add pressure of their majority religion to nearby cities.",
+        "Each city provides +6 pressure per turn to all cities within 10 tiles, though the exact amount of pressure depends on the game speed.",
+        "This pressure can also be seen in the city screen, and gives you an idea of how religions in your cities will evolve if you don't do anything.",
+        "Holy cities also provide +30 pressure of the religion founded there to themselves, making it very difficult to effectively convert a holy city.",
+        "Lastly, before founding a religion, new cities you settle will start with 200 pressure for your pantheon.",
+        "This way, all your cities will starting following your pantheon as long as you haven't founded a religion yet.",
+        "",
+        "Inquisitors are the last religious unit, and their strength is removing other religions.",
+        "They can remove all other religions from one of your own cities, removing any pressures build up and reclaiming the city for your religion.",
```
SomeTroglodyte, line 216: "built up". Reclaiming as in converts immediately probably only if your religion has enough pressure left?

android/assets/jsons/Tutorials.json
```diff
@@ -168,5 +168,53 @@
...
+        "Great prophets also have this ability, removing all other religions when spreading their religion.",
+        "Additionally, when an inquisitor is stationed in or directly next to a city center, units of other religions cannot spread their faith there."
```
SomeTroglodyte, line 218: I wonder - Only units? Or is nearby city pressure muted too? (I mean I could look up the code, but maybe other readers might also wonder at this point)

SomeTroglodyte (approved):

android/assets/jsons/Tutorials.json
xlenstra: I don't think it does in the original, and at least its not something I coded in for unciv

android/assets/jsons/Tutorials.json
```diff
@@ -192,16 +193,16 @@
...
-        "Missionaries can be bought in cities following a major religion, and take the religion of that city.",
-        "So do take care where you are buying them! If another civilization has converted one of your cities to their religion, missionaries bought their will follow their religion.",
-        "Great prophets always have your religion, regardless if they are bought or spawned naturally, but captured great prophets do retain their original religion.",
-        "Both great prophets and missionaries are able to spread religion to cities when they are inside their borders, even cities of other civilizations.",
+        "Missionaries can be bought in cities following a major religion, and will take the religion of that city.",
+        "So do take care where you are buying them! If another civilization has converted one of your cities to their religion, missionaries bought there will follow their religion.",
+        "Great prophets always have your religion when they appear, even if they are bought in cities follow other religions, but captured great prophets do retain their original religion.",
```
Mape6, line 202: 'in cities follow other religions' from my point of view, it should be 'in cities following other religions' or 'in cities that follows other religions'

SomeTroglodyte:
Yup, 'ing' is better. But - waaaaait a minute. All I did so far was nitpicking, and this deserves - a big praise. Good job. Seriously. Players without original Civ5 experience will need this, and the balance succinct/verbose is good, which is not easy. I'll have to see it in-game for a second proofread pass, after all, fixing english after the translators start on it is not optimal...

Mape6:
> 
> 
> Yup, 'ing' is better. But - waaaaait a minute. All I did so far was nitpicking, and this deserves - a big praise. Good job. Seriously. Players without original Civ5 experience will need this, and the balance succinct/verbose is good, which is not easy. I'll have to see it in-game for a second proofread pass, after all, fixing english after the translators start on it is not optimal...

You are totally right. @xlenstra Thank you very much for that hard work! We will appreciate that!

SomeTroglodyte:
Do we need to graft the Uniques system to Tutorials now, so we can hide some when the Religion master switch is off? :cold_sweat:

xlenstra:
Thanks for your kind words :)

SomeTroglodyte:
Now that I see it white on blue...
GP can "construct a holy city"? Meaning the improvement "Holy site", or did I miss something major..
GP keep being born... faith you have to save up will "be higher": hmm... 'increase with each one'? Or shock 'em with the exact formula - shall I make Civilopedia [TeX](https://en.wikipedia.org/wiki/TeX)-enabled?
Both GP ... spread religion to cities ... when inside its...: 'its' refers to what? 
You can even enter tiles of...: Maybe a smidgen clearer as "Your religious units can..."?
The next to last paragraph on GP removing other religions while spreading - still doesn't sound quite right. But I can't really suggest sth better. Period instead of comma and "they remove" instead of "removing", and "Therefore (thus, hence?) they almost always succeed in converting a city when they spread their religion"? Not entirely happy with that one either. OK, I'll leave it to you.
Last paragraph: "un**a**ffected"

Some of that really feels like full ICivilopediaText conversion and therefore linking and formatting abilities would be a big improvement - don't you think? I repeatedly put it off because we would have to completely reformat the json (can be done automated) _and_ split lines using `\n` (could be automated too) and so invalidate translations. Well, the `\n` thing _does_ have alternatives.

PR #5133, opened 2021-09-07
Implement limitations for pledge/revoke protection

SimonCeder:
As described in #4602
This PR adds a timer for 10 turns before you can revoke protection, and 20 turns before you can re-pledge after revoking.
The logic for when pledging/revoking is split out to a separate function so code isn't duplicated between logic/AI/diplomacy interface.


ravignir:
But can you remove these 9 lines (I guess) that make you insta declare war on the civ that attacked CS protected by you? XD

diplomacymanager.kt lines 550-559.

This is not in civ5 and it is super annoying.

PR #5134, opened 2021-09-07
CS vulnerable to ally unhappiness

SimonCeder:
As mentioned in #4602.
If the ally of a city state has negative unhappiness, the city state's units are affected as well.


xlenstra:
I was looking into this, and while this is indeed remarked on the wiki, I couldn't find any source for this in the source code. Did you find any code for it when implementing this?

SimonCeder:
I don't think so. Probably the wiki being wrong yet again. Relevant code seems to be `GetGenericMaxStrengthModifier` in CvUnit.cpp, there is a flat -33% penalty for very unhappy (<= -10 happiness) but only counting the happiness of the civ owning the unit.

PR #5136, opened 2021-09-08
Adding religous city states to unciv!

Interdice:
Hey guys. I am adding religious city states to unciv. For now this is a draft as I am not done yet.

EDIT: ready for merge

this relates to issue https://github.com/yairm210/Unciv/issues/4290


Interdice:
does this game have artifacts?

xlenstra:
I'd advise on using the friendbonus and allybonus other city states also use, they can be found in era's.json

xlenstra:
> does this game have artifacts?

What do you mean by artifacts?

SomeTroglodyte:
And I'd even get a bonus for _knowing_  such a city state when I'm at war with them...

This is expressly very early and unfinished, so I'll delay further critique. Welcome aboard!

ravignir:
not yet.

@xlenstra I guess he meant archaeology mechanic from bnw and the way antiquity sites work (created on city ruins/barb camps spots/tiles where battle was fought earlier in the game etc.). When you improve them with archaeologist you can dig up an artifact.

SomeTroglodyte:
Or maybe they meant build artifacts? As in those expressly excluded from Android Studio?

SimonCeder:
Thanks for the PR! As others said look at `getStatMapForNextTurn()` in `CivInfoStats.kt`. Since you've added the enum already, adding a couple lines to Eras.json should just work. You'll need to make sure religious city-states aren't generated in games with religion turned off, and also that religious city-states give some faith upon first contact.

Interdice:
https://civilization.fandom.com/wiki/Archaeology_(concept)_(Civ5)

Interdice:
@SomeTroglodyte Damn. Could have swore I put that if statement in

SimonCeder:
No, we are not currently doing any BNW stuff

Interdice:
@SimonCeder okay

Interdice:
@SimonCeder yeah will work on all that you stated

Interdice:
@SimonCeder dumb question but what will the statmap accomplish

xlenstra:
> @SimonCeder dumb question but what will the statmap accomplish

`getStatMapForNextTurn()` calculates the gold/science/culture/faith your civ will get at the end of your turn, split out into its contributing factors. The largest part of it is a for loop determining the stats gained from friendly/allied city states.

Interdice:
@SimonCeder ahh so you are saying refractor the code into that module got it

xlenstra:
> @SimonCeder ahh so you are saying refractor the code into that module got it

That shouldn't even be necessary, if you add the right lines to `eras.json` the code in `getStatMapForNextTurn()` should already automatically provide the correct bonuses.

SimonCeder:
Yes, you could put the code you wrote in the else clause together with the compatibility code for cultured CS, under the `// Deprecated, assume Civ V values for compatibility`. Otherwise the function should work as-is, you just need to add something like `"Religious": ["Provides [3] [Faith] per turn"],` to `friendBonus` and `allyBonus` in Eras.json

Interdice:
@SimonCeder okay

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -70,6 +72,7 @@ class ReligionManager {
 
     fun endTurn(faithFromNewTurn: Int) {
         storedFaith += faithFromNewTurn
+        storedFaith += getFaithFromCityStates()
```
SimonCeder, line 75: so now this is no longer necessary, since the values in Eras.son are already in the faithFromNewTurn :)

SomeTroglodyte:
> Damn. Could have swore I put that if statement in

:rofl:  I know the feeling, but then I'm already going 144... :birthday:

Interdice:
@SimonCeder Thanks I don't really know Json's this well.

Interdice:
@SomeTroglodyte lol

core/src/com/unciv/logic/civilization/ReligionManager.kt
Interdice: Is what i have done okay?

Interdice:
Stupid question but do i delete the function? I really don't know JSON :)

Interdice:
Do i have to write code to generate the new types of cities, because no religious cities are popping up

Interdice:
added faith on meet city states

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -401,9 +394,11 @@ class CivilizationInfo {
         val cityStateLocation = if (cities.isEmpty()) null else getCapital().location
 
         val giftAmount = Stats(gold = 15f)
+        val faithAmount = Stats(faith = 15f)
```
xlenstra, line 397: Why is this seperate from the giftAmount? They are used for the exact same thing and should also be doubled when meeting them for the first time. Adding them to the existing stats when the city state is religious should suffice

android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -1121,6 +1121,42 @@
 		"innerColor": [224,224,224],
 		"cities": ["Mogadishu"]
 	},
+	{
+		"name": "Bratislava",
+		"adjective": ["Bratislava"],
+		"cityStateType": "Religious",
```
xlenstra, line 1127: Preferably you make it so that religious city-states cannot spawn in games that have religion disabled

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
Interdice: It is separate because if i send a notification, it will print 15 gold

Interdice:
Can't filter the religious city states. When i try to filter out the religious city states i get the following error

kotlin.UninitializedPropertyAccessException: lateinit property ruleSet has not been initialized

Don't think there is any way around it.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
xlenstra: If you add the faith to it, it will print '15 gold, 15 faith'. It should also only be 4 faith iirc

android/assets/jsons/Civ V - Vanilla/Nations.json
Interdice: the problem is that the game does not know if the game has religion disabled when generating the map

Interdice:
sent a commit that shows the erroneous code

xlenstra:
> sent a commit that shows the erroneous code

I don't have time to do anything for the next ~12 hours, after that I'll see what I can do.

Interdice:
@xlenstra sure

android/assets/jsons/Civ V - Vanilla/Nations.json
yairm210: It does, this is all five in startNewGame

xlenstra: Now that religion is in beta anyways, its fine to remove all parts of `GameInfo.hasReligionEnabled` depending on the ruleset, everything should still work after that.

yairm210:
New city states should a. have unique inner colors and b. have unique quotes. The fact that many existing city states have the same quotes is a problem, not the correct state.

Interdice:
@yairm210 The inner colors are unique. also I don't know where to get the unique quotes. Do I just make them up?

Interdice:
Fixed all of the bugs

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -413,6 +408,12 @@ class CivilizationInfo {
         else
             otherCiv.addNotification(meetString, NotificationIcon.Gold)
 
+        if (otherCiv.isCityState() && otherCiv.cityStateType == CityStateType.Religious){
```
SimonCeder, line 411: otherCiv.canGiveStat(Stat.Faith) would be more robust

SimonCeder (changes requested):

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
Interdice: fixed

android/assets/jsons/Civ V - Vanilla/Nations.json
xlenstra: So you've solved it by excluding the names of these city-states? That doesn't work because of mods: if you create a mod where any of these is a cultural city state, it will never show up. Similarly, if you create another religious city-state, it will show up in games without religion. why not do `.filter { it.cityStateType == Religious }` instead?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
Interdice: fixed

android/assets/jsons/Civ V - Vanilla/Nations.json
Interdice: @xlenstra got a better implementation

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -197,9 +197,14 @@ object GameStarter {
...
         unusedMercantileResources.addAll(allMercantileResources.shuffled())
+        var numbersToTake = newGameParameters.numberOfCityStates
 
-        for (cityStateName in availableCityStatesNames.take(newGameParameters.numberOfCityStates)) {
+        for (cityStateName in availableCityStatesNames.take(numbersToTake)) {
             val civ = CivilizationInfo(cityStateName)
+            if (civ.cityStateType == CityStateType.Religious && !newGameParameters.religionEnabled){
+                numbersToTake++
+                continue
+            }
```
SimonCeder, line 207: Are you sure this works? Isn't the list for the for loop generated when entering the for loop, ie incrementing numbersToTake halfway through will have no effect?
It might be better to exclude the religious city states when first generating the `availableCityStatesNames`

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
xlenstra: This hasn't been adressed yet.

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -197,9 +197,14 @@ object GameStarter {
         val allMercantileResources = ruleset.tileResources.values.filter { it.unique == "Can only be created by Mercantile City-States" }.map { it.name }
         val unusedMercantileResources = Stack<String>()
         unusedMercantileResources.addAll(allMercantileResources.shuffled())
+        var numbersToTake = newGameParameters.numberOfCityStates
 
-        for (cityStateName in availableCityStatesNames.take(newGameParameters.numberOfCityStates)) {
+        for (cityStateName in availableCityStatesNames.take(numbersToTake)) {
             val civ = CivilizationInfo(cityStateName)
+            if (civ.cityStateType == CityStateType.Religious && !newGameParameters.religionEnabled){
```
xlenstra, line 204: I'd still advise on using `gameInfo.hasReligionEnabled()` instead of `!newGameParameters.religionEnabled` as it prevents code duplication, and also provides correct behaviour when starting in the industrial era or later.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -408,6 +410,12 @@ class CivilizationInfo {
         else
             otherCiv.addNotification(meetString, NotificationIcon.Gold)
 
+        if (otherCiv.isCityState() && otherCiv.canGiveStat(Stat.Faith)){
+            otherCiv.addNotification(religionMeetString, NotificationIcon.Faith)
```
xlenstra, line 414: I don't understand the need for the notifications to be split up. Why shouldn't they just be combined?

xlenstra:
> @yairm210 The inner colors are unique. also I don't know where to get the unique quotes. Do I just make them up?

Either I'm blind, or all inner colours are [255, 255, 255] and therefore not unique. On the unique quotes, be creative!

core/src/com/unciv/logic/GameStarter.kt
Interdice: if i use the gameInfo class, I will get a horde of uninitialized variable errors

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
Interdice: 1. the notification will be too long
2. one is a gold notification and one is a faith notification

Interdice:
@xlenstra ohh I thought faith civs all had to have 255 255 255. :)

core/src/com/unciv/logic/GameStarter.kt
xlenstra: I thought you could work around them, but looking closer that does seem impossible. I'll check out later today why the transients of the gameInfo aren't already set by this point, as especially the ruleSet should already be known at this point.

core/src/com/unciv/logic/GameStarter.kt
Interdice: fixed

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
xlenstra: 1. We have long notifications so that doesn't really matter imo
2. It could just be one stats notification with stats given as a token of goodwil? You can always add multiple icons if that is what you mean.

core/src/com/unciv/logic/GameStarter.kt
xlenstra: Alright, it's fine to add the line:
```gameInfo.ruleSet = ruleset // need to set this transient so that we can get the starting era```
Just before calling `addCivilizations` in `startNewGame` (line 60 or so), and that should fix all the `unitialized variable` errors

Comment (Interdice) [reply core/src/com/unciv/logic/GameStarter.kt]: fixed

**Comment by Interdice** [comment]
Why is this giving me an error? when i run it on my laptop is works perfectly

**Comment by SomeTroglodyte** [comment]
> Why is this giving me an error? when i run it on my laptop is works perfectly

"perfectly" in IDE and failing the github action is normal, as more build steps are run, among them the unit tests. Usually means you just don't see the "absolutely not perfectly" aspect by not testing enough. "Why" - go look and read the output. Threre's a "Details" link, you browse upwards, there's something like 2 of 60 unit tests failing, scroll some more, you'll see which and maybe some detail.

In this case, `serializedLaziesTest FAILED` makes the other test irrelevant. Sounds like You've added a `by lazy` into some class that gets serialized with a save game. You'll probably not really "work perfectly", just try saving a game. Background: Our serializer is written for Java and cannot cope with kotlin `by lazy`, so you have to help and tell it to ignore such a field with `@delegate:Transient`.

If you wish, you could try to learn how to run the same level of checks locally - see my scratchpad repo.

**Comment by Interdice** [comment]
@SomeTroglodyte I did not use by lazy though. I think is must be something to do with the gameinfo class

File: core/src/com/unciv/logic/GameStarter.kt (hunk 6 above)
Comment (Interdice) [reply core/src/com/unciv/logic/GameStarter.kt]: @xlenstra some other error is coming in the unit test

**Comment by Interdice** [comment]
There is no good way to find if city states are religious because
 "lateinit property nation has not been initialized"
I think Yarim should make religious city states standard in the game somehow

**Comment by Interdice** [comment]
@logicminimal 

There are 3 ways to determine if a city state is religous

1. Determine if the name of the city state is religious. This is bad because if modders create their own religious city states, the game wouldn't know

2. Determine if the type of city state is religious by looking at its class variables. Nations unfortunately gets initialised after the city state is set in stone and can't removed and replaced if religiousl

3. I guess i could use the JSON. But that is kind of a pain and should only be used if no other solution is found

File: core/src/com/unciv/logic/GameStarter.kt [hunk 8]
```diff
@@ -49,6 +49,7 @@ object GameStarter {
         runAndMeasure("addCivilizations") {
             gameInfo.tileMap = tileMap
             tileMap.gameInfo = gameInfo // need to set this transient before placing units in the map
+            //gameInfo.ruleSet = ruleset doesn't work for some reason
```
SomeTroglodyte, line 52: The reason is: this is early on in the construction of a card house on thin air - some things go first others come later, and in the middle you can't rely on anything existing unless you control the workflow such that it does. Anyway, you do not need to interfere here.

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -188,11 +189,22 @@ object GameStarter {
 
         val civNamesWithStartingLocations = gameInfo.tileMap.startingLocationsByNation.keys
 
-        val availableCityStatesNames = Stack<String>()
```
SomeTroglodyte, line 191: All the confusion "just" to exclude the new CS when Religion is off... OK. Actually, looks like an almost valid approach, though it can look much simpler. Revert all that and insert a filter in the chain feeding `availableCityStatesNames.addAll`: "get all ruleset nations... limit to city states ... shuffle" can do what you need right in the "limit to" part.
```kotlin
        availableCityStatesNames.addAll(
            ruleset.nations
            .filter {
                it.value.isCityState() &&
                (it.value.cityStateType != CityStateType.Religious || newGameParameters.religionEnabled)
            }
            .keys
            .shuffled()
            .sortedByDescending { it in civNamesWithStartingLocations }
        )
```
More readable like this perhaps.

Some of your misunderstandings here:
- `val` versus `var` for a collection: No need to var that Stack - val makes the container object immutable, not its contents - that depends on the collection type. IOW, you don't want to change the toybox, you only want to change the toys in it - so the box itself can be immutable. This distinction can be hard to grasp - it's a mindset.
- Nation vs CivilizationInfo: Nation is a template, CivilizationInfo a player, a participant. This code is about building the latter using the former. The property cityStateType - yes CivilizationInfo has it but it gets it from the Nation, so while in the process of creating the requested players, such filters should go by the original.
- for..remove: Look up `ConcurrentModificationException`. A loop uses an `Iterator` - an object responsible for "where am I, what's next and when am I done". The job of an iterator becomes harder when you modify the unterlying collection while you iterate - modifications might remove what the iterator expects to be next, or similar issues. So most of them explicitly guard against modification of "their" collection. Exceptions are very specific and need a very constrained usage, so yes I don't know how, never learned those. To keep it simple, never modify a collection you or someone else may be iteration at the same time. That's why you might discover loops iterating over List.toList() - they iterate over an independent clone.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SomeTroglodyte: I understand both views. This actually depends on how exactly Civ5 ticks for RCS, which I don't know... e.g. Gold is currently doubled for being the first major to meet a CS, Faith is not..
- Notification length is not an issue - it very much so but there's worse places already, so we need auto-wrapping ones anyway in some future. Simplicity trumps here.
- Displayed icons are vararg, you can have as many as you want. Doing so dynamically is a bit more than 10 chars code, though. Still, beauty would trump that consideration. _If_ doubling faith for a first-meet is OK this _could_ do it:
    ```kotlin
        val giftAmount = Stats(gold = 15f)
        if (cityStateType == CityStateType.Religious) giftAmount.faith = 4f
        var meetString = "[${civName}] has given us [${giftAmount}] as a token of goodwill for meeting us"
        if (diplomacy.filter { it.value.otherCiv().isMajorCiv() }.count() == 1) {
            giftAmount.timesInPlace(2f)
            meetString = "[${civName}] has given us [${giftAmount}] as we are the first major civ to meet them"
        }
        val icons = giftAmount.map { it.key.notificationIcon }.toTypedArray()
        if (cityStateLocation != null)
            otherCiv.addNotification(meetString, cityStateLocation, *icons)
        else
            otherCiv.addNotification(meetString, *icons)
        for ((key, value) in giftAmount)
            otherCiv.addStat(key, value.toInt())
    ```
... and yes, if you grok how and why that val icons works - congrats, level-up!
- `if (otherCiv.isCityState() && otherCiv.canGiveStat(Stat.Faith))` - well, at this point `otherCiv.isMajorCiv()` is guaranteed true...

SimonCeder:
@Interdice 
I've laid the groundwork for a simple way to exclude religious city-states in `CityStateFunctions.kt`, there's a TODO in `initCityState`. Just make a check for religious city states there, look at the check for militaristic above for reference. You might end up with something like
```
if (allPossibleBonuses.any { it.getPlaceholderText() == "Provides [] [] per turn" && it.getPlaceholderParameters()[1] == Stat.Faith.name}
            || (allPossibleBonuses.isEmpty() && cityStateType == CityStateType.Religious)) {
```
and then, if religion is disabled, `return false` and you're all set - no modifications to GameStarter.kt needed.

core/src/com/unciv/logic/GameStarter.kt
Interdice: Thanks for the help. I actually changed the val to a var in an attempt to determine if a city state is religious. I understand the difference between them.

I probably should learn more about concurrentmodificationexecptions

Thanks for the implementation

core/src/com/unciv/logic/GameStarter.kt
Interdice: thanks

core/src/com/unciv/logic/GameStarter.kt
Interdice: fixed

Interdice:
fixed

android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -1121,6 +1121,42 @@
...
+		"cities": ["Cahokia"]
+	},
+	{
+		"name": "Jerusalem",
+		"adjective": ["Jerusalem"],
+		"cityStateType": "Religious",
+
+		"declaringWar": "By god's grace we will not allow these atrocities to occur any longer. We declare war!",
+		"attacked": "May god have mercy on your evil soul.",
+		"defeated": "I for one welcome our new conquer overlord!",
```
yairm210, line 1155: Cute :)

yairm210:
Okay, I think this is finally ready!
Be aware that one this is released to users, there may be more things to solve, but that's just the way it is :)

PR #5140, opened 2021-09-08
Fixed all renamed units being called 'instanceName'

### LINKED ISSUE #5139: Unit's name becomes "instanceName" when unit has been renamed to anything
**Platform**
- Windows 10 Build 2004 (64-bit)
- Android 8.1.0 (64-bit ~~architecture unsure~~)

**Version**
- 3.16.12, from this GitHub repository on PC and from Google Play Store on Android

**Describe the bug**
- Renaming any unit that can be renamed doesn't really rename it -- rather, it is renamed to "instanceName" regardless of what was inputted. (As an example, renaming a Warrior to "test" *should* rename it to "test", but instead it is renamed to "instanceName")

**To Reproduce**
1. Start any game
2. Select any unit that can be promoted
3. Attempt to rename selected unit to any name except "instanceName"
4. Bug happens

- None; this happened on both devices in both generated and custom maps

**Expected behavior**
- It *should not* be "instanceName" unless if it was intended to be renamed as "instanceName"

**Screenshots**

**Additional context**

xlenstra:
Fixes #5139

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -111,7 +111,7 @@ class MapUnit {
     fun displayName(): String {
         val baseName =
             if (instanceName == null) "[$name]"
-            else "instanceName ([$name])"
+            else "$instanceName ([$name])"
 
         return if (religion == null) baseName
         else "$baseName ([${getReligionDisplayName()}])"
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -116,6 +116,11 @@ class MapUnit {
         return if (religion == null) baseName
         else "$baseName ([${getReligionDisplayName()}])"
     }
+    
+    fun shortDisplayName(): String {
+        return if (instanceName != null) "[$instanceName]"!!
+            else "[$name]" 
+    }
 
     var currentMovement: Float = 0f
     var health: Int = 100
--- a/core/src/com/unciv/logic/automation/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/UnitAutomation.kt
@@ -495,7 +495,7 @@ object UnitAutomation {
         if (unit.isDestroyed) return // Opening ruins _might_ have upgraded us to another unit
         if (unit.health < 80 && tryHealUnit(unit)) return
         if (tryExplore(unit)) return
-        unit.civInfo.addNotification("${unit.displayName()} finished exploring.", unit.currentTile.position, unit.name, "OtherIcons/Sleep")
+        unit.civInfo.addNotification("${unit.shortDisplayName()} finished exploring.", unit.currentTile.position, unit.name, "OtherIcons/Sleep")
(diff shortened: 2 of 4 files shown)


SomeTroglodyte (approved):
🏎️

xlenstra:
> racing_car

Gotta get third most commits somehow :)

SomeTroglodyte:
:3rd_place_medal:  well, if you can get github to book them over, you can have some of mine.

But - how can the i10l work better without the `[]`s? They can't match e.g. "[unit] finished exploring. = [unit] is gestopt met ontdekken." any longer?

xlenstra:
> 3rd_place_medal well, if you can get github to book them over, you can have some of mine.
> 
> But - how can the i10l work better without the `[]`s? They can't match e.g. "[unit] finished exploring. = [unit] is gestopt met ontdekken." any longer?

The problem is that the `displayName` can also contain square brackets of their own, for example for religion or the name of the unit. This mean that previously, notifications such as "[[musketman]] has stopped exploring" would be added, and after translating '[musketman] is gestopt met ontdekken" still contained square brackets. My second commit fixes this, removing the extraenous square brackets. There does however remain a problem, as now "$instanceName ([$name]) has stopped exploring" is untranslatable. I'll add some extra brackets and translations to solve this.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -111,11 +111,16 @@ class MapUnit {
...
             if (instanceName == null) "[$name]"
-            else "instanceName ([$name])"
+            else "$instanceName ([$name])"
 
         return if (religion == null) baseName
         else "$baseName ([${getReligionDisplayName()}])"
     }
+    
+    fun shortDisplayName(): String {
+        return if (instanceName != null) "[$instanceName]"!!
```
SomeTroglodyte, line 121: This is definitely a `UNNECESSARY_NOT_NULL_ASSERTION` - a string literal, even with templates, is never null. Any template placeholders evaluating to null will be replaced with the letters "null", so ensuring non-null on them is always optional, too. Just FYI.

SomeTroglodyte:
This still doesn't taste right. Shifting the responsibility for translation markers to a subroutine is less readable... Maybe we should have forbidden `{}[]` in displaynames, there's a ~~parliament of rooks~~ sorry ~~murder of crows~~ sorry! ~~party of ravers~~ sorry!! - _many_ typographical brackets out there we could have used for visuals instead... `❲❳⟦⟧⟮⟯❴❵❬❭⦗⦘〈〉《》【】〔〕﹛﹜﹝﹞［］｛｝` just a few of the many, without leaving UCS-16 single-codepoint, and none would trip the translator code. One or two would surely render on Linux, Mac and Lollipop...

But hey, not that important. As long as that dollah is back..

core/src/com/unciv/logic/map/MapUnit.kt
xlenstra: I have no idea how that even ended up there, I fully agree that it looks completely superfluous

PR #5141, opened 2021-09-08
Add a "Reset to defaults" button to the Newgame screen

SomeTroglodyte:
@ravignir inspired, not sure there's an issue to link. It's in the `rightSideGroup` which seems to me existed for that purpose.. The saved settings are not forgotten until you actually start a game with the freshly reset defaults, which is cool.

Conflicts with my earlier PR touching PickerScreen expected. Or maybe not - I made sure that visibility change is the same in both.


logicminimal:
can u also pls add a randomise seed button?

PR #5142, opened 2021-09-08
Adds support for opening debug menu on MacOS

Alex25820:
This PR adds support for opening the debug menu added in #5081 on MacOS.

The reason it wasn't possible to open this menu before on MacOS was because you needed to hold down `Shift Right` and `Control Right` whilst left clicking the options button on the main menu screen. When holding down `Control` on MacOS and left clicking it automaticity converts the click to a right click instead. (Known as a Control Click on MacOS.) This means the menu will not open since you now `Shift Right` + `Right Clicked` instead.

_As far as I could find there is no way to disable Control Click on MacOS without having to download external programs._

I choose to add `Alt Right` as I thought it kept with the theme of buttons on the right side of the keyboard. I wasn't sure if you wanted to keep the `Control Right` option too so I decided to make it possible to use both `Control Right `and `Alt Right`.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
@@ -78,7 +78,7 @@ class OptionsPopup(val previousScreen: CameraStageBaseScreen) : Popup(previousSc
                 if (modCheckFirstRun) runModChecker()
             }
         }
-        if (Gdx.input.isKeyPressed(Input.Keys.SHIFT_RIGHT) && Gdx.input.isKeyPressed(Input.Keys.CONTROL_RIGHT)) {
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
All that is not to be taken too seriously. I don't _click_ if I want that page, I use only keys - would those work the same on a mac? What _does_ ctrl-shift-O do on the [Macintosh](https://github.com/yairm210/Unciv/pull/5142)?

Alex25820:
Didn't think about that. I tried but it doesn't work either. Most Macs only have a Left Control key and it doesn't work. You would think that on a keyboard with only one Control key it would act as both Left and Right Control but that isn't the case for some reason.

SomeTroglodyte:
The 'some reason' is something like Gdx hat to make choices in cross-platform abstraction. The Gdx keyboard mapping stuff re-invents the wheel in parts to go cross-platform, and some of it is surprisingly wrong. But mapping one physical key to two keycodes would be even wrongier.

So - here's my official apology having completely forgotten about good old 68k CPU's!

I guess your approach is fine. My dice fell on right ctrl-shift _only_ because you can produce a Ctrl-O with them without it being easy to hit accidentally.

PR #5146, opened 2021-09-08
delete the annoying part of misimplemeted pledge to protect mechanic.

ravignir:
Maybe poll this on discord.


SimonCeder:
Don't know that a poll is needed, as you stated this is misimplemented. The correct implementation is to lower relationship values for many turns with the protectors, deets in CvDiplomacyAI.cpp, starting on line 20613 (wow)

yairm210:
Yeah sure
That was probably me, this is how I remembered the pledge to protect to work

PR #5147, opened 2021-09-08
Militaristic city states can give unique unit; refactor city state initialization

SimonCeder:
As described in #4602.
Militaristic city states get a unique unit chosen from among the major civs that are _not_ in the game, that they can provide if the ally has the correct tech. If able they will always give the unique unit instead of a random unit.

Also refactored the city state initializing code into CityStateFunctions. Right now the main benefit is a more robust init of mercantile city-states, but a way of eg restricting the spawning of religious city states does suggest itself.

Screenshot_2021-09-08_21-47-41
Militaristic city-states get an extra line in the diplomacy view explaining this. I know people have suggested that this screen is a bit cluttered as is, can we put this anywhere else? Bake it into the Friend/Ally bonus lines? In Civ V this information is displayed when mousing over the "Militaristic" in the city state screen.


ravignir:
Isn't the timer of unit gifts from militaristic city states random in civ5?

SimonCeder:
Hmm, you're right. We'd have to rewrite the bonus strings. "Provides military units", "Provides military units slightly more often"?

SomeTroglodyte:
> will always give the unique unit

I'm surprised, I'd have thought some randomness left. But you know C5 better than me.

Regarding cluttering of the screen -> reduce padding to _group_ lines and you both save space and improve visual - what's the name? I mean people will parse the info they're interested in easier. Type & Personality closer, group Ally/relationship/changes closer, and group bonusi closer perhaps? I'd suggest a small v-padding of 2.5f and whatever it is now as inter-group pad.

SimonCeder:
Screenshot_2021-09-08_23-50-07
Something like this?

The civ 5 mechanic is that you get units every 20 or 17 turns, +/-1 turn. I'm a bit torn between rewriting the strings to be vague or just implement it without changing the strings and having them be slightly misleading.

SomeTroglodyte:
Font size for 'able to' line and 'When RelationshipLevel' are the same? Then yes.

...`±1` or `≈` somewhere?

yairm210:
Does this mean that e.g. Dublin still continue providing ballista units into the modern era?
Because that seems unhelpful

SimonCeder:
> Font size for 'able to' line and 'When RelationshipLevel' are the same? Then yes.
> 
> ...`±1` or `≈` somewhere?
They do seem to end up slightly different sizes, but they're both just `.toLabel()`. I'll try specifying explicit font size.
`≈` seems good.

> Does this mean that e.g. Dublin still continue providing ballista units into the modern era?
> Because that seems unhelpful
No, they'll resume providing random units as soon as the unique unit is obsoleted, so as soon as Physics is researched in this case. I'll add a couple comments to make the mechanic more clear.

SimonCeder:
New and improved:
Screenshot_2021-09-09_13-02-51

SomeTroglodyte:
I know I'm late with this, I didn't pay enough attention.
- `initCityState` effectively deprecates Nation.cityStateType by recalculating them from the Eras - shouldn't we write the overridden type back into the CivilizationInfo?
- `initCityState` essentially does a `setNationTransient` temporarily - why not call it?
- `CityStateFunctions` needing all those wrappers - ugly. If this will never need its own fields (likely), a bunch of Extension functions with CivilizationInfo receivers living in an object would have been a bit more efficient (on bytecode level, too), at the expense of more imports.

Just food for thought and for much later.

SomeTroglodyte:
OK-ay... :+1: 
Nah, moving fields means different serialization + consequences. Skipping the wrapper "every time" would then be close to the other "managers"... Anyway, as long as the decision has considered all viable alternatives it's fine.

Sigh... <span title="C# has those"><code>partial class</code></span>...

logicminimal:
it isnt limited to one unique unit right?

SimonCeder:
Each city state gets one unique unit. It's randomly determined at game start.

logicminimal:
> Each city state gets one unique unit. It's randomly determined at game start.

oh, i thought it was more than one. also while researching this i found this paragraph of text:

"Friends to a Militaristic City-State will get a Free Unit on average, every 20 Turns. Allies get them every 17 Turns. These Units are typically modern, and may come from any Civilization - so you could end up with some Unique Units from other Civilizations. **_If your Military is getting too big and the maintenance costs too high, you can talk to a Militaristic City-State and ask them to stop sending you Units until you allow them to resume_** "

should i add this to the city states thread or is it a really minor side effect of militaristic city states?

SimonCeder:
Add it, but I think it's low priority.

ravignir:
ultra low priority. i don't even see how it is useful given you can disband these gifted units or send them back to city state for influence.

PR #5149, opened 2021-09-08
Religion goes from alpha into beta: available from start screen immediately

xlenstra:
As almost all the most important parts of religion are implemented at this point, I feel like it is time to open it up to a wider audience.
Some features are still lacking, (most of which can be found in #4290) making this the main reason the option is disabled by default.
After merging this PR, updates to religion should also not invalidate existing saves, and I'll take care to do that.

Starting from this PR, more intrusive features can be added, such as changing existing natural wonders to provide faith, or redesigning the piety tree to provide more faith-related bonuses, even in games with religion disabled.


SomeTroglodyte:
> all the most important parts of religion are implemented

Hey! Enhancer beliefs are missing an icon in 'pedia! :smiling_imp:  ([this one perhaps](https://thenounproject.com/term/pill/199308)?)

PR #5156, opened 2021-09-09
Fix Denunceation typo, reorg compatibility code

SomeTroglodyte:
... and bring warning count down another two. See previous chat in #5121.

- Extensive Diff for GameInfo is due to my new rule "If I see a pure functions region and a pure function _outside_ that region, and I feel like it, redo region use thoroughly, and mark this  as done by capitalizing region names.". This may be the second class so far this applied to...
- I used 'Deprecated after' instead of 'with' as I suspect a new version number rollover might be close. Feel free to tell meee otherwise.
- That unused `matchesCityStateRelationshipFilter` function - is such a todo good enough to suppress the public warning? Or look at blame and ask author what's up?

Remaining non-deprecation 'public' (as in, viewable without having a compiler) warnings after this:
```
w: Unciv/core/src/com/unciv/logic/battle/BattleDamage.kt: (285, 9): Parameter 'tileToAttackFrom' is never used
w: Unciv/core/src/com/unciv/logic/city/CityInfoConquestFunctions.kt: (83, 13): Variable 'goldPlundered' is never used
w: Unciv/core/src/com/unciv/models/ruleset/Building.kt: (73, 29): Parameter 'ruleset' is never used
w: Unciv/core/src/com/unciv/models/ruleset/Building.kt: (117, 45): Parameter 'ruleset' is never used
w: Unciv/core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt: (238, 85): Parameter 'worldScreen' is never used
```
Of these, I think the last three can be safely removed, but the first two might have some ulterior motive...


core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -341,7 +342,7 @@ object NextTurnAutomation {
     private fun chooseBeliefs(civInfo: CivilizationInfo, beliefContainer: BeliefContainer): HashSet<Belief> {
         val chosenBeliefs = hashSetOf<Belief>()
         // The 'continues' should never be reached, but just in case I'd rather have AI have a
-        // belief less than make the game crash. The 'continue's should only be reached whenever
+        // belief less than make the game crash. The 'continues' should only be reached whenever
```
yairm210, line 345: I actually disagree. One "continue", many "continue"s.

SomeTroglodyte: I just adapted the combo to the one on the line above, so Studio would shut up. That was a waait - do I need to think about this? Nah, don't wanna - look there's a precedent so we don't have to think -> all in ~600ms. With 3000ms,  I'd tend to an approach like  "let's pretend markup works here even if it doesn't" and use ` to demark the keyword and 's do make plurals...    So many words for a measly comment...   Mebbeth I should turn off spellchecking inspection globally instead.

xlenstra:
>     * That unused `matchesCityStateRelationshipFilter` function - is such a todo good enough to suppress the public warning? Or look at blame and ask author what's up?

Did someone call me?
The original idea with that filter was to parameterize relationships with city-states in cases where they appeared in uniques, but just before merging the PR I decided to remove them from that unique. Probably happened in the implementation of the patronage policy tree iirc. I decided to leave this function in, maybe for future use in parameterized uniques, but nothing ever happened with it. It is now on my list of filters to start using for the unique overhaul I've been working on the past few days, so it doesn't have to be removed just yet. If in the end I decide not to use it, I'll remove it myself at that time.

SomeTroglodyte:
...then `@Suppress("unused")  //todo` is perfectly adequate :+1:

SomeTroglodyte:
Though...
```kotlin
    fun matchesCityStateRelationshipFilter(filter: String) = relationshipLevel().let {
        filter == it.name || filter == it.synonym
    }
```
..and add synonym: String? to the enum, init with Allied for Ally, same for F(r)iendly and Unforgiv(ing/able).

PR #5160, opened 2021-09-09
Icons for city states

SimonCeder:
This PR adds icons to the city buttons for city-states so it's easier to tell their type at a glance.
Screenshot_2021-09-09_13-49-18


SomeTroglodyte:
Is this still pretty when they have a religion?

SimonCeder:
Reasonably:
Screenshot_2021-09-09_15-06-26

core/src/com/unciv/ui/tilegroups/CityButton.kt
```diff
@@ -237,6 +239,17 @@ class CityButton(val city: CityInfo, private val tileGroup: WorldTileGroup): Tab
         iconTable.addActor(cityStrengthLabel)        // We create this here to we can .toBack() it as well.
         cityStrengthLabel.toBack()
 
+        if (city.civInfo.isCityState()) {
+            val cityStateImage = when (city.civInfo.cityStateType) {
```
SomeTroglodyte, line 243: `val cityStateImage = ImageGetter.getImage("OtherIcons/${city.civInfo.cityStateType.name}").apply { color = secondaryColor }` would hold so far... And reusing a texture region under a different name is done by simply giving the packer a bit-identical copy. In case you would like that better.

SimonCeder: Yep, that's better, now it's future-proof for Religious city-states.

SomeTroglodyte:
> Reasonably

Do a padding inventory. From here I don't see the cell defaults, but yours is a right(10) while religion is left(5) so it's to be expected that can look inconsistent. Hmmm. I always wanted to try the usefulness of `Stage.setDebugUnderMouse` and `Stage.setDebugParentUnderMouse`...

SimonCeder:
Does this seem better?
Screenshot_2021-09-09_15-32-18

yairm210:
If we're adding an icon for each city state type, please add it to the enum as well (Troglodyte wins this round) so we'll be able to use them in the introduction and diplomacy as well :)

SimonCeder:
The `CityStateTypes` enum?

yairm210:
Yup

yairm210:
See example in Stat enum

SimonCeder:
Done. I also added icons to the diplomacy sidebar:
Screenshot_2021-09-09_19-57-22

yairm210:
Nice! 👍🏿

SomeTroglodyte:
Niiiiice, yes. But as devil's advocate I can say - it's visible that one of these is 35f and the other 30f.  It's not an optical illusion like I thought at first. Also - for the positioning it should be possible to let Gdx do the math, and save one under-the-hood notification. E.g. `questIcon.setPosition(civIndicator.width, 0f, Align.bottomRight)` works same as the existing `x=`, and yours should work fine with `typeIcon.setPosition(civIndicator.width, civIndicator.height, Align.topRight)`

PR #5165, opened 2021-09-09
Proper implementation of pledge to protect

SimonCeder:
This PR implements proper diplomacy penalties for aggressions against protected city-states.

~~HOWEVER, as you can see this is very complicated, includes a bunch of auxiliary timers and flags, and could alternatively be implemented _almost_ as well with the existing `revertToZero` function. Do we want the elegant and simple way or the technically accurate way?~~ Actually this is OK

Screenshot_2021-09-09_21-45-45


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -495,6 +495,8 @@ object Battle {
 
     fun destroyIfDefeated(attackedCiv: CivilizationInfo, attacker: CivilizationInfo) {
         if (attackedCiv.isDefeated()) {
+            if(attackedCiv.isCityState())
```
SomeTroglodyte, line 498: https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace

SomeTroglodyte:
With all this sets out to do I don't see how it could be much shorter.
Define "almost"? Refers to the flags counting down differently?

SimonCeder:
I mean we could dispense with `RememberBulliedProtectedMinor` and so on and just count the penalty down each turn like we do with all other penalties.

ravignir:
oz

You either take diplomatic penalty with a major civ or take -20 influence hit with city state, you stop protecting it and you can't protect it for next 20 turns. This only happens when they demand tribute.

When city state is attacked, city state will ask you for help. If you kill 3 units of a major civ that attacked them, you will gain 100 influence.
z

You can't pledge to protect CS if you have less than 0 influence with them. You can't revoke protection until 10 turns after you pledge.

SimonCeder:
> You either take diplomatic penalty with a major civ or take -20 influence hit with city state, you stop protecting it and you can't protect it for next 20 turns. This only happens when they demand tribute.

I'll implement this.

> When city state is attacked, city state will ask you for help. If you kill 3 units of a major civ that attacked them, you will gain 100 influence.

This is a quest, not related to pledge to protect. I'll have a look at city-state quests sometime in the future. Meanwhile could you add this to #4602 ?

> You can't pledge to protect CS if you have less than 0 influence with them. You can't revoke protection until 10 turns after you pledge.

Already implemented in #5133

yairm210:
I'm okay with the current implementation. Are the turns game-speed-agnostic?

SimonCeder:
In the civ V source there's no correction for game speed, however there's a comment about adding it sometime in the future. I've preserved this, but we could just as easily put in multipliers. Do we want to?

yairm210:
Nah, I think this is fine
Diplomacy being turn-dependent actually fit in with "player perception" of time as turns rather than time as "ingame years"

SimonCeder:
> You either take diplomatic penalty with a major civ or take -20 influence hit with city state, you stop protecting it and you can't protect it for next 20 turns. This only happens when they demand tribute.

Implemented:
Screenshot_2021-09-10_22-27-15

Also increased the resting point bonus for protected civs from 5 to 10.

logicminimal:
sorry for late response, does this implement this?

- [ ] When a City-State becomes the victim of bullying (another civ demands tribute from them/makes them afraid), they call for defense in one of the following manners, all of which only apply to you, and are valid for 30 turns:

   +Pledge protection to them. In this case only, the action will give you some Influence Influence.
   +Denounce the civilization that bullied them.
   +Donate money to help them recover. Unlike the "Public Works" quest, this one is a single-use, giving you +20  Influence Influence when you make the gift in addition to the regular influence effect. Also, the City-State becomes more resistant to tribute in the future from other players
    +When city state is attacked, city state will ask you for help. If you kill 3 units of a major civ that attacked them, you will gain 100 influence.

SimonCeder:
No, those will be implemented in a future PR dealing with quests.

Issue #5091, opened 2021-09-04
Uniques - some problems and possible solutions - input requested


xlenstra:
While the current system with uniques, placeholders and filters allows a lot of freedom for modders, in some areas it still has some problems. This issue will first give an overview of the most major problems with uniques, for modders, users and developers, and then proposes a vision of uniques where these problems will be solved.

Coders, what do you think? Mod makers, would this change be a major improvement to moddability? Do you expect to use this heavily if it were added?

### Problems with uniques

So, first, what problems are there with the current uniques?
- Combining filters is not possible. While this has been tried for `unitFilter`, by allowing `[{military} {water}]` as a unit filter, it was quickly discovered that the space in between the partial filters is depended on language. Thus, this could not become a general solution.
- A lot of uniques are very similar both in wording and effect, but every single unique needs to be checked independently. This was for example noticed for battle uniques in #4777, were @avdstaaij proposed a combined naming scheme for these.
- Uniques are strings. While this is great for storing in `.json` files and for modders, doing continuous string comparisons all over the code base will eventually add up to performance loss. Additionally, typos and miscapitalization might go unnoticed, possibly introducing hard to find bugs where certain uniques do not seem to work for no reason whatsoever. This point has been discussed before in a comment in #4890.
- Detecting the source of a unique is difficult. There are multiple places where we would like the source of a unique, for example when determining stat bonuses in the city screen, or combat bonuses for a unit. An example of this would be the `Haka War Dance` unique, which provides `-10% Strength to adjacent enemy units`. In the olden days where uniques had no placeholders, we could just use this unique and call the modifier 'haka war dance', which was descriptive. However, because in mods an entirely different source might provide this unique, this name would only create confusion. It was thus replaced with the less descriptive "Adjacent enemy units", which, although correct, is not clear to a user. Why only these adjacent enemy units? Why only here?
- Continuing the last point for determining where the stats from cities come from: in `CityStats.kt` there are a lot of functions whose only purpose is to check for a specific set of uniques of a specific object for matching uniques, and only add those stats to a `StatList` with a given tag. This means, that if a new source of stats is added, you have to test for this type in multiple functions and somehow get access to all uniques from that source. And some of these sources are combined for no reason whatsoever, such as `getGrowthBonusFromPoliciesAndWonders()`. It would seem all of this could be much simplified if uniques somehow were able to keep track of their source themselves.

### Proposed solutions

[...]

PR #5166, opened 2021-09-09
Very basic implementation of conditionals for uniques

xlenstra:
This PR implements one of the ideas brought up in #5091, optional conditionals for uniques.
It implements two of these, `<when at war>` and `<when not at war>`, and lays most of the groundwork for future implementation of other conditionals.
It implements the checking of these conditionals by adding `getMatchingApplyingUniques()` and `hasApplyingUnique()` functions, which function similar to `getMatchingUniques()` and `hasUnique()`, except that they also filter out uniques that don't apply in this situation. Currently, this can be determined only by passing the correct `civInfo` to these functions, but later on these functions will have a myriad of optional parameters that may be necessary to check in conditionals.

Although the amount of changes in this PR may look daunting, most of them are either replacing `getMatchingUniques()` with `getMatchingApplyingUniques()` or small coding style changes.

Deprecated uniques:
- [amount]% Growth [cityFilter] when not at war

Added conditionals:
- \<when at war\>
- \<when not at war\>


SimonCeder:
Nice.
Perhaps the city states `allyBonus` `friendBonus` per era can be simplified to a long list of uniques with <when allies> <during the Classical Era> and so on?

xlenstra:
> Nice.
> Perhaps the city states `allyBonus` `friendBonus` per era can be simplified to a long list of uniques with \<when allies\> \<during the Classical Era\> and so on?

I'd personally say that they should still be saved in the eras.json (unless we create a cityState.json), but they can be restated as `[stats] <from [relationFilter] City-States>` or something, so that would definitely be possible

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -269,7 +273,16 @@ fun String.tr(): String {
...
+fun String.getPlaceholderText() = 
+    this.replace(squareBraceRegex, "[]")
+        .replace(pointyBraceRegex, "")
+        // So, this is a quick hack, but it works as long as nobody uses word separators different from " " (space) and "" (none).
+        // According to https://linguistics.stackexchange.com/questions/6131/is-there-a-long-list-of-languages-whose-writing-systems-dont-use-spaces
+        // This is a reasonable but not fully correct assumption to make.
+        // By doing it like these, we exclude languages such as Tibetan, Dzongkha (Bhutan), and Ethiopian.
+        // If we ever start getting translations for these, we'll fix them then.
+        .replace("  ", " ")
+        .trim()
```
yairm210, line 285: So this is where the magic happens...

yairm210:
Instead of changing the function from hasMatchingUniques to hasMatchingApplyingUniques, which will require a huge rename, why not have an overload? (I assume more will follow)
getMatchingUniques(placeholderText)
getMatchingUniques(placeholderText, civInfo)

This will allow more fine-grained conditionals for specific uniques in the future, like:
getMatchingUniques(placeholderText, cityInfo)
getMatchingUniques(placeholderText, unitInfo)
etc.

yairm210:
This is especially relevant since the need for multiple filters was raised first and foremost through the multiple unit filters (military water units etc)

yairm210:
There's a lot of fluff here getting in the way of the actually important parts of the PR.
Please remove all renames, AND check the translation - I didn't see any changes in the .tr() function, and from what I see all translations with conditionals will be totally broken.
It seems to me that conditionals will need to go at the end of the line.

Unless ALL conditionals are set in a specific place, e.g. the end of the line, I don't see a way to handle this. You can't just "splice" an existing translation and insert conditionals in the middle and hope it parses...
And now the question becomes, "can all languages handle delimiter-conditionals". Sounds like a question for Discord.

xlenstra:
> There's a lot of fluff here getting in the way of the actually important parts of the PR.
> Please remove all renames, AND check the translation - I didn't see any changes in the .tr() function, and from what I see all translations with conditionals will be totally broken.
> It seems to me that conditionals will need to go at the end of the line.
> 
> Unless ALL conditionals are set in a specific place, e.g. the end of the line, I don't see a way to handle this. You can't just "splice" an existing translation and insert conditionals in the middle and hope it parses...
> And now the question becomes, "can all languages handle delimiter-conditionals". Sounds like a question for Discord.

Tbh, I completely forgot about everything that has to do with translations. I'll wait for the feedback in discord and recreate this PR based on that. I'll also remove all the unnecessary fluff and start with only a few uniques being able to use conditionals, which should remove most of the renames.

PR #5178, opened 2021-09-10
Fixed some bugs

xlenstra:
- Fixed bug where moving a unit stopped automation
- Fixed bug where chosing an icon last in the religion screen would not enable the 'found religion' button
- Fixed bug where buying units with an increasing cost would deduct the wrong amount
- Fixed bug making it impossible to pick free techs

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -503,9 +503,25 @@ class CityConstructions {
         if (!cityInfo.civInfo.gameInfo.gameParameters.godMode) {
             val construction = getConstruction(constructionName)
             if (construction is PerpetualConstruction) return false
-            val constructionCost = (construction as INonPerpetualConstruction).getStatBuyCost(cityInfo, stat)
+            val constructionCost =
+                (construction as INonPerpetualConstruction).getStatBuyCost(cityInfo, stat)
             if (constructionCost == null) return false // We should never end up here anyway, so things have already gone _way_ wrong
             cityInfo.addStat(stat, -1 * constructionCost)
+
+            if (cityInfo.civInfo.getMatchingUniques("May buy [] units for [] [] [] starting from the [] at an increasing price ([])")
+                .any {
+                    (
+                        construction is BaseUnit && construction.matchesFilter(it.params[0]) ||
+                        construction is Building && construction.matchesFilter(it.params[0])
+                    )
+                    && cityInfo.matchesFilter(it.params[3])
+                    && cityInfo.civInfo.getEraNumber() >= cityInfo.civInfo.gameInfo.ruleSet.eras[it.params[4]]!!.eraNumber
+                    && it.params[2] == stat.name
+                }
+            ) {
+                cityInfo.civInfo.boughtConstructionsWithGloballyIncreasingPrice[constructionName] =
+                    (cityInfo.civInfo.boughtConstructionsWithGloballyIncreasingPrice[constructionName] ?: 0) + 1
+            }
         }
 
         if (queuePosition in 0 until constructionQueue.size)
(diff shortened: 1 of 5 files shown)


SomeTroglodyte:
> moving a unit stopped automation

Well, manually moving an automated unit _should_ cancel automation? Like, either you want it to do _its_ thing _or_ you want it to do _your_ thing?

xlenstra:
> > moving a unit stopped automation
> 
> Well, manually moving an automated unit _should_ cancel automation? Like, either you want it to do _its_ thing _or_ you want it to do _your_ thing?

Problem was that when it move _as part of its automation_, it would also stop the automation. For now I've reverted to behaviour before my changes fixing unit movement bugs

SomeTroglodyte:
> as part

I expected that much. Why didn't I trust you to be clear about the other aspect? Dunno. Maybe because I'm now fighting scene2d layout, where _nothing_ can be trusted and _nothing_ does exactly what it sounds like it should do... Sorry.

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -503,9 +503,25 @@ class CityConstructions {
         if (!cityInfo.civInfo.gameInfo.gameParameters.godMode) {
             val construction = getConstruction(constructionName)
             if (construction is PerpetualConstruction) return false
-            val constructionCost = (construction as INonPerpetualConstruction).getStatBuyCost(cityInfo, stat)
+            val constructionCost =
+                (construction as INonPerpetualConstruction).getStatBuyCost(cityInfo, stat)
```
SomeTroglodyte, line 507: Wait - if PerpetualConstruction and INonPerpetualConstruction are mutually exclusive, inverting that test in 505 should make the cast redundant... ?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -503,9 +503,25 @@ class CityConstructions {
...
             if (constructionCost == null) return false // We should never end up here anyway, so things have already gone _way_ wrong
             cityInfo.addStat(stat, -1 * constructionCost)
+
+            if (cityInfo.civInfo.getMatchingUniques("May buy [] units for [] [] [] starting from the [] at an increasing price ([])")
+                .any {
+                    (
+                        construction is BaseUnit && construction.matchesFilter(it.params[0]) ||
+                        construction is Building && construction.matchesFilter(it.params[0])
```
SomeTroglodyte, line 515: Unique text says "[] units" - support `Building` anyway? And - the newbie-puzzling `||` wouldn't be necessary if `matchesFilter` were part of `INonPerpetualConstruction`, right?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -503,9 +503,25 @@ class CityConstructions {
...
+                    )
+                    && cityInfo.matchesFilter(it.params[3])
+                    && cityInfo.civInfo.getEraNumber() >= cityInfo.civInfo.gameInfo.ruleSet.eras[it.params[4]]!!.eraNumber
```
SomeTroglodyte, line 518: That chain is getting long... See also `civInfo.gameInfo.ruleSet.eras[policy.branch.era]!!.eraNumber` - maybe that can be prettified a little?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -503,9 +503,25 @@ class CityConstructions {
...
+                    && it.params[2] == stat.name
+                }
+            ) {
+                cityInfo.civInfo.boughtConstructionsWithGloballyIncreasingPrice[constructionName] =
+                    (cityInfo.civInfo.boughtConstructionsWithGloballyIncreasingPrice[constructionName] ?: 0) + 1
```
SomeTroglodyte, line 523: Sounds like what `Counter` was designed for (and yes I know Counter serializes garishly) - we should maybe prefer fixing Counter instead of circumventing it?

SomeTroglodyte (approved):
I see nothing really wrong here

Issue #5127, opened 2021-09-06
Do we want a zoomable Tech tree?


SomeTroglodyte:
On a scale from BS to must-have - what's this:

https://user-images.githubusercontent.com/63000004/132259218-ebf23ff5-2278-4748-97a8-999bd7e5f954.mp4

xlenstra:
As long as this doesn't use the scroll wheel on desktop it doesn't hurt. It looks kinda cool fully zoomed out, and checking icons on smaller screens by zooming in might also be useful

SomeTroglodyte:
Of course the wheel zooms - same widget as main map. Must be a facepalm-level blindness that I don't see why that would be bad? Right now the wheel scrolls the thing a few units up and down mostly, not even left or right. What the trivial patch can't do is to zoom while keeping the place under the mouse centered - the worldscreen doesn't either, but there it's OK.

xlenstra:
True, I guess scrolling for zooming does make sense. I would prefer it if the scroll wheel scrolls sideways, as that is most often what you want to do in the tree anyway, and is already the behaviour shown so when playing at large resolutions (1920x1280 for example). Just my two cents

SomeTroglodyte:
Ah yes the default ScrollPane scrolls sideways if there is not enough content to scroll vertically. But that depends on window, world resolution - and tech tree size. Anyway, shutting down for a day.

SimonCeder:
On desktop I'd normally expect wheel to scroll horizontally, shift+wheel to scroll vertically, and ctrl+wheel to zoom.

yairm210:
Looks good!

SomeTroglodyte:
Ouch! Now i'm obliged to polish this to perfection...

@Simon - actually that sounds very specific to some office package. I'd expect the wheel to go vertical - as a generic prejudice. With Ctrl, yes, that seems quite universal. I'll think about it. You know there's wheels that also go sideways? Wonder if that can be caught with Gdx...

SimonCeder:
Woops, I meant wheel to scroll vertically, and shift+wheel to scroll horizontally, of course. You see that in plenty of programs that have horizontal scroll bars, most text and image editors for example.

yairm210:
> Ouch! Now i'm obliged to polish this to perfection...

Absolutely not, i'd much rather we close some of the existing 100+ issues than open new ones 😅

SomeTroglodyte:
[...]

PR #5182, opened 2021-09-11
Zoomable tech tree

SomeTroglodyte:
This is ***broken*** - last version of #5127

I just present it like this in case someone finds out what's wrong with it (my guess: Gdx is what's wrong).

Symptom: Everything is fine until you click a tech while zoomed out. The ScrollPane clips to the size it would have had before the resize to respect scale. Size fixes itself with another zoom - but scroll position has already been broken. In the 'broken' state all dimensions visible in the debugger seem to be identical to the ones before breaking. Breaking occurs as soon as the description label is updated - and that's neither an ascendant nor descendant of anything important. Resizing after selectTechnology doesn't help. Overriding getPref* and returning _any_ values changes nothing.  Programatically zooming after the selectTechnology call doesn't help, neither does picking apart the widget hierarchy between splitPane and techTable, packing, reassembling, in various combinations.

I give up.

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
@@ -4,7 +4,7 @@ import com.badlogic.gdx.scenes.scene2d.ui.*
 import com.unciv.Constants
 import com.unciv.UncivGame
 import com.unciv.ui.utils.*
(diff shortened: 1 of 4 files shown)


SomeTroglodyte:
Case study has been up for inspection long enough.

logicminimal:
damn, unsolvable bugs in huge projects :<

PR #5187, opened 2021-09-11
Implementation of conditionals, but better than before

xlenstra:
This PR is a replacement for #5166, providing an actual basic implementation of conditional uniques.

It mostly fixes the translation problem, by allowing translators to set a global order in which all conditionals will be placed.
To further improve on the generation, translators are now also asked what the word boundary is in their language, and whether they capitalize the first letter of a sentence.
This is implemented by abusing the translation database by adding strings that aren't literal translations to them.
It also adds tests whether these translations are correctly formatted. To run these tests, Dutch translations were added.

The conditionals are still parsed in the same way as before, and still checked in the same way as well, though I'm now planning to work my way through existing uniques and add conditionals to a few uniques at a time, instead of all at once. This reduces an enormous amount of clutter existing in the previous PR.

Deprecated uniques:
- "+[amount]% growth [cityFilter]"
- "+[amount]% growth [cityFilter] when not at war"

Added uniques:
- "[amount]% growth [cityFilter]"

Added conditionals:
- \<when at war\>
- \<when not at war\>

Added conditional support for the following uniques:
- "[amount]% growth [cityFilter]"


yairm210:
Upper case vs lower case Growth in the Json uniques

core/src/com/unciv/models/ruleset/Unique.kt
```diff
@@ -7,23 +7,45 @@ import com.unciv.logic.map.MapUnit
...
+        condition: Unique, 
+        civInfo: CivilizationInfo? = null
+    ): Boolean {
+        when (condition.placeholderText) {
+            "when not at war" -> if (civInfo == null || civInfo.isAtWar()) return false
+            "when at war" -> if (civInfo == null || !civInfo.isAtWar()) return false
+            else -> return false
+        }
+        return true
+    }
```
SomeTroglodyte, line 48: Can't help it, my fingers are typing without my conscious support:
```kotlin
) = when (condition.placeholderText) {
            "when not at war" -> civInfo?.civInfo.isAtWar() == false
            "when at war" -> civInfo?.civInfo.isAtWar() == true
            else -> false
        }
```
..makes no bytecode difference.

core/src/com/unciv/models/ruleset/Unique.kt
```diff
@@ -7,23 +7,45 @@ import com.unciv.logic.map.MapUnit
...
     val stats: Stats by lazy {
         val firstStatParam = params.firstOrNull { Stats.isStats(it) }
         if (firstStatParam == null) Stats() // So badly-defined stats don't crash the entire game
         else Stats.parse(firstStatParam)
     }
+    val conditionals: List<Unique> = text.getConditionals()
+    
+    
+    // This function will get LARGE, as it will basically check for all conditionals if they apply
+    // This will require a lot of parameters to be passed (attacking unit, tile, defending unit, civInfo, cityInfo, ...)
```
SomeTroglodyte, line 28: I think it's good. The tests need access to a many objects and should stay agnostic on how to determine _which_ specific instances these are. Any 'elegant' approach using collections of polymorphic interfaces or worse would only complicate matters without significant gain. Also, can't get more than twice the number of Civilopedia pages. If they get passed around very often, wrap in a data class (data for the flexibility of the copy method, and for the free deconstruction support).

```kotlin
data class DataForConditionalTest(val civ: CivilizationInfo? = null, val city: CityInfo? = null, val attacker: ICombatant? = null,...)
fun conditionalsApply(data: DataForConditionalTest) = conditionals.all { conditionalApplies(it, data) }
fun conditionalApplies(condition: Unique, data: DataForConditionalTest): Boolean {
    (civ, city, attacker) = data
    return when (condition.placeholderText) {
...
}

```

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -196,6 +219,10 @@ val eitherSquareBraceRegex = Regex("""\[|\]""")
 // Analogous as above: Expect a {} pair with any chars but } in between and capture that
 val curlyBraceRegex = Regex("""\{([^}]*)\}""")
 
+// Analogous as above: Expect a <> pair with any chars but > in between and capture that
+val pointyBraceRegex = Regex("""\<([^>]*)\>""")
```
SomeTroglodyte, line 223: The inspections should complain that <> do not need escaping??

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -213,6 +241,57 @@ fun String.tr(): String {
...
+            conditionsWithTranslation[conditional.placeholderText] = conditional.text.tr()
+        
+        val translatedConditionals: MutableList<String> = mutableListOf()
+        
+        // Somewhere, we asked the translators to reorder all possible conditionals in a way that
+        // makes sense in their language. We get this ordering, and than extract each of the
+        // translated conditionals, removing the <> surrounding them, and removing placeholder text
+        // where if it exists.
+        val conditionalOrdering = UncivGame.Current.translations.getConditionalOrder(language)
+        for (placedConditional in pointyBraceRegex.findAll(conditionalOrdering).map { it.value.substring(1, it.value.length-1).getPlaceholderText() }.toList()) {
```
SomeTroglodyte, line 276: Hmmm, Without really checking - the toList should be redundant, you're not modifying anything the iterator sees. And removing first and last char should be possible to already do in the regex capturing - looks like it already does. The capture won't contain the braces. So don't use the entire match, use the capture. Or if you don't use the capture, don't capture.

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -269,7 +347,9 @@ fun String.tr(): String {
     return UncivGame.Current.translations.getText(this, UncivGame.Current.settings.language, activeMods)
```
SomeTroglodyte, line 347: Can't this reuse the cached language?

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -291,3 +371,17 @@ fun String.fillPlaceholders(vararg strings: String): String {
         filledString = filledString.replaceFirst(keys[i], strings[i])
     return filledString
 }
+
+fun String.getConditionals() = pointyBraceRegex.findAll(this).map { Unique(it.groups[1]!!.value) }.toList()
```
SomeTroglodyte, line 375: See, you _are_ getting the capture here. Am I missing something?

tests/src/com/unciv/testing/TranslationTests.kt
```diff
@@ -188,4 +185,48 @@ class TranslationTests {
...
+        }
+        
+        Assert.assertTrue(
+            "This test will only pass when the word boundrary translation succeeds",
+            allTranslationsCheckedOut
+        )
+    }
+    
+    @Test
+    fun allConditionalsAreContainedInConditionalOrderTranslation() {
```
SomeTroglodyte, line 211: Not sure whether relaxing this might be better. How often did we explain test failures to translators recently? You did the code to default to original ordering...

xlenstra: I don't have a particularly strong reasoning either way, but more relaxation does mean more possibility for unintended typo's, and translators than need to fix that anyway for it to work, regardless of how whether we have to explain that. At least this way we know that they made a mistake, instead of them being confused why three versions later it isn't working as they expected it to

core/src/com/unciv/models/translations/Translations.kt
xlenstra: Tbh, regex is complete magic to me and I just copy+pasted things until it worked. At my end Android Studio is complaining about every single of these regexes, so I just ignored it and hoped that either nobody would mind or someone else would later fix it :)

core/src/com/unciv/models/translations/Translations.kt
xlenstra: Dunno, when I used the `pointyBraceRegex` I always got it including the `<>`, maybe Kotlin regex is just wierd

core/src/com/unciv/models/ruleset/Unique.kt
xlenstra: The lack of conscious support was clear from the doubling of the `civInfo` :p
But in the end that would probably look cleaner, so I'll use this as inspiration

core/src/com/unciv/models/translations/Translations.kt
SomeTroglodyte: Debug, and Alt-F8 and play around. Or are you coding _everything_ using Vi from a command line? :rofl: 

Regex.findall().first().value is the first match. Regex.findall().first().groups[0].value is, too. Regex.findall().first().groups[1].value is the first capture of the first match... Regex("Hello (\w+) world").findAll("Hello kitty world").first().groups[1]!!.value == "kitty" ... 

Or use one of the offline Regex test playgrounds. What were they named... Is Expresso still open? Coach... http://weitz.de/regex-coach/ ... OK but win-crud... Mint offers github.com/artemanufrij/regextester - hey a YAK! To really get a feeling for Regexes you _really_ need to play with them interactively.

SomeTroglodyte: Oh, and the only point where kotlin is weird anywhere near Regexes is IMHO raw strings. Those are completely moronic. Try writing a pattern searching for a non-line-end `$`... (answer: `"""${'\'}${'$'}"""`)

core/src/com/unciv/models/ruleset/Unique.kt
SomeTroglodyte: > inspiration

wasn't meant as anything other

(yay! double negative!)

core/src/com/unciv/models/translations/Translations.kt
SomeTroglodyte: > hoped that nobody would mind

What an invitation for satire!

Let's just be glad that it doesn't refuse to _compile_ a perfectly valid Regex (again).

SomeTroglodyte:
> Received status code 500 from server: Internal Server Error

And here we go again with broken checks through no fault of our own...

android/assets/jsons/translations/template.properties
```diff
@@ -1,3 +1,12 @@
+# Language settings
+
+# Equivalent of a space in your language
+# If your language doesn't use spaces, just add "" as a translation, otherwise " "
+" " = 
+
+# If the first word in a sentence starts with a capital, translate this as 'true', otherwise as 'false'
```
yairm210, line 7: How about StartWithCapitalLetter, the word "Capital" has a different meaning in-game. I was confused by this.

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -177,7 +178,29 @@ class Translations : LinkedHashMap<String, TranslationEntry>(){
...
+    /** Returns the equivalent of a space in the given language
+     * Defaults to a space if no translation is provided
+     */
+    fun getSpaceEquivalent(language: String): String {
+        val translation = getText("\" \"", language, null)
+        return translation.substring(1, translation.length-1)
+    }
+    
+    fun shouldCapitalize(language: String): Boolean {
+        return get(shouldCapitalizeString, language, null)?.get(language)?.toBoolean() ?: true
```
yairm210, line 197: This requires protection against translators translating to "true" or "false" in their own language. The toBoolean can crash.
Also it should probably be explicit in the translation file, put like "'true' or 'false' IN ENGLISH"

yairm210:
Okay, I think this is almost ready to go! Getting close!

core/src/com/unciv/models/translations/Translations.kt
xlenstra: According to the documentation I could find in Android Studio, `.toBoolean()` only checks if it's "true", and if it's anything else it returns `false`, so no crashes should occur. I will add the note in the template.properties though

yairm210: Oh, nice!
I was going by toInt, I know that crashes 😅

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -177,7 +178,29 @@ class Translations : LinkedHashMap<String, TranslationEntry>(){
...
+    }
+    
+    companion object {
+        // Whenever this string is changed, it should also be changed in the translation files!
+        const val defaultConditionalOrderingString = "<UNIQUE> <when at war> <when not at war>"
```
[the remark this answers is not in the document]
yairm210: Choosing *where* the conditions go in the sentence is vital.
Choosing the order of them feels less so, and will lead to a 'renaming' every time we add a condition.
And then what? Either we rename the existing, and then the translation doesn't contain a location for the new condition
Or, we erase the old one, and then the translator needs to do everything from scratch, which is worse

I think this should be divided into 2:

A. [unique] [conditions]
B. Condition A, condition b, etc

The first, as I said, is vital
The second I feel is problematic

SomeTroglodyte:
> translators translating to "true" or "false" in their own language

Maybe 0 and 1?

core/src/com/unciv/models/translations/Translations.kt
SomeTroglodyte: You can Ctrl-B on that and check the source - yes it's non-throwing. Use toBooleanStrict for the crashing version.

yairm210:
Looks mergable, resolve conflicts :)

xlenstra:
> Looks mergable, resolve conflicts :)

Done!

xlenstra:
For me the workflow's don't show up, not sure if that's a problem on my end or just this PR being weird. At least locally the unit tests pass, so it should be fine?

SomeTroglodyte:
> the workflows don't show up

You mean "Workflow runs completed with no jobs"? We should be able to figure that one out, right?

xlenstra:
Reopening fixed it for now, no idea what caused this

xlenstra:
Merged in the browser, hoping it still compiles

xlenstra:
To avoid having to merge in master another time, are you fine with me merging this PR in myself when I've fixed the conflicts next time, Yairm?

yairm210:
I prefered to leave this until the next version.
Unfortunately, with 2 "patch" versions already, the version release is pushed back further than I expected.
So yeah, sure. Next version will be a monster though, expect many patch versions for that one as well...

PR #5190, opened 2021-09-12
Handle maps with invalid mapSize more gracefully

SomeTroglodyte:
The _extreme_ version of making Unciv handle bad maps such as the one in #5179 or the "Large_Africa" i found on my box (forgot source). Feel free to reject partially or entirely.

- Makes the save maps compressed switch available in-game from secret options (making it easier to manually fix maps - like the Gold not-ore in that Africa)
- Marks maps with the Unciv version they were created with as help for just such support tickets (new maps only) - needs debugging
- Rejects maps with mapSize not matching tile count and displays why on NewGameScreen
- Pops a toast and auto-fixes mapSize for these maps on load from map editor
- Silly earlier mistake: There are both "reset to default" and "reset to defaults" - reduced to plural variant even if that's more work for the translators as that fits both uses better
- padBottom of that reset to defaults button was wrong (implemented as inside not outside) - as there are no cells in that container I fear this is the smallest way to get that padding


xlenstra (approved):
Looks good!

PR #5198, opened 2021-09-12
City-States Influence rates; Wary status; Proximity calculations

SimonCeder:
Part of #4602 
This PR implements proper influence resting points, decay and regrowth rates.
It also implements the Wary Of mechanic where city states have a chance to grow wary of you after you've attacked at least two, this lowers their influence resting point by 20.

To do this I also had to implement proximity calculations where civs are graded as being Neighbors, Close, Far, or Distant to each other. I'm sure we'll find more uses for the in AI later, so it needed to be implemented anyway.

To do *this* I also had to implement a way to figure out if two cities are on the same continent. I decided to add a `continent` field to TileInfo which is assigned at map generation. Two tiles are on the same continent if their continent numbers are the same. Again I'm sure this will be useful down the line.

So quite a bit of code for a small functionality at this time.

Screenshot_2021-09-12_22-00-01

A note about Permanent War: In some sources you can read about city-states banding together in Permanent War against attackers. This is only true for vanilla Civ V, in G&K and later this was removed in favor of the Wary Of mechanic.

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -10,6 +10,7 @@ import com.unciv.logic.civilization.RuinsManager.RuinsManager
 import com.unciv.logic.civilization.diplomacy.DiplomacyFlags
 import com.unciv.logic.civilization.diplomacy.DiplomacyManager
 import com.unciv.logic.civilization.diplomacy.DiplomaticStatus
+import com.unciv.logic.map.MapType
 import com.unciv.logic.map.MapUnit
 import com.unciv.logic.map.TileInfo
 import com.unciv.logic.map.UnitMovementAlgorithms
@@ -32,6 +33,14 @@ import kotlin.collections.HashMap
 import kotlin.math.min
 import kotlin.math.roundToInt
 import kotlin.math.sqrt
+
+enum class Proximity {
+    None, // ie no cities
+    Neighbors,
+    Close,
+    Far,
+    Distant
+}
 
 class CivilizationInfo {
 
@@ -107,6 +116,7 @@ class CivilizationInfo {
     var victoryManager = VictoryManager()
     var ruinsManager = RuinsManager()
     var diplomacy = HashMap<String, DiplomacyManager>()
+    var proximity = HashMap<String, Proximity>()
     var notifications = ArrayList<Notification>()
     val popupAlerts = ArrayList<PopupAlert>()
     private var allyCivName: String? = null
@@ -164,6 +174,7 @@ class CivilizationInfo {
         toReturn.allyCivName = allyCivName
         for (diplomacyManager in diplomacy.values.map { it.clone() })
             toReturn.diplomacy[diplomacyManager.otherCivName] = diplomacyManager
+        toReturn.proximity.putAll(proximity)
         toReturn.cities = cities.map { it.clone() }
 
         // This is the only thing that is NOT switched out, which makes it a source of ConcurrentModification errors.
@@ -194,6 +205,9 @@ class CivilizationInfo {
 
     fun getDiplomacyManager(civInfo: CivilizationInfo) = getDiplomacyManager(civInfo.civName)
     fun getDiplomacyManager(civName: String) = diplomacy[civName]!!
+
+    fun getProximity(civInfo: CivilizationInfo) = getProximity(civInfo.civName)
+    fun getProximity(civName: String) = proximity[civName] ?: Proximity.None
 
     /** Returns only undefeated civs, aka the ones we care about */
     fun getKnownCivs() = diplomacy.values.map { it.otherCiv() }.filter { !it.isDefeated() }
@@ -386,10 +400,6 @@ class CivilizationInfo {
     fun makeCivilizationsMeet(otherCiv: CivilizationInfo) {
         meetCiv(otherCiv)
         otherCiv.meetCiv(this)
-
-        // Generate proximity rankings
-        getDiplomacyManager(otherCiv).updateProximity(
-            otherCiv.getDiplomacyManager(this).updateProximity())
     }
 
     private fun meetCiv(otherCiv: CivilizationInfo) {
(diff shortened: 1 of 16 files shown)


SimonCeder:
yep, I don't get why this is failing.

SomeTroglodyte:
> same continent

Refactored around that `landTilesInBigEnoughGroup` variable like I mentioned somewhere?

How do you not get complaints about `getContinent`, `setContinent`, `continent` and JVM signatures?

> throw java.lang

This is kotlin. Do not use Java min, use kotlin min. Likewise...

> runAndMeasure

Remind me to throw these functions out, kotlin has premade ones in Duration...

SomeTroglodyte:
> Could not GET 'https://jitpack.io/com/github/anuken/packr/-SNAPSHOT/maven-metadata.xml'.

Bad luck, the action VM didn't have healthy internet or jitpack.io was down...

SomeTroglodyte:
> private fun assignContinents

So, no, you copied it so now the costly cr.. ahem, runs three times instead of one. You know, I've meant to optimize that as it can waste around 25% of GameStarter time, but never got around to it? If you want, do it now. Move the assignContinents call to before addCivStartingUnits, produce both the continent numbers and that `landTilesInBigEnoughGroup` in one go, cache it and reuse in `getStartingLocations`. Don't forget to baseline properly so you can document the improvement here.

BTW, jitpack is still down.

SomeTroglodyte:
No, jitpack just takes over a minute to deliver a 300 byte file. I've retriggered the check though chances aren't good that the VM has a long enough timeout.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -895,6 +920,80 @@ class CivilizationInfo {
...
+                if (minDistance > distance) minDistance = distance
+            }
+        }
+
+        if (minDistance <= 7) {
+            proximity = Proximity.Neighbors
+        }
+        else if (connections > 0) {
+            val averageDistance = totalDistance / connections
+            val mapFactor = (height + width) / 2 // Perhaps slightly lower for hexagonal maps??
```
SomeTroglodyte, line 958: Look at `MapSize` - I'd use radius for hexagonal here. Or for all shapes HexMath.getDistanceFromEdge(Vector2.Zero). Maybe.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -895,6 +920,80 @@ class CivilizationInfo {
...
+                totalDistance += distance
+                connections++
+                if (minDistance > distance) minDistance = distance
+            }
+        }
+
+        if (minDistance <= 7) {
+            proximity = Proximity.Neighbors
+        }
+        else if (connections > 0) {
```
SomeTroglodyte, line 956: Kotlin style guide says no - same line

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -895,6 +920,80 @@ class CivilizationInfo {
...
+                Proximity.Close
+            else if (averageDistance <= farDistance)
+                Proximity.Far
+            else
+                Proximity.Distant
+        }
+
+        // Check if different continents (unless already max distance, or water map)
+        if (connections > 0 && proximity != Proximity.Distant
+            && gameInfo.tileMap.mapParameters.type != MapType.archipelago) {
```
SomeTroglodyte, line 972: archipelago with water level slider at -0.05 or below will have large landmasses, while default with that slider way to the right will be a bunch of islands...

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -895,6 +920,80 @@ class CivilizationInfo {
...
+
+            if (getCapital().getCenterTile().getContinent() != otherCiv.getCapital().getCenterTile().getContinent()) {
+                // Different continents - increase separation by one step
+                proximity = when (proximity) {
```
SomeTroglodyte, line 976: .. or `operator fun inc() = values()[ordinal+1]` in the enum class and `proximity++` here .. Surprisingly, that works if you know you can't hit the 'edges' .. For short cases like this when is probably clearer.

logicminimal:
so this implements landmasses?

SimonCeder:
Depends on what you mean. We will end up with a bunch of landmass IDs that are saved

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SimonCeder: Right, added an `isWaterMap` function that utilizes the continents data we generate instead.

SimonCeder:
BFS is now only run once. The performance gain is actually slim:

```
Master:
getStartingLocations BFS took 1.99ms
getStartingLocations BFS took 1.83ms

Intial PR:
MapGenerator.assignContinents took 2.45ms
getStartingLocations BFS took 1.58ms
getStartingLocations BFS took 1.39ms

Reviewed PR:
MapGenerator.assignContinents took 2.96ms
```
(When generating a huge pangaea, standard settings, seed 1111111111111)
While `spreadResources` takes 500-ish ms.

Each `TileInfo` now has a `continent` field for continent ID, and the `MapInfo` has a` continentSizes<Int, Int>` map that holds the number of tiles for each continent. Again this is potentially useful information. I'm using it in this PR to figure out if we are on a water map by checking if continents with more than 20 tiles > number of players.

core/src/com/unciv/logic/map/TileMap.kt
```diff
@@ -33,6 +33,7 @@ class TileMap {
     var mapParameters = MapParameters()
 
     private var tileList = ArrayList<TileInfo>()
+    val continentSizes = HashMap<Int, Int>()    // Continent ID, Continent size
```
SomeTroglodyte, line 36: This is what Mt. Fuji and Sri Pada placement could use :+1:

SomeTroglodyte:
> slim

Indeed. I remember baselining a monster map totaling 10 seconds where resources and rivers were 60-70% and a big chunk of the rest was those landmass-big-enough runs, so I assumed a total gain of at least 10% of the entire GameStarter should be possible. Maybe I measured wrongly.

SimonCeder:
Here's what I came up with. `assignContinents` goes back into `MapGenerator` so the data's avaliable for natural wonders. We have three possible scenarios:
* A freshly generated map. `assignContinents` is called from `MapGenerator`, we have a cached `landTilesInBigEnoughGroup`. All's well.
* A pre-made map without saved continent data (because it's old). `assignContinents` is called from `GameStarter`, we have a cached `landTilesInBigEnoughGroup`. Also works fine.
* A pre-made map *with* saved continent data. We don't have a `landTilesInBigEnoughGroup` cached. For this instance, the old BFS in `getStartingLocations` makes a return. Still only one BFS called per map!

yairm210:
This looks good!
The proximity calculation in particular will probably be useful in many future AI-related improvements, so kudos for that :)
I only have one minor point, which is - "random" outcomes should be seeded by something (tile location, turn number, etc) so players don't save-scum to get the optimal random result.

PR #5202, opened 2021-09-13
Fixed crash from new piety policy

### LINKED ISSUE #5199: Piety update crashes game
CityStats.getStatPercentBonusesFromUniques:
```kotlin
            for (unique in uniques.filter { it.placeholderText == "+[]% Production when constructing [] buildings" }) {
                val stat = Stat.valueOf(unique.params[1])
```
One can immediately see this will crash when fed "+[100]% Production when constructing [Shrine] buildings", "+[100]% Production when constructing [Temple] buildings"

Compare "Police State": "+[100]% Production when constructing [Courthouse]"

Actually, feeding "user input" (mods) to Enum.valueOf is an invitation for DOS attacks. Not that we can help it when we have directly deserialized enums, but in unique params that's too easy. :nerd_face: 

From my shelf:
```kotlin
inline fun <reified T: Enum<T>> enumValueOfOrNull(name: String?): T? {
    if (name == null) return null
    return try {
        enumValueOf<T>(name)
    } catch (ex: IllegalArgumentException) {
        println("Could not parse\"$name\" as ${T::class.java.simpleName}.")
        null
    }
}
```
And in Stat.Companion choose one of:
```kotlin
        fun valueOfOrNull(name: String) = enumValueOfOrNull<Stat>(name)
        fun safeValueOf(name: String, default: Stat = Gold) = enumValueOfOrNull(name) ?: default
```

xlenstra:
Fixes #5199

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Policies.json
+++ b/android/assets/jsons/Civ V - Vanilla/Policies.json
@@ -135,7 +135,7 @@
 	},{
 		"name": "Piety",
 		"era": "Classical era",
-		"uniques": ["+[100]% Production when constructing [Shrine] buildings", "+[100]% Production when constructing [Temple] buildings", 
+		"uniques": ["+[100]% Production when constructing a [Shrine]", "+[100]% Production when constructing a [Temple]", 
 			"Incompatible with [Rationalism]"],
 		"policies": [
 			{


SomeTroglodyte:
So there's "+[x]% Production when constructing a [y]", "+[x]% Production when constructing [y]", and "+[x]% Production when constructing [y] buildings"? Who's keeping track of the difference?

We need a new 'unique' system.

I propose we deprecate all of them. What we need is something that allows a single string to be parsed into unique type, attached-to scope, effect(s), effect scope, conditions and condition scopes, ***unambiguously***, and the way from the idea to a valid and unmistakeable unique string must also be ***intuitive***. Possible - if we ditch the expectation that these are gamer-readable.

What we have now not only encourages simple mistakes with embarrassing  consequences, it also does not allow a simple and efficient mod check that has a chance to catch every mistake...

xlenstra:
> So there's "+[x]% Production when constructing a [y]", "+[x]% Production when constructing [y]", and "+[x]% Production when constructing [y] buildings"? Who's keeping track of the difference?
> 
> We need a new 'unique' system.
> 
> I propose we deprecate all of them. What we need is something that allows a single string to be parsed into unique type, attached-to scope, effect(s), effect scope, conditions and condition scopes, _**unambiguously**_, and the way from the idea to a valid and unmistakeable unique string must also be _**intuitive**_. Possible - if we ditch the expectation that these are gamer-readable.
> 
> What we have now not only encourages simple mistakes with embarrassing consequences, it also does not allow a simple and efficient mod check that has a chance to catch every mistake...

Completely agree, and I'm already working on it: #5091, #5187.

PR #5204, opened 2021-09-13
Add missing Natural Wonders

ravignir:
This adds missing natural wonders from G&K and BNW. The latter are commented out, but I decided to leave them here anyway.

z

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -356,7 +356,7 @@
 		"name": "King Solomon's Mines",
 		"type": "NaturalWonder",
 		"production": 6,
-		"occursOn": ["Plains","Desert","Tundra","Snow","Grassland"],
+		"occursOn": ["Plains","Desert"],
 		"turnsInto": "Plains",
 		"impassable": true,
 		"unbuildable": true,
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -305,18 +305,85 @@
 		"impassable": true,
 		"unbuildable": true,
 		"weight": 10
-	}
+	},
+	{
+		"name": "Mount Kailash",
+		"type": "NaturalWonder",
+		"faith": 6,
+		"happiness": 2,
+		"occursOn": ["Plains","Grassland"],
+		"turnsInto": "Mountain",
+		"impassable": true,
+		"unbuildable": true,
+		"weight": 10
+	},
+	{
+		"name": "Mount Sinai",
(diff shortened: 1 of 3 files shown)


xlenstra (approved):
Indentation is off a bit due to mixing of tabs & spaces, but we'll fix that later. Looks good otherwise!

ravignir:
ok, fixed that :)

SomeTroglodyte:
Then you'll need a better editor - or better settings. Kotlin style guide expressly says [do not use tabs](https://kotlinlang.org/docs/coding-conventions.html#indentation) and prescribes 4 spaces per indent. If your editor can handle python without the code breaking after a few edits, you should be fine...

ravignir:
>
>     * What does "avoid ocean" mean? I see there's neither an implementation nor a todo.
> 
That basically means Natural Wonder  can't spawn next to coast, which was omitted up to this point - which is why i did not include it. 
`it.neighbors.none { neighbor -> neighbor.getBaseTerrain().name == Constants.coast }`

>     * "biggest landmass" can be done or soon - See [City-States Influence rates; Wary status; Proximity calculations #5198](https://github.com/yairm210/Unciv/pull/5198) where we discussed how landmass size was already done (twice) and discarded by GameStarter. So now we know that code should be refactored in a way that it can run within map generation _or_ on a map loaded from file, and in the case a generated map is not for the editor but 'started' directly, a side-channel to pass the info from map generation to gamestarter could save time and GC load.
> 
ok. So that is still ToDo later.

>     * The wonders urgently need "uniques" style conditions - one loop, no names special-cased as constants, no function per wonder, just json - should be possible, all those conditions are very similar.
> 
Civ5 actually does that, but that would require quite a lot of uniques. Unfortunately, I don't feel I am capable of doing that (yet). 

>     * `&& it.improvement == null` - that's just copied - but why is this here? There can't _be_ any improvements at this point - ancient ruins come last.
> 
ok. I did not know that. But If I recall correctly, it used to be the case in past.
>     * I see json and code, but no FantasyHex or atlas - but your screenshot must have had those, right?
> 
It was already added in #4916, so that is why there is only code and json :)


To sum up: I will add avoid ocean thing and check if reddit source is indeed right, because i recall Sri Pada spawning next to coast tiles in my civ5 games.

SomeTroglodyte:
> `it.neighbors.none { neighbor -> neighbor.getBaseTerrain().name == Constants.coast }`

Now you mention it, `neighbors` is a little weird. The actual materialized list is nowhere 'kept' explicitly, and we trust the GC to not collect it by seeing the connection through that Sequence offer... Do I understand that? No.

Anyway, if you use `neighbors` you can also use `isCoastalTile()`

SomeTroglodyte (approved):

PR #5206, opened 2021-09-13
Made CN tower functional, and free buildings are removed on capture

### LINKED ISSUE #5185: CN Tower does not provide free Broadcast Tower to cities despite this is declared in this wonder description in civilopedia
**Platform**
Android

**Version**
Google Play, 3.16.12

**Describe the bug**

The CN Tower wonder is declared as giving free Broadcast Tower to all cities, but is not doing so

**To Reproduce**
1) load the savefile [from here](https://disk.yandex.ru/d/hudcMrHhwJQJew) 
2) find Murmansk city in Russia, it has CN Tower built
3) find Tver city nearby Murmansk, it still has Broadcast Tower in construction queue and does not have it in the list of city buildings already built in the city.  Also there are several other cities wo Broadcast Tower

**Expected behavior*
CN tower should do what is declared as a feature of this Wonder

xlenstra:
This PR adds a CivConstructions class, which keeps track of constructions added to cities by global effects.
It also makes the CN tower functional, fixes #5185:
If the owner of the CN tower changes, the originally placed broadcast towers disappear, and broadcast towers are placed in all the cities of the new owner.
Whenever a city is captured, free buildings are now also removed from it, and new free buildings are added to it immediately.

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -43,7 +43,8 @@ class CityConstructions {
     val inProgressConstructions = HashMap<String, Int>()
     var currentConstructionFromQueue: String
         get() {
-            if (constructionQueue.isEmpty()) return "" else return constructionQueue.first()
+            return if (constructionQueue.isEmpty()) "" 
+                else constructionQueue.first()
         }
         set(value) {
             if (constructionQueue.isEmpty()) constructionQueue.add(value) else constructionQueue[0] = value
@@ -53,6 +54,9 @@ class CityConstructions {
     var productionOverflow = 0
     val queueMaxSize = 10
 
+    // Maps cities to the buildings they received
+    val freeBuildingsProvidedFromThisCity: HashMap<String, HashSet<String>> = hashMapOf()
+    
     //region pure functions
     fun clone(): CityConstructions {
         val toReturn = CityConstructions()
@@ -61,6 +65,7 @@ class CityConstructions {
         toReturn.currentConstructionIsUserSet = currentConstructionIsUserSet
         toReturn.constructionQueue.addAll(constructionQueue)
         toReturn.productionOverflow = productionOverflow
+        toReturn.freeBuildingsProvidedFromThisCity.putAll(freeBuildingsProvidedFromThisCity)
         return toReturn
     }
 
@@ -125,7 +130,7 @@ class CityConstructions {
      */
     fun getMaintenanceCosts(): Int {
         var maintenanceCost = 0
-        val freeBuildings = cityInfo.civInfo.civConstructions.getMaintenanceFreeBuildings(cityInfo.id)
+        val freeBuildings = cityInfo.civInfo.civConstructions.getFreeBuildings(cityInfo.id)
         
         for (building in getBuiltBuildings()) {
             if (building.name !in freeBuildings) {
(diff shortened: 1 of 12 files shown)


core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -61,17 +65,22 @@ class CityConstructions {
...
-            .asSequence().filter { it.isBuildable(this) }
+        .asSequence().filter { it.isBuildable(this) }
 
     fun getBasicCultureBuildings() = cityInfo.getRuleset().buildings.values
-            .asSequence().filter { it.culture > 0f && !it.isAnyWonder() && it.replaces == null }
+        .asSequence().filter { it.culture > 0f && !it.isAnyWonder() && it.replaces == null }
+    
+    fun getBasicStatBuildings(stat: Stat) = cityInfo.getRuleset().buildings.values
+        .asSequence()
+        .filter { !it.isAnyWonder() && it.replaces == null && it.getStats(cityInfo)[stat] > 0f }
```
SomeTroglodyte, line 83: `it[stat] > 0f` or this is no longer "basic"? You get a Belief that adds Culture to a Shrine and this function would return the Shrine as Culture building. If intentional - Kdoc what & why.

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -53,6 +54,9 @@ class CityConstructions {
     var productionOverflow = 0
     val queueMaxSize = 10
 
+    // Maps cities to the buildings they received
+    val freeBuildingsProvidedFromThisCity: HashMap<String, HashSet<String>> = hashMapOf()
```
SomeTroglodyte, line 58: Serialized. You sure it isn't preferable to re-calculate on game load?

xlenstra: That wouldn't work, because we have to remember somewhere which buildings we added to which cities. If there are two cities, one of which already contained a broadcast tower and one which was added through the CN tower, we have to differentiate these somehow

core/src/com/unciv/logic/city/CityConstructions.kt
xlenstra: Good point, didn't think about that, fixed

SomeTroglodyte: Except there is no such Belief - but for Temples :rofl:

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: Yeah, I was wondering. So I build a BT in 2 of 5 cities, then the CN. All 5 cities would show a BT in the city screen as built (right?). Can it show which are the free ones?

I now lose CN -> I'll still have the two BT I built first, only the rest is gone? Is it supposed to work like that? I guess you would know. So - OK something needs persistence, understood.

SomeTroglodyte (approved):
Playtested for a few turns - couldn't really verify goal reached, too lazy to artificially stage a situation, but found no regressions.

core/src/com/unciv/logic/civilization/CivConstructions.kt
```diff
@@ -0,0 +1,165 @@
...
+    
+    fun tryAddFreeBuildings() {
+        addFreeStatsBuildings()
+        addFreeSpecificBuildings()
+    }
+    
+    fun getFreeBuildings(cityId: String): HashSet<String> {
+        val toReturn = freeBuildings[cityId] ?: hashSetOf()
+        for (city in civInfo.cities) {
+            toReturn.addAll(city.cityConstructions.freeBuildingsProvidedFromThisCity[city.id] ?: hashSetOf())
```
yairm210, line 97: If this function was called frequently I'd quibble about the useless hashset creation for most cities, but since it's a rare one the effect is very minor

core/src/com/unciv/logic/civilization/CivConstructions.kt
```diff
@@ -0,0 +1,165 @@
...
+            .mapValues { unique -> unique.value.sumOf { it.params[1].toInt() } }
+
+        for ((building, amount) in buildingsUniquesData) {
+            addFreeBuildings(building, amount)
+        }
+    }
+    
+    private fun addFreeBuildings(building: String, amount: Int) {
+        for (city in civInfo.cities.take(amount)) {
+            if (freeSpecificBuildingsProvided[building]?.contains(city.id) == true || city.cityConstructions.containsBuildingOrEquivalent(building)) continue
```
yairm210, line 154: The `city.cityConstructions.containsBuildingOrEquivalent(building)` should cover all cases, no?

core/src/com/unciv/logic/civilization/CivConstructions.kt
```diff
@@ -0,0 +1,165 @@
...
+
+            (city.cityConstructions.getConstruction(building) as INonPerpetualConstruction).postBuildEvent(city.cityConstructions)
```
yairm210, line 156: This is correct, and this comment isn't aimed specifically at you, but the fact that you're using this here is indicative of a deeper problem in the code.
CityConstructions has an addBuilding function, which is called by the postBuildEvent of the Building.
But, the addBuilding is also called directly from other places.
This is a problem of a couple of sorts - if I, a coder, see the `Cityconstructions.addBuilding` function, my first thought would be that this is what I should call to add a building to the city. There's no way I would think to check the postBuildEvent of the city.
I'm not sure what the best way to solve this would be. The beauty of the postBuildEvent is that it lets the CityConstructions 'build' units and cities in the same way, but from a more holistic perpective this is a detriment to the general codebase...

PR #5207, opened 2021-09-13
Fix Khan unit

SimonCeder:
This PR fixes the Khan unit, so adjacent units only get the healing bonus when they would have healed otherwise.
Also fixes adjacent healing bonuses so they don't also affect the center tile in addition to neighboring tiles.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
+++ b/android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
@@ -146,13 +146,13 @@
 	{
 		"name": "Medic",
 		"prerequisites": ["Shock I", "Drill I", "Scouting II", "Survivalism II"],
-		"uniques": ["[+5] HP when healing", "All adjacent units heal [5] HP when healing"],
+		"uniques": ["All adjacent units heal [+5] HP when healing"],
 		"unitTypes": ["Sword","Gunpowder","Mounted","Scout"]
 	},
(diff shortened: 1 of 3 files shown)


yairm210:
I'm like 80% sure that the "adjacent" units includes the center tile. It doesn't make sense that Khan units heal units *next* to them, but not *on the same tile*.

SimonCeder:
It's counterintuitive, but it is how civ V does it, just checked. We could change it of course, but then we'd have to make sure that promotions like medic don't count "double"

ravignir:
o
it says **this** unit and **adjacent** units. Not all units on this tile. There are affects like `all units on this tile and adjacent tiles`, but it is not one of them.

SimonCeder:
Wait.. does medic give extra healing twice to the promoted unit? I'll have to check.

yairm210:
This is extremely odd, but OK.
Another point - the change you've made also affects Medic II, do the effects there still hold up to Civ V?

SimonCeder:
Checked with medic promotion in civ V, it is in fact the case that adjacent unit bonuses do not affect the unit itself.
However as ravignir pointed out, the "enhanced medic" promotion the khan has heals adjacent units, and *this* unit. I added a separate `"[+15] HP when healing"` to the Khan to implement this. Right now the only time it would come into play is if a Khan is damaged by a citadel.

PR #5209, opened 2021-09-13
Civilopedia quickie: Units show other direction of upgrade path

SomeTroglodyte:
image


xlenstra (approved):

yairm210:
Requires spacing so the "can upgrade from" is separated from above and below, it's currently run-on and slightly confusing to read

SomeTroglodyte:
> Requires spacing

OK - from the mail only I was confused but it's clear - I originally had only one "from" path which lists as a single line, and the 'upgrades to' doesn't have extra spacing and looked fine, so I didn't think the other variant would need some, but yes. Might be prettier if I moved it to get less blocks? No, the order feels natural as it is... So, v-pad if multiline, both sides, coming up.

SomeTroglodyte:
image

Too tired for anything else right now. Later alligater. Have phun!

PR #5210, opened 2021-09-13
Panzer can upgrade according to fandom wiki

SomeTroglodyte:


xlenstra:
My in-game civilopedia agrees

xlenstra (approved):

PR #5215, opened 2021-09-14
Embarked units only have 1 vision (except marines)

xlenstra:
Added uniques:
- "[] sight for all [] units"
- "Normal vision when embarked"

Deprecated uniques:
- "+[] sight for all [] units"


yairm210:
Do embarked units have 1 sight regardless of other "sight-enhancing" bonuses they may have?
If so this should be changed to "if(unit.isEmbarked()) return 1"

xlenstra:
I didn't think so, but after quickly checking this, it seems like they do. I'll change this

PR #5217, opened 2021-09-14
Update existing wonders to G&K stats, mostly providing faith instead of culture

xlenstra:


SimonCeder:
What happens if playing without religion?

xlenstra:
Then you will still get all these effects, same as in the original

SimonCeder:
Yes but won't gaining Faith in a religion-off game cause strangeness?
Long term I think we need to implement a vanilla/g&k ruleset switch.

xlenstra:
> Yes but won't gaining Faith in a religion-off game cause strangeness?

In some cases this was already possible, long running games might have the 'monastery' building from before it became a faith-purchasable building, generating faith every turn since them. Natural wonders also include faith generation and aren't removed from games without religion. For some time it was possible to gain great prophets from the liberty finisher or city-state gifts, allowing for the construction of holy sites and faith generation. This adds some additional ways for faith generation in new games without religion, but in my opinion religion should be enabled by default not too far into the near future. This is then a transition phase in which weird stuff can happen.

> Long term I think we need to implement a vanilla/g&k ruleset switch.

We have two options in that:
- Enable G&K by default and allow mods to disable religion, making it possible to make a vanilla mod. Imo we should create an official vanilla mod if we go this path
- Create a switch somewhere (options, new game screen, etc.) that switches between vanilla/G&K/BNW rulesets created by us

The downside of the second option, is mods. If someone makes an extension mod, that mod can probably only work together with one of these rulesets: some techs might not exist in vanilla, or maybe it adds ideology tenants which only work when BNW is active. This means either mod makers have to store the compatible rulesets for their extension mod, or allow the user to combine possibly incompatible extension mods with the base game. For now I belief that therefore the first option is the better

SomeTroglodyte:
See `enum class BaseRuleset` - there's the preparation for a Vanillin/Grub&Krud switch.
Mods - I think I'm repeating myself: have a compatibleWith: List<String> in modOptions. Every mod without gets BaseRuleset.Default and nothing else. If no base ruleset mod is selected default counts as selected. Then no combination is allowed where neither A lists B as compatible nor vice-versa. Solves this _and_ the two-Mexicoes problem.

yairm210:
> See `enum class BaseRuleset` - there's the preparation for a Vanillin/Grub&Krud switch.
> Mods - I think I'm repeating myself: have a compatibleWith: List in modOptions. Every mod without gets BaseRuleset.Default and nothing else. If no base ruleset mod is selected default counts as selected. Then no combination is allowed where neither A lists B as compatible nor vice-versa. Solves this _and_ the two-Mexicoes problem.

I actually don't remember you suggesting that, but that's _neat_
A simple solution to a complex problem

yairm210:
As for Vanilla / G&K, this was actually the plan from the get go :)
We even have the folder structure and translation generation supporting multiple internal base rulesets (ironically though, the folder hosting the main ruleset is still called "vanilla", since the change was gradual)
But since the tech tree for what was then Vanilla was still constantly changing, we felt it wasn't the time to start branching out.
Now, however, this is viable.

xlenstra:
> As for Vanilla / G&K, this was actually the plan from the get go :)
> We even have the folder structure and translation generation supporting multiple internal base rulesets (ironically though, the folder hosting the main ruleset is still called "vanilla", since the change was gradual)
> But since the tech tree for what was then Vanilla was still constantly changing, we felt it wasn't the time to start branching out.
> Now, however, this is viable.

Alright, I'm not sure this is something I have the time to work on, but I do agree that this would be the best course of action.
Should this PR wait on that to be implemented, or can this be merged and the switch made later?

There are still some questions on how the switch should take place, as most notably vanilla had units with a max of 10 HP and completely different combat mechanics, leading to different strength values being balanced. Many formula's had different numbers, the 'permanent war' mechanic existed instead of 'wary of' mechanic, and a lot of other small changes that are part of vanilla but we have not implemented existed in vanilla. 
Should all these also be switched out? That would likely lead to multiple behaviours being implemented all over the code base and then some ruleset-specific unique being checked to choose which should activate.
To determine what should be done, we should probably at least need something like a `settings.json` or `ModOptions.json` in these base rulesets for choosing which mechanics to activate and some `Constants.json` to make constants in formula's moddable or something. This sounds like yet another major undertaking to fully  implement and at least I don't have the time to worry about all of these details currently with uni.

yairm210:
We'll merge this if it's ready by your standard.
The current ruleset is headed towards G&K at full speed, any Vanilla-retconning we'll do later

yairm210:
And yes, modOptions for the base ruleset

logicminimal:
relates to #4290

PR #5218, opened 2021-09-14
Simplified check for enabled religion and added it in more places

xlenstra:
Adds mod options unique:
- "Disable religion"


SomeTroglodyte:
Yup, perfectionable-expialidocious.

Issue #5168, opened 2021-09-10
Generate translation files generates empty files.


ravignir:
I tried on 2 different devices, always ending up with he same result. Clicking generate translation files in the main menu generates translation files, but they are empty. Furthermore, it removes all strings from already existing files.
![z](https://user-images.githubusercontent.com/30041793/132820131-2fbd700e-6281-475c-bc21-cd814f7dbba6.png)

SomeTroglodyte:
Is that button even _supposed to_ work for mods...? I vote we move it to the "secret" page for starters.

yairm210:
It is supposed to work for mods. This is a bug.

SomeTroglodyte:
Confirmed. Additionally, it shouldn't create a translations folder for mods that don't already have one.

Side note: @ravignir , the system to separate textures into several atlases should work in mods, too - at least the packer. Not sure about the loader, might check later. Rekmod surely could use it.

SomeTroglodyte:
So far: This cannot ever have worked - is the superficial prejudiced impression. 

First it looks for the mod's jsons in the wrong folder. Then it arrives at TranslationFileWriter line 82 and does nothing because `linesFromTemplates` is _always_ empty thanks to line 64. And line 183 (I remember moving that - but it was always the vanilla ruleset) says the unique params will be named without knowledge of the actual mod itself, so in a total overhaul mod it would probably name most unique params "[param]"...

Anyone remember a version where this actually worked as can be expected? I'd like to read the diff...

ravignir:
I actually reported it over a year ago on discord. All I can say is it was not working on 3.10.4.

yairm210:
We could just remove this if it looks like it never worked

SomeTroglodyte:
Oh my. Multiple bugs before `readAllLanguagesTranslation` would even read mods correctly - e.g. one language completely overwriting the last or `createTranslations` using `containsKey` on vanilla instead of the mod...

But nevertheless I _do_ get the feeling this can be done in a few hours.

SomeTroglodyte:
Getting close. Questions:
- I now do write the completionPercentages for mods - I'd wiki that they're for the modder's reading only and can be removed for publication. As a tool category 'why not it's free' -> :+1: or :-1:?
- Only texts actually used in the mod are written, and no language is written unless the file already exists and has at least one `x = ` line.
[...]

PR #5219, opened 2021-09-14
Mod TranslationFileWriter re-work

SomeTroglodyte:
Should solve #5168, see comments over there, including open questions.


core/src/com/unciv/models/translations/TranslationFileWriter.kt
```diff
@@ -99,16 +115,25 @@ object TranslationFileWriter {
...
-                var translationValue = ""
+                // count translatable lines only once
+                if (languageIndex == 0) countOfTranslatableLines++
 
-                val translationEntry = translations[hashMapKey]
-                if (translationEntry != null && translationEntry.containsKey(language)) {
-                    translationValue = translationEntry[language]!!
-                    translationsOfThisLanguage++
-                } else stringBuilder.appendLine(" # Requires translation!")
+                var translationValue = fun (): String {
```
yairm210, line 121: This looks like it was written specifically to annoy me personally.
This is - and I don't use this lightly - barely readable.
The decision tree flow is obscured by the syntax.


Please make a function, call it a name.
Don't use .? , let, or also.
I assure you it will be much more readable.

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -86,58 +86,58 @@ class Translations : LinkedHashMap<String, TranslationEntry>(){
...
         // try to load the translations from the mods
         for (modFolder in Gdx.files.local("mods").list()) {
             val modTranslationFile = modFolder.child(translationFileName)
             if (modTranslationFile.exists()) {
-                val translationsForMod = Translations()
-                createTranslations(language, TranslationFileReader.read(modTranslationFile), translationsForMod)
-
-                modsWithTranslations[modFolder.name()] = translationsForMod
+                val translationsForMod = modsWithTranslations[modFolder.name()]
+                    ?: Translations().also {
```
yairm210, line 98: Again with this... I definitely feel 'if' would make this better

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -86,58 +86,58 @@ class Translations : LinkedHashMap<String, TranslationEntry>(){
...
                 translation.key.getPlaceholderText()
             else translation.key
-            if (!containsKey(hashKey))
-                targetTranslations[hashKey] = TranslationEntry(translation.key)
-
-            // why not in one line, Because there were actual crashes.
-            // I'm pretty sure I solved this already, but hey double-checking doesn't cost anything.
-            val entry = targetTranslations[hashKey]
-            if (entry != null) entry[language] = translation.value
+            val entry = this[hashKey] ?: TranslationEntry(translation.key).also { this[hashKey] = it }
```
yairm210, line 120: I'm sorry to harp in the same subject, but this is a good example.
For once, the .also allows for this to be shorter than the alternative.
But let's consider a hypothetical. Instead of an IDE, where mouseover solves my problems of typing, I happen to be viewing this from - I don't know - Github.
Then I come across this line. also{this...it}.
So wait, what's the 'it' in this case? What's the 'this'?
This is hardly the worst offender, but as an example, it serves

yairm210:
Apart from readability issues, the logic itself seems sound :)

SomeTroglodyte:
Has some similarity with the `Counter` problem: I have a map of collections or other objects supporting some aggregation, and simply want to add to one of the indexed collections. I need to make sure the target exists, though. As this can be seen frequently, and I want it done in the least amount possible of lookups, I came up with this get - elvis - constructor - also pattern, Which feels intuitive to _me_, because I recognize it.

I will rephrase those passages where I went overboard with "kotlin-y-elegant", no problem, and sorry. 

But in the mean time - any opinion on that question in 5168 on the problem where the mod references an unique, which is pulled into the strings to translate, then written with the translation pulled from vanilla into the mod's with a comment that it's a copy? This PR copies and comments, but in the next run it won't recognize the difference anymore and re-output them without the comment. I found that quite manageable, and could be useful if a mod wants to translate a bit differently to express some kind of scenario-like style. But I would also guess it quite likely others might prefer a different behaviour: Not writing those at all, or keep the comment by testing whether the mod's translation is still identical to the base game translation, or maybe some other way entirely.

yairm210:
You're right, this is quite a dilemma.
On the one hand, there *are* mods that overwrite existing translations, e.g. Deciv's Gold = Water. Any translation that already exists in the mod's translation files should be preserved.
On the other hand, copying *all* translations would make e.g. Rekmod's translation files *huge*. There are ~5k lines in Unciv's translation files, and that would clog up the mod's files.
What's worse is that the mod won't benefit from translation updates at all - once the lines have been copied over the first time, that's it, we now forever use that version of the translation!

All in all - don't copy, but do preserve overloads.

SomeTroglodyte:
Working as intended now, but I tested an idea for an extension that handles the "gimme a map entry, even if you have to create it first" pattern. I guess I'll remove it, because that _would_ be nice but better if we applied it throughout Unciv. It is:
```kotlin
/**
 * Get an entry from a hashmap or, if it is missing, create a new one, store it, and return that.
 *
 * @param createNew factory lambda used when a new instance<V> is needed
 */
inline fun <K, reified V> HashMap<K, V>.getExistingOrNew(key: K, createNew: ()->V): V {
    var entry = get(key)
    if (entry != null) return entry
    // kotlin weakness can't instantiate V: https://youtrack.jetbrains.com/issue/KT-6728
    entry= createNew.invoke()
    set(key, entry)
    return entry
}
```

Using that on your critiques 2 and 3 makes them really concise one-liners...
```kotlin
                val translationsForMod = modsWithTranslations.getExistingOrNew(modFolder.name()) { Translations() }
instead of
                var translationsForMod = modsWithTranslations[modFolder.name()]
                if (translationsForMod == null) {
                    translationsForMod = Translations()
                    modsWithTranslations[modFolder.name()] = translationsForMod
                }
(and possibility for non-null inference fails)
```

SomeTroglodyte:
`Placeholder `[]` not found` ... interesting!!! How come - I thought the TranslationFileWriter wasn't changed while I wasn't paying attention??? Let's see... The actual files _are_ OK, the test must be broken...

PR #5220, opened 2021-09-14
Initial unique compliance detection - buildings only.

yairm210:
This is the first level of compliance detection.
Using the translation files, where all existing unique formats should be stored, we can root out bad uniques when mod checking.
Furthermore, using the auto-generated *values* for placeholders, we can detect bad *values* of placeholders for uniques.

All of these should be warnings for several reasons:
A. When deprecating a unique, since its value  will not be used by the base game anymore, its translation will be auto-removed the next version. However, it will still work. Having the deprecated unique show up as a warning is a good thing in keeping mod uniques up to date, but this shouldn't stop the mods from working.
B. Some mod uniques are flavor text. This is absolutely fine!
C. Some uniques are special and used for filtering. For example, you can add a "Magic" unique to units so that you can catch them with a "[Magic] units" unitFilter.

Asking for your thoughts on this. Once the initial form of the compliance checking is in place, expanding it to include all other unique-containing ruleset objects, and expanding the placeholder checks to include e.g. Stat and Stats, should be simple enough.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1242,3 +1242,4 @@ This Unit upgrades for free =
 [stats] when a city adopts this religion for the first time = 
 Never destroyed when the city is captured = 
 Invisible to others = 
+Will not be displayed in Civilopedia = 
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -98,8 +98,8 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
         Gdx.graphics.isContinuousRendering = settings.continuousRendering
 
         thread(name = "LoadJSON") {
-            RulesetCache.loadRulesets(printOutput = true)
             translations.tryReadTranslationForCurrentLanguage()
+            RulesetCache.loadRulesets(printOutput = true) // needs to be after translation loading, for the ruleset checks
             translations.loadPercentageCompleteOfLanguages()
             TileSetCache.loadTileSetConfigs(printOutput = true)
 
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -3,6 +3,7 @@ package com.unciv.models.ruleset
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
 import com.unciv.JsonParser
+import com.unciv.UncivGame
 import com.unciv.logic.UncivShowableException
 import com.unciv.models.Counter
 import com.unciv.models.metadata.BaseRuleset
@@ -18,6 +19,7 @@ import com.unciv.models.stats.INamed
 import com.unciv.models.stats.NamedStats
 import com.unciv.models.stats.Stat
 import com.unciv.models.stats.Stats
+import com.unciv.models.translations.TranslationFileReader
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.colorFromRGB
 import kotlin.collections.set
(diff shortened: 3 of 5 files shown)


xlenstra:
Some remarks about the idea and not the code, take these as you will, this is your project after all and you can decide what is the way forward.

Taking the parameters of uniques out of the translation files has a few disadvantages:
- Some uniques have `[param]` as a parameter, either due the options not being checked in the TranslationFileWriter (e.g. `Earn [unitFilter] units [param] as [Stat]`, where `param` can be either `Strength` or `cost`), the TranslationFileWriter not understsanding them (e.g. `[{Miliatry} {water}] units`) or more than one type of object being allowed (e.g. `[stats] from every [param]`). These can only be checked by making specific exceptions for these
- Maybe we implement a unique to accept multiple types of inputs, such as we did for the `[stats] from every [param]`, but we only implement use them ourselves for a single type. Then the TranslationFileWriter would assume that that is the only type of object we can use, and taking things out of the translation would give out warnings for uniques that are perfectly fine
- If we implement a unique on request that we don't use ourselves and forget to add to `template.properties`, using it will give warnings, even though it will work, potentially confusing both mod makers and users

Showing a warning for every unknown unique will not only tell the mod maker but also the user that something is wrong. If a user uses a mod where half of the units have a `[Magic]` unique, his entire screen will be filled with "Warning: unknown unique `[Magic]` for unit 'bla'", which might scare the user off, even if the mod itself works perfectly. We can't detect what uniques are typo's and what are used as identifiers based only on the translation files, so we either require additional input (an `identifierUniques` field in the ModOptions?), potentially scare users off or can't generate warnings for missing uniques at all.

I may be biased as it was partly my idea, but implementing uniques as in #5091, would allow for a standardized place to note these parameters, namely the Enum containing all base uniques. This would solve the first three problems, though the last still remains. In that case it's also possible to display warnings when mod makers use deprecated uniques by setting a flag for these uniques in the enum (or something similar), though that would not detect typo's.

yairm210:
But that's the main problem, isn't it?
How can you differentiate between "Magic" as an "ok" non-standard unique, and an unintentional misspelling?

I agree that we may not want to display all potential problems to the user, that's something week need to solve independently of enum or not

I'm not entirely sure what you're suggesting.
An enum of uniques is fine, but each one will need to have a list-of-lists for acceptable types (multiple types for multiple params), is that what you're going for?

xlenstra:
> But that's the main problem, isn't it?
> How can you differentiate between "Magic" as an "ok" non-standard unique, and an unintentional misspelling?

You can't, I agree, and that is the problem. But isn't that just using the same data field for two different purposes? Maybe it would be better off split into two fields: `uniques` for unique abilities, and `idetifiers` for these kind of identifiers.
Alternatively, we could add a list somewhere of all uniques that are used solely as identifiers, and ignore those when checking uniques for validity.

I must admit that I also hadn't considered this for the Enum, as that will likely run into the same problem later down the line

> I agree that we may not want to display all potential problems to the user, that's something week need to solve independently of enum or not
>
> I'm not entirely sure what you're suggesting.
> An enum of uniques is fine, but each one will need to have a list-of-lists for acceptable types (multiple types for multiple params), is that what you're going for?

That would be the obvious implementation, probably hidden behind some data class, but I do agree that that feels somewhat cumbersome and I hadn't thought out the details on this yet. Using conditionals would also help here, as it will often split up uniques into parts with often only one or two parameters, and each of these will almost always accept only one type.

Now that I've thought more about it, it is clear that the enum isn't also the obvious silver bullet to these problems, though it does make for a standardized place to save the information on the parameter types. I don't know how useful that would be, as reading in jsons is basically the only place where that would matter, so maybe storing it somewhere else in a similar data format would make sense.

yairm210:
I do think the enum is a good idea, though.
It could serve as the true "source of truth" for what is and is not acceptable in uniques, what uniques can go where, and therefore also both highlight limitations in the current uniques, state deprecations, and serve as the datasource for autobuilding the "uniques" wiki page as a standalone webpage :)

SomeTroglodyte:
Enum or json-fed class with a usage close to enum. The important thing is that at run time the check for one is fast. That would mean immutable instances and === comparisons, or at least short-ish string "keys" needing no further parsing.

As for magic - require mods to declare them and nail it down that such uses cannot have parameters, as no code supports them. Or scan and count, and only complain about unknown strings occurring only once as singletons can not really have a meaning and are more likely to be typos? No, I like mandatory declaration better.

> what uniques can go where

Exactly. With current uniques one of the finer points irritating me most is a lack of clarity of _scope_: What will the effect(s) apply to, what are/can be the targets of conditions, and on which kinds of objects can a unique be declared? Having all that in a master table could help enormously with usage, or implementing new ones, or finding bugs.

SomeTroglodyte:
And when Scope has a runtime representation, plus the unique master table _declares_ what parameters a unique will parse, that checker code can be _one_ loop for all ruleset object types, finally refactoring that one inelegant piece in CivilopediaScreen - "how do I enumerate all ruleset objects" still uses its own master list, that could be a general utility.

yairm210:
Closing this in favor off rebuilding for the new unique enum

PR #5222, opened 2021-09-14
Added basic functionality for uniques enum

yairm210:
As discussed in #5220

This contains:
- Stating what parameter types are acceptable for each enum
- Auto-detection of parameter types by unique parameter text, with optional overloading for special params
- Compliance checks for strings against each parameter type

The only disadvantage that I see is that when working on a unique, you don't see the unique text "right in front of you", as it's hidden behind the enum, as captured nicely in this convenient side-by-side comparison:

image

This can be slightly improved by adding the text to the comment on the enum so it'll be ctrl+q visible, but if you're going to press extra buttons you may as well just go to the source.

But since we're discussing code readability, next to the *disadvantage* of less immediate readability there's the obvious advantage of confirmed *eventual* readability. I chose the Consumes unique because it appears in multiple places. and having them all available at a click is really useful.

Another advantage is that unlike the placeholderText currently used in the code, which does not inform you of the typings (so you have to guess from context and use), this method ensures that all parameter types are named.


core/src/com/unciv/models/ruleset/Unique.kt
```diff
@@ -14,16 +14,76 @@ import com.unciv.models.translations.hasPlaceholderParameters
...
+    ConsumesResources("Consumes [amount] [resource]");
+
+    /** For uniques that have "special" parameters that can accept multiple types, we can override them manually
+     *  For 95% of cases, auto-matching is fine. */
+    val parameterTypeMap= ArrayList<List<UniqueParameterType>>()
+    init {
+        for (placeholder in text.getPlaceholderParameters()) {
+            val matchingParameterType =
+                UniqueParameterType.values().firstOrNull { it.parameterName == placeholder }
+            if (matchingParameterType != null) parameterTypeMap.add(listOf(matchingParameterType))
```
SomeTroglodyte, line 56: Prime Elvis the King example: `parameterTypeMap.add(listOf(matchingParameterType ?: UniqueParameterType.Unknown))` - even a oneliner dropping matchingParameterType could be readable... Just teasing.

xlenstra:
I generally agree with the direction this is going, but again I have a few minor points, mostly on performance :)
- Currently we still do string comparions to determine the value of a unique, couldn't we just find the corresponding enum of its type and cache that in the class instead? Enum comparisons are likely faster than string comparisons
- If we do this, in the future we can maybe transition into a situation where unique parameters are checked when the ruleSet is loaded instead of when using the unique, removing the need for the `matches` function. In general I'd assume we compare uniques more often than that there are uniques in the ruleset, so that would increase performance.
- Minor point that currently has no effect and I'd assume that this will be added later on anyway, but the current list of unitFilters is incomplete

core/src/com/unciv/models/ruleset/Unique.kt
```diff
@@ -14,16 +14,76 @@ import com.unciv.models.translations.hasPlaceholderParameters
...
+
     /** This is so the heavy regex-based parsing is only activated once per unique, instead of every time it's called
      *  - for instance, in the city screen, we call every tile unique for every tile, which can lead to ANRs */
     val stats: Stats by lazy {
         val firstStatParam = params.firstOrNull { Stats.isStats(it) }
         if (firstStatParam == null) Stats() // So badly-defined stats don't crash the entire game
         else Stats.parse(firstStatParam)
     }
+
+    fun isOfType(uniqueType: UniqueType) = placeholderText == uniqueType.placeholderText
```
SomeTroglodyte, line 84: Could be performance-critical. (loud thinking) Can we avoid the char-by-char equality? Intern both uniqueType.placeholderText and Unique.placeholderText once and early and then use `===` here? Or cache the UniqueType per Unique? Both won't be good if Unique objects are short lived - are they?

SomeTroglodyte:
Overlap. Yeah, "Enum comparisons are likely faster than string comparisons" - because Enum instances are immutable (or should be? I hope kotlin will barf if I add a var), the compiler knows that and so it will compile a `===` or something close.

> removing the need for the matches function

And parameters could be pre-compiled so you wouldn't need toInt(), Stat.parse or anything the like where a unique is actually implemented.

yairm210:
Saving the enum in the Unique would make comparisons faster, totally :)
Precompiled values... Ehhhhh. I thought if that, but that would mean saving them as an Any type, and I inherently don't trust those.

yairm210:
As for incompleteness, indeed :)
I wanted to see if we were all in the same page before going whole hog

SomeTroglodyte:
> saving them as an Any type

Java is full of them - no, I wanted to say, sparse typed arrays. 8 params, one of them an Int, at position 3. paramsAsInt = Array<Int>(8){0} or so, only assign paramsAsInt[3]=parameters[3].toInt(), getIntParam(index) = paramsAsInt[index] - for Stats use null as the sparse default. Only pre-parse the more common/costly types - Int and Stats may be enough. (Did I express my dislike of all those uniques expressing 1 Gold as [amount] [stat] instead of as [stats]?)

yairm210:
That sounds like "more trouble than it's worth" category. I'd need to see a timing test to be convinced that creating the arrays doesn't require more time than it saves.

yairm210:
Well, since no comments have been forthcoming, in it goes

PR #5224, opened 2021-09-14
Carthage civ

SimonCeder:
This PR adds Carthage to the game.
Added uniques:
* "Land units may cross [] tiles after the first [] is earned"
* "Units ending their turn on [] tiles take [] damage" (All civs by default take 50 damage from mountains. Adding the unique "Units ending their turn on a Mountain take 50 damage", as in this case, is purely cosmetic and does nothing.)
* "Units ending their turn on this terrain take [] damage" (for Terrains)

carthage

After a bit of tinkering, the AI does a decent job at hopping across mountains when it needs to while avoiding taking too much damage from them.


android/assets/jsons/Civ V - Vanilla/Terrains.json
```diff
@@ -83,6 +83,7 @@
 	{
 		"name": "Mountain",
 		"type": "Land",
+		"movementCost": 1,
```
yairm210, line 86: This is default, no need :)

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -879,30 +879,40 @@ class MapUnit {
...
+            "Our [$name] took [$tileDamage] tile damage",
+            currentTile.position,
+            name
+        )
+    }
+
+    fun getDamageFromTerrain(terrainName: String = currentTile.baseTerrain): Int {
+        var totalDamage = 0
+        if (terrainName == Constants.mountain)
+            totalDamage += 50   // The norm
```
yairm210, line 904: If we're renovating all this fun stuff, this should be a unique as well.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -881,6 +896,12 @@ class CivilizationInfo {
             ?: return null
         if (unit.isGreatPerson()) {
             addNotification("A [${unit.name}] has been born in [${cityToAddTo.name}]!", placedUnit.getTile().position, unit.name)
+            for (unique in getMatchingUniques("Land units may cross [] after the first [] is earned")) {
```
yairm210, line 899: Uniques with tileFilters should be formatted "[tileFilter] tiles" for readability - also relevant to the "ending their turn" unique

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -622,6 +632,11 @@ class CivilizationInfo {
             cityInfo.civInfo = this // must be before the city's setTransients because it depends on the tilemap, that comes from the currentPlayerCivInfo
             cityInfo.setTransients()
         }
+
+        passThroughImpassableUnlocked = passableImpassables.isNotEmpty()
+        // All civs normally take 50 damage on a mountain. Cache whether this civ is an exception for performance reasons.
+        nonStandardTerrainDamage = getMatchingUniques("Units ending their turn on a [] take [] damage")
+            .any { it.params[0] != Constants.mountain || it.params[1].toInt() != 50 }
```
yairm210, line 639: Not sure how relevant this really is...

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -669,8 +669,8 @@ class MapUnit {
             }
         }
 
-        getCitadelDamage()
-        getTerrainDamage()
+        doCitadelDamage()
+        doTerrainDamage()
```
yairm210, line 673: Good renames!

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -879,30 +879,40 @@ class MapUnit {
...
+            name
+        )
+    }
+
+    fun getDamageFromTerrain(terrainName: String = currentTile.baseTerrain): Int {
+        var totalDamage = 0
+        if (terrainName == Constants.mountain)
+            totalDamage += 50   // The norm
+        if (civInfo.nonStandardTerrainDamage) {
+            for (unique in getMatchingUniques("Units ending their turn on a [] take [] damage")) {
```
yairm210, line 906: The way the unique is phrased, it sounds like "X damage" and not "X extra damage", so totalDamage =  and not totalDamage +=, no?

core/src/com/unciv/logic/map/MapUnit.kt
SimonCeder: The thing is, 50 damage from mountains is actually not unique to Carthage at all but rather shared for all civs, relevant since any civ can enter a mountain with a helicopter. Hence the half-implemented unique.
If we want to get rid of the hardcoding, we should probably add a `damagePerTurn` field to Terrains. The implementation of the unique `"Units ending their turn on a [] take [] damage"` would then be to overwrite this amount of damage for that civ. Meaning Carthages unique in this case will still be effectively meaningless since it means it will take 50 damage instead of 50 from mountains. So we can still set the `nonStandardTerrainDamage` to false for Carthage on base ruleset and save a bit of time in pathfinding.
The other alternative to get rid of hardcoded terrain damage would be to give _every_ civ a `"Units ending their turn on a [Mountain] take [50] damage"`, but that seems very inelegant.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SimonCeder: It's to avoid doing string comparisons during pathfinding.

SimonCeder:
Added `damagePerTurn` to the Terrain class so it's possible to change the values around (and even have healing terrains).
The damage from tiles is overridable on a civ basis by using the unique `"Units ending their turn on [] tiles take [] damage"`.

android/assets/jsons/Civ V - Vanilla/Terrains.json
```diff
@@ -83,9 +83,9 @@
 	{
 		"name": "Mountain",
 		"type": "Land",
-		"movementCost": 1,
 		"impassable": true,
 		"defenceBonus": 0.25,
+		"damagePerTurn": 50,
```
xlenstra, line 88: As this an 'ability' only a handful (namely one) of terrains have, I'd personally made this a unique instead of another field

SimonCeder: Alright but I still think it should be stored as an Int in the actual class since it's accessed often.

SimonCeder:
Another question, regarding `getShortestPath`.
I wrote in a recursive call so the function is called twice, first avoiding damaging terrain, and if no path is found, then again normally. This only happens if the civ has `passThroughImpassableUnlocked` to shave off a bit of time if the civs aren't Carthage. The hack is necessary for Carthage not to kill their units on mountains.
But it's possible that a mod will add a damaging terrain that everyone is able to enter. Should `getShortestPath` always do this double run to cover this eventuality?

android/assets/jsons/Civ V - Vanilla/Terrains.json
xlenstra: You can always add a `terrain.setTransients()` function called from `tile.setTransients()` or `map.setTransients()` depending on what already exists

yairm210:
getShortestPath is a heavy function as is, doubling it's runtime is definitely not something to do "just in case"
Looks like png rebuild and we're good to go!

SimonCeder:
I would just delete it from the PR but I don't know how. I'll overwrite it with the game.png from master when I get home.

SomeTroglodyte:
"delete" from a branch once it's online -> "Revert" the change locally (several options) and push that. File should disappear from a PR diff. Actually merging then repacking the atlas is just as easy or complicated. If you delete the online branch first, so the local branch knows (meaning if deleted online you need a fetch) it's not tracking an online one, you have more freedoms, and can "undo" as well. But - deleting a branch autocloses any PR associated with it.
:nerd_face: 

There are surely other options - e.g. I have never learned what exactly "drop commit" does. Undo is nice because it puts the changes back into a Changelist.

Check your commits that they contain exactly what you expect. Looking at the local Changelist (Alt-9, local changes tab) can be a little easier to use than the Ctrl-K window, sometimes. If you accidentally commit a file/change you didn't want to, undo and rollback/shelve. Commit messages can be edited without undoing. And so on... Oh, and my scratchpad has a section on how my working directory is *not* the repo clone's assets folder but a partial copy, partial symlink - I _cannot_ accidentally commit an atlas/texture change I didn't actually contribute to thanks to that, same with .gitignore (better use .git/info/exclude anyway) or any other file I want independent.

PR #5226, opened 2021-09-15
Unique enum compliance detection

yairm210:
Dependent on and extension to #5222 

Unique errors can be divided into 3 sorts - 
  - Outright ruleset-invariant errors, like inputting "Prodction" for a stat, or "fifty" for a number.
  Ruleset-invariant errors should be checked and warned for all mods
  - Ruleset-dependent but still outright errors, like inputting a resource that doesn't exist as a requirement for a building.
  Ruleset-dependent errors should be checked and warned for base ruleset mods (this includes "combined" mods, base + extras)
  - Warnings - these are what we discussed in #5220, that some filters can basically accept whatever. We *do* want to warn the mod creators if we don't recognize the parameter, but this is not necessarily an error, and so shouldn't be user-visible.
  We're not sure what to do about these for now, but they'll become relevant.

If multiple parameter types are acceptable (can be unitname OR buildingname), if any check is error-less, there's no error.
Otherwise, take the least severe error.


yairm210:
For now, check only the latest commit, the rest are just part of #5222 :)

yairm210:
No comments, anyone?

xlenstra:
My feeling two days ago was that it looks fine, and looking at it again I still feel the same. Only thing that might be difficult in this framework is extending the `UniqueParameterType` to accept `[param]` parameters

SomeTroglodyte:
Limited capacity at the moment, and for this I would want to be able to take my time and read it awake and concentrated.

Well, OK, just superficially:
`parameterName:String` - style guide
`Unknown("")` - that param name is meant for the translation writer? If yes - wouldn't the "param" default be better? Could mean trouble if a placeholder and a filled-out one were indistinguisheable - maybe.
`data class UniqueComplianceError` - will we actually use the extras `data` offers - free equals contract, deconstruction or copy()? If not, data is possibly just dead code and a way to block some class syntax.
`UniqueComplianceErrorType` - close to error 'severity' in other systems. Worth planning for more-dimensional error types? Axis "warn or bail" could be independent of "when can this be checked" someday? `ordinal` _is_ used as severity - better document in the class? Just food for thought.

Not even 20% scanned and my eyes are shutting down - sorry. So, what Xander said :smiley_cat:  - just plan this for shortly after a tag release, so we have ample time to play with it before the next.

yairm210:
2 things:
A. The reason that each parameter in UniqueType is so we can handle the "params", if we do our job sell there will be no "params" :)
B. Instead of having to manually add the various types in a param that can accept e.g. both tech name and policy name ("until [x]"), *this too* can be parsed from the unique text by having it slash-delimited, like "[stats] until [policyName/techName]" -> param.split("/").map {getParamTypeByName(it)} gets you "policyName/techName"->listOf(ParamType.policyName, ParamType.techName)

PR #5234, opened 2021-09-15
Implemented Unit Supply by r3versi, not by me

SomeTroglodyte:
For all info and discussion, see #4913


SomeTroglodyte:
Of course, the magnifying glass some /&%"§/%!" wanted in their mod sort&search PR 'broke' this. Verified, both r3versi's exclamation mark and that humpty-dumpty's magnifying glass are in the texture. But there's a new far-too-large icon - a loudspeaker at more than 270x250 px... Did I do a loudspeaker somewhere????

SomeTroglodyte:
Ouch, I *did* forget to scale that one down.

yairm210:
All else can be dealt with later
For now this is merged :)

PR #5235, opened 2021-09-15
Update Swedish.properties

SimonCeder:


xlenstra:
I'd reckon you already found yourself, but for completeness:
    Placeholder `[dimensions]` not found in `Swedish` for entry `Invalid map: Area ([area]) does not match saved dimensions ([dimensions]).`

PR #5236, opened 2021-09-15
Jungles on hills

ArchDuque-Pancake:
I hope there is no problem...

Based on fandom:
imagen


ravignir:
there is a problem. You need to make sure jungles don't spawn on tundra hills.

ravignir:
MapGenerator.kt line 419
```
    private fun spawnVegetation(tileMap: TileMap) {
        val vegetationSeed = randomness.RNG.nextInt().toDouble()
        val candidateTerrains = Constants.vegetation.flatMap{ ruleset.terrains[it]!!.occursOn }
        //Checking it.baseTerrain in candidateTerrains to make sure forest does not spawn on desert hill
        for (tile in tileMap.values.asSequence().filter { it.baseTerrain in candidateTerrains
                && it.getLastTerrain().name in candidateTerrains }) {
            val vegetation = (randomness.getPerlinNoise(tile, vegetationSeed, scale = 3.0, nOctaves = 1) + 1.0) / 2.0

            if (vegetation <= tileMap.mapParameters.vegetationRichness)
                tile.terrainFeatures.add(Constants.vegetation.filter { ruleset.terrains[it]!!.occursOn.contains(tile.getLastTerrain().name) }.random(randomness.RNG))
        }
    }
```
That function needs a rework.

ArchDuque-Pancake:
Ok, I'll take a look if I can do it. I have a space this afternoon NY time.

ArchDuque-Pancake:
My lord, this is embarrasing, now I'm making the "This dumb guy is trying to do something that he has no idea about"...

I have looked at this for some time and dont find any logic or clue to what I have to do to not let jungles spawn in desert and tundra. 
Sorry but, I did not think this far, honestly I tought it would be easy but here I am.
 Never again.

SomeTroglodyte:
Hey, no reason to insult yourself :wink:
Conceptually we could do a unique allowing that. Plus, Jungles should have a correlation with humidity, too, methinks - which isn't checked or even available in spawnVegetation yet either, as can readily be seen in the code snippet shown above. So, no way to do it json-only yet, needs coding.

ravignir:
> 
> 
> Hey, no reason to insult yourself 😉
> Conceptually we could do a unique allowing that. Plus, Jungles should have a correlation with humidity, too, methinks - which isn't checked or even available in spawnVegetation yet either, as can readily be seen in the code snippet shown above. So, no way to do it json-only yet, needs coding.

Well, that is A BIT harder to figure out... I really wish I had more time. I guess `applyHumidityAndTemperature` function has to be split into two lesser functions (generating humidity and temperature | applying base terrains) and then `spawnVegetation` needs to be changed accordingly.(?)

SomeTroglodyte:
Ah, reading this after mewling on #5247 - sorry. No, temp and humidity need to stay together due to the unique formulation - as I see it. Just refactor from `if (noTerrainUniques)` on with a helper fun for the fallback and a then much smaller tile.baseTerrain = when { fallback/match/grassland } so the setTerrainTransients at loop end is always used, then do vegetation right there? ...???

SomeTroglodyte:
>  I really wish I had more time

... I really wish I had ***less*** time - for Unciv - as that would mean I would be spending the time - e.g. outdoors?

yairm210:
Closing because #5247 is the viable replacement

PR #5237, opened 2021-09-15
Update and generalize Siam UP

SimonCeder:
This PR changes the first Siam unique from `"Food and Culture from Friendly City-States are increased by 50%"` to a generalized `"[]% [] from City-States"` and updates it to also include Faith as mentioned in #4602 .
I also added a function to the Diplomacy Screen to display the correct modified values if playing as a civ with one of these uniques, so you can see what you're getting.
While I was at it I uncommented `"preferredVictoryType":"Diplomatic"` from Siam and England who
 were the last two to have that line commented.

Screenshot_2021-09-15_23-01-01

Edit:
Also refactors the city state bonuses into Unique using the new UniqueType system. Added the utility function `undefinedCityStateBonuses()`.


core/src/com/unciv/models/stats/Stats.kt
```diff
@@ -112,15 +112,26 @@ open class Stats(
...
+            else -> {
+                production *= number
+                food *= number
+                gold *= number
+                science *= number
+                culture *= number
+                happiness *= number
+                faith *= number
+            }
+        }
```
xlenstra, line 134: When doing it for a just single stat, can't you just do `Stat.Production *= number`?

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -223,6 +231,52 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
         return diplomacyTable
     }
 
+    /** Given a list of city-state bonuses, returns the list with updated values for Siam-like uniques */
+    private fun getIncreasedBonuses(bonuses: List<String>): List<String> {
+        val newBonuses = ArrayList<String>()
+        println(bonuses)
```
xlenstra, line 237: This looks like debug printlns :) (also one at the bottom of this function)

xlenstra (commented):
Other than those two things: looks good!

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
SomeTroglodyte: Not without context in the output _and_ an if(const val)

core/src/com/unciv/models/stats/Stats.kt
SomeTroglodyte: :scream: Polished performance lost to a conditional. `fun timesInPlace(stat: Stat? = null, number: Float) = if (stat == null) timesInPlace(number) else this[stat] *= number` - at least most existing calls won't hit this overload. Untested though.

SomeTroglodyte: Moot anyway as Stat.valueOf(unique.params[1]) will crash instead of returning a null

yairm210:
To navy officer

yairm210:
City state bonuses *desperately* need to be converted to Uniques at the source
We're passing them like 10 times currently

SimonCeder:
The division into `friendBonuses` and `allyBonuses` kind of would require a parallel process though.
If you agree, I think we might be able drop those fields and put everything in `uniques` in Eras.json eg
```
uniques: ["[Cultured] city-states [When Friends]: {Provides [3] [Culture] per turn}",
          "[Cultured] city-states [When Allies] : {Provides [6] [Culture] per turn}",
          ...
```
While slightly less readable it would fit right into the current unique system, and remove the need for duplicated code in a couple places.

yairm210:
> The division into `friendBonuses` and `allyBonuses` kind of would require a parallel process though.
> If you agree, I think we might be able drop those fields and put everything in `uniques` in Eras.json eg
> 
> ```
> uniques: ["[Cultured] city-states [When Friends]: {Provides [3] [Culture] per turn}",
>           "[Cultured] city-states [When Allies] : {Provides [6] [Culture] per turn}",
>           ...
> ```
> 
> While slightly less readable it would fit right into the current unique system, and remove the need for duplicated code in a couple places.

We Can have 2 sets of Uniques like the two sets of strings, no problem

yairm210:
Much easier than parsing those kinds of strings, and we'll need to save them somewhere anyway

SimonCeder:
Here's a proposal.
The bonuses are objectified into the new `CityStateBonus` class. Since the number of possible bonuses are so small, I went ahead and sorted the possible string placeholders into a new enum class, similar as far as I understand to #5222. The class also has useful fields such as `amount` and `cityFilter` rather than params[1] or whatever. Also added the utility function `undefinedCityStateBonuses()`. This all makes the city-state code much neater.
Too fancy?

core/src/com/unciv/models/ruleset/Era.kt
```diff
@@ -46,3 +75,33 @@ class Era : INamed, IHasUniques {
     /** This is used for display purposes in templates */ 
     override fun toString() = name
 }
+
+class CityStateBonus(val text:String) {
```
yairm210, line 79: Why do we need another class for this, this is just Unique again
And if you want enum recognition, there's a new method for that as well with UniqueType

SimonCeder:
Alright, too fancy it is.
I refactored it all into Unique. I also rewrote the bonus strings from `[amount] [stat]` to `[stats]` so they'll work with `Unique.stats`.
I _did_ also write a new function for `Stats`, `toStringWithDecimals()` so we can get those nice "+4.5 Culture" strings for Siam.

SomeTroglodyte:
Just food for thought, in no way expressing a negative view of code quality:
- Seems like you could be running behind the Unique transformation train some more- you're using the enum but not throughout?
- Seems toStringWithDecimals is untranslateable - wait, are you assuming that Stat-toString override I mentioned as idea is already in? It isn't.
- Happiness is also a Stat. I guess there's a reason that hasn't made the same move to [Stats].

SimonCeder:
> Just food for thought, in no way expressing a negative view of code quality:

No need to apologize - I appreciate your reviews.

>     * Seems like you could be running behind the Unique transformation train some more- you're using the enum but not throughout?

I thought that "Provides a unique luxury" is so rare that it would be clogging up the enum for no good reason. But then enums can be as long as an Int right? Added it.

>     * Seems toStringWithDecimals is untranslateable - wait, are you assuming that Stat-toString override I mentioned as idea is already in? It isn't.

Good catch. For some reason I assumed that since "Culture" is translatable "+3 Culture" would be, but that's not the case. I was worried about a hypothetical double translation. Let's say there is a language in which "Culture" is called "Food". If you translate "Culture" first separately and then as a part of a string you would get a strange result! Added a `.tr()` to the function.

>     * Happiness is also a Stat. I guess there's a reason that hasn't made the same move to [Stats].

The main reason is that "Provides [+2 Happiness] per turn" would sound strange to me, while the other stats don't. Also Happiness is treated separately in CivInfoStats so it's somewhat practical.
..we could drop the " per turn" from the other bonuses, so it's more unified. We could even go for the minimalist `"Provides [+3 Faith] per turn"` -> `"[+3 Faith]"`. But that's perhaps the matter for another PR, maybe after a poll.

SomeTroglodyte:
> But then enums can be as long as an Int right

Close, but maybe a misunderstanding: In my b-box it says: As "long" as any Instance key - whatever that is on the platform - I imagine a CPU address but that's a prejudice too as who says the JVM needs to operate with the same addressing scheme as its host. Int as in CPU register word width because the enum value's _ordinal_ is relevant is C/C++ thinking. Enum as bitmasks is C/C++ thinking. In java/kotlin, the ordinal is more like a glued-on property, the individual enum "value" is an instance like any other. You can have as many of them as the VM can instantiate objects/classes, but sharing that pool with everything else...

See my perverted PR 5251 and its relation IRejectionReason to RejectionReason. Or look at EnumSet - very different kettle of fish than C++ bitmask flags enums.

> double translation

We have that problem, true. I remember catching one or two of these. Problem is - normally you get no visible consequences, but as you said it's not impossible. But a test is difficult to construct, too... Let's begin with more Kdoc stating where exactly translation is supposed to happen in some call chain?

core/src/com/unciv/models/ruleset/Era.kt
```diff
@@ -22,12 +27,36 @@ class Era : INamed, IHasUniques {
...
+        for (pair in bonusMap) {
+            objectMap[CityStateType.valueOf(pair.key)] = pair.value.map { Unique(it) }
+        }
+        return objectMap
+    }
+
+    fun getCityStateBonuses(cityStateType: CityStateType, relationshipLevel: RelationshipLevel): List<Unique> {
+        return when (relationshipLevel) {
+            RelationshipLevel.Ally   -> allyBonusObjects[cityStateType]!!
+            RelationshipLevel.Friend -> friendBonusObjects[cityStateType]!!
```
yairm210, line 51: Consider that for mods this may be null, and return a toList() instead

yairm210: emptyList, sorry
You can use the Elvis operator ?: here, this is a good use

core/src/com/unciv/models/ruleset/Era.kt
```diff
@@ -22,12 +27,36 @@ class Era : INamed, IHasUniques {
...
+    val friendBonusObjects: Map<CityStateType, List<Unique>> by lazy { initBonuses(friendBonus) }
+    val allyBonusObjects: Map<CityStateType, List<Unique>> by lazy { initBonuses(allyBonus) }
+
     var iconRGB: List<Int>? = null
     override var uniques: ArrayList<String> = arrayListOf()
     override val uniqueObjects: List<Unique> by lazy { uniques.map { Unique(it) } }
 
+    private fun initBonuses(bonusMap: Map<String, List<String>>): Map<CityStateType, List<Unique>> {
+        val objectMap = HashMap<CityStateType, List<Unique>>()
+        for (pair in bonusMap) {
```
yairm210, line 42: for((cityStateType, bonusList) in bonusMap) ...
Slightly more readable imo

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -223,6 +229,90 @@ class DiplomacyScreen(val viewingCiv:CivilizationInfo):CameraStageBaseScreen() {
...
+                }
+            }
+            // No matching unique, add it unmodified
+            if (!improved)
+                bonusStrings.add(bonus.text)
+        }
+        return bonusStrings
+    }
+
+    /*private fun getAdjustedBonuses(bonuses: List<CityStateBonus>): List<String> {
```
yairm210, line 275: And remove dead code :)

yairm210:
Okay, almost there!
Good work!

PR #5239, opened 2021-09-15
Arabic.properites

Kleznam:
This is an attempt to see if unciv can make Arabic work, but I only translated 60 lines, and iam afraid that I might have made one or to or more coding errors,so please if anyone sees any mistakes please correct me.and thanks


SomeTroglodyte:
We expressly state RTL languages are not supported in the wiki, and I just considered a while ago to amend that to 'you're welcome to try but don't expect text to flow or align properly'... I'm curious, do you just invert the string so the beginning of a sentence is to the right, maybe there are even tools for that?

Kleznam:
Mission failed we will get em next time

PR #5240, opened 2021-09-16
Update German translation

Mape6:


SomeTroglodyte:
Still `Could not GET 'https://jcenter.bintray.com/com/github/anuken/packr...` - page gives a 403 forbidden. Probably because jcenter  is _scheduled_ to die an ignoble death. :coffin: 

@yairm210  - https://mvnrepository.com/artifact/com.badlogicgames.packr/packr/1.2 ?

Didn't we yet reorder the repo sources to eliminate or at least order jcrap after mavencentral?

yairm210:
merge translations

yairm210:
Translations branch created

yairm210:
Translations PR created

yairm210:
@SomeTroglodyte  As for the Packr problem, I tried using the up-to-date 4+ Packr versions directly in the code instead of the Anuken wrapper, but it failed for baffling reasons - couldn't find a function in a library that Gradle insisted was already there. Didn't manage to get it to work.
I am seriously considering throwing Win32 support out since it's causing us a lot of hassle. Everything else is already converted to packr-from-cli, so if we want to ditch it we can remove the Packr library completely.

PR #5241, opened 2021-09-16
Arabic. properites

Kleznam:
So yesterday I translate arabic and I know github doesn't support Arabic but I tried to sync it in reverse and write in reverse iam not stupid :( ,but now I used the Franko script for Arabic which is a Latin script but I I didn't translate except like 20 lines or even less but I want review and I want Arabic people to say their opinion about this translation


SomeTroglodyte:
Welcome back. I hope my tone didn't come over too discouraging. What I meant is I welcome the attempt and would be very curious to see if can be pulled off. For hints on where the problems lie see https://github.com/libgdx/libgdx/issues/787 - they tried to do it "properly", but that doesn't mean a 'hack' style approach just delivering codes for the glyphs one wants to see as left-to-right couldn't work.

Sorry I can't help with actual knowledge of the language.

Kleznam:
There is an error with process code exit 1 ,but yeah :)  iam trying to think how to fix this..

SomeTroglodyte:
Hopefully helpful (or tell us what's bad or unclear or missing): https://github.com/yairm210/Unciv/wiki/Translating#pitfalls
There is - to the right of "Failing after 5m — build" is a Details link, and in all that console output the line "com.unciv.testing.TranslationTests > allTranslationsHaveCorrectPlaceholders STANDARD_OUT", and below that a few pointers.
Basically every [thing] in square brackets needs to be present verbatim in the translation.

Kleznam:
Ahh so is the problem mainly about it being RST language (written from right to left) or is it about the gylphs being connected?.

SomeTroglodyte:
Both I guess, _if_ you want to create a correct 'picture' for a given computer representation (in a standard encoding) of arabic text. What you're trying to do is trick a LTR renderer into producing a sufficiently similar picture by feeding it reversed code, and for glyphs that would normally be build by using a combination of code points you'd have to find a replacement that looks the same or similar enough... As I sad, it's just pretty squiggles to me, so I can't judge. I only know a portion of the code involved in getting translated text onto a screen - in Unciv.

yairm210:
If I was going about this I would do it in two stages -
1. Non-glyph LTR language, like Hebrew
2. Add glyph recognition

Kleznam:
Ahh i don't really understand what you mean yarim? :0

yairm210:
I'm saying that you have two separate problems to solve, and you can solve them once at a time instead if both at once

Kleznam:
Ohh I get what you mean, so do you want me to write from left to write with no gylhpes?...or wut..

Kleznam:
I changed the code a bit ,and added the thing that should make Arabic work, and translated 90 lines but I have two question why is the n in \n important the second as I said before should i but the code for Arabic up the translation or down the translation?

yairm210:
That's definitely not going to work... the code probably needs to be in the translation.kt file, and chances are it won't work there either, it seems to be dependant on some external class that isn't actually included here at all?

SomeTroglodyte:
Chuckle. There's code and then there's data processed by that code. The translation files are data. You added source code to data. 

Not meant disparaging - I see no inherent special value in having coding skills.

Look, maybe first thing you should try is a way to test your stuff without us merging it into master first. Installing Android Studio and cloning your fork there may take a few hours, but you could run the game directly with your added translations and **_see_** the effect. It's not trivial should you have zero experience with coding, but not impossible, as there's instructions, one of many in our wiki. The worst thing many people stumble on is after you clone the repo and then need the "Run configuration". There's a gradle sync there that needs to finish or else the instructions won't work cleanly, so just make a long pause at that point.

Kleznam:
Okey.

PR #5247, opened 2021-09-17
Vegetation unique, Jungle Hills

ravignir:


yairm210:
Nice!!

ravignir:
Now I am thinking, that would break base ruleset mods or mods that edit Jungles/Forests.

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -417,15 +417,17 @@ class MapGenerator(val ruleset: Ruleset) {
...
-                && it.getLastTerrain().name in candidateTerrains }) {
-            val vegetation = (randomness.getPerlinNoise(tile, vegetationSeed, scale = 3.0, nOctaves = 1) + 1.0) / 2.0
-
-            if (vegetation <= tileMap.mapParameters.vegetationRichness)
-                tile.terrainFeatures.add(Constants.vegetation.filter { ruleset.terrains[it]!!.occursOn.contains(tile.getLastTerrain().name) }.random(randomness.RNG))
+        val vegetationTypes = ruleset.terrains.values.filter {
+            it.type == TerrainType.TerrainFeature && it.uniques.contains("Vegetation") || it.name == Constants.jungle || it.name == Constants.forest
+        }
+        for (tile in tileMap.values.asSequence().filter { it.terrainFeatures.isEmpty() || it.isHill() }) {
+            val vegetation = (randomness.getPerlinNoise(tile, randomness.RNG.nextInt().toDouble(), scale = 3.0, nOctaves = 1) + 1.0) / 2.0
```
SomeTroglodyte, line 424: This un-perlins your perlin. That seed is not fed to a random generator, it's the Z axis the perlin "cloud" is cut on. Won't look as good as before.

SomeTroglodyte (commented):
Too much Hill hardcoding, though I can't really think of a better option. Or - group occursOn into base terrains and features, then test for `baseTerrain in first && (second.isEmpty && features.none() || features.any { it in second })`??? Or - change occursOn to apply only to baseTerrain only and create a unique for additional base/feature combos.

Also, likelihood could depend on temperature/humidity (must be same as routine before -> better integrate the two routines - or pass a seed) - for a potentially more natural look?

Generally - good direction, so I opt for "Approve"

AdityaMH:
Outdated?

yairm210:
I think we'll close this

PR #5248, opened 2021-09-17
Add a fixed base strength to motivationToAttack

SimonCeder:
The problem: The new force evaluation algorithm does not add city strengths. In very early game city-states have no army which makes the AI think it has overwhelming odds of attacking them.
The solution: As in civ V's equivalent function `CvDiplomacyAI::DoUpdateOnePlayerMilitaryStrength`, adds a base value of 30 (slightly more than a Warrior) to the force evaluations to reduce swinginess in early game.

### DIFF
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -498,8 +498,9 @@ object NextTurnAutomation {
     }
 
     private fun motivationToAttack(civInfo: CivilizationInfo, otherCiv: CivilizationInfo): Int {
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
> swinginess

:laughing: 

I trust you did consider and reject simply adding the base within `getMilitaryMight`? Would then affect 'protection' and AI/AI / AI/Player and Trade and... strength evaluations as well, which sounds like it might also be more of an improvement than a bad thing

SimonCeder:
Sure, the main reason I didn't is that civ V doesn't. I'll have a look through the source how these numbers are handled elsewhere, possibly adding it directly to the evaluation makes sense.

PR #5250, opened 2021-09-17
Nation picker portraits

### LINKED ISSUE #5211: New game nation picker alternatives
<details><summary>Some screenshots:</summary>

image
image
image
image
image
image

</details>

These were all done with [the result of this script](LeaderPortraitScraper.py) as perm visual mod, and varying code tweaks.

Which variant?

SomeTroglodyte:
Closes #5211 - see discussion & screens there. 

- The red X idea = Separate PR
- Some text looks less than ideal because it actually isn't ICivilopediaText but comes from the old description code and thus is indented by spaces, not matching the introducing line - not specific to this, won't change in _this_ PR
- The closing "X" position and the insistent wish in some open issue for a "confirm button" - not this PR, too.

<details><summary>Looks same as the last initial variant presented in the issue:</summary>

image
</details>

### DIFF
--- a/core/src/com/unciv/ui/newgamescreen/NationTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/NationTable.kt
@@ -8,6 +8,7 @@ import com.badlogic.gdx.utils.Align
 import com.unciv.Constants
 import com.unciv.models.ruleset.Nation
 import com.unciv.models.ruleset.Ruleset
+import com.unciv.ui.civilopedia.FormattedLine.IconDisplay
 import com.unciv.ui.civilopedia.MarkupRenderer
 import com.unciv.ui.utils.*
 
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/ui/newgamescreen/NationTable.kt
```diff
@@ -8,53 +8,86 @@ import com.badlogic.gdx.utils.Align
...
+| | | |     *nationIndicator*   *leaderDisplayLabel*   | | | |
+| | | |   center or left/wrap            0: all sides  | | | |
+| | | +------------------------------------------------+ | | |
+| | |     White                          Dark-blue       | | |
+| | |   MarkupRenderer.render(getCivilopediaTextLines)   | | |
+| | |     left/wrap                      10: all sides   | | |
+| | |                                                    | | |
+| | +----------------------------------------------------+ | |
+| +--------------------------------------------------------+ |
++------------------------------------------------------------+
```
yairm210, line 91: DAMN

SomeTroglodyte: Oh, sorry, I needed that to understand better where which padding affects what and that stuff... Could have taken it out after. The Scene2D debug flags don't always help.

PR #5259, opened 2021-09-18
Fixed intercept chance being inverted

xlenstra:
If I'm not mistaken, this fixes a bug where interceptions would only fail if there is more than 100% chance to intercept. I've made this edit on mobile, so I can't test the effects though


SomeTroglodyte (approved):
Without testing - you can't be wrong

PR #5261, opened 2021-09-18
Translation update

cjddig:


xlenstra:
Placeholder `[civName]` not found in `Korean` for entry `It's come to my attention that I may have attacked [civName], a city-state under your protection.
    While it was not my goal to be at odds with your empire, this was deemed a necessary course of action.`
    Placeholder `[amount]` not found in `Korean` for entry `Diplomatic Marriage ([amount] Gold)`

PR #5266, opened 2021-09-19
I Am Coding A New Drag And Drop Card Feature For The Game's Engine

Interdice:
I am coding a new drag and drop interface to be used in various parts of this game. This is a very early prototype to excuse any mistakes I make


Interdice:
image

This is an example of what i want to add

Interdice:
I might ask noob questions as this is pretty hard

Interdice:
Is there a way to use the stage from an actor?

Interdice:
stage from the parent class

logicminimal:
uh... this surely is a great addition for **mods** and you are free to implement it in this game, but we are heading for civ 5 so it will be disabled in un-modded games

Interdice:
@logicminimal 

This feature can be used in parts of the civ 5 features and improve the visuals and usability of the game. unfortunately i am not sure if drag and drop like civ 6 are possible with the libgdx framework i can't use the stage included with the actor to show the images or text

logicminimal:
> @logicminimal 
> 
> This feature can be used in parts of the civ 5 features and improve the visuals and usability of the game. unfortunately i am not sure if drag and drop like civ 6 are possible with the libgdx framework i can't use the stage included with the actor to show the images or text

i have no idea if you can implement it easily but im %99 sure that libgdx doesn't forbid implementing this, also what are some example usages?

Interdice:
@Iogicminimal

I haven't played ever played civ 5 as I only got into the franchise with civ 6. 

it can be used for managing artifacts in museums. I was considering using buttons but these look more cool and will be easier for the end user. This can also be used for policies if civ 5 has them

The reason i can't get this to work is because i am using an actor for the rectangle. The rectangle has limited use of the stage and can't use the stage to create more objects like images or text to put on the dragable rectangle.

I might give up on this as it is hard to implement but i will try for a few more days

logicminimal:
> @Iogicminimal
> 
> I haven't played ever played civ 5 as I only got into the franchise with civ 6. 
> 
> it can be used for managing artifacts in museums. I was considering using buttons but these look more cool and will be easier for the end user. This can also be used for policies if civ 5 has them
> 
> The reason i can't get this to work is because i am using an actor for the rectangle. The rectangle has limited use of the stage and can't use the stage to create more objects like images or text to put on the dragable rectangle.
> 
> I might give up on this as it is hard to implement but i will try for a few more days

oh, okay then

yairm210:
> Is there a way to use the stage from an actor?

Using Actor.stage you can add what you want

SomeTroglodyte:
I wouldn't dare try some stunt like this unless either I found a doc on how to in the Gdx wiki or a working example.
Also, adding freshly instantiated actors to the stage in an override draw sounds like stack overflow.
Actor.stage is null in the initialization phase and gets set by the addActor call - quite like Actor.parent really. If you use it too early you get the null pointer problems - but in such a case just set a breakpoint and look at the call stack and you normally can easily see what happened that you didn't expect.
Lastly, I wouldn't touch SpriteBatches & co, just Actor.setPosition _might_ suffice.

ravignir:
afaik, there is no drag and drop action in civ5. Everything is just click here and there.

Interdice:
@ravignir sure but this is a ui choice and doesn't affect gameplay

Interdice:
@SomeTroglodyte good idea might try

Interdice:
i am deciding to close this for now. I simply don't have the experience and this is complicated

PR #5267, opened 2021-09-19
Fixed a few bugs relating to earning great prophets

xlenstra:


SomeTroglodyte:
:+1: Was about to open an issue yesterday about the CS gifted prophets but fell asleep

PR #5270, opened 2021-09-19
Adds conditionals to most of the uniques currently in the enum

xlenstra:
Based of my PR moving the `unique` files to their own folder, so ignore the first commit.

Most of the ones without conditional support are harder to implement, as the unique is not checked where the behaviour is executed (e.g. city states providing military units), or run into difficulties requiring large refactors.

Added uniques:
- "[stats]"

Added conditionals:
- <if this city has at least [amount] specialists>

Deprecated uniques:
- "[stats] if this city has at least [amount] specialists"

- "[stats]"
- "[stats] [cityFilter]"
- "[amount] units cost no maintenance"
- "[amount]% maintenance costs for [mapUnitFilter] units"
- "Provides [stats] per turn"
- "Provides [stats] [cityFilter]"
- "Provides [amount] Happiness"


yairm210:
Aha
So you can have conditionals without uniques
So the getPlaceholderParams should be .removeConditionals before .replace, for runtime improvement (regex is heavy)

PR #5272, opened 2021-09-19
Fix translation tests and wrongly translated conditionals

xlenstra:
This PR fixes the tests that made 3.17.0 fail, as well as some problems with generating translatable strings for conditionals that apparently didn't work.
Branch is based off of 3.17.0, so merging this to another branch and publishing a release from there can be done without additional commits being added unintentionally.


yairm210:
Thanks!

PR #5276, opened 2021-09-20
Using reflection to find and report deprecated uniques

yairm210:
CLI output (for RekMOD as example):
`Manchuria's unique "-[100]% [Mounted] unit maintenance costs" is deprecated As of 3.16.16, replace with "[amount]% maintenance costs for [mapUnitFilter] units"`

Again we encounter the problem of "how do we display certain errors to the mod creator and not to the users"
For now I'm fine with using the cli, but suggestions are welcome

The reason I removed the deprecation replacement from the UniqueType is because I wanted to give a varargs of acceptable targets for the unique, and I couldn't do that with an "optional" deprecation unless I specified for each one that yes, this is really not deprecated. This way is definitely an improvement over that way.

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -23,7 +23,6 @@ import com.unciv.models.stats.Stats
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.colorFromRGB
 import kotlin.collections.set
-import kotlin.reflect.typeOf
 
 object ModOptionsConstants {
     const val diplomaticRelationshipsCannotChange = "Diplomatic relationships cannot change"
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -23,6 +23,7 @@ import com.unciv.models.stats.Stats
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.colorFromRGB
 import kotlin.collections.set
+import kotlin.reflect.typeOf
 
 object ModOptionsConstants {
     const val diplomaticRelationshipsCannotChange = "Diplomatic relationships cannot change"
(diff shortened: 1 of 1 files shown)


xlenstra:
Although this is fine, I don't know how many modders are comfortable enough with the command line to actually look at what is output there. Maybe also putting this on the 'check mod links' screen in the options menu would help in that regard, though I don't know how difficult it is to show these warnings there but not when downloading a mod or selecting it when starting a game.

SomeTroglodyte:
`UniqueTarget` is an enum, no? Then acceptable targets shouldn't be varargs, but an `EnumSet<UniqueTarget>`?
As UI - severities "Dev warning" < "User warning", configurable filter on the severity? Could be public, but users need to be very specific they want to see warnings intended for authors?

SomeTroglodyte:
Oh - did you test and actually _get_ an annotation at runtime? When I tried stuff with annotations, despite `@Retention(AnnotationRetention.RUNTIME)` they _never_ showed up in actual reflection.

yairm210:
Yup, tested this
Varargs rather than enumSet because it's almost always just one, and when it's multiple then it's easier like this
But I'm not picky, if there are advantages to enumSet then we can use that instead

yairm210:
> Oh - did you test and actually _get_ an annotation at runtime? When I tried stuff with annotations, despite `@Retention(AnnotationRetention.RUNTIME)` they _never_ showed up in actual reflection.

Yup, apparently for enums you need to add the .getField to get the "instance" of the enum. It's confusing.

yairm210:
> Although this is fine, I don't know how many modders are comfortable enough with the command line to actually look at what is output there. Maybe also putting this on the 'check mod links' screen in the options menu would help in that regard, though I don't know how difficult it is to show these warnings there but not when downloading a mod or selecting it when starting a game.
> 
> Other than that: this way of detecting looks good and extra behaviour can always be added in a later PR

Sounds to me like each error needs to come with a type, like "error", "userWarning", "modderWarning"

PR #5279, opened 2021-09-20
Natural Wonders un-hardcoded

SomeTroglodyte:
No more 1 fun per NW, or a big when emulating a vtable, all effects and filters moddable

- Nat W "Continent" todo's resolved (avoid largest landmass...)
- Extra: Simon's continent assigns done differently, can now be reused as map editor floodfill
- Untouched: Fountain of Youth effect still needs a better unique
- Open: These new uniques should be hidden from civilopedia - as maybe the other mapgen uniques too - add property to UniqueType?

~~OK so I forgot to teach the parameter type to accept the lists I used, but~~ - architecture acceptable so far?


core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -19,7 +19,7 @@ import kotlin.random.Random
 class MapGenerator(val ruleset: Ruleset) {
     companion object {
         // temporary instrumentation while tuning/debugging
-        const val consoleOutput = false
+        const val consoleOutput = true
```
xlenstra, line 22: Do we want this?

core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
```diff
@@ -22,410 +24,199 @@ class NaturalWonderGenerator(val ruleset: Ruleset, val randomness: MapGeneration
...
+                        if (tile.baseTerrain == convertNeighborsTo) continue
+                        if (tile.baseTerrain == convertNeighborsExcept) continue
+                        if (convertNeighborsTo == Constants.coast)
+                            for (neighbor in tile.neighbors) {
+                                // This is so we don't have this tile turn into Coast, and then it's touching a Lake tile.
+                                // We just turn the lake tiles into this kind of tile.
+                                if (neighbor.baseTerrain == Constants.lakes) {
+                                    neighbor.baseTerrain = tile.baseTerrain
+                                    neighbor.setTerrainTransients()
+                                }
```
xlenstra, line 134: So if we have a tile turn into coast that touches a lake that touches a lake, the second lake won't become coast? I feel if you do this you either have to convert all lakes (requiring a BFS or something) or no lakes to coast

xlenstra, line 134: Are resources already scattered at this point? Can this place a lake-only resource on land?

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
SomeTroglodyte: Maybe? No, in just for testing. Although three months ago the output was permanently on...

core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
SomeTroglodyte: No, spread resourecs comes after

core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
```diff
@@ -22,410 +24,199 @@ class NaturalWonderGenerator(val ruleset: Ruleset, val randomness: MapGeneration
...
-            for (tile in location.neighbors) {
-                if (tile.baseTerrain == Constants.coast) continue
-                tile.baseTerrain = Constants.coast
-                clearTile(tile)
-            }
-        }
-    }
 
-    /*
     Rock of Gibraltar: Specifics currently unknown
```
xlenstra, line 184: Based on the source code of the original:
- Spawns on land tiles
- Requires at least 3 adjacent coast tiles
- Requires exactly 1 adjacent land tile

SomeTroglodyte: OK, but I just migrated the existing base -> _now_ you can fix that by json only

xlenstra: Agree, just something I came across so feel free to ignore this for now

SomeTroglodyte:
github is Strange sometimes - my mail says there should be a comment about the lakes code, but the link doesn't point it out.... Anyway: that's existing code migrated, and tested so it being active for krakatoa too (unlike before, to save some conditionals) doesn't hurt. With number-to-spawn+=99, a 120x80 map and 12 runs the lake code never triggered, so: //todo think hard whether the situation it aims to fix is actually impossible...

xlenstra:
> github is Strange sometimes - my mail says there should be a comment about the lakes code, but the link doesn't point it out.... Anyway: that's existing code migrated, and tested so it being active for krakatoa too (unlike before, to save some conditionals) doesn't hurt. With number-to-spawn+=99, a 120x80 map and 12 runs the lake code never triggered, so: //todo think hard whether the situation it aims to fix is actually impossible...

Jup that was me, I deleted the comment after I realized I misunderstood the code and the problem I thought there was didn't actually exist :)

xlenstra:
Scanning the diff looks fine to me, also architectural-wise. Nice!

I'm in favour of adding a 'hidden in civilopedia' boolean for uniques, and it could also be used for 'hidden when religion is disabled' and similar uniques as well

SomeTroglodyte:
No lambasting for sneaking `or` logic into a filter? It's actually unclean in the sense the filtering code exists only privately in the NW generator, so the checker should test it specifically, as I wrote it, it doesn't 100% fit the _general_ tileFilter. Decision: No `or` -> multiply those uniques, works too, then no real need for the private filter. Alternately, move the capability to the 'public' filter. Also: didn't really think about how your conditionals would fit here.

xlenstra:
> No lambasting for sneaking `or` logic into a filter? It's actually unclean in the sense the filtering code exists only privately in the NW generator, so the checker should test it specifically, as I wrote it, it doesn't 100% fit the _general_ tileFilter. Decision: No `or` -> multiply those uniques, works too, then no real need for the private filter. Alternately, move the capability to the 'public' filter. 

The main problem I see here (but didn't think about when reading it through) is translation: for each of these parameters, specific translations will be generated, resulting in more duplicate work for translators. Using an 'or' and comma's to seperate them might also not work in some other languages, so on second thought this might actually be a problem. Either the Translations would need some logic to work with this case, or if possible the filters should be split up.

> Also: didn't really think about how your conditionals would fit here.

Don't think they would apply here anyway, at least not with anything I have in mind for it right now. Maybe later it turns out some conditionals would fit here, but we'll add them then.

SomeTroglodyte:
> The main problem I see here (but didn't think about when reading it through) is translation

Yup, all the more reason to hide them completely... If they were to be displayed, the parsed list should be translated individually same as currently, and maybe someone comes up with a clever hack for how to answer "how do you express a list of terms of comparable semantic type and significance in your language" per i10n...

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -464,16 +464,18 @@ class MapGenerator(val ruleset: Ruleset) {
...
+     */
     fun assignContinents(tileMap: TileMap) {
         var landTiles = tileMap.values
             .filter { it.isLand && !it.isImpassible()}
         var currentContinent = 0
 
         while (landTiles.any()) {
             val bfs = BFS(landTiles.random()) { it.isLand && !it.isImpassible() }
-            bfs.stepToEnd(currentContinent)
+            bfs.stepToEnd { it.setContinent(currentContinent) }
```
yairm210, line 478: As said above, this should be two lines.
Once to run the bfs, one to assign id on resulting tiles

core/src/com/unciv/logic/map/BFS.kt
```diff
@@ -25,11 +25,10 @@ class BFS(
 
     /** Process fully until there's nowhere left to check
      *  Optionally assigns a continent ID as it goes */
-    fun stepToEnd(continent: Int? = null) {
-        if (continent != null)
-            startingPoint.setContinent(continent)
+    fun stepToEnd(actionPerTile: ((TileInfo)->Unit)? = null) {
```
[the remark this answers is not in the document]
SomeTroglodyte: Well the BFS side effect wasn't my original idea, I just thought if it's already there at least make it reusable - like the Array initializer style.

core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
```diff
@@ -62,6 +64,31 @@ enum class UniqueParameterType(val parameterName:String) {
             return UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
         }
     },
+    TerrainFilter("terrainFilter") {
+        private val knownValues = setOf("All", "Coastal", "River", "Open terrain", "Rough terrain", "Water resource",
+            "Foreign Land", "Foreign", "Friendly Land", "Friendly", "Enemy Land", "Enemy")
+        override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
+                UniqueType.UniqueComplianceErrorSeverity? {
+            if ("," in parameterText || " or " in parameterText)
```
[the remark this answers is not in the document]
SomeTroglodyte: As said - the filter implementation is private to the generator at the moment, and these uniques shouldn't be shown to the user (question is - as the unique system is currently evolving - where will the switch go).

Plus - the whole thing is more of a proof of concept - it _can_ be done, and the resulting code has half the line count for a tad more complexity. I haven't forgotten you shoot these down (except for the precedent), but for a design study I "just" got it to work, then show, then style.

SomeTroglodyte (commented):
Hey github - you're wrong. I did **NOT** hit any review buttons, as this is my own PR I don't get those!

SomeTroglodyte:
merge+requested changes done.

FYI: Unit tests by now _succeed_ **despite** a crash in the mokk engine:
```
WARNING: Failed to transform class java/lang/Object
java.lang.IllegalArgumentException: Unsupported class file major version 60
	at net.bytebuddy.jar.asm.ClassReader.<init>(ClassReader.java:184)
```

<details><summary>Question remains what to - how to - do about:</summary>

image
</details

yairm210:
Hm, that doesn't look great.
I vote adding another "adjacent to [amount]" unique, since most uses have only 1 number, and not a range

SomeTroglodyte:
<details><summary>You vote for prettier text, I vote for back to:</summary>

image
</details>

...Maybe we should hide the other placement uniques, too. It's more of a "gaming experience" when you learn by exploring that these wonders actually follow a pattern and what that may be - and that the pattern has some resemblance to the Real World wonders. 

Code of this PR currently reflects that, except for _not_ generating translation prompts for those uniques.

So (A) - explain everything in Civilopedia, (B) - leave some mystery and hide, using a hardcoded helper as in this PR, or (C) - hide using a UniqueType property?
All three would need some more code changes, so draft it is.

yairm210:
I'm in favor of visibility as of now, if we have

SomeTroglodyte:
Might take a while. The music PR was a mistake - I just can't close Unciv down anymore after any atlas build or test or...

SomeTroglodyte:
<details><summary>Now looks like this in pedia</summary>

image
</details>

SomeTroglodyte:
For reference: [wiki updated](https://github.com/yairm210/Unciv/wiki/Uniques#terrain-uniques---map-generation)

PR #5280, opened 2021-09-20
Added support for conditionals to some more uniques

xlenstra:
Also moved the conditionals to the unique enum itself, as in the end they are uniques themselves, and implemented another conditional.

Added uniques:
- [amount]% [stat]

Added conditionals:
- \<while the empire is happy>

Deprecated uniques:
- [amount]% [stat] while the empire is happy

- [amount]% [stat]
- Provides [amount] [resource]
- Provides a free [buildingName] [cityFilter]
- Gain a free [buildingName] [cityFilter]

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1306,15 +1306,17 @@ ConditionalsPlacement =
 # However, you should not translate the parts between the brackets, only move them around so that when
 # translated in your language the sentence sounds natural.
 #
-# Note that every time a new conditional is added, it will be added below and this line will have to be retranslated.
-# As this will happen quite a lot in the near future, you may want to wait with translating this
-# until most conditionals have been added.
-#
 # Example: "+20% Strength <for [unitFilter] units> <when attacking> <vs [unitFilter] units> <in [tileFilter] tiles> <during the [eraName]>
 # In what order should these conditionals between <> be translated?
 # Note that this example currently doesn't make sense yet, as those conditionals do not exist, but they will in the future.
+#
+# As this is still under development, conditionals will be added al the time. As a result,
+# any translations added for this string will be removed immediately in the next version when more
+# conditionals are added. As we don't want to make you retranslate this same line over and over,
+# it's removed for now, but it will return once all planned conditionals have been added.  
 
-<if this city has at least [amount] specialists> <when at war> <when not at war> <while the empire is happy> = 
+
+# <if this city has at least [amount] specialists> <when at war> <when not at war> <while the empire is happy> = 
 
 
 
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1314,7 +1314,7 @@ ConditionalsPlacement =
 # In what order should these conditionals between <> be translated?
 # Note that this example currently doesn't make sense yet, as those conditionals do not exist, but they will in the future.
 
-<if this city has at least [amount] specialists> <when at war> <when not at war> = 
+<if this city has at least [amount] specialists> <when at war> <when not at war> <while the empire is happy> = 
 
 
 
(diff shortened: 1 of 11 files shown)


core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -162,7 +162,8 @@ class CityConstructions {
     
     fun addFreeBuildings() {
         // "Provides a free [buildingName] [cityFilter]"
-        for (unique in cityInfo.getMatchingUniques(UniqueType.ProvidesFreeBuildings)) {
+        for (unique in cityInfo.getLocalMatchingUniques(UniqueType.ProvidesFreeBuildings)) {
```
yairm210, line 165: GetlocalMatchingUniques should run conditional check itself, otherwise this line will appear in loads of places

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -161,6 +161,7 @@ class CityInfo {
         civInfo.civConstructions.tryAddFreeBuildings()
 
         for (unique in getMatchingUniques(UniqueType.GainFreeBuildings)) {
```
yairm210, line 163: Same, conditional check should be in the function

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -283,8 +283,13 @@ class CityStats(val cityInfo: CityInfo) {
                     stats.production += unique.params[0].toInt()
             }
 
-
+        for (unique in uniques.filter { it.placeholderText == UniqueType.StatPercentBonus.placeholderText }) {
```
yairm210, line 286: filter it.isOfType

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -283,8 +283,13 @@ class CityStats(val cityInfo: CityInfo) {
...
+            if (!unique.conditionalsApply(cityInfo.civInfo, cityInfo)) continue
+            stats.add(Stat.valueOf(unique.params[1]), unique.params[0].toFloat())
+        }
+        
         // For instance "+[50]% [Production]
+        // Shouldn't this be deprecated?
```
yairm210, line 292: Probably

android/assets/jsons/translations/template.properties
```diff
@@ -1314,7 +1314,7 @@ ConditionalsPlacement =
 # In what order should these conditionals between <> be translated?
 # Note that this example currently doesn't make sense yet, as those conditionals do not exist, but they will in the future.
 
-<if this city has at least [amount] specialists> <when at war> <when not at war> = 
+<if this city has at least [amount] specialists> <when at war> <when not at war> <while the empire is happy> =
```
yairm210, line 1317: Once again calling attention to the fact that when we rebuild translations, all existing order will be lost.
Translators will need to do it all user again for every new conditional.
I really think we should remove the sorting entirely, at least for now

xlenstra: I have remarked right above the translation that this is the case, so translators can choose for themselves whether or not to add these translations. I agree that that is imperfect in the short term, but in the end this will allow for the best translations, and for then this way the logic is already in place.

core/src/com/unciv/logic/city/CityConstructions.kt
xlenstra: I'll fix this in the next PR on conditionals, if you don't mind. To fix this properly a lot needs to be changed, and I'd rather fix that on its own

android/assets/jsons/translations/template.properties
xlenstra: I don't know whether it is actually useful in general, but I can imagine for some languages it alleviates at least some of the weirdness the translation would otherwise have. For now I'll comment out things related to this, but unless a better solution is found I feel like this should be enabled again once we've added all the conditionals we plan on adding.

Issue #4777, opened 2021-08-06
Battle modifier uniques review / roadmap


SomeTroglodyte:
### Inventory / current state of affairs
|placeholder|Att/Def|Unit/Civ|suggested|notes|
|-----------|-------|--------|---------|-----|
|+[]% Strength vs []|*|U,C|[]% Strength vs []|isProbablySiegeUnit will need to check param0 when this is generalized|
|-[]% Strength vs []|*|U,C|-|see above|
|+[]% Combat Strength|*|U|[]% Combat Strength||
|[] units deal +[]% damage|*|C|[] units deal []% damage||
|Bonus for units in 2 tile radius 15%|*|U|Bonus for units in [] tile radius []%|modChecker should ensure param1 in 1..5|
|[]% Strength when stacked with []|*|U|OK|implementation should check if a mod makes param1 apply to the unit itself|
|[]% Strength for [] units which have another [] unit in an adjacent tile|*|C|OK||
|[]% Strength for enemy [] units in adjacent [] tiles|*|U|OK||
|+10% Strength for all units during Golden Age|*|C|[]% Strength for all units during Golden Age||
|+30% Strength when fighting City-State units and cities|*|C|[]% Strength when fighting City-State units and cities||
|+[]% Strength in []|*|U|[]% Strength in []|tileFilter for defender's tile|
|+[]% Strength for units fighting in []|*|C|[]% Strength for units fighting in []|tileFilter for defender's tile|
|+[]% Strength if within [] tiles of a []|*|C|[]% Strength if within [] tiles of a []|tileFilter for defender's tile|
|+[]% Strength when attacking|A|U|[]% Strength when attacking||
|+[]% attack strength to all [] units for [] turns|A|C|[]% attack strength to all [] units for [] turns||
|+15% Combat Strength for all units when attacking Cities|A|C|[]% Combat Strength for all units when attacking [cityFilter]||
|+[]% attacking strength for cities with garrisoned units|A|City|[]% attacking strength for [cityFilter] with garrisoned units||
|[]% attacking Strength for cities|A|City|[]% Offensive Strength for [cityFilter]|for consistency with "Defensive"|
|+[]% Strength when defending|D|U|[]% Strength when defending||
|[]% Strength when defending vs [] units|D|U|OK||
|+[]% Defensive strength for cities|D|City|[]% Defensive strength for [cityFilter]||
|Defense bonus when embarked|D|U|Defence bonus when embarked|Works exactly like the next one - but - display wording|
[...]

PR #5281, opened 2021-09-21
Save the sources of uniques with the uniques themselves

xlenstra:
This PR tells each unique what type of object it source was, and was that instance of object is called, conform #5091.
This has multiple applications, none of which are implemented in this PR:
- Comparing with the types saved in the `UniqueType` enum to see if it is on a valid object
- Better determining the source of stat bonuses in the `CityStatsTable`
- After the unification of Strength bonuses for uniques (#4777), still having something reasonable to display as its source (e.g.: "Autocracy Complete (Policy)")


yairm210:
Hah! I started working on this exact pr as well :)
Looks good!

PR #5287, opened 2021-09-21
Nation Picker visual help where to click

### LINKED ISSUE #5229: Confirm Button when choosing Civilizations
**The Problem**:
For new players they sometimes seem to get trapped when choosing civilizations. 

**The Solution:**
Add a confirm button on the right most side of the panel. 

~~just a little QoL~~

SomeTroglodyte:
Closes #5229
- Apparently some people don't 'get' intuitively that the whole right side of that popup is active
- Adds a checkmark to clarify - green on mouseover - same onClick as the widget behind it
- Moves the close "X" to a symmetrical place, its position was inside the popup long ago, but the "cover screen" change of Popup probably undid that - was a little lost in an unexpected place
- Allows ESC / Back

<details><summary> Screenshot </summary>

image
</details>

Not 100% happy as the 'floating' character of these two symbols is visually not very clear. Add a drop shadow (easiest done in gimp, so 'close' would get a separate png) or a half-transparent square behind?


yairm210:
I think adding a circle behind the icons could help

SomeTroglodyte:
Last push corresponds to upper screenshot, the other is an experiment to compare how a real drop shadow (only a few pixels at alpha 67%, fading to nothing within 10 pixels) - gimped icons, not done in gdx - would look.

image
image

</details>

PR #5288, opened 2021-09-21
Update German.properties

Lukasdotcom:


SomeTroglodyte:
> Ich habe eine Idee une ich sotte es dir sagen

Two typos

> It is done. = Es ist fertig.

Not quite the same. Context is diplomacy not construction.

> Units ending their turn on [tileFilter] tiles take [amount] damage =

Trailing blank removed -> read the wiki

Lukasdotcom:
@SomeTroglodyte 
I think I fixed all the mistakes that you mentioned. I removed the It is done because I have now idea what it would translate to.

Mape6:
Hi Lukas, I added the translation for carthago this morning in my PR. See #5277

PR #5294, opened 2021-09-22
Update German translation

Mape6:


Mape6:
What does the error message mean?
# <if this city has at least [amount] specialists> <when at war> <when not at war> <while the empire is happy> = ends without a space at the end

SomeTroglodyte:
It means that the mentioned line ends with the equals sign, not with a space (duh), and despite the line being a comment the tests insist on a space after the equals. Needs either the test to learn what a comment is, or you to re-add the space, or someone to re-think if this kind of comment line couldn't be expressed otherwise (in the template file).........

PR #5296, opened 2021-09-22
Texture pack/load for mods also distributes by Images.textureGroup

### LINKED ISSUE #5238: Mod Images folders - documentation and reality
Roughly half of the mods I see, including 5Hex, have Image **sub**folders directly under the mod's root, but we document they should go one level deeper under mod/Images/subfolder.

While the placement is largely irrelevant as long as the atlas and texture are included and correct, we should get this straightened out. It pains me to say so, but I propose we make sure the code supports the Images/* placement, such that the packer works for that, and that only, even if it might mean more work for our stressed modders.

We might also consider warnings in the mod checker, and also warnings for json misplacement (yes they manage that, too, and the game simply ignores the jsons making it an empty mod).

SomeTroglodyte:
Closes #5238

This enables the texture packing and the loader to treat mods same as the base game - control over texture distribution, Atlases.json, etc. Example: 5Hex could ensure that users who turn off pixel units got a little performance boost - by eliminating the need for the 'spill' texture to swap in. Haven't measured it, though.

### DIFF
--- a/core/src/com/unciv/ui/utils/ImageGetter.kt
+++ b/core/src/com/unciv/ui/utils/ImageGetter.kt
@@ -1,6 +1,7 @@
 package com.unciv.ui.utils
 
 import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.graphics.Color
 import com.badlogic.gdx.graphics.Texture
 import com.badlogic.gdx.graphics.Texture.TextureFilter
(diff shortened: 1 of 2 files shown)


yairm210:
Have you seen that 5hex tiles are loaded as expected?

SomeTroglodyte:
Yup, tested with 5Hex as shown in that [Nemo screenie over there](https://user-images.githubusercontent.com/63000004/133747965-5e1c53c1-4571-44ba-909b-4935406a489d.png), also toggling pixel units. Actually played with it for 2 days until I saw an update of 5Hex...

SomeTroglodyte:
The documentation is in [the top 'Mods' wiki entry](https://github.com/yairm210/Unciv/wiki/Mods#more-on-images-and-the-texture-atlas) - does not merit its own page by length. I'll do a sidebar link anyway.

PR #5297, opened 2021-09-22
World and Natural Wonders Overview

### LINKED ISSUE #5256: Austria is Missing Buildings
Austria is missing the Pentagon and the building that comes with the railroad research in v3.16.14 on Android 8. It's a very nice game, but this was annoying.

SomeTroglodyte:
We should close #5256 - this is the code inspired by it that did those screenshots.

Open questions:
- logicminimal may have a point - currently all religion/victory enabled World Wonders are listed, just as there is 1 row for every Natural Wonder in the game (so you know how many you haven't found yet as the count / map size formula is no secret - but it _may_ disclose through sorting whether a not-found NW's name is alphabetically preceding or following a found one).
- Icon for the category button up there...


core/src/com/unciv/ui/overviewscreen/WonderOverviewTable.kt
```diff
@@ -0,0 +1,196 @@
...
+            .filter { it.isNaturalWonder() }
+            .map { it.naturalWonder!! to it }
+            .toMap()
+        val naturalsCount = allNaturalsMap.size
+
+        // Natural Wonders sort order index to name
+        val naturalsIndexMap: Map<Int, String> = allNaturalsMap.keys
+            .sortedWith(compareBy(collator, { it.tr() }))
+            .withIndex()
+            .map { it.index to it.value }
```
yairm210, line 115: .toMap accepts a mapping function iirc

yairm210:
Pyramids is good
Easily recognizable
I think it was the wonder sign in civ v somewhere as well
Maybe civpedia

core/src/com/unciv/ui/overviewscreen/WonderOverviewTable.kt
yairm210: But maybe pyramids on a .addCircle white background

SomeTroglodyte: Ha! .toMap() is quite rare - the only other one so far is `val roadsMap` in TileSetStrings... Did I perpetrate that, too?

... nope, there's seven versions of toMap(), four with no arguments except the generic type, and three with a destination to fill. None with any lambda in sight.

SomeTroglodyte:
`Image(ImageGetter.getDrawable("BuildingIcons/The Pyramids").tint(Color.WHITE))` is _still_ black, so a dupe it is, or have you got a better idea?

SimonCeder:
> Maybe civpedia

Yep
civpedia
Perhaps something for our civpedia as well?

SomeTroglodyte:
You mean decorate the category buttons in pedia too? Never thought of it. Way back they _did_ fit the bleaker "landscape"... Good idea. Maybe someday in a PR far far away... And now for something completely different... Found a regression in my music player :sob:  - that gets precedence. And today was movie day :scream:

SomeTroglodyte:
Y'all gotta try this:
image
Kawaiii!

SomeTroglodyte:
> Pyramids in a white circle

Saw that late, so this PR now has white Pyramids as can be seen smallish in the corner of that thumbnail up there. 75% sure I'd like the black on white circle less - visual simplicity trumps. probably.

xlenstra:
>     * logicminimal may have a point - currently all religion/victory enabled World Wonders are listed, just as there is 1 row for every Natural Wonder in the game (so you know how many you haven't found yet as the count / map size formula is no secret - but it _may_ disclose through sorting whether a not-found NW's name is alphabetically preceding or following a found one).

I agree with logicminimal, but also for another reason: when starting in the ancient era this list already contains all wonders in the entire game (if my diff scanning was correct), while at that point you only really care about Ancient Era wonders. I'd either only show constructed wonders, wonders in the current and all previous era's or wonders falling into either category

SomeTroglodyte:
... no helper functions to _directly_ get a wonder's era? Also, hiding wondahz already obsolete at game start seems like an unnecessarily long piece of code - I _must_ be missing something. Can't compile anyway since I blindly allowed the latest kotlin plugin to do its thing.

SomeTroglodyte:
Coming up:
- Omits the obsolete wonders when starting with a later era
- Grouped by era with a little era color
- Filter wonder.era <= viewingCiv.era (a wonder actually existing always trumps all other filters)
- Knowing about the existence of a natural wonder but not where it is now works (duh, as if *your* QuestManager could know anything about *your* Quests)
- Click on icon -> pedia, on location -> map

SomeTroglodyte:
<details><summary>Fresh screenie</summary>

image
</details>

Yes the test civ knows about Grand Mesa only thanks to a Quest...

PR #5299, opened 2021-09-22
Added the Byzantine empire

xlenstra:
Also some small refactores in choosing beliefs, most notably removing the BeliefContainer in favour of an Array and a Counter.

Added uniques:
- "May choose [amount] additional [beliefType] beliefs when [founding/enhacing] a religion"


android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -800,6 +800,32 @@
...
+		"declaringWar": "It is always a shame to destroy a thing of beauty. Happily, you are not one.",
+		"attacked": "Now darling, tantrums are most unbecoming.  I shall have to teach you a lesson.",
+		"defeated": "Like a child playing with toys you are.  My people will never love you, nor suffer this indignation gracefully.",
+		"introduction": "My, isn't this a pleasant surprise - what may I call you, oh mysterious stranger?  I am Theodora, beloved of Byzantium.",
+		"neutralHello": "Hello again.",
+		"tradeRequest": "I have heard that you adept at certain kinds of ... interactions.  Show me.",
+		"outerColor": [114, 162, 233],
+		"innerColor": [61, 0, 109],
+		"uniqueName": "Patriarchate of Constantinople",
+		"uniques": ["May choose [1] additional [Any] beliefs when [founding] a religion"],
```
yairm210, line 822: This looks ugly as text, but I understand the problem... I can't think of a good phraseology that would accept both e.g. Pantheon and Any..
Although I am generally loathe to do so, I think this is a case where we should split it into "May choose [1] additional belief(s) of any type when [founding] a religion"

android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -800,6 +800,32 @@
...
+		
+		"cities": ["Constantinople", "Adrianople", "Nicaea", "Antioch", "Varna", "Ohrid", "Nicomedia", "Trebizond", "Cherson", "Sardica", 
+			"Ani", "Dyrrachium", "Edessa", "Chalcedon", "Naissus", "Bari", "Iconium", "Prilep", "Samosata", "Kars", "Nicopolis", "Theodosiopolis", 
+			"Tyana", "Gaza", "Kerkyra", "Phoenice", "Selymbria", "Sillyon", "Chrysopolis", "Vodena", "Caesarea", "Traianoupoli", "Constantia", "Athens",
```
Mape6, line 826: THe city name `Athens` is already used for the Greece nation. Will that be a problem?

xlenstra: It's far enough down that it should almost never be reached, and even then the problems should be solely visual so I don't really mind them

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -339,35 +339,26 @@ object NextTurnAutomation {
         )
     }
 
-    private fun chooseBeliefs(civInfo: CivilizationInfo, beliefContainer: BeliefContainer): HashSet<Belief> {
+    private fun chooseBeliefs(civInfo: CivilizationInfo, beliefsToChoose: Counter<BeliefType>): HashSet<Belief> {
```
SomeTroglodyte, line 342: Is that the first use of `Counter<T>` where `T != String::class`? Wow! Congrats!

docs/Credits.md
```diff
@@ -42,7 +42,8 @@ Unless otherwise specified, all the following are from [the Noun Project](https:
 * [Bow](https://thenounproject.com/search/?q=bow&i=101736) By Arthur Shlain for Bowman
 * [Fishing Vessel](https://thenounproject.com/term/fishing-vessel/23815/) By Luis Prado for Work Boats
 * [Greek Trireme](https://thenounproject.com/search/?q=ancient%20boat&i=1626303) By Zachary McCune for Trireme
-* [Greek Trireme](https://thenounproject.com/search/?q=ancient%20boat&i=1626303) By Zachary McCune for Quinquereme
+* [Greek Trireme](https://thenounproject.com/search/?q=ancient%20boat&i=1626303) By Zachary McCune for Quinquereme // These credits are incorrect
+* [Norse Longship](https://thenounproject.com/zmccune/uploads/?i=1626302) by Zack McCune for Dromon
```
SomeTroglodyte, line 46: Way too Viking.

docs/Credits.md
```diff
@@ -42,7 +42,8 @@ Unless otherwise specified, all the following are from [the Noun Project](https:
 * [Bow](https://thenounproject.com/search/?q=bow&i=101736) By Arthur Shlain for Bowman
 * [Fishing Vessel](https://thenounproject.com/term/fishing-vessel/23815/) By Luis Prado for Work Boats
 * [Greek Trireme](https://thenounproject.com/search/?q=ancient%20boat&i=1626303) By Zachary McCune for Trireme
-* [Greek Trireme](https://thenounproject.com/search/?q=ancient%20boat&i=1626303) By Zachary McCune for Quinquereme
+* [Greek Trireme](https://thenounproject.com/search/?q=ancient%20boat&i=1626303) By Zachary McCune for Quinquereme // These credits are incorrect
```
SomeTroglodyte, line 45: @SimonCeder  - do you remember? Nice image, but I can't find it on tnp.

SimonCeder: The attribution is correct, I modified it a bit which I believe is allowed under the license. African Forest Elephant and Carthage symbol are also modifications.

SomeTroglodyte: You edited the second sail onto the Trireme... Nicely done. Should the credits mention that it's not a 'direct' usage?

docs/Credits.md
SomeTroglodyte: I looked at the wikipedia for general expectations and scanned pages of tnp icons - those with rows of oars are rare to begin with, and so I came up empty. Another gimp maybe.   (as long as it's not [this one](https://thenounproject.com/term/boat/1611891) )

docs/Credits.md
SimonCeder: The Creative Commons license suggests the phrase "The original work has been modified." I'll make a PR labeling my modifications.

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
yairm210: Yeah that turned out to be much less useful than I expected

SomeTroglodyte: On the other hand, Counter serializes horribly (using explicit class annotations for trivial Int's), and I always wanted to try how that would look if we un-generic'ed that class. Made some attempts at hooking into the serializer to teach it better manners, testing with precisely such a Counter<enum>, but got blocked by enum as an itself generic class type not being available at compile time to feed the setSerializer thingy, so I went with <*> and then couldn't access needed methods from the superclass due to some creepy "you're not allowed to call super because you're on Java 6", obviously untrue...

docs/Credits.md
xlenstra: I have to say that gimping isn't my strong suit, so I've for now taken a TNP image that roughly resembles the icon of the original

SomeTroglodyte: Why not - it's not a ship, but readily recognizable as different from other ships. And cute, too.

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -204,10 +204,23 @@ class ReligionManager {
...
+    fun getBeliefsToChooseAtFounding(): Counter<BeliefType> {
+        val beliefsToChoose: Counter<BeliefType> = Counter()
+        beliefsToChoose.add(BeliefType.Founder, 1)
+        beliefsToChoose.add(BeliefType.Follower, 1)
         if (shouldChoosePantheonBelief)
-            return BeliefContainer(pantheonBeliefCount = 1, founderBeliefCount = 1, followerBeliefCount = 1)
-        return BeliefContainer(founderBeliefCount = 1, followerBeliefCount = 1)
+            beliefsToChoose.add(BeliefType.Pantheon, 1)
+            
+        for (unique in civInfo.getMatchingUniques("May choose [] additional [] beliefs when [] a religion")) {
```
yairm210, line 214: Please enum-ify this into "May choose [amount] additional [beliefType] beliefs when [foundingOrEnhancing] a religion"
That last param will be parsed as an "unknown param", which is absolutely fine! That's why it exists in the first place :)

PR #5303, opened 2021-09-23
Precursor to mod target check -

yairm210:
Define acceptable targets for each unique holder
We can then compare each unique against the acceptable targets of the host and warn against incorrectly placed uniques

### DIFF
--- a/core/src/com/unciv/models/ruleset/IHasUniques.kt
+++ b/core/src/com/unciv/models/ruleset/IHasUniques.kt
@@ -1,6 +1,7 @@
 package com.unciv.models.ruleset
 
 import com.unciv.models.ruleset.unique.Unique
+import com.unciv.models.ruleset.unique.UniqueTarget
 import com.unciv.models.ruleset.unique.UniqueType
 
 /**
@@ -8,7 +9,9 @@ import com.unciv.models.ruleset.unique.UniqueType
  */
 interface IHasUniques {
     var uniques: ArrayList<String> // Can not be a hashset as that would remove doubles
+    // I bet there's a way of initializing these without having to override it everywhere...
     val uniqueObjects: List<Unique>
+    val uniqueTarget: UniqueTarget
     
     fun getMatchingUniques(uniqueTemplate: String) = uniqueObjects.asSequence().filter { it.placeholderText == uniqueTemplate }
     fun getMatchingUniques(uniqueType: UniqueType) = uniqueObjects.asSequence().filter { it.isOfType(uniqueType) }
--- a/core/src/com/unciv/models/ruleset/IHasUniques.kt
+++ b/core/src/com/unciv/models/ruleset/IHasUniques.kt
@@ -11,7 +11,7 @@ interface IHasUniques {
     var uniques: ArrayList<String> // Can not be a hashset as that would remove doubles
     // I bet there's a way of initializing these without having to override it everywhere...
     val uniqueObjects: List<Unique>
-    val uniqueTarget: UniqueTarget
+    fun getUniqueTarget(): UniqueTarget
     
     fun getMatchingUniques(uniqueTemplate: String) = uniqueObjects.asSequence().filter { it.placeholderText == uniqueTemplate }
     fun getMatchingUniques(uniqueType: UniqueType) = uniqueObjects.asSequence().filter { it.isOfType(uniqueType) }
--- a/core/src/com/unciv/models/ruleset/IHasUniques.kt
+++ b/core/src/com/unciv/models/ruleset/IHasUniques.kt
@@ -11,6 +11,9 @@ interface IHasUniques {
     var uniques: ArrayList<String> // Can not be a hashset as that would remove doubles
     // I bet there's a way of initializing these without having to override it everywhere...
     val uniqueObjects: List<Unique>
+    /** Technically not currently needed, since the unique target can be retrieved from every unique in the uniqueObjects,
+     * But making this a function is relevant for future "unify Unciv object" plans ;)
+     * */
     fun getUniqueTarget(): UniqueTarget
     
     fun getMatchingUniques(uniqueTemplate: String) = uniqueObjects.asSequence().filter { it.placeholderText == uniqueTemplate }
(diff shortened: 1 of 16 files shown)


core/src/com/unciv/models/ruleset/IHasUniques.kt
```diff
@@ -1,14 +1,17 @@
...
 import com.unciv.models.ruleset.unique.Unique
+import com.unciv.models.ruleset.unique.UniqueTarget
 import com.unciv.models.ruleset.unique.UniqueType
 
 /**
  * Common interface for all 'ruleset objects' that have Uniques, like BaseUnit, Nation, etc.
  */
 interface IHasUniques {
     var uniques: ArrayList<String> // Can not be a hashset as that would remove doubles
+    // I bet there's a way of initializing these without having to override it everywhere...
```
SomeTroglodyte, line 12: Never found it. Interfaces can have default implementations for methods, but not for properties with backing fields.

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -51,6 +51,8 @@ class ModOptions : IHasUniques {
     override var uniques = ArrayList<String>()
     // If this is delegated with "by lazy", the mod download process crashes and burns
```
SomeTroglodyte, line 52: Should be fixable, I mean the "crashes and burns" part

SomeTroglodyte:
I admit I don't understand the point. What is `target` meant to be? The scope where a unique can be attached? In that case the previous const implementation as far as the ruleset objects go is correct. Or the scope where the effect will apply? Then why start out with the other usecase and then switch over? Those scopes are independent, so you'll need both anyway - unless you can hide one behind hardcoding.

The inheritance thingy is a nice touch, definitely. Sorry, forgot, that word has been officially changed by majority vote into ["definately"](www.d-e-f-i-n-i-t-e-l-y.com)...

yairm210:
@SomeTroglodyte The problem we're trying to solve is - how do I warn modders that the unique they put on X is a unique that doesn't actually work on X?
The solution is to mark for each unique what types of objects it can go on - already done - and for each object, what type it is.
Step 2 will be checking (in rulesetCheck) each unique that belongs to a IHasUniques, and if the unique's type is not compatible with the object's type, add a warning.

This is slightly complicated by the fact that - and this is a good thing - uniques come in groups.
There are global uniques, which can be put on any of the global objects.
There are unit uniques, which can be put on any of the unit objects.
We still DO want to subdivide these, both so we can specify uniques relevant only to subparts AND to display to the user where the effects are coming from, therefore inheritance.

yairm210:
@xlenstra Oof. There's no problem having 2 types of belief uniques - `FounderBelief(Global)` and `FollowerBelief`. But if you say that there are city-local uniques, this becomes a problem, which is that buildings can contain BOTH local AND global uniques.
So then what?
```
Global,
Local,
Building(Global,Local),
FounderBelief(Global),
FollowerBelief(Local)
```
?

xlenstra:
> @xlenstra Oof. There's no problem having 2 types of belief uniques - `FounderBelief(Global)` and `FollowerBelief`. But if you say that there are city-local uniques, this becomes a problem, which is that buildings can contain BOTH local AND global uniques.
> So then what?
> 
> ```
> Global,
> Local,
> Building(Global,Local),
> FounderBelief(Global),
> FollowerBelief(Local)
> ```
> 
> ?

I may have written that message a bit hastely, the comparison I wanted to make was the global/local source distinction, where follower uniques were more akin to buildings in that they originate from a single city. However, unlike buildings, they only apply to that city, while all building uniques (as far as I'm aware and at least for now) apply globally, unless they have an "in this city" parameter, so it should be fine

yairm210:
Thanks for clarifying!

SomeTroglodyte:
> warn modders

Precisely why I ask - so from my confused comment we're staying with usecase 1. Data flows json -> uniques: Collection<String> -> uniqueObjects: Collection<Unique>, and there's no "sideload", everything comes from json. Thus, at the point where that last step runs for e.g. Nations, the target is static `UniqueTarget.Nation` - I don't see the need to make that overridable. If you're using a Unique in a Nation that can work on, say, a building, too - then that target would still be Nation and that works because the specific Unique's type is one that maps to Nation through its 'inheritance' hierarchy? Or so I would imagine...

yairm210:
AHA
I see what you mean, and you are CORRECT

yairm210:
I'm still in favor if keeping it, though, because of 
A. modOptions
B. Future uses for "potential" uniques (more modders)

yairm210:
Wait no, that would mean it needs to be a static for the class and not per instance...
Arrgh
How annoying

SomeTroglodyte:
Confusing enough so I don't get what you mean with modOptions or "potential" -> bring out the "KISS principle" weapon, that heavy brutal morningstar.

Step back, user story (user in this case being us): Global table of possible uniques, and we want to sound the alarm when someone attaches "can occur between temperature [A] and [B]" to - say a Policy. That's json file scope, and we want to have a "field" in that "database table" telling us which unique type is allowed where. Implementation detail - that table is actually an enum class + properties, and each row an enum value. Fine, works, promises to be efficient. The scope can be expressed as regex pattern for the filename, single-valued enum, enumSet, a delegate doing the check, whatever. We chose enum UniqueTarget with a parent-child hierarchy to express multitarget scoping. 

Valid implementation choice, user story fulfilled.

Now - the idea to generalize to allow more freedom -> need a clear picture what for, why, and why the advantages will outweigh the inconvenience of complicating a simple solution - and that's right where I can't follow your vision so I need to ask.

yairm210:
K
So I feel the current pr is in a good spot right now.
Can we dispose of the getType function in the IHasUniques and instead use the type that's in the uniqueObjects? Yes, yes we can.
But I still prefer to keep the function, because it will be relevant to my next PR.

SomeTroglodyte:
> dispose of the getType function in the IHasUniques ... keep

But when the upheaval has settled down, such synergies / potential deduplications should be checked. Maybe leave a todo...

PR #701, opened 2019-05-04
added air icons set

Am-per-Sand:
May be turns out handy one day.

Post your opinion below. However I promise to introduce corrections only after I will have seen them in game. Earlier it's hard to say if the images are crisp enough or not.

_Based on antique catalogue photos and public domain works by US military employees, so I release the to public domain too._

Zero chasing B-17:

PR #5304, opened 2021-09-23
"The original work has been modified." in attributions

SimonCeder:
The [CC-BY 3.0](https://creativecommons.org/licenses/by/3.0/us/legalcode) license suggests adding the phrase "The original work has been modified." when appropriate. Since I added a couple modified icons, this PR adds that phrase to those attributions.


SomeTroglodyte:
Sorry to have prodded you so impolitely - you see, I was researching other minor issues about the credits when I saw Xander's comment, so I had to bark...
(Those others: 1 dead link `* [Connection](https://thenounproject.com/search/?q=connection&i=1521886) By Travis Avery` and the icons from #701 aren't mentioned at all - the PR states the PR submitter @Am-per-Sand is the author themselves and they place the icons in the public domain, and for completeness I think that should be mentioned. As I haven't found out a better attribution for case 1, I haven't yet done anything about this)

SimonCeder:
While we're talking about it, the "Gift" icon (Present.png) is my own work, consider it public domain.

SomeTroglodyte:
:+1: - go ahead and attribute yourself. Or "the Unciv team". I imagine someone like an auditor coming along and checking the files one by one whether they conform to our chosen license model, and then it's best there is _some_ entry for everything. Even if we painted something ourselves. I may be paranoid, of course.

PR #5307, opened 2021-09-23
Music controller - Fix pause

SomeTroglodyte:
Small tuning and fixes pause.


AdityaMH:
Can i tell bug in this issue?
I can listen unciv music even on outside game (on android).
Version: 3.17.1

PR #5308, opened 2021-09-23
Softcode Fountain of Youth

SomeTroglodyte:
... Kilimanjaro should now be only a matter of json. Not that we want it now ...
- I did _not_ attempt to fix the deviation that if I remember correctly still exists from our template game: Fountain of Youth should give +10 Happy when owned, but cannot be worked by a city? Something like that.

Only some very quick testing so far.


core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
```diff
@@ -18,6 +18,10 @@ enum class UniqueParameterType(val parameterName:String) {
         private val knownValues = setOf("Wounded", "Barbarians", "City-State", "Embarked", "Non-City")
         override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
                 UniqueType.UniqueComplianceErrorSeverity? {
+            if ('{' in parameterText) // "{filter} {filter}" for and logic
+                return parameterText.removePrefix("{").removeSuffix("}").split("} {")
```
xlenstra, line 22: Does this still work for languages that don't use spaces?

SomeTroglodyte: And I thought that code I copied there was yours... But no, blame says Yair's.
This is just a counterpart to MapUnit.matchesFilter so the unique-auto-checks will accept the same input as the filter itself.
And this is pre-translation, so... Should not be a step back? But that's why I value input - sometimes I get the feeling to have missed some aspect and can't put my finger on it, this one especially so.

SomeTroglodyte:
'nuff testing. Most relevant result from testing - Those 3-4 "Double movement" uniques need merging, which emphasizes again some thinking about and/or for terrainFilters would be nice.

core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
yairm210: This looks good to me!
The translation of said parameters should be dealt with, but definitely outside this PR

PR #5312, opened 2021-09-24
Unified many implementations of common interfaces into abstract classes

yairm210:
Here it is ladles and jellyspoons, the abstract classes that probably should have been put in place ages ago.
Not only does this decrease the number of repetitive lines in the codebase and unify existing implementations, it will also allow us future, ruleset-object specific implementations.

### DIFF
--- a/core/src/com/unciv/models/ruleset/RulesetObject.kt
+++ b/core/src/com/unciv/models/ruleset/RulesetObject.kt
@@ -1,0 +1,28 @@
+package com.unciv.models.ruleset
+
+import com.unciv.models.ruleset.unique.Unique
+import com.unciv.models.stats.INamed
+import com.unciv.models.stats.NamedStats
+import com.unciv.ui.civilopedia.FormattedLine
+import com.unciv.ui.civilopedia.ICivilopediaText
+
+abstract class RulesetObject: INamed, IHasUniques, ICivilopediaText {
+    override lateinit var name: String
+    override var uniques = ArrayList<String>() // Can not be a hashset as that would remove doubles
+    @delegate:Transient
+    override val uniqueObjects: List<Unique> by lazy {
+        uniques.map { Unique(it, getUniqueTarget(), name) }
+    }
+    override var civilopediaText = listOf<FormattedLine>()
+    override fun toString() = name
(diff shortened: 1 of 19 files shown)


core/src/com/unciv/models/ruleset/RulesetObject.kt
```diff
@@ -0,0 +1,28 @@
...
+
+// Same, but inherits from NamedStats - I couldn't find a way to unify the declarations but this is fine
+abstract class RulesetStatsObject: NamedStats(), IHasUniques, ICivilopediaText {
+    override var uniques = ArrayList<String>() // Can not be a hashset as that would remove doubles
+    @delegate:Transient
+    override val uniqueObjects: List<Unique> by lazy {
+        uniques.map { Unique(it, getUniqueTarget(), name) }
+    }
+    override var civilopediaText = listOf<FormattedLine>()
+}
```
SomeTroglodyte, line 28: If prefer github to not display that red traffic sign when a file doesn't end in a linefeed (though the [kotlinlang.org conventions](https://kotlinlang.org/docs/coding-conventions.html#formatting) do not say anything on that matter) :laughing:

SomeTroglodyte (changes requested):
:+1:  - one "food for thought": For I while I had both `ICivilopediaText` and `CivilopediaText`, interface and abstract with the default implementations for fields. I dropped that mainly because you can implement as many ifaces as you want, but inherit only one superclass, and didn't want to impose limitations.

Here, it's definitely good this now provides *the* base, but maybe having the interface too as clearer documentation / for use in parameter passing...

Wait a minute - RulesetObject and RulesetStatsObject are independent! Ahem, changing Approve to Request changes...

`interface IRulesetObject: INamed, IHasUniques, ICivilopediaText` does compile, at a minimum...

yairm210:
How would you suggest I unify them? Just with an interface?

SomeTroglodyte (approved):
Exactly. Having ***one*** type - whatever kind of - apply to ***all*** of them is the goal, right? I can talk haughtily like that easily, as I won't be doing any testing on Unicv for a few hours :wink:

PR #5319, opened 2021-09-25
Double movement unique parameterized

SomeTroglodyte:
This should not change game behaviour*. It unifies all "double movement" uniques, and brings the block of cached flags UnitMovementAlgorithms uses to the UnitType system, and I tried to limit myself to just that...
- *: Except now "double" really means "double" when a mod alters movement costs of the affected terrains.

Open question: "Double movement in [Marsh]" will not be flagged, but will do nothing - the implementation still hardcodes only six terrains. Would a HashSet<String> still be fast enough?

Side Question: I still got in the back of my mind that using most of our objects on the key side of a HashCollection is asking for trouble, as they aren't properly implementing the equality contract. Loose Ideas: The interface might do it - just comparing type and name should be enough after all. Another loose idea: We still have `override equals` that miss the required signature and thus look like they would be used by an `==` (or contains or or or...) but aren't - should we ditch them some time?


core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -208,8 +206,16 @@ class MapUnit {
...
+        tempUniques.asSequence().filter { it.type == uniqueType } +
+            civInfo.getMatchingUniques(uniqueType)
+
     fun hasUnique(unique: String): Boolean {
-        return getUniques().any { it.placeholderText == unique } || civInfo.hasUnique(unique)
+        return tempUniques.any { it.placeholderText == unique } || civInfo.hasUnique(unique)
+    }
+
+    fun hasUnique(uniqueType: UniqueType): Boolean {
+        return tempUniques.any { it.type == uniqueType } || civInfo.hasUnique(uniqueType)
```
yairm210, line 218: getMatchingUniques.any() perhaps

yairm210: So we don't need to handle 'where are the uniques coming from' in 2 places

yairm210:
This looks good.
As for using hashsets - no, but actually maybe yes!
The reason this was hardcoded to start with us because of the "loads of units each creating hash in every possible movement equals loads of time"
BUT
If we use alayered approach - adding a "hasDoubleMovementOnAnyTerrain" boolean, for instance - that can be our initial check, returning false for most units, and then only the minority of units that *actually* need the hashset check can use it!

SomeTroglodyte:
> initial check + HashSet

Hey, simple ideas are always best - but also easy to overlook. :+1: May need two gates, however - one before `roughTerrainPenalty` and `ignoreHillMovementCost` one after? Or include those in the gate. I'll take a look.

> getMatchingUniques.any

- It's handled exactly that way elsewhere (`CivilizationInfo`).
- :+1: Simple.
- `getMatchingUniques(type/placeholder).any{param match}` is a pattern - rare (4 occurrences) but it exists
- `getMatchingUniques(type/placeholder).any()` has 5 occurrences not translated to `hasUnique`.
- `MapUnit` already had the independent approach, as does `IHasUniques`.
- More overhead. The Sequence `getMatchingUniques` enforces needs instantiation first before it even starts looking, and Sequences are pretty complex under the hood. `ArrayList.any` _may_ be much simpler under the hood as everything about the actual "structure" of the required iterator is known at compile time. May be moot as the none case is more likely and then it will pass on onto a `Sequence` anyway.

Hmmm ...... ..... .... ... .. Yes. Simpler is better.

Maybe some thoughts on standardizing?
- `IHasUniques` now has default functions for all four signatures mentioned above. Another pair of overloads may be useful - `getMatchingUniques(uniqueType: UniqueType, parameterPredicate: (Unique)->Boolean): Sequence<Unique>` and `hasUnique(uniqueType: UniqueType, parameterPredicate: (Unique)->Boolean): Boolean` - ?
- The `uniqueTemplate: String` overloads [cs]ould get a `@Deprecated` soon and then keep it for a good while?
- Individual implementations exist for CivilizationInfo, CityInfo and MapUnit - they need more complexity because target scope and effect scope of a unique can be different, so they need to "broaden" their search - and in the city case Xander has built a monster of a pass-along-through-layers system I wouldn't want to really understand in depth as I would be afraid of brain overheating. Powerful but not very transparent. Maybe think about a second scoping property for the effect locus? Then such nesting could be automatic and a generic implementation possible. Just food for thought...
- ***None*** of these individual implementations are overrides of the abstract+default in IHasUniques! City and Civ use differing signatures, while MapUnit doesn't inherit the interface. Needs some mulling over.

yairm210:
- I think parameter predicates are the same as chaining a .filter  after getting the initial sequence, so I don't see a great need for that. Filter does what it does impeccibaly.
- uniqueTemplate overloads  to get 'deprecated' - good idea, will only affect devs, and will show the correct way forward
- "in this city" and "in all other cities" make stuff complex, this does need simplifying, agree. I have one idea - marking the city each unique comes from - but I don't like it much.

SomeTroglodyte:
By the way:
image
... an absolute gem. I'll choose 'Add to dictionary'...

SomeTroglodyte:
OK, so "Double movement" can now handle any tileFilter. That part of the code hasn't seen any testing, though - would have to set up a mod just for that... Advantage: Needs no special treatment in the UniqueType system.

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -40,26 +47,38 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
...
+            return terrainCost * 0.5f + extraCost
 
-        return to.getLastTerrain().movementCost.toFloat() + extraCost // no road
+        if (unit.noFilteredDoubleMovementUniques)
+            return terrainCost + extraCost
+        if (unit.doubleMovementInTerrain.any { 
+            it.value == MapUnit.DoubleMovementTerrainTarget.Filter &&
+                to.matchesFilter(it.key)
+        })
+            return terrainCost * 0.5f + extraCost
```
yairm210, line 79: A. I think it's time we deprecated the old Hill, but that's for a different PR
B. if/else? If a unit has multiple, I don''t think they should stack. Maybe that's just me.

SomeTroglodyte: They don't stack - it's peppered with `return`s. Case for a `when` but when AS doesn't whine I tend to leave the old code structure.

And yes this still uses Constants.hill outside deprecation, but that needs more thorough thinking and a clear vision how-to which I currently don't see.

Issue #4987, opened 2021-08-25
Help Needed: Ideas for UI for seeing followers in foreign cities


xlenstra:
Recently, I've added a UI for viewing how many followers and pressures there are in your own cities. However, in the original game you could also view this for other cities, by hovering your mouse over the majority religion icon on citybuttons. Since 'hovering your mouse' is not a viable action to do in Android apps, this is not a viable way of implementing it into unciv. I have trouble coming up with other ways to do this, however, especially as I also don't want to compromise on the 'open diplomatic overview of civ' when clicking on their city buttons.
Does anyone have an idea for how to implement a UI for this?

logicminimal:
opening their diplomacy page takes 2 clicks, when someone clicks it once a ui showing how much followers every religion has. something like this pop ups either instantly or with an animation like pressing a city for the first time, ascending it, im sure you can make this fancier

![20210825_150843.jpg](https://user-images.githubusercontent.com/56484801/130787899-90b5a18f-9af3-4e2f-9548-f8c397712ac3.jpg)

(i actually didnt mean to add the red line but as an euroka moment it could be used to seperate the majority religion from others, which in this case is judaism (what happens if two religions have same amount of followers)

AdityaMH:
Add button like worker use tile improvement when click someone city and lead to pop-up like get notification from builded wonders.

I think we can set this for tile improment too instead blank blue color.

logicminimal:
on another thought, when you press any city (yes any city, including your own) a small pop up could appear on top of the screen or to the below of it, shoeing city name, population, and religions, i think it would be better if it had a brown background, like a wooden plank, no idea why i think brown would be good for this though

ravignir:
![ed](https://user-images.githubusercontent.com/30041793/130803292-3ea91f4e-6f76-4a45-90db-cc4094e6a15e.png)
I will be damn original.
The only problem i can see with this, the resolution may be too low.

AdityaMH:
Hm, still no idea?

xlenstra:
Closing this as it was added in #5321

PR #5321, opened 2021-09-26
Mini-UI to see Religion info on foreign cities

SomeTroglodyte:
Offer for #4987 

This is a Popup "inserted" between the second click on a foreign city button and the diplomacy. Pass-thru when Religion off.
Safer than putting it in UnitTable - that might be nicer for the user, but managing the situation where the "selection" is a foreign city gets complicated quickly.

Open question: What _else_ about a foreign city should a player be allowed to see that's currently unavailable? Wonders? Spies? :grin:

image

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityReligionInfoTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityReligionInfoTable.kt
@@ -1,0 +1,77 @@
+package com.unciv.ui.cityscreen
+
+import com.badlogic.gdx.graphics.Color
+import com.badlogic.gdx.scenes.scene2d.ui.Table
+import com.unciv.logic.city.CityInfoReligionManager
+import com.unciv.models.Religion
+import com.unciv.ui.utils.*
+
+class CityReligionInfoTable(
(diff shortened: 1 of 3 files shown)


SimonCeder:
Spies should allow you to see the entire city view. Your own spies should be marked on the city button. I've been working a bit on spies btw.

SomeTroglodyte:
Then, as interim solution, this better than nothing?

SimonCeder:
Sure, the one does not exclude the other.

yairm210:
@SimonCeder I won't be accepting Espionage PRs until 
more threads that are already open are resolved, we're already working on lots of things at the same time and espionage can really be put of until we've finished more things

SomeTroglodyte:
> won't be accepting Espionage

Valid good argument, but on the other hand - good to know someone's having the drive to start on that! :+1: Simon!

xlenstra:
One thing I did notice, is that an empty table is shown in cities in which there are no followers of any kind, is that the best approach? To me it feels nicer if the table was entirely absent in that case, just my two cents

PR #5325, opened 2021-09-26
Fix Polynesian vision when embarked

SimonCeder:
Properly adds the + vision when embarked bonus, even to units that are not marines.


core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -264,10 +264,15 @@ class MapUnit {
...
-        
         var visibilityRange = 2
+
+        if (isEmbarked() && !hasUnique("Normal vision when embarked")) {
+            visibilityRange = 1
+            for (unique in getMatchingUniques("[] Sight for all [] units"))
+                if (unique.params[1] == "Embarked")
+                    visibilityRange += unique.params[0].toInt()
+            return visibilityRange
+        }
```
xlenstra, line 275: I don't think this matches the original. If you now have an embarked unit that would have extra vision on land, it will now have extra vision in water tiles, which is not what happens in the original

xlenstra: I'd probably solve it for now with a new 'embarked units have normal vision' unique, which will then eventually be replaced by "Normal vision when embarked <for [all] units>"

SimonCeder: Look again - it's only counting bonuses that are explicitly for embarked units. I'll add a comment so it's more clear what's going on.

xlenstra: My vision probably needs to improve, my bad :)

SomeTroglodyte: Just add a "[+3] Sight for all [Dutch] units <when inspecting code>" unique to that slot at the base of your skull...

PR #5329, opened 2021-09-26
Proper great general points

SimonCeder:
Two modifications to great general points to bring them in line with the original (and make them much scarcer!):

- 200 points are needed for the first general (not 30)
- You can't get great general points from fighting barbarians


logicminimal:
both of those should be generalized in the future

PR #5331, opened 2021-09-27
Filipino.properties

SunnySunset:
Sorry if I was not able to continue this. I will try to continue it now.


xlenstra:
Welcome back!
The text in between the square brackets [] should not be translated, as that is replaced with an instance when we show it to the user.
For example, "Requires [buildingName] to be built in all cities" will always be converted to something like "Requires Library to be built in all cities". We use the text between the square brackets to determine what the part is that should be replaced, and for that it doesn't have to be translated.
This is the main reason your PR has that ❌ emoji.

More info on this can be found on the wiki:
https://github.com/yairm210/Unciv/wiki/Translating#pitfalls

SunnySunset:
How do I edit it now?

xlenstra:
Go here:
https://github.com/yairm210/Unciv/blob/master/android/assets/jsons/translations/Filipino.properties
And press the 'edit' pencil mark. Then once you're done, create new PR at the bottom of the page.
More information on how exactly this is done can be found on the wiki:
https://github.com/yairm210/Unciv/wiki/Translating

PR #5333, opened 2021-09-27
Fix Fountain of Youth yield as per wiki

SomeTroglodyte:
- F of Y "Tile provides yield without assigned population"
- Made that unique work on terrain
- Byzantium startBias fix - a pity - "Coats" sounds so nice. Should have done a terrain "Coats" instead.

Open question: This works to provide yield (and prevent worker assignment) when within the workable radius of a city and owned. How it should behave when owned by - 4 tiles from city center - I'm not sure.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Nations.json
+++ b/android/assets/jsons/Civ V - Vanilla/Nations.json
@@ -804,7 +804,7 @@
 		"name": "Byzantium",
 		"leaderName": "Theodora",
 		"adjective": ["Byzantine"],
-		"startBias": ["Coats"],
+		"startBias": ["Coast"],
 		"preferredVictoryType": "Cultural",
 		
 		"startIntroPart1": "All hail the most magnificent and magnanimous Empress Theodora, beloved of Byzantium and of Rome! From the lowly ranks of actress and courtesan you became the most powerful woman in the Roman Empire, consort to Justinian I. Starting in the late 520's AD, you joined your husband in a series of important spiritual and legal reforms, creating many laws which elevated the status of and promoted equal treatment of women in the empire. You also aided in the restoration and construction of many aqueducts, bridges, and churches across Constantinople, culminating in the creation of the Hagia Sophia, one of the most splendid architectural wonders of the world.",
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -251,8 +251,7 @@
 		"impassable": true,
 		"unbuildable": true,
 		"uniques": ["Must be adjacent to [0] [Coast] tiles",
-			"Grants [Rejuvenation] ([all healing effects doubled]) to adjacent [{Military} {Land}] units for the rest of the game",
-			"Tile provides yield without assigned population"],
+			"Grants [Rejuvenation] ([all healing effects doubled]) to adjacent [{Military} {Land}] units for the rest of the game"],
 		"weight": 1
 	},
 	{
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -251,7 +251,8 @@
 		"impassable": true,
 		"unbuildable": true,
 		"uniques": ["Must be adjacent to [0] [Coast] tiles",
-			"Grants [Rejuvenation] ([all healing effects doubled]) to adjacent [{Military} {Land}] units for the rest of the game"],
+			"Grants [Rejuvenation] ([all healing effects doubled]) to adjacent [{Military} {Land}] units for the rest of the game",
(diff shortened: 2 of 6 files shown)


xlenstra:
If I am not mistaken (but I very well could be), in general, happiness yields from all tiles inside your border is always provided, regardless of whether the tile is worked or not. For the fountain of youth, this turns out to be equivalent to have all tile yields always be provided, as it only produces happiness. In general, this is not the case, so there is an argument that this implementation isn't necessary when that behavior is implemented correctly

SomeTroglodyte:
> happiness yields from all tiles inside your border is always provided

OK that would be a whole different kettle of fish, implementation-wise. And this actually provides more than 10 happiness: 10 for the json entry, +1 happy just because it's a Nat Won, and +1 P +1 F you get for the turnsInto Plains...

OK, I suspect this is not original C5 (the game, not the explosive) behaviour -> but it shows what _could_ be done with that otherwise unused unique. Was that one done for some mod I don't know?

logicminimal:
what if there is also a yield when its being worked?

android/assets/jsons/Civ V - Vanilla/Nations.json
```diff
@@ -804,7 +804,7 @@
 		"name": "Byzantium",
 		"leaderName": "Theodora",
 		"adjective": ["Byzantine"],
-		"startBias": ["Coats"],
```
yairm210, line 807: Maybe they just happen to really like coats. Fair enough.

SomeTroglodyte:
OK - decision time. 11 Happiness displayed in two separate places +1 production +1 food, as long as in workable and owned space, but unable to be worked for any extra yield, is partially right but overall still wrong.

But the unique would be closer to what it advertises - "tile yield" does sound like terrain just as much as it sounds like an improvement - and it works as it is designed to do.

I vote (pushed version) the code to support the "yield without worker" unique on terrain stays in, but the Fountain doesn't get it. A potential fix to FoY would then be another PR and technically quite a bit different.

To reiterate what I now believe it should do:
- +1 Happiness when found and ever after (currently works)
- +10 Happiness _instead of_ +1 when in owned territory, no matter distance from owning city (nope)
- Cannot be assigned a worker (nope - currently you can, and then it totals +11 Happiness)

@xlenstra  or @ravignir - right or wrong?

android/assets/jsons/Civ V - Vanilla/Nations.json
SomeTroglodyte: Calls for a mod where Byzantium always spawns near a Natural Wonder named "Coat Rack" and gets a unique Worker replacement named "Coats" and a Warrior replacement named "Coåts" and an Archer replacement named "Cøats" and a Scout replacement named "Çoats" - all having the exact same icon and pixel unit graphic... and so on...

ravignir:
+10 Happiness when within borders (+20 as sPAIN), +1 when discovered (+2 when sPAIN). Natural wonders stats should override base terrain stats.

@SomeTroglodyte Off-topic: I was supposed to take a break from Unciv 😅

xlenstra:
> OK - decision time. 11 Happiness displayed in two separate places +1 production +1 food, as long as in workable and owned space, but unable to be worked for any extra yield, is partially right but overall still wrong.
> 
> But the unique would be closer to what it advertises - "tile yield" does sound like terrain just as much as it sounds like an improvement - and it works as it is designed to do.
> 
> I vote (pushed version) the code to support the "yield without worker" unique on terrain stays in, but the Fountain doesn't get it. A potential fix to FoY would then be another PR and technically quite a bit different.

Agree

> To reiterate what I now believe it should do:
> 
>     * +1 Happiness when found and ever after (currently works)
> 
>     * +10 Happiness _instead of_ +1 when in owned territory, no matter distance from owning city (nope)
> 
>     * Cannot be assigned a worker (nope - currently you can, and then it totals +11 Happiness)
> 
> 
> xlenstra or ravignir - right or wrong?

Based on a quick test I did, the first two seem correct: the +10 overrides the normal +1. Didn't have the time to test the third

SomeTroglodyte:
A break? Sorry to interrupt, then. Brakes are important! Especially downhill...

xlenstra:
> Based on a quick test I did, the first two seem correct: the +10 overrides the normal +1. Didn't have the time to test the third

Also tested the third, and this seems to indeed be the case. Moreover, happiness doesn't seem to even show up as a yield icon at all, and no tile with 0 yield appears to be able to be worked. I have also tested the first two further and the +10 does in fact _not_ override the +1 from finding the natural wonders. So if you have the fountain of youth inside your borders, you have +11 happiness from it.

yairm210:
That is all extremely counter intuitive. I think it makes for a worse gameplay experience, since it goes against everything the game has taught you so far. I'm against adding this to the game.

SomeTroglodyte:
OK. I'd still like to know what `TileProvidesYieldWithoutPopulation` is actually used for.

PR #5335, opened 2021-09-27
Rework game start again (continents)

SomeTroglodyte:
- Less memory - the only once used landTilesInBigEnoughGroup isn't kept in memory after use
- Instead generated from the continent ID's - slower but not much, cheaper than the repeat BFS runs from way before
- hardcoded minimum continent size 20 replaced by 75% landmass or each their island
- hardcoded Snow exclusion removed

And then something I tried with the Aliens mod:
- Natural wonder placement uniques expanded (mods only benefit)
- startBias = Natural wonder made possible (mods only benefit)

Those Aliens, mod updated and this branch running, start like this:
image

Don't know whether any other mods would want that, but it has great scenario potential.


SimonCeder:
Something that occurred to me - if someone generated a map in the map editor, and then edited it changing the landmasses, and saved it - the continent data would be wrong. We should probably regenerate it on all premade maps to be sure.

SomeTroglodyte:
> continent data would be wrong

I thought I had considered the case and it came up clean, but now that you're saying it seems obvious you're right. The editor could either clear continent data early on and rely on the loader to repopulate (not really what we went for making TileInfo.continent & TileMap.continentSizes serializable) - or rerun on save... Not today.

PR #5341, opened 2021-09-27
Civilopedia category icons and keyboard navigation

SomeTroglodyte:
Someone said something about reusing those pyramids on the pedia category buttons?
image
+ Does keyboard navigation including arrows now, and finally scrolls the preselected "Tutorials" button into view.

That's reusing anything I could find - across several texture atlases. Maybe not optimal. Ideas?

Instead of the city state capitol for "nations" I could offer:
image
... because: colors, reminiscent of nations but no repetition

And maybe someone knows an answer to that `//todo`.


logicminimal:
if the city state icons will be like ones in civ 6 (specific icons for every city state) im all in for it but if its just a one texture thats recoloured for every city state its not worth it

yairm210:
Images look good to me 🙂

xlenstra:
I personally have a small dislike for the resource & terrain icon having colour, while the others don't. I don't have better icons myself, so I'm fine with it as is, just a thought about consistency :)

SomeTroglodyte:
> Images look good

Better than nothing and easily changed later - but what's your take on them being in different texture packs? Irrelevant as pedia is not performance critical and depending on active category it will pull those textures anyway? Or reason enough to ensure the "game" pack can serve _all_ header icons?

> consistency

Glad I'm not the only one to think that way. Half the reason to paint that generic nation one. If above question comes out >50% yes I might go and _add_ colour to the currently white ones. For these, 50x50 should suffice so extra area in the texture isn't _that_ critical. One or two of the "abused" existing ones are 50x50 already.

SomeTroglodyte:
... on second look, there's quite a few, so maybe not. Some white alternative to the smiley for resources?

logicminimal:
actually i think the modders should be able to give custom icons amd colours to their city states, a new unciv project for me.

logicminimal:
actually, the people icon seem to be better for nations tab, i though u were gonna replace the city state capitol with it. also maybe use oil for esource and a mountain from nounproject for terrains? unciv sprites don't really fit here

SomeTroglodyte:
All in white best shot so far:
image

SomeTroglodyte:
Oops, this is still missing the atlas - was distracted by [this](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Demos/GeoffKnorrMusicScraper.md). Soonish.

PR #5343, opened 2021-09-28
implements isRoughTerrain for city expansion coefficient

logicminimal:
is a made up value for now as i dont have the data needed. Relates to #4697

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -147,6 +147,10 @@ object Automation {
             else if (distance <= 3) score -= 104
             
         } else {
+            // Rough tiles are meant to have less appeal for expansion than plain ones
+            // I literally just made up the the 15 here, need absolute values for this
+            // For now it acts as a placeholder score to mark off an element in #4697
+            if(tile.isRoughTerrain()) score += 15
             // Water tiles without resources aren't great
             if (tile.isWater) score += 25
             // Can't work it anyways


yairm210:
What? why?
Usually forest and hills give you sweet production, which is often the most important thing for an expanding city

logicminimal:
> 
> 
> What? why? Usually forest and hills give you sweet production, which is often the most important thing for an expanding city

that's how its in civ 5, i can change it to base hills if you want

yairm210:
But from a gameplay perspective, why?
Like, what does this add to the game?

xlenstra:
Every terrain has a specific value of how likely a city is to expand to this tile. The main reason for this is (I assume) to resolve ties that would otherwise appear between tiles. The values of these range from -3 to +3, and most of them are documented here:
https://www.reddit.com/r/civ/comments/4o4o6j/civ_v_border_expansion_algorithm/

logicminimal:
> 
> 
> Every terrain has a specific value of how likely a city is to expand to this tile. The main reason for this is (I assume) to resolve ties that would otherwise appear between tiles. The values of these range from -1 to +3, and most of them are documented here: https://www.reddit.com/r/civ/comments/4o4o6j/civ_v_border_expansion_algorithm/ At least, that is how it is implemented in the original

so should i change this pr to add everything in that table?

yairm210:
Sounds to me like a .random() (based on city id and number tile, perhaps, to remain "constant") would solve this equally well if not better.
Alternatively, making rough terrain *more* likely is also an option :)

xlenstra:
> > Every terrain has a specific value of how likely a city is to expand to this tile. The main reason for this is (I assume) to resolve ties that would otherwise appear between tiles. The values of these range from -1 to +3, and most of them are documented here: https://www.reddit.com/r/civ/comments/4o4o6j/civ_v_border_expansion_algorithm/ At least, that is how it is implemented in the original
> 
> so should i change this pr to add everything in that table?

That is what I was referring to in the 'missing features' issue. If you want to implement this, I'd do it with an extra value in the `terrain.json` files

logicminimal:
> 
> 
> > > Every terrain has a specific value of how likely a city is to expand to this tile. The main reason for this is (I assume) to resolve ties that would otherwise appear between tiles. The values of these range from -1 to +3, and most of them are documented here: https://www.reddit.com/r/civ/comments/4o4o6j/civ_v_border_expansion_algorithm/ At least, that is how it is implemented in the original
> > 
> > 
> > so should i change this pr to add everything in that table?
> 
> That is what I was referring to in the 'missing features' issue. If you want to implement this, I'd do it with an extra value in the `terrain.json` files. Though whether this is necessary is up for debate, as Yairm noted

maybe we can use weighted randomness along with those appeals?

yairm210:
I'm still not sure what problem we're trying to solve here, and the solution depends on the problem.
If we're trying to enable the computer to decide between tiles - which appears to be the case -  then adding json params for that is a hassle for modders, so random().
If we're trying to allow modders to super fine-tune the AI to prefer certain tiles above others, then json could be an option. But I really don't think that's *interesting* from a modding or gameplay perspective. So basically, "no".

xlenstra:
I'll be honest, I don't know why the original had this either. Given the very small effect it has on gameplay I'm fine with not including it.

SomeTroglodyte:
The only thing I would personally change in that function is to increase the weight at `if (distance > 3) score += 100` - which means leave it as it is unless you can improve runtimes.

logicminimal:
> 
> 
> I'll be honest, I don't know why the original had this either. Given the very small effect it has on gameplay I'm fine with not including it.

i actually want to implement that appeal thing for tiles to be generalisable

yairm210:
Before your do that, please state - what problem are you trying to solve?

logicminimal:
> 
> 
> Before your do that, please state - what problem are you trying to solve?

in missing features thread there was a mark to give less appeal to rough terrain, but since this doesnt matter rn i want the modders to be able to give terrains appeals themslves (appeal = how likely it will be chosen in auto expansion)

yairm210:
I consider that a non-problem

logicminimal:
> I consider that a non-problem

ill implement it anyway, im trying to work on simple things for nos

logicminimal:
i think the appeal thing wont have any proper use, im closing this

PR #5345, opened 2021-09-28
First step into unifying strength bonuses using conditionals

xlenstra:
This PR implements the first step into unifying the strength bonuses noted in #4777.
It unifies the "vs [combatantFilter]" combat bonuses into the new main unique for applying strength.

Added uniques:
- "[amount]% Strength"

Added conditionals:
- "vs cities"
- "vs [mapUnitFilter] units"

Deprecated uniques:
- "+[amount]% Strength vs [mapUnitFilter]"
- "-[amount]% Strength vs [mapUnitFilter]"
- "+[amount]% Combat Strength"

- "[amount]% Strength"


yairm210:
Looks really good! Excellent job!
I like the fact that the conditionals also have potential depreciation, we'll be able to add that when it becomes relevant ;)

SomeTroglodyte:
game.png changed despite no actual new graphics - which is why I *don't* run with 'Working directory' within the actual project...

SomeTroglodyte:
Style remarks:
- `StrengthMin` won't live long so it's OK but my first association was min() not minus()
- Comment emphasis (CONDITIONALS->Conditionals) - maybe - If we just want something to really stick out in the IDE, One yould use a `/** */` comment + bold markup. That works fine if and only if the next actual code element able to carry a Kdoc actually has its own Kdoc - I've recently seen at least one bad example.
- `ConditionalInTiles` - may have to differentiate whether aggressor tile or victim tile counts - do both when that goes live?

Issue #5309, opened 2021-09-24
Mods that add civilizations don't load (Android - Unciv 3.17.0)


osjasmine:
**Platform**
Android 10 (32 bit). 
Device: Samsung A10

**Version**
Unciv 3.17.0, Github (the bug was also present in F-Droid version - 3.16.15-patch2)

**Describe the bug**
I can download mods but in the new game   settings screen, when I check/select mods that add new civilizations, they don't load. 
For example when I check Ukraine Mod, I don't see Ukraine in the Civilizations list. 

**To Reproduce**
1. Download any of the many civilizations mods.
2. Start new game > Select the mod
3. Then check the civilization list. You will find that no new civilizations have been added. 

**Additional context**
This bug was present in all mods that add new civilizations except in [Kakaoalan](https://github.com/Cavenir/Kakaoalan). Kakaoalan is the only one that actually loads. 
Also the mod Difficulty_Sandbox doesn't work. Other mods like DecivRedux work.

I have already tried clearing data and reinstalling the app but no change.

osjasmine:
The last version where civilization mods work is 3.16.0. In 3.16.0-patch1 (and all subsequent versions) the mods stop working.

SomeTroglodyte:
That is very likely because many mods don't follow the rules. Which is because we're changing the rules - to make following the rules simpler in the future.

You _should_ get explicit reasons what's wrong on the "mod check" tab of the options dialog.

SomeTroglodyte:
Ukraine works perfectly fine for me. Difficulty_Sandbox works. Latin american civs work.

OK, some mods stumble over a regression due to Civilopedia text accessing a current player - which shouldn't be, Civilopedia is about globally valid info, not player-specific. But if that one hits, you'll crash, not search in vain for the nation in the picker. (fix underway)

osjasmine:
Thanks for trying to help me out.
I tried reproducing this another phone (Android 9) but I couldn't so its probably something wrong with my phone. I will try reproducing this issue on another phone running Android 10 and if Im not able to do so, you can close this issue.

Other info:
I found out that if I install a civilisation mod (like the Ukraine mod) in Unciv 3.16.0 and then update Unciv to the latest version on Fdroid (3.17.0), the mod still works (when it wouldnt work on fresh install of Unciv 3.17.0).

[...]

PR #5346, opened 2021-09-28
Fix Mod unpack folder move

SomeTroglodyte:
I got mod unpacking errors with ChooseYourDestinymod on desktop, and it boiled down to java rename not moving a folder but its contents to a destination in some cases but not in others. Related to #5309 but slightly different.

I originally inserted that helper because on Android the extra time for a copy-and-delete is noticeable. The OS _is_ capable of just moving a filesystem object without actually copying the data, so why does Gdx's moveTo avoid that? This variant, specifying the full destination name for folders (which at that point is guaranteed to not exist), fixed that for desktop runs without breaking any of the 20+ mods I tried. Tests on Android a few mods less, but all worked, including ChooseYourDestinymod.

***Question***: The more _careful_ choice would be to revert this helper entirely and put up with the ~100% longer unpacking time, increased SSD wear and tear, and increased battery drain... Your opinion?

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/GitHub.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/GitHub.kt
@@ -111,6 +111,9 @@ object Github {
     private fun FileHandle.renameOrMove(dest: FileHandle) {
         // Gdx tries a java rename for Absolute and External, but NOT for Local - rectify that
         if (type() == Files.FileType.Local) {
+            // See #5346: renameTo 'unpacks' a source folder if the destination exists and is an
+            // empty folder, dropping the name of the source entirely.
+            // Safer to ask for a move to the not-yet-existing full resulting path instead.
             if (isDirectory)
                 if (file().renameTo(dest.child(name()).file())) return
             else
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/GitHub.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/GitHub.kt
@@ -107,10 +107,15 @@ object Github {
 
         return finalDestination
     }
-    
+
     private fun FileHandle.renameOrMove(dest: FileHandle) {
         // Gdx tries a java rename for Absolute and External, but NOT for Local - rectify that
-        if (type() == Files.FileType.Local && file().renameTo(dest.file())) return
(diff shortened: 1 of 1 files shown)


yairm210:
Faster times and less wear sound like a real advantage, and the disadvantage is theoretical, no?
I'm in favor if sticking to the real
Also, please add a small comment like "see #5346"

SomeTroglodyte:
> disadvantage is theoretical

No, or we wouldn't need this PR? Again in other words: Back then I noticed Gdx.FileHandle.moveTo copies then deletes instead of moving if .type() == FileType.Local -> you can see the sources with Ctrl-B or [here](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/files/FileHandle.java#L617). I created that wrapper function that catches the case and tries the same Java call that External uses and falls back to Gdx.moveTo when that doesn't report success. _Seemed_ to work. But in some cases it doesn't - mod 'ChooseYourDestinymod' demonstrates - it has only the jsons subfolder nothing else, and that one does not get moved as asked, the function will move the folder contents instead. I'm convinced that happens if and only if the given destination is an empty folder.

So the Pro for this PR's current code is: Efficient as planned, can handle ChooseYourDestinymod, 5Hex, my music-only mods, RekMod, example map mod and whatever I tested.
The Con is: The Gdx authors had a reason to be inefficient at that point, and we don't know what exactly that is. Checks where mountpoints and volume boundaries are too expensive (but we know within our mods folder everythins is on the same volume) - or maybe there's yet another quirk in the java libraries we haven't hit yet.

An aside: java.io.File.renameTo states itself that java.nio.file.Files.move would be the better alternative, but referencing/including that is a can of worms I don't wanna open.

> I'm in favor if sticking to the real

Not sure what you mean.

> comment

Yes of course.

yairm210:
I see
I'm fine with adding this as is, and if we discover problems down the line we can dream with them then.
Right now, as you say, we have a problem and a possible solution, I don't think the potential damage is large enough to put this off for fear of the unknown

SomeTroglodyte:
'Fear of the Unknown' is like ['Tales from the Crypt'](https://en.wikipedia.org/wiki/Tales_from_the_Crypt_(comics)), right?

PR #5349, opened 2021-09-29
Add looping minimap viewport if worldwrap enabled

Thyrum:
The viewport now also appears on the other side of the minimap when viewing the border of the world with worldwrap enabled.

image

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/Minimap.kt
+++ b/core/src/com/unciv/ui/worldscreen/Minimap.kt
@@ -27,7 +27,7 @@ import kotlin.math.min
 class Minimap(val mapHolder: WorldMapHolder, minimapSize: Int) : Table(){
     private val allTiles = Group()
     private val tileImages = HashMap<TileInfo, Image>()
-    private val scrollPositionIndicator = ClippingImage(ImageGetter.getDrawable("OtherIcons/Camera"))
+    private val scrollPositionIndicators = ArrayList<ClippingImage>()
 
     init {
         isTransform = false // don't try to resize rotate etc - this table has a LOT of children so that's valuable render time!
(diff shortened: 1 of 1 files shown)


Thyrum:
The one problem with this approach is that it doesn't look good on very tiny maps with world wrap enabled (when you can see more than one copy of the map at once):
image

xlenstra:
> Possible solution: Stop drawing the sides (left & right) of the viewport rectangle once different viewports start to overlap. I don't know how to implement this yet though, any ideas would be appreciated.

I don't know how the code for drawing the minimap works, but I assume that somewhere the width of the minimap viewport is calculated, which can then be compared with the width of the entire minimap. If it's larger, you know the edges are going to overlap, and then can stop drawing the minimap viewport entirely?

SomeTroglodyte:
Hey, first of all: :+1:  good idea (had it a while ago, too, but was too lazy to find out how), and :+1: for actually pulling it off, no matter if there's room for improvement.
à propos: The whole new block lines 123-131 could be one loop too - just declare the [1] indicator is the "main" one. To avoid cursing mutability of Gdx geometry too much, `scrollPositionIndicator[i].setViewport(Rectangle(miniViewport).apply { x += (i-1) * allTiles.width })` will probably work.

yairm210:
Nice!

Thyrum:
> I don't know how the code for drawing the minimap works, but I assume that somewhere the width of the minimap viewport is calculated, which can then be compared with the width of the entire minimap. If it's larger, you know the edges are going to overlap, and then can stop drawing the minimap viewport entirely?

What I meant is I don't know how to only draw the top & bottom of the viewport (leaving out the sides). The rectangle is currently just an image which is scaled to the proper size. Personally I would want to impelement it as an actual rectangle built from lines (instead of a PNG), making it easier to leave out some lines. I would have to look into how drawing works using Kotlin though.

I don't think we should stop drawing the viewport whenever it overlaps. This will probably just create confusion since people will think the minimap glitches when the viewport overlay suddenly disappears.

yairm210:
This is already an improvement on the current state, I'm merging this and if you change to a "draw lines" approach that can be a separate PR :)

Issue #5116, opened 2021-09-06
Bug: Notifications does not work properly


touhidurrr:
**Platform**
Android 10 and Android 11, arm64

**Version**
3.15.x to 3.16.x (too many to list)

**Describe the bug**
Notifications does not work properly. The Game notifies players to join the game regardless of it's their turn or not.

**To Reproduce**
Just exit a Multiplayer Game and wait for Notifications. You will get it always regardless if it is your turn or not

**Comments**
I reported this on Discord before. But maybe developers forgets but if there is no Issue about it. There is always the desire to close issues yourself when there is an Issue maybe. Chats are easy to forget.

touhidurrr:
For those who might ask I was getting this before and after I updated my Android version. And its not encountered on a Virtual Environment. Nowadays all new Android Devices are arm64 based. I encountered it on Realme 3 Pro.

yairm210:
I checked the code now, it's actually the part I know least in Unciv because it's all Android
The logic *seems* sound. It should only notify when either
A. You pause a game, it's multiplayer, and it's your turn
B. Of the multiplayer games saved in the folder, one of them has your ID as the current ID.

So my first question is - what do you mean by "exit a multiplayer game"? What exactly are you doing?
Second question, to @GGGuenni - how can I debug this code?

GGGuenni:
Uh that's tough
Debugging that is really hard
As I've just finished my bachelor degree today, I will be back to work on Unciv tomorrow, though, so I can take a look at it 😁

Definitely need more info though
@touhidurrr do you get "normal" notifications? Or do you get multiple notifications at once? I heard of some occurrences of that but could not reproduce it. And you made sure there is really no game in the multiplayer list where it's your turn? Even when the game is maybe already over? I should definitely add the name of the first game found to the notification. That would probably help with debugging.

yairm210:
Congrats! 🥳

touhidurrr:
> I checked the code now, it's actually the part I know least in Unciv because it's all Android The logic _seems_ sound. It should only notify when either A. You pause a game, it's multiplayer, and it's your turn B. Of the multiplayer games saved in the folder, one of them has your ID as the current ID.

Maybe this is the problem. Normally you expect to get the notification of the game you are playing. Not all of them. Maybe we need a 'Get Notifications' making / joining a multiplayer game?

GGGuenni:
> Congrats! 🥳

Thanks 😁

[...]

PR #5350, opened 2021-09-29
Added game name to turn notification

GGGuenni:
To be able to better understand problems that are related to the turn notifier (like #5116) and give the user a better overview of why the notification pops, I've added the name of the game that triggers the notification to it.

I've also changed one of the turn checker start conditions from "the current game must be a multiplayer game" to "any multiplayer game exists in the multiplayer list". 
I'm not entirely sure why the condition "a current game must be initialized" is implemented though. Shouldn't the turn checker start every time the game is closed and not only when a game is initialized? Starting Unciv without starting a game currently does not enable the turn checker which might be irritating for a user who is waiting to be notified but isn't because the checker didn't start, only because no game was loaded.

### DIFF
--- a/android/res/values-de/strings.xml
+++ b/android/res/values-de/strings.xml
@@ -2,7 +2,7 @@
 <resources>
     <string name="app_name">UnCiv</string>
     <string name="Notify_YourTurn_Short">Unciv - Du bist am Zug!</string>
-    <string name="Notify_YourTurn_Long">Deine Freunde warten auf deinen Zug.</string>
+    <string name="Notify_YourTurn_Long">Deine Freunde warten auf deinen Zug in [gameName].</string>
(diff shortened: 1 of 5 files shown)


SomeTroglodyte:
It's good to see someone taking up improving the multiplayer aspect!
Made me think: _abandoned_ MP games. There's currently no provision to deal with those, right?

GGGuenni:
Yea that's true that's not really dealt with.
I've once implemented the resign feature but that still takes manual action of the person abandoning the game and does not really help with games that are already abandoned

Do you have any idea on how to improve on it?

SomeTroglodyte:
Idea - not really. Ask dropbox api for last modified date - somewhere the check and wait for reply doesn't really hurt - some background thread? Opening multiplayer screen? Nothing stands out as obviously workable approach.

PR #5354, opened 2021-09-29
Barbarian spawning and camp placements

SimonCeder:
Reworks spawning of barbarians and encampments to be more like the original.
Main effects include encampments spawning farther apart but respawning quicker and generally spawning more barbarians.
There is also a game option, "Raging Barbarians", which increases spawn rates further.

Most of the changes are in a new file which was split off from `GameInfo.kt`.

Also implements "No improvements except Roads may be built on this tile" unique for Oasis and Fallout. Sorry, this means no more great person improvements on Oases.

Differences from the original:
- The implementation in the original adds two extra fields to every tile in the game, holding time until next spawn as well as number of barbarians spawned from that tile. I decided to spin this data into a class instead, thought that was more elegant.
- In the original there is also a counter that ensures no new camp can go on the *exact tile* of a destroyed encampment for 15 turns. This would have meant adding another field to TileInfo and I felt this was unlikely anyway.
- The original weights spawned units by AI types which means more Brutes and less Scouts. Perhaps someday.


yairm210:
"here" is not a good word to use since it's ambiguous.

Please change to "No improvements may be built on this tile"

SimonCeder:
It's now "No improvements except Roads may be built on this tile" as the original does permit Roads on oases.

xlenstra:
What about "No improvements except [improvementFilter] may be built on this tile"?

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    fun setTransients(gameInfo: GameInfo) {
+        this.gameInfo = gameInfo
+        this.tileMap = gameInfo.tileMap
+
+        // Add any preexisting camps as Encampment objects
+        for (tile in tileMap.values) {
+            if (tile.improvement == Constants.barbarianEncampment
+                && camps.none { it.position == tile.position }) {
```
SomeTroglodyte, line 35: faster if camps where a HashMap. Vector2 obeys the equality contract and is safe to use as Hash key.

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,265 @@
...
+                val newCamp = Encampment()
+                newCamp.position = tile.position
+                newCamp.gameInfo = gameInfo
```
SomeTroglodyte, line 38: Done again 6 lines down

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,265 @@
...
+
+        // Update all existing camps
+        for (camp in camps) {
+            camp.update()
+        }
+    }
+
+    /** Called when an encampment was attacked, will speed up time to next spawn */
+    fun campAttacked(position: Vector2) {
+        camps.firstOrNull { it.position == position }?.wasAttacked()
```
SomeTroglodyte, line 66: Would also profit from HashSet<Vector2, Encampment>...

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,265 @@
...
+        val fogTiles = tileMap.values.filter { it.isLand && it !in allViewableTiles }
+
+        val fogTilesPerCamp = (tileMap.values.size.toFloat().pow(0.4f)).toInt() // Approximately
+
+        // Check if we have more room
+        var campsToAdd = (fogTiles.size / fogTilesPerCamp) - camps.size
+
+        // First turn of the game add 1/3 of all possible camps
+        if (gameInfo.turns == 1) {
+            campsToAdd /= 3
```
SomeTroglodyte, line 86: shouldn't this round up so an original campsToAdd == 2 wouldn't end up doing nothing on turn 1?

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,265 @@
...
+        camps.firstOrNull { it.position == position }?.wasAttacked()
+    }
+
+    fun placeBarbarianEncampment() {
+        // Before we do the expensive stuff, do a roll to see if we will place a camp at all
+        if (gameInfo.turns > 1 && Random().nextBoolean())
+            return
+
+        // Barbarians will only spawn in places that no one can see
+        val allViewableTiles = gameInfo.civilizations.filterNot { it.isBarbarian() || it.isSpectator() }
```
SomeTroglodyte, line 75: civilizations.asSequence so only the end set takes tons of heap space?

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,265 @@
...
+        // First turn of the game add 1/3 of all possible camps
+        if (gameInfo.turns == 1) {
+            campsToAdd /= 3
+        } else if (campsToAdd > 0)
+            campsToAdd = 1
+
+        if (campsToAdd <= 0) return
+
+        // Camps can't spawn within 7 tiles of each other or within 4 tiles of major civ capitals
+        val tooCloseToCapitals = gameInfo.civilizations.filterNot { it.isBarbarian() || it.isSpectator() || it.cities.isEmpty() || it.isCityState() }
```
SomeTroglodyte, line 93: Depending on the sizes of the tile collections it may be more efficient to not keep two overlapping sets of tiles to check for non-inclusion - keeping the capitals as iterable may suffice, then use .none { ... aerialDistanceTo ... }. Otherwise, when staying with sets, sets of Vector2 are more lightweight than those of TileInfo - and HashSets or -Maps of TileInfo are everywhere but theoretically unreliable as TileInfo doesn't do correct implementations of hashCode and equals - and building larger Sets has a smaller footprint if sequenced. May be academic. Thoughts triggered, goal of comment already reached.

SimonCeder: .none { ... aerialDistanceTo ... } was actually my first attempt, but this is significantly faster, likewise with mapping to Vector2. Of course this is on a PC and memory may be more of an issue on old phones, I'm not knowledgeable enough to tell. I'll let this stand unless you think it will actually be an issue for some users.

SomeTroglodyte: No no I just like to push, but in the end I trust you to use your brain.
Measuring memory is - hard to do on the JVM. Easier to (guess the memory footprint of a single instance) * (guess cardinality) * (JVM collections overhead). With Vector2 ~8 Byte and TileInfo wild guess 100 Bytes - a large map can have 5000 - still less than 1 MB. But if you allocate and GC-abandon that regularly, markedly increased GC pauses may crop up much later...

SomeTroglodyte: A way to test would be - console log a marker, deploy & run on a stone age AVD with dalvik and watch the GC messages on the logcat close to that marker, before/after. :grin: :skunk:

SimonCeder:
> What about "No improvements except [improvementFilter] may be built on this tile"?

I did "Only [improvementFilter] improvements may be built on this tile" which perhaps sounds a bit stilted but fits the filter.
"Only [All Road] improvements may be built on this tile"

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,269 @@
...
+    fun clone(): Encampment {
+        val toReturn = Encampment()
+        toReturn.position = position
+        toReturn.countdown = countdown
+        toReturn.spawnedUnits = spawnedUnits
+        return toReturn
+    }
+
+    fun update() {
+        if (countdown <= 0) {
```
yairm210, line 176: Switch around for if ... Else if ... For less indentation and easier readability

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,269 @@
...
+            .filter { it.isMilitary() }
+            .filter { it.isBuildable(barbarianCiv) }
+
+        var unit = if (naval)
+            unitList.filter { it.isWaterUnit() }.randomOrNull()
+        else
+            unitList.filter { it.isLandUnit() }.randomOrNull()
+
+        if (unit == null) // Didn't find a unit for preferred domain
+            unit = unitList.random() // Pick another, or crash the game vOv
```
yairm210, line 244: How about adding if unitlist isempty return on line 237 to defend against mad modders

core/src/com/unciv/logic/BarbarianManager.kt
```diff
@@ -0,0 +1,269 @@
...
+        countdown += gameInfo.ruleSet.difficulties[gameInfo.gameParameters.difficulty]!!.barbarianSpawnDelay
+        // Quicker if this camp has already spawned units
+        countdown -= min(3, spawnedUnits)
+
+        countdown *= when (gameInfo.gameParameters.gameSpeed) {
+            GameSpeed.Quick -> 67
+            GameSpeed.Standard -> 100
+            GameSpeed.Epic -> 150
+            GameSpeed.Marathon -> 400 // sic!
+        }
```
yairm210, line 266: I would add the /100 here

core/src/com/unciv/logic/BarbarianManager.kt
SomeTroglodyte: randomOrNull should exist as well

core/src/com/unciv/logic/BarbarianManager.kt
SimonCeder: Doesn't work, it's an Int so it would just round down to 0 for Quick

core/src/com/unciv/logic/BarbarianManager.kt
SimonCeder: I would expect the Set of TileInfos to just contain references to the actual TileInfos held by TileMap. Experimentally doing something silly to the tiles in the final viableTiles list has an effect on the game map.

SomeTroglodyte: :facepalm: oops silly me. Of course those maps don't hold clones. So, just keep the equality issue in mind as decision argument. Why Maps of TileInfo work at all may be luck, may be some deterministic secret (probably no map ever sees both a tile and a clone of it), but in theory - I had that kind of failure in KeyPressDispatcher and took the easy route to make KeyCharAndCode properly obey equality (data).

PR #5355, opened 2021-09-29
Fix Fountain of Youth

SimonCeder:
Fixes a couple problems with the Fountain of Youth:
- Now gives happiness without needing to be worked
- The base Plains terrain now does not add its stats to the tile, ie implemented the `overrideStats` field for Natural Wonders
- The promotion now actually affects all sources of healing including pillaging and Jaguar-like effects.


SomeTroglodyte (approved):
Thanks for the endorsement.
A FoY 4 tiles from city center but owned will probably still not actually provide the yield, but that's minor.

yairm210:
I assume you've seen https://github.com/yairm210/Unciv/pull/5333

yairm210:
That had the genius TileInfo.hasUnique(), which I think should replace the two usages here that check both terrains and improvements

SimonCeder:
> I assume you've seen #5333

I actually didn't, or I did and somehow forgot about except subcounciously. Now I understand SomeTroglodyte's comment.

SomeTroglodyte:
Arriving independently at almost identical code somehow proves it's good? Or - goes to show there's still problems. Looking for differences, I arrive at the conclusion with your code it would be possible to reap the benefits of one FoY up to 3 times. That's why I closed my PR - as I understand it, it needs treatment civ-wide, not in city code, and I was too lazy to test that out.

SimonCeder:
> That had the genius TileInfo.hasUnique(), which I think should replace the two usages here that check both terrains and improvements

Added.


> Arriving independently at almost identical code somehow proves it's good? Or - goes to show there's still problems. Looking for differences, I arrive at the conclusion with your code it would be possible to reap the benefits of one FoY up to 3 times. That's why I closed my PR - as I understand it, it needs treatment civ-wide, not in city code, and I was too lazy to test that out.
> 
> Correction: 4 times. 5 once we implement city distance 3 is allowed if on different continents.

I had to extra check, but no it works just fine:
Screenshot_2021-09-30_22-51-02

The magic is in `it.owningCity == cityInfo` in `getStatsFromTiles`. Since each tile only can have one owningCity the wonder only gets counted once.

SomeTroglodyte:
> `it.owningCity == cityInfo`

Exactly - I didn't see it applying to the TerrainTarget unique check, only to the ImprovementTarget one. Looking now it's clear enough. :+1:

PR #5361, opened 2021-09-30
Simplified combat uniques further

xlenstra:
This PR continues the simplification of combat uniques started in #5345, see also #4777.

Added conditionals
- \<when attacking\>
- \<when defending\>

Deprecated uniques:
- +[amount]% Strength when attacking
- +[amount]% Strength when defending
- +[amount]% Strength when defending vs [unitFilter] units


core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -597,21 +597,29 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
 
             when {
                 unique.isOfType(UniqueType.Strength) && unique.params[0].toInt() > 0 -> {
-                    if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsCity) })
-                        power += (power * unique.params[0].toInt()) / 200
-                    else if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) })
-                        power += (power * unique.params[0].toInt()) / 400
+                    when {
+                        unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) } ->
```
yairm210, line 601: Sounds like only the vs city is the odd one out
So if else

SomeTroglodyte: I'd foremost unify the uniques and promotions loop. Why should 'May Paradrop' have a value to the AI when it's on a unique but not then it's on a promotion? The other differences in these loops are academic, too. Pity the duplication of `power += (power * unique.params[0].toInt()) /` can't be trivially simplified.

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -215,14 +227,16 @@ object BattleDamage {
...
-                modifiers.add("Defender Bonus", unique.params[0].toInt())
-            }
+            // Deprecated since 3.17.4
+                for (unique in defender.unit.getMatchingUniques("[]% Strength when defending vs [] units")) {
+                    if (attacker.matchesCategory(unique.params[1]))
+                        modifiers.add("defence vs [${unique.params[1]}] ", unique.params[0].toInt())
+                }
+    
+                for (unique in defender.unit.getMatchingUniques("+[]% Strength when defending")) {
+                    modifiers.add("Defender Bonus", unique.params[0].toInt())
```
yairm210, line 237: Please add uniquetypes for deprecated uniques, so modders get warning signs for them

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -597,21 +597,28 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
...
                 //
-                unique.placeholderText == "+[]% Strength when attacking" // Attack - half the bonus
-                    -> power += (power * unique.params[0].toInt()) / 200
-                unique.placeholderText == "+[]% Strength when defending" // Defense - half the bonus
-                    -> power += (power * unique.params[0].toInt()) / 200
+                // Deprecated since 3.17.4
+                    unique.placeholderText == "+[]% Strength when attacking" // Attack - half the bonus
+                        -> power += (power * unique.params[0].toInt()) / 200
+                    unique.placeholderText == "+[]% Strength when defending" // Defense - half the bonus
+                        -> power += (power * unique.params[0].toInt()) / 200
```
yairm210, line 620: Here too - uniquetypes

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -626,21 +633,28 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
...
+                        unique.placeholderText == "+[]% Strength vs []" && unique.params[1] == "City" // City Attack - half the bonus
+                            -> power += (power * unique.params[0].toInt()) / 200
+                        unique.placeholderText == "+[]% Strength vs []" && unique.params[1] != "City" // Bonus vs something else - a quarter of the bonus
+                            -> power += (power * unique.params[0].toInt()) / 400
+                    //
+                    // Deprecated since 3.17.4
+                        unique.placeholderText == "+[]% Strength when attacking" // Attack - half the bonus
+                            -> power += (power * unique.params[0].toInt()) / 200
+                        unique.placeholderText == "+[]% Strength when defending" // Defense - half the bonus
+                            -> power += (power * unique.params[0].toInt()) / 200
```
yairm210, line 656: etc etc

core/src/com/unciv/logic/battle/BattleDamage.kt
xlenstra: Good point, I'll fix this later today

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -597,21 +597,28 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
 
             when {
                 unique.isOfType(UniqueType.Strength) && unique.params[0].toInt() > 0 -> {
-                    if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsCity) })
-                        power += (power * unique.params[0].toInt()) / 200
-                    else if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) })
-                        power += (power * unique.params[0].toInt()) / 400
+                    if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) } ) { // Bonus vs some units - a quarter of the bonus
```
yairm210, line 600: Should convert to int, divide by 4, and only them convert to percent

This will give you, for a 4% bonus that should become a 1%:
(1.04)/4=0.26 instead of 1.01

PR #2323, opened 2020-04-02
New music player

rh-github-2015:
Brewing for a while, @Iver88 has co-tested, and now re-based on 3.6.15.
Resolves #2232 partly. Some planning for expanded music downloader are done, later, separately.

To properly test it might be easiest to supply a copy of my music folder, but I'm hesitant even though everything is from civ-related open sources... I'd have to compile a precise credits file.

PR #5364, opened 2021-09-30
MusicController tweaks and hooks for mood - War and Peace

SomeTroglodyte:
With this, and music tracks from:
- [These instructions](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Demos/Fetch%20Civ%20V%20Soundtrack%20from%20Y-vomit-tube.md)
- [This scraper script](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Demos/GeoffKnorrMusicScraper.py) ([mini howto](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Demos/GeoffKnorrMusicScraper.md))
- More "Ambient" tracks - Work in progress. My "Additional Music *" mods ~~need renames~~ now fit, but thatched villagers could do with a rename so they all are more cleanly separated from the diplomatic tracks. Since everything is fuzzy logic, there'll be no failures (should be, knock on wood) - worst case music plays that doesn't really fit the situation.

.. you get most of the musical experience this 'project' is aiming for.

Disclaimer: This is a re-do of #2323 with _implied_ permission by the original author.

Borderline decision: Fuzzy based on strings and file names - could have gone for Enums for Context / Mood instead. Pro: simple, no additional json needed.

Open ideas: Hooks on Wonder built, tech researched, selecting a perm av mod, city state specific, quests, whatever. ~~A generic defeat dirge (hook is in this PR, I just haven't fetched an appropriate file - anything named "whatever - Defeat" will be used if found)~~ Some appropriate marches are now included in my Various mod. ...Wiki documentation!

Draft: I believe this is fine, but still want to test some more. Opportunity for input.


SomeTroglodyte:
Some important bugs found - foremost the TimSort one. The kludgy retries when OpenAL throws up buffer allocation are not fully tested, I might go for leaving a 'Silence' state instead, delaying a retry and having it choose a different track... But that could be done later as well.

yairm210:
The suffixes you use are Theme, War, Peace, and only them, and they see many uses.
Const strings, perhaps?

SomeTroglodyte:
True. I may have expected each to be needed less often...

SomeTroglodyte:
Okay, now the MusicMood dimension is almost an enum... And even the single use ones make for easier where-used searches and later renaming if need be.

logicminimal:
btw if you run unciv jar file u got from android studio desktop build and have groove enabled with some music in it (or if u have music in musics directory probably) these music play which is fine i guess but they also play when the game is at the background as well

SomeTroglodyte:
Your third party malware is none of our concern. But thanks for the idea - we might wish to put a .nomedia file in music folders for ~~morons~~ users that allow the default Android media indexer to roam free.

yairm210:
> Your third party malware is none of our concern. But thanks for the idea - we might wish to put a .nomedia file in music folders for ~morons~ users that allow the default Android media indexer to roam free.

Big oof

SomeTroglodyte:
The first hit when I tried to look up what "groove" is, was - silicone wedding rings.

logicminimal:
> Your third party malware is none of our concern. But thanks for the idea - we might wish to put a .nomedia file in music folders for ~~morons~~ users that allow the default Android media indexer to roam free.

... groove is an win10 inbuilt app

and the problem as i think arises from musics folder rather than groove

SomeTroglodyte:
After a few hours I couldn't repro that AL exception, so let's go with the Silence-retry approach.

Now I remember - Groove was a Windoze system service you got saddled with from certain SharePoint or Office installs, always went to 100% CPU sooner or later bogging down your servers... Malware, as I said. As I hear it went up in OneDrive Enterprise, even worse malware.

logicminimal:
>  exception, so let's go with the Silence-retry approach.\n\nNow I remember - Groove was a Windoze system service you got saddled with from certain SharePoint or Office installs, always went to 100% CPU sooner or later bogging down your servers... Malware, as I said. As I hear it went up in OneDrive Enterprise,

what am i meant to understand from that

AdityaMH:
Any wiki about this?

PR #5367, opened 2021-10-01
The Celtic People Reboot

SomeTroglodyte:
See #4658. Now uses a Conditional, and most code changes are linting or evolution of the system that's due anyway.

Maybe remove unused `ConditionalNeighborTiles` - typed when I assumed I'd find a nicer alternative for "and".


logicminimal:
did unimproved forest work?

SomeTroglodyte:
Todo:
- Edinburgh and Dublin need to be removed from City states in a way that doesn't immediately break saved games -> CS list needs G&K workover anyway - separate PR.
- Pictish Warrior icon redo - done.
- Find the other unique with a kind of terrain filter "and" and compare: It's Petra, and the scope is different but the approach comparable -> leave proposal as is.

AdityaMH:
Lastest commit make me laugh

Mugi1234:
```
```bbbmmmmmmll
Mmmmmmm. Nnnnnnnnnnmm

, ..   
```
```

SomeTroglodyte:
Reference, among many:
https://en.wiktionary.org/wiki/Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch

The original Civ5 made an achievement out of that, so it's worth mentioning. Also, the village itself is a tourism magnet - though it is not at all located in Boudicca's historical domain.

logicminimal:
> 
> 
> Reference, among many: https://en.wiktionary.org/wiki/Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch
> 
> The original Civ5 made an achievement out of that, so it's worth mentioning. Also, the village itself is a tourism magnet - though it is not at all located in Boudicca's historical domain.

we should also add achievements and/or easter eggs sometime.

AdityaMH:
> Reference, among many: https://en.wiktionary.org/wiki/Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch
> 
> The original Civ5 made an achievement out of that, so it's worth mentioning. Also, the village itself is a tourism magnet - though it is not at all located in Boudicca's historical domain.

But still funny and just imagine how to read this f****ng names.

PR #5368, opened 2021-10-01
RFC: Remove TranslationFileWriter knowledge about param values..

SomeTroglodyte:
.. rely only on UniqueParameterType

As you can see from the comments another 8 lines or so could go easily, the TFW would still recognize more than before - but some templates would change param names invalidating existing xlations. "building" / "buildingName" being the prime example - somehow a bit pointless.

What do you think? Retro-match a few UniqueParameterType secondary names to keep more xlt's?

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
@@ -100,6 +100,13 @@ enum class UniqueParameterType(val parameterName:String) {
             return UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
         }
     },
+    Technology("tech") {
+        override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
+            UniqueType.UniqueComplianceErrorSeverity? = when (parameterText) {
+                in ruleset.technologies -> null
+                else -> UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
+            }
+    },
     Promotion("promotion") {
         override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
             UniqueType.UniqueComplianceErrorSeverity? = when (parameterText) {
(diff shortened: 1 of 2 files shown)


yairm210:
Pay of the translation magic is if it recognizes that a param name has changed, it also changes the cake in the translation ;)
But this requires testing
That is, running translation then tests

SomeTroglodyte:
> recognizes that a param name has changed

Oh, yes, so that is much easier even than it seemed, and it seemed somehow obvious. Thus the 5min-RFC, wanted to see how it could work... Even if maybe that code will become obsolete itself in time.

As for testing - of course, at least four of my branches (only one local) need that, but pause first.

SomeTroglodyte:
Let's see... Take 'gerfinishen' as example... Remove all commented lines so only the two UniqueParameterType based sources remain...

`[param] Units adjacent to this city heal [amount] HP per turn when healing = Während Heilen, erhalten [param] Einheiten, die benachbart zu dieser Stadt sind, [amount] LP pro Runde`
...got converted to...
`[baseUnitFilter] Units adjacent to this city heal [amount] HP per turn when healing = Während Heilen, erhalten [baseUnitFilter] Einheiten, die benachbart zu dieser Stadt sind, [amount] LP pro Runde`
:+1:

`[stats] from [tileFilter] tiles without [tileFilter2] [cityFilter] = [stats] von [tileFilter] Feldern ohne [tileFilter2] [cityFilter]`
...got converted to...
<code>[stats] from [terrainFilter] tiles without [terrainFilter2] [cityFilter] = [stats] von [**tileFilter**] Feldern ohne [**tileFilter2**] [cityFilter]</code>
:-1:

All of the diff falls into the same categories of change, so overall very good, a lot of "param" now specific, a very few less specific but still OK for translator context.

Second example (rare) shows that the conversion code has a loophole - but I'd need incentive to debug it. Is this simplification worth it anyway? It could be obsolete in a few months, if we get to the point where `unique.type != null` is always `true`...

yairm210:
The magic only works if only one parameter has been renamed, if there's more than one then it doesn't "know" what was renamed to what, so that would indeed break things :(
And yes, I hope to get the the point of obsoletion as well :)

SomeTroglodyte:
OK, after enough mulling, I'll close this, it's too early, and maybe improving that 'magic' might be nice to have first. But it shows how _nicely_ the new UniqueType system would improve that part.

PR #5370, opened 2021-10-01
Reduce Mod checker messages for a set of deprecated UniqueTypes

SomeTroglodyte:
image
instead of
image


logicminimal:
btw i think this kind of a mod window would look great

PR #5371, opened 2021-10-01
Update template.properties

Smashfanful:
Added new translation line


android/assets/jsons/translations/template.properties
```diff
@@ -1277,6 +1278,8 @@ Status ￬ =
 Must be next to [terrain] = 
 Must be on [terrain] = 
 +[amount]% vs [unitType] = 
++[amount]% Strenght vs [unitType] = 
++[amount]% Strenght vs [unitType] units =
```
xlenstra, line 1282: These uniques are deprecated, so not having translations for them isn't a problem imo

android/assets/jsons/translations/template.properties
```diff
@@ -1190,6 +1190,7 @@ Water =
 water units = 
 wounded units = 
 Wounded = 
+Double movement in =
```
xlenstra, line 1193: Do we actually use this string? Or do we only use "Double movement in [tileFilter]"? That should already be automatically added

yairm210:
All of these translations should come from the uniques and not from the translations.
There was a change in uniques related to all of these, hence it's not surprising that older deprecated translations were auto-removed

PR #5375, opened 2021-10-02
Updated screenshots for google play & itch.io to better reflect the current state of the game

xlenstra:
While browsing through the google play page of unciv, I noticed that pretty much all screenshots were more or less out of date.
I've therefore taken new screenshots and replaced the old ones with these, not sure whether this is enough to get them to google play, though.
All of these were taken in the current version of master and should thus be compatible with 3.17.4 whenever that comes out.
The games were either games I played myself or of games that were send in by users on discord to report a bug.
I have not taken the width and height of the image into account, so if they need to have a specific size or size ratio, that still needs to be changed.

<details> <summary> Screenshots are also included below: </summary>

City Screen:
CityScreen
Civilopedia:
Civilopedia
Diplomacy Screen:
DiplomacyScreen
Map Editor:
MapEditor
Nation Picker:
NationPicker
Policy Screen:
PolicyScreen
Tech Tree:
TechTree
Trade Screen:
TradeScreen
World Screen:
WorldScreen

</details>


yairm210:
Looks good!
Only thing that won't work is Google Play feature graphic, which has a max size of 1024*500

yairm210:
And with this commit you're officially up to 4th most committed participant :)

xlenstra:
> And with this commit you're officially up to 4th most committed participant :)

:D

yairm210:
Also, itch.io recommended size is 630x500, so if you can...

PR #5379, opened 2021-10-02
Fixed bugs with fallout

### LINKED ISSUE #4300: Three fallout issues
1. Fallout cannot be placed in map editor. I put Grassland, select Fallout, click, and nothing.

2. (Not only for Fallout, but it's the only Vanilla here) Some tile layers can create negative yields instead of only decreasing.
Screenshot_20210629-121758_UnCiv

3. Fallout doesn't decrease production (as seen on Screenshot)

### LINKED ISSUE #5353: Fallout Bug(?) on Map Editor
You cannot place Fallout on map editor. This issue has last for a long time.

xlenstra:
This PR fixes a few bugs to do with fallout.
Fixes #5353 and part 1 and 3 of #4300, and as the second part was already solved, this fixes #4300
Additionally, I got distracted while adding the new uniques, and accidentally reordered all the uniques in the `UniqueType` enum, and added comments to distinguish different sections, hopefully increasing readability.

Added uniques:
- "Nullifies all other stats anything on this tile provides"
- "Doesn't generate naturally" - Doesn't actually do anything, but I'm not yet ready for reworking a large part of the system for generating terrain features. It is for now only used to hide the "occursOn" quickfix noted by @SomeTroglodyte in #5353, but it has potential to also be used later.


SomeTroglodyte:
Somehow I was sure the negative yield thing was already fixed long ago as my aliens mod depends on that...

xlenstra:
> Somehow I was sure the negative yield thing was already fixed long ago as my aliens mod depends on that...

It could very well be, I hadn't actually checked that.

SomeTroglodyte:
Ah, I suspected but didn't check either it's missing some minus stats.. Last thing I actually _checked_ was by how much the original C5 devs shortened that achievement-giving "longest" city name... (Answer: To roughly one [_third_](https://en.wiktionary.org/wiki/Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch))

'Check' - 'accidentally reordered all the uniques in the UniqueType' - reminds me: Did we want to allow Uniques with identical placeholderText but still distinct, by target type? If so, the only place where right now there's a UniqueType.values().first - would be incorrect... Unique.init assigns a UniqueType by comparing placeholderText alone. I almost expected that to be one step further.

android/assets/jsons/Civ V - Vanilla/Terrains.json
```diff
@@ -152,11 +152,12 @@
 	{
 		"name": "Fallout",
 		"type": "TerrainFeature",
-		"food": -3,
-		"production": -3,
-		"gold": -3,
 		"movementCost": 2,
 		"unbuildable": true,
+		"uniques": ["Nullifies all other stats anything on this tile provides", "Doesn't generate naturally"],
```
yairm210, line 157: I would remove the "anything on"

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -57,36 +57,60 @@ enum class UniqueTarget(val inheritsFrom:UniqueTarget?=null) {
...
 enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
 
+    //////////////////////////////////////// GLOBAL UNIQUES ////////////////////////////////////////
+    
+    /////// Stat providing uniques
+    
     Stats("[stats]", UniqueTarget.Global),
     StatsPerCity("[stats] [cityFilter]", UniqueTarget.Global),
+    @Deprecated("As of 3.16.16", ReplaceWith("[stats] <if this city has at least [amount] specialists>"), DeprecationLevel.WARNING)
+    StatBonusForNumberOfSpecialists("[stats] if this city has at least [amount] specialists"),
```
yairm210, line 67: Missing unique target

yairm210: This does make a difference IIRC between "warning" and "error"

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -57,36 +57,60 @@ enum class UniqueTarget(val inheritsFrom:UniqueTarget?=null) {
...
+    /////// Other global uniques
+    
     FreeUnits("[amount] units cost no maintenance", UniqueTarget.Global),
     UnitMaintenanceDiscount("[amount]% maintenance costs for [mapUnitFilter] units", UniqueTarget.Global),
-    BonusStatsFromCityStates("[amount]% [stat] from City-States", UniqueTarget.Global),
-    RemoveAnnexUnhappiness("Remove extra unhappiness from annexed cities", UniqueTarget.Building),
+    @Deprecated("As of 3.16.16", ReplaceWith("[amount]% maintenance costs for [mapUnitFilter] units"), DeprecationLevel.WARNING)
+    DecreasedUnitMaintenanceCostsByFilter("-[amount]% [mapUnitFilter] unit maintenance costs"),
+    @Deprecated("As of 3.16.16", ReplaceWith("[amount]% maintenance costs for [mapUnitFilter] units"), DeprecationLevel.WARNING)
+    DecreasedUnitMaintenanceCostsGlobally("-[amount]% unit upkeep costs"),
```
yairm210, line 92: Same here, add target

PR #5382, opened 2021-10-03
Solves bugs with unit movement through fog of war

### LINKED ISSUE #4241: Units ignore movement penalties when the tiles are not visible
a
z
I think it is self explanatory, there should be no way for that Horseman to get there in a single turn.

xlenstra:
Fixes #4241, but I'm not fully happy with the way its implemented.
I don't see a better easier way, though, so I've made this a PR anyway.
Feel free to leave comments or close

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -813,8 +813,9 @@ class MapUnit {
         if (civInfo.isMajorCiv() 
             && tile.improvement != null
             && tile.getTileImprovement()!!.isAncientRuinsEquivalent()
-        )
+        ) {
             getAncientRuinBonus(tile)
+        }
         if (tile.improvement == Constants.barbarianEncampment && !civInfo.isBarbarian())
             clearEncampment(tile)
 
--- a/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
+++ b/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
@@ -441,6 +441,8 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
         // Cache this in case something goes wrong
 
         var lastReachedEnterableTile = unit.getTile()
+        var previousTile = unit.getTile() 
+        var passingMovementSpent = 0f // Movement points spent since last tile we could end our turn on
 
         unit.removeFromTile()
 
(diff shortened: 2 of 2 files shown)


yairm210:
I think this is actually a really good fix, but as with anything that touches the movement mechanics, to what extent was this tested?

xlenstra:
I've run through the logic in my head for about 10 minutes and have tested that this fixes the bug in the save file provided in #4241, and it seems to pass the automated testing. I haven't actually ran a game or done additional tests. So feel free to not include it in the version being released now, and I'll do some more testing later

yairm210:
K :)

SomeTroglodyte:
As a band-aid on a band-aid, and from cursory code inspection only, I'd say this looks good. The whole pathfinding still needs a complete rewrite, though...

yairm210:
> The whole pathfinding still needs a complete rewrite, though...

I have tried, bit at this point I don't know where I would start.
If I were to start a clean slate and rewrite this from scratch, I would probably use the same setup the same functions, and it would come out very similar, at least.
What this needs is probably just "someone who isn't me" 😅 

But perhaps you can help! What parts do you find unclear? Where would you start?

SomeTroglodyte:
I have no more than a mental coarse concept:
- use the single-tile movement cost function more or less as is, that should also already cover ZOC. Forget everything else.
- write a new class similar to BFS in API, doing A*, and keeping a record per tile - how did I get here, how many of the movement points did it cost, how many turns did it take, a flag for can I only pass through / can I "land here" / must I pause here on a multi-turn move / would this be an attack, and how much damage will a multi-turn move cost to get here. As such, it can't store alternate paths and must decide immediately between two alternates when it encounters them - whether this is possible with acceptable results for multi-turn moves incurring damage remains to be seen.
- write a unit test container setting up complex problems and running the new class against those, until happy
- Think about making it cacheable and able to "fast-forward", meaning able to discard and re-calculate only part of its decision tree to react to changes on the "game board" - may be possible, but may require different storage structure not based on tiles, but paths, same as above with the damaging paths. Caching with total discard on any tileMap change would already help, but not needing a full flush could mean an order of magnitude.
- Then, only then, think about the API structure and how it fits in with existing AI and UI.

I'm more amenable for the fun parts...

An aside: Testing is the big blocker, thus the idea to have a non-live class to begin with, in time there could be an advanced option defaulting false, and so on. And dry testing via unit test wrappers could use a better map editor allowing unit placement for the "setting up complex problems" part. So - I'd go for the map editor first, the ideas are there, but the scope is still worth a week or two.

ravignir:
Finally!

xlenstra:
I've tested it some more and don't believe there are any noticeable bugs in it.
I did, however, find another bug related to movement that is not related to this PR.
In the situation in the included picture, clicking the highlighted tile doesn't move the work boat at all.
This is the result of a combination of factors, were `getTileToMoveToThisTurn` doesn't know what to do when a tile and all adjacent tiles are impossible to move to. Normally this is fine, but if the tile and all adjacent tiles are obscured by fog of war, this is unwanted behaviour.
<details> <summary> Picture: </summary>


</details>

> someone who isn't me

I _might_ be interested in looking into this, but for now I'm more focusing on uniques and uni, maybe later when I have more time

PR #5385, opened 2021-10-03
New map for map editor shares settings storage

SomeTroglodyte:
Small thing: Map settings for a new map editor map share (load and save) persistence with the new game screen


SomeTroglodyte:
An aside: I just deleted a 30-line shelved diff that detected water-less or almost so maps in RiverGenerator and made them into "bathtub plug" ones. I had those on the screen - verry cute. Triggered by testing this PR - the RiverGenerator throws a message-less exception when it can't find water and actually finding the source of that "looks like..." message took a few minutes. Today I couldn't repro no matter how hard I tried.

So - that RiverGenerator line 58 is a potential problem, to be remembered.
Also - do we want a River density slider in adv map properties?

PR #5386, opened 2021-10-03
Sort City States, replace+deprecate Dublin and Edinburgh

SomeTroglodyte:
... and teach new game screen city state count slider how to calculate max properly, including changes to religion enabled.

Notes:
- Rounds off the Celts
- Sorting done so we can more easily compare with the [List on the wiki](https://civilization.fandom.com/wiki/Valletta_(Civ5)) - on each CS's entry bottom
- Just added 2 militaristic so the count stays the same, kept the colours of Dublin and Edinburgh for them
- Saved games with CS Dublin or Edinburgh can still be finished
- UniqueType-converting that hide from civilopedia unique is not trivial, so postponed
- We still have a lot of differences in our city states to that list... Lhasa for instance, or Bogota, or Cahokia... Do we want them to conform or is only the balance - counts of types - important?

### DIFF
--- a/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
@@ -2,10 +2,12 @@ package com.unciv.ui.newgamescreen
 
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.unciv.UncivGame
+import com.unciv.logic.civilization.CityStateType
 import com.unciv.models.metadata.BaseRuleset
 import com.unciv.models.metadata.GameSpeed
 import com.unciv.models.ruleset.RulesetCache
 import com.unciv.models.ruleset.VictoryType
+import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.ui.audio.MusicMood
 import com.unciv.ui.audio.MusicTrackChooserFlags
 import com.unciv.ui.utils.*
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -89,23 +92,32 @@ class GameOptionsTable(
...
+    private fun numberOfCityStates() = ruleset.nations.values.count {
+        it.isCityState() &&
+                (it.cityStateType != CityStateType.Religious || gameParameters.religionEnabled) &&
+                !it.hasUnique(UniqueType.CityStateDeprecated)
+    }
+
+    private fun Table.addReligionCheckbox(cityStateSlider: UncivSlider?) =
+        addCheckbox("Enable Religion", gameParameters.religionEnabled) {
+            gameParameters.religionEnabled = it
+            cityStateSlider?.run { setRange(0f, numberOfCityStates().toFloat()) }
```
yairm210, line 105: why .run not .apply?

yairm210:
The exact names aren't important, they're all interchangable

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
SomeTroglodyte: For me, expresses that I'm not using the return value, `Unit` in this case, while `apply` is more like a 'I do something to it - pardon, to a shadowed this - but am not done with it afterwards'... Not really important, taste mostly. Like coming from M$ languages I thought `with` to be a good thing at first, now it's off the palette for me, `run` instead. Or the use of `also` for the sole reason I don't _like_ to use that shadowed `this` explicitly (often meaning on the right side of the expression or as function parameter), but implicit is fine. That's the whole 'why' in a nutshell.

PR #5387, opened 2021-10-03
Ethiopia - Nation

SomeTroglodyte:
Todo:

- [x] Nation data
- [x] Nation unique code
- [ ] Unit
- [ ] Building
- [ ] Icons and Pixel Unit
- [ ] Atlas
- [x] LeaderIcon - my [scraper](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Demos/LeaderPortraitScraper.md) already does Selassie, just like Boudicca and Pacal
- [ ] Music Themes?
- [ ] Test

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -54,6 +54,11 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
                 state.defender != null && state.defender.matchesCategory("City")
             UniqueType.ConditionalVsUnits ->
                 state.defender != null && state.defender.matchesCategory(condition.params[0])
+            UniqueType.ConditionalVsLargerCiv -> {
+                val yourCities = state.civInfo?.cities?.size ?: 1
+                val theirCities = state.defender?.getCivInfo()?.cities?.size ?: 0
+                yourCities < theirCities
+            }
             UniqueType.ConditionalNeighborTiles ->
                 state.cityInfo != null &&
                 state.cityInfo.getCenterTile().neighbors.count {
--- a/core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
+++ b/core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
@@ -557,7 +557,7 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
             isRanged()
             && (uniqueObjects + getType().uniqueObjects)
                 .any { it.isOfType(UniqueType.Strength)
-                    && it.params[0].toInt() > 0    
+                    && it.params[0].toInt() > 0
                     && it.conditionals.any { conditional -> conditional.isOfType(UniqueType.ConditionalVsCity) } 
                 }
         )
@@ -593,14 +593,24 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
             power += 4000
 
         // Uniques
-        for (unique in uniqueObjects) {
+        val allUniques = uniqueObjects.asSequence() +
+            promotions.asSequence()
+                .mapNotNull { ruleset.unitPromotions[it] }
+                .flatMap { it.uniqueObjects }
 
+        for (unique in allUniques) {
             when {
                 unique.isOfType(UniqueType.Strength) && unique.params[0].toInt() > 0 -> {
-                    if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsCity) })
-                        power += (power * unique.params[0].toInt()) / 200
-                    else if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) })
-                        power += (power * unique.params[0].toInt()) / 400
+                    when {
+                        unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsCity) } ->
+                            power += (power * unique.params[0].toInt()) / 200
+                        unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) } ->
+                            power += (power * unique.params[0].toInt()) / 400
+                        unique.conditionals.any() -> // e.g. ConditionalVsLargerCiv
+                            power += (power * unique.params[0].toInt()) / 200
+                        else ->
+                            power += (power * unique.params[0].toInt()) / 100
+                    }
                 }
                 // Deprecated since 3.17.3
                     unique.placeholderText == "+[]% Strength vs []" && unique.params[1] == "City" // City Attack - half the bonus
@@ -621,34 +631,6 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
             }
         }
 
-        // Base promotions
-        for (promotionName in promotions) {
-            for (unique in ruleset.unitPromotions[promotionName]!!.uniqueObjects) {
-                when {
-                    unique.isOfType(UniqueType.Strength) && unique.params[0].toInt() > 0 -> {
-                        if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsCity) })
-                            power += (power * unique.params[0].toInt()) / 200
-                        else if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) })
-                            power += (power * unique.params[0].toInt()) / 400
-                    }
-                    // Deprecated since 3.17.3
-                    unique.placeholderText == "+[]% Strength vs []" && unique.params[1] == "City" // City Attack - half the bonus
-                        -> power += (power * unique.params[0].toInt()) / 200
-                    unique.placeholderText == "+[]% Strength vs []" && unique.params[1] != "City" // Bonus vs something else - a quarter of the bonus
-                        -> power += (power * unique.params[0].toInt()) / 400
-                    //
-                    unique.placeholderText == "+[]% Strength when attacking" // Attack - half the bonus
-                        -> power += (power * unique.params[0].toInt()) / 200
-                    unique.placeholderText == "+[]% Strength when defending" // Defense - half the bonus
-                        -> power += (power * unique.params[0].toInt()) / 200
-                    unique.placeholderText == "[] additional attacks per turn" // Extra attacks - 20% bonus per extra attack
-                        -> power += (power * unique.params[0].toInt()) / 5
-                    unique.placeholderText == "+[]% Strength in []" // Bonus in terrain or feature - half the bonus
-                        -> power += (power * unique.params[0].toInt()) / 200
-                }
-            }
-
-        }
         cachedForceEvaluation = power
     }
 }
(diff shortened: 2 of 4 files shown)


logicminimal:
damn with this pace of development we will finish this game entirely within less than a year

SomeTroglodyte:
@xlenstra - say I forgot about "Hidden when religion is disabled" for the Celts entirely, and the Stele is pretty obviously something needing that unique. What do you say - ignore and expecto patronum, uh, expect the switch to be gone soonish, or implement properly, meaning _all_ nations with direct properties depending on faith existing should be marked the same, and teach newgame to filter in the player picker too?

ravignir:
> damn with this pace of development we will finish this game entirely within less than a year

I called it first. The whole tourism thing is the most challenging thing to do atm. Espionage and ideologies aren't as complex.

SomeTroglodyte:

image
image
image
image

</details>

... That colour scheme doesn't go too well. Took it directly with the gimp picker from [this](https://static.wikia.nocookie.net/civilization/images/9/92/Ethiopian_(Civ5).png)... Just darken the green and lighten the red into pinkish?

ravignir:
Capture+_2021-10-03-18-22-00
These are original colors.

xlenstra:
I've mostly used colours from this Google sheet which appears roughly correct in my experience:
https://docs.google.com/spreadsheets/d/1ronxbl50Evz2OZ4khHlsfEvbDru_Vtd1z9UBQ0BQE_k/edit#gid=0

xlenstra:
> @xlenstra - say I forgot about "Hidden when religion is disabled" for the Celts entirely, and the Stele is pretty obviously something needing that unique. What do you say - ignore and expecto patronum, uh, expect the switch to be gone soonish, or implement properly, meaning _all_ nations with direct properties depending on faith existing should be marked the same, and teach newgame to filter in the player picker too?

Main thing that needs to be done before the switch is removed is finishing the AI, and a few small bug fixes, all other things are implemented. I already have a branch somewhere which does most of the choosing beliefs parts, which leaves only buying with faith and using missionaries & inquisiteurs for spreading/removing religions. If I have the time for that it shouldn't take too long, but having time is kinda the issue.

Long story short: I don't think it's that bad, worst case you have a nation that has no bonuses for a little while. It's at least not worth the effort filtering all these out at the player picker, maybe for the AI chosen nations too.

SomeTroglodyte:
What I found: That "Near Capital" bonus works, the linear decrease by 3/tile from that steem source is credible, but it is calculated for the unit's position before combat. And if it can move two tiles for the attack, then the starting tile may have the "wrong" distance. Not too sure how this could be solved.

SomeTroglodyte:
> Google sheet

Then I was close, the red can go a little brighter, the green darker and a touch colder. Pretty much as I said earlier. But that green from that sheet seems too dark:
image
no, I'll go with 67% that sheet + 33% my original values which looks like:
image

SomeTroglodyte:
> These are original colors

Thanks, looks pretty much identical to what the spreadsheet said, not quite to my taste. But you can vote for dark, I would listen...

> not worth the effort filtering all these out at the player picker

And I just went to the trouble with the city state slider max :rofl:  Jokes aside, I actually begin to think keeping that switch permanently, just changing its default soon, would even have a certain charm. Prophets are always the little guy from Life of Brian's stoning scene for me, and it might be refreshing to do a game without being reminded of ['romani ite domum'](https://upload.wikimedia.org/wikipedia/commons/thumb/d/d5/Romani_ite_domun_HER_Museum_6_July_2018.jpg/800px-Romani_ite_domun_HER_Museum_6_July_2018.jpg) from time to time?

... D*mn. No Mayas today, it's movie rerun time.

android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
```diff
@@ -538,5 +538,9 @@
 	{
 		"name": "Pictish Courage", // only for Pictish Warrior and subsequent upgrades
 		"uniques": ["No movement cost to pillage", "+[20]% Strength in [Foreign Land]"]
+	},
+	{
+		"name": "Home Sweet Home", // only for Mehal Sefari and subsequent upgrades
+		"uniques": ["[+30]% Strength decreasing with distance from the capital"]
```
xlenstra, line 544: Is it reasonably doable to make this something like "[+30]% Strength <decreasing over [3] tiles from the capital>" to allow extra moddability? That would also increase how easily it can be combined with other strength conditionals: `[+30]% Strength <when attacking> <decreasing [...]>` sounds better than `[+30]% Strength <decreasing [...]> <when attacking>` in my opinion, and this might change in other languages

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -593,15 +590,28 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
...
+                .flatMap { it.uniqueObjects }
 
+        for (unique in allUniques) {
             when {
                 unique.isOfType(UniqueType.Strength) && unique.params[0].toInt() > 0 -> {
-                    if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsCity) })
-                        power += (power * unique.params[0].toInt()) / 200
-                    else if (unique.conditionals.any { it.isOfType(UniqueType.ConditionalVsUnits) })
-                        power += (power * unique.params[0].toInt()) / 400
+                    power += (power * unique.params[0].toInt()) / when {
```
xlenstra, line 601: According to my math, this is actually equivalent to `power *= unique.params[0].toPercent / 2f`, which looks more readable in my opinion. I've also changed this in #5361, so these might conflict because of that. Otherwise, looks fine :+1:

xlenstra (commented):
Looks good!

android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
SomeTroglodyte: Conditionals at the moment are booleans testing something. I didn't want to change that system too fundamentally.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
SomeTroglodyte: I only touched that because Studio had some squigglies on that 'when'... I try to make one inspections check for each touched file after I'm done, not always treating each complaint, but at least considering each. We still have that bunch of 8-12 warnings actually appearing in github checks, but those I left at type 'unused parameter' only.

android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
xlenstra: Good point, maybe I'm too overzealous converting everything to conditionals and should actually think before propsing :)

SomeTroglodyte:
Conflix again? I was busy doing this:
image
... yes that's a working partial Maya calendar ...

android/assets/jsons/Civ V - Vanilla/UnitPromotions.json
SomeTroglodyte: Thinking of Uniques as potentially json-programmable by recombination is good, but we're still a long way from what e.g. the aurora toolset allowed in the neverwinter series...

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -57,6 +57,11 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
                 state.defender != null && state.defender.matchesCategory(condition.params[0])
             UniqueType.ConditionalAttacking -> state.combatAction == CombatAction.Attack
             UniqueType.ConditionalDefending -> state.combatAction == CombatAction.Defend
+            UniqueType.ConditionalVsLargerCiv -> {
+                val yourCities = state.civInfo?.cities?.size ?: 1
```
yairm210, line 61: I don't like the implementation. It would return true by default, which is the opposite of what we want.
More like "if state.civInfo null or state.defender null return false"

ravignir:
> I've mostly used colours from this Google sheet which appears roughly correct in my experience: https://docs.google.com/spreadsheets/d/1ronxbl50Evz2OZ4khHlsfEvbDru_Vtd1z9UBQ0BQE_k/edit#gid=0

Yup. These are extracted from the game. The problem is they are hsl values, so after converting them to rgb, they might be a little off.

ravignir:
> Conflix again? I was busy doing this: image ... yes that's a working partial Maya calendar ...

Pls no. Civ5 used numbers for mayan long count calendar so it at least was somewhat intuitive (not to mention it also displayed traditional time)

307b6bd(1)

SomeTroglodyte:
> it also displayed traditional time

But then it's not as easter-egg-y :sob:  -> I'll probably do a nice popup onClick.

But tell me: Of the two choices:
a) You research Theology and get your first pick. 394 years later you get the next.
b) You research Theology. When the maya calendar rolls over the next time to 0 k'atun, 0 tun you get to pick a GP.
From my reading I thought it would be a, but from that screenshot it sounds like b.

core/src/com/unciv/models/ruleset/unique/Unique.kt
SomeTroglodyte: No it returns false by default as 1 < 0 is false. Shouldn't ever trigger as I recall.
Problem is - 'defender' is a little misnomer here, as I expect and from my inspection/testing am almost sure the effect works both on offense and defense (and it should), and the defender is sometimes the aggressor really. But that part of the codebase is not too transparent to me, so if I understood that wrong, tell me.

ravignir:
> > it also displayed traditional time
> 
> But then it's not as easter-egg-y 😭 -> I'll probably do a nice popup onClick.
> 
> But tell me: Of the two choices: a) You research Theology and get your first pick. 394 years later you get the next. b) You research Theology. When the maya calendar rolls over the next time to 0 k'atun, 0 tun you get to pick a GP. From my reading I thought it would be a, but from that screenshot it sounds like b.

It is b. The count starts at 4000 BC (and once you research theology it is displayed in the long count). If you research theology late, you may miss one or more picks.

SomeTroglodyte:
> It is b

Good thing, too, that eases implementation, go by year only without need for an additional persistent field, or the flags system, or a need to do an inverse of the turn-to-year function. Still need a field for the each GP only once until all picked rule. Pity that part of the Uniques code is such a shambles at the moment (stuff in trigger activation that's not once-only, and once-only stuff hardcoded outside, and no trace of UniqueTypes yet...), maybe postpone & do fixes first.

core/src/com/unciv/models/ruleset/unique/Unique.kt
yairm210: I can't math apparently

SomeTroglodyte:
> merging this as is, if you are

Yes. My unresolved concern is way off, namely there's no public source for the nation-specific music tracks. And I still need to put my Celts-specific music _somewhere_...

Issue #3506, opened 2021-01-05
Couple of "Issues"


ravignir:
Here are couple of things I decided to elaborate on.

Helicopter Gunship unit should not be able to take cities. It has it's own unit type (Helicopter Unit) in civ5 which is basically an air unit that can end a turn on a tile. It does receive additional damage from Anti-Air units like Fighters, Anti-Aircraft guns, which is not present in unciv.
![image](https://user-images.githubusercontent.com/30041793/103679821-a1e02c80-4f85-11eb-9e06-230df2bfc763.png)

Computers tech provides +10% science and +10% production in all cities. Where did that came from? There is nothing like that in civ5.

In Unciv, Great Persons points are accumulated per Empire, while they are accumalated per city in civ5 - except for Great Prophets and Generals/Admirals. It means having 4 Scientist Specialists in three cities will not grant you a Great Scientist faster than one city with 4 Scientists Specialists. It only matters when you want to get multiple scientists one after another (count only resets in a city which created the GP). This forces the players to have multiple cities, each focused on a different thing. This also makes you know where the GP will be born.

Strategic Resource quantity is really scarce in Unciv. In civ5 Strategic resources spawn in quantities choosen from the list of possible values (each resource has its own predefinied max/min value) : https://forums.civfanatics.com/threads/resource-quantity.416107/#post-10329758 According to what I've experienced playing civ5, linked values are correct (Each resources appears either in max value or min value, depending on the resource amount setting, i.e Iron might either appear in small stack of 2 or major stack of 6 on a map with resource amount set to standard).
Oil on water tiles always appears in only one stack variation - 4 for sparse/default and 6 for abundant settings.

Speaking of Luxuries, you almost always get a couple copies (usually 3) of the same luxury near your starting location, so each player has "its own" luxury to trade + sometimes you get 1 copy of another luxury for happiness. The "Starting Luxuries" don't usually appear anywhere else on the map.

ravignir:
Another thing: In Civ5 when you found a city it automatically creates Road improvement on the same tile. This road is maintenance free (as long as the city is here) and stays there even when you raze the city (but then you will have to pay for it). Roads under cities are automatically upgraded to Railroads as soon as you get the tech that unlocks them.


![roads](https://user-images.githubusercontent.com/30041793/103752985-b87c9700-500a-11eb-8b97-e7d79a6bbec4.png)
[...]

PR #5393, opened 2021-10-04
Added `in [tileFilter] tiles` conditional to combat uniques

### LINKED ISSUE #4508: Promotion bugs
Ranged/Siege Promotion "Accuracy I" and  Melee/Mounted/Armor Promotion "Shock I" are different in original game, but now their effect are both "+[15]% Strength in [Open terrain]"
In original game, "Accuracy I" is:
if our Ranged/Siege unit attacks the enemy unit which is in Open terrain, our Ranged/Siege unit +[15]% Strength
Compare with "Accuracy I", "Shock I" is:
if our Melee/Mounted/Armor unit is in Open terrain, and then it attacks the enemy unit, our Melee/Mounted/Armor unit +[15]% Strength

xlenstra:
This PR simplifies combat uniques further using conditionals.
Fixes #4508, and adds the bonus vs cities of siege units only applying when attacking, as noted in #3506.

Added conditionals:
- "in [tileFilter] tiles"

Deprecated uniques:
- "+[amount]% defence in [tileFilter] tiles"
- "+[amount]% Strength in [tileFilter]"
- "[amount]% Strength for [mapUnitFilter] units in [tileFilter]"
- "+15% Combat Strength for all units when attacking Cities"

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Beliefs.json
+++ b/android/assets/jsons/Civ V - Vanilla/Beliefs.json
@@ -248,7 +248,7 @@
     {
         "name": "Defender of the Faith",
         "type": "Enhancer",
-        "uniques": ["[+20]% Strength for [All] units in [Friendly Land]"],
+        "uniques": ["[+20]% Strength <for [All] units> <in [Friendly Land] tiles>"],
         // ToDo: Should only be friendly territory of cities that follow this religion
     },
     {
@@ -264,7 +264,7 @@
     {
         "name": "Just War",
         "type": "Enhancer",
-        "uniques": ["[+20]% Strength for [Non-City] units in [Enemy Land]"],
+        "uniques": ["[+20]% Strength <for [All] units> <in [Enemy Land] tiles>"],
         // ToDo: Should only be enemy territory of cities that follow this religion
     },
     {
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -756,7 +756,7 @@
 		"culture": 3,
 		"isWonder": true,
 		"greatPersonPoints": {"Great Engineer": 2},
-		"uniques": ["+[15]% Strength for [Non-City] units fighting in [Friendly Land]",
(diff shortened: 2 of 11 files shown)


yairm210:
How about "for combat in [] tiles"?

Combat tile is always the defender's tile so I think this makes it clear that the relevant told is where the combat occurs and not where the unit is standing

SomeTroglodyte:
> Combat tile is always the defender's tile

#4508 said otherwise?

xlenstra:
> > Combat tile is always the defender's tile
> 
> #4508 said otherwise?

My testing in the original does seem to agree with Yairm's, so I don't know what the point was there. I've here fixed the bug that Ravignir mentioned in that issue, assuming it was the same as the original post

SomeTroglodyte:
'for' 80%, 'during' 5%, 'when' 15%.  ... but ...  'combat' 5%, 'bloodbath' 30%, 'massacres' 65% - so my judgement may not count. 'fisticuffs'! 

Why not just 'in [tileFilter] tiles'? Design the Conditional implementation such that it works no matter what the Conditional is attached to? Or - are we really relying on placeholderText being without duplicates over that entire enum? Or could we make it so a combination of placeholderText and scoping attributes must be unique (meant like for a sql index here), but the plchldrtxt alone can be defined repeatedly, e.g. once for Conditionals that apply to combat, once for Conditionals that apply to Stat boni?

yairm210:
The problem in the issue mentioned is clarity.

In Civ V, if an archer gets a combat bonus on plains, then if he's on plains and attacking - no bonus, if he attacks a plains tile - bonus.
If he's attacked in a plains tile, he gets the bonus.
The reason being, the tile the combat occurs in is the tile of the defender.
Same issue is also relevant for melee attacks.

xlenstra:
I'm personally leaning towards 'fighting in [] tiles', which may not sound great but I don't think there is a perfect option either

yairm210:
"When fighting in forest tiles", "when fighting in enemy land tiles"... Works for me, good one!

SomeTroglodyte:
Yes, I think that last change will prove worth it. :+1:

PR #5389, opened 2021-10-03
Added two more conditionals; updated unit movement & sight

xlenstra:
This PR continues adding more conditionals to combine combat uniques, and also refactors the movement and sight of units.

Added uniques:
- "[amount] Sight"
- "[amount]% Religious Spread Strength"

Added conditionals:
- \<during a Golden Age\>
- \<for [mapUnitFilter] units\>

Deprecated uniques:
- "[amount] Visibility Range"
- "[amount] Sight for all [unitFilter] units"
- "+[amount] Movement for all [unitFilter] units"
- "+1 Movement for all units during Golden Age"
- "[amount]% Spread Religion Strength for [unitFilter] units"
- "+10% Strength for all units during Golden Age"
- "[unitFilter] units deal +[amount]% damage"

PR #5397, opened 2021-10-04
UniqueType-i-fying UniqueTriggerActivation

SomeTroglodyte:
Begs the question - guideline to group and comment UniqueTypes... But somehow there's something amiss in that diff.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
@@ -15,6 +15,10 @@ enum class UniqueParameterType(val parameterName:String) {
             else null
         }
     },
+    NumberOrAll("amountOrAll") {
+        override fun getErrorSeverity(parameterText: String, ruleset: Ruleset) =
+            if (parameterText == "All") null else Number.getErrorSeverity(parameterText, ruleset)
+    },
     MapUnitFilter("mapUnitFilter"){
         private val knownValues = setOf("Wounded", "Barbarians", "City-State", "Embarked", "Non-City")
         override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
@@ -115,6 +119,13 @@ enum class UniqueParameterType(val parameterName:String) {
                 in ruleset.unitPromotions -> null
                 else -> UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
             }
+    },
+    Era("era") {
+        override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
+                UniqueType.UniqueComplianceErrorSeverity? = when (parameterText) {
+            in ruleset.eras -> null
+            else -> UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
+        }
     },
     /** Behaves like [Unknown], but states explicitly the parameter is OK and its contents are ignored */
     Comment("comment") {
--- a/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
@@ -15,9 +15,10 @@ enum class UniqueParameterType(val parameterName:String) {
             else null
         }
     },
-    NumberOrAll("amountOrAll") {
+    // todo potentially remove if OneTimeRevealSpecificMapTiles changes
+    KeywordAll("'all'") {
         override fun getErrorSeverity(parameterText: String, ruleset: Ruleset) =
-            if (parameterText == "All") null else Number.getErrorSeverity(parameterText, ruleset)
+            if (parameterText == "All") null else UniqueType.UniqueComplianceErrorSeverity.RulesetInvariant
     },
     MapUnitFilter("mapUnitFilter"){
         private val knownValues = setOf("Wounded", "Barbarians", "City-State", "Embarked", "Non-City")
@@ -177,6 +178,8 @@ enum class UniqueParameterType(val parameterName:String) {
             "in City-State cities",
             "in cities following this religion",
         )
+
+        fun safeValueOf(param: String) = values().firstOrNull { it.parameterName == param } ?: Unknown
     }
 }
 
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -240,7 +240,8 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
     OneTimeGainStatRange("Gain [amount]-[amount] [stat]", UniqueTarget.Ruins),
     OneTimeGainPantheon("Gain enough Faith for a Pantheon", UniqueTarget.Ruins),
     OneTimeGainProphet("Gain enough Faith for [amount]% of a Great Prophet", UniqueTarget.Ruins),
-    OneTimeRevealSpecificMapTiles("Reveal up to [amountOrAll] [tileFilter] within a [amount] tile radius", UniqueTarget.Ruins),
+    // todo: The "up to [All]" used in vanilla json is not nice to read. Split?
+    OneTimeRevealSpecificMapTiles("Reveal up to [amount/'all'] [tileFilter] within a [amount] tile radius", UniqueTarget.Ruins),
     OneTimeRevealCrudeMap("From a randomly chosen tile [amount] tiles away from the ruins, reveal tiles up to [amount] tiles away with [amount]% chance", UniqueTarget.Ruins),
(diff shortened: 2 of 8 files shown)


SomeTroglodyte:
Ah, #5389 already fixed the diff I was missing here, so that's OK then.

This is intended as groundwork, I did my best to not change behaviour in any way.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -220,6 +220,43 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
     // tile conditionals
     ConditionalNeighborTiles("with [amount] to [amount] neighboring [tileFilter] tiles", UniqueTarget.Conditional),
     ConditionalNeighborTilesAnd("with [amount] to [amount] neighboring [tileFilter] [tileFilter] tiles", UniqueTarget.Conditional),
+
+    ///////////////////////////////////////// TRIGGERED ONE-TIME /////////////////////////////////////////
+
+    OneTimeFreeUnit("Free [baseUnitFilter] appears", UniqueTarget.Global),  // used in Policies, Buildings
+    OneTimeAmountFreeUnits("[amount] free [baseUnitFilter] units appear", UniqueTarget.Global), // used in Buildings
+    OneTimeFreeUnitRuins("Free [] found in the ruins", UniqueTarget.Ruins), // Differs from "Free [] appears" in that it spawns near the ruins instead of in a city
```
yairm210, line 228: Missing

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -220,6 +220,43 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
...
+    OneTimeFreePolicy("Free Social Policy", UniqueTarget.Global), // used in Buildings
+    OneTimeAmountFreePolicies("[amount] Free Social Policies", UniqueTarget.Global),  // Not used in Vanilla
+    OneTimeEnterGoldenAge("Empire enters golden age", UniqueTarget.Global),  // used in Policies, Buildings
+    OneTimeFreeGreatPerson("Free Great Person", UniqueTarget.Global),  // used in Policies, Buildings
+    OneTimeGainPopulation("[amount] population [cityFilter]", UniqueTarget.Global),  // used in CN tower
+    OneTimeGainPopulationRandomCity("[] population in a random city", UniqueTarget.Ruins),
```
yairm210, line 234: Missing

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -220,6 +220,43 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
...
+    OneTimeRevealEntireMap("Reveals the entire map", UniqueTarget.Global),  // used in tech
+    OneTimeGainStat("Gain [amount] [stat]", UniqueTarget.Ruins),
+    OneTimeGainStatRange("Gain [amount]-[amount] [stat]", UniqueTarget.Ruins),
+    OneTimeGainPantheon("Gain enough Faith for a Pantheon", UniqueTarget.Ruins),
+    OneTimeGainProphet("Gain enough Faith for [amount]% of a Great Prophet", UniqueTarget.Ruins),
+    OneTimeRevealSpecificMapTiles("Reveal up to [amountOrAll] [tileFilter] within a [amount] tile radius", UniqueTarget.Ruins),
+    OneTimeRevealCrudeMap("From a randomly chosen tile [amount] tiles away from the ruins, reveal tiles up to [amount] tiles away with [amount]% chance", UniqueTarget.Ruins),
+    OneTimeTriggerVoting("Triggers voting for the Diplomatic Victory", UniqueTarget.Global),  // used in Building
+
+    OneTimeUnitHeal("Heal this unit by [] HP", UniqueTarget.Promotion),
```
yairm210, line 247: We should probably as a test for [] in unique type texts

SomeTroglodyte:
OK - another thought: If we rely on placeholderText or (placeholderText+target) to be unique - ahem - a unit test for that?

core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
```diff
@@ -15,6 +15,10 @@ enum class UniqueParameterType(val parameterName:String) {
             else null
         }
     },
+    NumberOrAll("amountOrAll") {
```
yairm210, line 18: Denied, this is a clear "amount". "Reveal up to All" throws human parse errors

SomeTroglodyte:
```json
    {
        "name": "reveal nearby Barbarian camps",
        "notification": "You find evidence of Barbarian activity. Nearby Barbarian camps are revealed!",
        "uniques": ["Reveal up to [All] [Barbarian encampment] within a [10] tile radius"]
    },
```
Not mine

SomeTroglodyte: > throws human parse errors

:grin:  and the human throws his breakfast in to the exception bucket (Mr Creosote)

yairm210: Eeww yukkk
"amount/param" then, and please add a todo to that

SomeTroglodyte: Wait - the "/" is parsed as alternate types? OK, forgot, that would obviously obviate the need for _that_ kludge type

yairm210: That is, this shouldn't be an enum value

SomeTroglodyte: No "/" isn't parsed - didn't someone code that? Thought I saw it earlier. Do we want? Right in UniqueType.init?

SomeTroglodyte: And that init could really use that safeValueOf generic I got in stash - time to unstash?
```kotlin
inline fun <reified T: Enum<T>> enumValueOfOrNull(name: String?): T? {
    if (name == null) return null
    return try {
        enumValueOf<T>(name)
    } catch (ex: IllegalArgumentException) {
        println("Could not parse\"$name\" as ${T::class.java.simpleName}.")
        null
    }
}
```
Each enum then can do a proxy easily as one-liner...

xlenstra: Hey, that looks like code I've written :p
On second thought, it's probably best to split this out into two uniques?

SomeTroglodyte: Our unit tests do not balk at parameter type violations as defined in UniqueParameterType.getErrorSeverity at all - uh, where were we? That still on the to-do list? Anyway, separating that all barb camp (plural is another readability problem) ruin reward - not this PR?

SomeTroglodyte: OK, `checkModLinks` does not check the _ruin rewards_ and I was only looking for a test failure manipulation those... Yet another PR.

SomeTroglodyte: Used a simpler safeValueOf in that latest push, also implements "/" for autoparsed placeholder params, so now those lists can have size>1

SomeTroglodyte:
Of course Unciv fails tests once `checkModLinks` covers _all_ unique-containing ruleset objects and once `@Test fun uniqueTypesHaveNoUnknownParameters()` is in there... On it.

SomeTroglodyte:
```
Error: Air Targeting I requires promotion Dogfighting I which does not exist!
Error: Sortie requires promotion Dogfighting II which does not exist!
Warning: Fertility Rites's unique "[+10]% growth [in this city]" cannot be put on this type of object!
Warning: Guruship's unique "[+2 Production] <if this city has at least [1] specialists>" cannot be put on this type of object!
Warning: Swords into Ploughshares's unique "[+15]% growth [in this city] <when not at war>" cannot be put on this type of object!
```
```
StatPercentBonus param[1] type is unknown
ConsumesResources param[1] type is unknown
ProvidesResources param[1] type is unknown
FreeExtraBeliefs param[1] type is unknown
FreeExtraBeliefs param[2] type is unknown
FreeExtraAnyBeliefs param[1] type is unknown
StrengthPlusVs param[1] type is unknown
StrengthMinVs param[1] type is unknown
NaturalWonderConvertNeighbors param[0] type is unknown
NaturalWonderConvertNeighborsExcept param[0] type is unknown
NaturalWonderConvertNeighborsExcept param[1] type is unknown
```
That's not too bad... The second test could pull out the actually used names though.

PR #5398, opened 2021-10-04
Sharpen unit tests for UniqueTypes (and mod checker too)

SomeTroglodyte:
... ***will*** fail github actions as this is tests only, no fixing yet -> next commit soon

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -59,15 +59,15 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
 
     //////////////////////////////////////// GLOBAL UNIQUES ////////////////////////////////////////
 
-    
+
     /////// Stat providing uniques
 
-    Stats("[stats]", UniqueTarget.Global),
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -59,15 +59,15 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
 
     //////////////////////////////////////// GLOBAL UNIQUES ////////////////////////////////////////
 
-    
+
     /////// Stat providing uniques
 
-    Stats("[stats]", UniqueTarget.Global),
+    Stats("[stats]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
```
xlenstra, line 65: In my opinion this should also be `UniqueTarget.Building`

SomeTroglodyte: UniqueTarget.Building is included in Global, but FollowerBelief was/is not - FollowerBelief(Global) would have been the other solution AFAIK, but this seemed more - defensive?

xlenstra (approved):
I haven't written the code on detecting problems with uniques so I can't say if this 100% correct, but to my untrained eye it looks fine

PR #5399, opened 2021-10-04
Updated amphitheater replacements to temple replacements

xlenstra:
This has the potential to break existing save files


SomeTroglodyte (approved):
First thought: Why? Second: Answer: Leftover json from pre-Religion times.

PR #5400, opened 2021-10-05
Added an advanced system for choosing beliefs

xlenstra:
This PR strikes one of the last remaining points of #4290 off: the AI will now actually choose decent beliefs instead of completely random ones.
I don't quite like this PR as it doesn't solve the real problem at hand: calculate a score for every possible unique and conditional combination. This PR only adds those present in vanilla beliefs, which will not work for modders. Despite this, it is still an improvement over the current situation and as such I've still made this PR.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Beliefs.json
+++ b/android/assets/jsons/Civ V - Vanilla/Beliefs.json
@@ -46,7 +46,7 @@
         "name": "God of War",
         "type": "Pantheon",
         "uniques": ["Earn [50]% of [Military] unit's [Strength] as [Faith] when killed within 4 tiles of a city following this religion"]
-    }
+    },
     {
         "name": "Goddess of Festivals",
         "type": "Pantheon",
@@ -82,6 +82,8 @@
         //"uniques": ["[+15]% Production when constructing [{Ancient Era} {Wonders}]",
         // "[+15]% Production when constructing [{Classical Era} {Wonders}]"]
         // For now this feels like overkill, but I'll leave this here for the future
+        
+        // Alternatively, we could approximate this with "[+15]% Production when constructing [Wonders] <during the [Ancient era]>"
     },
     {
         "name": "One with Nature",
@@ -248,7 +250,7 @@
     {
         "name": "Defender of the Faith",
         "type": "Enhancer",
-        "uniques": ["[+20]% Strength for [All] units in [Friendly Land]"],
+        "uniques": ["[+20]% Strength for [All] units in [Friendly Land]"]
         // ToDo: Should only be friendly territory of cities that follow this religion
     },
     {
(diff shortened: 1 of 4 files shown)


yairm210:
I do like this, and this can be a basis for future evaluations as well, for example choosing policies or wonder benefits :)

android/assets/jsons/Civ V - Vanilla/Beliefs.json
```diff
@@ -264,13 +266,13 @@
...
         "type": "Enhancer",
-        "uniques": ["[+20]% Strength <for [All] units> <when fighting in [Enemy Land] tiles>"],
+        "uniques": ["[+20]% Strength <for [All] units> <when fighting in [Enemy Land] tiles>"]
         // ToDo: Should only be enemy territory of cities that follow this religion
     },
     {
         "name": "Messiah",
         "type": "Enhancer",
-        "uniques": ["[+25]% Spread Religion Strength <for [Great Prophet] units>", "[-25]% Faith cost of generating Great Prophet equivalents"]
+        "uniques": ["[+25]% Spread Religion Strength <for [Great Prophet] units>", "[-25]% Faith cost of generating Great Prophet equivalents", "[Faith] cost for [Great Prophet] units [-25]%"]
```
SomeTroglodyte, line 275: Isn't this a dupe? Or does equivalents not cover the original itself?

core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
```diff
@@ -0,0 +1,185 @@
...
+        if (belief.type == BeliefType.Pantheon)
+            score *= 0.9f
+        
+        return score
+    }
+    
+    private fun beliefBonusForTile(belief: Belief, tile: TileInfo, city: CityInfo): Float {
+        var bonusYield = 0f
+        for (unique in belief.uniqueObjects) {
+            when (unique.placeholderText) {
```
SomeTroglodyte, line 46: I wouldn't have left it at that but ensured the UniqueType can be used

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -364,7 +362,7 @@ object NextTurnAutomation {
                     .flatMap { religion -> religion.getBeliefs(beliefType) }.contains(it.value) 
             }
             .map { it.value }
-            .randomOrNull() // ToDo: Better algorithm
+            .maxByOrNull { ChooseBeliefsAutomation.rateBelief(civInfo, it) }
```
SomeTroglodyte, line 365: No fuzzy factor at all left? Won't players notice the boring predictability?

android/assets/jsons/Civ V - Vanilla/Beliefs.json
xlenstra: There is both the natural generation of great prophets until the industrial era and the buying of great prophets with faith starting from the industrial era, and both need to have a reduction in price

core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
xlenstra: 75% of this PR was made before that was introduced. I agree that it would be better, but most of these aren't on the enum yet anyway and adding all of them required a lot of additional code that is rather add in another PR

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
xlenstra: Random factor is done when rating the belief. Not perfect, but good enough for now imo

xlenstra: And also, the rating is also based on what tiles you own and other factors that change betwee games, so there is always some fuzziness in that

PR #5401, opened 2021-10-05
AI rationing of strategic resources; Hydro Plant re-enabled

SimonCeder:
This PR adds code for the AI to ration certain strategic resources. Gets rid of the old hardcoded "Aluminum" reserves.
Resources that will be needed for space victory are saved for that purpose and even units disbanded and buildings sold when really needed. AI will not trade space race resources when Apollo Project is built. The AI is now a contender in the space race.
Otherwise, when a resource is needed both for non-obsolete units and for buildings (mainly coal and aluminum in the base game), the AI will try to use 50% of its resource for each.
Adds back in the Hydro Plant which the AI is now able to build responsibly.

### DIFF
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -302,7 +302,7 @@ object NextTurnAutomation {
     }
 
     /** If we are able to build a spaceship but have already spent our resources, try disbanding
-     *  a unit and selling a building to make room. */
+     *  a unit and selling a building to make room. Can happen due to trades etc */
     private fun freeUpSpaceResources(civInfo: CivilizationInfo) {
         // Can't build spaceships
         if (!civInfo.hasUnique("Enables construction of Spaceship parts"))
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -45,6 +45,7 @@ object NextTurnAutomation {
             exchangeLuxuries(civInfo)
             issueRequests(civInfo)
             adoptPolicy(civInfo)  // todo can take a second - why?
+            freeUpSpaceResources(civInfo)
         } else {
             civInfo.getFreeTechForCityState()
             civInfo.updateDiplomaticRelationshipForCityState()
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -300,6 +301,39 @@ object NextTurnAutomation {
...
+            return
+
+        for (resource in civInfo.gameInfo.spaceResources) {
+            // Have enough resources already
+            if (civInfo.getCivResourcesByName()[resource]!! >= Automation.getReservedSpaceResourceAmount(civInfo))
+                continue
+
+            val unitToDisband = civInfo.getCivUnits()
+                .filter { it.baseUnit.getResourceRequirements().containsKey(resource) }
+                .toList().minByOrNull { it.getForceEvaluation() }
```
yairm210, line 318: Why tolist?

yairm210:
There's a lot of use of x.getResourceRequirements().containsKey(), I think this could benefit from a .requiresResource() function

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
SimonCeder: Because you can't minby a sequence?

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -681,6 +687,20 @@ class CivilizationInfo {
         // Cache whether this civ gets nonstandard terrain damage for performance reasons.
         nonStandardTerrainDamage = getMatchingUniques("Units ending their turn on [] tiles take [] damage")
             .any { gameInfo.ruleSet.terrains[it.params[0]]!!.damagePerTurn != it.params[1].toInt() }
+
+        // Cache the last era each resource is used for buildings or units respectively for AI building evaluation
+        for (resource in gameInfo.ruleSet.tileResources.values.filter { it.resourceType == ResourceType.Strategic }.map { it.name }) {
```
yairm210, line 692: I doubt it's really necessary to have these values cached, it's possible this is just function inefficiency

Instead of iterating for each resource for each building and unit, you can iterate only on each building and unit sorted by era, and assign resources that you find

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
yairm210: Can't you? I recall doing so

SomeTroglodyte: sequences are almighty! Of course you can aggregate any way you want - you can even asSequence() them!
<details><summary>implementation</summary>

```kotlin
/**
 * Returns the first element yielding the smallest value of the given function or `null` if there are no elements.
 *
 * The operation is _terminal_.
 * 
 * @sample samples.collections.Collections.Aggregates.minByOrNull
 */
@SinceKotlin("1.4")
public inline fun <T, R : Comparable<R>> Sequence<T>.minByOrNull(selector: (T) -> R): T? {
    val iterator = iterator()
    if (!iterator.hasNext()) return null
    var minElem = iterator.next()
    if (!iterator.hasNext()) return minElem
    var minValue = selector(minElem)
    do {
        val e = iterator.next()
        val v = selector(e)
        if (minValue > v) {
            minElem = e
            minValue = v
        }
    } while (iterator.hasNext())
    return minElem
}
```
</details>

SimonCeder: don't know why I thought that then!

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
SimonCeder: Necessary no, but it makes the function a lot faster.
It makes little sense to build the whole resource structure anew for each time the function is called as for the vast majority (all?) of cases there is only a single required resource.
My first implementation was something like
```
val neededForUnits = ruleSet.units.values.any { (it.obsoleteTech == null || !civInfo.tech.isResearched(it.obsoleteTech!!))
                    && resource in it.getResourceRequirements().keys }
```
which is decently fast, but the cached values version takes only 30% as much time for the entire function.

Still the function is not that slow in absolute terms and not called _that_ often (a couple times per turn), so idk? What do you think? I'd add in the `.requiresResource()` of course.

PR #5404, opened 2021-10-05
Minimum city distance across continents

SomeTroglodyte:
... and Type-ify a few uniques on the way (settler+worker).
Should have separated the two.
Runs and unit tests, but has seen no playtesting so far.
Todo: link the issue where the rule was established

Question: I suspect the GameStarter changes for AI capital location are not 100% effective, and that depends on where the CivilizationInfo.units transient may be filled - Alt-F7 seems to fail in more recent Android Studio patches, at least I can't find any writes besides addUnit and removeUnit, and that can't be enough to recreate the collection after a load, or can it (is it TileInfo.setUnitTransients?)? The suspicion is: A tile with high score is chosen. The neighbors will have lower scores. Starting units are placed - on higher difficulties several settlers around the initial best tile. Founding happens in AI automation and goes by unit order in that units collection... So we have a guarantee the original 'best' tile will be settled only if the 'best' settler placed by starter code still is first in order when automation kicks in...


xlenstra (approved):

SomeTroglodyte:
Testing this properly is tricky and time consuming, so I propose to lower our standards and fix bugs later. Several hundred turns of simulation on lowered-water-level archipelago huge maps cropped up mostly no change in the usual city placement patterns, but I got this:
image

PR #5408, opened 2021-10-05
Mayas

SomeTroglodyte:
OK, time to show this off.

Todo:
- [ ] Test "rollover".
- [ ] Check effect on "normal" Great Person brewing thresholds (pretty sure it's correct, but did not yet explicitly compare sources).
- [ ] Decide which texture atlas the extras go into.
- [ ] Notifications needing non-standard icons to "just work" - alternatives.
- [ ] Decide: A new Notification type to make the notifications clickable (and maybe one for policy too)?
- [ ] All those template lines - so now is the time to get the wordung fine-tuned

Digging up the properties and useable images was not hard, their special unit and building are run-off-the-mill and needed no code. But The Long Count... It's a bit over the top, but I like it.

image
That link symbol on the ability unique is no accident.

image

image
Yes you can unlock the ability and get the first GP in the same turn - exactly on turn 23

image
See - Turn 23 is the beginning of creation (that overlap is normal in low resolution)

image
After this I decided to make the one line with the glyphs considerably larger in this popup

</details>


SimonCeder:
I like the long count a lot.
The nation icon is a bit busy but I couldn't really find anything better.

From looking at source it seems GP spawned by long count work just like ones spawned naturally wrt increasing thresholds and such.

SomeTroglodyte:
> GP spawned by long count work just like ones spawned naturally wrt increasing thresholds 

... and they should, right?

SomeTroglodyte:
The close name duality Pyramid-Pyramids and Chichen Itza 2x irks me a bit, but it seems the original had the same. There were maybe a total of 12 tourists on the site when I was there.

SomeTroglodyte:
Notifications need black icons while most Unciv icons are white (Gdx color attribute or tinting works on the white ones but not on the black ones) - I had planned to find a way to bypass the need to redo a black copy of something already there in white to have it show up on notifications - nope not that easy. Notifications could do with some groundwork synergy with our ruleset object superclass for icon extraction, just as civilopedia could. Not in this PR.

android/assets/jsons/translations/template.properties
```diff
@@ -709,6 +712,13 @@ Do you want to exit the game? =
 Start bias: = 
 Avoid [terrain] = 
 
+# Maya calendar popup
+
+The Mayan Long Count = 
+Your scientists and theologians have devised a systematic approach to measuring long time spans - the Long Count. During the festivities whenever the current b'ak'tun ends, a Great Person will join you. = 
+While the rest of the world calls the current year [year], in the Maya Calendar that is: = 
+"[amount] b'ak'tun, [amount] k'atun, [amount] tun =
```
yairm210, line 720: amount, amount2, amount3, otherwise the translation won't work properly

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -116,6 +116,9 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
     FreeExtraBeliefs("May choose [amount] additional [beliefType] beliefs when [foundingOrEnhancing] a religion", UniqueTarget.Global),
     FreeExtraAnyBeliefs("May choose [amount] additional belief(s) of any type when [foundingOrEnhancing] a religion", UniqueTarget.Global),
 
+    MayanGainGreatPerson("Receive a free Great Person at the end of every [comment] (every 394 years), after researching [tech]. (Each bonus person can only be chosen once.)", UniqueTarget.Nation),
```
yairm210, line 119: Remove parentheses from the last part

android/assets/jsons/translations/template.properties
SomeTroglodyte: :facepalm:  of course

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -103,6 +103,9 @@ class GameOptionsTable(
         addCheckbox("Enable Religion", gameParameters.religionEnabled) {
             gameParameters.religionEnabled = it
             cityStateSlider?.run { setRange(0f, numberOfCityStates().toFloat()) }
+            with(cityStateSlider) {
```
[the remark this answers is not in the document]
SomeTroglodyte: Ouch. Just a kludge for kotlin source sigging I was _sure_ I'd removed immediately.

core/src/com/unciv/ui/pickerscreens/GreatPersonPickerScreen.kt
```diff
@@ -7,37 +7,42 @@ import com.unciv.models.UncivSound
...
 
     init {
-        closeButton.isVisible=false
+        closeButton.isVisible = false
         rightSideButton.setText("Choose a free great person".tr())
 
         val greatPersonUnits = civInfo.getGreatPeople()
-        for (unit in greatPersonUnits)
-        {
+        val useMayaLongCount = civInfo.greatPeople.mayaLimitedFreeGP > 0
```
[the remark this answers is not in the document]
SomeTroglodyte: CivInfo flag decides as early as possible extra effort for the calendar is not necessary. This one is something completely different and means the current free GP that is to be chosen should obey The Long Count rules, namely look into the already gotten ones and disable those in the Picker UI - so each type of GP can only be chosen once - until you're out of them and can start over.

It's a bit kludgy, as in a 'clean' api the free GP picker class should have a new parameter to disable offers, and not side-look into the GPP manager for the Maya data, but I was lazy.

yairm210:
In general I don't like the fact that Civ V felt the need to get all special with this civ, but that's a spec problem, not an implementation problem.

SomeTroglodyte:
In a way I can understand - their other unique stuff is unspectacular, meaning they were likely a bit out of ideas, so they had to think of _something_ to make them less boring. Which is why I went overboard, too - so it's at least worth it playing them once :)

SimonCeder:
The Mayas are tied with Poland for best civ in the game imo. The Long Count is absurd if you get Theology early.

PR #5409, opened 2021-10-05
Prevent city states from taunting you when declaring war

SimonCeder:
From Discord: a major civ with city state allies attacking your protected city state can lead to city states phoning you up and taunting you about it. While amusing, this is not intended behavior.

ur


xlenstra (approved):

SomeTroglodyte:
So these City States don't have a Charisma >= 13 or the Intimidate proficiency and are therefore not allowed to cast that spell?

PR #5410, opened 2021-10-05
Barbarians capture civilians and take gold from cities

SimonCeder:
- A barbarian melee unit that would have captured a city will instead take up to 200 gold from the owner, and then disappear
- Barbarians can now capture civilian units. The captured unit will head towards the closest reachable barbarian camp where it will stay until recaptured.

Screenshot_2021-10-05_23-22-12


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -468,12 +479,12 @@ object Battle {
...
         when {
-            // Uncapturable units are destroyed (units captured by barbarians also - for now)
-            defender.unit.hasUnique("Uncapturable") || attacker.getCivInfo().isBarbarian() -> {
+            // Uncapturable units are destroyed
+            defender.unit.hasUnique("Uncapturable") -> {
                 capturedUnit.destroy()
             }
-            // Captured settlers are converted to workers.
-            capturedUnit.name == Constants.settler -> {
+            // Captured settlers are converted to workers unless captured by barbarians (so they can be returned later).
```
xlenstra, line 486: But if you recapture your settler from a barbarian, it would still turn into a worker, right?

SimonCeder: Yes, for now. A future PR will add memory to civilian units so that you can return them their original owner when defeating the barbarians. A settler returned in this way does not become a worker.

PR #5411, opened 2021-10-06
Implemented holy warriors follower belief

xlenstra:
This PR continues development of religion (#4290) by implementing the last missing belief.

Added Uniques:
- May buy [baseUnitFilter] units with [stat] for [amount] times their normal Production cost

Added Conditionals:
- Before the [era]
- During the [era]
- Starting from the [era]


SomeTroglodyte:
The familiar, well-loved "does not end with a space" on an old unchanged line... Never even open a language file in the IDE I say. Those are notepad++ only... As for trailing blanks, if they become scarce, you could take one of those you just left on BaseUnit line 346 :grin:

PR #5412, opened 2021-10-06
Added a label displaying religion is off when it is off

xlenstra:
image
Original:
image


SomeTroglodyte:
Hey the original had capital "F"s

AdityaMH:
> Hey the original had capital "F"s

No different, no objection

PR #5414, opened 2021-10-06
Civ icon redirects to civilopedia

logicminimal:


yairm210:
Please remove the comments.
If you have a way to un-harcode the xp, then great. Otherwise, this is not something that any modder requested, so I'm fine with hardcoding.

logicminimal:
> 
> 
> Please remove the comments. If you have a way to un-harcode the xp, then great. Otherwise, this is not something that any modder requested, so I'm fine with hardcoding.

the problem is, i didnt intend to put those comments in the pr, i literally cant remove them right now, if u can, thats great but i am unable to due to how git works and i doubt if i would be able to fix it without deleting my fork

logicminimal:
@yairm210 merge?

PR #5416, opened 2021-10-06
Fix Educated Elite

### LINKED ISSUE #5306: Great Person gifts from City States do not result in what the Notification promises
Steps to reproduce the behavior (:
1. Play as non-Mongolia, acquire Patronage complete, ally all the CS
2. Wait for "Sydney gave us a Khan as a gift!"
3. Click the notification twice
4. See it's a Great General - or it's entirely missing

image

Yes - that's the only Great General/equivalent in the Civ, either both messages refer to the same unit creation, or the GG and the Khan were independent, and the Khan disappeared.

Will debug myself in time unless I forget...

SimonCeder:
Fixes #5306 
Credit to @SomeTroglodyte for finding the bug - sorry if you were working on a different fix.

Solves the bug by reverting from addUnit to placeUnitNearTile which will not exchange the unit for an equivalent. Great Prophets are not supposed to be gotten from Educated Elite, even in games with religion.


SimonCeder:
Sure, added "May enhance a religion" as well because why not.

PR #5436, opened 2021-10-08
When choosing military units, 'carrying' units are only chosen if they provide needed 'slots' for carrying units.

### LINKED ISSUE #5427: AI creates too much useless carriers
**Platform**
Android 8.1.

**Version**
UnCiv F-Droid v3.17.2 (latest).

**Describe the bug**
AI creates too much useless carriers (a dozens of them) which don't have aviation and can't attack anything. This is literally carriers spam, which makes performance worse. Sorry, but I saw it earlier and right now I don't have corresponding save game to show it to you.

**To Reproduce**
1. Just play the game on the technological era when carriers is opened by bots and then sometimes you will see that bug during the peace time or war time.

**Expected behavior**
AI just shouldn't spam carriers if states don't have aviation for them. So, there is neccessary to add some restrictions or checks.

yairm210:
Should resolve #5427

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -4,10 +4,12 @@ import com.unciv.logic.city.CityInfo
 import com.unciv.logic.city.INonPerpetualConstruction
 import com.unciv.logic.civilization.CivilizationInfo
 import com.unciv.logic.map.BFS
+import com.unciv.logic.map.MapUnit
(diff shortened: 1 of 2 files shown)


yairm210:
It does indeed, good catch, I'll filter for "cannot attack" as well :)

xlenstra:
@yairm210 I may be mistaken, but it seems like this was merged as-is so including the problem I mentioned before, and no other commit seems to have touched this code either?

PR #5437, opened 2021-10-08
Return Civilians captured by Barbarians to original owner; Capture stacked civilians

SimonCeder:
Relates to #4602 as well as #4697.

This PR implements a popup asking the player whether to return a civilian unit to its original owner when recapturing it from barbarians (only). This gives bonus influence or a diplomatic modifier as appropriate. A new field in `MapUnit` is introduced for this purpose, old saves will work but only units built after this will be able to be returned in this way.

Screenshot_2021-10-08_13-33-38

I also resolved a bug which I didn't at first realize wasn't intended behavior - in the original, when a civilian and military unit are stacked on the same tile and you defeat the military unit with a melee attack, you are supposed to move into the tile and capture the civilian unit in the same attack, rather than having to make a separate attack to capture the civilian unit.
Looking at the code of `postBattleMoveToAttackedTile` it seems this was always the intention, but because `canMoveTo` was called before capturing the civilian, it would always return false.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/AlertPopup.kt
+++ b/core/src/com/unciv/ui/worldscreen/AlertPopup.kt
@@ -324,7 +324,7 @@ class AlertPopup(val worldScreen: WorldScreen, val popupAlert: PopupAlert): Popu
                 }).row()
             }
             AlertType.RecapturedCivilian -> {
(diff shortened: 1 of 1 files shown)


yairm210:
Looks good!

PR #5446, opened 2021-10-09
Fixed the visual gaps in territory borders

avdstaaij:
This change is probably best explained with an image:

Comparison2

This was surprisingly complicated to fix.

To achieve this, I added a distinction between "convex" and "concave" border segment sides. A border segment is left-concave if the left shared neighbor of the tile inside of the border segment and the tile outside of the border segment belongs to the border's civ. Otherwise, it is left-convex.

<details>
  <summary>Example of a left-concave right-convex border segment</summary>

LeftConcaveRightConvexBorder

</details>

Instead of a single inner and outer border image, there are now three variants of each: a convex-convex variant, a convex-concave variant and a concave-concave variant. In principle, this could have been done with only two variants: a convex half-segment and a concave half-segment. That would however have doubled the amount of image instances used for borders.

Because diagonal border images are not aligned with the pixels, there is a small imperfection at concave corners:

<details>
  <summary>Concave corner example</summary>

Concave-imperfection

</details>

This was as close as I could get it. At any rate, I think this is an improvement over the gaps. Perhaps a different kind of inner border image could fix this imperfection.

The convex border image is shaped in the same way as the concave one: `\____/`. You might have expected it to be shaped like `/____\` instead. The reason for this is that the overlap this causes in the convex corners actually looks better than the alternative.

<details>
  <summary>Convex corner example</summary>

Convex-corner

</details>

This trick doesn't work so well for the concave corners, so this doesn't solve the mentioned imperfection.

Finally, the logic used by the added functions `get[Left/Right]SharedNeighbor` could perhaps be used to implement zone of control more efficiently as well (see #4085).


yairm210:
Nice!
Looks good!
Just move the clock position function and map, and I'll merge :)

avdstaaij:
> So much brain matter for such a little thing. That deserves my utmost respect. I would probably have gone with some blurring, so the gaps are even wider but not as sharp-edged.... But the results look stunning, even those demonstrating the compromises...

In hindsight, this was probably more trouble than it was worth, but I am quite happy with the result :)

PR #5449, opened 2021-10-10
Likely fixed a bug where AI nukes could hit nations they had a peace treaty with

xlenstra:


yairm210:
Please typify the blast radius unique :)

PR #5458, opened 2021-10-10
Update template.properties

Smashfanful:
Added new translation line


SomeTroglodyte:
Thanks to that "Hidden when [Diplomatic] Victory is disabled" unique this is automatically available, but perhaps it is cleaner this way...

PR #5479, opened 2021-10-13
Updated google play & itch.io screenshots to recommended sizes

xlenstra:
As discussed in #5375


logicminimal:
wpuldnt fantasyhex be better?

xlenstra:
Good point, completely missed the fact I was on the other tileset. I'll update them momentarilly.

PR #5486, opened 2021-10-14
Update Czech.properties

stoupa007:
- many new strings translated


xlenstra:
Placeholder `[stats]` not found in `Czech` for entry `[stats] whenever a Great Person is expended`
    Placeholder `[tech]` not found in `Czech` for entry `Receive a free Great Person at the end of every [comment] (every 394 years), after researching [tech]. Each bonus person can only be chosen once.`

PR #5492, opened 2021-10-15
Allow the player to see which era the opposing civ in diplomacy screen is in

logicminimal:
Warning:(195, 76) Conditional branch result of type String is implicitly cast to Any

Warning:(195, 63) Conditional branch result of type Era is implicitly cast to Any

i hope those warnings dont cause problems


logicminimal:
also removed the extra .onclick

logicminimal:
> 
> 
> My oh my. Still, for a beginner you seem to have a certain knack, your main problem seems to be some... avoidance of hard thinking. So let me encourage you: That pair of warnings _can_ be understood, just look closely, bottom up, what you're asking kotlin to do. What's this, what's that, and what will I get then I apply a certain operator to those two, and what will a string template do with the result... If you get why the compiler warns, 5 points. When you get why it works _nevertheless_, 50 points. It's still kinda impolite towards the poor runtime to leave it like that, moreso because the result is not translatable (as in Unciv translations) as it is.
> 
> And with the onClick's - don't guess. True, having both is a problem as predicting what will happen when a pixel belonging to the areas of both nested actors is clicked may not be intuitive - will Gdx fire events at both, or the bottom one or the top one? However, in the resulting code - what will now happen when we click/tap that label, and what will happen when we click the nation icon, and is that exactly what we want?

right... i need to add translations for it. 

no idea about that onclick lines

the nation icon works very fine, except if we call it from the main game menu like you mentioned in a suggestion about adding civilopedia to the main page, but it cant be done so

i will check those warnings tomorrow, probably will struggle though, i guess i need to first check if it absolutely is a "string" then officially convert it to string. thanks for help

logicminimal:
btw @SomeTroglodyte im trying to really get into the development of this game but i dont know almost anything about the project structure.. any tips?

PR #5494, opened 2021-10-16
Fixed more broken uniques, e.g. polynesian and ethopian national abilities

xlenstra:


Kleznam:
Yup good job xlenstra but still Naya's great prophet s from the long count don't work

xlenstra:
> Yup good job xlenstra but still Naya's great prophet s from the long count don't work

Did you choose it after there were already `amount of players / 2 + 1` religions in the game? If so, no bug. Otherwise, please post this on discord or open another issue.

Kleznam:
Ahh there only was another religion tho

yairm210:
This is a much more common use case than I expected

yairm210:
Also, unit.getMatchingUniques can provide its own conditional state, so you don't need to set it anywhere else

yairm210:
But only for uniqueType ;)

Kleznam:
I apologize Maya's seem to work with prophets

PR #5508, opened 2021-10-17
Provide more information to waiting players in multiplayer

thepianoboy:
A few things were bothering me when I was waiting for the next turn in a multiplayer game and I decided to fix them.

---

Summary of changes:

- GameInfo.kt
  - Add timestamp for the start of the last turn to save info (I hope to use this to add a multiplayer turn timer in the near future)

- MultiplayerScreen.kt
  - The current turn indicator for multiplayer games will show for how long it has been the current player's turn

- WorldScreen.kt
  - Next Turn Action will display which player we are waiting for rather than the generic "Waiting for other players..."

Notes:

- Visually checked on all supported resolutions, doesn't seem to overcrowd the UI.
- Tested in live multiplayer game for 10 or so turns; not a significant enough change to warrant an extended test, IMO.

---

This is a great project  :+1:, I hope I can contribute something more substantial soon.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -683,7 +683,8 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Cam
 
     private fun getNextTurnAction(): NextTurnAction {
         return when {
-            !isPlayersTurn -> NextTurnAction("Waiting for [${gameInfo.currentPlayerCiv}]...", Color.GRAY) {}
+            !isPlayersTurn && gameInfo.gameParameters.isOnlineMultiplayer -> NextTurnAction("Waiting for [${gameInfo.currentPlayerCiv}]...", Color.GRAY) {}
+            !isPlayersTurn && !gameInfo.gameParameters.isOnlineMultiplayer -> NextTurnAction("Waiting for other players...", Color.GRAY) {}
 
             viewingCiv.shouldGoToDueUnit() ->
                 NextTurnAction("Next unit", Color.LIGHT_GRAY) {


yairm210:
Tests are my bad

SomeTroglodyte:
So, how does one go about testing this properly, anyway? Is there some trick to avoid some of the overhead in getting an MP game set up and at least two environments (desktop debugger, avd, hardware, whatever) to run the same codebase and that same test game?

... and regarding "from thepianoboy:master" -> see [some old notes of mine](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Unciv-notes/Working%20in%20the%20IDE%20and%20Contributing.md#branches-are-important). Something a reviewer notices once they actually try to run a PR's code, as it requires extra key strokes...

yairm210:
Single player should have a "waiting for other players" message

thepianoboy:
> What about undiscovered knowledge about who else is in the game, that we shouldn't disclose here? 

The game is already doing that with respect to the "Current Turn:" display in MultiplayerScreen. I guess you could just _not_ look at it, but there are other points of failure here too: for example, if players want to start as specific Civs, the player who sets up the game needs to know all of them anyways. 

This is probably a matter for a different discussion, but perhaps letting a player set an in-game nickname would be a viable way to hide strategic information like that? I wouldn't mind working on that for a future PR, I'm currently working out a way to implement a multiplayer turn timer, so it would probably be something I could do while I'm experimenting with the save-synchronization behavior.

> What I saw was, however, Alice displaying "waiting for Alice" meaning "waiting for dropbox to become reachable" - but that "waiting for Alice" would still seem bullshit to the _human_ reader.

You're right that it reads like bullshit, but shouldn't the player be **much** more concerned about the popup dialog saying that they can't sync the game state? It's probably _less_ BS, but "Waiting for other players..." isn't exactly accurate either. This might be something for a different discussion as well; I don't know if this is something I'm equipped to help with.

> Nah, I'm not good at this. Most of the multiplayer code in Unciv still is too much of a mystery to me.
Ha, you can say that again! It's pretty arcane, especially coming from the perspective of a new contributor.

GGGuenni:
@thepianoboy 
> I wouldn't mind working on that for a future PR, I'm currently working out a way to implement a multiplayer turn timer, so it would probably be something I could do while I'm experimenting with the save-synchronization behavior.

Maybe a bit offtopic to the PR but I would love to discuss the implementation of a turn timer with you if you are interested. I'm currently thinking of a way to enable players to vote kick other players that abanded the game without resigning and I think we are onto something similar! If you are on discord you can hit me up there (same name). I would say I've got some solid understanding of the unciv multiplayer by now and I'm willing to help😄

PR #5513, opened 2021-10-18
Enumified all terrain uniques in terrains.json

xlenstra:


yairm210:
@SomeTroglodyte  lWhat's nice about using the enums and not text is for can quickly find all references, so even if you're missing context you can give it very quickly

PR #5523, opened 2021-10-20
Enumified all remaining resource & improvement uniques

xlenstra:
Turns out there were a lot more of these then I thought, but they're all there now

Added conditionals:
- \<before discovering [tech]>
- \<after discovering [tech]>
- \<before adopting [policy]>
- \<after adopting [policy]>


SomeTroglodyte:
> there were a lot more of these

... earning you the sisyphus award? :smile_cat:

PR #5527, opened 2021-10-20
Improved MultiplayerScreen performance

GGGuenni:
Achieved by introducing a separate GameInfo for multiplayer games. This new GameInfo contains only the necessary data to load a multiplayer game (gameID) and to display the current player. Additionally, it holds the multiplayer settings to turn off notifications per game which will be introduced later (discussed in #5116). 

The new MultiplayerGameInfo makes the MultiplayerScreen much faster especially with a lot of games as all GameInfos together are probably <1KB instead of up to 1MB per game for huge games. Old save files in the Multiplayer folder are loaded without problems since GameInfo files can be loaded as MultiplayerGameInfo objects since all necessary variables are labeled the same. They will be converted on the next refresh automatically.

I've also added a popup when joining a multiplayer game to cover up some ANRs. Implementation is a bit weird though.

Some new data that is introduced in #5508 has yet to be added to the MultiplayerGameInfo so I will have to wait for that PR to be merged first.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -410,6 +410,8 @@ class GameInfoPreview() {
         currentPlayer = gameInfo.currentPlayer
         turns = gameInfo.turns
         currentTurnStartTime = gameInfo.currentTurnStartTime
+        //We update the civilizations in case someone is removed from the game (resign/kick)
+        civilizations = gameInfo.getCivilizationsAsPreviews()
 
         return this
     }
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
`class GameInfoPreview` - same idea same purpose, expand that instead?

GGGuenni:
Yea true never had seen that before 
Then the question is how should I expand on it:
- just derive MultiplayerGameInfo from GameInfoPreview
- put the multiplayer setting into GameInfoPreview (don't think it belongs there)
- split it into two completely different things with an extra multiplayer settings file

GGGuenni:
Changed everything from MultiplayerGameInfo to GameInfoPreview.

I should also note that I had to change the icon that is used to mark the games, which are waiting for the player, from the nation icon, which is transient, to something else. Went with the exclamation mark now.
NewTurnImage

PR #5535, opened 2021-10-21
Inner Sea map type

SimonCeder:
Adds a new map type, Inner Sea.
It's essentially an inverted Pangaea, with a higher randomness factor to allow some bays and islands to form in the center.

<details><summary>Examples</summary>

Screenshot_20211021_212310

Screenshot_20211021_212529

Screenshot_20211021_212641

</details>


SimonCeder:
About as you'd expect:
Screenshot_20211021_225048

(to be clear, there's only one sea, the wrap displays the world several times)

Issue #5440, opened 2021-10-08
City UI slow in late game. Stats depsgraph?


will-ca:
**Platform**
Android

**Version**
3.17.4, Play Store

**Describe the bug**
The build queue and population assignments UIs are very slow for large cities in the late game. It feels like every selected or added item is triggering a full recomputation of the entire empire's population and stats, or something. (I suppose that's probably how it updates the displayed city stats, accounting for all social policies, wonders, etc.)

Technically, the UI itself remains semi-interactive, in that it can sometimes be scrolled and clicked. But the unit/building selector layout element doesn't update until several seconds after the current queue display expands, so there's a broken layout for quite a while. And the scroll position also resets itself once the operation is completely processed— So trying to use it in that time is pointless and frustrating since it will jump around.

This is true for assigning specialists, queuing production, and working tiles.

**Expected behavior**
Expensive calculations should never be done in the middle of interacting with a UI widget. If they must, it should ideally be more asynchronous, so layout and interactivity aren't broken during the computation.

If the expensive calculations are not needed for the interactive display, then they should be deferred and done only once when exiting the city screen IMAO. If they are needed, E.G. to iterate over all buildings in all cities for happiness and thus production penalty... Then I suppose they should be cached at the start of the turn, and recomputed only when something like social policy adoption or building purchase changes them (probably not when reassigning population, and definitely not when doing a UI-only change like adding to the build queue).

**Additional context**
Normally I'd be hesitant to complain about a performance issue, but there does seem to be precedent showing it's accepted here.

Perhaps it would be elegant to implement a dependency graph for game state— Cache all values, and recursively check if their inputs have changed to decide whether to recompute? If it's not already done, it could probably improve performance in more places than just the city screen, while still using lots of dynamic calls that always deliver the most up to date values.

---

I suppose this is actually a single effect with multiple issues behind it:

* [ ] Make UI more asynchronous from empire state, so the City screen stays interactive even with lots of computations.
* [ ] Reduce the amount of computation done during city screen interaction— Cache when entering and defer until exiting where possible.
* [x] Maybe implement game-wide stats cache and dependency graph.

K3XRad:
I can confirm that issue. I even got some crashes with that slow UI when I tried to buy some buildings in a cities. Sorry, but I don't saved the logcat output with those crashes.

yairm210:
Recent performance improvements should have solved this

PR #5536, opened 2021-10-21
City construction speedup with caching stats from tiles

yairm210:
May not completely *resolve* #5440, but certainly goes a long way towards it.

Background:
image

Performance profiling on a particularly large game shows that out of ~8.9 seconds it took this city to gather the construction information, which is almost all citystats.update (8.4s), a whopping 5.4s was spent calculating tile yields, BOTH in updateBaseStatList AND in updateCityHappiness.
This is due to the fact that for each construction, we want to know how many turns it will take to construct, therefore we run the whole gamut of gathering all yields, checking all uniques, etc. Gathering tile yields is going over *every* tile and for each one going over all uniques to try and find, like, 2 or 3 specific ones.
So:
Foreach construction
Foreach worked tile
Foreach unique
And it does this TWICE!
No wonder it takes so long!

I have already typeified the uniques there, leading to a decrease from 9.1s to 8.9s, but that's petty change compared to this.

This PR does 2 things:
1. Saves tile stats in cityStats, effectively reducing the 2 calls to getTileStats() to one. This by itself already reduces the function time by 2.7s/8.4s = 32%
2. Makes the turnsToConstruction *always* used cached values, reducing it by another 32% to around *a third* of its previous time.

"Slowness" issues are difficult to quantify (how fast is fast enough? When is this resolved?), but I think this is the major landmark.
I highly doubt we'll be able to find anything else that will so drastically improve performance, except for maybe having a cached uniqueMap in the civInfo, which probably requires having cached unique values in all its subset first, so that's a huge project.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -337,7 +337,9 @@ class GameInfo {
             civInfo.initialSetCitiesConnectedToCapitalTransients()
 
             // We need to determine the GLOBAL happiness state in order to determine the city stats
-            for (cityInfo in civInfo.cities) cityInfo.cityStats.updateCityHappiness()
+            for (cityInfo in civInfo.cities) cityInfo.cityStats.updateCityHappiness(
+                cityInfo.cityConstructions.getStats()
+            )
 
             for (cityInfo in civInfo.cities) {
                 /** We remove constructions from the queue that aren't defined in the ruleset.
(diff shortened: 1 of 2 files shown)


yairm210:
image

Another victory for the mastermind
DTO gathering itself is down to a final of 2.7s from 8.9s, or in other words - 30%.
Further improvement seems to all lead to getMatchingUniques, or in other works - typifying uniques so they use enum comparison and not string comparison, which before Unciv I never *imagined* was such a heavy part of performance improvements.

yairm210:
Graphs are with "profile" - cpu - "record" - flame graph
That gets me the best bang for buck, other visualizations aren't as immediately clear

PR #5556, opened 2021-10-25
Regions part 1 - subdivide generated maps into regions, and use to place civs

SimonCeder:
Sorry about large PR, wanted to get a minimum viable product that actually did something for unciv before I did a PR.

During map generation, the map is divided into a number of rectangular regions with equal fertility (roughly, unimproved yield). As many regions are generated as there are major civs in the game.
Regions are classified as Tundra, Desert, etc, etc, or Hybrid if very mixed.
A good starting position is found for each region, close to the center of the region if possible.
Civs are then allocated to one region each, taking into account their start biases.

World wrap, and rectangular vs hexagonal maps all work.

How slow is this? Huge map, 10 civs:
```
MapGenerator.generateRegions took 69.53ms
MapGenerator.assignRegions took 28.13ms
```
Total time spent is similar to spreadResources. There is a slight time saving when looking for start positions in GameStarter as the major civs have predetermined starts. In the end there might not be meaningful performance difference if working luxuries placement into the regions code saves time in spreadResources.

Right now the main effects of this PR are:
- Civs don't necessarily spawn right on top of their preferred terrain, rather they will spawn in a region with much of that terrain
- Civs are more evenly distributed across the map, and the placement takes into account the size of the continents they are placed on (so if continent A is three times the size of continent B, continent A will get 3x as many civs).

Problems:
- Uncivs terrains are much more spread out than in civ V, for example our Jungles don't really cluster as they do in the original. For this reason it is rare for a region to be classified as Jungle.
- Similarily, Unciv loves tundra and snow on many map types, so it is common to get several Tundra regions.
- I expect both of the above to be resolved with further refinement to map generation.
- There's a lot of technical data stored in "uniques", which kind of clutters the civilopedia. Do we want to create a "don't display anything below this" unique? A conditional hacked as a "don't display this in civpedia" tag?

Future:
- Place city states using regions
- Place luxuries using regions (so it is possible that eg citrus only grows in one region)
- Rework start normalization (the old consolation prize function)


SimonCeder:
Added the new HideInCivilopedia flag to the uniques with region information so they don't mess up the civilopedia so bad.

yairm210:
Reminding you that if you're planning on adding all these, to add an autodetect of types to reduce the number of specified uniques (for instance all grassland uniques should not need to be there at all)

yairm210:
"can act as a region" and "a region is formed" uniques always come together, can be merged to one unique

core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
```diff
@@ -155,6 +155,26 @@ enum class UniqueParameterType(val parameterName:String) {
...
+            return UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
+        }
+    },
+    /** Used for start placements */
+    TerrainQuality("terrainQuality") {
+        private val knownValues = setOf("Undesirable", "Food", "Desirable", "Production")
+        override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
+                UniqueType.UniqueComplianceErrorSeverity? {
+            if (parameterText in knownValues) return null
+            return UniqueType.UniqueComplianceErrorSeverity.RulesetSpecific
```
yairm210, line 175: Should be invariant, since this parameter is setting no matter what ruleset

SimonCeder:
> "can act as a region" and "a region is formed" uniques always come together, can be merged to one unique

Ok, done

> Reminding you that if you're planning on adding all these, to add an autodetect of types to reduce the number of specified uniques (for instance all grassland uniques should not need to be there at all)

I added a guessing function as fallback in case there are no defined qualities for the terrain type, and it gives OK results and should work well for mods unless they have a different baseline for tile yields.
But those uniques do need to be there if we want to preserve the Civ V behavior, so for example in regions that have been defined as plains the start position tends to avoid grassland and prefer plains..

yairm210:
Code looks okay 👍
What happens when creating a nap in the map editor?
Can you upload an image of a generated map?

SimonCeder:
Creating a map in the map editor bypasses the region generation entirely. Starting a game on a premade map uses the same placement algorithm as before.

Generated maps (with a visualizing function to color in regions, so it's more clear what's going on (the visualizer is not in the PR)):
<details><summary>Huge Images</summary>

Screenshot_20211104_124335

Screenshot_20211104_124619
(Netherlands is getting a bit screwed here, not being able to access some of "their" land - snaky continents is a weakness of the algorithm)

Screenshot_20211104_124839

Screenshot_20211104_125021

</details>

SomeTroglodyte:
Sorry for the late nitpick:
```kotlin
                    val effectiveTotalProd = if (totalProd >= totalFood * 2) totalProd
                        else (totalFood + 1) / 2 // Can't use all that production without food
```
(MapRegions.evaluateTileForStart, Ring-specific scoring ring 2)

... Sounds like it's completely missing the intention. Sounds more like `totalProd.coerceAtMost(totalFood * 2)` ...?

PR #5581, opened 2021-10-28
Declarative Mod Compatibility

SomeTroglodyte:
Unfinished draft - as PR only for easier inspection, @xlenstra  go ahead take what you find useful for #5577 or followups.


yairm210:
I think this is no longer relevant
I think the ruleset should determine the compatibility, since that's more dynamic and liable to change

PR #5593, opened 2021-10-29
Added Server Sided Mutex

GGGuenni:
Added server-sided mutex to prevent race conditions especially since we have two files now which should stay synced. Allows us to use resign/kick even on someone else's turn. Tested with a private Dropbox account and worked flawlessly. I've put everything into dropbox.kt to be bundled for this draft PR but @thepianoboy will make some improvements to it.


yairm210:
Look okay, doesn't look like you're actually using this lock anywhere, I assume that's why this is still a draft, yes?

GGGuenni:
> Look okay, doesn't look like you're actually using this lock anywhere, I assume that's why this is still a draft, yes?

Yea I wanted to wait for thepianoboy to work on the usage and coordinate collaboration with him, hence the draft.
It will have to be used everywhere a file is uploaded

thepianoboy:
I wrote up a simple wrapper over the weekend that works a lot like the [Mutex](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/index.html) from Kotlin's coroutines library. It feels about right in terms of how it looks in the code, but I would like to make sure everything works in Android Studio before I push it back upstream.

GGGuenni:
PR should be good to review now 
Or should we implement the use of the mutex in this PR too?

core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
```diff
@@ -156,3 +183,109 @@ class OnlineMultiplayer {
...
+
+    /**
+     * Release a server lock acquired by tryLock or lock
+     * @see tryLock
+     * @see lock
+     */
+    fun unlock() {
+        DropBox.deleteFile("${OnlineMultiplayer().getGameLocation(gameInfo.gameId)}_Lock")
+        locked = false
+    }
```
xlenstra, line 282: Shouldn't you make sure you hold the lock before trying to unlock?

xlenstra (approved):

core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
```diff
@@ -151,3 +173,112 @@ class OnlineMultiplayer {
...
+
+    /**
+     * Release a server lock acquired by tryLock or lock
+     * @see tryLock
+     * @see lock
+     */
+    fun unlock() {
+        if (!locked)
+            return
```
will-ca, line 271: Personal preference: I'd throw an exception to crash early here if any users of this lock tried to unlock it without having acquired it.

…Hm… This might be horrible practice, so maybe don't do this, but I think throwing an exception in a new thread would get us Google Play logs without crashing the client?

core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
```diff
@@ -151,3 +173,112 @@ class OnlineMultiplayer {
...
+}
+
+/**
+ *	Wrapper around OnlineMultiplayer's synchronization facilities.
+ *
+ *	Based on the design of Mutex from kotlinx.coroutines.sync, except that when it blocks,
+ *	  it blocks the entire thread for an increasing period of time via Thread.sleep()
+ */
+class ServerMutex(val gameInfo: GameInfoPreview) {
+    private var locked = false
```
will-ca, line 194: Personally I found "`locked`" to be just a little confusing, as I expected it to be more global, I.E. "Is this lock currently held by *any* client?/Does this file exist on DropBox?", but it looks to be more like "Is this client currently holding the lock?"

will-ca (approved):
Looks good to me, aside from some nitpicky stuff.

However, I'm not comfortable merging myself as multiplayer is an area of the code that I've yet to work on or even look at much.

xlenstra:
One other problem I recently thought of, what if you lock a game, but the game crashes or is closed before the file is removed. Then the multiplayer game stays locked forever, right?

GGGuenni:
> One other problem I recently thought of, what if you lock a game, but the game crashes or is closed before the file is removed. Then the multiplayer game stays locked forever, right?

We could check the creation date of the lock file in dropbox in the tryLock function as we download the file metadata with fileExist anyway and just delete it if it's let's say 1 minute old.

core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
GGGuenni: I don't think calling unlock when not previously locked is a problem we really need to care about. 
The mutex should be run in a separate thread anyway as lock() blocks it, which would result in ANRs and that thread probably also has a catch inside it as it's mandatory when working with the OnlineMultiplayer class as basically everything throws exceptions, so I don't think we would get Google Play logs. Or would we? When do Google Play logs get generated?

will-ca: > […] so I don't think we would get Google Play logs. Or would we? When do Google Play logs get generated?

I haven't confirmed this personally, but it seems that log entries are generated whenever any thread (including the main thread) exits due to an uncaught exception, which also means that you can force a log entry by rethrowing an exception in a short-lived thread (`catch (e: Throwable) { thread { throw e } }`).

https://github.com/yairm210/Unciv/pull/5833#issuecomment-1001064624

https://support.google.com/googleplay/android-developer/answer/9859174

core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
```diff
@@ -151,3 +179,137 @@ class OnlineMultiplayer {
...
+            val date = metaData?.getServerModified()
+            println(date?.time)
+            println(System.currentTimeMillis())
+            // 30 seconds should be more than sufficient for everything lock related
+            // so we can assume the lock file was forgotten if it is older than 30 sec
+            if (date != null && System.currentTimeMillis() - date.time < 30000) {
+                return locked
+            } else {
+                DropBox.deleteFile(fileName)
+            }
```
GGGuenni, line 237: I tried to implement the lock release for forgotten lock files but the Date object returned by String.parseDate() returns an epoch timestamp which is offset by one hour (UTC -1) compared to System.currentTimeMillis(). Shouldn't epoch time be timezone independent? Don't get what's going on, so just wanted to ask if someone has a clue or if I just have to check for < 3630000 instead of 30000 ¯\_(ツ)_/¯.

PR #5594, opened 2021-10-29
Remove Discord RPC checks for unsuitable devices

asda488:
Hi, 
Firstly, I'm not a Kotlin developer so sorry if the style is out of order, please change things as necessary. I have changed `RaspberryPiDetector` to a general `DiscordCompatibleDetector.kt`, which uses `System.properties("os.arch")` instead of `/etc/os-release` to find the system (installed Java) architecture, in this case ARM as there are no Discord and RPC binaries available.

Thank you,
asda488.


asda488:
That is a good point, and I would say there's two ways ahead from here (albeit with my no Kotlin ability):
One, we iterate through the files in the `discord_rpc` folder and check if the operating system/architecture is compatible, else just skip (and print warning?), if compatible load.

Two, we leave the code "as it is". I would suggest this as the code is working fine on all desktop archs/OSes (win x86, win x64, linux x86_64, linux x64, linux armhf, linux arm64), and anyone wanting it to work on a different arch (or OS even) would either have the knowledge to fix this, or open another issue.

Apologies if two sounds stupid.

yairm210:
K, looks fine to me
If we can catch the errors at load that also future-proofs us to future architectures and devices (and operating systems?)
Less code, less bugs

PR #5602, opened 2021-10-30
Show construction icons in Cities Overview.

will-ca:
Personally I find icons way easier to scan in a big table than text that has alternating lines between it and doesn't always look aligned (due to different lengths).

It's especially easier to find the city specializations (E.G. "Produce Science"), which you probably want to be the most aware of due to their potential to waste dozens of turns, as they have different colours.

I think Civ V shows icons.

The icons are non-interactive. The city buttons right next to them can be clicked anyway.

Also, I've often been thinking: Good job to whoever designed the icons. Dozens of different options are all fairly distinctive and recognizable with only a tiny silhouette.

`*0.8f` puts them at the same size as in `CityConstructionsTable.kt`.

Before:

image

After:

image

First time doing anything with Kotlin. First time writing in anything where typing is an integral part of the program logic, actually.. May have copied and pasted/poked through some Java a while ago.

### DIFF
--- a/core/src/com/unciv/ui/overviewscreen/CityOverviewTable.kt
+++ b/core/src/com/unciv/ui/overviewscreen/CityOverviewTable.kt
@@ -141,6 +141,8 @@ class CityOverviewTable(private val viewingPlayer: CivilizationInfo, private val
             }
             citiesTable.add(button)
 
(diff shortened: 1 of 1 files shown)


xlenstra (approved):

SomeTroglodyte:
In your screenshot, Eregil has an empty white circle. I would probably avoid that and leave the cell empty instead.
Hints: Table.add() or Table.add(null) are allowed and do that - empty cell. I don't remember whether `city.cityConstructions.currentConstructionFromQueue` will be null or an empty string in that case - it's an empty string.

Issue #5552, opened 2021-10-25
Command Line Interface and API-GUI convergence?


will-ca:
**Is your feature request related to a problem? Please describe.**
Managing large empires is tedious and painful. A lot of stuff, like build order, unit promotions, finding resources or freeing them up, is highly repetitive.

I like a slick UI as much as the next guy. But in an open-sourced program with clean code, it's frustrating to have to spend several minutes and hundreds of mouse clicks to do something that could be described with a single line of scripting or with a reusable function.

**Describe the solution you'd like**
It would be great to have a CLI that can be used to access all information and perform all actions that are currently available in the UI.

E.G.:

`len(Empire.Cities)` to count the number of cities to better plan strategic resource use, instead of having to manually count in the Overview screen.

`[city for city in Empire.Cities if any("Aluminium" in building.Consumes for building in city.Buildings)]` to get a list of all cities that have Spaceship Factories or Hydro Plants, instead of having to click into dozens of cities and scroll through the entire building list for each.

`Empire.Cities.map((city) => city.Governor.SetPriority("Food"))` to tell all cities to stop assigning scientists and focus on population growth (once governors get implemented).

`[city for city in Empire.Cities if city.Tiles.has("Aluminium")]` to figure out where your strategic resources are, instead of having to stare at the map for several minutes, checking all city centers separately, and still missing them.

`UITools.Messages.ResourceDiscovery("Uranium", target="Mongolia")` to manually re-trigger the "You have discovered [X] sources of [Resource]!" notification, but have it target and cycle through resources on visible tiles near a foreign country, instead of having to spend several minutes staring at every tile and then finding out that you've still missed some after you've already declared and ended a war.

`for i, city in enumerate(sorted(Empire.Cities, key=lambda city: -city.Stats.Production.Base)); do if i < Empire.Resources["Coal"]; then city.Production.prepend("Factory"); fi; city.Production.append("NukeSub"*3, "Cruiser"); done` to automatically build as many factories in your largest cities as you have coal resources, and queue up three nuclear submarines and one cruiser to be built in all cities.

`buildorder = ["Monument", "Granary", ...]; for city in Empire.Cities: if not city.Production: for building in buildorder: try: city.Production.append(building)` to have cities follow a custom automatic build order.

`UITools.Map.HighlightCoords(tile.Coord for tile in Map.Tiles if "Coal" in tile.Resources)` to display some sort of visual overlay for all visible coals resources, instead of again having to scour every tile by eye.

[...]

PR #5608, opened 2021-10-30
Command-line interface, reflective IPC protocol, and scripting language bindings

will-ca:
Implementation of #5552.

---

**Docs/Specs:**

* **[Kotlin and Specs](https://github.com/will-ca/Unciv/blob/cli/core/Module.md)**:
  * **[Design Principles](https://github.com/will-ca/Unciv/blob/cli/core/Module.md#design-principles)**
  * **[Class Overview](https://github.com/will-ca/Unciv/blob/cli/core/Module.md#class-overview)**
  * **[REPL Loop](https://github.com/will-ca/Unciv/blob/cli/core/Module.md#repl-loop)**
  * **[IPC Protocol](https://github.com/will-ca/Unciv/blob/cli/core/Module.md#ipc-protocol)**
* **[Python Binding Implementation](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md)**
  * **[Overview](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#overview)**
  * **[Foreign Object Wrappers](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#foreign-object-wrappers)**
  * **[Foreign Object Tokens](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#foreign-object-tokens)**
  * **[Assigning Tokens to Paths to Get Wrappers](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#assigning-tokens-to-paths-to-get-wrappers)**
  * **[API Modules](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#api-modules)**
  * **[Performance and Gotchas](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#performance-and-gotchas)**
  * **[Error Handling](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#error-handling)**
  * **[Other Languages](https://github.com/will-ca/Unciv/blob/cli/android/assets/scripting/enginefiles/python/PythonScripting.md#other-languages)**

---

**Examples:**

* [Player automation macros.](https://github.com/yairm210/Unciv/pull/5608/#issuecomment-962827276)
* [Generate map from heightmap, or math function.](https://github.com/yairm210/Unciv/pull/5608/#issuecomment-972555750)
* [Generate map from colour-coded image of biomes.](https://github.com/yairm210/Unciv/pull/5608/#issuecomment-974409933)
* [Mod script handlers.](https://github.com/yairm210/Unciv/pull/5608#issuecomment-994405224)
* [Story event popups.](https://github.com/yairm210/Unciv/pull/5608#issuecomment-994395007)
* [Test suite.](https://github.com/yairm210/Unciv/pull/5608#issuecomment-980037830)

---

**UI**

* [Warnings, startup scripts, and error handling.](https://github.com/yairm210/Unciv/pull/5608#issuecomment-994403155)

---

**Execution Model/REPL Loop**

This simple, thirty-step process…

(There may be more calls to TokenizingJson than shown in the above diagram, because it's used as the universal serialization tool for encoding and decoding every packet, as well as the arbitrarily-structured JsonElement data in every packet.)

---

**Adds:**

* Console-emulator screen reached by pressing the backtick (`/~) key on the map screen.
* TAB-autocomplete and up/down command history navigation.
* Very simple basic API for extending the console— Adding a new language/backend is done by basically writing a single new function/generator for its REPL. Take a string, return a string, and change whatever stats and call whatever bindings in between.
* Simultaneous running of multiple backend instances of different or same types, and UI for starting, switching between, and terminating them.
* Path parsing and resolution for properties, indices, and keys— Allows accessing arbitrary properties as E.G. JSON requests, or code strings (E.G. `"civInfo.cities[0].cityStats.happinessList"`)— Basis for IPC protocol with interpreters.
* Setting deep properties and performing function/method calls with multiple dispatch by path too.
* Integration with system Python, QuickJS, and Lua interpreters. Currently the Python bindings are the most advanced. They have iteration, calls, containment checks, magic methods for operators, and (I think) most major program flow control tools working with Kotlin objects.
* Base classes for building REPLs with external processes' `STDIN`/`STDOUT`. Interface accepting anything that provides an input stream and an output stream.
* Simple, language-agnostic and transport-agnostic JSON-based protocol for IPC and dynamic access to properties­. You can also just eval raw code in the interpreter (which is the current status of the JS and Lua "backends"), but then there's no access to the JVM.
* Completely dynamic scripting API based on the IPC protocol and Kotlin/JVM-side reflection, at least for the Python bindings. Because the API is basically automatically generated at evaluation time by Python using IPC requests for Kotlin reflection, there is no need to update the exposed scripting API with game code changes. Any new members, properties, and other changes to the main game code will be immediately, naturally, and automatically reflected in the scripting API. Not sure how viable this would be for JS ("`Proxy()`"s and `.valueOf()`?) and Lua ("Metatables"?), but the protocol supports it at least.
* Modular design. If you wrap up a Python interpreter in the JNI/NDK/whatever, I think the existing code should basically all work on Android automatically. Likewise, as long as you implement the JSON-based IPC protocol, it doesn't matter if the programming language you're trying to support is INTERCAL or the IPC mechanism is messenger pigeons.
* Autocomplete and exception handling across script interpreter and Kotlin. …Hm. Kotlin/JVM-side exceptions during a reflective request will tag and return the error message to the script interpreter, which in Python's case raises it in Python (which, if unhandled, gets caught again, stringified, and sent back to be printed in the REPL). Python-side errors are currently caught and handled in Python, either by the running script or by the REPL (which stringifies them and sends them back to Kotlin be printed), or, if they're not caught in the Python REPL, then they raise an exception in Kotlin, but only because they look like malformed/missing IPC packets. Anyway, it should be pretty easy to add an `"Exception"` flag to the IPC packet spec if it's needed.
* Isolated codebase. GitHub tells me this PR replaces only 5 lines, and I think most of those are comments or build configuration. So I think this shouldn't be able to break anything else, or increase the maintenance load on the rest of the codebase.
* Basically adds everything needed for a fully programmable modding API. You pretty much just have to spawn a new instance of `ScriptingState()` and run its `exec()` method periodically in a separate `Thread` with mod-defined handlers— And maybe add some new bells and whistles like functions for creating event popup GUIs, and a `Map<String, Any>` for storing arbitrary mod-controlled data in save files— In order to implement events, scenarios, custom behaviour, etc... Hm. Security notwithstanding, that is.
* Docs describing the class structure, IPC protocol, and scripting API, which can be added straight to the wiki.
* Example Python scripts demonstrating player automation macros, modding-aimed scripted events, script-managed external process interface, and scripted map editor tools/map generation.
* Test suite in Python and (WIP) Kotlin.

image

---

**Impact:**

Size of `Unciv-debug.apk` as compared to master: 14.4MB instead of 12.3MB as of this writing.

---

Still a bit of a mess in several places. ~3,000 lines in a new language in about a week will do that.

Don't actually know if this is even wanted, or possibly horribly offensive to everything that is good and stable.


will-ca:
* [ ] Nothing preventing the history from going on forever.
* [ ] Code is full of empty lines with spaces.
* [ ] Brackets are behaving weirdly.... I suppose `.toLabel()` runs replacements for Uniques and such? Will replace with `Label()`, then.
* [ ] Stability of `ConsoleState` should be looked at— Direct access from `ConsoleScreen`, accessing and skipping/automatically clipping indices when out of bounds of `.consoleBackends[]`, etc.
* [ ] Not sure about the consistency of all the names.
* [ ] Styling is very austere with no margins— Not a huge deal IMO, for what it is.
* [ ] Exiting screen breaks keyboard bindings.

SomeTroglodyte:
> Brackets are behaving weirdly.... I suppose .toLabel() runs

.. tr(), yes. that's its _main_ function. Don't be surprised by crashes when you use Label() instead - an UI using only .toLabel, .toTextButton and its cousins can manage without maintaining a valid skin, with standard Gdx widgets you'll need that, just pass the singleton from CameraStageBaseScreen's companion.

> breaks keyboard bindings

I tied keyPressDispatcher to CameraStageBaseScreen, but the listener lives on the Stage. If you switch pre-existing Screen instances onto a Stage, it's unsurprising some management will be needed. Possibly it's enough to call KeyPressDispatcher's uninstall and install...

> choosing backed types

Why choose? Python is the obvious single choice :smiling_imp:  - see Civ 4

Please also check out the 'secret' page on the Options popup and what implications there may be...

> horribly offensive to everything that is good

Very likely each one of us will be deeply offended to the core of our soul, if any. That said, this definitely (IMHO) should not be accessible to the casual Google Play Store user; and the expert can always do most of what you aim for using Alt-F8 on a breakpoint or global-search on a save loaded into a good editor, or somesuch. On the other hand, many major games if not all have such a tool, so, "a million flies can't be wrong, eat --- solyent brown"?

SomeTroglodyte:
... but having Python bindings is something I have repeatedly thought about, but more in a modding context: scenarios, campaigns, events for those, e.g. tied to tile objects, scripted map generation and so forth ...

will-ca:
@SomeTroglodyte Thanks for the pointers. Saved me quite a bit of time digging around, I'm sure, especially with the keyboard thing.

I'm used to Python personally, but it's not clear to me that it'd be the best language to embed. Hard to sandbox, takes up 8MB of RAM on my machine right after launch— Much better at only 170kB for MicroPython, though. Also, not sure how to integrate with Kotlin/JVM in a cross-platform app.

I've been thinking that whatever wrappers and structures exist in the scripting environment would probably best be automatically generated anyway, using a JSON definition or mirroring the Kotlin code or some such— And that IPC with a separate interpreter process might be more flexible and easier to maintain than trying to properly "embed" anything. So then, there wouldn't be too much overhead from supporting multiple scripting languages, as you would just have to write the code generator and IPC protocol once per language, and feed them all the same API spec.

Anyone here have any ideas how to run and communicate with an interpreter written in C, in a way that works on both Android and Desktop?

I had a look at the code for the secret Options tab. I can't seem open it though. I'm only seeing the hashed version of the password stored in the code?

I mostly agree about hiding a console by default, but I also wanted this mostly from a perspective as a user. I suppose adding an opt-in and disclaimer in the Options menu would probably be enough?

Scriptability for modding is something that I have in my notes for this too, though at the periphery of its scope. I suppose the only real difference between `PlayerAPI`/`CheatAPI` and `ModAPI` would be the creation of handlers that let code be run automatically.

SomeTroglodyte:
> I can't seem open it though

(R-Ctrl or R-alt) + R-Shift + O. PW is _of course_ hash only in source, but real easy to guess, especially for the boss - or just bruteforce it. Even written in kotlin with zero optimization, single-threaded on a low-end nuc10 that came up with a different, much longer PW overnight - so ensure your bruteforcer is breadth-first :rofl:

> Thanks for the pointers

Nah. Normally I would have checked the branch out and analyzed a little, but at the moment I'm on low-burn mode for Unciv, barely reading the mail, so ideas is all you get, sorry.

core/src/com/unciv/scripting/ScriptingBackend.kt
```diff
@@ -0,0 +1,300 @@
...
+                appendOut(buildingcities.joinToString(", "))
+            }
+            "listcities" -> {
+                if (!(scriptingScope.isInGame)) { appendOut("Must be in-game for this command!"); return out }
+                appendOut(scriptingScope.civInfo!!.cities
+                    .map { city -> city.name }
+                    .joinToString(", ")
+                )
+            }
+            "cheatson" -> {
```
xlenstra, line 147: Would it be possible to disable this command in multiplayer games?

will-ca: Hm. Good point. Though I mostly mean for the built-in commands to be a demo until I or someone else cooks up a good cross-platform IPC system for a Python/Lua/JS interpreter, and I've also added another backend that gives near-full reflective access to the entire program state, so I'd prefer to not to set a precedent of implementing too much functionality like permissions in the command parsers themselves. IMO permissions and action legality should be handled on the internal API level, ideally in a way that can eventually be unified with the main game GUI.

I'll think about starting to implement a `CheatAPI` controller in the command backend scope, or maybe just disable the console screen entirely on multiplayer.

will-ca:
I think this is mostly feature-complete now. There's still a ton of details to fix or implement, and much of the code is still a bit of a mess. But the basic idea of an interactive, dynamic REPL (currently in Python) with rich read-write access to the game state is implemented, and I don't think the protocols and interfaces are going to change too much more.

will-ca:
Example code:

```python3
from unciv.api import *

def gatherBiggestCities(cities, ratio, stat='production'):
	"""Return the biggest cities from given cities that collectively comprise at least a given ratio of the total stats of all the cities. E.G: `gatherBiggestCities(civInfo.cities, 0.25, 'production')` will return a list of cities that together produce 25% of your empire's production."""
	assert 0 < ratio <= 1
	getstat = lambda c: real(getattr(c.cityStats.currentCityStats, stat))
	total = sum(getstat(c) for c in cities)
	inorder = sorted(cities, key=getstat, reverse=True)
	currsum = 0
	threshold = total*ratio
	selected = []
	for c in cities:
		if currsum >= threshold:
			break
		selected.append(c)
		currsum += getstat(c)
	assert sum(getstat(c) for c in cities) >= threshold
	print(f"Chose {len(selected)}/{len(inorder)} cities, representing {str(currsum/total*100)[:5]}% of total {total} {stat}.")
	return selected

def clearCitiesProduction(cities):
	"""Clear given cities of all queued and current production, and return the iterable of cities."""
	for i, city in enumerate(cities):
		city.cityConstructions.constructionQueue.clear()
	print(f"Cleared production from {i+1} cities.")
	return cities

def addCitiesProduction(cities, queue=()):
	"""Add given construction items to the construction queues of given cities, and return the iterable of cities."""
	for i, city in enumerate(cities):
		for build in queue:
			city.cityConstructions.addToQueue(build)
	print(f"Set {i+1} cities to build {queue}.")
	return cities
```

Save this as `democomms.py` somewhere.

Add it to `sys.path`, and import it from Unciv:

`import os, sys; sys.path.append(os.path.join(os.getcwd(), "../..")); from democomms import *`

Then use it to quickly calculate and switch your biggest cities to focus on war with a single line:

`[real(c.name) for c in addCitiesProduction(clearCitiesProduction(gatherBiggestCities(civInfo.cities, 0.5)), ('Missile Cruiser', 'Nuclear Submarine', 'Mobile SAM'))]`

image

logicminimal:
i think the console window looks a bit ugly, maybe colours and a bit of padding should be added between each paranthesis?

will-ca:
@logicminimal Ouch? Lol.

The console doesn't have any awareness about the parentheses or decimals that are printed on it. It just receives the command as a text string, sends that to the Python interpreter (or whatever backend), and then displays the response it receives from the backend, also as a text string.

So, if you want to limit the number of decimals printed, you would have to do it in the user script. In Python, for example, that could be `str(number)[:5]`. That's how the percentage is clipped, but I forgot to clip the sum in the demo script. But that's not really part of the PR, framework, protocol, API, or feature, as it's just a demo.

Maybe I could change to a monospaced font or something. Personally I don't have any trouble reading the printout, but I'll look into if that's possible and might improve legibility.

I briefly entertained adding support for `bash`-style colour-formatting codes, but it would be a lot of work both to add support for it in the display and use it in each backend, for relatively little gain. It looks like GDX does have easy support for in-line colours, but they don't follow any external standards, which means that implementing them would still be the same amount of work and also break compatibility with external script interpreters (for debug, etc) while making future maintenance harder.

For serious scripts, you should probably be editing your text in an external program with syntax highlighting anyway.

I do think the layout is a bit rough. But any added margins/padding would also decrease the number of lines that can be displayed, and I'm not sure that's worth it.

will-ca:
@logicminimal Actually, I think coloured syntax highlighting could be doable, and doable cleanly, if I add a way to define a list of REGEXs/transforms that are run at the display level instead of trying to do anything at the backend or script level. That would be separated enough to avoid a lot of increase in maintenance difficulty, and it would probably also let me reuse highlighting rules from existing open-source projects, alleviating much of the implementation effort.

I was a little frustrated at first because I viewed the GUI as a relatively minor part of this PR, but thanks for reminding me about it and getting me thinking about how to improve it, I guess.

SomeTroglodyte:
> external program with syntax highlighting

I concur. Don't overdo it. As it is, this has little chance of being merged simply because it's so BIG already.
I should probably check out, analyze and test this, but I'm on the latter part of the manic-depressive cycle as far as Unciv is concerned right now...

> looks like GDX does have easy support for in-line colours

Really? Where? I searched for such a feature when we did the spruced up Civilopedia and found nothing except dead attempts to embed browser-like renderers on a way lower level than Scene2D...

will-ca:
@SomeTroglodyte 

> I concur. Don't overdo it. As it is, this has little chance of being merged simply because it's so BIG already.

Yeah, I'm a bit worried about that. But I was making major changes to the protocol and class structure well into the thousands of lines (and still kinda am), so I'm not sure breaking it up into many smaller PRs would have been any better. I kinda needed to get it this big in order to test things enough out in practice to get a stable working design. (Either that, or spend a lot more time on planning.)

If it helps, I think this could be thought of as more like an external library than a change to Unciv itself. I think the vast majority of the code doesn't really have anything to do with the game or game state directly— It's basically just a language-agnostic IPC protocol/foreign function interface, with reference implementations that happen to be in Kotlin/Python, and helper classes that can be used to build an REPL in any Kotlin application. And of the stuff that is tied directly to Unciv, a lot of the LoC are just repeated implementations of relatively few and simple interfaces/ABCs.

I'm going to gradually stop adding new features now/soon™, and start thinking about cleaning it up and maybe breaking it up into many smaller PRs now that it's mostly feature-complete.

> I should probably check out, analyze and test this, but I'm on the latter part of the manic-depressive cycle as far as Unciv is concerned right now...

Ehh, I'd definitely appreciate tests and comments on the general design/architecture/behaviour, but do be aware the code still needs a LOT of cleanup before I'd consider it worth your time to carefully analyse. It's full of potential problems I didn't fix when writing because I wasn't sure if I'd end up replacing them completely, idioms in need of rewrite, features/cleanup I started but haven't yet finished, etc. Just the other day, I found out about the existence of `.withIndex()`, `.indices`, and `.lastIndex`, so that's one more thing I need to go back and clean up instead of doing `0..Array.size-1`.

> Really? Where? I searched for such a feature when we did the spruced up Civilopedia and found nothing except dead attempts to embed browser-like renderers on a way lower level than Scene2D...

https://github.com/libgdx/libgdx/wiki/Color-Markup-Language

The search term I used was `"gdx inline formatting"`, which brought me to this StackOverflow answer:

https://stackoverflow.com/questions/33465564/libgdx-multi-color-text-in-one-label

I haven't tried it yet, but I *think* this is the same GDX, and it looks easy enough.

For this console, I'm thinking I can just have a `.JSON` file associating each language with a big stack of REGEXs stolen from KWrite/GTKSourceView (depending on license) in a later PR.

StephenHogg:
Apropos of nothing, this would allow AI people like me to use Civ V as a testbed for designing game-playing AI. Please don't let this drop - even if it winds up being a separate package instead of merged here, it's potentially very, very valuable. Let me know if you'd like to chat more about this :)

will-ca:
@StephenHogg Yep. That was a thought in my original feature request as well. I was thinking of trying to make a new Civ V AI before I discovered Unciv, for which the first step would have basically been to program Unciv myself but as a server only, before I got fed up enough with late-game clicking around and pivoted to user-scripted automation instead.

Although, I'm personally too jealous (and terrified of democratic backsliding and such) to be a fan of ongoing AI research in human civilization, so that might actually be a deterrent against finishing this or freely licensing it lmao.

Anyway, the latest commit starts adding in an example for making a user script interact with the main Unciv process's STDIN and STDOUT, and I added a line in my notes about command-line options to launch Unciv directly into a script interpreter, which I figure may be helpful for that kind of automation-heavy research-oriented work.

will-ca:
This is going to be well over 5,000 LoC by the time it's ready for review.

But somewhere over a 1,000 of those will be code comments/KDocs, a lot more yet will be example Python scripts and reference backend implementations that nothing else relies on, and I'm trying my best to clean, document, and organize it all as best I can before I take it out of draft status— And the systems are very separate from the core game code, so there's not much concern about breaking anything else— So hopefully it won't be too painful/unappealing to review.

(`ScriptingProtocol.kt` is 400 lines, but the majority of that is just different packet types. `ScriptingBackend.kt` (pre-split) is 600 lines, but the actual interfaces/ABCs are only a couple dozen, with most of the rest being test/demo backend implementations.)

And IMO it brings a lot of potential functionality. By integrating dynamic languages, most of the groundwork is laid for everything from a scriptable modding API to rapidly prototyped map editor tools and even machine learning research, all doable by the user without any core code/developer work (hence the extensive WIP demo scripts, which include some form of example for all of those).

will-ca:
> [….] scripted map generation [….]

@SomeTroglodyte 

Generated based on a heightmap from disk/assets, using only 30 lines of example Python:

image

Scales automatically to map size:

image

Original heightmap:

EarthTopology

Generated Mandelbrot Set:

image

SomeTroglodyte:
Still not in the right mindset to fully appreciate this (meaning pulling the branch and dig in), my brain is infested with distractions like - can I do that 100km roundtrip (97% offroad and >800m denivellation) on one ebike battery... But from what I see here this is  still the jaw-dropper-WOW vs is-it-worth-it battle. Makes me wonder whether I should foist my unfinished work on you to bring up to par...

But you seem to be using map generator fragments in a strange way - uranium on ocean or rivers not ending on water wouldn't be possible with current master branch. Not meant as criticism, of course all this is demo level.

will-ca:
@SomeTroglodyte Ah, I'm not using map generator fragments at all. The structures/interface I've created are all rather "low level"— Meaning raw read-write access to Kotlin/JVM instances, without any particularly specialized rules or frameworks for how they're used in the context of Unciv— So currently the demo works by just mutating `.baseTerrain` and `.terrainFeatures` on an existing `TileMap`, hence the resources and rivers from the original, "real" map are still left over.

In fact, there are no limits to what the script can do. The scripts' functions accept/are oblivious to using `TerrainFeature`s as `.baseTerrain`, base terrains can be stacked on top of each other, and I guess you could even have one tile with every Natural Wonder stacked on top of it, as long as it's permitted in the Kotlin/JVM API.

That's not to say that it couldn't also make use of the existing map generation stuff though. I think that would probably be just a question of making sure they're exposed from the Kotlin side, and, in the case of having scripted maps that integrate with the "New Game"/"New Map" menus, letting mods register a handler that gets run when it's needed. And if the Kotlin code throws an exception when being set to an illegal value or some such, it's propagated and is currently able to be handled from the script too, so restrictions on permissible actions can be added on the Kotlin side (possibly through a separate "ModdingApi" utility function) if needed.

Good luck with your trip! Although, having spent much of last year on an e-bike, maybe bring a spare battery, or pack a charger or something...

will-ca:
More scripted map generation/editing, this time by matching pixel colours to the closest values in a dictionary of terrain types:

```python3
loadImageColours(imagepath="StarryNight.jpg")
```

image

StarryNight


The terrain colour matching space is by default automatically generated by analyzing the FantasyHex tileset. Custom mappings of colours per terrain type can also be supplied. This lets a colour-coded image be used to directly define biomes/terrains in an Unciv map:

```python3
loadImageColours(imagepath="EarthTerrainFantasyHex.png")
```

image

image


Adding dithering support was only an additional 8 lines:

```python3
loadImageColours(imagepath="TurboRainbow.png", maxdither=0.4)
```

image

image


Something like this would probably be the way to to ship custom maps that should always look the same, but can scale to different sizes. Draw a colour-coded PNG by terrain type, define a JSON of what terrain each colour should have, and then sample from the image at the projected points of each tile.

SomeTroglodyte:
> memory leak

PixMap was that one class where the caller is responsible for cleanup, wasn't it? Unciv has no probs ATM with that as they get used directly only for Fonts, and that's one-time.

will-ca:
Test suite implemented in Python. Should be fairly straightforwards to integrate into Kotlin build tests, once script-to-Kotlin exception propagation is implemented (which itself should also be fairly straightforwards, and is needed for modding API use):

```python3
>>> from unciv_scripting_examples.Tests import *
>>> TestRunner.run_tests()
```

```
Python test PASSED: NewGameTest
Python test PASSED: gatherBiggestCities
Python test PASSED: clearCitiesProduction
Python test PASSED: addCitiesProduction
Python test PASSED: clearCitiesSpecialists
Python test PASSED: focusCitiesFood
Python test FAILED: buildCitiesQueue
	NotImplementedError()
Python test FAILED: rebaseUnitsEvenly
	NotImplementedError()
Python test FAILED: spreadResources-InGame
	NotImplementedError()
Python test FAILED: spreadResources-InMapEditor
	NotImplementedError()
Python test FAILED: dilateTileTypes-InGame
	NotImplementedError()
Python test FAILED: dilateTileTypes-InMapEditor
	NotImplementedError()
Python test FAILED: erodeTileType-InGame
	NotImplementedError()
Python test FAILED: erodeTileType-InMapEditor
	NotImplementedError()
Python test FAILED: floodFillSelected-InGame
	NotImplementedError()
Python test FAILED: floodFillSelected-InMapEditor
	NotImplementedError()
Python test PASSED: makeMandelbrot-InGame
Python test PASSED: makeMandelbrot-InMapEditor
Python test PASSED: graph2D-InGame
Python test PASSED: graph2D-InMapEditor
Python test PASSED: graph3D-InGame
Python test PASSED: graph3D-InMapEditor
Python test PASSED: loadImageHeightmap-InGame
Python test PASSED: loadImageHeightmap-InMapEditor
Python test PASSED: loadImageColours-InGame
Python test PASSED: loadImageColours-InMapEditor
Python test FAILED: extendTechTree
	NotImplementedError()
Python test PASSED: clearTechTree
Python test FAILED: scrambleTechTree
	NotImplementedError()


12 Python tests FAILED: ['buildCitiesQueue', 'rebaseUnitsEvenly', 'spreadResources-InGame', 'spreadResources-InMapEditor', 'dilateTileTypes-InGame', 'dilateTileTypes-InMapEditor', 'erodeTileType-InGame', 'erodeTileType-InMapEditor', 'floodFillSelected-InGame', 'floodFillSelected-InMapEditor', 'extendTechTree', 'scrambleTechTree']
```

image

Kotlin-side unit testing for `InstanceMethodDispatcher` is also planned. The tests are kinda already written, but they're in a scrappy `.kts` file so I have to rename everything, put them into methods, etc.

core/src/com/unciv/scripting/ScriptingBackend.kt
will-ca: @xlenstra Right, finally did this as below:

https://github.com/yairm210/Unciv/blob/f1abea3dfe41c5c1227759347ac83fe6f15af84d/core/src/com/unciv/scripting/ScriptingState.kt#L129-L133

This is at the entrypoint for every script/command that can/should get run, so it should blanket block everything with a nice notification. A further check will have to prevent the game from trying to automatically run/load mod script handlers (which will fail), once those are implemented. I'm not worried about pass-and-play, where there's easier ways both to cheat and be caught cheating anyway.

However, can you confirm that `UncivGame.Current.gameInfo.gameParameters.isOnlineMultiplayer` is the right check to make here?

will-ca:
> ... but having Python bindings is something I have repeatedly thought about, but more in a modding context: scenarios, campaigns, events for those, e.g. tied to tile objects, scripted map generation and so forth ...

@SomeTroglodyte 

Did anybody say EvEnt [PopUpS]???

**Basic popup from Python:**

https://github.com/yairm210/Unciv/blob/fa85106025549689e991d9c99a8c299918a085bb/android/assets/scripting/enginefiles/python/unciv_scripting_examples/EventPopup.py#L14-L31

image

**Paradox-style/Beyond Earth-style story-driven event popup:**

https://github.com/yairm210/Unciv/blob/fa85106025549689e991d9c99a8c299918a085bb/android/assets/scripting/enginefiles/python/unciv_scripting_examples/EventPopup.py#L48-L147
image
image


All of the options are fully functional. The resistance turns get added to the already existing number per city at the instant when the popup is spawned.

The action taken when selecting a button is specified as a JVM-space lambda function. In the example above, I used Kotlin-space helper functions from Python to generate lambdas that run entirely in the JVM, but there's also a helper function and usage example for creating callbacks which themselves also run in the scripting backend.

On desktop, the first time showing the popup takes maybe half a second, but all subsequent runs feel pretty instant, which is true of most scripted actions. Maybe the JVM's JIT'ing, maybe the game data's initializing, or maybe it's the caching that I implemented in some places (though I don't think those should have near that impact yet).

Because the scripting bindings aim mostly to expose existing functionality from the Kotlin code, scripts will have to provide their own implementations of the event popup layout using the GDX widgets. If the Unciv core gets universal branching events at some point, then that will also be usable from scripts. But I'm not planning on providing any kind of Python/JS standard Unciv library or dependency resolution between mods, although modders can always copy and paste code from the examples.

For a yet more extreme demonstration of dynamically injected GUIs with full scripting integration, I was also thinking about doing an example script/reference mod for a Tic Tac Toe minigame where the winner gets what they want for free from a balanced trade proposal. Should be entirely possible to do (without hacks or Kotlin changes) using the current API status— Only normal Python code needed.

As with previously, the example code serves at least triple purposes:
1. To inform the API design. A lot of the scripting API tools that make it possible/convenient, and a number of small bugfixes, didn't exist or didn't have a clear design until I had a practical test of what they could be used for.
2. As a stock example script. When/If this gets merged, I imagine that having examples that modders can look at to see how to do common or varied tasks would be helpful for adoption.
3. As a unit test. Running a diverse sampling of practical code is one way to catch bugs and breaking changes at any point in the rather dynamic Kotlin-Reflection-IPC-Binding loop, and the Python tests are designed with that in mind. So in theory, every example script also stabilize future versions of the API.

will-ca:
By default, opening the scripting console now displays a warning popup message:

image

The warning can be disabled in the new Scripting tab in Options, where a similar warning is also on prominent and permanent display:

image

> That said, this definitely (IMHO) should not be accessible to the casual Google Play Store user;

Concern with regards to users on Android is also alleviated by the fact that I think most mobile users probably literally *can't* "Press <kbd>**~**</kbd> to activate" with their devices. It has been a deliberate choice to restrict ways of opening the scripting console to a keyboard button that is only usually available on Desktop, and not add any on-screen buttons for it. Maybe there could be a downloadable mod that uses scripting handlers to inject a button into the main screens in case anyone *really* wants to use the console on Android.

The opt-in to let mods run scripts (currently half implemented but not hooked up to anything) also display a similar warning:

image
 
Selecting "NO" resets both the settings flag and the button.
An additional opt-in tick box is planned on a per-mod basis in the mod manager page, which will also show a confirmation. The `GameSettings` property will show the ModManagementScreen checkbox, and the ModManagementScreen checkbox (`&&` the GameSettings property) will allow mods to register and run scripted handlers.

So all told:
- In order to get at the scripting console, 2 explicitly affirmative user actions will be needed: Enable in OptionsPopup button, accept ConsoleScreen warning.
- In order to start letting mods run scripts, 4 explicitly affirmative user actions will be needed: Click OptionsPopup button, accept options warning, tick mod checkbox, accept mod warning.

I know warning fatigue is a thing, but how else are you going to deal with the fundamentally unsafe operation of running untrusted code? The warnings also create friction mainly at the start, to make sure the user has a chance to understand and make an informed choice about what they're doing: Disabling the warning lets the ConsoleScreen be used immediately, and enabling mod scripting is a one-time action per game install in the options and a one-time action per mod install in the mod manager.

---

The "Scripting" tab in OptionsPopup also exposes a chart for setting startup scripts for different backends in ConsoleScreen:

image

Users can use this to E.G.: Set their `PYTHONPATH`, load custom automation macros/UI customization injections/research projects, turn cheats on automatically for the backends that have such a flag, and so on.

Sensible defaults are defined for each available backend type. Clicking the <kbd>**X**</kbd> resets the command's startup script to the default. Applies only to the console; Mods will have to do their own imports.

---

Most of the ways that modded scripts can fail also produce a popup with a highly legible explanation plus smaller and scrollable error output/stack trace— The other side of "How do I run untrusted code safely" being "How do I prevent the user from blaming the core devs when stuff inevitably breaks because of unverified code?"

 image

 
image

will-ca:
And that brings me to the other prematurely implemented feature: Mod handlers are now supported.

Currently I'm reading from a handful of hard-coded JSON "mods" for test purposes, but hopefully it will be fairly straightforwards to just read from actual files from downloaded mods instead. The format is simple: Each mod gets to specify an engine/language type to use, plus lists of background and foreground commands to run for each handler:

https://github.com/yairm210/Unciv/blob/3ec2701f64a2b2b702ffa7eeb0c5c438636088f2/core/src/com/unciv/models/modscripting/ModScriptingRegistrationManager.kt#L11-L47

(Background for most stuff. Foreground if touching UI elements, due to OpenGL context availability.)

The handler types are defined from Kotlin (for now) using the "builders" thing that I finally decided to figure out now after SomeTroglodyte pointed me at it some weeks ago:

https://github.com/yairm210/Unciv/blob/3ec2701f64a2b2b702ffa7eeb0c5c438636088f2/core/src/com/unciv/models/modscripting/ModScriptingHandlerTypes.kt#L22-L47

And likewise, they're triggered at the point where they're called by using the names/IDs passed to their builders:

https://github.com/yairm210/Unciv/blob/3ec2701f64a2b2b702ffa7eeb0c5c438636088f2/core/src/com/unciv/ui/consolescreen/ConsoleScreen.kt#L199-L200

(The only downside is that due to the dynamic nature of the builders, the arguments passed to `runHandler()`— Which specify the handler type to run and an object that gets used by the handler type definition to generate parameters which are exposed to the scripts— Aren't statically checked. The handler definitions themselves are type-checked, but uses of them aren't, so they could result in runtime exceptions. But the use of `enum`s for the handler type selection, and run-time checking of the base parameter type, should make any typos there easy to find— It wouldn't be any trickier to flesh out and maintain than the Uniques system, I'd say.)

The way the scripting system has evolved means that any script interpreters belonging to mods are treated similarly to those launched by the player. Their commands, output, results, and errors are shown in the console screen, and they can be selected and have their internal state inspected or modified just like any manually launched interpreter:

image

I've greyed out their "Terminate" buttons, and added text communicating what mods they come from, but other than that they're just like any backend. Useful for debugging and developing mod scripts, I imagine.

Also WIP: Dokka integration, to expose the scripting API to modders without needing them to dig through the Kotlin in AS themselves:

https://github.com/will-ca/Unciv/tree/dokka/

https://will-ca.github.io/Unciv/dokka/

will-ca:
The build tests for this PR are going to start failing now, because I've added tests that launch a full running instance of the game (requiring X11, OpenGL, Python, and so on).

Since this scripting API by nature tries to expose as much of the game state as possible, this was what I decided makes the most sense to test it. I've also added and still have planned some more tests for certain critical subsystems, but my biggest concern about future stability is the sum behaviour of all the stuff around the IPC protocol. The packet types aren't checked by the IDE, and the bindings and operator overloading in Python/JS *definitely* aren't checked against against the Kotlin code that they interact with, so the only way to know whether they work is to try to use them. Hardcoded tests would be restricted to covering only very specific edge cases in unrealistic circumstances; Actually running example scripts can hit a much larger area of the scripting API, in all the dynamicism of the scripting languages and IPC protocol, but it requires launching the full application to do that.

Fortunately, it doesn't seem like making the Github workflow set up a software-emulated X11 and OpenGL environment will be too hard. E.G:
- https://github.com/konserw/mre/blob/master/.github/workflows/linux.yml
- https://github.com/marketplace/actions/setup-opengl

I suppose I would do that in my branch, and then `cherry-pick` it into a separate PR. I'd probably try to set it up as a separate action or something, so the normal tests and other existing automations wouldn't be broken by X11 or OpenGL failures.

@yairm210 Thoughts?

SomeTroglodyte:
> Did anybody say EvEnt [PopUpS]???

ROFL - all the world says "reins" in that context yet your spelling is theoretically correct - you're insulting the MacDo majority!

SomeTroglodyte:
Holy you-know-what. I would fear for my soul if I were able to gain that much insight into these technologies. You sure you're not an avatar of QC's Yay?

will-ca:
> Holy you-know-what. I would fear for my soul if I were able to gain that much insight into these technologies. You sure you're not an avatar of QC's Yay?

Careful now. My self-esteem doesn't know what to do with this.

Is Yay like Schlock's Petey/Chinook?

SomeTroglodyte:
> doesn't know what to do

Precisely!

That [Shlock](https://www.schlockmercenary.com/) webcomic is new to me. But, according to some tropes site yes, maybe. Links: [QC](https://questionablecontent.net/) [Yay](https://questionablecontent.fandom.com/wiki/Yay_Newfriend)

PR #5613, opened 2021-10-30
Made rich presence text not change with language

logicminimal:
made the game not change the text shown in rich presence with the current language, added a language override functionality to String.tr() which is probably inefficient as hell, also the reason why i had to keep the .tr in leader name is because it showed the parantheses

also, i dont know why git randomly removes that onclick setscreen to civ civilopedia article line and that italian.properties line...


logicminimal:
about italian.properties file: git says that the contents are identical what the hell is going on

yairm210:
I'm a bit confused - what problem are you trying to solve here?

yairm210:
If it's just to display all text in English, then ignore the translation function entirely, and display text without using the function at all

yairm210:
Take what you want to display, which is in English anyway, and just set the text

logicminimal:
> 
> 
> If it's just to display all text in English, then ignore the translation function entirely, and display text without using the function at all

i was about to do that, but if i remove the tr it shows like [maria theresa] of [austria] which we dont want

yairm210:
What I'm saying is you can take the leader name directly from the Civ name :)
Like
"${civInfo.nation.leaderName} of ${civInfo.nation.name}"
You don't need those []'s if you aren't translating

logicminimal:
> 
> 
> What I'm saying is you can take the leader name directly from the Civ name :) Like "${civInfo.nation.leaderName} of ${civInfo.nation.name}" You don't need those []'s if you aren't translating

well, ill fix it according to that tomorrow, also, i guess we dont need a language override for .tr()?

yairm210:
Nope, no file needs to be changed apart from the specific function.
But I'll poll this on Discord to see if it's even a good idea.

logicminimal:
> 
> 
> Nope, no file needs to be changed apart from the specific function. But I'll poll this on Discord to see if it's even a good idea.

so... are we going with add parameter to .tr() way or with  "${civInfo.nation.leaderName} of ${civInfo.nation.name}" ?

xlenstra:
> > Nope, no file needs to be changed apart from the specific function. But I'll poll this on Discord to see if it's even a good idea.
> 
> so... are we going with add parameter to .tr() way or with "${civInfo.nation.leaderName} of ${civInfo.nation.name}" ?

The latter is way simpler and for the first there will be almost no other usecases, so the latter

logicminimal:
done.. i guess...

PR #5618, opened 2021-10-31
Reworked buying buildings & units with stats a bit

xlenstra:
Reworked the system for buying units & buildings a bit, unifying uniques for them, enumifying them, adding support for conditionals and fixing a few bugs.

Bugs fixed:
- Faith cost for buying faith buildings should increase based on the era you're in
- Trying to buy a great artist before the industrial era crashes the game

Uniques now supported:
- May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] at an increasing price ([amount])
- May buy [baseUnitFilter] units for [amount] [stat] [cityFilter]
- May buy [baseUnitFilter] units with [stat] [cityFilter]
- May buy [baseUnitFilter] units with [stat] for [amount] times their normal Production cost
- All of the above, but for buildings instead of units.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -150,12 +150,20 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
     FreeExtraBeliefs("May choose [amount] additional [beliefType] beliefs when [foundingOrEnhancing] a religion", UniqueTarget.Global),
     FreeExtraAnyBeliefs("May choose [amount] additional belief(s) of any type when [foundingOrEnhancing] a religion", UniqueTarget.Global),
 
+    // There is potential to merge these
     BuyUnitsIncreasingCost("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] at an increasing price ([amount])", UniqueTarget.Global),
     BuyBuildingsIncreasingCost("May buy [buildingFilter] buildings for [amount] [stat] [cityFilter] at an increasing price ([amount])", UniqueTarget.Global),
-    @Deprecated("As of 3.17.9", ReplaceWith ("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] at an increasing price ([amount]) <starting from the [era]>"))
-    BuyUnitsIncreasingCostEra("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] starting from the [era] at an increasing price ([amount])", UniqueTarget.Global),    
+    BuyUnitsForAmountStat("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -150,12 +150,20 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
...
     FreeExtraAnyBeliefs("May choose [amount] additional belief(s) of any type when [foundingOrEnhancing] a religion", UniqueTarget.Global),
 
+    // There is potential to merge these
     BuyUnitsIncreasingCost("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] at an increasing price ([amount])", UniqueTarget.Global),
     BuyBuildingsIncreasingCost("May buy [buildingFilter] buildings for [amount] [stat] [cityFilter] at an increasing price ([amount])", UniqueTarget.Global),
-    @Deprecated("As of 3.17.9", ReplaceWith ("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] at an increasing price ([amount]) <starting from the [era]>"))
-    BuyUnitsIncreasingCostEra("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter] starting from the [era] at an increasing price ([amount])", UniqueTarget.Global),    
+    BuyUnitsForAmountStat("May buy [baseUnitFilter] units for [amount] [stat] [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
+    BuyBuildingsForAmountStat("May buy [baseUnitFilter] buildings for [amount] [stat] [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
```
yairm210, line 157: Buildings and unit filter?

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -250,48 +250,51 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
...
-                && it.params[1] == stat.name
-            }
-        ) return true
-
-        return super.canBePurchasedWithStat(cityInfo, stat)
-    }
-
-    private fun getCostForConstructionsIncreasingInPrice(baseCost: Int, increaseCost: Int, previouslyBought: Int): Int {
-        return (baseCost + increaseCost / 2f * ( previouslyBought * previouslyBought + previouslyBought )).toInt()        
+        return (
```
yairm210, line 256: Equality comparisons before filter comparisons, for faster false results

yairm210:
Looks OK
I tried to think of a way to reduce the number of uniques used but could not do so, mayhaps future generations will be more effective

xlenstra:
> Looks OK I tried to think of a way to reduce the number of uniques used but could not do so, mayhaps future generations will be more effective

The only idea I have for this would be one I've stolen from @avdstaaij: nestable unique parameters. Then you could unify the "[unitFilter] units" and "[buildingFilter] buildings" into a single parameter, halving the amount of uniques.

PR #5623, opened 2021-11-01
"Never appears as a Barbarian unit" for Unique Units

SimonCeder:
Adds `"Never appears as a Barbarian unit"` to Unique Units replacing such units. Purely cosmetic as Barbarians will never use Unique units anyways, but to get rid of this nonsense:
Screenshot_20211101_205949


SomeTroglodyte:
Actually, sounds more like we should graft that "internal" or "don't display" flag onto the UniqueType enum after all? Would solve the "nonsense" more cleanly and there's others that don't help much in Civilopedia...

yairm210:
WOW
Flags for unique types that determine UI functionality
GENIUS

PR #5625, opened 2021-11-02
Unique flags

SimonCeder:
Building on discussions in #5556 and #5623.
Adds a "flags" value to UniqueType and implements one such, HideInCivilopedia. For now I've set that flag for "never appears as a barbarian unit" as well as some of the more map-generation technical details of terrains.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -55,7 +55,11 @@ enum class UniqueTarget(val inheritsFrom:UniqueTarget?=null) {
     }
 }
 
-enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
+enum class UniqueFlag {
+    HideInCivilopedia,
+}
+
+enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags: List<UniqueFlag> = emptyList()) {
 
     //////////////////////////////////////// GLOBAL UNIQUES ////////////////////////////////////////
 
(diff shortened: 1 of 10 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -55,7 +55,11 @@ enum class UniqueTarget(val inheritsFrom:UniqueTarget?=null) {
     }
 }
 
-enum class UniqueType(val text:String, vararg targets: UniqueTarget) {
+enum class UniqueFlag {
+    HideInCivilopedia,
+}
+
+enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags: List<UniqueFlag> = emptyList()) {
```
SomeTroglodyte, line 62: List? I had a discrete Boolean in mind, but if we want to be extensible, this is what EnumSet was created for, even if it is more legacy Java than idiomatic Kotlin... As always, just food for thought.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -133,8 +134,10 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
             textList += FormattedLine(replacementTextForUniques)
         } else if (uniques.isNotEmpty()) {
             textList += FormattedLine()
-            for (uniqueObject in uniqueObjects.sortedBy { it.text })
-                textList += FormattedLine(uniqueObject)
+            uniqueObjects.sortedBy { it.text }.forEach {
```
SomeTroglodyte, line 137: Makes you think - why in two handfuls of repeated 'pedia code displaying uniques is this the only repetition that sorts, and then by the english version? Most are intentionally in json order, so json has some control over the looks. I really can't remember why. Were these uniques long ago a set thus not having stable order??? Might warrant a little test / code merge into some superclass method..

PR #5629, opened 2021-11-02
Quest fixes

### LINKED ISSUE #5583: City states should send notifications when ending quests
In the original, city-states also showed notifications when ending quests, either by having them fulfilled by other players, cancelling them because they are obsolete or it being a global contest that is won by another player. These notifications aren't in unciv as of yet, resulting in quests seemingly disappearing randomly.

SimonCeder:
Fixes #5583 
Displays a notification when a quest has ended due to expiring or becoming invalid.
Also fixes the strings for the contest quests so that the current leader texts are translatable.


android/assets/jsons/translations/template.properties
```diff
@@ -631,6 +631,8 @@ By expending your [unit] you gained [Stats]! =
 Clearing a [forest] has created [amount] Production for [cityName] = 
 [civName] assigned you a new quest: [questName]. = 
 [civName] rewarded you with [influence] influence for completing the [questName] quest. = 
+[civName] no longer needs your help with the [questName] quest. = 
+The [questName] quest for [civName] has ended. It was won by =
```
yairm210, line 635: "It was won by [civName]", surely

yairm210: Or civNames since it can be multiple

SimonCeder:
Better placeholder string, works and translates fine.

Issue #4430, opened 2021-07-08
Keyboard controls only work with english layout


sockybob:

SomeTroglodyte:
I'd say they work as described. They can definitely be operated with french or german layouts, so "only english" isn't strictly true. I don't even know how you would generate an ascii "C" with a cyrillic layout, but the libGdx functions used and their documentation say that's what I'm looking for when a settler is selected to found a city. We're not looking for the fourth key to the right of the left 'shift', where many western layouts have the "C" - which would also be possible. If you can show otherwise I'd be highly interested.

That the chosen letters are not adaptable to language, culture, user preference or the like is by design for the time being and an entirely different matter.

sockybob:
> I'd say they work as described. They can definitely be operated with french or german layouts, so "only english" isn't strictly true. I don't even know how you would generate an ascii "C" with a cyrillic layout, but the libGdx functions used and their documentation say that's what I'm looking for when a settler is selected to found a city. We're not looking for the fourth key to the right of the left 'shift', where many western layouts have the "C" - which would also be possible. If you can show otherwise I'd be highly interested.
> 
> That the chosen letters are not adaptable to language, culture, user preference or the like is by design for the time being and an entirely different matter.

I never encounter such weird behaviour in other games. All games i played so far are not dependent of which layout i'm currently switched to, so i assumed that the Unciv gets input from keyboard in a wrong way, like reading letters i typing in and doing certain action if it's equal to some latin letter (like if i were typing in a text field). Imo it's completely wrong approach and should be changed to process keycodes instead. Also i don't quite get what you mean but i will try to answer anyway. When i want to press button with certain latin letter i would press the same button as i were on english layout, so it's the same key with the same keycode, but the very fact that i'm on cyrillic layout makes the game ignore keyboard input until i switch to english. I'm not asking to make the game also understand those cyrillic letters which are similar to the latin ones.

yairm210:
Some small experimenting lead to the fact that Input.Keys has a .toString() function and that if you replace line 33 in KeyPressDispatcher with `       if (event == null) '_' else Input.Keys.toString(event.keyCode).first().toLowerCase(),` all single-letter keys work on all layouts, but all non-letter keys (e.g. the esc key) don't, so it still needs some refining.

sockybob:
Just whenever you process input from keyboard compare keycode integers instead of comparing strings/chars.

yairm210:
[...]

PR #5633, opened 2021-11-03
Fix Spacebar key binding in LWJGL3; Switch to integer/named key.

### LINKED ISSUE #5632: Spacebar broken for Next Turn.
**Platform**
Linux

**Version**
Git Master

**Describe the bug**
Spacebar binding for "Next Turn"'s broken in `WorldScreen`.

**Additional context**
Presumably due to #5614.

PR incoming, probably.

will-ca:
Fix #5632.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -565,7 +565,7 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Cam
         nextTurnButton.labelCell.pad(10f)
         val nextTurnActionWrapped = { nextTurnAction() }
         nextTurnButton.onClick(nextTurnActionWrapped)
-        keyPressDispatcher[' '] = nextTurnActionWrapped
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -565,7 +565,7 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Cam
         nextTurnButton.labelCell.pad(10f)
         val nextTurnActionWrapped = { nextTurnAction() }
         nextTurnButton.onClick(nextTurnActionWrapped)
-        keyPressDispatcher[' '] = nextTurnActionWrapped
+        keyPressDispatcher[Input.Keys.SPACE] = nextTurnActionWrapped
```
SomeTroglodyte, line 568: KeyCharAndCode has a prefab for that, also often saves an import

SomeTroglodyte:
OK, now I've seen the code live I suspect this regresses #4542, which was made for issue #4430 involving cyrillic keyboards. Anyone (@sockybob?) can confirm this is still OK in the situations described in #4430?

PR #5638, opened 2021-11-05
Update Ukrainian.properties

Marerjh:


android/assets/jsons/translations/Ukrainian.properties
```diff
@@ -2,8 +2,7 @@
 
 # Equivalent of a space in your language
 # If your language doesn't use spaces, just add "" as a translation, otherwise " "
- # Requires translation!
-" " = 
+" " = " "
```
xlenstra, line 5: There is a trailing space at the end of this line, which breaks the tests

PR #5640, opened 2021-11-06
can remove fallout on oases

SimonCeder:
An unintended side effect of blocking improvements on oases.


core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -87,6 +87,7 @@ class TileImprovement : RulesetStatsObject() {
             "All" -> true
             "All Road" -> isRoad()
             "Great Improvement", "Great" -> isGreatImprovement()
+            "Remove" -> name.startsWith("Remove ")
```
yairm210, line 90: I'm not sure we need this, did this serve some other purpose?

PR #5647, opened 2021-11-08
Second version of G&K / vanilla split

xlenstra:
This is an updated version of my previous attempt to split the G&K and vanilla rulesets (#5577).
It is ready to be reviewed (and quite large, sorry for that), but not yet to be merged as I'd first want to test it more thoroughly.

It fully implements a way to choose between G&K and vanilla rulesets, as well as other rulesets downloaded as mods by the user. This change can be made when starting a new game, when opening the map editor and when changing rulesets of an existing map in the map editor. Old savegames automatically use the G&K ruleset, as that is the closest to what they currently have.

<details> <summary> Some pictures for good measure: </summary>

image

image

image

image

image

</details>


yairm210:
Let's go :)

xlenstra:
I'm currently still testing & solving bugs, but I'll put these in a different PR

yairm210:
Perhaps there should be a function gameParameters.modsWithBaseRuleset() since you seem to create hashsets for that combo several times

PR #5649, opened 2021-11-08
Removed the Hanse, a wrongly implemented BNW-only building

### LINKED ISSUE #5639: Should not German Hanse be removed?
As it was introduced in BNW.

xlenstra:
Hanse is a BNW-only building that should provide production boosts only when connecting to city-states via trade routes made by caravans or cargo ships. However, due to those not being in unciv, it was implemented as providing the bonus for any city connection. As city connections are made over water via harbors, this means that in larger games Germany could reliably get +50% production or more in all their cities, which is obviously overpowered.

Fixes #5639.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -709,19 +709,19 @@
 		"requiredBuilding": "Market",
 		"requiredTech": "Banking"
 	},
-	{
-		"name": "Hanse",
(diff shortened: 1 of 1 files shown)


yairm210:
Confirmed - code in GameInfo.setTransients() should deal with existing buildings, in-progress buildings, and accumulated progress

yairm210:
ALSO we should remove the associated unique. That's a unique that should have been generalized, but since it wasn't, we can just remove it now completely.

PR #5660, opened 2021-11-09
Update Russian.properties

AlimovAR:


xlenstra:
Placeholder `[param]` not found in `Russian` for entry `We have found survivors in the ruins! Population added to [param].`
    Placeholder `[param]` not found in `Russian` for entry `We have found a stash of [param] Gold in the ruins!`

PR #5663, opened 2021-11-10
Regions part 2 - City state placements, start normalization

SimonCeder:
- Implements Region based city state placements. City states are divided between regions and "uninhabited" continents so there's a guarantee of somewhat fair distribution. Poor regions can get an extra city state as compensation. City states are attempted to be placed at the edges of regions and prefer coastal tiles. There is a minimum distance to major civ starts to avoid interfering too much with players.
- Redoes start normalization for both major civs and city states. This replaces the old "consolation prize" function. The new normalization has more tricks at its disposal including adding hills and changing around grasslands and plains to ensure enough food and production in the very early game. Also does not interfere with pre-made maps.
- Lays a bunch of groundwork for luxury and strategic resource placement. I've left the actual placements for another PR to avoid too huge code dumps at once.

**Question:**
In Civ V, a city state that could not be placed in an acceptable location (can settle turn 1, not too close to major civ) was simply dropped. We currently force the requested number of city states in anyway. Which way should we go on this?

<details><summary>Images</summary>

Screenshot_20211110_210309

Screenshot_20211110_210827
City states nicely spread across the map, without cramping major civs too much.

The old algorithm:
Screenshot_20211110_214116-1
The previous algorithm would sometimes create clusters of city states.

</details>


yairm210:
I think place them anyway, if that's an option.
If the player asked for n city states, he'll get n city states.

SimonCeder:
Sure, that's the behavior in the PR as it stands. They get discarded at map generation, but GameStarter will fill the list of city states back up and place them with its own placement algorithm.

logicminimal:
look great but is it deterministic?

SimonCeder:
The assignment of city states to each region vs uninhabited lands is mostly deterministic. The actual start position within those parameters is random.
The amount of bonuses a start position gets is deterministic, where the bonuses end up is random.

PR #5671, opened 2021-11-14
Fixed a crash when changing the base ruleset while in portrait mode

xlenstra:


logicminimal:
i didnt understand why this even happens, care to explain please?

xlenstra:
The `GameOptionsTable` class had an component called `modCheckboxes`, which was only ever initialized if the game was NOT in portrait mode. This was as the entire Start Game Screen (of which the `GameOptionsTable` is a part) is drawn completely differently when in portrait mode, and the `modCheckboxes` was later added on as a different component. 
However, when clicking any of the mod checkboxes, the game should test whether this extension mod is compatible with the currently selected mods. For this it needs access to the base ruleset. Before the G&K-vanilla split, the base ruleset was itself a mod, or, if no base ruleset was selected, the vanilla ruleset. This could thus still be saved in the `modCheckboxes` component, and so it was done.

However, after the G&K-vanilla split, the base ruleset was a selector box in the `GameOptionsTable` itself. To fix this, I added a function on the `modCheckboxes` that changes the base ruleset. Every time the users changes the base ruleset in the selector box, this function is called, setting the new base ruleset, so that mod compatability can be correctly checked. 

This is the line:
```Kotlin
modCheckboxes!!.setBaseRuleset(newBaseRuleset)
```

However, this assumes that the `modCheckboxes` component is initialized, which wasn't the case when loading the screen in portrait mode. In tha case, `modCheckboxes` would be `null`, and thus this would throw an exception.

This PR makes it so that even in portrait mode, the `GameOptionsScreen` still owns the `modCheckboxes`, but if it is loaded in portrait mode, it isn't added visually to the main menu. The `StartGameScreen` than gets the `modCheckboxes` from the `GameOptionsScreen`, and adds these to the screen itself. As thus in all cases the `modCheckboxes` are initialized, the previous exception never appears again, solving the problem.

logicminimal:
@xlenstra so basically modCheckboxes is a lateinit that was initialised in all cases before the vanilla and gnk split, but after this more cases were added which the initialisation of modcheckboxes was not implemented in

xlenstra:
> @xlenstra so basically modCheckboxes is a lateinit that was initialised in all cases before the vanilla and gnk split, but after this more cases were added which the initialisation of modcheckboxes was not implemented in

Not necessarily a lateInit, but a null-init that wasn't set to an actual value. It also wasn't initialized in all cases before the split, but the cases where it wasn't initialized didn't matter yet then. During the split I, however, did assume it was initialized in all cases, which lead to crashes after the split.

PR #5690, opened 2021-11-17
Regions part 3 - resource placement, resource settings

SimonCeder:
This is the last substantive part of the MapRegions class, I don't see a point to migrating natural wonder placement here as well.

As always, there are fallback functions for missing json definitions.

Implements resource placement according to region theming and adds the possibility for quite intricate rules for resource generation with a slight abuse of conditionals.

Also implements resource settings:
- Sparse, fewer resources and lower quantities in each strategic deposit. "Unciv classic"
- Default, Civ V base values
- Abundant, more and higher quantity resources.
- Strategic Balance, guarantees starting with Iron, Horse, and Oil nearby
- Legendary Start, makes starts very very good.

Also updates `terrainsCanBeFoundOn` for a bunch of resources. There *are* other terrains that resources can be found on in some cases in Civ V, either due to placing strategic resources for city states or placing the aforementioned Strategic Balance resources. I've chosen to ignore this and simply allow putting Strategic Balance resources on the "wrong" tile if there are no valid options.

Draft for 2 reasons mainly:

1. `terrainsCanBeFoundOn` is still slightly inaccurate, because Civ V uses rules such as "Forest, but not on Tundra", or "Grassland Hills with Fresh Water" for resource placements. See my comments in the G&K `TileResources.json` We can either live with the inaccuracy (and our tundra citruses), change the way `terrainsCanBeFoundOn` works to allow compound definitions, or move it all to uniques.
2. `placeStrategicAndBonuses()` is slightly slow, due to the need to build a large number of lists of the entire map. This list-building takes more than half the time of the entire resource placement code. I haven't been able to find a way around this. Still, the time might be acceptable as a one-time hit when starting the game.

<details><summary>How slow?</summary>

Running the old spreadResources side-by-side for comparison:

MapGenerator run with parameters (Huge Hexagonal R40Default)
MapGenerator.MapLandmassGenerator took 2.96ms
MapGenerator.raiseMountainsAndHills took 14.51ms
MapGenerator.applyHumidityAndTemperature took 4.50ms
MapGenerator.spawnLakesAndCoasts took 57.20ms
MapGenerator.spawnVegetation took 1.49ms
MapGenerator.spawnRareFeatures took 0.97ms
MapGenerator.spawnIce took 2.67ms
MapGenerator.assignContinents took 4.23ms
MapGenerator.RiverGenerator took 5.93ms
MapGenerator.generateRegions took 69.65ms
MapGenerator.assignRegions took 27.32ms
**MapGenerator.placeResourcesAndMinorCivs took 149.66ms** (100 ms of this is building lists)
MapGenerator.spreadResources took 136.81ms
MapGenerator.NaturalWonderGenerator took 11.18ms
MapGenerator.spreadAncientRuins took 13.53ms
</details>


SimonCeder:
Solved problem 1) with moar uniques, now all resources are placed correctly.
Problem 2) (slow speed) is still there, but might not be significant.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -795,21 +797,30 @@ open class TileInfo {
...
     fun setTileResource(newResource: TileResource, majorDeposit: Boolean = false) {
         resource = newResource.name
-        
+
         if (newResource.resourceType != ResourceType.Strategic) return
-        
+
         for (unique in newResource.getMatchingUniques(UniqueType.ResourceAmountOnTiles)) {
-            if (matchesTerrainFilter(unique.params[0])) {
+            if (matchesFilter(unique.params[0])) {
```
yairm210, line 804: ?? In map generation you don't have improvements, so why not use the more specific terrainFilter?

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
```diff
@@ -495,7 +506,20 @@ class MapRegions (val ruleset: Ruleset){
             }
         }
 
-        // TODO: Strategic Balance Resources
+        // Place Strategic Balance Resources
+        if (tileMap.mapParameters.mapResources == MapResources.strategicBalance) {
+            val candidateTiles = startTile.getTilesInDistanceRange(1..2).shuffled() + startTile.getTilesAtDistance(3).shuffled()
```
yairm210, line 511: Why the split between layers 1,2 and letter 3? This is the same as is you would have taken all the layers at once

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
```diff
@@ -787,35 +808,41 @@ class MapRegions (val ruleset: Ruleset){
         // If there are any weightings defined in json, assume they are complete. If there are none, use flat weightings instead
         val fallbackWeightings = ruleset.tileResources.values.none {
             it.resourceType == ResourceType.Luxury &&
-                (it.hasUnique(UniqueType.LuxuryWeighting) || it.hasUnique(UniqueType.LuxuryWeightingForCityStates)) }
+                (it.uniqueObjects.any { unique -> unique.isOfType(UniqueType.ResourceWeighting) } || it.hasUnique(UniqueType.LuxuryWeightingForCityStates)) }
```
yairm210, line 811: Why not hasUnique?

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
```diff
@@ -787,35 +808,41 @@ class MapRegions (val ruleset: Ruleset){
...
 
-        val maxRegionsWithLuxury = if (regions.count() > 12) 3 else 2
+        val maxRegionsWithLuxury = when {
```
yairm210, line 813: Regions.count/4... ?

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
SimonCeder: The placement respects the order of the list, so all tiles in ring 1 and 2 will be tried before any in ring 3 are tried.

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
SimonCeder: Because they have conditionals and so `hasUnique` will always evaluate to false when not passed a `StateForConditionals`

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
SimonCeder: More like `max(1, min(3, (Regions.count() - 1) / 4))` but surely this is clearer?

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
yairm210: But that's what getTilesInDistance does as well

SimonCeder: No, getTilesInDistance does not shuffle the tiles. Small difference perhaps, but it might be noticeable if more resources generally spawn at 6' o clock than in other directions.

yairm210: Ah I see, because it the shuffle
Nice

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -530,6 +530,8 @@ open class TileInfo {
             resource -> observingCiv != null && hasViewableResource(observingCiv)
             "Water resource" -> isWater && observingCiv != null && hasViewableResource(observingCiv)
             "Natural Wonder" -> naturalWonder != null
+            "Featureless" -> terrainFeatures.isEmpty()
+            "Fresh Water" -> isAdjacentToFreshwater
```
xlenstra, line 534: Wouldn't this create an infinite loop before? As `isAdjacentToFreshWater` is a lazy that calls `matchesTerrainFilter`?

PR #5694, opened 2021-11-19
Request for take-over: Rewritten Map Editor

SomeTroglodyte:
Original comment half-outdated...

This was almost-done about october and I can't seem to scrounge up the energy to finish polishing it. So, if anyone wants to take over, go ahead and take it, no attribution expected. Add me as remote, [check out the branch](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Unciv-notes/Working%20in%20the%20IDE%20and%20Contributing.md#testing-another-collaborators-work), try and go... That branch will be your local copy, and you can easily push updated versions of it to your own repo and PR from there - in case you didn't know.

Hints:
- ~~Currently this runs separately from the old editor and launches from a fourth button of the map edit menu.~~ It's meant to replace, and open directly - no menu at all. That means all files currently in master branch under the mapeditor folder go, except MapEditorScreen and EditorMapHolder, which are replaced by this.
- All todo's live in the main MapEditorScreenV2 file on top. Other todo's should correspond to one of those to mark places, nothing more.
- Merged to current ~~3.18.3~~ ~~3.18.7~~ ~~3.18.10~~ ~~3.18.11-patch1~~ ~~3.18.13~~ ~~3.18.14~~ ~~3.18.18~~ ~~3.19.3~~ ~~3.19.10~~ 4.0.5
- That 'drawer' widget is not quite what I had in mind. Look and feel, yes, but the way it interacts with child actors regarding sizing was meant to be more flexible. I spent ages on it, and had one as derivate of SplitPane with nulled left child that was quite fine - except I couldn't for religious-icon's sake get its listeners to leave the left part of the screen alone and not eat mouse/tap activity away... So forgive if it seems like a kludge with too much code for the result - it is.

Forgotten in the to-do list:
- Redo some icons


AdityaMH:
Outdated?

will-ca:
@AdityaMH It's been regularly updated, it adds some pretty enticing features, and the code looks solid.

SomeTroglodyte:
Way I see it biggest holes are 1 - the function to enforce valid tiles and 2 - lots of small cleanup like proper mod handling, oversights in isDirty handling and such. Point 1 meaning - you take a tile, the brush paints something on it, then said function resolves the possibly invalid result back to a valid combo. The old editor doesn't do that 100% correctly (e.g. allowing oases with resources), one of the motivations to begin this. I found no existing function that works properly, so I had planned to write a new one and lost motivation midway.

Here, I have prepared a setting that balances which way the reconciliation goes, e.g. brush is "forest" and you paint over desert. On one end of the setting the tile stays unchanged because forest can't go on desert, on the other the desert becomes plains so a forest becomes possible - making that flexible enough to work with any mod is the challenge.

Some new options are already quit fun to use - like the two-click "make me a river from here to there" or flood fill. But it needs polish and thorough testing.

No conflicts while merging up to 3.18.14 plus up to #5897

SomeTroglodyte:
OK, I'll be gone a while, so time to consider whether this is already better than the old one. IMHO it is. Really missing are the "drag paint" ability of the old one, not that hard to migrate, and a better fitting normalizeToRuleset function, but the old editor had problems with that too.

[The short demo video](https://www.dropbox.com/s/l4s4jnbsqtici23/Peek%202022-04-11%20Mapedit.mp4?dl=0) is too big for github, so you'll have to DL it. Shows just a few features.

AdityaMH:
Can you add starting location for random civ?

SomeTroglodyte:
Uh-oh. Merged - so I'll have to somehow find the time to work on those TODO's..

> Can you add starting location for random civ?

Nope, just as with the old editor. I get the idea, but that might not be trivial. I guess it would behave the same as if you placed a start for  _every_ possible nation on the desired tile, but that's tedious, I know.

PR #5697, opened 2021-11-19
`optipng` everything.

will-ca:
Not as much change as I was hoping for. Surely less when stripping out extraneous files for the Github page and such. May have no impact at all if atlases/packing are regenerated at build or run time and the optimizations don't carry over.

But I already ran the commands and set up the branch, so meh.

```bash
$ shopt -s globstar

$ du -bc **/**.png | tail -1
14673137	total

$ for f in **/**.png; do echo "$f"; done | xargs -d \\n -n 3 -P 6 optipng -preserve -o3 2>&1

$ du -bc **/**.png | tail -1
12996803	total
```

Haven't done pixel-by-pixel comparison of everything. But OptiPNG's supposed to be lossless, and the game still seems to run file for me.

Close if pointless.


SomeTroglodyte:
Impact is near zero as
- The game needs very few of the original png's at run time (ExtraImages)
- The packaged jar omits the unneeded png 'sources'
- The atlas(es/i/anti/whatever-the-plural) may profit, but load and run performance gain is definitely zero, and I doubt the bytes saved are worth it

will-ca:
Yeah, figures.

I started by doing a test that saw total PNG size drop from 18MB to 13MB, which seemed like it might be worth it even if most of it didn't carry over after packaging given the APK size of ~8MB. But it turns out I accidentally included huge modded tiles in that, so the gains in core files are much less.

SomeTroglodyte:
But kudos for the way of thinking. Actually, I would have treated android/assets/ExtraImages/World_Screen.png this way had I thought of it. Instead, I tried and compared 250k and 350k jpeg's and decided legibility after scaling is borderline anyway, so I stayed with lossless.

PR #5702, opened 2021-11-22
Counteroffer mechanic, updated trade valuations

SimonCeder:
EDIT: When making an offer to the AI, the AI will - unless accepting your trade proposal as is - get back to you with a counteroffer on their turn that they find acceptable. This is presented in the same way as other trades initiated by the AI.

Screenshot_20211122_223609

Note: This is not exactly a Civ V mechanic, Civ V had an instant "what do you want for this?" button. This system would not work without modifications in Unciv since our AIs respond to trade requests during their turn, not the player's.

Also updates some trade evaluations, mainly not giving extra value to multiple copies of luxuries since they can no longer be traded on, and Netherlands-type civs being more willing to sell their last copy of a luxury.


yairm210:
I think AI should always counteroffer. Worst case you won't accept, but it shows more agency on the AI's paylrt, which is what players read as 'good AI'

SimonCeder:
Sure, I added a notification string to make it more obvious what's going on.

AdityaMH:
Also AI always denied all trade counteroffer even fair.

PR #5705, opened 2021-11-24
We Love The King Day

SimonCeder:
Part of #4697
Implements We Love The King Day, cities will periodically demand new luxuries. Acquiring the luxuries by any means causes the city to go into WLTKD for 20 turns, giving +25% bonus to growth.

Adds a tutorial to explain the concept the first time a city demands a luxury.

Also updates the growth penalty for unhappiness - it should not block other bonuses than WLTKD, and goes as far as -100% at -10 happiness.


core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -550,33 +549,38 @@ class CityStats(val cityInfo: CityInfo) {
...
-        // Since growth bonuses are special, (applied afterwards) they will be displayed separately in the user interface as well.
-        if (totalFood > 0 && !isUnhappy) { // Percentage Growth bonus revoked when unhappy per https://forums.civfanatics.com/resources/complete-guide-to-happiness-vanilla.25584/
-            val foodFromGrowthBonuses = getGrowthBonusFromPoliciesAndWonders() * totalFood
-            newFinalStatList.add("Growth bonus", Stats(food = foodFromGrowthBonuses))  // Why Policies? Wonders can also provide this?
-            totalFood = newFinalStatList.values.map { it.food }.sum() // recalculate again
+        // Apply growth modifier only when positive food
+        if (totalFood > 0) {
+            // Since growth bonuses are special, (applied afterwards) they will be displayed separately in the user interface as well.
+            // All bonuses except We Love The King do apply even when unhappy
+            val foodFromGrowthBonuses = Stats(food = getGrowthBonusFromPoliciesAndWonders() * totalFood)
```
xlenstra, line 568: According to
https://forums.civfanatics.com/resources/complete-guide-to-happiness-vanilla.25584/
All positive growth modifiers are not applied when -10 < happiness < 0 (as was the case in the old code)
Is that source only valid for vanilla?

SimonCeder: Seems that it was never the case, relevant lines:
[Vanilla, starting on 5646](https://github.com/Gedemon/Civ5-DLL/blob/master/CvGameCoreDLL/CvCity.cpp)
[G&K, starting on 6294](https://github.com/Gedemon/Civ5-DLL/blob/master/CvGameCoreDLL_Expansion1/CvCity.cpp)
only specifically WLTKD bonus is supressed.

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -624,6 +683,42 @@ class CityInfo {
...
+            // 2. Not a city-state only resource eg jewelry
+            !it.hasUnique(UniqueType.CityStateOnlyResource) &&
+            // 3. Not the same as last
+            it.name != demandedResource &&
+            // 4. Not found nearby
+            getCenterTile().getTilesInDistance(3).none { nearTile -> nearTile.resource == it.name }
+        }.shuffled()
+
+        // Keep trying until we get one that actually exists on the map
+        val chosenResource = candidates.firstOrNull { tileMap.values.any { tile -> tile.resource == it.name } }
```
yairm210, line 699: Wildly inefficient
I think we save all natural wonders in a lazy transient in the tilemap, we can do the same with resources

core/src/com/unciv/logic/trade/TradeEvaluation.kt
```diff
@@ -91,14 +91,16 @@ class TradeEvaluation {
...
-                    0
+                val weLoveTheKingPotential = civInfo.cities.count { it.demandedResource == offer.name } * 50
+                return weLoveTheKingPotential +
+                    if(!civInfo.hasResource(offer.name)) { // we can't trade on resources, so we are only interested in 1 copy for ourselves
+                        when { // We're a lot more interested in luxury if low on happiness (AI is never low on happiness though)
+                            civInfo.getHappiness() < 0 -> 450
+                            civInfo.getHappiness() < 10 -> 350
+                            else -> 300 // Higher than corresponding sell cost since a trade is mutually beneficial!
+                        }
+                    } else
```
yairm210, line 102: This feels redundant, if() wltkd += when{} would be better

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -624,6 +683,42 @@ class CityInfo {
         civInfo.updateDetailedCivResources() // this building could be a resource-requiring one
     }
 
+    private fun demandNewResource() {
+        val candidates = getRuleset().tileResources.values.filter {
+            // 1. Luxury
+            it.resourceType == ResourceType.Luxury &&
+            // 2. Not a city-state only resource eg jewelry
+            !it.hasUnique(UniqueType.CityStateOnlyResource) &&
```
will-ca, line 691: Perhaps I'm just missing it, but I don't see anywhere that checks that you don't already have the resource globally?

IIRC in Civ V it will never demand a resource you have in *any* city or trade deal, not just in your current city, unless you aleady have *every* resource, in which case it may be possible to enter a semi-permanent WLTK cycle with instant completion as soon as it's available. Wikia doesn't say much, but that's what I think I noticed from playing. (Not sure about the semi-pemanent cycle— Definitely remember it, but it may be due to leaving a single resource untapped and buying it through a trade deal each time WLTK restarted.)

It also seemed to choose the globally furthest resources, instead of just one that isn't near the affected city. More like `sorted(ruleset.resources, key=(res -> min(cities.map(city -> city.distancetoresource(res)))))[randomindexnearend]` or something than `.getTilesInDistance(3).none(resource)`. Hm. Compare this approach with Yair's comment, though.

Perhaps build a transient that maps resource types to distance to closest empire city whenever any change in empire city locations is detected? Hm. Not sure what would be best way to do that. Iterate over all cities for all resources, for `O(resourcecount*tilecount*citiecount)` performance? Check only against capital, as a faster heuristic (`O(resourcecount*mapsize)`)? Or dilate from set of all empire/city center tiles, counting how many iterations it takes to find the first copy of each resource (Faster algorithm, *except* that dilation sounds expensive).

Or leave it as it is, at least for now, either because it already behaves correctly or because it behaves incorrectly but scope creep is lethal.

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -624,6 +683,42 @@ class CityInfo {
...
+            // 3. Not the same as last
+            it.name != demandedResource &&
+            // 4. Not found nearby
+            getCenterTile().getTilesInDistance(3).none { nearTile -> nearTile.resource == it.name }
```
will-ca, line 695: Perhaps I'm just missing it, but I don't see anywhere that checks that you don't already have the resource globally?

IIRC in Civ V it will never demand a resource you have in *any* city or trade deal, not just in your current city, unless you aleady have *every* resource, in which case it may be possible to enter a semi-permanent WLTK cycle with instant completion as soon as it's available. Wikia doesn't say much, but that's what I think I noticed from playing. (Not sure about the semi-pemanent cycle— Definitely remember it, but it may be due to leaving a single resource untapped and buying it through a trade deal each time WLTK restarted.)

It also seemed to choose the globally furthest resources, instead of just one that isn't near the affected city. More like `sorted(ruleset.resources, key=(res -> min(cities.map(city -> city.distancetoresource(res)))))[randomindexnearend]` or something than `.getTilesInDistance(3).none(resource)`. Hm. Compare this approach with Yair's comment, though.

Perhaps build a transient that maps resource types to distance to closest empire city whenever any change in empire city locations is detected? Hm. Not sure what would be best way to do that. Iterate over all cities for all resources, for `O(resourcecount*tilecount*citiecount)` performance? Check only against capital, as a faster heuristic (`O(resourcecount*mapsize)`)? Or dilate from set of all empire/city center tiles, counting how many iterations it takes to find the first copy of each resource (Faster algorithm, *except* that dilation sounds expensive).

Or leave it as it is, at least for now, either because it already behaves correctly or because it behaves incorrectly but scope creep is lethal.

Sorry for the multiple deletes/recomments. I used the wrong button to post, then highlighted the wrong lines.

SimonCeder: You're right, it's supposed to be a resource you *don't* have.
Distance is not a factor beyond "not within 3 tiles of this city"

SimonCeder: It seems that if you have every resource on the map, the WLTKD just keeps restarting with the same resource over and over.

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -624,6 +683,38 @@ class CityInfo {
...
+        }
+
+        val chosenResource = candidates.randomOrNull()
+        /* What if we had a WLTKD before but now the player has every resource in the game? We can't
+           pick a new resource, so the resource will stay stay the same and the city will demand it
+           again even if the player still has it. But we shouldn't punish success. */
+        if (chosenResource != null)
+            demandedResource = chosenResource.name
+        if (demandedResource == "") // Failed to get a valid resource, try again some time later
+            setFlag(CityFlags.ResourceDemand, 15 + Random().nextInt(10))
```
xlenstra, line 703: So if you found a new city after having all luxuries, it will never celebrate we love the king day?

xlenstra (approved):

core/src/com/unciv/logic/city/CityInfo.kt
SimonCeder: Yes, and I wasn't able to test this in-game in civ V, but from CvCity.cpp it is the behavior.
It is slightly inelegant, I'll admit. Other thinkable solutions:

1. A new city forces a luxury to be picked, making the behavior similar to old cities.
2. Once all luxuries are controlled, no cities demand more luxuries and so there is no WLTKD. This is perhaps the most intuitive solution, but it is punishing success, and encourages gamey behavior like intentional pillaging or breaking trades to get new WLTKD:s.
3. As long as all luxuries are controlled, all cities enter a permanent WLTKD of sorts without demanding any particular resource. Mainly a cosmetic difference from option 1, might reduce notification spam in the late game.

xlenstra: I don't think it matters all that much, as having all resources under your control almost always means you have basically won the game anyways (though excessive trading might be exception to this). In my personal opinion, option 3 would make the most sense as long as this behaviour is explained in the tutorial section on WLTKD. This also incentives trading for resources more than just the +4 happiness, as that often doesn't matter all that much.

core/src/com/unciv/logic/city/CityInfo.kt
will-ca: I probably got the impression of distance weighting just because Luxury Resources spawn regionally and I tend to gobble up any on my side of the map ASAP.

Restarting the same resource matches observed behaviour (from a long time ago), I think. IIRC for me every city would re-enter WLTK with whales at the same time whenever it ended, probably because that was the last luxury I got.

PR #5707, opened 2021-11-25
Added health conditionals

xlenstra:
Added conditionals:
- \<when above [amount] HP>
- \<when below [amount] HP>

Those both only apply to our unit, so for a pikeman with "[+20]% Strength <vs [All] units> <when below [20] HP>" attacking a swordsman, the pikeman would only receive the bonus if it less than 20 HP, regardless of the HP of the swordsman. For this reason, the conditionals are automatically reordered such that these conditionals are always in front of the 'vs' conditionals: "[+20]% Strength <when below [20] HP> <vs [All] units>".
As requested in #3242


yairm210:
A. Naming consistency - Health, not HP (iirc)

Adding "when" would make it clearer
Compare "vs all units above 20 health" to "vs all units when above 20 health"

xlenstra:
> A. Naming consistency - Health, not HP (iirc)

Currently, all healing uniques use HP instead of Health, and Health is only used in the unit overview screen, for city health and in the code base

> Adding "when" would make it clearer Compare "vs all units above 20 health" to "vs all units when above 20 health"

Only thing I worry about is having too many conditionals start with when: "[+20]% Strength <for [All] units> <when above [20] HP> <when fighting in [Marsh] tiles> <when attacking> <when at war>" starts sounding repetitive.

PR #5709, opened 2021-11-26
Add `interface IScreenHistoryAccessible{}` to unify "Return to previous screen" behaviour.

will-ca:
Re: https://github.com/yairm210/Unciv/commit/bdeccb87e827b3aad1a17ed8171f493ac79092a9

Basically does the same thing. But defines interface with extension method so the original screen doesn't have to explicitly pass `this` to the newly opened screen, another extension method to abstract the "keep track of and go to previous screen" closing behaviour so it doesn't have to be coded separately into E.G. CivilopediaScreen or (probably?) PickerScreen, and uses `closeAction: () -> Unit` with default value of `{ game.setScreen(this) }` instead of `previousScreen: BaseScreen` so things like WorldScreen can pass `{ game.setWorldScreen() }`.

This, like so much else, was simple in my head but feels costlier reified. It's 50% blank lines and KDocs, though.

Marked as draft, but I'm not planning changing anything more with this version of the idea.

### DIFF
--- a/core/src/com/unciv/ui/utils/IScreenHistoryAccessible.kt
+++ b/core/src/com/unciv/ui/utils/IScreenHistoryAccessible.kt
@@ -53,15 +53,14 @@ interface IScreenHistoryAccessible {
      * @throws IllegalStateException if there are no previous screens to return to.
      */
     fun BaseScreen.closeToPreviousScreen() {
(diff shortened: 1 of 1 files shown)


will-ca:
No interest, and wasn't as clean of an implementation as I'd intended either.

PR #5718, opened 2021-11-28
Show which cities are missing required buildings for National Wonders.

will-ca:
Hides the Unique's text and shows list of cities missing required building if any:

image

If all cities have the required building, then behaviour is unchanged:

image

Surprisingly legible, even with small screen size and lots of cities:

image

Could also truncate list if too long, I guess. But that may require adding something like "and {n} other cities…" to be clearest, which itself is worth a couple city names in length, and would make translation more complicated.

Civ V (at least G&K/BNW/BE) does something similar.

Line count/Diff is inflated by editor stripping trailing spaces.


yairm210:
Nice QOL improvement!

PR #5720, opened 2021-11-28
Strip trailing spaces in Kotlin files and changelog.

will-ca:
Why?

For cleaner future PRs/diffs.

I had a Python script to verify that the diff doesn't do anything except strip trailing spaces:

`$ python3 -ic 'difflines = open("/tmp/diff").readlines(); originals, news = ([l.rstrip("\n") for l in difflines if l and l[0] == char and not (l.startswith("+++ ") or l.startswith("--- "))] for char in "-+"); replacements=[(line, stripped in news and (news.remove(stripped) or stripped)) for line in originals for stripped in ["+"+line[1:].rstrip(" ")]]; assert not news; assert all(v for l, v in replacements)'`

I haven't checked exhaustively for multi-line strings, but the game looked fine when I ran it, and I'm not sure if that could even be an issue.

Markdown does use trailing spaces as a part of its syntax, but not AFAICT in any part of `changelog.md`.


will-ca:
Broken by merge.

will-ca:
@yairm210 Worth making another PR for? Or should I just drop this?

The effect on diffs is annoying. And I could see feature branches potentially ending up with merge conflicts that are entirely due to stripped trailing whitespaces.

SomeTroglodyte:
First, all contributors should have consistent editor settings, and conventions should govern this (including the LF before EOF) - the default Android Studio editors, at least since Arctic Fox, _will_ produce new trailing whitespace - lots of it.

I, when I still had the energy to produce, ended up with a personal decision to a) just have AS _display_ spurious whitespace and b) manually remove it in those source portions I touch, maybe a page up and down.

yairm210:
I think that using AS defaults is probably the way to go
I highly advise that this be your IDE anyway

Issue #5722, opened 2021-11-28
Easy pickings for colourblindness support?


will-ca:
**Version**
Git master.

**Describe the bug**
I wrote a quick Python script to find nation colours that might be easily confused with different types of colourblindness.

It turns out that for deuteranomaly, which is apparently by far the most common type of colourblindness, there's only a handful of countries that are probably the most problematic:

```python3
>>> testNations(filternations=isPlayableNation, mindifference=0.1, minprevalence=0.01)

India and Denmark may be hard to tell apart with:
	Deuteranomaly. (Diff: 0.099, Affects: ~4.0%)
Rome and Sweden may be hard to tell apart with:
	Deuteranomaly. (Diff: 0.089, Affects: ~4.0%)
Rome and Korea may be hard to tell apart with:
	Deuteranomaly. (Diff: 0.081, Affects: ~4.0%)
China and America may be hard to tell apart with:
	Deuteranomaly. (Diff: 0.07, Affects: ~4.0%)
The Netherlands and Austria may be hard to tell apart with:
	Deuteranomaly. (Diff: 0.058, Affects: ~4.0%)
India and Ethiopia may be hard to tell apart with:
	Deuteranomaly. (Diff: 0.033, Affects: ~4.0%)

Found 6 potentially problematic pairings.
```

For every other pairing of countries, the difference in either inner colour or outer colour is at least 10% of the dimensional width of the simulated colourblind colour-space (in which, for deuteranomaly, I'm currently just zeroing the Green RGB channel and then converting to YIQ).

For reference, the most similar nation colours with "normal"/"healthy" human vision, as measured in the same colour-space, have a distance of 0.11. The next most similar nation colours all have distances of at least 0.18.

```python3
>>> print(formatNationTestResults(dict([*testNations(filternations=isPlayableNation, mindifference=2, minprevalence=0.5).items()][-5:])))

Spain and Mongolia may be hard to tell apart with:
	Normal_Vision. (Diff: 0.2, Affects: ~100.0%)
Germany and The Maya may be hard to tell apart with:
	Normal_Vision. (Diff: 0.19, Affects: ~100.0%)
Korea and Ethiopia may be hard to tell apart with:
	Normal_Vision. (Diff: 0.18, Affects: ~100.0%)
Iroquois and Sweden may be hard to tell apart with:
	Normal_Vision. (Diff: 0.18, Affects: ~100.0%)
Babylon and America may be hard to tell apart with:
[...]

Issue #5721, opened 2021-11-28
City states with identical colours.


will-ca:
**Version**
Git master.

**Describe the bug**
Writing a colourblindness tester, I've found three sets of city states that have either completely identical or very similar colours:

```python3
>>> testNations(minprevalence=0.5)
Kuala Lumpur and Edinburgh may be hard to tell apart with Normal_Vision. (Difference: 0.0)
Kuala Lumpur and Valletta may be hard to tell apart with Normal_Vision. (Difference: 0.0)
Belgrade and Dublin may be hard to tell apart with Normal_Vision. (Difference: 0.0)
Edinburgh and Valletta may be hard to tell apart with Normal_Vision. (Difference: 0.0)
Cahokia and Jerusalem may be hard to tell apart with Normal_Vision. (Difference: 0.0703970853264033)
Found 5 potentially problematic pairings.
```

Since there are hundreds of possible pairings over dozens of city states, this is actually quite good.

Still, they should probably be changed.

SomeTroglodyte:
Edinburgh and Dublin should be out ("Will not be chosen for new games" unique - might be time to treat that as a deprecation that can be removed) as they're now in Celt's city list. I vaguely remember choosing Valletta (definitely) and Belgrade (? did I? Yup, those tongue-in-cheek texts sound like mine) as direct replacements. And the religious CS were originally all white if I remember correctly, so them still being close isn't that surprising. Some CS replacements should probably be reverted for the new Vanilla ruleset...

yairm210:
Until this is solved (disclosure - I'm not planning on dealing with this), let's merge all relevant information into a single issue
I think #5722 's title is more apt for what we're going for

Issue #231, opened 2018-08-31
Add symbols for colorblind unit recognition


yairm210:

yairm210:
To be honest I'm not quite sure what to do with this.
I'm not sure that the game can even be changed to be colorblind-friendly without a massive change in UI because both civ and tile type recognition are color-dependent...

L314:
One Idea I had for this is to have different Civs have different shapes as their bg shape. Another easier fix is to label all tiles with who it belongs to. Like the units are right now.

theodorthegreathe:
I beleive UnCiv is playable even on E-Ink screens. I can  provide you with a YotaPhone 2 for tests. Obviously, the terrain is not the biggest problem, one just have to issue monochrome Tileset which only requires cosmetic changes on Default or ThorfMaps.
An elegant solution to distinguish ownership is the usage of geometric 🚩🔰📍 or symbolic ($£@₪₹¤¥) flags on an occupied tile.Right part of the hexagon is vacant for that  at the moment.
When decreasing scale, one may want to make this flags larger.

kfish610:
In order to not clutter up the display for non-colorblind users, maybe the changes would be an option in the settings?

i-need-to-tell-you-something:
## Solution 1: 
You can use something like [this](https://www.colorhexa.com/68bf26#blindness-simulator) to test your color scheme for different types of color blindness. There are probably even better tools somewhere. I can ask around if need be.

## Solution 2:
### For tile types:
The tileset should have ever so slightly more details to it than just the color. Similarly to e.g. [these](https://duckduckgo.com/?q=civ+6+strategic+view&t=canonical&iar=images&iax=images&ia=images&iai=http%3A%2F%2Fi.kinja-img.com%2Fgawker-media%2Fimage%2Fupload%2Ft_original%2Fduhy5ncusnlkinkmrr6h.png).

### For tile ownership:
[...]

PR #5723, opened 2021-11-28
Add Python script to check colourblindness.

will-ca:
1. Skim script to make sure I'm not doing anything nefarious.
2. Run directly from terminal to walk through the automated wizard, or run with `python -i` to call `testNations()` yourself.

Uses only Python Standard Library imports, I believe.

Uses very simplistic colour models. Alternative would probably be external import and PIP dependency. But easy to swap out if getting serious.

Related: #231, #5721, #5722.

### DIFF
--- a/NationColourHelper.py
+++ b/NationColourHelper.py
@@ -1,0 +1,239 @@
+#!/usr/bin/env python3
+
+"""
+This script is intended to help detect nation colours that may be appear very similar depending on the viewer's colour vision.
+
+Due to the extremely high number of nations included in the game, and due to the high reliance on hue to distinguish player and terrain data, full colourblindness support is not currently considered feasible in Unciv:
+
+https://github.com/yairm210/Unciv/issues/231
+
+However, my hope is that this script can still be useful.
+
+Simply run this script from the command line to interactively print a report for similar colours between player civilizations that may affect more than 1% of the human population.
+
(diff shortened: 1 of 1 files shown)


yairm210:
This looks like a nice solution, but I'm not going to add a Python script into this repo
Kotlin ensures we can later connect it to user output, tests perhaps, and makes it truly cross platform, which Python isn't
If we add this as Python it'll stay as Python...

SomeTroglodyte:
Actually, python could easily be argued to be more cross-platform than kotlin. I would be all for using whatever suits best for tool scripts as long as it supports the same dev platforms kotlin does, and to have them version-controlled.

This is not a unit test, it does not try to prevent release of flawed code - thus "just" a tool. That said, I think it quite valuable. 

But it should definitely not live in the project root folder. A subfolder that clearly indicates "tools" use and is excluded properly from distributions - or a separate repo. Such a dedicated place would be welcome to take ownership of the Unciv-related helpers I did in python...

will-ca:
@yairm210  Fair enough.

Porting it to Kotlin would also make it just a small step further to, E.G., analyze what nation colours are used when loading a save and mutate any that are too close together depending on user's vision setting. Just one more feature…

I also agree with @SomeTroglodyte's assessment in more general terms, though. You wouldn't require the IDE or build system to run on Android, so why not allow tool scripts as long as they run on all the same dev platforms? But in this case, there are also specific benefits that could be realized by doing the checks in Kotlin.

PR #5731, opened 2021-11-29
Merged some uniques and added a new one for not allowing to build units

xlenstra:
This PR merges a few uniques using conditionals, and also adds a new one, requested in #3242.

Added uniques:
- "Cannot build [unitFilter] units"

Deprecated uniques:
- "Cannot enter ocean tiles until Astronomy" -- replaced with "Cannot enter ocean tiles \<before discovering [Astronomy]>"
- "Cannot be built on [tileFilter] tiles until [tech]" -- replaced with "Cannot be built on [tileFilter] tiles \<before discovering [tech]>"


core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -283,8 +283,10 @@ class MapUnit {
 
         //todo: consider parameterizing [terrainFilter] in some of the following:
         canEnterIceTiles = hasUnique(UniqueType.CanEnterIceTiles)
-        cannotEnterOceanTiles = hasUnique(UniqueType.CannotEnterOcean)
-        cannotEnterOceanTilesUntilAstronomy = hasUnique(UniqueType.CannotEnterOceanUntilAstronomy)
+        cannotEnterOceanTiles = hasUnique(UniqueType.CannotEnterOcean, StateForConditionals(civInfo=civInfo, unit=this))
```
yairm210, line 286: Does this update when we gain the required tech?

xlenstra: Yes, but there are bigger problems. It doesn't currently work at all, as the 'cannot enter ocean tiles' does not apply before researching astronomy, so you can always enter ocean tiles. After researching Astronomy you have researched Astronomy, so you can enter ocean tiles.
I'll probably need to rework most of this PR for this to work correctly, until then I'll mark this as draft.

xlenstra: I've returned to this PR and found that what I thought was happening, did in fact not happen, but I just made a typo.

This also doesn't update when we gain the required tech, but that behaviour is not changed from before this PR: if you researched astronomy mid-turn your ships also wouldn't be able to enter ocean tiles immediately. If need be I can implement this, but it will cost performance, as the only way to do this correctly would be to update the uniques of all units every time a tech is discovered

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -176,9 +176,12 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags:
     EnablesResearchAgreements("Enables Research agreements", UniqueTarget.Global),
     TriggersVictory("Triggers victory", UniqueTarget.Global),
     TriggersCulturalVictory("Triggers a Cultural Victory upon completion", UniqueTarget.Global),
-
+    
+    CannotBuildUnits("Cannot build [baseUnitFilter] units", UniqueTarget.Nation),
```
yairm210, line 180: Why not Global? If you can build a building that grants you bonuses but prevents certain units that's an interesting trade off!

PR #5732, opened 2021-11-30
Fixed a bug where statue of Zeus would not work

xlenstra:
Also removed some code belonging to removed uniques


yairm210:
Please create unique types, so modders recognize that they have deprecated uniques and know to replace them

xlenstra:
The removed uniques already had unique types, namely `StrengthPlusVs`, `StrengthMinVs` and `CombatBonus` that are marked as 'removed' in the UniqueType file since 3.17.13. However, the uniques here still used the default string, probably a result of me forgetting to replace them when doing the strength conditionals.

yairm210:
K :)

PR #5738, opened 2021-12-02
Recycling center

itanasi:
Adds Recycling Center
Blocks showing the building in the choices if 5 or more are built
If 5 are building, still shows but with a Rejection Text
Image included thanks to @SimonCeder. Credit added to Credits.md


core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -454,6 +454,12 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
     override fun shouldBeDisplayed(cityConstructions: CityConstructions): Boolean {
         if (cityConstructions.isBeingConstructedOrEnqueued(name))
             return false
+        if (hasUnique(UniqueType.MaxNumberBuilding)) {
```
yairm210, line 457: Unnecessary if

android/assets/jsons/Civ V - Gods & Kings/Buildings.json
```diff
@@ -1049,6 +1049,12 @@
 		"requiredTech": "Ecology",
 		"uniques": ["Must be next to [Desert]", "Cannot be built with [Nuclear Plant]"]
 	},
+	{
+		"name": "Recycling Center",
+		"maintenance": 3,
+		"requiredTech": "Ecology",
+		"uniques": ["Provides [2] [Aluminum]","Can have [5] of this building in your empire at most"]
```
yairm210, line 1056: Change to "Limited to [] per Civilization" so we can reuse this for units, improvements etc

itanasi:
Resolved requests. Added code to destroy building if exceeds max allowed (including enqueued). Tested in Human v Human game both captured and traded.

Future improvement: perhaps remove (and refund) queued buildings first before resorting to destroying the new one in the captured/traded city. But that's a bit complicated for today's work

core/src/com/unciv/models/ruleset/Building.kt
itanasi: Done

Issue #3927, opened 2021-05-13
Very old android versions - worth debugging?


SomeTroglodyte:
Running the game in a KitKat AVD (silly, and I have no idea what drove me to try - it even takes more disk space than the Oreo one):
- Mod manager runs into a problem because the openssl version can't cope with the upgraded minimum TLS demanded by github. 
     I can catch this (`catch (ex: SSLHandshakeException)`) but then what? Trying http (`return download(URL("http", url.host, url.file), action)`) just gives a 301 permanently moved - back to https.
- `CameraStageBaseScreen.closeAllPopups` crashes, a `.toList()` fixes the crash

No big magic, but before I debug everything I find - _is supporting Android 4.4 kitkat worth it_? Any input welcome.

If answers tend to yes, _only then_ I'll try to find out the actual android version levels where anything changes, and start implementing solutions. #1 would get a useful user message, #2 is trivial, ...

yairm210:
Google no longer supports it, and most major things stopped supporting it 2 years ago... I see no point tbh

SomeTroglodyte:
Still thinking - maybe just add a one-time warning? Otherwise, we could / should bump up the minimum API... In the meantime, until I manage to prybar my fingers  away from that nostalgic AVM I'll collect what I find right here. Do you still get mails if I just _edit_ a comment? No, wait, I'll move this data [to my own repo](https://github.com/SomeTroglodyte/Scratchpad/issues/1).

Issue #5737, opened 2021-12-01
Broken at API Level 17.


will-ca:
**Platform**
Android Studio/AVD-emulated Nexus 4 with `"Jelly Bean" API Level 17 x86 Android 4.2 (Google APIs)` image.

**Version**
Git master, within one or two commits.

**Describe the bug**
I suppose that officially, API Level 17 is supported by the `minSdk` value.

However, the game doesn't seem to actually work in an emulator at API Level 17:

![image](https://user-images.githubusercontent.com/37680486/144161965-850fde94-95af-473b-93c8-8dfdbb543908.png)

```
W/System.err: java.lang.NumberFormatException: Invalid int: "+3"
        at java.lang.Integer.invalidInt(Integer.java:138)
        at java.lang.Integer.parse(Integer.java:375)
        at java.lang.Integer.parseInt(Integer.java:366)
        at java.lang.Integer.parseInt(Integer.java:332)
        at com.unciv.logic.map.TileInfo.getTileFertility(TileInfo.kt:367)
        at com.unciv.logic.map.mapgenerator.MapRegions.generateRegions(MapRegions.kt:84)
        at com.unciv.logic.map.mapgenerator.MapGenerator$generateMap$10.invoke(MapGenerator.kt:85)
        at com.unciv.logic.map.mapgenerator.MapGenerator$generateMap$10.invoke(MapGenerator.kt:84)
        at com.unciv.logic.map.mapgenerator.MapGenerator.runAndMeasure(MapGenerator.kt:107)
        at com.unciv.logic.map.mapgenerator.MapGenerator.generateMap(MapGenerator.kt:84)
        at com.unciv.logic.GameStarter$startNewGame$3.invoke(GameStarter.kt:54)
        at com.unciv.logic.GameStarter$startNewGame$3.invoke(GameStarter.kt:51)
        at com.unciv.logic.GameStarter.runAndMeasure(GameStarter.kt:127)
        at com.unciv.logic.GameStarter.startNewGame(GameStarter.kt:51)
        at com.unciv.ui.newgamescreen.NewGameScreen.newGameThread(NewGameScreen.kt:196)
        at com.unciv.ui.newgamescreen.NewGameScreen.access$newGameThread(NewGameScreen.kt:24)
        at com.unciv.ui.newgamescreen.NewGameScreen$3$4.invoke(NewGameScreen.kt:130)
[...]

PR #5744, opened 2021-12-03
Desugar Java 8 APIs for older Android devices.

will-ca:
[Recent-ish](https://developer.android.com/studio/releases/gradle-plugin#4-0-0) versions of the Android Gradle plugin purport to support certain Java 8 features through some compiler translation thing, even on phones that only support Java 7:

https://developer.android.com/studio/write/java8-support

Also switched to `minSdk = 21`, as otherwise `multiDexEnabled = true` would have to set for desugaring. 20's only for wearables, and there are unresolved critical bugs at 19 anyway (#3927).

APK size impact:

```bash
$ du -s master/**/*.apk desugar/**/*.apk
12004	master/android/build/outputs/apk/debug/Unciv-debug.apk
12400	desugar/android/build/outputs/apk/debug/Unciv-debug.apk
```

MultiDex would, IIRC, add around another ~0.5MB.

Tested on 24 as that's the AVD I have. If merged, I will/can install a 21 AVD and see if everything works there.

Some relevant code paths/commits:

https://github.com/yairm210/Unciv/commit/cbeb9a96a64948c10cf983313488102ee16f0615

https://github.com/yairm210/Unciv/commit/24d71c345037522e67350bca03c6be7a81799180

https://github.com/yairm210/Unciv/blob/7ddbac1f07dd6a836830c5454bcb3312216cf56b/core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt#L118-L124

(I just searched for "Java 8".)

There's also an unrelated minor change to `gradle.properties` here. I mean, it builds fine with 1.5GB of memory, and the 4GB setting was causing pretty bad swapping and freezing my fairly high-RAM laptop, but I can remove or separate the change if that would be preferred.

### DIFF
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -24,7 +24,7 @@ android {
     }
     defaultConfig {
         applicationId = "com.unciv.app"
-        minSdk = 17
+        minSdk = 21
         targetSdk = 30 // See #5044
         versionCode = BuildConfig.appCodeNumber
         versionName = BuildConfig.appVersion
@@ -60,6 +60,7 @@ android {
         disable("MissingTranslation")
     }
     compileOptions {
+        isCoreLibraryDesugaringEnabled = true // https://developer.android.com/studio/write/java8-support
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
@@ -124,4 +125,5 @@ dependencies {
     //   run `./gradlew build --scan` to see details
     implementation("androidx.core:core-ktx:1.6.0")
     implementation("androidx.work:work-runtime-ktx:2.6.0")
+    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:1.1.5")
 }
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,18 @@
+## For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+#
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx1024m -XX:MaxPermSize=256m
+# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+#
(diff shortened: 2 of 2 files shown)


yairm210:
I'm entirely unsure what your aim is here.
Raising the minimum api level I understand, but what do we gain from the other changes?

will-ca:
`.replaceAll` and `java.time`, for example. Everything else on the table of desugarable APIs: https://developer.android.com/studio/write/java8-support-table

Usually `.replaceAll` would require API level 24+, and `java.time` would require 26+. With "desugaring", they can apparently be used on *any* API level, so the three commits I linked (and probably others) can be rewritten/reverted to be more concise, and future code can freely use the many Java 8 features in that support table too.

Presumably Google would update their desugaring system with new API levels, so this would make backwards compatibility easier in the future too— If they release some fancy new thing for API Level 35 or whatever, we would immediately be able to use it without having to increase the minimum API level.

So there's kinda three semi-related changes here:
1. Enable desugaring. This lets Java 8 APIs be used regardless of minimum API level. So E.G. Supporting 21 doesn't create any additional burden over supporting 29, because incompatible Java 8 APIs are translated at compile time, so you can just write for Java 8 regardless of `minSdk`.
2. Raise API level *for desugaring*. I bumped it up to 21 here not just because of the conversation in #5737, but also because Google says that using desugaring at any lower level would require MultiDEX to also be used.
3. Decrease maximum Gradle memory. An unrelated change. I can separate this or remove it, but my laptop was freezing, the lower memory limit seems to work fine, and I didn't feel it was worth making a separate PR.

This is based on my understanding of the linked pages, so it's possible I've completely misunderstood what it will do or misconfigured it, but it all looks straightforwards enough— Well, I trust myself to flip a build switch more than I trust myself to refactor and test existing code in a part I haven't touched before.

yairm210:
I'd rather raise the api level directly to 24 than try and shim in missing java 8 functionality for the bottom 3% of devices, many of which may not even be played at all

SomeTroglodyte:
I thought we had isolated _all_ date formatting to the extension functions, if only to make any migrations like this one easier... You say I overlooked one, in the mod manager no less? No, that's from before that move. OK...

will-ca:
@SomeTroglodyte Oh, yeah. Sorry. It is indeed in `ExtensionFunctions.kt`; That's how I know about it and why I started thinking about Java 8. I went digging through the Git history to see if there was any further context when it was added.

PR #5764, opened 2021-12-06
Sort Maintenace Using Fixed Point

itanasi:
Using suggestion from @SomeTroglodyte, try sorting using fixed point value to avoid float bugginess


yairm210:
Please add a comment explaining why, because this reads as completely arbitrary

itanasi:
The theory is that with floating point, certain values will be too close and so when comparing A to B you'll have a different result than B to A. So perhaps sorting using fixed point will ensure consistent value ordering.

yairm210:
I mean in the code :)
Imagine being the person who reads this line and it's like: ????

SomeTroglodyte:
And remember, I pointed out a similarity, there's no proof this was behind any crashes, that should also be documented under source control

itanasi:
Ah, good thought. Done!

PR #5777, opened 2021-12-09
Update Japanese.properties

satorope:
As far as I can tell, I corrected the unnatural translation.
I changed the space from "" to " " because the grammar is English.
Translated the inherent abilities of each civilization and the greeting immediately after the start of the game.
Added missing translation. Move it to the appropriate line.
I translated many other things.

Is "Great People" not used ?

### DIFF
--- a/android/assets/jsons/translations/Japanese.properties
+++ b/android/assets/jsons/translations/Japanese.properties
@@ -2,7 +2,7 @@
 
 # Equivalent of a space in your language
 # If your language doesn't use spaces, just add "" as a translation, otherwise " "
-" " = ""
+" " = " "
 
 # If the first word in a sentence starts with a capital in your language, 
 # put the english word 'true' behind the '=', otherwise 'false'.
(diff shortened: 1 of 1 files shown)


xlenstra:
Placeholder `[cityFilter]` not found in `Japanese` for entry `[amount]% unhappiness from population [cityFilter]`
    Placeholder `[tileFilter]` not found in `Japanese` for entry `No Maintenance costs for improvements in [tileFilter] tiles`
    Placeholder `[amount]` not found in `Japanese` for entry `May buy [baseUnitFilter] units with [stat] for [amount] times their normal Production cost`

xlenstra:
The "" should only be replaced by a " " if Japanese uses spaces, as then they are automatically inserted in some cases. Although I don't know Japanese, the existing translations don't seem to use spaces, so this should probably remain ""?

Great People is used as a header for one of the tutorials, and probably in other places as well

yairm210:
Still relevant?

yairm210:
Closing, you can reopen of relevant

PR #5778, opened 2021-12-09
Update Nations.json

Smashfanful:
Took out two unecessary spaces


SomeTroglodyte:
Strange. Existing translations are present for the fixed variants - why would TranslationFileWriter not pick up the wrong, pre-this variant?

xlenstra:
> Strange. Existing translations are present for the fixed variants - why would TranslationFileWriter not pick up the wrong, pre-this variant?

Different vanilla & G&K versions?

PR #5782, opened 2021-12-11
Disable `-XstartOnFirstThread` when not on Mac.

will-ca:
I've never actually been able to use `desktop:run`, because the `-XstartOnFirstThread` argument is apparently highly non-standard:

```bash
$ ./gradlew desktop:run

> Task :desktop:run
Unrecognized option: -XstartOnFirstThread
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

So, add it only on Mac.

Don't have $1000 to spend on a Apple machine to test this, and documentation is surprisingly sparse, so basing it off the below examples:

https://mkyong.com/java/how-to-detect-os-in-java-systemgetpropertyosname

https://dirask.com/posts/Java-How-to-check-OS-name-and-if-I-run-application-on-Windows-Linux-or-Mac-OS-zjMJa1


desktop/build.gradle.kts
```diff
@@ -23,9 +23,11 @@ val deployFolder = file("../deploy")
 
 // See https://github.com/libgdx/libgdx/wiki/Starter-classes-and-configuration#common-issues
 // and https://github.com/yairm210/Unciv/issues/5679
-val jvmArgsForMac = listOf("-XstartOnFirstThread", "-Djava.awt.headless=true")
+val customJvmArgs = mutableListOf("-Djava.awt.headless=true")
+if ("mac" in System.getProperty("os.name").toLowerCase())
+    customJvmArgs.add("-XstartOnFirstThread") // Non-standard, only available (and necessary) on Mac.
```
yairm210, line 28: Also the second variable is only necessary for Mac

will-ca: Changed.

yairm210 (changes requested):
Haha my bad, I didn't realize I started the review but didn't finalize it so you didn't see my comment yet 😅

desktop/build.gradle.kts
```diff
@@ -136,7 +138,7 @@ for (platform in PackrConfig.Platform.values()) {
                         " --classpath $jarFile" +
                         " --mainclass $mainClassName" +
                         " --vmargs Xmx1G " +
-                        (if (platform == PackrConfig.Platform.MacOS) jvmArgsForMac.joinToString(" ") {
+                        (if (platform == PackrConfig.Platform.MacOS) customJvmArgs.joinToString(" ") {
```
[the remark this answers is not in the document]
will-ca: Haha, no worries!

Change done. But TBH I'm not totally clear on how Packr works or what this statement does. Hopefully it works.

The type cast (in new commit) is vaguely dangerous, because theoretically I suppose the Gradle builder scope could implement a custom getter that chews up the `MutableList` I tried to assign and returns something weird— Keep a `val` if so— Probably not worth prematurely worrying about, though.

PR #5789, opened 2021-12-12
Improvements to TurnChecker data usage

### LINKED ISSUE #5533: TurnNotification uses a lot more mobile data than necessary
Here is the problem:
I currently got 3 games in my multiplayer list with one around 400 KB and two around 800 KB in size. When they are zipped for upload and download they are around 70KB/140KB respectively. So if I turn on the Notifications set to every 5 minutes and it's not my turn for a whole day, I'm downloading 350 KB every 5 minutes only to get the current player. That is 288*350KB=~100 MB spread over the day which is around 10% of my mobile data volume. Sure I've got WiFi in most places but I think this amount of data usage is just overkill. 

So my proposal is to use the GameInfoPreview (around 400 Byte in size for a 5 player game zipped) by uploading it together with the GameInfo and using it instead for the MultiplayerTurnCheckerWorker and the MultiplayerScreen to update the player info

### LINKED ISSUE #5756: Multiplayer Section doesn’t update even when turn notifications are on (can be an improvement request)
**Platform**
Android 11

**Version**
3.18.6 and a few previous releases

**Describe the bug**
For the `out-of-game turn notifications` to work, the game needs to download `Multiplayer Games Saves` from `Dropbox` and parse them to know if the state of the game change. This means the Multiplayer section should update automatically also as the data has already been downloaded. It shouldn’t be necessary to press `Refresh list` and wait for the game to download those files again.

But in reality, this doesn’t happen. Players need to press `Refresh list` manually to see changes in Multiplayer UI.

This also happens when players are inside of a game. If a game state changes, those changes are not reflected in Multiplayer  UI. So, unless you press `Refresh list`, basically all games in the Multiplayer UI would be in the same state as they were when first added.

I don’t know if It's a bug or hasn’t been implemented yet. In case It hasn’t been implemented yet this issue might be of an improvement rather than of a bug, or maybe both.

**To Reproduce**
Steps to reproduce the behaviour:
1. Enable turn notifications: `Options` > `Multiplayer` > `Enable out-of-game turn notifications` > `yes`
2. Press `Refresh list` in Multiplayer
3. Wait for a game state to change
4. Check `Mulplayer` UI

**Expected Behaviour**
Changes displayed in Multiplayer UI

**Actual Behavior**
None of the changes is reflected

**Screenshots**
Example of the Multiplayer UI

GGGuenni:
Finally implemented the use of GameInfoPreview for the TurnChecker and MultiplayerScreen which I wanted to do almost two months ago. (Oops 😬)

Resolves #5756 as the saving of the files after the TurnChecker download has been reimplemented. I hope it does not make any problems again, but it shouldn't (fingers crossed) since the preview files are way smaller and thus saved faster.
Also closes #5533 as the download size per game is reduced to a few hundred bytes.

### DIFF
--- a/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
+++ b/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
@@ -16,6 +16,7 @@ import com.unciv.logic.GameInfo
 import com.unciv.logic.GameSaver
 import com.unciv.models.metadata.GameSettings
 import com.unciv.ui.worldscreen.mainmenu.OnlineMultiplayer
+import java.io.FileNotFoundException
 import java.io.PrintWriter
 import java.io.StringWriter
 import java.io.Writer
@@ -175,6 +176,7 @@ class MultiplayerTurnCheckWorker(appContext: Context, workerParams: WorkerParame
                         gameIds[count] = gamePreview.gameId
                         gameNames[count] = gameFile.name()
                         count++
+                    }
                 } catch (ex: Throwable) {
                     //only loadGamePreviewFromFile can throw an exception
                     //nothing will be added to the arrays if it fails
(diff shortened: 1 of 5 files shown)


yairm210:
Looks good
Was this tested on existing multiplayer games?

GGGuenni:
I've tested with newly created games and one old game which does not have a preview in dropbox and where it's also not my turn. Unfortunately, I could not find an old game where it is my turn, but that should not make a difference

PR #5793, opened 2021-12-13
Shrink maximum Gradle RAM use.

will-ca:
image

I mean… 7.5GB just to get a build running in an AVD? Isn't that a bit obscene? I barely managed to run my last PR, and my SSD is just a little bit closer to breaking now for all the processes that ended being pushed into swap.

This is the size I've been using in my feature branch, where I haven't had any problems with it. Maybe release builds might take more? Idk. Android Studio says the default is 1.5GB, though the comments it added to the file seem to say something else.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,18 @@
+## For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+#
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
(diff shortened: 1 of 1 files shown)


yairm210:
I disagree here. I have quite a slow computer, and allowing gradle to use more ram is definitely beneficial for me
I'll try to provide numbers

will-ca:
Odd. I suppose there may be something about my system that causes it to use RAM more zealously with the same limit? (By which I mean: Oversized pagefile.)

Would it be possible to perhaps add `gradle.properties` to `.gitignore`? Presumably that would let new branches use whatever the defaults are, which should work for the maximum number of people?

PR #5802, opened 2021-12-16
Chain Military and Civilian Units Together

Interdice:


Interdice:
still very much a draft

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -170,6 +170,8 @@ class MapUnit {
     var religion: String? = null
     var religiousStrengthLost = 0
 
+    var isLinked = false
```
will-ca, line 173: `linkedTo: MapUnit?`?

Interdice: Does it matter if we list what it is linked to? we can always call the function "getUnitNotOfType" to get the linkedunit.

yairm210: Interdice is correct here - we want to preserve linking between turns, so we need to save the link in a way that's json serializable

SomeTroglodyte:
`allUnitActionsHaveTranslation FAILED` - I remember thinking before that this test is overly picky

Interdice:
@SomeTroglodyte Yeah I was going to add the translations at the end.

docs/uniques.md
```diff
@@ -1069,4 +1069,4 @@ Applicable to: Conditional
  - "[amount] Visibility Range" - Deprecated As of 3.17.5 - removed 3.18.5, replace with "[amount] Sight"
  - "Limited Visibility" - Deprecated As of 3.17.5 - removed 3.18.5, replace with "[-1] Sight"
  - "[stats] on [tileFilter] tiles once [tech] is discovered" - Deprecated As of 3.17.10, replace with "[stats] from [tileFilter] tiles <after discovering [tech]>"
- - "Deal 30 damage to adjacent enemy units" - Deprecated As of 3.17.10, replace with "Adjacent enemy units ending their turn take [30] damage"
\ No newline at end of file
+ - "Deal 30 damage to adjacent enemy units" - Deprecated As of 3.17.10, replace with "Adjacent enemy units ending their turn take [30] damage"
```
Interdice, line 1072: idk why this is here

Interdice:
How would I save the "isLinked" variable to the save

xlenstra:
> How would I save the "isLinked" variable to the save?

As long as you do not put '@Transient' before it, and add a way to copy it in the `.clone()` method of the unit, it should work automatically.

SomeTroglodyte:
> How would I save the "isLinked" variable to the save?

If you're curious about the why - saves are serialized through Gdx.Json, which has [documentation](https://github.com/libgdx/libgdx/wiki/Reading-and-writing-JSON) - and has sources so you can actually trace its code in Android Studio. In a nutshell - recursive Java reflection _tries_ to scan your object and understand the fields/types. Very specialized collections might give surprises. And the reflection code tries to include everything it can, thus the @Transient annotation to give back some control.

docs/uniques.md
will-ca: Documentation for Uniques is automatically generated with Gradle runs:

https://github.com/yairm210/Unciv/commit/4617bc21a70f4f4bd29dc80ba7d648349c9fc3f8

I `git add` selectively to keep them out of commits. E.G. `git add core`.

PR #5803, opened 2021-12-16
Autogenerate KDocs for Github Pages with Dokka.

will-ca:
Why? The direct reason is that I realized the other day that a way of specifying reflective operations I added early on in the scripting API branch is actually a domain-specific language that itself can already be used for a significant amount of scripting, so I'm probably going to be able to have a (series of many) PR(s) ready for Android-compatible mod-scripted event handlers somewhat sooner than I'd originally expected— Which means that I should also make sure there's a way for modders to explore the APIs available to them in scripting languages without having to dig through or read the Kotlin code.

But aside from there, there's also been a couple times while I was coding where I thought it would be helpful if I could get an overview of a class's members without having to find it in the source.

The generated HTML includes links to Github sources, Kotlin docs, Java 7 docs, Android SDK docs, and GDX docs, where applicable.

Example output currently hosted at the GH page of my fork:

https://will-ca.github.io/Unciv/dokka/

To generate/update docs:

```bash
$ ./gradlew dokkaRoot
```

The final assembled web files go into `/docs/dokka/`, as that can easily be exposed by Github Pages. Intermediate files stay in `build/` in each subproject.

Dokka takes up to a couple minutes for this project, so it's probably best to like, either manually or automatically update the docs before every major release or something.

HTML output from Dokka currently works best with an HTTP server. The one that comes with Python can be used for testing. This should expose the same directories as do Github pages:

```bash
$ python3 -m http.server 8000 --bind 127.0.0.1 -d docs
```

Then go to `http://localhost:8000/dokka/` to view them.

~~A separate PR will include the docs as web files.~~ Actually, no. No need to PR 100K LoC of opaque code when it can just be generated from this one.

Recommend: Making a separate `dokka-html` branch in this repository, running the task only in there so the `master` branch history and forks aren't bombarded with millions of lines of HTML, and setting the Github pages of the project to point there:

image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,11 @@
 import com.unciv.build.BuildConfig.gdxVersion
-import com.unciv.build.BuildConfig.roboVMVersion
+import com.unciv.build.BuildConfig
+import java.net.URL
 
+plugins {
+    kotlin("jvm") version com.unciv.build.BuildConfig.kotlinVersion
+    id("org.jetbrains.dokka") version (com.unciv.build.BuildConfig.dokkaVersion) apply false
+}
 
 // You'll still get kotlin-reflect-1.3.70.jar in your classpath, but will no longer be used
 configurations.all { resolutionStrategy {
@@ -31,6 +36,8 @@ buildscript {
 
         // This is for wrapping the .jar file into a standalone executable
         classpath("com.github.anuken:packr:-SNAPSHOT")
+
+        classpath("org.jetbrains.dokka:dokka-gradle-plugin:${com.unciv.build.BuildConfig.dokkaVersion}")
     }
 }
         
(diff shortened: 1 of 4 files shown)


yairm210:
To be honest I'm not sure what problem you're trying to solve with this
Searching functions in dokka feels similar to searching the functions in github, what advantages does it confer?

will-ca:
I don't think Github is able to search for names within files, or generate links between different documentation pages based on type hierarchies and such? It's at least as big a difference as the difference between using a text editor and using an IDE.

Though that said, mainly I'm thinking ahead to mod scripting API stuff. Being able to quickly see at a glance what names are available, and click on the links to explore the API, without having to read Kotlin or figure out what file a class is from— Especially if the scripting language is Python or JS, as doc page versus Kotlin source makes a much larger difference when you're not fluent in Kotlin. [E.G.: ScriptingScope](https://will-ca.github.io/Unciv/dokka/core/com.unciv.scripting.api/-scripting-scope/). For existing uses I actually think Android Studio does a much better job.

will-ca:
Closing for now.

will-ca:
Pretty big benefit IMO would be unifying the [online JSON docs](https://github.com/yairm210/unciv/wiki/Civilization-related-JSON-files#nationsjson) with in-code and IDE-accessible KDocs.

Currently there's like three or four different places to check just to figure out what a single property does:

1. Code comments/KDocs.
2. Use in code.
3. Use in stock rulesets/mods.
4. Wiki.

Personally, I find this makes it a lot more trouble to integrate code with existing behaviour and avoid "adding" already-existing features.

(Not reopening though because I'm not ready to maintain this PR.)

PR #5804, opened 2021-12-16
Add new universal crash handlers and error reporting screen.

will-ca:
image

image

Gives user explanation.

May make crashes easier to report.

Should catch *most* errors and exceptions, I think. Wraps `Stage`'s event handling, and `Game`'s rendering, the two of which I think most code goes through. Prominent remaining gap is exceptions in threads— Crash handling is an extension on function types, so should be easy enough to wrap the lambda passed to threads too.

"Open Issue Tracker" shows a ToastPopup asking user to copy the report first if it hasn't already been copied.

IIRC I also read somewhere Google adjusts Play store rankings based on app stability. So this should turn most strikes in that regard into graceful exits.

Tested both the event and the renderer crash catchers, on both Desktop and AVD.

Ulterior motive: In the scripting API branch, I catch and handle all exceptions that occur during script execution. But it's still possible for misbehaving scripts to modify data in ways that lead to exceptions later on, E.G., by recklessly mutating tile or ruleset data. So, I'm trying to make up for the potential future decrease in stability by making existing crashes better handled and easier to report too.

…Honestly I wasn't aware of, or had forgotten about, the existing CrashController until after I was basically done this PR. But it looks like that one only handles crashes in a specific part of `WorldScreen.nextTurn()`? And it only shows a dialog on the next game launch/save load?

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -42,7 +42,7 @@ You can also report this on the issue tracker. =
 Copy = 
 Error report copied. = 
 Open Issue Tracker = 
-Please copy the error message first. = 
+Please copy the error report first. = 
 Close Unciv = 
 
 # Buildings
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -34,6 +34,17 @@ See your stats breakdown!\nEnter the Overview screen (top right corner) >\nClick
 Oh no! It looks like something went DISASTROUSLY wrong! This is ABSOLUTELY not supposed to happen! Please send me (yairm210@hotmail.com) an email with the game information (menu -> save game -> copy game info -> paste into email) and I'll try to fix it as fast as I can! = 
 Oh no! It looks like something went DISASTROUSLY wrong! This is ABSOLUTELY not supposed to happen! Please send us an report and we'll try to fix it as fast as we can! = 
 
+# Crash screen
+
+An unrecoverable error has occurred in Unciv: = 
+If this keeps happening, you can try disabling mods. = 
+You can also report this on the issue tracker. = 
+Copy = 
+Error report copied. = 
+Open Issue Tracker = 
+Please copy the error message first. = 
+Close Unciv = 
+
 # Buildings
 
 Unsellable = 
--- a/core/src/com/unciv/CrashScreen.kt
+++ b/core/src/com/unciv/CrashScreen.kt
@@ -26,6 +26,7 @@ class CrashScreen(message: String): BaseScreen() {
     val text = generateReportHeader() + message
 
     var copied = false
+        private set
 
     fun generateReportHeader(): String {
         return """
@@ -69,8 +70,8 @@ class CrashScreen(message: String): BaseScreen() {
             ).padTop(15f)
                 .width(stage.width)
                 .row()
-            layoutTable.add(Table().also { table ->
-                table.add(
+            layoutTable.add(Table().also { buttonsTable ->
+                buttonsTable.add(
                     "Copy".toButton()
                         .onClick {
                             Gdx.app.clipboard.contents = text
@@ -81,7 +82,7 @@ class CrashScreen(message: String): BaseScreen() {
                             )
                         }
                 ).pad(10f)
-                table.add(
+                buttonsTable.add(
                     "Open Issue Tracker".toButton(icon = "OtherIcons/Link")
                         .onClick {
                             if (copied) {
@@ -96,10 +97,10 @@ class CrashScreen(message: String): BaseScreen() {
                 ).pad(10f).also {
                     if (isCrampedPortrait()) {
                         it.row()
-                        table.add()
+                        buttonsTable.add()
                     }
                 }
-                table.add(
+                buttonsTable.add(
                     "Close Unciv".toButton()
                         .onClick {
                             Gdx.app.exit()
--- a/core/src/com/unciv/CrashScreen.kt
+++ b/core/src/com/unciv/CrashScreen.kt
@@ -13,6 +13,7 @@ import java.io.StringWriter
 
 /** Screen to crash to when an otherwise unhandled exception or error is thrown. */
 class CrashScreen(message: String): BaseScreen() {
+    constructor(exception: Throwable): this(exception.stringify())
 
     private companion object {
         fun Throwable.stringify(): String {
@@ -22,10 +23,7 @@ class CrashScreen(message: String): BaseScreen() {
         }
     }
 
-    constructor(exception: Throwable): this(exception.stringify())
-
     val text = generateReportHeader() + message
-
     var copied = false
         private set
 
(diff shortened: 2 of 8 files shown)


SomeTroglodyte:
> existing CrashController

My first association reading the mail subject for this. Yes please overhaul / analyze the background of that one. A solid last-ditch exception handler would be a very good thing.

That reminds me, not directly related: I repeatedly wished non-nerds simply running us from g00 play store could provide console logs - I always thought you can't get at them without root or a full debugging environment. I recently learned there's an intermediate option - still not within the mental capabilities of the average g00kl follower, but fwiw there's simply a pm command to add the READ_LOGS permission to a package. Or - install LineageOS 18.1 without Magisk (= NO root), then get the FOSS log viewer "MatLog" from F-Droid - it will tell you how. And it works. You still need adb and enable debugging, but it's once-only and can be done without Android Studio. Something to store in the back of our brains just in case.

android/assets/jsons/translations/template.properties
```diff
@@ -34,6 +34,17 @@ See your stats breakdown!\nEnter the Overview screen (top right corner) >\nClick
 Oh no! It looks like something went DISASTROUSLY wrong! This is ABSOLUTELY not supposed to happen! Please send me (yairm210@hotmail.com) an email with the game information (menu -> save game -> copy game info -> paste into email) and I'll try to fix it as fast as I can! = 
 Oh no! It looks like something went DISASTROUSLY wrong! This is ABSOLUTELY not supposed to happen! Please send us an report and we'll try to fix it as fast as we can! = 
 
+# Crash screen
+
+An unrecoverable error has occurred in Unciv: = 
+If this keeps happening, you can try disabling mods. = 
+You can also report this on the issue tracker. = 
+Copy =
```
yairm210, line 42: "Copy to clipboard" already exists, we can use that

core/src/com/unciv/CrashScreen.kt
```diff
@@ -0,0 +1,113 @@
...
+                        buttonsTable.add()
+                    }
+                }
+                buttonsTable.add(
+                    "Close Unciv".toButton()
+                        .onClick {
+                            Gdx.app.exit()
+                        }
+                ).pad(10f)
+            }).padTop(10f)
```
yairm210, line 109: This is kind of horrendous to read. 
Please use `val`s to separate the parts, call them by name, activate functions etc

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -93,6 +93,16 @@ fun Actor.addBorder(size:Float, color: Color, expandCell:Boolean = false): Table
...
 
+/** Wrap an [Actor] in a [Group] of a given size */
+fun Actor.sizeWrapped(x: Float, y: Float) = Group().also {
+    it.isTransform =
+        false // performance helper - nothing here is rotated or scaled
+    it.setSize(x, y)
+    this.setSize(x, y)
+    this.center(it)
+    it.addActor(this)
+}
```
yairm210, line 104: If you're using both `it` and `this` in the same mini lambda it gets confusing, how about a regular function where you change the actor and then wrap it in a group and return the group?

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -284,3 +309,36 @@ object UncivDateFormat {
...
+            if (postToMainThread) {
+                Gdx.app.postRunnable {
+                    UncivGame.Current.setScreen(CrashScreen(e))
+                }
+            } else {
+                UncivGame.Current.setScreen(CrashScreen(e))
+            }
+            null
+        }
+    }
```
yairm210, line 335: Why not always post to main thread? That way you don't need to worry about whether you're running from the main thread or not

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -284,3 +309,36 @@ object UncivDateFormat {
...
+
+/**
+ * Returns a wrapped a version of a Unit-returning function which safely crashes the game to [CrashScreen] if an exception or error is thrown.
+ *
+ * @param postToMainThread Whether the [CrashScreen] should be opened by posting a runnable to the main thread, instead of directly. Set this to true if the function is going to run on any thread other than the main loop.
+ * */
+fun (() -> Unit).wrapCrashHandlingUnit(postToMainThread: Boolean = false): () -> Unit
+    = with(this.wrapCrashHandling(postToMainThread)) // Don't instantiate a new lambda every time.
+        { { this() ?: Unit } }
```
yairm210, line 344: Sorry this is just unreadable.

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
will-ca: I was going to argue for the original code, but then I got confused myself, so yeah lol. Changed.

android/assets/jsons/translations/template.properties
will-ca: I think "Copy to clipboard" is too long? The current buttons have a nice combined length IMO; See the screenshot.

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
will-ca: `postRunnable` goes through a different call path than calling directly. Since this is meant to be done after a crash, where the application state is undefined, I'm trying to play things safe by avoiding different lower-level behaviour where possible.

E.G.:
- `postRunnable`, I think, means that whatever called the wrapped function will get and process its returned result *before* the crash is later handled in the next game loop. In such a case, the return result is always null, so the behaviour after the return during this game loop is kinda undefined and in theory I think it's safer to handle the crash right away when possible than wait until the next frame.
- Usually everything on the main thread goes through either `Stage.[touchUp/touchDown/someEvent]()` or `Game.render()`. The stuff given to `postRunnable` skips that. I don't know of any specific problems that would cause, but why not play it safe by minimizing the number of different GDX API paths we're using when the game is already in an undefined state?

will-ca: That said, having taken a closer at the stack, I think the benefit might be marginal if any… With LWJGL3, at least, the calls for `postRunnable` are a subset of those used for events and rendering, diverging at `Application.loop()` and jumping straight into the lambda.

LMK If I should always post, given that.

android/assets/jsons/translations/template.properties
yairm210: K fair enough

core/src/com/unciv/CrashScreen.kt
```diff
@@ -0,0 +1,160 @@
...
+import com.badlogic.gdx.scenes.scene2d.ui.Label
+import com.badlogic.gdx.scenes.scene2d.ui.Table
+import com.badlogic.gdx.utils.Align
+import com.unciv.models.ruleset.RulesetCache
+import com.unciv.ui.utils.*
+import java.io.PrintWriter
+import java.io.StringWriter
+
+/** Screen to crash to when an otherwise unhandled exception or error is thrown. */
+class CrashScreen(message: String): BaseScreen() {
```
yairm210, line 15: Much better!
Still, this function is about 25% longer than in needs to be (in lines) because of all the unnecessary 'enter's, which makes it harder to parse
So please remove gratuitous line breaks :)

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -93,6 +93,21 @@ fun Actor.addBorder(size:Float, color: Color, expandCell:Boolean = false): Table
     return table
 }
 
+/** Wrap an [Actor] in a [Group] of a given size */
+fun Actor.sizeWrapped(x: Float, y: Float): Group {
+    val wrapper = Group().apply {
+        isTransform = false // performance helper - nothing here is rotated or scaled
+        setSize(x, y)
+    }
+    this.apply {
```
yairm210, line 102: ...why would you need this.apply?

will-ca: Oh. Lmao. …I stole/moved this from existing code, so in my fear of breaking anything, I tried to preserve the existing call stack apparently without paying any attention at all to what it was actually doing. Oops.

core/src/com/unciv/CrashScreen.kt
will-ca: Funny. I actually thought those would improve readability by separating different elements… IIRC My first/only other GH contributions were to a project with lots of line breaks— I think I went overboard with putting every first function's argument on its own line.

android/assets/jsons/translations/template.properties
will-ca: FWIW I think there's a semantic/applicability difference too… "Copy" when it's visually obvious that there's a text field somewhere, "Copy to clipboard" when the user doesn't think of the current UI context as displaying text (E.G. save files).

I would also want to use "Copy" in at least one other place in my scripting API branch (if I ever get that done), for example.

yairm210:
There is a major problem with this screen, in that it hides all crashes from the Google analytics
That's one of my major sources of bug locating
For Android, at least, closing the screen should throw the exception we caught in the first place, so the analytics will display it

PR #5814, opened 2021-12-18
Refactor MiniMapHolder's little green map overlay toggle icons.

will-ca:
There was a lot of repetition.

Also removed a redundant `yieldImage.actor.color.a = if (settings.showTileYields) 1f else 0.5f` from the `onClick`s, which wasn't needed because the `worldScreen.shouldUpdate` was already propagating to `MiniMapHolder.update()`.

…I can't believe the line count actually increased??? Ah well, there's some KDocs, and the repetition and complexity per added button is way down.


yairm210:
There's a lot of line spaces where you really don't need them, that probably contributed to the line count
But this looks fine

SomeTroglodyte:
Didn't I do something similar a while back that got ultimately reverted? Might be worth checking what happened why?

PR #5815, opened 2021-12-18
Fixed a bug where an empty improvement picker screen could open

xlenstra:
This happened as for some reason unbuildable improvements (academy, landmark, city center, ancient ruins, etc.) were added to the list of buildable improvements in the action button check, but excluded in the picker screen. This PR fixes this, by also excluding these from the action button check.


yairm210:
If we want the same results in two different places, shouldn't we be using the same function in both places, not duplicating the function?

xlenstra:
True, I'll look into it later today

xlenstra:
Hotfixed it for now, the real solution should probably be to merge all checks made everywhere into this function. But as this is mess right now anyway with fishing boats and other things, I'd rather not touch it without completely overhauling it, for which I have no time today

PR #5821, opened 2021-12-18
Changed remove heresy sound effect

xlenstra:
Better save than sorry


SomeTroglodyte:
The more torture-chamber-y one I did back then was better :clown:
Seriously: Wouldn't the libre spirit be server even better if the credits made the association to the actual file a little easier? For some of our credits the correlation in-game use - file is obvious, for some it's not. Worth improving?

will-ca:
> Seriously: Wouldn't the libre spirit be server even better if the credits made the association to the actual file a little easier? For some of our credits the correlation in-game use - file is obvious, for some it's not. Worth improving?

I considered using a one-off Python script to try to convert `Credits.md` into a `Credits.json`/`Credits.yaml` when I compiled the list of technically illegal files. Something like:

```
-	Name: Image Icon
	Assets:
	-	android/assets/Images/File.png
	-	android/assets/Images/FileWithBackground.png
	-	android/assets/ImageSizes/File/*
	Source: https://www.example.com/icons/imageicon.html
	Author: Sam Jones
	License: CC-BY-4.0
```

`Credits.md` looked structured enough to at least get some of those fields filled in… The URL, definitely, and from there license, author, and original title looked easily scrapable from both Freesound and TheNounProject. Actual association to the asset path would have been the hardest part, but some sort of fuzzy matching can probably get it right for most cases and flag potential problems with a comment for manual review.

And then there would of course also be a build-time test to make sure that all files in the assets directory have at least one matching entry in the credits file.

PR #5822, opened 2021-12-18
Fixes warnings in `desktop-file-validate`

touhidurrr:
Trying to validate this desktop entry with `desktop-file-validate` (default tool for desktop entry validation) shows the following warnings.
```bash
~$ desktop-file-validate unciv.desktop
unciv.desktop: warning: key "Encoding" in group "Desktop Entry" is deprecated
unciv.desktop: warning: boolean key "Terminal" in group "Desktop Entry" has value "0", which is deprecated: boolean values should be "false" or "true"
```
Thus deprecated lines were removed or edited.


touhidurrr:
My first pr so be kind (emotional blackmail, lol).
I tested the updated desktop entry. It works. The lines that were removed were not necessary after all.

Issue #3910, opened 2021-05-10
Show attacked city or unit animation. 


AdityaMH:
**Is your feature request related to a problem? Please describe.**
When my troops or my units are attacked by the enemy, I am confused as to which area they are attacking, even though there is a notification on the right screen. If they are destroyed, it still remains confused as well. 

**Describe the solution you'd like**
Please display the screen towards the unit or city that is attacked when it's the enemy's turn to play, even if it's just an arrow direction icon. 

**Describe alternatives you've considered**
Or it is better to display the amount of health lost above unit head or city like in Civ5 Mobile.

PR #5824, opened 2021-12-19
Show arrows on map for unit actions.

### LINKED ISSUE #5431: Show hexagons that will be occupied by units
The problem is that when I'm moving a lot of units to a well-defined positions, then I need to see which hexagons already occupied to not send two units on the same hexagon. I need this for prevent superfluous actions when units will arrive to their positions.

1. The game shouldn't allow to send two units of the same type (military or civil, technically, a Great Generals are civil units too) to one hexagon.
2. Additionally, the game should show hexagons that will be occupied by units, for example, using temporarily white circles.
3. Finally, it could be possible to add userful function. When you tap on the occupied hexagon two times, then the game should move the camera to the unit that is moving to this hexagon.

Image_0
Image_1

will-ca:
This PR includes only code. #5825 includes the code in this PR plus assets.

---

Adds `.addArrow()` and `.resetArrows()` methods to `TileGroup()`, similar to the existing `.showCircle()` and `.hideCircle()` methods.

Uses them to show past and planned unit movements and attacks. Closes #5431, and provides (IMO) easier-to-use solution for many cases of #3910.

Toggled by new button next to minimap. Adds "Footprints" icon to repository, but unused because I decided the existing "Movement" icon is clearer— It's there and I tweaked it for legibility if anyone wants to include it in the future.

---

I've realized since opening this PR these arrows are also basically unit attack and movement graphics. So mods that override them can add bullets, arrows, explosions, footprints, etc.— Followup PR is planned to make sure they always point the right way up, and let arrow variants be selected based on `UnitType` and `BaseUnit` when available.

---

image

image

image

Melee withdrawal:

image

Maximum number of movement arrows on one unit (Withdrawal+Move+IsMovingTo):

image

---

A list is kept of recent positions for each unit, and how they got there.

Arrows are drawn from each position in the list until the next, and then an arrow is drawn from the last position in the list to the unit's current position. The arrows use different colours/textures for different types of movement. (Regular, attack, withdrawal, paratroop, etc.)

A new entry is added to the list at the start and at the end of the unit's turn. The list can never contain more than two items, corresponding to one segment.

So in the vast majority of cases, each unit can have only at most one arrow behind it, as due to the order of `.startTurn()`, `.endTurn()`, and unit movement, either the two positions on the list will be the same (I.E. during the unit's turn, relevant for player units), or the last position on the list will be the same as the unit's current position (I.E. after the unit's turn, relevant for foreign units).

In exceptional circumstances where a unit moves in between turns, it can have two arrows behind it. I think that applies sans mods to only Caravels/Destroyers, and maybe a couple select land units, that can withdraw from melee combat. …Actually open borders ending would do it too.

---

An additional arrow of a separate style can be drawn pointing to where a unit is currently headed.

---

In addition, a red arrow of a separate style is shown for each attack that a unit performed since the start of its most recent turn (including this one), pointing from the unit's current tile to the tile it attacked.

---

Arrows of zero length are always skipped.

---

The arrow image is wider than the default textures need it to be, so mods have room to add their own aesthetic. E.G. Picture tank treads, foot prints... Maybe a teleportation portal at each end for "paratroopers" in a magic-themed mod— Rocket smoke or wake turbulence for attacks? Or, more abstractly, maybe just a wider flare at the ends for visual prominence.

I originally wanted to tile the arrow image, but honestly, stretching is not only easier and probably more performant, but also makes it easier to make the ends look clean, allows mods to control how the path will look relative to the whole path (E.G. Fade in at start, fade out at end), and gives a way to visually assess how far away the unit is moving from without saccading all the way to the end, just by seeing how stretched the arrow is.

~~I tried to make the default images visible on diverse terrain types, and have visual prominence proportional to importance. E.G. The edges have constant hue for colour-coding and visibility in most cases, while the center has either `0` or `1` value with partial alpha so it should still be visible even if the background has the same colour, and important movement types (attack, escape) have high brightness and alpha in their value-extreme centers while normal movement types have more neutral tones.~~ I was going to do this, but honestly the flat ones look fine with the default tileset and GUI, and it would be a whole new development effort. …I did, however, check the flat arrows I made both in RGB and perceptually to make sure they're sufficiently distinct from the closest terrain types while looking alright (…in multiple channels, actually… Though I wasn't aiming for it, I think that might make this colourblind-accessible— Quick desaturate in GIMP seems to confirm that— Arrow types could still get mixed up, but very different arrow types also have different silhouettes).

---

Default arrow meanings:

- Blue: Regular move (past, behind unit).
- Green: Melee withdrawal (past, behind unit).
- Red: Melee attacked and moved into tile (past, behind unit).
- Braided: Teleport (Paradrop, expulsion from foreign borders, air rebase) (past, behind unit).

- Thin, red, beaded: Attacked, but either didn't immediately move into tile or then moved away from tile (past, away from unit).
- Thin, white: Is moving (future, away from unit).

---

If the overlay is enabled:
- All past and planned movement for the player civilization is visible.
- Past movement for foreign civilizations is visible if the unit is visible and both the starting point and the ending point of all segments in its history are visible.
- All attacks for the current civilization are visible.
- Attacks for foreign civilizations are visible if either their targeted point is visible or the attacker's tile is visible.

Planned future movement for foreign civilizations is never visible, except for in spectator mode.

This is intended to be similar to the information exposed by animations in Civ V.

---

Hidden and logically redundant arrows are a deliberate part of the design:
- When a melee unit defeats another unit and moves into the other unit's tile, two red arrows get added: A thick one for the movement and a thin one for the attack. This is fine because the attack arrow is suppressed due to zero length as long as the unit stays in the tile, and once it moves out of the tile, its movement arrow for the turn gets reset to being blue, so the attack arrow pointing back to the tile is needed again for communicating what the unit did.
- As mentioned above, *most* of the time at least one zero-length arrow will be attempted to be drawn (and silently skipped) per unit, because turn end and turn start both create new steps, even though usually nothing changes, and because a line is always drawn from the most recent step to the current position, even though those are usually the same for foreign civs. The main purpose of this is to support both foreign units and own units with the same code: For your own units, the "turn end to turn start" arrow is zero-length and hidden so the one you see is the "turn start to current position" arrow, and for foreign civs, the "turn end to current position" arrow is zero-length and hidden so the one you see is the "turn start to turn end" arrow before that. It also lets the times when something does change between turns (melee withdrawal, open borders end) be shown.

Known bug: Very long arrows are culled when panning too far away from their starting points. In the third screenshot, there should be a couple more braided white/thin red arrows visible due to stealth bomber attacks and rebases from halfway across the map, for example. This behaviour is rare enough, opaque enough, low-level enough, and probably tied closely enough to performance that I stopped investigating it after checking the most probable causes that I found. Left a `// FIXME` describing both the bug and also why it may not be worthwhile or advisable to fix.

Known bug: For some reason the unit withdrawal arrow isn't shown to the *attacker* on the turn of withdrawal. It's still shown to the defender on their turn, though. So it's a marginal case of an already very marginal feature.

Known edge case: When you expose a unit, you can immediately see where it came from last turn if you can also see the tile it came from, kinda leaking half a turn's worth of information… I don't think that's a big deal, especially considering that there's also information which was exposed in Civ V but not here (E.G. you can't see the paths of units that either entered or emerged from the Fog of War, which I think you could see in Civ V as animations).

Known edge case: I suppose if a ranged unit kills something and then ends its turn on the tile of the thing it killed, then no arrow will be drawn for its attack, because the thin red arrow in front of the unit will be zero-length and the thick red arrow behind a unit is only for `postBattleMoveToAttackedTile`. Whatever. Conflating movement arrows and attack arrows would be far worse in this case— It doesn't make sense and it would spaghettify the code to show a melee movement arrow when there wasn't any melee combat. Likewise if a melee unit kills, moves off, then moves back on. The vast majority of units don't even have enough MP for these scenarios to be possible.

Known limitations: Cities bombard doesn't show arrows— Fine IMO; City bombard has always been distinct from unit attack even in Civ V, and conflating them would be confusing. Arrows are associated with units, so you won't see who attacked you if they were then killed. Not sure if civilian capture shows the thick melee `postBattleMoveToAttackedTile` red arrow— Would only be one LoC to make sure it does, but not sure if it should, or only military combat should.

---

Pretty interesting in spectator mode. You can quickly scan around for red to see where the action is. I'm kinda impressed with how coordinated the AI seems to be— Lots of big rivers of blue and focused rays of red.

Haven't played much yet, but just from testing I feel it may be somewhat more immersive to have this on during gameplay too. Your brain kinda fills in the animations when it sees a movement arrow, and it gives context and can be both foreboding or a relief to see how your enemy's focusing their attacks and watch the red attack-indicating arrows build up during your own turn.


will-ca:
Hm. Since I'm already using `TileSetStrings`, it would only be a couple more lines to let  tilesets define different arrow textures per unit type, defaulting to the unqualified arrows as a fallback. So modders could put tank threads, horse footprints, ocean wakes, etc behind all of the appropriate units, with missile cruisers that actually fire missiles, archers that shoot arrows, knights that stab with their lance, paratroopers that drop down from a stream of jets above them, etc.

will-ca:
@ravignir You make one of the most popular/best/biggest tilesets, right? Curious what you'd think of this (see last comment).

xlenstra:
First thoughts: One thing I am concerned about is how much this could clutter the map. Also, does this work correctly with world wrap?

will-ca:
@xlenstra 

1. Likewise. But really, in that case the screen is *already* cluttered— This just gives the user a new tool to pick through the clutter if they turn it on. Swarms happen at the game data level; Not having a way to visualize their movement just forces the player to hide the clutter in their brains (I.E. by futilely trying to work out what just happened) instead of having it out on the screen.

2. Yes, I believe. My calculation for wrap doesn't feel like the most efficient thing, but it's there and it seems to work.

will-ca:
https://github.com/libgdx/libgdx/wiki/Ninepatches

Hm.

Nah. Maybe later.

core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
```diff
@@ -432,6 +433,56 @@ class WorldMapHolder(internal val worldScreen: WorldScreen, internal val tileMap
...
+
+    /**
+     * Add arrows to show all past and planned movements and attacks, if the the options setting to do so is enabled.
+     *
+     * @param pastVisibleUnits Sequence of [MapUnit]s for which the last turn's movement history can be displayed.
+     * @param targetVisibleUnits Sequence of [MapUnit]s for which the active movement target can be displayed.
+     * @param visibleAttacks Sequence of pairs of MapUnits to the target coordinates of attacks that they have done and can be displayed.
+     * */
+    internal fun updateMovementOverlay(pastVisibleUnits: Sequence<MapUnit>, targetVisibleUnits: Sequence<MapUnit>, visibleAttacks: Sequence<Pair<MapUnit, Vector2>>) {
+        if (UncivGame.Current.settings.showUnitMovements) {
```
yairm210, line 461: Reverse if to exit quickly

core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
```diff
@@ -432,6 +433,56 @@ class WorldMapHolder(internal val worldScreen: WorldScreen, internal val tileMap
...
+            for (unit in pastVisibleUnits) {
+                if (unit.movementMemories.isNotEmpty()) {
```
yairm210, line 463: Same

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -363,6 +364,24 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
         }
     }
 
+    /** @return Whether a unit's past movements should be visible to the player. */
+    private fun isUnitPastVisible(unit: MapUnit): Boolean {
```
yairm210, line 368: Why is this function in the world screen and not in the holder?

yairm210:
I think this can be merged relatively quickly, and then we can ask players on Discord what they think.
This is easily disable-able, and doesn't change core gameplay, so we can afford to be more experimental with this.

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
will-ca: It uses `WorldScreen.viewingCiv`, and it requires `.gameInfo` where it's used since it filters through all units… It needs the same data/properties as `WorldScreen`? Idk. It just didn't feel like a good fit for the map holder to me— Deciding what information to expose is too close to game rules instead of just the map display— Though looking at WorldScreen.kt again, I think I was mistaken to think that's a better place for it…

…Hm. I've moved it into a separate helper class, to keep all the "Is this visualization visible?" logic distinct, à la UnitMovementAlgorithms.

will-ca:
> I think this can be merged relatively quickly, and then we can ask players on Discord what they think.
> This is easily disable-able, and doesn't change core gameplay, so we can afford to be more experimental with this.

Sounds good.

#5825 will be needed too. Once this PR is merged, #5825 should have only asset files in its diff.

yairm210:
Seems to conflict with something

will-ca:
> Seems to conflict with something

Hm? Is it not working?

yairm210:
I fixed the conflict, working now

will-ca:
@yairm210 Cool.

Hm. I will rebase the assets PR, hold on.

AdityaMH:
Should i write issue.
No bug about this feature.
But that arrow also showing for who not supposed to be showed.
For example moved/attacking/attack unit position who need to discovered first by own unit.

will-ca:
@AdityaMH That would be a bug if they're showing when they shouldn't. Screenshot?

AdityaMH:
Before Horsemen move:
IMG_20211226_101529

After horsemen move:
IMG_20211226_101322


In civ5 animation (afaik), enemy moving animation only visible (arrow showing) when enemy unit appear/moving on horseman sight, but i can see that skirmisher movement (arrow) after horsemen move to undiscovered tiles.

P.S.: Actually, that just illustration, real scene can't be screenshoted because different barbarians move.

will-ca:
@AdityaMH Right. It's a known limitation.

> Known edge case: When you expose a unit, you can immediately see where it came from last turn if you can also see the tile it came from, kinda leaking half a turn's worth of information… I don't think that's a big deal, especially considering that there's also information which was exposed in Civ V but not here (E.G. you can't see the paths of units that either entered or emerged from the Fog of War, which I think you could see in Civ V as animations).

I don't see it as having any significant tactical implications. Changing it would involve a lot of complexity for relatively little gain.

AdityaMH:
I think we need create this issue since have known case. And fix it one by one.

And thanks anyway for this great feature.

~~I think next project will be damage or heal total on upper arrow 😆~~

PR #5831, opened 2021-12-21
Update Italian.properties

Smashfanful:


xlenstra:
Placeholder `[amount2]` not found in `Italian` for entry `From a randomly chosen tile [amount] tiles away from the ruins, reveal tiles up to [amount2] tiles away with [amount3]% chance`

PR #5833, opened 2021-12-21
Re-throw exception from CrashScreen on Android.

will-ca:
As discussed in https://github.com/yairm210/Unciv/pull/5804#issuecomment-998672736.

Decided to keep the graceful exit when not on Android, so exit cleanup can still be run.

### DIFF
--- a/core/src/com/unciv/CrashScreen.kt
+++ b/core/src/com/unciv/CrashScreen.kt
@@ -1,11 +1,15 @@
 package com.unciv
 
+import com.badlogic.gdx.Application
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.graphics.Color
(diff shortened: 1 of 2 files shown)


yairm210:
Already done
But I'm still getting very few reports, I think I'll throw the exception in a separate thread, we'll see if that helps

will-ca:
@yairm210 See the comment I left on the `3.18.11-patch1` commit.

Just throwing the exception makes it get caught again by `CrashHandlingStage`. This PR overrides that behaviour by making `CrashScreen` use the normal `Stage` instead.

Throwing in a separate thread would just kill that thread, I think. Not sure if it would result in Google Play report. But it wouldn't close the main Unciv thread/app.

yairm210:
That's the point, I don't *want* to kill the main app, just report it, so a separate thread does the job just fine :)

Issue #5835, opened 2021-12-21
More natural biome borders. (Terrain height in feature draw order?)


will-ca:
I'm not suggesting we add edge-smoothing, texture splatting, or anything so radical to the Unciv core.

But a lot is already *almost* possible with the existing systems.

Proof of concept. Smooth, blended biome edge version of 5Hex:

![image](https://user-images.githubusercontent.com/37680486/146852088-b65d1102-8a84-48bc-b69f-ab704eb0dad7.png)

Mask:

![image](https://user-images.githubusercontent.com/37680486/146852310-03ae4dff-c318-48dd-9d67-b79aec9cf7ae.png)

Proof of concept. Fractal biome edge version of 5Hex:

![image](https://user-images.githubusercontent.com/37680486/146852176-9e9f83e8-9247-4c0b-a6a0-a3e5c6870112.png)

Mask:

![image](https://user-images.githubusercontent.com/37680486/146852270-16c793bc-bd82-42f4-b83d-10b1ff5ac318.png)

These work by making the tile textures go slightly past the hexagon, so there's always an overlapping region where a gradual or fractal transition breaks up the hard edges.

The only problem is that doing it this way makes rivers and terrain features get covered up:

![image](https://user-images.githubusercontent.com/37680486/146852413-e6187634-5e4c-4ab2-a9c2-1247bd2fa910.png)

![image](https://user-images.githubusercontent.com/37680486/146852469-b698a4dc-1091-4daf-9eeb-c4ba2964e419.png)

**Possible solution: Take terrain height into account for ordering terrain features?**

(It would also be useful to have a ~10% margin off to the side of each hexagon texture, but that would pose bigger compatibility concerns.)

will-ca:
Alternate solution found within existing capabilities: Tileset can shift entire entire river image up to northwards of the covered/transition area.

AdityaMH:
And how to add this mask for tileset mods?

PR #3231, opened 2020-10-05
Pixel unit colors based on civilization colors

r3versi:
The idea is to have pixel unit sprites composed of 3 layers (it is not mandatory):
- the **first layer** is the base sprite we are used to: for example for Warrior ("TileSets/FantasyHex/Units/Warrior");
>Warrior
- the **second layer** is a white sprite (can have transparency for _pro_ shading) that will be colored with the nation **inner color** ("TileSets/FantasyHex/Units/Warrior-1");
> Warrior-1
- the **third layer** is just like the second, but it will be colored with the nation **outer color** ("TileSets/FantasyHex/Units/Warrior-2").
> Warrior-2

I implemented a generic method `getLayeredImageColored` that could be useful in the future to support image layers for different kind of entities.

The rules for naming layers are:
* A layer name must start with the same name as the base layer. I.e. `Warrior`, followed by a `-` and any number of digits.
* The layers are ordered alphabetically by their names. 

Here is an example of the effect obtained by using the following two detail layers for warriors:

warriors-shaded

PR #5837, opened 2021-12-21
Randomize textures from variants.

will-ca:
Pretend the letters are slightly different textures:

image

Adds `ImageGetter.getRandomImageVariant(baseFileName: String, seed: Int? = null): String`, which randomly chooses an image variant from all available, following the pattern: `{"baseName", "baseName-A", "baseName-B", …, "baseName-Z"}`.

Short-circuits quickly and should add only minimal overhead if there are no variants available.

Returned result is a new image path String, so it can further be used in other image-getting functions, like `getLayeredImageColored`. (Note the nation-coloured letters above.) (#3231 is also partly inspiration for this.)

Used in all invocations of `.getImage` in `TileGroup`, and only in `TileGroup`.

Hashcode of tile position `Vector2` is used as seed for terrain-y images to make sure the map doesn't change during gameplay.

Added a `var randomnessSeed: Int` to `MapUnit` that is set on construction and preserved on `.clone()` for the same purpose. Currently `@Transient`, deciding a visual effect isn't worth a save file field. I could see myself getting attached to a special Frigate and getting sad when it's changed on a save reload or something, though.

I actually skimmed through all 118 invocations of `.getImage()`, and `TileGroup` was really the only place where it made sense to use random variants. It's just as well to have such a clearly defined scope.

Test files in above screenshot are available as example mod: https://github.com/will-ca/Unciv-Tileset-Variants-Example. If you fork it and tag the fork, I'll delete my repo to keep all the examples consolidated.

---

Why? Abstract reasons: Perfectly repeating patterns are very unnatural and jarring in textures. Randomly mixing just two or three variants is often enough to break up large-scale patterns a lot. Likewise clones don't exist, so slightly different units are always a really cool touch IMO.

Direct reason/Example: I was experimenting with breaking up coastlines/biomes with fractal and blended boundaries (#5835). Some results looked fairly natural. But repeating textures and patterns are still a big immersion breaker— What are the odds of there being fifty of the exact same, very distinctive mountain?

FantasyHex does a really good job hiding the repetition where possible and working it into the visual style otherwise. That gets harder with more realistic tilesets. I notice a lot more repetition with 5Hex. Contemplated CGI-generating a "photorealistic" tileset— Repetition would be a nightmare unless variants are possible.

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -23,6 +23,7 @@ import com.unciv.models.ruleset.unit.UnitType
 import com.unciv.ui.utils.toPercent
 import java.text.DecimalFormat
 import kotlin.math.pow
+import kotlin.random.Random
 
 /**
  * The immutable properties and mutable game state of an individual unit present on the map
@@ -116,6 +117,11 @@ class MapUnit {
     @Transient
     var hasUniqueToBuildImprovements = false    // not canBuildImprovements to avoid confusion
 
+    @Transient
+    /** Random integer for internal use. Shouldn't change. Not guaranteed to be unique. */
+    var randomnessSeed = Random.nextInt()
+        private set
+
     /** civName owning the unit */
     lateinit var owner: String
 
@@ -176,6 +182,7 @@ class MapUnit {
         toReturn.baseUnit = baseUnit
         toReturn.name = name
         toReturn.civInfo = civInfo
+        toReturn.randomnessSeed = randomnessSeed
         toReturn.owner = owner
         toReturn.originalOwner = originalOwner
         toReturn.instanceName = instanceName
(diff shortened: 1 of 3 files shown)


ravignir:
It does not solve the issue for 5Hex without introducing new problems, that is forest/jungle tiles. When these tiles are improved, some trees stay on the same position (for example, compare lumber mill tiles with forest tiles, the trees would be misplaced if the game were to choose completely random texture from the available list).

will-ca:
…So, there's an `Ancient ruins.png` and an `Ancient ruins2.png`, both of which get used… Which means that somewhere, in some way, some kind of functionality something like this is already in the game…

https://github.com/yairm210/Unciv/commit/4fd869dbeee22d6e2f195e0b77183f3c62c39d76

https://github.com/yairm210/Unciv/commit/0c06237570f6f04abd454e94b0bb86ce5431ee7e

I did search both the code and wiki for keywords… Evidently not thoroughly enough.

Well, compare and contrast, I guess. The code looks basically the same, just moved and used in more places. …IMO Letters would be a clearer convention since unit colours already use numbers, and I guess only one texture would have to be renamed…

will-ca:
@ravignir This PR doesn't change how images are retrieved for improvements, though. It just adds more places where different variants can be provided by the tileset.

yairm210:
We already have this - see TileGroup:254

PR #5842, opened 2021-12-24
Added score and time victory

xlenstra:
This PR implements the last missing victory type, time victory.
When enabled, the game will always end after a predetermined amount of turns, which can be chosen at the start of the game.
The winner is determined by whoever has the highest score, which can now also be found in multiple places.

<details> <summary> Pictures </summary>

Game Options with time victory enabled:
image

Game Options with time victory disabled:
image

Max turn options:
image

Rankings screen:
image

Diplomacy table:
image

Victory Screen after winning:
image

</details>

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -308,6 +308,7 @@ Game Options =
 Civilizations = 
 Map Type = 
 Map file = 
+Max Turns = 
 Could not load map! = 
 Invalid map: Area ([area]) does not match saved dimensions ([dimensions]). = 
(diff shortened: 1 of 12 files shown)


SomeTroglodyte:
Did you consider/compare making the max turns a slider, matching city state count? The widget supports mapping discrete values to display strings, which could be used to do a progression - or look under the hood to the Gdx original which I seem to remember also allowed clamping to an array of values or somesuch. I think minimap size is/was done the first way.

Of course, this positively begs for a score breakdown display, but I expect you're planning on one anyway.

PR #5843, opened 2021-12-24
Converted the max turns drop down to a slider, idea courtesy of SomeTroglodyte

xlenstra:
Could probably be changed so the step size changes when more to the right on the slider, but for now this works fine.
Also includes a score breakdown in the stat screen menu

<details> <summary> Pic: </summary>

image

</details>

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -3,10 +3,8 @@ package com.unciv.logic
 import com.unciv.Constants
 import com.unciv.UncivGame
 import com.unciv.logic.BackwardCompatibility.removeMissingModReferences
-import com.unciv.logic.BackwardCompatibility.replaceDiplomacyFlag
 import com.unciv.logic.automation.NextTurnAutomation
 import com.unciv.logic.civilization.*
-import com.unciv.logic.civilization.diplomacy.DiplomacyFlags
 import com.unciv.logic.city.CityInfo
 import com.unciv.logic.map.TileInfo
 import com.unciv.logic.map.TileMap
@@ -290,7 +288,7 @@ class GameInfo {
         
         val winningCiv = civilizations
             .filter { it.isMajorCiv() && !it.isSpectator() && !it.isBarbarian() }
-            .maxByOrNull { it.calculateScore() } 
+            .maxByOrNull { it.calculateScoreBreakdown().values.sum() } 
             ?: return // Are there no civs left?
         
         winningCiv.victoryManager.hasWonTimeVictory = true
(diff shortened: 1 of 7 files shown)


SomeTroglodyte:
<details><summary>Should you wish your original allowed values...</summary>

```patch
Index: core/src/com/unciv/ui/utils/UncivSlider.kt
IDEA additional info:
<+>UTF-8
diff --git a/core/src/com/unciv/ui/utils/UncivSlider.kt b/core/src/com/unciv/ui/utils/UncivSlider.kt
--- a/core/src/com/unciv/ui/utils/UncivSlider.kt	(revision cd8b60ff3d798574fa09841a121417706498db6b)
+++ b/core/src/com/unciv/ui/utils/UncivSlider.kt	(date 1640385510201)
@@ -73,15 +73,21 @@
             slider.stepSize = value
             stepChanged()
         }
+    /** Returns true if the slider is being dragged. */
     val isDragging: Boolean
         get() = slider.isDragging
     var isDisabled: Boolean
         get() = slider.isDisabled
         set(value) { slider.isDisabled = value }
+    /** Sets the range of this progress bar. The progress bar's current value is clamped to the range. */
     fun setRange(min: Float, max: Float) {
         slider.setRange(min, max)
         setPlusMinusEnabled()
     }
+    /** Will make this progress bar snap to the specified values, if the knob is within the threshold. */
+    fun setSnapToValues(values: FloatArray?, threshold: Float) {
+        slider.setSnapToValues(values, threshold)
+    }
 
     // Value tip format
     var tipFormat = "%.1f"
Index: core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
IDEA additional info:
<+>UTF-8
diff --git a/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt b/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
--- a/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt	(revision cd8b60ff3d798574fa09841a121417706498db6b)
+++ b/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt	(date 1640386403279)
@@ -136,11 +136,13 @@
         if (!gameParameters.victoryTypes.contains(VictoryType.Time)) return null
 
         add("{Max Turns}:".toLabel()).left().expandX()
-        val slider = UncivSlider(250f, 1500f, 25f) {
+        val slider = UncivSlider(250f, 2000f, 25f) {
             gameParameters.maxTurns = it.toInt()
         }
         slider.permanentTip = true
         slider.isDisabled = locked
+        val snapValues = floatArrayOf(250f,300f,350f,400f,450f,500f,550f,600f,650f,700f,750f,800f,900f,1000f,1250f,1500f,2000f)
+        slider.setSnapToValues(snapValues, 250f)
         slider.value = gameParameters.maxTurns.toFloat()
         return slider
     }
```
</details>

Should work with Studio `Git-Patch-Apply Patch from Clipboard`.

xlenstra:
> 

Thanks!

PR #5336, opened 2021-09-27
Change NationIcon for 4 nation

AdityaMH:

PR #5844, opened 2021-12-24
Better Icon

AdityaMH:
Don't forget to update png, and atlas


SomeTroglodyte:
I like the "old" one better - yes it's quite cluttered, but it's unequivocally Maya. The RekMod one could just as well fit the Olmecs, Toltecs, or just about all of the mesoamerican cultures...

ravignir:
It is not my icon. It is an Icon from civ5 expansion mod.

xlenstra:
If that is the case, we don't have explicit permission from the author to use it, and thus probably legally cannot use it either way

AdityaMH:
> It is not my icon. It is an Icon from civ5 expansion mod.

And also @k4zoo never touch his mod too.
Also we need change this too.
https://github.com/yairm210/Unciv/blob/master/android/Images.Construction/UnitIcons/Horse%20Archer.png

will-ca:
@xlenstra I looked at it this morning. It's actually very likely a blurred version of an exact trace from Civ V, and thus not legally usable for us even if the mod author tried to give permission. In fact I think this is probably a rare case where the icon design itself is distinctive/novel enough to have copyright protection for Firaxis, so even if we remade it from scratch we still couldn't use it.

Options:
- Keep current.
- Prove that Firaxis *don't* own the design— E.G. By finding an identical museum artefect— And then remake it from scratch to clean the specific file's copyright stuff.
- Make a similar icon that draws inspiration from it while being novel, distinct, or generic enough to not infringe the copyright.

xlenstra:
In my personal opinion, the current image is fine and I don't see the extra value of creating an entirely new one. I'll close this for that reason, if more discussion on this pops up, I'm fine with reopening it again.

This also reminds me that a few months ago, a similar PR was accepted which also changed a few icons to their RekMod versions, in #5336. That would be the following images:
[China](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/China.png), [Inca](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/Inca.png), [Iroquois](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/Iroquois.png), [Songhai](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/Songhai.png).
@ravignir, are those also not your original works? If so and we also don't know the actual author, these should probably for legal reasons* either be reverted or replaced by some new icons.

*Note: I am not a lawyer

ravignir:
> In my personal opinion, the current image is fine and I don't see the extra value of creating an entirely new one. I'll close this for that reason, if more discussion on this pops up, I'm fine with reopening it again.
> 
> This also reminds me that a few months ago, a similar PR was accepted which also changed a few icons to their RekMod versions, in #5336. That would be the following images: [China](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/China.png), [Inca](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/Inca.png), [Iroquois](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/Iroquois.png), [Songhai](https://github.com/ravignir/5Hex-Tileset/blob/master/Images/NationIcons/Songhai.png). @ravignir, are those also not your original works? If so and we also don't know the actual author, these should probably for legal reasons* either be reverted or replaced by some new icons.
> 
> *Note: I am not a lawyer

China and Inca are 100% my original works. Songhai and Iroquois are heavily based of Civ5 icons (so i can see it might not be ok to use them).

will-ca:
I couldn't find any pre-existing uses of the Iroquois and Songhai symbols before Civ V. In my not-a-lawyer opinion, that means that the designs are owned by Firaxis and probably can't be used even if modified or re-created.

E.G. Precedent: https://www.martindale.com/legal-news/article_hinshaw-culbertson-llp_931362.htm

The Iroquois, at least, apparently have a pretty cool-looking IRL symbol that dates back to well into Public Domain times, with an .SVG already available on Wikimedia Commons:

https://en.wikipedia.org/wiki/Flag_of_the_Iroquois_Confederacy

That does raise the question of why Firaxis decided to make their own symbol— Would there be non-IP, cultural issues with using this symbol?

The Songhai seem trickier to me. I can't find anything on their symbology. Their predecessor and successor states both have banners, but there's not so much as a single Songhai coin from Googling.

…It looks like Firaxis may have mashed together a bunch of geometric motifs from traditional West African textiles: https://forums.civfanatics.com/threads/what-do-all-of-the-civ-insignias-mean.443101/page-4#post-11611058

So could make a new design using those shapes.

AdityaMH:
Think about it, America, Arabian, Aztec, Byzantium, Japan has same icon as Firaxis have. Should we change it too?

will-ca:
@AdityaMH No. "Same as Firaxis" isn't what matters. Problem is copyright status. The ones you mentioned are mostly pre-existing historical or generic symbols that Firaxis has no ownership over, and were sourced from The Noun Project, where the authors of the files used have explicitly granted a license. Neither of these facts are true for the Iroquois and Songhai.

AdityaMH:
That's not i mean. thenounproject created by unknown user.
should we ask the creator where/how they find or created that icon?
Or should our member place all icon to thenounproject for legit copyright?

also, i not lawyer too.

ravignir:
Iroquois
Songhai
What about these?

xlenstra:
Top one looks good, the bottom icon kind of reminds me of the confederate flag, though that might be just me :)

will-ca:
Looks pretty cool IMO. Creative, original, distinct, and with historical inspiration.

Maybe a bit elaborate— Thought not any more so than some of the existing icons. Are they good zoomed out on a map?

Didn't see the Confederate flag myself, though I guess there is a certain resemblance. Maybe with just a dot and not the star in the middle?

PR #5845, opened 2021-12-24
Fixed a rare diplomacy voting bug in one-more-turn mode

xlenstra:
After conquering all other major civs, the check for voting for the diplomatic victory would not stop, which could soft-lock the game in a state where players were required to vote but couldn't vote as there was no one to vote for. This PR fixes this bug.

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -928,6 +928,8 @@ class CivilizationInfo {
     fun mayVoteForDiplomaticVictory() =
         getTurnsTillNextDiplomaticVote() == 0 
         && civName !in gameInfo.diplomaticVictoryVotesCast.keys
+        // Only vote if there is someone to vote for, may happen in one-more-turn mode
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -928,17 +928,21 @@ class CivilizationInfo {
...
     fun diplomaticVoteForCiv(chosenCivName: String?) {
         if (chosenCivName != null) gameInfo.diplomaticVictoryVotesCast[civName] = chosenCivName
     }
 
     fun shouldShowDiplomaticVotingResults() =
          flagsCountdown[CivFlags.ShowDiplomaticVotingResults.name] == 0
+         && gameInfo.civilizations.any { it.isMajorCiv() && !it.isDefeated() && it != this }
 
     // Yes, this is the same function as above, but with a different use case so it has a different name.
     fun shouldCheckForDiplomaticVictory() =
```
yairm210, line 943: So can't we just return the results of showDiplomatic... etc?

yairm210, line 943: So can't we just return the results of showDiplomatic... etc?

PR #5846, opened 2021-12-25
Fixed a few combat bugs and changed the religions founded label

xlenstra:
Fixes include:
- Fix `Strength \<in [Friendly Land] tiles\>` not working due to the relevant civ not being passed to `tile.matchesFilter`. This fixes Hemeji castle, Nationalsm & some belief not working.
- Fix the temporary autocracy bonus to apply & groundwork for when I'll turn the 'for [amount] turns' into a conditional
- Renamed the 'religions founded' label to 'religions to be founded', as that is the stat it is actually displaying

### DIFF
--- a/core/src/com/unciv/logic/civilization/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/PolicyManager.kt
@@ -47,8 +47,11 @@ class PolicyManager {
     }
 
     fun addPolicyToTransients(policy: Policy) {
-        for (unique in policy.uniqueObjects)
-            policyUniques.addUnique(unique)
+        for (unique in policy.uniqueObjects) {
+            // Should be replaced with a conditional of the same form later
+            if (!unique.text.contains(Regex("for \\[[0-9]*\\] turns")))
+                policyUniques.addUnique(unique)
+        }
     }
 
     fun addCulture(culture: Int) {
--- a/core/src/com/unciv/logic/civilization/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/PolicyManager.kt
@@ -28,7 +28,11 @@ class PolicyManager {
     var shouldOpenPolicyPicker = false
         get() = field && canAdoptPolicy()
 
-
+    // Only instantiate a single value for all policy managers
+    companion object {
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -47,8 +47,11 @@ class PolicyManager {
     }
 
     fun addPolicyToTransients(policy: Policy) {
-        for (unique in policy.uniqueObjects)
-            policyUniques.addUnique(unique)
+        for (unique in policy.uniqueObjects) {
+            // Should be replaced with a conditional of the same form later
+            if (!unique.text.contains(Regex("for \\[[0-9]*\\] turns")))
```
yairm210, line 52: Please make this a static const

yairm210: The regex that is
Making the regex every time we add a policy is very inefficient

SomeTroglodyte: True, Regex objects _can_ and in any case with reuse _should_ have a compiled form of the pattern.
As for the pattern itself - I would recommend standardizing on a "transport" form throughout a project, meaning its presentation to the compiler. I'm slightly in favour of using const val with a raw string everywhere, despite kotlin's raw strings being severely broken. const val would be to enable debugger inspection to make sure the pattern actually contains exactly what's intended.
Lastly, that pattern matches ""for [] turns", probably slightly unintended. `?` = 0 or 1, `+` = 1..∞, `*` = 0..∞, `{n}` = exactly n, `{n,m}` = n..m repetitions...

xlenstra:
Failed checks are in master, I'll wait a few mins before closing & reopening

SomeTroglodyte:
You know, I suspect close/reopen isn't the only/cleanest/fastest way to provoke a fresh check run. There's a re-run jobs button on the Checks tab... Any one of us ever tried that?

xlenstra:
> You know, I suspect close/reopen isn't the only/cleanest/fastest way to provoke a fresh check run. There's a re-run jobs button on the Checks tab... Any one of us ever tried that?

Huh, I never noticed that button before

Issue #5834, opened 2021-12-21
`TileSetConfig` not set on first load.


will-ca:
With `./gradlew desktop:run`.

Freshly downloaded 5Hex is looking way darker in-game than the images suggest it should?

![image](https://user-images.githubusercontent.com/37680486/146807492-188204c4-a469-4b14-aae7-b967aab3a245.png)

Feels like all the RGBs are squared.

The `game.png` seems fine:
![game](https://user-images.githubusercontent.com/37680486/146807816-8d338fd9-8d78-452f-bf41-79ea28a29a89.png)

After a restart it's fine:

![image](https://user-images.githubusercontent.com/37680486/146807944-1c152127-d74a-432b-a1a2-d63cd8fac0ba.png)

Not sure how to reproduce… But I think it's happened before too, but with a JAR.

And Pix5, on Android:

![image](https://user-images.githubusercontent.com/37680486/146980995-112dc8b6-74b2-4753-a700-3af38f3a57e0.png)

How it's supposed to look:

![image](https://user-images.githubusercontent.com/37680486/146981183-caae7193-1ea8-4cb7-a827-0282f808845c.png)

---

Yeah. This is supposed to have the same colours as FantasyHex:

![image](https://user-images.githubusercontent.com/37680486/146993287-1fe1a91c-1c7f-4ca9-8b8b-7e7a76644fef.png)

xlenstra:
If I'm not mistaken, this has been brought up before and the conclusion was that this was the result of the tileset trying to override the default fantasyHex tileset

will-ca:
5Hex doesn't do that AFAICT thoiugh, and this is specifically after Pix5 was updated (https://github.com/AdityaMH/Pix5-Tileset/pull/2) to no longer do it, nor does the tile variants example.

AdityaMH:
> 5Hex doesn't do that AFAICT thoiugh, and this is specifically after Pix5 was updated ([AdityaMH/Pix5-Tileset#2](https://github.com/AdityaMH/Pix5-Tileset/pull/2)) to no longer do it, nor does the tile variants example.

I'm just download strike from github website, place it on folder, start game, greeny occur, delete from mod manager, restart, fixed automatically.

AdityaMH:
> If I'm not mistaken, this has been brought up before and the conclusion was that this was the result of the tileset trying to override the default fantasyHex tileset

Yeah, but @will-ca suggest me to not override.

will-ca:
[...]

PR #5852, opened 2021-12-26
Use tileset fallbacks for missing images.

will-ca:
Lets tilesets specify a `fallbackTileSet` to use when no valid images are found. Defaults to `"FantasyHex"`. `null` disables.

General control flow:
1. Go through all existing checks, fallbacks, etc, to find an image either provided as a file or explicitly specified through configuration in the active tileset.
2. If no image is found or specified anywhere in the active tileset, then look in the fallback tileset. So custom (and even invalid) `ruleVariants`, those giant `when () {}` blocks for pixel units, etc, should all work same as before when images are present.

In "Modding requests": https://github.com/yairm210/Unciv/issues/3242#issuecomment-1000976826

Discussion:
https://github.com/AdityaMH/Pix5-Tileset/pull/2#issuecomment-998692019
https://github.com/AdityaMH/Pix5-Tileset/issues/3#issuecomment-1001080034

(Not just the new arrows, though. Previously, missing roads would also be white boxes, missing tiles would be blank hexagons, missing pixel units would be… Missing, etc.)

Also made the checks for `unitWithSprite` around `TileGroup:680` use `imageExists` instead of `unitIconExists`… I think the previous behaviour might have been incorrect; Unit icons have nothing to do with the value that check is supposed to validate AFAICT.

The fallback is configurable and disableable in order to not break the old "Default" tileset. Modded tilesets can also disable this, if they wish, by setting `"fallbackTileSet": null`.

May have some issues with #5834 in rare cases.

### DIFF
--- a/core/src/com/unciv/ui/tilegroups/TileSetStrings.kt
+++ b/core/src/com/unciv/ui/tilegroups/TileSetStrings.kt
@@ -7,10 +7,10 @@ import com.unciv.models.tilesets.TileSetConfig
 import com.unciv.ui.utils.ImageGetter
 
 /**
- * @param tileSet Name of the tileset. Defaults to active.
+ * @param tileSet Name of the tileset. Defaults to active at time of instantiation.
  * @param fallbackDepth Maximum number of fallback tilesets to try. Used to prevent infinite recursion.
  * */
-class TileSetStrings(val tileSet: String = UncivGame.Current.settings.tileSet, fallbackDepth: Int = 1) {
+class TileSetStrings(tileSet: String = UncivGame.Current.settings.tileSet, fallbackDepth: Int = 1) {
 
     // this is so that when we have 100s of TileGroups, they won't all individually come up with all these strings themselves,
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/ui/tilegroups/TileSetStrings.kt
```diff
@@ -66,4 +72,30 @@ class TileSetStrings {
...
+     * */
+    fun orFallback(image: String, fallbackImage: TileSetStrings.() -> String): String {
+        return if (fallback == null || ImageGetter.imageExists(image))
+            image
+        else
+            fallback!!.run(fallbackImage)
+    }
+    /** @see orFallback */
+    fun orFallback(image: TileSetStrings.() -> String, fallbackImage: TileSetStrings.() -> String)
+            = orFallback(this.run(image), fallbackImage)
```
will-ca, line 96: I wonder if using extension function syntax (`fallback!!.fallbackImage()`, `this.image()`) might allow more compiler optimizations. But `.run()` should make it fairly immune to name collisions.

PR #5854, opened 2021-12-26
Move wiki pages into repository, fix dozens of broken links.

### LINKED ISSUE #5847: Move wiki to `/docs/`?
There have been cases where I've wanted to document features that I was not yet able to find on the wiki, but I've always thought better of it due to the lack of a review or approval system.

Moving the wiki into the main repository would let pull requests be opened against it.

Since GH renders MD, no readability should be lost by moving the wiki pages into `/docs/`. It could even be set up as a GH Page (though I don't see much benefit in that).

Since GH apparently exposes wikis as their own `git` repos, this move may be as simple as cloning the wiki repo and copying its files into the main repo, then disabling the wiki.

Caveats: May break some links. Would get the code tree navigator, instead of the wiki page search (but with equivalent functionality). Couldn't use the "Wiki" link in the navbar.

If approved I'll open a PR for this, I guess.

will-ca:
I have now added a bot/GitHub Action to automatically sync the GH Wiki with `/docs/wiki` after merging a PR.

To use:

1. Open a PR against this repository.
2. Merge the PR into `master`.
3. Leave a comment that says "update wiki".

Example use: https://github.com/will-ca/Unciv/pull/33
(Also shows check for commit privileges, and check that PR has already been merged.)

So we can have the best of both worlds: Editing with IDE support and reviews via PR, plus reading with the Wiki's searchability and presentation.

Example results: https://github.com/will-ca/Unciv/wiki/Home/_history

The bot clobbers the previous state of the wiki with the state of `/docs/wiki`, but it should leave the wiki's previous revision history untouched.

AS gives really helpful linting to autocomplete and repair links in Markdown files, but its URL resolution is slightly different from the GH Wiki with regard to filename extensions and the URL root. The bot does two very simple REGEXes on all Markdown files pushed to the wiki to transform them from the normal scheme to the wiki's. These are explained in comments in the lines with `sed`.

---

Closes #5847. Some reasons why I think this would be better are described in there. Well, workflow of IDE → PR → Review is vastly superior to trying to get things right on the wiki, and GH's Markdown rendering and browsing from the code tree isn't really any worse than from the wiki.

If you don't squash on merge, this should preserve the edit history of all wiki pages.

Another benefit: Android Studio has linting and even autocompletion to detect and fix broken Markdown links and anchors, which can be used if the Markdown files are kept in the repo. (It turns out maybe a third to two thirds of all links were broken. This PR also fixes most of those.)

If the wiki browsing experience is still better, I think a GH workflow could be used to automatically push from the repo into the wiki.


SomeTroglodyte:
Good idea... But why is it based 16 months back and thus sports all those 300 commits?

will-ca:
@SomeTroglodyte To preserve the edit history, and make it transparent what I've changed myself in this PR?

Well, I guess it can probably be squashed into a single commit on merge if the idea of suddenly adding 300 commits into the repository isn't considered desirable for some reason.

yairm210:
I will squash this, since the edit history of the wiki isn't that interesting, tbh.
The main issue is traversability. In the wiki, you have a sidebar where you can get from any wiki page to any wiki page, and here it'll probably be manual searching unless we make a central lndex page with every wiki page starting with an Index link.

will-ca:
@yairm210 Ah, I've got it.

So set the Wiki to read only, and then, whenever anyone makes a change to the docs in the main repo, run:

```bash
CODE_REPO="Unciv"
WIKI_REPO="Unciv.wiki"
rsync -a --delete --exclude '.git*' "$CODE_REPO/docs/wiki/" "$WIKI_REPO/" &&
cd "$WIKI_REPO" &&
for f in *.md; do echo "Formatting $f links."; cat "$f" | sed -ie 's/\(](\..*\)\.md/\1/g' "$f"; done &&
git add . &&
git commit -m 'Synchronize wiki with repository.' &&
git push -f
```

Example result: https://github.com/will-ca/Unciv/wiki

Could set up an automation for it, but honestly putting it in a shell script/batch file would probably be just as convenient.

Only caveat I'm seeing so far: Link resolution works differently. Android Studio expects "`.md`" at the ends of relative URLs, and I think the GH code explorer expects the same, but the wiki requires absolute or relative web URLs without file extension to show rendered pages from Markdown. So the REGEX/`sed` in the middle strips `.md` from relative links before commiting to the wiki's repository. (But it's worth noting that before, there was *no* IDE validation of links at all. So even without the link conversion, this would still be an improvement.)

will-ca:
Will try to include Github automation for updating the Wiki in this PR.

will-ca:
Includes bot to sync wiki with repo now.

yairm210:
You know what let's give this a go
Worst case, we can ditch the wiki altogether and stick with the /docs folder with autogenerated headers as a workaround

yairm210 (approved):

xlenstra:
I suppose this doesn't contain the ~4 commits I made to the wiki since this PR was opened?

will-ca:
> I suppose this doesn't contain the ~4 commits I made to the wiki since this PR was opened?

Nope, nor the change to the version deployment process Yair made.

The Wiki's just a Git repository, so I'll manually copy and paste from the diffs to get those… Actually, I'll do that now and open a follow-up PR.

xlenstra (approved):

PR #5855, opened 2021-12-26
Save attacks per civ for arrows for cities, missiles, dead units.

will-ca:
First of two feature expansions planned for #5824.

This one keeps tracks of recent attacks on a per-civ basis. Used for two things:

1. To show attack arrows for cities' bombardment.
2. To show attack arrows for units that have been killed— Implicitly enables arrows for single-use missile units.

The basic feature might still need time to stabilize or be decided on, I know. But honestly, I had a branch with a WIP commit where I already had most of this written and commented out, so may as well get the a PR ready.

Draft at first just in case I decide to change a name or a signature for forwards compatibility.


AdityaMH:
For killed unit, can used 50~70% tranparent icon/unit pixel.

yairm210:
Sure why not

PR #5860, opened 2021-12-27
Add `interface ICloneable` and extension functions to unify cloning behaviour.

will-ca:
There's at least 34 implementations of `fun clone()` in the codebase, and at least 8 uses of `.map { it.clone() }` or similar.

So let's unify it all with an interface, and a couple extension functions for Collections.


yairm210:
What does this achieve? It seems to me that it just adds unneeded lines

will-ca:
That is... A good point.

I guess mostly in practice it's just the `.copy()` extensions for Collections... Which also, TBF, are much less significant than they felt to me in theory.

PR #5862, opened 2021-12-28
Try to fix #5729 as concurrency issue.

### LINKED ISSUE #5729: Unit maintenance costs errors
```
java.lang.IllegalArgumentException: 
  at java.util.TimSort.mergeHi (TimSort.java:899)
  at java.util.TimSort.mergeAt (TimSort.java:516)
  at java.util.TimSort.mergeForceCollapse (TimSort.java:457)
  at java.util.TimSort.sort (TimSort.java:254)
  at java.util.Arrays.sort (Arrays.java:1492)
  at java.util.ArrayList.sort (ArrayList.java:1470)
  at java.util.Collections.sort (Collections.java:206)
  at kotlin.collections.CollectionsKt__MutableCollectionsJVMKt.sortWith (MutableCollectionsJVM.kt:42)
  at kotlin.sequences.SequencesKt___SequencesKt$sortedWith$1.iterator (_Sequences.kt:638)
  at kotlin.sequences.DropSequence$iterator$1.<init> (Sequences.kt:490)
  at kotlin.sequences.DropSequence.iterator (Sequences.kt:489)
  at com.unciv.logic.civilization.CivInfoStats.getUnitMaintenance (CivInfoStats.kt:55)
  at com.unciv.logic.civilization.CivInfoStats.getStatMapForNextTurn (CivInfoStats.kt:170)
  at com.unciv.logic.civilization.CivilizationInfo.updateStatsForNextTurn (CivilizationInfo.kt:288)
  at com.unciv.logic.city.CityStats.update (CityStats.kt:516)
  at com.unciv.logic.city.CityConstructions.turnsToConstruction (CityConstructions.kt:298)
  at com.unciv.logic.city.CityConstructions.getTurnsToConstructionString$core (CityConstructions.kt:181)
  at com.unciv.ui.cityscreen.CityConstructionsTable.getConstructionButtonDTOs (CityConstructionsTable.kt:168)
  at com.unciv.ui.cityscreen.CityConstructionsTable.access$getConstructionButtonDTOs (CityConstructionsTable.kt:30)
  at com.unciv.ui.cityscreen.CityConstructionsTable$updateAvailableConstructions$1.invoke (CityConstructionsTable.kt:208)
  at com.unciv.ui.cityscreen.CityConstructionsTable$updateAvailableConstructions$1.invoke (CityConstructionsTable.kt:206)
```

will-ca:
See https://github.com/yairm210/Unciv/issues/5729#issuecomment-1001805689 for explanation.

Tries to get rid of anything that could be modified concurrently.

Also cuts down on the loops and Unique param parsing.

Sort is done with a List instead of a Sequence, but I looked at the Kotlin sources and sorting a Sequence just calls `.toMutableList()` anyway.

Did not test for the crash I'm trying to fix, due to lack of reproducer. If the theory behind this patch is correct, then reliably reproducing the crash is basically impossible due to requiring millisecond-accurate coincidence between threads. So I have no way to tell if this will actually fix it; I assume that's what the Google Play logs are for?

(…Hm. Could stick a semaphore before the area of concern to make the timing likelier, I guess.)

Please verify that this will produce the same results as the original code!

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivInfoStats.kt
+++ b/core/src/com/unciv/logic/civilization/CivInfoStats.kt
@@ -35,27 +35,32 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
         // Handle unit maintenance discounts
         // Have to capture global and per-unit
         // Free Garrison already removed above from sequence
-        // Initialize maintenance cost per unit, default 1 aka 100%
+        // Initialize base maintenance cost empire-wide, default 1 aka 100%
+        // Then calculate per-unit discounrts
         // Note all discounts are in the form of -X%, such as -25 for 25% reduction
-        for (unit in unitsToPayFor){
-            unit.maintenance = 1f
+
+        var civWideMaintenance = 1f
+        // Calculate global discounts
+        for (unique in civInfo.getMatchingUniques(UniqueType.UnitMaintenanceDiscountGlobal, StateForConditionals(civInfo))) {
+            civWideMaintenance *= unique.params[0].toPercent()
+        }
+
+        val costsToPay = ArrayList<Float>()
+
+        for (unit in unitsToPayFor) {
+            var unitMaintenance = civWideMaintenance
             for (unique in unit.getMatchingUniques(UniqueType.UnitMaintenanceDiscount)){
-                unit.maintenance *= unique.params[0].toPercent()
+                unitMaintenance *= unique.params[0].toPercent()
             }
+            costsToPay.add(unitMaintenance)
         }
         // Apply global discounts
-        for (unique in civInfo.getMatchingUniques(UniqueType.UnitMaintenanceDiscountGlobal, StateForConditionals(civInfo))) {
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
A superficial look says - seems thought-out well enough. Butting in cuz: `ArrayList<Float>()` - I just met `FloatArray` in Gdx source - did you check as thorough as you usually do / is the prejudice 'first one is pure kotlin, second pure java, and therefore less desirable for typical predicate chaining stuff' close to truth?
Oh, and - `discounrts`

will-ca:
Oh, you think too highly of me. Honestly, I just didn't know about the GDX one, and probably would have (wrongly) assumed it to be unresizable even if I had known. I am inclined to agree with what you said now that you've mentioned it, though— Would probably use the GDX array if I expect to pass it to a GDX method at any point, or maybe if there were more floats and it's more compact.

I did notice the typo at some point. Decided to practice letting go of, or perhaps merely hiding, my perfectionism a little bit rather than fire up the IDE again…

Issue #3997, opened 2021-05-26
Show damage text


AdityaMH:
Please add small damage number, if user confirm attack other units and small heal number if wounded unit going to fortify/sleep. Just like other turn-based strategy games (maybe all turn-based games).
So, player will know damage overall if user attack maori warrior using wounded spearman.

ravignir:
This is actually present in civ5:
![image2](https://user-images.githubusercontent.com/30041793/119681993-cf4d4a00-be42-11eb-8130-fe2476a71a89.png)
![obraz](https://user-images.githubusercontent.com/30041793/119682241-03c10600-be43-11eb-8dc8-20b344bc075f.png)
and whenever any unit heals/deals damage a number is displayed next to it, positive and green when it heals or negative and red when it is dealt damage.

ravignir:
![obraz](https://user-images.githubusercontent.com/30041793/119682620-526ea000-be43-11eb-95b0-9313f67c7d5a.png)

ravignir:
another thing, can we please have city strength always displayed? and when you click enemy unit, make it display its promotion icons.

ravignir:
![obraz](https://user-images.githubusercontent.com/30041793/119683475-f7897880-be43-11eb-80a6-1e769b4f1742.png)

avdstaaij:
Another thing: since damage involves a random factor, it would also be nice to have an option to display the possible damage range, though I'm not sure if the original game has this. The current behavior, where the predicted damage shown also involves the random factor, is certainly _not_ the best UI: it should at least be changed to show the average damage done.

AdityaMH:
> Another thing: since damage involves a random factor, it would also be nice to have an option to display the possible damage range, though I'm not sure if the original game has this. The current behavior, where the predicted damage shown also involves the random factor, is certainly _not_ the best UI: it should at least be changed to show the average damage done.

I think will be better if show text "minor, decent, or dominant" victory like original civ5 (just show for melee battle). Damage range can be incorrect because critical hit can be more high than damage range.

yairm210:
This is currently impossible, and making it possible is more trouble than it's worth.

AdityaMH:
Wow, unciv now can create number text pop-up on above City-State scrolling. No doubt we can create show damage result.

AdityaMH:
#5856

PR #5865, opened 2021-12-28
Show unit damage/heal labels.

will-ca:
#3997

image

With arrows:

image

Idk. I wasn't really seeing the need for this personally, but it seemed like easy pickings.

Could make last turn's health Transient, to not add the save field. Meh, I kinda like the save-reload persistence.

Displays delta since last turn end of the unit. So it doesn't show damage taken by foreign attackers during their own turn, or taken by foreign units that were attacked before their turn. Would probably have to add a history/memory queue for health like I did for the movement arrows to change this, and altered as it's not just the unit's turn but all civs' turns that matter— Not terribly complicated, but also not really worth it. For the purposes of assessing damage taken, healed, and dealt by your own units the current system should already be fine.

Togglable in options. Only for military units.

Unlike Civ, the text is permanent and cumulative over the course of a turn. Could also make it reset on unit movement or selection/click (or even have a thread sleep to reset it after X seconds), I guess.

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -214,6 +214,9 @@ class MapUnit {
     /** Array list of all the tiles that this unit has attacked since the start of its most recent turn. Used in movement arrow overlay. */
     var attacksSinceTurnStart = ArrayList<Vector2>()
 
+    /** This unit's [health] at the end of its last turn. Used to display damage/heal text. */
+    var lastTurnHealth: Int? = null
+
     //region pure functions
     fun clone(): MapUnit {
         val toReturn = MapUnit()
@@ -236,6 +239,7 @@ class MapUnit {
         toReturn.movementMemories = movementMemories.copy()
         toReturn.mostRecentMoveType = mostRecentMoveType
         toReturn.attacksSinceTurnStart = ArrayList(attacksSinceTurnStart.map { Vector2(it) })
+        toReturn.lastTurnHealth = lastTurnHealth
         return toReturn
     }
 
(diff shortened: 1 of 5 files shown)


yairm210:
I think this is more confusing than it is helpful to be honest. The board keeps getting stuffed with more and more information...
Civ V's solution of having it be a one-time popup was correct

SomeTroglodyte:
At first glance looks less pleasing than the arrows... How about...
- Animate them: Fade+Zoom in and out like the tooltips do (Gate with the continuous rendering option?) - modify to add movement upwards on fade-out
- Trigger for fade-in is a click on a notification, maybe add notifications like "28 units have healed since last turn", a few existing ones are obvious
- Timed fade-out, can be preempted by panning or zooming

Doable, but could well end up quite complex. Take it as food for thought. Raises questions like - could the numbers "live" in the notifications instead? That could allow attempts to actually show how X was destroyed by N attacks (thus the movement component to the animation). Con: battles during the user's turn not trivially supported.

... as Yair says. Transient or easy short-term display, at least with another minimap toggle might be nicer.

will-ca:
Hm. Yeah, UI is a pretty big hurdle here.

Civ V can show the popup right after the change in health happens, because usually there's an animation that pans the camera and catches the player's eye. There's generally not an equivalent in Unciv, so when to show the number and what exactly the number means are more nebulous and detached from the event that resulted in the change in health. Have to trigger externally, a single delta event could be shown more than once depending on trigger while others may not be shown at all, the number shown might be (and in this PR, is) a cumulative delta from everything that happened during an arbitrary length of time, etc. Notifications could work, but personally I've never found the association between notifications and map locations to be particularly intuitive in either Civ V or Unciv.

…I don't actually necessarily think numbers, even if done in a robust way, would really communicate any information that isn't already exposed to the player in more readable ways. HP bars, UnitTable, BattleTable, and now the arrows— I usually have a pretty okay idea of what's happening with my units already. Idk. Maybe if a more convenient solution or a greater need presented itself.

AdityaMH:
So, this can be good if damage showed when player attacking and clicking notification.
Can add this feature as experimental

PR #5867, opened 2021-12-28
Let tilesets customize movement/arrow graphics by base unit, era, and unit type.

will-ca:
Second of the two feature expansions planned for #5824.

This one's heavily focused on visuals. It adds highly granular and flexible ways for tilesets to customize the attack, movement, etc. graphics shown in different situations.

Arrows for MapUnits are resolved using the `class ImageAttempter` from #5897 to the first image of the following that exists:
- `Units/[BaseUnit]-[ArrowType]`
- `UnitTypes/[UnitType]-[UnitEra]-[ArrowType]`
- `UnitTypes/[UnitType]-[ArrowType]`
- `Units/[UnitMovementType]Unit-[UnitEra]-[ArrowType]`
- `Units/[UnitMovementType]Unit-[ArrowType]`
- `[ArrowType]`

Arrows for Cities are resolved to the first image of the following that exists:
- `City/[ArrowType]-[Era]` `#TODO: Flip this.`
- `City/[ArrowType]`
- `[ArrowType]`

These paths are all relative to `TileSets/ActiveTileSet/Arrows/`.

E.G.:

```
TileSets/Default/Arrows/Units/Machine Gun-UnitMoved
TileSets/Default/Arrows/UnitTypes/Ranged Gunpowder-Modern era-UnitMoved
TileSets/Default/Arrows/UnitTypes/Ranged Gunpowder-UnitMoved
TileSets/Default/Arrows/Units/LandUnit-Modern era-UnitMoved
TileSets/Default/Arrows/Units/LandUnit-UnitMoved
TileSets/Default/Arrows/UnitMoved
```

`MapArrowType` definitions are somewhat more structured to enable this.

As with #5852, the entire order of candidates in the active tileset is traversed and tried before switching to the fallback tileset.

Arrows pointing to the left side of the screen are flipped upside-down in their local space. This is meant to keep ocean wakes, archers' parabolic arcs, high-flying jet planes, etc, all pointing the right way up.

Code basically done, but draft for now. Would like to prepare an example/reference mod that replaces the arrows for most/all built-in unit(type)s with pixel art.


AdityaMH:
Hmmm...
Maybe add costumize city center by builded building like wall?

will-ca:
@AdityaMH Not relevant to this PR. But interesting idea. Probably less than 10 LoC to check for image per building and add if available… Ordering could be weird.

AdityaMH:
I think would be better if arrow make it same like road route.
So if speed unit type moving too long, not making stretch too long too.

will-ca:
> I think would be better if arrow make it same like road route. So if speed unit type moving too long, not making stretch too long too.

My original plan was similar to that, but reasons for abandoning it were explained in the original PR.

It would be more complicated to program, slower to run, less visually informative, and less customizable for mods.

Additionally, it would make the main purpose of this PR impossible.

PR #5868, opened 2021-12-28
Move BorderImages to tileset. Set origin in `setHexagonImageSize()`.

will-ca:
The border shapes are hex-sized, and really a big part of the look and feel of the map. So this lets tilesets do fun things with them.

E.G. Smooth borders:

image

Bubbly borders:

image

Greasy borders:

image

Sketchy borders:

image

Also:
- Changes border placement code to use `setHexagonImageSize`.
- Changes `setHexagonImageSize` to always put the origin of the image at the center of the tile— This make further rotation and scaling transforms easier to work with, gets rid of some of the "crazy voodoo magic" in the rest `TileGroup`, and will help with another tileset feature I want to try.
- Adds new border image shape "ConcaveConvex". Previously a flipped "ConvexConcave" would be used for this case, but tilesets should be able to create designs with directionality around the border IMO. (See Sketchy borders above for example.)

Planned:
- Do the same for Crosshairs, Circle, Fog, maybe more. These are all as much part of the look and feel of the map as tiles and units, and they're already hex-based, just with duplicate code.
- Probably upload the three border sets above as example mods.


SomeTroglodyte:
> upload the three border sets above

Condolences on the loss of those fingers. Chainsaw accident? :carpentry_saw: (:v: = five beers please for the sawmill crew) :innocent:

yairm210:
OH DAMN this is NEAT

yairm210:
Look at those bubbly borders
B E A Utiful

yairm210:
Unfortunately I was already on the rollout roll when this was PR'd, would have loved to include it in this release :)

yairm210:
Where did you get these images from? Self-made?

will-ca:
Eh, s'all good. It looks like it's already a pretty big release as is.

I whipped up the segments yesterday morning! Took me a couple tries to get everything joining up, but once I did it was just a filter and a node offset for each different style. They're actually not as clean as I would have done for something more than a demonstrative— Different segments overlap/bleed into each other, the softness is an offset blur instead of a true feather/gradient, some angles might not be perfect, and other such details, though it does seem to all work out.

I should maybe clarify that this PR doesn't include the border images in the screenshots, just the option for tilesets to provide their own. I've uploaded them separately as mods under CC-BY, though, so anyone can take them and use them either directly or as a reference for their own:

https://github.com/will-ca/Sleek-Borders-Example
https://github.com/will-ca/Bubbly-Borders-Example
https://github.com/will-ca/Greasy-Borders-Example
https://github.com/will-ca/Sketchy-Borders-Example

PR #5874, opened 2021-12-29
Add `tileScale` in `TileSetConfig`.

will-ca:
Adds `tileScale` TileSetConfig option for all images passed through `setHexagonImageSize` to be effectively scaled up, with the origin at the center of the hex.

Default value does nothing.

This allows for some potentially quite powerful effects to be done on a per-tileset basis, and more easily:

screenshot

screenshot2

Super-blended terrain transitions:

image

Mountain shadows:

image

City search lights:

image

~~(In these experiments/examples, the terrain is visually misaligned with the hex grid. But I believe that is a limitation of my tileset experiments, not of this PR. Because the scaled-up tile can extend below each hex, the transition zones can (and, in an actual tileset, should) just be shifted south)~~

### DIFF
--- a/core/src/com/unciv/models/tilesets/TileSetConfig.kt
+++ b/core/src/com/unciv/models/tilesets/TileSetConfig.kt
@@ -8,6 +8,7 @@ class TileSetConfig {
     var fogOfWarColor: Color = Color.BLACK
     /** Name of the tileset to use when this one is missing images. Null to disable. */
     var fallbackTileSet: String? = "FantasyHex"
+    var tileScale: Float = 1f
     var ruleVariants: HashMap<String, Array<String>> = HashMap()
 
     fun updateConfig(other: TileSetConfig){
--- a/core/src/com/unciv/models/tilesets/TileSetConfig.kt
+++ b/core/src/com/unciv/models/tilesets/TileSetConfig.kt
@@ -8,6 +8,7 @@ class TileSetConfig {
     var fogOfWarColor: Color = Color.BLACK
     /** Name of the tileset to use when this one is missing images. Null to disable. */
     var fallbackTileSet: String? = "FantasyHex"
+    /** Scale factor for hex images, with hex center as origin. */
     var tileScale: Float = 1f
     var ruleVariants: HashMap<String, Array<String>> = HashMap()
 
--- a/core/src/com/unciv/ui/tilegroups/TileGroup.kt
+++ b/core/src/com/unciv/ui/tilegroups/TileGroup.kt
@@ -246,11 +246,11 @@ open class TileGroup(var tileInfo: TileInfo, val tileSetStrings:TileSetStrings, 
     /** Used for: Underlying tile, unit overlays, border images, perhaps for other things in the future.
      Parent should already be set when calling. */
     private fun setHexagonImageSize(hexagonImage: Image) {
-        // Using "scale" can get really confusing when positioning, how about no
         hexagonImage.setSize(hexagonImageWidth, hexagonImage.height * hexagonImageWidth / hexagonImage.width)
         hexagonImage.setOrigin(hexagonImageOrigin.first, hexagonImageOrigin.second)
         hexagonImage.x = hexagonImagePosition.first
         hexagonImage.y = hexagonImagePosition.second
+        hexagonImage.setScale(tileSetStrings.tileSetConfig.tileScale)
     }
 
     private fun updateTileImage(viewingCiv: CivilizationInfo?) {


ravignir:
OMG, that blending looks great. However, I wish it only affected the base terrain sprites and used special texture when water blends with land tiles.

will-ca:
@ravignir Hm.. Basic way to blend only base terrain sprites would be to just make sure terrain feature sprites never cover the southern transition areas, I think. (Instead shift it into northern transition zone for edge-to-edge coverage.) You'd have to play around with the parameters of the mask to make it all work; I basically just ran everything through a Python/PIL script and called it a day for this example/demo/test.

Could try drawing all terrain features on top of all base terrains too, but then— Are mountains base terrains?— There's probably cases where you'd still want the base terrain to cover up terrain features that are "behind" it.

Using different sprites for water and land blending would actually require somewhere between 6 and 64 extra sprites per coastal tile though, I think, since each of the 3-6 blended edges of every hex might have a different land/water neighbour type. I do have another, much more radical idea which would enable (amongst other things) tiles to customize appearance depending on what's next to them— All these graphical PRs I've been submitting are preparatory cleanup ahead of yet another Super Secret Evil Plan— But that's something that's much further off, if at all.

SomeTroglodyte:
A SSEP? Is that akin to a SEP[?](https://en.wikipedia.org/wiki/Somebody_else%27s_problem)

ravignir:
> Hm.. Basic way to blend only base terrain sprites would be to just make sure terrain feature sprites never cover the southern transition areas, I think. (Instead shift it into northern transition zone for edge-to-edge coverage.)

That would not work imo, there are river tiles and some tiles like forests would simply look bad (it would be hard to make them connect and make them look good at the same time).

>Could try drawing all terrain features on top of all base terrains too, but then— Are mountains base terrains?— There's probably cases where you'd still want the base terrain to cover up terrain features that are "behind" it.

Maybe specify in the tileset config file which tiles should blend (Coast, Ocean, Lake, Snow, Tundra, Desert, Plains, Grasslands).
Imo tiles that cover the entire tile can be reworked anyway (natural wonders, mountains).

>Using different sprites for water and land blending would actually require somewhere between 6 and 64 extra sprites per coastal tile though, I think, since each of the 3-6 blended edges of every hex might have a different land/water neighbour type. 

Just use slightly brighter desert tile to simulate sand for all base terrains. That is what civ5 does :D

will-ca:
> That would not work imo, there are river tiles and some tiles like forests would simply look bad (it would be hard to make them connect and make them look good at the same time).

Rivers are indeed tricky, but I think you would just have to offset their entire images north to be clear of the transition area. I'll see if I can make an updated proof-of-concept that does this for terrain features, and fix the textures to actually line up with the hex board.

> Maybe specify in the tileset config file which tiles should blend (Coast, Ocean, Lake, Snow, Tundra, Desert, Plains, Grasslands).
> Imo tiles that cover the entire tile can be reworked anyway (natural wonders, mountains).

Oh, Unciv doesn't actually do any blending! This is the exact same technique as the fuzzed edges that you posted in Discord some time ago, and I tried again more recently at #5835— Just with a much wider fuzz area.

Here's an example, of what the Grassland tile looks like:

Grassland

All this PR does is scale all tile images up by a constant tileset-defined factor (centered at the hex's center, after positioning them normally), so the blended areas of the textures extend all the way into neighbouring tiles.

So terrain features, meanwhile, look something like this:

Forest

(Transparent background.)

The terrain features show the true bounds of the hex, with everything in the margins extending into neighbouring tiles.

IMO Unciv should refrain from implementing too much special-cased rendering code beyond the basic "Place image here" operation. …It's my understanding that that's been tried before, during the age of fixed-function rendering pipelines, and nearly every program with graphics ended up finding that it was easier and more powerful to let assets themselves specify how they want to be rendered. For example: Should tilesets need to specify every possible pairing of terrains? Because that means 36 pairings of just base terrains, and quadratic growth with every new terrain. Would there need to be a Turing-complete pair-checking mechanism to make that practical? What if you still want to blend but just want a shorter transition, or a jagged coastline-like edge with a custom shape, or you want to blend like standard but also want to add E.G. a custom coastal beach graphic or a gradient tint near the border or something? I think adding generic *tools* like image scaling give a lot more artistic power relative to their implementation costs, whereas adding specific rendering *mechanisms* like tile blending tends to just create more things that need to be special-cased.

> Just use slightly brighter desert tile to simulate sand for all base terrains. That is what civ5 does :D

Oh. Huh. Is that what that is? I always thought it was supposed to be some weird chalky rock Lol.

But I mean even so, each of the 3-6 edges still has an independent state of either needing blending or needing a hard edge, and that's a lot of combinations to represent with sprite variants. (Plus, that comes back to special casing: If the base ruleset needs water-land boundaries to be able to behave differently, then the core code implementation should IMO support arbitrary types of boundaries being rendered differently— Which would then also exacerbate the sprite count problem.)

---

FWIW: I've been looking into letting tilesets provide GLSL shaders. Should be possible as a wrapper on GDX Actors, meaning the existing code, logic, and behaviour wouldn't really have to change, and I've got a custom rendering pipeline drafted up that should be quite powerful and performant if I manage to implement it. So in that case, basic behaviours like tile blending would be done by just extending the tile textures, and special-casing would be handled dynamically by transforming that texture in shaders where needed. The main app could include a couple commonly used shaders to make the process convenient for artists, but they would be interchangeable asset files instead of core logic.

(This doesn't mean I've abandoned the scripting API— My original timeline for that was "months or years", and that's the case with this too.)

will-ca:
> A SSEP? Is that akin to a SEP?

@SomeTroglodyte More like the opposite, if anything.

ravignir:
> But I mean even so, each of the 3-6 edges still has an independent state of either needing blending or needing a hard edge, and that's a lot of combinations to represent with sprite variants.

I've completely missed #5835 . What if you just used masks to represent each variant (and fill it with texture) instead of having multiple sprites for each terrain type and for each case.

will-ca:
> I've completely missed #5835 . What if you just used masks to represent each variant (and fill it with texture) instead of having multiple sprites for each terrain type and for each case.

Could work. Could also represent each edge as a separate mask so you only need three masks (for the three top edges), and composite them together at runtime to get the right variant.

But again, I would see that as a form of special-casing if implemented in core game code. And if I'm going to work on the improving graphics, I'd personally rather work towards just giving mods near-complete control via shaders than trying to provide dynamic effects through hard-coded behaviour.

---

Marking as draft because I didn't test the behaviour of this given the many Groups that have `isTransform = false` for performance.

xlenstra (approved):

PR #5876, opened 2021-12-30
Dispose object Graphics2D if it isn't used

lishaoxia1985:


will-ca:
There's another memory leak too in `Fonts.extractPixmapFromTextureRegion()` (and possibly one more where it's used, not sure). Happens a fixed number of times with every application launch, I think. I did some slight refactoring and fixed it in #5608 while exposing internal TextureRegions to Python/JS scripts, but didn't split the change off into its own PR. (Well, I think I imagined then that I would get that one ready for merge a lot sooner.)

Link to fix:

https://github.com/yairm210/Unciv/commit/0b5099c6172bb160d52dee8c57141066f6a21a2e#diff-33406f3c57bd42bc6ba64c0c6dfdca0de4c005aabdb7ea1260cc2528ab023c03

PR #5881, opened 2021-12-31
Translations update

xlenstra:
Already added now, as a Portuguese translations update had not been merged automatically for ~a month, most likely due to the user being marked as 'spammy'. As the translations updated looked fine, I believed they should be merged, but the only way to do this was manually, which is what I did here.


yairm210 (approved):

xlenstra:
Translation updates are a bit of a weird one, as it's not actually my code despite being my PR. So if I also agree with this, should I just merge it, despite also technically being the 'author' of this PR?

yairm210:
Yes

> Translation updates are a bit of a weird one, as it's not actually my code despite being my PR. So if I also agree with this, should I just merge it, despite also technically being the 'author' of this PR?

Yes

xlenstra:
summary

PR #5880, opened 2021-12-31
Removed periodic saving of TurnChecker again

GGGuenni:
It also did not work out with smaller files. Seems like I have to find a proper solution to implement #5756

PR #5883, opened 2021-12-31
Revert "Removed periodic saving of TurnChecker again"

touhidurrr:
Reverts yairm210/Unciv#5880
Why I want to do this? Check the discussion at #5880.


yairm210:
@GGGuenni what say you?

GGGuenni (approved):
Yea should be reverted as the commit was made based on a misunderstanding.

PR #5892, opened 2022-01-01
Revamped nukes again to closer match the original

### LINKED ISSUE #5888: Strength 1 Nukes with benefits over Strength 2
Strength 1 nukes have some seemingly out-of-place benefits over nukes of strength 2.

~~Nukes of strength 1 will destroy cities below 5 population if the unique `"Population loss from nuclear attacks []% []"` is not applied, while nukes of strength 2 will not.~~
Similarly, nukes of strength 2 will be prevented from decreasing population below 5 if the unique is applied, while nukes of strength 1 will not.

I'd imagine that these were supposed be reversed, as I doubt a strength 2 nuke was supposed be worse at destroying things than a strength 1 nuke.

#### Application of population damage:
in `nukeStrength1Effect()` https://github.com/yairm210/Unciv/blob/1f153cc86e5bc1fa3f6773c7e015cd4dceb84dd1/core/src/com/unciv/logic/battle/Battle.kt#L704-L710
in `nukeStrength2Effect()` 
https://github.com/yairm210/Unciv/blob/1f153cc86e5bc1fa3f6773c7e015cd4dceb84dd1/core/src/com/unciv/logic/battle/Battle.kt#L774-L777

xlenstra:
Fixes #5888 
Also reduced code duplication and fixed a bug where bomb shelters did not work against nuclear missiles

### DIFF
--- a/docs/uniques.md
+++ b/docs/uniques.md
@@ -23,17 +23,17 @@ Applicable to: Global, FollowerBelief, Improvement
 #### [stats] [cityFilter]
 Example: "[+1 Gold, +2 Production] [in all cities]"
 
-Applicable to: Global
+Applicable to: Global, FollowerBelief
 
 #### [stats] from every specialist [cityFilter]
 Example: "[+1 Gold, +2 Production] from every specialist [in all cities]"
 
-Applicable to: Global
+Applicable to: Global, FollowerBelief
 
 #### [stats] per [amount] population [cityFilter]
 Example: "[+1 Gold, +2 Production] per [20] population [in all cities]"
 
-Applicable to: Global
+Applicable to: Global, FollowerBelief
 
 #### [stats] in cities with [amount] or more population
 Example: "[+1 Gold, +2 Production] in cities with [20] or more population"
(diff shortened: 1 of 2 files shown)


yairm210 (approved):

docs/uniques.md
```diff
@@ -23,17 +23,17 @@ Applicable to: Global, FollowerBelief, Improvement
 #### [stats] [cityFilter]
 Example: "[+1 Gold, +2 Production] [in all cities]"
 
-Applicable to: Global
+Applicable to: Global, FollowerBelief
```
GGGuenni, line 26: The changes to uniques.md are completely additional and are not really a part of the nuke revamp right? I've got no clue how the uniques work as I've never worked with them.

xlenstra: Jup, this is an automatically generated documentation file. It should have been updated as part of #5882, but I forgot to include the file then, so it's included here instead

GGGuenni:
Another question: We Squash and merge right?

xlenstra:
I think so

SomeTroglodyte:
> Squash and merge

I'd say - Either take care the commit history is already locally on a minimal meaningful granularity -or- yes squash on PR merge. In other words - having one PR consisting of several steps in the final history might have some value, but that needs conscious evaluation and preparation? Am I formulating that right?
However, I haven't seen the process yet, and the repo settings controlling merge squash are not visible to us, so that may actually be irrelevant.

PR #5894, opened 2022-01-01
Added more yield icons to the city stats table

xlenstra:
E.g.:
image
image
image

Alternatively, I could only put the culture and food icons after the second number, or use the happiness icon for an active 'we love the king' day.

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
@@ -63,7 +63,7 @@ class CityStatsTable(val cityScreen: CityScreen): Table() {
                 } else "Stopped expansion".tr()
         if (cityInfo.expansion.chooseNewTileToOwn() != null)
             turnsToExpansionString +=
-                    " (${cityInfo.expansion.cultureStored}♪/${cityInfo.expansion.getCultureToNextTile()}♪)"
+                    " (${cityInfo.expansion.cultureStored}${Fonts.culture}/${cityInfo.expansion.getCultureToNextTile()}${Fonts.culture})"
 
         var turnsToPopString =
                 when {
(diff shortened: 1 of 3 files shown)


yairm210:
All the icons look unaligned to the text :/
I remember I had to manually change the small icons to raise them, did someone remake the icons from the originals or something?

yairm210:
Also, please use Fonts.x instead of the character itself, so it's obvious what's going on

SomeTroglodyte:
My vague memory says I re-did them once, but it also says there should be separate versions for font (50px) and larger (100? reduced from 200 when we had but one atlas spilling?) use, and I also vaguely remember tweaking the 50px ones for text baseline???

yairm210:
My issues with the font shouldn't prevent this from being merged
But I'm not going to merge this myself to see if the '2 other collaberators merge PRs' system works :)

yairm210 (approved):

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
```diff
@@ -74,17 +74,26 @@ class CityStatsTable(val cityScreen: CityScreen): Table() {
...
-            innerTable.add("Demanding [${cityInfo.demandedResource}]".toLabel()).row()
+        
+        val tableWithIcons = Table()
+        tableWithIcons.defaults().pad(2f)
+        if (cityInfo.isInResistance()) {
+            tableWithIcons.add(ImageGetter.getImage("StatIcons/Resistance")).size(20f)
+            tableWithIcons.add("In resistance for another [${cityInfo.getFlag(CityFlags.Resistance)}] turns".toLabel()).row()
+        }
+        if (cityInfo.isWeLoveTheKingDay()) {
+            tableWithIcons.add(ImageGetter.getStatIcon("Food")).size(20f)
```
will-ca, line 90: Nitpicky/highly subjective: Maybe the `Population` stat icon?

will-ca (approved):

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
xlenstra: Was my first idea as well, but we don't have that as an emoji yet

xlenstra: To me food seems more logical, as the main benefit is the +20% Growth (surplus food) it gives, but an argument could be made for both

SomeTroglodyte:
According to my mails there was mention that 'Population' is not yet there as emoji/font glyph.
- I did push the redone gimps to the repo, they're in [ImagesToNotAddToGame](https://github.com/yairm210/Unciv/tree/master/android/ImagesToNotAddToGame)
- To make a character from that, I take the full xcf, do a scale to 42px (NoHalo), paste as layer over the happiness 50px png, nudge the thing up using layer transparency till they align, throw away the helpers and export... Dunno whether github preserves png bits unchanged here, but you could try:
Population

PR #5895, opened 2022-01-01
GH Actions: Check user privileges before running bots.

will-ca:
I don't think any of the existing actions would do much harm even if a would-be vandal did run them, but Idk it still seems suboptimal to let any random account sic a bot on the repo.

I have tested the condition expressions in my fork:

https://github.com/will-ca/Unciv/pull/33

Can confirm that the original code let any random account do stuff to the summaries and translations, and this PR stops that while still letting privileged accounts run the bots:

https://github.com/will-ca/Unciv/pull/35

### DIFF
--- a/.github/workflows/uncivbot.yml
+++ b/.github/workflows/uncivbot.yml
@@ -6,7 +6,8 @@ on:
 
 jobs:
   summary:
-    if: github.event_name == 'issue_comment' && github.event.comment.body == 'summary'
+    if: github.event_name == 'issue_comment' && github.event.comment.body == 'summary' && contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
+    # This is the only place I could find an apparent list of valid author associations. Also, at least they're not case-sensitive: https://docs.github.com/en/graphql/reference/enums#commentauthorassociation https://docs.github.com/en/actions/learn-github-actions/expressions#contains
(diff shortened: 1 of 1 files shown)


yairm210:
Um, do you need that fromJson() if you can just put the array directly in there as json and not as a string?
IIRC the whole whole line is parsed as javascript

will-ca:
@yairm210 I didn't try it due to every change needing the whole ordeal of being pushed to my fork's master and run on GH, but Google told me there's no array literal syntax in GH Actions:

https://github.community/t/passing-an-array-literal-to-contains-function-causes-syntax-error/17213

It's not super clearly documented IMO, but my reading of the GH Actions "expressions" doc is that it's actually its own scripting language and not JS:

https://docs.github.com/en/actions/learn-github-actions/expressions

E.G. It describes semantics that are very much not JS semantics:

> GitHub ignores case when comparing strings.

And that also explains why it's `fromJSON`/`toJSON` instead of `JSON.parse`/`JSON.stringify`, `contains(a, v)` instead of `v in a`, and so on.

So it seems to me like the `script:` value for `actions/github-script@v3` gets run as JS on a Node instance, but the rest of the GH action itself is actually this weird custom language that's exactly similar enough to JS to be confusing as heck (and which doesn't have array literals).

yairm210 (approved):

will-ca:
I suppose I'll wait a day then squash merge?

PR #5897, opened 2022-01-01
Refactor pixel unit image resolving. (Includes `ImageAttempter`.)

will-ca:
There were a lot of repeated string operations before.

This replaces them with a builder-style syntax and helper class. Greatly improves readability and reliability IMO, as the exact same expression is no longer hardcoded twice, and the value used in `.imageExists` should now be the exact same as that used in `.getImage`. May also improve performance, as it cuts out the repeated string operations and the method call syntax may be more amenable to JIT'ing in some way or another— Does vastly increase the number of function calls in some scenarios, but when those calls are extraneous they basically only check a single Boolean and then exit immediately, which I hear/read is *probably* near-free on the JVM.

Tested with FantasyHex, Default, Pix5, 5Hex, and Civ Army Color Style Sheet.

The `ImageAttempter` class and the commit that add it are shared with a couple of other WIP PRs. Those are basically done too, but I want to make some vanilla assets and example mods before I submit them.

### DIFF
--- a/core/src/com/unciv/ui/tilegroups/TileGroup.kt
+++ b/core/src/com/unciv/ui/tilegroups/TileGroup.kt
@@ -11,6 +11,7 @@ import com.badlogic.gdx.utils.Align
 import com.unciv.UncivGame
 import com.unciv.logic.HexMath
 import com.unciv.logic.civilization.CivilizationInfo
+import com.unciv.logic.map.MapUnit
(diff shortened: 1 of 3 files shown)


GGGuenni (approved):
After googling the run scope function, I think I understand what's going on 😅 But I must say it looks really clean, well documented, and nice to use! Kudos to you!

PR #5900, opened 2022-01-02
Expanded the buildingFilter to include options for national wonders

xlenstra:
As implicitely requested on discord

### DIFF
--- a/core/src/com/unciv/models/ruleset/Building.kt
+++ b/core/src/com/unciv/models/ruleset/Building.kt
@@ -765,6 +765,8 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
             name -> true
             "Building", "Buildings" -> !isAnyWonder()
             "Wonder", "Wonders" -> isAnyWonder()
+            "National Wonder" -> isNationalWonder
+            "Non-national Wonder" -> isWonder
             replaces -> true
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -765,6 +765,8 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
             name -> true
             "Building", "Buildings" -> !isAnyWonder()
             "Wonder", "Wonders" -> isAnyWonder()
+            "National Wonder" -> isNationalWonder
+            "Non-national Wonder" -> isWonder
```
will-ca, line 769: I believe in Civ V these are called "World Wonders"— Though I think I lean towards "Non-national Wonder" myself.

will-ca (approved):
Obligatory "Idk about Uniques", but looks good to me!

core/src/com/unciv/models/ruleset/Building.kt
yairm210: This is a bit confusing though, as it could be taken to mean "all buildings that are not national wonders".
"World wonder", "Global wonder", "International wonder"? "World" seems like the best fit

PR #5903, opened 2022-01-02
Update Turkish.properties

HariboSever:
I translated again.This time i translated some error scripts


will-ca:
merge translations

PR #5904, opened 2022-01-02
3.18.14

xlenstra:
Updated translation files, version number for unciv & play store so this should be all, right?


yairm210:
Looks good!
I'm not 100% sure what will happen with the Google play since they don't allow texts longer than 500 characters through the UI, so I always shorten the texts

xlenstra:
> Looks good! I'm not 100% sure what will happen with the Google play since they don't allow texts longer than 500 characters through the UI, so I always shorten the texts

That's good to know, I'll shorten the text a bit and merge it in with the right tag

xlenstra:
Closing as I've merged it locally to add a tag to the commit

PR #5906, opened 2022-01-03
Ranged capture

itanasi:
Adding ability for Ranged Map Units to capture Civilian units
Added proper checks in BattleTable
Added exception to allow Ranged Units to move after fighting if just captured Civilian. Does NOT move if attacked a guarding unit
Note that City bombards will still kill Civilian units


itanasi:
Hold up. Didn't handle Air Units. I assume they can't capture lol

itanasi:
And need to double check I didn't break Naval units

will-ca:
Does this turn all ranged attacks on civilians into captures? Sometimes I still want to just shoot them, and not capture. …Jeez that sounds bad.

Is there a check to make sure ranged units can actually reach/move to the civilian units they capture, and code to move them?

itanasi:
Yeah, gotta redo some stuff after some thought.
New plan, give Ranged Units option to either attack (kill) Civilian unit, or move on to undefended Civilian units to capture

itanasi:
Not ready. Need to rework

PR #5908, opened 2022-01-03
Update Nations.json

Smashfanful:
Removed superfluous spaces


xlenstra:
One problem with this is that it invalidates existing translations of these lines. If it's not too much of a hassle, could you also remove the superfluous space in the translation files?

will-ca:
> One problem with this is that it invalidates existing translations of these lines.

Meaning's the same, though. Search and replace for those lines before merge?

will-ca (approved):

xlenstra:
> > One problem with this is that it invalidates existing translations of these lines. If it's not too much of a hassle, could you also remove the superfluous space in the translation files?
> 
> Okay. Interesting. It appears that the translation files already have the superfluous spaces removed, and `Nations.json` is the only place they show up. I assume the translation file generator thingy's probably collapsing consecutive whitespace?

Huh, looks like you're right

Issue #5901, opened 2022-01-02
Dropbox has run out of space (again)


xlenstra:
Same issue as #3986, basically.

Checking the allocated space on the [dropbox API tester](https://dropbox.github.io/dropbox-api-v2-explorer/#users_get_space_usage), yields:
```
{
  "used": 2684366459,
  "allocation": {
    ".tag": "individual",
    "allocated": 2684354560
  }
}
```
i.e., we're using more space than we have been allocated

yairm210:
I'm pretty sure we can create a github action as a cron job, to delete old saves every X amount of time
We can add args to the desktop runner to do so sorts of things, I think that's a nice extendable solution :)

xlenstra:
My knowledge of github actions and how to work with them isn't very good, so for now I've solved it with the python script below. I'm in favour of keeping this open as we do need some kind of solution like that.

<details> <summary> Python script: </summary>

Requires `pip3 install dropbox` to work.
This will download all deleted files, in case something does go wrong. Due to the likely large number of files that will be downloaded, I'd advise to run this script from inside an empty folder.

```python
LAST_YEAR_TO_REMOVE = 2021
LAST_MONTH_TO_REMOVE = 1
LAST_DAY_TO_REMOVE = 1
LAST_HOUR_TO_REMOVE = 12

import dropbox
dbx = dropbox.Dropbox("LTdBbopPUQ0AAAAAAAACxh4_Qd1eVMM7IBK3ULV3BgxzWZDMfhmgFbuUNF_rXQWb")

for entry in dbx.files_list_folder('/MultiplayerGames').entries:
    print("Found file {0}".format(entry.name))
    metadata = dbx.files_download_to_file(entry.name, "/MultiplayerGames/{0}".format(entry.name))
    print("Downloaded file {0}".format(entry.name))
    if (metadata.client_modified.year <= LAST_YEAR_TO_REMOVE
        and metadata.server_modified.year <= LAST_YEAR_TO_REMOVE
        and metadata.client_modified.month <= LAST_MONTH_TO_REMOVE
        and metadata.server_modified.month <= LAST_MONTH_TO_REMOVE
        and metadata.client_modified.day <= LAST_DAY_TO_REMOVE
        and metadata.server_modified.day <= LAST_DAY_TO_REMOVE
        and metadata.client_modified.hour <= LAST_HOUR_TO_REMOVE
        and metadata.server_modified.hour <= LAST_HOUR_TO_REMOVE
[...]

PR #5911, opened 2022-01-04
DropBox GH Action.

will-ca:
#5901

Python script by @xlenstra taken basically verbatim from the above issue.

Tested in my fork. Looks alright.

Has a lot of placeholders. E.G. Needs GH secrets, optional backup repository, etc. that should be simple enough in principle but I can't add through a PR.

Has code to archive old saves to a GH repository. Can either uncomment it and fill in all the placeholders to enable, or just delete those parts to reject. I like the idea of rescuing old saves. But GH provides only finite space for action runners.

Still needs to be changed to automatically set the date.

Currently the line that deletes DB files is commented out. IDK How DropBox works or is used. Recommend letting it run for a few days (or manually triggering), then uncommenting when logs look right.

---

EDIT: Well, I'll leave this open as an option, but I like @SomeTroglodyte's decentralized approach better:

> Didn't I have a branch months ago that would distribute the deletion load to the players? Like everyone who creates a new MP runs a little snippet that deletes the oldest 10 files? [*]
> 
> […]
> 
> [*] Kept it as patch in case anyone wants a read: [Automatic_DropBox_cleanup_on_newgame1.patch.zip](https://github.com/yairm210/Unciv/files/7809803/Automatic_DropBox_cleanup_on_newgame1.patch.zip)

https://github.com/yairm210/Unciv/issues/5901#issuecomment-1005092095

### DIFF
--- a/.github/workflows/cleanOldDropbox.yml
+++ b/.github/workflows/cleanOldDropbox.yml
@@ -1,0 +1,95 @@
+name: Clean old DropBox files.
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  clean_dropbox:
+    if: always()
+    runs-on: ubuntu-latest
+    env:
+      # TODO: Setup repo, use real secret with key?
+      GH_ARCHIVE_URL: https://secrets.PlaceholderString@github.com/account/repo.git
+      ARCHIVE_DIR: SomeArchiveRepository
+    steps:
+      - name: Setup Python.
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install DropBox.
+        shell: bash
+        run: "pip3 install dropbox"
+      - name: Setup Archive.
+        shell: bash
+        # TODO: Archive old saves in GH repo?
+        run: |
+          set -euo pipefail
+          echo "Cloning archive repository."
+          echo git clone "$GH_ARCHIVE_URL" "$ARCHIVE_DIR"
+          mkdir -p "$ARCHIVE_DIR"
+      - name: Clean DropBox.
+        shell: python
+        # TODO: Put key in secrets instead of hardcoded.
+        env:
+          DROPBOX_KEY: secrets.PLACEHOLDER_KEY
+        run: |
+          import dropbox, os
+          from datetime import datetime
+
+          os.chdir(os.environ['ARCHIVE_DIR'])
+          print(f"In {os.getcwd()}.")
+
+          LAST_YEAR_TO_REMOVE = 2021
+          LAST_MONTH_TO_REMOVE = 3
+          LAST_DAY_TO_REMOVE = 31
+
+
+          last_date_to_remove = datetime(LAST_YEAR_TO_REMOVE, LAST_MONTH_TO_REMOVE, LAST_DAY_TO_REMOVE)
+
+          dbx = dropbox.Dropbox("LTdBbopPUQ0AAAAAAAACxh4_Qd1eVMM7IBK3ULV3BgxzWZDMfhmgFbuUNF_rXQWb")
+          #TODO: Move this to GH secrets.
+          print(f"dbx = dropbox.Dropbox({os.environ['DROPBOX_KEY']})")
+
+          stop = False
+
+          i = 0
+          while True:
+              print("Starting with page {0}".format(i))
+              i += 1
+
+              # We can just use the list_folder and ignore the cursor
+              # as we delete all files anyway
+              folderList = dbx.files_list_folder('/MultiplayerGames')
+
+              for entry in folderList.entries:
+                  print("Found file {0}".format(entry.name))
+                  metadata = dbx.files_download_to_file(entry.name, "/MultiplayerGames/{0}".format(entry.name))
+                  print("Downloaded file {0}".format(entry.name))
+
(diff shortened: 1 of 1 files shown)


yairm210:
Why not both?
I'm in favor of trying this out as a GH action

yairm210:
If and when we have it decentralized we can disable the action and see how things go, to deprecate it gradually
Currently, we don't have that, and we do have this, so...

yairm210 (approved):

xlenstra:
I'm not very well versed in how exactly github actions work, but looking through the code there are still a lot of placeholders & commented out code. If I read your comment correctly, these will be fixed after this PR? If so, I approve and feel free to merge this in.

For now I've put all the save files I've downloaded in a git repo [here](https://github.com/xlenstra/unciv-save-archive). I'm fine with that also being the repo used to store future save files, but I'm not sure on how to set that up. I can send you a collaboration request, or I'm fine with you downloading all the save files and adding these to another git repo.

will-ca:
> If I read your comment correctly, these will be fixed after this PR?

That's the idea. The most important behaviour (`dbx.files_delete()`, line `72`) is commented because I can't fully test the bot (except for on my fork) until it's merged, so will check the logs first to make sure it's acting right before having it delete anything. Some things are commented because I either can't or didn't think I should set them up myself (GH secrets, archive repository). And the backup script steps are commented because of that, plus not being sure if that's a feature we even want— If not, then those lines can just be deleted.

> I'm fine with that also being the repo used to store future save files, but I'm not sure on how to set that up. I can send you a collaboration request, or I'm fine with you downloading all the save files and adding these to another git repo.

Um... I think you would actually need to send @yairm210 the collaboration request, as I believe that's the PAT that's set in this repository's secrets. So that would let the GH Action use the existing PAT to upload backups. Although, it would also be helpful if you could send me a collaboration request as well, as that way I would be able to set up and use my own PAT to test or improve the upload steps in my fork before merging this.

@xlenstra How big are all the save files? I will definitely have to switch to a different way of updating the repo if they're in the multiple-gigabyte range, due to GH runner space limits. Also, are the filenames practically guaranteed to be unique? That may affect whether we can put everything in a flat directory, or have to put differerent batches in timestamped directories. Would you be okay with me restructuring them into directories as/if needed?

This PR will change significantly if we're going to store the old saves from day one, so marking as draft.

xlenstra:
> > If I read your comment correctly, these will be fixed after this PR?
> 
> That's the idea. The most important behaviour (`dbx.files_delete()`, line `72`) is commented because I can't fully test the bot (except for on my fork) until it's merged, so will check the logs first to make sure it's acting right before having it delete anything. Some things are commented because I either can't or didn't think I should set them up myself (GH secrets, archive repository). And the backup script steps are commented because of that, plus not being sure if that's a feature we even want— If not, then those lines can just be deleted.

:+1:

> > I'm fine with that also being the repo used to store future save files, but I'm not sure on how to set that up. I can send you a collaboration request, or I'm fine with you downloading all the save files and adding these to another git repo.
> 
> Um... I think you would actually need to send @yairm210 the collaboration request, as I believe that's the PAT that's set in this repository's secrets. So that would let the GH Action use the existing PAT to upload backups. Although, it would also be helpful if you could send me a collaboration request as well, as that way I would be able to set up and use my own PAT to test or improve the upload steps in my fork before merging this.

I'll send both of you a collaboration request

> @xlenstra How big are all the save files? I will definitely have to switch to a different way of updating the repo if they're in the multiple-gigabyte range, due to GH runner space limits. Also, are the filenames practically guaranteed to be unique? That may affect whether we can put everything in a flat directory, or have to put different batches in timestamped directories. Would you be okay with me restructuring them into directories as/if needed?

The 3 months worth of data adds up to ~470 MB of data. The filenames are randomly generated UUID's, so for all practical purposes those should be unique. Feel free to do anything to the repo you like. I'll keep a local copy of all the save files in case anything goes wrong.

> This PR will change significantly if we're going to store the old saves from day one, so marking as draft.


Alternatively, I'm fine with just running the script as is every ~3 months as long as I'm still with unciv, so don't feel obligated to get something like this working if its too much of a hassle :)

yairm210:
Is there anything preventing us from adding this in?

xlenstra:
Looking at the action, currently the line deleting the files is commented out, deleted files aren't backed up as of yet, and the maximal date till which saves are deleted is still hardcoded in the github action.

will-ca:
Will prioritize.

touhidurrr:
And so where actually this GH Action intends to backup old save files?
Please don't use yairm210/Unciv

yairm210:
The save files are just for being careful, once we see that we can consistently delete old save files we won't need any extra safety measures.
Looking at this again, I think the 'safe' method would be to extract all the python to a separate file, have the python file receive the date-to-delete-older-than as an ISO date string or something, and then it'll be perfectly testable locally and the only thing we'll need to worry about with the Github Action would be sending it a correct ISO string.

touhidurrr:
Btw, @yairm210 
Did you guys implement the Dropbox backup thing?
I am getting this:
image

yairm210:
Nope, we don't have anything that auto deletes old Dropbox files, but we should

touhidurrr:
@yairm210 can I try to implement it then?

Something like a script that runs everyday, and..
if Dropbox usage > 95%
then sort files by updated Timestamp.
bulk delete files old files to make space 90%.

xlenstra:
I'll boot up my python script again locally, should (again) be at least a temporary fix

yairm210:
Yeah sure, more solutions is better :D

PR #5915, opened 2022-01-05
Transform some var to unique in Building/Unit,  some modification for the fun getRejectionReasons

lishaoxia1985:
I don't know how to tackle with replacementTextForUniques, it seems not to be used in the app.
If it is relate to mod, when we load the mod, we can do this: "unique = replacementTextForUnique", we should only tackle with unique.

### DIFF
--- a/core/src/com/unciv/models/ruleset/Building.kt
+++ b/core/src/com/unciv/models/ruleset/Building.kt
@@ -49,10 +49,12 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
     var isNationalWonder = false
     fun isAnyWonder() = isWonder || isNationalWonder
     var requiredBuilding: String? = null
+    // Deprecated
     var requiredBuildingInAllCities: String? = null
+    //
 
     /** A strategic resource that will be consumed by this building */
-    private var requiredResource: String? = null
+    var requiredResource: String? = null
 
     /** City can only be built if one of these resources is nearby - it must be improved! */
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -49,10 +49,12 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
     var isNationalWonder = false
     fun isAnyWonder() = isWonder || isNationalWonder
     var requiredBuilding: String? = null
+    // Deprecated
```
will-ca, line 52: Annotation?

lishaoxia1985: you can ask @yairm210, it has been replaced with "Requires a [﹉] in all the cities＂

yairm210:
Most of these are changes that haven't been made.
There are three instances where we have 2 different ways of specifying things.
requiredBuilding and requiredBuildingInAllCities - should be replaced by unique, first step is to add a ruleset warning about this in checkRuleset, and mark the parameters as deprecated.

The second thing is the resources - this is actually a good question, I'm not sure if deprecating the old method is the way to go.

As for the others, I don't think those uniques exist?


In any case this should be split into at least 2 PRs, one for requiredBuildingInAllCities and requiredBuilding, and I've for resources.
The others would require a change to each, and we'd need to discuss them.

PR #5918, opened 2022-01-06
Disables '[cityState] is afraid of your military power' for spectators & other non-major civs

xlenstra:
Also for a few other similar notifications

### DIFF
--- a/core/src/com/unciv/logic/civilization/QuestManager.kt
+++ b/core/src/com/unciv/logic/civilization/QuestManager.kt
@@ -500,7 +500,7 @@ class QuestManager {
     fun cityStateConquered(cityState: CivilizationInfo, attacker: CivilizationInfo) {
         val matchingQuests = assignedQuests.asSequence()
             .filter { it.questName == QuestName.ConquerCityState.value }
-            .filter { it.data1 == cityState.civName && it.assignee == attacker.civName}
+            .filter { it.data1 == cityState.civName && it.assignee == attacker.civName }
 
         for (quest in matchingQuests)
             giveReward(quest)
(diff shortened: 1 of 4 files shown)


yairm210 (approved):

yairm210:
The thing is, that these notifications won't really make a difference, since AI chica receiving notifications just throw them away in "addNotification"
This is just a matter of "does this save runtime", and on a second pass, I'm not sure it does

yairm210:
But for spectators it is important

xlenstra:
> But for spectators it is important

True, this is mostly for spectators. I could make the check only exclude spectators, if that's better. I'm not sure this is a huge performance save/detriment either way

yairm210:
Nah, leave it as is

core/src/com/unciv/logic/civilization/QuestManager.kt
```diff
@@ -500,7 +500,7 @@ class QuestManager {
     fun cityStateConquered(cityState: CivilizationInfo, attacker: CivilizationInfo) {
         val matchingQuests = assignedQuests.asSequence()
             .filter { it.questName == QuestName.ConquerCityState.value }
-            .filter { it.data1 == cityState.civName && it.assignee == attacker.civName}
+            .filter { it.data1 == cityState.civName && it.assignee == attacker.civName }
```
GGGuenni, line 503: Unrelated to this PR and without knowing what it's for (perhaps just because of that), but data1 really sound like a suboptimal variable name

xlenstra: Wait until you see its counterpart, `data2` (yes, this exists) :upside_down_face: 
If I read the code correctly, the variables are used to store data for quests that differs for each quest. E.g., in the quest for finding a civ, which civ should be found, or which natural wonder should be found for the 'find natural wonder' quest, etc. While it is possible to make variables for all these, it is easier (though costing readability) to use a multi-functional variable, especially as you then don't have variables that are only used for a single quest and unused outside of that.

SomeTroglodyte: In other words, a plainer way to do what subclassing is for?

PR #5920, opened 2022-01-06
Deprecated old uniques

yairm210:


core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -48,6 +48,7 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
     var isWonder = false
     var isNationalWonder = false
     fun isAnyWonder() = isWonder || isNationalWonder
+    @Deprecated("Since 3.18.15 - replaced with RequiresAnotherBuilding unique")
     var requiredBuilding: String? = null
```
xlenstra, line 52: Why not remove this outright, if its uses have been removed?

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -48,6 +48,7 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
     var isWonder = false
     var isNationalWonder = false
     fun isAnyWonder() = isWonder || isNationalWonder
+    @Deprecated("Since 3.18.15 - replaced with RequiresAnotherBuilding unique")
```
GGGuenni, line 51: I'm really bad at this whole unique thing but didn't you remove the requiredBuilding instead of deprecating it? Or are lines 252-259 something different? Couldn't find any other depreciation in the commits since 3.18.14 as this is marked `since 3.18.15`

core/src/com/unciv/models/ruleset/Building.kt
yairm210: Tru dat

yairm210:
Actually this is not a good way of doing it. We should definitely first only warn mod creators that these fields are deprecated and only much later remove them entirely.

xlenstra (approved):

xlenstra:
> Actually this is not a good way of doing it. We should definitely first only warn mod creators that these fields are deprecated and only much later remove them entirely.

Ah, missed this as I was already reading through code. I do agree though

yairm210:
I'm actually uncertain about this.
'Requires a X in all cities" is a relatively rare one, so that should obviously be a unique,
But most buildings require another building, so should we really convert that to a unique permanently and not allow the old way of doing it? I don't know, the more I think of it the less sense it makes for me

xlenstra:
~28% of buildings require another building to be built, so it isn't really a 'unique property' in that sense. It's also a question of where we draw the line between uniques and attributes. In theory we could make virtually every attribute a unique*, but while that does allow for easier generation of UI strings (they already exist), I'm not sure whether it makes the lives of modders easier

<details> <summary> * e.g., for a market: </summary>

```json
{
  "name": "Market",
  "gold": 2,
  "specialistSlots": {"Merchant": 1},
  "hurryCostModifier": 25,
  "percentStatBonus": {"gold": 25},
  "requiredTech": "Currency"
},
```
Could be turned into
```json
{
  "name": "Market",
  "requiredTech": "Currency",
  "uniques": [
    "[+2 Gold]", "[+25]% [Gold] [in this city]", 
    "[+1] [Merchant] Specialist slot", "[+25]% [Gold] cost when bought"
  ]
},
```
</details>

PR #5921, opened 2022-01-06
Added a way to add moddable constants

xlenstra:
The original game had a file named `globalDefines.xml`, which contained a lot of constants that could easily change the formula's the game uses. Until now, Unciv didn't have any equivalent of that, which for example required the techstopper mod series to add a lot of bogus techs to the end of the tech tree, in order to avoid city strength getting to information-era levels in the renaissance.
This PR creates a way to change these constants, by adding them as optional parameters in the `ModOptions.json` file. It can later be extended in a similar manner to allow for more moddability of similar formula's. An additional idea for the `ModConstants` file is to be understandable by non-coders, and reference all formula's the constants are used in in there as well.

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -417,7 +417,10 @@ object Battle {
     private fun addXp(thisCombatant: ICombatant, amount: Int, otherCombatant: ICombatant) {
         var baseXP = amount
         if (thisCombatant !is MapUnitCombatant) return
-        if (thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.maxXPfromBarbarians
+        if (
+            (thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.maxXPfromBarbarians
+                || thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.constants.maxXPfromBarbarians
+            )
             && otherCombatant.getCivInfo().isBarbarian()
         ) {
             return
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -417,10 +417,8 @@ object Battle {
     private fun addXp(thisCombatant: ICombatant, amount: Int, otherCombatant: ICombatant) {
         var baseXP = amount
         if (thisCombatant !is MapUnitCombatant) return
-        if (
-            (thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.maxXPfromBarbarians
-                || thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.constants.maxXPfromBarbarians
-            )
+        val modConstants = thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.constants
+        if (thisCombatant.unit.promotions.totalXpProduced() >= modConstants.maxXPfromBarbarians
             && otherCombatant.getCivInfo().isBarbarian()
         ) {
             return
--- a/core/src/com/unciv/models/ModConstants.kt
+++ b/core/src/com/unciv/models/ModConstants.kt
@@ -1,0 +1,13 @@
+package com.unciv.models
+
+class ModConstants {
+    // Max amount of experience that can be gained from combat with barbarians
+    val maxXPfromBarbarians = 30
+
+    // Formula for city Strength:
+    // Strength = baseStrength * (%techs * multiplier) ^ exponent
+    // If no techs exist in this ruleset, %techs = 0.5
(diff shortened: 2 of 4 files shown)


yairm210 (approved):

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -417,7 +417,10 @@ object Battle {
     private fun addXp(thisCombatant: ICombatant, amount: Int, otherCombatant: ICombatant) {
         var baseXP = amount
         if (thisCombatant !is MapUnitCombatant) return
-        if (thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.maxXPfromBarbarians
+        if (
+            (thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.maxXPfromBarbarians
+                || thisCombatant.unit.promotions.totalXpProduced() >= thisCombatant.unit.civInfo.gameInfo.ruleSet.modOptions.constants.maxXPfromBarbarians
```
will-ca, line 422: Equivalent to `totalXpProduced() >= min(modOptions.maxXPfromBarbarians, modOptions.constants.maxXPfromBarbarians)`? I think mods would have to define both properties to be able to increase the threshold.

core/src/com/unciv/models/ModConstants.kt
```diff
@@ -0,0 +1,13 @@
...
+
+class ModConstants {
+    // Max amount of experience that can be gained from combat with barbarians
+    val maxXPfromBarbarians = 30
+
+    // Formula for city Strength:
+    // Strength = baseStrength * (%techs * multiplier) ^ exponent
+    // If no techs exist in this ruleset, %techs = 0.5
+    val cityStrengthFromTechsMultiplier = 5.5
+    val cityStrengthFromTechsExponent = 2.8
```
will-ca, line 11: Does `val` work for deserialization?

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -46,13 +47,16 @@ class ModOptions : IHasUniques {
...
  
+    @Deprecated("As of 3.18.15")
     val maxXPfromBarbarians = 30
 
     override var uniques = ArrayList<String>()
     // If this is delegated with "by lazy", the mod download process crashes and burns
     override var uniqueObjects: List<Unique> = listOf()
     override fun getUniqueTarget() = UniqueTarget.ModOptions
 
+    val constants = ModConstants()
```
will-ca, line 58: Also.

will-ca (commented):
Pretty glad to see this feature personally. That's the last noticeable part of hardcoding gone, I think?

core/src/com/unciv/logic/battle/Battle.kt
xlenstra: The `modOptions.maxXPfromBarabrians` is only temporary and should be removed once most mods have updated, so upping both should be unnecessary.
It should be a `min` if you assume mods will lower the amount, and a `max` otherwise, so I'll work something out for that before merging

core/src/com/unciv/models/ModConstants.kt
xlenstra: I think so -- I've had cases where the type of an object was changed when deserializing so I assume the deserializer will also ignore whether something is a val or var.

core/src/com/unciv/logic/battle/Battle.kt
will-ca: Oh, I meant the current behaviour is equivalent to a min, which probably isn't intended as it prevents the mod from being able to increase the value unless it increases both properties.

Maybe checking only the new property, and setting it from the old one on init if the old one's changed and the new one's unchanged?

core/src/com/unciv/models/ModConstants.kt
xlenstra: I'll test it and see what happens

xlenstra: Seems to work fine

core/src/com/unciv/models/ruleset/Ruleset.kt
SomeTroglodyte: That "crashes and burns" comment still there? It's wrong. A by lazy and deserialization mandates a transient annotation like everywhere else, that's all...

core/src/com/unciv/models/ModConstants.kt
SomeTroglodyte: Gdx Json operates entirely through reflection- the debugger is also able to modify val's. _const_ val's would be another matter, but those can't live in a serializable context anyway.

Question is - readability?

will-ca: Right... Not sure if Kotlin reflection (rather than Java reflection) can naïvely modify `val`s. If not, that may be why I assumed GDX can't.

On readability: IDE/Compiler has some errors based on mutability like "Cannot smart-cast because might have changed". I suppose those wouldn't work when the property is marked as a `val`. I could see myself introducing edge-case bugs in code I'm unfamiliar with by assuming they can't change and using them in a constructor before GDX can set them or something, for example. So... Which is more readable when they in practice both do change and usually shouldn't change?

PR #5924, opened 2022-01-07
Converted "Can construct [improvementName]" to a uniqueType

yairm210:


xlenstra (approved):
Would it make sense to also include a StateForConditionals in all the `.getMatchingUniques()` calls now that we use a UniqueType?

yairm210:
MapUnit.getMatchingUniques already contains this by default so I'm not sure which call you're referring to

xlenstra:
> MapUnit.getMatchingUniques already contains this by default so I'm not sure which call you're referring to

True, and there probably aren't any other things that should be added to the state, so it's fine :+1:

PR #5927, opened 2022-01-07
Simplified conditional checks - where I felt it would make it more readable

yairm210:
Also added checks to ensure that era exists when checking era.

Big think - many tile conditionals contain 'if attackedTile OR unit tile' checks.
This is probably not intended behavior since if I'm attacking another unit, I should only get bonuses for the tile the combat occurs in, and not for the tile I'm currently in, if I understand correctly,

We should probably have something like `val relevantTile = state.attackedTile ?: state,unit?.getTile()`

And then run all checks on that tile
That'll also simplify the logic of those checks by a lot, since there's a lot of reused code in tile checks due to this

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -49,19 +49,25 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
         condition: Unique,
         state: StateForConditionals
     ): Boolean {
+
+        val ruleset = state.civInfo?.gameInfo?.ruleSet
+
         return when (condition.type) {
             UniqueType.ConditionalWar -> state.civInfo?.isAtWar() == true
             UniqueType.ConditionalNotWar -> state.civInfo?.isAtWar() == false
-            UniqueType.ConditionalHappy -> 
+            UniqueType.ConditionalHappy ->
                 state.civInfo != null && state.civInfo.statsForNextTurn.happiness >= 0
             UniqueType.ConditionalGoldenAge ->
                 state.civInfo != null && state.civInfo.goldenAges.isGoldenAge()
             UniqueType.ConditionalBeforeEra ->
-                state.civInfo != null && state.civInfo.getEraNumber() < state.civInfo.gameInfo.ruleSet.eras[condition.params[0]]!!.eraNumber
+                state.civInfo != null && ruleset!!.eras.containsKey(condition.params[0])
+                        && state.civInfo.getEraNumber() < ruleset.eras[condition.params[0]]!!.eraNumber
             UniqueType.ConditionalStartingFromEra ->
-                state.civInfo != null && state.civInfo.getEraNumber() >= state.civInfo.gameInfo.ruleSet.eras[condition.params[0]]!!.eraNumber
+                state.civInfo != null && ruleset!!.eras.containsKey(condition.params[0])
+                        && state.civInfo.getEraNumber() >= ruleset.eras[condition.params[0]]!!.eraNumber
             UniqueType.ConditionalDuringEra ->
-                state.civInfo != null && state.civInfo.getEraNumber() == state.civInfo.gameInfo.ruleSet.eras[condition.params[0]]!!.eraNumber
+                state.civInfo != null && ruleset!!.eras.containsKey(condition.params[0])
+                        && state.civInfo.getEraNumber() == ruleset.eras[condition.params[0]]!!.eraNumber
             UniqueType.ConditionalTech ->
                 state.civInfo != null && state.civInfo.tech.isResearched(condition.params[0])
             UniqueType.ConditionalNoTech ->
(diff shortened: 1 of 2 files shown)


xlenstra:
> Big think - many tile conditionals contain 'if attackedTile OR unit tile' checks. This is probably not intended behavior since if I'm attacking another unit, I should only get bonuses for the tile the combat occurs in, and not for the tile I'm currently in, if I understand correctly,
> 
> We should probably have something like `val relevantTile = state.attackedTile ?: state,unit?.getTile()`
> 
> And then run all checks on that tile That'll also simplify the logic of those checks by a lot, since there's a lot of reused code in tile checks due to this

As it's currently implemented, `unit` is used exclusively outside of combat, and `ourCombatant` and `theirCombatant` are used exclusively inside of combat, so there is no unintentional behaviour with `unit.getTile()` being used when attacking. I do agree that it'd be an improvement for the logic

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -49,19 +49,25 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
...
             UniqueType.ConditionalWar -> state.civInfo?.isAtWar() == true
             UniqueType.ConditionalNotWar -> state.civInfo?.isAtWar() == false
-            UniqueType.ConditionalHappy -> 
+            UniqueType.ConditionalHappy ->
                 state.civInfo != null && state.civInfo.statsForNextTurn.happiness >= 0
             UniqueType.ConditionalGoldenAge ->
                 state.civInfo != null && state.civInfo.goldenAges.isGoldenAge()
             UniqueType.ConditionalBeforeEra ->
-                state.civInfo != null && state.civInfo.getEraNumber() < state.civInfo.gameInfo.ruleSet.eras[condition.params[0]]!!.eraNumber
+                state.civInfo != null && ruleset!!.eras.containsKey(condition.params[0])
```
xlenstra, line 63: Shouldn't these kinds of checks be made when building the ruleset instead of during the game?

yairm210: Yes, which makes it all the stranger that the Google Play Console shows this isn't the case:

```
java.lang.NullPointerException: 
  at com.unciv.models.ruleset.unique.Unique.conditionalApplies (Unique.kt:62)
  at com.unciv.models.ruleset.unique.Unique.conditionalsApply (Unique.kt:43)
  at com.unciv.logic.civilization.CivilizationInfo$getMatchingUniques$2.invoke (CivilizationInfo.kt:405)
  at com.unciv.logic.civilization.CivilizationInfo$getMatchingUniques$2.invoke (CivilizationInfo.kt:404)
  at kotlin.sequences.FilteringSequence$iterator$1.calcNext (Sequences.kt:171)
  at kotlin.sequences.FilteringSequence$iterator$1.hasNext (Sequences.kt:194)
  at kotlin.sequences.FlatteningSequence$iterator$1.ensureItemIterator (Sequences.kt:316)
  at kotlin.sequences.FlatteningSequence$iterator$1.hasNext (Sequences.kt:303)
  at com.unciv.models.ruleset.unit.BaseUnit.canBePurchasedWithStat (BaseUnit.kt:705)
  at com.unciv.logic.city.INonPerpetualConstruction$DefaultImpls.canBePurchasedWithAnyStat (IConstruction.kt:51)
  at com.unciv.models.ruleset.unit.BaseUnit.canBePurchasedWithAnyStat (BaseUnit.kt:26)
  at com.unciv.models.ruleset.unit.BaseUnit.shouldBeDisplayed (BaseUnit.kt:371)
  at com.unciv.ui.cityscreen.CityConstructionsTable$getConstructionButtonDTOs$1.invoke (CityConstructionsTable.kt:166)
  at com.unciv.ui.cityscreen.CityConstructionsTable$getConstructionButtonDTOs$1.invoke (CityConstructionsTable.kt:166)
  at kotlin.sequences.FilteringSequence$iterator$1.calcNext (Sequences.kt:171)
  at kotlin.sequences.FilteringSequence$iterator$1.hasNext (Sequences.kt:194)
  at com.unciv.ui.cityscreen.CityConstructionsTable.getConstructionButtonDTOs (CityConstructionsTable.kt:166)
  at com.unciv.ui.cityscreen.CityConstructionsTable.access$getConstructionButtonDTOs (CityConstructionsTable.kt:30)
  at com.unciv.ui.cityscreen.CityConstructionsTable$updateAvailableConstructions$1.invoke (CityConstructionsTable.kt:208)
  at com.unciv.ui.cityscreen.CityConstructionsTable$updateAvailableConstructions$1.invoke (CityConstructionsTable.kt:206)
```

xlenstra: If I'm not mistaken, we don't actually check the validity of parameters of conditionals yet, which would also explain the crash

yairm210: Oh.
Actually it appears we don't check this.
Ruleset.checkUniques runs unique.type.getComplianceErrors
But you know what doesn't run? 
unique.conditionals

SomeTroglodyte:
Haven't followed closely - but removing the null guard from `eras[condition.params[0]]!!` in that last commit - is it already guaranteed the exceptions from malformed mods are caught at load time?

yairm210:
Yes, in another PR we created parameter checks fire conditionals, so you won't even be able to load a ruleset where the era is not defined

xlenstra (approved):

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -70,61 +74,55 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
...
-            UniqueType.ConditionalForeignContinent -> 
-                state.civInfo != null 
-                && state.unit != null
-                && (state.civInfo.cities.isEmpty() 
-                    || state.civInfo.getCapital().getCenterTile().getContinent() != state.unit.getTile().getContinent()
-                )
+            UniqueType.ConditionalForeignContinent ->
+                state.civInfo != null && state.unit != null
+                        && (state.civInfo.cities.isEmpty()
+                        || state.civInfo.getCapital().getCenterTile().getContinent()
```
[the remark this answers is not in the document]
yairm210: I just use that AS recommends usually

PR #5934, opened 2022-01-09
Update French.properties

Newexae:
Sorry for the 8 commits, I'm not used to Github and I did some testing.


yairm210:
Placeholder `[unitName]` not found in `French` for entry `Return [unitName] to [civName]?`
    Placeholder `[cityName]` not found in `French` for entry `We Love The King Day in [cityName] has ended.`

android/assets/jsons/translations/French.properties
```diff
@@ -693,9 +693,9 @@ The [questName] quest for [civName] has ended. It was won by [civNames]. =
 The resistance in [cityName] has ended! = la résistance dans [cityName] est finie!
 [cityName] demands [resource]! = [cityName] demande [resource]!
  # Requires translation!
-Because they have [resource], the citizens of [cityName] are celebrating We Love The King Day! = 
+Because they have [resource], the citizens of [cityName] are celebrating We Love The King Day! = Parce qu'ils ont [resource], les citoyens de [cityName] célèbrent la Fête du Roi !
  # Requires translation!
-We Love The King Day in [cityName] has ended. = 
+We Love The King Day in [cityName] has ended. = La Fête du Roi est terminée à [cityName].
```
SomeTroglodyte, line 698: Majority of the other translations here uses a lower case 'roi' - consistency is good

android/assets/jsons/translations/French.properties
```diff
@@ -1043,17 +1043,17 @@ Undefeated civs = Civilisations invaincues
  # Requires translation!
 Turns until the next\ndiplomacy victory vote: [amount] = 
  # Requires translation!
-Choose a civ to vote for = 
+Choose a civ to vote for = Choississez une civilisation pour laquelle voter
  # Requires translation!
-Choose who should become the world leader and win a diplomatic victory! = 
+Choose who should become the world leader and win a diplomatic victory! = Coississez qui devrait devenir le dirigeant mondial et gagner une victoire diplomatique !
```
SomeTroglodyte, line 1048: Petit Typo

android/assets/jsons/translations/French.properties
```diff
@@ -1474,11 +1474,11 @@ Gain a free [building] [cityFilter] =
 
 Only available after [] turns = Seulement disponible après [] tours
  # Requires translation!
-This Unit upgrades for free = 
+This Unit upgrades for free = Cette unité s'est améliorée gratuitement
```
SomeTroglodyte, line 1477: Isn't the left one talking about a not yet realized possibility while your translation sounds past tense?

Newexae: I thought it was a summary of the precedent turn, sorry.

SomeTroglodyte: Nonono I formulated that as question because I did not actually check the context... OK checking now... Actually that's used for Ancient Ruins rewards. Those are instantaneous, so present/past might be better? But then, I _think_ you see that text only in the Civilopedia listing of possible ruin rewards... Where the potential version would look better?

Hmmm, the Pedia entry does use different texts altogether??? Sorry, then I don't know the context where this string is seen. (OK, this exact one is commented as `// Not used in Vanilla`, but its sibling `"This Unit upgrades for free including special upgrades"` should be seen somewhere)

PR #5949, opened 2022-01-11
Added a few missing template strings from the 'about' page of the settings

xlenstra:


SomeTroglodyte (approved):
And I was sure I tested that in kraut back when I built it...

PR #5951, opened 2022-01-11
Typed a few unit uniques

xlenstra:


SomeTroglodyte:
Interesting quotes in `CanOnlyAttackUnits param[0] type "mapUnitFilter/"City""`... and I don't see any logic supporting that Unique?

xlenstra:
> Interesting quotes in `CanOnlyAttackUnits param[0] type "mapUnitFilter/"City""`... and I don't see any logic supporting that Unique?

Ah, that was probably the reason why I didn't commit this earlier :sweat_smile:

SomeTroglodyte:
`CanOnlyAttackUnits` still appears nowhere else, so it would be non-functional for now...?

SomeTroglodyte (approved):

will-ca (approved):

PR #5952, opened 2022-01-11
More unit unique typing

yairm210:
When will the madness end

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
+++ b/core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
@@ -284,9 +284,10 @@ object UnitActions {
                     tile.improvement = null
                     if (tile.resource != null) tile.getOwner()?.updateDetailedCivResources()    // this might take away a resource
 
-                    val freePillage = unit.hasUnique("No movement cost to pillage")
-                        || unit.civInfo.hasUnique("No movement cost to pillage")
-                        || (unit.baseUnit.isMelee() && unit.civInfo.hasUnique("Melee units pay no movement cost to pillage"))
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -284,9 +284,10 @@ object UnitActions {
                     tile.improvement = null
                     if (tile.resource != null) tile.getOwner()?.updateDetailedCivResources()    // this might take away a resource
 
-                    val freePillage = unit.hasUnique("No movement cost to pillage")
-                        || unit.civInfo.hasUnique("No movement cost to pillage")
-                        || (unit.baseUnit.isMelee() && unit.civInfo.hasUnique("Melee units pay no movement cost to pillage"))
+                    val freePillage = unit.hasUnique(UniqueType.NoMovementToPillage)
+                        || unit.civInfo.hasUnique(UniqueType.NoMovementToPillage)
```
xlenstra, line 288: Shouldn't these be combined into `unit.hasUnique(NoMovementToPilalge, checkCivUniques=true)`?

xlenstra (approved):

SomeTroglodyte:
> When will the madness end

Do you ***really*** want to know?
 <sub><sup>When entropy finally wins I'd guess.</sup></sub>

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
yairm210: Perhaps, that's a change I was thinking of as well
So far those seems to be the only usage it would have though, but if I see another place I'll make the change

PR #5956, opened 2022-01-12
Fix incorrect Longswordman obsolete

AdityaMH:


AdityaMH:
Since longswordman and musket strength is same, should i change it too?

Pros:
 - For balancing

Cons:
 - Make upgrade to musket become useless

yairm210:
Upgrade to musket is useful! It means you don't need resources!

ravignir:
I would not change anything in jsons (especially to custom values) without adjusting combat calculations to vanilla ones.

xlenstra:
I'm not sure whether these changes make a lot of sense. Musketman are, until the strength values are adjusted, objectively better than longswordman and the unique units replacing it. Having both available for building thus doesn't really make a lot of sense. When both have the same strength value, there is some merit to having to spend a resource to lower its cost slightly. However, as this won't happen until the combat calculations are redone, which depends on a lot of constants becoming moddable and probably having different formula's in some places all-together, I don't see the direct merit of this change.

yairm210:
Currently denied, at least until we change the battle system to correctly reflect Vanilla

PR #5959, opened 2022-01-12
Resolved #5930 - removed extraneous stat icons + resolved sorting problems

### LINKED ISSUE #5930: Some emoji stat icons appear in unexpected places
I'd assume most of these shouldn't be here.
Examples I've found, include:
- The empire stats overview screen:
image
- The tutorial section of the civilopedia
image
- The gold in the trade screen
image
- The ranking screen (though this is debatable)
image
- Cities named after a stat
image
- When founding a religion
image
- In a lot of places actually
image
image
image

There's probably a few more, but these were all that I could find

yairm210:
image

image

### DIFF
--- a/core/src/com/unciv/logic/trade/TradeOffer.kt
+++ b/core/src/com/unciv/logic/trade/TradeOffer.kt
@@ -45,7 +45,7 @@ data class TradeOffer(val name:String, val type:TradeType, var amount:Int = 1, v
         }.tr()
 
         if (type.numberType == TradeTypeNumberType.Simple || name == Constants.researchAgreement) offerText += " ($amount)"
-        else if (type.numberType == TradeTypeNumberType.Gold) offerText += " ($amount${Fonts.gold})"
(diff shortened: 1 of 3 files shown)


xlenstra (approved):

SomeTroglodyte:
`.replace(Fonts.happiness.toString(),"").replace(Fonts.faith.toString(),"")` - pretty kludge-y. Wouldn't have happened with the Stat.toString approach I suggested ages ago - though I didn't foresee this back then either, just had the gut feeling.

We should for now just remember and consider when similar spots crop up repeatedly.

yairm210:
The stat tostring won't catch many places where the stat name appears, though
Menus and Uniques for instance

SomeTroglodyte:
Of course, said so back then - much more manual work but less risk of regressions. As I said, I'm fine with the `tr` approach, just remember this kind of workaround and if they get out of hand consider a decoration-free `tr` variant instead.

Issue #5866, opened 2021-12-28
Multiplayer sync via filesystem?


will-ca:
Hm. Most popular network protocols that users regularly set up themselves have some kind of wrapper/virtual filesystem that exposes the server as a user-space directory.

So why not allow users to run multiplayer games via "folders" on their device? I think this would basically mean automatic support for SSH, FTP, Google Drive, NextCloud, SyncThing, Bonjour/Avahi, RDP, and basically every protocol ever, with no barriers for users to set up their own home or cloud servers running whatever system they want. 

I haven't looked at this part of the code. But since DropBox already exposes a filesystem-like interface, I cautiously assume code reuse may be plentiful and extra maintenance burden minimal. Not sure how filesystem access works on Android. But supporting nearly every protocol that will ever exist with such a relatively minor change (if it is indeed minor) is probably worth it IMO even if limited to desktop.

refola:
This looks like a simpler fix for #2679 than was suggested at the time. I expect there'd need to be a folder watching service for next turn notification, but dealing with files sounds much easier (Android `sdcardfs` mangling notwithstanding) if someone's up to implementing it.

yairm210:
Dropping this due to lack of interest. You can already save games to folders and create your own servers so if you want anything else, that's it of scope for me

Issue #2679, opened 2020-06-01
Change Off Dropbox


grravity:
Would be very appreciated to see a more foss friendly solution to multiplayer that doesn't involved dropbox. Maybe a VPS or Selfhosting. It's a shame because everything else is great!

lishaoxia1985:
I think onedrive is good choice because it can be used in China too.

refola:
Could it work with Syncthing? That way it's peer-to-peer instead of centralized. It would also automatically work on a LAN even without an Internet connection.

grravity:
If that's the case, using Bittorrent may work better for a P2P protocol, idk if that's even possible though.

refola:
> If that's the case, using Bittorrent may work better for a P2P protocol, idk if that's even possible though.

If by "Bittorrent" (a protocol made for massive P2P distribution of fixed, predetermined data) you mean "BitTorrent Sync", then you should know that it's now called "Resilio Sync" and, per [its Wikipedia article](https://en.wikipedia.org/wiki/Resilio_Sync), is proprietary and thus a non-starter for Unciv.

That said, embedding Syncthing could be overkill for Unciv when something smaller like [rsync](https://en.wikipedia.org/wiki/Rsync) is *much* smaller (< 1MB on my system vs 23MB for Syncthing; Unciv itself is 35MB). I have no idea how Unciv's syncing protocol works or if there's a smaller "library" or "embedded" version of Syncthing that could be used. I only mentioned Syncthing because that's what I'm familiar with using on my own systems and it's been really reliable for P2P syncing between *n* devices.

It may be worth skimming [the open-source section of Wikipedia's comparison of file synchronization software](https://en.wikipedia.org/wiki/Comparison_of_file_synchronization_software#Open-source). Both "DirSync Pro" and "SymmetricDS" are written in Java, which might help with embedding them into Unciv. :shrug:

natrius:
> Multiplayer takes advantage of Dropbox, which is non-free software, for syncing purposes.

I have not tried Multiplayer yet, but i read that part i quoted above and wanted to ask if you will make it possible with other service, Nextcloud for example as this would be an open source and free software. I could provide you with an account on my instance for development purposes, if needed. 
Just asking, because Nextcloud is working quite similiar to Dropbox, depends on what you need. But it could be a good alternative.

yairm210:
From what I see, it's not a good alternative.
It may be open source, but it's not free.

natrius:
[...]

PR #5960, opened 2022-01-12
Abstracted file storage from the multiplayer data retrieval - this sh…

yairm210:
…ould allow implementations of both #5866 and #2679 for other implementations of IFileStorage

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/DropBox.kt
@@ -116,8 +116,29 @@ object DropBox {
 
 }
 
+interface IFileStorage {
+    fun saveFileData(fileName: String, data: String)
+    fun loadFileData(fileName: String): String
+}
+
+class DropboxFileStorage:IFileStorage{
+    // This is the location in Dropbox only
+    fun getLocalGameLocation(gameId: String) = "/MultiplayerGames/$gameId"
+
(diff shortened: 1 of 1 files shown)


xlenstra (approved):
This looks surprisingly simple for something on first sight I'd assumed to be quite complex

yairm210:
The real mess isn't in implementing the file saving, it's in handling the different data stores. Do you allow multiple games each from a different store? I think yes, therefore each new multiplayer game you'll need to specify which data store it uses, you'll need a way to configure data stores, it's a whole new mess

will-ca:
Would anything be needed beyond a single UI "Use custom file server" button similar to the single-player "Load from custom location", and I guess either one or two properties in the game specifying the type and state of the FileStorage?

(Hm… I'm thinking maybe one string property for FileStorage type, and a second string property to either pass to a constructor/factory or deserialize as JSON for that type?)

will-ca (approved):

SomeTroglodyte:
I had always imagined the "store" would be one per-setup global. A player wouldn't be able to easily go back and forth from a game on the  "east-coast server" to one on the "west-coast server", but that's a small price to pay for simplicity???

yairm210:
Works for me, and works as a first step

xlenstra:
One downside of having only a single server to connect to is that you can only play games on that server. If, for example, you and a few friends have their own nextcloud server and play on that, you wouldn't also be able to play games with people from the discord.

PR #5969, opened 2022-01-14
Tint colour and not alpha in CityTileGroup.

### LINKED ISSUE #5954: Don't use alpha for unowned tiles in `CityTileGroup`.
#5874 made it possible for tilesets to blend tile images by overlapping each other.

That can sometimes look wrong in the CityScreen though, as some of tiles are rendered with transparency there so the overlapped regions accumulate instead of blending. The transparency can be somewhat designed around on the mod's end, but it would be cleaner to not need to do that (and instead set E.G. colour or CrossHatchHexagon in the `CityTileGroup` code).

I've meant to do this for a couple days now, so opening ticket and self-assigning just to say "This is a concern and I will be addressing it in a PR".

will-ca:
Closes #5954.

#5874 made it possible for tilesets to blend tile images by overlapping each other.

However, CityTileGroup uses alpha to dim tiles, which means that can lead to unexpected results. So this PR changes the dimming to work through non-alpha components instead.

Secondary effect: Fixes layered rendering in dimmed areas. Before, GDX seemed to do something weird to the children of a Group with fractional alpha, which looks more like additive or divisive blending than alpha compositing. (Something, something, straight premultiplied.) See: ghostly white hills, lime green forests, very soft iconographic overlays in Default, strangely bright boundaries between tiles.

Secondary effect: Terrain features that extend "above" the hex cover up the hexes to the north instead of being translucent. In a way this is more realistic, because it's like the entire tile is in shadow. See the mountain southeast of the city in the comparisons.

Visual difference: Alpha fadeout squeezes both white and black to the background colour at full strength. Multiplicative tinting remaps white but leaves black at black. Hence this PR may look darker in some cases (while preserving the same amount of contrast). See the Pix5 oceans.

Tested with FantasyHex, Default, (blended) 5Hex, and Pix5. Checked existing calls/assignments to the `.color`s being set, clicked around a bit on the CityScreen, didn't find any conflicts.

Previous is left. This PR is right. Clearer at full size:

Dim

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityTileGroup.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityTileGroup.kt
@@ -6,6 +6,7 @@ import com.unciv.logic.city.CityInfo
 import com.unciv.logic.map.TileInfo
 import com.unciv.ui.tilegroups.TileGroup
 import com.unciv.ui.tilegroups.TileSetStrings
+import com.unciv.ui.utils.BaseScreen
 import com.unciv.ui.utils.ImageGetter
 import com.unciv.ui.utils.centerX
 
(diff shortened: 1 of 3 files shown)


yairm210:
Nice job, both on the bug location and the resolution!

yairm210 (approved):

PR #5974, opened 2022-01-15
Capturing Civilians Not an Attack

itanasi:
Change so that attacking (capturing) Civilian units doesn't count as an attack.
Based on #4697

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -398,9 +398,12 @@ object Battle {
     private fun reduceAttackerMovementPointsAndAttacks(attacker: ICombatant, defender: ICombatant) {
         if (attacker is MapUnitCombatant) {
             val unit = attacker.unit
-            if (!defender.isCivilian()) {  // capturing Civilian doesn't count as attack
-                unit.attacksThisTurn += 1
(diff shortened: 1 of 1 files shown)


xlenstra (approved):

yairm210:
Does capturing civilians not reduce movement points?

ravignir:
> Does capturing civilians not reduce movement points?

capturing an unit is not an action in civ5. It is just something that automatically happens when you move a military unit into a tile already occupied by an enemy civilian unit.

https://user-images.githubusercontent.com/30041793/149635629-f54ba569-764d-4ed2-8e15-b879945d8048.mp4

itanasi:
The previous code had it that capturing only happened during an attack. Which would also set you to 0 movement. My change says it doesn't count as an attack (not attacksThisTurn increments) and doesn't set you to 0 movement. But you'll still use the movement point for moving to the new tile as part of your attack.

A more robust solution is capturing automatically as part of movement. But that requires the other Ranged Capture code merge.

ravignir:
I thnk there is a problem: you can capture an anit even if you are not at war :|

yairm210:
That's not part of this pr, that's a different pr
This one is about expending movement

ravignir:
right, I had them two mashed together :D

Issue #3477, opened 2020-12-28
separate "capture" button next to "attack" button


b-1:
sometimes instead of capturing a worker / settler with melee unit, you just need to kill him. now you can only capture and dissolve it. but during a battle, a tile with an enemy builder may not be safe. I want to deprive the enemy of a worker, but I do not want to expose my unit to fire by occupying this tile. you need to give the player a choice (grab or attack) using two separate buttons.

on the other hand, the ranged unit will be able not only to kill workers, but also to capture them. as well as capturing cities too. what could be done in the original Civ5 by giving the order to move (with left menu), and not by clicking on the unit itself, where the attack action is by default.

krushia:
Would a melee unit now be able to gain XP from attacking a worker?

ravignir:
I have run couple of tests in civ5:
- Archers and any ranged unit can move onto a tile with worker/settler etc and capture it. CAPTURE(move) action is the default action for ranged units.
- Great Person can not be captured, they instantly die when enemy military unit enters the tile.
- Only capturable units are: Settler (turns ino worker), Worker, Missionary, Archaeologist, Great Prophet.
- Capturing an unit costs no movement points and unit can still attack another military unit.
- Melee units can't attack civilians.
- Civilian units always lose 40 HP when attacked by ranged attack (regardless if they are attacked by the city/archer/rocket artillery). This will also earn XP points.

b-1:
I think any unit has the right to both capture and attack civilians or a city.

I'm not sure about gaining experience for such an attack. but, I think, if a unit spent its turn on an attack, then it should get experience for it. on the other hand, they offer no resistance. but with a ranged attack, the enemies also do not fight back.

ravignir:
Not in civ5. All ranged units have "may not melee attack" unique/promotion, which makes them unable to attack cities.

Also you DO NOT ATTACK a civilian unit when you capture it, you just move to the same tile - and depending on the civilian unit - you either automatically capture it or automatically destroy (this provides no XP). Any military unit that can move, can also capture a civilian.

Alternatively you can attack any civilian unit (but only with RANGED attack) and it will always deal 40 damage to them. This will make your unit gain XP.

So no, melee unit can not attack civilian units at all.

ravignir:
![obraz](https://user-images.githubusercontent.com/30041793/103281490-2bff2280-49d3-11eb-9014-f640ff81adb2.png)

[...]

PR #5975, opened 2022-01-15
Ranged capture

itanasi:
Adds ability for Ranged Units to Move onto unguarded Civilian Unit
If a unit moves on top of a Civilian Unit, execute a captureCivilian action (this should be free and thus consistent if also merging #5974 )
This still allows Ranged Units to Attack an unguarded Civilian unit. So it can be destroyed, all movement used, but doesn't move the Ranged Unit. That is, no change to Ranged Battle code.


xlenstra (approved):

yairm210:
And what does the battle table display to the user when you attempt to do this?

itanasi:
I should have included a screen shot.

You'll have two options. There's the Battle Tile dialog to attack (and thus kill the Civilian unit). Or you can move to the tile and capture the Civilian Unit as part of movement.  This is a bit inconsistent with melee units still doing the attack says Capture prompt. I could rework the whole system to be more consistent.

itanasi:
image

Obviously this breaks the whole can't move through enemy unit though...

itanasi:
Beware, this merge now also edits one of the tests. Highly recommend reviewing!

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -619,6 +619,9 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
         if (!unit.canEnterForeignTerrain && !tile.canCivPassThrough(unit.civInfo)) return false
 
         val firstUnit = tile.getFirstUnit()
+        // allow Ranged Units to capture Civilian
+        if (firstUnit != null && firstUnit.isCivilian() && unit.isRanged() )
```
yairm210, line 623: This is missing a check that the tile doesn't contain a military unit - don't count on getFirst to return the military unit if there is one

yairm210: ALSO
you don't check if the unit is off an ENEMY civ! You shouldn't be able to capture allied civilians with impunity

itanasi:
Hm, didn't think about all of that. My bad.

Guess I should turn this into a revamp and move all civilian captures to movement. But also then add all the necessary checks and how it will trigger War as well.

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
itanasi: > This is missing a check that the tile doesn't contain a military unit - don't count on getFirst to return the military unit if there is one

Why is that not guaranteed? Is this a multithreading thing? When I read getFirst() seemed it went in order.

yairm210: It is said that "programmers won't depends on what your library says it does, that will depend on observed results. Ordered results will be relief on to be ordered, and random ordering will be taken as a source of randomness".
Basically - if we want to be about to change the implementation of getFirst, we need to ensure that functions don't depend on a byproduct of its internal logic.

itanasi:
I get the need for robustness and can make the change. But I guess I don't
understand then what the promised behavior is of getFirst() since the name
implies an ordering.

will-ca:
> But I guess I don't understand then what the promised behavior is of getFirst() since the name implies an ordering.

Looks like it's meant for blocking movement through enemies. So in that case the promised/specified/relevant behaviour is that it will return a unit (belonging to the correct civInfo) if there are any units, or `null` if there are no units.

Compare to `.first()` and `.firstOrNull()` on unordered Kotlin `Iterable`s like `HashSet`. The "ordering" is "iteration order", but "iteration order" itself doesn't have any promises.

itanasi:
Ah, thanks for the explanation! Makes sense.

Still new to Kotlin and it's conventions.

yairm210:
> Hm, didn't think about all of that. My bad.
> 
> Guess I should turn this into a revamp and move all civilian captures to movement. But also then add all the necessary checks and how it will trigger War as well.

There's no need for that, we can easily have the tile move-through-able only if it contains a single, enemy, civilian unit, just by adding the 'enemy' check

itanasi:
That would be functionally correct. But I'm starting to feel like there's a
lot of extra checks and exceptions. So want to see if unifying Civilian
Capture folded into the movement code (for both melee and ranged) would be
more robust.

Especially since we now also have Ranged Attack is 40hp and Civilian Cap
isn't an Attack. Basically fully close @ravignir original ticket. #3477

itanasi:
Pretty sure the logic is correct now.
Running on an emulated save with a bunch of AIs in Modern Era, hard to tell how long canPassThrough takes since the AIs seem to take different tacts/choices.

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -618,8 +619,14 @@ class UnitMovementAlgorithms(val unit:MapUnit) {
         if (!unit.canEnterForeignTerrain && !tile.canCivPassThrough(unit.civInfo)) return false
 
         val firstUnit = tile.getFirstUnit()
-        if (firstUnit != null && firstUnit.civInfo != unit.civInfo && unit.civInfo.isAtWarWith(firstUnit.civInfo))
-            return false
+        if (firstUnit != null) {
```
yairm210, line 622: I would put the `unit.civInfo != firstUnit.civInfo` check into the if as well

yairm210 (approved):
Definitely different, but this method allows the AI to both ignore capturable units and to do multiple things per turn, so well done!

xlenstra (approved):

PR #5976, opened 2022-01-15
Explain when cities can't be razed.

### LINKED ISSUE #5971: Destroy option no respond & can't function in the city menu
**Platform**
Android
**Version**
Play store,3.18.16
**Describe the bug**
Destroy city option can't function in city menu
**To Reproduce**
1. Go to 'City menu'
2. Click on 'Destroy city' 
3.See error

```

```
</details>

**Expected behavior**
'Destroy city' should function properly like usual
**Screenshots**
Screenshot_20220115-085548

**Additional context**

will-ca:
Closes #5971.


android/assets/jsons/translations/French.properties
```diff
@@ -6137,8 +6137,8 @@ Unlike Luxury Resources, each Strategic Resource on the map provides more than o
...
...n be mitigated by building a courthouse in the city, returning the citizen's unhappiness to normal.\n\nPuppeting the city will mean that you have no control on the city's production.\nThe city will not increase your tech or policy cost, but its citizens will generate 1.5x the regular unhappiness.\nA puppeted city can be annexed at any time, but annexed cities cannot be returned to a puppeted state!\n\nRazing the city will lower its population by 1 each turn until the city is destroyed!\nYou cannot raze a city that is either the starting capital of a civilization or the holy city of a religion. = Fantochiser la ville signifie que vous n'avez pas de contrôle sur la production de la ville.\n la ville n'augmentera pas le coût de vos Recherches ni de vos Doctrines, mais ses citoyens génèreront 1,5x le mécontement habituel.\nAnnexer la ville vous donnera le control sur la production, mais doublera le mécontentement de ses citoyens !\n Cela peut être résolu en construisant un Tribunal dans la ville, ramenant le mécontentement des citoyens à son niveau normal.\nUne ville fantoche peut être annexée à tout moment, cependant, une fois annexée, elle ne pourra pas redevenir une ville fantoche !
```
xlenstra, line 6141: The downside of changing the text in the .proporties files, is that now the existing translations will not be correct. My French isn't great, but it looks like even though the English string was changed, the French was not, and thus this translation doesn't correctly translate the string anymore. It's also not visible to the translators that this is the case, as the translation writer will not add a ` # requires translation` before it. The translators themselves will probably also not spot it, unless they switch between English and their language in-game on the exact screen this string is used, which is very unlikely.

will-ca: The two translations changed always appear right next to each other in the UI, and I mostly just reordered some sentences around (and added the new sentences about razing and liberating), so while the translations are no longer verbatim they do still say the same general things/make sense in the UI.

Hm. If I manually add a `# Requires update`, would it be preserved by the translation file writer? Maybe I could add "(Requires new translation!)" to the end of the translation string itself, so that way the existing and almost-complete translations aren't lost but translators can also see that it's changed?

will-ca: Right, I added "(Translation update needed!)" into each of the translation strings, so the change is highlighted to translators without not throwing away the existing translations.

yairm210:
Why does the 'liberating a city' start with an enter?

will-ca:
> Why does the 'liberating a city' start with an enter?

It's always show immediately after the "When conquering a city[…]", so the new line at the start separates them:

image

Significant because I added internal "\n\n"s to separate the descriptions of the four conquest actions, so not having the leading line break would mean the space between the intro and the descriptions is smaller than the space between each of the descriptions.

yairm210 (approved):

PR #2114, opened 2020-03-10
Translation refactoring - part 2 (Tutorials)

JackRainy:
Here I continue the idea initiated in the PR #2096 .

The goal is to get rid of multilanguage additional files like `Nations_Language.json`, `Tutorials_Language.json`, also to minimize the `template.properties` and the manual work of its maintaining.
In the long term perspective it can be even removed (or minimized as much as possible) while `Language.properties` will be generated fully automatically (and localized by translators, of course).

**BEFORE:**
json-properties-before

**AFTER:**
json-properties

The technical improvements in this PR:
* `Tutorials_Languages.json` files are removed
* Their translations are migrated to `Language.properties`
** Notice: Nothing changed to the `template.properties` !
* The `Language.properties` are automatically generated based on `template.properties` + `Nations.json` + `Tutorials.json`
* The `Tutorials.json` is simplified: no more *maps of Array of Arrays* :)
* The loader of tutorials does not refer to the file (the slowest operation of possibles!) each time when the dialog should be displayed.
* etc.

How it works now (notice the update of the city name and the tutorials messages):

PR #5977, opened 2022-01-15
Update English.properties. (Remove no-op translations.)

will-ca:
REGEX search: `^(.+) = \1`

Skipped the AfterConquering tutorial strings as they would conflict with #5976.

Four translations remain in English.properties, which seem to add spaces/remove a word from "WaterCivilian", "WaterMelee", "WaterRanged", and "WaterSubmarine".

Found and may fix one apparent case where the template string was updated at some point without updating the English translation, which would have probably led to the updated version being hidden. ("Cities cannot be built within 3 tiles of…")

I don't know what the original reason was for duplicating so many strings… Seems a holdover from an old refactor of the translations system (#2114), maybe a convention that was never adopted or a precaution that became unnecessary?

### DIFF
--- a/android/assets/jsons/translations/English.properties
+++ b/android/assets/jsons/translations/English.properties
@@ -1321,12 +1321,12 @@ Stop automation =
 Construct road = 
  # Requires translation!
 Fortify = 
-Fortify until healed = Fortify until healed
+Fortify until healed = 
  # Requires translation!
 Fortification = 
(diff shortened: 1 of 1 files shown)


yairm210:
merge translations

yairm210:
Translations branch created

yairm210:
Translations PR created

PR #5978, opened 2022-01-15
Made attacked civilians lose 40 hp as in civ5.

ravignir:
Is it the smallest commit ever? xD

### DIFF
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -307,6 +307,7 @@ object BattleDamage {
                 attacker,
                 defender
             )
+        if (defender.isCivilian()) return 40
         return (damageModifier(ratio, false) * getHealthDependantDamageRatio(attacker)).roundToInt()
     }
 


yairm210:
Not enough, since battle take still reports that you will kill the unit if you make this attack

ravignir:
I don't understand , what is the problem here? :|
All it changes is the damage dealt to a civilian unit to always equal 40. As it works now, the civilian unit has 0 strength (as no assigned strength defaults to 0), meaning it always loses 100 HP. If you assign civilian unit strength value in json, it will behave exactly like any other military unit.
az

yairm210:
The problem is that I'm not sure what the label is when a ranged unit attacks a civilian, if it's "Killed" or something then that needs to be changed.

ravignir:
there is no label.
z

yairm210:
Well OK then :)

yairm210 (approved):

AdityaMH:
> there is no label. z

How about great people or embarked unit test?
I'm afraid, civilian unit always on 40 HP. Not losing 40 HP.

itanasi:
Glad someone else did this fix for me

itanasi:
Wait, this makes so returned value for an attack is 40hp. Which is good for Ranged. But right now melee "attacks" capture. Is that still handled OK? Need to reread the code.

AdityaMH:
> Wait, this makes so returned value for an attack is 40hp. Which is good for Ranged. But right now melee "attacks" capture. Is that still handled OK? Need to reread the code.

That's what i'm afraid before.

ravignir:
> Wait, this makes so returned value for an attack is 40hp. Which is good for Ranged. But right now melee "attacks" capture. Is that still handled OK? Need to reread the code.

Does not affect melee, since their attack counts as a capture. All this commit changes is damage dealt to civilians from 100 (results from 0 strength calculations) to 40.

Another thing is to delete line that makes embarked strength not apply to civilians (as it does apply in civ5). So I probably should change battledamage.kt once again if I were to implement this as well (so it only returns 40 damage when civilian unit has 0 strength)

itanasi:
Embarked overall needs an overhaul right? There's a note about military and
civilian taking up the same slot while embarked. I have to spin up Civ5 to
refresh on the logic

itanasi:
Hm, will need to double check all my logic for Civilian Capture.

In general though we want to stick with Melee Attacks Civilian is capture right? Not have an option to hit for 40hp and stay on tile? Or separate out Melee Attack vs Melee Capture? Feel like Civ 5 was always Melee did capture

PR #5982, opened 2022-01-15
civInfo.exploredTiles immutability

yairm210:
exploredTiles being mutable has been causing concurrency errors when autosaving for a while now. it's time to come to terms with the fact that exploredTiles needs to be immutable as well, and switched out upon change.

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -154,7 +154,9 @@ class CivilizationInfo {
     // This is basically a way to ensure our lists are immutable.
     var cities = listOf<CityInfo>()
     var citiesCreated = 0
-    var exploredTiles = HashSet<Vector2>()
+    /** Do not add tiles to this directly - instead, create a copy, add to the copy, and switch to retain immutability
+     * Unfortunately we can't make this a setOf since the json parser doesn't know what to make of existing data... */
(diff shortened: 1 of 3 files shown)


yairm210:
I need to check this - if every time a unit passes through a tile we create a new hashmap this could have severe performance implications

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -240,7 +242,7 @@ class CivilizationInfo {
         // This is the only thing that is NOT switched out, which makes it a source of ConcurrentModification errors.
         // Cloning it by-pointer is a horrific move, since the serialization would go over it ANYWAY and still lead to concurrency problems.
         // Cloning it by iterating on the tilemap values may seem ridiculous, but it's a perfectly thread-safe way to go about it, unlike the other solutions.
-        toReturn.exploredTiles.addAll(gameInfo.tileMap.values.asSequence().map { it.position }.filter { it in exploredTiles })
+        toReturn.exploredTiles = exploredTiles
```
SomeTroglodyte, line 245: Wouldn't this change invalidate the comments above?

yairm210: Correct

yairm210:
Perhaps not. This doesn't seem worth the risk.

PR #5989, opened 2022-01-16
Resolved #5988 - Allow modded improvements

### LINKED ISSUE #5988: Stupid workers in DeCiv Redux won't build Moisture traps
A persistent bug I run into in DeCiv is that automated workers won't build Moisture traps. They will say "Worker has no work to do" even when there are multiple areas on the map where they could build a Moisture trap, which would make that area more useful.

On this map, there is a Rubble area to the southwest of Inheritance and a Ruins area to the west of Accord. Both of these tiles could benefit by having a Moisture trap added to them. But the automated workers insist they have no work to do.
[StupidWorkers.zip](https://github.com/yairm210/Unciv/files/7877558/StupidWorkers.zip)

yairm210:
Current worker automation assumes A LOT about the current ruleset.
This is only step 1, which allows skipping irrelevant improvements to reach actually useful improvements.
Comments welcome.

### DIFF
--- a/core/src/com/unciv/logic/automation/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/WorkerAutomation.kt
@@ -339,14 +339,19 @@ class WorkerAutomation(
         }
 
         // turnsToBuild is what defines them as buildable
-        val tileImprovements = ruleSet.tileImprovements.filter { it.value.turnsToBuild != 0 }
+        val tileImprovements = ruleSet.tileImprovements.filter {
+            it.value.turnsToBuild != 0 && tile.canImprovementBeBuiltHere(it.value, tile.hasViewableResource(civInfo)) }
         val uniqueImprovement = tileImprovements.values
             .firstOrNull { it.uniqueTo == civInfo.civName }
 
+        val currentlyBuildableImprovements = tileImprovements.values.filter { tile.canBuildImprovement(it, civInfo) }
+        val bestBuildableImprovement = currentlyBuildableImprovements.map { Pair(it, Automation.rankStatsValue(it, civInfo)) }
+            .filter { it.second > 0f }
+            .maxByOrNull { it.second }?.first
+
         val improvementString = when {
             tile.improvementInProgress != null -> tile.improvementInProgress!!
-            improvementStringForResource != null && tileImprovements.containsKey(improvementStringForResource)
-                    && tileImprovements[improvementStringForResource]!!.turnsToBuild != 0 -> improvementStringForResource
+            improvementStringForResource != null && tileImprovements.containsKey(improvementStringForResource) -> improvementStringForResource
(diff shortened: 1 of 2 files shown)


xlenstra (approved):

core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -362,11 +367,16 @@ class WorkerAutomation(
...
-            tile.baseTerrain in listOf(Constants.grassland, Constants.desert, Constants.plains) -> "Farm"
-            tile.isAdjacentToFreshwater -> "Farm"
-            tile.baseTerrain in listOf(Constants.tundra, Constants.snow) -> Constants.tradingPost
+            tile.terrainFeatures.contains(Constants.forest) && tileImprovements.containsKey("Lumber mill") -> "Lumber mill"
+            tile.isHill() && tileImprovements.containsKey("Mine") -> "Mine"
+            tile.baseTerrain in listOf(Constants.grassland, Constants.desert, Constants.plains)
+                    && tileImprovements.containsKey("Farm") -> "Farm"
+            tile.isAdjacentToFreshwater && tileImprovements.containsKey("Farm") -> "Farm"
+            tile.baseTerrain in listOf(Constants.tundra, Constants.snow) && tileImprovements.containsKey(Constants.tradingPost)
+                -> Constants.tradingPost
```
SomeTroglodyte, line 376: Maybe moving the literals to `Contants` and marking them as would-be-nice-to-get-rid-of there? With a nice name that makes it clear they're automation kludges? And: We could avoid the repeated re-allocation of those listOf's while at it?

Question: The love of contains/containsKey - is `in` not bytecode equivalent and more readable even for Map/ArrayList? Ctrl-B suggests the "equivalent" part is true - so it's a matter of taste?

yairm210: They're both equivalent, for 'key in map' is not as intuitive, since a map for me is made of pairs, not of keys
Agree about the single creation of lists :)
Constants would also be something to track globally - every hardcoded improvement name is a another impediment to modding

yairm210: Actually, you know what? No.
I hope to be able to replace all of this with weighing the *yield* of the improvements, making this moot.

yairm210: But that's for step 2 :)

SomeTroglodyte: The `in` question was meant globally, that's all over the project. Thanks for explaining.
And "moot" - precisely, but transitioning to symbols recognized by compiler and studio indexer as a stepping stone can help catch oversights. We've done it before.

SomeTroglodyte (approved):

PR #5992, opened 2022-01-17
Refactor and unify some buttons that have a label next to an icon.

will-ca:
This is something that shows up a fair bit in the UI.

Everything should be mostly visually equivalent:
- Some icons are sized at `18f` instead of `20f` due to using the `fontSize`.
- The padding in the Promotions and Improvement picker screens are a bit smaller (and unified with the Great Person picker).
- The trade deal item buttons have icons on the left instead of the right.

…Hm. I'm also noticing that some of the font sizes look smaller, but I'm just using the same `18` font size in those cases that's also the default in `.toLabel(…)`, so it seems that the two existing overloads for `.toLabel()` behave inconsistently, with the zero-arg one producing slightly bigger text. That may be an existing bug. Some of the WorldScreen top bar labels are bigger than the others too, depending on which `.toLabel()` they use. Will try to diagnose/document and open a separate PR or issue.

After one rebase I ran into a mysterious issue where it seemed Gradle wasn't adding `:core` to the classpath of `:desktop` or something… It spat out dozens of "Unresolved reference"s at first, and then it seemed to finish the build but crash with `NoClassDefFoundError: com/unciv/UncivGame` on run. Seems fine now.

### DIFF
--- a/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
+++ b/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
@@ -223,10 +223,7 @@ class CivilopediaScreen(
 
         var currentX = 10f  // = padLeft
         for (categoryKey in categoryToEntries.keys) {
-            val button = Button(skin)
-            if (categoryKey.headerIcon.isNotEmpty())
-                button.add(ImageGetter.getImage(categoryKey.headerIcon)).size(20f).padRight(5f)
-            button.add(categoryKey.label.toLabel())
+            val button = categoryKey.label.toButton(icon = if (categoryKey.headerIcon.isNotEmpty()) ImageGetter.getImage(categoryKey.headerIcon) else null)
             button.addTooltip(categoryKey.key)
 //            button.style = ImageButton.ImageButtonStyle(button.style)
             button.onClick { selectCategory(categoryKey) }
--- a/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
+++ b/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
@@ -223,7 +223,8 @@ class CivilopediaScreen(
 
         var currentX = 10f  // = padLeft
         for (categoryKey in categoryToEntries.keys) {
-            val button = IconTextButton(categoryKey.label, if (categoryKey.headerIcon.isNotEmpty()) ImageGetter.getImage(categoryKey.headerIcon) else null)
+            val icon = if (categoryKey.headerIcon.isNotEmpty()) ImageGetter.getImage(categoryKey.headerIcon) else null
+            val button = IconTextButton(categoryKey.label, icon)
             button.addTooltip(categoryKey.key)
 //            button.style = ImageButton.ImageButtonStyle(button.style)
             button.onClick { selectCategory(categoryKey) }
--- a/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
+++ b/core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
@@ -223,7 +223,7 @@ class CivilopediaScreen(
 
         var currentX = 10f  // = padLeft
         for (categoryKey in categoryToEntries.keys) {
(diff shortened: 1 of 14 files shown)


core/src/com/unciv/ui/civilopedia/CivilopediaScreen.kt
```diff
@@ -223,10 +223,7 @@ class CivilopediaScreen(
 
         var currentX = 10f  // = padLeft
         for (categoryKey in categoryToEntries.keys) {
-            val button = Button(skin)
-            if (categoryKey.headerIcon.isNotEmpty())
-                button.add(ImageGetter.getImage(categoryKey.headerIcon)).size(20f).padRight(5f)
-            button.add(categoryKey.label.toLabel())
+            val button = categoryKey.label.toButton(icon = if (categoryKey.headerIcon.isNotEmpty()) ImageGetter.getImage(categoryKey.headerIcon) else null)
```
yairm210, line 226: Separate the icon to a `val` before this line

core/src/com/unciv/ui/overviewscreen/EmpireOverviewScreen.kt
```diff
@@ -54,12 +54,7 @@ class EmpireOverviewScreen(private var viewingPlayer:CivilizationInfo, defaultPa
                 game.settings.addCompletedTutorialTask("See your stats breakdown")
             game.settings.lastOverviewPage = name
         }
-        val button = Button(skin)
-        if (iconAndKey.icon != "") {
-            val image = ImageGetter.getImage(iconAndKey.icon)
-            button.add(image).size(20f).pad(5f)
-        }
-        button.add(name.toLabel(Color.WHITE)).pad(5f)
+        val button = name.toButton(icon = if (iconAndKey.icon != "") ImageGetter.getImage(iconAndKey.icon) else null)
```
yairm210, line 57: Same here

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -196,18 +184,57 @@ fun Float.toPercent() = 1 + this/100
...
-/** Translate a [String] and make a [Button] widget from it, with control over font size, font colour, and an optional icon. */
-fun String.toButton(fontColor: Color = Color.WHITE, fontSize: Int = 24, icon: String? = null): Button {
+/**
+ * Builder-like configuration scope for [String.toButton].
+ *
+ * Properties: [button], [icon], [label], [iconCell], [labelCell]
+ *
+ * @param I [Actor] subclass of the icon given to [String.toButton].
+ * */
+interface IIconAndLabelCellsInButton<I: Actor?> {
```
yairm210, line 194: There is no need for this to be an interface if the only usage is to generate a single-instance class out of it.
Also, I doubt we're actually using all these cells. Please clean up all fields that are not in actual use - we can always add them later if needed.
The most maintainable code is that which isn't written :)

yairm210: Also, do we actually use any non-Actor functions of the icon anywhere? If not then we don't even need the templating. 
This is starting to look more like `IconLabelButton : Button(BaseScreen.skin){...}` which is much simpler.

core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
```diff
@@ -88,4 +89,16 @@ open class PickerScreen(disableScroll: Boolean = false) : BaseScreen() {
     fun removeRightSideClickListeners() {
         rightSideButton.clearListeners()
     }
+
+    companion object {
+        /** Icon size used in [getPickerOptionButton]. */
+        const val pickerOptionIconSize = 30f
+        /** Return a button for picker screens that display a list of big buttons with icons and labels. */
+        fun <I: Actor?> getPickerOptionButton(icon: I, label: String): Button {
```
yairm210, line 97: Here too we don't need renovating, we can accept an Actor

yairm210:
Okay! Looks good!

core/src/com/unciv/ui/pickerscreens/PickerScreen.kt
will-ca: Right. Pitfalls of automatic refactor. I skimmed the diff a couple times to try to find needed changes, but clearly missed this. :)

yairm210 (approved):

PR #5995, opened 2022-01-17
Missed a translation template.

will-ca:
Lol.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -429,6 +429,7 @@ Anything above 40 may work very slowly on Android! =
 
 # Multiplayer
 
+? = 
 Username = 
 Multiplayer = 
 Could not download game! = 


SomeTroglodyte:
"Lol"? Land of (the) Living? Lavender orange lime? Lusty old Limericks? Latu Orada Likto?
At least give a hint where that "?" is relevant - ?

will-ca:
Ah, sorry. This is a continuation of #5993 and a side quest of #5994. The "?" is used for the help info bubble in MultiplayerScreen. <kbd>CTRL</kbd>+<kbd>SHIFT</kbd>+<kbd>F</kbd> finds one more use in ImageGetter.getRandomNationIndicator().

Currently the use in MultiplayerScreen doesn't get translated. #5994 changes that.

The thinking is that some languages may not use the question mark in the same way, or may find it clearer to use a different symbol or a word. I tested with English.properties that the single character is successfully translated.

*However*, I'm now considering that the use in `MultiplayerScreen` meaning "Help" and the use in `getRandomNationIndicator` meaning "unknown quantity" may not take the same translations… Perhaps the MultiplayerScreen use should be switched to the word "Help" instead?

yairm210:
The multiplayer screen should have "help" and not "?"
If I were a player and entered the multiplayer screen, a "?" button would do little to dispel my confusion...

yairm210:
But I would add a harceded comment before this describing that it's the character on the Random civ

PR #5996, opened 2022-01-17
Remove `String.toLabel()` overload, fix uneven fonts?

will-ca:
I noticed the two overloads seem to produce different font sizes. `String.toLabel(…)` with the default arguments is not the same as `String.toLabel()`. The `.fontScale` properties don't make much sense either:

Before:
image
After:
image

Also visible in the WorldScreenTopBar. The turn count, happiness, and resource labels are slightly bigger than the rest of the stat labels:

Before:
image
After:
image

`String.toLabel()` commit: https://github.com/yairm210/Unciv/commit/9ba360408c8f8b631c04b0e37c2ad59613769462
`String.toLabel(fontColor: Color, fontSize: Int)` commit: https://github.com/yairm210/Unciv/commit/de3038253681c3597801503d9e80954ac4ce4cd0

Test code:

```kotlin
val labelTable = Table()
val label1 = "ZERO-ARG".toLabel()
labelTable.add(label1)
val label2 = "DEFAULT-ARGS".toLabel(Color.WHITE, 18)
labelTable.add(label2).row()
labelTable.add("fontScale: ${label1.fontScaleX}, ${label1.fontScaleY}".toLabel(fontSize = 10))
labelTable.add("fontScale: ${label2.fontScaleX}, ${label2.fontScaleY}".toLabel(fontSize = 10)).row()
labelTable.background = ImageGetter.getRoundedEdgeRectangle(Color.BLACK)
```

The zero-arg one apparently has ~227 uses plus at least another ~43 through `Int.toLabel()`, compared to ~97 for the configurable one. Not sure about call frequency. …Well, it's for building the UI, so I suppose the tiny CPU difference isn't going to be noticeable. Biggest concern aside from consistency and maintainability would perhaps be memory use, then? But it looks like the configurable one is already reusing stuff when it can. I suppose the other options are to either 1. Ignore the uneven spacing or 2. Keep both overloads and harmonize them.

### DIFF
--- a/core/src/com/unciv/ui/utils/ExtensionFunctions.kt
+++ b/core/src/com/unciv/ui/utils/ExtensionFunctions.kt
@@ -211,12 +211,10 @@ fun String.toButton(fontColor: Color = Color.WHITE, fontSize: Int = 24, icon: St
     return button
 }
 
-/** Translate a [String] and make a [Label] widget from it */
-fun String.toLabel() = Label(this.tr(), BaseScreen.skin)
 /** Make a [Label] widget containing this [Int] as text */
 fun Int.toLabel() = this.toString().toLabel()
 
-/** Translate a [String] and make a [Label] widget from it with a specified font color and size */
+/** Translate a [String] and make a [Label] widget from it with an optionally specified font color and size */
 fun String.toLabel(fontColor: Color = Color.WHITE, fontSize: Int = 18): Label {
     // We don't want to use setFontSize and setFontColor because they set the font,
     //  which means we need to rebuild the font cache which means more memory allocation.
--- a/core/src/com/unciv/ui/worldscreen/bottombar/BattleTable.kt
+++ b/core/src/com/unciv/ui/worldscreen/bottombar/BattleTable.kt
@@ -107,7 +107,7 @@ class BattleTable(val worldScreen: WorldScreen): Table() {
         add(attackerNameWrapper)
 
         val defenderNameWrapper = Table()
-        val defenderLabel = Label(defender.getName().tr(), skin)
+        val defenderLabel = defender.getName().toLabel()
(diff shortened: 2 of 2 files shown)


SomeTroglodyte:
(Not in-depth analyzed idea): You're not actually fixing the cause of the discrepancy. From your proof that there _is_ a discrepancy and looking at the code for the umpteenth time I'm getting the suspicion that trying a sequence of toLabels with a linear fontSize might yield nonlinear actual sizes around the eighteen... If not, the only explanation I could think of would be that the skin used is not identical.

will-ca:
The vastly different `fontScale`s originally put me off investigating any further— Makes it seem like there's an entire different scaling/sizing system, seemingly completely disconnected from the ones I could see but coincidentally almost correct just by luck— Easier to just switch completely to the method which I already know works and is widely used in the code.

Anyway, it turns out it's not so complicated. The default widget styles defined in BaseScreen try to assign some widgets to use 18 scale by default and some to use 20 scale by default. `String.toLabel()` uses 18, hence the discrepeancy. Normalizing them fixes it:

https://github.com/yairm210/Unciv/blob/ec59533f82ba07ee3038032045bdd8d3774d4aa7/core/src/com/unciv/ui/utils/BaseScreen.kt#L96-L103

(Side note: If I'm reading this correctly, actually only the last `.apply{}` will have any effect? That would explain why the `18` for `LabelStyle` doesn't seem to work.)

Will PR tomorrow. Still, I think it may be better to also remove the zero-arg overload. Once this is fixed, its behaviour will be a strict subset of the configurable one's.

SomeTroglodyte:
Sound like you hit the nail on the head. And yes, I would also guess there's a single font object that gets its scale changed back and forth - but I'd check in the debugger to be sure.

will-ca:
Closing in favour of new PR.

PR #5998, opened 2022-01-18
Worker AI for mods, step 2

yairm210:
Support for removing terrain features, both when we need to get rid of them for a resource, and when they plain reduce the yield of a tile

Note that the "Remove <improvement>" was implemented WAY before templating and uniques were a thing - the correct solution in these enlightened times would be to separate the improvement *name* from the *effect*, so we could have e.g. a "Deforestation" improvement which would contain botth ["Removes [Jungle]", "Removes [Forest]"] or whatnot.

For now I Constant'D the "Remove " so we can at least follow where it's used.

### DIFF
--- a/core/src/com/unciv/logic/automation/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/WorkerAutomation.kt
@@ -333,18 +333,12 @@ class WorkerAutomation(
 
         // turnsToBuild is what defines them as buildable
         val tileImprovements = ruleSet.tileImprovements.filter {
-            it.value.turnsToBuild != 0 && tile.canImprovementBeBuiltHere(
-                it.value,
-                tile.hasViewableResource(civInfo)
-            )
-        }
+            it.value.turnsToBuild != 0 && tile.canImprovementBeBuiltHere(it.value, tile.hasViewableResource(civInfo)) }
         val uniqueImprovement = tileImprovements.values
             .firstOrNull { it.uniqueTo == civInfo.civName }
 
-        val currentlyBuildableImprovements =
-            tileImprovements.values.filter { tile.canBuildImprovement(it, civInfo) }
-        val bestBuildableImprovement =
-            currentlyBuildableImprovements.map { Pair(it, Automation.rankStatsValue(it, civInfo)) }
+        val currentlyBuildableImprovements = tileImprovements.values.filter { tile.canBuildImprovement(it, civInfo) }
+        val bestBuildableImprovement = currentlyBuildableImprovements.map { Pair(it, Automation.rankStatsValue(it, civInfo)) }
                 .filter { it.second > 0f }
                 .maxByOrNull { it.second }?.first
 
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -357,25 +363,23 @@ class WorkerAutomation(
...
             // I think we can assume that the unique improvement is better
             uniqueImprovement != null && tile.canBuildImprovement(uniqueImprovement, civInfo)
                     && unit.canBuildImprovement(uniqueImprovement, tile) ->
                 uniqueImprovement.name
 
-            tile.terrainFeatures.contains("Fallout") -> "Remove Fallout"
-            tile.terrainFeatures.contains(Constants.marsh) -> "Remove Marsh"
+            lastTerrain.let {
+                isUnbuildableAndRemovable(it) && Automation.rankStatsValue(it, civInfo) < 0
+            } -> Constants.remove + lastTerrain.name
```
xlenstra, line 374: Note that fallout doesn't have negative yield anymore, but instead has the unique "Nullifies all other stats this tile provides", so it won't be caught by this check

yairm210:
The real question now is - given that we rank improvements by stats now, which of the orderings aren't ordered correctly?

Let's check!

- Jungle - since this is the only buildable improvement anyway, bestBuildableImprovement will be trading post
- Oasis - this should be converted to a general case of "if buildable improvements is empty return null" beforehand
- Forest and lumber mill - same, this is the only buildable improvement
- Hill and Mine... HMMM I don't recall if Grassland Hill tiles can build Farm, but if they CAN, then this is ordered backwards. @ravignir ?
- And... that's it! Automation will prefer Farm (1.2f) over all other default starting improvements so we're golden!

core/src/com/unciv/logic/automation/WorkerAutomation.kt
yairm210: Ayyyy good catch

ravignir:
>Hill and Mine... HMMM I don't recall if Grassland Hill tiles can build Farm, but if they CAN, then this is ordered backwards. @ravignir ?
>
Farms can be build only if the hill is next to fresh water tile. What about terrace farms?
Also what about trading posts in forests?

yairm210:
A. My goal here is to reproduce the current order, not necessarily to improve the ordering, although that's interesting as well :)
B. Terrace farms both in old and proposed solutions come before mines
C. Trading posts in forests are valued much lower than mines for good reason! 1 Production is *much* better than 1 gold!
D. Are Trading Posts buildable in Forests in Civ V? Because in Unciv they aren't currently (sounds like a different PR)

yairm210:
What say you, though? In your experience, is a Farm or a Mine better for a freshwater-adjacent hill?
(and don't "it depends", I'm asking in general)

xlenstra:
> What say you, though? In your experience, is a Farm or a Mine better for a freshwater-adjacent hill? (and don't "it depends", I'm asking in general)

I'd say, and I'm sorry for this, it depends. If you have a city with lots of hills, a farm is obviously better, in a city with lots of grasslands a mine would be better. However, as hills come in many different areas (snow, desert) where they may be fresh-water but no grasslands, I'd say that in general a farm would most of the time be better.

core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -357,25 +364,24 @@ class WorkerAutomation(
 
             // Defence is more important that civilian improvements
             // While AI sucks in strategical placement of forts, allow a human does it manually
-            !civInfo.isPlayerCivilization() && evaluateFortPlacement(tile, civInfo, false) -> Constants.fort
+            !civInfo.isPlayerCivilization() && evaluateFortPlacement(tile, civInfo,false) -> Constants.fort
```
xlenstra, line 367: Very minor point but basically the only thing of this PR I don't agree with :)

xlenstra (approved):

core/src/com/unciv/logic/automation/WorkerAutomation.kt
yairm210: That's what I get for trying to undo auto-formatting...

ravignir:
> 

farm is better.

yairm210:
If so then the next stage, getting rid of all hardcoded references, is as good as done :)

PR #5999, opened 2022-01-18
Iterate once through all uniques to find those relevant to sat percent bonuses, instead of per-type

yairm210:
The same change we made in the move to getStatsFromUniquesBySource()
This should be faster, more generic, more accurate (buildings vs wonders), and catch all usecases (i.e. stat bonuses from uniques attached to *techs* or who knows what else


xlenstra (approved):

yairm210:
This is of course step 1 out of 2 - setting the groundwork
Step 2 is taking ALL the uniques from getStatPercentBonusesFromUniques() and moving them to the new function, removing the references to the old function completely

SomeTroglodyte:
I guess nothing to do with satellites, just a missing 't' in the title :)

PR #6002, opened 2022-01-19
Update Indonesian.properties

Kensvin28:


yairm210:
Placeholder `[baseUnitFilter]` not found in `Indonesian` for entry `[stat] cost of purchasing [baseUnitFilter] units in cities [amount]%`

PR #6003, opened 2022-01-19
Removed ALL hardcoded decisions in worker AI - we're now fully moddable!

yairm210:
I tested this out with RekMOD, 150 simulated turns, to see if this was working.
Not only did I discover that workers were squatting on Antiquity Sites since they had resource improvements that they couldn't build, I also discovered something much worse.
Even after fixing that, most cities were woefully underimproved.
Turns out, the construction automation would limit worker construction to a measly *3 workers* even for a *10 city* civ!
After removing this limitation and making civs aim for a 1:1 ratio between cities and workers, everything started looking much, much better.
I'm not sure what the exact effect on the AI will be but I'm _sure_ that this leads to a major improvement. More improved tiles means more stats means more everything.

Before:
image

(yes, that's the CAPITAL you see barren before you)

After:
image

(results achieved by starting 4 civs, large map, RekMOD, and picking the one with the most cities after 150 turns)

(I just noticed that the second image is turn 166, which means I spent 16 turns following the automated workers to see what they were doing, but the point stands. Stats in everything are higher, you can see that the amount of cities for all civs except for the obligatory Culture one is higher, I think this is a huge boost in AI productivity.

### DIFF
--- a/core/src/com/unciv/logic/automation/ConstructionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/ConstructionAutomation.kt
@@ -168,12 +168,9 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
                 && it.isBuildable(cityConstructions)
                 && Automation.allowSpendingResource(civInfo, it) }
         if (workerEquivalents.isEmpty()) return // for mods with no worker units
-        if (civInfo.getIdleUnits().any { it.isAutomated() && it.hasUniqueToBuildImprovements })
-            return // If we have automated workers who have no work to do then it's silly to construct new workers.
 
-        val citiesCountedTowardsWorkers = min(5, cities) // above 5 cities, extra cities won't make us want more workers
-        if (workers < citiesCountedTowardsWorkers * 0.6f && civUnits.none { it.hasUniqueToBuildImprovements && it.isIdle() }) {
-            var modifier = citiesCountedTowardsWorkers / (workers + 0.1f)
+        if (workers < cities) {
+            var modifier = cities / (workers + 0.1f) // The worse our worker to city ratio is, the more desperate we are
             if (!cityIsOverAverageProduction) modifier /= 5 // higher production cities will deal with this
             addChoice(relativeCostEffectiveness, workerEquivalents.minByOrNull { it.cost }!!.name, modifier)
         }
--- a/core/src/com/unciv/logic/automation/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/WorkerAutomation.kt
@@ -333,15 +333,21 @@ class WorkerAutomation(
     private fun chooseImprovement(unit: MapUnit, tile: TileInfo): TileImprovement? {
 
         // turnsToBuild is what defines them as buildable
-        val tileImprovements = ruleSet.tileImprovements.filter {
-            it.value.turnsToBuild != 0 && tile.canImprovementBeBuiltHere(it.value, tile.hasViewableResource(civInfo)) }
-        val uniqueImprovement = tileImprovements.values
-            .firstOrNull { it.uniqueTo == civInfo.civName }
+        val potentialTileImprovements = ruleSet.tileImprovements.filter {
+            unit.canBuildImprovement(it.value, tile)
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -333,15 +333,21 @@ class WorkerAutomation(
...
 
         // turnsToBuild is what defines them as buildable
-        val tileImprovements = ruleSet.tileImprovements.filter {
-            it.value.turnsToBuild != 0 && tile.canImprovementBeBuiltHere(it.value, tile.hasViewableResource(civInfo)) }
-        val uniqueImprovement = tileImprovements.values
-            .firstOrNull { it.uniqueTo == civInfo.civName }
+        val potentialTileImprovements = ruleSet.tileImprovements.filter {
+            unit.canBuildImprovement(it.value, tile)
+                    && tile.canImprovementBeBuiltHere(it.value, tile.hasViewableResource(civInfo))
+                    && (it.value.uniqueTo == null || it.value.uniqueTo == unit.civInfo.civName)
```
xlenstra, line 339: Would it be better for performance to do this check first?

xlenstra (approved):

core/src/com/unciv/logic/automation/WorkerAutomation.kt
yairm210: Perhaps, I considered this an edge case but you're right

core/src/com/unciv/logic/automation/ConstructionAutomation.kt
```diff
@@ -168,12 +168,9 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
                 && it.isBuildable(cityConstructions)
                 && Automation.allowSpendingResource(civInfo, it) }
         if (workerEquivalents.isEmpty()) return // for mods with no worker units
-        if (civInfo.getIdleUnits().any { it.isAutomated() && it.hasUniqueToBuildImprovements })
-            return // If we have automated workers who have no work to do then it's silly to construct new workers.
```
will-ca, line 172: Huh. Neat. So I guess it was possible to flat-out disable an AI's improvement construction by trapping a worker in a bubble inside your borders and killing the other two. (Not a review comment, but I find the exploit interesting.)

core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -367,20 +378,12 @@ class WorkerAutomation(
...
             uniqueImprovement != null && tile.canBuildImprovement(uniqueImprovement, civInfo)
-                    && unit.canBuildImprovement(uniqueImprovement, tile) ->
-                uniqueImprovement.name
+                -> uniqueImprovement.name
 
             lastTerrain.let {
                 isUnbuildableAndRemovable(it) &&
                         (Automation.rankStatsValue(it, civInfo) < 0 || it.hasUnique(UniqueType.NullifyYields) )
             } -> Constants.remove + lastTerrain.name
-            tile.terrainFeatures.contains(Constants.jungle) -> Constants.tradingPost
```
will-ca, line 377: Android Studio says this is the only use of Constants.tradingPost, so can it be removed from Constants.kt?

core/src/com/unciv/logic/automation/WorkerAutomation.kt
```diff
@@ -333,15 +333,21 @@ class WorkerAutomation(
     private fun chooseImprovement(unit: MapUnit, tile: TileInfo): TileImprovement? {
 
         // turnsToBuild is what defines them as buildable
-        val tileImprovements = ruleSet.tileImprovements.filter {
-            it.value.turnsToBuild != 0 && tile.canImprovementBeBuiltHere(it.value, tile.hasViewableResource(civInfo)) }
-        val uniqueImprovement = tileImprovements.values
-            .firstOrNull { it.uniqueTo == civInfo.civName }
```
will-ca, line 339: Nitpicky, but I guess the comment is outdated?

will-ca (approved):

will-ca:
None of the outstanding comments point out anything that would stop this from working well. I'll get the linting in a separate PR myself instead of slowing down the merge process if it matters to me.

PR #6005, opened 2022-01-19
Added button to update an installed mod from its action menu

yairm210:
Much easier and more intuitive than clicking on it, then on the equivalent online mod, then 'download'


SomeTroglodyte:
Nice. Dunno why I decided against back then. I have the feeling, however, those `?.` came from testing, which would mean the `!!` might be easily crash-able. Tried selecting a non-installed mod _while_ an update downloads?

yairm210:
Yes, but I'll try and run some stressier tests

PR #6008, opened 2022-01-22
Removed a probably unnecessary check in unit.mayPassThrough()

xlenstra:
This also has as an advantage that units can move through non-coastal cities between lakes, e.g. fixing
image

I've run a few quick tests which showed that moving water units doesn't immediately lead to weird behaviour, but I have not simulated long games or done anything in-depth.


yairm210 (approved):

will-ca (approved):
Also comment outdated, but meh.

PR #6012, opened 2022-01-22
Miscellaneous linting, Unique typifying, refactoring, bugfix/consistency to avoid broken links to disabled religious UBs.

will-ca:
Started as nitpicky linting after #6003 and #6008.

Commit history is change list. May be easier to review per commit.

It's my vague presumption that terrains seem to be gradually moving away from being hardcoded?

Also typified a Unique that was in Constants.kt… Which then led to some other refactoring… And then a bugfix/consistency thing. First time actually even touching the Uniques system, I think? I did test to see that everything seems to still work, but please make sure I haven't disenfunctionalitified it. Though I guess it won't break game mechanics either way. (Original PRs: #3359, #4907… Looks like they were before `UniqueType` was added in #5222.)

This grew a bit out of hand as many things seem to… I could try splitting off the later and bigger commits into a separate PR if that would be preferred, but I'm not sure if they would be applicable to the current `master` on their own or if they depend on the current commits… Though the entire PR is 1:1 translations of existing expressions so hopefully it's alright.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -572,9 +572,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
 
     AvailableAfterCertainTurns("Only available after [amount] turns", UniqueTarget.Ruins),
     HiddenWithoutReligion("Hidden when religion is disabled", UniqueTarget.Unit, UniqueTarget.Building, UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
-    HiddenBeforePantheon("Hidden before founding a Pantheon", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
-    HiddenAfterPantheon("Hidden after founding a Pantheon", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
-    HiddenAfterGreatProphet("Hidden after generating a Great Prophet", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenBeforePantheon("Hidden before founding a Pantheon", UniqueTarget.Ruins),
+    HiddenAfterPantheon("Hidden after founding a Pantheon", UniqueTarget.Ruins),
+    HiddenAfterGreatProphet("Hidden after generating a Great Prophet", UniqueTarget.Ruins),
     HiddenWithoutVictoryType("Hidden when [victoryType] Victory is disabled", UniqueTarget.Building, UniqueTarget.Unit, flags = listOf(UniqueFlag.HiddenToUsers)),
     HiddenFromCivilopedia("Will not be displayed in Civilopedia", *UniqueTarget.values(), flags = listOf(UniqueFlag.HiddenToUsers)),
     
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -569,13 +569,12 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     //endregion
 
     ///////////////////////////////////////////// META /////////////////////////////////////////////
-    
-    
+
+    AvailableAfterCertainTurns("Only available after [amount] turns", UniqueTarget.Ruins),
     HiddenWithoutReligion("Hidden when religion is disabled", UniqueTarget.Unit, UniqueTarget.Building, UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -571,12 +571,13 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags:
...
-    HiddenBeforePantheon("Hidden before founding a Pantheon", UniqueTarget.Ruins),
-    HiddenAfterPantheon("Hidden after founding a Pantheon", UniqueTarget.Ruins),
-    HiddenAfterGreatProphet("Hidden after generating a Great Prophet", UniqueTarget.Ruins),
-    AvailableAfterCertainTurns("Only available after [amount] turns", UniqueTarget.Ruins),
-    HiddenWithoutVictoryType("Hidden when [victoryType] Victory is disabled", UniqueTarget.Building, UniqueTarget.Unit),
+    HiddenWithoutReligion("Hidden when religion is disabled", UniqueTarget.Unit, UniqueTarget.Building, UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenBeforePantheon("Hidden before founding a Pantheon", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenAfterPantheon("Hidden after founding a Pantheon", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenAfterGreatProphet("Hidden after generating a Great Prophet", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    AvailableAfterCertainTurns("Only available after [amount] turns", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
```
yairm210, line 578: This actually sounds like something the users should be aware of, no?

yairm210 (approved):

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
will-ca: Right.. Oops. Overzealous copy-pasting…

SomeTroglodyte:
A quick scan gives the impression a) thorough b) from what it does, risk-free or close. But - my _first_ association was: `UniqueFlag.HiddenToUsers` is missing an `r` to make it a proper Wallace & Gromit reference...

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -569,14 +569,14 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags:
...
-    HiddenBeforePantheon("Hidden before founding a Pantheon", UniqueTarget.Ruins),
-    HiddenAfterPantheon("Hidden after founding a Pantheon", UniqueTarget.Ruins),
-    HiddenAfterGreatProphet("Hidden after generating a Great Prophet", UniqueTarget.Ruins),
+
     AvailableAfterCertainTurns("Only available after [amount] turns", UniqueTarget.Ruins),
-    HiddenWithoutVictoryType("Hidden when [victoryType] Victory is disabled", UniqueTarget.Building, UniqueTarget.Unit),
+    HiddenWithoutReligion("Hidden when religion is disabled", UniqueTarget.Unit, UniqueTarget.Building, UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenBeforePantheon("Hidden before founding a Pantheon", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenAfterPantheon("Hidden after founding a Pantheon", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
+    HiddenAfterGreatProphet("Hidden after generating a Great Prophet", UniqueTarget.Ruins, flags = listOf(UniqueFlag.HiddenToUsers)),
```
xlenstra, line 577: impo these should also be visible to the user

will-ca: I think I'm inclined to agree. But that would also take this PR out of "linting" territory into changing existing behaviour, and I think it already has enough not-quite-related changes mixed in.

I'll take a closer look at what they're actually used for, whether it might make more sense to say "Unavailable" instead of "Hidden" if they'll be user-visible, etc.

will-ca: Right, unhid. But I haven't found anywhere those Uniques would be displayed in the UI anyway, and if a detailed "Ruin Rewards" Civilopedia section or something is added at some point then I think they should probably say "Unavailable" instead of "Hidden".

SomeTroglodyte: The current ""Ruin Rewards" Civilopedia" is only Improvements - Ancient ruins, TileImprovements.kt line 166+... And doesn't show text from Uniques?

will-ca: AFAICT It just shows the effects/names:

image

PR #6014, opened 2022-01-22
Fix uneven fonts, unify font sizes.

will-ca:
Does two things:
* Fixes uneven font sizes, as in #5996.
* Moves the 13 occurrences of `18` and 11 occurrences of `24` to Constants, to make future consistency easier.

May also be easier to review by commit. First one actually changes logic; Second one just swaps out a lot of integer literals for a constant.

### DIFF
--- a/core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
+++ b/core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
@@ -6,6 +6,7 @@ import com.badlogic.gdx.scenes.scene2d.Actor
 import com.badlogic.gdx.scenes.scene2d.Group
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.utils.Align
+import com.unciv.Constants
 import com.unciv.UncivGame
 import com.unciv.models.metadata.BaseRuleset
 import com.unciv.models.ruleset.Ruleset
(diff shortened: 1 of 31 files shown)


yairm210 (approved):

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
```diff
@@ -48,7 +49,7 @@ class FormattedLine (
     val extraImage: String = "",
     /** Width of the [extraImage], height is calculated preserving aspect ratio. Defaults to available width. */
     val imageSize: Float = Float.NaN,
-    /** Text size, defaults to 18. Use [size] or [header] but not both. */
+    /** Text size, defaults to [Constants.defaultFontSize]. Use [size] or [header] but not both. */
     val size: Int = Int.MIN_VALUE,
```
xlenstra, line 53: Is this changed in another piece of code, or does this default to `Int.MIN_VALUE` instead of `[Constants.defaultFontSize]`?

yairm210: Hm good point. This SHOULD be set for everyone, but why not avoid problems we didn't think of prematurely?

xlenstra (approved):

core/src/com/unciv/ui/civilopedia/CivilopediaText.kt
will-ca: I didn't investigate too closely, but the stuff I saw in my peripheral vision as I did the one-to-one replacements led me to believe that Int.MIN_VALUE as a kind of `null`-like value that gets replaced with the default size.

SomeTroglodyte: Precisely. Used as placeholder for "not set". Same with Float.NaN just above.

PR #5955, opened 2022-01-12
Typed all policy uniques

xlenstra:
Turns out there are a lot of uniques in the `policies.json` files.
All of them (except the timed bonus, which I'll handle separately sometime) should be typed now.
Also reworded a few to get rid of + and - signs outside of parameters, as well as turning 'increased by [amount]' into '[+amount]', etc.

PR #6016, opened 2022-01-22
Try to fix a potentially miscapitalized/broken policy cost unique.

will-ca:
Oh yeah. This was the other PR that I wanted to submit today.

---

The UniqueType had a lower-case "policies", but its uses in the JSONs mostly had capitals.

I *think* that can cause problems? Not sure. They're case-sensitive, right? I also mentioned this on #5955, but I worded it in a very nitpicky way and mixed it in with a bunch of less consequential things.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -203,8 +203,10 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags: 
     BonusHappinessFromLuxuryDeprecated("+[amount] happiness from each type of luxury resource", UniqueTarget.Global),
     LessPolicyCostFromCities("Each city founded increases culture cost of policies [amount]% less than normal", UniqueTarget.Global),
     LessPolicyCost("[amount]% Culture cost of adopting new Policies", UniqueTarget.Global),
-    @Deprecated("As of 3.18.17", ReplaceWith("[amount]% Culture cost of adopting new policies"))
+    @Deprecated("As of 3.18.17", ReplaceWith("[amount]% Culture cost of adopting new Policies"))
     LessPolicyCostDeprecated("Culture cost of adopting new Policies reduced by [amount]%", UniqueTarget.Global),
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -202,7 +202,7 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags:
     @Deprecated("As of 3.18.17", ReplaceWith("[+amount] Happiness from each type of luxury resource"))
     BonusHappinessFromLuxuryDeprecated("+[amount] happiness from each type of luxury resource", UniqueTarget.Global),
     LessPolicyCostFromCities("Each city founded increases culture cost of policies [amount]% less than normal", UniqueTarget.Global),
-    LessPolicyCost("[amount]% Culture cost of adopting new policies", UniqueTarget.Global),
+    LessPolicyCost("[amount]% Culture cost of adopting new Policies", UniqueTarget.Global),
     @Deprecated("As of 3.18.17", ReplaceWith("[amount]% Culture cost of adopting new policies"))
```
xlenstra, line 206: This `ReplaceWith` should also be changed then, and it might be worthwhile to also add a deprecated unique type without the capitalized letter so mod makers are aware that something has changed.

yairm210: Precisely what I was going to say :)
Remember that we're now a semi-platform - there are mods running on the logic of this game, so any change that would require changes in the mods means allowing a 'deprecation time' for the old functionality.
For Uniques, this means 
 - Creating a new unique
 - Marking the old one as 'deprecated as of <next version>' with a warning
 - Having both uniques still work

Then, after about a month (usually), we'll actually remove it, which means:
 - Marking the @Depecated as an ERROR and not a WARNING
 - Marking the version of removal on the Deprecation attribute
 - Removing references in code

Stage 3, which is really removing this completely so it doesn't appear even as a warning or in documentation, is then done like 2 months later, That's why all deprecated uniques have version of deprecation, and version of removal.

will-ca: Right…. I think the new commit should take care of this?

yairm210 (approved):

PR #6017, opened 2022-01-23
Typified some building uniques

yairm210:
I noticed that the code allows "[stats] from every [buildingFilter]" but the UniqueType was limited to "[stats] from every [buildingName]", so now that that's available we don't actually need the "[stats] from every Wonder" unique :)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -75,14 +75,18 @@ enum class UniqueType(val text:String, vararg targets: UniqueTarget, val flags:
     StatsFromCitiesOnSpecificTiles("[stats] in cities on [terrainFilter] tiles", UniqueTarget.Global, UniqueTarget.FollowerBelief),
     @Deprecated("As of 3.18.14", ReplaceWith("[stats] [in all cities] <before discovering [tech]> OR [stats] [in all cities] <before adopting [policy]>"))
     StatsFromCitiesBefore("[stats] per turn from cities before [tech/policy]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
-
-
+    StatsFromBuildings("[stats] from all [buildingFilter] buildings", UniqueTarget.Global, UniqueTarget.FollowerBelief),
+    @Deprecated("As of 3.19.1", ReplaceWith("[stats] from every [Wonder]"))
+    StatsFromWondersDeprecated("[stats] from every Wonder", UniqueTarget.Global, UniqueTarget.FollowerBelief),
+    StatsWithResource("[stats] with [resource]", UniqueTarget.Building),
```
xlenstra, line 81: If this only applies to buildings, shouldn't it be under the buildings section of uniques instead?

yairm210: Correct!

xlenstra (approved):

PR #6019, opened 2022-01-23
Start bias includes neighboring tiles for better effect

yairm210:
This also means we don't need a special coastal check, since it's caught by the regular check
IIRC there are nations in Original Civ that have 'mountain' start bias - if so this will also enable that :)

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -12,6 +12,8 @@ import com.unciv.models.ruleset.ModOptionsConstants
 import com.unciv.models.ruleset.Ruleset
 import com.unciv.models.ruleset.RulesetCache
 import com.unciv.models.ruleset.unique.UniqueType
+import com.unciv.models.translations.equalsPlaceholderText
+import com.unciv.models.translations.getPlaceholderParameters
 import java.util.*
 import kotlin.collections.HashMap
 import kotlin.collections.HashSet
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -436,13 +438,20 @@ object GameStarter {
...
+                startBias.equalsPlaceholderText("Avoid []") -> {
+                    val tileToAvoid = startBias.getPlaceholderParameters()[0]
+                    preferredTiles.filter { tile ->
+                        !tile.getTilesInDistance(1).any {
+                            it.matchesTerrainFilter(tileToAvoid)
+                        }
+                    }
                 }
-                startBias == Constants.coast -> preferredTiles.filter { it.isCoastalTile() }
                 startBias in tileMap.naturalWonders -> preferredTiles  // passthrough: already failed
```
xlenstra, line 449: (not specific to this PR, but this just caught my eye)
Doesn't `tileMap.naturalWonders` only contain the wonders that are actually placed on this map? If so, aren't those the exact same that would have been filtered before this `for`-loop, effectively making this dead code? I'd assume you want to check if they are a natural wonder, and if so, not do anything, but that would be `ruleSet.terrains[startBias]!!.type == TerrainType.NaturalWonder`, right?

Additionally, the code above this PR for determining the starting space next to natural wonders has a variable `wonderNeighbor` which would make it likely to contain the tiles adjacent to natural wonders. However, unless I read the code incorrectly, it only finds the exact tile the wonder is on

xlenstra (approved):

will-ca (approved):

SomeTroglodyte:
Didn't @SimonCeder code something to make start biases much friendlier, with that region stuff?

yairm210:
He did, but this is still the function that filters tiles based on nation biases

SimonCeder:
I'll point out that this function is used for placing starts on pre-made maps that don't have specified start locations.
When playing on a generated map all the regions code is run and returns a map with already defined start locations for each player, so the placement in GameStarter is not used in that case.

PR #6020, opened 2022-01-23
Typed some uniques, added more examples for parameters in unique documentation

xlenstra:
Also minor changes, such as replacing `uniqueObjects.any { it.isOfType(UniqueType) && condition }` with `getMatchingUniques(UniqueType).any { condition }`, etc.


android/assets/jsons/Civ V - Gods & Kings/Policies.json
```diff
@@ -18,7 +18,7 @@
 			},
 			{
 				"name": "Oligarchy",
-				"uniques": ["Units in cities cost no Maintenance", "+[50]% attacking strength for cities with garrisoned units"],
+				"uniques": ["Units in cities cost no Maintenance", "[+50]% Strength for garrisoned cities <when attacking>"],
```
yairm210, line 21: <for garrisoned cities> sounds like a conditional - I recall there's a happiness bonus for cities with garrison somewhere, and this also leads nicely into the 'ruleset-wide uniques with conditionals' that's been brewing for a while

desktop/src/com/unciv/app/desktop/UniqueDocsWriter.kt
```diff
@@ -16,17 +16,36 @@ class UniqueDocsWriter {
...
+                .replace("[promotion]","[Shock I]")
+                .replace("[era]", "[Ancient era]")
+                .replace("[improvementName]", "[Trading Post]")
                 .replace("[resource]", "[Iron]")
                 .replace("[beliefType]", "[Follower]")
+                .replace("[belief]","[God of War]")
+                .replace("[foundingOrEnhancing]", "[founding]")
+                .replace("[tech]", "[Agriculture]")
+                .replace("[specialist]","[Merchant]")
+                .replace("[policy]", "[Olichargy]")
```
yairm210, line 47: Oligarchy

yairm210 (approved):

will-ca:
Seems to have broken Medical Labs and Aqueducts.

will-ca:
Okay. Right. So I think it's a mix of both. The typified Unique is used in ConstructionAutomation, but not in PopulationManager or any of the JSONs. So the net effect would be that Aqueducts and Medical Labs still work, but not using the typed Unique, while AI and 

…Hm. But also, it actually looks like this was *already* the case, and this PR just typified one of the two disconnected Uniques, allowing the fuzzy spelling checker to detect that something was fishy.

PR #6022, opened 2022-01-23
Converted stat list to stat tree

yairm210:
current changes do not affect UI at all, since we're still going by the shallow mapping that existed beforehand

The problem:
image

I have no way of knowing which policies, techs, and wonders are contributing to my stat production.
Solution: Turn all lists of stats into trees of stats! This will also help us in the overview screen to dig deeper.

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -84,10 +84,10 @@ class CityConstructions {
     /**
      * @return [Stats] provided by all built buildings in city plus the bonus from Library
      */
-    fun getStats(): Stats {
-        val stats = Stats()
+    fun getStats(): StatTreeNode {
+        val stats = StatTreeNode()
         for (building in getBuiltBuildings())
-            stats.add(building.getStats(cityInfo))
+            stats.addStats(building.getStats(cityInfo), building.name)
         return stats
     }
 
--- a/core/src/com/unciv/logic/city/CityStats.kt
+++ b/core/src/com/unciv/logic/city/CityStats.kt
@@ -32,6 +32,13 @@ class StatTreeNode {
         children[childName]!!.addStats(newStats, *hierarchyList.drop(1).toTypedArray())
     }
 
+    fun add(otherTree: StatTreeNode){
+        for ((key, value) in otherTree.children) {
+            if (!children.containsKey(key)) children[key] = value
+            else children[key]!!.add(value)
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


yairm210:
image

Bless my autism
I don't think there's a good reason to break down the tile stats and specialists, since the city screen gives you a plain view of both of these from the get-go

yairm210:
We can do the exact same with the percentages, but let's do this one step at a time... this PR itself could already have been split into 2 PRs, one for each of the commits 😅

xlenstra:
I'm a bit worried about clutter, especially if this same thing is planned for the bonuses stat list...
image

yairm210:
Hrm you're right, that's a lot to take in, but then again - if someone is in the stat drilldown tab I assume they'd prefer the option to really drill down.
Maybe a toggle?

xlenstra:
> Hrm you're right, that's a lot to take in, but then again - if someone is in the stat drilldown tab I assume they'd prefer the option to really drill down. Maybe a toggle?

Most user intuitive would probably be a toggle for each category, similar to an expander tab but for just 'buildings' or 'policies' instead of for the entirety of culture. That's probably a lot of work to implement, so a global toggle might also suffice

Additionally, in your screenshot of the second number the values don't seem to add up? 7 + 11 + 5 + 3 = 26, not 38, while 7 + 11 + 2 + 3 + 2 + 3 + 5 + 5 = 38, but that double counts all the new shown values.

yairm210:
expanding_drilldown

will-ca:
Could there be a visual indicator for the toggle? Triangle icon next to the header, maybe? I think I would be a bit confused by the changes in layout until/unless I figured it out.

yairm210:
How would you indicate this? A triangle to me indicates 'up' and 'down', not 'expand' and 
'contract', and leaves me more confused than I started.

will-ca:
ExpanderTab already uses a right-pointing triangle to indicate "Closed", and a down-pointing triangle to indicate "Opened".

I think "+"/"-" or OtherIcons/New and OtherIcons/Close might work too.

yairm210:
expanding_drilldown_with_button

SomeTroglodyte:
I don't think there's anything to prevent _nesting_ ExpanderTab's?? Making the inner one visually smaller should also be mostly prepared for.

xlenstra (approved):

yairm210:
@SomeTroglodyte That won't keep the tidy organization of the inner table though. And I'm not overly fond of the way buildings currently expand, and that's nested expandertabs, so...

PR #6024, opened 2022-01-23
Repo wiki resync. — Requires #5854.

will-ca:
This is a version of #5854 that also includes the changes to the wiki since then.

Separate PR to avoid having to re-review and re-test everything from #5854.

Once #5854 is merged, the diff for this PR should include only Markdown changes. (I may need to rebase again onto `master` first.)

Checked by visually comparing diffs of Wiki and this branch side-by-side.

Oh, I think I also had a code editor add-on that automatically reformatted and updated the Table of Contents on one of hte pages… The result looks fine though.


yairm210:
Ayy self conflict

will-ca:
Cool. Rebase fixed it without me even having to do anything.

xlenstra (approved):

will-ca:
Well now… Let's see if this works.

will-ca:
update wiki

will-ca:
Hoorah!

…Oh, did I not mention the `.mde` files? AFAICT, those are just backups created by `sed` due to the REGEXes used to translate link formats from AS/GH tree format to GH Wiki format. They can be safely ignored.

I'll skim the diff filtered to `*.md`s to make sure there's nothing being clearly broken or accidentally reverted.

will-ca:
@yairm210 The `update wiki` run above also added a footer on all the Wiki pages saying to PR against `/docs/wiki` in the main repo instead. I think the Wiki should be [made publicly read-only](https://docs.github.com/en/communities/documenting-your-project-with-wikis/changing-access-permissions-for-wikis) now, as any edits to it would only conflict with and eventually be overwritten by the next `update wiki` PR (but still recoverable from the Wiki commit history).

PR #6027, opened 2022-01-23
Check rulesets for potential typos.

will-ca:
Follow-up to #6016.

image

<details><summary>Some algorithm test code.</summary>

```kotlin
fun testTextDist(text1: String, text2: String) {
    println()
    println(text1)
    println(text2)
    println("Absolute: ${getTextDistance(text1, text2)}\nRelative: ${getRelativeTextDistance(text1, text2)}")
}

testTextDist(
        "The quick brown fox jumps over the lazy dogs.",
        "The quick brown fox jumps over the lazy dogs."
) // This is actually a really bad/atypical example due to lack of repeating characters.
testTextDist(
        "The quick brown fox jumps over the lazy dogs.",
        "The quick Brown fox jumps over the Lazy dogs."
)
testTextDist(
        "The quick brown fox jumps over the lazy dogs.",
        "The quic brwn fox jumpsover the lazy dogs"
)
testTextDist(
        "The wolf had the fox with him, and whatsoever the wolf wished, that the fox was compelled to do, for he was the weaker, and he would gladly have been rid of his master.",
        "Here are not many people—and as it is desirable that a story–teller and a story–reader should establish a mutual understanding as soon as possible, I beg it to be noticed that I confine this observation neither to young people nor to little people, but extend it to all conditions of people: little and big, young and old: yet growing up, or already growing down again—there are not, I say, many people who would care to sleep in a church. I don’t mean at sermon–time in warm weather (when the thing has actually been done, once or twice), but in the night, and alone."
) // Wikisource random.
testTextDist(
        "[]% Culture cost of adopting new Policies",
        "[]% Culture cost of adopting new policies"
) // https://github.com/yairm210/Unciv/pull/6016
testTextDist(
        "[]% maintenance cost for buildings []",

        "[]% maintenance cost for buildings []"
)
testTextDist(
        "[]% maintenance cost for buildings []",

        "[]% maintenance cost forBuildings []"
) // Does poorly because missing space followed by wrong character means it skips all the way to the next space.
testTextDist(
        "[]% maintenance cost for buildings []",

        "[]% maintenance cost forbuildings []"
)
testTextDist(
        "[]% maintenance cost for buildings []",

        "[]% maintenaince cost for buildings []"
)
testTextDist(
        "[]% maintenance cost for buildings []",

        "+[]%maintenaince cost for buildings []"
)
```
</details>

<details><summary>Algorithm test output.</summary>

```
The quick brown fox jumps over the lazy dogs.
The quick brown fox jumps over the lazy dogs.
Absolute: 0
Relative: 0.0

The quick brown fox jumps over the lazy dogs.
The quick Brown fox jumps over the Lazy dogs.
Absolute: 2
Relative: 0.044444444444444446

The quick brown fox jumps over the lazy dogs.
The quic brwn fox jumpsover the lazy dogs
Absolute: 3
Relative: 0.06976744186046512

The wolf had the fox with him, and whatsoever the wolf wished, that the fox was compelled to do, for he was the weaker, and he would gladly have been rid of his master.
Here are not many people—and as it is desirable that a story–teller and a story–reader should establish a mutual understanding as soon as possible, I beg it to be noticed that I confine this observation neither to young people nor to little people, but extend it to all conditions of people: little and big, young and old: yet growing up, or already growing down again—there are not, I say, many people who would care to sleep in a church. I don’t mean at sermon–time in warm weather (when the thing has actually been done, once or twice), but in the night, and alone.
Absolute: 651
Relative: 1.7690217391304348

[]% Culture cost of adopting new Policies
[]% Culture cost of adopting new policies
Absolute: 1
Relative: 0.024390243902439025

[]% maintenance cost for buildings []
[]% maintenance cost for buildings []
Absolute: 0
Relative: 0.0

[]% maintenance cost for buildings []
[]% maintenance cost forBuildings []
Absolute: 19
Relative: 0.5205479452054794

[]% maintenance cost for buildings []
[]% maintenance cost forbuildings []
Absolute: 1
Relative: 0.0273972602739726

[]% maintenance cost for buildings []
[]% maintenaince cost for buildings []
Absolute: 1
Relative: 0.02666666666666667

[]% maintenance cost for buildings []
+[]%maintenaince cost for buildings []
Absolute: 3
Relative: 0.08
```
</details>

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -311,7 +311,7 @@ class Ruleset {
                 val similarUniques = UniqueType.values().filter { getRelativeTextDistance(it.placeholderText, unique.placeholderText) <= uniqueMisspellingThreshold }
                 val equalUniques = similarUniques.filter { it.placeholderText == unique.placeholderText }
                 if (equalUniques.isNotEmpty()) {
-                    lines.add(
+                    lines.add( // This should only ever happen if a bug is or has been introduced that prevents Unique.type from being set for a valid UniqueType, I think.
                         "$name's unique \"${unique.text}\" looks like it should be fine, but for some reason isn't recognized.",
                         RulesetErrorSeverity.OK
                     )
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -25,6 +25,7 @@ import com.unciv.models.stats.Stats
 import com.unciv.models.translations.fillPlaceholders
 import com.unciv.models.translations.tr
 import com.unciv.ui.utils.colorFromRGB
+import com.unciv.ui.utils.getRelativeTextDistance
 import kotlin.collections.set
 
 object ModOptionsConstants {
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -297,13 +298,39 @@ class Ruleset {
...
+     * Roughly corresponds to the fraction of the Unique placeholder text that can be different/misspelled, but with some extra room for [getRelativeTextDistance] idiosyncrasies. */
+    private val uniqueMisspellingThreshold = 0.15 // Tweak as needed. Simple misspellings seem to be around 0.025, so would mostly be caught by 0.05. IMO 0.1 would be good, but raising to 0.15 also seemed to catch what may be an outdated Unique.
 
     fun checkUniques(uniqueContainer:IHasUniques, lines:RulesetErrorList,
                      severityToReport: UniqueType.UniqueComplianceErrorSeverity) {
         val name = if (uniqueContainer is INamed) uniqueContainer.name else ""
 
         for (unique in uniqueContainer.uniqueObjects) {
-            if (unique.type == null) continue
+            if (unique.type == null) {
```
xlenstra, line 310: If this is your only check for whether a unique exists, that will probably give a lot of false positives until all uniques are typed. How much output does this give for the vanilla & G&K rulesets, including the 'unique looks like it should be fine, but for some reason isn't recognized'?
Additionally, we allow using uniques as a sort of 'identifier', similar to what we do with 'spaceship part'. Here we also catch these.

will-ca: See the screenshot. False positives are avoided by excluding anything that's *too* different. Basically, spelling similarity, not `unique.type == null`, is used as the main check for whether a unique exists, so the spelling has to be close enough to feasibly be a typo but not identical to count. If we start to see false positives, then we can just decrease `uniqueMisspellingThreshold` until they disappear. I think 0.1 would be good, but 0.05 would also work. Currently at 0.15, which is rather high, but based on the output in the screenshot it seems to still work fine.

No "Looks like it should be fine"s are currently generated for the stock rulesets. In in theory I don't think they should actually ever be generated, as the check for `null` I think should exclude all uniques that can produce them— I added it because I felt like I was going crazy trying to notice what was actually different about the spellings, so *not* seeing it affirms that there is indeed a spelling difference. A bug would have to be introduced that prevents `unique.type:UniqueType` from being set, I think, in order for it to actually ever show up.

xlenstra: Ah, I misread the code for when that line is placed, my bad. The ones in the screenshot (after close inspection :upside_down_face:) are all indeed valid misspellings so I might have overreacted a bit

xlenstra (approved):

xlenstra:
Would it be worthwhile to run the same check for the conditionals of the unique?

core/src/com/unciv/models/ruleset/Ruleset.kt
will-ca: Ah, no worries. My original notes for this PR had a heavy emphasis on false positives too, so I definitely think it's a valid concern— Hundreds of UniqueTypes with short strings made of just a couple stat and placeholder names, so you'd think some of them would have very similar names to untyped Uniques, right? But no, it just worked out better than either of us expected.

will-ca:
> Would it be worthwhile to run the same check for the conditionals of the unique?

Uh… Not sure. I hadn't actually looked much (at all) into how conditionals work until five minutes ago. Could it just be recursed through `unique.conditionals`?

But also, while doing #6028, I got another SeverityLevel-Red ERROR when I accidentally missed the square brackets in a conditional. So is there already another check that catches invalid conditionals (using exact spelling matches, which is probably better for that case)?

yairm210:
The maintenance unique looks exactly like what it thinks it's a typo of, what's the discrepancy?

PR #6028, opened 2022-01-23
Try to fix potential typos in stock rulesets.

will-ca:
Uses information from #6027. Removes all the messages shown on the screenshot over there.

I did notice the happiness thing in a recent game too.

Also makes the "No problems found." Label translatable.

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Buildings.json
+++ b/android/assets/jsons/Civ V - Vanilla/Buildings.json
@@ -690,7 +690,7 @@
 	},
 	{
 		"name": "Kremlin",
-		"cost": 500
+		"cost": 500,
 		"culture": 3,
 		"cityStrength": 12,
 		"isWonder": true,
@@ -905,7 +905,7 @@
 		"isWonder": true,
 		"culture": 3,
 		"greatPersonPoints": {"Great Engineer": 2},
-		"uniques": ["[-33]% Gold cost of upgrading [Military] units"],
+		"uniques": ["[-33]% Gold cost of upgrading <for [Military] units>"],
 		"requiredTech": "Radar",
 		"quote": "'In preparing for battle I have always found that plans are useless, but planning is indispensable.' - Dwight D. Eisenhower"
 	},
@@ -995,6 +995,6 @@
 		"isNationalWonder": true,
 		"uniques": ["Hidden until [5] social policy branches have been completed", "Triggers a global alert upon build start",
 			"Triggers a Cultural Victory upon completion", "Hidden when [Cultural] Victory is disabled"]
-	},
+	}
 	
 ]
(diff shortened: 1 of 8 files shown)


android/assets/jsons/Civ V - Vanilla/Buildings.json
```diff
@@ -691,7 +691,7 @@
 	{
 		"name": "Kremlin",
 		"cost": 500
-		"culture": 3,
+		,"culture": 3,
```
xlenstra, line 694: :upside_down_face:

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -188,9 +188,11 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     // Should be replaced with moddable improvements when roads become moddable
     RoadMovementSpeed("Improves movement speed on roads",UniqueTarget.Global),
     RoadsConnectAcrossRivers("Roads connect tiles across rivers", UniqueTarget.Global),
-    RoadMaintenance("[amount]% maintenance on road & railroads", UniqueTarget.Global), 
+    RoadMaintenance("[amount]% maintenance on roads & railroads", UniqueTarget.Global),
     @Deprecated("As of 3.18.17", ReplaceWith("[-amount]% maintenance on road & railroads"))
```
xlenstra, line 192: Could be useful to also update the replaceWith message in this unique

xlenstra (approved):

android/assets/jsons/Civ V - Vanilla/Buildings.json
will-ca: I wish I could take credit for this chaos, but I think it was probably some kind of auto-format.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -188,9 +188,11 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     // Should be replaced with moddable improvements when roads become moddable
     RoadMovementSpeed("Improves movement speed on roads",UniqueTarget.Global),
     RoadsConnectAcrossRivers("Roads connect tiles across rivers", UniqueTarget.Global),
-    RoadMaintenance("[amount]% maintenance on road & railroads", UniqueTarget.Global), 
-    @Deprecated("As of 3.18.17", ReplaceWith("[-amount]% maintenance on road & railroads"))
+    RoadMaintenance("[amount]% maintenance on roads & railroads", UniqueTarget.Global),
```
yairm210, line 191: If you're already making a new unique, why not "[amount]% maintenance on [improvementName]"? 
Not only does this enable specificity (e.g. Rome gets -50% maintenance on roads, Germany gets -25% maintenance on Railroads),
That way when we eventually redo the road status improvements to make them not hardcoded (and this will happen eventually) we'll be ready for it

will-ca: This PR also fixes a couple other Social Policies, and I'm still not entirely comfortable digging into the game mechanics, so I'll get rid of the new UniqueType for now to leave room to change it separately.

Issue #6032, opened 2022-01-24
copper


FriedrichWilhelm123:
https://civilization.fandom.com/wiki/Copper_(Civ5)
Mine available with Mining
+1 
![Screenshot_2022-01-24-10-23-44-42](https://user-images.githubusercontent.com/90401287/150714063-46ce41b3-659c-45b4-8cf1-fd5c7bbe84ab.jpg)

PR #6033, opened 2022-01-24
Update Copper Mine to add only 1 production (match Civ 5)

itanasi:
As noted in #6032, updating Copper Mine to only provide +1 Production instead of +2 to match Civ5


yairm210:
https://civilization.fandom.com/wiki/Copper_(Civ5) confirms this and that's good enough for me

yairm210 (approved):

PR #6042, opened 2022-01-24
Make Indirect Fire a starting promotion instead of Unique for Artillery, Battleship, etc.

will-ca:
No point exposing the promotion in the picker screen if all it does is give a Unique the unit already has.

Plus, this gives better linking in the Civilopedia, is consistent with how the Privateer works, and shows a nice icon in the selected unit promotions table (which IIRC Civ V also did).

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Units.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Units.json
@@ -1140,7 +1140,8 @@
 		"requiredTech": "Dynamite",
 		"upgradesTo": "Rocket Artillery",
 		"uniques": ["[+200]% Strength <vs cities> <when attacking>","No defensive terrain bonus",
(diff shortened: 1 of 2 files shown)


yairm210:
Do these units upgrade into a unit that *doesn't* have this promotion?

will-ca:
> Do these units upgrade into a unit that _doesn't_ have this promotion?

Artillery upgrades into Rocket Artillery, which also has the promotion. Battleship and Missile Cruiser don't upgrade.

yairm210:
In that case it doesn't make any gameplay difference, so why not

yairm210 (approved):

will-ca:
Realized this will need backwards compatibility code, which will also slightly improve robustness in future. WIll PR later today or early tomorrow.

PR #6046, opened 2022-01-24
Wiki: Remove outdated translation info?

will-ca:
As a result of AS linting, I found what appeared to be some very outdated information on translations from before strings were automatically scraped from JSONs.

(I probably found them and tried to figure them out in the original PR too, but just didn't know what they were… Actually, looks like I did leave those TODO comments.)


yairm210:
Fair enough
I needed to go through the wiki and change all looks to my "example mod" to instead direct to rekmod, which is just much better all around

xlenstra (approved):

PR #6049, opened 2022-01-25
Wiki: Mention nation colour layers.

will-ca:
Would you believe that wanting to make this edit was the entire reason I started the wiki migration into the repo?

Also fixes some inter-wiki links.


xlenstra (approved):

will-ca:
update wiki

PR #6052, opened 2022-01-25
Tiny/inconsequential debug crashes.

will-ca:
Tiny and inconsequential. But also one of the oldest items to cross off on my personal Unciv list, and safer to make impossible.


yairm210:
Au contraire, I WANT these edge cases to crash so we have crash reports, not to hide

PR #6055, opened 2022-01-26
Removed a function that is no longer used and should not be used imo

xlenstra:
I feel like all queries for uniques should either go through `getMatchingLocalUniques` instead of manually querying the uniques and then filtering through those


yairm210:
I think I removed the last surviving use of this today so you're right

PR #6062, opened 2022-01-27
Possibly fixed a crash accessing gameInfo before it is initialized

xlenstra:
From the crash logs of Google play:
```
kotlin.UninitializedPropertyAccessException: 
  at com.unciv.UncivGame.getGameInfo (UncivGame.kt:37)
  at com.unciv.logic.trade.TradeOffer.<init> (TradeOffer.kt:21)
  at com.unciv.logic.trade.TradeOffer.<init> (TradeOffer.kt:11)
  at com.unciv.logic.trade.TradeLogic.getAvailableOffers (TradeLogic.kt:35)
  at com.unciv.logic.trade.TradeLogic.<init> (TradeLogic.kt:13)
  at com.unciv.logic.automation.NextTurnAutomation.potentialLuxuryTrades (NextTurnAutomation.kt:518)
  at com.unciv.logic.automation.NextTurnAutomation.exchangeLuxuries (NextTurnAutomation.kt:562)
  at com.unciv.logic.automation.NextTurnAutomation.automateCivMoves (NextTurnAutomation.kt:46)
  at com.unciv.logic.GameInfo.nextTurn (GameInfo.kt:228)
  at com.unciv.ui.worldscreen.WorldScreen$nextTurn$1.invoke (WorldScreen.kt:642)
  at com.unciv.ui.worldscreen.WorldScreen$nextTurn$1.invoke (WorldScreen.kt:635)
  at com.unciv.ui.utils.ExtensionFunctionsKt$wrapCrashHandling$1.invoke (ExtensionFunctions.kt:306)
  at com.unciv.ui.utils.ExtensionFunctionsKt$wrapCrashHandlingUnit$1.invoke (ExtensionFunctions.kt:333)
  at com.unciv.ui.utils.ExtensionFunctionsKt$wrapCrashHandlingUnit$1.invoke (ExtensionFunctions.kt:333)
```
The line `tradeOffer.kt:21` is:
```kotlin
val gameSpeed = UncivGame.Current.gameInfo.gameParameters.gameSpeed
```

This seems to indicate that `UncivGame.Current` is not yet initalized when this object is created. A comment direclty above the line of the crash warns for exactly this:
```
// Do *not* access UncivGame.Current.gameInfo in the default constructor!
```
yet it is still done.

This PR moves this dependency from object creation time to property access time (not sure if those are official names), when `UncivGame.Current` should always be initialized.
When testing I was unable to see any differences in behaviour with before this PR.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -395,10 +395,10 @@ class DiplomacyManager() {
     private fun remakePeaceTreaty(durationLeft: Int) {
         val treaty = Trade()
         treaty.ourOffers.add(
-            TradeOffer(Constants.peaceTreaty, TradeType.Treaty, _duration = durationLeft)
+            TradeOffer(Constants.peaceTreaty, TradeType.Treaty, duration = durationLeft)
         )
         treaty.theirOffers.add(
-            TradeOffer(Constants.peaceTreaty, TradeType.Treaty, _duration = durationLeft)
+            TradeOffer(Constants.peaceTreaty, TradeType.Treaty, duration = durationLeft)
         )
         trades.add(treaty)
         otherCiv().getDiplomacyManager(civInfo).trades.add(treaty)
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -395,10 +395,10 @@ class DiplomacyManager() {
     private fun remakePeaceTreaty(durationLeft: Int) {
         val treaty = Trade()
         treaty.ourOffers.add(
-            TradeOffer(Constants.peaceTreaty, TradeType.Treaty, duration = durationLeft)
+            TradeOffer(Constants.peaceTreaty, TradeType.Treaty, _duration = durationLeft)
         )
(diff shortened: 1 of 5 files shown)


yairm210:
A data class can have a constructor IIRC. I think that besides the regular constructor which doesn't have any logic there should be another constructor that accepts a gameSpeed parameter - so the TradeOffer never directly accesses the UncivGame itself, it's always created from within an existing game, so the configuration can trickle downwards.

The very fact that this problem exists is because I didn't think of trickle down configurations when making the game because I was a noob 😓

xlenstra:
My understanding of the problem, is that the default constructor is used when loading a save, and somehow the `duration` variable is not set when doing this. If and when this happens is unclear to me (maybe if it's 0?), but if then this variable is queried before `UncivGame.Current` is initalized, it crashes. Adding another constructor doesn't seem to solve that problem, unless we use it to save the gameSpeed inside each trade offer, which seems like overkill. I may just be misunderstanding either your point or the problem, though :)

yairm210:
I'm saying that we can have a default duration of -1, and if save files contain that then that's how they'll remain.
The turn modifier will only be queried when constructing a TradeOffer from code, not from json.

xlenstra:
If I understand you correctly, this is basically what you had in mind?

yairm210 (approved):
Yes :)

SomeTroglodyte:
`TradeOffer` Seems not far at all to a clean and immutable class.

PR #6063, opened 2022-01-27
Converted era parameter to a transient var

yairm210:
The cost of casting getEra() constantly is definitely non-trivial, and why do we even generate the era every time? It only changes when we add a new tech! So we can save it as a transient in the civInfo.tech and be done with it!

Before:
image

After:
image

As a reminder, yesterday this same save took about 40s to setTransients, and now we're down to 8s! All the little microimprovements compound on each other :)

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -572,30 +572,7 @@ class CivilizationInfo {
         else -> getCivUnits().none()
     }
 
-    fun getEra(): Era {
-        if (gameInfo.ruleSet.technologies.isEmpty() || tech.researchedTechnologies.isEmpty())
-            return Era()
-        val maxEraOfResearchedTechs = tech.researchedTechnologies
-            .asSequence()
-            .map { it.column!! }
-            .maxByOrNull { it.columnNumber }!!
-            .era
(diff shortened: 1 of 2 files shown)


xlenstra (approved):

yairm210:
Because techs equal points, and researching recurring techs (a la Future Tech) should give you more points
We COULD change this all to a HashMap<String,ArrayList<String>> and I thought of that, but with this change the marginal benefit will be trivial to non existent

xlenstra:
One thing to note is that `researchedTechnolofies` is a transient, and recreated from `techsResearched`, a HashSet of Strings, so recurring techs are already not saved more than once.
For Score calculation I had to add another variable (`repeatingTechsResearched`) specifically for tracking how many techs were researched more than once

yairm210:
Aha
So maybe we could convert the whole thing into a hashset
Though as stated, profiling doesn't show that that will be any considerable improvement


In other news, my new enemy is tileInfo.hasUnique, and apparently the greatest time-waster there is that we rebuild the getAllTerrains every time we call it.
The correct solution is to have a transient allTerrains var in the tile, have terrainFeatures be private set, and have addTerrainFeature and removeTerrainFeature functions... but that's going to be a hassle :/

will-ca (approved):

PR #6067, opened 2022-01-28
Current map generation is extremely dependant on specific terrains existing in the ruleset.

yairm210:
This attempts to eliminate those dependencies.
All changes indicate areas where a crash occured before the change.

I still encounter problems in generateRegions when trying to generate a map with no water tiles - @SimorCedar I think the splitRegion doesn't like the fact that there are land tiles on edges of the map?

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
@@ -16,17 +16,21 @@ class RiverGenerator(
     }
 
     fun spawnRivers() {
+        if (tileMap.values.none { it.isWater }) return
         val numberOfRivers = tileMap.values.count { it.isLand } / MAP_TILES_PER_RIVER
 
         var optionalTiles = tileMap.values.asSequence()
-                .filter { it.baseTerrain == Constants.mountain && it.isFarEnoughFromWater() }.toMutableList()
+            .filter { it.baseTerrain == Constants.mountain && it.isFarEnoughFromWater() }
+            .toMutableList()
         if (optionalTiles.size < numberOfRivers)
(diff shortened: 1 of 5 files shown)


yairm210:
Ah no I see what's going on.
SplitRegions tries to split *by fertility*, and so when all tiles are the same it defaults to split by the edge.
So it splits into 2 rects - one is the original rect, and the other is 'maxX maxY' etc.
The solution I have is to take the split that's closest to the center, out of all possible optimal splits.

core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
```diff
@@ -16,17 +16,21 @@ class RiverGenerator(
     }
 
     fun spawnRivers() {
+        if (tileMap.values.none { it.isWater }) return
         val numberOfRivers = tileMap.values.count { it.isLand } / MAP_TILES_PER_RIVER
 
         var optionalTiles = tileMap.values.asSequence()
-                .filter { it.baseTerrain == Constants.mountain && it.isFarEnoughFromWater() }.toMutableList()
+            .filter { it.baseTerrain == Constants.mountain && it.isFarEnoughFromWater() }
```
will-ca, line 23: Why not use TileInfo.height?

will-ca (approved):

SomeTroglodyte:
Reminds me that the intent behind #5247 is still sorely missed...

core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
yairm210: Good question, but not part of this pr :)

PR #6072, opened 2022-01-29
Resources can now again provide uniques applying to the entire civ

xlenstra:
This fixes Marble not providing the 15% Production bonus.
One downside of this implementation, is that Marble must provide bonuses that are anti-local.
For example, even if it has "[+15]% Production [in this city]" as a unique, that will still apply to all cities. What logical way could you define 'in this city' if you received this resource via trading otherwise?

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -403,6 +403,11 @@ class CivilizationInfo {
         if (religionManager.religion != null)
             yieldAll(religionManager.religion!!.getFounderUniques().filter { it.isOfType(uniqueType) })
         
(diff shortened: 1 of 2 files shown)


will-ca:
> For example, even if it has "[+15]% Production [in this city]" as a unique, that will still apply to all cities. What logical way could you define 'in this city' if you received this resource via trading otherwise?

I still haven't dabbled in the conditionals system much, and am currently thinking of something else, so may be talking completely out of my keister. But conditional based on owning tile, applying only to friendly stats?

E.G.:
- Marble in your own city. → Boost in the city that owns a tile.
- Marble from trading. → You get nothing.
- Traded away last copy of marble. → You also get nothing.

Idk. I guess the proper action would be to see what Civ V does.

will-ca (approved):

PR #6076, opened 2022-01-30
Caught more mod failure conditions, removed certain assumptions from map creation

yairm210:
My 'example base ruleset', which I used to perform these tests, is now as minimal as I can make it
I plan to make it a template repo so people can start making base rulesets less painfully than has been the case so far

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -266,7 +266,10 @@ class CivilizationInfo {
     //region pure functions
     fun getDifficulty(): Difficulty {
         if (isPlayerCivilization()) return gameInfo.getDifficulty()
-        return gameInfo.ruleSet.difficulties["Chieftain"]!!
+        // TODO We should be able to mark a difficulty as 'default AI difficulty' somehow
+        val chieftainDifficulty = gameInfo.ruleSet.difficulties["Chieftain"]
+        if (chieftainDifficulty != null) return chieftainDifficulty
+        return gameInfo.ruleSet.difficulties.values.first()
     }
 
     fun getDiplomacyManager(civInfo: CivilizationInfo) = getDiplomacyManager(civInfo.civName)
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -2,6 +2,7 @@ package com.unciv.models.ruleset
 
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
+import com.unciv.Constants
 import com.unciv.JsonParser
 import com.unciv.logic.UncivShowableException
 import com.unciv.models.Counter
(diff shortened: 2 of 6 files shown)


SomeTroglodyte:
> template repo

_excellent_ idea

> able to mark a difficulty as 'default AI difficulty' somehow

`ModConstants.defaultDifficultyForAI` or `Difficulty.isDefaultForAI` - both would need checker code, first to disallow random text, second to disallow >1 `true`s..

yairm210:
https://github.com/yairm210/example-base-ruleset

This is the boi

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -266,7 +266,10 @@ class CivilizationInfo {
     //region pure functions
     fun getDifficulty(): Difficulty {
         if (isPlayerCivilization()) return gameInfo.getDifficulty()
-        return gameInfo.ruleSet.difficulties["Chieftain"]!!
+        // TODO We should be able to mark a difficulty as 'default AI difficulty' somehow
```
xlenstra, line 269: ModConstants maybe?

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -604,7 +611,7 @@ class Ruleset {
             for (building in era.settlerBuildings)
                 if (building !in buildings)
                     lines += "Nonexistent building $building built by settlers when starting in ${era.name}"
-            if (era.startingMilitaryUnit !in units)
+            if (era.startingMilitaryUnitCount != 0 && era.startingMilitaryUnit !in units)
```
xlenstra, line 614: I don't think this catches all uses, as they can also be used in `difficulties.json`

xlenstra, line 614: I don't think this catches all uses, as `Difficulties.json` can also use the Era Starting Unit

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
yairm210: Perhaps, that's for a different PR though :)

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -611,7 +611,7 @@ class Ruleset {
             for (building in era.settlerBuildings)
                 if (building !in buildings)
                     lines += "Nonexistent building $building built by settlers when starting in ${era.name}"
-            if (era.startingMilitaryUnitCount != 0 && era.startingMilitaryUnit !in units)
+            if (era.startingMilitaryUnitCount != 0 && era.startingMilitaryUnit != Constants.eraSpecificUnit && era.startingMilitaryUnit !in units)
```
xlenstra, line 614: I think this check is redundant? If we need to add `Constants.eraSpecificUnit`, it is actualy replaced with era.startingMilitaryUnit?

Hello200000: thanks for the warning

yairm210: Then I don't understand your previous comment.

xlenstra: What I meant is that only checking whether the era specific unit exists if that unit is used in the era is incorrect, as it is possible that it is not used in the era's, but is used in by AI's on higher difficulties, meaning that also in those cases the unit should still exist

xlenstra (approved):

PR #6077, opened 2022-01-30
Replace old uniques with *parameter-filled* new uniques!

yairm210:
Since we have the old unique typed, and the new replacement unique typed, we can show modders EXACTLY the new unique they'll need to have!

The final step of this process is having a 'replace old uniques' button which will automagically replace old uniques in a mod with the new versions. This will of course require some manual intervention since it's possible we added some parameters that didn't exist in the old one, but should work well for 95%+ of uniques.

Behold:

image


xlenstra:
Does this deal correctly with uniques with multiple parameters of the same type?
E.g., `[amount]% [stat] from every follower, up to [amount]%`

xlenstra (approved):

yairm210:
> Does this deal correctly with uniques with multiple parameters of the same type? E.g., `[amount]% [stat] from every follower, up to [amount]%`

Probably not?

PR #6084, opened 2022-01-31
Gradle build tools upgrade

yairm210:
Upgrading to 7.1 gives us the fun java.lang.NoSuchMethodError: 'void org.apache.commons.compress.archivers.zip.ZipFile.<init>(java.nio.channels.SeekableByteChannel) error so let's not

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -26,7 +26,7 @@ buildscript {
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${com.unciv.build.BuildConfig.kotlinVersion}")
         classpath("de.richsource.gradle.plugins:gwt-gradle-plugin:0.6")
-        classpath("com.android.tools.build:gradle:7.0.2")
+        classpath("com.android.tools.build:gradle:7.0.4")
         classpath("com.mobidevelop.robovm:robovm-gradle-plugin:2.3.1")
 
         // This is for wrapping the .jar file into a standalone executable


SomeTroglodyte:
SeekableByteChannel is the same error as when you do _not_ have the expected platform tools installed and it fails downloading them on-the-fly... I still suspect gradle to be directly responsible.

SomeTroglodyte:
The wrapper interfering? My impression was that the wrapper is the most resilient component in the gradle mess, as in: has the best up- and downwards compatibility.

Unsorted observations:
- Our wrapper jar is currently **6.9.2** as per its SHA256 in https://gradle.org/release-checksums/ - mebbeth make that easier to see via a comment???
- A wrapper upgrade via e.g. `./gradlew wrapper --gradle-version 7.2 --distribution-type bin` will yield a wrapper jar with a sha256 **not** listed on the releases page above and thus must be counted as malware. Doing the same for 7.3.3 will be clean, as in - result matches published digests.
- Trying a build with 7.3.3 both bin and wrapper will force a 32GB machine into out-of-memory in mere seconds, mouse freezes, only a killall java from a separate console will recover control. After a rollback the project fails to sync, which is then fixed by force-updating the IDE and plugins.

Conclusion: I admire your chuzpah to even try to keep up with that pain generator. I'll go sandblast my fingers and gargle with sulfuric acid to get rid of that unclean feeling, which is normally reserved for recovery from working with tinysquishy software.

yairm210:
'tinysquishy' is a wonderful phrase, I'll try to remember it :)

SomeTroglodyte:
'mouminuscule'.fr, 'chiquitablanda'.es, 'winzigweich'.de  :nerd_face:

PR #6099, opened 2022-02-02
Options checks mods gradually and has expander tabs

yairm210:
I have a lot of mods and need to check some of them sometimes. It's annoying that I need to wait for everything to load, and that I need to scroll endlessly to reach what I want.

image

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
@@ -4,7 +4,6 @@ import com.badlogic.gdx.Application
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.Input
 import com.badlogic.gdx.graphics.Color
-import com.badlogic.gdx.scenes.scene2d.Actor
 import com.badlogic.gdx.scenes.scene2d.ui.*
 import com.badlogic.gdx.utils.Align
 import com.unciv.Constants
@@ -27,7 +26,6 @@ import com.unciv.ui.utils.LanguageTable.Companion.addLanguageTables
 import com.unciv.ui.utils.UncivTooltip.Companion.addTooltip
 import com.unciv.ui.worldscreen.WorldScreen
 import java.util.*
-import kotlin.concurrent.thread
 import kotlin.math.floor
 import com.badlogic.gdx.utils.Array as GdxArray
 
@@ -279,8 +277,6 @@ class OptionsPopup(val previousScreen: BaseScreen) : Popup(previousScreen) {
             for (mod in RulesetCache.values.sortedBy { it.name }) {
                 var noProblem = true
                 val lines = ArrayList<FormattedLine>()
-                // Appending {} is a dirty trick to deactivate the automatic translation which would drop [] from unique messages
-                lines += FormattedLine("$mod", starred = true, header = 3)
 
                 val modLinks =
                     if (complex) RulesetCache.checkCombinedModLinks(linkedSetOf(mod.name))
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
```diff
@@ -294,27 +292,39 @@ class OptionsPopup(val previousScreen: BaseScreen) : Popup(previousScreen) {
...
+                    val loadingLabel = modCheckResultTable.children.last()
+                    modCheckResultTable.removeActor(loadingLabel)
+                    modCheckResultTable.add(expanderTab).row()
+                    modCheckResultTable.add(loadingLabel).row()
+                    modCheckCheckBox!!.enable()
+                }
             }
-            if (noProblem) lines += FormattedLine("No problems found.".tr())
 
+            // done with all mods!
```
xlenstra, line 325: I'm not entirely clear on how threads are run, but if each mod is run inside a separate thread, doesn't the main thread just continue? And if so, and it reaches the `postCrashHandlingRunnable` below, is it required that that `postRunnable` is run last, and also guaranteed to do so?

SomeTroglodyte: As far as I can understand from inspection only, the actual checks, while in a separate thread from the main UI thread, are sequential. I'll assume postCrashHandlingRunnable tries to mimic Gdx's postRunnable functionally, which would make the blocks it runs sequential again, but decoupled by a kind of fifo? No real concurrency here, except allowing the UI to stay responsive while the checks run. If there were, we'd have real problems - 317-320 would need some guard to ensure atomicity and thread-safety, and we'd need an upper limit to the number of worker threads. In short, go coroutines where you practically get thread pools for free if I remember right..

, but the rendering into an ExpanderTab may be parallel. That is assuming postCrashHandlingRunnable causes control to be passed to the Gdx UI thread using a new thread to do the _deferring_ - another assumption being that  in that respect. If that's the case, then line 317 may be reached by

will-ca: > postCrashHandlingRunnable tries to mimic Gdx's postRunnable functionally,

`postCrashHandlingRunnable` just calls `postRunnable` after wrapping its lambda in a last-ditch `try`/`catch`, IIRC.

yairm210: There are only 2 threads that are at work here - the ModChecker thread, which gathers data, and the main thread, which changes the UI.
`PostRunnable` takes a Runnable, or a piece of work to be done, and adds it to the main thread's todo list. The next time the main thread is before a render action, it will execute all items on its todo list. This means that posted runnables are both synchronous and in-order of being posted.

In our case, the ModChecker thread goes over each mod, gets the error data to display, and adds a "put this data in the UI" task on the main thread's todo list. Once it reaches the end, it adds a "remove last actor" to the main thread to get rid of the 'checking mods for errors' label

I originally implemented this with a thread per mod, but that was kinda overkill and was less readable 😅

core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
```diff
@@ -294,27 +292,39 @@ class OptionsPopup(val previousScreen: BaseScreen) : Popup(previousScreen) {
...
+                            label.wrap = true
+                            it.add(label).width(stage.width / 2).row()
+                        }
+                    }
+
+                    val loadingLabel = modCheckResultTable.children.last()
+                    modCheckResultTable.removeActor(loadingLabel)
+                    modCheckResultTable.add(expanderTab).row()
+                    modCheckResultTable.add(loadingLabel).row()
+                    modCheckCheckBox!!.enable()
```
xlenstra, line 321: Then one last question: should this enabling of the mod check checkbox be done for each mod, or only once at the end? Seeing as its only disabled once outside the entirety of this thread I'd have assumed the latter

yairm210: Oh yes that's a mistake :)

xlenstra (approved):

yairm210:
If I manage to get the next step working you're going to love it :)

core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
SomeTroglodyte: Very well worded description.

PR #6104, opened 2022-02-03
Fixed a bug where units requiring nearby units for bonuses could find themselves

xlenstra:
For example the discipline policy wanted 'an adjacent [Melee] unit' for the bonus to take effect. However, the 'adjacent' was determined from the tile where the battle would take place, to which the unit was adjacent. The condition would therefore be satisfied, by having found themselves.

Save file (send in by discord user):
[Greece -  6 turns.txt](https://github.com/yairm210/Unciv/files/7993947/Greece.-.6.turns.txt)


yairm210:
I think it should be dependent on the tile where the battle will take place, just we should filter out "and the unit we found is not this unit"

PR #6105, opened 2022-02-03
Find all replaceable uniques, generate replacement text, check for compatibility, and replace!

yairm210:
All of the work up until now has lead up to this point - since we already know what's out and what to replace it with, we can do 90%+ of the work ourselves and spare the sanity of our modders

For an example run:
- Download your favorite mod
- Check its deprecated uniques in the options menu
- Run the autoupdater from the same menu
- See how they disappear
- Bask in the glory


xlenstra:
Some padding between the button and the unique texts might be nice
image

xlenstra:
Few things:
- For modders this is a great update, maintaining mods in an ever-changing landscape of uniques was probably the most uninteresting necessary task
- Sometimes the button shows when no uniques can be updated, but there are errors:
image
- Not all objects with uniques are caught, resulting in a few uniques that for no reason seem to not update when pressing the button:
image

core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
```diff
@@ -329,6 +340,94 @@ class OptionsPopup(val previousScreen: BaseScreen) : Popup(previousScreen) {
...
+        for (fileName in filesToReplace) {
+            val file = jsonFolder.child(fileName)
+            if (!file.exists() || file.isDirectory) continue
+            var newFileText = file.readString()
+            for ((original, replacement) in deprecatedUniquesToReplacementText) {
+                newFileText = newFileText.replace("\"$original\"", "\"$replacement\"")
+            }
+            file.writeString(newFileText, false)
+        }
+        val toastText = "Replaced the following uniques:\n" +
```
xlenstra, line 423: Due to the options itself being a popup, this toast doesn't show up until after closing the options screen

xlenstra (approved):
Other than those points, nice!

PR #6112, opened 2022-02-06
Big update for French.properties

Ouaz:
- Try as much as possible to match Civ 5 french translation
- Modify some translations to better fit the UI
- Modify some translations to improve some original Civ 5 french entries which were poorly translated
- Make translations consistent through all the file
- Fix all broken translations due to bad word/key ordering
- Some quotes are now real quotes from french books (when available) rather than translated from english quotes
- Fix many typos

If needed, detailed diffs can be checked here: https://github.com/Ouaz/FRunciv/commits/master

### DIFF
--- a/android/assets/jsons/translations/French.properties
+++ b/android/assets/jsons/translations/French.properties
@@ -15,83 +15,83 @@ StartWithCapitalLetter = true
 
 # Tutorial tasks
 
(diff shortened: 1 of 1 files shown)


xlenstra:
Placeholder `[amount]` not found in `French` for entry `Nuclear weapon of Strength [amount]`

Ouaz:
Oops, my bad, really sorry. Fixed.

Issue #5985, opened 2022-01-16
As warning on battle table when attacking barbarians with a unit with 30 xp


yairm210:
This is something that comes up a lot on discord, which means it's not presented clearly to users.
A simple "units with more than (amount) xp will not gain xp from attacking barbarians" should suffice

xlenstra:
Maybe an easier solution is to always show how much XP attacking would earn in the battle table, and then put that line somewhere in the tutorials section?

yairm210:
Hmmm also a good point

alexban011:
shouldn't this be closed ?

PR #6115, opened 2022-02-07
Notify when Barbs don't give more XP

itanasi:
Addition to BattleTable entry to notify players when you can't gain more XP from Barbarians

image

### DIFF
--- a/core/src/com/unciv/logic/battle/ICombatant.kt
+++ b/core/src/com/unciv/logic/battle/ICombatant.kt
@@ -30,4 +30,6 @@ interface ICombatant {
         return this is CityCombatant
     }
     fun isCivilian() = this is MapUnitCombatant && this.unit.isCivilian()
+
+    fun getXP() = (this as MapUnitCombatant).unit.promotions.totalXpProduced()
 }
--- a/core/src/com/unciv/ui/worldscreen/bottombar/BattleTable.kt
+++ b/core/src/com/unciv/ui/worldscreen/bottombar/BattleTable.kt
@@ -163,6 +163,14 @@ class BattleTable(val worldScreen: WorldScreen): Table() {
             row().pad(2f)
         }
 
+        // from Battle.addXp(), check for can't gain more XP from Barbarians
+        val modConstants = attacker.getCivInfo().gameInfo.ruleSet.modOptions.constants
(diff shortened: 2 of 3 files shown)


itanasi:
Solves #5985

core/src/com/unciv/logic/battle/ICombatant.kt
```diff
@@ -30,4 +30,6 @@ interface ICombatant {
         return this is CityCombatant
     }
     fun isCivilian() = this is MapUnitCombatant && this.unit.isCivilian()
+
+    fun getXP() = (this as MapUnitCombatant).unit.promotions.totalXpProduced()
```
yairm210, line 34: Surely this should be a function of MapUnitCombatant and not ICombatant

xlenstra (approved):

core/src/com/unciv/logic/battle/ICombatant.kt
itanasi: That's fair since ICombatant doesn't always have XP

itanasi:
Manual edit from phone/web, so will check later or someone else can pull the branch and double check

core/src/com/unciv/ui/worldscreen/bottombar/BattleTable.kt
```diff
@@ -163,6 +163,14 @@ class BattleTable(val worldScreen: WorldScreen): Table() {
             row().pad(2f)
         }
 
+        // from Battle.addXp(), check for can't gain more XP from Barbarians
+        val modConstants = attacker.getCivInfo().gameInfo.ruleSet.modOptions.constants
+        if (attacker is MapUnitCombatant && (attacker as MapUnitCombatant).unit.promotions.totalXpProduced() >= modConstants.maxXPfromBarbarians
+                && defender.getCivInfo().isBarbarian()){
+            add("Cannot gain more XP from Barbarians".tr().toLabel(fontSize = 16).apply { wrap=true }).width(quarterScreen)
```
yairm210, line 170: Add to translation file

itanasi: Derp. Forgot it auto pulls on JSONs

itanasi:
Fixed

xlenstra (approved):

PR #6116, opened 2022-02-07
Can't Melee Attack if Can't Move to Tile

### LINKED ISSUE #6111: Barb camp - capture missionary in camp bug
**Platform**
Windows
**Version**
3.19.5 - playing Gods&Kings

**Describe the bug**
Barb camp with a swordsman & a missionary - attack to kill swordsman, attack (with boat) to capture missionary. Barb camp at that point still exists with my missionary in it. Next turn: barb unit spawned, so is on the same tile as my missionary... Looks really weird. Either a) the capture of the unit captures the camp or b) the captured unit is kicked out to another tile?
image

itanasi:
Fixes #6111
Battle code works, but need to update UI to now show attacks that aren't valid

### DIFF
--- a/core/src/com/unciv/logic/automation/BattleHelper.kt
+++ b/core/src/com/unciv/logic/automation/BattleHelper.kt
@@ -98,6 +98,10 @@ object BattleHelper {
             if (combatant.isRanged()) return false
         }
 
+        if (combatant is MapUnitCombatant && combatant.isMelee()
+                && (!combatant.unit.movement.canPassThroughIfEmpty(tile) && !tile.isCityCenter()))
+            return false
+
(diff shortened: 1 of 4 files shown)


itanasi:
Small note. BattleTable still shows up even if you can't attack them, which may be confusing. But no Attack Crosshair and Attack Button is disabled

image

yairm210:
Hm this is very different from what I was thinking of.
I think I'll poll this.

yairm210:
Almost 50 results in, after a lot of talk, the poll is much more definite than when it started out - units *can* attack neighboring tiles they can't enter, and the fix you made ensures that they won't *capture* there, so I think we're done

PR #6120, opened 2022-02-07
Added conditional for 'first civilization to research a tech'

yairm210:
Techs with these kinds of conditional bonuses exist in Civ IV and VI
Phrasing also allows long-term benefits for techs which no one else has researched, like "[+10%] Strength <if no other Civilization has researched this>"
A nicer sounding but more limiting phrasing is "<for the first civilization to research this>", but since we don't currently hold information on who first researched each tech, that could confuse players that this will continue to work even after others have researched it because we were the first.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -42,7 +42,7 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
         if (state == null) return conditionals.isEmpty()
         if (state.ignoreConditionals) return true
         for (condition in conditionals) {
-            if (!conditionalApplies(condition, state)) return false
+            if (!conditionalApplies(condition, state, this)) return false
         }
         return true
     }
@@ -71,7 +71,8 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
 
     private fun conditionalApplies(
         condition: Unique,
-        state: StateForConditionals
+        state: StateForConditionals,
+        baseUnique: Unique
     ): Boolean {
 
         fun ruleset() = state.civInfo!!.gameInfo.ruleSet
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -71,7 +71,8 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
 
     private fun conditionalApplies(
         condition: Unique,
-        state: StateForConditionals
+        state: StateForConditionals,
+        baseUnique: Unique
```
xlenstra, line 75: This parameter feels redundant; it is a private function so only accessible from within this class, and is only ever accessed from `conditionalsApply`, where the value `this` is passed, which can also just be called in this function

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -584,6 +584,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     ConditionalOnWaterMaps("on water maps", UniqueTarget.Conditional),
     ConditionalInRegionOfType("in [regionType] Regions", UniqueTarget.Conditional),
     ConditionalInRegionExceptOfType("in all except [regionType] Regions", UniqueTarget.Conditional),
+    ConditionalFirstCivToResearch("if no other Civilization has researched this", UniqueTarget.Conditional),
```
xlenstra, line 587: Very minor point, but I think this unique should be put in the `/////// civ conditionals` section instead of the `/////// area conditionals` section of the `uniqueType` enum

core/src/com/unciv/models/ruleset/unique/Unique.kt
yairm210: Good point :)

xlenstra (approved):

PR #6121, opened 2022-02-07
Update Portuguese.properties

chclxds:


yairm210:
Placeholder `[amount]` not found in `Portuguese` for entry `Current leader is [civInfo] with [amount] [stat] generated.`
    Placeholder `[amount]` not found in `Portuguese` for entry `Current leader is [civInfo] with [amount] Technologies discovered.`

PR #6125, opened 2022-02-07
Translation: Update español

ArchDuque-Pancake:


yairm210:
Conflicts with another Spanish translation in a way that I can't autoresolve, you each made your own similar translations to several lines :)

PR #6127, opened 2022-02-08
Add Amphibious penalty to Land attacking into Water and vice versa

itanasi:
Right now there's a penalty for moving from Water (Embarked) to Land. But no penalty for Land Melee attacking into Water nor Naval units attacking on to Land.
Apply (and rename) penalty to all attacks that cross the land/water boundary


android/assets/jsons/translations/Brazilian_Portuguese.properties
```diff
@@ -1069,7 +1069,7 @@ defence vs ranged = defesa vs ataque à distancia
 [percentage] to unit defence = [percentage] a defesa da unidade
 Attacker Bonus = Bônus de Atacante
 Defender Bonus = Bônus de defesa
-Landing = Aterrissagem
+Amphibious = Aterrissagem
```
xlenstra, line 1072: I usually dont like doing this, as it gives no indication that translators need to update this translation.
Especially if it's only a single word and because it completely changes the meaning of it, this can be problematic and retranslating it from scratch doesn't take that long.
I'd have removed the word from the template file and then this new change would be distributed next version update.

Additionally, if I'm not mistaken 'amphibious' is the name of a promotion and should as such already be translated. Renaming these translations could then result in these strings having two different translations, of which the wrong one might be discarded when regenerating the translation files.

yairm210:
I'm against this change entirely - "Amphibious" is what lets you *not* have this penalty, it's definitely not a good name for the penalty itself!
IMO attacking land-to-water should have a "Boarding" penalty, similar to the "Landing" penalty for water-to-land

itanasi:
Makes sense. This was more a what-if, so we can workshop a bit more

yairm210:
If you're going to change an existing unique, create a unique type for the old and new versions and mark the old version as deprecated with'replace with' of the new unique
That was people can auto update their mods

android/assets/jsons/Civ V - Gods & Kings/UnitPromotions.json
```diff
@@ -227,6 +227,11 @@
 		"uniques": ["[+20]% Strength <vs cities>", "Earn [33]% of the damage done to [City] units as [Gold]"],
 		"unitTypes": ["Melee Water"]
 	},
+	{
+		"name": "Amphibious",
+		"uniques": ["Eliminates combat penalty for attacking across a coast"],
+		"unitTypes": ["Melee Water"]
```
yairm210, line 233: Wiki says this is for melee units?

xlenstra: This PR adds an additional Amphibious promotion, different from the already existing one for melee units

yairm210: In that case it's even worse, two promotions with the same name won't work, the latter will override the former
We save the promotions in a String - Promotions hashmap

itanasi: Oooh, ok. Will fix

itanasi:
Icon isn't working. Thought just needed to add using same name to appropriate folder?

xlenstra:
> Icon isn't working. Thought just needed to add using same name to appropriate folder?

Most of the time it should, but sometimes the game bugs out and doesn't regenerate the game.png and game.atlas file. Try removing those manually (should be saved in `./android`) and restarting your game

yairm210 (approved):

xlenstra (approved):

PR #6128, opened 2022-02-08
Sea Unit can't capture Land Civilian (and vice versa)

### LINKED ISSUE #6111: Barb camp - capture missionary in camp bug

itanasi:
Fix #6111 without changing all the combat rules.
Add in a check to prevent capturing Civilians you can't reach
Note if they were defended, still won't capture since won't move on to tile

### DIFF
--- a/core/src/com/unciv/logic/automation/BattleHelper.kt
+++ b/core/src/com/unciv/logic/automation/BattleHelper.kt
@@ -48,7 +48,7 @@ object BattleHelper {
                         // capture enemy units since we can't move through them!
                         && !it.canCivPassThrough(unit.civInfo)
                         // Land Unit can't capture Naval and vice versa
-                        && (unit.type.isLandUnit() && it.isWater || unit.type.isWaterUnit() && it.isLand)
+                        && !(unit.type.isLandUnit() && it.isWater || unit.type.isWaterUnit() && it.isLand)
             }
 
         val rangeOfAttack = unit.getRange()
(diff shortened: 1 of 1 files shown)


yairm210:
This is a good compromise :)
The poll got a lot more attention than I expected

yairm210 (approved):

core/src/com/unciv/logic/automation/BattleHelper.kt
```diff
@@ -47,6 +47,8 @@ object BattleHelper {
                         // DO NOT use "!unit.movement.canPassThrough(it)" since then we won't be able to
                         // capture enemy units since we can't move through them!
                         && !it.canCivPassThrough(unit.civInfo)
+                        // Land Unit can't capture Naval and vice versa
+                        && !(unit.type.isLandUnit() && it.isWater || unit.type.isWaterUnit() && it.isLand)
```
xlenstra, line 51: I may be mistaken, but isn't this line in the function for getting the attackable enemies? And so doens't this also stop melee attacks into/from the water?

yairm210: No, since this is in the "mapCombatant.isCivilian" filterNot check

xlenstra: Ah, my bad

itanasi:
Ooh, just realized maybe could add an exception if the unit has Amphibious?

yairm210:
No I think this is fine

PR #6129, opened 2022-02-08
Typified Obsolete unique, made luxury resources obsoletable

yairm210:
Things that were luxuries in the past become uninteresting as time goes on


core/src/com/unciv/models/ruleset/tech/Technology.kt
```diff
@@ -106,7 +109,10 @@ class Technology: RulesetObject() {
      */
     // Used for Civilopedia, Alert and Picker, so if any of these decide to ignore the "Will not be displayed in Civilopedia" unique this needs refactoring
     fun getObsoletedBuildings(civInfo: CivilizationInfo) = getFilteredBuildings(civInfo)
-        { it.uniqueObjects.any { unique -> unique.placeholderText == "Obsolete with []" && unique.params[0] == name } }
+        { it.getMatchingUniques(UniqueType.ObsoleteWith).any { it.params[0]  == name } }
+
+    fun getObsoletedResurces(ruleset: Ruleset) = ruleset.tileResources.values.asSequence()
```
xlenstra, line 114: *Resources

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -330,6 +330,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     MustNotBeNextTo("Must not be next to [terrainFilter]", UniqueTarget.Building),
 
     Unsellable("Unsellable", UniqueTarget.Building),
+    ObsoleteWith("Obsolete with [tech]", UniqueTarget.Building, UniqueTarget.Resource),
```
xlenstra, line 333: It looks like this can also apply to improvements (`tileInfo.kt:457`)? iirc maybe also units?

yairm210: Units already have an 'obsolete' field, and it's ubiquitous enough that I don't feel the need to convert it
Improvements...? Interesting

yairm210: OH I see what you mean
It was ALREADY for improvements 
I'll add that in then, and rewrite some of the common aspects

xlenstra (approved):

android/assets/jsons/Civ V - Gods & Kings/Units.json
```diff
@@ -1308,10 +1308,11 @@
 		"unitType": "Gunpowder",
 		"movement": 2,
 		"strength": 65,
-		"cost": 400,
+		"cost": 160,
 		"requiredTech": "Pharmaceuticals",
 		"promotions": ["Amphibious"],
-		"uniques": ["Normal vision when embarked", "Defense bonus when embarked"],
+		"uniques": ["Normal vision when embarked",
+			"[+50]% Strength <vs [Artillery] units>", "[+50]% Strength <vs [Machine Gun] units>"],
```
xlenstra, line 1315: I assume you're still working on this, but if not, this seems like debug code?

yairm210:
That last commit was a mistake while working on something else :(
I'll fix this when I get back to it

PR #6133, opened 2022-02-09
Unified "X is only available under Y conditions" into a single unique

yairm210:
There were a few problems with existing uniques - they weren't really composable, the offered things they didn't keep, etc

For example, "Incompatible with [policy/tech/promotion]", UniqueTarget.Policy, UniqueTarget.Tech, UniqueTarget.Promotion. In fact, promotions only checked promotion incompatibility, promotions - promotion incompat, etc

Additionally, with a few more changes, this could cover several other uniques - "Hidden until [amount] social policy branches have been completed", "Requires at least [amount] population", perhaps others

I have to say I think conditionals are the best thing ever and they make amazing composability possible :)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -279,10 +279,11 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     PopulationLossFromNukesDeprecated("Population loss from nuclear attacks -[amount]%", UniqueTarget.Global),
     
     NaturalReligionSpreadStrength("[amount]% Natural religion spread [cityFilter]", UniqueTarget.FollowerBelief, UniqueTarget.Global),
-    @Deprecated("as of 3.19.3", ReplaceWith("[amount]% Natural religion spread [cityFilter] <after discovering [tech]> OR [amount]% natural religion spread [cityFilter] <after adopting [policy]>"))
+    @Deprecated("as of 3.19.3", ReplaceWith("[amount]% Natural religion spread [cityFilter] <after discovering [tech/policy]> OR [amount]% natural religion spread [cityFilter] <after adopting [tech/policy]>"))
```
xlenstra, line 282: Why `tech/policy` in both cases? Due to the 'OR' automatic replacement probably won't work anyway

tests/src/com/unciv/testing/BasicTests.kt
```diff
@@ -134,8 +134,8 @@ class BasicTests {
                 println("${uniqueType.name}'s deprecation text does not match any existing type!'")
                 allOK = false
             }
-            if (replacementTextUnique.getDeprecationAnnotation() != null){
-                println("${uniqueType.name}'s deprecation text references another deprecated unique!'")
```
xlenstra, line 138: One thing to note is that it could then be the case (if I read the replacing code correctly) that a deprecated unique is upgraded to another deprecated unique, requiring the auto unique updater to be run again, which may not be intuitive for the user

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: Because then you can click "autoreplace" and manually remove the irrelevant part, instead of manually typing in everything from scratch

tests/src/com/unciv/testing/BasicTests.kt
yairm210: I agree but the alternative is that each Unique update requires updating the deprecation text of all the tail, which is considerable more work than pushing a button several times

yairm210: It is indeed possible and that's a good point :)

xlenstra: I hadn't approved this yet waiting for this to be added, but I suppose that can be a different PR

PR #6134, opened 2022-02-09
Arbitrary conditionals can become timed conditionals!

yairm210:
triggerCivwideUniques is activated in 3 situations:
- Tech complete
- Policy adopted
- Building complete

So for each of the unique containers for these I added a catch saying 'if this unique is a temporary unique then don't save it yourself

This has been lying around in todos for a while and generalizes one of the nasty hardcoded messes that I hate about Civ V. Why would you introduce ONE timed effect in the whole game?! You can use them all the time like Humankind or never use them like Civ IV, but why have just one whyyy

One more thing in this PR, that needed to be solved by the by, is allowing parameters to have a '2' on the end so we can autoreplace multiple parameters of the same type.
Using a regex is slightly more inefficient, but since this is A. only used once per UniqueType.kt, and B. allows us to have as many 'amount's as we want which is important if we have a lot of conditionals, I feel this is justified.

### DIFF
--- a/core/src/com/unciv/logic/civilization/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/PolicyManager.kt
@@ -54,8 +54,9 @@ class PolicyManager {
 
     fun addPolicyToTransients(policy: Policy) {
         for (unique in policy.uniqueObjects) {
-            // Should be replaced with a conditional of the same form later
-            if (!unique.text.contains(turnCountRegex))
+            // Should be deprecated together with TimedAttackStrength so I'm putting this here so the compiler will complain if we don't
+            val rememberToDeprecate = UniqueType.TimedAttackStrength
+            if (!unique.text.contains(turnCountRegex) && unique.conditionals.none { it.type == UniqueType.ConditionalTimedUnique })
                 policyUniques.addUnique(unique)
         }
     }
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -225,7 +225,7 @@ class UniqueMapTyped: EnumMap<UniqueType, ArrayList<Unique>>(UniqueType::class.j
 
 
 // Will probably be allowed to be used as a conditional when I get the motivation to work on that -xlenstra
-class TemporaryUnique() {
+class CivwideUnique() {
 
     constructor(uniqueObject: Unique, turns: Int) : this() {
         unique = uniqueObject.text
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -85,6 +85,7 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
         }
 
         return when (condition.type) {
+            UniqueType.ConditionalTimedUnique -> true
             UniqueType.ConditionalWar -> state.civInfo?.isAtWar() == true
             UniqueType.ConditionalNotWar -> state.civInfo?.isAtWar() == false
             UniqueType.ConditionalWithResource -> state.civInfo?.hasResource(condition.params[0]) == true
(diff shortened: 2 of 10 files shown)


core/src/com/unciv/logic/civilization/PolicyManager.kt
```diff
@@ -54,8 +54,9 @@ class PolicyManager {
 
     fun addPolicyToTransients(policy: Policy) {
         for (unique in policy.uniqueObjects) {
-            // Should be replaced with a conditional of the same form later
-            if (!unique.text.contains(turnCountRegex))
+            // Should be deprecated together with TimedAttackStrength so I'm putting this here so the compiler will complain if we don't
+            val rememberToDeprecate = UniqueType.TimedAttackStrength
+            if (!unique.text.contains(turnCountRegex) && unique.conditionals.none { it.type == UniqueType.ConditionalTimedUnique })
```
xlenstra, line 59: `it.isOfType()`?

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -225,7 +225,7 @@ class UniqueMapTyped: EnumMap<UniqueType, ArrayList<Unique>>(UniqueType::class.j
 
 
 // Will probably be allowed to be used as a conditional when I get the motivation to work on that -xlenstra
```
xlenstra, line 227: This comment should also be removed now that it's implemented :upside_down_face:

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -225,7 +225,7 @@ class UniqueMapTyped: EnumMap<UniqueType, ArrayList<Unique>>(UniqueType::class.j
...
-class TemporaryUnique() {
+class CivwideUnique() {
```
xlenstra, line 228: What is the reason this is renamed?

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -548,6 +548,8 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     ConditionalPolicy("after adopting [policy]", UniqueTarget.Conditional),
     ConditionalNoPolicy("before adopting [policy]", UniqueTarget.Conditional),
 
+    ConditionalTimedUnique("for [amount] turns", UniqueTarget.Conditional),
```
xlenstra, line 551: I think this unique should always return 'true' in the 'when' statement of `conditionalApplies`, as otherwise it will default to false and the unique will be ignored?

core/src/com/unciv/models/ruleset/unique/Unique.kt
yairm210: A part of my nefarious plan to use this for permanent conditionals as well.
I'm still planning that but this change should go in a different pr :)

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: Damn you're right

core/src/com/unciv/logic/civilization/PolicyManager.kt
yairm210: I regret creating that function, it adds nothing and equality is really understood

yairm210: I'll probably remove it completely in the future

xlenstra (approved):

docs/uniques.md
```diff
@@ -1351,6 +1346,11 @@ Example: "<before adopting [Oligarchy]>"
 
 Applicable to: Conditional
 
+#### <for [amount turns>
+Example: "<for [amount turns>"
```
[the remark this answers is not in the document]
yairm210: Yup, that was fixed in the code, but I didn't regenerate the md file

PR #6146, opened 2022-02-11
Add uniques and constants for unit supply

xlenstra:
Also a bit of code readability, though I haven't checked performance impact of that

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -288,9 +288,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     StartsWithTech("Starts with [tech]", UniqueTarget.Nation),
     ResearchableMultipleTimes("Can be continually researched", UniqueTarget.Global),
 
-    BaseUnitSupply("[+amount] Unit Supply", UniqueTarget.Global),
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -288,6 +288,10 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     StartsWithTech("Starts with [tech]", UniqueTarget.Nation),
     ResearchableMultipleTimes("Can be continually researched", UniqueTarget.Global),
 
+    BaseUnitSupply("[amount] Unit Supply", UniqueTarget.Global),
+    UnitSupplyPerPop("[amount] Unit Supply per [amount] population", UniqueTarget.Global),
```
yairm210, line 292: This feels like it should have a cityfilter

xlenstra: I was planning so at first too, but removed it after deciding that would complicate the calculation for supply a lot. Basically for each unique I would have to iterate over all cities and add their population only if the city matched the filter (O(n^2)). If you want I can still add it in

yairm210: I think we should add it, it's O(n) per unique which is exactly the complexity of a unique that sums population count anyway :)

xlenstra: Should be added now

PR #6149, opened 2022-02-12
Show Improvements that are buildable after Removing TerrainFeature

itanasi:
Checks and shows if you could build something after Clearing a Forest/Jungle/Marsh. Heavily abuses getLastTerrain(), so won't work if stacked unbuildable features.

Partial fix for the whole Chop + Build Improvement option

image


itanasi:
I tried using a .clone() version of TileInfo. Would be safer and faster. But for some reason didn't work. My guess .clone() doesn't copy all internal variables. I added a few obviously missing ones and still didn't get it to work.

core/src/com/unciv/ui/pickerscreens/ImprovementPickerScreen.kt
```diff
@@ -89,9 +105,12 @@ class ImprovementPickerScreen(val tileInfo: TileInfo, unit: MapUnit, val onAccep
                     && improvement.name != Constants.cancelImprovementOrder)
             if (tileInfo.improvement != null && removeImprovement) labelText += "\n" + "Replaces [${tileInfo.improvement}]".tr()
 
-            val pickNow = if (tileInfo.improvementInProgress != improvement.name)
+            val pickNow = if (suggestRemoval)
+                (Constants.remove + tileInfo.getLastTerrain().name + " first").toLabel()
```
yairm210, line 109: This isn't translatable

yairm210:
The problem isn't .tr(), since .toLabel() find that. The problem us that the text you generate don't be translated with .tr() since it doesn't have the square brackets.
Try changing the language (and adding the relevant translation line) and you'll see that it isn't translated

yairm210:
You need to add square brackets around the feature name, in the label text on line 109

itanasi:
Translation working. Optimized performance

PR #6153, opened 2022-02-12
Update and rename Spanish.properties 

Hello200000:
Fixed some bugs in translations


Hello200000:
THEN I WILL NAME IT

yairm210:
We'll try to merge this, if it doesn't conflict things should be fine

PR #6157, opened 2022-02-13
Trigger uniques by sacrificing units with conditional

yairm210:
Triggered uniques currently come from researching techs, adopting policies, and building buildings.

This adds a third way of triggering uniques, by attaching them to units with a "by consuming this unit" conditional, which will be added as a possible unit action.
So for example, "[amount] Free Technologies \<by consuming this unit>", "Reveals the entire map \<by consuming this unit>" etc.

I debated whether to add a tilefilter or a cityfilter to the triggering and for now decided against. I also don't want the triggering to be dependent on external conditionals like "Reveals the entire map \<by consuming this unit> \<in the capital>" because then you won't have a way to differentiate between conditionals *for the trigger* and conditionals *for the action* and that's not good.
I think if this goes over and we need to, we can add "by consuming this unit in [tilefilter] tiles" and "by consuming this unit [cityFilter]" if necessary.

image

In addition to this relatively simple PR, I'd also like your thoughts on how to generalize the space race victory.

Currently the science victory is *extremely* hardcoded.
I'm thinking of changing this gradually in a number of ways, which will open up Civ IV and Civ VI science victory conditions:

- "Triggers victory" and "Triggers a Cultural Victory upon completion" unified to "Triggers a [victoryType] Victory"
- A resource conditional similar to "with [resource]" like "with [amount] [resource]"
- Spaceship parts have "Provides [1] [nameOfSpaceshipPart]", all spaceship parts are bonus resources (or perhaps a fourth type of resource - "Special" resources?), and the general list of uniques contains "Triggers a [Scientific] Victory \<with [1] [SS Engine]> \<with [3] [SS Booster]>" etc
- And finally, once that's done, we can convert the SS parts to a unit that triggers "Provides [1] [nameOfSpaceshipPart] \<by consuming this unit [in the capital]>"


xlenstra:
> In addition to this relatively simple PR, I'd also like your thoughts on how to generalize the space race victory.

I was also working on a branch making those more modifiable, but due to spending a lot of time on my bachelor thesis, I basically stashed away all large PR's like that for now.
Basic idea there, was to make both the VictoryScreen and requirements for them moddable.
It adds another json file for the victory types (called `VictoryTypes.json`), which then replaces the VictoryType enum. It specifies the 'milestones' each player must reach for each victory type, such as 'build [Apollo Project]', 'add [3] [SS booster] units in the capital', etc. and automatically generates the VictoryScreen to contain those milestones. The global overview then shows which milestone each player needs to complete next and the amount of milestones already completed (e.g. 3/7 if you've build the Apollo Project and added 2 spaceship parts to the spaceship). It would also contain the strings that are displayed at each column in the Victory Screen, which are still hardcoded as of now (See also: #4293).
This json would then also contain the units that would be required for a space victory and how many of each.

I'll push the branch later today so you can see what progress I made with that. Iirc, I was at the point where I needed to ensure only a maximum of `3` SS boosters may exist at the same time, and still needed to do a lot of automatically generating the victory screen.

> "Triggers victory" and "Triggers a Cultural Victory upon completion" unified to "Triggers a [victoryType] Victory"

Agree
> A resource conditional similar to "with [resource]" like "with [amount] [resource]"

Sounds useful in general, regardless of usefulness for victories

> Spaceship parts have "Provides [1] [nameOfSpaceshipPart]", all spaceship parts are bonus resources (or perhaps a fourth type of resource - "Special" resources?), and the general list of uniques contains "Triggers a [Scientific] Victory <with [1] [SS Engine]> <with [3] [SS Booster]>" etc

To me, this kinda feels like pushing a square block through a circular hole. I suppose it would work, bu I don't feel like the resource system would be the most logical way to implement this.

> And finally, once that's done, we can convert the SS parts to a unit that triggers "Provides [1] [nameOfSpaceshipPart] <by consuming this unit [in the capital]>"

If we do go this route, this would be the logical way to do it.


I'll look at the rest of the PR later, these were just my thoughts on the victory system

xlenstra (approved):

yairm210:
Test failed because of missing translation. which was the reason I had to do the truffle shuffle to begin with... you can't win
Actionable fix: translation test should check if all unit actions are in translation.properties? Or maybe writer should add them instead

PR #6165, opened 2022-02-15
Some edits for fun getRejectionReasons in class building/unit and fun checkMod

lishaoxia1985:
1. Some edits for fun getRejectionReasons in class building/unit, now check the entire unique in one loop
2. delete InvalidRequiredBuilding(false, "Required building does not exist in ruleSet!") because the relevant unique will be checked in fun checkmod


yairm210:
If you want to deprecate "Requires []", which I agree is a good idea, the correct way would be to
A. Add "Requires [eraName/policyName/buildingName]" as a UniqueType
B. Add a new conditional of "if [buildingName] has been built"
C. Add a deprecation warning for "Requires []" with replacement text "Only available <if [buildingName] has been built> OR Only available <starting from the [eraName]> OR Only available <after adopting [policyName]>"

This should probably be its own PR, because your current PR is packed and I can't tell what's supposed to connect with what :/

yairm210:
Creating a uniqueType for the different filtering uniques will ensure that they're ALL typechecked upon initialization, so we won't need to worry about adding a "requires a [whatever] that does not exist in the ruleset" check to everything :)

lishaoxia1985:
@yairm210 I haven't deprecated "Require []" in this commit, those comments is the suggestion to improve in the future.
This commit gets only 2 target above.
About UnitType, I agree.
But maybe we have a better method if we use Table-Driven Methods:
1. Create a hashmap<String, ()->Unit)>, The String key is unique, ()->Unit value is unique effect, it a fun or expression.
2. We add all unique and its effect into this hashmap.
It have 2 advantages:
1. Can check whether the unique has effect, if not, we can give a warning.
2. If the hashmap key doesn't contain the unique, we can use this to check ruleset.
Maybe not all the unique can use this - Table-Driven Methods, I'm not sure.
Sorry, I'm tired of thinking this, my computer is broken, the another older one has Low performance, so editing code spends more time.

yairm210:
The same deprecation process applies to "Unlocked with []", we don't just delete uniques, we deprecate them with the steps outlined for "Requires []"
This is so that mods can automatically update from the old deprecated uniques to the new ones with a single button press

lishaoxia1985:
@yairm210 OK, I have edited this.

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -575,9 +575,18 @@ class Ruleset {
...
                     lines += "${building.name} creates a ${unique.params[0]} improvement which does not exist!"
+                if (unique.placeholderText == UniqueType.RequiresAnotherBuilding.placeholderText
+                        || unique.placeholderText == UniqueType.RequiresBuildingInAllCities.placeholderText
+                        || unique.placeholderText == UniqueType.RequiresBuildingInSomeCities.placeholderText
+                ) {
+                    val buildingName = unique.params[0]
+                    if (!buildings.containsKey(buildingName))
+                        lines += "${building.name} requires $buildingName which does not exist!"
+                }
+            }
```
yairm210, line 589: This is unnecessary, due to the checkUniques function below, which will do this anyway

lishaoxia1985: OK, I delete it.

lishaoxia1985:
@yairm210 👀😀

PR #6176, opened 2022-02-16
Damage in battle table is now the average damage done

xlenstra:
Instead of the actual damage


yairm210 (approved):
Looks fine but why is the N is in randomness capitalized

PR #6182, opened 2022-02-17
Translation: Update Spanish for 3.19.10

ArchDuque-Pancake:


xlenstra:
This will conflict with #6173

ArchDuque-Pancake:
> This will conflict with #6173

Once we resolve it, how can we 'merge' the two? Or one will have to delete theirs and the other update it?

PR #6190, opened 2022-02-18

Rodolfz147:


yairm210:
Merging this - what was changed on the English side will just be ignored, there's no reason to put off the majority of the correct translations

PR #6193, opened 2022-02-19
Make multiple tile defense bonuses stack

### LINKED ISSUE #6189: Double defensive bonus on hill tiles.
**Platform**
Android 11 (F-Droid)
Linux (Flatpak)

**Version**
3.19.7

**Describe the bug**
Individually, forests/jungle and hills tiles give a 25% defensive bonus to terrain units, but also I can improve it building a fort on those tiles to reach 75%. However, when a forest/jungle has been generated on a hill, defensive bonus is just 25% and should be 50%.

**Expected behavior**
Forests/jungles generated on a hill, should provide a double defensive bonus. (Also, if I have built a fort on a hill with a forest/jungle, should I get a 100% defensive bonus)

itanasi:
Fix #6189

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -595,7 +595,7 @@ open class TileInfo {
 
     fun getDefensiveBonus(): Float {
         var bonus = 0F
-        for (terrain in terrainFeatures){ bonus += ruleset.terrains[terrain]!!.defenceBonus }
+        getAllTerrains().forEach{bonus += it.defenceBonus}
         val tileImprovement = getTileImprovement()
         if (tileImprovement != null) {
             for (unique in tileImprovement.getMatchingUniques(UniqueType.DefensiveBonus))
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -594,7 +594,8 @@ open class TileInfo {
             tileMap.getTilesAtDistance(position, distance)
 
     fun getDefensiveBonus(): Float {
-        var bonus = getLastTerrain().defenceBonus
+        var bonus = 0F
+        for (terrain in terrainFeatures){ bonus += ruleset.terrains[terrain]!!.defenceBonus }
         val tileImprovement = getTileImprovement()
         if (tileImprovement != null) {
             for (unique in tileImprovement.getMatchingUniques(UniqueType.DefensiveBonus))


SomeTroglodyte:
base terrains and nat wonders can't have defensive boni?

Also - just a little thought experiment - shortening and the battle against sumOf not having a Float overload...
```kotlin
inline fun <T> Sequence<T>.sumOf(selector: (T) -> Float) = fold(0f) { sum, element -> sum + selector(element) }
(...)
fun getDefensiveBonus() = getAllTerrains().sumOfFloat { it.defenceBonus } + (getTileImprovement()?.getMatchingUniques(UniqueType.DefensiveBonus)?.sumOfFloat { it.params[0].toFloat() } ?: 0f)
```
... interesting learning experience (kotlin actively defends against extending its libraries), works, not worth it.

yairm210:
Good point, all time layers should be able to contribute
Sounds like getTileLayers.sumOf to me, don't see why this isn't an int?

SomeTroglodyte:
> why this isn't an int

No idea, it just ***is*** a Float at the moment. Maybe the code god likes dice after all.

..Which piqued my interest to see if the missing library overload can be added by non-jetbrain people - it can't. My sample above is incorrect. That sumOf, no matter the annotations or file placement, makes the compiler forget all the predefined ones, simplifying just a little, so the "works" comment was using a renamed function...

yairm210 (approved):

xlenstra:
`sequence<T>.sumOf{}` does have an overload for double, so casting `defenceBonus.toDouble()` makes it possible to use `.sumOf{}` normally
Can also go in another, PR, so for now:

xlenstra (approved):

itanasi:
Yeah, another PR could be revamping the type. Need to double check how stuff is populated to prevent a JSON overhaul

SomeTroglodyte:
> does have an overload for double

Yes, well-known. That's why the missing Float version irks. Casting to and fro is just ugly and slightly inefficient. At the moment I'd tend to `fold` (as the resulting code might be close to what a library-provided sumOf<Float> should give) , but the cast juggling is surely already used elsewhere in the project, too. Generally, I'm anti-32-bit-ieee, in fact in SQL I'd _never_ use `real` even under threats of physical violence, but Gdx has such a lopsidedly strong penchant for Float it seems to irrevocably have infected Unciv...

The current code is perfectly fine except for a space I'd add between 'h' and '{'. And the convention for the Float literal modifier favours lowercase 'f'.

Issue #6178, opened 2022-02-17
Music keeps playing even after pressing 'home' button


jahodovykolac:
**Platform**
Android 10, Android 5.1

**Version**
3.19.8 and some earlier

**Describe the bug**
Music is still playing even if app is minimized using 'home' button.

**To Reproduce**
1. Open app
2. Press home button
3. Wait a while
4. Music is playing

**Expected behavior**
Music is paused until opening the app.

**Additional context**
I already reported this in #5438, but that was closed because of old android 5.1.

SomeTroglodyte:
- Oreo 8.0 AVD - ~~No repro, pauses just fine.~~ see below
- Q 10.0 AVD - ~~pauses just fine.~~ ditto
- LOS 16.1 Hardware - pauses just fine but _after configured delay_ plays a new song. Swiping Unciv out of the app history stops music.

jahodovykolac:
Still happening on 3.19.11

PR #6195, opened 2022-02-19
Fix for #6178 music resumes after minimizing on android

SomeTroglodyte:
Only superficially tested. Other testers welcome.


yairm210 (approved):

AdityaMH:
Still keeping play ambient or random music until track end while minimizing.
But after back to game again.
music become shutdown completely.

Issue #5891, opened 2022-01-01
Suffix Bug Name for Music Track (Android only?)


AdityaMH:
**Platform**
Android 10

**Version**
3.18.13

**Describe the bug**
I'm trying to make music mod with suffix name
`[CivName] - [SongName] - [Conditional].[FileFormat]`

But really didn't work for me

I'm trying to make music name include `Main Menu`, America and Aztecs (also with `Peace` and `War`) but when playing as Aztecs music player keep playing randomly also America, or Main Menu.

Here my investigation:
`Main Menu` - Worked
`Ambient` - (Random even has another suffix name like civ name) when app minimized
`Theme`/`Peace` - (Keep random playlist) Only occur when Technology researched and at peace (after track end still keep playing random track) and diplomacy screen (sometime playing `War` song for selected civ).
`War` - (Keep random playlist) Only occur when Technology researched and at peace after track end still keep playing random track and diplomacy screen.

SomeTroglodyte:
I can't interpret this to any degree of meaningful precision, as I'm a bit distracted.
- Setting MusicController.consoleLog to true and testing with your actual data, reading the console log might help most.
- I _had_ written some documentation, though I can't find it at the moment...
```kotlin
    /**
     * Chooses and plays a music track using an adaptable approach - for details see the wiki.
     * Called without parameters it will choose a new ambient music track and start playing it with fade-in/out.
     * Will do nothing when no music files exist or the master volume is zero.
     * 
     * @param prefix file name prefix, meant to represent **Context** - in most cases a Civ name
     * @param suffix file name suffix, meant to represent **Mood** - e.g. Peace, War, Theme, Defeat, Ambient
     * (Ambient is the default when a track ends and exists so War Peace and the others are not chosen in that case)
     * @param flags a set of optional flags to tune the choice and playback.
     * @return `true` = success, `false` = no match, no playback change
     */
    fun chooseTrack (
        prefix: String = "",
        suffix: String = "Ambient", 
        flags: EnumSet<MusicTrackChooserFlags> = EnumSet.noneOf(MusicTrackChooserFlags::class.java)
    ): Boolean {
```
... yes sounds like there should be some more description ...

AdityaMH:
I think i need to re-write my result.
Condition: "If you play as civ A but no A civ on world screen."
If mod contain "Ambient", will play as "Ambient", if no "Ambient" inside mod. Will play music randomly.

Condition: "If you play as civ A but warring civ B"
[...]

PR #6197, opened 2022-02-19
Wiki chapters for context-sensitive music

SomeTroglodyte:
Thanks to #5891 I remembered I had prepared a few lines of markdown for the music track choosing algorithm, and that I never ensured they got published. Here, I just appended them to the "Audiovisual mods" wiki page.

~~Draft because I'm pretty sure it's by now a bit incomplete - I remember the game can make better use of some of the official music tracks than possible with the listed stuff ("official" as fetched if you follow my scratchpad instructions and scrapers). Thus, needs another few Alt-F7 scans, but I'm too tired right now.~~ 
Code comparison done, now up-to-date, and footnote syntax adapted for github.


yairm210:
Update wiki

PR #6199, opened 2022-02-20
Terrainfeature objects transient

yairm210:
For performance reasons, as uncovered by both memory and CPU profiling, it's better to save terrain feature objects permanently so we can iterate on them constantly without a significant performance overhead
Separated into 2 commits for convenience

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -66,6 +66,10 @@ open class TileInfo {
     var position: Vector2 = Vector2.Zero
     lateinit var baseTerrain: String
     var terrainFeatures: List<String> = listOf()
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -66,6 +66,10 @@ open class TileInfo {
     var position: Vector2 = Vector2.Zero
     lateinit var baseTerrain: String
     var terrainFeatures: List<String> = listOf()
+        private set
+    
+    @Transient
+    var terrainFeatureObjects: List<Terrain> = listOf()
```
xlenstra, line 72: Would it add to also mark this `private set`?

xlenstra (approved):

PR #6161, opened 2022-02-14
Moddable victories

xlenstra:
As already remarked in #6157.
This PR makes victories moddable, and the changes mod makers make will also be shown on the victory screen.

<details><summary> Pictures </summary>

image
Victory screen without any milestones completed

image

image

Victory screen with a few milestones completed

image
Global victory screen

image

image

</details>

I've tested scientific, cultural, diplomatic and time victory after most major changes were made, so I'm pretty certain they work. Save files from before these changes were implemented load just fine, and creating new games should also work without problem.

PR #6203, opened 2022-02-20
Made spaceship parts units instead of buildings

xlenstra:
Cherry-picked from #6161 so at least this can already be used and we don't have to wait until I actually have time to finish that PR.
Existing progress towards scientific victory _should_ not be lost, but I haven't tested this.


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -451,7 +451,8 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     CannotBeBarbarian("Never appears as a Barbarian unit", UniqueTarget.Unit, flags = listOf(UniqueFlag.HiddenToUsers)),
     
     ReligiousUnit("Religious Unit", UniqueTarget.Unit),
-    SpaceshipPart("Spaceship part", UniqueTarget.Building, UniqueTarget.Unit),
+    SpaceshipPart("Spaceship part", UniqueTarget.Unit),
```
yairm210, line 454: By removing the "Building" target you make all existing mods throw errors

android/assets/jsons/translations/template.properties
```diff
@@ -726,6 +726,8 @@ Sleep until healed =
 Moving = 
 Set up = 
 Paradrop = 
+Add in capital =
```
yairm210, line 729: I think the 'every action is translated' test is causing us to have fake translations... this text would never actually be shown, if I understand correctly

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -475,6 +475,7 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
                     )
                         rejectionReasons.add(RejectionReason.ShouldNotBeDisplayed)
 
+                // Shouldn't this be "Enables nuclear weapon_s_"?
```
yairm210, line 478: I personally want this to be removed entirely and replaced by "Only available \<when [Apollo Program] has been completed]>" or somesuch

yairm210: Currently we have the nuclear UNITS saying they need the building, and the BUILDING saying it enables the units, with no direct connection, which is suboptimal

yairm210 (changes requested):
Please return the "Spaceship part" unique to units to not break existing mods

android/assets/jsons/translations/template.properties
xlenstra: Correct, I think this also happened before for the great engineer bonus iirc, where `Hurry Construction` is in the `template.properties` file, while only ever `Hurry Construction (+[amount]⚙)` is actually shown to the user

core/src/com/unciv/models/ruleset/Building.kt
xlenstra: Agree, though for user clarity having it both ways might be benficial

PR #6204, opened 2022-02-20
Translation: Update of Spanish

ArchDuque-Pancake:


SomeTroglodyte:
`Remove [feature] first` is a suggestion on the improvement picker screen - like "you could build a farm here but you would need to remove that forest first"... Your tr() doesn't sound quite like that, but I couldn't come up with a sufficiently short one either.

ArchDuque-Pancake:
Perhaps, altough to me it doesn't look like that in english either. 
But I'll try to think about another option, if not, I would need to see it in game to have a better view.

yairm210:
merge translations

yairm210:
Translations branch created

yairm210:
Translations PR created

PR #6208, opened 2022-02-21
Autoupdates to uniques will multiple solutions now replace to the correct one

yairm210:
Problem - some deprecated uniques can lead to multiple possibilities of replacements, depending on the parameter type
This lead to replacements in the jsons that were either unparseable entirely or were causing errors, both of which needed to be corrected by hand
We now separate such deprecations into their constituent potential replacement uniques, and try and take only the unique that doesn't cause any errors
Works like a charmander :)

### DIFF
--- a/core/src/com/unciv/models/translations/Translations.kt
+++ b/core/src/com/unciv/models/translations/Translations.kt
@@ -262,7 +262,7 @@ fun String.tr(): String {
     }.toHashSet()
     val language = UncivGame.Current.settings.language
 
-    if (contains('<')) { // Conditionals!
+    if (contains('<') && contains('>')) { // Conditionals!
         /**
          * So conditionals can contain placeholders, such as <vs [unitFilter] units>, which themselves
          * can contain multiple filters, such as <vs [{Military} {Water}] units>.
--- a/core/src/com/unciv/Constants.kt
+++ b/core/src/com/unciv/Constants.kt
@@ -59,6 +59,8 @@ object Constants {
     const val lowering = "Lowering"
     const val remove = "Remove "
     
+    const val uniqueOrDelimiter = "\" OR \""
+    
     const val minimumMovementEpsilon = 0.05
     const val defaultFontSize = 18
     const val headingFontSize = 24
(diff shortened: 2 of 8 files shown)


core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -262,7 +262,7 @@ fun String.tr(): String {
     }.toHashSet()
     val language = UncivGame.Current.settings.language
 
-    if (contains('<')) { // Conditionals!
+    if (contains('<') && contains('>')) { // Conditionals!
```
xlenstra, line 265: Technically this doesn't check the order of these, so it could still be that `.removeConditionals()` returns the same string, leading to infinite recursion of `.tr()` calls on the next line. But this is a minor point, as this PR fixes the main way this can happen anyway

SomeTroglodyte:
As troglo, I _do_ like char meanders!

`return uniquesToUnify.joinToString("\", \"")` - will produce literal `A", "B", "C` from collection `["A", "B", "C"]` - right? Use prefix and suffix args of joinToString?

`replacementUniqueText.split("\" OR \"")` and `@Deprecated("as of 3.19.8", ReplaceWith("...")) NotDisplayedWithout` - these match, but it's a bit ugly that the simpler split actually _requires_ the ReplaceWith string to use quotes around the OR's but _omit_ the first and last quote character that would follow from typical conventions... `"A" OR "B" OR "C"` would look nicer in some obscure context... I repeat, technically clean but might look a little strange. ...Also, how about `const`'ing the delimiter and reuse the const in the ReplaceWith?

yairm210:
I did think of const'ing it, but then I checked and was there were only 2 uses - one in the test and the other in getDeprecationText - and decided it wasn't worth it

yairm210:
The alternative is to replace with non- \" OR which now that we should have only 1 correct deprecation text could work well actually... but sounds like a job for a different PR, I will add the constant to make work on future PR easier

SomeTroglodyte:
I meant the const could also make the definition a little prettier like:
```kotlin
    @Deprecated("as of 3.19.8", ReplaceWith(
            "Only available <after adopting [buildingName/tech/resource/policy]>" +
            Constants.uniqueOrDelimiter +
            "Only available <with [buildingName/tech/resource/policy]>" +
            Constants.uniqueOrDelimiter +
            "Only available <if [buildingName/tech/resource/policy] is constructed>" +
            Constants.uniqueOrDelimiter +
            "Only available <after discovering [buildingName/tech/resource/policy]>"))
    NotDisplayedWithout("Not displayed as an available construction without [buildingName/tech/resource/policy]", UniqueTarget.Building, UniqueTarget.Unit),
```
... no matter. just a matter of taste.

PR #6210, opened 2022-02-21
Reorganized the way city states grant resources & rounded a float to an int

### LINKED ISSUE #6168: City state UI ideas
Screenshot_20220215-113207_UnCiv

- [x] Influence gets approximated, so here it shows 77 instead of 76.9988
- [x] Resources get subtracted. This can be really annoying in late game when a city states says 3 coal, 1 coal, -2 coal, -1 coal, so you have to calculate it for yourself. (My 2 braincells can't do that)

xlenstra:
Basically fixes #6168 by implementing all its ideas. Also fixes that the resources of the city-state are actually duplicated instead of just ignoring their buildings in the calculation.

<details> <summary> Pics: </summary>

Before:
image

After:
image

The -1 has disappeared as this resource was from a horseman.

</details>

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -237,15 +237,20 @@ Hostile =
 Irrational = 
 Personality = 
 Influence = 
+
+Ally: [civilization] with [amount] Influence = 
 Reach 30 for friendship. = 
 Reach highest influence above 60 for alliance. = 
 When Friends: = 
 When Allies: = 
 The unique luxury is one of: = 
-
 Demand Tribute = 
 Tribute Willingness = 
 At least 0 to take gold, at least 30 and size 4 city for worker = 
+Take [amount] gold (-15 Influence) = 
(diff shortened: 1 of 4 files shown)


android/assets/jsons/translations/template.properties
```diff
@@ -237,15 +237,20 @@ Hostile =
...
 Reach 30 for friendship. = 
 Reach highest influence above 60 for alliance. = 
 When Friends: = 
 When Allies: = 
 The unique luxury is one of: = 
-
 Demand Tribute = 
 Tribute Willingness = 
 At least 0 to take gold, at least 30 and size 4 city for worker = 
+Take [amount] gold (-15 Influence) =
```
yairm210, line 250: You sure that negative influence doesn't scale with gold taken?

yairm210: Sorry now I see you're just moving this around, nvm

yairm210 (approved):

PR #6211, opened 2022-02-21
Added mod constants for the distance between two cities

xlenstra:


SomeTroglodyte:
Looks perfectly fine, though it's a pity the knowledge that 2<3 and the optimizations that follow are lost.

Issue #6186, opened 2022-02-18
Petra is unavailable on single desert


saud2410:
3.19.9 apk
on Android 11, no mods

[Greece - 1 turns.zip](https://github.com/yairm210/Unciv/files/8094504/Greece.-.1.turns.zip)
ZIP contains 2 jsons; medieval era and information era.

Currently, Petra is unavailable on single desert.
Below shows Greece - 1 turns medieval.json.
![Screenshot_20220218-110507_UnCiv](https://user-images.githubusercontent.com/64586749/154626485-f1a106ef-2dec-406f-ad98-e8f6b73c1849.jpg)

I made a new game on a custom map in which Greece starts on a single desert.
I started from Medieval and Information era, and both Petra and solar plant is unavailable on a single desert.
But according to Civ wiki, they are available "near or on the desert."
[Petra](https://civilization.fandom.com/wiki/Petra_(Civ5)) : ...City must be built on or next to desert
[Solar Plant](https://civilization.fandom.com/wiki/Solar_Plant_(Civ5)) : ...However, it can only be constructed in a city which is on or adjacent to a desert tile.

SomeTroglodyte:
Other uses of the `MustBeNextTo` unique use tile filters 'Coast', 'Fresh water', or 'Mountain'. As far as I can guess, those would not be affected, as cities cannot be built on Coast or Mountain; and a single Fresh water tile without any of its neighbors also being freshwatered is impossible - is it (= are Gardens affected and can't be built on the only Oasis in a riverless desert - I don't think so)?

Any negative side effects or negative political implications to change the Unique to _mean_ 'next to or on' without deprecating it or changing its wording?

Both implemetations (building and improvement) rely on `TileInfo.isAdjacentTo` which uses `neighbors` - and those do not include the center tile. `isAdjacentTo` in turn is used only by this Unique and the `ChangesTerrain` Unique. For the latter, changing the meaning of `isAdjacentTo` could be detrimental as I see it. Could - not in the present rulesets as those use only 'River' as tilefilter which takes a different path within `isAdjacentTo`.

Clean and simple might be to create a `TileInfo: fun isAdjacentToOrOn(terrainFilter: String) = isAdjacentTo(terrainFilter) || matchesFilter(terrainFilter)` and use that for `MustBeNextTo` but not `ChangesTerrain`? ...????

Umm... @  Unciv-Experts: Why doesn't `isAdjacentToFreshwater` recurse over the entire map?

[...]

PR #6219, opened 2022-02-23
Freshwater crisis fix

yairm210:
Fresh water is currently a mess, as outlined in #6186 
Step 1 of resolving #6186 freshwater crisis is to unify adjacency to a single function, no more isAdjacentToFreshwater.

Remaining problems, not part of PR but here for discussion:
 - "[+1 Food] from [Fresh Water] tiles" and even worse, "[+1 Food] from [non-freshwater] tiles" - replaced with new tile conditionals, \<in tiles adjacent to [Fresh water]> and \<in tiles not adjacent to [Fresh water]>
 - "Can also be built on tiles adjacent to fresh water" that enables farms on *hills* can be replaced with *enabling* hills and adding "Cannot be built on [tileFilter] tiles <in tiles not adjacent to [Fresh water]>"
 
That should get rid of existing problems while also adding adjacency conditionals and removing an extremely badly-worded unique

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -51,14 +51,6 @@ open class TileInfo {
     @Transient
     var isOcean = false
 
-    // This will be called often - farm can be built on Hill and tundra if adjacent to fresh water
-    // and farms on adjacent to fresh water tiles will have +1 additional Food after researching Civil Service
-    @delegate:Transient
-    val isAdjacentToFreshwater: Boolean by lazy {
-        matchesTerrainFilter("River") || matchesTerrainFilter("Fresh water")
-                || neighbors.any { it.matchesTerrainFilter("Fresh water") }
-    }
-
     var militaryUnit: MapUnit? = null
     var civilianUnit: MapUnit? = null
     var airUnits = ArrayList<MapUnit>()
@@ -398,7 +390,7 @@ open class TileInfo {
                     .sumOf { it.params[0].toInt() }
         }
         if (isAdjacentToRiver()) fertility += 1
-        if (isAdjacentToFreshwater) fertility += 1 // meaning total +2 for river
+        if (isAdjacentTo(Constants.freshWater)) fertility += 1 // meaning total +2 for river
         if (checkCoasts && isCoastalTile()) fertility += 2
         return fertility
     }
@@ -435,8 +427,8 @@ open class TileInfo {
                     // Freshwater and non-freshwater cannot be moved to matchesUniqueFilter since that creates an endless feedback.
                     // If you're attempting that, check that it works!
                     // Edit: It seems to have been moved?
-                    || unique.params[1] == "Fresh water" && isAdjacentToFreshwater
-                    || unique.params[1] == "non-fresh water" && !isAdjacentToFreshwater
+                    || unique.params[1] == Constants.freshWater && isAdjacentTo(Constants.freshWater)
+                    || unique.params[1] == "non-fresh water" && !isAdjacentTo(Constants.freshWater)
                 )
                     stats.add(unique.stats)
             }
@@ -496,10 +488,12 @@ open class TileInfo {
         }
     }
 
+    // This should be the only adjacency function
     fun isAdjacentTo(terrainFilter:String): Boolean {
-        if (terrainFilter == "Fresh water" && isAdjacentToFreshwater) return true
-        return if (terrainFilter == "River") isAdjacentToRiver()
-        else neighbors.any { neighbor -> neighbor.matchesFilter(terrainFilter) }
+        // Rivers are odd, as they aren't technically part of any specific tile but still count towards adjacency
+        if (terrainFilter == "River") return isAdjacentToRiver()
+        if (terrainFilter == Constants.freshWater && isAdjacentToRiver()) return true
+        return neighbors.any { neighbor -> neighbor.matchesFilter(terrainFilter) }
     }
 
     /** Without regards to what CivInfo it is, a lot of the checks are just for the improvement on the tile.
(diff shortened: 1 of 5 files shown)


SomeTroglodyte:
So `ConditionalInTilesNot` was defined but not actually implemented before? Strange.
And - will this count an Oasis itself as having fresh H<sub>2</sub>O? It should, right?

yairm210:
Yes, and Oasis has the "Fresh water" unique, so yes.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -528,14 +522,14 @@ open class TileInfo {
...
             improvement.terrainsCanBeBuiltOn.isEmpty() && improvement.turnsToBuild == 0 && isLand -> true
             improvement.terrainsCanBeBuiltOn.contains(topTerrain.name) -> true
-            improvement.uniqueObjects.filter { it.type == UniqueType.MustBeNextTo }.any {
+            improvement.getMatchingUniques(UniqueType.MustBeNextTo).any {
                 !isAdjacentTo(it.params[0])
             } -> false
             !isWater && RoadStatus.values().any { it.name == improvement.name && it > roadStatus } -> true
             improvement.name == roadStatus.removeAction -> true
             topTerrain.unbuildable && !improvement.isAllowedOnFeature(topTerrain.name) -> false
             // DO NOT reverse this &&. isAdjacentToFreshwater() is a lazy which calls a function, and reversing it breaks the tests.
```
SomeTroglodyte, line 531: Comment obsolete

SomeTroglodyte (approved):
I didn't test, but can't see how this could go wrong

xlenstra:
> "Can also be built on tiles adjacent to fresh water" that enables farms on hills can be replaced with enabling hills and adding "Cannot be built on [tileFilter] tiles <in tiles not adjacent to [Fresh water]>"

Doesn't the current setup also allow building farms on snow, tundra, desert (next to oasis), etc., which would then be lost? I'm not sure what the original does, but at least for modding deprecating that unique would be a loss.
Additionally, "[...] on hills tiles in tiles adjacent [...]" to me sounds a bit weird, with  'tiles' being in the same sentence so close together. Maybe make a new unique "adjacent to [tileFilter]" that behaves the same so modders can choose which sounds better? Or just fix this in `uniqueReplacementText` or however it is called?

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -574,7 +568,7 @@ open class TileInfo {
             "Water resource" -> isWater && observingCiv != null && hasViewableResource(observingCiv)
             "Natural Wonder" -> naturalWonder != null
             "Featureless" -> terrainFeatures.isEmpty()
-            "Fresh Water" -> isAdjacentToFreshwater
+            "Fresh Water" -> isAdjacentTo(Constants.freshWater)
```
xlenstra, line 571: So we still have both "Fresh water" and "Fresh Water" and support only one or both depending on the unique?

SomeTroglodyte: How about `const`ing that one as well, then naming one freshWaterSource and the other freshWaterAccess (the latter the one with capital Water), irrespective of actual string value - so the uses are traceable and changing the actual keys, if at all, becomes more readable?

yairm210: The goal is to remove all support for "Fresh Water", but this will I think require adding a warning in the ruleset checks

xlenstra (approved):

SomeTroglodyte:
Some more loud thinking:
- Capitalized "Fresh Water" is now _only_  used in the conditional of the resource generation uniques, only implemented in `TileInfo.matchesTerrainFilter` and allowed by `UniqueParameterType.TerrainFilter.knownValues` - consistent and clean. A comment pointing out the special role of the "W" in matchesTerrainFilter might help.
- Gardens (and Floating Gardens) - will use `cityCenter.isAdjacentTo("Fresh water")` which, in absence of rivers, will allow the building next to a lake or oasis but not if the city is _on_ the source and no neighbor has one. Petra issue again.
- Farm (and Terrace farm) - uses `ImprovementStatsOnTile` - the one with the "endless feedback" comment. The question of the center tile is moot, no Farm on Oasis...
- Lower case "Fresh water" now means "this tile is a source" or "check for access to" depending on circumstances. A Tilefilter testing whether "this" tile is a source without looking at the neighbors is no longer possible.
- The `FreshWater`UniqueType could use the const, too

Tldr - more analysis and my approval still stands.

I'd say we need unit tests for a few typical vanilla ruleset situations - There's already one test afaik that loads an entire map from a gzipped string, we could go that route...

SomeTroglodyte:
<details><summary>Toying with this <code>FreshWaterTests.kt</code>:</summary>

```kotlin
package com.unciv.logic.map

import com.unciv.Constants
import com.unciv.logic.MapSaver
import com.unciv.models.ruleset.Ruleset
import com.unciv.models.ruleset.RulesetCache
import com.unciv.testing.GdxTestRunner
import org.junit.Assert
import org.junit.Before
import org.junit.Test
import org.junit.runner.RunWith

@RunWith(GdxTestRunner::class)
class FreshWaterTests {
    private var map = TileMap()
    private var ruleSet = Ruleset()

    companion object {
        private const val testingMap = "H4sIAAAAAAAA/51TvU7DMBB+lTyAOzSI5UYKFUOkooDKUHUw5CAnnDiyz4Vg5d1xlIn8tc0SyfH3p/vOvpDVkzSyQEZjwXNdITwUFdci3DzTL4I3MiNnIRbflHEOtyJH+swZbkQZeLBxlnXRiDdpMXUKLTJs6BTto1W0lyUpJRvBpDAhy3DwLe4FjZFUwn1AGxYGrXbmHeE1R8mN8JW2xKRL8D+wWos6fDqD/8QesgWex02JzaWY0J0PvoByfbLAidv5xKNE7o7bwHJBBA47ackeewqtQI+fyC+0o0VcaPQYeu/7rIc+l9MnrWdru5rSpZzoIZf2TnPY9gQ/OKUTGmDjcGxOZwTS9gVNKixIff1wut7Hcw762Cqts6hS4edgfboFXCR0bP4A33i02IEEAAA="
    }

    @Before
    fun initTheWorld() {
        RulesetCache.loadRulesets()
        ruleSet = RulesetCache.getVanillaRuleset()
        map = MapSaver.mapFromSavedString(testingMap, false)
        map.setTransients(ruleSet, false)
    }

    @Test
    fun isAdjacentToFreshWater() {
        for (tile in map.values) {
            val isFresh = tile.isAdjacentTo(Constants.freshWater)
            val hasWheat = tile.resource == "Wheat"
            Assert.assertFalse("Tile $tile has fresh water but should not", isFresh && hasWheat)
            Assert.assertFalse("Tile $tile should have fresh water but doesn't", !isFresh && !hasWheat)
        }
    }

    fun canBuildFarm() {
        //TODO
    }

    fun canBuildGarden() {
        //TODO
    }
}
```
</details>
<details><summary>Which uses this map:</summary>

image
</details>
(I misused Wheat as a marker for tiles that shouldn't have fresh water)

...Fails because loading a map and setTransients on it will still leave all the `terrainFeatureObjects` empty, and if one fixes that, then...

Issue #6192, opened 2022-02-19
"Convert gold to science" slider causes AI to go bankrupt


SpacedOutChicken:
Ever since adding the "Can convert gold to science with sliders" to Smoky Skies Redux, AI civilizations have started going bankrupt late in the game, which they were not doing before. I believe that they are converting too much gold into science, and the AI is not smart enough to realize that it doesn't have to go into debt.

Ideally, I'd like the AI to be able to recognize when it's losing money and adjust its science conversion accordingly, but in the short term, I have a simple solution: set the default level of gold-to-science to zero.

SomeTroglodyte:
Might be easier to check out if there were a save & explicit info whether the mod used matches its repo 100% or is modified

SpacedOutChicken:
I've attached a mod with the slider enabled here, along with a saved game from that mod. Just 18 turns in, the AIs are all deep in debt with no end in sight.
[Debt.zip](https://github.com/yairm210/Unciv/files/8121374/Debt.zip)

SomeTroglodyte:
Thx. I might look into it if I _fail_ to get my priorities straight...

SomeTroglodyte:
Just 'thinking' aloud:
- The AI never checks the `convertGoldToScience` 'Unique'
- The AI never touches `TechManager.goldPercentConvertedToScience`
- The default is 0.6f for all civs. The var is saved, so changing the default could potentially annoy someone.
- stack from where the value is applied: CityStats.updateFinalStatList - CityStats.update - 33 call sites, I pick CityInfo.startTurn - CivilizationInfo.startTurn - GameInfo.nextTurn
- So where would we hook in an AI routine for that percentage? At the point where the 'slider' is used during nextTurn, neither the initial treasury at the start of this AI's turn nor the final treasury at the end of switchTurn are known to any degree of approximation, and the value changes for each city called. Deciding a percentage ad hoc, per city, right before it is used seems to depend on too many random or inscrutable factors. So - at the start of the AI's turn, just after any existing `updateStatsForNextTurn` call, and do some logic from that gold surplus estimate, and maybe era...

SomeTroglodyte:
With that PR, income stabilizes somewhat _after_ one turn, right after loading it seems to base the displayed stats on the saved (=default) conversion rate, but that's fine by me.

PR #6222, opened 2022-02-23
UncivSlider fixes

SomeTroglodyte:
See also commit messages. I noticed one little imperfection long ago and forgot until now, and another thanks to #6192.

Old UncivSlider has issues combining the added +/- buttons with `isDisabled` and `setSnapToValues` - this makes a disabled UncivSlider "immutable" by drag _and_ +/-, and makes +/- use snap positions if they're defined. Also, a helper for easier value tip display as percentage.

Affects sliders: Max turns for time victory, Convert gold to science, Music and Sound effect volumes.

Too much code for little improvement, but as is the widget doesn't keep all its promises, which may stump future uses.


yairm210:
Looks can't getTipText always be the float format? It looks overridable but never actually overridden, which to me indicates it shouldn't be overridable at all. YAGNI &c

SomeTroglodyte:
Default format is float, number of decimals depends on step size. getTipText _is_ used. What's not really used atm is the client modifying the format string property `tipFormat` without replacing its use via getTipText - a percentage would be representable as format string in .net if I remember correctly, but I can't find the java equivalent. Ron Jeffries? True, I tend to not do XP when offering reusable widgets. In this case, making the property private is possible, but removing it entirely would break something.

Wait, maybe I misunderstand and you suggesting to always go through getTipText and defaulting it to a stub doing the `tipFormat.format(slider.value)` instead of routing via if(null)... I suppose that would save a line or two, yes. Might look at it later.

SomeTroglodyte:
Got the variant working (the one making `getTipText` the only way to influence formatting for clients) - but there were so many surprises on the way I'm not happy to present that as "the" better way. Decisions: don't make `getTipText` a var, don't change `getTipText`'s signature to give it instance access, and consequently do make stepSize a val to avoid inconsistencies, needing changes to MapParametersTable...

So what I just pushed is just comments, plugging a microscopic cosmetic bug, and making the format string private...

SomeTroglodyte:
<details><summary>The discarded variant as patch if someone's interested:</summary>

```patch
Index: core/src/com/unciv/ui/utils/UncivSlider.kt
IDEA additional info:
<+>UTF-8
diff --git a/core/src/com/unciv/ui/utils/UncivSlider.kt b/core/src/com/unciv/ui/utils/UncivSlider.kt
--- a/core/src/com/unciv/ui/utils/UncivSlider.kt	(revision 11d69bedb6bf6c9ec8c3ac727db303334543cf73)
+++ b/core/src/com/unciv/ui/utils/UncivSlider.kt	(date 1645735555214)
@@ -41,7 +41,7 @@
     plusMinus: Boolean = true,
     initial: Float = min,
     sound: UncivSound = UncivSound.Slider,
-    private val getTipText: ((Float) -> String)? = null,
+    private val getTipText: ((Float) -> String) = getFormatter(step),
     onChange: ((Float) -> Unit)? = null
 ): Table(BaseScreen.skin) {
     companion object {
@@ -49,6 +49,12 @@
         fun formatPercent(value: Float): String {
             return (value * 100f + 0.5f).toInt().toString() + "%"
         }
+        private fun getFormatter(step: Float): ((Float) -> String) = when {
+            step > 0.99f -> { value: Float -> "%.0f".format(value) }
+            step > 0.099f -> { value: Float -> "%.1f".format(value) }
+            step > 0.0099f -> { value: Float -> "%.2f".format(value) }
+            else -> { value: Float -> "%.3f".format(value) }
+        }
         // constants for geometry tuning
         const val plusMinusFontSize = Constants.defaultFontSize
         const val plusMinusCircleSize = 20f
@@ -81,12 +87,8 @@
             slider.value = newValue
             valueChanged()
         }
-    var stepSize: Float
+    val stepSize: Float
         get() = slider.stepSize
-        set(value) {
-            slider.stepSize = value
-            stepChanged()
-        }
     /** Returns true if the slider is being dragged. */
     @Suppress("unused") // Part of the Slider API
     val isDragging: Boolean
@@ -110,9 +112,6 @@
         slider.setSnapToValues(values, threshold)
     }
 
-    // java format string for the value tip, set by changing stepSize 
-    private var tipFormat = "%.1f"
-
     /** Prevents hiding the value tooltip over the slider knob */
     var permanentTip = false
 
@@ -128,8 +127,6 @@
             }
         }
 
-        stepChanged()   // Initialize tip formatting
-
         if (plusMinus) {
             minusButton = "-".toLabel(Color.BLACK, plusMinusFontSize)
                 .apply { setAlignment(Align.center) }
@@ -208,11 +205,7 @@
 
     // Visual feedback
     private fun valueChanged() {
-        if (getTipText == null)
-            tipLabel.setText(tipFormat.format(slider.value))
-        else
-            @Suppress("UNNECESSARY_NOT_NULL_ASSERTION") // warning wrong, without !! won't compile
-            tipLabel.setText(getTipText!!(slider.value))
+        tipLabel.setText(getTipText(slider.value))
         if (!tipHideTask.isScheduled) showTip()
         tipHideTask.cancel()
         if (!permanentTip)
@@ -229,17 +222,6 @@
         plusButton?.apply {circle.color.a = if(enablePlus) 1f else 0.5f}
     }
 
-    private fun stepChanged() {
-        tipFormat = when {
-            stepSize > 0.99f -> "%.0f"
-            stepSize > 0.099f -> "%.1f"
-            stepSize > 0.0099f -> "%.2f"
-            else -> "%.3f"
-        }
-        if (getTipText == null)
-            tipLabel.setText(tipFormat.format(slider.value))
-    }
-
     // Attempt to prevent ascendant ScrollPane(s) from stealing our focus
     private val killedListeners: MutableMap<ScrollPane, List<EventListener>> = mutableMapOf()
     private val killedCaptureListeners: MutableMap<ScrollPane, List<EventListener>> = mutableMapOf()
Index: core/src/com/unciv/ui/newgamescreen/MapParametersTable.kt
IDEA additional info:
<+>UTF-8
diff --git a/core/src/com/unciv/ui/newgamescreen/MapParametersTable.kt b/core/src/com/unciv/ui/newgamescreen/MapParametersTable.kt
--- a/core/src/com/unciv/ui/newgamescreen/MapParametersTable.kt	(revision 11d69bedb6bf6c9ec8c3ac727db303334543cf73)
+++ b/core/src/com/unciv/ui/newgamescreen/MapParametersTable.kt	(date 1645735455361)
@@ -241,8 +241,8 @@
         table.add("RNG Seed".toLabel()).left()
         table.add(seedTextField).fillX().padBottom(10f).row()
 
-        fun addSlider(text: String, getValue:()->Float, min:Float, max:Float, onChange: (value:Float)->Unit): UncivSlider {
-            val slider = UncivSlider(min, max, (max - min) / 20, onChange = onChange)
+        fun addSlider(text: String, getValue: ()->Float, min: Float, max: Float, step: Float = 0f, onChange: (value:Float)->Unit): UncivSlider {
+            val slider = UncivSlider(min, max, if (step == 0f) (max - min) / 20 else step, onChange = onChange)
             slider.value = getValue()
             table.add(text.toLabel()).left()
             table.add(slider).fillX().row()
@@ -265,11 +265,11 @@
         addSlider("Rare features richness", {mapParameters.rareFeaturesRichness}, 0f, 0.5f)
         { mapParameters.rareFeaturesRichness = it }
 
-        addSlider("Max Coast extension", {mapParameters.maxCoastExtension.toFloat()}, 0f, 5f)
-        { mapParameters.maxCoastExtension = it.toInt() }.apply { stepSize = 1f }
+        addSlider("Max Coast extension", {mapParameters.maxCoastExtension.toFloat()}, 0f, 5f, 1f)
+        { mapParameters.maxCoastExtension = it.toInt() }
 
-        addSlider("Biome areas extension", {mapParameters.tilesPerBiomeArea.toFloat()}, 1f, 15f)
-        { mapParameters.tilesPerBiomeArea = it.toInt() }.apply { stepSize = 1f }
+        addSlider("Biome areas extension", {mapParameters.tilesPerBiomeArea.toFloat()}, 1f, 15f, 1f)
+        { mapParameters.tilesPerBiomeArea = it.toInt() }
 
         addSlider("Water level", {mapParameters.waterThreshold}, -0.1f, 0.1f)
         { mapParameters.waterThreshold = it }
```
</details>

yairm210:
All methods are fine, none of this changes the big picture, which is still "things work".

yairm210 (approved):

PR #6224, opened 2022-02-25
Block Embarked from capturing Civilians on Water

itanasi:
Naval can capture Embarked Civilian
image
Naval can't capture Civilian on Land
image
Embarked can't capture Civilian on Water
image
Embarked can capture Civilian on Land
image
Land can capture/attack Land Civilian, and water units, and Embarked Civilian
image

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -40,7 +40,7 @@ class MapUnit {
     lateinit var baseUnit: BaseUnit
 
     @Transient
-    internal lateinit var currentTile: TileInfo
+    lateinit var currentTile: TileInfo
 
     @Transient
     val movement = UnitMovementAlgorithms(this)
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -850,8 +850,6 @@ class MapUnit {
     }
 
     fun removeFromTile() = currentTile.removeUnit(this)
-    // ONLY FOR DEBUG/Unit Testing
-    fun testingSetCurrentTile(tile: TileInfo) { currentTile = tile }
 
     fun moveThroughTile(tile: TileInfo) {
         // addPromotion requires currentTile to be valid because it accesses ruleset through it.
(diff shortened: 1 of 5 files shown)


itanasi:
Draft until better way to fix the unit test

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -850,6 +850,8 @@ class MapUnit {
     }
 
     fun removeFromTile() = currentTile.removeUnit(this)
+    // ONLY FOR DEBUG/Unit Testing
+    fun testingSetCurrentTile(tile: TileInfo) { currentTile = tile }
```
yairm210, line 854: I disagree with this existing, There must be a nicer way to do this.

SomeTroglodyte: Tentative opinion: Unit tests should conform to useful APIs??? So if the test needs to move a unit, use an existing move function... Or if the test just needs a unit with certain characteristics on the new tile, (re-)create one...

itanasi: Understandable. Just means that the tile I use has to be more fully fleshed out in order to use putInTile

SomeTroglodyte: "fleshed out" - maybe be lazy and use micro-maps loaded using the std map loader [like here](https://github.com/yairm210/Unciv/blob/535c4e9ab0552daa46dca1f0371dc2c773040fb2/tests/src/com/unciv/logic/map/FreshWaterTests.kt). Main caveat: Such a map isn't readily visualizable to the coder. Yes, a TileMap loaded from json _can_ include units, we just do not have the editor for them.

(Idea: Build an Android Studio plugin that actually draws such a map from the string literal...)

Don't take me too seriously here!

itanasi: The real problem is that this check requires the MapUnit to check it's currentTile. So one way to set currentTile is via moveThroughTile. Which also needs the ruleset and to run updateVisibleTiles which needs to know the Current civilization to evaluate. Which means we basically need a full game running.
The other option is via the TileInfo.setUnitTransients(). Except for some reason it keeps complaining I haven't set a ruleset property even though I have. Ugh

itanasi: Quickest way is make currentTile not internal

PR #6230, opened 2022-02-25
Fresh water unit tests

SomeTroglodyte:
Follow-up to ideas presented in #6219, solves #6186. Nevertheless, please evaluate carefully (actual play testing welcome) - the PR is meant more as idea to further the discussion. I did playtest just a little setup specifically to check the Petra issue.

To consider:
- the setTerrainFeatureTransients thingy is needed so the isAdjacentTo function actually works in that unit test; I believe the issue is likely to affect regular gameplay too (see linked issues), therefore this approach.
- the actual Petra fix is just "the seventh (day, son, seal, samurai, tile)" -  I see no more reason why isAdjacentTo shouldn't also check the center tile, but I might be wrong.
- The test is meant to be expandable, but I lack ideas. The mini-map it sets up is the one from #6219 with all Wheat removed. e.g. would it be worth it to actually test farm buildable on hill with fresh water?

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -821,12 +821,17 @@ open class TileInfo {
         isWater = getBaseTerrain().type == TerrainType.Water
         isLand = getBaseTerrain().type == TerrainType.Land
         isOcean = baseTerrain == Constants.ocean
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
OK methinks splitting this up is better.

Issue #6226, opened 2022-02-25
Align Embarked Unit Strengths


itanasi:
Some notes from testing in Civ 5

In Civ 5, Embarked Units have reduced strength
![image](https://user-images.githubusercontent.com/44038014/155671930-5ceee1ab-d8d5-44a2-a821-5029a799043e.png)
![image](https://user-images.githubusercontent.com/44038014/155672313-6b3cbd5f-9234-4ede-8d46-d40eb107a329.png)
Embarked Workers have a strength?
![image](https://user-images.githubusercontent.com/44038014/155673076-ba57f67e-8088-4b89-b6fc-304167531633.png)
Amphibious Landing penalty is stacked on top of regular strength
![image](https://user-images.githubusercontent.com/44038014/155673414-e826b691-a992-4ed3-8a82-4116fa5e3daf.png)
Songhai have neither penalty when landing
![image](https://user-images.githubusercontent.com/44038014/155675185-44a1d96b-f1cb-494c-9dae-c29d7175be66.png)
Songhai have double the base Embarked Strength
![image](https://user-images.githubusercontent.com/44038014/155675593-77558523-6754-4c69-a499-90617e17cf4d.png)
![image](https://user-images.githubusercontent.com/44038014/155675656-6dbd4b63-78a9-4f22-a9f5-d318cb3d702b.png)
Which does oddly means somehow they have more def strength Embarked as an Archer than base Archer Ranged Attack
![image](https://user-images.githubusercontent.com/44038014/155675751-cdfe0dca-54b9-4cec-b0f7-37c4879d9dd2.png)
Interestingly, Songhai Workers don't get the War Canoe promotion
![image](https://user-images.githubusercontent.com/44038014/155675909-6b59dfae-2ef5-4090-b794-359c1c91ca16.png)
Medieval it's Str 6
![image](https://user-images.githubusercontent.com/44038014/155676651-97ec27d2-65a9-4b70-8dfb-23c3a952f913.png)

itanasi:
Information Age base embarked is 20str
Marines get Embarkation w/ Defense, doubling their Embark Strength
![image](https://user-images.githubusercontent.com/44038014/155677335-fbe2d979-2116-49d5-ae56-2d1460c001bd.png)

PR #6231, opened 2022-02-26
Align (reduce) Embarked Defense Strength per Era

itanasi:
Based on #6226 findings from game

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Eras.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Eras.json
@@ -11,6 +11,7 @@
         "startingMilitaryUnit": "Warrior",
         "settlerPopulation": 1,
         "baseUnitBuyCost": 200,
+        "embarkDefense": 3,
         "startPercent": 0,
         "friendBonus": {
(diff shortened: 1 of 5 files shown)


SomeTroglodyte:
Sounds like a case to move the value to a json-read Era property. Also, `}else code` is twice unpretty. Other that that, looks clean and should work just fine.

yairm210:
I completely agree with Trog.
This is a classic value to put in the eras, like embarkedUnitStrength or something

itanasi:
Heh. Was thinking similarly, but didn't want to learn the JSON import code yet. I'll take a look

SomeTroglodyte:
> learn the JSON import code

You don't need to. Era import is already there and will recognize new fields. Unless you use outlandish types or disrupt the default constructor it's auto-magic based on reflection. Since you just want one Int - no worries one should hope.

SomeTroglodyte:
Linked the issue for you. Ya might be interested in [this](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) article.

yairm210 (approved):

PR #5614, opened 2021-10-30
Upgrading to LWJGL 3, which is now the default, as of LibGDX 1.10.1

yairm210:
MacOS running problem solved by adding JVM args to desktop build.gradle
Should resolve #5601

PR #6232, opened 2022-02-26
Fix control-key bindings

SomeTroglodyte:
#5614 seems to have regressed several documented key bindings.
This offers a way to reactivate the Ctrl combos.

This is also a _request for ideas/help_ to solve the issue that all keys are now treated as if the user had a !/&"%§/!&"%§ yankee QWERTY keyboard, which is the reason many of us can't answer a yes/no popup with "Y" anymore. It seems neither Gdx nor ljwrgl3 handle layout translation, that seems left to GLFW, and I can't find either an interface to query the layout not one to do the translation. Interestingly, the GLFW sources (as seen by following call stacks down) contain Jdoc that says layout mapping has already happened at the keycode level before key events are queued... Though I can't find the specific comment at the moment.

AFAIK it has still not been decided whether we want e.g. a yes/no box translated to oui/non to react to "O" instead of "Y" to match the translated button, and how to achieve that. Several ideas offer themselves - first letter of button text, entries in the translation files...

### DIFF
--- a/core/src/com/unciv/ui/utils/KeyPressDispatcher.kt
+++ b/core/src/com/unciv/ui/utils/KeyPressDispatcher.kt
@@ -46,7 +46,7 @@ data class KeyCharAndCode(val char: Char, val code: Int) {
             else -> "\"$char\""
         }
     }
-    
+
     companion object {
         // Convenience shortcuts for frequently used constants
         /** Android back, assigns ESC automatically as well */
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/utils/KeyPressDispatcher.kt
```diff
@@ -66,9 +66,16 @@ data class KeyCharAndCode(val char: Char, val code: Int) {
         /** mini-factory for control codes - case insensitive */
         fun ctrl(letter: Char) = KeyCharAndCode(Char(letter.code and 31), 0)
```
yairm210, line 67: Why are we reducing all chars to only the first 5 bits? Is this just because we needed an arbitrary segment of chars that wouldn't collide with any other chars, or is there some deeper meaning?
In either case, I think a comment would be appropriate

yairm210:
Looks good

SomeTroglodyte:
ctrl('r') == ctrl('R') == KeyCharAndCode(Char(0x12), 0) - and 0x12 is ASCII Ctrl-R. Contrary to EBCDIC, in ASCII letter cases and their control codes can be translated using bit masks. Yes that factory doesn't guard against meaningless input. e.g. ctrl('[') would work to give a Char-based representation of ESC, but the event code wouldn't match that.

I actually tend 51:49 to do away with KeyCharAndCode and use a KeyCodeWithModifiers approach instead, but that's a middling to major refactor. Remember, all this was born when keyTyped was the handled event and matching in the encoded Char realm was preferred. The quick fix just generates Char-based KeyCharAndCode in the event for Ctrl-combos only so we have something distinguishable from the naked key code; and the client's mapping orders go through this factory which already matched that model.

I still haven't found ***any*** handle on how to either determine kdb layout platform-independently or have an existing library translate for us, so let's merge this as what it is...

(For the problem "why can't a kraut hit his "Y" key to say "Yes"" I'm almost convinced giving up and offer configurable key bindings instead, maybe even allow mods to provide a binding json - do not heal the pain of the victim but offer the option to suffer another pain once to make the chronic one go away... Like the [chainsaw](https://www.stihl.de/de/p/kettensaegen-motorsaegen-ms-261-105118#ms-261-c-m-105118) to [cure](https://www.imdb.com/title/tt0365748) [headaches](https://en.wikipedia.org/wiki/Intracranial_aneurysm)).

PR #6233, opened 2022-02-26

heipizhu4:
I have translated all I can, but there're still some of them out of my ability. Note: How to translate ID?


yairm210:
I don't know, don't you have a word for like an identification number or name?

PR #6238, opened 2022-02-26
Fix Polynesia's Wayfinding

### LINKED ISSUE #6202: Polynesian naval units can no longer enter oceans immediately
The new format for the Polynesian unique takes away an ability they had before: the ability for all naval units, even Triremes and such, to immediately enter ocean tiles. I suggest going back to the old unique, or creating a new unique for overriding the "Cannot enter ocean tiles" unique.

SomeTroglodyte:
Fix #6202 by introducing a UnitsMayEnterOcean UniqueType with mapUnitFilter support which Polynesia can use with "All" instead of "Embarked"

Tested:
* [x] Polynesia's embark and trireme can sail ocean immediately
* [x] Non-Polynesia:  Astronomy still as expected
* [x] Modded `UnitsMayEnterOcean` with a unit filter that's neither "All" nor "Embarked"

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Nations.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Nations.json
@@ -488,7 +488,7 @@
 		"innerColor": [255,255,78],
 		"favoredReligion": "Christianity",
 		"uniqueName": "Wayfinding",
-		"uniques": ["Enables embarkation for land units <starting from the [Ancient era]>", "Enables embarked units to enter ocean tiles <starting from the [Ancient era]>", "Normal vision when embarked <for [All] units>", "+[10]% Strength if within [2] tiles of a [Moai]"],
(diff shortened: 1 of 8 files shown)


yairm210:
Nice fix!

yairm210 (approved):

SomeTroglodyte:
In that case - the last test can't affect existing games - ready

SomeTroglodyte:
Nah, the first mod came to mind showed a mistake with that 'specificUnitsCanEnterOcean' path - just a minute...

SomeTroglodyte:
So this now works as expected with a modded polynesia using a "Enables [{Embarked} {Melee}] units to enter ocean tiles <starting from the [Ancient era]>" unique.

I don't like the extra lazy overhead there, but it's clearer than predicting more precisely when the evaluation is needed and it should be faster than potentially evaluating twice. Thought about longer term caching and dismissed it (a mapunit won't change any mapUnitFilter result unless it moves, fights, upgrades or something; and the evaluation is repeated often during a single route-finding call - but imo not worth the risk to miss a cache invalidation reason). Moot anyway unless someone mods something using the feature.

PR #6244, opened 2022-02-27
Fix terrainFeatureObjects not set by setTransients

SomeTroglodyte:
The part of #6230 that resolves #6234, resolves #6235, resolves #6240

PR #6245, opened 2022-02-27
Fix Petra and Garden not allowed in some cases

### LINKED ISSUE #6186: Petra is unavailable on single desert

SomeTroglodyte:
The tiny part of #6230 that fixes #6186 - now separate due to being lower priority than #6244

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -493,7 +493,7 @@ open class TileInfo {
         // Rivers are odd, as they aren't technically part of any specific tile but still count towards adjacency
         if (terrainFilter == "River") return isAdjacentToRiver()
         if (terrainFilter == Constants.freshWater && isAdjacentToRiver()) return true
-        return neighbors.any { neighbor -> neighbor.matchesFilter(terrainFilter) }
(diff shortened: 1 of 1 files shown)


yairm210:
This does change all 'adjacency' checks to include the current tile, I personally can't think of a case where this is not what we want and so I'm okay with it

yairm210 (approved):

SomeTroglodyte:
Yes - I also thought a while on this and the worst anti-argument I got was that the unique text now isn't really saying what it does exactly.

PR #6246, opened 2022-02-27
Unit test for Fresh water filter

SomeTroglodyte:
Last part of #6230, even lower priority. Draft as it will fail without #6244 and #6245 (or remove (-2,-2) and (2,2) from the set to make it run without the latter).


SomeTroglodyte:
The green check re-proves my transient fix was unnecessary... The general approach to this test too whacky? I'd prefer it if we could somehow include images in comments - with the map visualized it's much clearer what happens.

SomeTroglodyte:
OK, `Tile TileInfo @(1.0,-1.0), Desert should have fresh water but doesn't` - I'll check in a few minutes

SomeTroglodyte:
Surprise! That fail @(1,-1) is due to #6266 - before 6266, rivers *on a map border* would sometimes work as fresh water, namely when a wrapped map of the same size would have had a tile on the other side of the river, or sometimes not. Now, no river on a map border will make the tile it is on to fresh water - and that's what I have in that testing map!
<details><summary>The map</summary>

image
</details>
... @(1,-1) is that tile on the very left corner, with the river right on the map border. Thing is, it's not that isConnectedByRiver was called for that nonexistent neighbor, it was taking the wrong code path for the neighbor in the other direction...

**Question** is - do we ***want*** rivers on map edges to work as fresh water or not?
- If yes - `private val isAdjacentToRiverLazy by lazy { hasBottomLeftRiver || hasBottomRiver || hasBottomRightRiver || neighbors.any { isConnectedByRiver(it) } }` [(TileInfo.690)](https://github.com/yairm210/Unciv/blob/03aa489dcaf6847a65844608f8a86a3f6da4c8a9/core/src/com/unciv/logic/map/TileInfo.kt#L690) would do the trick, double tests grudgingly accepted for simplicity (might even be faster on average).
- If no, I'd change the map and the `shouldHaveFreshWater` Set in here.

yairm210:
I'm fine with it not being considered freshwater, this is an extreme edge case

PR #6249, opened 2022-03-01
Return Stacking Terrain Bonus to Civ5 Rules

itanasi:
Natural Wonder could still stack (if they were passable terrain). But now only taking best of terrainFeatureObjects. So that Forest Hills aren't +50%.


itanasi:
Forest Plains Hill has only +25% Defense Bonus
image

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -609,8 +609,12 @@ open class TileInfo {
             tileMap.getTilesAtDistance(position, distance)
 
     fun getDefensiveBonus(): Float {
-        var bonus = 0F
-        getAllTerrains().forEach{bonus += it.defenceBonus}
+        var bonus = baseTerrainObject.defenceBonus
+        if (terrainFeatureObjects.isNotEmpty()) {
+            val otherTerrainBonus = terrainFeatureObjects.maxOf { it.defenceBonus }
+            if (abs(otherTerrainBonus) > 0.000001f) bonus = otherTerrainBonus // never compare to 0f
```
yairm210, line 615: Why not compare to 0f?  Did you see problems with this?

SomeTroglodyte: We had [a discussion about float rounding and what Epsilon to use some time ago](https://github.com/yairm210/Unciv/issues?q=epsilon) (tldr: 1f.ulp might be a good one. Float.MIN_VALUE isn't). However, if the aggregations before the comparison do not include actual math, it's pointless - I see max and abs here, neither would ever round.

itanasi: More just paranoia about float comparison. I can switch to 0f for simplicity

yairm210 (approved):

PR #6250, opened 2022-03-01
Adding more polish translations

simur407:
Added some more polish translations, also fixed some typos from previous translators.


jahodovykolac:
Line 189 has [civState] instead of [civName] in translated part, fix that to make the translation mergeable.

heipizhu4:
Maybe you can try to change fewer lines and update them one by one.

jahodovykolac:
> com.unciv.testing.TranslationTests > allTranslationsHaveCorrectPlaceholders STANDARD_OUT
    Placeholder `[civName]` not found in `Polish` for entry `I thought you might like to know that I've launched an invasion of one of your little pet states.
    The lands of [civName] will make a fine addition to my own.`

The check gives error message only here, saying that in the line there is missing placeholder [civName]. I checked the line and it really says [civState] instead, probably a typo

simur407:
Yup, Made a typo, probably copy-pasted wrong value. It's now fixed. ;)

PR #6253, opened 2022-03-01
WLTK decorations

SomeTroglodyte:
WLTK day seemed a little lacking in clarity and presence, so I tried a few ideas. On the way I changed/fixed some minor things:

* The Pedia links on the TileInfoTable were commented out - fixed.
* Pedia had different return behaviour for Back/ESC and the "Close" button. I refrained from renaming it "Back" as then some might expect it to provide internal back navigation within the Pedia.
* Tooltips had a chance to be clipped if their target sits on a ScrollPane or other clipping widget - fixed.
* Removed outdated comment - Pedia _can_ return to a CityScreen - so no need for a "sorry"

<details><summary>Screenies:</summary>

image
    ...Size of Tooltip should probably be a little larger
image
</details>

* As the screenie says, I'm not happy with the party happening right next to the raze button, but somehow they need to be up top... Input welcome.
* The WLTK column has tips for the header icon and demand, too
* One assignment uses too 'clever' coding style, I intended to simplify but forgot until after the push
* The WLTK string could already be a candidate for Constants?


SomeTroglodyte:
OK, the check on my own fork failed too earlier, but that one was a "[Could not find org.codehaus.groovy](https://stackoverflow.com/questions/50763121/could-not-find-org-codehaus-groovygroovy-all2-4-12)" type - this one looks like foul magic to me, I didn't change any build parameters after all. Runs and tests fine locally.

PR #6262, opened 2022-03-02
Fix new game screen not always respecting translations of selected mods

SomeTroglodyte:
Steps to reproduce:
- Load a base mod translating the Barbarians to something else, e.g. the one from #6192 - and run in English
- Create a game with that mod
- (while in New game screen observe how changing the base ruleset changes the barbarians options to their translated form and back - cool!)
- Quit Unciv and relaunch
- Go to New Game
- Observe the mod is pre-selected but its tranlsation isn't active (raging Barbarians instead of raging Luddites)

- Enter any game
- From the world menu go to new game
- Observe the Barbarians or Luddites options no longer change with the base ruleset

Cause for #1: GameOptionsTable has the code to update translationActiveMods in onChooseMod, but that one does not run during init.
Cause for #2: Translations uses translationActiveMods _only_ while there's no game running under the hood.

This solution solves #2 with an override flag - the danger being missing setting it back. That dispose capture works in testing, but I'm not 100% sure this is the best solution. I encourage thought, own testing, or even rejection as "not worth it" ;)

### DIFF
--- a/core/src/com/unciv/models/translations/Translations.kt
+++ b/core/src/com/unciv/models/translations/Translations.kt
@@ -29,8 +29,8 @@ import kotlin.collections.LinkedHashSet
  *
  *  @see    String.tr   for more explanations (below)
  */
-class Translations : LinkedHashMap<String, TranslationEntry>(){
-    
+class Translations : LinkedHashMap<String, TranslationEntry>() {
+
     var percentCompleteOfLanguages = HashMap<String,Int>()
             .apply { put("English",100) } // So even if we don't manage to load the percentages, we can still pass the language screen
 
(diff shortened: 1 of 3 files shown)


yairm210:
I'm on the fence about this.
If there was a nice way to declare a 'parameter scope', so that the tr() function could check if it's within a scope and act accordingly, that would be one thing
But I already don't really like the fact that we use what is basically static globals, aka the root of all evil, to determine translation. Adding more complexity to that system sounds like a bad idea to me.

Alternative approach - have the Translation module *always* have a "mods to translate" parameter, always use the same one. Set it when entering a game, save it when opening the new game screen, and set it before configuring each mod UI piece, and then reset it to saved value when exiting the screen, That'll also solve the existing "if/else" withing the tr() function currently.

SomeTroglodyte:
>root

True. Very. Thing is, `tr` is an extension to a built-in primitive, which practically forces any additional dynamic context to be - un-dynamic. Any 'OOP best practice' fix wouldn't allow the same call pattern???

Since the defect needs so outlandish conditions to be actually seen, maybe we should forget it. At least the issue is now 'documented' here.

yairm210:
Fair enough
I don't mean to brag but ignoring small problems is one of my superpowers ;)

SomeTroglodyte:
> superpowers

You Luddite!

PR #6263, opened 2022-03-02
Locate Mod Errors choose base ruleset to do complex check against

SomeTroglodyte:
... see also commit titles.
image

Not mentioned side effect: A "complex" check is omitted for base ruleset mods. I opted to skip them instead of staying with the almost-same check (`RulesetCache.checkCombinedModLinks(linkedSetOf(mod.name))` _does_ end up doing the same as `mod.checkModLinks()` - _if_ it's a base ruleset mod, so the only difference is `forOptionsPopup = true` which the "more thorough" run omits, probably by oversight).

I also did not touch the broken auto-updater (uniques with multiple parameters are split into one per parameter with that parameter's original value passed along while all other parameters get replaced with the generic placeholder name) - not the place. I needed the base-ruleset choice to even start playing with the thingy.


yairm210:
Does this still allow the autoupdate and spelling correction check?

SomeTroglodyte:
The auto update isn't changed. The spell checks aren't influenced. But..
- The spelling correction check is new to me, and the messages I've seen so far from it were BS. I haven't yet learned anything about its mechanisms.
- As far as I can tell, they depend on the forOptionsPopup parameter of checkModLinks, and when the complex ruleset check is activated, that param doesn't get passed through. But that's the same with or without this PR. That's what I meant above by "probably by oversight". I was tempted to fix it, but refrained - a matter for a different PR, and I'd like to know which way it was intended first.

yairm210 (approved):

PR #6267, opened 2022-03-04
Sort maps & accelerate playing a newly edited map

SomeTroglodyte:
See also commit messages. Previously, the order was 100% controlled by the underlying libraries/OS stack, that is, more or less random to the user.

### DIFF
--- a/core/src/com/unciv/ui/newgamescreen/MapOptionsTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/MapOptionsTable.kt
@@ -1,10 +1,10 @@
 package com.unciv.ui.newgamescreen
 
-import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/newgamescreen/MapOptionsTable.kt
```diff
@@ -71,6 +71,10 @@ class MapOptionsTable(private val newGameScreen: NewGameScreen): Table() {
             newGameScreen.updateTables()
         }
 
+        // Pre-select custom if any map saved within last 15 minutes
```
yairm210, line 74: NICE QOL improvement!

SomeTroglodyte: Thanks but - see comment `
        // This is a kludge - the better way would be to have a "play this map now" menu button in the editor
        // (which would ideally not even require a save file - which makes implementation non-trivial)` in there...

PR #6268, opened 2022-03-04
Update Polish.properties - translation

FancyDot:
There's not much changed, 'cause it is my first day of translation, but I'll work further on this project if it comes to translation to Polish as my native language:D


yairm210:
merge translations

PR #6271, opened 2022-03-05
Move automated units button was showing when it didn't do anything

SomeTroglodyte:
To repro in current master: Give the first Warrior at turn 0 a multi-turn move order, the "Move automated units" appears and does nothing as the unit already did what it could this turn...

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -791,7 +791,7 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
                 }
 
             !viewingCiv.hasMovedAutomatedUnits && viewingCiv.getCivUnits()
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -791,7 +791,7 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
                 }
 
             !viewingCiv.hasMovedAutomatedUnits && viewingCiv.getCivUnits()
-                .any { it.isMoving() || it.isAutomated() || it.isExploring() } ->
+                .any { it.currentMovement >= 0.1f && (it.isMoving() || it.isAutomated() || it.isExploring()) } ->
```
xlenstra, line 794: I personally prefer `currentMovement > 0` or `currentMovement >= Constants.MinimumMovementEpsilon` here, but I suppose this has been merged for a few days already

PR #6273, opened 2022-03-05
Added deep linked multiplayer games

GGGuenni:
This allows the turn notification to open a multiplayer game directly without having to go through Main Menu->Multiplayer->Multiplayer Game->Join Game.
It also allows the app to open links like https://unciv.app/multiplayer?id=c7b4a08c-1420-4fc3-bfcb-8bc27b0d8d1d to download the specified game and open it. This could be used to make sharing games between players easier. To make it safer for everyone using it I will probably buy the unciv.app domain to redirect it to Github before implementing the share feature in case someone opens such a link in the browser instead of with the app. It's quite cheap (around 1€ per month) so that's not a problem.

[Unciv.com](https://unciv.com/) however :upside_down_face:

### DIFF
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -19,8 +19,9 @@ import com.unciv.ui.audio.MusicMood
 import com.unciv.ui.utils.*
 import com.unciv.ui.worldscreen.PlayerReadyScreen
 import com.unciv.ui.worldscreen.WorldScreen
+import com.unciv.ui.worldscreen.mainmenu.OnlineMultiplayer
+import java.lang.Exception
 import java.util.*
-
 
 class UncivGame(parameters: UncivGameParameters) : Game() {
     // we need this secondary constructor because Java code for iOS can't handle Kotlin lambda parameters
@@ -34,6 +35,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
     val customSaveLocationHelper = parameters.customSaveLocationHelper
     val limitOrientationsHelper = parameters.limitOrientationsHelper
 
+    var deepLinkedMultiplayerGame: String? = null
     lateinit var gameInfo: GameInfo
     fun isGameInfoInitialized() = this::gameInfo.isInitialized
     lateinit var settings: GameSettings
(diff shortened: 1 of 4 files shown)


SomeTroglodyte:
Serial fraudsters like [Mike Mann](https://domaingang.com/tag/mike-mann/) (do *not* visit link without good ad and script blocker) have their advantages - load their page, see what third-party code they're trying to pull, block all of them - safer future browsing.

SomeTroglodyte:
> `import java.lang.Exception`

Don't needlessly bind to java libraries in kotlin code- kotlin has its own Exception, Error, Throwable etc... (and yes for generic catch-alls Throwable may be preferable) And the try-catch itself is repeated, could be a helper function?

> `game: Pair<String, String>`

Since it's passed around a bit, better a `data class MultiplayerGameIdentifier { val name: String, val id: String }` ? Right up to the two Arrays?

core/src/com/unciv/UncivGame.kt
```diff
@@ -121,7 +123,17 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
 
                 if (settings.isFreshlyCreated) {
                     setScreen(LanguagePickerScreen())
-                } else { setScreen(MainMenuScreen()) }
+                } else {
```
yairm210, line 126: Flatten, needlessly nested

core/src/com/unciv/UncivGame.kt
```diff
@@ -159,6 +171,17 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
         super.resume()
         musicController.resume()
         if (!isInitialized) return // The stuff from Create() is still happening, so the main screen will load eventually
+
+        // This is also needed in resume to open links and notifications
+        // correctly when the app was already running. The handling in onCreate
+        // does not seem to be enough
+        if (deepLinkedMultiplayerGame != null) {
```
yairm210, line 178: Since this is called twice how about making it a function, tryLoadDeepLinkedGame() or something?

yairm210:
It's more important for me to merge this into the next version than it is to fix this minor issue though :)

PR #6274, opened 2022-03-05
update Simplified_Chinese translation

heipizhu4:


heipizhu4:
Note: Maybe we can unite the translation of"improvement"

Issue #6145, opened 2022-02-11
Faith Building Summary


itanasi:
Currently it can be difficult to remember what the different Faith buildings do when picking your Follower Belief.

Would be handy to include their production stats so you don't have to check in the Civilopedia

From Discord:
![image](https://user-images.githubusercontent.com/44038014/153543002-ff132f99-daad-4854-a600-2307078323b6.png)

itanasi:
Looking at the code this might be harder than anticipated since the string is using a standardized Unique already. "May buy [] buildings with [] []"

xlenstra:
My first idea would be to also show the effects of each belief at the bottom of the screen, maybe even instead of showing it on the buttons themselves. Then we can use links to the civilopedia, similar to how to tech tree might do it, for people to to find out what stats buildings provide. Still requires closing the civilopedia and reopening the religion founding screen and also loses previously selected beliefs/religion icon/religion name, but should be relatively easy to implement

PR #6287, opened 2022-03-06
Improve handling of Religion PickerScreen

### LINKED ISSUE #6288: Silly mouse hover over TextButton side-effects bug
**Platform**
Mint 20.3 (Ubuntu 20.4 based)

**Version**
Current master

**Describe the bug**
Once the religion picker bottom-right button is active, mouse-over-ing over it not only changes that button's style, but changes all un-resized and un-recolored Label instances on the screen as well to Color.GREY

**Expected behavior**
No side effects when hovering over a Button outside of said Button.

**Screenshots**
<details><summary>Screen</summary>

image
</details>

**Additional context**
The Close button will not quirk out, it's fine.

A possible """fix"" is to replace `String.toLabel()` with `fun String.toLabel() = Label(this.tr(), Label.LabelStyle(BaseScreen.skin[Label.LabelStyle::class.java]))` - this does make the Religion picker labels invulnerable to mouse-over of the `TextButton`. OK - just writing this sparked a much better idea (forcing just the button in question to use a cloned LabelStyle), but the correlations are still ***vexing***.

Even `Label` instances ***not*** built from that function and therefore even in the 'fixed' version carrying no own clone of the `LabelStyle` (e.g. `WrappableLabel`) ***are*** affected by the bug ***and*** the fix. That seems impossible to me.
I've now tried a very long time to get any understanding for this, long debug sessions digging deep into libGdx sources, and couldn't find the culprit. The Color class not being immutable has something to do with this (moronic decision), but I don't understand how exactly.

Any Input on how to actually _understand_ this?

SomeTroglodyte:
<details><summary>Visible changes:</summary>

- Religion icon chooser scrolls horizontally
- Religion icon chooser doesn't get squashed by the middle part when belief lists get filled
- Religion icon chooser shows unavailable religions in dark red
- Belief slot chooser (left side) shows selected Button by disabling it
- Belief slot chooser shows previous choices with a green background
- Belief chooser (right side) shows selected Button by disabling it
- Belief chooser no longer suppresses choices because another Civ has them or because you already chose it (the latter had very quirky effects)
- Beliefs unavailable because another Civ has them are shown dark red
- Beliefs unavailable because you already chose it in another "slot" are shown dark green
- 3 AutoScrollPanes so mouse wheel easier
- All three ScrollPanes scroll to show selected entry
- Now also fixes #6288
</details>

<details><summary>Under the hood:</summary>

- Dropped index-to-type kludge, better store types beforehand
- Use cell defaults instead of repeated formatting
- "OK" calls dispose like "Close" already did
- Removed outdated comment
- Fixed the most misleading variable names
- Trailing spaces away (not only blank lines either)
- Unified common code setupChoosableReligionIcons / setupVisibleReligionIcons
</details>

<details><summary>To-do:</summary>

- Check Pantheon chooser
- Button class ignores setBackground - but I'm sure I've seen (or even perpetrated) Unciv code that calls it (on Table that works fine, copying to the Button subclass - nope).
- Label wrap is not recalculated on window resize.
- Rethink for portrait mode - much better than before this is though.
</details>

<details><summary>Screenshot</summary>

image
</details>


itanasi:
Willing to do help handle #6145 while you're at it?

SomeTroglodyte:
No. That sat at be back of the skull while doing this, but I intentionally excluded it from this PR's focus. Complex enough, and result more playable.

Moving the uniques from the buttons to `PickerScreen.descriptionLabel` would not be hard, but need some testing and fiddling to actually feel better. That predefined Label is just big enough to hold _one_ Belief, so where do you list all of them? Drop template and do a new layout from scratch, maybe tabbed (where's that widget? I think it's public only in that RFT PR - which should get precedence anyway)? - gets more work than this quickly. Also, the Pantheon dialog, while not a hindrance, should get a matching look first.

SomeTroglodyte:
> Willing to do help handle #6145 while you're at it?

OK, maybe I will. Using the BeliefType icons some more and allowing some Pedia links might be nice after all...

PR #6289, opened 2022-03-06
Fix Mughal Fort using UniqueType Stats instead of StatsPerCity

### LINKED ISSUE #6284: "[+1 Gold] <after discovering [Flight]>" applying globally instead of locally
This unique is causing bonuses to apply in all cities instead of just the one city where the Mughal Fort is being built. It's affecting both Android and Windows, on 3.19.13 and 3.19.14.

SomeTroglodyte:
Hopefully fixes #6284. Untested.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Buildings.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Buildings.json
@@ -601,7 +601,7 @@
 		"culture": 2,
 		"hurryCostModifier": 25,
 		"requiredBuilding": "Walls",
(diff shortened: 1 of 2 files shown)


yairm210:
Looks correct

SomeTroglodyte:
Still trying to repro/confirm - turn 72 and still no Flight...

SomeTroglodyte:
Tested - repro confirmed, fix works. 6 Indian cities one without Mughal Fort, Flight discovered. master: All cities list +5 gold for the Mughal; this: Only the cities 'with' list +1 for the Mughal, the new city doesn't.

PR #6296, opened 2022-03-07
Add "Starts with [policy] adopted" unique

HaneulCheong:
This pseudocloses https://github.com/yairm210/Unciv/issues/3242#issuecomment-1051560640

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -96,6 +96,10 @@ object GameStarter {
             addCivStats(gameInfo)
         }
 
+        runAndMeasure("Policies") {
+            addCivPolicies(gameInfo, ruleset)
+        }
+
         if (tileMap.continentSizes.isEmpty())   // Probably saved map without continent data
             runAndMeasure("assignContinents") {
                 tileMap.assignContinents(TileMap.AssignContinentsMode.Ensure)
@@ -170,6 +174,25 @@ object GameStarter {
                     civInfo.tech.addTechnology(tech.name)
 
             civInfo.popupAlerts.clear() // Since adding technologies generates popups...
+        }
+    }
+
+    private fun addCivPolicies(gameInfo: GameInfo, ruleset: Ruleset) {
+        for (civInfo in gameInfo.civilizations.filter { !it.isBarbarian() }) {
+
+            // generic start with policy unique
+            for (unique in civInfo.getMatchingUniques(UniqueType.StartsWithPolicy)) {
+                // get the parameter from the unique
+                val policyName = unique.params[0]
+
+                // check if the policy is in the ruleset and not already adopted
+                if (ruleset.policies.containsKey(policyName) && !civInfo.policies.isAdopted(policyName)) {
+                    civInfo.policies.freePolicies++
+                    val policyToAdopt = ruleset.policies[policyName]
+                    if (policyToAdopt != null)
+                        civInfo.policies.adopt(policyToAdopt)
+                }
+            }
         }
     }
 
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -187,10 +187,11 @@ object GameStarter {
 
                 // check if the policy is in the ruleset and not already adopted
                 if (ruleset.policies.containsKey(policyName) && !civInfo.policies.isAdopted(policyName)) {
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
Looks perfectly workable but...

- `// get the parameter from the unique` is a bit too 'repeating the obvious' - you know the special superpower of that primitive race way off in the unfashionable side arm of the galaxy...
- going through `civInfo.policies.adopt` will trigger global alerts if the policy has them - I assume you know that and the effect is intentional
- containsKey then [] != null is redundant... Well - a `val policyToAdopt = ruleset.policies[policyName] ?: continue` right after `val policyName =` would do it and make policyToAdopt a non-nullable type. The rest could then reduce to `civInfo.policies.run { if !isAdopted(policyName) { freePolicies++; adopt(policyToAdopt) } }`. Not saying that would be best, just compare the possibilities and decide for yourself. (semicolon is definitely not good practice, that was just to make it a one-liner)

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -173,6 +177,25 @@ object GameStarter {
...
+            // generic start with policy unique
+            for (unique in civInfo.getMatchingUniques(UniqueType.StartsWithPolicy)) {
+                // get the parameter from the unique
+                val policyName = unique.params[0]
+
+                // check if the policy is in the ruleset and not already adopted
+                if (ruleset.policies.containsKey(policyName) && !civInfo.policies.isAdopted(policyName)) {
+                    civInfo.policies.freePolicies++
+                    val policyToAdopt = ruleset.policies[policyName]
+                    if (policyToAdopt != null)
```
yairm210, line 192: This if will always be false, stove we already checked it exists in the previous if

yairm210: Always true, sorry
That is to say we can remove the check

SomeTroglodyte: At that point kotlin _will_ infer policyToAdopt as Policy? and thus insist on something...

HaneulCheong: I was trying to avoid nullable type mismatch back then when I wrote that... Yeah perhaps I'll just change it to something simpler, such as `val policyToAdopt = ruleset.policies[policyName]!!`.

SomeTroglodyte:
> altered version

Hah! Talk about me trying to see the bigger picture and get others to do the same, and here you did look over the garden fence and I didn't. Well done.

> silence any alert

Nah. If we don't already know mad modders will want that, keep it simple and maybe document.

HaneulCheong:
I should say I'd like to leave the comment in... It certainly helped me understand the code quickly without delving into more references.

SomeTroglodyte (approved):
Nice `above.py`...

I hope that `run{}` works, came from a brain short-circuit. We mostly do `apply`. OK, yes, run is `this`-based and just discards any return instead of passing it out.

My first comment was meant in part to emphasize the synergy between kotlin Elvis, kotlin type inference, and kotlin's collection indexed access typically not throwing on a key that's not in the collection, but returning null instead. Which is why checking key presence, then doing an indexed access, then testing on null is twice redundant - two lookups, two checks where one each is enough. And the Elvis was born to rock - no to handle exactly this situation. Try that `val policyToAdopt = ruleset.policies[policyName] ?: continue` before the `if` and see how Studio's hover tells you it's now a `Policy` instead of a `Policy?` - no key presence check needed, it's already done. Sure, this works just as well, and some prefer it. Me, I'm happy about any `!!` avoided, personal taste.

Again, meant as learning nudge, not to imply any lesser quality.

HaneulCheong:
@SomeTroglodyte Thanks, tho I might have to update it since I stopped learning Rust for now...

I remember reading "Don't use .let{} and ?:" from the project's [Coding standards](https://yairm210.github.io/Unciv/Developers/Coding-standards/#dont-use-let-and) so that's why I used the `!!` operator. But I should also admit that I am currently a Kotlin beginner. Would `?:` be more suitable here instead of `!!`?

SomeTroglodyte:
Maybe we should update that coding standards entry... Yea, one can easily go way over the top with Elvises and go way cryptic, but the key in that paragraph is "readability" - and properly used, an Elvis or five can improve readability vastly. If you look up early discussions between me and Yair, you'll find plenty about me exaggerating Elvises or other clever constructs.

It all depends, as always.

PR #6297, opened 2022-03-07
Replace YesNoButton with Checkbox

lishaoxia1985:
YueR


SomeTroglodyte:
Why not... I doubt a tiny bit, wrappability for narrow screens seems lost, and why that colspan(2)

PR #6303, opened 2022-03-08
Religion overview improved

SomeTroglodyte:
Depends on #6302, thus draft for now. Moves text from left to top so the whole thing practically never needs horizontal scrolling, and top-aligns the page so selecting religions doesn't make stuff jump around vertically.


SomeTroglodyte:
Merged and some last-second tweaks. One idea ignored - for the selection to stay on resize the container needs a minor update, and since it's in need of a major one anyway - postponed. German some minor tweaks 'cuz I also tested translatability.

<details><summary>screenie</summary>

image
</details>

PR #6304, opened 2022-03-08
Unique units abilities that should be inherited by upgrades

### LINKED ISSUE #6281: Mohawk Warrior should keep its Forest+J bonus on upgrade
[This link](https://steamcommunity.com/sharedfiles/filedetails/?id=233028579) suggests that the Mohawk Warrior's +33% in Jungle/Forest should be implemented as Promotion, not direct Unique. It also suggests the "does not need Iron" unique should have a quirk - namely you can't upgrade Warriors to it with zero Iron though it won't be consumed. Haven't tested Unciv's behaviour on the latter.

SomeTroglodyte:
Fixes #6281

image

To do: 
- ~~Verify BackwardCompatibility guaranteeUnitPromotions catches these~~ tested successfully
- Maintain table in issue, did we catch all correctly (e.g. some +sight boni to Knight alternates sound like they should inherit too)?
- ~~Edit back units that already had their boni inheritable to use the no promotion icon needed ability???? Pictish Warrior, Slinger?~~ No

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/UnitPromotions.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/UnitPromotions.json
@@ -547,5 +547,54 @@
 	{
 		"name": "Home Sweet Home", // only for Mehal Sefari and subsequent upgrades
 		"uniques": ["[+30]% Strength decreasing with distance from the capital"]
+	},
+	{
+		"name": "Mohawk Warrior ability",
+		"uniques": ["[+33]% Strength <when fighting in [Forest] tiles>", "[+33]% Strength <when fighting in [Jungle] tiles>"]
+	},
+	{
+		"name": "Jaguar ability",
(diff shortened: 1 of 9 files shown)


yairm210:
"[unitName] ability" for single translation

yairm210:
Also free promotion in Civilopedia should show promotion's effects, but that's probably a different PR

yairm210:
Instead of "[unitName] ability" perhaps "[unitName] unique promotion", YMMV

yairm210:
Otherwise no complaints :)

SomeTroglodyte:
> "[unitName] ability" for single translation

:face_with_spiral_eyes: I knew I forgot an important aspect...

> "[unitName] unique promotion"

I see why, but on the other side - length and fixed Pedia geometry on the left side; and I think "ability" just may have some existing association for veteran gamers?

You happen to know quickly where the documentation would need to be updated - the bit that promotions can pull the unit's icon if they have that format?

yairm210:
I'm not sure we have a ready-made place that fits

PR #1456, opened 2019-12-14
Natural Wonders 

r3versi:
* Created models/ruleset/tile/NaturalWonder and assets/json/NaturalWonders
* MapGenerator now spawns NaturalWonders in random compatible random locations
* MapParameters has a noNaturalWonders option
* TileInfo has naturalWonder property (set to null if none)
* TileInfo and  UnitMovementAlgorithms correctly manages stats and consider the tile unpassable and unimprovable

TODO list:
- [x] Update MapGenerator to spawn wonders in suitable locations 
- [x] El Dorado grants 500 gold to the first civilization to discover it
- [x] Fountain of Youth bonus: adjacent military land units heal at double rate for the rest of the game
- [x] Spanish bonus 
- [x] Great Barrier Reef spawns on two adjacent tiles
- [x] Grant +1 permanent happiness on discovery
- [x] Show Natural Wonder icon (maybe Mount Fuji like?)  #124
- [x] Images  #124
- [x] Issue notification on discovery

PR #6309, opened 2022-03-08
Update ModConstants and start its documentation

SomeTroglodyte:
Wiki was still ignorant of ModOptions, and I found a few map generation constants that might be useful there.

### DIFF
--- a/core/src/com/unciv/models/ModConstants.kt
+++ b/core/src/com/unciv/models/ModConstants.kt
@@ -25,11 +25,25 @@ class ModConstants {
     // unitSupplyBase and unitSupplyPerCity can be found in difficulties.json
     // unitSupplyBase, unitSupplyPerCity and unitSupplyPerPopulation can also be increased through uniques
     val unitSupplyPerPopulation = 0.5
-    
+
     // The minimal distance that must be between any two cities, not counting the tiles cities are on
     // The number is the amount of tiles between two cities, not counting the tiles the cities are on.
     // e.g. "C__C", where "C" is a tile with a city and "_" is a tile without a city, has a distance of 2.
     // First constant is for cities on the same landmass, the second is for cities on different continents.
     val minimalCityDistance = 3
     val minimalCityDistanceOnDifferentContinents = 2
+
+    // NaturalWonderGenerator uses these to determine the number of Natural Wonders to spawn for a given map size
+    val naturalWonderCountMultiplier = 0.13133208f
+    val naturalWonderCountAddedConstant = 0.56128831f
+    // MapGenerator.spreadAncientRuins: number of ruins = suitable tile count * this
+    val ancientRuinCountMultiplier = 0.02f
+    // MapGenerator.spawnIce: spawn Ice where T < this, with T calculated from temperatureExtremeness, latitude and perlin noise. 
+    val spawnIceBelowTemperature = -0.8f
+    // MapGenerator.spawnLakesAndCoasts: Water bodies up to this tile count become Lakes
+    val maxLakeSize = 10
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/models/ModConstants.kt
```diff
@@ -25,11 +25,25 @@ class ModConstants {
...
-}
\ No newline at end of file
+
+    // NaturalWonderGenerator uses these to determine the number of Natural Wonders to spawn for a given map size
+    val naturalWonderCountMultiplier = 0.13133208f
+    val naturalWonderCountAddedConstant = 0.56128831f
+    // MapGenerator.spreadAncientRuins: number of ruins = suitable tile count * this
+    val ancientRuinCountMultiplier = 0.02f
+    // MapGenerator.spawnIce: spawn Ice where T < this, with T calculated from temperatureExtremeness, latitude and perlin noise. 
+    val spawnIceBelowTemperature = -0.8f
```
xlenstra, line 42: To me this looks like it should rather become the temperatur-Humidit-Rang-Unique applied on the ice feature?

yairm210: Agree

core/src/com/unciv/models/ModConstants.kt
```diff
@@ -25,11 +25,25 @@ class ModConstants {
...
     // e.g. "C__C", where "C" is a tile with a city and "_" is a tile without a city, has a distance of 2.
     // First constant is for cities on the same landmass, the second is for cities on different continents.
     val minimalCityDistance = 3
     val minimalCityDistanceOnDifferentContinents = 2
-}
\ No newline at end of file
+
+    // NaturalWonderGenerator uses these to determine the number of Natural Wonders to spawn for a given map size
+    val naturalWonderCountMultiplier = 0.13133208f
+    val naturalWonderCountAddedConstant = 0.56128831f
```
yairm210, line 38: These are very long floats, where did they come from? Can't they be simplified to 0.13 and 0.56?

SomeTroglodyte:
I just _moved_ those numbers, they were already there... OK, ~~there's a comment containing links that should be moved together with the nat won constants,~~ (no the linked posts are more related to stuff in the jsons) - giving more digits than a Float can hold is silly. Not upgrading to a Unique - I tried to keep this small, but I agree wholeheartedly on the Ice thingy. I'll take it out of this PR, to be done separately.

SomeTroglodyte:
OK, looks like the default constants are wrong anyway. Since [this commit](https://github.com/yairm210/Unciv/commit/60aeebd3bbeff013320345eb914c8b29d03127dd) disrupts the history, I can't even find who established them - no, we can, it was #1456, [last commit](https://github.com/yairm210/Unciv/pull/1456/commits/de429cfbe492524b959614b3b38e79265f27b773), by @r3versi.
If our predefined sizes are OK (given I trust Federico, not quite unlikely), then the formula gives 5 NW for Huge while the linked sources say 7.

SomeTroglodyte:
OK, our predefined map sizes are much smaller than original ones. With original dimensions the formula is very close to correct but not quite. naturalWonderCountMultiplier = 0.122 and naturalWonderCountAddedConstant = 0f would hit that mark exactly. I assume the author tweaked a bit to avoid a sudden jump (small->2, medium->2, large->4)...

So - do we patch the predefined sizes to the original's sizes? If yes, change constants to those above. If no, keep & document - or move the responsibility to MapSize classes, "interpolate" for Custom, and use formula only for Custom > Huge?

SomeTroglodyte:
OK, that's the minimal approach using the stuff just learned.

PR #6327, opened 2022-03-10
Global review of Constants use, linting, maxXPfromBarbarians

SomeTroglodyte:
Looks big, but is almost without functional impact.

The one change means the deprecated maxXPfromBarbarians is now correctly transferred to the ModConstants one, and even written that way after Mod download.

The rest is Constants:
- "Coast" and "Coastal" are const with Kdoc so the functional difference is clear, and used throughout.
- Same for "Fresh Water" and "Fresh water" - I hope the difference in const name makes distinction easier to understand. You might have different ideas - those would be welcome.
- "Custom" has two different meanings, and I moved the Constants definition to make the distinction un-missable.
- All other existing checked for missed uses. Strings with different semantics - e.g. "Research Agreement" as key vs. display only on a button or modifier list - those I left alone.

... and one or two squiggly-line-inspections obeyed
... and a lot of trailing whitespace linting in files I touched anyway.


SomeTroglodyte:
Sorry not doing merges myself, thanks for your extra work - was having fun in knee-deep snow without snowshoes all day.

Issue #6333, opened 2022-03-11
Spectator / Id from clipboard


HistericalCat:
**Platform**
Android, Redmi 8A, 64 bit.

**Version**
PlayStore - V3.19.12

**Describe the bug**
Playing the game with you as a spectator and pressing next turn, it will continuously say waiting for spectator.
When choosing Id for human, doing Id from clipboard and creating the game will have you in a spectator mode where you can the entire map but can't do anything

**To Reproduce**
1. Go to create game
2. Click on make human spectator
3. Create game
4. Click next turn
5. See bug

Second bug
1. Click to create game
2. Click for human take Id from clipboard
3. Start game
4. Click next turn
5. See bug

If game data is required, please add it here. From the menu, choose save game (or map), copy to clipboard, and paste here between the backtick blocks (or, if no such save is required, you may delete this):
<details><summary>Saved game</summary>

```

```
</details>

**Expected behavior**

**Screenshots**

**Additional context**

SomeTroglodyte:
Insufficient Data. From your steps, in 1 you would be starting a non-onlinemultiplayer game with 1 human spectator and N AI players with the G&K base ruleset and zero mods. Nope, works just fine.

Actually, you're running a 1-Spectator, N-AI game, and when the spectator's player ID is your own, you simply relinquish control until one of the AI's wins. That's right, "continuously" is not infinite. When the ID is not yours, it needs to check whether someone over the wire updates the game... Which can't happen unless the owner of that ID loads it and runs the mentioned simulation-until-AI-wins.

You tell a mad politician to jump off a cliff - and if they obey, you cry "they have a bug"! I would rejoice instead... Sorry.

So, everyone -
- should we block the Start Game button when no player fulfills `it.playerType == PlayerType.Human && !(it.chosenCiv == Constants.spectator && gameSetupInfo.gameParameters.isOnlineMultiplayer && it.playerId != UncivGame.Current.settings.userId)` ?
- should we reconsider GameInfo.226 (don't stop if `gameParameters.isOnlineMultiplayer && (thisPlayer.isDefeated() || thisPlayer.isSpectator())` )?

SomeTroglodyte:
I'll close this as the 6337 PR should have improved things, 8 days is enough for the issue OP to check and come back...

PR #6337, opened 2022-03-11
Fix 1 Spectator N AI games in seemingly endless loop

SomeTroglodyte:
Coarse guess for #6333 ... Base for discussion, nothing more.


xlenstra (approved):

yairm210:
Looks like a good change in any case

yairm210 (approved):

SomeTroglodyte:
> good change in any case

Wasn't too sure about that as my gut feeling says 'you really don't know enough to see the entire thing' - but OK then I'll 'ready' it

PR #6343, opened 2022-03-12
Add Filipino.png since the Translation file is already there

SomeTroglodyte:
No credits entry necessary, it's from [the same source](https://www.flaticon.com/premium-icon/philippines_4855761), even if that site is a royal spam PITA.


yairm210:
If you find a good other source for flags we can change them all, I'm not particularly attached

PR #6345, opened 2022-03-12
Enable ModOptions uniques and ModConstants from non-base mods

SomeTroglodyte:
Currently, ModOptions.uniques and ModConstants do work in base ruleset mods, but not in add-on mods. This fixes that, so one mod could do maxXPfromBarbarians = 10 and another could do minimalCityDistance = 2 and both would be respected.


SomeTroglodyte:
P.S.: GameInfo.setTransients calling RulesetCache.getComplexRuleset on each turn seems like a mighty waste (or lots of unnecessary GC stress potential). Why not keep the ruleset instance when cloning and do it in GameInfo.setTransients only when uninitialized?

core/src/com/unciv/models/ModConstants.kt
```diff
@@ -11,25 +12,42 @@ class ModConstants {
...
+        if (other.cityStrengthFromTechsMultiplier != defaults.cityStrengthFromTechsMultiplier) cityStrengthFromTechsMultiplier = other.cityStrengthFromTechsMultiplier
+        if (other.cityStrengthFromTechsExponent != defaults.cityStrengthFromTechsExponent) cityStrengthFromTechsExponent = other.cityStrengthFromTechsExponent
+        if (other.cityStrengthFromTechsFullMultiplier != defaults.cityStrengthFromTechsFullMultiplier) cityStrengthFromTechsFullMultiplier = other.cityStrengthFromTechsFullMultiplier
+        if (other.cityStrengthFromGarrison != defaults.cityStrengthFromGarrison) cityStrengthFromGarrison = other.cityStrengthFromGarrison
+        if (other.unitSupplyPerPopulation != defaults.unitSupplyPerPopulation) unitSupplyPerPopulation = other.unitSupplyPerPopulation
+        if (other.minimalCityDistance != defaults.minimalCityDistance) minimalCityDistance = other.minimalCityDistance
+        if (other.minimalCityDistanceOnDifferentContinents != defaults.minimalCityDistanceOnDifferentContinents) minimalCityDistanceOnDifferentContinents = other.minimalCityDistanceOnDifferentContinents
+    }
```
xlenstra, line 48: Is this really the best approach? The very-late end goal of the constants object, is to contain almost every single magic number in the entire game, so having to copy all of them over manually sounds like it will use an enormous amount of space.

Additionally, does it even make sense to allow non-baseruleset mods to change this? What if two non-baseruleset mods change the same constant, which one would the correct one be?

SomeTroglodyte: Exactly! I had a variant making the class immutable, nice and allows a shorter merge with a helper function that also decides "which one would the correct one be" - but you'd still need a decision per member where I see no prepared shortcut short of reflection. The immutable variant was about 2 lines longer in the end so I ditched it. It involved using the `copy`offered by data classes to e.g. fix in the deprecated maxBarbXp. But you had one central place that could decide first, last or even average...

It all comes down to - if anyone simply publishes a mod that aims to change one of the values - and I can imagine changing that barbarians thingy to 0 or very large being popular - do we want them to work or not?

If not, we should warn!

PR #6346, opened 2022-03-13
Resolved #6335 - now shows conditionals precluding construction of units

### LINKED ISSUE #6335: Unable to build/buy settlers
**Platform**
Android phone, 8.1.0 version

**Version**
Google play store,  3.19.15 game version

**Describe the bug**
Settler is no longer showing up in units I can build/buy.

**To Reproduce**
Idk how, just noticed it when I was in the modern era trying to get more oil and the AI is still producing settlers when I can't

```

```
</details>

**Expected behavior**
Settlers should show up in the unit list to be bought/built.

**Screenshots**

**Additional context**

yairm210:
Built simply so that if we decide to expand the conditionals that should be shown it will simply be building a hashset of conditionalsToShow


core/src/com/unciv/logic/city/IConstruction.kt
```diff
@@ -129,7 +129,7 @@ class RejectionReasons: HashSet<RejectionReason>() {
 
 // TODO: Put a wrapper class around this containing the errorMessage, so that we don't
 // change the value of a enum constant sometimes.
-enum class RejectionReason(val shouldShow: Boolean, var errorMessage: String) {
+enum class RejectionReason(var shouldShow: Boolean, var errorMessage: String) {
```
xlenstra, line 132: I would like to note that changing the `shouldShow` variable for one instance of the enum, changes it for all instances iirc. I.e., every unit with the `CannotBuildUnits` unique will be shown, with the same error message as the one that is blocked by having to little happiness. This is probably not expected behaviour.

yairm210: Oh
That makes sense in retrospect but is not what I want

xlenstra (approved):

SomeTroglodyte:
So you dislike `in`? Will have to remember. But then why not `it in this` -> `contains(it)` instead of -> `this.contains(it)`?

core/src/com/unciv/logic/city/IConstruction.kt
SomeTroglodyte: I think I warned long ago that overwriting vars on enums would come back to haunt us - it's not 'all instances' but 'a singleton per enum entry'.

yairm210:
> So you dislike `in`? Will have to remember. But then why not `it in this` -> `contains(it)` instead of -> `this.contains(it)`?

`it in this` would not compile in this case
The this.contains(it) is indeed unnecessary

SomeTroglodyte:
Was talking about fun `getMostImportantRejectionReason` - and yes `it in this` compiles within any it-defining lambda hosted in a block whose receiver has a contains method - `in` _is_ a `contains` synonym. All about readabilty - and by the way IMHO explicitly stating a redundant `this` _can_ in some cases improve readability.

PR #6354, opened 2022-03-15
Double Zoom Out Range

### LINKED ISSUE #6294: Further Zoom
Seems a very popular request to enable further zoom out in both game and Map editor.

Ideally after a certain point unit icons can not be rendered to prevent clutter.

Quick messing with `ZoomableScrollPane` zoom limits doesn't seem to work. Lowering limit lower than 0.5 causes the world screen to also shrink
image

itanasi:
Also increase the padding so that center of screen ends up at map edge.
Fixes #6294


SomeTroglodyte:
Any measurements on memory impact / performance?

yairm210 (approved):

SomeTroglodyte:
Pity I didn't pull and test this before:
image
Note the "camera" rectangle on the minimap corresponds to the screen's borders, not the visible portion, and it overlaps itself - so that is as should be, only the map is clipped to half of what we want to see - so it must be something simple?

Looking in the debugger, scaleX/Y is 0.26-something, world coords screen height is 800, the ScrollPane(=WorldMapHolder)'s height is 1600, and the widgetCullingArea it passed to the TileGroupMap is [1869.0,1507.0,2695.9707,1600.0] - so yes what we see is clipped to half the actual screen. But why doesn't it fill the stage? Oh my.

itanasi:
This is with the Xwrap on? Curses, I didn't test that part

SomeTroglodyte:
> [Curses](https://en.wikipedia.org/wiki/Curses_(programming_library))

So you're an old-school Unixer? Kudos! :balloon:

PR #6368, opened 2022-03-18
Empire Overview Units redo

SomeTroglodyte:
* Fixed header
* Clicking a unit button now _selects_ garrisoned units, even aircraft. Planes in Carriers so far untested.
* Page maintains vertical scroll from one invocation to next - quite helpful IMHO for the units, and those only. Weaknesses: Does not reset on loading a different game, or follow changes in unit order due to upgrades and the like.
* Unit Supply is an Expander - open and Alarm-Iconed only when in deficit - in the fixed header, not pushing the list to the right.
* "Next Unit" candidates (due and idle) have their name in white, others in light gray
* Closest City clickable, and tinted light green if the unit is actually garrisoned there
* Can promote star dimmed when forbidden due to no movement left or attacked
* Upgrade separate column, dimmed if action not possible at the moment, clickable otherwise, no "you sure you wanna spend the moolah" question - that's responsibility of the UnitAction

**Question**: Drop the health column as we also have the "bar" on the button?
**Question**: Ability to promote is indicated with a star - same as Quick Study promotion, but Civilopedia uses the Mobility icon as category decoration... Unify?

image
(sorry no promotable units in this test game)
image
(this one has - dimmed because they already moved)
</details>

### DIFF
--- a/core/src/com/unciv/ui/overviewscreen/EmpireOverviewCategories.kt
+++ b/core/src/com/unciv/ui/overviewscreen/EmpireOverviewCategories.kt
@@ -34,8 +34,8 @@ enum class EmpireOverviewCategories(
                 = TradesOverviewTab(viewingPlayer, overviewScreen),
         fun (viewingPlayer: CivilizationInfo) = viewingPlayer.diplomacy.values.all { it.trades.isEmpty() }.toState()),
     Units("OtherIcons/Shield", 'U',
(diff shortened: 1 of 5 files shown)


yairm210:
Good point, health seems unnecessary to me as well

SomeTroglodyte:
Only the number is a tad more exact than a health bar... Well, you can always select it and see the number on the WorldScreen's UnitTable. Later.

yairm210 (approved):

PR #6372, opened 2022-03-18
Android TV D-pad controls

Pitel:
#1236

Currently just proof-of-concept.

- Cursor/circle is displayed
- Cursor can be moved
- You can click

## TODO

(In no particular order.)

- [ ] Adjustable sensitivity/increase speed when holding
- [ ] Dialog warning that this sucks and you should get a proper mouse or something
- [ ] Dragging/moving map

### DIFF
--- a/android/src/com/unciv/app/AndroidLauncher.kt
+++ b/android/src/com/unciv/app/AndroidLauncher.kt
@@ -4,8 +4,11 @@ import android.content.Intent
 import android.content.pm.ActivityInfo
 import android.net.Uri
 import android.os.Bundle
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
Hey, welcome! Looks quite plausible and well done. I would probably have marshalled the CursorView.cursor.x/y properties to the CursorView level including the invalidate() in the setter; and not used a lazy there, but - hey.

Idea: Dragging might be harder to do than panning when the cursor tries to leave the worldscreen view (the cursor not leaving the App area is done by android.view.View? Or is that missing?)...

PR #6375, opened 2022-03-18
Empire Overview Diplomacy

SomeTroglodyte:
Looking at the next Empire Overview Tab, Diplomacy, the foremost thing to improve was making it usable in portrait. To get there I needed to understand it and then ended up reorganizing to the point of leaving little in place. But that affects the Tab only, the rest this PR touches is easier to see, and I separated that into commits.

#### EmpireOverviewScreen, DiplomacyOverviewTab:
* Civs now sorted instead of json order, unified with DiplomacyScreen sorter
* Line colour = Relation colour as used on the Diplomacy screen. I'm very unsure about this. Add a legend, or a tooltip per line? The current Tooltip system would need a new listener to be able to recognize line mouseovers...
* Civs icons in the polyhedron show the civ name as tooltip
* Portrait mode moves the list down under the polyhedron, gives it the ability to use more columns, and switches scroll responsibility from the list to the entire tab
<details><summary>Portrait Screenie</summary>

image
</details>

#### DiplomacyScreen
* Links into the DiplomacyScreen now scroll the "selected" civ into the center on the left side. Could still use some clearer visual marker which nation is the selected one, but I haven't a good idea what would look both clear and unobtrusive. Works for the links from DiplomacyOverviewTab, but also counteroffers to trades or any other way into the DiplomacyScreen with context.
* "Selected" civ icon still not _kept_ in view on resize - since we're not recreating on resize. Still small price to pay for resizes not disrupting trades and so on.
* But this _does_ fix DiplomacyScreen Close button wandering off on resizes.

#### Left undone
* Make it clearer those numbers are the Score?
* ~~Civ list has a high chance to not mix major and minor civs in a row, but forgot to write the code line to ensure.~~

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -283,15 +283,15 @@ class GameInfo {
                 }
         )
     }
-    
+
     private fun checkForTimeVictory() {
         if (turns != gameParameters.maxTurns || !gameParameters.victoryTypes.contains(VictoryType.Time)) return
-        
+
         val winningCiv = civilizations
             .filter { it.isMajorCiv() && !it.isSpectator() && !it.isBarbarian() }
-            .maxByOrNull { it.calculateScoreBreakdown().values.sum() } 
+            .maxByOrNull { it.calculateTotalScore() } 
             ?: return // Are there no civs left?
-        
+
         winningCiv.victoryManager.hasWonTimeVictory = true
     }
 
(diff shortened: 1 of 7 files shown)


SomeTroglodyte:
FYI: There's a new bug where it is possible that the TabbedPager widget miscalculates dimensions - symptom e.g. the city table can get wider than the screen but the ScrollPane won't scroll that (because parts were pushed _out_ of the screen)... The fix is in the works, I'll do that in a cleanup PR once the last "page rework" is in.

Issue #6376, opened 2022-03-19
Custom desktop font?


HAHH9527:
My friend complained to me that the font was not clear.
![image](https://user-images.githubusercontent.com/27153191/159105198-e1de4a3b-5e71-4a50-81ca-b830c7babf2e.png)
![image](https://user-images.githubusercontent.com/27153191/159105212-4ae0df8e-274c-420d-bf48-ea7636c06fd2.png)

So we can add a custom font setting in desktop.
Even if this setting is not displayed.

One of my idea:
```patch
Index: desktop/src/com/unciv/app/desktop/NativeFontDesktop.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/desktop/src/com/unciv/app/desktop/NativeFontDesktop.kt b/desktop/src/com/unciv/app/desktop/NativeFontDesktop.kt
--- a/desktop/src/com/unciv/app/desktop/NativeFontDesktop.kt	(revision fef993d5dca34f871f7a45a585cae7a73be77d6b)
+++ b/desktop/src/com/unciv/app/desktop/NativeFontDesktop.kt	(date 1647660638627)
@@ -1,6 +1,10 @@
 package com.unciv.app.desktop
 
+import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.graphics.Pixmap
+import com.unciv.JsonParser
+import com.unciv.logic.GameSaver
+import com.unciv.models.metadata.GameSettings
 import com.unciv.ui.utils.NativeFontImplementation
 import java.awt.*
 import java.awt.image.BufferedImage
@@ -8,7 +12,11 @@
 
 class NativeFontDesktop(private val size: Int) : NativeFontImplementation {
     private val font by lazy {
-        Font("", Font.PLAIN, size)
+        val settings = JsonParser().getFromJson(
+            GameSettings::class.java,
+            FileHandle(GameSaver.settingsFileName)
+        )
+        Font(settings.desktopFontFamily, Font.PLAIN, size)
     }
     private val metric by lazy {
         val bi = BufferedImage(1, 1, BufferedImage.TYPE_4BYTE_ABGR)
Index: core/src/com/unciv/models/metadata/GameSettings.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
[...]

PR #6377, opened 2022-03-19
Custom desktop font

HAHH9527:
#6376

### DIFF
--- a/core/src/com/unciv/ui/utils/Fonts.kt
+++ b/core/src/com/unciv/ui/utils/Fonts.kt
@@ -18,7 +18,7 @@ import com.unciv.models.stats.Stat
 interface NativeFontImplementation {
     fun getFontSize(): Int
     fun getCharPixmap(char: Char): Pixmap
-    open fun getAvailableFont(): Collection<FontData> = emptyList()
+    fun getAvailableFont(): Collection<FontData>
 }
 
 // If save in `GameSettings` need use enName.
--- a/core/src/com/unciv/ui/utils/Fonts.kt
+++ b/core/src/com/unciv/ui/utils/Fonts.kt
@@ -24,7 +24,7 @@ interface NativeFontImplementation {
 // If save in `GameSettings` need use enName.
 // If show to user need use localName.
 // If save localName in `GameSettings` may generate garbled characters by encoding.
-data class FontData(val localName: String, val enName: String) {
+data class FontData(val localName: String, val enName: String = localName) {
     override fun equals(other: Any?): Boolean {
         return if (other is FontData) enName == other.enName
         else super.equals(other)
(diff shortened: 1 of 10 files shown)


SomeTroglodyte:
Could that also work on Android?
Also, I'd not read the settings here, I'd pass them down from DesktopLauncher.main where the file is read anyway.

HAHH9527:
@SomeTroglodyte 

> Could that also work on Android?

I think this setting only run with desktop.
Because desktop default fonts are easily unclear in non-English.

>Also, I'd not read the settings here, I'd pass them down from DesktopLauncher.main where the file is read anyway.

Your mean is this?
```kotlin
class NativeFontDesktop(private val size: Int) : NativeFontImplementation {
    private val font by lazy {
        val settings = UncivGame.Current.settings
        Font(settings.desktopFontFamily, Font.PLAIN, size)
    }
...
```
or this?
```kotlin
class NativeFontDesktop(private val size: Int, private val settings: GameSettings) : NativeFontImplementation {
    private val font by lazy {
        Font(settings.desktopFontFamily, Font.PLAIN, size)
    }
...
```

HAHH9527:
And this code only run in `desktop` module.
I didn't change other modules.

So if you don't want change setting page, We can use this setting as a hidden option.
We just need to inform the player in the docs.

Or we can add an text prompt in setting page.

HAHH9527:
@SomeTroglodyte
Look the new commit.
We can use `getDesktopAllFonts()` to show setting.
If the list is empty, the setting does not require a custom desktop font to be displayed.

core/src/com/unciv/ui/utils/Fonts.kt
```diff
@@ -18,8 +18,14 @@ import com.unciv.models.stats.Stat
 interface NativeFontImplementation {
     fun getFontSize(): Int
     fun getCharPixmap(char: Char): Pixmap
+    open fun getDesktopAllFonts(): List<DesktopFont> = emptyList()
 }
 
+// If save in `GameSettings` need use enName.
+// If show to user need use localName.
+// If save localName in `GameSettings` may generate garbled characters by encoding.
+data class DesktopFont(val localName: String, val enName: String)
```
HAHH9527, line 27: @SomeTroglodyte 
`localName` is name of language used by the user. But that maybe not ASCII code.
If use this name save in setting, this value will become garbled.

`enName` is ASCII code. But users may not necessarily recognize what this is.
Like `黑体` 's `enName` is `SimHei`.
`微软雅黑` is `Microsoft YaHei`

yairm210 (approved):

SomeTroglodyte (approved):
That `open fun` in an interface (redundant) isn't mine, is it? Otherwise, of course I approve.

SomeTroglodyte:
People are perhaps going to notice the extra delay opening OptionsPopup on >= Q, though. 4s on a slow AVD.

PR #6379, opened 2022-03-19
Unified & generalized a few uniques

xlenstra:


SomeTroglodyte (approved):
Cursory inspection:
- ["[+1 Happiness] <in cities with at least [6] [Population]"] is missing a ">"
- More not less trailing blanks
- Are those getPopulationFilterAmount keys translatable 'cuz the automatic picks them up or should they get templates?

Other than that: :+1:

yairm210:
Good point, we should add a warning error in checkUniques it the number of < and > are different
(For a different pr obviously)

PR #6380, opened 2022-03-19
Unified two uniques into one more descriptive unique

### LINKED ISSUE #6347: Jungle and forest are resistent and destroyable by nukes at the same time
**Platform**
Android 10

**Version**
Google Play, 3.19.16

**Describe the bug**
Jungle and forest have uniques "Can be destroyed by nukes" and "Resistant to nukes" at the same time

**Expected behavior**
They're resistant or destroyable

**Screenshots**
Screenshot_20220313-130550_UnCiv
Screenshot_20220313-130544_UnCiv

**Additional context**
It looks pretty random which one wins.

### LINKED ISSUE #6355: Fallout can be created on city tiles, but cannot be removed
**Platform**
Android 10

**Version**
3.19.16, Google Play

**Describe the bug**
Nukes can create fallout on city tiles, but it cannot be removed, because it's on city tile.

**Expected behavior**
Fallout isn't created on city tiles. 

**Screenshots**
Screenshot_20220315-123902_UnCiv

xlenstra:
Uniques were for determining when terrain features should be removed from a tile

Edit: closes #6347 
Edit: now also closes #6355 (untested, but it should work)

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
@@ -181,7 +181,7 @@
 		"uniques": ["Rough terrain",
 					"Provides a one-time Production bonus to the closest city when cut down",
 					"Blocks line-of-sight from tiles at same elevation",
-					"Resistant to nukes", "Can be destroyed by nukes",
+					"[25]% Chance to be destroyed by nukes",
 					"A Region is formed with at least [30]% [Forest] tiles, with priority [3]",
 					"A Region is formed with at least [35]% [Forest] tiles and [Jungle] tiles, with priority [3]",
 					"A Region can not contain more [Jungle] tiles than [Forest] tiles",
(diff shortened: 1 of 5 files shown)


yairm210 (approved):

SomeTroglodyte:
There was an issue we could close with this... #6347? And did you look for #6355 while there? The function you touched has no checks for city centers, so the code outside should handle that when it builds `hitTiles` - does it? Don't see it. OK, not this PR.
```
            if (tile.hasUnique(UniqueType.DestroyableByNukesChance)) { ... }
            if (!tile.hasUnique(UniqueType.DestroyableByNukesChance)) return;
```
Strange. Should still make the code for the deprecated unreachable unless the new unique is present too

xlenstra:
> There was an issue we could close with this... #6347? 

Ah, that's were I remembered this from. Linked, thx for reminding.

> And did you look for #6355 while there? The function you touched has no checks for city centers, so the code outside should handle that when it builds `hitTiles` - does it? Don't see it. OK, not this PR.

I'll take a quick look, but that should have been the exact reason the `Indestructible` unique exists iirc.

> 
> ```
>             if (tile.hasUnique(UniqueType.DestroyableByNukesChance)) { ... }
>             if (!tile.hasUnique(UniqueType.DestroyableByNukesChance)) return;
> ```
> 
> Strange. Should still make the code for the deprecated unreachable unless the new unique is present too

Whoops, the second one should have been `DestroyableByNukes`, I'll change that right away.

PR #6381, opened 2022-03-19
Fixed a bug where the resource supply overview would not add up

xlenstra:
This was as when multiple city-states provided a resource, only the count of one of them would be used in the overview.


SomeTroglodyte (approved):
That change to [ResourcesOverviewTable.kt](https://github.com/yairm210/Unciv/pull/6381/files#diff-36fe3f4e4fdd64276af29f24f5504095b39867b9a55d6e9e765458f751e4e882) is redundant - the whole wrap-in-holder at that point is a dead remnant from before the click was changed from display-a-name-label to go-to-worldscreen-and-notify-where-they-are... It's gone in my WIP, which waits on the other EmpireOverview PR's.

Issue #6360, opened 2022-03-17
Improvements to online multiplayer system


GGGuenni:
Ideas to improve the multiplayer. Some things were already planned and pursued in #9497 

- [ ] Resign at any time
- [ ] Multiplayer lobbies
- [ ] Public/Private lobbies with a list for public lobbies to easily join games 
- [ ] Kick players from games
- [ ] invite links
- [ ] Turn notifications per game
- [ ] Option to allow players to join already running games (Take over AI)
- [ ] Move multiplayer game related settings (kick, invite, resign, etc.) to WorldScreenMenuPopup (right side?)
- [ ] Usernames to hide UUIDs
- [x] #8482
- [ ] disable charts for non spectator players until the game ends.

2. Multiplayer lobbies: A lobby before the game starts, where everyone can choose their Civ themselves and only the leader can change game related settings 
3. Public lobbies + list: Is this worth the implementation effort? Would this even be used? Maybe ask in a Poll if people would use this
4. Kick players: Should it be a democratic vote or leader decides? Maybe Poll this

ravignir:
Imo, the most important thing is to implement some anti-cheating features.

GGGuenni:
> Imo, the most important thing is to implement some anti-cheating features.

Sadly not possible as far as I can tell. Players will always have the ability to modify the game files or even create a cheating version of unciv on their own repo. There is not much we can do about that.
But that's one of the reasons I don't know if public lobbies would even be used.

xlenstra:
I don't think we have to make it completely fool-proof, but making it more difficult might already be enough to stop most ways of cheating. Currently, in order to start a game you need to send the host your userID, which the host can then later use at any moment to log into the game and mess up your empire.
One way to prevent this, is by not sharing userID's, but by sharing usernames (as you're already planning) and not making these one-to-one translatable to userID's. For example, adding a file in the dropbox saving usernames, hash-encrypted passwords and password-encrypted userID's. This will also allow users to easily sync multiplayer games across different devices; simply logging in with the same account on a different device automatically gives access to the userID, no copy+pasting between different devices required.
We could then also save the username, password & userId in `gameSettings.json`, so we can automatically try logging in again whenever the multiplayer tab is opened.

This would, of course, still leave open the possibility for people to clone the repo and create a version where all passwords are accepted, but the only way out of that is to run all authentication server-side, which is in the current server setup impossible.

GGGuenni:
[...]

PR #6384, opened 2022-03-20
Unciv server

yairm210:
I think I'm not extracting the contents on home save properly in the server, hence why the files returned are empty
To use: run UncivServer and UncivDesktop together, options-multiplayer, copy ip and paste ip to "server ip"

UI of options looks bad, as can be expected

### DIFF
--- a/desktop/src/com/unciv/app/desktop/UncivServer.kt
+++ b/desktop/src/com/unciv/app/desktop/UncivServer.kt
@@ -1,19 +1,22 @@
 package com.unciv.app.desktop
-import com.badlogic.gdx.Gdx
-import com.badlogic.gdx.Net
 import io.ktor.application.*
 
 import io.ktor.response.*
 import io.ktor.routing.*
 import io.ktor.server.engine.*
 import io.ktor.server.netty.*
+import io.ktor.util.*
 import io.ktor.util.cio.*
 import io.ktor.utils.io.jvm.javaio.*
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import java.io.File
 import kotlin.Exception
 
 fun main() {
     val fileFolderName = "MultiplayerFiles"
+    File(fileFolderName).mkdir()
+    println(File(fileFolderName).absolutePath)
     embeddedServer(Netty, port = 8080) {
         routing {
             get("/isalive") {
@@ -21,14 +24,21 @@ fun main() {
             }
             post("/files/{fileName}"){
                 val fileName = call.parameters["fileName"] ?: throw Exception("No fileName!")
-                val recievedText = call.request.receiveChannel().toInputStream().toString()
-                File(fileFolderName, fileName).writeText(recievedText)
+                withContext(Dispatchers.IO){
+                    val recievedBytes = call.request.receiveChannel().toInputStream().readAllBytes()
+                    val textString = String(recievedBytes)
+                    println("Recieved text: $textString")
+                    File(fileFolderName, fileName).writeText(textString)
+                }
             }
             get("/files/{fileName}"){
                 val fileName = call.parameters["fileName"] ?: throw Exception("No fileName!")
+                println("Get file: $fileName")
                 val file = File(fileFolderName, fileName)
                 if (!file.exists()) throw Exception("File does not exist!")
-                call.respondText(file.readText())
+                val fileText = file.readText()
+                println("Text read: $fileText")
+                call.respondText(fileText)
             }
             delete("/files/{fileName}"){
                 val fileName = call.parameters["fileName"] ?: throw Exception("No fileName!")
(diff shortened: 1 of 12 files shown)


desktop/src/com/unciv/app/desktop/UncivServer.kt
```diff
@@ -0,0 +1,41 @@
...
+fun main() {
+    val fileFolderName = "MultiplayerFiles"
+    embeddedServer(Netty, port = 8080) {
+        routing {
+            get("/isalive") {
+                call.respondText("true")
+            }
+            post("/files/{fileName}"){
+                val fileName = call.parameters["fileName"] ?: throw Exception("No fileName!")
+                val recievedText = call.request.receiveChannel().toInputStream().toString()
```
yairm210, line 24: This line is probably what's causing the problems, you probably need to read it too a buffer or idk

SomeTroglodyte:
Umm, "http POST" and the concept "upload" are not the same... A POST is simply put just a GET with parameters moved from the URL to request headers. ?right?  [this](https://ryanharrison.co.uk/2018/09/20/ktor-file-upload-download.html) looks plausible...

yairm210:
I'm a believer in http verbs having meaning, but in any case this should be a post, since the data is passed by content and not by url
And no, not exactly. Both verbs can accept headers, but headers have a limited size. For large data transfer you need content.

SomeTroglodyte:
I still believe I'm right. POST is like an API call, classically created in conjunction with Client-side html FORM where fields under that form are automatically included in the request, but not via the URL. The URL is the server-side entity that will process said data, e.g. a php/aspx/whatever script updating a database with contact information from that form. If you use POST for uploads, the semantics must be akin to said database example, e.g. the URL is "/fileservice", there's a "gameInfo" field in the POST parameters, then you save that in the filesystem instead of a database - or in sql after all, the client won't know. As I said, service/rpc semantics. Routing "files/<filename>" to your service is strange, as there may be no actual entity at that URL... The semantics you are using are more like those of the PUT verb.

Nevertheless, since routing has the last word, the construct should still work.

maybe `val receivedText = call.receiveText()` as the dox say "To access the raw bits of the payload, you can use receiveChannel, but it is directly part of the call instead of call.request"

SomeTroglodyte:
Congrats :party:, looks like going in the right direction. I assume code cleanup (println, package locality) is planned. You didn't adopt the coroutine approach to the server side, which is OK as the examples yield every 4MB and our files are small -> Idea turn on gzip for them to further move load away from the server?

yairm210:
All files send to/from devices are gzipped anyway to minimize network load ;)

yairm210:
This is a complex decision, but I think it's actually better to leave the server as part of the *general* jar rather than separate it to a different jar.

Pros:
Our current production stream gets it out to all clients (git, itch,io, flatpak) and changing all of them to include a new jar will make things both confusing to 'basic' players

Cons:
Larger jar size - if this was APK size then I'd have an issue with it, but honestly desktops have so much space they don't know what to do with it
Can't specify is as main class so it's not just `java -jar UncivServer.jar` - this is a real point, but honestly I feel that if someone is active enough to want to run a server he can read the correct command anyway.

yairm210:
In the future, @GGGuenni can use this as a base to implement many of the points of #6360, thus giving people a more selfish reason to migrate away from Dropbox. Want a multiplayer lobby, sorry pal that's only for private servers not for Dropbox ;)

SomeTroglodyte:
Agree 100%, zero-delay w/ pro/cons. Maybe one day make it runnable from the Options menu, even on Droid? Yea, joking.

xlenstra:
The only alternative I could think of that would not confuse players while also not having those cons, would be to have two jar's. These should then either distribute these independently, e.g., two different games on itch, or the server should only be downloadable after following an in-game button that redirects you to a download & instruction page on our github, thus not requiring additional distribution work on our part.

Though having them in the same jar is probably fine.

touhidurrr:
> I'm a believer in http verbs having meaning, but in any case this should be a post, since the data is passed by content and not by url And no, not exactly. Both verbs can accept headers, but headers have a limited size. For large data transfer you need content.

Normally in RESTful API's they does it like this:

`GET` get one data
`POST` create one data
`PATCH` edit one data
`PUT` create / edit multiple data (if data doesn’t exist then create new otherwise edit)
`DELETE` delete one data

touhidurrr:
> All files send to/from devices are gzipped anyway to minimize network load ;)

The Gzipping feels a little weird to me. Cause we gzip and then base64 it. The base64 part feels unnecessary if you implementing a Gzip library anyways.

SomeTroglodyte:
> base64

Methinks.... Because that is also used for issue communication right here - those copy to clipboard buttons in the game. You are right this need not be the format on-disk or over-the-wire. Feel up to designing an enum FileCompressionLevel {ClearText, GZipBase64, GZipBinary}, teaching it to GameSaver and MapSaver (you'll find not everything relevant is properly encapsulated in those), etc etc...?

Besides, the main source of bloat is long field names being repeated verrrrrrrryyy often in json, so the zipping part is exceptionally efficient and the 4:3 penalty of Base64 is far less than that.

PR #6387, opened 2022-03-20
The name of religion is translated into Japanese.

kobon2015:
I am referring to civ5. Therefore, the meaning is different from the literal translation.If there are no problems, translate the rest.I hope it helps you :)

Discord アーリア=kobot#8096

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Beliefs.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Beliefs.json
@@ -1,80 +1,80 @@
 
 [
     {
-        "name": "Ancestor Worship",
-        "type": "Pantheon",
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
No no no no njet ikke nein いいえ.. That belongs in [Japanese.properties](https://github.com/yairm210/Unciv/blob/master/android/assets/jsons/translations/Japanese.properties) - see [Translating!](https://yairm210.github.io/Unciv/Other/Translating/). Sorry for wasted work, but you don't want to force all the nice people playing in italiano, español, français, or even yankees, to have to learn Japanes to understand those?

yairm210:
As stated, this is not the place for these translations :)

PR #6389, opened 2022-03-20
Split 6 tiles visible unique into its parts, making it more moddable

xlenstra:
Most notably split the "6 tiles in every direction always visible" into "[+4] Sight", "Can see over obstacles".
Also added the "No sight" unique, which is unused, but could be cool for modding.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/UnitTypes.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/UnitTypes.json
@@ -59,7 +59,7 @@
     {
         "name": "Fighter",
         "movementType": "Air",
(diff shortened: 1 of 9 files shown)


SomeTroglodyte (approved):
Rationale behind the Belief/Policy reorder?
Json order is used in Pedia (afair), so best have all aircraft use the same order?

xlenstra:
> Rationale behind the Belief/Policy reorder? 

Made alphabetic so you can easier find whether or not something is in the list

> Json order is used in Pedia (afair), so best have all aircraft use the same order?

Sounds good, I'll fix that in a bit

yairm210 (approved):

Issue #6391, opened 2022-03-20
ctrl crashes the game on Ubuntu


yairm210:
To reproduce:
Start Unciv on Desktop
Click ctrl+whatever (ctrl+v for example)
Behold

```
java.lang.NoSuchMethodError: 'int kotlin.CharCodeKt.access$getCode$p(char)'
	at com.unciv.ui.utils.KeyCharAndCode$Companion.ctrl(KeyPressDispatcher.kt:67)
	at com.unciv.ui.utils.KeyCharAndCode$Companion.ctrlFromCode(KeyPressDispatcher.kt:73)
	at com.unciv.ui.utils.KeyPressDispatcher$install$1.keyDown(KeyPressDispatcher.kt:184)
	at com.badlogic.gdx.scenes.scene2d.InputListener.handle(InputListener.java:53)
	at com.badlogic.gdx.scenes.scene2d.Actor.notify(Actor.java:188)
	at com.badlogic.gdx.scenes.scene2d.Actor.fire(Actor.java:152)
	at com.badlogic.gdx.scenes.scene2d.Stage.keyDown(Stage.java:418)
	at com.unciv.CrashHandlingStage.access$keyDown$s80204510(CrashHandlingStage.kt:10)
	at com.unciv.CrashHandlingStage$keyDown$1.invoke(CrashHandlingStage.kt:30)
	at com.unciv.CrashHandlingStage$keyDown$1.invoke(CrashHandlingStage.kt:30)
	at com.unciv.ui.utils.ExtensionFunctionsKt$wrapCrashHandling$1.invoke(ExtensionFunctions.kt:329)
	at com.unciv.CrashHandlingStage.keyDown(CrashHandlingStage.kt:30)
	at com.badlogic.gdx.InputEventQueue.drain(InputEventQueue.java:58)
	at com.badlogic.gdx.backends.lwjgl3.DefaultLwjgl3Input.update(DefaultLwjgl3Input.java:189)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:394)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:143)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:116)
	at com.unciv.app.desktop.DesktopLauncher.main(DesktopLauncher.kt:60)
```

xlenstra:
[...]

PR #6392, opened 2022-03-20
Prevent Char-to-code crash

SomeTroglodyte:
For #6391 to try


yairm210:
Sure why not, better than a crash in any case

Issue #6383, opened 2022-03-20
DeCiv crashing unexpectedly


SpacedOutChicken:
Attached is the saved game. The game wouldn't let me copy the crash report.
[The.Patriots.-.85.turns.zip](https://github.com/yairm210/Unciv/files/8310615/The.Patriots.-.85.turns.zip)

HanuelChong has analyzed the situation here: https://github.com/SpacedOutChicken/DeCiv-Redux/issues/48

They believe that what's happening is that a city-state is attempting to gift a unit, but it has no unit it can give as a gift.

HaneulCheong:
Yeah, basically, it seems the game crashes when no unit is available to be gifted by a city-state. Possible quick solution would be catching the `NoSuchElementException` and printing some sort of warning log while ignoring it in game. There could be a better solution, but this is what comes to my mind right now.

SomeTroglodyte:
Nah, that .random() in line 98 of giveMilitaryUnitToPatron is the probable culprit. Make it a `.randomOrNull() ?: return`, done. But I'd probably make that multiline statement readable first.

PR #6393, opened 2022-03-20
Fix CS unit gift crash

SomeTroglodyte:
One way to solve #6383 (and another loophole when a CS-unique has no tech prereq). All creds to @HaneulCheong


HaneulCheong:
So this would also close [SpacedOutChicken/DeCiv-Redux#48](https://github.com/SpacedOutChicken/DeCiv-Redux/issues/48) I guess? Anyway, thanks for the fix.

yairm210 (approved):

SomeTroglodyte:
It was throw away or upload... Your work. Come to think of it, CivFlags.CityStateGreatPersonGift in CivilizationInfo.startTurnFlags - has a _safe_ .filter.random, but the flow could be much clearer (_why_ is it safe? Not immediately obvious) / repeated code unified...

PR #6400, opened 2022-03-21
Implement a rudimentary CrashReportSysInfoDesktop

SomeTroglodyte:
... why not?

Surprise: a `null!!` compiles (w/ tons of unreachable code warnings of course) and gives:

image
</details>

... I wiped the Windoze setup I once had to test, so no specific provisions for it. I'd like to ensure the 1804/1909/21H2 version thingy is in, the build number, and at most perhaps the chosen servicing ring or somesuch... Not in this PR. "os.version" should be better than nothing. Volunteers welcome.

### DIFF
--- a/desktop/src/com/unciv/app/desktop/CrashReportSysInfoDesktop.kt
+++ b/desktop/src/com/unciv/app/desktop/CrashReportSysInfoDesktop.kt
@@ -2,6 +2,7 @@ package com.unciv.app.desktop
 
 import com.badlogic.gdx.files.FileHandle
 import com.unciv.ui.utils.CrashReportSysInfo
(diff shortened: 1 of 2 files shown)


SomeTroglodyte:
That second commit uses running a cmd shell to fetch the desired windoze version info - maybe risky doing that while crashing. An alternative would be to run that beforehand when the class is instantiated by the desktop launcher and cache the string until needed - easy with the current schema.

Example output:
```
OS: Windows 10
	Windows 10 Enterprise Version 1909 (Build 18363)
Java: Oracle Corporation 11.0.11+9-b60-7590822
	Max Memory: 4016 MB
```

PR #6402, opened 2022-03-21
Fix crash on loading a save where Machu Picchu is built

### LINKED ISSUE #6401: Macchu Picchu now crashes us
FYI: 4.0.0 has a tiny regression, fix in testing

SomeTroglodyte:
fixes #6401


SomeTroglodyte:
Dumb idea: A unit test that loads a forged save that has every building built and all nations and units present. Ensures all transient updaters can run? Downside: Major rule changes (new or removed objects) would need an update to said construct. But such a test would have caught this one.

yairm210:
About half a week

SomeTroglodyte:
> About half a week

Then we _should_ do a unit test that could have caught this. And better get the ball rolling, this is going to hit _everyone_ sooner or later - except those playing base ruleset mods not using that unique.

PR #6403, opened 2022-03-21
Empire Overview: Resources

SomeTroglodyte:
This has waited long enough, and needs more waiting [^X], but I thought I could show it to interested parties.
- Linting
- Ability to transpose to vertical layout (handies in portrait might profit...)
- Extra data - now it could e.g. help locating the one Ivory your cities demand for WLTK which you haven't yet conquered/improved...

image
image
</details>

Note: "unimproved" isn't capitalized 'cuz the translation stuff for that one was already there :-O Input on wordings or choices welcome.

[^X]: It _really_ needs the fixed TabbedPager widget to shine - or staying in landscape high resolution. And that one was _far_ trickier than I anticipated, still needs testing and polish.


itanasi:
unimproved being a negative number just feels weird.

WLTK values are a bit unintuitive

SomeTroglodyte:
> unimproved being a negative number just feels weird

OK - actually that's exactly the reason to ask, perfect feedback, even though I originally did that consciously as it's "bad for you" - just like the WLTK demand, which is a count of cities wishing for that resource, while the other WLTK column is the number of cities celebrating due to that resource... The WLTK column headers in vertical mode already have tooltips (which benefit desktop only I'm aware) - maybe do them on both orientations and a little more verbose?

Anyway, pulling this, needs repairing the damage you get when you don't listen to your own advice and allow AS to open a translation file, and a little reorg with the other Empire Overview patches.

Issue #6405, opened 2022-03-22
Custom port pls


PlexPt:
https://github.com/yairm210/Unciv/blob/5fa10348b59747bbed203d503b0878687840e5cd/desktop/src/com/unciv/app/desktop/UncivServer.kt#L20

PlexPt:
Exception in thread "main" java.net.BindException: Address already in use

PR #6406, opened 2022-03-22
Custom server port

HAHH9527:
#6405


PlexPt:
use yml config pls

SomeTroglodyte:
> yml

no

PlexPt:
){(2~Z VH S WHWB%WM%Q_E

PlexPt:
There may be some parts of the client connection that need to be modified

yairm210:
Since the server is run via cli, configs will be cli-based

desktop/src/com/unciv/app/desktop/UncivServer.kt
```diff
@@ -12,12 +12,31 @@ import java.io.File
 
 
 internal object UncivServer {
+    private var serverPort = 8080
+
     @JvmStatic
-    fun main(arg: Array<String>) {
+    fun main(args: Array<String>) {
+        args.forEach { arg ->
+            when {
```
yairm210, line 20: Way too much nesting here. This can be cut down by half.

desktop/src/com/unciv/app/desktop/UncivServer.kt
```diff
@@ -12,12 +12,31 @@ import java.io.File
...
+                arg.startsWith("-port=") -> {
+                    try {
```
yairm210, line 22: If we already know that it might not be a number, we can use toIntOrNull and get rid of try/catch

desktop/src/com/unciv/app/desktop/UncivServer.kt
```diff
@@ -12,12 +12,31 @@ import java.io.File
...
+                        with(arg.removePrefix("-port=").toInt()) {
+                            if (this !in 1.rangeTo(65535)) println("'port' must in 1-65535")
```
yairm210, line 24: Spelling mistake :)
Also after refactor there should be only one error message for incompatible values, so that should be a problem

desktop/src/com/unciv/app/desktop/UncivServer.kt
```diff
@@ -12,12 +12,25 @@ import java.io.File
...
+    private var serverPort = 8080
+
     @JvmStatic
-    fun main(arg: Array<String>) {
+    fun main(args: Array<String>) {
+        args.forEach { arg ->
+            when {
+                arg.startsWith("-port=") -> with(arg.removePrefix("-port=").toIntOrNull() ?: 0) {
+                    if (this in 1.rangeTo(65535)) serverPort = this
+                    else println("'port' must be between 1 and 65535")
```
yairm210, line 23: 1024 to 49151 should be the range for pets. Below are reserved ports, above are dynamic ports.

yairm210:
This PR in only half of the required outcome - we need to add a way for clients to specify the port of the server.
Using the same textbox as the ip, we can have ip:port, and if the port isn't there then default to 8080

desktop/src/com/unciv/app/desktop/UncivServer.kt
PlexPt: @yairm210 You don't need to care which one your users use, they use whichever they like

yairm210: I disagree, that's a myopic view. Users who would use my game and encounter an error because they accidentally took a system-dedicated port and it caused them problems *would* open a ticket, and rightly so. Why not avoid problems for other people?

yairm210: Users *will* shot themselves in the foot in any way you allow them to. There is no real benefit to the other ports, anyone who knows enough to know he really needs them would be able to port forward, and so the obvious conclusion is to derby

yairm210 (approved):
Looks good!

SomeTroglodyte (approved):
Not sure I like having a _new_ ExtensionFunctions file when there is already [ExtensionFunctions.kt](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/utils/ExtensionFunctions.kt) - OK, splitting extensions according to usecase is a reason, but... Maybe check the /ui/ one later for functions that have nothing to do with UI and need moving, and lead the files with a fat comment outlining the existence and distinction of two files?

Also, what `getUrlWithPort` does might surprise - so Kdoc it. Consider implementation through standards? There should be a kind of "new URI; uri.parse(this); uri.port = 1234; uri.toString()" type of approach like every framework has (and that's the potential surprise right there - what would that pseudocode do differently). Check and if unsuitable comment why?

"-port=whatever" will try to run the server on port 0 - better fall back to 8080?

Spacing around ':' is usually: one after for type annotation, both before+after for superclass. Readability of N.rangeTo(M) vs. N..M is a matter of taste.

Lastly, if this ever gets bigger, whe should consider kotlinx-cli or even Clikt -> Those cater to standard 'nix notation easier as to the '=' used here. Just as food for thought. That said, I'd probably have done a mini-cli-mapper...
<details><summary>...roughly like this:</summary>

```kotlin
internal object UncivServer {
    @JvmStatic
    fun main(arg: Array<String>) {
        val argMap = (
                System.getenv().asSequence().map { it.key to it.value } +
                arg.asSequence()
                    .mapNotNull {
                        it.split('=', limit = 2).run {
                            if (this.size < 2) null else this[0] to this[1]
                        }
                    }
            )
            .toMap()

        fun String?.checkValidPort(): Int? {
            if (this == null) return null
            if (this.toIntOrNull() == null) {
                println("Port option '$this' is not a valid number.")
                return null
            }
            val port = this.toInt()
            if (port in 1024..49151) return port
            println("Port number '$port' is not in the valid range (1024..49151).")
            return null
        }

        val portNumber = argMap["-port"].checkValidPort()
            ?: argMap["-p"].checkValidPort()
            ?: argMap["UncivServerPort"].checkValidPort()
            ?: 8080

        val fileFolderName = argMap["-folder"] ?: argMap["-f"] ?: argMap["UncivServerFolder"] ?: "MultiplayerFiles"
        File(fileFolderName).mkdir()

        println("Starting UncivServer for ${File(fileFolderName).absolutePath} on port $portNumber\n(You can use command line option -port=X or environment UncivServerPort=X to use a custom port.)")
        embeddedServer(Netty, port = portNumber) {
```
</details>

desktop/src/com/unciv/app/desktop/UncivServer.kt
```diff
@@ -12,12 +12,25 @@ import java.io.File
...
 
 internal object UncivServer {
+    private var serverPort = 8080
+
     @JvmStatic
-    fun main(arg: Array<String>) {
+    fun main(args: Array<String>) {
+        args.forEach { arg ->
+            when {
+                arg.startsWith("-port=") -> with(arg.removePrefix("-port=").toIntOrNull() ?: 0) {
```
yairm210, line 21: I agree with @SomeTroglodyte that the default value for a non-parsable port should be 8080, which means instead of ?:-ing the result we can check it below (if this != null && this in 1024..49151)
However I think the current PR is useful enough to enter this version, and you can fix this minor issue in another insta-PR

PR #6413, opened 2022-03-22
Improved Widgets - Fixing Tabbed Pager Scrolling

SomeTroglodyte:
All of the widget patches from the Empire Overview work - getting the Cities page right was a major headache.

This has probably seen well over 200 debug launches, at least 30 hours of fiddling, trying and discarding different approaches, including getting event bubbling to work like in a proper windowing framework, or reimplementing the ScrollPane from scratch, or hacking protected values via reflection. This is a monster, but it conforms OK with the Gdx design philosophy, IMHO. Well, the ExpanderTab notifying its parent it needs more room is still unorthodox but not new. I had the "proper" Gdx way for that usecase (with invalidate / layout overrides) working, but it was noticeably slower, and logging showed an incredible number of events and re-layout taking place, so I ditched that.

Long story short - this is not final, but runs as is, and is better. Especially the City page in cramped portrait mode goes from barely useful to fine, even if the latest patches for that will come later.

### DIFF
--- a/core/src/com/unciv/ui/utils/AutoScrollPane.kt
+++ b/core/src/com/unciv/ui/utils/AutoScrollPane.kt
@@ -32,22 +32,22 @@ import com.badlogic.gdx.scenes.scene2d.utils.ClickListener
  */
 
 open class AutoScrollPane(widget: Actor?, style: ScrollPaneStyle = ScrollPaneStyle()): ScrollPane(widget,style) {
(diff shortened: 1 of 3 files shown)


SomeTroglodyte:
Clip:

https://user-images.githubusercontent.com/63000004/159545810-e31b8369-1878-4003-8507-eb84f2f6d6b6.mp4

As you can see, header and content are horizontally in sync, where my earlier code fails woefully.

Issue #6131, opened 2022-02-08
Missing entries in template.properties that make lines not translatable


Ouaz:
Here's a list of entries missing in template.properties, making them appear still in english when the game is set to another language.
There is a screenshot (in french) for each entries under the spoiler tag.
Maintainers can edit the message to tick checkboxes when fixed (I think)

_New !_

- [ ] **Friends list**
<details>
  <summary>Details</summary>
- In "Multiplayer" game menu, the new "Friends list" button is not translated. It is in the stringtables though.<br>
<img src="https://user-images.githubusercontent.com/11946570/171749807-e9d44c9c-1ce1-482c-bf22-83b6419867e0.png">
</details>

_New !_

- [ ] **[saveFileName] deleted successfully.**
<details>
  <summary>Details</summary>
- In "Load" game menu, when deleting a save, the "success" line is not translated. It is in the stringtables though.<br>
<img src="https://user-images.githubusercontent.com/11946570/171749523-50d412e4-864c-44ac-aa01-e0272a12cd3a.png">
</details>

_New !_

- [ ] **Please input player ID**
<details>
  <summary>Details</summary>
- In "New Game" menu, when creating a new multiplayer game. I noticed it a long time ago, but never reported it, as it was very minor. The string is not in the translation file.<br>
<img src="https://user-images.githubusercontent.com/11946570/171750169-15bc3c4d-c30c-405b-b0c5-10bb0647b2cb.png">
</details>

- [ ] **May buy [baseUnitFilter] units with [stat] for [amount] times their normal Production cost**
<details>
  <summary>Details</summary>
- In the Holy Warriors belief article, one of the keys is not translated (stat=Faith), as the sentence itself. String is in the translation file. I think there was no problem before.<br>
<img src="https://user-images.githubusercontent.com/11946570/167232620-61e9e156-a4d4-4f95-8b1a-428d65875a9d.png">
</details>

__________________________________

**- _Fixed_ -**

- [x] **Wait**
<details>
  <summary>Details</summary>
- The new "Wait" command in unit action list, is not translatable.<br>
<img src="https://user-images.githubusercontent.com/11946570/170373136-99cd4209-bfc9-48fc-b133-30df2957ca04.png">
</details>

- [x] **Consumes 1 [resource]**
<details>
  <summary>Details</summary>
- In the City screen, when choosing an unit to build, in the lower-right info panel. The string already exists in the translation file, and there was no problem in previous versions.<br>
[...]

PR #6421, opened 2022-03-24
Improve TranslationFileWriter (less hardcoding, fixes PolicyBranch)

SomeTroglodyte:
Fixes PolicyBranch names as mentioned in a late comment on #6131 
Not reaching the full scope of what the branch set out to find and fix, but as is already an improvement.

Open questions:
- ~~Why does the xltfw write the chance to be destroyed by nukes as requires translation even when already translated (I know why, what's mysterious is why not _all_ uniquetypes are doubled up)~~
- Conditionals seem to have xlt problems


SomeTroglodyte:
Next discovery: `TranslationFileWriter` processes `UniqueTypes`, OK, and there's some hitch there as shown by `DestroyableByNukesChance` being output twice, once with existing translation copied, and once with "requires translation"... But it also processes `UniqueParameterType` knownValues lists, partially, in a very inconsistent way, which is the root behind "Enemy Land" not being requested for translation.

Unsure how to go about that:
- Remove and make sure the uniques, including conditionals and their parameters, are processed completely
- Make knownValues part of the public class contract and have `TranslationFileWriter` scan all of them

SomeTroglodyte:
The next two pushes now do:
- Beginning of a framework to eliminate `TranslationFileWriter` hard-coding stuff depending on `UniqueParameterType`
- Use it to fix "The Spaceship" and the Fountain of Youth unique not translating - that's a giant hammer to flatten a tiny booger and could instead be done as templates, but - well - allows for more forgetfulness in the future
- A few missed entries that do belong in templates
- A translation run will be neutral on the included German

Next steps can already be seen as comments. Actually I had the system complete when Studio rolled back all my new code when I had only the translation folder right-clicked for Rollback... Getting that to be "neutral" meaning producing identical results won't make sense as the hard-coding is too off-the-mark, e.g. using a different placeholder than the typed Uniques for the same thing, or clear mapping where the hard-coded values belong in the UniqueParameterType's isn't always clear cut, so one has to move forward. Test results were almost neutral, good enough, and promise to be more future-proof. Remember, the whole thing where TranslationFileWriter hard-codes its guessing will be obsolete when all uniques are typed, anyway. At that point, `getTranslationWriterStringsForMatching` can go.

SomeTroglodyte:
> DestroyableByNukesChance being output twice

Actually now I see that's a "Chance"/"chance" case difference, meaning an actual bug probably, thus an easy thing to get rid of

SomeTroglodyte:
Much simpler, all the heavy hard-coding is gone from TranslationFileWriter, and it's a little faster despite less hash based lookups and more looping. The C/c fix follows existing translations and typical [n%] X precedents.

Next - scan conditionals to get that Enemy Land into the xlatoons.

SomeTroglodyte:
OK should have committed all languages. Who'd have thought the test would compare the left side of a french xlt line with the right side of a german equivalent line and vice versa.

SomeTroglodyte:
Hey wow - actually I'm five steps further on, also fixing Enemy Land, Rough terrain/Rough Terrain, "<sigh>", some trailing blanks that treating every line as Unique and therefore .trim()ing it have hidden, and so on. Soon, gotta pause first.

yairm210:
Not a problem, you can have multiple PRs from the same branch, but I see no reason to not include changes so far :)

SomeTroglodyte:
> multiple PRs from the same branch

Sure, but I preferred stashing and rebasing offline, keeping stuff up to date in an already published / partially merged branch gets hard quickly, and one _wants_ to test against the freshest codebase.

PR #6424, opened 2022-03-25
Update Hosting_a_Multiplayer_server.md

heipizhu4:
I suppose some noobs (like me) will have a hard time running cmd. Is it neccessary to give them some tutorials?


SomeTroglodyte:
No. Do **N O T** encourage Windoze users to run code from the root of the system drive. That's encouraging malware. That Win~crap~dows even allows such without jumping through hoops is proof of bad decisions. Also, 'dictionary' for directories/folders is a very uncommon wording.

Say, fellow coders, should we perhaps make the Server not a .main() sporting entry class but call it from DesktopLauncher - if Ctrl-Alt or somesuch is pressed do not even initialize Gdx but ask, maybe through awt or swing, whether they want the Server instead?

PR #6427, opened 2022-03-25
UncivServer update

HAHH9527:
- Add `clikt` lib.
- Add custom multiplayer file's folder.

The help text could be better, but my ability is limited.
Do we need i18n for server help text?


yairm210:
Fair warning - depending on how much this adds to the final JAR file, we may decide to go without this lib

PR #6428, opened 2022-03-25

HAHH9527:


heipizhu4 (approved):
Finally, there's someone who can translate with me together. Anyway, thanks for your translation.

heipizhu4 (changes requested):
#6444 You may check my pull for some translation change.
Pls remove 'translation required' after translation.

HAHH9527:
> #6444 You may check my pull for some translation change. Pls remove 'translation required' after translation.

Doc say `Each untranslated phrase will have a "requires translation" line before it, so you can quickly find them. You don't need to remove them yourself if you don't want to - they will be automatically removed the next time we rebuild the file.`

HAHH9527:
@heipizhu4 
I'm a Chinese programmer, translation is not my strong suit.
> 我是个程序员，翻译并不是我的强项。

SomeTroglodyte:
"automatically removed" means that tool on the desktop version's options pages (that the boss runs just before a release) will take a structure "how we want it to be" that is built from the jsons and the templates file, then fill in the existing translations. Meaning all comments are lost except those the generator does itself, as is ordering. But when/if PR 6433 gets merged, there will be one exception.
That also means if you want to use comments in translation file pushes between releases to communicate, why not, but github comments are easier :wink:

PR #6433, opened 2022-03-26
Translation writer conditionals

SomeTroglodyte:
I think that's all for the translation writer I'm gonna do for now.

#### Notes
- Again, since that unit test compares parameter placeholders across languages, all of them are included, and that will break open translation PR's not based on this, and merging will be _quite_ difficult. I recommend merge this and have people start over, _or_ wait for a lull in translation PR's, have them all merged, then fix this by pulling those updates and simply re-running the generator over those.
Merging older translation PR's after this will be tricky because:
- This moves stuff around, since the scanner is more capable, stuff can appear earlier. I think that's good (else I could have switched processing order), uniques appear under the object where they are first used, so the xl'tor has more context, instead of in one long list.
- The Unique() treatment including [] and <> processing is no longer applied to _everything_. This means Caesar's war declaration <sigh> came through in that step, and keys are no longer trimmed. This should fix ` OR `, but a) two flavour texts, on austria and india, were now not-matching since the xlt had the trimmed string but json an extra space. Removed the space in json so the xlt stays valid. b) that <sigh> still falls through when actual xlations are matched, and at that point the context is gone, so I opted to side-step that problem and make it a (sigh).
- The Unique() treatment _is_ applied to uniques collections, obviously, but also to promotion names - a kludge to not make our idea for those `[unit] ability` promotions uglier in the xlt files (still worked, but the lines were all `[Maori Warrior] ability = ...` - almost self-explaining why).
- The Unique() treatment _is not_ applied to Quest description, which will copy the original json [] content to the xlt files (still auto-matching the old translations) - including the one hardcoded [50]%. I think that's just fine, all the others are clearer now.

- **Unmatched translations are no longer lost** - see the wiki changes included.

- Minor tweaks...
- Performance is better between 20 and 50% I guess, fluctuates too much on my box to be more precise.

#### *Not* treated here

- Will fix some of the stuff mentioned in #6131 if not all - I have _not_ checked, sorry.
- Actually fixing the UniqueParameterType knowledge of how the filters work, or UniqueTypes choosing a different filter in the declaration and implementation. See my issue for input on that. I fixed only tiny stuff necessary to get nice and stable translation files (like Rough terrain / Rough Terrain).
- Forgot: I declared Impassable to be a valid TerrainFilter mainly so no template entry is necessary because it could well _be_ one, I should have actually implemented the option.

### DIFF
--- a/core/src/com/unciv/models/translations/TranslationFileWriter.kt
+++ b/core/src/com/unciv/models/translations/TranslationFileWriter.kt
@@ -3,7 +3,6 @@ package com.unciv.models.translations
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.utils.Array
-import com.unciv.Constants
 import com.unciv.JsonParser
 import com.unciv.models.metadata.BaseRuleset
 import com.unciv.models.ruleset.*
@@ -15,8 +14,6 @@ import com.unciv.models.ruleset.unique.*
 import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.ruleset.unit.Promotion
 import com.unciv.models.ruleset.unit.UnitType
-import com.unciv.models.stats.Stat
-import com.unciv.models.stats.Stats
 import java.lang.reflect.Field
 import java.lang.reflect.Modifier
 
@@ -43,7 +40,8 @@ object TranslationFileWriter {
 
             return "Translation files are generated successfully."
         } catch (ex: Throwable) {
-            return ex.localizedMessage
+            ex.printStackTrace()
+            return ex.localizedMessage ?: ex.javaClass.simpleName
         }
     }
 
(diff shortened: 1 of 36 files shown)


core/src/com/unciv/models/translations/TranslationFileWriter.kt
```diff
@@ -177,13 +176,24 @@ object TranslationFileWriter {
...
 
-                val lineToWrite = translationKey.replace("\n", "\\n") +
-                        " = " + translationValue.replace("\n", "\\n")
-                stringBuilder.appendLine(lineToWrite)
+                stringBuilder.appendTranslation(translationKey, translationValue)
             }
 
             countOfTranslatedLines[language] = translationsOfThisLanguage
 
+            // As a courtesy to translators, add in any deletions as comments
```
yairm210, line 184: I'm not sure this is a good idea. We ourselves add and remove lines a lot, so every time we deprecate a unique the old translation line will be added at the bottom? That's worse than just deleting it in my opinion

SomeTroglodyte: True but a) we can fix spelling on the english side with complete impunity b) if it gets out of hand we'll introduce automatic cleanup. I _almost_ added a date...

Question is - should we use more imperative wording on the wiki or even other channels to get our translators to accept the cleanup responsibility?

PR #6439, opened 2022-03-26
Reworked city name generation

HaneulCheong:
- **Reworked city name generation**

### DIFF
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -103,7 +103,9 @@ class CityInfo {
 
         name = generateNewCityName(
             civInfo.nation.cities,
-            civInfo.cities.map { cityInfo -> cityInfo.name },
+            civInfo.gameInfo.civilizations.flatMap { civilization ->
+                civilization.cities.map { city -> city.name }
+            },
             civInfo.hasUnique(UniqueType.BorrowsCityNames)
         )
 
@@ -187,7 +189,7 @@ class CityInfo {
      * predefined prefix to the list of names until an unused name is successfully generated.
      *
      * @param nationCityNames Every city name of the nation.
-     * @param usedCityNames Used city names of the nation.
+     * @param usedCityNames Every city name in use including foreign cities.
      * @param hasBorrowsCityNames Whether this nation borrows city names or not
      * ([UniqueType.BorrowsCityNames]).
      * @return A new city name in [String].
@@ -201,7 +203,7 @@ class CityInfo {
 
         // If all names are taken and this nation has the BorrowsCityNames unique,
         // return a random borrowed city name
-        if (hasBorrowsCityNames) return generateBorrowedCityName()
+        if (hasBorrowsCityNames) return generateBorrowedCityName(usedCityNames)
 
         /*
         If the nation doesn't have the unique above,
@@ -225,30 +227,32 @@ class CityInfo {
         }
     }
 
-    /** Generates and returns a city name borrowed from another nation. */
-    private fun generateBorrowedCityName(): String {
+    /**
+     * Generates and returns a city name borrowed from another nation.
+     *
+     * @param usedCityNames Every city name in use including foreign cities.
+     * @return A borrowed city name in [String].
+     */
+    private fun generateBorrowedCityName(usedCityNames: List<String>): String {
         // Attempt to return a random name from the pool of last unused city name for every *other*
         // nation in this game, skipping nations whose names are exhausted
-        val usedNames = civInfo.gameInfo.civilizations.flatMap { civilization ->
-            civilization.cities.map { city -> city.name }
-        }
         val otherNations = civInfo.gameInfo.civilizations.filter { nation ->
             nation.isMajorCiv() && nation != civInfo
         }
-        var unusedNames = otherNations.mapNotNull {
-            it.nation.cities.lastOrNull { city -> city !in usedNames }
+        var unusedCityNames = otherNations.mapNotNull {
+            it.nation.cities.lastOrNull { name -> name !in usedCityNames }
         }
-        if (unusedNames.isNotEmpty()) return unusedNames.random()
+        if (unusedCityNames.isNotEmpty()) return unusedCityNames.random()
 
         // As per fandom wiki, once the names from the other nations in the game are exhausted,
         // names are taken from the rest of the nations in the rule set
         val missingNations = getRuleset().nations.filter { nation ->
             nation.key !in civInfo.gameInfo.civilizations.map { civ -> civ.nation.name }
         }
-        unusedNames = missingNations.values.flatMap { nation ->
-            nation.cities.filter { city -> city !in usedNames }
+        unusedCityNames = missingNations.values.flatMap { nation ->
+            nation.cities.filter { city -> city !in usedCityNames }
         }
-        if (unusedNames.isNotEmpty()) return unusedNames.random()
+        if (unusedCityNames.isNotEmpty()) return unusedCityNames.random()
 
         // If all else fails (by using some sort of rule set mod without city names),
         // just return something so we at least have a name
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -107,7 +107,7 @@ class CityInfo {
                 civilization.cities.map { city -> city.name }
             }.toSet(),
             civInfo.hasUnique(UniqueType.BorrowsCityNames),
-            listOf("New ", "Neo ", "Nova ", "Altera ")
+            listOf("{New }", "{Neo }", "{Nova }", "{Altera }")
         )
 
         isOriginalCapital = civInfo.citiesCreated == 0
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -105,7 +105,7 @@ class CityInfo {
             civInfo.nation.cities,
             civInfo.gameInfo.civilizations.flatMap { civilization ->
                 civilization.cities.map { city -> city.name }
-            },
+            }.toSet(),
             civInfo.hasUnique(UniqueType.BorrowsCityNames)
         )
 
@@ -188,14 +188,14 @@ class CityInfo {
      * [hasBorrowsCityNames] and, if true, returns a borrowed name. Else, it repeatedly attaches a
      * predefined prefix to the list of names until an unused name is successfully generated.
      *
-     * @param nationCityNames Every city name of the nation.
-     * @param usedCityNames Every city name in use including foreign cities.
+     * @param nationCityNames A [List] of every city name of the nation.
+     * @param usedCityNames A [Set] of every city name in use including foreign cities.
      * @param hasBorrowsCityNames Whether this nation borrows city names or not
      * ([UniqueType.BorrowsCityNames]).
      * @return A new city name in [String].
      */
     private fun generateNewCityName(
-        nationCityNames: List<String>, usedCityNames: List<String>, hasBorrowsCityNames: Boolean
+        nationCityNames: List<String>, usedCityNames: Set<String>, hasBorrowsCityNames: Boolean
     ): String {
         // Attempt to return the first missing name from the list of city names
         // ex) Rome, Antium, ... , Interrama, Adria
@@ -219,10 +219,11 @@ class CityInfo {
         var candidate: String?
         var number = 1
         while (true) {
+            val currentPrefix: String = prefix.repeat(number)
             candidate = nationCityNames.firstOrNull { cityName ->
-                (prefix.repeat(number) + cityName) !in usedCityNames
+                currentPrefix + cityName !in usedCityNames
             }
-            if (candidate != null) return prefix.repeat(number) + candidate
+            if (candidate != null) return currentPrefix + candidate
             number++
         }
     }
@@ -230,10 +231,10 @@ class CityInfo {
     /**
      * Generates and returns a city name borrowed from another nation.
      *
-     * @param usedCityNames Every city name in use including foreign cities.
+     * @param usedCityNames A [Set] of every city name in use including foreign cities.
      * @return A borrowed city name in [String].
      */
-    private fun generateBorrowedCityName(usedCityNames: List<String>): String {
+    private fun generateBorrowedCityName(usedCityNames: Set<String>): String {
         // Attempt to return a random name from the pool of last unused city name for every *other*
         // nation in this game, skipping nations whose names are exhausted
         val otherNations = civInfo.gameInfo.civilizations.filter { nation ->
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -62,7 +62,6 @@ class CityInfo {
     var turnAcquired = 0
     var health = 200
 
-
     var religion = CityInfoReligionManager()
     var population = PopulationManager()
     var cityConstructions = CityConstructions()
@@ -102,7 +101,11 @@ class CityInfo {
         location = cityLocation
         setTransients()
 
-        setNewCityName(civInfo)
+        name = generateNewCityName(
+            civInfo.nation.cities,
+            civInfo.cities.map { cityInfo -> cityInfo.name },
+            civInfo.hasUnique(UniqueType.BorrowsCityNames)
+        )
 
         isOriginalCapital = civInfo.citiesCreated == 0
         if (isOriginalCapital) civInfo.hasEverOwnedOriginalCapital = true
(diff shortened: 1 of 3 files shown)


HaneulCheong:
> This method attempts to return the first unused name in the given [nationCityNames] taking [usedCityNames] into consideration. If that fails, it then checks whether the nation [hasBorrowsCityNames] and, if true, returns a borrowed name. Else, it repeatedly attaches a predefined prefix to the list of names until an unused name is successfully generated.
> \- `CityInfo.generateNewCityName()`'s docstring

In short: Rome, Antium, ... , Adria, New Rome, New Antium, ... , New Adria, New New Rome, New New Antium, ...

This should more closely resemble the way how things are done in the original game.

HaneulCheong:
Bug found: My method currently doesn't account for a name being available again by the city using it being captured.

SomeTroglodyte:
> better way to do this

Sequences probably. The "definition how to iterate" can have infinite depth, and a terminal transform does the trick. Still, we should probably have some upper boundaries on city count and city name length. Does the city name edit box impose a max?

`usedCityNames` sounds like it should be a set. That said, generating it using flatMap should use sequences and .toSet() at the end. That said, this set generator should be part of the specialized function, not on the call site. Yes, those parameter names tell the caller nicely what the function does and how, but still...

You dropped "Neo". Don't like the Matrix movies? OK... Still, it was a nice option. Better yet, *do* implement that newCityNamesPrefixes: ArrayList<String>? moddable source on Nation? Or suffixes too? Or one list, if it starts with a capital letter it's prefixed else it's suffixed? Would that even work in your language?

HaneulCheong:
@SomeTroglodyte 

> Sequences probably. The "definition how to iterate" can have infinite depth, and a terminal transform does the trick. Still, we should probably have some upper boundaries on city count and city name length. Does the city name edit box impose a max?

*What if I told you* I tried every possible way of generating a sequence in my knowledge and every time the project failed to build again and again so I'm on the verge of posting a question on Stack Overflow oh god...

Here's one of my attempts:

```kotlin
val seq = sequence {
    var number = 1
    while (true) {
        yield(number)
        number++
    }
}
seq.forEach { number ->
    candidate = nationCityNames.firstOrNull { cityName ->
        (prefix.repeat(number) + cityName) !in usedCityNames
    }
    if (candidate != null) return prefix.repeat(number) + candidate
}
```

I can simply add an unreachable return statement at the end as a last resort, but that feels like me losing against Kotlin :(

> `usedCityNames` sounds like it should be a set. That said, generating it using flatMap should use sequences and .toSet() at the end. That said, this set generator should be part of the specialized function, not on the call site. Yes, those parameter names tell the caller nicely what the function does and how, but still...

I intentionally pulled them out in an attempt to make the function as pure as possible since it's private (and I would've succeeded if it wasn't that `generateBorrowedCityName()`). I can indeed change `usedCityNames` to a `Set` though.

> You dropped "Neo". Don't like the Matrix movies? OK... Still, it was a nice option. Better yet, _do_ implement that newCityNamesPrefixes: ArrayList?

The old method allowed infinite duplicates of "Neo (name here)". The reworked method would instead generate "New New (name here)", "New New New (name here)" and so on, hence ensuring that every city name is unique.

Also, italic text ;)

> moddable source on Nation? Or suffixes too?

This is well beyond my ability since I recently joined the project, and I'll be honest; I don't want to destroy the whole production while tweaking a rather trivial part of the game lol. I'll still keep this in mind; I'm a big fan of "lore"-friendliness when it comes to gaming you know.

> Or one list, if it starts with a capital letter it's prefixed else it's suffixed? Would that even work in your language?

I...don't think that would end well; Many writing systems (including mine) don't have distinct "capitals".

HaneulCheong:
#### TODO
- [ ] Find a way to generate an infinite sequence that actually builds
- [ ] Change `usedCityNames` to `Set`

SomeTroglodyte:
> Many writing systems (including mine) don't have distinct "capitals".

Thanks for confirming.

> oh god

D%&$mn, now I'll have to check out your branch to see for myself...

SomeTroglodyte:
Just noticed: Old and new code will not allow city name translation once a prefix is added... "{New} {$cityName}"? Then `New = ` would need to go into templates.

```kotlin
        val prefixes = listOf("New", "Neo", "Nova", "Altera") // don't take this too seriously

        return (1..Int.MAX_VALUE).asSequence()
            .flatMap { number ->
                prefixes.asSequence()
                .flatMap { prefix ->
                    nationCityNames.asSequence()
                        .map { cityName ->
                            "{$prefix} ".repeat(number) + "{$cityName}"
                        }
                }
            }.firstOrNull {
                it !in usedCityNames
            } ?: "The City without a Name"
```
seems just fine... I mean true, the compiler wants to know you'll return something, so if you're sure the loop/sequence scan will always give a value, you'll have to tell it - here it's easy, Elvis or !!. I prefer Elvis in case we limit the number of prefixes to limit the max name length...

With the builder, that could be:
```kotlin
        return sequence {
            for (number in 1..Int.MAX_VALUE) {
                for (prefix in prefixes) {
                    for (cityName in nationCityNames) {
                        val candidate = "{$prefix} ".repeat(number) + "{$cityName}"
                        if (candidate !in usedCityNames) {
                            yield(candidate)
                            return@sequence
                        }
                    }
                }
            }
        }.first()
```
 ...slight variation, would operate pretty much the same under the hood, but crash in the first() operator if the sequence were ever to exhaust.

Simpler yet, do those three for loops directly (no Sequence) and `return candidate` - but you would indeed have to supply a fallback return at the end of the loops. If that weren't the last alternative in the decision making, you'd wrap in a function.

All this just because you asked. Use you own judgement..

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -102,7 +101,13 @@ class CityInfo {
         location = cityLocation
         setTransients()
 
-        setNewCityName(civInfo)
+        name = generateNewCityName(
+            civInfo.nation.cities,
+            civInfo.gameInfo.civilizations.flatMap { civilization ->
+                civilization.cities.map { city -> city.name }
```
xlenstra, line 107: Does this mean that if I rename my cities, the same name will be chosen again? E.g., I found the city of Rome, then rename it to 'My Favourite City', and then found another city. To me it looks like Rome is then once again a valid choice, and as it is the first unused name, will be chosen again?

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -175,60 +180,85 @@ class CityInfo {
...
+        TODO: Make the prefix moddable per nation?
+         Also, if then, what about other languages using suffixes?
+        TODO: Could've used a [sequence] from 1 to [int.MAX_VALUE] instead of the while loop but my
+         IDE wouldn't stop complaining about a return expression being required. I wonder whether
+         this is the best or there is a better way to do this.
+        */
+        val prefix = "New "
+        var candidate: String?
+        var number = 1
+        while (true) {
```
xlenstra, line 221: I'm a bit hesitant with `while (true)` loops as for some reason they tend to repeat indefinitely even when that should be impossible. Maybe a for loop repeating 10 times and otherwise returning `The City without a Name` as used in `generateBorrowedCityName`? Having a city called "New New New New New New New New New New New New Rome" probably also isn't exactly great either imo

core/src/com/unciv/logic/city/CityInfo.kt
HaneulCheong: Yes, and that's how it worked in the original game if I recall correctly.

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -175,60 +180,85 @@ class CityInfo {
...
-        // If for some reason we have used every single city name in the game,
-        // (are we using some sort of baserule mod without city names?)
+        // names are taken from the rest of the nations in the rule set
+        val missingNations = getRuleset().nations.filter { nation ->
+            nation.key !in civInfo.gameInfo.civilizations.map { civ -> civ.nation.name }
+        }
+        unusedCityNames = missingNations.values.flatMap { nation ->
+            nation.cities.filter { city -> city !in usedCityNames }
+        }
+        if (unusedCityNames.isNotEmpty()) return unusedCityNames.random()
```
xlenstra, line 255: I personally do not necessarily feel that rewriting this function by adding additional variables but keeping the logic the same improves readability. This is partly as I've written this function myself. If anyone else does feel that way, I'm fine with changing it, but as it is I do not necessarily see the added value of adding additional vars and vals that are only ever used once :)

core/src/com/unciv/logic/city/CityInfo.kt
HaneulCheong: > The old method allowed infinite duplicates of "Neo (name here)". The reworked method would instead generate "New New (name here)", "New New New (name here)" and so on, hence ensuring that every city name is unique.
> \- https://github.com/yairm210/Unciv/pull/6439#issuecomment-1079749064

On a second thought, I think I'll just use a list of prefixes as you and @SomeTroglodyte advised...

core/src/com/unciv/logic/city/CityInfo.kt
xlenstra: I'll check that as that sounds unintuitive.

SomeTroglodyte: I think that's a good thing actually.

xlenstra: > Yes, and that's how it worked in the original game if I recall correctly.

Turns out you're right

HaneulCheong: Lol I remember that because I once renamed my capital while playing the original game only to find out my next newly-founded city will be named as the very name I've replaced. Maybe we can make an exception for capital names (i.e. the first element of `nationCityNames`), but this should be conferred with more people in my opinion.

SomeTroglodyte: No, for that I'd vote for a persisted nextCityNameIndex. Yes the Huns could (should be able to?) steal the name right next turn, and that I'd find quite nice indeed.

HaneulCheong:
> All this just because you asked. Use you own judgement..

I just accepted my defeat against Kotlin and used an ending return statement at this point... Still, thanks. I actually learned a lot from your code and googling some parts.

> // don't take this too seriously

*How about I do anyway*

SomeTroglodyte:
> learned a lot

That's the whole point.

SomeTroglodyte (changes requested):
As you requested the review, I'll go from a purely code view point of view only, though I really should test myself for that rating. I suggest _you_ give it a thorough testing (do a mod with a few nations with very short city name lists then play easy difficulty advanced start or so and see that everything is as expected, including doing a local translation and checking that part).

Trailing spaces on translatable units work in theory - but see a little detail on one of my latest translation PR's where they got .trim() ed (and worked nevertheless because all components trimmed - but templates aren't). Fragile. Also, there are languages not using spaces as guidance, which would mean you're right. Hmmm.

So - those ruleset changes: No. a) forEach when a for is just as easy is a No. b) always treat what's read from json as immutable even when someone made it not so (tidbit: the json reader mutates val's through reflection). And c) No! The city names (everything I think) go through translation when it's time to be displayed. Try it. Play english, have a city you know has a translation in your language, save, change language, load and look...

Also, I think you could learn about what List, Map, Set and Sequence really mean for memory allocation and performance. filter, map and especially flatMap operations running on a List will materialize (allocate memory) every intermediate result. The next map/whatever will run after that is done. Not so on Sequences which are essentially instructions and do not actually move data until they are iterated. The example with the range, prefixes and citynames - the sequence version uses not much more memory (they do live on the heap as complex delegate constructs or so I imagine) than the loop version, it does _not_ store the cartesian product of all possible numbers and prefixes anywhere before the next level is run. Sequences also have some kinds of optimizations built in, with different terminating transforms you get different delegates earlier in the processing - very opaque, but simplified they have a chance to short-circuit and bypass unnecessary parts of the "phase space". The big "con" is - building those delegates constituting the processing logic is some overhead.

So, in generateBorrowedCityName, val otherNations is simple, a filter on an List then alocated and stored as a List, order of magnitude less than 100. Cheap. But only used once. Next step runs a double iteration over it and a few ArrayLists (oreder of magnitude <100), materializing the result. Last operation says I take only one of them (at most) but it's random - OK, a sequence might optimize that some, but the "classic" version you wrote is on the scale of not much more than a thousand iterations (and allocations), cheap enough to forego a Sequence, maybe borderline. But you have to understand to see that. Gist: Verbose, just fine, alternate approaches might be better but probably not far off, maybe readability trumps.

Now I look at missingNations =, and see a filter running on a List doing a contains on... a mapped list built on the fly. Will the compiler see that the sub-List won't change over the outer loop and optimize that to only one materialization of the sub-List? No, civilizations is marked mutable. Meaning it will build and discard one List per outer iteration. Plus contains on a List can be far slower on larger Lists than on Map/Set. What we want is akin to Set subtraction, so why not do one? Should work with a suitable unique key (names aren't always). Or at least materialize the inner lookup ourselves and speed up the contains... Now a Sequence solution to Set subtraction is possible but you'd have to be careful not to build two independent ones  where the outer must terminate the inner per iteration, nothing gained. Could be worth it when chaining the following flatMap + filter + random though. Maybe. Then again, readability. Gist: could save on allocations, specifically that inner 'nations in use' lookup.

The thing about _not_ preferring `forEach` over `for` [is official on kotlinlang.org](https://kotlinlang.org/docs/coding-conventions.html#loops) - and you're doing the reverse. Are they equivalent for the result? Maybe, if the compiler is clever enough. But it starts out with different instructions - for says compile a simple iteration with a loop variable. forEach says call one of several generic library functions and pass it an Iterable and a delegate as parameters. In other words, the {} of the for is a block for the compiler, trivial - the {} of forEach is a lambda that needs to exist on the heap before it can be used, and marshalling closures into lambdas can be a tad more complicated than calling upon them in a block............. ..... ..

OK, I'll shut up. You're only getting such a long-winded rant because you can learn. Watch Studio's tooltips, and watch what your code does in the debugger - one can _see_ what one's source will actually cause under the hood. Use Ctrl-B and Alt-F7 extensively. And so on... What I want is to stop paying so much attention, which I can do when I see you're thinking for yourself. To express it in a satirically narcissistic way.

The only actual potentially breaking problem I see is _when_ city names get translated. Trailing blanks on translation units - I see why but I still needed to warn. Without the TranslationFileWriter changes I'm proposing in #6433 I think they will fail, with them they should work and allow more nicely for languages not using spaces like we do - but I still don't like them. Pity we didn't use json as format for the translation files, then that wouldn't be such a problem. The rest is - well, matter of taste, I don't see why it shouldn't work as intended.

HaneulCheong:
> Trailing spaces on translatable units work in theory - but see a little detail on one of my latest translation PR's where they got .trim() ed (and worked nevertheless because all components trimmed - but templates aren't). Fragile. Also, there are languages not using spaces as guidance, which would mean you're right. Hmmm.

Languages such as Japanese and Chinese don't use spaces at all as far as I know and there are some translation keys intentionally including leading/trailing spaces to mitigate that. Even in my language spaces are removed from most toponyms (ex. "New York (City)" -> "紐約(市)", "ニューヨーク(市)", "뉴욕(시)") I don't think trimming those spaces was a good idea...

> Blanks: Watch out for blanks at the start of a line or two of them before the equals sign. If you got such a line - those blanks are part of the translation key and must not be deleted on the left side, and you should probably also include them in your translation (unless your language doesn't need spaces to separate things).
> \- https://yairm210.github.io/Unciv/Other/Translating/#pitfalls

---

> Also, I think you could learn about what List, Map, Set and Sequence really mean for memory allocation and performance.
> . . .

So it's something like generators in Python.

> The thing about _not_ preferring `forEach` over `for` [is official on kotlinlang.org](https://kotlinlang.org/docs/coding-conventions.html#loops) - and you're doing the reverse.
> . . .

Also, didn't know about the latter one at all. I'll keep these in mind.

---

> So, in generateBorrowedCityName, val otherNations is simple, a filter on an List then alocated and stored as a List, order of magnitude less than 100.
> . . .
> Gist: could save on allocations, specifically that inner 'nations in use' lookup.

Honestly [all I did](https://github.com/HaneulCheong/Unciv/compare/2ef8581b52b8127b177acbe34c1f99a5b73900ba...56804044cdbfd7a8bdc6b605aefb78895891f710#diff-fbe720e9817ffdc802bb40557a08a86a431127c60d4225d7f4f1a4591f1bbc6cL197-R265) to `borrowCityName()` (currently `generateBorrowedCityName()`) is:

 - Changed `usedCityNames` in line 198 to a method argument
 - Divided `newNames` in line 202 to `otherNations` (line 247) and `unusedCityNames` (line 250)
 - Divided `newNames` in line 214 to `missingNations` (line 257) and `unusedCityNames` (line 260)
 - Changed `.map { /* . . . */ }.flatten()` in line 218-221 to `.flatMap { /* . . . */ }` (line 260) and that's done by Android Studio auto formatting, not me

Basically the core of the method is not mine; I just thought some more readability for a method closely related to my PR would be nice...... I'll attempt to restructure and optimize the whole method based on your review, or maybe I can just revert it back to its previous state...

---

> So - those ruleset changes: No.
> . . .

> The only actual potentially breaking problem I see is _when_ city names get translated.
> . . .

I'm currently trying to come up with a solution but it would take some serious amount of time if I have to guess; Save files are allowed to have something like `"{New }{Rome}"`, right?

HaneulCheong:
Update: This already spiraled way out of my ability since https://github.com/yairm210/Unciv/pull/6439#issuecomment-1079749629 and I already spent three quarters of my friday plus weekend; At this point I am sure I can't fixing everything mentioned here without breaking something, and some issues brought up even predates this PR. I'm going to close this and open a new PR to focus on what I originally intended to deliver. All the other issues can (and hopefully will) be solved one step at a time, a.k.a. TODO.

---

Found issues and wanted features to wrap up the conversation above:
- City name prefixes are not translatable
- `borrowCityName()` is sub-optimal
- Support moddable prefixes (and suffixes if below is implemented) per nation
- Support city name suffixes

PR #6443, opened 2022-03-26
Generalized a few nation uniques

xlenstra:
Most notably, added the conditional `<with [amount]% chance>`.
Also refactored some code in `battle.kt` to remove duplication of spawning captured enemy units.
Most of these ideas were recently requested in #suggestions on discord.
I'm open to better wordings of Japans unique: "Damage is ignored when determining unit Strength <for [non-air] units>". While it contains all the relevant information, I'm not sure if the general user immediately understands what it means


SomeTroglodyte (approved):
Wording? Umm - aren't conditionals technically working everywhere within the unique string? Not only the end?
"enemey"? Sounds like a musical dialect. Pity template doesn't match code.
Changing the OnlyAvailableWhen line only to add a trailing blank? Ctrl-F, Check ".*" button, search for " $"
Modifying Windozer-only gradlew.bat? Won't survive the next Gradle upgrade.
Make GainFromDefeatingUnit stack but only for the Stats part? Nice to not overlook that.
`tileBasedRandom by lazy` - Think contextBasedRandom and do a proper hashCode over the complete `state`? The hashCode is   specifically meant to support the equality contract, meaning a clone should keep its hashCode, and since Vector2 is not by us and therefore equality contract abiding, position.hashCode is just as stable and cheaper than position.toString().hashCode(). And if we just let the Studio wizard generate equality support for StateForConditionals and all its parts, state.hashCode() would be best... I'm still waiting for all those Map/Set<TileInfo> to bite us in the posterior due to no proper support for the Map/Set contract - which says keys must support equality.

SomeTroglodyte:
Oh, and <> in github markdown -> within `` they're pretty safe unlike "&", no need to `\` escape; outside they might get lost including contents... _if_ they could be any xml element. If in doubt, use <code>&amp;lt;</code> (which I wrote < code > & amp ; lt ; < / code > without the spaces) :fearful:

xlenstra:
> Wording? Umm - aren't conditionals technically working everywhere within the unique string? Not only the end?

True, but when translated they're always put at the end of the unique.

> "enemey"? Sounds like a musical dialect. Pity template doesn't match code. 

Good catch, I'll change it

> Changing the OnlyAvailableWhen line only to add a trailing blank? Ctrl-F, Check ".*" button, search for " $" 

At first I removed the new line at the end, but later decided to revert that. Turns out I didn't fully revert it. I'll change this.

> Modifying Windozer-only gradlew.bat? Won't survive the next Gradle upgrade. 

My game was running extremely slow without this change, even though I'm on a linux machine. Not sure how, but changing this did fix it

> Make GainFromDefeatingUnit stack but only for the Stats part? Nice to not overlook that. `tileBasedRandom by lazy` - Think contextBasedRandom and do a proper hashCode over the complete `state`? The hashCode is specifically meant to support the equality contract, meaning a clone should keep its hashCode, and since Vector2 is not by us and therefore equality contract abiding, position.hashCode is just as stable and cheaper than position.toString().hashCode(). 

I've just copy-pasted this from `UniqueTriggerActivation.kt`, but I suppose I might as well change it.

> And if we just let the Studio wizard generate equality support for StateForConditionals and all its parts, state.hashCode() would be best... I'm still waiting for all those Map/Set to bite us in the posterior due to no proper support for the Map/Set contract - which says keys must support equality.
> Oh, and <> in github markdown -> within `` they're pretty safe unlike "&", no need to `\` escape; outside they might get lost including contents... _if_ they could be any xml element. If in doubt, use <code>&amp;lt;</code> (which I wrote < code > & amp ; lt ; < / code > without the spaces) :fearful: 

Thx, changed.

SomeTroglodyte:
I didn't mean just _using_ state.hashCode _as is_ without further research... OK, it's a data class, so we have the data class default implementation, a priori a good thing, which afaik calls hashCode for all members, multiplies each by a different power of 31 (starting with 31^0), and sums up - or so. But most members of state (OK, all, makes little sense counting ignoreConditionals) don't have that luxury and thus get the Any.hashCode implementation which does - what? Calls [identityHashCode](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native/identity-hash-code.html) - good luck understanding that description. I assume it will do something from the _memory address_ of the instance... Which is not good for what we want - which is same random after a reload.

I hope I said _**if**_ our RulesetObjects (at least those contained in StateForConditionals) all had their own conforming euqals+hashCode implementation - **_then_** that would be a better seed.

From another angle: We don't implement equality, thus `equals` calls are not obligated to say "yes equal" on an original and its clone - and the default implementation likely won't. Using such an object as key on a HashSet or HashMap won't find a stored value using a clone as key. That seems to work for us as such maps/sets are rebuilt whenever stuff is cloned, and no cloning seems to be done without rebuilding such maps/sets. Breaking that means nasty hard to find bugs. When `equals` says Dolly and its clone aren't equal, hasCode isn't required to be equal either. Which means serialization/deserialization will change the hashCodes. Which means reload spamming to get different random results based on a generator seeded with such hashes will eventually succeed.

Why would Vector2.hashCode or String.hashCode **work** to prevent reload spamming? 'cuz they _do_ implement semantic equality which means same hash for the same value independent of cloning or memory address or serialization cycles.

PR #6446, opened 2022-03-27
Improved city name generation

HaneulCheong:
- Improved city name generation
  - `generateNewCityName()`, previously `setNewCityName()`, attempts to return the first unused name in the given `nationCityNames` (every city name of the nation), taking `usedCityNames` (every city name in use including foreign cities) into consideration. If that fails, it then checks whether the nation `hasBorrowsCityNames` and, if true, returns a borrowed name. Else, it repeatedly attaches one of the given `prefixes` (currently "New", "Neo", "Nova", and "Altera") to the list of names up to ten times until an unused name is successfully generated. If all else fails, it returns `"The City without a Name"` as a fail safe.
  - The previous version of the method (2ef8581b52b8127b177acbe34c1f99a5b73900ba) basically only took how many cities are created into consideration. The improved method now more closely resembles that of the original game.

Note: This is a simplified version of #6439. Issues and features brought up in https://github.com/yairm210/Unciv/pull/6439#issuecomment-1079895472 may be fixed in the future PRs.

### DIFF
--- a/core/src/com/unciv/logic/city/CityInfo.kt
+++ b/core/src/com/unciv/logic/city/CityInfo.kt
@@ -102,7 +102,14 @@ class CityInfo {
         location = cityLocation
         setTransients()
 
-        setNewCityName(civInfo)
+        name = generateNewCityName(
+            nationCityNames = civInfo.nation.cities,
+            usedCityNames = civInfo.gameInfo.civilizations.asSequence().flatMap { civilization ->
+                civilization.cities.asSequence().map { city -> city.name }
+            }.toSet(),
+            hasBorrowsCityNames = civInfo.hasUnique(UniqueType.BorrowsCityNames),
+            prefixes = arrayOf("New ", "Neo ", "Nova ", "Altera ")
+        )
 
         isOriginalCapital = civInfo.citiesCreated == 0
         if (isOriginalCapital) civInfo.hasEverOwnedOriginalCapital = true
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -175,60 +181,105 @@ class CityInfo {
...
+            absentMajorNations.flatMap { nation ->
+                nation.cities.asSequence().filter { city -> city !in usedCityNames }
+            }.toSet()
+        if (newCityNames.isNotEmpty()) return newCityNames.random()
+
         // If for some reason we have used every single city name in the game,
-        // (are we using some sort of baserule mod without city names?)
-        // just return something so we at least have a name
-        return "The City without a Name"
+        // (are we using some sort of rule set mod without city names?)
```
xlenstra, line 279: Very minor point, but iirc 'ruleset' is a single word

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -175,60 +181,105 @@ class CityInfo {
...
-        }
+        // names are taken from the rest of the major nations in the rule set
+        val absentMajorNations: Sequence<Nation> =
+            getRuleset().nations.values.asSequence().filter { nation ->
+                nation.isMajorCiv() && nation !in aliveMajorNations
+            }
+        newCityNames =
+            absentMajorNations.flatMap { nation ->
+                nation.cities.asSequence().filter { city -> city !in usedCityNames }
+            }.toSet()
```
xlenstra, line 275: If you're changing the type anyways, why not a hashset?

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -175,60 +181,105 @@ class CityInfo {
...
+        We take the last unused city name for each other major nation in this game,
+        skipping nations whose names are exhausted,
+        and choose a random one from that pool if it's not empty.
+        */
+        val otherMajorNations: Sequence<Nation> =
+            aliveMajorNations.filter { nation -> nation != foundingCiv.nation }
+        var newCityNames: Set<String> =
+            otherMajorNations.mapNotNull { nation ->
+                nation.cities.lastOrNull { city -> city !in usedCityNames }
+            }.toSet()
```
xlenstra, line 263: Same here

xlenstra (approved):

core/src/com/unciv/logic/city/CityInfo.kt
SomeTroglodyte: My take is - Set is a much more general contract than HashSet. With toSet you leave some choice to the compiler, which is good. And lookup on a generic Set can't be less performant than on a HashSet, or else JetBrains would have none. Look under the hood - it's implemented as a LinkedHashSet with optimizeReadOnlySet called on it. Next version might implement it differently if they find some room for improvement...

So, unless you know you need a specific Set implementation, toSet ***is*** best practice.

HaneulCheong: My original thought process here was 'I need an immutable or read-only set'. Also, aren't LinkedHashSets slightly faster than normal HashSets? Then again HashSets consume less memory than the other...so maybe I should indeed convert the result to HashSet.

core/src/com/unciv/logic/city/CityInfo.kt
HaneulCheong: I knew I cannot rely on Android Studio...

core/src/com/unciv/logic/city/CityInfo.kt
HaneulCheong: I've eventually decided to leave it as `toSet()`. After all it's just some sets of strings and this method is called once in every 5 minutes or so, which is not much considering there are methods and functions called every single game loop. I'm gonna believe in the compiler.

core/src/com/unciv/logic/city/CityInfo.kt
HaneulCheong: https://github.com/yairm210/Unciv/pull/6446#discussion_r838406998

core/src/com/unciv/logic/city/CityInfo.kt
HaneulCheong: Since this is a comment I'll fix this in one of the future PRs; I have a feeling that I'll come back to these methods really soon... Thanks a lot for pointing this out though; I definitely missed it without you.

PR #6449, opened 2022-03-28
Version rollout

yairm210:


SomeTroglodyte:
On my side this is failing tests - both `noTwoPlaceholdersAreTheSame` where it looks the test is claiming bullshit as well as allTranslationsHaveCorrectPlaceholders where it seems somehow the "save removed entries" feature lost the parameter names - in my testing they were preserved... I'll need to investigate.

yairm210:
With all due respect if the 'save old lines' is preventing rollout I'm going to remove that :)

SomeTroglodyte:
Very old idea and I included it only because it turned out to be so _easy_ thanks to lines beginning with '#' but containing '=' already being part of Translations, not being thrown out. A writer run would throw them out because no template... That made the feature cheap.

Still, I am not convinced it is at fault per se. I tested repeatedly and those lines stayed identical, not tripping tests... Was I lucky only to "deprecate" parameter-less lines in testing??? I would investigate as soon as Studio is free to switch branches.

SomeTroglodyte:
Yes I was unlucky to have only parameter-less removals in testing. Fix is trivial, use `entry` instead of key. Question is - what to do about the removals I already bungled with those []?
a) Revert all language files to just before the TranslationFileWriter run, merge the fix, rerun TranslationFileWriter
b) Manual removal of the offending lines
c) Patch test to ignore comments

I'll prepare a PR doing the fix _and_ (c)

PR #6438, opened 2022-03-26
Some translation and translation requirement fixes

SomeTroglodyte:
Requires #6433 and #6435 (because I was lazy testing)
Closes #6131

Open question see [this comment on 6131](https://github.com/yairm210/Unciv/issues/6131#issuecomment-1079693634)

PR #6450, opened 2022-03-28
Update German translation

Mape6:


SomeTroglodyte:
Would be better to wait for #6451 or even #6438

PR #6451, opened 2022-03-28
Patch TranslationFileWriter preserve removed feature

SomeTroglodyte:
As promised in [#6449](https://github.com/yairm210/Unciv/pull/6449#issuecomment-1080964570)

Re-adding the param names in the now truncated removals should be trivial, I've probably got them from testing this, but am not sure I took the correct master state to generate them.. Now which commit exactly has all new translations but not the writer run?...


SomeTroglodyte:
This second commit is the broken translations lines fixed. I originally used 4.0.1 for testing, wrong, pulled [7abb982657](https://github.com/yairm210/Unciv/tree/7abb9826573745f71e3cf73b2c1ecc23a6e346b3) for basing this on instead.

yairm210:
Sorry, I really don't agree with the concept. It's not actually helpful to translators (going back and forth takes more time than thinking of a retranslation), and it clogs up the file endlessly. People will need to dedicate time to clean this up, and future translators won't want to touch it because "it was here so it might be important IDK".

SomeTroglodyte:
Merging this and _then_ reverting that component might have been easier. A revert is just a few line deletions, and the writer will clean up automatically next run. I for one have missed such an option in the past and have then used git blame instead, which is much more of a hassle. But I hear ya - and accept pro/con may end up at well below 50:50.

PR #6453, opened 2022-03-29
Fixed the official wiki

HaneulCheong:
-   [x] Fix broken links
-   [x] Fix the tables in [Civilization-related-JSON-files.md](https://github.com/yairm210/Unciv/blob/5ab44dd945af96a46e97e5a48e2ecec3f691c597/docs/Other/Civilization-related-JSON-files.md)
-   [x] Remove all redundant "table of contents" bulleted lists
-   [x] Make the image in [Regions.md](https://github.com/yairm210/Unciv/blob/5ab44dd945af96a46e97e5a48e2ecec3f691c597/docs/Other/Regions.md) properly be displayed
-   [x] Fix lists with invalid indentation
    -   It seems Python Markdown (the underlying engine behind MkDocs) is very strict when it comes to indentation of list items. I had no choice but to standardized all of them with the following format:
    ```markdown
    -   Item A
    -   Item B
        -   Item B.A
        -   Item B.B
    ```
-   Fix headings
    -   [x] Add a title to every document as level-1 heading
    -   [x] Chapters should start with a level-2 heading
-   [x] Fix footnotes with invalid syntax


HaneulCheong:
Context:

The absolute state of the wiki

SomeTroglodyte:
Looks well researched and executed!

... makes me think - around "Create a Github account, if you don't already have one" - you can easily clone the original without forking or compile without cloning and thus without account... But no sense mentioning that there, is there? The described way is best practice if you want to push and PR from your local code... And not even that as really "best" would mean teaching people how to set up the `git@github.com:/you/Unciv.git` instead of `https://github.com/you/Unciv.git` way.

HaneulCheong:
I remember having a really bad time typing stuff into Git Bash and realizing I just reverted hours of my work myself when I was learning Git. Perhaps we could include a link or two leading to guides on how to use Git and GitHub?

HaneulCheong:
I'll keep testing the wiki in my local mkdocs server to find more "bugs" hidden here and there, but for now I think this PR is ready for review and merge. Also, I left `uniques.md` unchanged since that seems to be auto-generated.

SomeTroglodyte:
> guides on how to use

Not quite on topic, but in case you haven't found that before: I kept notes while getting set up, with some learning experience actually documented well enough _for myself_ that I lost no time when I had to start from scratch. Very specialized stuff in parts. Look for the "Scratchpad" repo on my profile. Anything in there is property of the unciv team, so you can take and reuse any parts if you want. Haven't maintained that in ages, though.

HaneulCheong:
I'll check it out; Thanks.

yairm210 (approved):

xlenstra (approved):
I haven't looked through all 1500-3000 changed lines, but what I did look at looked good

HaneulCheong:
What.

xlenstra:
This should update the old wiki to also include the new changes, but for some reason it didn't work. Maybe cause the normal checks where still running? It probably doesn't matter anyways now that we have the new wiki and that is already seemingly updated

SomeTroglodyte:
Well, we can go into actions and look up a little more detail.
```
Running `rsync`.
rsync: change_dir "/home/runner/work/Unciv/Unciv//CodeRepo/docs/wiki" failed: No such file or directory (2)
sending incremental file list

sent 20 bytes  received 12 bytes  64.00 bytes/sec
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1207) [sender=3.1.3]
total size is 0  speedup is 0.00
Error: Process completed with exit code 23.
```
... looks like maybe some environment variable did not get set properly?

PR #6460, opened 2022-03-29
Tabbed pager architecture update

SomeTroglodyte:
Original Reason: Can't currently merge forward that Map Editor, as my decrepit brain had forgotten that code needed a better TabbedPager and included a better approach than I did later for the EmpireOverview pages - same usecase and I still forgot. This reworks the widget closer to the approach MapEditor uses. Some code gets much simpler, two places need an extra private class not needed before, still a good balance imho.

Tangible improvements:
- OptionPopup mod checker tab now only scrolls the mod list, and the tab is hidden with no mods as it was before the G&K/Vanilla split.
- Keyboard Left/Right works just as in Civilopedia (now making that one use the same widget is for another day) - options and overview, though it has a funny effect on the options when called from world - of course - map panning is independent and stays active. Cute, so I left it. Wouldn't be surprised others may find that offensive.
- Server IP box big enough for an actual live IPv4 address - by filling the space used above which by lucky chance fits nicely - at least in certain languages. Getting TextBox to actually play nice and report a content-dependent pref/minWidth was too frustrating, it stubbornly stays at 150f.


SomeTroglodyte:
OK, fixing what's wrong with the merged version is not scope of this PR. It begins with object SimpleHttp belonging in com.unciv.logic.multiplayer, NOT in com.unciv.ui.worldscreen.mainmenu.OptionsPopup, then unsanitized user input being passed to the non-forgiving URL constructor instead of a more forgiving parser,... I'll stash and roll back what I did to it and limit this branch to what it was meant to do.

PR #6463, opened 2022-03-30
Typed denmarks unique and generalized it

xlenstra:


yairm210 (approved):

SomeTroglodyte:
Looks good... If I were to actually _search_ for excuses to nitpick...:
` ?: 100f` - why not do that when setting the var, make it non-nullable and init with 100f?
Don't see the compatibility code to keep `DisembarkCostDeprecated` working for a while. I could be blind of course.

xlenstra:
> Looks good... If I were to actually _search_ for excuses to nitpick...: ` ?: 100f` - why not do that when setting the var, make it non-nullable and init with 100f? 

I'd say that performance-wise this is negligible (don't have the data to back that up). Additionally, the `100f` is arguably a magic number, and by keeping it in the function we at least keep that magic number inside the function, where it is used any time we want to remove all movement points.

> Don't see the compatibility code to keep `DisembarkCostDeprecated` working for a while. I could be blind of course.

Whoops, missed that, I'll add it right away

SomeTroglodyte:
Hmmm... The 100f might be too low for mad mods... :cold_face:

xlenstra:
Tbh, if anybody makes a unit with 103 movement points, imo, they deserve to skip past zoc and (dis)embarking, just as how fast they are going. So I'd see it as a feature, actually 🙃

PR #6464, opened 2022-03-30
Added a unique for attacking when embarked

xlenstra:
As requested in #3242

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -77,7 +77,7 @@ object Battle {
         // Withdraw from melee ability
         if (attacker is MapUnitCombatant && attacker.isMelee() && defender is MapUnitCombatant) {
             val withdraw = defender.unit.getMatchingUniques(UniqueType.MayWithdraw)
-                .sumOf{ it.params[0].toInt() }  // If a mod allows multiple withdraw properties, ensure the best is used
+                .maxOfOrNull { it.params[0].toInt() } ?: 0 // If a mod allows multiple withdraw properties, ensure the best is used
             if (withdraw != 0 && doWithdrawFromMeleeAbility(attacker, defender, withdraw)) return
         }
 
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -448,6 +448,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     @Deprecated("as of 3.19.8", ReplaceWith("Eliminates combat penalty for attacking across a coast"))
     AttackFromSea("Eliminates combat penalty for attacking from the sea", UniqueTarget.Unit),
     AttackAcrossCoast("Eliminates combat penalty for attacking across a coast", UniqueTarget.Unit),
+    AttackOnSea("Can attack normally when embarked", UniqueTarget.Unit),
```
yairm210, line 451: Maybe without the "normally", just "May attack when embarked"

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -77,7 +77,7 @@ object Battle {
         // Withdraw from melee ability
         if (attacker is MapUnitCombatant && attacker.isMelee() && defender is MapUnitCombatant) {
             val withdraw = defender.unit.getMatchingUniques(UniqueType.MayWithdraw)
-                .sumOf{ it.params[0].toInt() }  // If a mod allows multiple withdraw properties, ensure the best is used
```
SomeTroglodyte, line 80: That was done on purpose, [see here](https://github.com/yairm210/Unciv/commit/644c61a85196842fae5d805e280189cd68968a3b) - only the comment was left from the old maxOfOrNull, no longer matching reality

xlenstra: In that case it's still a bit weird; if a caravel has an 80% chance to withdraw and it is from a nation with a 40% chance to withdraw, should it have a 120% chance to withdraw? To me this didn't make sense, so I reverted it. But I suppose that in that case you could say it should have a `1-(1-80%)*(1-40%)` chance to withdraw. I'll chance the code accordingly

yairm210 (approved):

PR #6466, opened 2022-03-30
Remove re-orientation for OptionsPopup

### LINKED ISSUE #6456: Options menu crashes Android S if...
**Platform**
Android S AVD, "G00gle APIs" x86_64 revision 9 system image, wipe data before test (but don't forget to re-remove the horrible Assistant from the power button if you do that)

**Version**
current master

**To Reproduce**
Set resolution to x700 or less, turn on enable portrait, turn AVD to portrait, re-open Options.
Orientation changes to landscape and the options popup is re-rendered, then the game crashes. Logs show our crashhandlers are trying but failing, the process gets killed by the OS anyway. Last exception seen: RulesetCache.getVanillaRuleset() on an empty RulesetCache -> null pointer.

Tested on L, O, Q and R: All do that re-orientation thingy just fine.

Possible approaches:
- Hand the API level out of the activity context and tell BaseScreen to skip the re-orientation when it's 31 (S) or higher
- Remove re-orientation altogether (but not the portrait feature). Some option tabs are a bit ugly at minimum resolution and portrait, but bearable (and could be mitigated by fiddling, that Advanced tab has gotten an aesthetical nightmare anyway)

<br/>
Opinions???

SomeTroglodyte:
This is the branch behind the screenshot in #6456. One way to fix #6456.

### DIFF
--- a/core/src/com/unciv/ui/utils/LimitOrientationsHelper.kt
+++ b/core/src/com/unciv/ui/utils/LimitOrientationsHelper.kt
@@ -8,12 +8,6 @@ import com.unciv.models.metadata.GameSettings
  *  position changes to automatically result in App orientation changes. This is about limiting that feature.
  */
 interface LimitOrientationsHelper {
-    /** Set a specific requested orientation or pull the setting from disk and act accordingly
-     *  @param newOrientation A SCREEN_ORIENTATION_* value from [ActivityInfo]
-     *          or SCREEN_ORIENTATION_UNSPECIFIED to load the setting
-     */
-    fun limitOrientations(newOrientation: Int)
-
     /** Pass a Boolean setting as used in [allowAndroidPortrait][GameSettings.allowAndroidPortrait] to the OS.
      *  @param allow `true`: allow all orientations (follows sensor as limited by OS settings)
      *          `false`: allow only landscape orientations (both if supported, otherwise default landscape only)
--- a/android/src/com/unciv/app/AndroidLauncher.kt
+++ b/android/src/com/unciv/app/AndroidLauncher.kt
@@ -1,7 +1,6 @@
 package com.unciv.app
 
 import android.content.Intent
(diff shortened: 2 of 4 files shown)


yairm210 (approved):

core/src/com/unciv/ui/utils/LimitOrientationsHelper.kt
```diff
@@ -8,12 +8,6 @@ import com.unciv.models.metadata.GameSettings
  *  position changes to automatically result in App orientation changes. This is about limiting that feature.
  */
 interface LimitOrientationsHelper {
-    /** Set a specific requested orientation or pull the setting from disk and act accordingly
-     *  @param newOrientation A SCREEN_ORIENTATION_* value from [ActivityInfo]
-     *          or SCREEN_ORIENTATION_UNSPECIFIED to load the setting
-     */
-    fun limitOrientations(newOrientation: Int)
-
```
xlenstra, line 16: Does this interface still make sense if we remove one of the only two functions in it?

xlenstra (approved):
I'm not 100% certain about what is going on, but the code looks good on a cursory reading, and if you've tested it, I believe it should be fine.

core/src/com/unciv/ui/utils/LimitOrientationsHelper.kt
SomeTroglodyte: That's still the platform-specific / platform-independent boundary pattern. The interface is the platform-agnostic common ground, even if no other platform implements it.

Hmm... There were tiltable monitors for PC's and some communication and API's so the desktop / running software could react. We don't need to cater for that as it would be seen as resize events already. But imagine those had a motor and an API for that so software could _set_ its orientation... :rofl: Never leave a full coffee mug near a tilt monitor. Even a non-motorized one.

PR #6475, opened 2022-04-01
Update Spanish.properties

Phoenix7g:
General grammar improvement and some rewordings.


SomeTroglodyte (approved):
Lista -> Cola en #6461 y ahora Cola -> Lista?
But overall looks like thorough work. And tons of it.

Phoenix7g:
> Lista -> Cola en #6461 y ahora Cola -> Lista? But overall looks like thorough work. And tons of it.

Oh, hi! Yeah, I know I've reverted my previous change over cola/lista. At the end, I felt that "lista" is more common than "cola", although the latter is more precise and technical in this context. Any of them will work, anyway.

Btw I couldn't find the criteria for the English capitalization, so at some point I just stopped struggling with it :D

ArchDuque-Pancake:
Yeah, "Lista" is more apropiate. "Cola" can mean "butt" in some countries as well.
An alternative can be also "Fila"

PR #6479, opened 2022-04-01
Move SimpleHttp object

SomeTroglodyte:
Based on #6460, thus the diff shows stuff not really belonging here until that is in.
- object SimpleHttp moved to package com.unciv.logic.multiplayer
- fun getIpAddress moved into SimpleHttp (an UI page shouldn't import networking stuff)
- sendRequest now has a simple "friendly" interpreter, so "localhost" in the server box _would_ actually connect to a locally running UncivServer, and pure IP values left in setting won't crash the game on 'test connection'.

Un-touched:
- I see you're suggesting https:// now. As I understand it (might be wrong) UncivServer serves only cleartext http, and actually playing off it won't work using https. Getting ktor/Netty to speak SSL/TLS is doable but not trivial, and would probably involve a self-signed cert, which to be of real use would need to be (its public key part) included in the android resources and referenced in the android manifest.


yairm210:
I'm not sure we even use the get Ip anymore, it night be removable
Also, I am suggesting https, but if you notice I do not give instructions anywhere on how to create an accessible URL, because there are *so many ways*. And since that's the case, there are also many ways to hide the http of ktor behind https, so encouraging that direction even without giving instructions feels like the correct way forward.

yairm210:
'http://localhost' should have worked anyway, no?
Or do you mean "even without the protocol specified"

SomeTroglodyte:
> so many ways

I must have missed something. UncivServer doesn't speak https. And the client is geared to UncivServer in its requests, simple though they are - is it that easy to create a server reacting in a way the client is happy with? Hmmm... I suppose it is, at least I could imagine doing one without coding in IIS, not that I touched that contraption in years... Now that is a nice side-project - create other UncivServers through wiki documentation alone.

> should have worked anyway

Yes, even without the protocol specified is the point. Before would crash, the URL constructor is unforgiving unlike the URI one (which is kind of almost intuitive).

yairm210:
We have a Discord user that did just that, https://uncivserver.xyz
That's what I conducted checks with ;)

SomeTroglodyte:
Yup, I'll join you on Discord once the client is on the official Ubuntu repo, or the browser based thing runs cleanly under µBlock Origin, or maybe even when the Weendohs client doesn't do evil malware things like registering system-wide protocol handlers. :wink:  Btw, I thought that domain was a typical imaginary example thingy when I saw it mentioned somewhere in source. Wow, cloudflare signed, not even letsencrypt.

Mape6:
The Discord user wrote that unciv server with Cloud flare pages and some JS as far as I remember. That's why the cert for TLS is from Cloudflare.
I wrote a small http server for UnCiv in Python with the http.server module. That was easier for me to deploy it on Linux because most distros come with python 3 pre-installed. I put it behind a Reverse proxy that's gets the cert from let's Encrypt automatically. So the server is not directly visible from the internet but reachable over https.

SomeTroglodyte:
> I wrote a small http server for UnCiv in Python

Wiki? Or repo it in your account (without specific configuration of course) and we link from the wiki as "semple implamentations" or somesuch?

Mape6:
It is already in my Github Account: https://github.com/Mape6/Unciv_server

I'm not sure if it's production ready. I couldn't test it with real players til now because Play Store seems not to ship 4.0.3 to public. And I will add the delete function the next days. But as far as I know, it's not in use currently.
so yes, it could be an example implementation. I thought about a Docker container with that script and a pre configured reverse proxy in it.
Any Security hints?

Mape6:
I did a refactor of the whole code to add more checks on the URI.
@SomeTroglodyte Do you like to check the code? Hows your Python knowledge? :-)

SomeTroglodyte:
Rusty. Only a few months, rust comes so quickly these days... Last time I noticed just two things. Message saying PUT in the GET handler and trusting the headers for length (but no idea how the rest of the code would do if that header were lying)... OK let's look at the update.
`uuid_regex` the `\b` seem redundant at a hex-digit "-" boundary.
preview/full file needn't be separate, a `(?:_[pP]review)?` would do - but then maybe later distinct treatment might be desirable, so it's all sausage.
I would probably have started out with a `Begin` thingy to support command line parameters... `import begin` but to get it I believe you had to `pip3 install Begins` or somesuch... Memory is fuzzy... Then later it would go
```python
@begin.start
def main (
		verbose: "Be chatty" = False,
		salty: "Mix salt into your ice cream" = False,
		flavour: "How many scoops?" = "strawberry",
		scoops: "What's 7 times 9?" = 42,
		result: "Mr Creosote" = "BOOM" ):
	"Silly icecream consumption simulator"
	do_eat_icecream(flavour, salty, result, verbose)
```

More interesting would be - does SimpleHTTPRequestHandler protect against overrun attacks - at the do_* level it's already too late (I'd still actively reject any query longer than say 1k - or maybe 80 byte in our case - in the handler before firing a regex on it). That and how do you get this thing to use _which_ certificate? No idea, never used libraries in that direction.

Mape6:
> `uuid_regex` the `\b` seem redundant at a hex-digit "-" boundary. preview/full file needn't be separate, a `(?:_[pP]review)?` would do - but then maybe later distinct treatment might be desirable, so it's all sausage. 

You are totally right regarding the regex. I copied the regex from [https://ihateregex.io/expr/uuid/](https://ihateregex.io/expr/uuid/) 
I combined both files with an `or` in the if loop, now.

> I would probably have started out with a `Begin` thingy to support command line parameters... `import begin` but to get it I believe you had to `pip3 install Begins` or somesuch... 

I didnt know the package and searched for it. On the first look it definitively helps a lot. On the second look its quiet old...
Python comes with `argsparse` module. Thats how I will do it. 
 
> More interesting would be - does SimpleHTTPRequestHandler protect against overrun attacks - at the do_* level it's already too late (I'd still actively reject any query longer than say 1k - or maybe 80 byte in our case - in the handler before firing a regex on it). 

Unfortunately there is not so much implemented regarding length checking in the http.server module. The developers wrote also a warning message in their documentation that it should not be used in production. See [http.server](https://docs.python.org/3/library/http.server.html)
I added a path length check (max. 128 characters) and a content-length check (max. 1048576 Bytes so 1 MByte)
I checked them with cURL and altered the content-length parameter in the header. If the value is shorter than the data, that will be ok.
If the value is larger than the data, the TCP session will stay open and waits for additional data. Until the end of the world probably.

> how do you get this thing to use _which_ certificate? No idea, never used libraries in that direction.

There is no support for TLS in that module. It's only a simple HTTP server for development testing. As already written, it's not for productive environments. I managed to get the Unciv server reachable over https with an reverse proxy. In my case traefik that runs on docker. In my environment the Unciv server is reachable only via a dedicated FQDN through that reverse proxy. The reverse proxy is checking the requested FQDN, if it is not the right one, no one will reach the unciv server.
Somit denke ich, ist die Angriffsfläche sehr gering. 😊

PR #6484, opened 2022-04-01
Update template.properties

Ouaz:
- Fix untranslatable entry "Revoke protection for" (see comment in #6131)


SomeTroglodyte:
Problem is, all is case sensitive, and the code uses a lower case 'p' there... The inconsistency with the other lines I noticed only after that last post. So, either merge this in, start collecting translations nobody will see until the code (DiplomacyScreen.331) is changed to use the capital 'P', or...

Ouaz:
Geez, I hesitated to change the capitalization... Because of "Declare **P**rotection of" the line below... :(
I should have not.
I push an extra commit.

PR #6493, opened 2022-04-06
Fix Goddess of Protection - maybe

### LINKED ISSUE #6488: Defence goddess belief doesn't work
**Platform**
Android 12

**Version**
Download here 4.0.3

**Describe the bug**
I choose the pantheon belief "defence goddess" (+30% damage of city when attack), but it is not displayed in attack overview (is it really work?). It can be seen in attached screenshot

Screenshot_20220404-233430

SomeTroglodyte:
Untested, may perhaps fix #6488

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Beliefs.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Beliefs.json
@@ -60,7 +60,7 @@
     {
         "name": "Goddess of Protection",
         "type": "Pantheon",
-        "uniques": ["[+30]% attacking Strength for cities"]
+        "uniques": ["[+30]% Strength for cities <when attacking>"]
     },
     {
         "name": "Goddess of the Hunt",
--- a/core/src/com/unciv/ui/utils/TextSimilarity.kt
+++ b/core/src/com/unciv/ui/utils/TextSimilarity.kt
@@ -37,6 +37,10 @@ fun getTextDistance(text1: String, text2: String): Int {
         if (char1 == char2) {
             i1++
             i2++
+        } else if (char1.lowercaseChar() == char2.lowercaseChar()) {
+            dist++
+            i1++
+            i2++
         } else {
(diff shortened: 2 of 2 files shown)


yairm210:
Hm this feels like a unique that wasn't correctly deprecated, or alternatively that our religion unique tests aren't working 100%

SomeTroglodyte:
Methinks the `baseRulesetHasNoBugs` test should run checkModLinks in a form that runs all checks so they return _all_ UniqueComplianceErrorSeverity types instead of just one severityToReport, that's where most checker intelligence gets lost, UniqueType.UniqueComplianceErrorSeverity.WarningOnly is never evaluated as far as I can see.

SomeTroglodyte:
... but yes this one wouldn't have made it into mod check results anyway. I wonder why. Ah yes - because "[+30]% attacking Strength for cities" is untyped, the deprecated _typed_ unique would have been "[+30]% Attacking Strength for cities". So it just gets the spellcheck treatment which is silent.

SomeTroglodyte:
Yes, `getRelativeTextDistance(UniqueType.StrengthForCitiesAttacking.placeholderText, "[]% attacking Strength for cities")` == `getRelativeTextDistance("[]% Attacking Strength for cities", "[]% attacking Strength for cities")` = 0.1818 which is above the threshold. Ouch. @will-ca - your code? How come one Char different in case makes such a large "distance"? Cuz it fails to see any similarity after the A/a mismatch probably, it gets derailed. Might a dumb `} else if (char1.lowercaseChar() == char2.lowercaseChar()) { dist++; i1++; i2++ } else...` in there help? Yup, relative dist goes to 0.03 and:
<details><summary>screen</summary>

image

</details>

SomeTroglodyte:
Since that upper/lower case fragment improves things even without touching the rest of the ruleset check cardhouse, pushed. With the old Json you _would_ see the spellcheck message for the G&K ruleset in the UI, in green, so you'd really have to go look for it, but it's there.

PR #6497, opened 2022-04-06
Make Icons clickable in religions per city display

SomeTroglodyte:
Small idea - you see nice icons and follower count and pressure, but which religion is it? What are its effects? So I made the icons - just the icons, not the row - clickable and lead them to a relevant screen. Works from here: 
image
or the detail icon (not the expander of course) here: 
image
Leads to the selected Religion on the Overview Screen - or Civilopedia for mere Pantheons.

Doubts:
- Worth it?
- Leaving Overview (but not pedia) won't get you back to your City screen, the return architecture of pedia isn't copied in overview. From the world popup it's just fine.
- How to communicate the possibility to users...?


yairm210:
Religion is something that was constructed almost entirely by people who are not me, so I'll wait for feedback from other people on this :)

SomeTroglodyte:
> Religion is something that was constructed almost entirely by people who are not me

Amen! :grinning:

Ouaz:
> Worth it?

I like that, and actually, I wished it was possible to open the religion info panel either from the city screen or the city popup, to check what are its effects on the city.

> How to communicate the possibility to users...?

We could add a sentence in the "Spreading Religion" articles:

_This pressure can also be seen in the city screen, and gives you an idea of how religions in your cities will evolve if you don't do anything._ **You can also click on a religion icon to check its effects.**

SomeTroglodyte:
> either from the city screen or the city popup

It is, kind of. The content is the same except one flag - what was it - showMajority - only the wrapper is an expander within the city and it's in a plain table cell within the popup - originally a kludge to let you see religion status for cities not belonging to you (until a better idea came along - I had a variant running where it was the same expander but bottom left in the "unit info", but that had more hacky compromises and thus was more prone to regressions)

> Tutorial

OK, yea, not everybody reads those, but there should be some mention there.

PR #6468, opened 2022-03-30
Move UncivServer to own module (and jar)

SomeTroglodyte:
Might solve #6458 

This even produces independent packr'ed releases, though to get that to run without errors I had to make `fun String.runCommand` not throw on exit code 255. :desktop:packr* still does the same - fail after producing a valid zip. Might want to revert that part, maybe move the harcoded name to BuildConfig.serverName, maybe...

Android Studio users will **absolutely need** to do a Build/clean and resync Gradle, preferrably with a 'Invalidate Caches and Restart' in between.

This comments out the workarounds for the ClassNotFoundExceptions from `Char(Int)` and `Char.code` and leaves the workaround functions as a stub. In my tests these problems were gone with the server separation. I hope we'll all be soon convinced these can go permanently, to be done in a separate PR.

I urge you to go over the build processes and fix what I may have bungled or avoided to touch at all (will the new jar appear in github releases??)

That `println` in `get("/isalive")` is a glitch - I forgot to stash it at some point, But it's not harmful either.

Oh, and I snuck in a new hint on how to run Unit Tests from Studio. Afaik well-known by many but not yet in our wiki.

Issue #5725, opened 2021-11-29
Game crashes whenever switching soundtracks


NoviceCoder2000:
**Platform**
Windows 10, 64 Bits

**Version**
3.18.15

**Describe the bug**
There's a chance that the game will crash whenever switching soundtracks/in-game music, within the options. Other than that, there's nothing much to report.

**Expected behavior**
Game crashes.

SomeTroglodyte:
Current: `const val appVersion = "3.18.7"` - You got 3.18.15 via time machine?
Jokes aside - console log please? All Gdx.Music calls are try/catch wrapped, but some of the OpenAL failures were very hard to catch back when I wrote the MusicController. I thought I ended up with something that at worst does a long pause then moves on to another track...

Newexae:
I have a similar bug with my Rpi 4 where the game crash itself. According to the log, it is due to the music.

Exception playing music: Unable to allocate audio buffers. AL Error: 40963
Exception playing music: Error reading audio data.
Exception in thread "main" com.badlogic.gdx.utils.GdxRuntimeException: Error reading audio data.
	at com.badlogic.gdx.backends.lwjgl3.audio.Mp3$Music.read(Mp3.java:90)
	at com.badlogic.gdx.backends.lwjgl3.audio.OpenALMusic.fill(OpenALMusic.java:253)
	at com.badlogic.gdx.backends.lwjgl3.audio.OpenALMusic.update(OpenALMusic.java:237)
	at com.badlogic.gdx.backends.lwjgl3.audio.OpenALLwjgl3Audio.update(OpenALLwjgl3Audio.java:253)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:132)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:116)
	at com.unciv.app.desktop.DesktopLauncher.main(DesktopLauncher.kt:61)
Caused by: javazoom.jl.decoder.BitstreamException: Bitstream errorcode 102
	at javazoom.jl.decoder.Bitstream.newBitstreamException(Bitstream.java:418)
	at javazoom.jl.decoder.Bitstream.readFrame(Bitstream.java:329)
	at com.badlogic.gdx.backends.lwjgl3.audio.Mp3$Music.read(Mp3.java:67)
	... 6 more

HaneulCheong:
[...]

PR #6499, opened 2022-04-06
Harden Gdx Music and simplify MusicController

SomeTroglodyte:
Please everybody at all curious, _please_ read the code. I'd love to hear any input.

Result from #5725 research - not final, works, but shown as learning example only. Contains #6468 just so I can debug at all.

Please see commit history - each is one major step. As I said the first two are not related.

- Harden that part of Audio (desktop only) that runs from the main application loop where Exceptions are not catchable - and makes them catchable forcibly, using reflection to replace the Audio instance Gdx places there with a subclass that does the catching for us. Nice piece of work if I say so myself, and without it I wouldn't have reached any satisfactory conclusion. I'm still debating whether to use this in the next branch to solve the music issues, just for backup safety.
- Next - remove the timer from MusicController, thereby also removing the fadein/fadeout and pause between tracks capability. This reduced the frequency of the (now logged and recovered from) music crashes, but not by much. It now became clearer which actions provoked the instabilities, so some tweaks trying to limit some stuff went in against click-spamming the now playing label or the repeated loads provoked by quickly raising volume from zero.
- Lastly I had enough and _had_ to conclude the whole OpenAL cardhouse with virtual method calls on long vars through JNI just isn't thread-safe _at all_ (the Android backend _is_) - I removed the threaded background loader and - poof - no more crashes logged. Just like that. Of course you can just barely notice the slight UI stuttering on track changes, but oh well.

So I'll drop this and start from scratch _only_ un-threading the thing and test again. I won't be listening in soon for over two weeks - some RL to do over easter. So if I don't present a fix PR soon, go ahead try your hand.

Question: Opinions on leaving the `HardenGdxAudio` class in? It would then only serve the function to guard against crashes resulting from actual corrupted music files which survive the initial play() call.


SomeTroglodyte:
.. okay, `override fun newAudioDevice` in `class HardenedGdxAudio` is garbage leftover from trying to read underlying sources, ignore that.

PR #6500, opened 2022-04-07
Make uniques.md more friendly

heipizhu4:
I can't understand why all uniques examples with a [number] are filled with 20. It's illogical sometimes. Such as [20] sight. 
And it sometimes confuses modding beginners like '[20]% Strength'. Does it mean +20% strength or -80% strength?

(Still in progress)

### DIFF
--- a/docs/Modders/uniques.md
+++ b/docs/Modders/uniques.md
@@ -655,17 +655,17 @@ Simple unique parameters are explained by mouseover. Complex parameters are expl
 	Applicable to: Global
 
 ??? example  "[amount] Movement"
-	Example: "[20] Movement"
+	Example: "[3] Movement"
 
 	Applicable to: Global, Unit
 
 ??? example  "[amount] Sight"
-	Example: "[20] Sight"
+	Example: "[5] Sight"
 
 	Applicable to: Global, Unit, Terrain
 
 ??? example  "[amount] Range"
-	Example: "[20] Range"
+	Example: "[3] Range"
 
 	Applicable to: Global, Unit
 


ymorgenstern:
That file is autogenerated by UniqueDocsWriter, and 20 is a number that fits in most of the uniques.
I haven't found a good way to programatically differentiate between uniques where 20 would fit vs 3

SomeTroglodyte:
```
enum class UniqueParameterType(
    var parameterName:String,
    val displayName: String = parameterName
) {
    Number("amount") {
                UniqueType.UniqueComplianceErrorSeverity? {
            return if (parameterText.toIntOrNull() == null) UniqueType.UniqueComplianceErrorSeverity.RulesetInvariant
            else null
        }
    },
```

SomeTroglodyte:
@yairm210  Maybe 
```kotlin
enum class UniqueParameterType(
    var parameterName: String,
    val docExample: String,
    val displayName: String = parameterName
) {
    Number("amount", "3") {
                UniqueType.UniqueComplianceErrorSeverity? {
            return if (parameterText.toIntOrNull() == null) UniqueType.UniqueComplianceErrorSeverity.RulesetInvariant
            else null
        }
    },

    RelativeNumber("relativeAmount", "+20") {
                UniqueType.UniqueComplianceErrorSeverity? {
            return if (parameterText.toIntOrNull() == null) UniqueType.UniqueComplianceErrorSeverity.RulesetInvariant
            else null
        }
    },
```
, move that stuff from replaceExamples to that new docExample property, patch the UniqueTypes where +20 sounds better to use "relativeAmount", replace the `replaceExamples` call with 
```kotlin
                if (uniqueType.parameterTypeMap.isNotEmpty()) {
                    val paramExamples = uniqueType.parameterTypeMap.map { it.map { pt -> pt.docExample }.joinToString { "/" } }.toTypedArray()
                    lines += "\tExample: \"${uniqueType.placeholderText.fillPlaceholders(*paramExamples)}\"\n"
                }
```
 etc etc. Hmm... String.fillPlaceholders, String.getPlaceholderParameters and so on are often wasteful as pre-processed param data is available in the context they're used in in the majority of cases. I also noticed the "we can override them manually" comment on parameterTypeMap which is at the moment technically untrue. To make it true...
```kotlin
    /** For uniques that have "special" parameters that can accept multiple types, we can override them manually
     *  For 95% of cases, auto-matching is fine. */
    val parameterTypeMap = parameterTypeMapInitializer()
    open fun parameterTypeMapInitializer(): ArrayList<List<UniqueParameterType>> {
        val map = ArrayList<List<UniqueParameterType>>()
        for (placeholder in text.getPlaceholderParameters()) {
            val matchingParameterTypes = placeholder
                .split('/')
                .map { UniqueParameterType.safeValueOf(it.replace(numberRegex, "")) }
            map.add(matchingParameterTypes)
        }
        return map
    }
    val targetTypes = HashSet<UniqueTarget>()

    init {
        targetTypes.addAll(targets)
    }
```
Just food for thought.

PR #6504, opened 2022-04-07
New nation icons

ravignir:
Old icons resemble Firaxis' icons a bit too much.


SomeTroglodyte:
Some before/after might help, and - you sure sprites don't work better if they have some transparent border pixels left? I usually work from 480x480 then expand canvas to 500x500 and scale down. leaves 1 px transparent for 50x50 result or 2 for the typical 100x100...

yairm210:
I personally don't see a problem with the old icons, and the similarity helps Civ V players to instantly recognize them.
I'm not sure this is an improvement,

ravignir:
> Some before/after might help, and - you sure sprites don't work better if they have some transparent border pixels left? I usually work from 480x480 then expand canvas to 500x500 and scale down. leaves 1 px transparent for 50x50 result or 2 for the typical 100x100...

i've started with 400x400, into 200x200, then 100x100. I don't think I have my original images anymore tho.

> I personally don't see a problem with the old icons, and the similarity helps Civ V players to instantly recognize them. I'm not sure this is an improvement,

I've made this PR just in case it was a problem.

SomeTroglodyte:
I just mentioned because - if you look at the atlas texture you get what looks like artifacts on those originals without a transparent border, and they are absent from the originals with such a border:
image
Yea, these artifacts are on purpose and somehow needed, but I don't trust that entirely. Got a prejudice when that scales and is antialiased onto some background, then the others might look a tad better.

yairm210:
I think I'll close this as unneeded.
I do want to say that I *really* appreciate the work, even though it's not going in the game - this and that are two completely different things :)

PR #6506, opened 2022-04-07
Added borders to minimap

yairm210:
image

Also general cleanup of the file, the new stuff is in lines 167-198

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/Minimap.kt
+++ b/core/src/com/unciv/ui/worldscreen/Minimap.kt
@@ -179,7 +179,7 @@ class Minimap(val mapHolder: WorldMapHolder, minimapSize: Int) : Group(){
                 
                 val hexagonEdgeLength = tileImages.tileHexagonImage.width/2
 
-                borderImage.setSize(hexagonEdgeLength, hexagonEdgeLength/10)
+                borderImage.setSize(hexagonEdgeLength, hexagonEdgeLength/4)
                 borderImage.setOrigin(Align.center)
                 val hexagonCenterX = tileImages.tileHexagonImage.x + tileImages.tileHexagonImage.width/2
                 borderImage.x = hexagonCenterX - borderImage.width/2
--- a/core/src/com/unciv/ui/worldscreen/Minimap.kt
+++ b/core/src/com/unciv/ui/worldscreen/Minimap.kt
@@ -17,21 +17,22 @@ import com.unciv.logic.civilization.CivilizationInfo
 import com.unciv.logic.map.MapShape
 import com.unciv.logic.map.MapSize
 import com.unciv.logic.map.TileInfo
-import com.unciv.ui.tilegroups.ActionlessGroup
-import com.unciv.ui.utils.IconCircleGroup
-import com.unciv.ui.utils.ImageGetter
-import com.unciv.ui.utils.onClick
-import com.unciv.ui.utils.surroundWithCircle
+import com.unciv.ui.utils.*
+import kotlin.math.PI
+import kotlin.math.atan
 import kotlin.math.max
 import kotlin.math.min
 
 class Minimap(val mapHolder: WorldMapHolder, minimapSize: Int) : Group(){
     private val allTiles = Group()
-    private val tileImages = HashMap<TileInfo, Image>()
+    class MinimapTileImages(val tileHexagonImage:Image) {
+        var cityCircleImage: IconCircleGroup? = null
+        var owningCiv: CivilizationInfo? = null
+        var neighborToBorderImage = HashMap<TileInfo,Image>()
+    }
+    private val tileinfoToImages = HashMap<TileInfo, MinimapTileImages>()
     private val scrollPositionIndicators = ArrayList<ClippingImage>()
-
-    private val cityIconsGroup = ActionlessGroup()
-
+    
     init {
         isTransform = false // don't try to resize rotate etc - this table has a LOT of children so that's valuable render time!
 
@@ -68,7 +69,7 @@ class Minimap(val mapHolder: WorldMapHolder, minimapSize: Int) : Group(){
                 mapHolder.setCenterPosition(tileInfo.position)
             }
             allTiles.addActor(hex)
-            tileImages[tileInfo] = hex
+            tileinfoToImages[tileInfo] = MinimapTileImages(hex)
 
             topX = max(topX, hex.x + groupSize)
             topY = max(topY, hex.y + groupSize)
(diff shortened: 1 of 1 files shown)


yairm210:
image

Heavier borders give a much better feel actually

SomeTroglodyte:
Looks good...
- I don't see any write to MinimapTileImages.owningCiv?
- having a full TileInfo as key to neighborToBorderImage seems a little heavy-weight. Since we don't do them ourselves, hashCode likely calculates hashCode for each and every field and does its powers-of-31 thing with them, and equals will probably also get an expensive default impelementation. Vector2 hashCode is likely cheaper.

yairm210:
Since tileInfo is a regular class and not a data class, it's hash is based on its physical location, so it's actually a lot simpler than that

SomeTroglodyte:
True, the powers of 31 thing is a `data class` thing, and otherwise it falls through to java.lang.Object which says:
<details><summary>Jdoc</summary>

```java
    /**
     * Returns a hash code value for the object. This method is
     * supported for the benefit of hash tables such as those provided by
     * {@link java.util.HashMap}.
     * <p>
     * The general contract of {@code hashCode} is:
     * <ul>
     * <li>Whenever it is invoked on the same object more than once during
     *     an execution of a Java application, the {@code hashCode} method
     *     must consistently return the same integer, provided no information
     *     used in {@code equals} comparisons on the object is modified.
     *     This integer need not remain consistent from one execution of an
     *     application to another execution of the same application.
     * <li>If two objects are equal according to the {@code equals(Object)}
     *     method, then calling the {@code hashCode} method on each of
     *     the two objects must produce the same integer result.
     * <li>It is <em>not</em> required that if two objects are unequal
     *     according to the {@link java.lang.Object#equals(java.lang.Object)}
     *     method, then calling the {@code hashCode} method on each of the
     *     two objects must produce distinct integer results.  However, the
     *     programmer should be aware that producing distinct integer results
     *     for unequal objects may improve the performance of hash tables.
     * </ul>
     * <p>
     * As much as is reasonably practical, the hashCode method defined
     * by class {@code Object} does return distinct integers for
     * distinct objects. (The hashCode may or may not be implemented
     * as some function of an object's memory address at some point
     * in time.)
     *
     * @return  a hash code value for this object.
     * @see     java.lang.Object#equals(java.lang.Object)
     * @see     java.lang.System#identityHashCode
     */
    @HotSpotIntrinsicCandidate
    public native int hashCode();
```
</details>


"may or may not be implemented as some function of an object's memory address" - now that I found this official juicy tidbit I hopefully will remember.

SomeTroglodyte:
I just tested this...
- My doubt concerning no write to MinimapTileImages.owningCiv seems justified. Removing the field and making the reads essentially null tests - looks exactly the same.
- viewEntireMapForDebug won't show you any of the stuff in tileinfoToImages in the unexplored regions, only hex colors.
- Some nations work far better than others. e.g. Russia: average, Austria, Korea, Spain: excellent, Maya and Japan: below average, Greece: practically invisible. OK, that's an archipelago map so Greece and its blue hues has the disadvantage obviously. I have no idea how some leveling of the field could be done for this effect.
- The "camera" rectangle(s) are drawn _below_ these borders (and the cities). That looks slightly off, for cities it's almost OK, for borders less so. The actors aren't in the same container as these here, so zIndex after the update won't help, and changing their parent conflicts with their coords calculation and clipping... Hmmm...
- Imho making the city circles bigger looks better, and making capitals even bigger than the others actually helps. Near line 155 (lacking x/y correction so it would as is shift the cities a little to top right):
```kotlin
                val size = (if (tileInfo.getCity()!!.isCapital() && tileInfo.getOwner()!!.isMajorCiv()) 1.667f else 1.25f) * hex.width
                val nationIcon= ImageGetter.getCircle().apply { color = nation.getInnerColor() }
                    .surroundWithCircle(size, color = nation.getOuterColor())
```
- Water inside borders is never touched, but _some_ nation hue... very much a matter of taste, but I tend to like it. Could use some tileInfo.getOwner caching anyway, so I tried code like this (also affects viewEntireMapForDebug):

<details>

```kotlin
    fun update(cloneCivilization: CivilizationInfo) {
        val viewEntireMapForDebug = UncivGame.Current.viewEntireMapForDebug
        fun getTileColor(tileInfo: TileInfo): Color {
            val owner = tileInfo.getOwner() ?: return tileInfo.getBaseTerrain().getColor().lerp(Color.GRAY, 0.5f)
            return when {
                tileInfo.isCityCenter() && !tileInfo.isWater -> owner.nation.getInnerColor()
                !tileInfo.isWater -> owner.nation.getOuterColor()
                else -> tileInfo.getBaseTerrain().getColor().lerp(owner.nation.getOuterColor(), 0.333f)
            }
        }
        for ((tileInfo, tileImages) in tileinfoToImages) {
            if (!(viewEntireMapForDebug || cloneCivilization.exploredTiles.contains(tileInfo.position))) {
                tileImages.tileHexagonImage.color = Color.DARK_GRAY
                continue
            }
            tileImages.tileHexagonImage.color = getTileColor(tileInfo)
            ....
```
</details>

SomeTroglodyte:
Here's how nation-tinted water and fat cities look like in comparison (camera disabled):
Minimap
You see some edges flipping, which are edges at the border between two nations, not between a nation's territory and unowned land, often but not always when a mountain is near. No idea why the decision which nation's color wins isn't deterministic, but that's it: they do compete, they don't stay each inside their own "side" of the edge.

All this just food for thought. No critical bugs, count my test results as approval.

SomeTroglodyte:
... could use a comment for the `/ relativeWorldPosition.x` for the atan call - that only never crashes (divide by zero) because our hexagons are "lying down" not "standing on an apex".

yairm210:
I like the city sizes and I'll be taking that, but the water tinting in my eyes make it less immediately visible where the water of the world is, because I need to readjust to each tinted water

SomeTroglodyte:
:smile_cat:  As I said, matter of taste. But I think my sample code wasn't complete? Is there a `shift` val? A moment...

SomeTroglodyte (approved):

Ouaz:
I really like that, except the tinted water, that is too much confusing.
Otherwise, it really succeeds to show a civ territory in a glimpse (Land and water tiles). Thanks!

SomeTroglodyte:
Now that it's in, kudos, definite improvement. To give what I remember as remaining problems a place, this..
<details><summary>..screenshot collage:</summary>

image
</details>
That's enlarged minimap and main map scaled to show the same square. On can see two details:

- As mentioned the 'camera' rectangle is below the border/city layer
- Borders nation-nation are on top of one another and one 'wins' hiding the other. Drawing them 'in' by a few percent towards 'their' hex center works partially, I couldn't get the joints nice without going into the whole concave/convex overhead the main map borders do.

Just food for thought as always, I'll shut up now.

PR #6518, opened 2022-04-10
Update Filipino.properties

shannaurelle:
New translations for Filipino


shannaurelle (commented):
All from the new version should be copied

xlenstra:
You seem to have conflicts with the current version, are you able to resolve these?

PR #6526, opened 2022-04-11
MusicController - Version 5

### LINKED ISSUE #5725: Game crashes whenever switching soundtracks

SomeTroglodyte:
This is my latest take on how to fix the problems first mentioned on the lwjgl3 PR (#5614), hopefully fixes #5725, closes #6499.

I called the commit "Can catch exceptions from Gdx.app.loop and replaces Timer on desktop with a callback from OpenALLwjgl3Audio.update" - that's the tech shorthand. In the meantime I coded, heavily tested and rejected three other branches with different concepts - all without the brutal reflection hooks. One did away with all background processing and used `setOnCompletionListener` to chain tracks. Best test result - survived sitting there for over 10 hours before it crashed anyway (with the div/0 in line 904 of javazoom.jl.decoder.LayerIIIDecoder). Also had noticeably more "choppiness" during next turn or changing screens. One tried to move all Music calls to one common background thread with its own postRunnable system - fail. One tried to use Gdx's postRunnable to get the few calls that couldn't be made to run on the main thread by other means over to that safe thread - fail, too.

So I went back to the Big Fat Hammer and polished it.

This has so far run music from a jar (on linux+openjdk11) for hours, and seen extensive Android hardware (LOS16) and AVD (only the Q one as it's so much faster) tests. Runs from the Studio debugger have been much shorter.

Sorry it hasn't become much simpler but more complicated, but it's the nicest _and_ most stable MusicController I have gotten so far.

New details:
- thatched finally renamed so it will have a chance to play even when mods having "Ambient" suffixed tracks are active. Name is also nicer for the 'currently playing' label.
- Thatched can still be downloaded when one installs music mods before using "that button". You can even download while mod music is playing and when it's done it will fade-over nicely. Maybe rename the button in time to something like "Download _default_ music" or "Download classic Unciv theme"?
- Fade-overs where one track fades out at the same time the replacement is fading in are now unblocked and tested. Previously, I had taken out a tiny enabling line which meant fade-in was done after fade-out finished. Fade-over working means I could extend the default fade time a little, sounds cool.
- The manual track change from options now uses default fade speed, not the 3s it did before.

If you really want a far simpler MusicController, the last two commits of #6499 would be a good start, plus disabling mp3 music support (!), then redoing the fade capability as you said with Gdx Actions - hoping setVolume (actually `alSourcef(..AL_GAIN..)`) isn't one of the AL10 functions with thread-safety problems - or Action code always runs on the main thread. This one was easier for me as I already had most elements ready somewhere, and the massive hours invested were mostly testing and monitored runs in the debugger.

### DIFF
--- a/core/src/com/unciv/ui/utils/AudioExceptionHelper.kt
+++ b/core/src/com/unciv/ui/utils/AudioExceptionHelper.kt
@@ -1,0 +1,10 @@
+package com.unciv.ui.utils
+
+import com.badlogic.gdx.audio.Music
+
+interface AudioExceptionHelper {
+    fun installHooks(
+        updateCallback: (()->Unit)?,
+        exceptionHandler: ((Throwable, Music)->Unit)?
+    )
+}
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -37,6 +37,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
     val consoleMode = parameters.consoleMode
     val customSaveLocationHelper = parameters.customSaveLocationHelper
     val limitOrientationsHelper = parameters.limitOrientationsHelper
+    private val audioExceptionHelper = parameters.audioExceptionHelper
 
     var deepLinkedMultiplayerGame: String? = null
     lateinit var gameInfo: GameInfo
@@ -97,6 +98,10 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
         settings = GameSaver.getGeneralSettings() // needed for the screen
         screen = LoadingScreen()  // NOT dependent on any atlas or skin
         musicController = MusicController()  // early, but at this point does only copy volume from settings
+        audioExceptionHelper?.installHooks(
+            musicController.getAudioLoopCallback(),
+            musicController.getAudioExceptionHandler()
+        )
 
         ImageGetter.resetAtlases()
         ImageGetter.setNewRuleset(ImageGetter.ruleset)  // This needs to come after the settings, since we may have default visual mods
--- a/core/src/com/unciv/UncivGameParameters.kt
+++ b/core/src/com/unciv/UncivGameParameters.kt
@@ -2,6 +2,7 @@ package com.unciv
 
 import com.unciv.logic.CustomSaveLocationHelper
 import com.unciv.ui.crashhandling.CrashReportSysInfo
+import com.unciv.ui.utils.AudioExceptionHelper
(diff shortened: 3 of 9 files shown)


yairm210:
I'll be honest my brain zoned out while trying to read this :/
But I trust you enough and see that you put a ridiculous amount of effort into this so I'll take your word for it

core/src/com/unciv/ui/utils/AudioExceptionHelper.kt
```diff
@@ -0,0 +1,10 @@
+package com.unciv.ui.utils
+
+import com.badlogic.gdx.audio.Music
+
+interface AudioExceptionHelper {
+    fun installHooks(
+        updateCallback: (()->Unit)?,
+        exceptionHandler: ((Throwable, Music)->Unit)?
+    )
+}
```
xlenstra, line 10: Would it make more sense to place this file in `/ui/audio/` instead of `/ui/utils/`?

xlenstra:
As I haven't worked with/understood the previous code for the Music Player I have no idea how this relates and whether it could fix it, so I don't see myself giving in depth feedback on this, though I trust you that this should be fine

core/src/com/unciv/ui/utils/AudioExceptionHelper.kt
SomeTroglodyte: True. I was halfway between pre-reorg and the idea all platform-dependent-helpers need to be grouped into their own package (still well distributed) - there's a word in psy for it - ya know, a predator threatens a cock potentially defending his hens, and he can't decide whether to attack or run, so he starts picking grains instead?

SomeTroglodyte:
How about testing [the jar of this](https://www.dropbox.com/s/h9kat1m0ic5dz6h/Unciv.jar?dl=0)? Run it, ensuring it sees the assets folder you want it to, giving it plenty of music, and letting it run playing muzak a few hours on your ubuntu-based distros? You know the old code won't make an hour.

SomeTroglodyte:
My test has now run >24 hours playing music, stable. Memory and "Disk read" have quickly converged to constants, with (Memory - Disk read) staying perfectly constant over the run duration - which to me means no leaks and the java runtime caches those files. I even had two of them running at the same time for quite a while - nice cacophony.

PR #6530, opened 2022-04-11
fix Byzanty intrduction translation error

vankos:
delete redundant comma for Byzanty intrduction:
aqueducts, bridges, and churches -> aqueducts, bridges and churches


SomeTroglodyte:
Sorry, this is not the way that stuff works. You're changing the "input" side of a russian translation where you're supposed to change only the "output" side. That change would simply disappear along with the russian side and reappear with the extra comma and a '#requires translation' the next time the generator runs.

English text needs to be fixed at its _source_ first, and in this case that's [Nations.json](https://github.com/yairm210/Unciv/blob/84ca296a5c0bead8e3db9eb4aa20a9034cfff98b/android/assets/jsons/Civ%20V%20-%20Gods%20%26%20Kings/Nations.json#L840). And once that change goes by the translation generator, _all_ existing translations of that line will disappear and be replaced by the changed one + '#requires translation'.

PR #6533, opened 2022-04-12
Air Units take damage in Combat

itanasi:
Air Units now take damage during combat.
Edits to both takeDamage() calculation and calculateDamageToAttacker()
image

Other changes:
- [x] Interception uses highest chance unit to Intercept
- [x] Interception by defending unit not possible
- [x] Interception grants 2 XP
- [x] Air units can't Intercept if they've moved that turn

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -221,7 +221,7 @@ object Battle {
 
         if (defender is MapUnitCombatant && defender.unit.isCivilian() && attacker.isMelee()) {
             captureCivilianUnit(attacker, defender)
-        } else if (attacker.isRanged()) {
+        } else if (attacker.isRanged() && !attacker.isAirUnit()) {  // Air Units are Ranged, but take damage as well
             defender.takeDamage(potentialDamageToDefender) // straight up
         } else {
             //melee attack is complicated, because either side may defeat the other midway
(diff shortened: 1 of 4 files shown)


xlenstra:
> Is this the way it works in Civ V?

Looking through the code of the original, there are different `Attack` and `AttackAir` functions, as well as `getMeleeCombatDamage` and `GetAirCombatDamage` functions, which at first glance have different calculations. So probably not.
`GetAirCombatDamage` is probably the most relevant function for this change, which can be found in `/CvGameCoreDLL_Expansion1/CvUnit.cpp:9369`

itanasi:
So a few subtle differences.
1) Air units actually get penalized double for lower HP when attacking. And Intercept damage is applied first (obv)
2) Base damage is lower (250 vs 400)
3) Variance is lower (300 vs 400)
4) Defender for Air uses Ranged Combat Strength (except for WATER units)

itanasi:
Oh, even worse. The equations are different for Air Strike DefenderDamageInflicted. Everything reduced (200 base, 200 variance)

itanasi:
Code seems different in the C++ SDK. But testing it in game sure seems like Air Combat damage calc is very similar to Melee.
image
Dumped some expected combat data, all at full HP, into a plot
image
Interestingly the City Bombard ratio seems a bit different. Centered more at 25 than 30

itanasi:
Found the proper XML, and Air, Melee, and Ranged Combat ass use the same 24 base 12 potential extra dmg equations. So this should match behaviors

xlenstra:
This screenshot does seem to be made in BNW instead of G&K, could that explain the difference?

xlenstra (approved):

PR #6543, opened 2022-04-14
Updating XP Rewards. Interception now gives XP

itanasi:
Add isAirUnit() Boolean for ICombatant
Not including AirSweep (6XP for attacker if draw Air Intercept, 6XP for Air Defending Intercept, No XP for Ground Intercept Defending) yet

Interception gives 2XP if deals dmg
Ranged Attack on City is 3XP (was 2XP)
Air Attacks are 4XP. Defender gets 2XP

Interesting note. Evasion in the Civ5 code is actually chance to not get Intercepted, not a damage reduction. So UnCiv will give out extra Intercept XP


xlenstra (approved):

itanasi:
Goofed the Melee XP. Will revisit

PR #6545, opened 2022-04-14
Update Simplified_Chinese translation

heipizhu4:
Can anyone tell me what is 'WLTK+'? I've played a new game and didn't find it.

### DIFF
--- a/android/assets/jsons/translations/Simplified_Chinese.properties
+++ b/android/assets/jsons/translations/Simplified_Chinese.properties
@@ -732,10 +732,8 @@ Sleep until healed = 休眠直至完全回复
 Moving = 移动中
 Set up = 架设
 Paradrop = 降落伞
- # Requires translation!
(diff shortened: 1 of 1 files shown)


xlenstra:
> Can anyone tell me what is 'WLTK+'? I've played a new game and didn't find it.

It appears to be used internally for the "We Love The King Day" label in the resource overview, but at first glance it doesn't seem to be ever shown to the user

SomeTroglodyte:
> 'WLTK+'?

When you switch the resource overview to vertical layout, to keep columns narrow. Chinese will probably look just fine with clear text. There's a tooltip - should be next to it - to explain a bit better. What I forgot is to check whether translators can disable a tooltip completely - no, I guess not, you can't translate to a zero-length string. If desired, we could map "" to the empty string then check whether such a tooltip is intelligent enough to omit the background and be really invisible...

PR #6390, opened 2022-03-20
Add Moddable Policy Priorities

HaneulCheong:
With this PR, modders would be able to implement custom policy priorities with the following syntax:
```json
[
    ". . .",
    {
        "name": "Freedom",
        ". . .",
        "priorities": {
            "Neutral": 10,
            "Cultural": 10,
            "Diplomatic": 10,
            "Domination": 10,
            "Scientific": 10
        },
        ". . ."
    },
    ". . ."
]
```

- **Fixed G&K policy eras**
  - Patronage should be unlocked after the Medieval era
  - Freedom should be unlocked after the Industrial era
  - Source: [Fandom Wiki](https://civilization.fandom.com/wiki/Social_policies_(Civ5))
- **Added 'priorities' object to policy branch entries**
  - [x] TODO: Write about this change to the official wiki - efd6ad7e00676adc634cc1147016a454373014fd
  - [x] TODO: Set actual priorities for each branch, both in the Vanilla and G&K - 83f822985abbe4f17d28e4373162abd378eb3dd6
- **Fixed Vanilla policy eras**
  - Freedom should be unlocked after the Renaissance era
  - Source: [Fandom Wiki](https://civilization.fandom.com/wiki/Social_policies_(Civ5))
- **Partial code formatting**
  - Followed the Kotlin Style Guide...as far as I can tell.
  - I don't know most of the work was done by Android Studio.
- **Reworked how the AI chooses a policy to newly adopt**
  ```markdown
  # Branch-based policy-to-adopt decision
  Basically the AI prioritizes finishing incomplete branches before moving on, \
  unless a new branch with higher priority is adoptable.
  
  - If incomplete branches have higher priorities than any newly adoptable branch,
    - Candidates are the unfinished branches.
  - Else if newly adoptable branches have higher priorities than any incomplete branch,
    - Candidates are the new branches.
  - Choose a random candidate closest to completion.
  - Pick a random child policy of a chosen branch and adopt it.
  ```
- **Removed debugging codes**
- **Update Civilization-related-JSON-files.md**
- **Assigned actual priorities to each branch**
  - See https://github.com/yairm210/Unciv/pull/6390#issuecomment-1073784701
  - Also fixed a debugging value in `Ruleset.kt` (-1 -> 0)
- **Merge branch 'master' into policy-ai**
  - ...
- **Merge branch 'master' into policy-ai**
  - ...

PR #6552, opened 2022-04-15
Fixed vanilla policies regression

xlenstra:
Should now contain the vanilla policies instead of the G&K policies (regression from #6390)

### DIFF
--- a/android/assets/jsons/Civ V - Vanilla/Policies.json
+++ b/android/assets/jsons/Civ V - Vanilla/Policies.json
@@ -217,16 +217,16 @@
             "Scientific": 0
         },
         "uniques": [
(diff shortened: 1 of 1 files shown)


heipizhu4 (changes requested):
image
I've just checked the wiki, and 'tradition complete' in vanilla only gives [+2] food in all cities, no additional growth is provided. (line65)

xlenstra:
From your screenshot:
> Adopting all policies in the Tradidition tree will grant a +15% Growth in all cities and:

Also, according to my local copy:
image

xlenstra:
> Vanilla doesn't provide the unique in line 550

Fixed

heipizhu4 (approved):

PR #6564, opened 2022-04-18
Add checks against 0 Strength combat

itanasi:
Adding checks against weird cases with 0 Str Defender or Attacker

Possibly not actually needed?


xlenstra:
Any unit that has 0 strength is automatically declared a civilian (this is actually the definition), so I don't think these are needed?

yairm210:
Agree with xlenstra, this *shouldn't* be needed, it what situation did you encounter this?

itanasi:
Someone in the Discord somehow got this to happen after modifiers. So some sort of crazy mod that set strength to 0.

Agree, it is very edge case and perhaps we just rule that people shouldn't have modifiers that set strength to 0.

xlenstra:
Like -100% strength or something? I'd say we should just cap the strength after modifiers to be at least 1

Issue #6588, opened 2022-04-23
City should not starve when settler is being built


0avsz:
**Platform**
Operating system type, major version, architecture (64/32 bit).

Android

**Version**
Where you got the game and the exact version number.

4.0.6

**Describe the bug**

During settler construction, city should not starve despite having insufficient food.

**To Reproduce**
1. Go to city and build settler
2. Remove all food tiles and select prodution tiles
3. CIty wil starve

If game data is required, please add it here. From the menu, choose save game (or map), copy to clipboard, and paste here between the backtick blocks (or, if no such save is required, you may delete this):
<details><summary>Saved game</summary>

```
[...]

Issue #4408, opened 2021-07-07
About settlers


logicminimal:
Is this implemented?

"__While a city is producing Settlers, population growth is paused. This prevents a city from growing during Settler production but also prevents it from starving. Any surplus  Food in the city is converted into  Production at a diminishing rate, with a point of  Production granted at the first, second, fourth, and eighth surplus  Food and every fourth  Food onward.[1] This conversion process helps reduce the negative impact of Settler production in cities with more excess  Food, so that it is not always optimal to produce a Settler in a city with the most  Production and least excess  Food.__"

also when i settled a city in the future era i got 5 population when i settled it, is this how it is implemented in civ 5? i couldn't find anything about cities starting with more than 1 population

ravignir:
Yairm said no to this for whatever reason. 

Whatever reason means: food to conversion is too complicated for players to understand and preventing city from starving makes no sense in his opinion.

xlenstra:
> also when i settled a city in the future era i got 5 population when i settled it, is this how it is implemented in civ 5? i couldn't find anything about cities starting with more than 1 population

This is implemented in 3.15.8

> While a city is producing [...] least excess Food

The current implementation converts excess food to production on a 1-to-1 ratio, so 10 excess food is 10 extra production for settlers.

logicminimal:
> Yairm said no to this for whatever reason. 
> 
> Whatever reason means: food to conversion is too complicated for players to understand and preventing city from starving makes no sense in his opinion.

1:1 seems too high compared to the original game though,
"Not exactly 4:1 theres a ramp up conversion price making it potentially best in early game


-    food overflow/hammers
-    1/1
-    2-3/2
-    4-6/3
-    7/4

    +1h per 4 food beyond 7 (so at 11 15 19 etc)"

yairm210:
I have no problem changing the ratio, but it should be linear, and specified in a unique.
I'm closing this since, as indeed stated, the main point is one that I disagree with strongly. "My city is starving, maybe I should construct a settler, that will solve my problem" - statements dreamed up by the utterly Deranged.

logicminimal:
> I have no problem changing the ratio, but it should be linear, and specified in a unique.
> I'm closing this since, as indeed stated, the main point is one that I disagree with strongly. "My city is starving, maybe I should construct a settler, that will solve my problem" - statements dreamed up by the utterly Deranged.

what about just disabling no starvation while production

PR #6590, opened 2022-04-23
Correct calculation of the production for the Settlers

JackRainy:
The fix for #6588 

First of all, it fixes the calculation of the excessive food conversion to the production for the Settlers.
~Secondly, it avoids the city of starvation during the Settlers production:~ https://civilization.fandom.com/wiki/Settler_(Civ5)

Also, a minor fix for the caption in the city: when it produces the Settler there was no message about conversion of food since it was overridden by starvation/growing checks.

### DIFF
--- a/core/src/com/unciv/logic/city/CityStats.kt
+++ b/core/src/com/unciv/logic/city/CityStats.kt
@@ -64,8 +64,6 @@ class CityStats(val cityInfo: CityInfo) {
 
     var baseStatTree = StatTreeNode()
 
-    var statPercentBonusList = LinkedHashMap<String, Stats>()
-
     var statPercentBonusTree = StatTreeNode()
 
     // Computed from baseStatList and statPercentBonusList - this is so the players can see a breakdown
@@ -74,8 +72,6 @@ class CityStats(val cityInfo: CityInfo) {
     var happinessList = LinkedHashMap<String, Float>()
 
     var statsFromTiles = Stats()
-
-    var foodEaten = 0f
 
     var currentCityStats: Stats = Stats()  // This is so we won't have to calculate this multiple times - takes a lot of time, especially on phones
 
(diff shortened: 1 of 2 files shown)


xlenstra:
#4408 might be relevant as it discusses exactly this

xlenstra:
> #4408 might be relevant as it discusses exactly this

I was mostly referring to yairms comment, but I suppose the calculation would be an improvement as well

JackRainy:
> I was mostly referring to yairms comment

Yes, I see that. It is the same what itanasi refers in #6588 - the trick to avoid starvation seems unnatural.
But still would like to keep it closer to the original. Also, I do not believe it changes much since the starvation situation is quite rare.

SomeTroglodyte:
A little aside: At least one other, older Civ (not sure which if not all of CTP, C4, C3, C2, my guess would be C4) did indeed reduce the population by 1 the moment the settler was produced, and there were micromanagement tips on how to fiddle food so it grows back in 1 or 2 turns; and using that to manage starvation was normal and documented in guides as well (thanks to the odd yield problem most cities went into starvation sooner or later). That same Civ also gave you a Settler when you razed a city to dust, if my crumbling memory still serves...

JackRainy:
Following all the advises and having the short voting in Discord, I tend to revert the denying of starvation.
However, the correct calculation of the production conversion is still there, and the UI minor fix as well.

yairm210 (approved):

xlenstra:
Would it be better if we make the unique reflect these changes in formula? `Excess food partially converted to Production when under construction` or something? Otherwise I'm fine with this being merged in

yairm210:
I'm against changing the unique, I can't think of a way making it clearer, any changes would just make it more confusing :/
'partially' makes it sound like "maybe only half is converted? What's going on?"

xlenstra:
I mean, I can make the same argument for keeping it this way. Currently it sounds like all of it is converted (especially as that is also the old behaviour and the unique for converting surplus happiness to culture have the same wording), so when that doesn't happen it would also confuse other players :)

PR #6601, opened 2022-04-25
Typed all belief automation uniques

xlenstra:


SomeTroglodyte (approved):
Looks straightforward enough to not need my testing.

I'd just say: `if(hasUnique || hasUnique)` will use `getMatchingUniques` which is a `Sequence` builder so if you'd cache the "recipe" of the two concatenated sequences before calculating `statsGranted`, an `any()` on that would be the exact same `if` condition with less allocations/constructors called. You only really need the `if` because the `reduce` would get sick given an empty iterable (`fold` would be immune)... No I don't mean eliminate the test, the rest of the former loop body justifies it anyway.

By the way, I suspect having an overload of `getMatchingUniques` with a vararg UniqueType parameter would allow to simplify quite a number of places in Unciv... Except a caller would need to use named arguments for the following parameters, and the change would need to trickle through all RulesetObjects, and maybe isOfType and hasUnique too - nah, too extensive.

xlenstra:
> Looks straightforward enough to not need my testing.
> 
> I'd just say: `if(hasUnique || hasUnique)` will use `getMatchingUniques` which is a `Sequence` builder so if you'd cache the "recipe" of the two concatenated sequences before calculating `statsGranted`, an `any()` on that would be the exact same `if` condition with less allocations/constructors called. You only really need the `if` because the `reduce` would get sick given an empty iterable (`fold` would be immune)... No I don't mean eliminate the test, the rest of the former loop body justifies it anyway.
> 
> By the way, I suspect having an overload of `getMatchingUniques` with a vararg UniqueType parameter would allow to simplify quite a number of places in Unciv... Except a caller would need to use named arguments for the following parameters, and the change would need to trickle through all RulesetObjects, and maybe isOfType and hasUnique too - nah, too extensive.

Sounds interesting, might do something with that later

PR #6604, opened 2022-04-25
Display wonders built in other civs correctly

JackRainy:
When you enter the era (e.g. Ancient), usually you see the whole list of the wonders with "Not built" status.
As soon as the game goes, some civilizations build the Wonders and they get assigned to the civilization that builds them or ... just disappear from the list as "Unknown", if they built by unknown civiliazation.

Screenshot 2022-04-25 17-32-18

That's a bug I am fixing right here.

Screenshot 2022-04-25 18-24-20

Now we know it is built (in the far far lands) but it is not found where it is yet.


SomeTroglodyte (approved):
And I thought I had tested that stuff back then... :+1:

PR #6606, opened 2022-04-25
Use both icon and roundIcon properties

### LINKED ISSUE #6587: Icon is missing on F-Droid
The app icon is missing on F-Droid.
photo_2022-04-22_21-24-45

JackRainy:
Here is my assumption how to fix #6587

See also: https://developer.android.com/about/versions/nougat/android-7.1.html#circular-icons

> When a launcher requests an app icon, the framework returns either android:icon or android:roundIcon, depending on the device build configuration. Because of this, apps should make sure to define both android:icon and android:roundIcon resources when responding to launcher intents.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -13,6 +13,7 @@
     <application
         android:allowBackup="true"
         android:icon="@drawable/uncivicon2"
+        android:roundIcon="@drawable/uncivicon2"
         android:label="@string/app_name"
         android:isGame="true"
         android:appCategory="game"


SomeTroglodyte:
Now that's an idea. F-Droid trying to find an icon without fastlane and failing only due to a missing Nougat+ metadata entry? Why not. Can't hurt. Using the same asset might look wrong - maybe I'll try this in an AVD, you should see the effect when dragging the app icon around on the "desktop"...

SomeTroglodyte:
OK - this does not change the icon into an adaptive one. By simply following [this](https://developer.android.com/studio/write/image-asset-studio#launcher) (cross-checking [this](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive)) and [this](https://developer.android.com/studio/write/image-asset-studio#create-notification) (cross-checking [this](https://developer.android.com/training/notify-user/build-notification)) I got a version that _does_ have adaptive icon behaviour, and the notification icon should also follow material design.

https://user-images.githubusercontent.com/63000004/165173709-4ba10234-3664-4474-b910-fb2ca86b995b.mp4

It doesn't specify android:roundIcon because the dox say you _must_ not include that unless your design won't allow the automatic round/square choices. And yes, whether it shows as square, rounded corners, squircle or circle follows the launcher's wishes (the >=Oreo AVD's are all similar, only my hardware with LOS on it uses squircles, and the jiggly effect is just peachy)

JackRainy:
My initial point was an assumption that F-droid is looking for the "wrong" icon (round not regular). 
That's great you have managed to create a new "live" icon both squared and rounded.
@SomeTroglodyte , would you create a new PR with your changes or should I cherry-pick yours?

SomeTroglodyte:
I was too lazy to polish that, so if you want go ahead take what you want however you want. No attribution needed. By laziness I mean mostly documentation and clearing up questions like
- How to deal with the Google Play Icon the wizard generated and placed in res - there's also one in extraImages - so how does which one get to the Spam Store in that workflow???
- is it desired by all to have a material (ie monochrome boring flat) icon on notifications instead of the coloured (if android even lets that through) classic Unciv hex?
- Clean up of unused extraImages??? I Gimp'ed v4 and left it in the branch as v5 to remove the black rhomboids at the corners so it would 'float' a tad nicer
- Scale is perhaps, just maybe, still to big, foreground leaves the background area and gets clipped a little when "wiggling" the "finger" fast enough. Means re-running the wizard.
- Is that papyrus-inspired background colour fine or would different tastes prevail?

JackRainy:
While I have no passion of playing with icons right now, and the #6587 is already closed by #6608 (tested as of today), I suggest closing this PR till better times.

PR #6611, opened 2022-04-26
Update to note you need Android 11.0 SDK Platform

itanasi:


SomeTroglodyte:
FYI - That gives you the Android SDK Platform + Sources for Android, of which only the first is mandatory for building. Maybe add to Gradle versions "These are under File - Project Structure - Project Tab. Do not change unless you know what you're doing and plan on pushing the upgrade to a PR." - or so?

SomeTroglodyte:
Confused. I have "Android SDK Platform 30" Revision 3 installed. That's _one_ entry in the detailed view of the SDK manager SDK Platforms tab. Which "Platform for 30" is different from that?

itanasi:
Under SDK Platforms, I had to install the Android 11.0 (R) which is API 30, rev 3. Along with the SDK Tools for 30.0.3 (which were already part of the Wiki instructions).

Just installing the SDK Tools 30.0.3 as the Wiki currently recommends isn't sufficient to fix the error and compile the project

SomeTroglodyte:
So you reiterated what this PR already explains better - I misunderstood that you did the "nuke and start over" _after_ the PR and at then noticed something missing in the result. Why isn't this approved already?

SomeTroglodyte (approved):

PR #6612, opened 2022-04-26
Healing parity

letstalkaboutdune:
Adjusted healing values for parity with CV

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -712,12 +712,12 @@ class MapUnit {
         val isFriendlyTerritory = tileInfo.isFriendlyTerritory(civInfo)
 
         var healing = when {
-            tileInfo.isCityCenter() -> 20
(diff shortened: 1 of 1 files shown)


xlenstra (approved):
Adjusting the comments isn't really necessary, but otherwise looks good!

yairm210:
This is good enough. Anyone who wants to mod this can add global healing uniques based on tile location.

letstalkaboutdune:
Thank you both! I can resubmit a version without the comments if you prefer. Since it's my first contribution I'm still getting a feel for the style and amount of documentation.

xlenstra:
Imo best is to just add another commit to this PR removing these comments; I'd say documentation should describe the code as it is and not as it used to be so I don't see the added benefit of these comments.
On coding style: I try to follow the official kotlin style guide, but I have some disagreements over indentation with it so I differ a bit from it. Just doing your own thing while keeping relatively close to the style guide should be fine.

SomeTroglodyte:
Welcome Sandworm! I say: History is always viewable through git. Mark some code, Git -> Current File -> Show History for Selection is really wondrous.

letstalkaboutdune:
Thanks for the welcome! That's a really good tip. I suppose since the pull request has already been approved and merged I'll just leave it and know for next time.

PR #6613, opened 2022-04-26
Changed milestone build globally text to make it more understandable

xlenstra:
I could deprecate this, but have no easy way of telling it to modders anyway until I make a wiki page for this file, so I've just changed it directly for now

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/VictoryTypes.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/VictoryTypes.json
@@ -27,8 +27,8 @@
     },
     {
         "name": "Diplomatic",
-        "hiddenInVictoryScreen": true,
-        "milestones": ["[United Nations] build globally", "Win diplomatic vote"],
+        "hiddenInVictoryScreen": false,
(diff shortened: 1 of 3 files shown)


yairm210:
I would change to "[building] is built by anyone" personally, but at the very least change "build" to "built"

xlenstra:
I've tried to make it more imperative, to make it similar to all other milestones ('do this', 'complete that', 'build that'), but the weird thing here is that its an imperative to all players. I'd say 'anyone do this' so 'anyone build UN' would then be the closest version. '[building] is built by anyone' is something you'd say when you've completed the milestone, which is not how all other milestones are worded, which is why I chose for the other version.

yairm210:
Ah
I see

yairm210 (approved):

SomeTroglodyte:
I'd support Yair's sugggestion. That still doesn't sound right. Maybe there are just too few words so we read it out of context and therefore it feels weird... I'd throw "Anyone should build" into the ring.

xlenstra:
image
It becomes a bit wide that way, but I suppose its better.

SomeTroglodyte (approved):

PR #6614, opened 2022-04-26
Making SpecialistAllocationTable an ExpanderTab

ultradian:
Hi,
I'm new to this project, and really enjoy the game you have created. I found that sometimes the CityStatsTable gets crowded on the right with the SpecialistAllocationTable taking up space like this:
uncivScreenshot_20220422-153443

I made this PR to enable the player to fold up that section to regain a little bit more view of the map. Let me know if you think there are better approaches to this, whether from a UI design or coding viewpoint. This is my first time coding in Kotlin (and libGDX).

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
@@ -46,10 +46,8 @@ class CityStatsTable(val cityScreen: CityScreen): Table() {
         innerTable.addSeparator()
         addText()
         if (!cityInfo.population.getMaxSpecialists().isEmpty()) {
-            innerTable.addSeparator()
-            innerTable.add(SpecialistAllocationTable(cityScreen).apply { update() }).row()
+            addSpecialistInfo()
         }
-        
         if (cityInfo.religion.getNumberOfFollowers().isNotEmpty() && cityInfo.civInfo.gameInfo.isReligionEnabled())
             addReligionInfo()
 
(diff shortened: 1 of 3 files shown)


yairm210:
Hard to judge UI changes by code, pics please :)

SomeTroglodyte (changes requested):
Welcome brainiac! Some untested thoughts:
- ~~give the expander a persistence ID~~
- ~~the lambda parameter works on an empty Table from a container in "not finished initializing" state - no reliable stage if I remember correctly. Thus a) pack() is not helpful, too early~~ ... oh, wait, that's not the original constructor call at all, both points moot! Well done.
- Got ya: "Specialists:" isn't translatable this way. Either add to template.properties or use the existing "Specialists" <br/>    (`"{Specialists}:"`)

Looks clean, managed to avoid a typical pitfall nicely - I don't think I need to test this (either something fails spectacularly right on opening the city, or it's just fine, and I can assume you tested at least once). Fix translatability however you see fit, then treat as approved by me.

ultradian:
Pics, of course!!
image

and
image

I figured default open would let users figure it out for themselves.

ultradian:
Thanks for the feedback @SomeTroglodyte. Your crossed out comments are educational for my understanding of the component interactions. I made the change you suggested which works, although I am still trying to figure out where in the codebase I would find the list of translated terms  (if I'm not putting them in _template.properties_)
image

SomeTroglodyte (approved):

SomeTroglodyte:
Simplifying, template needs to list everything shown from literals in code, while stuff from jsons is automatic. See TranslationFileWriter. The wiki should have good info too.

PR #6615, opened 2022-04-26
Updated documentation to contain information on victory types

xlenstra:


SomeTroglodyte:
"file contains of strings" -> consists of / contains ~~of~~
"Become the world religion" is "coming soon", I expect -o patronum
"when that amount is reached" -> "when that turn is reached" ?

PR #6623, opened 2022-04-27
Harden composite unit filters

### LINKED ISSUE #6619: Crash on launch with certain Mods
Modchecker runs into an infinite loop, a regression that must be caused by me with #6516. You can get it on game launch or by running "Locate Mod Errors" after freshly installing the Mod. The Mod in question uses "{Military}{Land}" instead of the documented "{Military} {Land}", which with a little thought is _plain and obvious_ that it will break the recursion used in `BaseUnitFilter.getErrorSeverity`. I had a gut feeling back then that `split("} {")` is bad but didn't set a `todo` immediately and then forgot to check later...

We (I) will need to ensure immunity against such trifles.

Devs: I'd tend to not only harden the helper, but in the process maybe also allow that syntax. I'll also look to make sure nested or unmatched braces lead to a RulesetInvariant error. Also - a regex could be the easiest way to do the split safely - too expensive?

SomeTroglodyte:
One way to fix #6619. Now _allows_ whitespace, but catches mistakes such as "{Military} {Water{" which would previously have failed _silently_ ("Water{" never passes matchesFilter).

### DIFF
--- a/core/src/com/unciv/ui/utils/ExtensionFunctions.kt
+++ b/core/src/com/unciv/ui/utils/ExtensionFunctions.kt
@@ -360,15 +360,25 @@ fun (() -> Unit).wrapCrashHandlingUnit(
     return { wrappedReturning() ?: Unit }
 }
 
-/** For filters containing '{', apply the [predicate] to each part inside "{}" and aggregate using [operation];
- *  otherwise return `null` for Elvis chaining of the individual filter. */
-fun <T> String.filterCompositeLogic(predicate: (String) -> T?, operation: (T, T) -> T): T? {
-    val elements: List<T> = removePrefix("{").removeSuffix("}").split("} {")
+private val compositeLogicCheckRegex = Regex("""^\s*(?:\{[^{}]+}\s*)+$""")
+private val compositeLogicSplitRegex = Regex("""\{([^{}]+)}""")
+/** For filters containing '{', apply the [predicate] to each part inside "{}" and aggregate using [operation].
+ * 
+ *  A syntax check is performed, if the string doesn't consist only of non-empty, non-nested {} pairs separated by optional whitespace, then [invalidSyntaxResult] is returned. 
+ *  If valid, this returns `null` if and only if all [predicate] results were `null`, otherwise the aggregation by the [operation] lambda of non-null [predicate] results. */
+fun <T> String.filterCompositeLogic(
+    predicate: (String) -> T?,
+    invalidSyntaxResult: T? = null,
+    operation: (T, T) -> T
+): T? {
+    if (!compositeLogicCheckRegex.matches(this)) return invalidSyntaxResult
+    val elements: Sequence<T> = compositeLogicSplitRegex.findAll(this)
+        .map { it.groups[1]!!.value }
         .mapNotNull(predicate)
-    if (elements.isEmpty()) return null
+    if (elements.none()) return null
     return elements.reduce(operation)
 }
 /** If a filter string contains '{', apply the [predicate] to each part inside "{}" then 'and' (`&&`) them together;
  *  otherwise return `null` for Elvis chaining of the individual filter. */
 fun String.filterAndLogic(predicate: (String) -> Boolean): Boolean? =
-    if (contains('{')) filterCompositeLogic(predicate) { a, b -> a && b } else null
+    if (contains('{')) filterCompositeLogic(predicate, false) { a, b -> a && b } else null
--- a/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
@@ -75,7 +75,7 @@ enum class UniqueParameterType(
         override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
                 UniqueType.UniqueComplianceErrorSeverity? {
             if ('{' in parameterText) // "{filter} {filter}" for and logic
-                return parameterText.filterCompositeLogic({ getErrorSeverity(it, ruleset) }) { a, b -> maxOf(a, b) }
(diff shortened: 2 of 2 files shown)


core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -360,15 +360,25 @@ fun (() -> Unit).wrapCrashHandlingUnit(
     return { wrappedReturning() ?: Unit }
 }
 
-/** For filters containing '{', apply the [predicate] to each part inside "{}" and aggregate using [operation];
- *  otherwise return `null` for Elvis chaining of the individual filter. */
-fun <T> String.filterCompositeLogic(predicate: (String) -> T?, operation: (T, T) -> T): T? {
-    val elements: List<T> = removePrefix("{").removeSuffix("}").split("} {")
+private val compositeLogicCheckRegex = Regex("""^\s*(?:\{[^{}]+}\s*)+$""")
+private val compositeLogicSplitRegex = Regex("""\{([^{}]+)}""")
```
yairm210, line 364: You'll need a \ before the final { - AS lies to you and says it's unneeded, but on Android it is needed. Same thing on the line above,

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -360,15 +360,25 @@ fun (() -> Unit).wrapCrashHandlingUnit(
...
+/** For filters containing '{', apply the [predicate] to each part inside "{}" and aggregate using [operation].
+ * 
+ *  A syntax check is performed, if the string doesn't consist only of non-empty, non-nested {} pairs separated by optional whitespace, then [invalidSyntaxResult] is returned. 
+ *  If valid, this returns `null` if and only if all [predicate] results were `null`, otherwise the aggregation by the [operation] lambda of non-null [predicate] results. */
+fun <T> String.filterCompositeLogic(
+    predicate: (String) -> T?,
+    invalidSyntaxResult: T? = null,
+    operation: (T, T) -> T
```
yairm210, line 372: Since the operation is always max, why even have this as an option?

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
```diff
@@ -360,15 +360,25 @@ fun (() -> Unit).wrapCrashHandlingUnit(
...
-    val elements: List<T> = removePrefix("{").removeSuffix("}").split("} {")
+private val compositeLogicCheckRegex = Regex("""^\s*(?:\{[^{}]+}\s*)+$""")
+private val compositeLogicSplitRegex = Regex("""\{([^{}]+)}""")
+/** For filters containing '{', apply the [predicate] to each part inside "{}" and aggregate using [operation].
+ * 
+ *  A syntax check is performed, if the string doesn't consist only of non-empty, non-nested {} pairs separated by optional whitespace, then [invalidSyntaxResult] is returned. 
+ *  If valid, this returns `null` if and only if all [predicate] results were `null`, otherwise the aggregation by the [operation] lambda of non-null [predicate] results. */
+fun <T> String.filterCompositeLogic(
+    predicate: (String) -> T?,
+    invalidSyntaxResult: T? = null,
```
yairm210, line 371: Same for this, if functions that are always called with the same value for a parameter that probably means the parameter is unneeded, YAGNI etc.

yairm210: Waitasec
Are you planning to use this function somewhere else? Because maybe this should just be collapsed into filterCompositeCheckErrors

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
SomeTroglodyte: I remember the comments elsewhere, yes. Sure that's still relevant, as we have bumped minimum versions since those days? Didn't the broken regex only appear on Java6-based Dalvik engines?  ... Still, you're right, stay defensive. After all, a similar fuzzy motivation keeps me from using `RegexOption.COMMENTS`.

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
SomeTroglodyte: max or `{ a, b -> a && b }`

yairm210:
See #6632

core/src/com/unciv/ui/utils/ExtensionFunctions.kt
SomeTroglodyte: Two values: `RulesetInvariant` and `false`

... and yes the vision that it could be useful for future expansions is fuzzy but existing. For example, I still imagine filters with 'or' capability would in some cases improve in-game readability quite a bit - but for that the templating function would need to use parameterized RegExes as well. It pains me two regexes are needed, but alas this is not mono, or I haven't found the way.

SomeTroglodyte:
To do #6632 with RegEx one would need lookaheads and/or lookbehinds, with the danger those old engines requiring redundant escapes might choke on them - but that's a common use of Regexes. Or do you mean we should refactor _that_ into being reusable for _this_?

SomeTroglodyte:
Question: Is a unit test like (in UnitUniquesTests):
```kotlin
    @Test
    fun malformedUnitFiltersDoNotCrash() {
        val civInfo = game.addCiv("[Gold] cost of purchasing [{Military}{Land}] units [-90]%")
        val tile = game.getTile(Vector2.Zero)
        val city = game.addCity(civInfo, tile)
        val construction = game.ruleset.units["Warrior"]!!
        try {
            construction.getStatBuyCost(city, Stat.Gold)
        } catch (exception: Throwable) {
            fail("getStatBuyCost with a malformed unique throws ${exception.javaClass.simpleName}")
        }
    }
```
... _supposed_ to succeed?

PR #6631, opened 2022-04-27
Allow removing Fallout from the city tiles

JackRainy:
I have faced the issue that the Fallout is the only "improvement" that can be "built" in the city but then there is no way to remove it. This PR fixes this problem.

P.S> During creation of the PR, I decide to look through the issues and found that #6355 was not fixed but just closed without proper testing. Now it is fixed-fixed.

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -229,7 +229,7 @@ open class TileInfo {
         }
     }
 
-    fun isEnemyTerritory(civInfo: CivilizationInfo): Boolean {
+    private fun isEnemyTerritory(civInfo: CivilizationInfo): Boolean {
         val tileOwner = getOwner() ?: return false
         return civInfo.isAtWarWith(tileOwner)
     }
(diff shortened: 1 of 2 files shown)


yairm210:
??? Fallout shouldn't spawn in city tiles at all.

JackRainy:
They are not spawn themselves but after nuking the city, of course, they appear there.

yairm210:
Yes
The city center improvement should prevent fallout from spawning

xlenstra:
As far as I'm aware, as city centers have the 'indestructible' unique, it should be impossible for fallout on them. Were your tests done with mods that don't have this unique?

SomeTroglodyte:
But UniqueType.Indestructible doesn't work, as can readily be seen from code inspection: All uses are in `doNukeExplosionForTile` and the gating of "Fallout ~~New Vegas~~" is attempted via `tile.hasUnique()` which looks in `getAllTerrains()` - not in improvement.

JackRainy:
@xlenstra , I have used none of mods. It reproduces quite easily - just 1-2 nuclear missiles on the city.
Will revert the changes while doing the fix on another end of the problem :)

SomeTroglodyte:
> It reproduces quite easily

See my comment just above yours

JackRainy:
Closing this, since the problem I was fixing, was not relevant.
The solution comes from the another side - #6640

SomeTroglodyte:
But thanks for stirring the pot!

Issue #6474, opened 2022-04-01
City names with prefixes should be translatable


HaneulCheong:
**Is your feature request related to a problem? Please describe.**
City names with (what I call) overflow prefixes such as "New Rome" cannot be translated as of now, even after the rework done in 4.0.3.

**Describe the solution you'd like**
There are several different ways to implement this feature, which is why I thought the whole thing should be consulted with other collaborators.

For instance we could implement a new internal name format for city names such as:

-   `"{New }{Rome}"` (New&nbsp; = New&nbsp;)
    -   Straight forward to translators
    -   Ugly
-   `"{New}{Rome}"` (New = New&nbsp;)
    -   Pretty
    -   Confusing to translators
-   `"{New} {Rome}"` (New = New)
    -   Straight forward to everybody
    -   Cannot accommodate languages without spaces

**Describe alternatives you've considered**
We can always scrap the whole prefix mechanic and fallback to the Civ 5's way of dealing this: borrowing city names from absent nations if all names are consumed.

**Additional context**
This bug thingy was found in #6439.

yairm210:
Alternate city naming:
New [cityName] = 
Neo [cityName] = 

That way we can accomodate languages without spaces AND have nice translations

HaneulCheong:
> New [cityName] =
> Neo [cityName] =

So this is the language of Gods I guess... Jokes aside, I'll do some tests then bring the result soon.

yairm210:
SomeTrog is right, I didn't think that through, it's actually not a great idea

SomeTroglodyte:
No, all ideas are good while we're missing the 'eureka' moment...

yairm210:
Been possible for a while with nested translations

PR #6632, opened 2022-04-27
Added placeholder parameter location in a way that enables nested translations

yairm210:
It's kind of pathetic, I've been avoiding having to touch this issue because I thought it would be really complex, but it's actually super easy, barely an inconvenience.
Got major flashbacks to learning Compilers in college, where we had to build our own high-to-mid and mid-to-low level compilers using a randomly chosen language. I wonder if they still teach that.

### DIFF
--- a/core/src/com/unciv/models/translations/Translations.kt
+++ b/core/src/com/unciv/models/translations/Translations.kt
@@ -392,10 +392,9 @@ fun String.tr(): String {
  * allowing us to have nested translations!
  */
 fun String.getPlaceholderParametersIgnoringLowerLevelBraces(): List<String> {
-//    val parameters = ArrayList<String>()
+    val parameters = ArrayList<String>()
     var depthOfBraces = 0
     var startOfCurrentParameter = -1
-    val parameterLocations = ArrayList<Pair<Int,Int>>()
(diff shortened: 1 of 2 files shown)


yairm210:
There are two places where this could be used effective immediately: city names as per #6474 and multi-term filters
(there should already be a "[] []" translation for "[stats] [cityFilter]" so that can also be used to parse and display correcly spaced vs unspaced languages, e.g. [{Military} {Water}] -> [[Military] [Water]]. Won't work for more than 2 parameters but really who does that, and if they do we can add it as well, or they can do like "[[conditionA] [[conditionB] [conditionC]]]  units" - ugly but it works.

SomeTroglodyte (approved):
Niiiiice test. And if the function passes that, I can approve without testing myself :smile_cat: 
Only thing I can think of - the wiki update to cover this. Or, to annoy y'all, how about the clever way to express getPlaceholderText?
```kotlin
fun String.getPlaceholderText() = removeConditionals().run {
    getPlaceholderParametersIgnoringLowerLevelBraces().fold(this) {
        stringToReturn, placeholderParameter -> stringToReturn.replace("[$placeholderParameter]", "[]")
    }
}
```

PR #6641, opened 2022-04-28
Map editor2.2

### LINKED ISSUE #6635: Can't load saved map
**Platform**
All

**Version**
4.0.10

**Describe the bug**
I can't load saved map from map editor

image

SomeTroglodyte:
Smaller than the last one, the commit messages say it all.

Fixes #6635
Missing ruleset _objects_ no longer prevent map load, they get removed and toasted instead.
e.g. The 'Unciv Europe Map Example' I have on disk still has 'Gold' instead of 'Gold ore' resources. Now loads just fine without the resource and without the starting locations.

### DIFF
--- a/core/src/com/unciv/ui/mapeditor/MapEditorLoadTab.kt
+++ b/core/src/com/unciv/ui/mapeditor/MapEditorLoadTab.kt
@@ -100,8 +100,8 @@ class MapEditorLoadTab(
             val newBaseRuleset = map.mapParameters.mods.filter { it !in missingMods }.firstOrNull { RulesetCache[it]!!.modOptions.isBaseRuleset }
             if (newBaseRuleset != null) map.mapParameters.baseRuleset = newBaseRuleset
             //
-
             if (map.mapParameters.baseRuleset !in RulesetCache) missingMods += map.mapParameters.baseRuleset
+
             if (missingMods.isNotEmpty()) {
                 Gdx.app.postRunnable {
                     needPopup = false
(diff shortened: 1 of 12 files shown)


core/src/com/unciv/ui/mapeditor/MapEditorLoadTab.kt
```diff
@@ -118,7 +118,17 @@ class MapEditorLoadTab(
                         map.mapParameters.mods -= modBaseRuleset
                     }
 
-                    editorScreen.loadMap(map)
+                    val ruleset = RulesetCache.getComplexRuleset(map.mapParameters)
+                    val rulesetIncompatibilities = map.getRulesetIncompatibility(ruleset)
+                    if (rulesetIncompatibilities.isNotEmpty()) {
+                        map.removeMissingTerrainModReferences(ruleset)
+                        val message = "{This map has errors:}\n\n".tr() +
+                                rulesetIncompatibilities.sorted().joinToString("\n") { it.tr() } +
```
xlenstra, line 126: So you first sort and then translate? Would that still look good in non-English languages?

xlenstra (approved):

core/src/com/unciv/ui/mapeditor/MapEditorLoadTab.kt
SomeTroglodyte: No oversight, I was reluctant to pull the collator for something as rarely seen. I wanted it deterministic and a bit grouped, actual alphabetics - shrug.

PR #6608, opened 2022-04-25
Implement fastlane step 1 - minimal framework

SomeTroglodyte:
This just might close #6587. Perhaps. Almost no idea how to actually _test_ (I could merge into _my_ master then fake a few commits for a changelog then go through the steps to trigger a version 701 there, just to see the script creating a file, but that's a _lot_ of work, especially getting my master back into sync. Testing the F-Droid side by forking their server code and setting up my own f-repo is even more disproportionate).

#### Done
- Minimal required fastlane structure
- Coded generation of changelogs/*.txt discrete files ([this guide](https://gitlab.com/-/snippets/1895688) lists them as 'nice to have', but I think their importance is higher)
- Made sure image sizes match recommendations where any can be found (192² for icon, 1024x500 for featureGraphic, both already available)
- Fulfill requiremets to make it into the "Latest" spam [as per this](https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#latest-tab-criteria)

#### To-Do after this PR
- [ ] File a PR against [Unciv's fdroiddata yml](https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.unciv.app.yml) to _remove_ `Description` entirely (it overrides fastlane, all languages)
- [ ] Automate syncing android/res/drawable-xxxhdpi/uncivicon2.png to fastlane/metadata/android/en-US/images/icon.png (In that js script? Copy or symlink?)
- [ ] Automate syncing extraImages/Feature graphic - Google Play.png to fastlane/metadata/android/en-US/images/featureGraphic.png
- [ ] Automate syncing extraImages/Screenshots to fastlane/metadata/android/en-US/images/phoneScreenshots
- [ ] Place hooks into template.properties to ask for translations for the full_description.txt and short_description.txt files
- [ ] Automate copying these translations into the fastlane structure (In TranslationFileWriter?)

PR #6643, opened 2022-04-28
Implement fastlane step 2: Translating

SomeTroglodyte:
Follow-up on #6608 - just one of many possible ways to get our translating contributors to supply the translations for Fastlane. Please review carefully and do not hesitate to offer better approaches.

- The magic happens with the same button that launches TranslationFileWriter - a 'why not' and 'one object I need is already instantiated there' decision.
- The changes to Options re-enable faster access to the translation generation using the F12 key. Regressed due to the controls and binding created significantly _after_ the Popup opens. I moved the decoupling thread a bit down, so that the whole advanced page is created immediately as before, only the cells where font label and select go are empty at first, the thread then places the actors in the cells. This makes a very quick O-F12 possible again (no you don't need to actually go to the advanced page for F12 to work), and to clarify what happens I also decoupled the actual TranslationFileWriter run and made sure a "Working..." is shown. None of the changes to that file are actually required for what this PR aims to do.
- German is included because I proved for myself that a TranslationFileWriter run leaves unchanged content in the already existing Fastlane files (still does touch the modification time).


SomeTroglodyte:
And @yairm210 - please - the full description in the G00gle Play St🤮re is mostly identical to the F-Droid one but never got the updated "internet permissions" paragraph - where exactly is all that maintained (manually @'goo I guess) and can something be automated to actually use the Fastlane content for the Play St0re too?

xlenstra:
In the google play console all of that can be changed, but I couldn't quickly find a way to do this automaticlaly. Assuming we don't change these too often I'm fine with manually inputting all of the strings for different languages

SomeTroglodyte:
| F-Droid start screen | F-Droid app screen | F-Droid categories | Aurora Store |
|----|----|----|----|
| Screenshot_20220501-171919_F-Droid | Screenshot_20220501-171944_F-Droid | Screenshot_20220501-172020_F-Droid | Screenshot_20220501-172403_Aurora_Store |

So [step 1](https://github.com/yairm210/Unciv/pull/6608) was a full success; time to get this rolling.

android/assets/jsons/translations/German.properties
```diff
@@ -10,6 +10,18 @@
 StartWithCapitalLetter = true
 
 
+# Fastlane
+# These will be automatically copied to the fastlane descriptions used by F-Droid. Their keys are not as usual the english original, please read those directly as linked.
+
+# Documentation: https://f-droid.org/en/docs/Build_Metadata_Reference/#Summary
+# English to translate: https://github.com/yairm210/Unciv/blob/master/fastlane/metadata/android/en-US/short_description.txt
+Fastlane_short_description = 4X-Spiel: Zivilisationen bauen
```
xlenstra, line 18: Isn't this exactly what we try to avoid with these translation files? So that translators can easily see what needs to be translated and don't have to open another link (isn't always what you want when you only have a mobile device)?

xlenstra, line 18: Isn't this exactly what we try to avoid with these translation files? So that translators can easily see what needs to be translated and don't have to open another link (isn't always what you want when you only have a mobile device)?

xlenstra: Other than this looks fine

SomeTroglodyte: Translators working from mobile? On a wearable maybe? Masochists!

Motivation: a) stands out, wake them up a little, b) full is awfully long and _much_ easier to read in the already separately existing file... c) make code independent from changes to the english text. All other translation keys are also fallbacks for immediate display (except for `" " = ` and StartWithCapitalLetter - who did those anyway?) ... so these, however long, must automatically be part of the release packages. The Fastlane english originals are _not_ displayed anywhere by Unciv itself, so I don't need the actual english content for anything other than the one offer for fastlane-speaking stores. Having the non-english text in the codebase and memory _is_ a kludge to shoehorn our translation system for these, saves some separate system, a small price to pay - that was the basic idea to this PR. Alternate concepts welcome, but puhlease don't make me escape and one-line that full_desc.

P.S.: There's still some strings in android/res/values/strings.xml _not_ using our translation central... Go take a look - abstract keys on the left.

yairm210 (approved):

SomeTroglodyte:
As of 2022-05-04 the [tcherman F-Droid page for Unciv](https://f-droid.org/de/packages/com.unciv.app/) already shows up translated (nothing to do with this PR, but with the change to the yml done by linsui).

Shows I was naïvely expecting markdown to work in the full desc. @Mape6 - wanna work it over?

PR #6644, opened 2022-04-29
Fix TileInfo.neighbors being incomplete in rare circumstances

### LINKED ISSUE #6531: Rare failure of BFS in assignContinents
**Version**
My map editor branch merged to 4.0.5 (meaning the faulty code is exactly as in current master)

**Describe the bug**
Two maps in my mods folder, from the 'Great Community Maps' mod as loaded on 2021-12-01: 3_Continents.txt and Indonesia.txt, show this when loaded and assignContinents(Ensure) is run on them. Master code is crashed by the "Continent already assigned" throw in TileInfo.setContinent. This is because the assignContinents code assumes BFS is commutative, meaning if A can be reached starting at B, then B can be reached starting at A. This is **not** the case on those maps, for whatever reason.

<details><summary>screenshot</summary>

image
</details>
The marked tiles on that "continent" can be reached from the unmarked ones at its right corner, but the inverse is not true. likewise, the rightmost tile, the one without a railroad, can reach the rest but it isn't BFS-reached from there. That's my current interpretation, I don't have the time to investigate further.

#
Map file:
[Indonesia.txt](https://github.com/yairm210/Unciv/files/8464842/Indonesia.txt)

SomeTroglodyte:
Fixes #6531 and possibly a quite number of unnoticed quirks.

Symptom I chased: Loading certain maps led to "continent already assigned" exceptions, the continents at that point were not up to expectations. Race condition due to the lazy being triggered while tileMatrix is still being built, which in turn was caused by convertHillToTerrainFeature. Other triggers due to setTransients using neighbor-checking uniques may have been possible or not, and even if not they might have been added in the future. By separating tileMatrix init and per-tile setTransients into two loops this is now all right.


SomeTroglodyte:
Using the continent visualization of map editor:
image
... in 'Before' that continent is missing three tiles.

xlenstra (approved):

yairm210:
Please add comments in code, I have no idea what problem this solves

yairm210:
Like, why are we forcing the x and y of the matrix to be at least a certain size?

SomeTroglodyte:
> I have no idea what problem this solves
Building tileMatrix (filling in the TileInfo references) and calling setTransients on those TileInfos -in the same loop- is fine as long as setTransients is guaranteed to leave neighbors uninitialized. I found a case where it queries and thus initalizes neighbors - convertHillToTerrainFeature. Uniques may be another trigger. And if that happens, some logical neighbors will be missing from the `neighbors` list because their tileMatrix position is still null. And they would _stay_ that way.

> why are we forcing

Pre-allocating. We know the final size exactly. Without specifying an initial capacity, ArrayList will start at 10 entries then re-allocate to 15 and copy over for the 11th item, then to 24, to 36, to 54,... Thus typically the old code would discard the storage and Arrays.copyOf the elementData into new storage four times. Unnecessary stress for the GC if you know the size beforehand.

PR #6645, opened 2022-04-29
New Demographics Scoreboard

letstalkaboutdune:
I've implemented a Demographics Scoreboard similar to the one from Civ V seen below:

Demographics

I feel that this view which only shows the current player's rank, value, and overall best/average/worst in the game metrics is more competitive than what's currently shown in the rankings of the victory screen in Unciv, which gives you detailed information for all civilizations in the game. 

To use, enable the checkbox in the new game setup screen. The placement here is intentional so that it can't be changed during a game. By default it is deselected, which uses the old ranking screen.

Game Options

In my implementation, the table is transposed compared to Civ V in order to stay more consistent with the other screens in Unciv. Note that the button on the top of the screen changes from "Rankings" to "Demographics" when this scoreboard is used. The formatting is basic and I'm not sure the current order of rows is optimal. Please let me know any feedback on the UI or code, as I'm not very experienced and more than happy to make any changes.

Scoreboard

### DIFF
--- a/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
@@ -96,11 +96,11 @@ class GameOptionsTable(
             { gameParameters.oneCityChallenge = it }
 
     private fun Table.addNuclearWeaponsCheckbox() =
-            addCheckbox("Enable nuclear weapons", gameParameters.nuclearWeaponsEnabled)
+            addCheckbox("Enable Nuclear Weapons", gameParameters.nuclearWeaponsEnabled)
             { gameParameters.nuclearWeaponsEnabled = it }
 
     private fun Table.addDemographicsCheckbox() =
-        addCheckbox("Demographics-style ranking", gameParameters.demographicsEnabled)
+        addCheckbox("Demographics Scoreboard", gameParameters.demographicsEnabled)
         { gameParameters.demographicsEnabled = it }
 
 
--- a/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
@@ -68,6 +68,7 @@ class GameOptionsTable(
         checkboxTable.addNuclearWeaponsCheckbox()
         checkboxTable.addIsOnlineMultiplayerCheckbox()
         checkboxTable.addReligionCheckbox(cityStateSlider)
+        checkboxTable.addDemographicsCheckbox()
         add(checkboxTable).center().row()
 
         if (!isPortrait)
@@ -97,6 +98,10 @@ class GameOptionsTable(
     private fun Table.addNuclearWeaponsCheckbox() =
             addCheckbox("Enable nuclear weapons", gameParameters.nuclearWeaponsEnabled)
             { gameParameters.nuclearWeaponsEnabled = it }
+
+    private fun Table.addDemographicsCheckbox() =
+        addCheckbox("Demographics-style ranking", gameParameters.demographicsEnabled)
+        { gameParameters.demographicsEnabled = it }
 
 
     private fun Table.addIsOnlineMultiplayerCheckbox() =
--- a/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
+++ b/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
@@ -174,6 +174,12 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
     }
 
     private fun setCivRankingsTable() {
+        if (gameInfo.gameParameters.demographicsEnabled ) {
+            // use new demographics table
+        } else {
+            // use old rankings table
+        }
+
         val majorCivs = gameInfo.civilizations.filter { it.isMajorCiv() }
         val civRankingsTable = Table().apply { defaults().pad(5f) }
 
--- a/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
+++ b/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
@@ -174,29 +174,34 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
     }
 
     private fun setCivRankingsTable() {
-        if (gameInfo.gameParameters.demographicsEnabled ) {
-            // use new demographics table
-        } else {
-            // use old rankings table
-        }
+        val majorCivs = gameInfo.civilizations.filter { it.isMajorCiv() } //creates element of only major civs from gameInfo
+        val civRankingsTable = Table().apply { defaults().pad(5f) } //creates table with padding of 5f
 
-        val majorCivs = gameInfo.civilizations.filter { it.isMajorCiv() }
-        val civRankingsTable = Table().apply { defaults().pad(5f) }
+        if (gameInfo.gameParameters.demographicsEnabled ) { //if demographics option is enabled...
+            val demographicsHeaders = arrayOf("Demographic","Rank","Value","Best","Average","Worst")
 
-        for (category in RankingType.values()) {
-            val column = Table().apply { defaults().pad(5f) }
-            column.add(category.name.replace('_',' ').toLabel()).row()
-            column.addSeparator()
-
-            for (civ in majorCivs.sortedByDescending { it.getStatForRanking(category) }) {
-                column.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX().row()
+            for (i in demographicsHeaders.indices) { //for every index in DemographicsHeaders array...
+                val column = Table().apply { defaults().pad(5f) } //create a column with padding of 5f
+                column.add(demographicsHeaders[i].replace('_',' ').toLabel()).row() //pulls name of category and creates label with it
+                column.addSeparator() //adds a separator line
+                civRankingsTable.add(column) //add resulting column to table
             }
+        } else { //if demographics option is disabled...
+            for (category in RankingType.values()) { //for every category in RankingType class (Score, Population, etc.)...
+                val column = Table().apply { defaults().pad(5f) } //create a column with padding of 5f
+                column.add(category.name.replace('_',' ').toLabel()).row() //pulls name of category and creates label with it
+                column.addSeparator() //adds a separator line
 
-            civRankingsTable.add(column)
+                for (civ in majorCivs.sortedByDescending { it.getStatForRanking(category) }) { //for every civ in majorCivs element, get ranking for this category and sort descending
+                    column.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX().row() //add the civ name, background, and rank value to a row in the column
+                }
+
+                civRankingsTable.add(column) //add resulting column to table
+            }
         }
 
-        contentsTable.clear()
-        contentsTable.add(civRankingsTable)
+        contentsTable.clear() //clear previous contents of table
+        contentsTable.add(civRankingsTable) //add civRankingsTable to table
     }
 
     private fun getCivGroup(civ: CivilizationInfo, afterCivNameText: String, currentPlayer: CivilizationInfo): Table {
--- a/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
+++ b/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
@@ -176,42 +176,40 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
     private fun setCivRankingsTable() {
         val majorCivs = gameInfo.civilizations.filter { it.isMajorCiv() }
         val civRankingsTable = Table().apply { defaults().pad(5f) }
+        val rankLabels = arrayOf("Demographic","Rank","Value","Best","Average","Worst")
 
         if (gameInfo.gameParameters.demographicsEnabled ) {
-            val demographicsHeaders = arrayOf("Demographic","Rank","Value","Best","Average","Worst")
+            for (i in rankLabels.indices) {
+                if (i == 0) {
+                    val demoColumn = Table().apply { defaults().pad(5f) }
+                    demoColumn.add(rankLabels[i].toLabel()).row()
+                    demoColumn.addSeparator()
+                    civRankingsTable.add(demoColumn)
+                    for (category in RankingType.values()) {
+                        val headers = Table().apply { defaults().pad(5f) }
+                        headers.add(category.name.replace('_',' ').toLabel()).row()
+                        headers.addSeparator()
+                        civRankingsTable.add(headers)
+                    }
+                } else {
+                    civRankingsTable.row()
+                    val demoRow = Table().apply {defaults().pad(5f)}
+                    demoRow.add(rankLabels[i].toLabel()).row()
 
-            for (i in demographicsHeaders.indices) {
-                val demoColumn = Table().apply { defaults().pad(5f) }
-                demoColumn.add(demographicsHeaders[i].toLabel()).row()
-                demoColumn.addSeparator()
-                civRankingsTable.add(demoColumn)
+                    //add rank and other info here
+
+                    civRankingsTable.add(demoRow)
+                }
             }
-
-            /*
-            val catColumn = Table().apply { defaults().pad(5f) }
-            for (category in RankingType.values()) {
-                catColumn.add(category.name.replace('_',' ')).fillX().row()
-            }
-            civRankingsTable.add(catColumn)*/
-
-                //pull categories and add to col. 0 of civRankingsTable: Score, Population, Yield, etc.
-                //determine player's civ
-                //determine player's rank in all categories, add to col. 1
-                //show player's value in all categories, add to col. 2
-                //determine best civ in all categories, show in col. 3 (hide civ if unmet)
-                //calculate average score in all categories, show in col. 4
-                //calculate worst civ in all categories, show in col. 5
 
         } else {
             for (category in RankingType.values()) {
                 val column = Table().apply { defaults().pad(5f) }
                 column.add(category.name.replace('_',' ').toLabel()).row()
                 column.addSeparator()
-
                 for (civ in majorCivs.sortedByDescending { it.getStatForRanking(category) }) {
                     column.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX().row()
                 }
-
                 civRankingsTable.add(column)
             }
         }
--- a/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
+++ b/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
@@ -174,34 +174,50 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
     }
 
     private fun setCivRankingsTable() {
-        val majorCivs = gameInfo.civilizations.filter { it.isMajorCiv() } //creates element of only major civs from gameInfo
-        val civRankingsTable = Table().apply { defaults().pad(5f) } //creates table with padding of 5f
+        val majorCivs = gameInfo.civilizations.filter { it.isMajorCiv() }
+        val civRankingsTable = Table().apply { defaults().pad(5f) }
 
-        if (gameInfo.gameParameters.demographicsEnabled ) { //if demographics option is enabled...
+        if (gameInfo.gameParameters.demographicsEnabled ) {
             val demographicsHeaders = arrayOf("Demographic","Rank","Value","Best","Average","Worst")
 
-            for (i in demographicsHeaders.indices) { //for every index in DemographicsHeaders array...
-                val column = Table().apply { defaults().pad(5f) } //create a column with padding of 5f
-                column.add(demographicsHeaders[i].replace('_',' ').toLabel()).row() //pulls name of category and creates label with it
-                column.addSeparator() //adds a separator line
-                civRankingsTable.add(column) //add resulting column to table
+            for (i in demographicsHeaders.indices) {
+                val demoColumn = Table().apply { defaults().pad(5f) }
+                demoColumn.add(demographicsHeaders[i].toLabel()).row()
+                demoColumn.addSeparator()
+                civRankingsTable.add(demoColumn)
             }
-        } else { //if demographics option is disabled...
-            for (category in RankingType.values()) { //for every category in RankingType class (Score, Population, etc.)...
-                val column = Table().apply { defaults().pad(5f) } //create a column with padding of 5f
-                column.add(category.name.replace('_',' ').toLabel()).row() //pulls name of category and creates label with it
-                column.addSeparator() //adds a separator line
 
-                for (civ in majorCivs.sortedByDescending { it.getStatForRanking(category) }) { //for every civ in majorCivs element, get ranking for this category and sort descending
-                    column.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX().row() //add the civ name, background, and rank value to a row in the column
+            /*
+            val catColumn = Table().apply { defaults().pad(5f) }
+            for (category in RankingType.values()) {
+                catColumn.add(category.name.replace('_',' ')).fillX().row()
+            }
+            civRankingsTable.add(catColumn)*/
+
+                //pull categories and add to col. 0 of civRankingsTable: Score, Population, Yield, etc.
+                //determine player's civ
+                //determine player's rank in all categories, add to col. 1
+                //show player's value in all categories, add to col. 2
+                //determine best civ in all categories, show in col. 3 (hide civ if unmet)
+                //calculate average score in all categories, show in col. 4
+                //calculate worst civ in all categories, show in col. 5
+
+        } else {
+            for (category in RankingType.values()) {
+                val column = Table().apply { defaults().pad(5f) }
+                column.add(category.name.replace('_',' ').toLabel()).row()
+                column.addSeparator()
+
+                for (civ in majorCivs.sortedByDescending { it.getStatForRanking(category) }) {
+                    column.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX().row()
                 }
 
-                civRankingsTable.add(column) //add resulting column to table
+                civRankingsTable.add(column)
             }
         }
 
-        contentsTable.clear() //clear previous contents of table
-        contentsTable.add(civRankingsTable) //add civRankingsTable to table
+        contentsTable.clear()
+        contentsTable.add(civRankingsTable)
     }
 
     private fun getCivGroup(civ: CivilizationInfo, afterCivNameText: String, currentPlayer: CivilizationInfo): Table {
--- a/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
+++ b/core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
@@ -32,7 +32,13 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
         if (!playerCivInfo.isSpectator()) tabsTable.add(setMyVictoryButton)
         val setGlobalVictoryButton = "Global status".toTextButton().onClick { setGlobalVictoryTable() }
         tabsTable.add(setGlobalVictoryButton)
-        val setCivRankingsButton = "Rankings".toTextButton().onClick { setCivRankingsTable() }
+
+        val rankingLabel = if (gameInfo.gameParameters.demographicsEnabled ) {
+            "Demographics"
+        } else {
+            "Rankings"
+        }
+        val setCivRankingsButton = rankingLabel.toTextButton().onClick { setCivRankingsTable() }
         tabsTable.add(setCivRankingsButton)
         topTable.add(tabsTable)
         topTable.addSeparator()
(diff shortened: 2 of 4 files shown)


letstalkaboutdune:
Also, I'm pretty new to Github and not sure why the old commit for "Healing Parity" is getting added to this, since it was already approved and merged.

SomeTroglodyte:
> why the old commit

A branch has a 'base', that won't change automatically. That's what rebase is for. I have my own strong recommendations on when to rebase or not - I may have documented that in my "[Scratchpad](https://github.com/SomeTroglodyte/Scratchpad/tree/master/Unciv-notes)" repo - not maintained for a while though.

If github lists no conflicts, it would be fine to leave it. Since it does - merge forward (that's the essence of those self-recommenations: rebase before publishing but stop rebasing once PR'ed, only use merge / if pushed by not PR'ed delete the online copy rebase and repush). And rebase on what /merge what? My take is - you local master which has just pulled from the upstream master then pushed to your online master, others "grok" it differently.

SomeTroglodyte:
- "Enable Nuclear Weapons" - a changed case will need updating the corresponding template row if it exists - or is that a fix to actually use an existing one?
- "Demographics" and the checkbox label will need translation templates
- "Demographics Scoreboard" checkbox - players won't immediately understand something is _replaced_ (but keep it short - thus I have no idea how to improve except a desktop-only tooltip)
- I would have done two separate factory functions for the alternate layouts
- `else -> {}`: else -> Unit / @Suppress("non exhaustive when") / Isn't the exhaustiveness nagging only for expression-when not block-when?
- val rankingLabel in 1 line wouldn't be too bad - and there's an extra blank "if (expression )"
- Most operators should have blanks on both sides (e.g. `if (n==0)`) and even `replace('_',' ')` would get one by the style guides I respect, `if(civ.isCurrentPlayer()){` two blanks...
- Now an idea not necessarily to everyone's liking - when(n) 1.. 2.. 3.. 4.. boom: Having a local helper `fun` for each branch could make the comments redundant and tighten readability a bit - and the daring would even do an array indexing instead of a when

Just food for thought, haven't got a Studio here to actually _look_ properly.

ravignir:
can't we change it further, so the values of certain things will be more in line with civ5? 
like population listed as millions, soldiers listed as thousands, approval(happiness) and literacy(science) as percentage, land in square meters etc.

SomeTroglodyte:
> millions square inches ...

That's something I'd be 66% against - we have less help text so that leaves some people wondering "HTF is this calculated"...

I think I'll sneak in a RankingType.Exploration sometime - % of tiles explored - which reminds me : width vs. height - see overview resources and its new "flip axes" button, this could do with something like that. Maybe I should wrap the axis-flippable spreadsheet concept into a generic Widget...

letstalkaboutdune:
> > why the old commit
> 
> A branch has a 'base', that won't change automatically. That's what rebase is for. I have my own strong recommendations on when to rebase or not - I may have documented that in my "[Scratchpad](https://github.com/SomeTroglodyte/Scratchpad/tree/master/Unciv-notes)" repo - not maintained for a while though.
> 
> If github lists no conflicts, it would be fine to leave it. Since it does - merge forward (that's the essence of those self-recommenations: rebase before publishing but stop rebasing once PR'ed, only use merge / if pushed by not PR'ed delete the online copy rebase and repush). And rebase on what /merge what? My take is - you local master which has just pulled from the upstream master then pushed to your online master, others "grok" it differently.

I appreciate the guidance. Unfortunately due to my novice-ness I only understand a little bit of this. I'll read up on branches, bases, etc.

letstalkaboutdune:
> * "Enable Nuclear Weapons" - a changed case will need updating the corresponding template row if it exists - or is that a fix to actually use an existing one?
> * "Demographics" and the checkbox label will need translation templates
> * "Demographics Scoreboard" checkbox - players won't immediately understand something is _replaced_ (but keep it short - thus I have no idea how to improve except a desktop-only tooltip)
> * I would have done two separate factory functions for the alternate layouts
> * `else -> {}`: else -> Unit / @Suppress("non exhaustive when") / Isn't the exhaustiveness nagging only for expression-when not block-when?
> * val rankingLabel in 1 line wouldn't be too bad - and there's an extra blank "if (expression )"
> * Most operators should have blanks on both sides (e.g. `if (n==0)`) and even `replace('_',' ')` would get one by the style guides I respect, `if(civ.isCurrentPlayer()){` two blanks...
> * Now an idea not necessarily to everyone's liking - when(n) 1.. 2.. 3.. 4.. boom: Having a local helper `fun` for each branch could make the comments redundant and tighten readability a bit - and the daring would even do an array indexing instead of a when
> 
> Just food for thought, haven't got a Studio here to actually _look_ properly.

Master Jedi, thank you for all this feedback! I have some homework to do looking into all of these points.

letstalkaboutdune:
> can't we change it further, so the values of certain things will be more in line with civ5? like population listed as millions, soldiers listed as thousands, approval(happiness) and literacy(science) as percentage, land in square meters etc.

I'm actually pretty against this, as I think Civ V's way is unnecessarily abstracted. The math was found and posted [here](https://civilization.fandom.com/wiki/Demographics_(Civ5)) in the Civ V wiki, but does it really add anything?

SomeTroglodyte:
> Master Jedi

Nah not even Padawan. I learned by doing just like everybody else. For this PR, I can't even really guide you on how to resolve those conflicts up there in a way that maximizes useful lessons learned - you worked in your _master_, and the way I know needs the local master to be separate from the new work (and synced to the main trunk master). There may be ways to... Like I at times do when I F-upped and pushed non-public code to my online master... Involves renaming, pulling, pushing, reassigning what the master is supposed to be then deleting - advanced. Or... Nah. Or maybe... Dunno.

If you want to use Android Studio to merge (IMHO way better than github online ways), maybe check out this repo's master _under a new name_ then merge that into your local master, push, and after this PR is either merged or shot down start over. Certainly not the only option, but I try to limit my git expertise to what is easily done within the Studio GUI.

letstalkaboutdune:
> If github lists no conflicts, it would be fine to leave it. Since it does - merge forward (that's the essence of those self-recommenations: rebase before publishing but stop rebasing once PR'ed, only use merge / if pushed by not PR'ed delete the online copy rebase and repush). And rebase on what /merge what? My take is - you local master which has just pulled from the upstream master then pushed to your online master, others "grok" it differently.

Well in this case the merge was trivial (removal of some comments), I was able to resolve it within GitHub. Moving forward I've created a new branch for this feature and will continue the dev there.

letstalkaboutdune:
> * I would have done two separate factory functions for the alternate layouts

This is the only bit of your feedback I have yet to address (because it's the most involved XD). Does this mean you would move the bulk of the code for each layout to its own function, and just call the functions from SetCivRankingsTable()? So ultimately it would look something like...

```
if (demographicsEnabled) {
  call buildDemographicsTable function
} else
  call buildRankingsTable function
}
```

Is that the idea?

SomeTroglodyte:
> Is that the idea

Yes but only if while doing it one gets the impression it will indeed look simpler. If there's too many things not readily available e.g. in `this` or lots of common preparation needing multiple parameters to pass into the alternatives, that could end up harder to read. You get the idea. Btw, leftover debug- or devonly- comments don't help readability. :rofl: :+1:

> barely spell "Kotlin." 

Bah, mental ability to juggle puzzle pieces is more important than spelling
["Kotlin"](https://goo.gl/maps/MKkDSdngvuHZrcedA) -no- ["Котлин"](https://goo.gl/maps/tCx5QGU71ArkTJ76A).

SomeTroglodyte:
My earlier comment on blanks and comma - I thought "after not before" is documented best practice - but actually [the doc I like](https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace) doesn't seem to explicitly mention that case...

letstalkaboutdune:
Ok, I'm ready for review again! I think the organization into functions works pretty well and makes the higher-level `setCivRankingsTable()` function very simple to interpret. The only UI/text change was the addition of this supplemental text to the Options screen. Thoughts?

image

SomeTroglodyte:
The option with a newline works surprisingly OK. Since I nagged that much and the "ready" was predictable I already checked out your branch and the compiler is running...

The first thing I see (in Studio, here I was blind) - `val rankLabels = arrayOf` will also need templates. None of those are already there verbatim. "Military Rank = " yes, "Rank = " no... And that whole when(index) thing still irks a little. If one were to build the thing RankingType as outer, rankLabels inner and transpose afterwards - then inner need not even be a loop, no when no if == 0. But actually coding that leads to a bunch of small snags. Nah, draw a line and leave for optionally later.

By the way - arrayOf() will create an instance  every time the line runs. In this case that's very cheap, but there are cases where moving such a thing up in the hierarchy, possibly all the way to a singleton, would be a worthwhile reduction in allocations. You can get that e.g. with a companion object. Or - make that an enum. Enum.values() then _is_ effectively allocated only once. And  it would read nice:
<details>

```kotlin
    enum class RankLabels { Demographic, Rank, Value, Best, Average, Worst }
    private fun buildDemographicsTable(majorCivs: List<CivilizationInfo>, playerCiv: List<CivilizationInfo>): Table {
        val demographicsTable = Table().apply { defaults().pad(5f) }
        for (rankLabel in RankLabels.values()) {
            if (rankLabel == RankLabels.Demographic) {
                val demoLabel = Table().apply { defaults().pad(5f) }
                demoLabel.add(rankLabel.name.toLabel()).row()
                demoLabel.addSeparator().fillX()
                demographicsTable.add(demoLabel)
                for (category in RankingType.values()) {
                    val headers = Table().apply { defaults().pad(5f) }
                    headers.add(category.name.replace('_', ' ').toLabel()).row()
                    headers.addSeparator().fillX()
                    demographicsTable.add(headers)
                    }
            } else {
                demographicsTable.row()
                demographicsTable.add(rankLabel.name.toLabel())
                for (category in RankingType.values()) {
                    @Suppress("NON_EXHAUSTIVE_WHEN")  //  RankLabels.Demographic treated above
                    when (rankLabel) {
                        RankLabels.Rank -> { //finds the current player's rank in each category
                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
                                if (civ.isCurrentPlayer()) {
                                    val rank = (rankCount + 1).toLabel()
                                    demographicsTable.add(rank)
                                }
                            }
                        }
                        RankLabels.Value -> { //finds the current player's value in each category
                            for (civ in playerCiv) demographicsTable.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX()
                        }
                        RankLabels.Best -> { //finds civ with the best value in each category
                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
                                if (rankCount == majorCivs.indices.first) {
                                    demographicsTable.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX()
                                }
                            }
                        }
                        RankLabels.Average -> { //calculates the average value in each category
                            var totalScore = 0
                            for (civ in majorCivs) totalScore += civ.getStatForRanking(category)
                            val averageScore = totalScore / (majorCivs.lastIndex + 1)
                            demographicsTable.add(averageScore.toLabel())
                        }
                        RankLabels.Worst -> { //finds civ with the worst value in each category
                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
                                if (rankCount == majorCivs.lastIndex) {
                                    demographicsTable.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX()
                                }
                            }
                        }
                    }
                }
            }
        }

        return demographicsTable
    }
```
</details>

SomeTroglodyte (approved):
OK, tested and works. I do see a bunch of possibilities for visual niceties[^1], but unless y'all say the whole thing is not worth the extra 1k jar size, it's fine.

[^1]: Some headers auto-decorated via Stat name - the others could do with manual icons; starts out such that the "Tab" buttons are too far off to the right; Shows dead civs.

letstalkaboutdune:
Thanks again for the feedback @SomeTroglodyte! Should I make the changes you suggested before this gets merged? I agree that using the enum looks nice and I still need to update the template for the rank labels.

SomeTroglodyte:
Last two cents: The commit history _in Studio_ seems to show two authors which probably means you work from two locations/boxes, and one has a slightly different github config. Seems to inadvertently expose your RL name. 
And the missing trailing blank on templates - I fall for that every time myself, so I set it up to be able to run all unit tests locally and every new branch defaults to that run config. Still neglect to run it once a while. A howto should be in my scratchpad.

Have phun!

yairm210 (changes requested):
The 'demographics scoreboard enabled' does not actually change the game, only visualization.
Therefore it should be an *options -> visualization* checkbox and not a 'new game' checkbox.

Reasons:
- Using new demographics table in older games will give much more useful feedback than having to create new games and get to that point
- Different users in multiplayer can decide to show or not, not everyone may share the same preference
- In general, the new game screen has too much going on and we try to make it as simple as possible

android/assets/jsons/translations/template.properties
```diff
@@ -625,7 +625,8 @@ Show minimap =
 off = 
 Show pixel units = 
 Show pixel improvements = 
-Enable nuclear weapons = 
+Enable Nuclear Weapons = 
+Demographics Scoreboard\n(replaces Rankings)" =
```
yairm210, line 629: "Experimental Demographics scoreboard" - this should be for the testing phase only, after which it will be on by defualt

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -95,9 +96,13 @@ class GameOptionsTable(
             { gameParameters.oneCityChallenge = it }
 
     private fun Table.addNuclearWeaponsCheckbox() =
-            addCheckbox("Enable nuclear weapons", gameParameters.nuclearWeaponsEnabled)
+            addCheckbox("Enable Nuclear Weapons", gameParameters.nuclearWeaponsEnabled)
```
yairm210, line 99: We need to be careful with changes like these to avoid losing all existing translations - I'll deal with it on the next version release but this is something we need to be aware of

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +177,89 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+        val playerCiv = gameInfo.civilizations.filter { it.isCurrentPlayer() }
+
+        contentsTable.clear()
+
+        if (gameInfo.gameParameters.demographicsEnabled) contentsTable.add(buildDemographicsTable(majorCivs, playerCiv))
+        else contentsTable.add(buildRankingsTable(majorCivs))
+    }
+
+    private fun buildDemographicsTable(majorCivs: List<CivilizationInfo>, playerCiv: List<CivilizationInfo>): Table {
+        val rankLabels = arrayOf("Demographic", "Rank", "Value", "Best", "Average", "Worst")
```
yairm210, line 189: Are all these translatable?

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +177,89 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                                    val rank = (rankCount + 1).toLabel()
+                                    demographicsTable.add(rank)
+                                }
+                            }
+                        }
+                        2 -> { //finds the current player's value in each category
+                            for (civ in playerCiv) demographicsTable.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX()
+                        }
+                        3 -> { //finds civ with the best value in each category
+                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
```
yairm210, line 220: We don't need to go over all of them, all we need is the first. So just maxBy{}, not for...if

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +177,89 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                    headers.addSeparator().fillX()
+                    demographicsTable.add(headers)
+                    }
+            } else {
+                demographicsTable.row()
+                demographicsTable.add(rankLabels[n].toLabel())
+                for (category in RankingType.values()) {
+                    when (n) {
+                        1 -> { //finds the current player's rank in each category
+                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
```
yairm210, line 209: All you need is the current player, so just currentPlayer.getStatForRanking()

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +177,89 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+        contentsTable.clear()
+
+        if (gameInfo.gameParameters.demographicsEnabled) contentsTable.add(buildDemographicsTable(majorCivs, playerCiv))
+        else contentsTable.add(buildRankingsTable(majorCivs))
+    }
+
+    private fun buildDemographicsTable(majorCivs: List<CivilizationInfo>, playerCiv: List<CivilizationInfo>): Table {
+        val rankLabels = arrayOf("Demographic", "Rank", "Value", "Best", "Average", "Worst")
+        val demographicsTable = Table().apply { defaults().pad(5f) }
+        for (n in rankLabels.indices)   {
```
yairm210, line 191: WAYYY too much indenting in this function, which means it's hard to understand.
Why are you iterating by index instead of by value? The index is meaningless, it's the value that's important!

Separate n==0 out to a separate `for` if you think it's different enough for an if/else

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +177,89 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                        3 -> { //finds civ with the best value in each category
+                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
+                                if (rankCount == majorCivs.indices.first) {
+                                    demographicsTable.add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX()
+                                }
+                            }
+                        }
+                        4 -> { //calculates the average value in each category
+                            var totalScore = 0
+                            for (civ in majorCivs) totalScore += civ.getStatForRanking(category)
```
yairm210, line 228: I'm pretty sure there's an average() function for lists, so majorCivs.map{it.getStatForRanking(category)}.avg() or something should do it

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +177,89 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                            val averageScore = totalScore / (majorCivs.lastIndex + 1)
+                            demographicsTable.add(averageScore.toLabel())
+                        }
+                        5 -> { //finds civ with the worst value in each category
+                            for ((rankCount, civ) in majorCivs.sortedByDescending { it.getStatForRanking(category) }.withIndex()) {
```
yairm210, line 233: Same as above, you only need one, so minBy{} instead of maxBy{}

yairm210 (changes requested):

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
SomeTroglodyte: I remember my idea to save such xlations... But that only shows fixing case for strings should, when possible, be separate PR's also fixing the translations, and be submitted right after a release..

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: Nope, already pointed out

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: > make the changes you suggested before this gets merged

Well here it says "yes"

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: Can't find such an average - sumOf / count would iterate twice

letstalkaboutdune:
Thanks for all the feedback @yairm210! I may have some questions as I work to implement all of your feedback.

letstalkaboutdune:
@yairm210 @SomeTroglodyte I've implemented your feedback. However I have a couple of questions:

@yairm210 regarding the too much indenting - how do you suggest reducing the indentations from the buildDemoGraphicsTable function? I'm not seeing a way around it since I have so much control logic.

Also, what do you think about this? At the beginning of the game when scores are the same, duplicates can show up like below. Is this okay or should there be something to force showing different civs even if the scores are the same?

Duplicates

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,84 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                      
+                        RankLabels.Best -> { // finds civ with the best value in each category
+                            val bestCiv = majorCivs.filter{ it.isAlive() }.maxByOrNull { it.getStatForRanking(category) }!!
+                            demographicsTable.add(getCivGroup(bestCiv, ": " + bestCiv.getStatForRanking(category).toString(), playerCivInfo)).fillX()
+                        }
+                      
+                        RankLabels.Average -> { // calculates the average value in each category
+                            var totalScore = 0
+                            for (civ in majorCivs.filter{ it.isAlive()} ) totalScore += civ.getStatForRanking(category)
+                            val averageScore = totalScore / (majorCivs.filter{ it.isAlive() }.lastIndex + 1)
```
SomeTroglodyte, line 227: `(majorCivs.filter{ it.isAlive() }.lastIndex + 1)` == `majorCivs.count { it.isAlive() }`?

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,84 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                    }
+            } else {
+                demographicsTable.row()
+                demographicsTable.add(rankLabel.name.toLabel())
+
+                for (category in RankingType.values()) {
+                    @Suppress("NON_EXHAUSTIVE_WHEN") // RankLabels.Demographic treated above
+                    when (rankLabel) {
+                        RankLabels.Rank -> { // finds the current player's rank in each category
+                            val rank = majorCivs.filter{ it.isAlive() }.sortedByDescending { it.getStatForRanking(category) }.indexOfFirst { it.civName == gameInfo.currentPlayerCiv.civName } + 1
```
SomeTroglodyte, line 211: `it.civName == gameInfo.currentPlayerCiv.civName` == `it == gameInfo.currentPlayerCiv` ? the civ instances are not cloned while the player can interact, so even a `===` should see the identity?? Dunno. I'd just check in the debugger.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,84 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                val demoLabel = Table().apply { defaults().pad(5f) }
+                demoLabel.add(rankLabel.name.toLabel()).row()
+                demoLabel.addSeparator().fillX()
+                demographicsTable.add(demoLabel)
+                for (category in RankingType.values()) {
+                    val headers = Table().apply { defaults().pad(5f) }
+                    headers.add(category.name.replace('_', ' ').toLabel()).row()
+                    headers.addSeparator().fillX()
+                    demographicsTable.add(headers)
+                    }
```
SomeTroglodyte, line 202: indent of that `}`

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: ... and totalScore probably == `majorCivs.filter { it.isAlive() }.sumOf { it.getStatForRanking(category) }`?

letstalkaboutdune:
Thanks @SomeTroglodyte, I've addressed all these. I'll hold off for any more commits just in case there's more feedback.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,84 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                for (category in RankingType.values()) {
+                    @Suppress("NON_EXHAUSTIVE_WHEN") // RankLabels.Demographic treated above
+                    when (rankLabel) {
+                        RankLabels.Rank -> { // finds the current player's rank in each category
+                            val rank = majorCivs.filter{ it.isAlive() }.sortedByDescending { it.getStatForRanking(category) }.indexOfFirst { it.civName == gameInfo.currentPlayerCiv.civName } + 1
+                            demographicsTable.add(rank.toLabel())
+                        }
+
+                        RankLabels.Value -> { // finds the current player's value in each category
+                            demographicsTable.add(getCivGroup(gameInfo.currentPlayerCiv, ": " + gameInfo.currentPlayerCiv.getStatForRanking(category).toString(), playerCivInfo)).fillX()
```
SomeTroglodyte, line 216: local `fun Table.addRankCivGroup(civ: CivInfo) = add(getCivGroup(civ, ": " + civ.getStatForRanking(category).toString(), playerCivInfo)).fillX()` such that it sees `category` - define inside for - and reuse for 3 of the 5 rows? Maybe, could shorten the code lines here?

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Yes indeed, thank you for these simplifications. :)

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: And caching `majorCivs.filter{ it.isAlive() }` outside the loops could also shorten the code here. Plus - reuses.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,84 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+        
+        if (UncivGame.Current.settings.useDemographics) contentsTable.add(buildDemographicsTable(majorCivs))
+        else contentsTable.add(buildRankingsTable(majorCivs))
+    }
+
+    enum class RankLabels { Demographic, Rank, Value, Best, Average, Worst}
+    private fun buildDemographicsTable(majorCivs: List<CivilizationInfo>): Table {
+        val demographicsTable = Table().apply { defaults().pad(5f) }
+        for (rankLabel in RankLabels.values())   {
+            if (rankLabel == RankLabels.Demographic) {
```
SomeTroglodyte, line 192: The indent level of this if could be saved - rankLabel == RankLabels.Demographic is the static header row, and by pushing the header row block outside the loop and having the loop start one later (`in RankLabels.values().drop(1)` maybe?) you could do without the if? Besides that, the function already looks near fine readability-wise as is.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,84 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                        }
+                      
+                        RankLabels.Worst -> { // finds civ with the worst value in each category
+                            val worstCiv = majorCivs.filter{ it.isAlive() }.minByOrNull { it.getStatForRanking(category) }!!
+                            demographicsTable.add(getCivGroup(worstCiv, ": " + worstCiv.getStatForRanking(category).toString(), playerCivInfo)).fillX()
+                        }
+                    }
+                }
+            }
+        }
```
SomeTroglodyte, line 238: Such a stack of closing braces is a classic readability hindrance. In some projects / languages it's common to comment each with what it is closing - so far haven't seen that in Unciv, but if someone likes it I wouldn't veto. Just an idea.

A tricky idea to _really_ trim that down would be to add 'cell evaluator' lambdas to the RankLabels enum, makes the horizontal loop something akin to `for (category..) { render(category.evaluator(params)) }`, but that enum would then be the ugly part... Nah.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: Wait no getCivGroup is already class-private and seems to be always called with that ": "+ part, even in the older buildRankingsTable, so it could just learn to do some of the repeated calling pattern. Maybe. Would need to see the whole thing in Studio, too lazy.

SomeTroglodyte:
You know what? None of the suggestions actually affect function, so I stand by my approval. If it really has a large optimization potential a near future PR will surely pick it up.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
SomeTroglodyte: > Yes indeed

I trust you _check_ such stuff in some form.

letstalkaboutdune:
> You know what? None of the suggestions actually affect function, so I stand by my approval. If it really has a large optimization potential a near future PR will surely pick it up.

Well I implemented just about all this stuff, will re-commit (since I will be changing machines next time I work on this) and see what you and @yairm210 think. :)

SomeTroglodyte:
> changing machines

Can be done without much ado. Machine A -> commit and push before you leave, then pull from yourself (the online origin) on B, aka 'update' in studio. And vice versa. No need to close PR's or recreate branches. As long as you use the same github account, or give the sub-account enough permissions on your fork.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: I did. Got a couple of save files that I've been using to test. Everything seems to check out as far as I can tell.

letstalkaboutdune:
Yep, I just commit and push on machine A and pull on machine B.

letstalkaboutdune:
Okay, ready for review again. The checkbox has been moved to Options -> Display, making it toggleable during the game. All new and updated text has been added to the translation template.

image

All the code suggestions you both made should be implemented now. I also adjusted how the old scoreboard handles defeated civs. Previously, their values for each category were retained. Now I zero them out so they appear at the bottom of the scoreboard. In the new Demographics scoreboard, dead civs are ignored.

Zeroed Out

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -644,17 +644,20 @@ class CivilizationInfo {
...
-            RankingType.Production -> statsForNextTurn.production.roundToInt()
-            RankingType.Gold -> gold
-            RankingType.Territory -> cities.sumOf { it.tiles.size }
-            RankingType.Force -> getMilitaryMight()
-            RankingType.Happiness -> getHappiness()
-            RankingType.Technologies -> tech.researchedTechnologies.size
-            RankingType.Culture -> policies.adoptedPolicies.count { !Policy.isBranchCompleteByName(it) }
+        return if (isDefeated()) 0
+        else {
+            when (category) {
```
xlenstra, line 649: This is a very minor coding style point, so I'm fine with you not changing it, but I would personally put the `when` directly behind the `else`:
```
else when (category) {
```
and then indent the entire function one level less

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -175,22 +178,90 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+                    val headers = Table().apply { defaults().pad(5f) }
+                    headers.add(category.name.replace('_', ' ').toLabel()).row()
+                    headers.addSeparator().fillX()
+                    demographicsTable.add(headers)
+                }
+            } else {
+                demographicsTable.row()
+                demographicsTable.add(rankLabel.name.toLabel())
+
+                for (category in RankingType.values()) {
```
xlenstra, line 209: Another approach to reduce indentation would be to extract the entirety of this for loop to a different function, `getRankingColumn(rankingType): Table` or something, which computes the column for that ranking type and returns it, so you can add it to the inner table.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
xlenstra: I'd personally just remove `Demographic` from the entire Enum and just handle that column seperately. It's very different from all other columns anyway

letstalkaboutdune:
Now there are no duplicates when finding best or worst, since I do it by sorting the list of civs once and simply pulling out the first or last item.

image

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
yairm210: I agree with xlen on this :)

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
yairm210: One really simple way is to take the Value, Best and Worst and just one-line them - they don't actually need braces :)

yairm210:
Definitely much cleaner than it was! I think once the current open review notes are closed it'll be just about mergable

android/assets/jsons/translations/template.properties
letstalkaboutdune: Done.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Done.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Done.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Moved code for static header row to new function, simplified other code and removed several indentations.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Code simplified to one line with no loops.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Done.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Cached filtered majorCivs outside the loop for reuse and simplified player equality check.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Fixed.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Added local function addRankCivGroup which is reused to grab and format civ data.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Demographic removed from the enum and its static header row moved to new function buildDemographicsHeaders.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
letstalkaboutdune: Done.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
letstalkaboutdune: Done, moved to new function buildDemographicsHeaders.

letstalkaboutdune:
> Definitely much cleaner than it was! I think once the current open review notes are closed it'll be just about mergable

Thanks @yairm210! All feedback has been implemented and reviews resolved, so it's ready for another review.

yairm210 (approved):

Issue #6144, opened 2022-02-11
City Production Focus (UI)


itanasi:
Civ 5 (and others in the series) gave the option to focus all production via population and specialists towards one metric. Either maximize Food, Production, Gold, or Science. And obviously revert to Balanced

Would be a nice feature to have.

Could also include UI to handle locking tiles as well (and unlock all tiles), and ideally locking Specialists too.

xlenstra:
I think we already have a UI for locking/unlocking tiles, though whether tiles are locked is not actually used anywhere as far as I know. Focusing population to a single stat is also already marked as todo in #4697.

itanasi:
We have locking and unlocking individual tiles already. But no button to unlock all tiles yet.

yairm210:
The problem isn't the UI, it's that the feature isn't well defined.
What does 'focuses on X' actually mean?
The main problem here is regarding food, I think. You need a minimum amount of food so the city doesn't starve. 
A way around this would be to try the naive 'each unit of X is worth 2 instead of 1' in rankStatsValue (that means a new focus:Stat? parameter), and if there's enough food then great
If NOT, then take food-heavy tiles until you have enough to break even, and only then prioritize tiles based on the focused stat

itanasi:
I haven't found any code about how the Focus works exactly in Civ 5 as basis. I feel like it does the latter. Make sure you don't starve, then max focused stat, then use standard algo for the rest. I should boot it up and fiddle around

xlenstra:
Looking a bit through the source code distributed with the game, the relevant function appears to be this:
<details> <summary> GetPlotValue() </summary>

```c++
/// What is the overall value of the current Plot?
int CvCityCitizens::GetPlotValue(CvPlot* pPlot, bool bUseAllowGrowthFlag)
{
	int iValue = 0;

	// Yield Values
	int iFoodYieldValue = (/*12*/ GC.getAI_CITIZEN_VALUE_FOOD() * pPlot->getYield(YIELD_FOOD));
	int iProductionYieldValue = (/*8*/ GC.getAI_CITIZEN_VALUE_PRODUCTION() * pPlot->getYield(YIELD_PRODUCTION));
	int iGoldYieldValue = (/*10*/ GC.getAI_CITIZEN_VALUE_GOLD() * pPlot->getYield(YIELD_GOLD));
[...]

Issue #6458, opened 2022-03-29
Strange observation concerning UncivServer


SomeTroglodyte:
**Platform**
You know I'm running Mint 20.3... This mentions running in the debugger, so
Android Studio about says:
```
Android Studio Bumblebee | 2021.1.1 Patch 2
Build #AI-211.7628.21.2111.8193401, built on February 17, 2022
Runtime version: 11.0.11+0-b60-7590822 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Linux 5.13.0-37-generic
GC: G1 Young Generation, G1 Old Generation
Memory: 8192M
Cores: 12
Registry: external.system.auto.import.disabled=true, ide.tooltip.initialDelay=590
Non-Bundled Plugins: org.jetbrains.kotlin (211-1.6.10-release-923-AS7442.40), org.intellij.plugins.markdown (211.7142.37)
Current Desktop: X-Cinnamon
```
Gradle Java is: `OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)`
Run Java is: `OpenJDK Runtime Environment Temurin-16.0.2+7 (build 16.0.2+7)`

**Version**
Current master

**Describe the bug**
I run the game in the Debugger as always, with unchanged run configuration. Large map, not huge, world wrap, max zoom setting = 2. One-click moves are off. Selecting a unit, selecting a destination, waiting for the move icon to show up and then clicking it takes around 6 (Six) seconds. Also, remember we have horrible kludges wrapping Char(Int) and Char.code calls (#6391, #6419, #6435).

Now remove UncivServer <details><summary>as follows:</summary>

```patch
Index: build.gradle.kts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/build.gradle.kts b/build.gradle.kts
--- a/build.gradle.kts	(revision 5ab44dd945af96a46e97e5a48e2ecec3f691c597)
+++ b/build.gradle.kts	(date 1648589933302)
@@ -70,13 +70,6 @@
         }
 
         "implementation"("com.github.MinnDevelopment:java-discord-rpc:v2.0.1")
-
-        // For server-side
-
-        "implementation"("io.ktor:ktor-server-core:1.6.8")
-        "implementation"("io.ktor:ktor-server-netty:1.6.8")
[...]

PR #6650, opened 2022-04-30
Assign Population Improvements

itanasi:
Solves #6144 
With a lot of help from @SomeTroglodyte 
- [x] Make Tile Lock matter for reassignPopulation()
- [x] Reassign on new trigger at start of NextTurn. Population Growth, Improvement completed/pillaged, new Tech
- Make Unassigned Population text also reassign population
- [x] Add Reset Tiles button for quick tile unlocks and assignment
- [x] Add Manual vs Auto Specialists functionality and controls
- Auto switch to Manual when adding/removing Specialists
- Click on text to toggle back to Manual. Need to work on making that clear
- [x] Update Locked icon
- [x] Change manual population override to Lock Tile on first click
- This would match Civ5. But do people really want that?
- [x] Focus Production UI and Equations (stretch goal)
- [x] Avoid Growth (stretch)
- [x] Translation templates

Example:
image

TODOS for future contributions:
- Make ScrollPane to support smaller resolutions, or ExpanderTab
- AI usage of Focuses
- More speed improvements
- JSON the weights of Stats for mod support
- Great Person Focus


itanasi:
Could use feedback on UI text color choices
image
image

itanasi:
Changed to look more like a button
image

SomeTroglodyte:
Nice. I always thought _some day_ I'd try to do something akin to Civ IV's governors... Your "stretch goal" seems to go in that direction. Cheers!

> feedback

- "Copy RGB" sounds like bad practice. We already have too much mix of visuals living in an incomplete skin and hardcoded (guilty myself)... You _can_ extract any skin element easily in code wherever you need it. `BaseScreen.skin.get(TextButton.TextButtonStyle::class.java).fontColor` or somesuch.
- I'd look to get a RoundedEdgeRectangle in there so the "toggle button" is unmistakable - actually _make it_ a TextButton then tweak it to shrink the space needed and make it less obtrusive, less padding, maybe make the RER color.a=0.5f...
- "Reassign on new trigger. Population Growth?" -> Improvement finished or pillaged...
- "Reset Tiles" in that location maybe mistakable to mean "un-own and re-acquire the numer of tiles I'm due" - like an RPG respec. "Reassign worked tuiles[^1]"? "Reset workers"?
- "Make Unassigned Population text also" - an UI label with a functionality is easily missed, unless you know of a common expectation... (guilty myself - e.g. some unmarked pedia links) In this case, it's close. Some _will_ guess without help and be extra happy about their discovery - we're now a <i>5</i>X game.
- https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace
- `posFromEdge + max(tileTable.height, selectedConstructionTable.height)` that max is for convenience, actually only one of the  two is visible at any one time, right? I know, annoying that Gdx doesn't do a true "anchor" like windoze forms does.
- `.tr().toLabel(Color.WHITE)` two redundancies?

[^1]: Some regional "tuiles d'amandes" are sooooooo delicious... (tuiles: frog = shingles, like those on a roof)

itanasi:
Thanks for the feedback. Especially the tip on the skin color. Might be more clear as a really tight TextButton. Need to play with it more.

Updating on Improvements and Tech would be nice, but much more complex to keep track of. Figuring out which City it "belongs" to and then marking it Dirty. Tech would have to always fire before cities and mark all cities Dirty. So more a stretch goal for me right now given dev time.

Triggering the auto assignment is the true UX thing I'm noodling. Reset clears locks, but we also should have an easy way to apply Population without messing with the locks. One indirect way is actually to switch to Auto Specialists, but that's weird. In Civ5 you could click on the Focus options, but likely that will be a dropdown here to save screen space. The UI text is definitely a bit hidden, not to mention small. But haven't figured out a cleaner solution yet.

Using max() was the hack. Only one should be visible, but tracking which is on needs some more surgery.

itanasi:
Semi crazy idea: clicking on the production stats in the UI could expand and reveal the Production Focus options

itanasi:
Uh, can't figure out how to extract a Up button's background color. TextButton only has Color for Font, and Button has Drawable and relies on Batch.getColor()

SomeTroglodyte:
I know what you mean. Either "assume" the `button-c: {color: color}` part of Skin.json and do `skin.get("color", Color::class.java)` or... reflection? No, cast: `(skin.get(Button.ButtonStyle::class.java).up as NinePatchDrawable).patch.color`

I admit, a bit extreme.

itanasi:
The first one works, but is a hack. Would be nice if I could reference the up state and tie that back to the button-c in Skin.json

SomeTroglodyte:
Yes our ninepatches have very limited scalability, by design, especially down. For much smaller radii you have no other option than what the notification scroll does - wrap and scale.

itanasi:
OK, now updating on new Techs and Improvements built near city

itanasi:
Latest version with new Lock icon and buttons
image

itanasi:
Buttons! Highlighting! Changing Focuses!
image

SomeTroglodyte:
Toying with the branch... At first glance, ***verrrry*** nice. Yes, indeeeeed. Which is why I can't help but think it could be nicer. :smile_cat: 

- Citizen Management as a button... Dunno. I'd have begun with a SelectBox hidden in an ExpanderTab up in the CityStatsTable. The reset button shouldn't be that permanent - inside said expander or CitizenManagementTable might be better .. ... ....?
- Can mods do something like "Choose a mayor: {Homer Simpson (default) / Bill Clinton (growth) / Isaac Newton (science) / Uncle Scrooge (Gold) / Ferdinand II of Aragon (Faith)}"? Label+SelectBox could do that.
- Testing with that Civ6 mod some issue foisted on me... :+1: no conflicts I can see. More specialists, so the Citizen management button can overlap CityStatsTable - still works.
- CitizenManagementTable: Icons left of buttons better? My opinion is near 51:49
- Citizen Management button - marginally irritating if clicked repeatedly - unless UI redesign -> hide if CitizenManagementTable visible?
- Reset Citizens button - marginally irritating if clicked repeatedly - cache result of first click for the lifetime of CityScreen, then compare and if equal disable button? Lots of effort for something only slightly cooler.

In case I forgot to say so: Well done.

- `if (surplusFood < 0){ foodYield * 8 } else { foodYield /= 2 }` - is that yours? I mean, * in that place does nothing (very curious there's neiter a warning nor an inspection for this). The whole function tastes icky... If it deals in Stats, why copy out the individual ones into separate vars in the first place? Stats was built for operations like Stats[focus] *= 3 (if focus _were_ a Stat, but one could easily set up mappable enums)
- `autoAssignPopulation` is shockingly slow. Pity `rankTileForCityWork` won't stay static for the duration of the function, which prevents some effective optimizations... One could save one rankTileForCityWork call and one rankSpecialist call per loop iteration, but that's not much.

Did I say this is ***neat***?

SomeTroglodyte:
<details><summary>Food for thought: What could be done with...</summary>

```kotlin
enum class CityFocus(val stat: Stat? = null) {
    NoFocus {
        override fun getStatMultiplier(stat: Stat) = 1f  // actually redundant, but that's two steps to see
    },
    FoodFocus(Stat.Food),
    ProductionFocus(Stat.Production),
    GoldFocus(Stat.Gold),
    ScienceFocus(Stat.Science),
    CultureFocus(Stat.Culture),
    GoldGrowthFocus {
        override fun getStatMultiplier(stat: Stat) = when(stat) {
            Stat.Production, Stat.Food -> 2f
            else -> 1f
        }
    },
    ProductionGrowthFocus {
        override fun getStatMultiplier(stat: Stat) = when(stat) {
            Stat.Production, Stat.Food -> 2f
            else -> 1f
        }
    },
    FaithFocus(Stat.Faith),
    GreatPersonFocus;

    open fun getStatMultiplier(stat: Stat) = when (this.stat) {
        stat -> 3f
        else -> 1f
    }

    fun applyWeightTo(stats: Stats) {
        for (stat in Stat.values()) {
            stats[stat] *= getStatMultiplier(stat)
        }
    }
}
```
</details>

core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
```diff
@@ -0,0 +1,73 @@
...
+        val defaultCell = Table()
+        defaultCell.add(focusLabel)
+        defaultCell.touchable = Touchable.enabled
+        defaultCell.onClick { city.cityAIFocus = CityFocus.NoFocus; city.reassignPopulation(); cityScreen.update() }
+        if (city.cityAIFocus == CityFocus.NoFocus)
+            defaultCell.background = ImageGetter.getBackground(BaseScreen.skin.get("selection", Color::class.java))
+        else
+            defaultCell.background = ImageGetter.getBackground(BaseScreen.skin.get("color", Color::class.java))
+        innerTable.add(defaultCell).growX().pad(3f).row()
+        for ((stat, _) in city.cityStats.currentCityStats) {
```
SomeTroglodyte, line 41: This iterates over the Stat's that the city has > 0f, not all of them. It effectively calls `Stats.asSequence()` through `Stats.iterator()`. Look it up - Stats.kt lines 178, 153. I'd iterate over `CityFocus.values()` and add a `disabled` property to the focus enum for a `continue`, or a function if we wanted to move the `isReligionEnabled` test there. The latter maybe not.

core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
```diff
@@ -0,0 +1,73 @@
...
+            if (stat == Stat.Happiness) continue
+            if (stat == Stat.Faith && !city.civInfo.gameInfo.isReligionEnabled()) continue
+            val label = "${stat.name} Focus".tr().toLabel().addBorder(5f, Color.CLEAR)
```
SomeTroglodyte, line 44: Double `tr()` (toLabel does one) and for the complex governors maybe another property on the enum.

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -638,21 +660,42 @@ class CityInfo {
...
+                population.specialistAllocations.clear()
+            population.autoAssignPopulation(foodWeight)
             cityStats.update()
 
             foodPerTurn = foodForNextTurn().toFloat()
             foodWeight += 0.5f
         }
     }
 
+    fun isFocus(stat: Stat): Boolean {
```
SomeTroglodyte, line 690: The CityInfo regions are already broken, but no need to place this in "state-changing functions" - could be obsolete with my enum fields suggestion anyway

core/src/com/unciv/ui/cityscreen/CityScreen.kt
```diff
@@ -81,10 +92,28 @@ class CityScreen(
 
         //stage.setDebugTableUnderMouse(true)
         stage.addActor(cityStatsTable)
+        val resetCitizensButton = "Reset Citizens".tr().toTextButton()
```
SomeTroglodyte, line 95: Double tr(), citizenManagementButton as well

itanasi:
Appreciate the feedback! My original idea was an expand tab, but did this version for quicker development.

Switching to enum class seems like a much better solution. I obviously was kludging a way to associate the Stat.<type> to the focuses.

The equations are very much not ideal yet and going to need more tweaks.

I don't think I'll put in the effort in this Pull for the AI to also try switching modes. That's what a few of the extra options are really for as I read the Civ5 GnK and Vox Populi equations.

Hopefully time later tonight

SomeTroglodyte:
I'll show you my 'ideas' as code in a few moments.

SomeTroglodyte:
... Nope, sorry, I mismanaged and the code is _gone_. All thanks to me not being able to debug at all without #6458, which I removed from my local copy of your branch the wrong way. Sorry. Irrecoverable except from memory, and I prefer preparing the weekend, RL calls.

Ideas lost:
- CityStatsTable stat icons click selects focus (with CityFocus.valueOf(Stat): CityFocus?)
- Tried .surroundWithCircle instead of square focus indicator - not bad, a bit light
- Moved skin get colors out of loops
- The spurious two Automation.rank* calls in that assign function - can go with `val (best, bestValue) = things.filter(...).associateWith { rank }.maxOrNull { it.value } ?: Map.Entry(null, 0f)` - except one needs a trick to get a Map.Entry compatible (null to 0f)
- Moved reset button into CitizenManagementTable
- Hide citizenManagementButtonHolder when citizenManagementVisible
- folded common code in rankStatsForCityWork and rankSpecialistForCityWork, partially into two CityFocus based functions
- both use a val Stats.clone() in place of those seven vars, finishing with .values.sum()

I'm sure you'll get how specifically from that badly worded brainstorm and the other comments.

itanasi:
Sounds like I have at least two evenings of work then haha

itanasi:
Got most of the ideas done. Ranking equations not quite done. But did muck a bit with adding base weights. May want those JSON-able eventually

I'd rather the Reset Tiles button sit outside the Citizen Table. Most people will be manually moving workers around, locking them, and may need quick access to unlock all tiles. I had considered a SelectBox initially. Difficultly is it would then take 2 actions to change Focus instead of 1 with the pseudo-radio button system now. And with the Citizen Management Button it's very obvious that this feature exists. Did hide when the Table is open though. ExpanderTab is on the TODO.

>     * The spurious two Automation.rank* calls in that assign function - can go with `val (best, bestValue) = things.filter(...).associateWith { rank }.maxOrNull { it.value } ?: Map.Entry(null, 0f)` - except one needs a trick to get a Map.Entry compatible (null to 0f)

Not sure about this one? I don't quite follow

AdityaMH:
Move "reset citizens" to inside border 🤔

SomeTroglodyte:
> don't quite follow

lemme see what I can remember. What are the problems in autoAssignPopulation and downstream?
- `rankStatsForCityWork` and `rankSpecialistForCityWork`have _lots_ of repeated code.
- Both use 7 discrete var's where a `val rankingYields = stats.clone()` as start would be more readable. That thing _is_ designed so property access is fast, a local var wouldn't be that much faster.
- Both have one `foodYield * 8` unused expression doing nothing. Was `*=` intended yes or no? Not sure. I guessed yes.
- `autoAssignPopulation` calls both first within a `.maxByOrNull` not keeping the function result after the comparison, calling it again for the winner instead. One could save the value calculated during the maxByOrNull loop:
```kotlin
// now
            val bestTile: TileInfo? = cityInfo.getTiles()
                    .filter { it.aerialDistanceTo(cityInfo.getCenterTile()) <= 3 }
                    .filterNot { it.providesYield() }
                    .maxByOrNull { Automation.rankTileForCityWork(it, cityInfo, foodWeight) }
            val valueBestTile = if (bestTile == null) 0f
            else Automation.rankTileForCityWork(bestTile, cityInfo, foodWeight)
// saving one rankTileForCityWork call
            val (bestTile, valueBestTile) = cityInfo.getTiles()
                .filter { it.aerialDistanceTo(cityInfo.getCenterTile()) <= 3 }
                .filterNot { it.providesYield() }
                .associateWith { Automation.rankTileForCityWork(it, cityInfo, foodWeight) }
                .maxByOrNull { it.value }
                ?: object : Map.Entry<TileInfo?, Float> {
                    override val key: TileInfo? = null
                    override val value = 0f
                } 
```
I know the fallback given to the Elvis is ugly but it's the best clear and kotlin-y way I found to always initialize both vals without crashing - if max was unsuccessful to null and 0f  respectively. And that's how to save one rankTileForCityWork call because the one used for maxing is stored directly. Dunno whether rankTileForCityWork is expensive enough to justify the ugliness, I'd guess yes.

Oh, and candidate tiles in this block are instantiated by ownership then filtered by distance, but instantiating by distance then filtering by ownership _might_ get cheaper in cities with many tiles if aerialDistanceTo is relatively expensive - Vector2 in HashSet<Vector2> might be far cheaper. Or not. Code like this:
```kotlin
            val tilesDistanceFirst = cityInfo.tileMap.getTilesInDistance(cityInfo.location, 3)
                .filter { it.position in cityInfo.tiles }
            val tilesOwnedFirst = cityInfo.getTiles()
                .filter { it.aerialDistanceTo(cityInfo.getCenterTile()) <= 3 }
            val (bestTile, valueBestTile) = (if (cityInfo.tiles.size > 18) tilesDistanceFirst else tilesOwnedFirst)
                .filterNot { it.providesYield() }
                .associateWith { Automation.rankTileForCityWork(it, cityInfo, foodWeight) }
                .maxByOrNull { it.value }
                ?: object : Map.Entry<TileInfo?, Float> {
                    override val key: TileInfo? = null
                    override val value = 0f
                }
```
... would need thorough measurements to determine if there's a worthwhile gain at all, and where the statistically best turnover point may be.

Too tired to actually inspect your progress, sorry. later.

SomeTroglodyte:
- Consider `fun statToFocus(stat: Stat) = values().firstOrNull { it.stat == stat } ?: NoFocus` - But I would probably call it CityFocus.safeValueOf(stat) because it's so close to what Enum.valueOf(String) does and the name has previous uses. Using CityFocus as 'namespace' works by placing the fun in the Enum's companion object.
- `for (stat in CityFocus.values()) {` - I would rename the var, it's no longer :Stat. Let's say `focus`...
- ... then `val label = "${stat.stat.name} Focus"` -> `focus.label` would IMHO be better. More code - another Enum property and filling it for each instance - but more flexible and easier to maintain
- ... then have NoFocus.label be "Default Focus" and see how much code you can save in CitizenManagementTable.update. Might need another property 'disabled' or 'hidden' instetad of stat==null
- `if (city.cityAIFocus == stat) cell.background =`... -> `cell.background = ImageGetter.getBackground(if (city.cityAIFocus == stat) colorSelected else colorButton)`? Interesting though why Studio does not pop the 'Lift assignment out of if' inspection - probably because the one-liner is 105 chars long.
- `val cell = Table()` if it's not a Cell<Actor> why call it cell? It's practically a Button - which in turn merits a comment why not  use Gdx TextButton (I know why but the next reader will wonder) - a comment which could be a `fun String.toSlimRectangularButton() = Table().apply { add(this.toLabel().addBorder(5f, Color.CLEAR); touchable = Touchable.enabled }` instead - name and whether to actually place it besides the existing .toTextButton and whether to give it an `action: ()->Unit` parameter and/or an `highlight: Boolean` one - up to debate.
- `surroundWithCircle` - I did something like `color = if(cityInfo.cityAIFocus.stat == stat) selected else Color.Clear` so no outer if or onClick duplication (or toggle behaviour) but it only looked OK without tweaking size - likely because I forgot the resizeActor parameter. Food for thought for readability, but your result is likely better.
- Losing foodWeight as parameter is interesting (as is commented out `food = (2.4f +...` or `if.. /=` inversion), I trust you'll do the balancing better than I would. There's two details remaining:  `yieldStats.food * 8` is still an unused expression, and you sure including happiness in rankStatsForCityWork is OK? The old code did it only for specialists, what reason could that have had...

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -19,43 +20,130 @@ import com.unciv.ui.victoryscreen.RankingType
...
+            } else {
+                yieldStats.food /= 2
+            }
+        } else {
+            // NoFocus or Food/Growth Focus. Target +2 Food Surplus
+            if (surplusFood < 2) {
+                yieldStats.food * 8
+            } else if (cityAIFocus != CityFocus.FoodFocus) {
+                yieldStats.food /= 2
+            }
```
yairm210, line 54: There's too much indentation here for me, please remove the brackets on the inner ifelses

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -19,43 +20,130 @@ import com.unciv.ui.victoryscreen.RankingType
...
-            rank += stats.culture / 2
-            rank += stats.gold / 5 // it's barely worth anything at this point
+            //yieldStats.food *= 1.2f
+            yieldStats.science /= 2
+            yieldStats.culture /= 2
+            yieldStats.gold /= 5 // it's barely worth anything at this point
         } else {
-            if (stats.food <= 2 || city.civInfo.getHappiness() > 5) rank += stats.food * 1.2f * foodWeight // food get more value to keep city growing
-            else rank += (2.4f + (stats.food - 2) / 2) * foodWeight // 1.2 point for each food up to 2, from there on half a point
+            //if (stats.food <= 2 || city.civInfo.getHappiness() > 5) yieldStats.food *= 1.2f // food get more value to keep city growing
```
yairm210, line 64: I assume you're going to remove dead comments later but just making sure :)

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -19,43 +20,130 @@ import com.unciv.ui.victoryscreen.RankingType
...
+            } else {
+                yieldStats.food /= 2
+            }
+        } else {
+            // Target +2 Food Surplus
+            if (surplusFood < 2) {
+                yieldStats.food *= 8
+            } else if (cityAIFocus != CityFocus.FoodFocus) {
+                yieldStats.food /= 2
+            }
```
yairm210, line 121: Same here, remove braces from inner if/elses

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -19,43 +20,130 @@ import com.unciv.ui.victoryscreen.RankingType
...
+            // "small city" - we care more about food and less about global problems like gold science and culture
+            yieldStats.food *= 1.2f
+            yieldStats.science /= 2
+            yieldStats.culture /= 2
+            yieldStats.gold /= 5 // it's barely worth anything at this point
+        } else {
+            if (stats.food <= 2 || city.civInfo.getHappiness() > 5) yieldStats.food *= 1.2f // food get more value to keep city growing
+            else yieldStats.food = (2.4f + (stats.food - 2) / 2) // 1.2 point for each food up to 2, from there on half a point
+
+            if (!(city.civInfo.gold < 0 && city.civInfo.statsForNextTurn.gold <= 0))
```
yairm210, line 134: if(!(a && b)) -> if(!a || !b) for better readability

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -19,43 +20,130 @@ import com.unciv.ui.victoryscreen.RankingType
...
+                yieldStats.gold /= 3 // 3 gold is worse than 2 production
+
+            if (!(city.tiles.size < 12 || city.civInfo.wantsToFocusOn(Victory.Focus.Culture)))
```
yairm210, line 137: Same

core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
```diff
@@ -0,0 +1,64 @@
...
+
+        if (!visible) {
+            isVisible = false
+            return
+        }
+        isVisible = true
+
+        val focusLabel = "Default Focus".toLabel().addBorder(5f, Color.CLEAR)
+        val defaultCell = Table()
+        defaultCell.add(focusLabel)
```
yairm210, line 33: Instead of adding a 'clear border' you can defaultCell.add(focusLabel).pad(5f) - saves unnecessary objects

core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
```diff
@@ -0,0 +1,64 @@
...
+        defaultCell.touchable = Touchable.enabled
+        defaultCell.onClick { city.cityAIFocus = CityFocus.NoFocus; city.reassignPopulation(); cityScreen.update() }
+        if (city.cityAIFocus == CityFocus.NoFocus)
+            defaultCell.background = ImageGetter.getBackground(BaseScreen.skin.get("selection", Color::class.java))
+        else
+            defaultCell.background = ImageGetter.getBackground(BaseScreen.skin.get("color", Color::class.java))
```
yairm210, line 39: How about backgroundName = if/else; defaultCell.background = ImageGetter...backgroundName

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
```diff
@@ -35,9 +37,25 @@ class CityStatsTable(val cityScreen: CityScreen): Table() {
...
+                    cityInfo.reassignPopulation(); cityScreen.update()
+                }
+                miniStatsTable.add(icon).size(27f).padRight(5f)
+            } else {
+                icon.add(ImageGetter.getStatIcon(stat.name))
+                icon.onClick {
+                    cityInfo.cityAIFocus = statToFocus(stat)
+                    cityInfo.reassignPopulation(); cityScreen.update()
+                }
+                miniStatsTable.add(icon).size(20f).padRight(5f)
```
yairm210, line 57: Since this happens anyway, remove from if/else

itanasi:
Equations consolidated and updated. Added Avoid Growth feature. All other advice done. I don't like my GPP hack, wish there was a better way to cache it

Haven't done ExpanderTab reorg

SomeTroglodyte:
> Haven't done ExpanderTab reorg

That was just an idea that might not even look good. Leave it for later. Go fix it up. But - I'll be away for almost a day for review.

itanasi:
I need to rethink through the weights a bit more.

The expected behavior is that when you Focus, you maximize that metric (without starving). So all weights should be such that x3 their weight they beat everything else.

However, when you incorporate GPP, Happiness, and the Food bonuses, very quickly Specialists can dominate as the best choice. Especially Merchants late game (+3 Gold, +2 Science is A LOT). Which doesn't feel quite right to me?

SomeTroglodyte:
As long as
a) if you don't use focus the choices aren't worse and
b) if you focus the associated metric is almost guaranteed to not drop

.. and I'd guess that is fulfilled.. Fine-tuning can come later. Heck, the UI with the original weighting (except the focus x3) would already be welcome.

itanasi:
So doing some profiling and this doubles the amount of processing time for autoAssignPopulation(). From my test case save, ~1sec to ~2sec

itanasi:
K, putting it up for review.

Future TODOs by the enterprising:
* Move into a ExpanderTab
* AI integration
* More efficient speedups
* JSON the weights

itanasi:
Cancel that. Thought of an idea to handle Food

itanasi:
Ok, quick sanity check to make sure I'm not doing something dumb.

How about at the start we calculate current Food Surplus based on however many Citizens are assigned. Save this value. Then as we assign Citizens increment Food Surplus by amount generated, and potentially extra for the Specialist Unique. It's not as bulletproof as doing a full recalc. Especially if there's bonuses (Artemis). But gets us close enough imo.

Downside is approximating quirks like Food Cost per Population and doubles some of the logic for Food calculations (terrible tech debt mine). But this prevents doubling the compute time.

itanasi:
It's not an AI aspect.

Adding this in is a decent performance hit because we recalculate all of the City Stats each time we assign a Population. This keeps things up to date and let's us make decisions based on how much Food we're generating so that we can hit >=0 Food (not starving).

So my fix would be instead of calculating all of the Stats, get them at the start then have a rough approximation of Food Generated by tracking the additional food from each Population assigned. Downside is duplicates some logic (Specialist Food bonus), and won't capture special +Food bonuses

SomeTroglodyte:
... can't really help at the moment, not enough resource governor assigned to Unciv ;)

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -89,21 +130,23 @@ object Automation {
...
-            BFS(city.getCenterTile()) { it.isWater || it.isCityCenter() }
+                BFS(city.getCenterTile()) { it.isWater || it.isCityCenter() }
         findWaterConnectedCitiesAndEnemies.stepToEnd()
         if (findWaterConnectedCitiesAndEnemies.getReachedTiles().none {
-                (it.isCityCenter() && it.getOwner() != city.civInfo)
-                        || (it.militaryUnit != null && it.militaryUnit!!.civInfo != city.civInfo)
-            }) // there is absolutely no reason for you to make water units on this body of water.
+                    (it.isCityCenter() && it.getOwner() != city.civInfo)
+                            || (it.militaryUnit != null && it.militaryUnit!!.civInfo != city.civInfo)
+                }) // there is absolutely no reason for you to make water units on this body of water.
```
xlenstra, line 142: Are all these extra indentations android studio auto-formatting? I personally don't like them tbh, and the kotlin style guide agrees with me on that. Its a style thing so this won't affect my approval regardless

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -21,41 +21,82 @@ import com.unciv.ui.victoryscreen.RankingType
...
-            else rank += stats.gold / 3 // 3 gold is worse than 2 production
-
-            rank += stats.production
-            rank += stats.science
-            if (city.tiles.size < 12 || city.civInfo.wantsToFocusOn(Victory.Focus.Culture)) {
-                rank += stats.culture
-            } else rank += stats.culture / 2
+                yieldStats.gold *= 2 // We have a global problem
+
+            if (city.tiles.size < 12 || city.civInfo.wantsToFocusOn(Focus.Culture))
```
xlenstra, line 92: You could look into also taking the `Victory.Focus.Science` and other stats into account

core/src/com/unciv/ui/cityscreen/SpecialistAllocationTable.kt
```diff
@@ -14,7 +15,18 @@ class SpecialistAllocationTable(val cityScreen: CityScreen): Table(BaseScreen.sk
 
     fun update() {
         clear()
-
+        // Auto/Manual Specialists Toggle
+        // Color of "color" coming from Skin.json that's loaded into BaseScreen
+        // 5 columns: unassignButton, AllocationTable, assignButton, SeparatorVertical, SpecialistsStatsTabe
+        if(cityInfo.manualSpecialists) {
+            val manualSpecialists = "Manual Specialists".tr().toLabel().addBorder(5f, BaseScreen.skin.get("color", Color::class.java))
```
xlenstra, line 22: `.toLabel()` already has an internal `.tr()` call, so no need to do it twice

core/src/com/unciv/logic/city/PopulationManager.kt
```diff
@@ -109,34 +111,63 @@ class PopulationManager {
...
+        // Sweden UP
+        for (otherCiv in cityInfo.civInfo.getKnownCivs()) {
+            if (!cityInfo.civInfo.getDiplomacyManager(otherCiv).hasFlag(DiplomacyFlags.DeclarationOfFriendship))
+                continue
+
+            for (ourUnique in cityInfo.civInfo.getMatchingUniques(UniqueType.GreatPersonBoostWithFriendship))
+                cityInfo.currentGPPBonus += ourUnique.params[0].toInt()
+            for (theirUnique in otherCiv.getMatchingUniques(UniqueType.GreatPersonBoostWithFriendship))
+                cityInfo.currentGPPBonus += theirUnique.params[0].toInt()
+        }
```
xlenstra, line 131: Isn't this straight up copy + pasted from `CityInfo:493::getGreatPersonPointsForNextTurn()`? I'd rather extract it to a new method and call that in both places, if we at any point in the future want to change this code we would otherwise have to do it in two places (and probably forget)

core/src/com/unciv/logic/city/PopulationManager.kt
```diff
@@ -109,34 +111,63 @@ class PopulationManager {
...
+        var specialistFoodBonus = 0f
+        for (unique in cityInfo.getMatchingUniques(UniqueType.FoodConsumptionBySpecialists))
+            if (cityInfo.matchesFilter(unique.params[1]))
+                specialistFoodBonus = -(unique.params[0].toFloat() / 100f) * 2f // base 2 food per Pop, and need to inverse negative
```
xlenstra, line 135: We set this each time instead of it adding together?

xlenstra (commented):
Few nitpicky things, but without removing the code duplication for calculating great person points I can't approve this. I'll also test it once that is addressed to see how the system works in practice.

It also seems like the AI doesn't actually choose a focus. Is that intentional (I'm fine with that, this is already an improvement as is) or are you still planning something with that?

core/src/com/unciv/logic/automation/Automation.kt
xlenstra: Though that might be more something for AI and not for players

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Yeah, did an Android Studio auto format pass. Can clean up a bit

SomeTroglodyte: I even think the defaults for that changed ArcticFox - Bumblebee - Chipmunk. But - we could make these settings part of the project under version control.

core/src/com/unciv/ui/cityscreen/SpecialistAllocationTable.kt
SomeTroglodyte: Similar to the little hint [in my comment 17 days back](https://github.com/yairm210/Unciv/pull/6650#issuecomment-1113937648)

SomeTroglodyte:
> the AI being able to pick a Focus

And it shouldn't, it should simply optimize without prejudice. If the code here helps, fine. But that's better left for a later PR.

> I'll also test it

Just for reference, I tested the state of the branch about 11 days ago and it was stable, helpful and from what I could see not breaking the game's rules.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -21,41 +21,82 @@ import com.unciv.ui.victoryscreen.RankingType
...
-            else rank += (2.4f + (stats.food - 2) / 2) * foodWeight // 1.2 point for each food up to 2, from there on half a point
+            if (cityAIFocus != CityFocus.NoFocus && cityAIFocus != CityFocus.FoodFocus && cityAIFocus != CityFocus.ProductionGrowthFocus && cityAIFocus != CityFocus.GoldGrowthFocus) {
+                // Focus on non-food/growth
+                if (surplusFood < 0)
+                    yieldStats.food *= 8 // Starving, need Food, get to 0
+                else
+                    yieldStats.food /= 2
+            } else if (!city.avoidGrowth) {
+                // NoFocus or Food/Growth Focus. Target +2 Food Surplus
+                if (surplusFood < 2)
```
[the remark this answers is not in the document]
itanasi: The point is to *8 until you hit 2 Surplus. That is, rank Food as the ultimate overriding need, until+2 Surplus. Then the weights carry it from there in a more balanced fashion. Or using Food Focus still gives a *3

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -302,6 +351,8 @@ class CityInfo {
         toReturn.isOriginalCapital = isOriginalCapital
         toReturn.flagsCountdown.putAll(flagsCountdown)
         toReturn.demandedResource = demandedResource
+        toReturn.updateCitizens = updateCitizens
+        toReturn.cityAIFocus = cityAIFocus
```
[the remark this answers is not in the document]
itanasi: Good catch

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -21,41 +21,82 @@ import com.unciv.ui.victoryscreen.RankingType
...
+        if (cityInfo.getRuleset().specialists.containsKey(specialist)) { // To solve problems in total remake mods
+            val specialistInfo = cityInfo.getRuleset().specialists[specialist]!!
+            gpp = specialistInfo.greatPersonPoints.sumValues().toFloat()
+        }
+        gpp = gpp * (100 + cityInfo.currentGPPBonus) / 100
+        rank += gpp * 3 // GPP weight
+        return rank
+    }
+
+    private fun rankStatsForCityWork(stats: Stats, city: CityInfo, specialist: Boolean = false): Float {
```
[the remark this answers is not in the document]
itanasi: Done

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Done

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Or at least mention in the Wiki. Done on my editor and fixed

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -34,6 +35,49 @@ enum class CityFlags {
...
+        override fun getStatMultiplier(stat: Stat) = 1f  // actually redundant, but that's two steps to see
+    },
+    FoodFocus("${Stat.Food.name} Focus", true, Stat.Food),
+    ProductionFocus("${Stat.Production.name} Focus", true, Stat.Production),
+    GoldFocus("${Stat.Gold.name} Focus", true, Stat.Gold),
+    ScienceFocus("${Stat.Science.name} Focus", true, Stat.Science),
+    CultureFocus("${Stat.Culture.name} Focus", true, Stat.Culture),
+    GoldGrowthFocus("Gold Growth Focus", false) {
+        override fun getStatMultiplier(stat: Stat) = when (stat) {
+            Stat.Production, Stat.Food -> 2f
```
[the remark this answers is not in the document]
itanasi: Yup

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -585,7 +636,11 @@ class CityInfo {
         tryUpdateRoadStatus()
         attackedThisTurn = false
 
-        if (isPuppet) reassignPopulation()
+        if (isPuppet) reassignAllPopulation()
+        if (updateCitizens){
+            reassignPopulation()
```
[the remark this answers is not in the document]
itanasi: Good catch. Changed

core/src/com/unciv/logic/city/PopulationManager.kt
itanasi: Pulled out now

core/src/com/unciv/logic/city/PopulationManager.kt
itanasi: Good catch. Should be cumulative multiplication, same as in CityStats. Fixed

core/src/com/unciv/logic/city/PopulationManager.kt
```diff
@@ -109,34 +111,63 @@ class PopulationManager {
...
-                if (bestTile != null)
+                if (bestTile != null) {
                     cityInfo.workedTiles = cityInfo.workedTiles.withItem(bestTile.position)
-            } else if (bestJob != null) specialistAllocations.add(bestJob, 1)
+                    cityInfo.currentSurplusFood += bestTile.getTileStats(cityInfo, cityInfo.civInfo)[Stat.Food]
+                }
+            } else if (bestJob != null) {
+                specialistAllocations.add(bestJob, 1)
+                cityInfo.currentSurplusFood -= specialistFoodBonus
+            }
```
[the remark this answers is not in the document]
itanasi: Logic bug!

core/src/com/unciv/ui/cityscreen/SpecialistAllocationTable.kt
itanasi: Yup

itanasi:
Hookay. All comments sorted

core/src/com/unciv/logic/city/PopulationManager.kt
```diff
@@ -109,34 +112,49 @@ class PopulationManager {
...
+                valueBestSpecialist = Automation.rankSpecialist(bestJob, cityInfo)
             }
 
             //assign population
             if (valueBestTile > valueBestSpecialist) {
-                if (bestTile != null)
+                if (bestTile != null) {
                     cityInfo.workedTiles = cityInfo.workedTiles.withItem(bestTile.position)
-            } else if (bestJob != null) specialistAllocations.add(bestJob, 1)
+                    cityInfo.cityStats.currentCityStats[Stat.Food] += bestTile.getTileStats(cityInfo, cityInfo.civInfo)[Stat.Food]
```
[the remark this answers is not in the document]
itanasi: It should run and clean up my temp use of it at the end when it runs the update(). I opted to just reuse this instead of copying to a Transient Stats variable in CityInfo.

The only stat that the Automation needs is Food. Replicating all of the stats if a lot of redundant code. I could run the update() after each Citizen is assigned, but that doubled the runtime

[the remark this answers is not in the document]
itanasi: I can do the change later tonight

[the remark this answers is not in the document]
itanasi: Done

SomeTroglodyte:
- Getting `Gold Growth Focus` and P G Focus into templates
- Getting the lock into the atlas

..Later? Those Foci - they weren't actually shown when last I played this. I'll have to give it another short spin.

> `Conversation 91 Commits 39 Checks 1 Files changed 15`

Getting that first counter to exactly :100: the moment of the merge will give an extra special reward by the github team - that's what you mean by ding ding ding? Or only if all four reach 100, thus 4 times the emoji?

SomeTroglodyte (approved):
Tests just fine. The atlas isn't critical as there's the old "Lock" in exactly the same place of the texture; and the un-templated strings indeed aren't visible.

itanasi:
> * Getting `Gold Growth Focus` and P G Focus into templates
> 
>     * Getting the lock into the atlas
> 
> 
> ..Later? Those Foci - they weren't actually shown when last I played this.

Figured the atlas would regenerate. Easy fix to push a compiled one.

As for the spare Foci, those are more for the AI. Too many options in the Table will make it waaaay too big for smaller resolutions (already pretty big, would like ScrollPane to work but couldn't quite get it last night).

itanasi:
I mean I figured the release process would recompile those assets. Otherwise sorta annoying to keep merge conflict fixing. Whatever

itanasi:
> > Otherwise sorta annoying to keep merge conflict fixing
> 
> You don't, you just do either `pick remote` or `pick ours` and just regenerate.

Correct, just did that. I just was assuming someone/something else was doing a quick regeneration as part of release

SomeTroglodyte:
Nobody volunteered to make that 100th discussion entry? Aw.

All right, here it is. Grace period expired, in this goes.

PR #6652, opened 2022-04-30
Fixed a crash when a plane tried to enter a full city

xlenstra:
This would only occur if the plane was damaged and looking for a spot to heal, and had no tile to go to with >0 healing. It would then try to head towards the nearest city, regardless of whether it could enter that. As it's an air unit, 'head towards' means 'fly there directly this 
turn', which obviously caused crashes if the unit couldn't enter that tile.

On a glance it looks like this entire function is not designed to handle air units, as these don't have to care about ranged or bombarding units at all anyway and also doesn't use the normal movement rules. I've only patched this crash, but realistically, this entire function should be redesigned to also handle air units.


SomeTroglodyte (approved):
So "empty glass" is the pessimist's euphemism for "not quite full to the brim glass"?
You know what - approved. This time that doesn't mean I did everything necessary to convince me this is an improvement without regressions, but, hey, nothing really suspicious stands out :ghost: !

PR #6653, opened 2022-04-30
Update libGDX version

### LINKED ISSUE #6609: Can't run the game in Mac OS with M1 CPU
**Platform**
Mac OS, Darwin Kernel Version 21.3.0, AArch64 (Apple Silicon (M1))

**Version**
4.0.9 compiled by myself and from github releases

**Describe the bug**
I just can't run the game:(

**To Reproduce**
1. Get Unciv.jar
2. Download liblwjgl.dylib and put it in folder with game (my first error was missed liblwjgl.dylib)
3. Run `java -XstartOnFirstThread -Djava.awt.headless=true -jar Unciv.jar`

**Expected behavior**
The game just needs to start

**Screenshots**
image

**Additional context**
I tried to run the game with different versions of java and even with Temurin, as a result I have the same stdout + stderr

adatan:
Fixes #6609

### DIFF
--- a/buildSrc/src/main/kotlin/BuildConfig.kt
+++ b/buildSrc/src/main/kotlin/BuildConfig.kt
@@ -6,6 +6,6 @@ object BuildConfig {
     const val appCodeNumber = 702
     const val appVersion = "4.0.10-patch1"
 
-    const val gdxVersion = "1.10.0"
+    const val gdxVersion = "1.10.1-SNAPSHOT"
     const val roboVMVersion = "2.3.1"
 }


SomeTroglodyte:
The question was meant to incite other devs to comment - maybe a base library upgrade _is_ reason enough. Still, I'll "Approve and run" the workflow, curious as to whether the Github VM's will be able to pull the snapshot too.

How extensively did you test? Desktop Mac obviously. Any others? Such a step calls for code coverage, or at least exercising every UI at least once. And other Gdx services - IO, Input...

SomeTroglodyte:
First problem I see: Sorting icons in mod manager missing, a font thing, and from what I can tell already regressed before this.

Azzurite:
I would advise against depending on SNAPSHOT versions, especially in production. Unless we want to help the GDX guys with some testing of course :D

SomeTroglodyte:
> help the GDX guys

Which they could really use some more of...

I playtested for an hour or so and found nothing out of order. I only refrained from formally approving as you're right there's a kind of political aspect to this and there I'm ambivalent.

yairm210:
I looked into it a bit.To the best of my understanding, this is de facto a nightly build, that means it could change at any moment, potentially breaking our game with it.
I understand the need but it's not worth the risk of breaking the game for everyone.

PR #6657, opened 2022-04-30
Adds a debug option to simulate up to some amount of turns

xlenstra:
Adds the option for simulating turns to the developer debug menu.
image


SomeTroglodyte:
This sounds like not enough sleep. Just put the control right on the getDebugTab Table. If you think the AskNumberPopup offers too much value, instantiate one, don't open, steal its innerTable (if that was the name), then discard.

yairm210 (approved):

PR #6664, opened 2022-05-01
Fixed `provides [amount] [resource]` on buildings not accepting conditionals

xlenstra:
This was due to it still using `getLocalMatchingUniques` without providing a `StateForConditionals`, and then manually filtering them if they didn't provide conditionals. This of course doesn't work anymore, as `getLocalMatchingUniques` then assumes that all conditionals aren't satisfied.
Also refactored a loop over `uniqueObjects` with `isOfType()` to just use `getMatchingUniques()`.


SomeTroglodyte:
Looks "harmless"...
Two loops instead of one - pico performance hit -> OK, neglect - maybe one copy of StateForConditionals instead of two?

SomeTroglodyte (approved):

PR #6666, opened 2022-05-01
Fixed game freeze when trying to start a mp game with no internet connection

### LINKED ISSUE #6649: Game freezes when a Network Error occurs when creating a new Game on Multiplayer Servers
**Platform**
Raspberry Pi OS armhf
Android 6
(Tested on two separate devices)

**Version**
4.0.10-patch1

**Describe the bug**
I was trying to create a game on my local server. Suddenly an error 'Couldn't Upload Game' was shown and the game froze. Later I Discoverd that I actually had the server turned off.
The close button on error ui was not working.

Later I discovered that this happens every time a Network error occurs when creating a new game with on a Multiplayer Server. (Not sure about Dropbox)

**To Reproduce**
1. Turn off your internet
2. Try to create a new game on a Multiplayer Server

alexban011:
Fixes #6649 
prevents people from starting a new multiplayer game if an internet connection is not available.

### DIFF
--- a/core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
+++ b/core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
@@ -11,18 +11,19 @@ import com.unciv.UncivGame
 import com.unciv.logic.*
 import com.unciv.logic.civilization.PlayerType
 import com.unciv.logic.map.MapType
+import com.unciv.logic.multiplayer.OnlineMultiplayer
 import com.unciv.models.metadata.GameSetupInfo
 import com.unciv.models.ruleset.RulesetCache
 import com.unciv.models.translations.tr
-import com.unciv.ui.pickerscreens.PickerScreen
-import com.unciv.ui.utils.*
-import com.unciv.logic.multiplayer.OnlineMultiplayer
 import com.unciv.ui.crashhandling.crashHandlingThread
 import com.unciv.ui.crashhandling.postCrashHandlingRunnable
 import com.unciv.ui.images.ImageGetter
+import com.unciv.ui.pickerscreens.PickerScreen
 import com.unciv.ui.popup.Popup
 import com.unciv.ui.popup.ToastPopup
 import com.unciv.ui.popup.YesNoPopup
+import com.unciv.ui.utils.*
+import java.io.IOException
 import java.util.*
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
@@ -37,6 +38,12 @@ class NewGameScreen(
     private val newGameOptionsTable: GameOptionsTable
     private val playerPickerTable: PlayerPickerTable
     private val mapOptionsTable: MapOptionsTable
+
+    @Throws(InterruptedException::class, IOException::class)
+    fun isConnected(): Boolean {
+        val command = "ping -c 1 github.com"
+        return Runtime.getRuntime().exec(command).waitFor() == 0
+    }
 
     init {
         updateRuleset()  // must come before playerPickerTable so mod nations from fromSettings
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
```diff
@@ -38,6 +39,12 @@ class NewGameScreen(
     private val playerPickerTable: PlayerPickerTable
     private val mapOptionsTable: MapOptionsTable
 
+    @Throws(InterruptedException::class, IOException::class)
+    fun isConnected(): Boolean {
+        val command = "ping -c 1 github.com"
```
SomeTroglodyte, line 44: I'm all for Wind-hose bashing, but I still know -c is a routing compartment on Woodledoze boxes, so that will fail unless the process has admin and elevation. And github is not the most reliable round-the-clock icmp partner. And does that even start on non-rooted Android? Even if, pinging is highest permissions on droid too.

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
```diff
@@ -71,6 +78,13 @@ class NewGameScreen(
         rightSideButton.setText("Start game!".tr())
         rightSideButton.onClick {
             if (gameSetupInfo.gameParameters.isOnlineMultiplayer) {
+                if(!isConnected()){
+                    val invalidPlayerIdPopup = Popup(this)
```
SomeTroglodyte, line 82: Name describes something else

SomeTroglodyte:
Welcome by the way. Good idea, but I fear correct platform-independent implementation will require a whole different kettle of fish (see all those proxies for platform-specific code in UncivGameParameters) - unless Gdx has already done the abstraction and offers some tool.

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
alexban011: I tested it on a non-root galaxy A52S and it worked with no problem
You're right about github.com, maybe use 8.8.8.8 instead? (the Google DNS)
I was thinking of making it ping 2 sites and consider it offline only if both fail... maybe 8.8.8.8 and 1.1.1.1

SomeTroglodyte: New, on Droid there are platform-specific API's for exactly this usecase and those should be used. And as I said - fails on Win:vomiting_face: , and W-can't say it - is still a supported platform.

alexban011: > Naw, on Droid there are platform-specific API's for exactly this usecase and those should be used. And as I said - fails on Win🤮 , and W-can't say it - is still a supported platform.

ohhh I see, I'll try to fix that.
Would checking for the OS be a good way to fix this? for android it will run -c and for windows -n

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
alexban011: I'm not quite sure what you mean here :)

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
SomeTroglodyte: For starters, yes. Detection is already somewhere in the codebase, look for.... CrashReportSysInfoDesktop. That one can also be used to learn how platform-specific stuff is handled around here - I mean how it is called and where it lives and so on.
And droid should start at: https://developer.android.com/reference/android/net/ConnectivityManager

alexban011: I'm running into issues. I try to import android.net.ConnectivityManager and it gives me "unresolved refrence: net". Even if I try android.* it says "unresolved refrence: android". I do have the right sdk and sdk tools installed and after multiple tries of invalidating caches and resyncing gradle files it still doesn't work. Do you happen to have an idea of whats going on here?

SomeTroglodyte:
Probably trying to from the core module. You work out how e.g. CrashReportSysInfo, CrashReportSysInfoAndroid and CrashReportSysInfoDesktop are tied together and how CrashScreen.formatReport uses them? Just one example.

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
SomeTroglodyte: See linked PR - import is no issue at all within the android module.

PR #6668, opened 2022-05-01
Use the scrolling for the oversized popups

### LINKED ISSUE #5046: City conquest screen doesn't scroll.
**Version**
Play Store, 3.16.7

**Describe the bug**
The city conquest options ("Annex", "Raze", "Puppet", "Liberate") don't scroll for me.

In Landscape orientation on a small phone, much of the text was clipped, although the main buttons were still visible.

**To Reproduce**
1. Conquer a city.

**Expected behavior**
Scroll view, like in the rest of the UI.

### LINKED ISSUE #5811: Shush now. (Non-scrollable text pushes button below screen in Popup, can't start game.)
Lmao. I mean… This is clearly a problem, right?

image

You can still get out of this by pressing the OS "Back" button, but there should probably be an `AutoScrollPane`.

See: #5046.

New issue because I'm not sure the same solution would apply— Wrapping *all* `Popup`s in `ScrollPane`s might break other popups that have their own inner scroll, and ideally it should only be the content body that scrolls (so the title and buttons are always visible). Perhaps a `Popup` constructor argument that defaults to `true`, with a comment saying it's better for specific Popups to implement scrolling themselves on only the variable-length part?

JackRainy:
Fixes #5811 and, probably, fixes #5046

The fix is simple and elegant, however, I feel it needs more testing than just a few popups.

<details><summary>See video here</summary>

Screencast 2022-05-01

</details>

As far as you see, the small popups are not affected and behave as usual (no scrolling), however, large oversized ones are able to be scrolled, so the large text on the game's start is not blocker anymore.

### DIFF
--- a/core/src/com/unciv/ui/popup/Popup.kt
+++ b/core/src/com/unciv/ui/popup/Popup.kt
@@ -28,12 +28,14 @@ open class Popup(val screen: BaseScreen): Table(BaseScreen.skin) {
         // Set actor name for debugging
         name = javaClass.simpleName
 
+        val scrollPane = AutoScrollPane(innerTable, BaseScreen.skin)
+
         background = ImageGetter.getBackground(Color.GRAY.cpy().apply { a=.5f })
         innerTable.background = ImageGetter.getBackground(ImageGetter.getBlue().darken(0.5f))
 
         innerTable.pad(20f)
         innerTable.defaults().pad(5f)
-        super.add(innerTable)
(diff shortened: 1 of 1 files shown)


SomeTroglodyte (approved):
Sounds too easy to just work like that, right?

JackRainy:
> Sounds too easy to just work like that, right?

Haha, yes, exactly! )))

PR #6670, opened 2022-05-01
Font choice rework

SomeTroglodyte:
#### Features
- Font list significantly shorter
        - By doing families instead of individual fonts
        - By trying to skip Script-Specific fonts if and only if the current locale has no interest in them
- Selecting a Font activates immediately (screen flashes a bit due to still needing to discard all screens and start over)

#### Notes
- Immediately switching Font is done through reflection - best hack I could come up with.
- The old code mixed font and family - you could select Bold and get Regular or vice versa. I chose one direction, the other would also have been possible, 
- Some acceleration should be there as one static cache is now reused as opposed to being used but once previously (droid only)
- On droid, the app will notice a change in system primary Language choice while running, but not installation of additional fonts. Secondary languages are ignored. Tested with "עִברִית" - offers two fonts more, not that using them could really be that useful, after all we lack a translation for that. Seemed easier than testing JP.
- Most intelligence lives in the platform-specific NativeFontImplementations - both a bit kludgy as neither platform seems to have helpful APIs. e.g. Droid's font list - the come with locales, and those have the same type as system locales, but there's zero correspondence. A font's locales break all ISO rules, they have a Script property but no language and toString to "unk-Xxxx"... and getFirstLocale will happily return a locale not even contained in the font's locale list... And so on.
- Originally, I tried to come up with a test to see whether the mod manager sorting symbols are present in a font - fail, all fonts I tested that render nothing for those code points still claim to _have_ the code, provide dimensions and advancement - but the rendered bitmap is empty.

#### Contains code common with #6643 (but not as common commit) -> after merging one the other _will_ have conflicts. [^1]

[^1]: It's the way the advanced options page is thread-decoupled, the fastlane PR also re-enables the quick O-F12 for a translation run and this one also needs the advanced tab to be created right during the constructor else the "show font immediately" screen reload lands on an empty tab.


android/src/com/unciv/app/NativeFontAndroid.kt
```diff
@@ -6,20 +6,42 @@ import android.graphics.Paint
...
+                val name = file?.nameWithoutExtension ?: return Int.MAX_VALUE
+                if (name == family) return 0
+                if (!name.startsWith("$family-")) return Int.MAX_VALUE
+                if (style.weight == FontStyle.FONT_WEIGHT_NORMAL && style.slant == FontStyle.FONT_SLANT_UPRIGHT) return 1
+                return 2 +
+                        abs(style.weight - FontStyle.FONT_WEIGHT_NORMAL) / 100 +
+                        abs(style.slant - FontStyle.FONT_SLANT_UPRIGHT)
+            }
+            val font = fontList.mapNotNull {
+                val q = it.matchesFamily(fontFamily)
```
yairm210, line 42: Rename q to something more significant

android/src/com/unciv/app/NativeFontAndroid.kt
```diff
@@ -64,13 +82,35 @@ class NativeFontAndroid(private val size: Int, private val fontFamily: String) :
...
+            val dashPos = indexOf('-')
+            if (dashPos < 0) return this
+            return this.substring(0, dashPos)
         }
+
+        // To get _all_ Languages a user has in their Android settings, we would need more help
+        // from the launcher: (Activity).resources.configuration.locales
+        val languageTag = Locale.getDefault().toLanguageTag()  // e.g. he-IL, corresponds to the _first_ Language in Android settings 
+        val supportedLocales = arrayOf(languageTag, "en-US")
+        val supportedLanguages = supportedLocales.map { it.take(2) }
```
yairm210, line 99: I'm confused. What does this line achieve?

yairm210:
Nice job with the options pop-up :)
I can't say I completely understand all the font x locale logic but it looks like you've accounted for more edge cases than I have

android/src/com/unciv/app/NativeFontAndroid.kt
SomeTroglodyte: String.take(2) is a non-throwing substring(0,2) to make "en-US" -> "en", and somewhere I stumbled over a string shorter than 2 chars, so the substring threw the Exception, but I don't remember the details. May have been elsewhere and I just stored `String.take` as equivalent to .net Left() in my brainbox.

android/src/com/unciv/app/NativeFontAndroid.kt
SomeTroglodyte: Of course, sounds like an interim never meant to stay. Still busy on another branch that like this one started from a "that's easy" trigger and got way complicated really fast... Soon.

SomeTroglodyte:
Need a bit to fix a regression: The "Default" entry shows as empty, shouldn't be.

<details><summary>A clip in the meantime</summary>

https://user-images.githubusercontent.com/63000004/166604623-a3b942f5-d718-4283-9080-4fa7e37479d5.mp4
</details>

Shows the bug, and how one can recover from mistakes easier than before. The pauses are due to me looking for "Oldania" but forgetting the name...

SomeTroglodyte:
So, turns out there was a **_entertaining lesson_** to be learned. Kotlin equality is not Java library equality, and actually having a library function to offer a choice between using (Java) `==` and an `equals()` call (com.badlogic.gdx.utils.Array.contains) won't always help.

PR #6672, opened 2022-05-01
Unciv Adaptive Android Icon

SomeTroglodyte:
Inspired by #6606 I did these a while ago. See clip over there. Would be a pity to just ditch.

#### References used
- https://developer.android.com/studio/write/image-asset-studio#launcher
- https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive
- https://developer.android.com/studio/write/image-asset-studio#create-notification
- https://developer.android.com/training/notify-user/build-notification

Also copies the 192² round version to fastlane - from what I can see, devs distribute almost evenly circles and rounded rectangles on fdroid, slight trend to circle.

#### Open questions:
- Papyrus-inspired background color agreeable? For modern devices, a change is just one xml; only legacy and now fastlane use pre-rendered.
- Cleanup unused? 'Unciv icon v3 for google play' is _not_ on Google play...
- Integration into non-fdroid publishing?


yairm210:
Clean up unused - sure
Non-fdroid - interesting, I wonder if Butler for itch has image uploads as well?

PR #6677, opened 2022-05-02
Internet connectivity test before starting a new mp game

alexban011:
It checks for internet connection before starting a new mp game to avoid #6649 

Many thanks to @SomeTroglodyte for most of the code and for explanations (#6675 )

PR #6679, opened 2022-05-02
check for internet before starting a mp game to avoid freeze

alexban011:
Fixed issue #6649 by checking for internet connection before starting the game if "online multiplayer" is selected.

Many thanks to @SomeTroglodyte, again,  for most of the code and for explanations. 
A slight modification has been done to #6675 to catch a crash caused by InetAddress on Windows and to display a Popup when trying to create a game with no access to the Internet, like the popup when the player ID is invalid

The difference between this and #6677 is the crash prevention and code cleanup


SomeTroglodyte:
Yup, probably best to catch - but what kind of Exception did you see???

And surprise: `Exception` is not the top-level catching everything, that would be `Throwable`. We all learned that here in this project a while ago... I think somewhere around the Dropbox code...

And take a look at (https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace): `try {` etc... Or somebody else will 'lint' that the next time they come around. Normal procedure, but since you're learning...

alexban011:
> Yup, probably best to catch - but what kind of Exception did you see???
> 
> And surprise: `Exception` is not the top-level catching everything, that would be `Throwable`. We all learned that here in this project a while ago... I think somewhere around the Dropbox code...
> 
> And take a look at (https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace): `try {` etc... Or somebody else will 'lint' that the next time they come around. Normal procedure, but since you're learning...

I posted the error on pastebin to keep it short here: https://pastebin.com/T8qNN8dc

SomeTroglodyte:
> pastebin

"Unrecognized Windows Sockets error: 11050: no further information" is the only interesting part... Maybe explicitly asking for an IPV4 address helps? Wait - that code wouldn't be able to handle a box where _only_ IPV6 is bound. Though such a box might trip tons of other software too. Querying interfaces just to decide on a ping target is no fun.

alexban011:
> And surprise: `Exception` is not the top-level catching everything, that would be `Throwable`. We all learned that here in this project a while ago... I think somewhere around the Dropbox code...
> 
> And take a look at (https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace): `try {` etc... Or somebody else will 'lint' that the next time they come around. Normal procedure, but since you're learning...

I made changes for these 2 and thanks again!

alexban011:
> > pastebin
> 
> "Unrecognized Windows Sockets error: 11050: no further information" is the only interesting part... Maybe explicitly asking for an IPV4 address helps? Wait - that code wouldn't be able to handle a box where _only_ IPV6 is bound. Though such a box might trip tons of other software too. Querying interfaces just to decide on a ping target is no fun.

After searching on the internet for this I saw some people having issues with VPNs. Then I enabled my vpn on my phone but kept the data/wifi turned off and it looks like its freezing again like before. (That is only if you disable your internet while you're connected to the VPN)

SomeTroglodyte (approved):

SomeTroglodyte:
VPN - yup, the new helpers _could_ detect that if we needed that. Saw some VPN flag in the Andoid classes, and nmcli would show them, and W has some API somewhere not far from what you need to do to determine what's metered what's not. I almost did something to limit metered traffic months ago when I had that splashscreen-displayig mod manager in the works. But - if you pull out the underlying transport from under a VPN it'll die. Quickly with keepalives configured, or else it could be a zombie for a long time, but it'd be dead. No matter. I'll leave further experiments to you.

xlenstra (approved):

SomeTroglodyte:
> used properly

proper or improper - you can check such links or create them on the right side under "Development"

Issue #6576, opened 2022-04-19
"Creates a [tileFilter] improvement on a specific tile" doesn't work


ArchDuque-Pancake:
Trying to use the unique of "Creates a [tileFilter] improvement on a specific tile", and the game reads it, but upon **clicking one of the green circles the game crashes**.

**To Reproduce**
1. Add the unique to a building, with any improvement: `"Creates a [Farm] improvement on a specific tile"`
2. New game, construct the building
3. When clicking one of the green circles, see crash

**Crash Log**

_(Unciv discord server, bugs channel)_
https://discord.com/channels/586194543280390151/588357826758574106/966057960067432468

**Expected behavior**
Game shouldn't crash _obviously_, and it should then place the improvement on the selected tile when the building finishes.

SomeTroglodyte:
Well, the code looks a bit unfinished. There's at least one expression evaluated without side effects or storing the results... Blame says Yair's work 14 months ago. Might have regressed in that time. Still, I would doubt at least one design decision: Which tile was selected is stored as improvementInProgress with turnsToImprovement==-1 and no _direct_ association with the construction queue entry. So, even with the crash fixed, I'd expect weird behaviour when a Worker lands on such a tile, or when queueing several buildings that create a farm and then choosing the same tile for all of them and so on.

Is this reproducible with the public+current version of Civilization 6 Mod or would one need a manual modification, as the OP seems to suggest?

ArchDuque-Pancake:
> Is this reproducible with the public+current version of Civilization 6 Mod or would one need a manual modification, as the OP seems to suggest?

The test district is in the public ver. of Civ 6 mod, but is currently blacked out. So you need to just remove the /* and load the game.

SomeTroglodyte:
> public ver

Does not show in mod manager??? [This one I assume](https://github.com/k4zoo/Civilization-6-Mod) - but it's not playable _at all_ due to too many modcheck failures (from newgame - loading above game unsurprisingly crashes in TechManager.updateEra). That can't be 4.0.6 and the original Civilization 6 Mod. Meaning I'm no longer interested, amigo :colombia:

ArchDuque-Pancake:
No, [This one](https://github.com/DW8030/Civ6-mod) bruh

SomeTroglodyte:
Then you should have said so - besides, the name differs from what the linked message says, so that's another hidden manipulation to cunfose oss

xlenstra:
[...]

PR #6680, opened 2022-05-02
Added uniques related to improvements

xlenstra:
More specifically, adds a unique for consuming the builder when finishing building something and for the 'take over adjacent tiles' ability of citadels.
Might have conflicts with whatever @SomeTroglodyte is doing with #6576. I'm fine with waiting for that PR to be merged and changing around it (should only be a renaming of a unique).

Part 1 of me going over the way improvements are build and fixing a few simple things (with the end goal of making it more intuitive and moddable, so that for example https://github.com/yairm210/Unciv/pull/4889#issuecomment-900612942 doesn't happen anymore. Additionally, I want to make it finally possible for multiple improvements to improve the same resource, allowing both oil wells and offshore platforms to improve oil). I'm planning on adding these other changes to this same PR, but this is technically fine to be merged as-is.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
@@ -250,7 +250,7 @@
 	},
 	{ 
 		"name": "City center", 
-		"uniques": ["Unpillagable", "Indestructible"],
+		"uniques": ["Unpillagable", "Irremovable"],
         "civilopediaText": [
 			{"text":"Marks the center of a city"},
 			{"text":"Appearance changes with the technological era of the owning civilization"}
--- a/android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
@@ -82,7 +82,7 @@
 		"terrainsCanBeBuiltOn": ["Coast"],
 		"food": 1,
 		"techRequired": "Sailing",
-		"uniques": ["[+1 Gold] <after discovering [Compass]>"]
+		"uniques": ["[+1 Gold] <after discovering [Compass]>", "Consumes builder when build"]
 	},
 
 	// Military improvement
@@ -174,31 +174,34 @@
 	{
 		"name": "Academy",
 		"science": 8,
-		"uniques": ["Great Improvement", "[+2 Science] <after discovering [Scientific Theory]>", "[+2 Science] <after discovering [Atomic Theory]>"]
+		"uniques": ["Great Improvement", "[+2 Science] <after discovering [Scientific Theory]>", "[+2 Science] <after discovering [Atomic Theory]>", "Consumes builder when build"]
 	},
 	{
 		"name": "Landmark",
 		"culture": 6,
-		"uniques": ["Great Improvement"]
+		"uniques": ["Great Improvement", "Consumes builder when build"]
 	},
 	{
 		"name": "Manufactory",
 		"production": 4,
-		"uniques": ["Great Improvement", "[+1 Production] <after discovering [Chemistry]>"]
+		"uniques": ["Great Improvement", "[+1 Production] <after discovering [Chemistry]>", "Consumes builder when build"]
 	},
 	{
 		"name": "Customs house",
 		"gold": 4,
-		"uniques": ["Great Improvement", "[+1 Gold] <after discovering [Economics]>"]
+		"uniques": ["Great Improvement", "[+1 Gold] <after discovering [Economics]>", "Consumes builder when build"]
 	},
 	{
 		"name": "Holy site",
 		"faith": 6,
-		"uniques": ["Great Improvement"]
+		"uniques": ["Great Improvement", "Consumes builder when build"]
 	},
 	{
 		"name": "Citadel",
-		"uniques": ["Great Improvement", "Gives a defensive bonus of [100]%", "Adjacent enemy units ending their turn take [30] damage", "Can be built just outside your borders"],
+		"uniques": [
+			"Great Improvement", "Gives a defensive bonus of [100]%", "Adjacent enemy units ending their turn take [30] damage", 
+			"Can be built just outside your borders", "Adds adjacent tiles to your closest city", "Consumes builder when build"
+		],
         "civilopediaText": [{"text":"Constructing it will take over the tiles around it and assign them to your closest city"}]
 	},
 
--- a/android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
@@ -82,7 +82,7 @@
 		"terrainsCanBeBuiltOn": ["Coast"],
 		"food": 1,
 		"techRequired": "Sailing",
-		"uniques": ["[+1 Gold] <after discovering [Compass]>", "Consumes builder when build"]
+		"uniques": ["[+1 Gold] <after discovering [Compass]>", "Consumes builder when built"]
 	},
 
 	// Military improvement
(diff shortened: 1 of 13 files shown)


SomeTroglodyte:
"Consumes builder when build": "built"
I chose to make the citadel unique text identical to the old civilopediaText, so xlators wouldn't need to re-translate - it _is_ clearer. And the civilopediaText is made obsolete by the unique. But I like your enum name better.
Indestructible deprecated becomes Indestructable?
Units that are _not_ great people but can place something and get consumed - without triggering Halicarnassus - should be possible. This reverts that option (UnitActions.683).

xlenstra:
Thanks for the comments!
> "Consumes builder when build": "built" 

Fixed, for some reason I always mess this conjuagtion up
> I chose to make the citadel unique text identical to the old civilopediaText, so xlators wouldn't need to re-translate - it _is_ clearer. 

I try to keep them relatively short, but I agree on the clarity - changed.
> And the civilopediaText is made obsolete by the unique. 

Good point, I've removed it
> Indestructible deprecated becomes Indestructable? 

I'm not sure what happened, I thought I had autocorrected those afterwards, but somehow their not in this PR. Fixed nonetheless.
> Units that are _not_ great people but can place something and get consumed - without triggering Halicarnassus - should be possible. This reverts that option (UnitActions.683).

I think they still can? On the new line `UnitActions:669` the function `improvement.handleImprovementCompletion(unit)` is called, which in turn calls `unit.consume()` which destroys the unit and, if it is a great person, provides stats?

SomeTroglodyte:
> for some reason I always mess this conjuagtion up

"conjuagtion" -> "conjugation" :rofl: :partying_face: :tada: :+1: :love_you_gesture: :heart_eyes_cat:  ... and isn't it "action" not "agtion"? I imagine _conjuaction_ could have some relation to the 5th pantheon Belief in our ruleset.

> unit.consume() which destroys the unit and, if it is a great person, provides stats?

From the diff display it looked as if your new consume() doesn't test if it's a GP, oh well

> try to keep them relatively short

I admit in this case I thought 'well if that already fits the display then it will just as well from the other source' - without actually looking at a game screen

SomeTroglodyte:
You know, stuff is piling up - don't we _have_ another third cerberus head with collaborator rights to fill up the three merge votes? You, me, who else is active these days and where can I see a list - @JackRainy I guess? Maybe I should put more minutes into review instead of playing.

xlenstra:
> > for some reason I always mess this conjuagtion up
> 
> "conjuagtion" -> "conjugation" rofl partying_face tada +1 love_you_gesture heart_eyes_cat ... and isn't it "action" not "agtion"? I imagine _conjuaction_ could have some relation to the 5th pantheon Belief in our ruleset.

I'll choose not to respond to that :joy:

> > unit.consume() which destroys the unit and, if it is a great person, provides stats?
> 
> From the diff display it looked as if your new consume() doesn't test if it's a GP, oh well

Yup, but that's intentional as that is already done inside the `addStatsPerGreatPersonUsage()` function itself

> > try to keep them relatively short
> 
> I admit in this case I thought 'well if that already fits the display then it will just as well from the other source' - without actually looking at a game screen

image
It's a bit long, but its probably fine

SomeTroglodyte:
`if (!isGreatPerson()) return` - I see.

SomeTroglodyte (approved):
UniqueDocsWriter result not up to date, but I don't care. The Indestructible deprecation  is a bit fast in 1 step to ERROR, but what the heck, you'll have reasons.

Tiny thing: My nose twitches whenever a pattern just remotely looks like 'dispose - continue using', and the consume case is such a thing. Now our MapUnits are just fine with that, but one _could_ fear some future coder has the same pattern expectations and goes and finds a reason to mess with data after 'destroy' because hey it's disposed anyway... So I'd probably cache out that home tile before destroying, without thinking twice. Nevermind.

`val turnsToBuild` - you know, fields accept Kdoc's and _do_ show them in tooltips (that comment might  be worth one - maybe). What I just learned: `[`references`]` in Kdoc participate in Studio's Alt-F7 "Find Usages" in both directions, and Kdoc style `/** */` _within_ code blocks - so they _don't_ appear as doc anywhere - still work for "Usages". I'm beginning to think that could be very _very_ useful...

SomeTroglodyte:
> [Updated docs I suppose](https://github.com/yairm210/Unciv/pull/6680/commits/be4eddfd9b2c5ea1f2e26963746f17f7f39260b9)

Come to think of it, for feature PR's I think that file should/can be excluded. I do so unless asleep. For any release the run to run TranslationFileWriter will update it anyway - right? I'm actually doubting having TranslationFileWriter in the GUI makes much sense anymore - command line parameter, call it from the auto workflows. And gate UniqueDocsWriter with the same approach, steals debug time.

JackRainy:
> You know, stuff is piling up - don't we _have_ another third cerberus head with collaborator rights to fill up the three merge votes? You, me, who else is active these days and where can I see a list - @JackRainy I guess? Maybe I should put more minutes into review instead of playing.

I was out of the project for a year or more, so I am trying to get in back by looking around what have changed (Diplomatic Victory, Rivers, the new approach with Uniques - are a few of the most obvious innovations). As soon as I more less understand what's going on, I will return for code reviews as well: it's hard doing a _good_ review, something more than static syntax check :) while have no clue what's going on around.

SomeTroglodyte:
> was out of the project

Thought so. Don't worry, you're clever, that clue is just around a corner.

yairm210:
Ah, nice! I see what you did here.
Previous conception was that there were 2 types of improvements - "instant build and consume" and "long build and no consume". What this does is separate the two - "instant build vs long build" is a property of *the builder*, and "consume vs no consume" is a property of *the improvement*.

What we gain in moddability is "instant build no consume" and "long build and consume", what we lose is the ability to have one improvement that can be built by one unit with consumption and by another without consumption. Seems like a fair trade-off.

yairm210:
I'm not sure if to merge this or not, given SomeTrog's parallel work. I approve, but I'll leave the decision to merge up to y'all.

yairm210 (approved):

xlenstra:
I'm fine with this being merged for now, the plans I had for fixing tile improvements/splitting oil well & offshore platform will go in a new PR anyway due how hard it got out of scope

SomeTroglodyte:
> lose the ability to have one improvement that can be built by one unit with consumption and by another without consumption

Uh, I'd be quite against that, if true. That would be an attractive in-game tradeoff for mods. "Look - build time is 20 turns and you can reuse the Builder, or you spend 500 shields and poof need to build another one"...

> SomeTrog's parallel work

Don't mind me, such can always be synced or reverted from a conflicting PR or... As long as we communicate ideas, the best one will win (idea that is).

SomeTroglodyte:
Build / construct - wording - again, not immediately clear to newcomers. "Can construct" (or build), "Can place immediately", and "Can place immediately consuming" maybe, and a count limitation could - syntactically - be expressed like a conditional??? Dunno.

I worked on my Aliens long ago, wanted them to have limited Workers (because they're lazy). So their specific Worker replacement needed 2x or more shields, had a high hurry modifier, and was only able to do a simple subset of what a Worker could do, and took 50% longer for that - making capturing true Workers a worthwhile adventure. In return, they could place/consuming that penguin hatchery, made to be not quite a bargain via their high price. I had that working just fine, though I don't remember whether the code patches for that were already PR'ed or are still in a certain sleeping local branch (which primarily deals with start locations, idea Aliens have their Spaceship wreck and it's always near their start - startbias = natural wonder).

I'll respect your opinion. If you think the choices available to modders will still be sufficient, go ahead. Code clarity and reliability could be a good tradeoff for limiting modder freedom.

xlenstra:
I agree with SomeTroglodyte that restricting modders in things that used to be possible isn't good, especially if I'm planning on completely reworking the feature the next release again. Thus I would remove the `consumes Builder when built` unique from improvements. As SomeTroglodyte also already made a different PR for the culture bomb unique of the citadel, having a PR only for that also wouldn't make any sense.
Removing those two features from this PR would remove basically everything, so I'm closing this PR. The changes in behaviour will be added to the next PR I make, which will be based of off this PR, but reverting those changes and applying those of SomeTroglodyte, assuming that is merged by the time I open that PR (it probably will be).
It will then also have the uniques for build/build immediately ([amount] times), so that modders will only get more options instead of different ones.

PR #6682, opened 2022-05-03
Create turn notifier for when the game is running for Windows

Azzurite:
If playing on Desktop, you often put the game into background, but still want to know if it's your turn. A standard Windows function for that is `FlashWindow` from winuser.h, which is implemented here.

Maybe someone else can add something for some of the window managers on linux or MacOS stuff or whatever, but that's not my domain.

### DIFF
--- a/core/src/com/unciv/ui/utils/GeneralPlatformSpecificHelpers.kt
+++ b/core/src/com/unciv/ui/utils/GeneralPlatformSpecificHelpers.kt
@@ -11,7 +11,12 @@ interface GeneralPlatformSpecificHelpers {
      *  @param allow `true`: allow all orientations (follows sensor as limited by OS settings)
      *          `false`: allow only landscape orientations (both if supported, otherwise default landscape only)
      */
-    fun allowPortrait(allow: Boolean)
+    fun allowPortrait(allow: Boolean) {}
 
     fun isInternetConnected(): Boolean
+
+    /**
+     * Notifies the user that it's their turn while the game is running
+     */
+    fun notifyTurnStarted() {}
 }
--- a/desktop/src/com/unciv/app/desktop/MultiplayerTurnNotifierWindows.kt
+++ b/desktop/src/com/unciv/app/desktop/MultiplayerTurnNotifierWindows.kt
@@ -1,0 +1,58 @@
+package com.unciv.app.desktop
+
+import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.backends.lwjgl3.Lwjgl3Graphics
+import com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window
+import com.badlogic.gdx.backends.lwjgl3.Lwjgl3WindowAdapter
+import com.sun.jna.Native
+import com.sun.jna.Pointer
+import com.sun.jna.platform.win32.User32
+import com.sun.jna.platform.win32.WinNT
+import com.sun.jna.platform.win32.WinUser
+import com.unciv.ui.MultiplayerTurnNotifier
+import org.lwjgl.glfw.GLFWNativeWin32
+
+class MultiplayerTurnNotifierWindows: Lwjgl3WindowAdapter(), MultiplayerTurnNotifier {
+    companion object {
+        val user32: User32? = try {
+            Native.load(User32::class.java)
+        } catch (e: UnsatisfiedLinkError) {
+            println("Error while initializing turn notifier: " + e.message)
+            null
+        }
+    }
+    private var window: Lwjgl3Window? = null
+    private var hasFocus: Boolean = true
+
+    override fun created(window: Lwjgl3Window?) {
+        this.window = window
+    }
+
+    override fun focusLost() {
+        hasFocus = false
+    }
+
+    override fun focusGained() {
+        hasFocus = true
+    }
+
+    /**
+     * See https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-flashwindowex
+     *
+     * We should've used FlashWindow instead of FlashWindowEx, but for some reason the former has no binding in Java's User32
+     */
+    override fun turnStarted() {
+        try {
+            if (user32 == null || hasFocus) return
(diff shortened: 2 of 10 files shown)


SomeTroglodyte:
A new helper... Couldn't this live in GeneralPlatformSpecificHelpers? And flashing windoes goes back to 16-bit 2.0 or so, few people know about it anymore methinks. I already thought we should have support for system Toasts aka desktop notifications. Tux has them, Winz has them, Droid obviously. Shouldn't be too hard.

SomeTroglodyte:
> Didn't know that existed

Fresh rename of orientations helper, to host a do-we-have-internet test. The others are historic - and their usecases/scope/complexity do justify separation a teeny weeny bit more.

> :D

I _do_ ignore when Visual Studio flashes its taskbar at me.. Mails arriving are so much  easier to see, even if it's highway robbery Outlook  (aka 365 :vomiting_face: )

> couldn't find

I said _should_ as in new feature

SomeTroglodyte:
Tux: Close to `Runtime.getRuntime().exec("notify-send --urgency=low --icon=\"ExtraImages/icon.png\" \"$message\"")` (but - jar packaged icon??)
W'oze: Maybe learn [from Nels](https://github.com/nels-o/toaster/blob/master/toast/Program.cs)...

SomeTroglodyte:
> should I move the behavior to `GeneralPlatformSpecificHelpers` for this to be merged

I will respect whatever architecture you prefer if you can present valid arguments for it. And - nothing is final except entropy.

<details><summary>But how about a signature like this, even if at first only partially implemented?</summary>

```kotlin
/** Interface to support various platform-specific tools */
interface GeneralPlatformSpecificHelpers {
    /** Pass a Boolean setting as used in [allowAndroidPortrait][GameSettings.allowAndroidPortrait] to the OS.
     *
     *  You can turn a mobile device on its side or upside down, and a mobile OS may or may not allow the
     *  position changes to automatically result in App orientation changes. This is about limiting that feature.
     *  @param allow `true`: allow all orientations (follows sensor as limited by OS settings)
     *          `false`: allow only landscape orientations (both if supported, otherwise default landscape only)
     */
    fun allowPortrait(allow: Boolean)

    /** Determine if connections to the Internet have high chances to succeed */
    fun isInternetConnected(): Boolean

    enum class SystemType { Unknown, Android, IOS, Linux, MacOS, Windows }
    /** Determine Platform Sub-Type (differs on Gdx 'desktop') */
    fun getSystemType(): SystemType

    enum class MessageUrgency { Low, Normal, Critical }
    /** Show a system message however the best integrated method may be for the platform -
     * e.g. Android Toast, Linux notify-send, Windows desktop notifications with blinking the taskbar icon when Critical...
     * A `null` [message] may for example do the taskbar blinking only */
    fun showSystemNotification(message: String, iconPath: String? = null, urgency: MessageUrgency = MessageUrgency.Low): Boolean
}
```
</details>

core/src/com/unciv/ui/utils/GeneralPlatformSpecificHelpers.kt
```diff
@@ -11,7 +11,12 @@ interface GeneralPlatformSpecificHelpers {
...
      */
-    fun allowPortrait(allow: Boolean)
+    fun allowPortrait(allow: Boolean) {}
 
     fun isInternetConnected(): Boolean
+
+    /**
+     * Notifies the user that it's their turn while the game is running
+     */
+    fun notifyTurnStarted() {}
```
SomeTroglodyte, line 21: That's allowed? I though default implementations necessitate abstract class instead of interface... And this should definitely stay a pure interface.

desktop/src/com/unciv/app/desktop/MultiplayerTurnNotifierWindows.kt
```diff
@@ -9,10 +9,9 @@ import com.sun.jna.Pointer
 import com.sun.jna.platform.win32.User32
 import com.sun.jna.platform.win32.WinNT
 import com.sun.jna.platform.win32.WinUser
-import com.unciv.ui.MultiplayerTurnNotifier
 import org.lwjgl.glfw.GLFWNativeWin32
 
-class MultiplayerTurnNotifierWindows: Lwjgl3WindowAdapter(), MultiplayerTurnNotifier {
+class MultiplayerTurnNotifierWindows: Lwjgl3WindowAdapter() {
```
SomeTroglodyte, line 14: I see. I had the vague idea the meat and bones of this was small enough to be folded into the helper - but yea, if it also needs to implement an API surface in another direction maybe better separate those.

desktop/src/com/unciv/app/desktop/PlatformSpecificHelpersDesktop.kt
```diff
@@ -15,4 +17,9 @@ class PlatformSpecificHelpersDesktop : GeneralPlatformSpecificHelpers {
             false
         }
     }
+
+    override fun notifyTurnStarted() {
+        turnNotifier.turnStarted()
```
SomeTroglodyte, line 22: I still expected something along `System.getProperty("os.name")?.startsWith("Windows")` - the `Native.load(User32::class.java)` _might_ `null` out cleanly on *ix - hopefully? The OS flavour distinction is in crash info, and some refactoring to have it in one place might be nice, but I have no obvious picture how. `object` level probably.

core/src/com/unciv/ui/utils/GeneralPlatformSpecificHelpers.kt
SomeTroglodyte: Ah now I remember. Of course. It's _fields_ that must stay abstract.

desktop/src/com/unciv/app/desktop/PlatformSpecificHelpersDesktop.kt
SomeTroglodyte: Imagine on Tux the "Preparing Wine Prefix..." dialog popping up :rofl:

SomeTroglodyte (approved):

PR #6687, opened 2022-05-03
Rework of the "Creates improvement on a specific tile" Unique

### LINKED ISSUE #6576: "Creates a [tileFilter] improvement on a specific tile" doesn't work

SomeTroglodyte:
Fixes #6576 

The original bug was a tiny mistake, passing the _improvement_ name instead of the _building_ name to addToQueue, but when I analyzed the basic design of how all that was meant to work, I found holes, and the scope gradually exploded. Sorry, all in one commit, and no screenies / clips for now. The checklist in that issue gives an idea.


SomeTroglodyte:
Since testing was good (see issue) and I can no longer remember some points my mind came up with (all code architecture), let's get this on the road.

Much too much code for such a borderline unique, but - the potential for ways to leave that turnstoimprovement = -1 in the map with no way to finish it needed addressing, and I could think of no more elegant way. Alternatives - extra TileInfo serializable field, or remove the UniqueType.

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -696,7 +696,10 @@ class CityInfo {
 
         // The relinquish ownership MUST come before removing the city,
         // because it updates the city stats which assumes there is a capital, so if you remove the capital it crashes
-        getTiles().forEach { expansion.relinquishOwnership(it) }
+        for (tile in getTiles()) {
+            tile.removeCreatesOneImprovementMarker()
+            expansion.relinquishOwnership(tile)
+        }
```
xlenstra, line 702: Would it make sense to put the `tile.removeCreatesOneImprovementMarker()` call in the `relinquishOwnership()` function?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -480,26 +502,42 @@ class CityConstructions {
...
-        if (!(getConstruction(constructionName) as INonPerpetualConstruction).postBuildEvent(this, stat))
+        val construction = getConstruction(constructionName) as? INonPerpetualConstruction ?: return false
+
+        // Support UniqueType.CreatesOneImprovement: it is active when getImprovementToCreate returns an improvement
+        val improvementToPlace = (construction as? Building)?.getImprovementToCreate(cityInfo.getRuleset())
+        if (improvementToPlace != null) {
+            // If active without a predetermined tile to place the improvement on, automate a tile
+            val finalTile = tile ?: Automation.getTileForConstructionImprovement(cityInfo, improvementToPlace)
+            if (finalTile == null) {
+                println("${cityInfo.name} tried to buy $constructionName without having any suitable tile for a ${improvementToPlace.name}.")  // todo remove when debugged
```
xlenstra, line 524: :upside_down_face:

xlenstra (approved):
I've looked through this code but haven't tested it. Doesn't look like anything obvious would go wrong, but it's too large to get the full picture in my head at once so I'm not 100% certain. I've assumed you've tested it so it's probably fine.

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: Ah yes a breakpoint anchor... Do we not have something like `asm { emit(0xCC) }` (if memory serves...) or `ILGenerator.Emit(OpCodes.Break);` ??

That breakpoint never hit.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -254,6 +281,15 @@ object Automation {
         }
     }
 
+    /** Support [UniqueType.CreatesOneImprovement] unique - find best tile for placement automation */
+    fun getTileForConstructionImprovement(cityInfo: CityInfo,  improvement: TileImprovement): TileInfo? {
+        return cityInfo.getTiles().filter {
+            it.canBuildImprovement(improvement, cityInfo.civInfo)
+        }.associateWith {
```
yairm210, line 288: No need to associate, just .maxBy{rankTileFirCityWork(it)}

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -915,11 +929,39 @@ open class TileInfo {
...
     }
 
+    /** Sets tile improvement to pillaged (without prior checks for validity)
+     *  and ensures that matching [UniqueType.CreatesOneImprovement] queued buildings are removed. */
+    fun setPillaged() {
+        // http://well-of-souls.com/civ/civ5_improvements.html says that naval improvements are destroyed upon pillage
+        //    and I can't find any other sources so I'll go with that
+        if (isLand) {
+            // Setting turnsToImprovement might interfere with UniqueType.CreatesOneImprovement
+            removeCreatesOneImprovementMarker()
```
yairm210, line 945: What happens in this case if the city completes construction?

core/src/com/unciv/ui/cityscreen/CityScreen.kt
```diff
@@ -160,20 +188,37 @@ class CityScreen(
     }
 
     private fun updateTileGroups() {
-        val nextTile = city.expansion.chooseNewTileToOwn()
+        fun isExistingImprovementValuable(tileInfo: TileInfo, improvement: TileImprovement): Boolean {
```
yairm210, line 191: Change "improvement" to "improvementPendingConsttruction" or something, the function and parameter names give the wrong idea

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -469,6 +456,41 @@ class CityConstructions {
                     builtBuildingUniqueMap.addUnique(unique)
     }
 
+    fun addFreeBuildings() {
+        // "Gain a free [buildingName] [cityFilter]"
+        val freeBuildingUniques = cityInfo.getLocalMatchingUniques(UniqueType.GainFreeBuildings, StateForConditionals(cityInfo.civInfo, cityInfo))
+
+        for (unique in freeBuildingUniques) {
+            val freeBuildingName = cityInfo.civInfo.getEquivalentBuilding(unique.params[0]).name
+            val citiesThatApply = when (unique.params[1]) {
```
yairm210, line 465: When { unique.isLocal, unique.isAntiLocal

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -469,6 +456,41 @@ class CityConstructions {
...
+                    freeBuildingsProvidedFromThisCity[city.id] = hashSetOf()
+
+                freeBuildingsProvidedFromThisCity[city.id]!!.add(freeBuildingName)
+            }
+        }
+
+        // Civ-level uniques - for these only add free buildings from each city to itself to avoid weirdness on city conquest
+        for (unique in cityInfo.civInfo.getMatchingUniques(UniqueType.GainFreeBuildings, stateForConditionals = StateForConditionals(cityInfo.civInfo, cityInfo))) {
+            val freeBuildingName = cityInfo.civInfo.getEquivalentBuilding(unique.params[0]).name
+            if (cityInfo.matchesFilter(unique.params[1])) {
```
yairm210, line 484: Reverse if with 'continue' for less indentation

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -469,6 +456,41 @@ class CityConstructions {
...
+                if (cityInfo.id !in freeBuildingsProvidedFromThisCity)
```
yairm210, line 485: These three lines are exactly the same as above, not sure if it's worth separating tbh

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -604,10 +640,61 @@ class CityConstructions {
...
+        val indexToRemove = constructionQueue.withIndex().mapNotNull {
+            val construction = getConstruction(it.value)
+            val buildingImprovement = (construction as? Building)?.getImprovementToCreate(ruleset)?.name
+            it.index.takeIf { buildingImprovement == improvement }
+        }.firstOrNull() ?: return
+
+        constructionQueue.removeAt(indexToRemove)
+
+        if (constructionQueue.isEmpty()) {
+            constructionQueue.add("Nothing")
```
yairm210, line 686: Const?

core/src/com/unciv/logic/city/CityInfo.kt
SomeTroglodyte: ~~I wanted to have the "what's that 'improvement marker'" knowledge in one place.~~ Oh you mean the call not the function. Okay, yes, that would also solve the case where a citadel drop steals the marked tile, which I'm almost certain the current state of the PR would miss.

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: Can't since the marker removal also removes the queued building. I didn't expand on that in the comment as the Kdoc says it and it's only twelve lines down to see how.

core/src/com/unciv/ui/cityscreen/CityScreen.kt
SomeTroglodyte: Correct (I allow myself some sloppiness in such a case (like naming a variable a, b, i, j, k, x,...) because it's a local sub-function and all uses (one) are just twelve lines away).

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: Not my code - I said elsewhere "either do a lint PR or concentrate on the matter at hand and keep your grubby fingers off unrelated lines even if it's tempting" - not that I really abide by that tenet - but since you're asking...

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: Not my code, and ... <sub><sup>(wtf is that isLocalEffect/isAntiLocalEffect anyway)</sub></sup> ... True, equivalent _and_ more efficient. OK, now I learned about those new (to me) helpers, I can include that.

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: :rage:  f*ing github just ate a detailed answer 'cuz I tried to avoid typing the filename and copy it. Lesson learned: _never_ compose text in the browser, use a notepad++ tab. Summary: Nope, I just moved the function as it violated the (in this file) still mostly correct regions (now correct). You know, a state-changing function in the pure region. Which I still didn't touch in MapUnit - a mess. I should learn that tenet I just postulated and keep my grubby fingers off such finds.

Any functional cleanup of that (`getLocalMatchingUniques` / `getMatchingUniques` / `cityInfo.matchesFilter` ignores "in this city" and "in other cities") mess needs **_testing_** which is not the scope of this PR. By someone who understands the architecture, which (seeing that) I'm not. Pity git blame is unable to track that relocation, I would need to switch branches to see whodunit.

So, do I revert the region cleanup or leave it - no, by now clearer to leave it in.

core/src/com/unciv/logic/city/CityConstructions.kt
SomeTroglodyte: Yeah, good one, though "not this PR's change" should trump.

Six uses, all the same meaning, so `Constants.kt` _or_ reuse one of them as `PerpetualConstruction.idle.name` the other five places. Not here please.

SomeTroglodyte:
I'm so sloppy - that last commit with the review changes has a completely unrelated name - how do I fix that? Where is the github-side rename for commits?

Note: I reverted all changes to `addFreeBuildings` it's still _only_ moved _verbatim_ in this PR.

Issue #5798, opened 2021-12-15
Musketman liberated a City State and got knocked into the sea


Mykro:
3.18.8 on Android 

My musketman captured a city state back from Mongolia. I chose to return the city to its original owner, so the borders went up and the musketman got dumped offshore and unable to move.  I even tried setting it to "Explore" and several turns later it hadn't gone anywhere.  Ended up having to disband the unit.

![Screenshot_20211215-082907-297](https://user-images.githubusercontent.com/1117944/146142424-d0afc5cf-0da3-4b07-87c2-0cffec8978eb.png)

xlenstra:
Can you provide a save file for this? Had you already researched Astronomy?
If not, it apppears your musketman is trapped there as it cannot ender the borders of Siam. Setting it to explore or moving it yourself then both don't work as it simply has no tiles to move to. The easy solution for this is trading for open borders with Siam, after which you can move your musketman as if nothing happened

yairm210:
I see, interesting
This could occur on land tiles as well which makes it extra tricky, and in other situations

I think that the teleport requires a general rethink, and since it's relatively rare we can afford to take more time to reach a decision
 - get all cities in contiguous enemy border
 - find closest city of ours with getClosestCities
 - try dropping unit near closest city, with priority to land tiles if land unit

will-ca:
Flood fill from each city into a Set until all cities in the set, and then pick the nearest tile in the set?

Cache for the turn, since usually there's more than one unit to teleport.

heipizhu4:
I wonder what happens in Civ V. Can anyone give it a test?[My PC can't run CivV. It's too big:( ]

0avsz:
![Imagepipe_0](https://user-images.githubusercontent.com/103357381/162610239-745c6d2c-7765-4b6a-98ad-4e9984e70121.jpg)
![Imagepipe_1](https://user-images.githubusercontent.com/103357381/162610245-b77517f7-fea2-4ea9-bb90-7f3f03adba92.jpg)

Military units stack on top of enemy civilian units after liberating or trading a city without killing them.

yairm210:
Should be solved by #6698

PR #6698, opened 2022-05-04
Reviewed version of the unit's teleportation

### LINKED ISSUE #4923: When the open borders agreement stopped, my submarine was moved from the sea into the lake not connected to the sea
**Platform**
Android 

**Version**
3.16.3

**Describe the bug**
Submarine was in neighbours aquatory when open borders treaty was on. After open borders is finished, submarine was automatically moved out of the borders, which is expected. But in result it was placed to the lake which is not connected to the sea

Screenshot_2021-08-20-12-37-07-129_com unciv app
Screenshot_2021-08-20-12-47-08-897_com unciv app

```

```
</details>

**Expected behavior**

Submarine is placed to sea instead of lake

JackRainy:
Fixes #4923 and resolves issue mentioned in https://github.com/yairm210/Unciv/issues/5798#issuecomment-1094220756 

Here we solve 2 problems:
1) When the military unit was teleported onto the enemy civilian, the latest was not captured.
Now it is.
2) When the unit was teleported it might appear in the place of no return, where it could not reach technically, e.g. in trap between mountains or a ship into the lake, etc.
Now the path to the destination point is something real which could be made by a unit being "escorted" from the expelled territory.

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -868,8 +868,12 @@ class MapUnit {
         }
         if (tile.improvement == Constants.barbarianEncampment && !civInfo.isBarbarian())
             clearEncampment(tile)
+        // Check whether any civilians without military units are there.
+        // Keep in mind that putInTile(), which calls this method,
+        // might has already placed your military unit in this tile.
+        val unguardedCivilian = tile.getUnguardedCivilian(this)
         // Capture Enemy Civilian Unit if you move on top of it
-        if (isMilitary() && tile.getUnguardedCivilian() != null && civInfo.isAtWarWith(tile.getUnguardedCivilian()!!.civInfo)) {
+        if (isMilitary() && unguardedCivilian != null && civInfo.isAtWarWith(unguardedCivilian.civInfo)) {
             Battle.captureCivilianUnit(MapUnitCombatant(this), MapUnitCombatant(tile.civilianUnit!!))
         }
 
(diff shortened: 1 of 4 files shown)


SomeTroglodyte:
What if the unit could not leave its current tile in any way by normal means? With "path to the destination point is something real" such a unit could not be displaced at all - will it stay, die, change owners, crash the game or teleport the old way?

JackRainy:
If it cannot leave the tile "by meaningful way" there is a fallback (same as it was before) - try to teleport it to one of your cities.
Furthermore, if you have no cities (can it happen?) or the territory of your cities is already occupied with other units (quite a rare case) - it will just gloriously die being tortured in captivity. 😢

SomeTroglodyte:
> tortured in captivity

Do we have a "resistant to torture" unique?

SomeTroglodyte (approved):
Maximizing English.words.invent("naggyness"):
- "might have" not "Might has" :rofl: 
- My nose twitches at `getUnguardedCivilian` and city tiles If that function can be called at all for city tiles - should a city itself not count as protection? And why test airunits? Isn't a carrier always military? - I'm too lazy to follow that thought.
- `getPathBetweenTiles` - didn't the BFS have a safety limit? Maybe this could do with a max of say 91 (getNumberOfTilesInHexagon(5)) or so?

JackRainy:
> "might have" not "Might has" 🤣

Damn... I need more practice in writing code comments to improve my English :)

>  `getUnguardedCivilian` and city tiles

`canMoveTo` is called before that and contains such check.

> didn't the BFS have a safety limit? 

It has `maxSize` property which may be used as a limit. Probably, it makes sense to use it since in the worst rare case it will go through the whole map and stop there. Not sure how it affects performance, though, let's see.

PR #6712, opened 2022-05-06
Improved clarity and moddability of building improvements

xlenstra:
This PR changes a lot of things on how improvements are placed, among them the following:
- Renamed `tile.hasUnique` to `tile.terrainhasUnique` as that is the more descriptive name. To also include the uniques of the resource and improvement, use the new `tile.hasUnique`.
- Made the citadel culture bomb ability a unique that can be applied to any improvement.
- Added "Land" and "Water" as options to `TileImprovement.terrainCanBeBuiltOn`, which act as general filters for any land and water tile
- Rewritten the entire `tile.canImprovementBeBuildHere` function for clarity, removing the "exquisite spaghetti" described in https://github.com/yairm210/Unciv/pull/4889#issuecomment-900612942
- Made it so resources can have multiple improvements, and used this to add offshore platforms in the correct way

This PR has spiraled out of control with every feature I've added, so I've split it into a few commits, each of which compiles and runs on its own, to make it slightly more readable. The first of these comments is what remains of #6680.

For this, I had to rewrite multiple tests, which I _think_ are still correct, but a second pair of eyes over them couldn't hurt. Additionaly, I'm 80% sure something will break when loading an old save in this game (at least water oil wells being removed, probably worse). I'll look into compatability upgrades later this weekend.
Additionally, mods might have to change a few things to get tile improvements working again. I'm ~20% certain ancient ruins (and city centers & barbarian encampments) might have become buildable unless specifically marked `Unbuildable`, and ~60% certain great improvements cannot be built anymore.
Regardless, in my opinion the code in `TileInfo.canImprovementBeBuildHere` is so much cleaner that this offsets the temporary problems modders might have.


SomeTroglodyte:
"Improvid" is a pun [on...](https://en.wikipedia.org/wiki/COVID-19)?  🦠 <sup><sub><- this is nice, renders completely different in editor and preview...</sup></sub>
"spaghetti" :rofl: - boy am I happy that it wasn't me being _that_ blunt :hammer: 
"~20% certain" yup, that cinches it - **properly reviewing** is ~two day's work :eyes:

xlenstra:
> "Improvid" is a pun [on...](https://en.wikipedia.org/wiki/COVID-19)?  🦠 <sup><sub><- this is nice, renders completely different in editor and preview...</sup></sub>

Nope, that's just me not being able to type properly 😅. Fixed!

SomeTroglodyte:
> Fixed

Awwwwww 😭😭😭 ...no more improvised beer...

yairm210:
I'm willing to overlook problems that *modders* may encounter, but breaking old games is bad, so until we know they arenlt broken this should be a draft

xlenstra:
> Additionaly, I'm 80% sure something will break when loading an old save in this game (at least water oil wells being removed, probably worse). I'll look into compatability upgrades later this weekend. 

It turns out that improvements aren't normalized when loading a game, so nothing bad actually happened. Oil wells remain on water oil tiles and provide that oil, but can be replaced with offshore platforms by work boats, which sounds fine to me.

> Additionally, mods might have to change a few things to get tile improvements working again. I'm ~20% certain ancient ruins (and city centers & barbarian encampments) might have become buildable unless specifically marked `Unbuildable`,

This turned out not to be the case
> and ~60% certain great improvements cannot be built anymore. Regardless, in my opinion the code in `TileInfo.canImprovementBeBuildHere` is so much cleaner that this offsets the temporary problems modders might have.

This turned out to be the case, but can easily be fixed by modders adding the line `canBeBuiltOn: ["Land"]` to every great improvement. When this PR is merged, I'll warn about this in the discord \#game-mod-discussion channel.

To me this sounds like this is mergeable, so I'll unmark this PR as draft.

yairm210:
In order to not break all existing mods (including roads?) and let modders have a transition period, I suggest the following: make all improvements with no terrainsCanBe... and no "built on resource" unique automatically act like they do have the "land" terrain.
Even if we add this unique automatically when parsing the ruleset.
This is a good change, bit we can easily avoid breaking everyone's games, so why not

SomeTroglodyte:
> BasicTests > baseRulesetHasNoBugs FAILED

Makes you wonder why in that place a `lines +=` can't leave a severity right a way...

xlenstra:
Problem is very likely that base rulesets can also have improvements marked 'unbuildable', which shouldn't have a terrainsCanBeBuiltOn or the resource unique. I'll fix this later today when I'm home again

SomeTroglodyte:
I want to approve. I almost did. I believe it's good.
**_But:_** wouldn't the "W" key for Offshore Platforms be more intuitive?

No, seriously:
Needs merge (This cannot load pre-#6720+#6724 games). Merge challenges mainly come from trailing blanks again. I can get it to run, but some merge choices I did sloppily, wasn't sure immediately. So - you do it. My merge plays the Civ6 mod just fine, though.
Also - did we **_really_** not have any vision how to allow mods where one improvement can be time-built _or_ placed, on land or water? 'Consumes' on the improvement still pretty much precludes such attempts, right?

xlenstra:
> Also - did we really not have any vision how to allow mods where one improvement can be time-built or placed, on land or water? 'Consumes' on the improvement still pretty much precludes such attempts, right?

I'm planning on adding the uniques mentioned before to units, by mixing & matching does I believe all configurations should be possible. But different PR, as this is large enough as it is

yairm210 (approved):

PR #6715, opened 2022-05-06
vpn/proxy issues when creating new mp game

### LINKED ISSUE #6709: Can't create new Multiplayer Games from Desktop
**Platform**
Raspberry Pi OS armhf (32-bit linux arm)

**Version**
4.0.12, 4.0.13

**Describe the bug**
After #6679 i cannot create new games from Desktop in the first place!!!
The Client shows no network error regardless of my internet being turned on or off. It cannot detect network in my Raspberry Pi. Not sure if other Desktops are affected too or not.

**To Reproduce**
Try to create any Online Multiplayer Game from a custom server in Desktop client.

alexban011:
resolves #6709 by opening a connection to dropbox.com

Tested on windows and linux (Zorin OS). Seems to be working now.
I removed the redundant ping and replaced it with opening a connection to dropbox

I've also fixed the unnecessary commits and will close the other PR

### DIFF
--- a/desktop/src/com/unciv/app/desktop/PlatformSpecificHelpersDesktop.kt
+++ b/desktop/src/com/unciv/app/desktop/PlatformSpecificHelpersDesktop.kt
@@ -2,7 +2,7 @@ package com.unciv.app.desktop
 
 import com.badlogic.gdx.backends.lwjgl3.Lwjgl3ApplicationConfiguration
 import com.unciv.ui.utils.GeneralPlatformSpecificHelpers
-import java.net.InetAddress
+import java.net.URL
 
 class PlatformSpecificHelpersDesktop(config: Lwjgl3ApplicationConfiguration) : GeneralPlatformSpecificHelpers {
     val turnNotifier = MultiplayerTurnNotifierDesktop()
@@ -12,7 +12,10 @@ class PlatformSpecificHelpersDesktop(config: Lwjgl3ApplicationConfiguration) : G
 
     override fun isInternetConnected(): Boolean {
         return try {
-            InetAddress.getByName("8.8.8.8").isReachable(500)  // Parameter timeout in milliseconds
+            val u = URL("http://www.dropbox.com")
+            val conn = u.openConnection()
+            conn.connect()
+            true
         } catch (ex: Throwable) {
             false
         }


touhidurrr:
firstly, I think I suggested that you use `https://content.dropboxapi.com` for connection checking instead of `http://www.dropbox.com`. The api and their main website is entirely two different thing altogether.
Also i suggested that you use `com.unciv.UncivGame.Current.settings.multiplayerServer` when available.

touhidurrr:
I suggest this changes!!!
https://github.com/alexban011/Unciv/pull/2/files

alexban011:
> I suggested that you use https://content.dropboxapi.com
I'll change it now but

> com.unciv.UncivGame.Current.settings.multiplayerServer
not sure how to check for that

touhidurrr:
> not sure how to check for that

Did you check my pr in your fork?

alexban011:
> Did you check my pr in your fork?

I did and merged it, thanks !

Mape6:
Looks good from my I-don't-know-anything-about-Kotlin point of view.

touhidurrr:
> Looks good from my I-don't-know-anything-about-Kotlin point of view.

Lol. I used https://play.kotlinlang.org to test some assumed lines of kotlin code and checked if those worked before doing these commits.

I know I used to say that. I will change 'anything' to 'much' from now on maybe. 🤣

alexban011:
> > Looks good from my I-don't-know-anything-about-Kotlin point of view.
> 
> Lol. I used https://play.kotlinlang.org & https://www.google.com to test some assumed lines of kotlin code and checked if those worked before doing these commits.
> 
> I know I used to say that. I will change 'anything' to 'much' from now on maybe. 🤣
> 
> I saw a YouTube video on Kotlin basics months ago. Maybe I should add that too.

yeah I never wrote Kotlin before committing to this repo, I just assumed its somewhat similar to java and so far it seems to be. (more like python variables combined with java but yeah) :))

SomeTroglodyte:
> somewhat similar

Kotlin is an Island and a language. Java is an Island and a language. Intentional. The Kotlin compiler emits Java VM bytecode - unless a different backend is used.

touhidurrr:
This new code looks a lot of sus to me. How does it work?
Also it it freezes for a few seconds then why not use promise and loop through promise.all output?

touhidurrr:
Also if internet is not working shouldn’t the previous changes be enough to check it?
Do we really need to check internet connectivity further like this?
Feels like a waste of time to me.

Or is the previous version is not enough to detect proxy?

alexban011:
> 

well its the code that runs on desktop, but I'm not sure why it freezes.
Removing the android SDK stuff also works the same way, so the freezing seems to be coming from connect()
My guess is that there is a timeout for that function and the freeze comes from waiting for that to finish

> Or is the previous version is not enough to detect proxy?
No, the previous version was not working when connected to proxy

alexban011:
So I could remove the SDK code. That would make the function identical to the desktop one so It could be moved back to NewGameScreen to avoid duplicate code? Or leave it in PlatformSpecificHelpers?

I kept it there in case that fails to continue to the SDK code, but I probably should have added another try-catch inside that one. Might be redundant though.

touhidurrr:
> Keep the android SDK code or get rid of it

Probably something that runs on both is better. Android Specific codes are a problem.

Also if the previous changes then Unciv shouldn’t work on proxy in the first place. URL comes directly from Java so i don't get why it wouldn’t work.

touhidurrr:
Maybe check Dropbox codes and use the network utilities used there? They should work regardless of platform a they are already working properly currently right?

android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
```diff
@@ -34,13 +36,23 @@ Sources for Info about current orientation in case need:
     }
 
     override fun isInternetConnected(): Boolean {
```
yairm210, line 38: If you're changing this to not only check internet connectivity, there are actually 2 fail states and so A. the function name should be changed (checkConnectionToMultiplayerServer or something) and B. we should tell the user if it's a problem with his internet or the server itself, so either return a fail string or an enum or something.

This also means we can keep the already-translated line of "No internet connection!" and just add another translation line of "The server is unreachable" or something

android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
```diff
@@ -34,13 +36,23 @@ Sources for Info about current orientation in case need:
...
+
+            val connectivityManager = activity.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+            for (network in connectivityManager.allNetworks) {
+                val networkCapabilities = connectivityManager.getNetworkCapabilities(network) ?: continue
+                val isInternet = networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+                val info = connectivityManager.getNetworkInfo(network) ?: continue
+                if (isInternet && info.isConnected) return true
+            }
+
+            true
```
yairm210, line 53: And obviously this should be *first*.
First check if we even have internet, if no - failure A. 
Then check if we can reach server - if no, failure B.

alexban011: I tried with that first but the code fails (the freeze is back)

android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
alexban011: yeah I'm gonna split the 2 in 2 separate functions

alexban011:
ignore my last commit :)

SomeTroglodyte:
Sorry, in its current state I'd call this a no-go, wrong design.
- Java URL.openConnection is not exactly platform-dependent. Not as cross-platform as Gdx APIs maybe, but still. So - doesn't really belong in platform-specific code if there are no other compelling reasons.
- Querying the Connectivity manager do decide whether or not to attempt an actual internet connection is good practice. It's meant to determine stuff like whether config allows, and interfaces are up, and should typically do its job without any packet on the line (and if it sends stuff that's the OS guy's decision and none of our business). So before, not after, and it _is_ the correct thing to do. The equivalent on Linux would be to query the Network manager, maybe even by calling nmcli and parsing the result; on Windows one could query whether a default route exits and its associated interface it up; or the API offers something equivalent (which probably would do similar things under the hood.)
- If you're sure the usecase justifies putting a packet on the line without prior test whether it even  has a chance to reach its destination, go ahead, you don't need the whole PlatformSpecificHelpers stuff. Unless existing Java or Gdx apis are lacking a feature.
- Uh, and a tiny thing: `?: "Dropbox"; if (multiplayerServer != "Dropbox")` is silly. `val url = URL(UncivGame.Current.settings.multiplayerServer ?: "https://content.dropboxapi.com")` would do. And the static dripbox address there could be clearer as a static const val somewhere appropriate.

Ah, parallel comments & commits, OK pick what's still relevant.

alexban011:
I moved the functions in NewGameScreen, where they are being called. Was there a better place to put them?

touhidurrr:
> So - doesn't really belong in platform-specific code if there are no other compelling reasons.

That was something I was wondering too. Why the code was being classified as Platform specific helpers if we were using `java.net.URL` anyways. Then alex used some Android specific code and looked like it made some sense (?)

> Querying the Connectivity manager do decide whether or not to attempt an actual internet connection is good practice.

I think it's unnecessary whether or not it's a good practice or not. If we can figure out why the connectivity test that is implemented now is not working behind a proxy it would enough to just ping `multiplayerServer` or `https://content.dropboxapi.com` to check it they are working. Stuff like determining which connection to use is for OS to decide.

> silly

That was because I was not sure what the default value of `multiplayerServer` is. I thought it might be `null` or `Dropbox` since when unset it is shown as `Dropbox` in the UI. Thus the extra caution. Also new to Kotlin.

alexban011:
> Then alex used some Android specific code and looked like it made some sense (?)

actually the URL code is new. The Android SDK code was there from the original PR so back than it did make sense

touhidurrr:
Anyways at this point this won't be done before the next version is rolled out.
This issue looks very messy. We are adding and removing codes like nothing in the same pr which is kinda weird.
Also your last commit makes no sense to me.
It happened like this:
1. First we decided that we would ping multiplayerServer or Dropbox endpoint
2. Then you began checking for all available network connections and see if any of them has network connection because you thought that 1 is not enough
3. Then @yairm210 told us that we should do separate checks for network and server and show proper erros
4. @SomeTroglodyte praised the usage of network scanning also that platformSpecificHelpers made no sense
5. And now we are suddenly pinging both multiplayerServer or Dropbox (?), ignoring the initial proxy concerns on point 2 (??), kind of messing up @yairm210's suggestion at point 3 (???), acknowledging @SomeTroglodyte's concern about platformSpecificHelpers but ditching network scanning anyways (????) and implementing some Hybrid solution that we don't even know if it would solve the issues or not (assuming from your chats where you said that this doesn't work on proxy / vpn) (????).

So, maybe we need a fresh start?

Also why `checkConnectionToInternet()` is pinging `Dropbox` while `checkConnectionToMultiplayerServer()` is pinging `multiplayerServer`? Practically speaking does they server any different purpose? While checking internet connectivity isn't both the same? so do we even need a separate `checkConnectionToInternet()` in the first place pinging `Dropbox` when we were already had `isInternetConnected()` which did both? This kind of feels redundant to me.

touhidurrr:
maybe something like this would do
```kt
import java.net.URL
import com.unciv.UncivGame.Current.settings.multiplayerServer

private fun connectivityChecker(): Boolean {
    try {
        val u =  URL(multiplayerServer ?: "https://content.dropboxapi.com")
        u.openConnection()

        true
    } catch {
        val noInternetConnectionPopup = Popup(this)
        val label = if (multiplayerServer) "Couldn't connect to the Multiplayer Server!" else "Couldn't connect to Dropbox!"
        noInternetConnectionPopup.addGoodSizedLabel(label.tr()).row()
        noInternetConnectionPopup.addCloseButton()
        noInternetConnectionPopup.open()
        return@onClick

        false
    }
}
```

touhidurrr:
Bur then again i don't know if it would solve anything. 🤣

alexban011:
> 2 

I didn't began checking for that. Like I said. The android specific code is already in the release so 1 happened after 2

> 3 and 4

thats what I did in the last commit. Separated the functions, got them out of platformSpecificHelpers and displayed the right message

> 5 

Not sure if I understand this. The new code doesn't ping both of those, just the right one according to the settings.

> but ditching network scanning anyways

if you're talking about the android SDK code again, thats because that code does not work when under proxy. The new code (how the PR looks now, at [5338e5c] ) does work from my tests: on android and windows with and without proxy

@SomeTroglodyte 's suggestion probably works very well but java.net.URL is platform independent (from my knowledge) so this should also work. 

> Also why checkConnectionToInternet() is pinging Dropbox while checkConnectionToMultiplayerServer() is pinging multiplayerServer? Practically speaking does they server any different purpose? While checking internet connectivity isn't both the same? so do we even need a separate checkConnectionToInternet() in the first place pinging Dropbox when we were already had isInternetConnected() which did both? This kind of feels redundant to me.

I probably should have kept them in 1 function, but I thought it might be better to keep them separate for some reason.

> val u =  URL(multiplayerServer ?: "https://content.dropboxapi.com")

My IDE says it always returns the left operand, so I don't think the code would ever check the connection to dropbox, unless the default value for multiplayerServer is https://content.dropboxapi.com.

alexban011:
>if (multiplayerServer) 

multiplayerServer cannot be null so it would have to be something like multiplayerServer != "Dropbox" (or whatever the default value is)

alexban011:
`    private fun connectivityChecker(): Boolean {
        return try {
            val u: URL = if (UncivGame.Current.settings.multiplayerServer == "Dropbox") {
                            URL("https://content.dropboxapi.com")
                        } else {
                            URL(UncivGame.Current.settings.multiplayerServer)
                        }
            u.openConnection()

            true
        } catch (ex: Throwable) {
            val noInternetConnectionPopup = Popup(this)
            val label = if (UncivGame.Current.settings.multiplayerServer != "Dropbox") "Couldn't connect to the Multiplayer Server!" else "Couldn't connect to Dropbox!"
            noInternetConnectionPopup.addGoodSizedLabel(label.tr()).row()
            noInternetConnectionPopup.addCloseButton()
            noInternetConnectionPopup.open()

            false
        }
    }`

I edited it to this but it doesn't work at all..

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
```diff
@@ -206,6 +219,28 @@ class NewGameScreen(
...
+        } catch (ex: Throwable) {
+            false
+        }
+    }
+
+    private fun checkConnectionToMultiplayerServer(): Boolean {
+        return try {
+            val u = URL(UncivGame.Current.settings.multiplayerServer)
+            val conn = u.openConnection()
+            conn.connect()
```
Mape6, line 237: What does the connect do exactly? Does it try to open the TLS sessions without any HTTP request?
If it does a HTTP request, please consider to request https://serverurl/isalive
Like in [Unciv_server.kt](https://github.com/yairm210/Unciv/blob/e1363d92f9d7bde846218123da90f14e2d66a311/desktop/src/com/unciv/app/desktop/UncivServer.kt ) line 45.

SomeTroglodyte:
> Not sure if we can create popups anywhere

In platform-specific code: NO - UI doesn't belong there, except for maybe unavoidable debugging console logs.
In Gdx app code: Only on the main thread. Without reading context, a name like `connectivityChecker` sounds more like a non-UI helper -> move UI to caller. Or, document (in `/** Kdoc */`) the absolute requirement that the function is only called from the main thread. Or do UI under postCrashHandlingRunnable - and still doc that can happen (and be aware your code will continue to run _before_ the UI stuff happens). Now since the function is private, some of this may perhaps be simplified a bit if readability and thread-safety is preserved.

alexban011:
> In platform-specific code: NO - UI doesn't belong there

so I guess its good where it is now (next to the invalid ID code)

yairm210 (approved):

touhidurrr (changes requested):
simplify and redo it!!

touhidurrr:
> but it doesn't really matter if it can't check for dropbox)

@alexban011 I think I have found the correct way to do this. See this:
https://github.com/yairm210/Unciv/blob/7b4833741d1e424e7c68d0c9192f8a477dbd3dc5/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt#L273

So, I think something like this would do:
```kt
import java.net.URL
import com.unciv.Constants
import com.unciv.UncivGame

private fun connectivityChecker(): Boolean {
    try {
        val multiplayerServer = UncivGame.Current.settings.multiplayerServer
        val isDropbox = multiplayerServer == Constants.dropboxMultiplayerServer
        val u =  URL(if (isDropbox) "https://content.dropboxapi.com" else multiplayerServer)
        u.openConnection().setTimeout(5000).connect()

        true
    } catch {
        val noInternetConnectionPopup = Popup(this)
        val label = if (isDropbox) "Couldn't connect to Dropbox!" else "Couldn't connect to Multiplayer Server!"
        noInternetConnectionPopup.addGoodSizedLabel(label.tr()).row()
        noInternetConnectionPopup.addCloseButton()
        noInternetConnectionPopup.open()
        return@onClick

        false
    }
}
```

alexban011:
Just tried it on android and it seems to work
I did had to make a slight modification: I moved the popup in the main function in an if(!isConnectionToServerPossible()) cause it was putting an error at @onClick 

did you tested it on the Pi?

touhidurrr:
> did you tested it on the Pi?

I can compile the source for reasons. Give me a jar to test.

alexban011:
> I can't compile the source for reasons. Give me a jar to test.

I'm trying. I'm getting this:
Kotlin could not find the required JDK tools in the Java installation. Make sure Kotlin compilation is running on a JDK, not JRE.

I'll try reinstalling the JDK as people on forums said this should solve this

alexban011:
Sorry for mediafire link, I have no clue how else to send it:
[Download link](https://www.mediafire.com/file/xhdbuckdez15xj7/Unciv.jar/file)

SomeTroglodyte:
> putting an error at `@OnClick`

You're lucky (or maybe unlucky) there seems to be no actual user with that name around, or you would have spammed them. And the jar location is in the wiki not far under the <a href="https://yairm210.github.io/Unciv/Developers/Building-Locally/#linuxmac-os"><code>gradlew syntax</code></a>.

SomeTroglodyte:
>mediafire

otnolatrnup.com, doubleclick.net and aaxads.com are all quite malicious. do not visit without a good blocker.

alexban011:
> You're lucky (or maybe unlucky) there seems to be no actual user with that name around, or you would have spammed them. 

I made sure I'm not pinging anyone when I sent that :))

alexban011:
So now it works on Windows and Android even with proxy (both Mobile Hotspot and 1.1.1.1 vpn). 

Now the "Couldn't connect to Dropbox!" makes more sense than the old "No internet connection!", but should I change it back to the old one to keep the translation or keep the new one as its more informative of what the issue is?

alexban011:
Last commit is just cleanup and I replaced the setConnectionTimeout because it was giving a warning. 

Also I set the timeout to 3000 instead of 5000 since if the server doesn't respond in 3 seconds I really doubt it will in 5. 
The freeze still happens on my phone but only if it was connected to the vpn before losing connection, so I don't think we should worry about it. Weirdly enough it seems to only freeze once and then the next time I press "start game" it displays the message instantly.

touhidurrr:
Never pull updates in a branch!!!

desktop/src/com/unciv/app/desktop/PlatformSpecificHelpersDesktop.kt
```diff
@@ -2,20 +2,11 @@ package com.unciv.app.desktop
...
 
 class PlatformSpecificHelpersDesktop(config: Lwjgl3ApplicationConfiguration) : GeneralPlatformSpecificHelpers {
     val turnNotifier = MultiplayerTurnNotifierDesktop()
     init {
-        config.setWindowListener(turnNotifier);
-    }
-
-    override fun isInternetConnected(): Boolean {
-        return try {
-            InetAddress.getByName("8.8.8.8").isReachable(500)  // Parameter timeout in milliseconds
```
touhidurrr, line 15: Sigh could it be that this is the reason why my checks always failed on Pi?
Proxy increases latency by a good margin and 500 is really not enough.
It is not enough even without proxy. And will fail 5-10% of times.
Who even sets timeout to only 0.5 seconds, I am astonished that nobody pointed this out.

alexban011:
> Never pull updates in a branch!!! But only in master / main.

Noted.

touhidurrr:
@alexban011 use this https://www.file.io/

alexban011:
> @alexban011 use this https://www.file.io/

https://file.io/uQ3O9ABTwnyx

touhidurrr:
@alexban011 I tested your jar and it works fine on raspberry pi

SomeTroglodyte:
> Never pull updates in a branch

Disagree. There are some very valid reasons to do so. Own work in two+ studio installations, or testing someone's branch and they pushed, or mixing features to see if they work together.......

touhidurrr:
[Merge branch 'yairm210:master' into master-proxyIssues](https://github.com/yairm210/Unciv/pull/6715/commits/ef4e4a552b21ac8ae0d5cfd2c632e62239a29910)

I said that cause I don't like this kind of commits.

SomeTroglodyte:
> I said that cause I don't like

Then a wording like "Never" gives a misleading impression, the sentence was an imperative. With three Char(0x21) to boot. Simply "I never..." would have hit the mark, right? Online, culture barriers and so on, be careful..

PR #6718, opened 2022-05-07
Assist on downloading of the missing mods

JackRainy:
Let's imagine the story: 
You are eager to fix yet another pesky bug and (what a luck!) you even got the saved file to reproduce it.
You try to load it and... 💥  

| Have you ever seen this screen? | ...or if you lucky enough, you know the reason, at least: |
| - | - |
| Screenshot 2022-05-07 19-34-02 |  Screenshot 2022-05-07 19-34-28 |

Previously, you had to go to the mod options, seaching for the given mods (if you are lucky to remember all their names), download them and return back to the load screen to try again. Too looooong and complicated.

Now you are 2 clicks away from loading the game since the magic button "Download missing mods" has appeared.

Just click on it and the missing mods will be downloaded for you to try loading the game once again.
P.S. Of course, it appears only if the loading error is caused by missing mods.

### DIFF
--- a/core/src/com/unciv/ui/saves/LoadGameScreen.kt
+++ b/core/src/com/unciv/ui/saves/LoadGameScreen.kt
@@ -12,12 +12,15 @@ import com.unciv.UncivGame
 import com.unciv.logic.GameSaver
 import com.unciv.logic.MissingModsException
 import com.unciv.logic.UncivShowableException
+import com.unciv.models.ruleset.RulesetCache
 import com.unciv.models.translations.tr
 import com.unciv.ui.crashhandling.crashHandlingThread
 import com.unciv.ui.crashhandling.postCrashHandlingRunnable
 import com.unciv.ui.images.ImageGetter
+import com.unciv.ui.pickerscreens.Github
 import com.unciv.ui.pickerscreens.PickerScreen
 import com.unciv.ui.popup.Popup
+import com.unciv.ui.popup.ToastPopup
 import com.unciv.ui.utils.*
 import com.unciv.ui.utils.UncivDateFormat.formatDate
 import java.util.*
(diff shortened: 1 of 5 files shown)


yairm210:
🤩 
This will be a HUGE timesaver!!!

yairm210 (approved):
Usually I would tell you that the "missing mods translated successfully" and "Could not load the missing mods!" should be translated, or you could replace them with already-translatable strings, but in this specific case - it's a feature meant for devs and mod makers so it'll be fine.

core/src/com/unciv/ui/saves/LoadGameScreen.kt
```diff
@@ -138,6 +144,52 @@ class LoadGameScreen(previousScreen:BaseScreen) : PickerScreen(disableScroll = t
...
+    }
+
+    private fun loadMissingMods() {
+        loadMissingModsButton.isEnabled = false
+        descriptionLabel.setText("Loading...".tr())
+        crashHandlingThread(name="DownloadMods") {
+            try {
+                val mods = missingModsToLoad.replace(' ', '-').lowercase().splitToSequence(",-")
+                for (modName in mods) {
+                    val repos = Github.tryGetGithubReposWithTopic(10, 1, "$modName+") // + is important here!
```
SomeTroglodyte, line 165: `+` why? ... I see. OK that should be the responsibility of the function not the caller.

SomeTroglodyte, line 165: `+` why? ... I see. OK that should be the responsibility of the function not the caller.

SomeTroglodyte:
Had the idea repeatedly, just never dared attempt it.
I believe Mod manager is threaded in a way (though it's not advertised and the UI gets confusing) that - if fast enough - you can download several mods in parallel and their installation would even complete if you left the manager before they finish arriving. This one - threaded and sequential, good enough.

core/src/com/unciv/ui/saves/LoadGameScreen.kt
JackRainy: Actually, it has been working even without `+` on my tests but I decide to stay on the safe side and added that.
And yes, I agree, it's better to have it _within_ the function but this is quite small and I was too lazy.
Will fix next time working there.

PR #6720, opened 2022-05-07
Fix Barbarian Camp Spawned notification not revealing the camp on the map in multiplayer

Azzurite:
Currently, in singleplayer, when you have the "Honor" policy that shows you when barbarian camps spawn, that camp is revealed in fog, so you can remember where it is. However, in multiplayer, the camp is not revealed. You still get the notification and you can click it and see where the camp spawned because the notification has a tile attached, but the "improvement icon" and pixel art of the camp is not shown.

**Why that happens**

The reason for that is this:

https://github.com/yairm210/Unciv/blob/822ce5e041240fddf1394006c95b2d69c1d6c50f/core/src/com/unciv/logic/civilization/CivilizationInfo.kt#L169-L175

Sort of, yeah. Because in 

https://github.com/yairm210/Unciv/blob/822ce5e041240fddf1394006c95b2d69c1d6c50f/core/src/com/unciv/logic/BarbarianManager.kt#L163-L164

only the transient field is set. This transient field is then persisted to the `lastSeenImprovementSaved` during the next update (Remember this for later!).

But when does the next update actually happen after barbarian camps spawned? In short, the next turn gets handled in `WorldScreen#nextTurn`, which then creates a new `WorldScreen` with all the updated data, and after the new `WorldScreen` is initialized, an update is requested and it updates on the next frame.

This is no problem in singleplayer, because in `WorldScreen#nextTurn`, the `gameInfo` gets cloned locally and the next screen is built from this local clone. The cloning process copies the transient field.

In multiplayer, the data for the next turn gets uploaded to the server in `WorldScreen#nextTurn`. Which is unfortunately before the next update. And now remember, the transient field only gets moved over to the non-transient one during update! So whatever we uploaded does not contain the updates to `lastSeenImprovement`! And since, after uploading, we immediately download that data again and use that downloaded data to make a new `WorldScreen`, all the data in `lastSeenImprovement` is lost.

Fun, fun, fun.

**Solution**

So, there are multiple solutions for this. 

One easy one would be to keep `lastSeenImprovement` and `lastSeenImprovementSaved` better in sync. Maybe create a custom setter or some shit that automatically updates both.

That would've worked with very little structural changes. The problem is that we have to do this for every field like this. And it requires us to do more work during the game. And it might still lead to hard to diagnose bugs like this one.

Another solution would've been to fix this for all maps automatically generically. This would've been my preferred solution, i.e. just look at a map via reflection, if its keys are not `String`s, then use a custom serialization, if the keys *are* `String`, just use the default one from GDX.

However, the problem is that the `Json` class from GDX is very problematic to extend. All the default code is in the `read/writeValue` functions. And inside these functions, the only way to add custom serialization behavior is to use a custom serializer for a *specific* class, i.e. you can't just say "use this serialization for all values that `extends Map`, you have to actually do it for specific classes like `HashMap`. But who knows if we're actually going to use `HashMap`s everywhere? Probably not, I don't even know what kotlin uses under the hood by default.

So I decided against implementing it generically, and instead created a custom class `HashMapVector2<T>` that always has `Vector2` as its key, and that's it. The `NonStringKeyMapSerializer` is then added to all usages of `Json` in our code, which correctly serializes this map. This is easily extensible with maps that have other key types, should we ever need them. I think this is the simplest solution without having to re-implement any code in `Json` or do any other hacky stuff.

**Open Questions**

1. Most instance creations of `Json` came from `fun GameSaver.json()`, but there are some classes, like `ImagePacker`, `JsonParser` and `CrashScreen` (among a few others) that instantiate their own `Json` objects. Should these also be changed to the new `UncivJson.json()` now?
2. Is the way that I did the game save migration correct? I'm not sure if that's how I am supposed to do it.


SomeTroglodyte:
> for some reason

I know, not your comment. Reason is - JSON rules.

> that the Json class from GDX is very problematic to extend

You too? Maybe I should have warned you - I tried, too, and gave up.

Q1 - ImagePacker just reads a simple array, CrashScreen -> definite yes, JsonParser -> merge with Yours and use everywhere I'd say

yairm210 (approved):
When I saw '27 files changed' I thought this would be a massive headache but it looks okay
Existing games continue as normal, I assume?

SomeTroglodyte:
> looks okay

To me too, in fact far better than okay because it replaces a maximally ugly kludge I always thought would come back to bite us with something elegant. No formal approval from me _only_ because I'm still reading / experimenting.

core/src/com/unciv/json/HashMapVector2.kt
```diff
@@ -10,12 +10,14 @@ import java.util.HashMap
 class HashMapVector2<T> : HashMap<Vector2, T>() {
     companion object {
         init {
+            @Suppress("UNCHECKED_CAST") // kotlin can't tell that HashMapVector2 is also a MutableMap within generics
```
SomeTroglodyte, line 13: "Honestly I've no fucking clue there" - you got it :+1: :+1: :+1:. Goes to show why justifying every single suppress is a good thing (and I just found a violation of that  elsewhere - whodunnit? me)

SomeTroglodyte:
> extend the serializer selection without re-implementing

Yeah, slowly arriving at the same conclusion. Except maybe - are all subclasses of something in a module, as the compiler would know when it's done, be available at runtime? Then fixing setSerializer to add them all (unless a specific one already exists) could be an option... Nah. Still a freakin' annoying _pity_ 😤 we can't arrive at the same result but leave the Map<Vector2, *> as they are...

SomeTroglodyte:
> freePolicies = 1

:fearful: but that's cheating! :scream:  <sub><sup>and exactly the way I kickstart new testing games 😇 </sup></sub>

SomeTroglodyte (approved):
Since I approve and am the third vote I should _do_ the merge - never merged here before - do we squash?

yairm210:
We do indeed squash

SomeTroglodyte:
I hope you like the squashed commit rename

PR #6724, opened 2022-05-08
Fix old barbarian camps not being properly loaded after 86d5011

Azzurite:
In https://github.com/yairm210/Unciv/pull/6720#issuecomment-1120311197, I said that barb migration should not be necessary, but I tested wrongly. The `camps` property actually needs migration from the old data format.

The problem is that when changing the type of the property and supplying this new custom serializer, of course the new serializer is always used. But since the data is still in the old format, the serializer doesn't deserialize correctly since it assumes the new format.

So I had to add a switch in the serializer itself, changing its behavior depending on if there are still `String` keys in the map to use the old deserialization (this is a direct copy of the code in `Json.readValue` for all `instanceof Map` values) or the new one if there are not.

### DIFF
--- a/core/src/com/unciv/logic/BackwardCompatibility.kt
+++ b/core/src/com/unciv/logic/BackwardCompatibility.kt
@@ -1,6 +1,9 @@
 package com.unciv.logic
 
 import com.badlogic.gdx.math.Vector2
+import com.badlogic.gdx.utils.JsonValue
+import com.unciv.json.HashMapVector2
+import com.unciv.json.json
 import com.unciv.logic.city.CityConstructions
 import com.unciv.logic.city.PerpetualConstruction
 import com.unciv.logic.civilization.CivilizationInfo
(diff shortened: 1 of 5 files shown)


SomeTroglodyte:
> I thought would be running before everything.

The wonders never cease. I'd have sworn, too, that all statics are ready before business as usual begins...

... Call a dummy in _your_ companion from _UncivGame_'s companion?

core/src/com/unciv/logic/BackwardCompatibility.kt
```diff
@@ -159,4 +162,38 @@ object BackwardCompatibility {
...
+            }
+
+            camps.clear()
+            camps.putAll(newFormat)
+        }
+    }
+
+    private fun isOldFormat(manager: BarbarianManager): Boolean {
+        val keys = manager.camps.keys as Set<Any>
+        val iterator = keys.iterator()
```
SomeTroglodyte, line 190: Advantage over `return keys.any { it is String }`?

SomeTroglodyte:
Old format in json: `camps:{(8.0,-5.0):{position:{x:8,y:-5},countdown:7,spawnedUnits:3},...`
New format in json: `camps:{class:com.unciv.json.HashMapVector2,entries:[[{x:1,y:2},{class:com.unciv.logic.Encampment,position:{x:1,y:2},countdown:4,spawnedUnits:2}],...`
Hmmmmmmm... I suppose that's exactly as it must be, no simplifying shortcuts come to mind... Well, making HashMapVector2 open and deriving a non-generic from it without changing implementation per usecase could drop the need to qualify the Encampent. Not worth it probably.

SomeTroglodyte:
Didn't see in first PR: `createSerializer` doesn't use `getSerializerClass`.. Could improve readability by 0.0001%?

SomeTroglodyte (approved):
Looks good. Gives itchy fingers, though. Their tips are crying 'hey, let's whip that Gdx.Json into [DWIM](https://en.wikipedia.org/wiki/DWIM)!'... Shut up Fingers, you don't know nothing. That thing resists common sense!

core/src/com/unciv/logic/BackwardCompatibility.kt
yairm210: Nope, it's the same

PR #6725, opened 2022-05-08
Leftovers from previous PRs

JackRainy:
Minor "nice-to-have" feedback received on post-merge code reviews


yairm210:
"Could not find a mod named '[modName]'", perhaps

SomeTroglodyte (approved):
An exception is a technical thing, so an exception message maybe should be culture-agnostic until the very point it is shown to the user? Dunno. Tinysquishy certainly doesn't think so... They rabidly (yes the foam-at-muzzle variant) push everything thru their abysmal machine translator until a coder has no chance at all to understand a message in their native language, unless it's en-US. :see_no_evil:

PR #6730, opened 2022-05-09
Vector2Serializer for more compact save games

SomeTroglodyte:
This one is a kind of an easter egg. @Azzurite I hope it amuses, and thanks.

Largest game in my saves (shape:Rectangular,mapSize:{radius:40,width:87,height:57,name:Huge},worldWrap:true): 
| variant | bytes | sample |
| ------- | ----- | ------ |
| default | 1279592 | exploredTiles:[{x:10,y:-7},{x:-10,y:-7},{x:41,y:15},{x:-41,y:-15},{x:44,y:2},... |
| array | 1109037 | exploredTiles:[[10,-7],[-10,-7],[41,15],[-41,-15],[44,2],... |
| string | 1007319 | exploredTiles:[10~-7,-10~-7,41~15,-41~-15,44~2,... |
| base32 | 918233 | exploredTiles:[q9,69,9hfg,nehf,ch2g,... |
| base64 | 864595 | exploredTiles:[qZ,WZ,pgPg,Xfxf,sgCg,... |

Sorry I didn't do any size tests for gzip compression, or performance measurements. All formats do a save/load cycle just fine, and the time that takes at least 'feels' normal.

There's an ugly hard switch to choose output format, but it accepts all formats on load (except base32 which I replaced after measuring). If a reduction in (uncompressed) save size of ~33% with complex code; or ~22% with easy code; or ~13% using still easily readable json - is attractive, I would trim this down to just one output format and two input (new and legacy).

### DIFF
--- a/core/src/com/unciv/json/UncivJson.kt
+++ b/core/src/com/unciv/json/UncivJson.kt
@@ -4,7 +4,6 @@ import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.utils.Json
-import com.badlogic.gdx.utils.Json.Serializer
 
 
 /**
@@ -14,6 +13,7 @@ fun json() = Json().apply {
     setIgnoreDeprecated(true)
     ignoreUnknownFields = true
 
+    setSerializer(Vector2::class.java, Vector2Serializer())
     setSerializer(HashMapVector2.getSerializerClass(), HashMapVector2.createSerializer())
 }
 
--- a/core/src/com/unciv/json/Vector2Serializer.kt
+++ b/core/src/com/unciv/json/Vector2Serializer.kt
@@ -1,0 +1,155 @@
+package com.unciv.json
+
+import com.badlogic.gdx.math.Vector2
+import com.badlogic.gdx.utils.Json
+import com.badlogic.gdx.utils.JsonValue
+import com.badlogic.gdx.utils.SerializationException
+import kotlin.math.max
+
+
+class Vector2Serializer : Json.Serializer<Vector2> {
+    companion object {
+        private const val outputFormat = '0'  // {=object, [=array, ~=compact, 0=base64
+
+        private fun Int.toBase64Char() = Char(when(this) {
+            in 0..25 -> 65 + this  // 'A'..'Z'
+            in 26..51 -> 97 - 26 + this  // 'a'..'z'
+            in 52..61 -> 48 - 52 + this  // '0'..'9'
+            62 -> 64  // '@', bash flavour
+            63 -> 95  // '_', bash flavour
+            else -> 61  // '='
+        })
+        private fun Int.toBase64String(): String {
+            val sb = StringBuilder()
+            var residue = this
+            while (true) {
+                sb.append(residue.mod(64).toBase64Char())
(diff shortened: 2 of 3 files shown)


yairm210:
I see no reason not to go for base64

yairm210:
Obviously the most important thing is that we retain the ability to load old games

SomeTroglodyte:
OK, I'll tweak the thing a bit and do some of the missing measurements (if only to reassure myself)... There's still a percent or two to be tickled out of it.

And - since the old format is an object to Gdx json but the new is plain string - no problem loading 'old' games.

JackRainy:
Just to be sure, did you consider any other [3rd-party serializers](https://github.com/eishay/jvm-serializers/wiki) before inventing own one?
For example, [Kryo](https://libgdx.com/wiki/utils/saved-game-serialization) which is 100% compatible with libGdx (assuming they have the same authors).

SomeTroglodyte:
| variant | Size | Load [ms] | Save [ms] |
| ------- | ----:| ---------:| ---------:|
| master | 1279592 | 271 | 341 |
| string | 1007319 | 462 | 150 |
| base64 | 855550 | 485 | 401 |
| master+gzip | 248948 | 333 | 355 |
| string+gzip | 222424 | 436 | 231 |
| base64+gzip | 207732 | 256 | 227 |

Lazily done with math in my head and ~6 repeats per row and load/save, then discarding any lower outlier if too extreme and taking the average of the two best runs. Variance was extreme, the first load always being > 3 times the best one. Measurements include json, gzip and disk but not setTransients.

Some surprising values, but that's what I saw. How can base64 be such a slowdown unzipped but much faster zipped? Can't be pure disk I/O! Strange. The bad 'Load' time for string+gzip at first was ~480 which surprised me enough to rewrite the parser to use indexOf(Char) instead of split, but it didn't get that much better.

Remember, only the first two are human-readable as @Azzurite mentioned. Votes? First thing I'll do anyway is pick out the switchable zipping and PR that.

SomeTroglodyte:
> saving to file the entire string is base64'd and gzip'd anyway

Not quite. It is gzipped then base64'ed _**if**_ for Multiplayer or Clipboard, not for normal saves, neither android nor desktop. The base64 here takes the bits of a int-converted Vector2 and serializes them as base64 -> `{x:41,y:15}` becomes "p8h". Taking the entire json and zip+base64 it afterwards is independent and was done in the last line of the statistics.

Having the clipboard copy readable plain text is attractive in issues, but we wanted to maximize chances github doesn't deny service due to size, right?

SomeTroglodyte:
Hey y'all - another save game size saving idea: `CivilizationInfo.exploredTiles` takes up large amounts of space which the original approach reduced at the Vector2 level. But subclassing HashSet<Vector2>() to something with (transient) knowledge about map dimensions could allow a custom serializer for that class which does a bitmap over all tiles then base64-codes that. Transparently with old saves. Meaning a typical late-game exploredTiles of 60k (<5000 tiles) would go down to 820 chars, staying constant over the whole game, in turn meaning early saves are larger until ~80 tiles explored per civ are exceeded, ~10 turns I'd say. In the save I used for this estimation, civ data is ~80% map ~20% and each civ has at most ~12k of non-exploredTiles data (city states much less), that would mean... a lot. 

Madness or Genius or simply sad not to have a Real Life:tm:?

PR #6735, opened 2022-05-09
Switchable gzipping of saved games

SomeTroglodyte:
- Saves by default are gzipped, switchable per session in debug settings.
    - Affects user saves, autosaves, CustomLocation
    - Not under zip switch control: OnlineMultiplayer (always zipped, now always routed through GameSaver)
    - Not under zip switch control: Copy & Paste (always zipped, now always routed through GameSaver)
- Leaving WorldScreen to MainMenuScreen didn't Autosave - now does
- Default save name now translatable (why not - but honestly I wanted the double space gone. Re-add in English.properties if so desired?)
- 11 uses of "Autosave" now const (almost gave the Thread name a separate symbol)
- GameSaver object region-ized

Questions:
- Home of "Autosave" const val? GameSaver as pushed or Constants instead?
- Default state of GameSaver.saveZipped? If we change it to false, save behaviour is unchanged (most of that tested).

Independent of the puzzle explored in #6730

image image
</details>

### DIFF
--- a/android/assets/jsons/translations/German.properties
+++ b/android/assets/jsons/translations/German.properties
@@ -581,6 +581,8 @@ Days = Tage
 Current saves = Gespeicherte Spiele
 Show autosaves = Zeige automatisch gespeicherte Spiele an
 Saved game name = Name des gespeicherten Spiels
+# This is the save game name the dialog will suggest
+[player] - [turns] turns = [player] ([turns] Runden)
 Copy to clipboard = In die Zwischenablage kopieren
 Copy saved game to clipboard = Gespeichertes Spiel in die Zwischenablage kopieren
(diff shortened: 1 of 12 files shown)


SomeTroglodyte:
> should at least be a setting in the options

See screenshutz, it's RCtrl-RShift-O only

> feature

One vote for default false :wink:

yairm210:
I think it's fine to have default false, usually the couple of MB you save per fine is negligible on an FS, and this is coming from someone who half a year ago had to remove apps from his phone if he wanted to download another app.

In my eyes the biggest gain from this PR is actually the try/catch on the loading, so we can load games in an unzipped format, which solves the 'unzipped Dave uploaded to pastebin' problem for me

yairm210 (approved):

PR #6741, opened 2022-05-09
Fix spectator not being able to view tile improvements or city buttons

Azzurite:
I'd really like to refactor vision to a separate file that is solely responsible for vision, and then pulling vision out of all the other classes.

But that's for another time :D This is just to fix the spectator not being able to view improvements or city buttons. Pretty straightforward.

### DIFF
--- a/core/src/com/unciv/ui/tilegroups/WorldTileGroup.kt
+++ b/core/src/com/unciv/ui/tilegroups/WorldTileGroup.kt
@@ -31,11 +31,15 @@ class WorldTileGroup(internal val worldScreen: WorldScreen, tileInfo: TileInfo, 
                 && city!!.civInfo == viewingCiv)
             icons.addPopulationIcon()
         // update city buttons in explored tiles or entire map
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/ui/tilegroups/WorldTileGroup.kt
```diff
@@ -31,11 +31,15 @@ class WorldTileGroup(internal val worldScreen: WorldScreen, tileInfo: TileInfo,
                 && city!!.civInfo == viewingCiv)
             icons.addPopulationIcon()
         // update city buttons in explored tiles or entire map
-        if (showEntireMap || viewingCiv.exploredTiles.contains(tileInfo.position))
+        if (showEntireMap
+                || viewingCiv.exploredTiles.contains(tileInfo.position)
+                || viewingCiv.isSpectator()
```
yairm210, line 36: This line is not needed due to the next line, no?

yairm210: Gotcha

yairm210 (approved):

PR #6742, opened 2022-05-09
Hide Unmet Civ and Capital Names in the Victory Status Screen

OptimizedForDensity:
Obscures the names of civs the player has not met in the "Our Status" tab of the Victory Status screen by replacing their names with "???" (```Constants.unknownNationName```). In the event that the civ dies before the player meets them, the civ name will be revealed anyways. Since the other Victory Status tabs hide the unmet civ names I don't see any reason why this one tab should spoil them.

Performs the same obscuring with the Religion victory type (which is currently unimplemented in the vanilla game). ~~The same could be done for the take-all-capitals victory condition but since there could be cases where the city name becomes unobscured then reobscured (plus with this victory type being unimplemented) I decided to leave that for a later time.~~ EDIT: Capital names are now replaced with a ```Constants.unknownCityName``` string (currently "???") until the player has seen the capital city tile.

### DIFF
--- a/core/src/com/unciv/models/ruleset/Victory.kt
+++ b/core/src/com/unciv/models/ruleset/Victory.kt
@@ -213,7 +213,10 @@ class Milestone(val uniqueDescription: String, private val parentVictory: Victor
             MilestoneType.CaptureAllCapitals -> {
                 val originalCapitals = civInfo.gameInfo.getCities().filter { it.isOriginalCapital }
                 for (city in originalCapitals) {
-                    buttons.add(getMilestoneButton("Capture [${city.name}]", city.civInfo == civInfo))
+                    val milestoneText =
+                        if (civInfo.exploredTiles.contains(city.location)) "Capture [${city.name}]"
+                        else "Capture [${Constants.unknownNationName}]"
+                    buttons.add(getMilestoneButton(milestoneText, city.civInfo == civInfo))
                 }
             }
             
--- a/core/src/com/unciv/models/ruleset/Victory.kt
+++ b/core/src/com/unciv/models/ruleset/Victory.kt
@@ -215,7 +215,7 @@ class Milestone(val uniqueDescription: String, private val parentVictory: Victor
                 for (city in originalCapitals) {
                     val milestoneText =
                         if (civInfo.exploredTiles.contains(city.location)) "Capture [${city.name}]"
-                        else "Capture [${Constants.unknownNationName}]"
+                        else "Capture [${Constants.unknownCityName}]"
                     buttons.add(getMilestoneButton(milestoneText, city.civInfo == civInfo))
                 }
             }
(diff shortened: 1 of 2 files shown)


SomeTroglodyte (approved):
Welcome Mr. Black Hole! 
Looks simple enough to approve untested. Reminds me `unknownNationName` could be a `ModConstant`.

core/src/com/unciv/models/ruleset/Victory.kt
```diff
@@ -223,10 +227,13 @@ class Milestone(val uniqueDescription: String, private val parentVictory: Victor
...
                 for (civ in majorCivs) {
+                    val milestoneText =
+                        if (civInfo.knows(civ) || !civ.isAlive()) "Majority religion of [${civ.civName}]"
+                        else "Majority religion of [${Constants.unknownNationName}]"
                     if (civReligion == null || (civReligion.isPantheon() && civInfo != civ) || !civ.religionManager.isMajorityReligionForCiv(civReligion))
-                        buttons.add(getMilestoneButton("Majority religion of [${civ.civName}]", false))
+                        buttons.add(getMilestoneButton(milestoneText, false))
                     else
-                        buttons.add(getMilestoneButton("Majority religion of [${civ.civName}]", true))
+                        buttons.add(getMilestoneButton(milestoneText, true))
```
yairm210, line 236: Send to me the second parameter passed to this function is the reverse if the if, so we can replace the if with a parameter

yairm210: Those isn't a problem with your code though, existed before, so I approve :)

yairm210 (approved):

xlenstra:
Shouldn't the city names on line 216 also be obfuscated, as you can otherwise still learn the nations from their capital names?

core/src/com/unciv/models/ruleset/Victory.kt
OptimizedForDensity: Done, thanks to De Morgan's laws.

OptimizedForDensity:
They should and I had some ideas on how to do it (Does the founding civ still own the city? Does the player know the current owner? Has the player revealed the capital city tile?) but I ran into edge cases where the city name could revert into a hidden state after being revealed (namely, if a civ that the player has met captures another civ's capital but then the original owner recaptures it). Between this and the fact that the vanilla game doesn't have this victory condition enabled, I decided to leave it for later but I can go back and change it if others think it should have some level of obfuscation.

xlenstra:
> They should and I had some ideas on how to do it (Does the founding civ still own the city? Does the player know the current owner? Has the player revealed the capital city tile?) but I ran into edge cases where the city name could revert into a hidden state after being revealed (namely, if a civ that the player has met captures another civ's capital but then the original owner recaptures it). Between this and the fact that the vanilla game doesn't have this victory condition enabled, I decided to leave it for later but I can go back and change it if others think it should have some level of obfuscation.

I'd say that checking if the player has revealed the capital city tile it should be shown, regardless of the current owner of the city. I don't think that should ever revert? If you want I can merge this right away and look into this myself.

OptimizedForDensity:
That's probably a better condition than what I was thinking with civilization contact (in my head, it seemed reasonable that an explorer would talk about how grand their civ's capital is but now I can see why going with a simpler case would be better as a start). I added that condition as well as an ```unknownCityName``` condition in ```Constants``` to allow for reuse/easy renaming but I can remove if this is unnecessary.

xlenstra (approved):

PR #6743, opened 2022-05-10
Autofix Multiplayer Server URL on input

### LINKED ISSUE #6583: Autofix URLs from user input at `Options` > `Multiplayer` >  `Server address`
So, I have a Multiplayer Server with the URL `https://uncivserver.xyz`.
Some people tries to connect to the server with the URL `https://uncivserver.xyz/` (with a trailing `/`) and says that it is not working.
Some might even input unnecessary spaces ?

I actually tried to fix this from server side using the following logic:
1. If the URL Path contains Repeated `/` than replace them with a single `/`
2. Remove all whitespaces from Request Path

On Commit: https://github.com/touhidurrr/UncivServer.xyz/commit/257b209be79d85d93ebae2fd3fe1cbf4f6e9147e

But for some reason this kind Autofix attempt from server side doesn't work. (Maybe beacuse of some kind of Clientside logic)
More specifically this did work for `GET //isalive` but not on `GET //files/:fileName` and `PUT //files/:fileName` for some reason **from Unciv**.
If you try to test it yourself by visiting URLs like:

- https://uncivserver.xyz///files/startup or
- https://uncivserver.xyz////files//////startup

You would find that it does indeed work so I would assume that this autofix attempt didn't work for some clientside logic maybe.

Anyways, since this kind of typos are quite common, I suggest URLs should be automatically formatted from Client Side instead.
The Fixes I suggest are:
1. Remove any whitespaces from the URLs
2. Remove Repeated `/` characters (Cause if the client sends urls like `https://example.com//isalive` it's probably a mistake)

touhidurrr:
resolves #6583

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
@@ -962,13 +962,11 @@ class OptionsPopup(
         // replace multiple slash with a single one
         result = Regex("/{2,}").replace(result, "/")
 
-        // remove trailing slash and return
-        result = Regex("/$").replace(result, "")
-
-        // reinstate protocol & return
+        // remove trailing slash, reinstate protocol & return
         // all the formatting above makes "https://" -> "http:/"
-        // so we need to fix that before returning
-        return Regex(":/").replaceFirst(result, "://")
+        // also people might leave a slash at end my mistake
+        // so we need to fix those before returning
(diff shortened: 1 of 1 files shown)


touhidurrr:
Test
```kt
val input = "    https://  unci vser ver. xyz////fi les//////startup//      "
println(formatMultiplayerUrlInput(input))
// prints https://uncivserver.xyz/files/startup
```

yairm210 (approved):

PR #6744, opened 2022-05-10
Bump kotlin language version to latest 1.6.21

Azzurite:
After running migrations, the only thing reported was that non-exhaustive when's will be an error in kotlin 1.7, so I added `else` branches to those.


SomeTroglodyte:
So you're the pro-active guy. How about running our gradle scripts with full logging and prepare them to be compatible with gradle 8/9? (But don't bump that version - syntax only)

.. and I personally would use `Unit` instead of `{}`

SomeTroglodyte:
It's something I have at the back of my head for a while, not sure how I did that last time. Builds will warn that the scripts are using deprecated notations. Gradle 6 to 7 was similar, but once you find how to get the logs to show which lines precisely they dislike, you get a migration hint or link with it.

..No of course not. Sound argument? Fine. Teaser: I'd bet `;` would do it too. Or `null`. Or `@Suppress("UNUSED_EXPRESSION") "This line intentionally left blank"`  ...  Wait - best argument: The kotlin author's choice for what should rightfully be called `void` is so abstruse it should be celebrated!

SomeTroglodyte:
> brb changing it right now

Hey I wasn't serious

> gradle

Yes, completely unrelated, just some task needed anyway sooner or later..

SomeTroglodyte:
```
./gradlew --warning-mode all --stacktrace tasks

The JavaExec.main property has been deprecated. This is scheduled to be removed in Gradle 8.0. Please use the mainClass property instead. See https://docs.gradle.org/7.2/dsl/org.gradle.api.tasks.JavaExec.html#org.gradle.api.tasks.JavaExec:main for more details.
        at org.gradle.api.tasks.JavaExec.setMain(JavaExec.java:420)
        at org.gradle.api.tasks.JavaExec_Decorated.setMain(Unknown Source)
        at Build_gradle$3.invoke(build.gradle.kts:35)
        at Build_gradle$3.invoke(build.gradle.kts:3)
```
That's not much, and the stacktrace points to a mystery - no JavaExec.main at all over there. Task 'tasks' is probably not the right place to start looking for gradle deprecations. Now _how_ does one run a gradle _sync_ with those options??? That's what I want to see...

By the way:
```
w: ~/Work/Unciv/core/src/com/unciv/json/NonStringKeyMapSerializer.kt: (46, 34): Unnecessary non-null assertion (!!) on a non-null receiver of type JsonValue
w: ~/Work/Unciv/core/src/com/unciv/json/NonStringKeyMapSerializer.kt: (53, 26): Unchecked cast: MT to MutableMap<String, Any>
w: ~/Work/Unciv/desktop/src/com/unciv/app/desktop/DesktopLauncher.kt: (73, 20): 'loadLibrary(String!, Class<T!>!): T!' is deprecated. Deprecated in Java
```
...would be some more places to proactively look for code to clean. Dunno whether I'll be able to try my hand at the research soon.

SomeTroglodyte:
But please, Pavlovically Currect language. "The intercourse", not "The f**k"!

I'll need to try this separately.

SomeTroglodyte:
Trying the migration from master myself, I can say I fully agree with treating that idiotic deprecation of `@Suppress("NON_EXHAUSTIVE_WHEN")` like this. Kudos for even 'migrating' that suppress comment that still applies.

You know what I think? PR the warnings treatment without the buildConfig change - proactively treating those is definitely good :tm: - then we can look at the language bump separately.

FYI
- `milliseconds(Long): Duration' is deprecated. Use 'Long.milliseconds' extension property from Duration.Companion instead` in Simulation.kt L122 --> it wants `totalDuration = (endTime - startTime).milliseconds`.
- DesktopLauncher.kt L73 `Native.load` works
- TranslationTests.kt L173 unused variable `translation` -> `_`
- There's another in class Milestone L185 - I would add the `null` the the first branch
- DiplomacyManager L492 fold filter+any
- Victory L143 fold filter+count
Together with the two in NonStringKeyMapSerializer.kt that should reduce warnings to deprecations only??? What do you think?

yairm210:
Looks fine, as usual or policy is to be on the latest stable versions of everything, so if there are problems we'll solve them.

PR #6745, opened 2022-05-10
Add UncivServer.jar to github release

Azzurite:
Please someone test it I have no idea if it works aaaaaaa

Tested and works: https://github.com/Azzurite/Unciv/releases/tag/test_server_jar

### DIFF
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -160,7 +160,17 @@ jobs:
           ./butler push deploy/Unciv-Windows64.zip yairm210/unciv:Windows64 --userversion ${{steps.tag.outputs.tag}}
 
           mv desktop/build/libs/Unciv.jar deploy/Unciv.jar
-          
+
+## Server
+
+      - name: Build UncivServer.jar
+        continue-on-error: true
+        if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
+        run: |
+          ./gradlew server:dist
+          mv server/build/libs/UncivServer.jar deploy/UncivServer.jar
+
+## Finalize
 
       - name: Upload binaries to Github release
         if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')


SomeTroglodyte:
No idea how, but later. Seems you've found an approach to something I wanted to do in that surgery PR but didn't investigate.

SomeTroglodyte:
> :laughing: 

:rofl: :zany_face:

SomeTroglodyte:
Ah. of course, yml. Yair's build scripts are off by default in our forks, but you can turn them on. Then most will run. e.g. I push a bungled templates.properties I get a mail _before_ the branch is PR'ed, if I take too long. Since our config isn't complete - no secrets - nothing breaks, or at least I hope that's how it goes (e.g. no itch.io).

... that should maybe be an approach to test this. In my case, I just have to push the right combo to my fork... breaking master sync. hmmm...

touhidurrr:
Also, is there any need to pack UncivServer.jar with windows zips?

touhidurrr:
I think and Wiser approach would be describing Windows users how to install Java Runtime in docs, download UncivServer.jar and running it from command line. As most users wouldn't even be interested in this anyways.

SomeTroglodyte:
> If so, you'd be the one to know, because you made the PR that creates UncivServer.jar

That was SomeTroglodyte, not touhidurr. And the exe-creating thing + MacOS packaging worked when I tested _that_ one back then, but not out of the box as some of the steps can't easily repro locally, so I supplied jre-macOS.tar.gz and packr-all-4.0.0.jar manually. Haven't documented these things in my scratchpad... maybe I should. It was _not_ simply `./gradlew desktop:packr`...

touhidurrr:
> Oh right :D sorry.

Lol i misread too. I was quickly scrolling through notifs so I thought you wrote something like "You would know better since you were discussing about it". 🤣

My bad too.

alexban011:
How do you start the server? Since this doesn't work I'm guessing this isn't the way to do it right?
`
>java UncivServer.jar
Error: Could not find or load main class UncivServer.jar
`

alexban011:
> [wiki](https://yairm210.github.io/Unciv/Other/Hosting-a-Multiplayer-server/) or right here in this discussion???

I got it in the meantime thanks! Should have checked there first before commenting

SomeTroglodyte:
So, itch.io won't get the server jar, and nobody gets packr'ed servers. If we were to want itch to get the server jar, we could move the lines to where butler is ready and butler it with the others. We could also packr the server, I remember preparing the requirements but stopping at that because locally packr'ing Unciv itself felt strange - see original PR. Still, `./gradlew server:packrLinux64`, `./gradlew server:packrWindows32`, `./gradlew server:packrWindows64` and `./gradlew server:zipLinuxFilesForJar` _are_ prepared as tasks, and should still test out with a little help locally (do the wget's beforehand). The code to build packr'ed Unciv is in the itch.io section - copying to do server too would not be that much.

I'm fine with this little step, but what do you think...?

SomeTroglodyte:
> think before hollering

.. now where's the spanking emoji?

yairm210:
Anyone trying to run the UncivServer has to know what they're doing. Running a jar from cmd is the least they'll need to do, there's no need to packr them up.

touhidurrr:
> Anyone trying to run the UncivServer has to know what they're doing. Running a jar from cmd is the least they'll need to do, there's no need to packr them up.

Yes! That's exactly what I was trying to say!!! The summary of it!!!
Thanks for your sweet & brief words!!!!!

touhidurrr:
> there's no need to packr them up

Btw, did you mean there is no need pack in with zip files or add to releases?

SomeTroglodyte (approved):

SomeTroglodyte:
> no need pack in

"packr" means https://github.com/Anuken/packr here

PR #6721, opened 2022-05-08
Missing uniques for Great Persons and others - part 1

JackRainy:
While this PR resolves #6719 issue, however, it is only a first part of the changes.
The second part will be to utilize the newly added uniques properly.

PR #6749, opened 2022-05-10
Missing typed uniques for the units - part 2

JackRainy:
This is the continuation of migration for the typed uniques.
See also, the part 1 - #6721

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -452,7 +452,6 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     HealOnlyByPillaging("Can only heal by pillaging", UniqueTarget.Unit, UniqueTarget.Global),
     HealsEvenAfterAction("Unit will heal every turn, even if it performs an action", UniqueTarget.Unit),
     HealAdjacentUnits("All adjacent units heal [amount] HP when healing", UniqueTarget.Unit),
-    HealInstantly("Doing so will consume this opportunity to choose a Promotion", UniqueTarget.Unit),
 
     NormalVisionWhenEmbarked("Normal vision when embarked", UniqueTarget.Unit, UniqueTarget.Global),
     DefenceBonusWhenEmbarked("Defense bonus when embarked", UniqueTarget.Unit, UniqueTarget.Global),
@@ -734,6 +733,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     OneTimeUnitUpgrade("This Unit upgrades for free", UniqueTarget.Global),  // Not used in Vanilla
     OneTimeUnitSpecialUpgrade("This Unit upgrades for free including special upgrades", UniqueTarget.Ruins),
     OneTimeUnitGainPromotion("This Unit gains the [promotion] promotion", UniqueTarget.Triggerable),  // Not used in Vanilla
+    SkipPromotion("Doing so will consume this opportunity to choose a Promotion", UniqueTarget.Promotion),
 
     UnitsGainPromotion("[mapUnitFilter] units gain the [promotion] promotion", UniqueTarget.Triggerable),  // Not used in Vanilla
     @Deprecated("as of 3.19.8", ReplaceWith("[+amount]% Strength <when attacking> <for [mapUnitFilter] units> <for [amount2] turns>"))
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -477,6 +477,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     DamageFromInterceptionReduced("Damage taken from interception reduced by [relativeAmount]%", UniqueTarget.Unit),
     DamageWhenIntercepting("[relativeAmount]% Damage when intercepting", UniqueTarget.Unit),
     ExtraInterceptionsPerTurn("[amount] extra interceptions may be made per turn", UniqueTarget.Unit),
+    CannotBeIntercepted("Cannot be intercepted", UniqueTarget.Unit),
 
     UnitMaintenanceDiscount("[relativeAmount]% maintenance costs", UniqueTarget.Unit, UniqueTarget.Global),
     UnitUpgradeCost("[relativeAmount]% Gold cost of upgrading", UniqueTarget.Unit, UniqueTarget.Global),
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -442,6 +442,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
 
     NoMovementToPillage("No movement cost to pillage", UniqueTarget.Unit, UniqueTarget.Global),
     CanMoveAfterAttacking("Can move after attacking", UniqueTarget.Unit),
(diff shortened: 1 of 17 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -451,6 +452,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     HealOnlyByPillaging("Can only heal by pillaging", UniqueTarget.Unit, UniqueTarget.Global),
     HealsEvenAfterAction("Unit will heal every turn, even if it performs an action", UniqueTarget.Unit),
     HealAdjacentUnits("All adjacent units heal [amount] HP when healing", UniqueTarget.Unit),
+    HealInstantly("Doing so will consume this opportunity to choose a Promotion", UniqueTarget.Unit),
```
xlenstra, line 455: Not sure I agree with this unique name and targets. It's use case is allowing modders to make promotions that can be taken multiple times, which isn't restricted at all to only healing instantly. Maybe `RepeatingPromotion` to make it similar to a repeating tech?
Additionally, it should only be allowed to put in Promotions, as it doesn't make sense for other Unit-like objects to have this unique.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -667,7 +667,7 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
                 unique.isOfType(UniqueType.StrengthNearCapital) && unique.params[0].toInt() > 0 ->
                     power *= (unique.params[0].toInt() / 4f).toPercent()  // Bonus decreasing with distance from capital - not worth much most of the map???
 
-                unique.placeholderText == "May Paradrop up to [] tiles from inside friendly territory" // Paradrop - 25% bonus
+                unique.placeholderText == UniqueType.MayParadrop.placeholderText // Paradrop - 25% bonus
```
xlenstra, line 670: `unique.isOfType(UniqueType.MayParadrop)`?
Or, even better, changing this entire when to query the type of a unique (`when (unique.type) {`)?

tests/src/com/unciv/testing/BasicTests.kt
```diff
@@ -157,6 +157,21 @@ class BasicTests {
...
+        for (promotion in promotions) {
+            for (unique in promotion.uniques) {
+                if (!UniqueType.values().any { it.placeholderText == unique.getPlaceholderText() }) {
+                    println("${promotion.name}: $unique")
+                    allOK = false
+                }
+            }
+        }
+        Assert.assertTrue("This test succeeds only if all uniques of promotions are presented in UniqueType.values()", allOK)
+    }
```
xlenstra, line 173: Not sure I agree with this test, as untyped uniques can be used as identifiers by modders

xlenstra (commented):
Other than these comments: looks good!

tests/src/com/unciv/testing/BasicTests.kt
yairm210: I agree with this test, since it's talking about the base game.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
SomeTroglodyte: After all, that's the point of the exercise.

tests/src/com/unciv/testing/BasicTests.kt
SomeTroglodyte: But sounds more like something for the modchecker? There is already a unit test calling it. Then maybe do a "completeness" check for each and every RulesetObject class, with for the moment differentiated expected results for base and mods? Overall uniquetypification completeness can be done using a debug option already.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
JackRainy: > it should only be allowed to put in Promotions, as it doesn't make sense for other Unit-like objects to have this unique.

Not sure how to declare where the promotion is allowed to be put. To use a different `UniqueTarget`? I still do not understand the difference between `UniqueTarget.Unit` and `UniqueTarget.Global`.

xlenstra: The `UniqueTarget` indeed means where uniques can be put. For example, `UniqueTarget.Nation` means that it can only be placed on nations, `UniqueTarget.Promotion` only means promotions, etc. The full list can be viewed on `UniqueType.kt:12`. You'll also notice that some unique targets have things between paranthesis, e.g. `Nation(Global)`. This indicates inheritance, i.e., that anything that has `UniqueTarget.Global` can also be put on a Nation. For example, `[stats] from every [tileFilter]` has as `UniqueTarget.Global` (and `UniqueTarget.FollowerBelief`), so you could put those on nations as well.

In this case, a unique that can be put on `UniqueTarget.Global` can be put on, for example, policies, techs, nations, wonders, buildings, eras, resources and founder beliefs as these all (indirectly) inherit from `UniqueTarget.Global`. Uniques that can be put on `UniqueTarget.Unit` can be put on units, unit types and promotions, as these all inherit from `UniqueTarget.Unit`.

For example the unique `[amount]% Strength` has both `UniqueTarget.Unit` and `UniqueTarget.Global` as, when combined with conditionals, it should be allowed on either. `[-33]% Strength <vs cities>` is a common unique for mounted _units_, so `UniqueTarget.Unit`, and `[+25]% Strength <for [All] units> <for [30] turns>`, the unique for finishing the autocracy policy branch, should be put on a `UniqueTarget.Policy` -> `UniqueTarget.Global`. 

Does this help in understanding?

SomeTroglodyte: > Does this help

_Now_ I'm confused. Just change the line to `UniqueTarget.Promotion` and move on...

JackRainy: Thanks Xander, now I know more... it smells like even more unit tests are needed to cover that, e.g. `UniqueTarget.Promotion` is not applied on the `Nation`, etc... but they go out of this PR, definitely.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
JackRainy: Initially, I thought about the same, however, `isOfType` does even more complex magic under the hood `UniqueType.values().firstOrNull { it.placeholderText == placeholderText }` which ends up with comparing the `placeholderText`, anyway. Thus I decided to keep it more obvious and simple.

JackRainy: Fair to say, I like the idea to have `when (unique.type) {` , probably, we can even get rid of `placeholderText` property at some point. I don't like the fact we store the same data in 2 places: in `placeholderText` and in `type.placeholderText` de facto.

SomeTroglodyte: `isOfType` using placeholderText is an interim kludge until all base game uniques are typed - or so I always thought? So using it while still biting your nails at its inefficiency is still a good thing :tm: 

`text` vs. `placeholderText` is not so easy - they're not the same data, and deriving the latter from the former on the fly is possible at a RegEx expense.

tests/src/com/unciv/testing/BasicTests.kt
JackRainy: > Then maybe do a "completeness" check for each and every RulesetObject class

I agree on that. The amount of those types (and types of types) is so large, that new developer will have a hard time to add/edit something there without a chance to mess up the things.

xlenstra (approved):

PR #6751, opened 2022-05-10
Great General UniqueTyped and improved moddability

SomeTroglodyte:
So far tested that G&K (new UniqueType) and Civ6 mod (old UniqueType) work as before, and a test mod shows a General can indeed learn "staggered" GG boni from promotions with this. Will obviously conflict with #6750.


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -412,6 +412,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     StrengthBonusVsCityStates("+30% Strength when fighting City-State units and cities", UniqueTarget.Global),
     StrengthForAdjacentEnemies("[relativeAmount]% Strength for enemy [combatantFilter] units in adjacent [tileFilter] tiles", UniqueTarget.Unit),
     StrengthWhenStacked("[relativeAmount]% Strength when stacked with [mapUnitFilter]", UniqueTarget.Unit),  // candidate for conditional!
+    @Deprecated("as of 4.0.15", ReplaceWith("[15]% Strength bonus for [Military] units in a [2] tile radius"))
+    BonusForUnitsInRadius("Bonus for units in 2 tile radius 15%", UniqueTarget.Unit),
+    GreatGeneralAura("[relativeAmount]% Strength bonus for [mapUnitFilter] units in a [amount] tile radius", UniqueTarget.Unit),
```
xlenstra, line 417: I personally prefer `[relativeAmount]% Strength for [mapUnitFilter] units in a [amount] tiles radius`, as the `+` in the relative amount should indicate the bonus. If this were to be negative (for some reason, don't ask me why you'd want that), a `[-20]% bonus` sounds a bit weird :upside_down_face:

SomeTroglodyte: Took me a second to grok - you suggest removing the word "bonus". OK, accepted. Yes, also matches the rest of the game better.

I'll wait for more feedback - is the maximized flexibility I went for overkill and JR's approach may please more? If so, point moot.

xlenstra:
I'm leaving this here as I've thought about this a bit on and off for a few months and is kinda relevant to this PR (though I don't want to imply this should be implemented in this PR if it all).

There are currently two uniques affecting units other than themselves. The one of great generals which is expanded in this PR, and the Haka War Dance unique. They both do the same thing, influence strength of units other than themselves, but for friendly and enemy units respectively. Imo they should also be treated the same, with the Haka War Dance unique being changed to have a radius of 1 instead of 'adjacent'. We could then have two uniques, `[relativeAmount]% Strength for friendly [mapUnitFilter] units in a [amount] tiles range` and `[relativeAmount]% Strength for enemey [mapUnitFilter] units in a [amount] tiles range`.

As for how this could be efficiently saved, one option is indeed in a full list in the gameInfo (or civInfo as they're civ specific?) as is done in this PR. Another option is to save them in the affected tiles, and remove/add these every time a great general is moved. Would incur a lot of adding/removing, but lookup is near instant, as it's just querying the tile the combat is on. Not sure if that offsets the adding/removing though.

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -157,9 +157,16 @@ object UnitAutomation {
...
                 return SpecificUnitAutomation.automateAddInCapital(unit)
-            
-            if (unit.hasUnique("Bonus for units in 2 tile radius 15%"))
-                return SpecificUnitAutomation.automateGreatGeneral(unit)
+
+            //todo this now supports "Great General"-like mod units not combining 'aura' and citadel
+            // abilities, but not additional capabilities if automation finds no use for those two
+            if (unit.hasGreatGeneralUnique && SpecificUnitAutomation.automateGreatGeneral(unit))
+                return
+            val isCitadelPlacer = unit.hasUnique(UniqueType.TakeOverTilesAroundWhenBuilt)
```
xlenstra, line 165: Wouldn't the improvement have that unique and not the unit?

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,141 @@
...
+            .filter { it.civInfo == civInfo && it.hasGreatGeneralUnique }
+        if (nearbyGenerals.none()) return 0
+
+        val greatGeneralModifier = nearbyGenerals
+            .flatMap { general ->
+                general.getMatchingUniques(UniqueType.GreatGeneralAura)
+                    .map { GeneralBonusData(general, it) } +
+                        general.getMatchingUniques(UniqueType.BonusForUnitsInRadius)
+                            .map { GeneralBonusData(general) }
+            }.filter {
```
xlenstra, line 45: Feel free to ignore this, but the staggering indentation to me seems to suggest more complicated behaviour than is really happening. Maybe
```
.flatMap { general ->
    general.getMatchingUniques(UniqueType.GreatGeneralAura)
        .map { GeneralBonusData(general, it) } +
    general.getMatchingUniques(UniqueType.BonusForUnitsInRadius)
          .map { GeneralBonusData(general) }
}.filter {
```
This way it's more clear that the two `getMatchingUniques` are at the same level and added together (I'd personally even put the + at the start to make that more clear, but that would indeed in some cases not work).

SomeTroglodyte:
> Haka War Dance

The relation didn't escape me. I preferred to not bloat this any more...

> Full list

No, "only" the upper bound of the radius to optimize the application in SpecificUnitAutomation _and_ BattleDamage. Storing in TileInfo - would have to be a Map<CivilizationInfo, Int> at least, and the more time-critical automation code wouldn't profit. Actual Battle damage calc is likely a lot less commmon than generals moving - my wild guess.

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,141 @@
...
+                // Support the border case when a mod unit has several
+                // GreatGeneralAura uniques (e.g. +50% as radius 1, +25% at radius 2, +5% at radius 3)
+                it.general.currentTile.aerialDistanceTo(unit.getTile()) <= it.radius
+                        && (it.filter == "All" || unit.matchesFilter(it.filter))
```
xlenstra, line 49: Shouldn't `it.filter == All` also be tested in the `unit.matchesFilter(it.filter)`? Or is this a surprisingly large optimization?

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,141 @@
...
+        val unitBonusRadius = generalBonusData.maxOfOrNull { it.radius }
+            ?: return null
+        return militaryUnitTilesInDistance
+            .maxByOrNull { unitTile ->
+                unitTile.getTilesInDistance(unitBonusRadius).sumOf { auraTile ->
+                    val militaryUnit = auraTile.militaryUnit
+                    if (militaryUnit == null || militaryUnit.civInfo != general.civInfo) 0
+                    else generalBonusData.firstOrNull {
+                        auraTile.aerialDistanceTo(unitTile) <= it.radius
+                                && (it.filter == "All" || militaryUnit.matchesFilter(it.filter))
```
xlenstra, line 95: Same here

core/src/com/unciv/logic/automation/UnitAutomation.kt
SomeTroglodyte: Good one. You're right, I don't remember testing how AI places citadels, focusing on modded boni of varying radii...

This code was a lot less flexible before, a modded unit able to place Citadels without the Strength bonus would never be usable to the AI, thus my changes - will need to re-think how to effectively cache that two-step test.

xlenstra (commented):
> I'll wait for more feedback - is the maximized flexibility I went for overkill and JR's approach may please more? If so, point moot.

Is it overkill? I'm not certain, it seems to work well and is definitely an improvement, while not feeling _overly_ complicated. I like it more than #6750 tbh. After the few comments I left are resolved, you can count this comment as an approval.

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
SomeTroglodyte: I think "All" is supported way down the call stack... BaseUnit.matchesFilter, and since the unmodded Great General should be efficient, yes, pure optimization. I remember leaving a comment - probably removed it thinking 'unnecessary'.

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
SomeTroglodyte: You think a comment is in order?

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,141 @@
...
+
+        // Get candidate units to 'follow', coarsely.
+        // The mapUnitFilter of the unique won't apply here but in the ranking of the "Aura" effectiveness.
+        val unitMaxMovement = general.getMaxMovement()
+        val militaryUnitTilesInDistance = general.movement.getDistanceToTiles().asSequence()
+            .map { it.key }
+            .filter { tile ->
+                val militaryUnit = tile.militaryUnit
+                militaryUnit != null && militaryUnit.civInfo == general.civInfo
+                        && (tile.civilianUnit == null || tile.civilianUnit == general)
```
JackRainy, line 80: What if the unit with such "Strength bonus in radius" unique will be not a civilian (as we got used for) but a military unit?

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,141 @@
...
+            civInfo.maxGeneralBonusRadius = civInfo.maxGeneralBonusRadiusBase
+            return
+        }
+
+        // Need to scan the units - might be faster looping getCivUnits only once, but far less readable
+        val maxPromotionGeneralBonusRadius = promotionsGrantingGeneralBonus.asSequence()
+            .filter { promotion ->
+                civInfo.getCivUnits().any { unit->
+                    promotion.name in unit.promotions.promotions
+                }
```
JackRainy, line 132: As far as I see "GG Aura" is a unique of the `Unit` type, not `Promotion` type. Why do we looking through the promotions here?

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -73,17 +70,11 @@ object BattleDamage {
...
-            val nearbyCivUnits = combatant.unit.getTile().getTilesInDistance(2)
-                .flatMap { it.getUnits() }.filter { it.civInfo == combatant.unit.civInfo }
-            if (nearbyCivUnits.any { it.hasUnique("Bonus for units in 2 tile radius 15%") }) {
-                val greatGeneralModifier =
-                    if (combatant.unit.civInfo.hasUnique(UniqueType.GreatGeneralProvidesDoubleCombatBonus)) 30 else 15
+                    modifiers["Missing resource"] = -25  //todo ModConstants
 
+            val greatGeneralModifier = GreatGeneralImplementation.getGreatGeneralBonus(combatant.unit)
+            if (greatGeneralModifier != 0)
                 modifiers["Great General"] = greatGeneralModifier
```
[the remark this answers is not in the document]
SomeTroglodyte: That's the text displayed on the battle table... I left it unchanged out of a fuzzy feeling that modifier lists are not directly comparable to units, and are sometimes aggregated, or queried with a const string. On second thought, only the first applies a bit, and the argument that the Stats Overview does the same.

I want that "Captain Spock" mod, though. My testing mod names them "Dà jiàng jūn"...

JackRainy:
> I'll wait for more feedback - is the maximized flexibility I went for overkill and JR's approach may please more? If so, point moot.

Generally, I believe it makes sense to have both mine #6750 and this one, since it adds a value of the additional GG Aura unique, while I address only the "bonus in radius" one.
And yes, some places (adding new properties to CivInfo) sound like an overkill for me. Too much attention (exceptions) as for just one particular unit.

SomeTroglodyte:
> I would rather remove the it.filter == "All" since this case is rare (taking into account the "Military" defaults for Vanilla and G&K)

Ah, now I see the a good catch - "All" was a lapse, I expected to put that in the json but then after some digging didn't. The default in GeneralBonusData for the legacy Unique and the short-circuit tests should of course be "Military".

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -123,10 +118,17 @@ class CivilizationInfo {
...
+
     @Transient
     var thingsToFocusOnForVictory = setOf<Victory.Focus>()
 
+    @Transient
+    /** Maximum radius of Great General or equivalent effect, from BaseUnit Uniques _only_. */
+    var maxGeneralBonusRadiusBase = 0
+    @Transient
+    /** Maximum radius of Great General or equivalent effect, from BaseUnit Uniques _or_ Promotion Uniques. 0 if none found!! */
+    var maxGeneralBonusRadius = 0
```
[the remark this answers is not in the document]
SomeTroglodyte: >  "just" for a speed optimization

Exactly. There's O(n²) or worse involved, so I decided to be careful.

> wording

:+1: 4 longer identifier names it is

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -342,6 +345,9 @@ class MapUnit {
         hasUniqueToBuildImprovements = hasUnique(UniqueType.BuildImprovements)
         canEnterForeignTerrain = hasUnique(UniqueType.CanEnterForeignTiles)
             || hasUnique(UniqueType.CanEnterForeignTilesButLosesReligiousStrength)
+
+        hasGreatGeneralUnique = hasUnique(UniqueType.GreatGeneralAura) ||
+                hasUnique(UniqueType.BonusForUnitsInRadius)
```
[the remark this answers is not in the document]
SomeTroglodyte: I explicitly aimed to support "GreatGeneralAura" from Promotions... And it works. I could imagine coding another mod-only feature to allow GG's participating in XP gained, but for now acquisition through buildings or natural wonders is an option.

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,141 @@
...
+                        general.getMatchingUniques(UniqueType.BonusForUnitsInRadius)
+                            .map { GeneralBonusData(general) }
+            }.filter {
+                // Support the border case when a mod unit has several
+                // GreatGeneralAura uniques (e.g. +50% as radius 1, +25% at radius 2, +5% at radius 3)
+                it.general.currentTile.aerialDistanceTo(unit.getTile()) <= it.radius
+                        && (it.filter == "All" || unit.matchesFilter(it.filter))
+            }.maxOfOrNull { it.bonus } ?: 0
+        if (unit.hasUnique(UniqueType.GreatGeneralProvidesDoubleCombatBonus, checkCivInfoUniques = true))
+            return greatGeneralModifier * 2
```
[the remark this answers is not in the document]
SomeTroglodyte: In a way you're right, that is a mod-unfriendly unique, used for the special of China. Implementation is unchanged. Should someone invent a "Great General" named unit with, say, only the Citadel placing unique and Haka War Dance - it won't apply. My thinking when I decided to support AI better when a unit doesn't have both the aura and the citadel ability was more inspired by Civ6 mod - it has the "Great General" without the Citadel ability. So I decided to leave the kludge as it is, to be revisited later.

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
SomeTroglodyte: Because `UniqueTarget.Promotion(Unit)` - they _are_ found and _do_ work from promotions. I didn't invest time in checking the implications of the "Aura" on UnitType - yea, `setMaxGreatGeneralBonusRadiusBase` should check them. Thanks!

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
SomeTroglodyte: I briefly thought about it, and am convinced such mods already exist. I dismissed the train of thought as battle damage will work, but automation wouldn't, as before. That isCivilian check early on... There's also that "AI wouldn't know what to do with your Archeologist" check in modchecker, and solving that would maybe become easier with a quite different structure - a little fuzzy, actually evaluating more alternatives and ranking them, that direction - not in this PR.

Until then, we have a guarantee that this code will only see civilians.

Even if - the check is to see whether the tile has "room" for the "general" assuming a general is always a civilian, and if it were not, it would still head toward that tile but stop short, a feature of UnitMovementAlgorithms.

SomeTroglodyte:
Summary of patches:
- Unique wording
- isCitadelPlacer correction
- GreatGeneralAura on _UniqueType_ now respected in radius caching
- Default mapUnitFilter now actually is an optimization
- Renames and linting
- The Chipmunk update seemed to have turned the auto-purge-imports option on without permission, didn't clean that up.

Actual Tests - give me a pause, later.

SomeTroglodyte:
Uh, it's bugged - but not this. #6712 broke normal improvement building except roads... Yup, no replacement for `tile.improvement = tile.improvementInProgress`. Darn, how could I overlook that!

SomeTroglodyte:
Merging both this and #6750 can be done, but I ended up accepting my version for all conflicts. This emphasizes where the differences have any relevance:
- Naming the Battletable modifier after the unit, arguments above, I'd follow a vote but vote a ~60% 'Nay' myself.
- Deciding the scope of GreatGeneralProvidesDoubleCombatBonus: Jack hangs it off the UniqueType.GreatPerson("War") _and_ StrengthBonusInRadius, while I have no additional gate, GreatGeneralAura is already checked. I'd say needs revisit when we have proper Great Admirals and find that the Chinese special should only apply to one of them - neither PR would do that.
- Jack also gates the AI automation by UniqueType.GreatPerson("War") - for the modcheck, the movement order and actual application of "move aura where it affects the most units" and citadel AI. I don't - I gate movement order and modcheck by the Aura unique, and AI by the individually relevant one. Which shows a weakness: A mod unit with only the citadel unique _will_ automate with this PR, but the modcheck will still complain.
- Jack gets away without caching a maximum range because at his point of applying the parameterized radius it's cheap and it works because the outer loop goes over combatant.getCivInfo().getCivUnits() - I try an outer scan by tile geometry _then_ check the units _then_ re-check radius as I find the uniques. This scales better as getCivUnits can grow quite large but a hex area with units tends to stay similarly sized - I hope. The penalty - a sensible limit for the outer loop is needed, with the convoluted way I dreamed up to cache that correctly even when the range changes over a game. I'd guess the average number of uniques scanned will stay lower with this PR despite aggregating to max and not taking a first (price for the ability to do e.g. +5% at distance 3, +15% at range 2 or +20% at range 1, and use the best even with several generals around).
- Lastly: Jack does compatibility replacement on the uniques in the Ruleset at game load, I use `@Deprecated`.. I may have a knot in my neurons, but that kind of compatibility fix is for breaking changes in game structures, not for replaced ruleset uniques, right?

JackRainy:
As far as I see the differences are:
This PR | My PR (#6750)
| --- | --- |
| **Pros:** | **Pros:** |
| * Besides the regular StrengthBonusInRadius the new GreatGeneralAura unique is added | * No additional logic, only basic existing mechanism of StrengthBonusInRadius |
| * The calculation does not depend on number of civ units | * The calculation is the same fast for different radius of aura since does not scan the tiles for units |
| * Citadel building AI logic is some kind of isolated | * Citadel building AI is the same as it was before linked to Great General |
| * Having several units with different bonuses the maximum will be taken | * The name of unit with StrengthBonusInRadius is shown to distinct which of those contributed to the battle stats |
| * The old deprecated unique is still available | * No need to keep the old unique since it will be automatically replaced with the new one in the rules |
| **Cons:** | **Cons:** |
| * Additional logic with external variables in the CivilizationInfo and their maintenance which increase the number of "states" and code "links" | * No special handling of Citadel building AI |
| * The algorithm does checks for 2 uniques of similar nature | * No additional unique to be customized in mods |
| * Having large radius of aura will decrease performance | * The performance decreases linearly with number of units in civ |
| * Several different units (e.g. Khan and Superman) with the aura will be displayed to user equally as "Great General" but effect will be taken from one of them | * In case of several units with the aura the bonus effect will be taken from one of the units, maybe, not the best one |
| * GreatGeneralProvidesDoubleCombatBonus will be triggered on other units with auras but not Great Generals | * The replacement of deprecated unique needs to be cleaned up anyway at some point |

I assume we need the performance tests on regular conditions and on large areas of aura with large number of civ units - to see whether the performance flaws are that much critical.
Besides this, I believe we can take the best from two worlds and merge two approaches in one.

SomeTroglodyte:
I can support Jack's analysis, feels like I said more or less the same.

Except maybe one detail: Merging both - please try it, you can do it locally, won't take that long. Only ~8 files to merge, less than 100 conflicts. You'll see - you end up deciding between the two approaches, getting either J with some S linting or S with some J linting...

JackRainy:
Challenge accepted! 💪🏻 
Actually, I have already merged them... and it took a bit more time than to find out a code-phrase for the Debug tab in Options by a given hash sum from the code :)
Now I need to clean up and optimise that a bit... something great is coming soon!

PR #6752, opened 2022-05-11
Update  Simplified_Chines…

xk730:
…e.properties to 人造人/资产/jsons/翻译/Simplified_Chinese.properties

修正了一部分翻译错误


heipizhu4 (changes requested):
Hi there, newcomer! Just a few to change:
1. Change the title, it's too long. You don't have to describe where the .propeties is. We all know
2. Change the decription. Use EN or EN+CH. It'll be better for others to understand.
3. You have put the file in the wrong dictionary.(wondering how you made it)

PR #6753, opened 2022-05-11

xk730:
Correct some translation errors


heipizhu4 (approved):
Becoming better after pull another request. Keep it up!
PS: I may translate less until July. Annoying standarlized tests.

yairm210:
Version rollout PR created

PR #6755, opened 2022-05-11
Proposal to add the current HexaRealm tileset as the built-in tileset.

GeneralWadaling:
Greetings, Yairm!

Although it is kind of a bit early to do that, here's my proposal for adding the HexaRealm tileset into the game.
All sprites are roughly 282 kb.

Currently just the tile sprites. There are concerns about mod compatibility, so besides the tile sprites it has no other sprites, the other sprites would be fetched from FantasyHex sprite files of base game and enabled mods.


GeneralWadaling:
Ahh shit, forgot to remove something, I'll come back later.

GeneralWadaling:
Whoops, I've forgotten to remove something. Will come back with another pull request.

PR #6756, opened 2022-05-11
Proposal to add the current HexaRealm tileset as the built-in tileset. (for serious this time.)

GeneralWadaling:
Greetings, Yairm!

Although it is kind of a bit early to do that, here's my proposal for adding the HexaRealm tileset into the game.
All sprites are roughly 282 kb.

Currently just the tile sprites. There are concerns from modders and players about mod compatibility, so besides the tile sprites it has no other sprites, the other sprites would be fetched from FantasyHex sprite files of base game and enabled mods.


yairm210:
I'm in favor :)
We'll need to see if this can all fit together in the 2048-px cube, but if not then we'll separate stuff

SomeTroglodyte:
Screens? ... or maybe I'll do some myself later. Then I could check atlas distribution as well.

SomeTroglodyte:
Reminder: Auto atlas grouping need not go by folder, having the same folder in different Image.xxx and sprinkling stuff over them is supported. If there's a consensus this is prettier -> easy, pack away all FantasyHex stuff not needed for fallback into Images.FantasyHex + flip the default in `GameSettings` + `credits.md`

yairm210:
When deciding what to group, the general direction is to minimize the largest source of user latency from image swapping, namely - [texture swaps while map rendering.](https://yairm210.github.io/Unciv/Developers/Map-rendering/)

Looking at the main game png, I see a couple of things that can easily be left out.
- Policies - they aren't rendered in the map at any point, easy win
- Religions & Nation icons - these *are* rendered in map, but only in the city buttons... which are a huge other mess, there are already at least 3 texture swaps per city button...I *think* that buttons that are out of sight are not called on to render, I hope I'm right

That should clear up a good quarter of the space in the game.png, I will run before/after tests to see if and how this affects rendering

yairm210:
Before - CityButton draw = 541ms / 4.47s, GLClear (basically 'wait for next frame') is 1.15/4.47
image


After: CityButton draw = 679/5.3s, GLclear = 1.44/5.3s.
image

Summary: Looks extractable :) And now we have about half the 2048 square clear :D

SomeTroglodyte:
Since the atlas itself wasn't part of this PR, you (YM😉) have now all the time in the world to tweak and merge those.
But don't forget that attribution in the credits is required..

PR #6759, opened 2022-05-11
see autofix attempts on real time when inputting multiplayerServer URL

touhidurrr:
I didn't know that `multiplayerServerTextField.text` was mutable. Then I saw this line and realized that it was....
https://github.com/yairm210/Unciv/blob/5a78c594053f89f8e8b86a8567d3952eec03b7b9/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt#L281

So, I think
https://github.com/yairm210/Unciv/blob/92d5b45e1bf08f347b350f37fd0e3763106d9bc5/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt#L284-L285
instead of just
https://github.com/yairm210/Unciv/blob/5a78c594053f89f8e8b86a8567d3952eec03b7b9/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt#L284
is a better way to do this cause players can see the autofixing in real time

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
+++ b/core/src/com/unciv/ui/worldscreen/mainmenu/OptionsPopup.kt
@@ -281,7 +281,8 @@ class OptionsPopup(
             multiplayerServerTextField.text = Gdx.app.clipboard.contents
         }).row()
         multiplayerServerTextField.onChange {
-            settings.multiplayerServer = formatMultiplayerUrlInput(multiplayerServerTextField.text)
+            multiplayerServerTextField.text = formatMultiplayerUrlInput(multiplayerServerTextField.text)
+            settings.multiplayerServer = multiplayerServerTextField.text
             settings.save()
             connectionToServerButton.isEnabled = multiplayerServerTextField.text != Constants.dropboxMultiplayerServer
         }


touhidurrr:
> Since I don't use the setting I don't care.
Copying and pasting is still possible, but typing is not - of course (the selection is reset). I'd suggest testing before pushing.

Lol.

touhidurrr:
wait @SomeTroglodyte, it seems like the pointer resets to me. Also typing the initial http:// is troublesome. Maybe we can fix that if we know how to preserve the pointer location.
Also autocomplete also looks possible. We can autocomplete the initial `http` if doesn't exists from beginning.

touhidurrr:
ok, so we have TextField.cursor, can anyone tell me if the initial value is 0 or 1, with an empty TextField

SomeTroglodyte:
> seems like the pointer resets

Other words for what I keep saying

> 0 or 1

Are you coding in Studio? Ctrl-B is your friend, goes to the _source_, e.g. from having a .cursor call marked; and from a little poking around you can quickly see it's 0-based. See `setSelection` - and for perfection you should also consider using selectionStart, which is <> cursor **if** a few characters are selected.

Again: onChange and using text, cursor, selectionStart is a high goal (kudos if you perfect that). A cheap char filter and validation outside onChange is safer and easier. I wouldn't be able to get either to work at all without using Android Studio, debug breakpoints, and Alt-F8 evaluation to actually _see_ what's happening.

PR #6770, opened 2022-05-11
Fixed a bug where roads seemingly removed improvements

xlenstra:


SomeTroglodyte (approved):
Regression from #6712, straightforward fix. :+1:

PR #6771, opened 2022-05-11
Resolved #6767 - AI will not declare war if it definitely can't take a city

### LINKED ISSUE #6767: Add condition for AI to declare war.
I'm sorry for the title. So:

The issue:

Currently, I'm sure of it, the only reason AI declares war on someone is just because they have a far superior army and that's it.
The problem; This leads to situations of:
- All or most civs ganging up to a single civ
- AI declaring war very early on without the proper army (they often just have 6 warriors and 3 archers and think thats enough to conquer you)
- AI making war even with very high relationship (I don't know if this point is still there, as I never get declared war on).

An example has been shown on Discord by the user @Dzar#9770 [over here.](https://discord.com/channels/586194543280390151/586194543716859916/973663905731735684), where he wrote down all wars in a game.
This image just proves points 1 and 2.imagen
Here we see that Cartage is being attacked by the whole world, why? Because they don't have an army; and we see the army of 4 nations, which are a bunch of warriors that ain't going to do anything.

And I do think it's a problem, because it makes AI so predicable, that once you learn to keep up your military, you never have to worry about getting attacked, which is not that fun.

The solution [I see]:

To add the condition so War can only be declared if relation with the enemy is 0 or below.
This comes with various birds down with one bullet, as it allows for:
- Denounce action becomes actually useful
- Doing good diplomacy becomes important
- Avoids absurd situation where your "friend" attacks you

This would only be a cap for AI, but it could also be added to players but an exception would have to be placed if we add the limit to human players as well, which would be for player be able to declare war if a Protected CS is being attacked by other, and to possibility for player to declare war when relationship is still neutral, so warmongering effect can take place.
What about the diplomatic debuff of "Your so called friendship is worth nothing"?
It would trigger when you attack a civ after having recently declare friendship with it. (with nukes, denouncing or broken promises)
But I think is better to leave the condition only for AI.

Another thing that *could* be done is to add a sort of check list of units to have for AI to declare war. Bt maybe that's to much, so probably not, I just leave it on top of the table as an option.

I'm no coder so this is just a wish letter :)

On the topic, denouncing currently doesn't remove the warmongering effect, as it suposedly does in Civ V.

yairm210:
Also simplified 'find city to attack' with existing function

### DIFF
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -707,7 +707,7 @@ object NextTurnAutomation {
         val ourCity = closestCities.city1
         val theirCity = closestCities.city2
         
-        if (civInfo.getCivUnits().none {
+        if (civInfo.getCivUnits().filter { it.isMilitary() }.none {
                 val damageRecievedWhenAttacking = 
                     BattleDamage.calculateDamageToAttacker(
                         MapUnitCombatant(it),
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -1,6 +1,9 @@
 package com.unciv.logic.automation
 
 import com.unciv.Constants
+import com.unciv.logic.battle.BattleDamage
+import com.unciv.logic.battle.CityCombatant
+import com.unciv.logic.battle.MapUnitCombatant
 import com.unciv.logic.city.CityInfo
 import com.unciv.logic.city.INonPerpetualConstruction
 import com.unciv.logic.city.PerpetualConstruction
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -687,17 +690,33 @@ object NextTurnAutomation {
...
-        if(otherCiv.isCityState() and otherCiv.getProtectorCivs().isNotEmpty()) {
+        if (otherCiv.isCityState() and otherCiv.getProtectorCivs().isNotEmpty()) {
             theirCombatStrength += otherCiv.getProtectorCivs().sumOf{it.getStatForRanking(RankingType.Force)}
         }
 
         if (theirCombatStrength > ourCombatStrength) return 0
 
+        val closestCities = getClosestCities(civInfo, otherCiv)
+        val ourCity = closestCities.city1
+        val theirCity = closestCities.city2
```
SomeTroglodyte, line 708: `val (ourCity, theirCity) = getClosestCities(civInfo, otherCiv)` if you want - data class supports deconstruction

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -260,7 +260,6 @@ object BattleDamage {
 
     fun calculateDamageToAttacker(
         attacker: ICombatant,
-        tileToAttackFrom: TileInfo?,
```
SomeTroglodyte, line 263: :+1: How many times did I read that warning...

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -687,17 +690,33 @@ object NextTurnAutomation {
...
+        
+        if (civInfo.getCivUnits().filter { it.isMilitary() }.none {
+                val damageRecievedWhenAttacking = 
+                    BattleDamage.calculateDamageToAttacker(
+                        MapUnitCombatant(it),
+                        CityCombatant(theirCity)
+                    )
+                damageRecievedWhenAttacking < 100
+            })
+                return 0 // You don't have any units that can attack this city without dying, don't declare war.
```
SomeTroglodyte, line 718: Ranged units with range 3 could... But then if they fail this test their damage inflicted would likely be negligible too.

SomeTroglodyte (approved):
Looks like nothing could go wrong

core/src/com/unciv/logic/automation/NextTurnAutomation.kt
yairm210: Their damage to attacker would be 0, they would pass
To fail this you would have to have no ranged units at all, which does happen

PR #6766, opened 2022-05-11
Update Simplified_Chinese.properties

heipizhu4:
Add some new lines.

PR #6775, opened 2022-05-12

xk730:
Correct language defects and ambiguities and fill in gaps


heipizhu4:
Already merged to #6766 , so that translations from 2 contributers can be merged at the same time.

PR #6783, opened 2022-05-12
Use ranged strength when defending against ranged attacks

OptimizedForDensity:
Ranged units (specifically, non-city units with ranged strength > 0 that aren't embarked) will use their ranged strength in defensive combat calculations when they are attacked by a ranged unit. This should have an effect on the survivability of ranged units including siege. Air units have the same strength and ranged strength so while they technically use their ranged strength as defense now, they aren't affected in game by the change. Checks off a box in #4697.

The ```BattleTable``` UI has been updated to show the correct defensive strength. To help the player understand which strength value is being used in defense, I've also modified the UI to show the ranged strength icon (a crosshair) when a unit is using ranged strength on defense. Cities and embarked ranged units will still use a shield icon. Ranged attackers will also have a ranged strength icon next to their attack strength. Below shows a side-by-side comparison of the interface changes. A simple ship style icon could be used to indicate embarked strength but I'll leave that change for the more artistically-inclined.

Interface changes

Behind the scenes, the ```ICombatant``` interface is modified so that ```getDefendingStrength()``` now takes a boolean ```attackedByRanged``` which has a default value of ```false```. The inheriting ```MapUnitCombatant``` class will then use this parameter to decide if it should return regular strength or ranged strength (easily accomplished by calling ```defender.getDefendingStrength(attacker.isRanged())``` for example). ```CityCombatant``` does not use this parameter at all since it uses a different strength calculation so it will always return the same strength as it did before this change.

Ranged units will still use their normal strength for capture calculations
https://github.com/yairm210/Unciv/blob/3ea9c6503ba86a4ef1efff32f17182a93023cb4e/core/src/com/unciv/logic/battle/Battle.kt#L207
which I believe is what civ 5 used for their calculations.
https://github.com/Gedemon/Civ5-DLL/blob/aa29e80751f541ae04858b6d2a2c7dcca454201e/CvGameCoreDLL_Expansion1/CvUnit.cpp#L3789


xlenstra (approved):

SomeTroglodyte (approved):
Extra points for exemplary PR documentation. :100: 
[`attackedByRange:Boolean`](https://kotlinlang.org/docs/coding-conventions.html#7dceef89)
So a city's bombard isn't ranged? :bow_and_arrow:  Interesting.

Issue #6785, opened 2022-05-12
Feature request: Embark Strength Listed in Civilopedia


OptimizedForDensity:
It would be nice if the embarkation civilopedia entry listed the embark strength for each era since otherwise you would need to look into the json files to see what its values are.

SomeTroglodyte:
Would be technically easier if we already had an Eras category. Adding one... Straightforward, and an Era entry has quite a lot of data by now, though most refers to starting in that Era, which I wouldn't necessarily want to display. name, color, researchAgreementCost, 
 baseUnitBuyCost, embarkDefense, CS boni, uniques... Would that look good? Or does the category buttons scroller get too crowded?

OptimizedForDensity:
That's probably a better idea to list all the era-dependent information instead of just embark strength, and it could probably go next to the difficulties category that's already in the civilopedia.

SomeTroglodyte:
<details><summary>How about...</summary>

![image](https://user-images.githubusercontent.com/63000004/168158709-3e8e397a-3085-4d75-b9cc-b7d61bc64546.png)
</details>

OptimizedForDensity:
It looks good!

SomeTroglodyte:
> [OptimizedForDensity](https://github.com/OptimizedForDensity) commented [23 minutes ago](https://github.com/yairm210/Unciv/issues/6785#issuecomment-1125456068)
> [SomeTroglodyte](https://github.com/SomeTroglodyte) commented [22 minutes ago](https://github.com/yairm210/Unciv/issues/6785#issuecomment-1125456777)

Look how fast I coded that :racehorse: :rocket: :mountain_biking_man:

PR #6788, opened 2022-05-12
Pedia from mainmenu

### LINKED ISSUE #5432: Add possibility to read Civilopedia from the main menu
It would be great if it will be possible to open and read Civilopedia from the main menu. At the moment it is possible only after loading or starting any game.

Add the little round button to the lower right corner of the main menu with big "C". Or just little standard button with "Civilopedia" text.

SomeTroglodyte:
I allowed Civilopedia access from Map Editor :see_no_evil:  - and there were a few silly crashes due to uncouth expectations on context, like UncivGame.Current.worldScreen being accessible...

<details><summary>Screenies...</summary>

image
image
</details>

Linking button broken - so let's try "Closes #5432"


SomeTroglodyte:
Sorry, the second half was meant to be a separate PR, forgot a step. Rollback and separate or take as is?

Second half adds Eras category, inspired by #6785 , with tech list and objects linked via `OnlyAvailableWhen` or `PolicyBranch.era`. Screenie in the issue is an early version missing the Policies, but you get the idea. Me making git mistakes means me go pause oogah oogah :orangutan: 

Rex icon for "Eras" header may not please, open for suggestions (I wanted the Rex _skeleton_, too cute, but it was too filigrane - could go for one of the excellent Rex skulls or even a simple "Calendar" symbol).

OptimizedForDensity:
The hour glass icon might work since it symbolizes time and is already in the game.

SomeTroglodyte:
That I used only in testing. Doesn't hold up to say the Wonders or Religions icons i thought - though a lot of the others could use ideas. Simpler more stylized - guide the user without distracting /vs/ more unique and decorative - good for headers? Something like that. Also - running sand = short, and I wanted looooooooong time. Thus the paleontology idea. My favourite of all is the one I gimped for map editor, the terrain one.

android/assets/jsons/translations/template.properties
```diff
@@ -1254,8 +1254,13 @@ Granted by: =
...
 Difficulty levels = 
 The possible rewards are: = 
+Eras = 
+Embarked strength: [amount]† = 
+Base unit buy cost: [amount]¤ = 
+Research agreement cost: [amount]¤ = 
 
-# Policies
+
+S# Policies
```
xlenstra, line 1263: S?

xlenstra (approved):

core/src/com/unciv/MainMenuScreen.kt
```diff
@@ -222,6 +238,13 @@ class MainMenuScreen: BaseScreen() {
         }
     }
 
+    private fun openCivilopedia() {
+        val ruleset =RulesetCache[game.settings.lastGameSetup?.gameParameters?.baseRuleset]
+            ?: RulesetCache[BaseRuleset.Civ_V_GnK.fullName]
+            ?: return
```
xlenstra, line 244: An alternative could be to open a mod selection screen, where you can select a base ruleset mod and extension mods.

yairm210:
Looks OK to me, can be improved in the future but let's get the base out now :)

core/src/com/unciv/MainMenuScreen.kt
SomeTroglodyte: As was suggested in one of the issues. Honestly, much to do and I think players will get used to it quickly and everybody has their favourite very likely selected in there. Ruleset selection - may be _in_ pedia, hidden when in-game. More important would be to check the pedia map editor loads shows the correct ruleset - don't remember what it does :grimacing:

android/assets/jsons/translations/template.properties
SomeTroglodyte: Oops, shrapnel typo. Good thing it's neutral - `#` is not the thing marking a comment line in translations, lack of `=` is.

PR #6793, opened 2022-05-13
Enable gradle caching for faster CI runs

touhidurrr:
Notes:
1. distribution is mandatory in v3
2. zulu is used because it was recommended in [wiki](https://yairm210.github.io/Unciv/Developers/Building-Locally/#without-android-studio)

### DIFF
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -30,9 +30,11 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v3
         with:
+          distribution: 'zulu'
           java-version: '11'
+          cache: 'gradle'
 
       - name: Compile kotlin and build classes
         run: ./gradlew classes
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -32,7 +32,7 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-          distribution: 'zulu'
+          distribution: 'temurin'
           java-version: '11'
           cache: 'gradle'
 


touhidurrr:
How how much does it affer CI runs? I think you can guess from this.
Actions History:
actions-history
Before Caching was Enabled, Compiling Classes of Unciv:
Before
After Caching was Enabled, Compiling Classes of Unciv:
After

touhidurrr:
> You sure this merits 'faster'?
> - succeeded 3 hours ago in 5m 43s [(last build in your fork)](https://github.com/touhidurrr/Unciv/runs/6423285778?check_suite_focus=true)
> - succeeded 6 minutes ago in 6m 17s [(check for this PR)](https://github.com/yairm210/Unciv/runs/6425632423?check_suite_focus=true)
> Which wiki recommends zulu and where? Ours? Entry may be outdated. Github's? As of today I'd trust azul about 666% less than adoptium, judging from the burden the site places on adblockers.

About your first question, see this: https://github.com/touhidurrr/buildUnciv/actions
And about the second, this: https://yairm210.github.io/Unciv/Developers/Building-Locally/#without-android-studio

SomeTroglodyte:
That's better. Easier to see what you mean. 

And the check for this PR wasn't faster because you turn it on for build, but not where the check is run? Don't remember the yml too well... Nah, it's two lines down same job. Curious.

touhidurrr:
> And the check for this PR wasn't faster because you turn it on for build, but not where the check is run? Don't remember the yml too well... Nah, it's two lines down same job. Curious.

Basically my action clones touhidurr/Unciv repo and do build and apk from it.
After enabling caching the first run would obviously not have access to the cahed files as no file had been cached yet.
The second and subseqent runs after enabling caching, the runs would have access to the cached files and thus we can see faster build times.
The improvement for building apk, jar & packr seems like around 66% faster.
For building kotlin & classes its like 95%.

So, why the huge improvement? Actually dependencies comes in a nested order. Each dependency might have some other dependencies, and those dependecies might have some other dependencies. So, in short we have to download thousands of files from 3rd party repos which takes a while. With caching enabled those files are not even requested. Github detects and caches gradle in its own server and restores caches from there and so we are downloading transferring those files over not from a 3rd party server but githubs own internal server and tranfering files among internal servers over a same local network is simply much much faster. From my tests it's like 1Gbps.
And thats why the improvement on runtime.

touhidurrr:
Also there is saving CPU times as gradle doesn't need to calculate from scratch the dependency graph and filehashes.
So, basically as soon as the cache is restored, gradle can start to compile things immediately.

SomeTroglodyte:
Each gh action boots up a docker image... So, that caching happens outside the VM, either via snapshots or - like you say locally networked shortcuts of any kind. I trust you so it's the latter. Which in turn means kicking the check run in _this_ PR has chances to get faster... kicked.

SomeTroglodyte (approved):
Can see the effect easily in my fork too, so whether zulu adopt temurin liberica microsoft or corretto - I vote 'in'

SomeTroglodyte:
For reference: [setup-java documentation is here](https://github.com/actions/setup-java#setup-java)

touhidurrr:
> Still, I'd go with distribution: 'temurin'

I prefer temurin too. I wrote zulu casue it linked in wiki. Should I change it?

SomeTroglodyte:
> Should I change it

With 2 : 0.1 votes... That zulu reference is old enough to devalue it by an order of magnitude :partying_face: . Would you propose a change for that doc as well please? I'm not convinced about the direction Adoptium as a company seems to be taking, but there's a noticeable trend to temurin when snooping around. I just don't have any evidence handy. Except their site uses only a single almost-harmless almost-tracker (googletagmanager), while azul is heavy with them.

touhidurrr:
What about the jre we use in packr?
Should we change them too?
https://github.com/yairm210/Unciv/blob/1ca1342816db2a0db4596a2606e8315d55160948/.github/workflows/buildAndDeploy.yml#L143

touhidurrr:
Personally I prefer pinging Github Api and check foe latest rather than hardcoded links.

SomeTroglodyte:
> licence is required

Not sure about the details and implications. If you want, do the research. I only know why nobody in their right mind is still using Oracle's... OK, I _do_ know of boxes still running their 8/201, decades old... For packr'ed bundles, all JRE's used here are outdated as far as I remember. I'd stay with an older level we know works, definitely LTE, but keep if to the latest patch - so the one you showed could go to 11.0.15+10 - but maintaining that is no fun so I never bothered.

touhidurrr:
Lets just open some pr for everything and see what others suggest.
Also I will use Temurin for this pr.

touhidurrr:
@SomeTroglodyte and so this pr https://github.com/yairm210/Unciv/pull/6798
But the problem is this:
https://github.com/touhidurrr/buildUnciv/runs/6430297670?check_suite_focus=true

PR #6795, opened 2022-05-13
Minor Map Editor improvements

### LINKED ISSUE #6610: Map editor scroll issue
OS: Android 12

Device: SM-G970N (2280x1080 screen)

ver: 4.0.9, APK from github

video: https://youtu.be/PNMUO2ns8mk

Description: If i select custom size on map editor, expansion leads to scroll bug.

**To Reproduce**
1. Open map editor.
2. Click 'Generate'.
3. Select 'Custom' on World Size.
4. See error.

**Expected behavior**
- If you select custom size, side menu expands, and scroll doesn't work. Thus you can't click 'Create' button.
- If you wanna generate custom-sized map, you must scroll down before choosing 'Custom'.

**Additional context**
Portrait orientation is useless because... 🤦

<details>
<summary>(See more)</summary>

Screenshot_20220426-093510_UnCiv
</details>

### LINKED ISSUE #6694: Map editor suggestion
Some suggestions about map editor.

SomeTroglodyte:
Closes #6610 , closes #6694 (partial improvement, rest is in the editor's source to-do list)

For 6610, the wrong ScrollPane 'taking' geometry changes when the MapParameter widgets change height - I couldn't get Gdx's methods to deal with relayouts to work, so I applied the lazy solution to simply take the inner tab content out of its parent and put it back in immediately.

For 6694, the Map Editor resource painting _and_ the old MapGenerator strategic resource placer, which is by now AFAIK _only_ used by the editor, will respect the resource abundancy setting and do major deposits randomly. From what I understand of the MapRegions code, editor won't be able to match the patterns of regions anyway, but I try to match the overall likelihood of major deposits.

To begin with, I had to """fix""" the tileResource object lazy - as commented elsewhere this left inconsistent tiles. From editor obviously, as it will trigger the lazy and still update the resource field in place. Other sources - I forgot them, marginal to nonexistent, but of course future code could have introduced more. Just try to make `resource` a `private set`... Took the verbose classical approach, will fix all possible sources of updates to `resource` except through reflection, and deserialization is fine anyway.


xlenstra (approved):

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -911,22 +928,29 @@ open class TileInfo {
...
+    private fun approximateMajorDepositDistribution(): Double {
+        // We can't replicate the MapRegions resource distributor, so let's try to get
+        // a close probability of major deposits per tile
+        return getAllTerrains()
+            .flatMap {
+                it.getMatchingUniques(UniqueType.MajorStrategicFrequency)
+            }.sumOf {
+                // The unique param is literally "every N tiles", so to get a probability p=1/f
+                // The null hoops are so we can sum without exceptions no matter what a mod may put there 
+                it.params[0].toIntOrNull()?.run { if (this <= 0) null else 1.0 / this }
```
yairm210, line 949: Too clever by half, i'd much prefer it to be readable to be kotliners.

yairm210:
Also conflicts

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: Heh. 'kotliners' rhymes with 'flatliners'. First one was the better one. :+1: 
I thought hard on how to aggregate probabilities when you get frequencies, and gave up and left what worked. I'll reboot that thought.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -69,6 +70,10 @@ open class TileInfo {
 
     var naturalWonder: String? = null
     var resource: String? = null
+        set(value) {
+            tileResourceCache = null
```
yairm210, line 74: I'm usually against side effects of setters but in this case this is an elegant solution to update caches

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: ... much better. Though flatMap now exceeds polite line length a bit - marginally preferred this to flat~lining~mapping via double loop: less indent and the sequence initialization overhead is already there. The alternate would have been a val for the sequence, and that was 51:49 nay...

yairm210 (approved):

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: Academics have a name for the pattern, right?

xlenstra (approved):

PR #6797, opened 2022-05-13
Possibility to liberate a traded city

### LINKED ISSUE #6722: Option to grant independence to former city-state acquired through trade agreement
I finally gained the upper hand in my 150-turn war, and used the opportunity to extract some territorial concessions from my foe. In this case, I asked for control of a city-state that he had captured, and he agreed. My intent was to "turn it loose" - that is, grant its independence, however it went straight to my control without giving me the option of what to do with it.

My memory has proven fallible numerous times, but I believe civ5 allowed me the option of "turning a city-state loose" that I had gained control of by way of a trade agreement. I'm also fairly certain that had I conquered this former city-state by force, I would have had this option, so it seems reasonable to allow this same option with trade agreements.

Screenshot_20220507-183132_UnCiv

JackRainy:
Fixes #6722

When you receive a city by trading you get the options to either liberate it or keep to yourself.

Screenshot 2022-05-14 00-20-28


xlenstra (approved):

yairm210:
Conflicts

PR #6798, opened 2022-05-13
Detect and Use Latest Temurin 11 builds for Packr

touhidurrr:
This PR uses Github API to detect latest Temurin 11 JRE builds and fetch them to be used with packr
Latest Releases: https://github.com/adoptium/temurin11-binaries/releases/latest
API URL: https://api.github.com/repos/adoptium/temurin11-binaries/releases/latest

### DIFF
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -150,8 +150,8 @@ jobs:
           ./butler push deploy/Unciv-Linux64.zip yairm210/unciv:Linux64 --userversion ${{steps.tag.outputs.tag}}
           ./gradlew desktop:zipLinuxFilesForJar
 
-          wget -q -O jdk-windows-32.zip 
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
> [But the problem is this:](https://github.com/touhidurrr/buildUnciv/runs/6430297670?check_suite_focus=true)

See https://github.com/libgdx/packr/issues/141 and the linked # 129

There were reasons that first wget is pulling an older packr, and that seems too dumb to deal with Java > 8 simply for prejudices on directory structure, and - well, sorry, my happiness demands from me not to invest many braincells on this. Maybe I left some notes in my scratchpad on what I learned months about packr, probably not.

touhidurrr:
Wait. Are we really using libgdx/packr or anuken/packr? what is this line for?
https://github.com/yairm210/Unciv/blob/8b9946e65e3356c7afa568066002f14494fcb35e/build.gradle.kts#L31

touhidurrr:
I tried building latest packr from source and using it but it didn't work
https://github.com/touhidurrr/packr/releases/tag/1653143085084336432

PR #6801, opened 2022-05-13
Use kotlin coroutines instead of raw threads (+ refactorings)

Azzurite:
So for that [one new feature]™ that I'm currently developing I did all these refactorings & kotlin coroutine addition, but yeah, it makes major sense to split that up.

So here are a bunch of refactorings/smaller improvements and the main thing: inclusion of kotlin coroutines within the project, and switching all usages of raw threads to coroutines which run on a cached thread pool (and coroutines also reuse threads anyway). This alone should vastly reduce our number of threads and thus likely increase performance.

When reviewing this, I would strongly suggest to do it commit-by-commit. I have neatly separated each refactoring/improvement into its own commit and the commit messages should be self-explanatory. I hope each commit compiles by itself, but I might have made mistakes, I split this into commits post-partum.

I tested pretty much all aspects of the game on desktop & android, should all be fine.

Here are some additional comments:

* cf293e835e972e7d27045ac4d67a7a51cb42b9bd: This is in preparation for my [one new feature]™ which is basically extracting all multiplayer logic into its own classes instead of it being spread around various `Screen`s
* fe30b6b0b76aecf634a8e16f74fc30b1421af8de: We do have `FileNotFound` exception handling in the multiplayer logic, it just makes sense to use 404's in the custom server implementation.
* 7ba4f69acd4749455ce9eef30568e148a571552a: There were a lot of instances of `GameSaver.saveGame` that did not add a `saveCompletionCallback`. These swallow all exceptions that happen on saving the game, which are some of the most important exceptions to handle, or at least get reports about from users. So I simply made the default handler throw the exception.
* 0be0b239ba74b45897ee1ac3a506b09beec9aa25 and 39b6c754d7d6278036401aac04a2fb2e64347b55: This is unused but will be used in my [one new feature]™
* bd015ec36259c72db1fb43fb274489f6e0369068: did you know that I hate unnecessarily large files?
* 43afe6cd49c2915bfcdf8f758782820d95c6c961: did you know that I hate violations of the single responsibility principle?
* 719809a9e315b688e43e0800248e788f8040a4ea: this was one instance where a game was loaded from multiplayer on the main thread - unnecessary.


touhidurrr:
Just interested, what is this `coroutines` stuff?

touhidurrr:
Also why do you do
```kt
stuff = newStuff {
  // .. moreStuff
  return@newStuff
}
```
in place of
```kt
stuff {
   // .. moreStuff
   return
}
```
does this change anything?

touhidurrr:
So, basically you believe that coroutines are a more lightweight form of handling parallel  processing and than Java Threads and implementing those for better performance? But since that would take explaining first that these would actually optimize stuff, you are shipping these with your **[one new feature]™**?
So, I would assume that your **[one new feature]™** would be something that can take advantage of coroutines massively. Interested to see what it is.

touhidurrr:
> So, basically you believe that coroutines are a more lightweight form of handling parallel processing and than Java Threads and implementing those for better performance?

Also pardon the sarcasm but coroutines looks like `async-await` to me. Because this:
```kt
fun stuff() = runBlocking {
    launch { someStuff() }
    launch { someMoreStuff() }
}
```
looks a lot like
```js
async function stuff() {
    await someStuff();
    await someMoreStuff();
}
```
Except based on docs, they are trickier and gives you more control over how your processor would actually handle them.
Also everything above is **[My Own Take on Coroutines]™**.

touhidurrr:
Also since I am new to kotlin it would be if you had some docs about `makeNewErrorScope("Stuff") {}` syntax for me.
This looks cool.

SomeTroglodyte:
> unless you pay me for my time

:rofl:  :+1: :+1: 

> basically you believe

Coroutines are _not_ a Religious matter, they're **_fact_**.  And they're _the_ way to get Unciv's CPU load to >16% where it's currently stopping even on the heaviest huge-map late-game next-turn. If only I had some ideas how to de-couple e.g. the units so one could automate without knowing whether Schrödinger's cat is alive or dead...

touhidurrr:
> unless you pay me for my time

🤣 👍 👍

Sigh. That's why I sometime wished for Github to have a realtime chat feature to that we could have such silly chats. Forums are not that good for chatting. And I was bored that time reading docs.
The docs was a little funny, first the author told that we don't have the async await stuff that you are looking for and instead we have coroutines. Then trying to convince how cool they were. Then ending with telling that they are lightweight comparing them with Threads as like the coolness didn't feel enough for them.
And so, I was wondering if turning Threads into coroutines was a common practice in Kotlin or not since azzurite just casually named the pr using coroutines while you ask me how much caching gradle would affect CI runs when I pr it. Sigh.

SomeTroglodyte:
So you didn't dare tackle the MusicController. I know it's intimidating, but the threading elements are really only a way to update volume repeatedly (fade) or leave a pause between tracks. And the latest complication is simpler than it looks too - lwjgl3 became less thread-safe than 2 _on desktop only_ so I hijacked the Gdx app loop instead of postRunnable everything. On desktop the only Thread is a recovery delay after a bad crash; but on Android that "fade" ticker is still a Timer. Come to think of it, `launchCrashHandling("Music")` - should be "MusicDownload".

Looks good, but I'm too tired right now to really switch on brain (hi brain where are you). Later.

yairm210:
How much does this raise the size of the apk?

touhidurrr:
> How much does this raise the size of the apk?

@yairm210 it is < 0.1 MB from my build tests.
See: https://github.com/touhidurrr/buildUnciv/releases

touhidurrr:
> This PR raises the size of the APK compared to the base commit of this branch by 20 kB, from 11,685 kB to 11,705 kB

Just wondering why so little. Maybe because coroutines are built on top of Java Threads or something like that?

yairm210:
😎👍

yairm210:
That was the major reason holding back coroutines however many years ago we tried it, of that's ok then in it goes
I'm not 100% sure about where it would be useful, but having it in opens a lot of options

SomeTroglodyte:
"Jar-Datei" - And I would have guessed you to be a [Tango-dancing](https://visittampere.fi/en/event/world-of-tango-festival/), Suomi full of ["Sisu"](https://en.wikipedia.org/wiki/Sisu) (ntbc w/ politics). Go prejudices go! :partying_face:

SomeTroglodyte:
> There are no threads being used within `MusicController`

But a kotlin.concurrent.timer.Timer? Is that integrated well enough w/ the coroutine thread pool to just leave it?

SomeTroglodyte:
@Azzurite  hey, we've had Music stutter on moderately large loads / nextturns for a while (on Mint at least), and now we're getting `[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)` console logs. Any idea on how to get music to pre-empt other threads (see https://github.com/kcat/openal-soft/issues/554: maybe set that RTkit RLIMIT_RTTIME for _our_ process though I haven't looked for a useable API for that yet)? I've looked into increasing the buffers instead, but my verdict was something along the line that the Gdx api offers some [buffer control](https://github.com/libgdx/libgdx/blob/e03b757caed03d69083ab9c4f5ffe9e594fef1e3/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALLwjgl3Audio.java#L84), they won't help and the buffers one would _really_ like to increase are [hard-coded to 3x 40k](https://github.com/libgdx/libgdx/blob/e03b757caed03d69083ab9c4f5ffe9e594fef1e3/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALMusic.java#L36).

Issue #6700, opened 2022-05-05
City States influence is too easily abusable unlike civ5.


ravignir:
As of now, when a player A declares war on B, player B permanently loses all influence with all the city states player A was allied with. This works different in civ5; for example:

situation 1: 
Player A has 90 influence in city state X
Player B has 80 influence in city state X

Player A and B start a war, triggering city state X to join on player's A side and TEMPORARY drop influence of player B to -60.

Player A and B sign peace treaty allowing player B to sign peace treaty with city state X.

After signing the peace treaty, player's B influence with city states goes back to where it would be if the war had never happened.

As of now in unciv, the last thing is not the case, making all the influence investments pointless, when player A can just reset it to -60. 

situation 2: 
Player A has 80 influence in a city state Y.
Player A declares war on city state Y, influence drops to -60.
Player A signs peace with city state Y. influence does NOT go back to where it would be if the war didn't happen.

This is working correctly afaik.

PR #6813, opened 2022-05-15
Disabled CS buttons when at war; CS keep influence when at war with ally

xlenstra:
This PR does two things:
- When you are at war with a city state, you can no longer give them money, demand stuff, etc.
- When at war with the ally of a city state, your influence is frozen, and after the end of the conflict set back to its original value.

To implement this last thing, I had to split the getter of `influence` off of the underlying field, as otherwise when copying over the old value in `.clone()`, we would copy over -60 when at war, instead of the original value. Most of the changes in that PR are a result of that.
Also resets the `simulateUntilTurns` variable after completing simulation of the turns.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -115,7 +115,9 @@ class DiplomacyManager() {
     var diplomaticModifiers = HashMap<String, Float>()
 
     /** For city-states. Influence is saved in the CITY STATE -> major civ Diplomacy, NOT in the major civ -> city state diplomacy.
-     *  Won't go below [MINIMUM_INFLUENCE]. Note this declaration leads to Major Civs getting MINIMUM_INFLUENCE serialized, but that is ignored. */
+     * Access via getInfluence() and setInfluence() unless you know what you're doing.
+     * Note that not using the setter skips recalculating the ally and bounds checks, 
+     * and skipping the getter bypasses the modified value when at war */
     private var influence = 0f
 
     /** Total of each turn Science during Research Agreement */
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -116,9 +116,7 @@ class DiplomacyManager() {
 
     /** For city-states. Influence is saved in the CITY STATE -> major civ Diplomacy, NOT in the major civ -> city state diplomacy.
      *  Won't go below [MINIMUM_INFLUENCE]. Note this declaration leads to Major Civs getting MINIMUM_INFLUENCE serialized, but that is ignored. */
-    var influence = 0f
-        private set
-        get() = if (civInfo.isAtWarWith(otherCiv())) MINIMUM_INFLUENCE else field
+    private var influence = 0f
 
     /** Total of each turn Science during Research Agreement */
     private var totalOfScienceDuringRA = 0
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -116,9 +116,7 @@ class DiplomacyManager() {
 
     /** For city-states. Influence is saved in the CITY STATE -> major civ Diplomacy, NOT in the major civ -> city state diplomacy.
      *  Won't go below [MINIMUM_INFLUENCE]. Note this declaration leads to Major Civs getting MINIMUM_INFLUENCE serialized, but that is ignored. */
```
SomeTroglodyte, line 118: Fix this comment??

SomeTroglodyte:
Food for thought:
- DiplomacyManager has those pure/state-changing <code><b>//region</b></code>'s that are already no longer perfectly appropriate. The problem is present in many parts of Unciv, and in some files I've touched I mopped up, others I shied away...
- Did you get those three influence writes not going through set/add?  ... the one line 700 yes. :+1:
- Tempting: leave the signature as is, but move to a new backing field, tweaking the serializer to rename it on the fly. Possible, but with Gdx Json a horror. Other frameworks have a `@ SerializeAs` annotation for that, here we would need to implement Gdx Serializable, which forces us to _rewrite_ the default implementations. Yes, e.g. a transparent Json.Serializable.write would be to call json.writeFields, and to tweak a 'real' to 'serailized' mapping of field names in - rewrite. So - correct approach.

SomeTroglodyte (approved):
Approval purely from code review. As said, architecture requiring tons of "chaff" diffs is unavoidable; some well-caught pure lints, and the actual "meat" - I trust Xander knows what he's doing here, actual testing would take me quite a while.

xlenstra:
> DiplomacyManager has those pure/state-changing //region's that are already no longer perfectly appropriate. The problem is present in many parts of Unciv, and in some files I've touched I mopped up, others I shied away...

To be completely honest, I never have paid much attention to these, so this is very likely partially my fault. I didn't even know what they were for for multiple months after joining this project and have never found them useful for anything. I can fix that if you want to, at least for the functions I've added here, but I personally find it more natural to have the `getInfluence()` and `setInfluence()` functions close together rather then in completely different sections of the file

> Did you get those three influence writes not going through set/add? ... the one line 700 yes. 

Yup, the only two raw writes left are in `setInfluence()` and `clone()`, and the only raw reads are in `getInfluence()`, `addInfluence()`, `clone()` and `nextTurnCityStateInfluence()`, as these are the only ones that should try to update the value when at war

SomeTroglodyte:
> functions close together

I know... As I said just food for thought. Regions can be very nice in the Structure window, but pure/nasty isn't the only approach. Region Reorgs probably should go in dedicated PR's. Then again - as long as the damn editor keeps closing the regions I just opened they're only 1/8th as useful as they could be.

JackRainy (approved):
Am I right saying that this change will fix #6700 ?
The code seems to be ok for me but did not test it though.

SomeTroglodyte:
> will fix https://github.com/yairm210/Unciv/issues/6700 ?

Looks very likely - let's link and see if @ravignir comes back :joy_cat:

ravignir:
> > will fix #6700 ?
> 
> Looks very likely - let's link and see if @ravignir comes back 😹

So as this commit is merged, the city state influence won't reset to -60 after the war is finished with their allied civ?

There is one more thing I see is missing here (but i did not correctly explain how it works in #6700 ) ; even when the player attacks CS directly, the -60 lasts until you sign peace treaty, and the real influence begins to increase towards resting point as soon as the war starts, e.g  when you sign a peace treaty after 5 turns of war, you will have -55 influence (given there are no modifiiers to influence growth). So yes, you still lose any influence you had, but you are not  at -60 when you sign peace.

xlenstra:
> So as this commit is merged, the city state influence won't reset to -60 after the war is finished with their allied civ?

Yup!
 
> There is one more thing I see is missing here (but i did not correctly explain how it works in #6700 ) ; even when the player attacks CS directly, the -60 lasts until you sign peace treaty, and the real influence begins to increase towards resting point as soon as the war starts, e.g when you sign a peace treaty after 5 turns of war, you will have -55 influence (given there are no modifiiers to influence growth). So yes, you still lose any influence you had, but you are not at -60 when you sign peace.

I _think_ I implemented that accidentally, but it might not work

PR #6814, opened 2022-05-15
More translations to Brazilian Portuguese

AndreLucas9310:


SomeTroglodyte:
~~Closing and reopening won't help. That can help to force a re-run of the checks, which aren't the issue here and aren't even started (besides, there's a button to re-run those if you know where to look - and maybe have enough permissions).~~

~~Seems like you just have bad luck that the Irremovable line got moved since you started the branch - go use that "Resolve conflicts" button if you can.~~

OK; you got it without that help. :+1:

android/assets/jsons/translations/Brazilian_Portuguese.properties
```diff
@@ -7594,65 +7594,68 @@ May buy [buildingFilter] buildings with [stat] for [amount] times their normal P
...
  # Requires translation!
 Creates a [improvementName] improvement on a specific tile = 
  # Requires translation!
 No defensive terrain penalty = 
  # Requires translation!
-May attack when embarked = 
+May attack when embarked = Pode atacar quando embarcado
  # Requires translation!
-No Sight = 
+No Sight =
```
SomeTroglodyte, line 7623: Don't remove trailing spaces from lines you don't translate... But, it might even be harmless here. Not sure.

AndreLucas9310: Ok, i will solve this.

PR #6816, opened 2022-05-15
Allow editor to generate rectangular Deciv redux maps

SomeTroglodyte:
Another tiny one, still best consumed in small bites. Feel free to omit from changelogs.

Fixes a silly (likely my bad) discrepancy in TileMap constructors that ultimately prevented map editor from generating a rectangular Deciv map.


xlenstra:
Similar to the TileMap constructor, there are also still constants in `generateLandCellularAutomata` preventing map generation of those

xlenstra (approved):

SomeTroglodyte:
Now where's that `generateLandCellularAutomata` comment gone to (good catch - I didn't even "look out of the moving train window")? Could have treated that quickly here, too.

PR #6817, opened 2022-05-15
Fix Gdx.files being garbage collected in MultiplayerTurnCheckWorker

Azzurite:
So the `Worker` in Android can be freely stopped & recreated by the `WorkManager`. It is supposed to be a standalone "application" that can handle re-initialization.

We do this slightly unconventional thing of making a new `OneTimeWorkRequestBuilder` recursively (every time it runs, a new work request is made for the next execution) instead of using `PeriodicWorkRequestBuilder`. I assume that because of that, our first `Worker` instance we started has been kept alive for longer than it was supposed to, and thus no new instance was ever created with usual player behavior.

But apparently after a long time (multiple hours of the turn worker running on my phone), our Worker instance is stopped temporarily and then re-initialized by the `WorkManager`. So all the initialization that happened by `AndroidLauncher` (which is an `AndroidApplication`) that initialized the GDX variables is gone. This includes `Gdx.files`. Then, when the new `Worker` is created, a new `GameSaver` instance is also created. But this instance has so far been statically accessing `Gdx.files`, which is now gone. 

So what the solution is here is to create & pass the correct `[gdx].Files` instance to the `GameSaver` on creation, instead of letting it use `Gdx.files` statically.

<details><summary>Exception</summary>

```
java.lang.NullPointerException: Attempt to invoke interface method 'com.badlogic.gdx.files.FileHandle com.badlogic.gdx.Files.local(java.lang.String)' on a null object reference
 at com.unciv.logic.GameSaver.getSave(GameSaver.kt:36)
 at com.unciv.logic.GameSaver.saveGame(GameSaver.kt:91)
 at com.unciv.logic.GameSaver.saveGame$default(GameSaver.kt:90)
 at com.unciv.app.MultiplayerTurnCheckWorker$doWork$1.invokeSuspend(MultiplayerTurnCheckWorker.kt:268)
 at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
 at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
 at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:279)
 at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:85)
 at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
 at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source:1)
 at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:38)
 at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source:1)
 at com.unciv.app.MultiplayerTurnCheckWorker.doWork(MultiplayerTurnCheckWorker.kt:238)
 at androidx.work.Worker$1.run(Worker.java:86)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
 at java.lang.Thread.run(Thread.java:919)
```

</details>
<details><summary>Old description</summary>
It seems like since the addition of `runBlocking` in `MultiplayerTurnCheckWorker`, the `AndroidApplication` that initially started the background worker gets garbage collected. Which is actually as it should be imo. But the `GameSaver` still needs to work, which is why we need to add a reference to `Gdx.files` to it so it doesn't get garbage collected.
</details>


core/src/com/unciv/logic/GameSaver.kt
```diff
@@ -20,30 +21,40 @@ object GameSaver {
     const val settingsFileName = "GameSettings.json"
     var saveZipped = false
 
-    @Volatile
-    var customSaveLocationHelper: CustomSaveLocationHelper? = null
+    /**
+     * This is necessary because the Android turn check background worker does not hold any reference to the actual [com.badlogic.gdx.Application],
+     * which is normally responsible for keeping the [Gdx] static variables from being garbage collected.
+     */
+    private lateinit var files: Files
```
SomeTroglodyte, line 28: Would that be a little clearer if we shortened the import and use "gdx.Files" here and below? Reminds me of the name-clashing Gdx Array :see_no_evil: - OK, this one is not _that_ bad.

android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
```diff
@@ -32,6 +33,7 @@ class MultiplayerTurnCheckWorker(appContext: Context, workerParams: WorkerParame
 
     companion object {
         const val WORK_TAG = "UNCIV_MULTIPLAYER_TURN_CHECKER_WORKER"
+        const val LOG_TAG = "UNCIV_TURN_CHECK"
```
SomeTroglodyte, line 36: This goes in the droid catlog? Verbatim? That log is getting hundreds of entries per µs, so some more is a good idea, but maximize readability anyway??

SomeTroglodyte: No I mean where does that literal ***go***? Is it visible in the catlog? Is it custom to have all-caps there? The Log.i doc reads as if a class name is customary...

SomeTroglodyte:
> Apparently this was not enough

Makes me curious at to what the symptoms actually look like - or what you see/don't see to determine "enough"

GGGuenni (approved):

SomeTroglodyte:
> no fucking reason

If you're looking for excuses for skin shenanigans on github, sorry, wrong place.

I ass-ume the conflicts are from #6847?

PR #6818, opened 2022-05-15
Great General typed uniques - remastered

JackRainy:
This PR is based on 2 previous implementations: #6751 and #6750 .

I have tried to get all the best from 2 versions + made a performance test of the both ways and took the simplest one.
The performance measurement were made in the next way:
```
val executionTime = measureTimeMillis {
    for (i in 0..100000) {
        val greatGeneralModifier =
            GreatGeneralImplementation.getGreatGeneralBonus(combatant.unit)
        if (greatGeneralModifier.second != 0)
            modifiers[greatGeneralModifier.first] = greatGeneralModifier.second + i - i
    }
}
println(executionTime)
```
while changing the function `getGreatGeneralBonus` within `GreatGeneralImplementation` class and that brought me to the next average results for 14 sequencial runs:
| S implementation with 80 units | J implementation with 80 units | J implementation with 200 units |
| --- | --- | --- |
| 502 ms | 410 ms | 530 ms |

Thus, the performance of the function is around 0.005 ms, which is absolutely acceptable in both implementations. I decided to choose mine as it takes less code = easier to maintain = less bugs in the future.

As a result we have here:
**Pros:**
* Minimum of the additional logic
* The calculation is the same fast for different radius of aura since it does not scan the tiles for units
* Citadel building AI logic is some kind of isolated
* Having several units with different bonuses the maximum will be taken
* The name of unit with `StrengthBonusInRadius` is shown to distinct which of those contributed to the battle stats
* No need to keep the old unique since it will be automatically replaced with the new one in the rules
* `GreatGeneralProvidesDoubleCombatBonus` will be triggered only for Great Generals (China has really great Great Generals but not other modded units)

**Cons:**
* The performance decreases linearly with number of units in civ (but it can be ignored since it takes less than 0.01 ms)
* The replacement of deprecated unique needs to be cleaned up anyway at some point

### DIFF
--- a/core/src/com/unciv/logic/BackwardCompatibility.kt
+++ b/core/src/com/unciv/logic/BackwardCompatibility.kt
@@ -144,6 +144,14 @@ object BackwardCompatibility {
                     unit.promotions.addPromotion(startingPromo, true)
     }
 
+    /** Upgrade the uniques from deprecated format to the new more general one **/
+    fun GameInfo.updateGreatGeneralUniques() {
+        ruleSet.units.values.filter { it.uniques.contains("Bonus for units in 2 tile radius 15%") }.forEach {
+            it.uniques.remove("Bonus for units in 2 tile radius 15%")
+            it.uniques.add("[+15]% Strength bonus for [Military] units in [2] tiles")
+        }
+    }
+
     /** Move max XP from barbarians to new home */
     @Suppress("DEPRECATION")
     fun ModOptions.updateDeprecations() {
--- a/core/src/com/unciv/logic/BackwardCompatibility.kt
+++ b/core/src/com/unciv/logic/BackwardCompatibility.kt
@@ -148,7 +148,7 @@ object BackwardCompatibility {
     fun GameInfo.updateGreatGeneralUniques() {
         ruleSet.units.values.filter { it.uniques.contains("Bonus for units in 2 tile radius 15%") }.forEach {
             it.uniques.remove("Bonus for units in 2 tile radius 15%")
-            it.uniques.add("[+15]% Strength bonus for [Military] units in [2] tiles")
+            it.uniques.add("[+15]% Strength bonus for [Military] units within [2] tiles")
         }
     }
 
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -163,8 +163,7 @@ object NextTurnAutomation {
         while (delta > 0) {
             // Now remove the best offer valued below delta until the deal is barely acceptable
             val offerToRemove = counterofferAsks.filter { it.value <= delta }.maxByOrNull { it.value }
-            if (offerToRemove == null)
-                break // Nothing more can be removed, at least en bloc
+                ?: break  // Nothing more can be removed, at least en bloc
             delta -= offerToRemove.value
             counterofferAsks.remove(offerToRemove.key)
         }
@@ -831,7 +830,7 @@ object NextTurnAutomation {
             when {
                 unit.baseUnit.isRanged() -> rangedUnits.add(unit)
                 unit.baseUnit.isMelee() -> meleeUnits.add(unit)
-                unit.hasUnique("Bonus for units in 2 tile radius 15%")
+                unit.hasGreatGeneralUnique
                     -> generals.add(unit) // Generals move after military units
                 else -> civilianUnits.add(unit)
             }
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -830,7 +830,7 @@ object NextTurnAutomation {
             when {
                 unit.baseUnit.isRanged() -> rangedUnits.add(unit)
                 unit.baseUnit.isMelee() -> meleeUnits.add(unit)
-                unit.hasGreatGeneralUnique
+                unit.isGreatPersonOfType("War")
                     -> generals.add(unit) // Generals move after military units
                 else -> civilianUnits.add(unit)
             }
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -73,8 +73,8 @@ object BattleDamage {
                     modifiers["Missing resource"] = -25  //todo ModConstants
 
             val greatGeneralModifier = GreatGeneralImplementation.getGreatGeneralBonus(combatant.unit)
-            if (greatGeneralModifier != 0)
-                modifiers["Great General"] = greatGeneralModifier
+            if (greatGeneralModifier.second != 0)
+                modifiers[greatGeneralModifier.first] = greatGeneralModifier.second
 
             for (unique in combatant.unit.getMatchingUniques(UniqueType.StrengthWhenStacked)) {
                 var stackedUnitsBonus = 0
--- a/core/src/com/unciv/logic/battle/BattleDamage.kt
+++ b/core/src/com/unciv/logic/battle/BattleDamage.kt
@@ -40,10 +40,9 @@ object BattleDamage {
 
         val conditionalState = StateForConditionals(civInfo, cityInfo = (combatant as? CityCombatant)?.city, ourCombatant = combatant, theirCombatant = enemy,
             attackedTile = attackedTile, combatAction = combatAction)
-        
-        
+
         if (combatant is MapUnitCombatant) {
-            
+
             for (unique in combatant.getMatchingUniques(UniqueType.Strength, conditionalState, true)) {
                 modifiers.add(getModifierStringFromUnique(unique), unique.params[0].toInt())
             }
@@ -61,9 +60,7 @@ object BattleDamage {
 
             //https://www.carlsguides.com/strategy/civilization5/war/combatbonuses.php
             val adjacentUnits = combatant.getTile().neighbors.flatMap { it.getUnits() }
-
-
-            for (unique in adjacentUnits.filter { it.civInfo.isAtWarWith(combatant.getCivInfo()) }
+            for (unique in adjacentUnits.filter { it.civInfo.isAtWarWith(civInfo) }
                     .flatMap { it.getMatchingUniques(UniqueType.StrengthForAdjacentEnemies) })
                 if (combatant.matchesCategory(unique.params[1]) && combatant.getTile()
                         .matchesFilter(unique.params[2])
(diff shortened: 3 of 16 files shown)


core/src/com/unciv/logic/automation/NextTurnAutomation.kt
```diff
@@ -831,7 +830,7 @@ object NextTurnAutomation {
             when {
                 unit.baseUnit.isRanged() -> rangedUnits.add(unit)
                 unit.baseUnit.isMelee() -> meleeUnits.add(unit)
-                unit.hasUnique("Bonus for units in 2 tile radius 15%")
+                unit.isGreatPersonOfType("War")
```
yairm210, line 833: I'm not sure we need this distinction between generals she other civilian units at all

SomeTroglodyte (approved):
First of all, sorry for missing your review request. Secondly, with all the discussion already invested, we should now at least get it perfect, so sorry if I seem a little picky. 

- `fun getGreatGeneralBonus(unit: MapUnit): Pair<String, Int>` - my, I think that Pair is mine. No biggie and certainly no influence on approval, but my take: `Pair` are cute easy and comfortable. But if you later go and do lots of `.first` and `.second` it is often clearer to give things a name (as in a dedicated trivial class). In this case, there's only one use, and that would be a little clearer with deconstruction: `val (unitName, bonus) = GreatGeneralImplementation.getGreatGeneralBonus(combatant.unit); if (bonus != 0) modifiers[unitName] = bonus`... And the word 'Percentage' on the @ return Kdoc seems misplaced.
- Unique on Promotion - will transients update - `addPromotion` hook removed - yes, updateUniques was always called and is what's needed.
- No Deprecation - You cling to your approach and it's cute, yes, but still a deviation from the normal way we established for how to replace uniques with better ones. I say the modders _need_ to get that message in Options-Check mods (and the offer to auto-fix), and that is driven by the @ Deprecated annotations. I don't mean to insist, if you stand by your approach after this last nag and nobody else votes I'll shut up. I know you do use your brain and trust is enough.
- Thanks for taking the time to test and measure performance. Clear picture. I was probably thinking of 300x200 tile maps with 30000 units on them...
- And (tada!) now for the biggest problem: The actual wording of the new strength unique. "in [2] tiles" is maybe 90% as clear as a long explanation. Of course I like mine best, but there were other suggestions. "within [2] tiles"? I'll be fine with whatever you decide, just if you'd be so kind an give it a short think-over?

~~I'll add a second approval but give you a bit of time - feel free to merge yourself.~~ Oops, I mistook Yair's comment for approval. Everything's good.

JackRainy:
> `val (unitName, bonus) = GreatGeneralImplementation.getGreatGeneralBonus(combatant.unit); if (bonus != 0) modifiers[unitName] = bonus`

Very good point! I didn't know this trick.

> And the word 'Percentage' on the @ return Kdoc seems misplaced.

True, will fix.
 
>     * Unique on Promotion - will transients update - `addPromotion` hook removed - yes, updateUniques was always called and is what's needed.

Not sure I understand what you mean. Could you, please, rephrase what is your suggestion.

>     * No Deprecation - You cling to your approach and it's cute, yes, but still a deviation from the normal way we established for how to replace uniques with better ones. 

Actually, I have seen both ways (you can see the `BackwardCompatibility.kt` file is not empty and have some tricks there). I can keep the Deprecated unique at place combining with my auto-replacement logic. Am I right saying that in such case the modder will get the message in "Options-Check mods" nudging for the fix, as well as the game will continue working without problems due to auto-replacement? It sounds like a "win-win".

>     * And (tada!) now for the biggest problem: The actual wording of the new strength unique. "in [2] tiles" is maybe 90% as clear as a long explanation. Of course I like mine best, but there were other suggestions. "within [2] tiles"? I'll be fine with whatever you decide, just if you'd be so kind an give it a short think-over?

Before this review I was thought that both and I are voting for `in [2] tiles` and now I hear your word for `within`. While my English is my 3rd language I tend to agree for the `within` wording. While `in 2` means `distance = 2` but `within` means `distance <= 2` which is more precise description.

SomeTroglodyte:
> rephrase what is your suggestion

Nothing. Internal monologue. This started out as a reminder to myself to check whether all transient stuff updates when it needs to - and it does. :+1: 

>  "win-win"

Why not, because the extra code will disappear in due time. And the "modder will get the message" looks like this (Civ6 mod not updated in a while):
<details>

image
</details>

> 3rd language

:see_no_evil: :hear_no_evil: :speak_no_evil: :ukraine: :+1: :1st_place_medal: 


We're at two votes. Go kick someone for a third :grinning: .

xlenstra:
> We're at two votes. Go kick someone for a third grinning .

On it :)

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -528,8 +529,6 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     RemoveOtherReligions("Removes other religions when spreading religion", UniqueTarget.Unit),
 
     CanActionSeveralTimes("Can [action] [amount] times", UniqueTarget.Unit),
-    // TODO needs to be more general
-    BonusForUnitsInRadius("Bonus for units in 2 tile radius 15%", UniqueTarget.Unit),
```
xlenstra, line 532: Instead of straight-up removing this, instead put this line above it:
```kotlin
@Deprecated("as of 4.1.0", ReplaceWith("[+15]% Strength bonus for [Military] units in [2] tiles"))
```

That way, modders get warnings they should update this unique. Additionally, when going to settings -> show mod errors -> a mod with this unique and press the `auto-update mods` button, it automatically replaces it in the mod itself, instead of the modder having to do it manually.

core/src/com/unciv/logic/BackwardCompatibility.kt
```diff
@@ -144,6 +144,14 @@ object BackwardCompatibility {
...
     }
 
+    /** Upgrade the uniques from deprecated format to the new more general one **/
+    fun GameInfo.updateGreatGeneralUniques() {
+        ruleSet.units.values.filter { it.uniques.contains("Bonus for units in 2 tile radius 15%") }.forEach {
+            it.uniques.remove("Bonus for units in 2 tile radius 15%")
+            it.uniques.add("[+15]% Strength bonus for [Military] units in [2] tiles")
+        }
+    }
+
```
xlenstra, line 154: I'd personally put a `// Deprecated since 4.1.1` above this function so we don't forget to remove it when deprecating uniques (searching for `deprecated` would then reveal it).
Alternatively, doing `it.uniques.any { it.isOfType(UniqueType.GreatGeneralDeprecated)) } ` instead of `it.uniques.contains("Bonus for units in 2 tile radius 15%) }` would also make the connection between this function and the deprecated unique more clear

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -73,17 +70,11 @@ object BattleDamage {
...
-                .flatMap { it.getUnits() }.filter { it.civInfo == combatant.unit.civInfo }
-            if (nearbyCivUnits.any { it.hasUnique("Bonus for units in 2 tile radius 15%") }) {
-                val greatGeneralModifier =
-                    if (combatant.unit.civInfo.hasUnique(UniqueType.GreatGeneralProvidesDoubleCombatBonus)) 30 else 15
-
-                modifiers["Great General"] = greatGeneralModifier
-            }
+            val greatGeneralModifier = GreatGeneralImplementation.getGreatGeneralBonus(combatant.unit)
+            if (greatGeneralModifier.second != 0)
+                modifiers[greatGeneralModifier.first] = greatGeneralModifier.second
```
xlenstra, line 77: I'd prefer extracting this variable immedatialy, e.g.,
```kotlin
val (unitName, modifier) = GreatGeneralImplementation.getGreatGeneralBonus([...])
```
But this is a minor point, feel free to ignore

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
```diff
@@ -0,0 +1,98 @@
...
+                    if (militaryUnit == null || militaryUnit.civInfo != general.civInfo) 0
+                    else generalBonusData.firstOrNull {
+                        // "Military" as commented above only a small optimization
+                        auraTile.aerialDistanceTo(unitTile) <= it.radius
+                                && (it.filter == "Military" || militaryUnit.matchesFilter(it.filter))
+                    }?.bonus ?: 0
+                }
+            }
+    }
+}
```
xlenstra, line 98: I'll assume this is unchanged from SomeTroglodyte's implementation, which looked fine last time I looked at this, so I'll skip it for now

xlenstra (commented):
Looks mostly fine!
You have my approval once the old great general unique is back in the `UniqueType` enum and marked as deprecated.

core/src/com/unciv/logic/battle/GreatGeneralImplementation.kt
JackRainy: No, it is changed a bit. See commit `f027696893f2e39019f74c2e3a3001f8de28b2b8` "[Revert of maxGreatGeneralBonusRadius logic](https://github.com/yairm210/Unciv/pull/6818/commits/f027696893f2e39019f74c2e3a3001f8de28b2b8)"

core/src/com/unciv/logic/BackwardCompatibility.kt
JackRainy: The current comment says "Upgrade the uniques from **deprecated** format to the new more general one", so the search should find it.

JackRainy:
@xlenstra , what about changing `in [2] tiles` for `within [2] tiles` ?

core/src/com/unciv/logic/battle/BattleDamage.kt
SomeTroglodyte: My words, if I remember correctly and didn't forget to hit the green button :+1: And 'datialy' sounds tasty. I'll leave the box and visit kitchen imme- wait - dates - ly. Or was that dates :cherries:  from Italy :it: ?

core/src/com/unciv/logic/BackwardCompatibility.kt
SomeTroglodyte: Or @ Deprecate the fun itself. Also has precedents if I remember correctly. But don't @ Deprecate `Stat.happiness`, the respect-lacking bigfarts are already trying that one and will hopefully fail soon.

xlenstra:
> @xlenstra , what about changing `in [2] tiles` for `within [2] tiles` ?

I'm in favour of that

xlenstra (approved):
Feel free to merge this or change the wording of this unique, you have my approval in both cases

JackRainy:
@SomeTroglodyte , @xlenstra , thank you! That was a truly team work on that!

PR #6821, opened 2022-05-15
Allow generation of "Default" Deciv redux maps

SomeTroglodyte:
Follow-up to @xlenstra comment on #6816.

This time I couldn't resist: That type ignored waterThreshold from the advanced MapParameters, now it doesn't. Effect of slider mapped to a "nice" looking range, comparing water percentages to other types. For that, I patched Map editor to display said water percentage. Might interest others, too. While there, I fixed a problem with "drag" painting not yet dealing with world wrap - never did.

That in turn shows - @will-ca 's arrow code came with the function I use, and I needed a kludge: The function is not meant to guarantee an actual tile at the returned coordinates, but IMHO with "closest to" target (0,0) it should. Doesn't. As I do not really understand all that HexMath, I left the kludge. Maybe someone can look into it?

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -498,6 +498,7 @@ Starting location(s): [param] =
 Continent: [param] ([amount] tiles) = 
 Change map to fit selected ruleset? = 
 Area: [amount] tiles, [amount2] continents/islands = 
+Area: [amount] tiles, [amount2]% water, [amount3] continents/islands = 
 Do you want to leave without saving the recent changes? = 
 Do you want to load another map without saving the recent changes? = 
 Invalid map: Area ([area]) does not match saved dimensions ([dimensions]). = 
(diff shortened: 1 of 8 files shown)


xlenstra (approved):

xlenstra:
> > Follow-up to @xlenstra comment on #6816.
> 
> There is no comment there? only yours?

I deleted the comment after posting as I figured it wasn't actually that relevant to the PR and that there were many spots `Constants.grassLand` and `Constants.ocean` were used. If you have mail notifications on for everything that happens in this repo (which someTroglodyte has), you can still see it as the email already had been sent.
Basically, I noted the exact problem this PR solves; default generation uses `Constants.grassLand` and `Constants.ocean`, which don't necessarily work with mods

SomeTroglodyte:
> email already had been sent

Exactly. Made sense. So I looked again, and thought one band-aid more would be worthwhile.

PR #6822, opened 2022-05-15
Fix coroutines not being added to desktop:dist

Azzurite:
The first commit in here is just the last coroutine PR. The second commit 38b3e14adedb8ad98877dc96134efaa69ad26791 is the interesting one.

For some reason that I don't care about, coroutine classes are only added to the runtime classpath, not the compile classpath. So I just simply add the runtime classpath to the files included in the jar. Which just makes sense, of course we want to have the classes that are required at runtime in the .jar that runs the program.

As for this: https://github.com/yairm210/Unciv/pull/6801#issuecomment-1126895447

> and now we're getting `[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)` console logs

Is this new with the coroutines or did this happen before? If it's happening only now, how do I reproduce it, i.e. which platform? It's not happening on Windows & Android (with this PR) as far as I can see.

> But a kotlin.concurrent.timer.Timer? Is that integrated well enough w/ the coroutine thread pool to just leave it?

I actually didn't check for `kotlin.concurrent.timer.Timer`. It's just a simple thread, one can still use normal threads of course, same behavior as before, so leaving it should simply change nothing. I'll see if it makes sense and how to migrate all the timers.


SomeTroglodyte (approved):
Yes, ALSOFT is a Linux OpenAL thingy.

And re Music - was just meant as trigger for joint brainstorms. Remember on desktop it _doesn't_ use a timer, piggybacking on Gdx's main loop instead. Took me months to figure out - and to balance the decision which compromise to PR.

And `runtimeClasspath + compileClasspath` sounds very much like "**_that's it_**"

SomeTroglodyte:
Say, coroutines - didn't they have specific requirements for androidx.work? And our work-runtime-ktx:2.6.0 is a bit old, but restricted by targetApi = 30. Is that enough for coroutines 1.6?

PR #6823, opened 2022-05-15
Great improvements can again be constructed on forest

xlenstra:
This PR fixes a bug where great improvements couldn't be build on forests/marshes/jungles/etc. It does so by creating a unique which specifically allows for removing features, and checking for that.
Additionally, we only remove these features when we have the tech to remove them. For example, you can no longer plonk an Academy down over a forest without having researched mining.

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -482,9 +482,16 @@ open class TileInfo {
     /** Returns true if the [improvement] can be built on this [TileInfo] */
     fun canBuildImprovement(improvement: TileImprovement, civInfo: CivilizationInfo): Boolean {
         val stateForConditionals = StateForConditionals(civInfo, tile=this)
+        val topTerrain = getLastTerrain()
+
         return when {
+            // Can't build if this improvement is unique to a nation other than us
             improvement.uniqueTo != null && improvement.uniqueTo != civInfo.civName -> false
+            // Can't build if we haven't researched the tech for this improvement
             improvement.techRequired != null && !civInfo.tech.isResearched(improvement.techRequired!!) -> false
+            // Can't build if the unique can't be build
+            improvement.hasUnique(UniqueType.Unbuildable, stateForConditionals) -> false
+            // Can't be build outside our borders, unless we can
             getOwner() != civInfo && !(
                 improvement.hasUnique(UniqueType.CanBuildOutsideBorders, stateForConditionals)
                 || ( // citadel can be built only next to or within own borders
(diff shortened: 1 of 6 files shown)


SomeTroglodyte:
Looks good but... Some readability points maybe. And something nagging in my noggin... Lemme collect my thoughts.

- `enum class RoadStatus.removeAction` - I see so much string compositing with that remove const - might be time the TileImprovement knows how it can be killed? Food for Thought. You know, the "You're not gonna like The Answer :tm:" Thought, first name "Deep"...
- "Can't build if we've obsoleted this unique" - :innocent:  It's not the Unique we've obsoleted... The Unique obsoleted US???
- The nice new "topTerrain.unbuildable && .. && " block: It looks correct, but hard to follow. Either prepare more val's or make absolutely un-overlookable where brackets begin and end + more comments / moved nearer to where they apply??? Will sound like reiterating the obvious to you, but might make sure no future reader stumbles? Maybe?
- `this.terrainFeatures = ` weren't we abstracting access? Is terrainFeatureObjects getting outdated here? I think that may be an actual flaw meriting non-approval?

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -491,13 +491,34 @@ open class TileInfo {
...
     /** Returns true if the [improvement] can be built on this [TileInfo] */
     fun canBuildImprovement(improvement: TileImprovement, civInfo: CivilizationInfo): Boolean {
+
+        fun TileImprovement.canBeBuildOnThisUnbuildableTerrain(civInfo: CivilizationInfo, stateForConditionals: StateForConditionals): Boolean {
+            val topTerrain = getLastTerrain()
+            // We can build if we are specifically allowed to build on this terrain
+            if (isAllowedOnFeature(topTerrain.name)) return true
+
+            // Otherwise, we can if this improvement removes the top terrain
+            if (!hasUnique(UniqueType.RemovesFeaturesIfBuilt, stateForConditionals)) return true
```
yairm210, line 501: ?
If we *don't* have the unique return true? False, surely?

SomeTroglodyte:
Good job!

SomeTroglodyte:
Seems to me this regressed Roads on Forest...? Goes into `canBeBuildOnThisUnbuildableTerrain` and comes to a final 'Nay' conclusion at `hasUnique(UniqueType.RemovesFeaturesIfBuilt)`? The pre-#6712 code had no `UniqueType.Unbuildable` test and went straight through to `canImprovementBeBuiltHere` where the RoadStatus code is...

SomeTroglodyte:
> shit

:hankey: :hankey: :hankey:     Xander will know what to do

PR #6827, opened 2022-05-15
Build warnings cleanup

SomeTroglodyte:
Meant as basis for discussion - and to document my take on what I found in build output. Take or leave or flame or whatever.

Separate commits for:
- Stuff deprecated by Gradle "will break in 8"
- The wrapper properties say 7.2 but the jar and gradle scripts weren't
- Kotlin compilation warnings not related to our own deprecations, again
- And getting the Android Lint step _empty_ - no lint report generated anymore for the moment.


gradlew.bat
```diff
@@ -33,7 +33,7 @@ set APP_HOME=%DIRNAME%
 for %%i in ("%APP_HOME%") do set APP_HOME=%%~fi
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS="-Xmx4G" "-Xms64m"
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
yairm210, line 36: I highly disagree. For gradle runs you *need* a lot of memory otherwise the build will take for *ever*

SomeTroglodyte: Wow. Yes, the automatic gradle self-update does _replace_ gradlew so we need to migrate individual settings. I didn't notice / remember we upped that. With your hint, I _do_ remember wondering how a ba can affect Linux results...

Upping gradle by just telling the wrapper to load a newer version by editing the config is not best practice, though it can be harmless for a time. Once in a while you should do a `./gradlew wrapper --gradle-version X` - add flavour X to taste. I just did that for the already "declared" gradle version, mainly so the wrapper _jar_ matches. Well, it also "upgrades" those scripts... Will patch soonish.

SomeTroglodyte: - We already have `org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=512m` in our root gradle.properties
- Rules for the following: DEFAULT_JVM_OPTS in both gradle scripts as per this PR, as placed by gradle wrapper update. Change jvmargs, invalidate cache and restart Studio, wait for indexing to finish. Gradle sync, then Build-clean and Build-rebuild. (Studio runs with -Xmx8g and actually uses <3g; system has 4g free and 13-14g assigned to cache over the entire test). 
- Rebuilding with `-Xmx6g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8` so I can be sure to _see_ these in system monitor
    - Sync took 49s
    - Rebuild took 5m14s
    - The daemon showed the -Xmx6g in its command line, but moved to a later part than the rest of org.gradle.jvmargs
    - The daemon spawns a sub-JVM for the actual compiler, it inherits -Xmx6g as can be seen in System Monitor
    - Daemon max actual Memory usage was 2.5G, the compiler around 1.5G, sum 4G - so -Xmx6g is way beyond saturation
- Switching to `-Xmx3072m`, clean, rebuild:
    - Sync took 53s
    - Rebuild took 46s - oops? 6x faster than 4g?
    - I get an additional daemon using -Xmx512m and not generating any disk read I/O
- Restart -Xmx3072m from 'invalidate cache'
    - Sync took 58s
    - Rebuild took 49s - oops? Despite invalidate and clean it still does not recompile and runs core:compileKotlin from cache instead.
- Restart -Xmx3072m, manual clean of ~/.gradle (yes it's outside the project what I need to clean, and yes it incurs 2G downloads, but they're outside measurement, the indexer already pulls them)
    - Sync took 53s
    - Rebuild took 5m16ss
- Switching to `-Xmx2g`
    - Sync same
    - 5m26s
- -Xmx1536m
    - Sync 9m23s (ignored as fluke)
    - Rebuild 4m57s
- -Xmx1g
    - Sync 1m5s
    - Rebuild 4m44s
    - Total processes 2.7g (gradle daemon 1.7, compiler 1)
- -Xmx768m
    - Sync 2m38s
    - Rebuild 5m1s
- -Xmx512m
    - Sync 2m51s
    - Rebuild: `java.lang.OutOfMemoryError: Java heap space` after 8m10s

- Conclusion: XX:MaxMetaspaceSize shows in memory usage; together with the observation that _two_ JVM's do the build, a -Xmx4g means a theoretical limit of ~10G memory usage. Our build already runs optimally with -Xmx2g, plus fuzzy factor -> 4g is perfect, maybe 3. The settings in our root gradle.properties _are_ respected, at least on Linux. I'd love to hear from a variety of platforms.

SomeTroglodyte:
The banner is singled out - you probably noticed - because the "hey you need all densities" warning gets suppressed, and the banner dox explicitly say you should only add one density.

The android lint warnings should be in github action build outputs, or `./gradlew build --scan` (but no need to upload the result to the kraken). The gradle whining is actually easiest to see when having the scripts themselves open and everything well synced by looking top right for the tiny yellow ! triangles - or opening the Alt+6 "Problems" pane directly. From gradle runs in my experience they're cached away too quickly, not worth insisting. Did I get that right from memory? Maybe.

As said, this is not meant as "hey this _**must**_ go in" but "hey these can be considered". Green only as it gets attention quicker - and it works fine after all. For me. Which is the only God that Counts. :zany_face:

SomeTroglodyte:
Does pure IntelliJ have the same inspection engine, and the inspections are provided by gradle itself, so some glue missing?? Very very unsure. Gradle's installation pages may have insight. Then again - gradle warnings are only 3 distinct ones, read first commit by itself???

SomeTroglodyte:
>  I don't think we should be supressing any of the lint things you added

Fruitful discussion! Yes, I agree. (But - I've seen automated build logs, maybe here maybe fdroid's, that claimed to fail on the androidLint task alone. So if we ever see such fails again we can try these suppressions I looked up).
Remove or leave comment? Comment you said. Will do, but still hoping for more discussion.

SomeTroglodyte:
That -Xmx4g thing in the gradle windoze bat - that was #6443 (sorry for taking so long to look that up), and @xlenstra said "My **game** was running extremely **slow** without this change, even though I'm on a linux machine. Not sure how, but changing this did fix it". Should we collect any more insight? I mean, that sounds so far-fetched, it might just have been a coincidence and your box could by now run just fine with the default script? If not, the vmopts for debug runs should have their own place not in a windows batch. For the **build** daemons I proved above that our existing settings work.

The IMHO correct place for **debug** jvmopts is in the run configuration, (regrettably) not the project. In Android Studio, edit configurations, in your 'desktop' one "Modify Options - Add VM Options" (or Alt-V), then put them in the new box.
<details>

image
</details>

If you can confirm that helps, we could get that tidbit into the wiki.

For jar runs or the packr'ed stuff the vmopts are yet again elsewhere...

xlenstra:
> That -Xmx4g thing in the gradle windoze bat - that was #6443 (sorry for taking so long to look that up), and xlenstra said "My **game** was running extremely **slow** without this change, even though I'm on a linux machine. Not sure how, but changing this did fix it". Should we collect any more insight? I mean, that sounds so far-fetched, it might just have been a coincidence and your box could by now run just fine with the default script?

I'll test it again, it could well be this was some kind of placebo effect.

gradlew.bat
JackRainy: > Rebuild took 46s - oops? 6x faster than 4g?

I would explain that as 3g were mapped fully in RAM, while 4g could use a swap file as a part of virtual memory, and of course, the file system is much slower than RAM.

SomeTroglodyte: :upside_down_face:  On a 32G box that doesn't even bother assigning the remaining 3G free to file system cache as it seems to feel 13G are enough of that. Nah, swapping to SSD is probably not the explanation, some other app more likely (I know which one - it does massive io every few hours for a few minutes :cinema: ).

SomeTroglodyte:
What we haven't discussed yet: The missing PendingIntent flags. From the docs, passing 0 hasn't been allowed since Marshmallow. But - I may not have chosen the most adequate type, because I don't in-depth understand the `MultiplayerTurnCheckWorker`. Expert opinions?

yairm210 (approved):

touhidurrr:
does gradle warnings even matters? was committing gradle.jar even necessary?

touhidurrr:
I think committing this folder is a mistake
https://github.com/yairm210/Unciv/tree/master/gradle/wrapper

PR #6829, opened 2022-05-16
Fix MayWithdraw modifier calculation

### LINKED ISSUE #6828: Survivalism III melee retreat not working
A unit with Suvivalism III i supposed to have a 75% chance of retreating before being attacked in melee. It was reported on Discord by Vsynk#7904 that it does not work. 

Might actually just be them hitting the 25% all the time :D I have not investigated this yet.

OptimizedForDensity:
Resolves #6828. Part of the calculation was off by a factor of 100 so a slinger with an 80% chance of withdrawal would have its base withdrawal chance calculated at -1900% so it would always fail to withdraw. All promotions or units which use ```UniqueType.MayWithdraw``` should be corrected (tested with slinger and survivalism III at least and those units will mostly withdraw with the occasional failure to withdraw).

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -77,7 +77,7 @@ object Battle {
         // Withdraw from melee ability
         if (attacker is MapUnitCombatant && attacker.isMelee() && defender is MapUnitCombatant) {
             val withdrawUniques = defender.unit.getMatchingUniques(UniqueType.MayWithdraw)
-            val baseWithdrawChance = 100-withdrawUniques.fold(100) { probability, unique -> probability * (100-unique.params[0].toInt()) }
+            val baseWithdrawChance = 100-withdrawUniques.fold(100) { probability, unique -> probability * (100-unique.params[0].toInt()) / 100 }
             // If a mod allows multiple withdraw properties, they stack multiplicatively
             if (baseWithdrawChance != 0 && doWithdrawFromMeleeAbility(attacker, defender, baseWithdrawChance)) return
         }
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
There's a toPercent extension somewhere - applicable?? And if you say "should be corrected" - if that means "in json" - do it for our 2 base rulesets and include. Unless - if you think "forcing" modders to re-visit by deprecating would be better??

yairm210:
I know the 'fold' existed before this PR, but I can't help feel that someone's being clever at the expense if readability.
Sounds to me like you can have 'var chanceToNotWithdraw = 100, foreach unique *= (1-it.param.topercent()' and that'll be much clearer

SomeTroglodyte:
> var chanceToNotWithdraw = 100, foreach

Which is exactly what fold _does_ and is designed for. Which may be a readability gain for some, or not so for others :laughing: . Point being - I am 99% sure there is no performance or even bytecode difference (here, having fold in a complex sequence chain would be a different story), so it's not critical?

That said - why doesn't the chance start at Zero for no Unique? Because the aggregation works on "chance _not_ to pull out". Just renaming the accumulation var, be it lambda parameter `probability` or explicit, to `probabilityToStayPut` would help, right?

And to answer my question: If the Slinger is _supposed_ to cowardly flee 4 times out of 5 (given enough room), then the uniques in json are correct and forget deprecating. Sounds right for the Slinger but never _felt_ right for the Ships. Oh well.

OptimizedForDensity:
I made some readability changes based on some of the suggestions in this thread.

> And if you say "should be corrected" - if that means "in json" - do it for our 2 base rulesets and include

By "should be corrected", I mean that I expect that all promotions and units with ```UniqueType.MayWithdraw``` will function as expected now. I now realize that "should" has a few too many meanings in this context :p

yairm210:
I do think that instead of `100-unique.params[0].toInt()) / 100` it should be `(1-unique.params[0].toPercent())`

PR #6833, opened 2022-05-16

xk730:


heipizhu4 (changes requested):
Nice！Line86 should be changed to "需要[buildingname]：”,没有倒装。

heipizhu4 (approved):

PR #6834, opened 2022-05-16
Add Build Artifacts in `.gitignore`

touhidurrr:
Details:
*.jar -> packr-all-4.0.0.jar
*.zip -> jdk-windows-32.zip, jdk-windows-64.zip
*.tar.gz -> jre-linux-64.tar.gz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,11 @@
 *.ear
 hs_err_pid*
 
+## Build Artifacts
+*.jar
+*.zip
+*.tar.gz
+
 ## Robovm
 /ios/robovm-build/
 


yairm210:
Disagree. We may need to include jar zip or tar.gz files in the future, and this will be an obstacle.
If you want to preclude specific files, then add those specific files to gitignore.

touhidurrr:
Well personally I thought that zip / tar.gz files ending up in the repository is highly unlikely.
Such large Build Artifact blobs are not to be added in a git repository.
Same goes for jars, if there is a 3rd-party jar necessary its most likely going to end up in .gradle cahces.
Thus I thought I would add zips and jars in master `.gitignore` files anyways.

SomeTroglodyte:
<project>/.git/info/exclude - is this the same as X's link? Yup. 
Excerpt from mine:
```
/android/assets/jsons/translations/bak/
/android/*.hprof
/jre-macOS.tar.gz
/packr-all-4.0.0.jar
...
```
Unsurprising, no? The repo-shared ignore should finger stuff *everybody* forking the project and playing around can produce and should definitely never try to push... And the build artifacts are already excluded by folder - except for those packr things, and those you won't get from running the project normally. In fact, the packr targets fail to run locally unless helped along by downloading these files manually. At least for me.

Revisiting the file and thinking about what is in there is good from time to time, thanks for the input. But me, too, I think those bases are covered. What I would change - who needs NetBeans here, some block under the NetBeans comment are needed but not thanks to NetBeans, and the next to last block - Unciv-saved files - needs its own comment header (and maps should move down there, and assets/fonts is outdated?).

PR #6841, opened 2022-05-16
Added more unit tests for uniques

xlenstra:


core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -311,7 +311,7 @@ class CityInfo {
     fun getWorkableTiles() = tilesInRange.asSequence().filter { it.getOwner() == civInfo }
     fun isWorked(tileInfo: TileInfo) = workedTiles.contains(tileInfo.position)
 
-    fun isCapital(): Boolean = cityConstructions.builtBuildings.contains(capitalCityIndicator())
+    fun isCapital(): Boolean = cityConstructions.builtBuildings.any { getRuleset().buildings[it]!!.hasUnique(UniqueType.IndicatesCapital) }
```
SomeTroglodyte, line 314: Debug fluke? But shows a few things: `capitalCityIndicator` would be specific to civ but lives in city, could use a cache, and is buggy: It could choose another civ's unique palace when 'your' civ has no unique palace.

core/src/com/unciv/logic/civilization/CivInfoStats.kt
```diff
@@ -191,26 +191,19 @@ class CivInfoStats(val civInfo: CivilizationInfo) {
...
-            for (unique in civInfo.religionManager.religion!!.getFounderUniques())
-                if (unique.placeholderText == "[] for every [] global followers []")
+                if (unique.isOfType(UniqueType.StatsFromGlobalFollowers))
                     statMap.add(
                         "Religion",
-                        unique.stats *
-                                civInfo.religionManager.numberOfFollowersFollowingThisReligion(
-                                    unique.params[2]
-                                ).toFloat() /
-                                unique.params[1].toFloat()
```
SomeTroglodyte, line 212: This is probably Studio auto-formatting? Could probably be prettier and still shorter. Maybe.

tests/src/com/unciv/uniques/TestGame.kt
```diff
@@ -35,29 +43,50 @@ class TestGame {
...
+        
+        for (row in tileMap.tileMatrix)
+            for (tile in row)
+                if (tile != null)
+                    tile.baseTerrain = baseTerrain
+
+        gameInfo.tileMap = newTileMap
+    }
+    
+    fun getTile(position: Vector2) = setTileFeatures(position)
```
SomeTroglodyte, line 77: This may be right, but it _sounds_ very wrong :laughing:

SomeTroglodyte (approved):
If changed unit tests pass with the current base ruleset, I don't care much if they're correct. They're a detector, and false positives can be dealt with as they come along, while false negatives - oh well, still better than no test at all??? A quick read seems fine.

tests/src/com/unciv/uniques/TestGame.kt
SomeTroglodyte: And the virgin went to the dentist expecting **_less_** pain...

tests/src/com/unciv/uniques/UnitUniquesTests.kt
```diff
@@ -24,9 +24,9 @@ class UnitUniquesTests {
 
     @Test
     fun `Sweden can gift Great Persons to City States`() {
-        // when
```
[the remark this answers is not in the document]
xlenstra: I completely missed the `then` later on and assumed it was a commented out `when()` statement instead. I'll readd it

PR #6846, opened 2022-05-17
Refactor: change GameSaver from singleton to single instance pattern & move autosave logic into GameSaver

Azzurite:
This basically changes `object GameSaver` to `class GameSaver` and instantiates it in `UncivGame.created`. But I also changed a bunch of method signatures & other smaller refactorings within the `GameSaver`. Functionality should be exactly the same, tested everything.

### DIFF
--- a/core/src/com/unciv/logic/GameSaver.kt
+++ b/core/src/com/unciv/logic/GameSaver.kt
@@ -47,7 +47,7 @@ class GameSaver(
 
     private fun getSave(saveFolder: String, gameName: String): FileHandle {
         val localFile = files.local("${saveFolder}/$gameName")
-        if (externalFilesDirForAndroid == "" || !files.isExternalStorageAvailable) return localFile
+        if (externalFilesDirForAndroid == null || externalFilesDirForAndroid.isBlank() || !files.isExternalStorageAvailable) return localFile
         val externalFile = files.absolute(externalFilesDirForAndroid + "/${saveFolder}/$gameName")
         if (localFile.exists() && !externalFile.exists()) return localFile
         return externalFile
@@ -65,7 +65,7 @@ class GameSaver(
 
     private fun getSaves(saveFolder: String): Sequence<FileHandle> {
         val localSaves = files.local(saveFolder).list().asSequence()
-        if (externalFilesDirForAndroid == "" || !files.isExternalStorageAvailable) return localSaves
+        if (externalFilesDirForAndroid == null || externalFilesDirForAndroid.isBlank() || !files.isExternalStorageAvailable) return localSaves
         return localSaves + files.absolute(externalFilesDirForAndroid + "/${saveFolder}").list().asSequence()
     }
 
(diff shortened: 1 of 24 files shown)


SomeTroglodyte:
For a "55 files changed" chimera coming with code snippets from all over the place this looks - from a 5-minute diff scan - surprisingly fine. But just a tiny idea: The maudlin thing wouldn't be a problem if it were a code- and const- only singleton, right? So maybe those three vars making it a stateful container should move instead??? They sound a bit like they should live in UncivGame anyway?

SomeTroglodyte:
> Wut

woot woot - I meant GameSaver is only three fields from being pure code (*). An "Instance" would then be irrelevant??? That may be from a wrong understanding of the VM/language, old COM vtables stuck in brain, but in my mind calling methods would still succeed with a stale, dead or even null `this` if `this` carried no data and wasn't used in the code... And (*): That was disregarding the `const` values which I would also expect to not be part of instance data.

> only 23 files

Yes I know but Buses are cool (Ghibli reference). Now I can say I not only _want_ to approve this as soon as it's done, but I am _confident_ it's fine and well executed.

SomeTroglodyte:
:o:h . Just goes to show I'm not the expert on all that multiplayer stuff. :+1: good to have you on the team.

core/src/com/unciv/logic/GameSaver.kt
```diff
@@ -47,7 +47,7 @@ class GameSaver(
 
     private fun getSave(saveFolder: String, gameName: String): FileHandle {
         val localFile = files.local("${saveFolder}/$gameName")
-        if (externalFilesDirForAndroid == "" || !files.isExternalStorageAvailable) return localFile
+        if (externalFilesDirForAndroid == null || externalFilesDirForAndroid.isBlank() || !files.isExternalStorageAvailable) return localFile
```
SomeTroglodyte, line 50: .isNullOrBlank?

core/src/com/unciv/logic/GameSaver.kt
```diff
@@ -65,7 +65,7 @@ class GameSaver(
 
     private fun getSaves(saveFolder: String): Sequence<FileHandle> {
         val localSaves = files.local(saveFolder).list().asSequence()
-        if (externalFilesDirForAndroid == null || externalFilesDirForAndroid.isBlank() || !files.isExternalStorageAvailable) return localSaves
+        if (externalFilesDirForAndroid.isNullOrBlank() || !files.isExternalStorageAvailable) return localSaves
```
SomeTroglodyte, line 68: Sometimes they copypasta the good parts of dotnet - pardon - mono

SomeTroglodyte:
Pity I don't know `:hankey:` about multiplayer stuff - tempted to approve just because it looks as if you know what you're doing. But approval sounds so official it feels like it needs another level of trust... @GGGuenni ? You have more experience in that area?

PR #6850, opened 2022-05-18
HexaRealm update

GeneralWadaling:
A major update on the HexaRealm tileset.
New additions of snow variant tile improvements as well as era variants of tile improvements.
Added the missing Offshore Platform sprite :)

Snow variants of City Centers has to be postponed as the version I have now isn't exactly satisfying.


xk730 (approved):

SomeTroglodyte:
I would feel better if we had this covered in [credits](https://github.com/yairm210/Unciv/blob/master/docs/Credits.md#icon-credits) - The whole file could do with a little work-over (but I wouldn't know which of the older parts may be outdated - "ThorfMaps tileset? Do we have that? Where?"), but just a line on why we're fine redistributing this and which license exactly applies?

xk730:
> > I would feel better if we had this covered in credits
> 
> I mean I'm not against this, but isn't this original content from the person who made the pull request? If yes, then why don't we put every code contributor in the credits as well?^^ I thought the credits only applied when we take something external, not something we made ourselves.
> 
> But yeah if @GeneralWadaling just pulled the images from somewhere else then I agree.

This is a material independently created by @GeneralWadaling , which was added to the game in an earlier version, tied with Fantasyhex and Default. His purpose of submitting this time is to improve the material

yairm210:
A. About credits, I'll add a line there, this is all original work by him. THE BIG QUESTION @GeneralWadaling is - would you be okay if *other people* used these tiles? If so, we'll include them under Unciv's license and call it a day

B. I saw you were concerned in the Discord about the size. We should have enough pixels in our 2048 cube to accommodate... 2048/64=32, 2048/56=~36, so that's about 1150 tiles. As long as you don't pass that number, we can take all the HexaRealm images and pack them separately, so no worries :)

SomeTroglodyte:
> from somewhere else

No I didn't fear that, just that - well, it's art, for art there are even more trouble hunters than for code, and some "contributed by original author" would be nice to show, especially if git blame shows said author added the line themselves.

GeneralWadaling:
> A. About credits, I'll add a line there, this is all original work by him. THE BIG QUESTION @GeneralWadaling is - would you be okay if _other people_ used these tiles? If so, we'll include them under Unciv's license and call it a day

By *other people* you mean people using them for external (outside-Unciv) purpose?

GeneralWadaling:
I'm okay if others used my tileset, BUT if it is used for **external (non-Unciv) purpose**, ***credits plz***. :)

yairm210:
For now I'll merge this and add a credits line that the images are under Creative Commons 3.0, if you want to change that to anything just say the word :)

yairm210:
I didn't know that existed 😓 
And it's been available since... the end of 2013? So I'm 9 years out of date? Welp

PR #6852, opened 2022-05-18

xk730:


xk730:
> 


我想这的确是个问题，当有了答案我将会回复你

xk730:
> Not sure whether that 'keep it' is right. Anyone knows where 'keep it' is? I may be slow to respond as my PC broke down and I use my phone to reply.

我想这的确是个问题，当有了答案我将会回复你
I think this is really a question. When I have the answer, I will reply to you

SomeTroglodyte:
> Keep it

[This line](https://github.com/yairm210/Unciv/blob/4513ddb75a35f9fd8415c6602c0b3bbb7eb19ce5/core/src/com/unciv/ui/worldscreen/AlertPopup.kt#L153) - AlertType.CityTraded, goes in a button after "What would you like to do with the city" and "Liberate (city returns to [originalOwner])"

SomeTroglodyte:
You know, I looked that up via an Android Studio Find-All for `"Keep it` (with the initial double quote). Without studio - a github search will find the thing too, it just has a different meaning for double quotes: [`https://github.com/search?q=%22Keep+it%22+repo%3Ayairm210%2FUnciv+language%3AKotlin&type=Code&ref=advsearch&l=Kotlin&l=`](https://github.com/search?q=%22Keep+it%22+repo%3Ayairm210%2FUnciv+language%3AKotlin&type=Code&ref=advsearch&l=Kotlin&l=).

Would that have helped you as well? Would it make sense to offer the possibility in our wiki for translators?

yairm210:
prepare version

PR #6854, opened 2022-05-18
Friends list, but still needs work

### LINKED ISSUE #6716: QoL - Friends list for remembering player IDs
Every time you want to start a game with friends you need to get their player IDs and paste them in

Have a way of adding the friends in a list and in NewGameScreen when adding player IDs there can be an option to add from friends list. Easiest way I can think of is to give the user who adds their friend the option to name them, so that we (the code) don't have to keep track of player names and IDs 
To make it clear: When adding a friend you set their name and their player ID. 

In addition the friends list can also remember:
- how many games you've played with X
- how many wins/losses

What do you think ?

alexban011:
resolves #6716 

So the few past days I'm been trying to implement this friends list. I made half the UI: it would still need a button to select a friend in newgamescreen and a screen to edit and show stats about the friend and the ViewFriendsListScreen would also need to load the friends from the FriendsList.json.

My first attempt was to store the friends list in a text file and serialize/deserialize from the string. This is obviously bad so I ditched it and started trying to use JSON. 

EDIT: I didn't mean to mention that issue, I forgot that if I add # it will automatically mention an issue/PR

TODO:
- [x] Create friends list with necessary functions
- [x] make UI for viewing, adding, removing and editing friends
- [x] make UI for selecting friends when creating a new game
- [x] use the right package for FriendsList related files
- [x] remove chosen friends from the list when selecting which friends you want to play with
- [x] don't display the friend selection button if friends list is empty
- [x] check validity of playerID when adding a new friend
- [x] update the ckeckmark after inserting the ID, right now it only updates if you move the cursor in the textbox
- [x] selecting a friend in a player slot where another friend is selected should add the old friend back in the list. Same when changing the id
- [x] make the friend select UI look nicer
- [x] save friendList in GameSettings instead of its own file
- [x] final cleanup  + edit help button
- [x] update template.properties
- [x] thorough testing of every feature / automated tests?


alexban011:
I also tried making the friend class like the following but it wasn't changing the name and ID:

```Kotlin
class Friend{
    var name: String= ""
    var playerID: String = ""

    constructor(name2: String, playerID2: String) {
        this.name = name2
        this.playerID = playerID2
    }
}
```

SomeTroglodyte:
Just some hints:
- `constructor(name: String) { this.name = name }` yes the compiler can tell field and parameter apart. Readability is a different matter.
- For deserialization a no-params constructor is a _must_. If using Gdx Json, which we do. `class Friend(val name: String = "", val playerID: String = "")` would do; `class Friend { lateinit var name: String; lateinit var playerID: String }` probably too.. Note I think these are both not pretty.
- `json.fromJson(MutableList::class.java, Friend::class.java, friendsListFileHandle)` there's such an overload? Didn't know that. Yes, there is... Anyway, a strongly typed fromJson with just one collection class is possible, but - kotlin refuses  `List<String>::class` while `Array<String>::class` is fine. No idea why. Extra zany because `@Suppress("UNCHECKED_CAST")  (Any() as KClass<List<String>>).java` would give you a `Class<List<String>>` just fine but is super ugly...

So I would probably arrive at:
<details>

FriendsList.kt
```kotlin
package com.unciv.ui.multiplayer

import com.badlogic.gdx.files.FileHandle
import com.unciv.json.fromJsonFile
import com.unciv.json.json

class FriendsList: ArrayList<FriendsList.Friend>() {
    companion object {
        private const val friendsListFileName = "FriendsList.json"
        private fun friendsListFileHandle() = FileHandle(friendsListFileName)
    }

    data class Friend(val name: String, val playerID: String) {
        constructor() : this("", "")
    }

    fun addNewFriend(friendName: String, playerID: String) =
        add(Friend(friendName, playerID))

    fun save() {
        json().toJson(this, friendsListFileHandle())
    }

    fun load() {
        val newFriends = json().fromJsonFile(Array<Friend>::class.java, friendsListFileName)
        clear()
        addAll(newFriends)
    }
}
```

FriendsListTest.kt
```kotlin
package com.unciv.testing

import com.unciv.ui.multiplayer.FriendsList
import org.junit.Assert
import org.junit.Before
import org.junit.Test
import org.junit.runner.RunWith
import java.util.*

@RunWith(GdxTestRunner::class)
class FriendsListTest {
    private val friendsList = FriendsList()
    private val friendsList2 = FriendsList()

    @Before
    fun initFriends() {
        friendsList.clear()
        friendsList.addNewFriend("Alice", UUID.randomUUID().toString())
        friendsList.addNewFriend("Bob", UUID.randomUUID().toString())
        friendsList2.addAll(friendsList)
    }

    @Test
    fun friendsRoundTrip() {
        friendsList.save()
        friendsList.clear()
        friendsList.load()
        Assert.assertTrue("This test will pass if a FriendsList serializes and deserializes properly",
            friendsList == friendsList2)
    }
}
```
</details>

... _if_ I wanted them in a separate file. But I'd probably nest them inside our normal GameSettings file.

alexban011:
Thanks @SomeTroglodyte 

Now it adds the friends to the list and displays them in the ViewFriendsListScreen. The refresh button doesn't work correctly at the moment but I gotta go to faculty and then work on a project so I'll continue tomorrow on this.

> if I wanted them in a separate file. But I'd probably nest them inside our normal GameSettings file

If we keep them in a separate file the players can easily import and export them from 1 device to another without importing the settings as well

alexban011:
> Oh, and if you produce a new json in /assets that belongs to the player not to the project, the .gitignore for it should be included in a PR. (remember .git/info/exclude and .gitignore work the same but the former is your private playground while the latter is shared by all here) 

I will add that tomorrow as well. But by "in a PR" you mean I should make a different PR for that or add it in the PR with the friends list ?

alexban011:
Changelog for last commit:
  1. Fixed refresh button and it now refreshes the screen after addin/editing a friend (would the refresh button even be usefull anymore?)
  2. Made load function load the FL correctly (I forgot to clear it before adding the deserialized data)
  3. Added edit button and functionality

alexban011:
Changelog for last commit:
    1. Checked if friendName/playerID of friend is already in friends list before adding it or if the name and ID are actually set
    2. Added button to newgamescreen to select friend from friendlist and a menu to choose the friend.

TODO:
-It needs to remove chosen friends from the list.
-It needs to notice the player to add friends if he tries to open the friend selection menu without having any friends in the friend list
-It needs to look nicer

alexban011:
I don't understand why its saying "Unresolved reference: ViewFriendsListScreen" when AddMultiplayerGameScreen works no problem

SomeTroglodyte:
I only see "Conflicting import, imported name 'PickerScreen' is ambiguous" - probably a bad merge choice. You're merging how? Web? CLI git? Studio? Studio would tell you quickest that there's _two_ identical imports in there, line 10 and line 15.

Actually the output of the Build and deploy action (Details link to the right) is pretty much to the point including line numbers:
```
> Task :core:compileKotlin
e: /home/runner/work/Unciv/Unciv/core/src/com/unciv/ui/multiplayer/MultiplayerScreen.kt: (10, 35): Conflicting import, imported name 'PickerScreen' is ambiguous

> Task :core:compileKotlin FAILED
8 actionable tasks: 8 executed
e: /home/runner/work/Unciv/Unciv/core/src/com/unciv/ui/multiplayer/MultiplayerScreen.kt: (15, 35): Conflicting import, imported name 'PickerScreen' is ambiguous
e: /home/runner/work/Unciv/Unciv/core/src/com/unciv/ui/multiplayer/MultiplayerScreen.kt: (120, 28): Unresolved reference: ViewFriendsListScreen
```
And "Unresolved reference" - well this file doesn't have a matching import, true. I could have failed an interview quiz believing same-package imports were unnecessary, but I don't need to know - Studio knows.

alexban011:
Not sure what I did wrong the first time I merged (Studio) but it should be fixed it now. Too bad I filled the branch with useless commits though

core/src/com/unciv/ui/multiplayer/AddFriendScreen.kt
```diff
@@ -0,0 +1,60 @@
...
+            } else if (friendlist.addFriendErrorType == "id") {
+                ToastPopup("Player ID is already in your friends list!", this)
+            } else if (friendlist.addFriendErrorType == "noName") {
+                ToastPopup("You have to write a name for your friend!", this)
+            } else if (friendlist.addFriendErrorType == "noID") {
+                ToastPopup("You have to write an ID for your friend!", this)
+            } else {
+                backScreen.game.setScreen(backScreen)
+                backScreen.refreshFriendsList()
+            }
```
alexban011, line 55: Should I replace with a when statement ?

alexban011:
I can't figure out how to update the check mark after selecting the friend. 
onPlayerIdTextUpdated() is doing that but I can't call it from outside the class and I can't put it outside the getPlayerTable function as it needs to know to which player it should update the checkmark
And if I call onPlayerIdTextUpdated() after opening the friend list popup it obviously doesn't work since the friend is not selected yet. Somehow I need to call that function after selecting the friend, but I can't think of any way of doing it

core/src/com/unciv/ui/multiplayer/AddFriendScreen.kt
SomeTroglodyte: Studio should automatically tell you to, and even do it for you with those little lightbulbs... You could probably even have only one ToastPopup call with something like `val errorMessave = when (friendlist.addFriendErrorType) { Name -> "Hey that dude's already in here!"... else -> { do and exit } }; Toast(errorMessave...` by no means a request, only food for thought. And that sting code collection reeks like a candidate for a local enum...

alexban011:
Btw why is color "0xff0000ff" not valid, but "0x00ff00ff" is? Is this a bug ?

SomeTroglodyte:
Unsigned Integers are experimental in kotlin, and Color is Gdx, and Gdx is not kotlin-friendly, they're java and have as far as I know not done a single commit to improve kotlin compatibility - from a certain viewpoint that's all right as kotlin is good at consuming java libraries. But java's laxness with signed / unsigned - well, just work around that one. 0xff0000ff is Color.RED for instance, or use public Color (float r, float g, float b, float a), or Color.lerp, Color.brighten or Color.darken your own..........

SomeTroglodyte:
> can't figure out...  onPlayerIdTextUpdated

Can't tell from the diff visible here, and sorry, not enough "drive" at the moment to take the time to pull the branch. Isn't the popup not existing at that point in time yet (or is just opened), and you want stuff to visibly react to outside stuff when you open the popup? Then pass it in in the constructor??? Also. looks long-ish that popup -> own file??

alexban011:
> Isn't the popup not existing at that point in time yet (or is just opened)

it does exist at that point, but if I check right after I open the popup, it checks the ID which isn't set yet

> you want stuff to visibly react to outside stuff when you open the popup?

no, I want it do react after I close the popup

> Also. looks long-ish that popup -> own file??

yes. Its similar to the NationPopup (where you select the nation), so since that one is in its own file I though this one should also be like that

For the copy from clipboard button it makes sense to check the Id right after the button is clicked since it pastes the ID just before the check. But for this it needs to check it after its set. The update() function should do the trick but I'm calling it and it doesn't

alexban011:
Omg I can't believe how easy it actually was. While writing my last comment I thought "if update() draws the entire table why not call the onPlayerIdTextUpdated() function right after declaring it". And yeah that does the trick...

alexban011:
Damn.. the next thing on the TODO list is way harder than I anticipated.. 

The furthest I got to fixing it is to make the friends reappear in the list if you select another friend / click on your user id / paste a user id with the from clipboard button, but changing the ID by hand typing it / CTRL+V doesn't put the selected friend back in the list.

alexban011:
After making some screenshots and/or some videos I'll open the PR.

Should I open this PR to keep the discussion or make a new one and just link this one over there, @SomeTroglodyte ?

SomeTroglodyte:
> keep the discussion or make a new one and just link this one over there, @

Your call. Could make scrolling to new discussion a bit tedious to keep it here, but concise history is also a pro. But be aware that when Issue+PR numbers reach 32768 the repo will crash and all content disappear since it's a 16-bit number.... Oops, 2months-1day late that.

PR #6857, opened 2022-05-18
Refactor: Split options into multiple files

Azzurite:
Functionally, nothing should have changed. I purposefully choose a very coupled design (i.e. optionsPopup is being passed into all functions) just so I had to barely make any architectural changes.


SomeTroglodyte:
Like the idea, like the new package path.

Don't you hate those red traffic signs: image?

Makes me regret I didn't make a `open class TabbedPager.Tab : Table(BaseScreen.skin)` as mandatory type for addPage etc instead of a weak optional `interface TabbedPager.IPageExtensions`! Allowed me some laziness, now allows you some laziness... In other words: Top level non-extension fun's? :vomiting_face:  :stuck_out_tongue_winking_eye:

SomeTroglodyte:
> custom style

`.no-nl-marker { display: none; }`, a little easier on the browser.

> now for you

#6862

SomeTroglodyte:
Another look, and I still would prefer every page file to be a class, But that doesn't change that the idea is sound and a direct solution can always be refactored to a different technique later. I'll approve when conflicks gone.

SomeTroglodyte (approved):
> unconflicksed

No more [Oceans](https://www.urbandictionary.com/define.php?term=Con%20artist) [Eleven](https://www.urbandictionary.com/define.php?term=flicks)

PR #6862, opened 2022-05-19
Unify editor style settings

SomeTroglodyte:
As far as I know both IntelliJ and Android Studio support all of these, and support for [EditorConfig](https://editorconfig.org) files is on by default. All settings open to discussion.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,76 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+ij_any_keep_indents_on_empty_lines = false
+trim_trailing_whitespace = true
+
+[*.{kt,kts}]
+ij_kotlin_code_style_defaults = KOTLIN_OFFICIAL
+ij_kotlin_space_after_comma = true
+ij_kotlin_space_after_extend_colon = true
+ij_kotlin_space_after_type_colon = true
+ij_kotlin_space_before_comma = false
+ij_kotlin_space_before_extend_colon = true
+ij_kotlin_space_before_type_colon = false
+ij_kotlin_space_before_catch_parentheses = true
+ij_kotlin_space_before_for_parentheses = true
+ij_kotlin_space_before_if_parentheses = true
+ij_kotlin_space_before_lambda_arrow = true
+ij_kotlin_space_before_when_parentheses = true
+ij_kotlin_space_before_while_parentheses = true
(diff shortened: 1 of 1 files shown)


yairm210:
Hard to tell until I see it in action, I'm fine with adding this and then nesting with it gradually

yairm210 (approved):

.editorconfig
```diff
@@ -0,0 +1,76 @@
...
+ij_kotlin_spaces_around_unary_operator = false
+ij_kotlin_spaces_around_when_arrow = true
+ij_kotlin_keep_blank_lines_before_right_brace = 1
+ij_kotlin_keep_blank_lines_in_code = 2
+ij_kotlin_keep_blank_lines_in_declarations = 2
+ij_kotlin_block_comment_at_first_column = true
+ij_kotlin_line_comment_add_space = true
+ij_kotlin_name_count_to_use_star_import = 999
+ij_kotlin_name_count_to_use_star_import_for_members = 999
+ij_kotlin_continuation_indent_size = 8
```
xlenstra, line 42: Didn't we want this at 4?

SomeTroglodyte:
> Hard to tell until I see it in action

That's a problem, And I'm a bit disappointed. Studio doesn't really show you in the Settings dialog. If they were heroes they'd have an indicator which switches are set in editorconfig and what the editorconfig setting is (but not hiding display of what 'unset' would do) plus ability to write back changes from the dialog to the file... Remember how M$ group policies would lead to "set by admin" messages in user dialogs. Sigh.

.editorconfig
SomeTroglodyte: Thought so to at first, and some people do it. Switching all of them off works the same - and I snuck the double indent in for unfinished `if` conditions - that might be the one place I would at the moment tend to like the bold in-your-face-this-is-different. Only a proposal, open to any opinions.

xlenstra: I prefer
```kotlin
if (someVeryLongBooleanExpression
    || anotherVeryLongBooleanExpression
) {
    doSomething()
}
```
much more than both of the ones above, and thus would like to go with 4 :slightly_smiling_face:.

xlenstra:
I was still writing a review, as I disagree with leaving the leading indent at 8, while this was merged. Should I open a new PR with the proposed change and my arguments for and against, or can we continue the discussion in this PR?

yairm210:
Sure
We can argue about specific values in separate PRs so as not to couple independent issues

PR #6865, opened 2022-05-19
Fixed a bug blocking special improvements on removable terrain features

xlenstra:
This should hopefully be the last bug fix necessary after the building improvements PR.
It makes it so that improvements like 'remove Forest', 'road', 'cancel improvement order' can again be built on forest

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -488,28 +488,17 @@ open class TileInfo {
         if (terrainFilter == Constants.freshWater && isAdjacentToRiver()) return true
         return (neighbors + this).any { neighbor -> neighbor.matchesFilter(terrainFilter) }
     }
-
+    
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -488,28 +488,17 @@ open class TileInfo {
         if (terrainFilter == Constants.freshWater && isAdjacentToRiver()) return true
         return (neighbors + this).any { neighbor -> neighbor.matchesFilter(terrainFilter) }
     }
-
+
```
SomeTroglodyte, line 491: Ha! Trailing whitespace snuck in!
...sorry. 

OK, chosen solution approach is a very simple concept, which is very likely a good thing :tm:. Let's just all go and keep an eye out for any unusual worker behaviour in any playtest. I could still bite myself for not noticing earlier.

PR #6869, opened 2022-05-19
Add more comments on Editor Settings

itanasi:


SomeTroglodyte:
Next time, we'll include Editor - General - Appearance - Show whitespaces (off, off, Trailing = on), OK? :wink:  there's no .editorconfig ij_* setting for that as it's purely visual.

PR #6870, opened 2022-05-19
Refactor: Extract PickerPane from PickerScreen

Azzurite:
This makes it possible to use the `PickerPane` outside of a `Screen`, for example in a `Popup` (which I'll be doing soon :D)

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/DiplomaticVotePickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/DiplomaticVotePickerScreen.kt
@@ -19,7 +19,7 @@ class DiplomaticVotePickerScreen(private val votingCiv: CivilizationInfo) : Pick
         val choosableCivs = votingCiv.gameInfo.civilizations.filter { it.isMajorCiv() && it != votingCiv && !it.isDefeated() }
         for (civ in choosableCivs)
         {
-            val button = getPickerOptionButton(ImageGetter.getNationIndicator(civ.nation, pickerOptionIconSize), civ.civName)
+            val button = PickerPane.getPickerOptionButton(ImageGetter.getNationIndicator(civ.nation, PickerPane.pickerOptionIconSize), civ.civName)
             button.pack()
             button.onClick {
                 chosenCiv = civ.civName
--- a/core/src/com/unciv/ui/pickerscreens/GreatPersonPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/GreatPersonPickerScreen.kt
@@ -19,7 +19,7 @@ class GreatPersonPickerScreen(val civInfo:CivilizationInfo) : PickerScreen() {
         val useMayaLongCount = civInfo.greatPeople.mayaLimitedFreeGP > 0
 
         for (unit in greatPersonUnits) {
-            val button = getPickerOptionButton(ImageGetter.getUnitIcon(unit.name), unit.name)
+            val button = PickerPane.getPickerOptionButton(ImageGetter.getUnitIcon(unit.name), unit.name)
             button.pack()
             button.isEnabled = !useMayaLongCount || unit.name in civInfo.greatPeople.longCountGPPool
             if (button.isEnabled) button.onClick {
--- a/core/src/com/unciv/ui/pickerscreens/ImprovementPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/ImprovementPickerScreen.kt
@@ -131,7 +131,7 @@ class ImprovementPickerScreen(
 
             regularImprovements.add(statIcons).align(Align.right)
 
-            val improvementButton = getPickerOptionButton(image, labelText)
+            val improvementButton = PickerPane.getPickerOptionButton(image, labelText)
             improvementButton.onClick {
                 selectedImprovement = improvement
                 pick(improvement.name.tr())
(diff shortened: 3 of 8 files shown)


JackRainy:
`PickerScreen.bottomTable` is never used and can be removed. Also, `PickerScreen.rightSideGroup` is used in few places and probably can be removed as well.

While the idea of split pickerPane out of the pickerScreen is super great, the idea of exposing the fields of pickerPane back to the screen smells badly.
I understand that it is absolutely not an easy excercise to split them right now, but while we can decrease the amount of exposing (ah... back to the encapsulation topic :) ) - the better.

JackRainy:
Just another thought: while `PickerScreen.pickerPane` is public and available from outside of the screen, why would you need all those `PickerScreen. ....` leftover-properties? Everybody from outside can reach them via `PickerScreen.pickerPane` !
However, if it is not the intention and we need them as getters, probably, it makes sense to have `pickerPane` private.

SomeTroglodyte:
PickerScreen design and concept has a few other problems. There's kludges (at least two my fault) that should be solved with any major redesign, like wanting two buttons bottom right and "hacking" them in (is that still live? where was it??? modmanager? No... can't really remember - or wait, resetToDefaultsButton in newgamescreen), or wanting a CivilopediaText render instead of a Label down there... Where was that... No matter, there's places where that isn't used but would be better to have. Most of them even maybe.

Just food for thought.

Lofty goal: Make a list of actual requirements, forget Unciv, design a WidgetGroup-based signature supporting the requirements, then come back and implement copying the good parts?

JackRainy:
> So I don't need to make a million changes everywhere.

I read that as "I am too lazy to do the things right, let's go for quick and dirty" :) Sorry for teasing)
Of course, the code does not become way too worse, while I would not dare to take that cleaning on myself, I would approve the code in this state, but isn't it better to clean up after oneself ?

> Make a list of actual requirements, forget Unciv, design a WidgetGroup-based signature supporting the requirements, then come back and implement copying the good parts?

That's a really good aproach. 👍🏻

SomeTroglodyte:
> Sorry for teasing

Sounds familiar. But honestly, I'm willing to support the approach, for an approval with just a bit of backing I would just check out the branch and run it, looking for those uses with the hacks, and if everything seems unchanged... Can't really go wrong in a surprising way if you test all instances once, irrespective of actual content / data??? Only - right now - distracted, lazy and tired. Later. Much later.

JackRainy (approved):
You are welcome to clean unused import directives in the `PickerScreen.kt` and it is ready to go, while you are not much eager to change that encapsulation issue.

SomeTroglodyte:
> don't want to make the design perfect

You're not a perfectionist? Could have fooled me. Reminds me of that DC9 that could't fly higher than 2500 metres anymore due to all the holes patched with tape, and still arriving on the other side of the 4000m mountain range...

SomeTroglodyte:
> Sarcasm

Always. If Sarcasm fails try Frumbling. If that fails too - Blarrbodoozle.

> civilopedia doesn't use a PickerScreen

The other way round. Search for `descriptionLabel.remove()`. The ReligionPicker hacks a `MarkupRenderer.render(belief.getCivilopediaTextLines...` into the place formerly known as pr... as descriptionLabel. Hack = fragile = test. Curiously, when I did that, I was sure I had done it before, but the other place just hacks a subclassed Label into the space, and my memory was saying a MarkupRenderer result... I mean, most pickers show RulesetObject.getDescription() in that Label on selection, and showing getCivilopediaTextLines in the space instead would _always_ be prettier, except pure text is more compact. A first step would be the Widget allowing a constructor to specify the kind of Actor to prepare or an actual Actor instance to insert inside that ScrollPane... The whole disableScroll thing could also be solved nicer. With some time and energy...

JackRainy:
> cleaning up after myself or cleaning up after someone else? Because I didn't design it. I'm just making it slightly more modular and less rigid. Sounds to me like I don't even need to clean up after myself because I didn't make a mess :D

That's true, sorry for bad wording. I meant a ["boy scout" approach](https://matheus.ro/2017/12/11/clean-code-boy-scout-rule/). If one goes for a picnic to the forest and finds the mess around their campfire place left by someone else, one cleans up after oneself (if needed) and that grub.

SomeTroglodyte:
I don't believe boy scouts do that anymore. The only kind of groups still doing that I know of are [spelunkers].(https://en.wikipedia.org/wiki/Caving).

SomeTroglodyte (changes requested):
PolicyPickerScreen regresses. Since in its constructor the scrollPane is no longer homed on a screen, it's dimensions are 150x150 (or 0x0 if you pack its parent), and nicely distributing the policy branches reverts to single row. No elegant idea. Separate init and update, call update when screen is done and contains pickerPane might work.

SomeTroglodyte:
> The real bug is PolicyPicker<a href="https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/Screen.java">Screen</a> doing its layout outside of an overridden [layout](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Layout.java) function like you're supposed to

Ahem... [:computer:](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/Screen.java) `!is` [:triangular_ruler:](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Layout.java), and [:triangular_ruler:](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Layout.java) `!is` [:performing_arts:](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/Actor.java) `!is`  [:triangular_ruler:](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/scenes/scene2d/utils/Layout.java)

SomeTroglodyte (approved):

PR #6876, opened 2022-05-19
Fix screens not correctly being dispose()d

Azzurite:
I ran into the problem while implementing my feature: de-registering of event listeners was not happening reliably. So this doesn't fix an actual problem right now, just a future one.

The reason for that is that `dispose()` was only being called on old `Game.screen`s, which denotes the current screen being shown. `UncivGame.worldScreen` is stored in a separate field to allow going to a different screen and then later returning to the `worldScreen`. But if the current `screen` is, let's say, `MainMenuScreen`, and something somewhere calls (the old) `worldScreen = WorldScreen(...); setWorldScreen()`, only the `screen` is being `dispose()`d, not the old `worldScreen`.

I fixed this by making the `worldScreen` setter private and instead adding a `WorldScreen` parameter to (now) `resetToWorldScreen`.


core/src/com/unciv/UncivGame.kt
```diff
@@ -67,6 +67,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
     val alertBattle = false
 
     lateinit var worldScreen: WorldScreen
+        private set
```
SomeTroglodyte, line 70: Me like

core/src/com/unciv/UncivGame.kt
```diff
@@ -155,18 +156,30 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
+    /**
+     * If called with null [newWorldScreen], disposes of the current screen and sets it to the current stored world screen. 
+     * If the current screen is already the world screen, the only thing that happens is that the world screen updates.
+     */
+    fun resetToWorldScreen(newWorldScreen: WorldScreen? = null) {
+        if (newWorldScreen != null) {
+            val oldWorldScreen = getWorldScreenOrNull()
+            worldScreen = newWorldScreen
+            // setScreen disposes the current screen, but the old world screen is not the current screen, so need to dispose here
+            if (screen != oldWorldScreen) {
```
SomeTroglodyte, line 179: I think the != is guaranteed to be true?

SomeTroglodyte: All right. I just looked at the diff and saw all calls either passing null or a freshly constructed instance. If you thought it through who am I to ... wait ... we're not arguing. I'm just lazily prodding to see _if_ you thought that through.

SomeTroglodyte (approved):
I'll approve untested. Can't really fail spectacularly from inspection; and to actually stage the scenario where I could _test_ and _see_ the difference... I _believe_ (now where's the BeliefType?) there's imperfections to Gdx.Screen.dispose() (it's not even an interface, it's guaranteed to not have any enforcement guarantee) and our use there of, that's enough for me.

GGGuenni (approved):
Looks good to me
But has a conflict

PR #6882, opened 2022-05-20
Go back to current game when you press "Resume" on the main menu (instead of re-loading)

Azzurite:
If you are in a game, go to main menu, and then press "resume" you currently load the autosave. Now you come back to the current game without re-loading.


SomeTroglodyte (approved):
Overlooked this too long. This is the kind "Nothing can possibly go wrong"...
Small aside: I like using :: too but in this case you _may_ have to look twice to see the method does actually get invoked. Wouldn't be an issue if there weren't the forEach. I've pointed repeatedly (cuz teaching myself too) to the kotlinlang.org place where they explicitly discourage... [this one](https://kotlinlang.org/docs/coding-conventions.html#loops)? No reason not to merge though.

PR #6886, opened 2022-05-20
Added more unit tests for uniques; added missing unique implementations

xlenstra:
Added 3 more unit tests, and while doing so found out that a unique wasn't actually implemented for all the cases it claimed to be implemented for.
I've also added the implementation in this PR, and the tests confirm the implementation is correct.

This will be the last PR I do for about a month. I have to focus on finish coding/writing my bachelor thesis, and unciv is too much a distraction to put the time in that needs to be put in. I'll also stop reading PR's and issues on github, and probably even mute the discord. If you need to contact me you can send me a DM on discord.
The only thing I'll probably continue doing is publishing the discord announcement after its been posted as I don't think that's been automated yet :upside_down_face:


tests/src/com/unciv/uniques/TestGame.kt
```diff
@@ -16,6 +16,7 @@ import com.unciv.models.Religion
 import com.unciv.models.metadata.BaseRuleset
 import com.unciv.models.metadata.GameSettings
 import com.unciv.models.ruleset.*
+import com.unciv.models.ruleset.tile.TileImprovement
```
SomeTroglodyte, line 19: Unused import number 4 in this file

tests/src/com/unciv/uniques/GlobalUniquesTests.kt
```diff
@@ -261,11 +242,92 @@ class GlobalUniquesTests {
...
+    fun statsSpendingGreatPeople() {
+        val civInfo = game.addCiv()
+        val tile = game.setTileFeatures(Vector2(0f,0f), Constants.desert)
+        val cityInfo = game.addCity(civInfo, tile, true)
+        val unit = game.addUnit("Great Engineer", civInfo, tile)
+        val building = game.createBuildingWithUnique("[+250 Gold] whenever a Great Person is expended")
+        cityInfo.cityConstructions.addBuilding(building.name)
+        civInfo.addGold(-civInfo.gold)
+
+        civInfo.addGold(-civInfo.gold) // reset gold just to be sure
```
SomeTroglodyte, line 328: Curious. Nothing can be running in parallel modifying that gold???

xlenstra: Nope, just me somehow missing that the above the new line is _exactly the same line_ already. I'll remove it

tests/src/com/unciv/uniques/TestGame.kt
xlenstra: I'll check it and remove it

docs/Modders/uniques.md
```diff
@@ -128,13 +128,13 @@ Simple unique parameters are explained by mouseover. Complex parameters are expl
 
 	Applicable to: Global, FollowerBelief
 
-??? example  "[relativeAmount]% [stat] from every [buildingFilter/tileFilter]"
-	Example: "[+20]% [Culture] from every [Culture]"
```
JackRainy: Just out of curiosity, how did it happen to generate [Culture] twice before? Was it taken from tileFilter?

SomeTroglodyte: No the buildingFilter example is "Culture"

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -319,7 +317,6 @@ open class TileInfo {
             tileUniques += city.getMatchingUniques(UniqueType.StatsFromObject, stateForConditionals)
             for (unique in localUniqueCache.get("StatsFromTilesAndObjects", tileUniques)) {
                 val tileType = unique.params[1]
-                if (tileType == improvement) continue // This is added to the calculation in getImprovementStats. we don't want to add it twice
```
JackRainy, line 322: Not sure I understand why it has been removed. According to the comment, will they be counted twice now?

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -464,17 +511,45 @@ open class TileInfo {
...
+    }
+    
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun getImprovementPercentageStats(improvement: TileImprovement, observingCiv: CivilizationInfo, city: CityInfo?): Stats {
+        val stats = Stats()
+        val conditionalState = StateForConditionals(civInfo = observingCiv, cityInfo = city, tile = this)
+        
+        // I would love to make an interface 'canCallMatchingUniques' 
+        // from which both cityInfo and CivilizationInfo derive, so I don't have to duplicate all this code
+        // But something something too much for this PR.
```
JackRainy, line 526: Double "something".

docs/Modders/uniques.md
JackRainy: Thanks. 🙇🏻‍♂️

core/src/com/unciv/logic/map/TileInfo.kt
xlenstra: The line below calls `matchesTerrainFilter`, which specifically excludes improvements from matching the unique. This line has since become unnecessary, so I removed it. I'm fine with adding it back 'just to be safe', though

core/src/com/unciv/logic/map/TileInfo.kt
xlenstra: [Intentional](https://www.urbandictionary.com/define.php?term=something%20something) though I agree it's a bit informal, I agree. I can clarify the comment if you want

core/src/com/unciv/logic/map/TileInfo.kt
JackRainy: I would keep it in the name of extra-tiny-minor optimization: I assume, it is a bit faster than `matchesTerrainFilter` and at least, you will not need to test for the "Natural Wonder" later.

JackRainy: Another option is to move a check for the natural wonder _before_ the terrain filter check and add `continue` in there since the wonder overrides its terrain properties (or not?).

core/src/com/unciv/logic/map/TileInfo.kt
JackRainy: Ouch.. did not know that slang. Keep it in then. But please, do not use in your thesis. :)

core/src/com/unciv/logic/map/TileInfo.kt
xlenstra: > since the wonder overrides its terrain properties

Not necessarily, only it `overrideStats` is set to true. I'll see if I can simplify the function a bit

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: [ooh I can picture the prof  :eyes:](https://translate.google.com/?source=osdd&sl=auto&tl=nl&text=something%20something%20thesis&op=translate)

core/src/com/unciv/logic/map/TileInfo.kt
xlenstra: I'll see whether I can reorder things for possible efficiency

core/src/com/unciv/logic/map/TileInfo.kt
SomeTroglodyte: Duplicate :link:

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -319,22 +317,29 @@ open class TileInfo {
             tileUniques += city.getMatchingUniques(UniqueType.StatsFromObject, stateForConditionals)
             for (unique in localUniqueCache.get("StatsFromTilesAndObjects", tileUniques)) {
                 val tileType = unique.params[1]
-                if (tileType == improvement) continue // This is added to the calculation in getImprovementStats. we don't want to add it twice
-                if (matchesTerrainFilter(tileType, observingCiv)) 
-                    stats.add(unique.stats)
-                if (tileType == "Natural Wonder" && naturalWonder != null && city.civInfo.hasUnique(UniqueType.DoubleStatsFromNaturalWonders)) {
+                // First check is just a small optimization, as we don't test improvements here.
+                // I'm uncertain as to whether this is worth the time, as string comparison is slow,
```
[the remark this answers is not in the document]
JackRainy: Choose any of the most popular values of tyleType here and it will be an optimization since `matchesTerrainFilter ` does N comparison of strings and having checked the only one the most popular will give its benefits.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -96,8 +95,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     // Stat percentage boosts
     StatPercentBonus("[relativeAmount]% [stat]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
     StatPercentBonusCities("[relativeAmount]% [stat] [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
-    StatPercentFromObject("[relativeAmount]% [stat] from every [tileFilter/specialist/buildingName]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
-    AllStatsPercentFromObject("[relativeAmount]% Yield from every [tileFilter]", UniqueTarget.FollowerBelief, UniqueTarget.Global),
+    // Todo: Add support for specialist next to buildingName
```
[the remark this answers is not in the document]
JackRainy: I understood that like:
> .... from every [tileFilter/buildingFilter/specialistFilter]", .....

SomeTroglodyte:
I was just going to suggest merging this before that conflict happened since Xander seems to do "the right thing"... I wouldn't know how to resolve them without changing authorship to me.

yairm210 (approved):

SomeTroglodyte (approved):
I finally took the time to playtest this for a round. Base is old, so one cannot road a forest, but otherwise no ill effects. The tests themselves of course run fine.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -347,17 +352,67 @@ open class TileInfo {
...
+        for ((stat, value) in getTilePercentageStats(observingCiv, city)) {
+            stats[stat] *= value.toPercent()
+        }
+
+        return stats
+    }
+    
+    // Also gets percentile tile improvement stats
+    @Suppress("MemberVisibilityCanBePrivate")
+    fun getTilePercentageStats(observingCiv: CivilizationInfo?, city: CityInfo?): Stats {
```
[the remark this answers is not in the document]
xlenstra: Yes

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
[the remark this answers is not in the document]
xlenstra: Basically what JackRainy said. `StatsFromObject` has support for `tileFilter`,`specialist` and `buildingFilter`, so it would make sense to also have support for all three of these for `StatPercentFromObject`.

xlenstra:
One thing that is changed in behaviour in this PR, is how percentage bonuses apply for improvements. If, for example, you have "[+20]% [Food] from every [Farm]", that used to only increase the Food output of the farm, but now also increases the Food output of the underlying tile. I made this change to match behaviour from buildings: "[+20]% [Production] from every [Factory]" increases the production of the city it is built in, not only the factory itself.

yairm210:
I *disagree* with this change. If a civ gets, say, "double food from every farm", I would expect the *farm improvement* to provide double food. If I wanted double food from the entire time, I'd go with "double food from every farm *tile*", which exists.

xlenstra:
> I'd go with "double food from every farm tile", which exists.

What unique is that? I couldn't find it in the `UniqueType` enum?

yairm210:
No, you're right, that combo only exists for flat bonuses not percentage ones
My point still stands though

xlenstra:
> My point still stands though

Yup, I've reverted it

PR #6887, opened 2022-05-20
Fix images not being able to have their size set

Azzurite:
This fixes images not being able to have their size set by themselves. 

Currently, they're always wrapped inside a `Table` `Cell` somewhere, and then the cell is being given a size, which completely ignores the child's preferences, so it doesn't matter what size the `Image` actually has.

But that basically makes `Image`s only be usable within `Table` `Cells`, which is not nice, as sometimes that's simply an unnecessary overhead.


itanasi:
Definitely like this idea. But I really don't understand all of the libgdx code yet to say if this works correctly

yairm210:
Looks good :)

PR #6889, opened 2022-05-20
Fix crashes related to getCapital()

### LINKED ISSUE #6844: Crash (probably) when liberating a dead Civ
From google play crash logs:

```
java.util.NoSuchElementException: 
  at com.unciv.logic.civilization.CivilizationInfo.getCapital (CivilizationInfo.kt:1404)
  at com.unciv.logic.civilization.CivilizationInfo.moveCapitalTo (CivilizationInfo.kt:1314)
  at com.unciv.logic.city.CityInfoConquestFunctions.moveToCiv (CityInfoConquestFunctions.kt:283)
  at com.unciv.logic.city.CityInfoConquestFunctions.conquerCity (CityInfoConquestFunctions.kt:105)
  at com.unciv.logic.city.CityInfoConquestFunctions.liberateCity (CityInfoConquestFunctions.kt:203)
  at com.unciv.logic.city.CityInfo.liberateCity (CityInfo.kt:741)
  at com.unciv.ui.worldscreen.AlertPopup$1.invoke (AlertPopup.kt:103)
  at com.unciv.ui.worldscreen.AlertPopup$1.invoke (AlertPopup.kt:102)
  at com.unciv.ui.utils.ExtensionFunctionsKt$onClick$1.invoke (ExtensionFunctions.kt:78)
  at com.unciv.ui.utils.ExtensionFunctionsKt$onClick$1.invoke (ExtensionFunctions.kt:78)
  at com.unciv.ui.utils.ExtensionFunctionsKt$onClickEvent$1.clicked (ExtensionFunctions.kt:71)
  at com.badlogic.gdx.scenes.scene2d.utils.ClickListener.touchUp (ClickListener.java:88)
  at com.badlogic.gdx.scenes.scene2d.InputListener.handle (InputListener.java:71)
  at com.badlogic.gdx.scenes.scene2d.Stage.touchUp (Stage.java:355)
  at com.unciv.ui.crashhandling.CrashHandlingStage.access$touchUp$s80204510 (CrashHandlingStage.kt:10)
  at com.unciv.ui.crashhandling.CrashHandlingStage$touchUp$1.invoke (CrashHandlingStage.kt:24)
  at com.unciv.ui.crashhandling.CrashHandlingStage$touchUp$1.invoke (CrashHandlingStage.kt:24)
  at com.unciv.ui.utils.ExtensionFunctionsKt$wrapCrashHandling$1.invoke (ExtensionFunctions.kt:333)
  at com.unciv.ui.crashhandling.CrashHandlingStage.touchUp (CrashHandlingStage.kt:24)
  at com.badlogic.gdx.backends.android.DefaultAndroidInput.processEvents (DefaultAndroidInput.java:426)
  at com.badlogic.gdx.backends.android.AndroidGraphics.onDrawFrame (AndroidGraphics.java:472)
  at android.opengl.GLSurfaceView$GLThread.guardedRun (GLSurfaceView.java:1581)
  at android.opengl.GLSurfaceView$GLThread.run (GLSurfaceView.java:1280)
```

On first glance this looks like liberating a city to a dead civ causes a crash, as the capital is tried to be moved to the newly liberated civ, but as it has no cities left, `getCapital()` crashes

OptimizedForDensity:
Resolves #6844, the first comment, and allows the posted save to be loaded. It's a lengthier than expected fix since it changes ```CivilizationInfo.getCapital()``` to allow it to return ```null``` instead of just crashing if there's no city where ```isCapital()``` is true. A lot of files see ```.getCapital()``` changed to ```.getCapital()!!``` while others had some if statement conditions changed to include ```getCapital() != null``` as a means of sanity checking, but hopefully all those changes are logical.

There was also a formatting change that was pushed recently that seems to have caused my Android Studio to add or remove spacing, so that also inflated the number of changes.

### DIFF
--- a/core/src/com/unciv/logic/BarbarianManager.kt
+++ b/core/src/com/unciv/logic/BarbarianManager.kt
@@ -102,8 +102,8 @@ class BarbarianManager {
         if (campsToAdd <= 0) return
 
         // Camps can't spawn within 7 tiles of each other or within 4 tiles of major civ capitals
-        val tooCloseToCapitals = gameInfo.civilizations.filterNot { it.isBarbarian() || it.isSpectator() || it.cities.isEmpty() || it.isCityState() }
-            .flatMap { it.getCapital().getCenterTile().getTilesInDistance(4) }.toSet()
+        val tooCloseToCapitals = gameInfo.civilizations.filterNot { it.isBarbarian() || it.isSpectator() || it.cities.isEmpty() || it.isCityState() || it.getCapital() == null }
+            .flatMap { it.getCapital()!!.getCenterTile().getTilesInDistance(4) }.toSet()
         val tooCloseToCamps = camps
             .flatMap { tileMap[it.key].getTilesInDistance(
                     if (it.value.destroyed) 4 else 7
@@ -134,7 +134,7 @@ class BarbarianManager {
                     tile = viableTiles.random()
             } else
                 tile = viableTiles.random()
-            
+
             tile.improvement = Constants.barbarianEncampment
             val newCamp = Encampment(tile.position)
             newCamp.gameInfo = gameInfo
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -139,7 +139,7 @@ class GameInfo {
         civilizations.firstOrNull {
             it.isSpectator() && it.playerId == playerId
         } ?:
-        CivilizationInfo(Constants.spectator).also { 
+        CivilizationInfo(Constants.spectator).also {
             it.playerType = PlayerType.Human
             it.playerId = playerId
             civilizations.add(it)
@@ -179,14 +179,14 @@ class GameInfo {
             }
         }
     }
- 
+
     fun getYear(turnOffset: Int = 0): Int {
         val turn = getEquivalentTurn() + turnOffset
         val yearToTurnList = YearsToTurn.getList(gameParameters.gameSpeed)
         var year: Float = -4000f
         var i = 0
         var yearsPerTurn: Float
- 
+
         // if macros are ever added to kotlin, this is one hell of a place for em'
         while (i < turn) {
             yearsPerTurn = yearToTurnList.firstOrNull { i < it.toTurn }?.yearInterval ?: 0.5f
@@ -292,7 +292,7 @@ class GameInfo {
                 }
         )
     }
-    
+
     fun getEnabledVictories() = ruleSet.victories.filter { !it.value.hiddenInVictoryScreen && gameParameters.victoryTypes.contains(it.key) }
 
     fun processDiplomaticVictory() {
@@ -344,11 +344,11 @@ class GameInfo {
 
         val exploredRevealTiles: Sequence<TileInfo> =
             if (ruleSet.tileResources[resourceName]!!.hasUnique(UniqueType.CityStateOnlyResource)) {
-                // Look for matching mercantile CS centers 
+                // Look for matching mercantile CS centers
                 getAliveCityStates()
                     .asSequence()
                     .filter { it.cityStateResource == resourceName }
-                    .map { it.getCapital().getCenterTile() }
+                    .map { it.getCapital()!!.getCenterTile() }
             } else {
                 tileMap.values
                     .asSequence()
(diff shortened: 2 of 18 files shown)


SomeTroglodyte:
> seems to have caused

Yes please, if any such formatting changes are wrong, do contribute to improving [/.editorconfig](https://github.com/yairm210/Unciv/blob/master/.editorconfig)

OptimizedForDensity:
That's a good point. I mainly did the separate checks to be redundant for situations where a civ had one or more cities but none were listed as the capital. In this case, ```getCapital() == null``` and ```cities.isEmpty() == false```. The posted save in the issue thread did not load at all since ```getCapital()``` was getting called in a place where there wasn't that redundancy, but now that you mention it, it might be a good idea to change ```getCapital()``` to try to set a capital if there isn't one then return that city, otherwise return null which would simplify those conditions.

SomeTroglodyte:
... then again silently fixing things, even when solution is obvious, will hide the source bug - how did that inconsistent state come to be in the first place???

Idea: Similar to unit tests, have a suite of GameInfo sanity checks. When they fail, block next turn and save-to-file, offering a save-to clipboard with new-issue link instead? No, wait, that is what happens when inconsistent state `throw`s...

Yes I know in this case and likely many others the original cause is long past, but still...

JackRainy:
> This PR changes CivilizationInfo.getCapital() to allow it to return null instead of just crashing

in combination with 

> A lot of files see .getCapital() changed to .getCapital()!!

sounds to me like a transfer of the problem from one place to ~another~ many of them.
The first thing I do not understand, why `civInfo.cities.size > 0` can be combined with `getCapital() == null` ? The civ must have a capital as long as there is at least one city.
The second one: we still do not know the origin of the bug #6844 - while loading the saved game I got a crash caused by having a city with no `capitalCityIndicator` - I believe that's the source of the problem: how did we end up with one city with no "Palace"? While the capital is captured it must be transferred to another city, and even when the nation is destroyed and revived by liberation - the capital must be restored.
I suggest fixing the original problem rather than adding hundred of `!!` to spread it further.

JackRainy:
>  add or remove spacing, so that also inflated the number of changes

I would recommend using `git add -p` (patch) or `git add -i` (interactive) with patch option. In this way, you get a full control of what you are commiting. Even if you still want to commit this change you can do that as a separate commit: e.g. "adding/removing spaces", "adding double exclamations", etc.

OptimizedForDensity:
The cause of the crash when destroying a capital in one city challenge mode (the first reply in the linked issue) stems from
https://github.com/yairm210/Unciv/blob/8b9946e65e3356c7afa568066002f14494fcb35e/core/src/com/unciv/logic/city/CityInfo.kt#L773
This line removes the city from the civ's list of cities. If this city is the capital of a civ, any use of ```getCapital()``` from that point until a capital building is added elsewhere in the civ's empire will trigger a crash. Here, this is caused by
https://github.com/yairm210/Unciv/blob/8b9946e65e3356c7afa568066002f14494fcb35e/core/src/com/unciv/logic/city/CityInfo.kt#L784
which in turn calls ```getCapital()``` for attempting to find the next best city (largest by population according to Civ 5 rules). This has been addressed by using ```isCapital()``` in the function ```moveCapitalToNextLargest()``` instead of ```getCapital()```.

The liberation crash also has the same cause where a previously-dead civ has the new city added to ```civInfo.cities``` then ```getCapital()``` is called before the new capital building can be placed. I regretfully missed both cases in testing when working on the capital movement PR but the core issues of those crashes ought to be fixed as far as I can tell. I'll have some more changes uploaded (mainly redundancy stuff) when I get some free time.

SomeTroglodyte:
Hey, an Idea I had outside: Official Support for mods that allow the palace to be destroyed and the victim civ to struggle to rebuild one. Wasn't the Palace a normal national wonder with 200 shields production cost in Civ3 or Civ2? If we analyze the requirements  for such a ruleset and how to make it work, we would also automatically fix this along? Without looking too closely, wasn't the original fix aimed at the situation where a new capital is to be "made", thus it would be intuitive that code should "unmake" an old capital only if there is one...

yairm210:
I'm going to be releasing a version tonight (so that's in like 5 hours or so), if you want this in then fix the conflicts :)
Again, no rush

JackRainy:
#6905 - seems to be also related to this.

SomeTroglodyte:
> seems to be also related to this

From the stacktrace - must be :+1:. I'll link it.

JackRainy:
> I'll link it.
Haven't I done that by mentioning the issue?
image

OptimizedForDensity:
> So does this mean not merge this PR and wait for that? Or do it separately?

Go ahead and merge if everything is ready now, I was a bit too busy yesterday to do any sort of code changes and I'd rather a fix get shipped out to stop the crash.

SomeTroglodyte:
> Haven't I done that by mentioning the issue

Curious. I thought that happens only with the magic keywords, and maybe only for the initial post. After F5 there was only one link in the PR, or the GUI wouldn't have allowed me to link.... ... .. Ah well, It's github. Behaves as if it were owned by Nanosponge... Or it's my providers today. About 5 reconnects per hour...

yairm210 (approved):

PR #6901, opened 2022-05-22
Fix multiplayer turn checker potentially stopping turn checks for everyone that plays that game if a file was not found remotely

Azzurite:
This is a commit I already had on my feature branch but that I should PR separately.

Basically, with #6826 (so this problem is not released yet) I removed all usages of `updateCurrentTurn`. `updateCurrentTurn` was only used because the turn checker set `GameInfoPreview.turnNotification` to `false` as a marker that this game should not be checked anymore because it was not found on the multiplayer server (dropbox).

However, this likely happened because the user messed with the multiplayer server settings and set it to a wrong server. So when they fix those settings, `turnNotification=false` potentially gets uploaded to the server if the game was up to date already and it was their turn, thus resulting in the `turnNotification` being `false` for everyone playing that game and no one getting turn notifications.

The original fix for this behavior was using `updateCurrentTurn` instead of simply replacing the `GameInfoPreview` (verified with the original author). I think using a simple field in the turn checker for this is a better solution, less code and no pollution of the `GameInfoPreview` with "useless" information. I just forgot to include this in the other PR where I removed `updateCurrentTurn`.

### DIFF
--- a/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
+++ b/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
@@ -190,11 +190,9 @@ class MultiplayerTurnCheckWorker(appContext: Context, workerParams: WorkerParame
             for (gameFile in gameFiles) {
                 try {
                     val gamePreview = gameSaver.loadGamePreviewFromFile(gameFile)
-                    if (gamePreview.turnNotification) {
-                        gameIds[count] = gamePreview.gameId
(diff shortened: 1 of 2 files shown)


yairm210 (approved):

GGGuenni (approved):
The original idea of the `GameInfoPreview` variable was to use it as a toggle inside the multiplayer menu game settings to be able to disable turn checks per game. Since it never saw the light of day and I can't find the feature request anyway, it's better to clean it up. 👍

PR #6907, opened 2022-05-22
Add notification on ’F ='

heipizhu4:
Inspired by my own translation fault. When I was translating the map editor, I translated the F into the meaning of ‘floodfill’， but I found I had been wrong because the fortify's shortcut key now says 'floodfill' as well. 
image
That's why I made a small pull here to notify other translators.
Edited: Just a little typo.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -486,7 +486,7 @@ Land or water only =
 
 ## Labels/messages
 Brush ([size]): = 
-# The letter shown in the [size] parameter above for setting "Floodfill"
+# The letter shown in the [size] parameter above for setting "Floodfill". It will also affect the shoutcut key which is shown on 'Fortify' and 'Sleep', so better be careful!
 F = 
 Error loading map! = 
 Map saved successfully! = 


SomeTroglodyte:
OK, but this should be solved at source. **My bad**. I believe I made the keys - all of them - translatable on purpose but didn't "publish" the possibility. And that abbreviation should be translatable _separately_.

IMHO Best would be to

1. Replace F with "Floodfill_Abbreviation" in code and template, and add `Floodfill_Abbreviation = F` to english.properties.
2. A version later, after translators have reacted, publish the key translatability by adding the templates.


Opinions?

heipizhu4:
Sounds good! So shall I close this PR and create a new issue?

yairm210:
Agree with the 'floodfill abbreviation' but not necessarily with the letter translation, we'll need to be careful with that. If the letter is "translated" for display, but clicking on the original letter will still activate the action, we've got a discrepancy between what we tell the user will happen and what actually happens

SomeTroglodyte:
> If the letter is "translated" for display, but clicking on the original letter will still activate the action, we've got a discrepancy between what we tell the user will happen and what actually happens

I think we discussed this marginally - it could be the translator's responsibility that does not happen, but would require a 1:1 correlation between language and keyboard, which is problematic. Perhaps that's why the keys were left translatable because that idea was not evaluated to conclusion... (yes addTooltip auto-translates just like toLabel & co by design)

Ouaz:
Haha, nice catch.

Fun fact: by coincidence, french translations for "Floodfill" (**R**emplir) and "Fortify" (**R**etrancher) start also by the same letter ^^

SomeTroglodyte:
Je suis sûr qu'il y a des alternatives...

PR #6909, opened 2022-05-22
Fix floodfill abbreviation translation showing on fortify key tooltips

SomeTroglodyte:
Credits to @heipizhu4 for #6907 
Untested - I'll test the next few minutes

### DIFF
--- a/android/assets/jsons/translations/English.properties
+++ b/android/assets/jsons/translations/English.properties
@@ -881,7 +881,7 @@ Land or water only =
  # Requires translation!
 Brush ([size]): = 
 # The letter shown in the [size] parameter above for setting "Floodfill"
-Floodfill_abbreviation = F
+Floodfill_Abbreviation = F
 # Requires translation!
 Error loading map! = 
  # Requires translation!
--- a/android/assets/jsons/translations/English.properties
+++ b/android/assets/jsons/translations/English.properties
@@ -881,9 +881,8 @@ Land or water only =
  # Requires translation!
 Brush ([size]): = 
 # The letter shown in the [size] parameter above for setting "Floodfill"
- # Requires translation!
-F = 
- # Requires translation!
+Floodfill_abbreviation = F
+# Requires translation!
 Error loading map! = 
  # Requires translation!
 Map saved successfully! = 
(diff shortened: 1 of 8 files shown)


SomeTroglodyte:
Case mismatch :see_no_evil: :woman_facepalming: - a second...

SomeTroglodyte:
Of course, this will look and handle horribly in languages where the translator hasn't touched that translation... What happens if we add a pre-filled key to templates??? Bad precedence, yes, but curious.

SomeTroglodyte:
I now also migrated the existing entries that had only 1 letter and made it clear via template comment that longer translations will make the slider to the right of it behave less than nicely... Sorry @heipizhu4 - go try dragging the slider to the end and back with a 3-symbol translation...

heipizhu4:
> Of course, this will look and handle horribly in languages where the translator hasn't touched that translation... What happens if we add a pre-filled key to templates??? Bad precedence, yes, but curious.

I suppose a safer way is to type the `Floodfill_Abbreviation = F` in every language after being merged?

SomeTroglodyte:
> I suppose a safer way is to type the `Floodfill_Abbreviation = F` in every language

Uh, didn't think of that... But it removes the prompt "please think about it", you know, the # requires line. Automatically on next run of TFW.

heipizhu4:
>  go try dragging the slider to the end and back with a 3-symbol translation...

You know what? I thought it was a special effect for entering floodfill, but it turns out to be a bug……Kind of intersting.

SomeTroglodyte:
.. _Of course_ that idea was considered - long ago when that was made translatable. There _will_ be languages where that isn't nice. Besides, `take(1)` would be equivalent to `if (isEmpty()) "" else get(0).toString()` - safer and sometimes easier to consume. I _thought_ about `"Floodfill_Abbreviation".tr().takeUnless { it == "Floodfill_Abbreviation" } ?: "Floodfill".tr().take(1)` or somesuch (as extension - possibly already existing somewhere)... Maybe if people complain. (Don't you love how kotlin sometimes feels like SQL? That's the good ol' `ISNULL(NULLIF(..)..)`...)

PR #6917, opened 2022-05-23

xk730:


android/assets/jsons/translations/Simplified_Chinese.properties
```diff
@@ -964,19 +960,13 @@ Worked by [cityName] = 被[cityName]开发
...
-Citizen Management = 
- # Requires translation!
-Avoid Growth = 
- # Requires translation!
-Default Focus = 
- # Requires translation!
-[stat] Focus = 
+Reset Citizens = 重新分配公民
+Citizen Management = 管理公民
+Avoid Growth = 停止增长
```
heipizhu4, line 965: Start a game and you will find that you cannot totally ‘stop' growth with this button. All you can do is to slow them down. So 人口控制或是减少增长会更好。

heipizhu4, line 965: Start a game and you will find that you cannot totally ‘stop' growth with this button. All you can do is to slow them down. So 人口控制或是减少增长会更好。

android/assets/jsons/translations/Simplified_Chinese.properties
```diff
@@ -964,19 +960,13 @@ Worked by [cityName] = 被[cityName]开发
...
+Default Focus = 默认的公民分配重点
+[stat] Focus = 目前分配重点：[stat]
```
heipizhu4, line 967: image
从这张截图可以看到focus不是只有当前会这么写，而是一个选项。所以直接写'注重[stat]就可以了。

heipizhu4 (changes requested):
2 errors. Maybe you should update your game before translating in order to avoid these mistakes?

android/assets/jsons/translations/Simplified_Chinese.properties
xk730: 谢谢更正；此外我一直有在更新游戏

heipizhu4 (approved):
Sorry for the wrong blame. Nice work now!

PR #6920, opened 2022-05-23

hhpy123:


heipizhu4 (commented):
Welcome, new translator! Hardly any person start their translation with these kinds of text, but they are good enough! The only problem is that you should remove some extra space after， 。and ?。They may make sentences hard to read.

PR #6922, opened 2022-05-23
Fix missing icons and translations in civilopedia accessed from main menu

### LINKED ISSUE #6913: Bugged icons in main menu civilopedia
**Platform**
Linux and Android

**Version**
Latest repo (desktop) and 4.1.1-patch2 (Android)

**Describe the bug**
Certain icons are missing in the civilopedia accessible from the main menu. Not all icons are affected, but ones in the Buildings, Wonders, Units, and Technologies tabs will sometimes display a white circle with what looks like a white square on top of it instead of the proper icon. The Terrains and Nations tabs also have some messed up icons but instead of showing the circle-square combination they just don't display any icon whatsoever.

The in-game civilopedia does not display this same behavior and everything works as expected.

**To Reproduce**
1. Click on ? on the main menu

**Expected behavior**
Proper symbols are displayed

**Screenshots**
this icon

OptimizedForDensity:
Fixes #6913. Seems like the game wasn't loading G&K icons. Credit to @SomeTroglodyte who wrote the one-line fix in the issue thread then provided help in this thread. Also ensures that the correct translation is applied to the civilopedia for mods including base ruleset mods.

### DIFF
--- a/core/src/com/unciv/MainMenuScreen.kt
+++ b/core/src/com/unciv/MainMenuScreen.kt
@@ -231,13 +231,11 @@ class MainMenuScreen: BaseScreen() {
     }
 
     private fun openCivilopedia() {
-        var ruleset: Ruleset? = RulesetCache[BaseRuleset.Civ_V_GnK.fullName]
         val rulesetParameters = game.settings.lastGameSetup?.gameParameters
-        if (rulesetParameters != null) {
-            ruleset = RulesetCache.getComplexRuleset(rulesetParameters)
-            UncivGame.Current.translations.translationActiveMods = LinkedHashSet(rulesetParameters.getModsAndBaseRuleset())
-        }
-        if (ruleset == null) return
+        val ruleset = if (rulesetParameters == null)
+                RulesetCache[BaseRuleset.Civ_V_GnK.fullName] ?: return
+                else RulesetCache.getComplexRuleset(rulesetParameters)
+        UncivGame.Current.translations.translationActiveMods = ruleset.mods
         ImageGetter.setNewRuleset(ruleset)
         setSkin()
         game.setScreen(CivilopediaScreen(ruleset, this))
--- a/core/src/com/unciv/MainMenuScreen.kt
+++ b/core/src/com/unciv/MainMenuScreen.kt
@@ -19,7 +19,6 @@ import com.unciv.ui.multiplayer.MultiplayerScreen
 import com.unciv.ui.mapeditor.*
 import com.unciv.models.metadata.GameSetupInfo
 import com.unciv.models.ruleset.Ruleset
-import com.unciv.models.translations.Translations
 import com.unciv.ui.civilopedia.CivilopediaScreen
 import com.unciv.ui.crashhandling.launchCrashHandling
 import com.unciv.ui.crashhandling.postCrashHandlingRunnable
--- a/core/src/com/unciv/MainMenuScreen.kt
+++ b/core/src/com/unciv/MainMenuScreen.kt
@@ -233,6 +233,7 @@ class MainMenuScreen: BaseScreen() {
         val ruleset =RulesetCache[game.settings.lastGameSetup?.gameParameters?.baseRuleset]
             ?: RulesetCache[BaseRuleset.Civ_V_GnK.fullName]
             ?: return
+        ImageGetter.setNewRuleset(ruleset)
         game.setScreen(CivilopediaScreen(ruleset, this))
     }
 
--- a/core/src/com/unciv/MainMenuScreen.kt
+++ b/core/src/com/unciv/MainMenuScreen.kt
@@ -18,6 +18,8 @@ import com.unciv.models.ruleset.RulesetCache
 import com.unciv.ui.multiplayer.MultiplayerScreen
 import com.unciv.ui.mapeditor.*
 import com.unciv.models.metadata.GameSetupInfo
+import com.unciv.models.ruleset.Ruleset
+import com.unciv.models.translations.Translations
 import com.unciv.ui.civilopedia.CivilopediaScreen
 import com.unciv.ui.crashhandling.launchCrashHandling
 import com.unciv.ui.crashhandling.postCrashHandlingRunnable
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
> All credit to

Nah. Was my bad anyway, one look and I knew. Sorry for pushing you to the effort. You could, however, do one better - what I'm still unclear on is the special font symbols when overridden by a mod. There was some nastiness with several copies of a singleton, and I'm unsure whether I found a robust solution. Test (provoke ? starting pedia on the Deciv Redux rules) or go check whether BaseScreen.setSkin() is somehow called... ??

yairm210 (approved):

OptimizedForDensity:
```BaseScreen.setSkin()``` isn't getting called, but after downloading Deciv redux it looks like the main menu civilopedia doesn't load the mod icons. I'm going to take a look into this to see if I can add this change to this PR.

**Comment by OptimizedForDensity** [comment]
The issue seems to be that ```RulesetCache``` doesn't store set the parameter ```mods``` when loading the list of mods so ```ImageGetter``` doesn't know where to find the images. 
https://github.com/yairm210/Unciv/blob/b710f8f6a17d32b8a7151d3f2870effdd13d08c4/core/src/com/unciv/ui/images/ImageGetter.kt#L69-L73

When starting a game, this variable is properly set.
https://github.com/yairm210/Unciv/blob/b710f8f6a17d32b8a7151d3f2870effdd13d08c4/core/src/com/unciv/models/ruleset/Ruleset.kt#L955-L964

Adding a line to set the ```mods``` variable to the ruleset name when loading the ruleset cache appears to fix the problem.

**Comment by SomeTroglodyte** [comment]
I feel this needs a step back and gain perspective and really look and understand the code. Too tired.

Keep in mind ImageGetter keeps ruleset around for looking up colors or differentiating units and buildings when getting a name that could be either - the actual image path to texture translations are in textureRegionDrawables, which setNewRuleset initializes. So the ruleset at that point needs to match the mods we want to display for... Wouldn't be too far a stretch to drop the var, have ImageGetter only receive ruleset or parts when it needs them. Pass actual objects instead of names to the getThingyIcon's and most if not all wouldn't need the var.

Also, openCivilopedia intends to show a _base_ ruleset without any extension mods to complicate matters. ruleset.mods is supposed to be empty for a pure base ruleset and these are what CruelsetCache caches. For Deciv, ImageGetter should then load a single atlas _set_ from the Deciv folder grace to the passed ruleset's baseRuleset field... That's what the missing line is supposed to do... To show the entire base+extensions from lastGameSetup.gameParameters you would want to call getComplexRuleset and that's not there.

But as I said, I'm happy to leave the heavy thinking to someone else. Thank you!

**Comment by SomeTroglodyte** [comment]
Yea, I was wrong. Ruleset doesn't have a baseRuleset field, but name, and it isn't used as I would expect. None of the entries in RulesetCache can be passed to ImageSetter as is, only getComplexRuleset results. Don't like that decision really. A game running only a base ruleset should be able to use the cached ruleset directly without a merge-clone. But even getVanillaRuleset() clones because we don't trust our code not to mutate the rules.

So easiest might be to show that 'pedia for the _complete_ last choice and go and feet it and ImageGetter a getComplexRuleset. Leaving only the Font glitch. Which setSkin() would fix brutally but efficiently.

**Comment by SomeTroglodyte** [comment]
And lastly the translation engine isn't initialized for the ruleset. Ugh. So either always display vanilla or (atrocious formatting, disregard)
```kotlin
    private fun openCivilopedia() {
        val lastParameters = game.settings.lastGameSetup?.gameParameters
        val ruleset = if (lastParameters == null)
            RulesetCache[BaseRuleset.Civ_V_GnK.fullName] ?: return
            else RulesetCache.getComplexRuleset(lastParameters)
        ImageGetter.setNewRuleset(ruleset)
        setSkin()
        game.translations.translationActiveMods = ruleset.mods
        game.setScreen(CivilopediaScreen(ruleset, this))
    }
```
So - getComplexRuleset combines mods and ensures ImageGetter sees the non-vanilla _base_ ruleset in the normal mods field. Without setSkin if you went to Deciv _and_ the vanilla gold was previously rendered, Deciv's Water would not get the water drop. And without translationActiveMods Deciv's water would still be called Gold.

SomeTroglodyte:
I guess you arrived at the same conclusions at the same time...

OptimizedForDensity:
Yeah, I found the code in the game setup screen that controlled translations but thanks for the sanity check/peer review. Everything is working now: translations are applied to the civilopedia in the main menu, mod entries have the correct images, all previously-loaded mods are included.

SomeTroglodyte:
My code still shows "Gold" instead of "Water" when pedia chooses Deciv... Is TranslationActiveModsCache failing??
And the main menu's ESC key and maybe other bindings regress after exiting pedia. Grrr!

core/src/com/unciv/MainMenuScreen.kt
```diff
@@ -230,9 +231,15 @@ class MainMenuScreen: BaseScreen() {
...
     private fun openCivilopedia() {
-        val ruleset =RulesetCache[game.settings.lastGameSetup?.gameParameters?.baseRuleset]
-            ?: RulesetCache[BaseRuleset.Civ_V_GnK.fullName]
-            ?: return
+        var ruleset: Ruleset? = RulesetCache[BaseRuleset.Civ_V_GnK.fullName]
+        val rulesetParameters = game.settings.lastGameSetup?.gameParameters
+        if (rulesetParameters != null) {
+            ruleset = RulesetCache.getComplexRuleset(rulesetParameters)
+            UncivGame.Current.translations.translationActiveMods = LinkedHashSet(rulesetParameters.getModsAndBaseRuleset())
+        }
```
JackRainy, line 239: Do we need to reassign `ruleset` if it has been obtained on the very first line?

OptimizedForDensity:
Hmm weird. I did get the regression on the escape key after exiting the civilopedia on the main menu.
Also noticed that sometimes the font icons don't get changed if you switch mods in the game setup screen then start a new game (Deciv to vanilla still shows the water symbol instead of the gold ingots). Adding a ```BaseScreen.setSkin()``` line to the game setup screen fixes that but there's probably a nicer, more reusable fix.

SomeTroglodyte:
> more reusable fix

Yup, move to ImageSetter.setNewThingy - but all the testing for such a wide-reaching change...

Also, do try to "fix" Deciv Redux by adding the missing Gold = Water english line then test whether the main menu pedia will display it.

> escape key

We suffer from Gdx "Widgets" not being key listeners as e.g. in WPF, only Screens can. There are one or two places already with a little kludge to accommodate that. But here we _do_ have an independent Screen... :puzzled:

core/src/com/unciv/MainMenuScreen.kt
OptimizedForDensity: Good point.

SomeTroglodyte:
> escape key

Must be something simple. New and Load screens do the same, while Mods and Map Editor _repair_ the key.

OptimizedForDensity:
> Also, do try to "fix" Deciv Redux by adding the missing Gold = Water english line then test whether the main menu pedia will display it.

I downloaded the Deciv Redux from the repo a few hours ago and with my code changes the Gold resource is renamed to Water in the civilopedia when you access it from the main menu. Same with Alpha Frontier, it renames Gold to Credits.

SomeTroglodyte:
To keep in mind: Key handling breaks because setScreen disposes the previous screen. Major flaw, maybe I shouldn't have tried to glue KeyPressDispatcher cleanup to that. Back then _everything_ was ditched and rebuilt all the time, so it wasn't a problem. Maybe tell setScreen with a flag the old screen is meant to be reused, or do a resurrection like `if (keyPressDispatcher.isNotEmpty()) keyPressDispatcher.install(screen.stage)` at the end of setScreen (kludge, the old listener might have had a captured ignore lambda), or use keyPressDispatcher.isPaused or...Not sure.

SomeTroglodyte (approved):

PR #6924, opened 2022-05-23

nightingale999:


android/assets/jsons/translations/Traditional_Chinese.properties
```diff
@@ -224,35 +224,23 @@ Provides [amountOfCulture] culture at 30 Influence = 當影響力不低於30時
...
-Build [improvementName] on [resourceName] (200 Gold) = 
- # Requires translation!
-Gift Improvement = 
+Protected by = 被保護
+Revoke Protection = 撤銷保護
+Revoke protection for [cityStateName]? = 撤銷對[cityStateName]的保護嗎?
+Pledge to protect = 請求保護
+Declare Protection of [cityStateName]? = 宣告對[cityStateName]的保護嗎?
+Build [improvementName] on [resourceName] (200 Gold) = 以[resourceName] (200金) 建造[improvementName] 
+Gift Improvement = 禮物升格
```
heipizhu4, line 236: ‘improvement' here is a noun. 'Gift Improvement’ means gifting improvement (like mines) to CS instead of 'giving something better'. So it should be ‘贈送設施'

heipizhu4 (changes requested):
Hi there! Just a small error, the rest is good enough.

android/assets/jsons/translations/Traditional_Chinese.properties
nightingale999: Thanks for the suggestion. Just updated it!

heipizhu4 (approved):

PR #6925, opened 2022-05-23
confirmation when pressing next turn

### LINKED ISSUE #6880: Feature request: optionally require confirmation for ending your turn
Buttons "Next unit" and "Next turn" are in the same place. On desktop both use `Space` key as shortcut. It is fairly easy to accidentally end your turn when you don't really mean so. 

**Request**: add an option "Require confirmation for ending the turn. By default could be unchecked, so that it behaves as now.

alexban011:
resolves #6880 

Adds an option in settings for getting an YesNoPopup when pressing next turn to avoid accidentally pressing it

By default is off

### DIFF
--- a/core/src/com/unciv/ui/options/GameplayTab.kt
+++ b/core/src/com/unciv/ui/options/GameplayTab.kt
@@ -33,4 +33,5 @@ fun gameplayTab(
         settings.automatedWorkersReplaceImprovements
     ) { settings.automatedWorkersReplaceImprovements = it }
     optionsPopup.addCheckbox(this, "Order trade offers by amount", settings.orderTradeOffersByAmount) { settings.orderTradeOffersByAmount = it }
+    optionsPopup.addCheckbox(this, "Ask for confirmation when pressing next turn", settings.confirmNextTurn) { settings.confirmNextTurn = it }
 }
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -39,6 +39,7 @@ import com.unciv.ui.pickerscreens.*
 import com.unciv.ui.popup.ExitGamePopup
 import com.unciv.ui.popup.Popup
 import com.unciv.ui.popup.ToastPopup
+import com.unciv.ui.popup.YesNoPopup
 import com.unciv.ui.popup.hasOpenPopups
 import com.unciv.ui.saves.LoadGameScreen
 import com.unciv.ui.saves.SaveGameScreen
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -822,8 +823,15 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
...
-                    game.settings.addCompletedTutorialTask("Pass a turn")
-                    nextTurn()
+                    if(game.settings.confirmNextTurn) {
+                        YesNoPopup("Confirm next turn", {
+                            game.settings.addCompletedTutorialTask("Pass a turn")
+                            nextTurn()
+                        }, UncivGame.Current.worldScreen).open()
+                    } else {
+                        game.settings.addCompletedTutorialTask("Pass a turn")
+                        nextTurn()
```
JackRainy, line 833: You can move these 2 lines to a separate function (e.g. `nextTurnClicked`) and reuse them, however, that's a minor thing.

JackRainy (approved):

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
alexban011: I was thinking about that but I wasn't sure if its worth it for just 2 lines of code which are only used in this if-else

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -822,8 +823,15 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
 
             else ->
                 NextTurnAction("${Fonts.turn}{Next turn}", Color.WHITE) {
-                    game.settings.addCompletedTutorialTask("Pass a turn")
-                    nextTurn()
+                    if(game.settings.confirmNextTurn) {
```
SomeTroglodyte, line 826: https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
SomeTroglodyte: And `game.settings` could be `UncivGame.Current.settings`... or was that the other way round? Or can a `this` be used somewhere? Maybe this looks tasty:?
```kotlin
                    val action = {
                        game.settings.addCompletedTutorialTask("Pass a turn")
                        nextTurn()
                    }
                    if (setting)
                        YesNoPopup("Confirm", action, this).open()
                    else action()
```

core/src/com/unciv/ui/options/GameplayTab.kt
```diff
@@ -33,4 +33,5 @@ fun gameplayTab(
         settings.automatedWorkersReplaceImprovements
     ) { settings.automatedWorkersReplaceImprovements = it }
     optionsPopup.addCheckbox(this, "Order trade offers by amount", settings.orderTradeOffersByAmount) { settings.orderTradeOffersByAmount = it }
-}
\ No newline at end of file
+    optionsPopup.addCheckbox(this, "Ask for confirmation when pressing next turn", settings.confirmNextTurn) { settings.confirmNextTurn = it }
```
SomeTroglodyte, line 36: Add to template.properties. Make sure your editor doesn't ruin the trailing blank.

alexban011:
Shouldn't I add the "confirm next turn" from the popup in template.properties as well?

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
alexban011: the game.settings was already used before so I just kept it as is. Its like this in the entire file
But I'll add the action like this cause it looks much better than an additional function

PR #6928, opened 2022-05-23
Adding the missing spaceship parts unit sprites for FantasyHex tileset

GeneralWadaling:
yeah it is simple as that


SomeTroglodyte (approved):
Cute lil' truckies.
We'll have to remember to redo the atlas before next release, no biggie.

PR #6929, opened 2022-05-24
Save and load reorg and keyboard handling

SomeTroglodyte:
Some rework of the SaveGameScreen and LoadGameScreen.
At first this was meant to not change any existing behaviour, but then... 

#### Debatable details
- (edited - GLFW is not to blame but lwjgl3 - or Gdx for not using the raw function) A drive-by shooting: [GLFW](https://www.glfw.org/docs/3.4/group__input.html#gaba1f022c5eb07dfac421df34cdcd31dd) is used under the hood by lwjgl3 and thus by Gdx for the clipboard. The [lwjgl3 glue function](https://github.com/LWJGL/lwjgl3/blob/9afbf8c3e0c980394a897af78db41b75e68a30df/modules/lwjgl/glfw/src/generated/java/org/lwjgl/glfw/GLFW.java#L4712) limits data size to the depth of a MemoryStack which defaults to 64k. Thus most games (and crashscreen copy) would error out and not copy. 384k is still stingy, but - why the hell were they so extremely stingy in the first place? Since that limit affects the GZipped size - could cover most games. No relation to Android version - different backend.
- The SaveGameScreen has a "Current Saves" header, Load does not - for the same content. I've kept the discrepancy on purpose. Could save a few lines if we were to equalize.

#### Features
- Modular, some code duplication reduced - some still to find (those three moved functions resume quickload and quicksave are for now only moved to the ui/saves package)
- Both can be _mostly_ operated by keyboard now. Ctrl-S, Enter - saved to default name. Ctrl-S, "Hello", Enter - saved as "Hello". Ctrl-L, Arrow down, Enter - loads most recent. Vertical arrow keys and PgUp/PgDown, Enter, Ctrl-C. Ctrl-V and Del do logical things. Ctrl-A is almost intuitive.
- More info when things go wrong
- Less dialog when loading a save with missing mods the "normal" way - Jack's code covered only from clipboard and ?from custom? - anyway, works the same for normal load now, meaning no error dialog but the red label and button to download.
- I think there were one or two places where the heavy lifting was still synchronous.

#### Discarded (maybe postponed) ideas
- Arrows could use a key repeat function. Far too much effort - you would have to do timer stuff yourself. Gdx first suppresses any key auto repeat the OS might do, then reimplements it for some widgets - try reading TextField source.
- Filter and Sort akin to Mod manager
- Confirm for Delete
- Forgot some...


SomeTroglodyte:
> Also, I've 100% copied to clipboard saves larger than 64kb before this PR. Guess there's another another implementation on Windows?

Makes me curious. Wanna go and set a breakpoint down there in GLFW where the MemoryStack is built and look at the first pointer returned? I did, and it _was_ just a few hundred below 64k. GLFW sources also confirm that is the default.

SomeTroglodyte:
- Take any `Gdx.app.clipboard.contents =`, e.g. old LoadGameScreen line 132
- Ctrl+B on the `contents`, takes you to com.badlogic.gdx.utils.Clipboard
- Select `Clipboard`, Ctrl-Alt-B, choose Lwjgl3Clipboard - now we're in com.badlogic.gdx.backends.lwjgl3
- Down on `GLFW.glfwSetClipboardString` a Ctrl-B - now in org.lwjgl.glfw
- Set the breakpoint right after the `try`, launch and copy something large
- In my choice I get `string.length() == 248864` and `stackPointer == 65536`
- F8 -> goes right to `finally`, no `nglfwSetClipboardString`

yairm210:
Conflicts

SomeTroglodyte:
> Conflicts

Pretty obvious, both touching MainMenu.autoLoadGame - but I'll re-test both desktop and droid.

SomeTroglodyte:
Good thing I tested again - there was a loophole where having autosaves _only_, then showing them, selecting one, un-showing them (so no saves visbible) then pressing an arrow key would crash: found & eradicated.

SomeTroglodyte:
If only we understood more. Will those 384k be added as TLS to _every_ thread in the process??? Can we adapt and set it dynamically? Could we bypass Gdx to get at GLFW.glfwSetClipboardString(ByteBuffer)? Sigh...

core/src/com/unciv/ui/saves/QuickSave.kt
```diff
@@ -0,0 +1,97 @@
...
+        launchCrashHandling("LoadGame") {
+            try {
+                val loadedGame = gameSaver.loadGameByName("QuickSave")
+                postCrashHandlingRunnable {
+                    toast.close()
+                    UncivGame.Current.loadGame(loadedGame)
+                    ToastPopup("Quickload successful.", screen)
+                }
+            } catch (ex: Exception) {
+                postCrashHandlingRunnable {
```
yairm210, line 46: Toast should be closed on unsuccessful attempt as well, maybe the whole thread should be taken out of try catch

core/src/com/unciv/ui/saves/QuickSave.kt
```diff
@@ -0,0 +1,97 @@
...
+                    loadingPopup.close()
+                    ToastPopup("Cannot resume game!", screen)
+                }
+                return@launchCrashHandling
+            }
+
+            if (savedGame.gameParameters.isOnlineMultiplayer) {
+                try {
+                    screen.game.onlineMultiplayer.loadGame(savedGame)
```
yairm210, line 82: We're still running in a thread, you suresure that loadmultiplayer doesn't require gl context?
Maybe the if else should be inside the try catch for simplicity

desktop/src/com/unciv/app/desktop/DesktopLauncher.kt
```diff
@@ -23,6 +23,8 @@ internal object DesktopLauncher {
         // Solves a rendering problem in specific GPUs and drivers.
         // For more info see https://github.com/yairm210/Unciv/pull/3202 and https://github.com/LWJGL/lwjgl/issues/119
         System.setProperty("org.lwjgl.opengl.Display.allowSoftwareOpenGL", "true")
+        // This setting (default 64) limits clipboard transfers. Value in kB!
```
yairm210, line 26: If this is an arbitrary number please write that so we know later we can change it
If not please look to relevant source

core/src/com/unciv/ui/saves/QuickSave.kt
SomeTroglodyte: This was just _moved_ code, no attempt yet to improve on it.

The whole file is a TODO meant for later, the advantage being to have all save/load _UI_ code in one package and folder, out of WorldScreen, making such improvements easier. And before anyone asks - yea I _know_ maps _aren't_ reusing this for the common usecases but should do so.

core/src/com/unciv/ui/saves/QuickSave.kt
SomeTroglodyte: Moved code last touched by Timo if remember the merge correctly. Didn't second-guess it.

desktop/src/com/unciv/app/desktop/DesktopLauncher.kt
SomeTroglodyte: Maybe even make it a setting to allow our support hotline to just say 'open your config in VIM and add...'... Nah. Maybe one day I have the Eureka on how to bypass Gdx calling the wrong lwjgl3 function.

core/src/com/unciv/ui/saves/QuickSave.kt
SomeTroglodyte: ...At least in the case the just loaded GameInfo passes the `hasLatestGameState` test (which runs synchronously), the OnlineMultiplayer code does the postRunnable. So, better not interfere.

desktop/src/com/unciv/app/desktop/DesktopLauncher.kt
SomeTroglodyte: The lwjgl3 glue layer to GLFW has a glfwSetClipboardString(long, CharSequence) and a glfwSetClipboardString(long, ByteBuffer) overload, and only the former does that despicable MemoryStack.nUTF8 conversion.

SomeTroglodyte: lwjgl3, not GLFW - the file is called that, but it's the lwjgl3 glue to the latter. Yes of course, I just meant Gdx would have a change to mitigate, but we no longer have. Not in a way I can think of. Well, short of of forking and patching. Or PR'ing a patch to them and waiting....................

SomeTroglodyte: Yea but... That clipboard stuff gets complicated once you dive deep, and some of that abstraction I wouldn't want to replicate. So -> ask lwjgl3 directly, but if we pull in those dependencies, will that coexist nicely??? That's where my brain says "SEP".

yairm210:
Is ready?

PR #6933, opened 2022-05-24
added optional on-screen buttons to zoom in and out

### LINKED ISSUE #4480: Zoom in/out buttons for stylus users
I am requesting dedicated zoom-in and zoom-out buttons in this game. This will make the game easier to play for stylus users by eliminating the one action where multiple fingers are required. This feature should not be enabled by default, as the majority of players likely do not have a stylus.

**Describe alternatives you've considered**
I believe that simple touchscreen buttons are the best way to add one-touch zooming, and I can't think of any other method of zooming that can be done with a stylus. Perhaps the volume rocker could be used to control zooming?

**Additional context**
I am testing this game on a Moto G Stylus 2021. While playing this game, the only action that I was unable to execute with the stylus was zooming.

alexban011:
resolves #4480

Added an option in the display tab of Options (by default off) to show buttons for zooming in and out

zoomButtons

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -651,6 +651,7 @@ Show pixel units =
 Show pixel improvements = 
 Enable Nuclear Weapons = 
 Experimental Demographics scoreboard = 
+Show zoom buttons in world screen = 
 Show tile yields = 
 Show unit movement arrows = 
 Continuous rendering = 
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -52,6 +52,7 @@ class GameSettings {
     var isFreshlyCreated = false
     var visualMods = HashSet<String>()
     var useDemographics: Boolean = false
+    var showZoomButtons: Boolean = false
 
 
     var multiplayerServer = Constants.dropboxMultiplayerServer
--- a/core/src/com/unciv/ui/options/DisplayTab.kt
+++ b/core/src/com/unciv/ui/options/DisplayTab.kt
@@ -34,6 +34,7 @@ fun displayTab(
     optionsPopup.addCheckbox(this, "Show pixel units", settings.showPixelUnits, true) { settings.showPixelUnits = it }
     optionsPopup.addCheckbox(this, "Show pixel improvements", settings.showPixelImprovements, true) { settings.showPixelImprovements = it }
     optionsPopup.addCheckbox(this, "Experimental Demographics scoreboard", settings.useDemographics, true) { settings.useDemographics = it }
+    optionsPopup.addCheckbox(this, "Show zoom buttons in world screen", settings.showZoomButtons, true) { settings.showZoomButtons = it }
 
     addMinimapSizeSlider(this, settings, optionsPopup.screen, optionsPopup.selectBoxMinWidth)
 
--- a/core/src/com/unciv/ui/worldscreen/minimap/MinimapHolder.kt
+++ b/core/src/com/unciv/ui/worldscreen/minimap/MinimapHolder.kt
@@ -7,6 +7,8 @@ import com.badlogic.gdx.utils.Align
 import com.unciv.UncivGame
 import com.unciv.logic.civilization.CivilizationInfo
 import com.unciv.ui.images.ImageGetter
+import com.unciv.ui.utils.onClick
+import com.unciv.ui.utils.toTextButton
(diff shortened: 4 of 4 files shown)


SomeTroglodyte:
The EditorMapHolder is a 66% copy of WorldMapHolder. might be simple. But maybe an actual UncivSlider? Comes with free +- buttons... Of course, how to map linear to nonlinear... And would be smaller. But if neolithic phone users can hit the little green ~~men~~ cows... (hey -"nonlinear"- let it be linear with snapto positions exactly where normal zoom could take you? Then with shift you could have fine tuning...)

alexban011:
>  But maybe an actual UncivSlider? Comes with free +- buttons...

yeah but I think the actual slider would clutter the screen without any benefits, and it would probably be harder to press the buttons

>  how to map linear to nonlinear

this one could be done the way it works for the number of turns, since that is also nonlinear. But again, not sure if it would look and work better than normal TextButtons

alexban011:
> For which use cases is this exactly? In what way wouldn't the other options to zoom work? I.e. why exactly have this? :D

If you want to play the game with a stylus on mobile, zooming in and out is impossible unless you use your fingers. 
On desktop I doubt there is much reason to use it, unless your mouse wheel is broken? :)) Or maybe if you play using the mouse pad and want to have more control over the zoom? (On my mouse pad dragging my fingers for 1 cm zooms in and out completely)

SomeTroglodyte:
> unless your mouse wheel is broken

Or it's [wheeeeeeeeeelllleeeesssss mouse](https://en.wikipedia.org/wiki/File:SRI_Computer_Mouse.jpg)

SomeTroglodyte:
Wait, stylus? _Now_ I know why I immediately _saw_ that slider in my mind. Wasn't there an attempt, or did I try by myself... And it got stuck learning how to reliably detect the stylus hardware. _That_ I remember. I even dug out a Note3 and tried stuff...

alexban011:
> Wait, stylus? _Now_ I know why I immediately _saw_ that slider in my mind. Wasn't there an attempt, or did I try by myself... And it got stuck learning how to reliably detect the stylus hardware. _That_ I remember. I even dug out a Note3 and tried stuff...

yep. You probably remember this from the issue I mentioned :)

SomeTroglodyte:
(That old issue - what a memory. Embarrassing).  But detecting a stylus would be :sunglasses:. OK, go OptionsPopup.

core/src/com/unciv/ui/worldscreen/ZoomButton.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.unciv.ui.utils.setFontSize
+
+/**
+ * Creates a zoom button
+ * zoomType is "in" for zooming in
+ * and "out" for zooming out
+ *
+ * @param zoomType
+ */
+class ZoomButton(zoomType: String, mapHolderEditor: EditorMapHolder? = null, mapHolderWorld: WorldMapHolder? = null): TextButton("", BaseScreen.skin) {
```
alexban011, line 18: I'm gonna change it a bit cause I really don't like this

alexban011: done

SomeTroglodyte:
Nonononono that's not _elegant_ at all (sorry but from a glance I'd go about it _very_ differently and to wrap that in helpful words is not easy - so you get unsorted chaos)...
- I see a good deal of code duplication and a missed chance for a helpful enum
- You know that render is heavy so anything not absolutely necessary in render shouldn't be in render... In fact, this kind of Widget shouldn't need overriding render at all methinks?
- "Wrong label for zoom button" shouldn't go on the console every time a bad zoom button is tried, but be an IllegalArgumentException or somesuch on construction. Which an enum would make unnecessary.
- Target is always a ZoomableScrollPane, which should be the way to eliminate all that duplication.
- And since the common concept the "Client" sees is always a pair of +/-, I'd start imagining a `ZoomInOutButtonPair(parent: ZoomableScrollPane)` or maybe call it ZoomButtons or ScaleControlWidget or whatever. Maybe a vertical: Boolean parameter, but not at first. Positioning responsibility of client is OK, but with just 1 Widget half the code. And that construct makes an in/out enum and a subclass for the single button overkill.
- There's a setPosition(x, y, align) which is technically crap but very good for readability.

SomeTroglodyte (approved):
I'll not merge immediately though - my vote doesn't really count as independent, and with 4.1.3 just through we can give it a little time for others to voice opinions? Edit: Nevermind. OP can do the conflix first anyway.

PR #6937, opened 2022-05-25
Zoom buttons

SomeTroglodyte:
For #6933


alexban011:
> Is there an unicode char with a bigger plus to match the dash?

maybe U+FF0B and then U+FF0D instead of the dash?

SomeTroglodyte:
> U+FF0B and then U+FF0D

I'd not. Stay in UCS-16 without combined chars should be safer.

Oh, and try to compare to the +/- stuff for city stats (expand all/none, and stats drilldown), didn't do that either.

alexban011:
Screenshot_20220525_200901

en-dash looks nice on arch, I'll test on android as well

alexban011:
but I don't understand why its centering the buttons for you but my + buttpm isn't centered (looks like the base of the + is centered not the + itself)

SomeTroglodyte:
> looks like the base of the + is centered

Which is actually as correct as centering the center of the glyph - namely not. The virtual room for descenders is/should be included in centering, and descenders should typically have far less room than ascenders. See tooltip code for some nice kludges involving that.

alexban011:
Well I tested it on android and it looks the same as on Ubuntu. Looks like its a (tiny) issue on arch only

PR #6940, opened 2022-05-25
AI counteroffers will no longer contain items already offered by the player

yairm210:
Inspired by Discord picture of double declaration of war in AI counteroffer


yairm210:
This is about how much they request of *you*, not the other way round.
But yes - this does mean that if you already offered them gold, they won't change the amount of gold they request, they will only add to that.
Still better than getting a counteroffer for more gold than you have, which is immediately invalidated by start-of-turn checks... gold offer is the least interesting option

PR #6944, opened 2022-05-25
Update Spanish.properties

ArchDuque-Pancake:
Question: Can translated lines have `" "` in them? Or does it break something?
I ask because in spanish the hierarchy of quotation marks is: «First order “Second ‘third order’ order”».
As probably most fonts don't have `« »` (neither does my keyboard), these (" ") are the "correct" ones to use instead of `' '`.


ArchDuque-Pancake:
Ok, entonces revisaré las comillas erroneas y ajustarlas a las comillas inglesas.

PR #6945, opened 2022-05-25
WordScreenTopBar rework, portrait-friendlier

SomeTroglodyte:
Now also fixes an outstanding minor issue - NotificationsScroll didn't re-layout on Minimap size changes. Now it does.

<details><summary>First, Why</summary>

image
</details>

Ugly overlap, possible even with pretty normal aspect ratios and resolutions; plus gaps in resources + year at turn 0 not centered in Vanilla (with Deciv the resource issue is more pronounced) and so forth.

<details><summary>The new code shifts stuff down only as necessary:</summary>

https://user-images.githubusercontent.com/63000004/170380871-1840c8b9-09bb-4fff-a70c-8d67d33ec23d.mp4

</details>

Now to the question: It's still kind of ugly. Getting those "stepped" background continuations to have a rounded corner might help, but that seems like daunting additional work. And maybe someone has a better idea.


SomeTroglodyte:
Well, rounded corners is possible:
<details>

https://user-images.githubusercontent.com/63000004/170802151-55112950-d71c-4969-ad62-c275532c005a.mp4
</details>
This has visual glitches at rare aspect ratios, likely the stock Actor.clipBegin miscalculating or some other misunderstanding... To get these out would be another few hours of poking in the dark.

SomeTroglodyte:
The stat icons are 'dancing', I didn't notice. Will need fine-tuning.

SomeTroglodyte:
Rounded corners are in, and some fixes (padding and I overlooked clear also clears listeners).

The glitch is still in, with a breakpoint where the glitch is visible I can see all coordinates match, and it still draws a gap between rectangles whose coordinates _do_ touch. Only happens once or twice when you drag the window width over the whole screen, and not in landscape or near 9:16 portrait.

yairm210:
Cute :)

core/src/com/unciv/ui/worldscreen/BackgroundActor.kt
```diff
@@ -0,0 +1,61 @@
...
+/** An Actor that just draws a Drawable [background], preferably a [NinePatchDrawable] created
+ *  by [BackgroundActor.getRoundedEdgeRectangle], meant to work in Table Cells and to be overlaid with other Widgets.
+ *  The drawable's center can be moved to any of the corners or vertex centers using `align`, which will also scale the
+ *  drawable up by factor 2 and clip to the original rectangle. This can be used to draw rectangles with one or two rounded corners.
+ *  @param align An [Align] constant - In which corner of the [BackgroundActor] rectangle the center of the [background] should be.
+ */
+class BackgroundActor(val background: Drawable, align: Int) : Actor() {
+    private val widthMultiplier = if (Align.isCenterHorizontal(align)) 1f else 2f
+    private val heightMultiplier = if (Align.isCenterVertical(align)) 1f else 2f
+    private val noClip = (align and Align.center.inv()) == 0
```
yairm210, line 22: This is obscure, don't use boolean logic on ints, even if it does work

SomeTroglodyte: The only test I wanted that Java hasn't in Align... :sob:  - 

but OK if we do not want to be hardened against align values Align.center and 0 being in fact equivalent to all other code we can just compare directly, much simpler. _Or_ Align.isCenterHorizontal(align) && Align.isCenterVertical(align) would be the same thing...

The whole class is not used for all it _can_ do - which this test showed: 
<details><summary>Screen</summary>

image
</details>
... but maybe we want such building blocks elsewhere in the future. That test helped a lot to think about the logic correctly.

PR #6946, opened 2022-05-25
puppet cities only focus on gold

### LINKED ISSUE #6939: Citizen management issues
**Platform**
Android and Linux

**Version**
4.1.2-patch1

**Describe the bug**
2 related issues
* Spectators can reassign citizen focus in cities (for example, switch from food focus to gold focus and avoid growth)
* Players can reassign citizens in puppeted cities while in Civ 5 these cities are switched to gold focus and are unable to be changed until annexed (https://civilization.fandom.com/wiki/Conquered_cities_(Civ5)#Puppet_cities, https://www.carlsguides.com/strategy/civilization5/war/annex-puppet-raze.php)

**To Reproduce**
1. Start a game as a spectator
2. Change the focus of a city

For the second issue
1. Puppet a city
2. Change the focus of the puppet

**Expected behavior**
* City focus should not change in either case
* Puppeted cities should be locked into gold focus

alexban011:
resolves #6939 (the unresolved part, the other part is resolved by #6954)

code checks if the player is a spectator and if the city is a puppet before changing the focus


alexban011:
Actually I don't think it sets puppet city's focus to gold yet, need to do some digging

itanasi:
Better to block the onClick actions in the City Screen via the canChangeState Boolean. Similar to the other logic in CityScreen

core/src/com/unciv/logic/city/CityInfo.kt
```diff
@@ -634,8 +634,10 @@ class CityInfo {
         tryUpdateRoadStatus()
         attackedThisTurn = false
 
-        if (isPuppet) reassignAllPopulation()
-        else if (updateCitizens){
+        if (isPuppet) {
+            reassignAllPopulation()
+            cityAIFocus = CityFocus.GoldFocus
```
itanasi, line 639: Set cityAIFocus before reassignAllPopulation()

core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
```diff
@@ -34,7 +35,13 @@ class CitizenManagementTable(val cityScreen: CityScreen) : Table() {
         val avoidCell = Table()
         avoidCell.touchable = Touchable.enabled
         avoidCell.add(avoidLabel).pad(5f)
-        avoidCell.onClick { city.avoidGrowth = !city.avoidGrowth; city.reassignPopulation(); cityScreen.update() }
+        avoidCell.onClick {
+            if (!UncivGame.Current.gameInfo.currentPlayerCiv.isSpectator() && !city.isPuppet) {
```
itanasi, line 39: Instead of checking isSpectator(), use cityScreen.canChangeState
There's also several other buttons you'll need to disable

itanasi (changes requested):

itanasi:
I handled all the buttons if you're a Spectator in #6954.

Take a look and add in the changes needed for Puppet

alexban011:
> Take a look and add in the changes needed for Puppet

Then I'll remove the changes for the buttons and only keep the changes for puppet, and rename the PR/make a new one to only mention the puppet changes

itanasi:
The way I'd do it is merge your branch up to master, resolve the conflicts, then adapt your code and recommit to the same branch so we have the discussion history in the same PR

itanasi:
Right now this sets the Focus between turns for the Puppetee Cities. However it leaves all of the buttons for Citizen Management active, meaning you can still use them to affect Cotizen Assignments.

Check out #6954 for all of the `OnClick()` locations that should be updated. This could be a separate PR if you'd like.

itanasi:
> > Check out #6954 for all of the OnClick() locations that should be updated
> 
> That's already merged :)

Correct, but only disabled the buttons if you're a Spectator. To complete the issue cited, we also want to disable all the buttons if the city is a Puppet

alexban011:
> disable all the buttons if the city is a Puppet

Correct. The fix should be easy enough, I'll add it in a couple hours when I get home

itanasi:
Sorry, just made a PR #6969 that moves a bunch of those buttons around.

yairm210 (approved):

PR #6952, opened 2022-05-26
Add simple logging solution

Azzurite:
The code already contains all kinds of `shouldConsoleLog` variables and `println()`. Adding some kind of common system for that has been figured out decades ago to make sense.

I've really really been debating not writing my own logging class. Of course that topic should've been figured out already, right? And yeah, of course it is. 

I've looked at various logging libraries.

* slf4j+Logback. 1MB added to the apk, just for some shitty logging. Slightly annoying to configure.
* log4j2: No backend to log to Android logcat properly
* Klogging: also no Android backend
* Timber: only Android.

And then I just gave up. This here now is self-written and just a 200 lines simple class. 

Every `println()` call has been replaced by either a `debug()` or `Log.error()` call. `debug()` only logs when the build is not a release build, `Log.error` always logs. 

All `shouldConsoleLog` variables were removed and instead moved to a set of classes to not log by default, located at `Log.disableLogsFrom`. This set can be changed at runtime while debugging with a watch or "Evaluate code", or if you're testing a specific system where you're restarting often, just add a `-DnoLog=class1,class2` list to your VM options or edit the file, whatever.

This is how it looks:

```
2022-05-26T01:37:20.310Z [main] [ImagePacker] Packing textures - 136ms
2022-05-26T01:37:21.743Z [crash-handling-daemon-0 @LoadJSON#2] [Ruleset] Loading ruleset - 58ms
```

as you can see, it nicely logs the timestamp, the thread including coroutine name and the class that made the logging call.

Blerg.

<sub><sup>Btw, perfect number of files changed.</sup></sub>

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -2,6 +2,7 @@ package com.unciv.logic
 
 import com.unciv.Constants
 import com.unciv.UncivGame
+import com.unciv.utils.debug
 import com.unciv.logic.civilization.*
 import com.unciv.logic.map.TileInfo
 import com.unciv.logic.map.TileMap
@@ -20,12 +21,11 @@ import kotlin.collections.HashSet
 
 object GameStarter {
     // temporary instrumentation while tuning/debugging
-    private const val consoleOutput = false
     private const val consoleTimings = false
 
     fun startNewGame(gameSetupInfo: GameSetupInfo): GameInfo {
-        if (consoleOutput || consoleTimings)
-            println("\nGameStarter run with parameters ${gameSetupInfo.gameParameters}, map ${gameSetupInfo.mapParameters}")
+        if (consoleTimings)
+            debug("\nGameStarter run with parameters %s, map %s", gameSetupInfo.gameParameters, gameSetupInfo.mapParameters)
 
         val gameInfo = GameInfo()
         lateinit var tileMap: TileMap
@@ -133,7 +133,7 @@ object GameStarter {
         val startNanos = System.nanoTime()
         action()
         val delta = System.nanoTime() - startNanos
-        println("GameStarter.$text took ${delta/1000000L}.${(delta/10000L).rem(100)}ms")
+        debug("GameStarter.%s took %s.%sms", text, delta/1000000L, (delta/10000L).rem(100))
     }
 
     private fun addPlayerIntros(gameInfo: GameInfo) {
@@ -333,7 +333,7 @@ object GameStarter {
                 var unit = unitParam // We want to change it and this is the easiest way to do so
                 if (unit == Constants.eraSpecificUnit) unit = eraUnitReplacement
                 if (unit == Constants.settler && Constants.settler !in ruleSet.units) {
-                    val buildableSettlerLikeUnits = 
+                    val buildableSettlerLikeUnits =
                         settlerLikeUnits.filter {
                             it.value.isBuildable(civ)
                             && it.value.isCivilian()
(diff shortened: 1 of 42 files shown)


yairm210:
Doesn't libgdx come with a prebuilt solution?
https://libgdx.com/wiki/app/logging

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -133,7 +133,7 @@ object GameStarter {
         val startNanos = System.nanoTime()
         action()
         val delta = System.nanoTime() - startNanos
-        println("GameStarter.$text took ${delta/1000000L}.${(delta/10000L).rem(100)}ms")
+        debug("GameStarter.%s took %s.%sms", text, delta/1000000L, (delta/10000L).rem(100))
```
SomeTroglodyte, line 136: This goes to an old-school C-style print? Won't that balk at longs passed for %s?

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -74,7 +83,7 @@ class MapGenerator(val ruleset: Ruleset) {
             return map
         }
 
-        if (consoleOutput || consoleTimings) println("\nMapGenerator run with parameters $mapParameters")
+        if (consoleTimings) debug("\nMapGenerator run with parameters %s", mapParameters)
```
SomeTroglodyte, line 86: || to &&?

core/src/com/unciv/logic/multiplayer/storage/DropBox.kt
```diff
@@ -41,10 +43,10 @@ object DropBox: FileStorage {
 
                 return inputStream
             } catch (ex: Exception) {
-                println(ex.message)
+                debug("Dropbox exception", ex)
```
SomeTroglodyte, line 46: much better, though now the tag might have been enough

core/src/com/unciv/ui/utils/Log.kt
```diff
@@ -0,0 +1,209 @@
...
+import java.util.regex.Pattern
+
+
+object Log {
+
+    /** Add -DnoLog=<comma-separated-list-of-class-names> to not log these classes. */
+    private val disabledLogsFromProperty = System.getProperty("noLog")?.split(',')?.toMutableSet() ?: mutableSetOf()
+
+    /** Log tags (= class names) **containing** the following Strings will not be logged.  */
+    val disableLogsFrom = if (disabledLogsFromProperty.isEmpty()) {
```
SomeTroglodyte, line 13: Who mutates this? Us when debugging?
If system properties is the main way to choose, I would prefer a more flexible design - you do either system props or defaults - I'd let props switch defaults individually.. Just an idea.
À propos - some dox for the dev "I want a run with Music logged: howTo" might be good to have, even right here in this file I'd find them. Plus wiki probably.

desktop/src/com/unciv/app/desktop/DesktopLogBackend.kt
```diff
@@ -0,0 +1,16 @@
+package com.unciv.app.desktop
+
+import com.unciv.utils.DefaultLogBackend
+import java.lang.management.ManagementFactory
+
+class DesktopLogBackend : DefaultLogBackend() {
+
+    // -ea (enable assertions) or kotlin debugging property as marker for a debug run.
+    // Can easily be added to IntelliJ/Android Studio launch configuration template for all launches.
+    private val release = !ManagementFactory.getRuntimeMXBean().getInputArguments().contains("-ea")
```
SomeTroglodyte, line 10: MrBean? Wow. Learned something new again.

core/src/com/unciv/ui/utils/Log.kt
SomeTroglodyte: I just meant one might wonder as to how one is 'supposed to' do this. Like "can one do run configs for this"?... You know, some people go looking for rules even if they see a viable way already. Might be the rules will work longer than the first shortcut...

yairm210 (approved):
I like the unification of the 'should log' to a single source, I'm not 100% sure that we need all the logging variants but can't see anything bad

PR #6961, opened 2022-05-26
Fix piety complete faith discount

OptimizedForDensity:
Fixes a bug mentioned in the discord by Vsynk and Ravignir where there was an extra space in the bonus string for completing the piety policy tree.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Policies.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Policies.json
@@ -264,7 +264,7 @@
             {
                 "name": "Piety Complete",
                 "uniques": [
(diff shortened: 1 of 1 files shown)


yairm210 (approved):
Looks good but this means we need to add ruleset tests to ensure validity of all religion uniques

SomeTroglodyte (approved):

SomeTroglodyte:
> add ruleset tests

Soon that will be the "No untyped uniqes in builtin rulesets" test - or does one exist already? - need not be part of this PR.

yairm210:
Hilariously, this actually did work until now and is *not* caught by tests, due to getPlaceholderText() activating removeConditionals() which has trim() at the end,
Tests do catch other bad uniques though.

PR #6962, opened 2022-05-26
Refactor: Make CustomFileLocationHelper "dumb", only handling Strings, no GameInfo objects

Azzurite:
Also fixes the `GameInfo.customSaveLocation` to work properly for Android.

Also fixes wrong handling for canceling on Desktop, checking for `selectedFile == null` is not correct, you have to check the return code of `JFileChooser.showSaveDialog`.

Also fixes AWT code not running on the AWT dispatch thread, which at least for me, caused errors.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Policies.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Policies.json
@@ -264,7 +264,7 @@
             {
                 "name": "Piety Complete",
                 "uniques": [
-                    "[Faith] cost of purchasing items in cities [-20]% ",
+                    "[Faith] cost of purchasing items in cities [-20]%",
                     "[+3 Gold, +3 Culture] from every [Holy site]"
                 ]
             }
--- a/android/src/com/unciv/app/AndroidLauncher.kt
+++ b/android/src/com/unciv/app/AndroidLauncher.kt
@@ -11,6 +11,7 @@ import com.unciv.UncivGame
 import com.unciv.UncivGameParameters
 import com.unciv.logic.GameSaver
 import com.unciv.ui.utils.Fonts
+import com.unciv.utils.Log
 import java.io.File
 
 open class AndroidLauncher : AndroidApplication() {
@@ -19,6 +20,7 @@ open class AndroidLauncher : AndroidApplication() {
     private var deepLinkedMultiplayerGame: String? = null
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        Log.backend = AndroidLogBackend()
         customFileLocationHelper = CustomFileLocationHelperAndroid(this)
         MultiplayerTurnCheckWorker.createNotificationChannels(applicationContext)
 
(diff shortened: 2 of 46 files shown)


desktop/src/com/unciv/app/desktop/CustomFileLocationHelperDesktop.kt
```diff
@@ -0,0 +1,116 @@
...
+                        gameData = it.readText()
+                    }
+            } catch (e: Exception) {
+                exception = e
+            }
+            callLoadCallback(loadCompleteCallback, file, gameData, exception)
+        }
+    }
+
+    private fun callLoadCallback(loadCompleteCallback: (SuccessfulLoadResult?, Exception?) -> Unit,
```
yairm210, line 72: Sounds like this is cross platform logic?

android/src/com/unciv/app/CustomFileLocationHelperAndroid.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    override fun loadGame(loadCompleteCallback: (SuccessfulLoadResult?, Exception?) -> Unit) {
+        val callbackIndex = createActivityCallback { uri ->
+            if (uri == null) {
+                loadCompleteCallback(null, null)
+                return@createActivityCallback
+            }
+
+            var gameData: String? = null
+            var exception: Exception? = null
```
yairm210, line 80: This whole logic pipeline also looks cross- platform

- get platform-specific input stream
- attempt to read
- return result depending on successfulness of attempt

yairm210 (changes requested):
Good change overall, I think the main logic can still be simplified to reduce platform-specific code

SomeTroglodyte (approved):
Load dialog _says_ "Save" in caption and OK button.
FYI: Still does `X Error of failed request:  BadWindow (invalid Window parameter)  Major opcode of failed request:  20 (X_GetProperty)` on Mint, so that one wasn't AWT threading, pity.
`SuccessfulLoadResult`: If you're wrapping for a result, why not wrap all three?
"looks cross- platform": I understand both sides: The Java.io.InputStream could be where the interface "draws the line" - but what if a platform decides Java streams aren't cutting it and comes with its own IO layer? Gdx will have a hard time, and we'd notice if that were ever the case as we use Gdx.FileHandle.file(), but still..
"callLoadCallback is cross-platform" - Looking at it, yea, hickups. But the interface "contract" says `Calls the [loadCompleteCallback] on the main thread` - and if Android can guarantee that without any postRunnable thanks to `startActivityForResult` already fulfilling that...

I'll "pre-"approve, wouldn't hurt _me_ (ok yes at least one "conflicts" extra), good approach and OP will have tested successfully on Windowz.

yairm210 (approved):
:)
Most changes are local so if they're not perfect it's simple to make the future changes
But code that needs knowledge of 3 files. in this case, is much harder to refactor once it's settled in

yairm210:
Check that I made the correct conflict resolution :)

SomeTroglodyte:
> Check that I made the correct conflict resolution :)

From the condensed view yes, but I fear that wouldn't cover simple accidental deletions. So - a short skim of the entire PR diff says yes no spurious bits or losses I can see. But Timo would know their code better.

PR #6965, opened 2022-05-27
More UniqueType unit tests and correct some beliefs

OptimizedForDensity:
Implements ```UniqueType.OneTimeGlobalAlert ("Triggers the following global alert: [comment]")``` unique type, mainly to move a hardcoded unique type into ```UniqueTypes.kt```. I'm definitely open to changing the ```UniqueType``` implementations.

This also fixes 3 beliefs by using the correct ```UniqueType``` and also removes an unused one from another belief.

Accordingly, more unit tests to check in the following areas
* Policies
* Policy branches
* Beliefs

These checks are applied to the Vanilla and G&K rulesets to correct for any mistakes. These sort of unit tests would have caught the typo corrected in #6961.

### DIFF
--- a/tests/src/com/unciv/testing/BasicTests.kt
+++ b/tests/src/com/unciv/testing/BasicTests.kt
@@ -2,6 +2,7 @@
 package com.unciv.testing
 
 import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.utils.compression.lzma.Base
 import com.unciv.Constants
 import com.unciv.UncivGame
 import com.unciv.UncivGameParameters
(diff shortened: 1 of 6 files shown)


tests/src/com/unciv/testing/BasicTests.kt
```diff
@@ -190,14 +190,70 @@ class BasicTests {
...
+        for (policy in policies) {
+            for (unique in policy.uniques) {
+                if (!UniqueType.values().any { it.placeholderText == unique.getPlaceholderText() }) {
+                    println("${policy.name}: $unique")
+                    allOK = false
+                }
+            }
+        }
+
+        for (policyBranch in policyBranches) {
```
yairm210, line 207: I think that ruleset.policies already contains all policy branches

yairm210 (approved):
More typing is always an advantage :)

yairm210:
Also, as a follow up to this, we can add a check for policy and belief uniques in the ruleset checks... ;)

tests/src/com/unciv/testing/BasicTests.kt
OptimizedForDensity: Deleted the duplicate check.

SomeTroglodyte:
I still say the `baseRulesetHasNoBugs` test should cover all vanilla json mistakes, and such tests should go via `checkModLinks`, making specific unit tests redundant.

SomeTroglodyte:
"[Forgot to push uniques.md](https://github.com/yairm210/Unciv/pull/6965/commits/db7cc2e63628924c99a84dcd52d9531a10c7d928)" - even less important than omitting an atlas. Will sort itself out on every release automatically, because Yair runs the TFW and for that the uniques.md generator will run unconditionally. Unlike atlas - checks file timestamps. I have excluded that file from any pushes (via my symlinking web)!

OptimizedForDensity:
> I still say the `baseRulesetHasNoBugs` test should cover all vanilla json mistakes, and such tests should go via `checkModLinks`, making specific unit tests redundant.

That definitely makes more sense, especially since `checkModLinks()` contains mechanisms to check for close misspellings. There are still a few untyped uniques (Carthage's movement over mountains, Incas' ignore hill movement cost, Iroquois' forest/jungle treated as roads when in borders, future tech's "Who knows what the future holds?", and the "Aircraft" unit type unique should be the last ones), so once those are in it's probably a good idea to move the unique checks to `checkModLinks()` or a similar function then have the function return errors instead of text warnings.

PR #6975, opened 2022-05-27
disabled focus buttons for puppet cities

alexban011:
this is a continuation of #6946 

disabled changing focus for puppet cities, since that was anyways going to be changed back on the next turn

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CitizenManagementTable.kt
@@ -21,9 +21,12 @@ class CitizenManagementTable(val cityScreen: CityScreen) : Table(BaseScreen.skin
         val resetLabel = "Reset Citizens".toLabel()
         val resetCell = Table()
         resetCell.add(resetLabel).pad(5f)
-        if (cityScreen.canChangeState) {
+        if (cityScreen.canChangeState && !city.isPuppet) {
             resetCell.touchable = Touchable.enabled
-            resetCell.onClick { city.reassignPopulation(true); cityScreen.update() }
+            resetCell.onClick {
+                city.reassignPopulation(true)
+                cityScreen.update()
+            }
         }
         resetCell.background = ImageGetter.getBackground(colorButton)
         add(resetCell).colspan(2).growX().pad(3f)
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
To me this begs the question - `cityScreen.canChangeState ... city.isPuppet` when and why is this ever true/true? OK, there may be a rule situation allowing that, but I can't at the moment remember. But if canChangeState is just a copy of the civ's or worldscreens flag but is misunderstandable in the local context... Replace?

itanasi:
canChangeState is purely if the Civ is the correct (active) Civ. That is, the active owner player. Not a Spectator.

We could change the Boolean to also include isPuppet as a more complete fix within CityInfo

itanasi:
Ooooh, one difference is the Annex City and Raze City Buttons. Obviously if you're a Puppet you can still Annex them.

Also, Change City Name.

But those are the exceptions and I think we can code around that

SomeTroglodyte:
I just meant take a second to consider - you found arguments against (:+1:) - go with whatever is simplest, could well be the current code.

itanasi:
Well, let's give @alexban011 a chance.

That being said, I believe there are also the code in CityStatsTable and CitySpecialistsTable that need to get isPuppet added. The code for the production icons being active, Unassigned Population, and the Auto/Manual Specialists button

yairm210 (approved):

itanasi:
Looks like you got em all now. Nice work

alexban011:
> should be one function

just did. I also used the function in other places where it should be.

itanasi:
> And maybe one can get rid of the `canChangeState` variable and just use `UncivGame.Current.worldScreen.canChangeState` directly in that new function, but I haven't checked all the code. Just a thought.

I did a Show Usage of canChangeState and there's a few spots that don't use isPuppet. And a few other gaps that probably should use isPuppet. Will likely try to clean up after this gets merged unless someone else does it.

Is there any performance reason to locally stash as a variable in CityScreen vs dereference each time? I expect it to he marginal

alexban011:
> Is there any performance reason to locally stash as a variable in CityScreen vs dereference each time?

I think it compiles to the same byte-code, but might be more readable

alexban011:
https://github.com/yairm210/Unciv/blob/8a5f7c7426b0c98bf8bde4fb2c121734288988da/core/src/com/unciv/ui/cityscreen/SpecialistAllocationTable.kt#L43-L45
the add__ functions check for isPuppet but maybe it should be in the if ? 

https://github.com/yairm210/Unciv/blob/8a5f7c7426b0c98bf8bde4fb2c121734288988da/core/src/com/unciv/ui/cityscreen/SpecialistAllocationTable.kt#L68
https://github.com/yairm210/Unciv/blob/8a5f7c7426b0c98bf8bde4fb2c121734288988da/core/src/com/unciv/ui/cityscreen/SpecialistAllocationTable.kt#L94

core/src/com/unciv/ui/cityscreen/CityScreen.kt
```diff
@@ -186,6 +186,10 @@ class CityScreen(
         }
     }
 
+    fun canCityBeChanged(): Boolean {
+        return UncivGame.Current.worldScreen.canChangeState && !city.isPuppet
```
yairm210, line 190: We already reference this and have a local variable of "canChangeState", there shouldn't be a reason to reference the worldscreen twice

alexban011:
> Android Studio -> Tools -> Kotlin -> Show Kotlin Bytecode

wow didn't know that was a thing. Nice

core/src/com/unciv/ui/cityscreen/CityScreen.kt
SomeTroglodyte: FYI: (BaseScreen L23) UncivGame.Current is available as `game`

PR #6980, opened 2022-05-27
Reorder TileGroup layers to draw borders under Pixel Units

### LINKED ISSUE #6837: Feature request: Render map units on top of cultural borders
**Problem:**
Unit sprites render beneath city cultural borders, blocking units from view. See the Archer in the example below.

Example

**Proposed solution:**
Render units on top of the cultural borders so they are not blocked from view.

SomeTroglodyte:
Closes #6837

@Azzurite  - if you would be pleased by a little puzzle (if not ignore): See that bit with the diy bucket sort beginning at TileGroupMap L91? Begs fixing code duplication. I tried this:

<details>

```kotlin
        tileGroups.asSequence()
            .flatMap {
                sequence {
                    yield(it)
                    if (worldWrap) yieldAll(mirrorTileGroups[it.tileInfo]!!.toList())
                }
            }
            .flatMap {
                val positionalSortKey = it.tileInfo.position.run { x + y }.toInt()
                sequence<Pair<Int, Group>> {
                    // now, we steal the subgroups from all the tilegroups, that's how we form layers!
                    yield(        - positionalSortKey to it.baseLayerGroup.apply { setPosition(it.x, it.y) })
                    yield(1000000 - positionalSortKey to it.terrainFeatureLayerGroup.apply { setPosition(it.x, it.y) })
                    yield(2000000 - positionalSortKey to it.miscLayerGroup.apply { setPosition(it.x, it.y) })
                    yield(3000000 - positionalSortKey to it.pixelMilitaryUnitGroup.apply { setPosition(it.x, it.y) })
                    // pixelCivilian can share pixelMilitary 'layer', so same major sort key
                    yield(3000000 - positionalSortKey to it.pixelCivilianUnitGroup.apply { setPosition(it.x, it.y) })
                    yield(4000000 - positionalSortKey to it.highlightFogCrosshairLayerGroup.apply { setPosition(it.x, it.y) })
                    // The above layers are for the visual layers, this is for the clickability of the tile
                    yield(5000000 - positionalSortKey to it)
                    // Aaand units above everything else.
                    yield(7000000 - positionalSortKey to it.unitLayerGroup.apply { setPosition(it.x, it.y) })
                    // This is so the individual textures for the units are rendered together
                    yield(8000000 - positionalSortKey to it.unitImageLayerGroup.apply { setPosition(it.x, it.y) })
                    // city buttons + clickability
                    yield(9000000 - positionalSortKey to it.cityButtonLayerGroup.apply { setPosition(it.x, it.y) })
                }
            }
            .sortedBy { it.first }
            .forEach { this@TileGroupMap.addActor(it.second) }
```

</details>
... replacing lines 91 to 141, just before the `// there are tiles "below the zero",`. Works just fine, and on initial instrumented runs goes from ~25ms to ~6.5ms. But later runs will deteriorate to ~150ms and show significant GC activity. Yup, all those Pairs probably. (And yes the still repeated setPosition could be folded too, but I instrumented first). So - I was out of ideas how to leverage intrinsic sort stability of TimSort to simplify away those wrappers, and to get the scope of the coordinate sort back down to the original element count. Must be something simple I don't see...

### DIFF
--- a/core/src/com/unciv/ui/map/TileGroupMap.kt
+++ b/core/src/com/unciv/ui/map/TileGroupMap.kt
@@ -1,5 +1,6 @@
 package com.unciv.ui.map
 
+import com.badlogic.gdx.graphics.g2d.Batch
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.scenes.scene2d.Group
 import com.unciv.logic.HexMath
@@ -163,6 +164,7 @@ class TileGroupMap<T: TileGroup>(
     fun getMirrorTiles(): HashMap<TileInfo, Pair<T, T>> = mirrorTileGroups
 
     // For debugging purposes
-//     override fun draw(batch: Batch?, parentAlpha: Float) { super.draw(batch, parentAlpha) }
-//     override fun act(delta: Float) { super.act(delta) }
+     override fun draw(batch: Batch?, parentAlpha: Float) { super.draw(batch, parentAlpha) }
+     @Suppress("RedundantOverride")
+     override fun act(delta: Float) { super.act(delta) }
 }
--- a/core/src/com/unciv/ui/map/TileGroupMap.kt
+++ b/core/src/com/unciv/ui/map/TileGroupMap.kt
@@ -1,21 +1,30 @@
 package com.unciv.ui.map
 
-import com.badlogic.gdx.graphics.g2d.Batch
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.scenes.scene2d.Group
 import com.unciv.logic.HexMath
 import com.unciv.logic.map.TileInfo
+import com.unciv.logic.map.TileMap
+import com.unciv.ui.cityscreen.CityTileGroup
 import com.unciv.ui.tilegroups.ActionlessGroup
 import com.unciv.ui.tilegroups.TileGroup
+import com.unciv.ui.tilegroups.WorldTileGroup
 import kotlin.math.max
 import kotlin.math.min
 
+
+/**
+ * A (potentially partial) map view
+ * @param T [TileGroup] or a subclass ([WorldTileGroup], [CityTileGroup])
+ * @param tileGroups Source of [TileGroup]s to include, will be **iterated several times**.
+ * @param tileGroupsToUnwrap For these, coordinates will be unwrapped using [TileMap.getUnWrappedPosition]
+ */
 class TileGroupMap<T: TileGroup>(
-    tileGroups: Collection<T>,
+    tileGroups: Iterable<T>,
     private val leftAndRightPadding: Float,
     private val topAndBottomPadding: Float,
     worldWrap: Boolean = false,
-    tileGroupsToUnwrap: Collection<T>? = null
+    tileGroupsToUnwrap: Set<T>? = null
 ): Group() {
     private var topX = -Float.MAX_VALUE
     private var topY = -Float.MAX_VALUE
@@ -33,7 +42,7 @@ class TileGroupMap<T: TileGroup>(
         }
 
         for (tileGroup in tileGroups) {
-            val positionalVector = if (tileGroupsToUnwrap != null && tileGroupsToUnwrap.contains(tileGroup)) {
+            val positionalVector = if (tileGroupsToUnwrap?.contains(tileGroup) == true) {
                 HexMath.hex2WorldCoords(
                     tileGroup.tileInfo.tileMap.getUnWrappedPosition(tileGroup.tileInfo.position)
                 )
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/ui/tilegroups/ActionlessGroup.kt
```diff
@@ -0,0 +1,39 @@
...
+
+/** A [Group] with [actions], [hit] and scaling effectively disabled, pre-sized.
+ *  @param groupSize [Sets size][setSize] initially */
+open class ActionlessGroupSized(groupSize: Float) : ActionlessGroup() {
+    init {
+        isTransform = false
+        @Suppress("LeakingThis")  // works by setting fields only
+        setSize(groupSize, groupSize)
+    }
+}
```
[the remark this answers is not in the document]
SomeTroglodyte: Didn't think of that

core/src/com/unciv/ui/tilegroups/ActionlessGroup.kt
```diff
@@ -0,0 +1,39 @@
...
+/** A with [touchable] and scaling disabled, pre-sized.
+ *  @param groupSize [Sets size][setSize] initially */
+abstract class GroupSizedTouchableDisabled(groupSize: Float) : Group() {
+    init {
+        isTransform = false
+        touchable = Touchable.disabled
+        @Suppress("LeakingThis")  // works by setting fields only
+        setSize(groupSize, groupSize)
+    }
+}
```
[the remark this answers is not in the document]
SomeTroglodyte: I probably expected reuse and wanted those "profiler names" classes to look similar... Will revisit later.

core/src/com/unciv/ui/map/TileGroupMap.kt
```diff
@@ -149,6 +163,6 @@ class TileGroupMap<T: TileGroup>(
     fun getMirrorTiles(): HashMap<TileInfo, Pair<T, T>> = mirrorTileGroups
 
     // For debugging purposes
-    override fun draw(batch: Batch?, parentAlpha: Float) { super.draw(batch, parentAlpha) }
-    override fun act(delta: Float) { super.act(delta) }
+//     override fun draw(batch: Batch?, parentAlpha: Float) { super.draw(batch, parentAlpha) }
+//     override fun act(delta: Float) { super.act(delta) }
```
yairm210, line 167: Why would you do this... this is uncommented be cause it helps debugging... it doesn't hurt performance at all

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -110,22 +108,25 @@ open class TileGroup(var tileInfo: TileInfo, val tileSetStrings:TileSetStrings,
     @Suppress("LeakingThis")    // we trust TileGroupIcons not to use our `this` in its constructor except storing it for later
     val icons = TileGroupIcons(this)
 
-    class UnitLayerGroupClass:Group(){
-        override fun draw(batch: Batch?, parentAlpha: Float) = super.draw(batch, parentAlpha)
-        override fun act(delta: Float) { // No 'snapshotting' since we trust it wil remain the same
+    class UnitLayerGroupClass(groupSize: Float) : GroupSizedTouchableDisabled(groupSize) {
+        //override fun draw(batch: Batch?, parentAlpha: Float) = super.draw(batch, parentAlpha)
```
yairm210, line 112: This too

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -110,22 +108,25 @@ open class TileGroup(var tileInfo: TileInfo, val tileSetStrings:TileSetStrings,
...
+        override fun act(delta: Float) { // No 'snapshotting' since we trust it will remain the same
             for (child in children)
                 child.act(delta)
         }
     }
 
-    class UnitImageLayerGroupClass:ActionlessGroup(){
-        override fun draw(batch: Batch?, parentAlpha: Float) = super.draw(batch, parentAlpha)
+    class UnitImageLayerGroupClass(groupSize: Float) : ActionlessGroupSized(groupSize) {
+        //override fun draw(batch: Batch?, parentAlpha: Float) = super.draw(batch, parentAlpha)
```
yairm210, line 120: ...

core/src/com/unciv/ui/map/TileGroupMap.kt
SomeTroglodyte: True, I doubted the "doesn't hurt performance at all" but came to no conclusive evidence.

SomeTroglodyte:
Good suggestions, and debug overrides reinstated. One of these days I'll learn how to use them...

letstalkaboutdune:
Thanks so much for implementing this, dev team! <3

PR #6985, opened 2022-05-27
units no longer explore on territory without open border agreement

### LINKED ISSUE #1814: Automated explore still trespasses city-states
**Describe the bug**
Automated explore trespasses if the tile it can explore is located behind the city-state territory.

**To Reproduce**
1. Load [the save game](https://pastebin.com/raw/5UUaucDn)
2. Press "Next turn"
3. Warrior is in Kathmandu

**Expected behavior**
Autoexplore doesn't trespass.

**Screenshots**
image

**Additional context**
The problem is `UnitAutomation.automatedExplore()` call. Algorithm only considers City States when choosing the destination tile, but not when deciding which route to take.

alexban011:
resolves #1814

units no longer pass through territory where they don't have an Open Borders agreement

Steps on how I tested:
- Created small world with 1 bot and 18 CS
- skipped 2 turns to let the CS build their cities (They can build their territory on a tile where you have a unit and they consider that as trespassing)
- put every unit on explore
- skipped a bunch of turns until the entire map is explored

### DIFF
--- a/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
+++ b/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
@@ -583,8 +583,8 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
         if (!assumeCanPassThrough && !canPassThrough(tile))
             return false
 
-        if(tile.getOwner() != null && !unit.civInfo.canEnterBordersOf(tile.getOwner()!!))
-            return false
+//         if(tile.getOwner() != null && !unit.civInfo.canEnterBordersOf(tile.getOwner()!!))
+//             return false
 
         // even if they'll let us pass through, we can't enter their city - unless we just captured it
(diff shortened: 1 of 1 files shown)


alexban011:
good thing we have automated tests

alexban011:
I give up. I fixed the check but now it doesn't work very well. 

The units still explore on city state tiles. Whats the chance of having 6 units explore on a map which is like 60-70% CS territory and have none of them enter CS territory in like 50+ turns? canPassThrough checks for this but it doesn't work every time for some reason...  

Maybe I'll look into it again in the future but right now I really can't tell why units still choose sometimes to enter tile they shouldn't

SomeTroglodyte:
The whole movement system needs a redo from scratch. Wrong paths, wrong predictions, doesn't see speed changes when em/disembarking, no fuzzy 'threat', 'repercussions', 'terrain damage'... factors when choosing routes... Can't _plan_ a swap... Sometimes can't move through friends on a railroad... Look for BFS and - what was it called... Forgot - something with A - in closed discussions.

PR #6990, opened 2022-05-28
Dynamically adjust StatsTable height

itanasi:
Dynamically set the StatsTable height based on whatever is in the lower right corner
image
image
image

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
@@ -20,27 +20,34 @@ import kotlin.math.round
 import com.unciv.ui.utils.AutoScrollPane as ScrollPane
 
 class CityStatsTable(val cityScreen: CityScreen): Table() {
-    private val innerTable = Table()
-    private val outerPane: ScrollPane
+    private val innerTable = Table() // table within this Table. Slightly smaller creates border
+    private val upperTable = Table() // fixed position table
+    private val lowerTable = Table() // table that will be in the ScrollPane
+    private val lowerPane: ScrollPane
     private val cityInfo = cityScreen.city
-    private val statsCell: Cell<ScrollPane>
+    private val lowerCell: Cell<ScrollPane>
 
     init {
         pad(2f)
         background = ImageGetter.getBackground(colorFromRGB(194, 180, 131))
 
         innerTable.pad(5f)
-        innerTable.defaults().pad(2f)
         innerTable.background = ImageGetter.getBackground(Color.BLACK.cpy().apply { a = 0.8f })
+        innerTable.add(upperTable).row()
 
-        outerPane = ScrollPane(innerTable)
-        outerPane.setOverscroll(false, false)
-        outerPane.setScrollingDisabled(true, false)
-        statsCell = add(outerPane)
+        upperTable.defaults().pad(2f)
+        lowerTable.defaults().pad(2f)
+        lowerPane = ScrollPane(lowerTable)
+        lowerPane.setOverscroll(false, false)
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
```diff
@@ -34,10 +36,11 @@ class CityStatsTable(val cityScreen: CityScreen): Table() {
         outerPane = ScrollPane(innerTable)
         outerPane.setOverscroll(false, false)
         outerPane.setScrollingDisabled(true, false)
-        add(outerPane).maxHeight(cityScreen.stage.height / 2)
+        statsCell = add(outerPane)
+        statsCell.maxHeight(cityScreen.stage.height / 2)
```
SomeTroglodyte, line 40: Likely redundant

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
```diff
@@ -34,10 +36,11 @@ class CityStatsTable(val cityScreen: CityScreen): Table() {
...
     }
 
-    fun update() {
+    fun update(height: Float = cityScreen.stage.height / 2) {
```
SomeTroglodyte, line 43: default really used??

SomeTroglodyte:
> [redo Gradle?](https://github.com/yairm210/Unciv/pull/6990/commits/8d6c3232c4eb85b33979765378e3e20ea3415fb7)

A non-collaborator can't go to the failed run and repeat it directly? Show checks - both on the left on the individual run and the top right:
image
... needs elevated permissions? Or not? Other people close and reopen or go draft-ready but I think if any contributor can, then we should wiki the cleaner way - prettier for history, not how to get there...

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
itanasi: No, just for safety

itanasi:
Nope, couldn't find the rerun button. Draft and back may work instead

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
itanasi: You mean setting maxHeight during init()? Are we guaranteed update()?

SomeTroglodyte: Exactly - are we?

SomeTroglodyte (approved):
Looks like it will work just fine.

core/src/com/unciv/ui/cityscreen/CityStatsTable.kt
itanasi: Functionally my guess is yes to populate the tables with values. But unsure. Likely safe to remove

SomeTroglodyte: If after analysis or debugging you're still unsure, leave it. If you're sure, ditch. Me, I just like to prod a little.

Ouaz:
Hello, can I make a request?

Could it possible to keep the stats lines always displayed, even when scrolling inside the table?

Because currently, when scrolling, the first stats line is hidden, and when modifying focus or specialists assignment, it's not possible to see at the same time what it changes in the stats line (it requires to scroll back up each time). With the stats lines always displayed, it will be more convenient to test different focus/specialists settings.

stats_scroll

SomeTroglodyte:
> make a request

Nice gimp/photoshop/whatever work. Thought so too earlier but forgot too quickly to comment.

itanasi:
I think that's doable

itanasi:
Behold
image

PR #6991, opened 2022-05-28
Maintain Fortify bonus after Fortify until Healed

itanasi:
Now if you immediately Fortify after finishing Fortify until Healed, the cumulative Fortify Bonus is maintained

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -106,10 +106,10 @@ class MapUnit {
 
     @Transient
     var canEnterForeignTerrain: Boolean = false
-    
+
     @Transient
     var costToDisembark: Float? = null
-    
+
     @Transient
     var costToEmbark: Float? = null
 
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
More serialized data is probably better than trying to do it without, that would be quite error-prone. But-
Existing saves? I see no perfect way - meaning will catch every game needing update _efficiently_ - except introducing a serialized flag that we deprecate immediately. GameInfo.needsFortificationMigration default true and after a few months drop it again... Ugly... Or leave as is an everybody loses their fortification boni once.

itanasi:
Could add a catch in the end_turn() maybe? Like if the value is still there, store it in the new variable?

SomeTroglodyte:
> end_turn

Just test.. Sounds like maybe, but boni lost for the turn you're loading? Right now they're "stored" in the action field, and you're making another field the relevant one? Not sure.

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -447,7 +449,10 @@ class MapUnit {
 
     fun getFortificationTurns(): Int {
         if (!isFortified()) return 0
-        return action!!.split(" ")[1].toInt()
+        // temporary backwards compatibility
```
SomeTroglodyte, line 452: Actually, the "p(r)op(p)er" backwards compatibility would be an all-units scan called from GameInfo.setTransients and homed in `object BackwardCompatibility` -> easier to find out-aged ones. Just saying. Sorry I didn't read your commits earlier. And there I'd probably `Regex("""^Fortify\s+(\d+)""")` and use the captured group or so.

itanasi:
Review request

PR #6992, opened 2022-05-28
Add Catalan translation and provide some strings.

J0anJosep:
I have copied the template.properties and started the Catalan localisation.

I will continue the translation after this PR, but I want to know whether I am doing it right.


SomeTroglodyte:
For a fresh start, that looks just fine.

There's a few elements for entirely new languages:
- The game will not offer languages without an entry in completionPercentages.properties
- This file is written by the TranslationFileWriter
- The TranslationFileWriter will only process Languages it knows about
- This is I believe a cycle that will need kickstarting, meaning the first entry in completionPercentages.properties will need to be pushed once. Just append Catalan = 0.
- For language-specific sorting and fastlane to work, an entry in GameSettings.LocaleCode is needed - the one for Catalan already exists: `Catalan("ca", "ES")` - and your language name and thus file name must match including case -> all fulfilled.
- The language picker in Options will want to show a flag... And new Art needs atlas regeneration, and attribution unless already covered by credits.md.

All of these can be completed by someone else when the time comes.

But there's one more detail: The template file isn't everything that needs translating. The mentioned TranslationFileWriter will add in the Ruleset stuff from the json files. So there's a _lot_ more work than you saw, but arguably the template stuff is a little more important for playability as it contains User Interface text.

You can run TranslationFileWriter locally, even from a jar: It's a button in Options - from the main menu hit O, F12. _Do_ make a backup copy of your file before doing so! From above, do the completionPercentages thing only, and if you can choose Catalan in Languages, and there is at least _one_ actual translation, TFW should process the file.

... wait, isn't all that in the wiki?

J0anJosep:
> For a fresh start, that looks just fine.
> 
> There's a few elements for entirely new languages:

Thank you. I will check these.

SomeTroglodyte:
... and the answer to "isn't all that in the wiki" is no - not yet ...

I wish you luck and courage, getting a new language to the point it's worthwhile playing with is no small feat. Concentrate on what is readily visible - and for that, actually running the game with your file _and_ having the TranslationFileWriter additions in can help a lot.

We'll help you integrate this into the releases, but you'll surely agree that makes sense only when we can hope someone else out there will like it.

J0anJosep:
I managed to run it as suggested and the missing strings have been added to the localisation file.

Tomorrow I will deal with the flag issue.

SomeTroglodyte:
I didn't find it (Andorra is there, but no Catalunya) on the source that's already in the credits (https://www.flaticon.com/packs/countrys-flags) - I'd probably gimp four red stripes on yellow background myself and list "Created by Unciv team" in the credits... Maybe add the Equus africanus asinus too.

J0anJosep:
> Equus africanus asinus

I also looked for it in flaticon and I didn't find any I like. I plan to create a svg file (and then export to png) with the four stripes and keeping the format of preexisting language flags. Donkey image will be left out so we have less complaints about political issues. :P

SomeTroglodyte:
El burro is the political detail?? Bah, politics. To think that one beautiful mountain with its iron caused so much trouble over the centuries... But the flags on the summit cross did indeed not have any burro last time I was up there.

J0anJosep:
Ok!

I have added the flag and edited the credits file. Please review the edit in docs/credits.md, as I don't know if I've done it correctly.

Also, I added the svg file of the Catalan flag. Maybe you want to keep that (and maybe move it to another folder) as it can be used for drawing other flags in the future.

I think the PR can be reviewed now.

SomeTroglodyte:
Wow. But I _can_ still nitpick... Let's see...
There's a [ImagesToNotAddToGame](https://github.com/yairm210/Unciv/tree/master/android/ImagesToNotAddToGame) folder where such a source asset could go, excluded from redistribution.
"licenced" - well, the rest of our English is Yankee, so "licensed"

:smile_cat:

SomeTroglodyte (approved):

SomeTroglodyte:
Looks like you're really aiming for 99% :100: :+1: [^1].

[^1]: Hint: The lists of city names per Nation - most can go translation = original, which one can copy from other language files. Individual translations for those only make sense when there really are local cultural expectations.

J0anJosep:
> Looks like you're really aiming for 99% 100 +1 [1](#user-content-fn-1-684b998e6feb6f8da1a2fc9c04435bfb).
> ## Footnotes

I want to play it in Catalan :P 

I force-pushed. I moved the svg file and changed licenced->licensed. So you should check again your approved review status.

PR #6993, opened 2022-05-28
Pillaging certain improvements may loot gold or other stat resources

OptimizedForDensity:
Improvements may now be given a `PillageYieldRandom` or `PillageYieldFixed` unique type which lets units loot different stats from them. Gold, culture (for now), faith, and science that are looted are transferred to the pillaging civ while any looted food or production is transferred to the nearest owned city (if found). All (lootable) vanilla improvements will only yield gold but modded ones may have more interesting yields. Ticks off a box in #4697. The pillaged victim now also receives a notification that they've been pillaged and clicking this notification takes them to the tile in question.

Example usage is seen here
<Details>

```

{
"name": "Pasture",
"turnsToBuild": 8,
"techRequired": "Animal Husbandry",
"uniques": ["[+1 Food] <after discovering [Fertilizer]>", "Can only be built to improve a resource", "Pillaging this improvement yields approximately [+10 Gold]", "Pillaging this improvement yields [+10 Food, +10 Culture]"],
"shortcutKey": "P"
}
```

</Details>

The math for randomized pillaging works similar to how it does in Civ 5. Two `Random.nextInt(statAmount + 1)` calls are made then the results added. This results in a triangular distribution with a mean value equal to `statAmount` defined in the JSON file and a possible outcome between 0 and 2 * `statAmount` resource (both endpoints are inclusive) on any given pillage. Fixed pillage yields are straightforward.

Improvements in the Civilopedia will list the lootable resources on the improvement's page and the Pillaging tutorial section has been tweaked to let the player know that resources can be pillaged.

Improvements definitions will throw an error for a negative stat amount (since trying to use `Random.nextInt` with a negative error crashes Java) and will complain with a warning if the improvement has `UniqueType.Unpillagable` and either of the `PillageYield` unique types. A unit test for looting was also added. It checks to make sure that gold is transferred to the pillaging civ and that pillaged food is transferred to a nearby city.

The original Civ 5 code can be found here for reference and the default ruleset values are taken from Civ 5 XML files
https://github.com/Gedemon/Civ5-DLL/blob/master/CvGameCoreDLL_Expansion1/CvUnit.cpp#L5455

In the future a `UniqueType.PillageModifier` could be added so that certain units/civs/promotions loot more resources. The AI also does not factor in pillageable gains when planning actions so that should be added at some point.

Can loot food too


yairm210:
This should be done with a unique that looks like "Yields [stats] when pillages" to allow for A. other stats and B. type checking
We don't add extra fields for moddable attributes, instead we use uniques to allow modifications :D

OptimizedForDensity:
That's a better idea (could loot culture or science, or maybe for a mod you could lose faith but gain gold by looting a temple-style improvement). I'll probably go with a `PillageYield` unique type that includes randomization then a `PillageYieldFixed` that results in the same looting yield each time so that there's finer control over how improvement looting works.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -296,6 +302,81 @@ object UnitActions {
...
+                        unit.civInfo.addStat(stat.key, looted)
+                        globalPillageYield[stat.key] += looted.toFloat()
+                    }
+                    Stat.Food, Stat.Production -> {
+                        if (closestCity != null) {
+                            closestCity.addStat(stat.key, looted)
+                            toCityPillageYield[stat.key] += looted.toFloat()
+                        }
+                    }
+                    else -> {}
```
yairm210, line 330: Instead of having this logic twice, have it once.
You can do this in a couple of ways, I would probably take all random stats, randomize them, then add them to static stats and iterate final stat list.

OptimizedForDensity: Ah I see what you mean.

OptimizedForDensity:
> * What went wrong:
Execution failed for task ':android:lintAnalyzeDebug'.
> Couldn't read file content: '/home/runner/work/Unciv/Unciv/android/assets/.attach_pid2577'.

This has been happening to me but I have no idea what's causing it (last time it happened it resolved itself)

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
```diff
@@ -5,100 +5,90 @@
...
 	},
 	{
 		"name": "Trading post",
 		"terrainsCanBeBuiltOn": ["Plains","Grassland","Desert","Hill","Tundra","Forest","Jungle","Flood plains"],
 		"gold": 1,
 		"turnsToBuild": 7,
-        "pillageGold": 20,
 		"techRequired": "Guilds",
-		"uniques": ["[+1 Gold] <after discovering [Economics]>"],
+		"uniques": ["[+1 Gold] <after discovering [Economics]>", "Yields on average [+20 Gold] when pillaged"],
```
SomeTroglodyte, line 39: Sorry, not sure this is the best possible wording. And as you know changing opinions after a release gets tricky... "Yields aproximately...", "Pillaging yields [20 Gold] on average"? I'm no native brit or yankee, but I hesitated an instant 'does the on refer to  yields or to average?'... "You can loot an average of X when pillaging"? Not sure. OK, I'll not veto any choice, just do what feels best.

SomeTroglodyte:
> > * What went wrong:
> >   Execution failed for task ':android:lintAnalyzeDebug'.
> >   Couldn't read file content: '/home/runner/work/Unciv/Unciv/android/assets/.attach_pid2577'.
> 
> This has been happening to me but I have no idea what's causing it (last time it happened it resolved itself). The code compiles and runs perfectly on my end.

Not new. Re-running now.

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
OptimizedForDensity: It's definitely difficult trying to get the point across in a succinct way that this unique type has randomization in (a non-uniform distribution too). I'll play around with it a bit but I like your suggestion of renaming the uniques to "Pillaging yields []..." since that also puts "pillaging" at the beginning of the line and makes it immediately clear to the reader what the line is talking about when it starts listing a bunch of stats.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
SomeTroglodyte: And wasn't there some code to support which stats have a global pool and can therefore be used for buying stuff? There may be something reusable there. And don't forget Stats is a nice and fast aggregator, so you could pull a Stats() add your loot to it _then_ forward the sums to civ/city pools _and_ use it to feed notifications - with consistent ordering regardless of unique order. (don't use values, that one includes zeroes. Normal iteration or asSequence won't)

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
```diff
@@ -5,100 +5,90 @@
...
+		"uniques": ["Can only be built to improve a resource", "Can only be built on [Water] tiles", "Yields on average [+20 Gold] when pillaged"],
 		"shortcutKey": "P"
 	},
 	{
 		"name": "Pasture",
 		"turnsToBuild": 8,
-        "pillageGold": 10,
 		"techRequired": "Animal Husbandry",
-		"uniques": ["[+1 Food] <after discovering [Fertilizer]>", "Can only be built to improve a resource"],
+		"uniques": ["[+1 Food] <after discovering [Fertilizer]>", "Can only be built to improve a resource", "Yields on average [+10 Gold] when pillaged", "Yields [+10 Food, +10 Culture] when pillaged"],
```
SomeTroglodyte: Is that your test? "Cul"-ture from manure?

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -133,10 +132,6 @@ class TileImprovement : RulesetStatsObject() {
         val statsDesc = cloneStats().toString()
         if (statsDesc.isNotEmpty()) textList += FormattedLine(statsDesc)
 
-        if (pillageGold > 0) {
-            textList += FormattedLine("Average pillageable gold: ¤[$pillageGold]", link="Tutorial/Pillaging")
```
SomeTroglodyte: Don't know a cheap solution to get that link back - except [civilopediaText](https://yairm210.github.io/Unciv/Other/Miscellaneous-JSON-files/#civilopedia-text)

SomeTroglodyte:
Wait a moment - this is not the first "looting" feature. Ships attacking cities loot, too - or was it Privateers only? Look for synergies. Somehow my brain says there already _was_ a decision to _not_ support looting stats without a global account because too complicated. Could be a while ago. Also look for a specific UniqueParameterType??

OptimizedForDensity:
I see what you mean now with the `[plunderableStat]` UniqueParameterType. I'm not familiar with any discussion about whether food and production should be plunderable/lootable or not, but I figure since production can already be "plundered" in a sense by cutting down a forest I figured that giving the food/production to the nearest friendly city would be a good way to represent lootable food/production.

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
OptimizedForDensity: Yeah it was haha, you can see it in the screenshot in the OP. Who knows, maybe it's some artistic manure.

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
OptimizedForDensity: Maybe by adding an optional link argument for `UniqueType(...)`? For most uniques this argument would be blank and the link would be generated by `getUniqueLink()` in `CivilopediaText.kt` as it's done now but for uniques with non-null link arguments it would use that link instead.

OptimizedForDensity: A quick and dirty implementation of what I mean.
<Details>

In UniqueTypes.kt
```
enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags: List<UniqueFlag> = emptyList(), val civilopediaLink: String? = null) {

[much later...]

PillageYieldRandom("Pillaging this improvement yields approximately [stats]", UniqueTarget.Improvement, civilopediaLink = "Tutorials/Pillaging"),
```

Then in CivilopediaText.kt

```
private fun getUniqueLink(unique: Unique): String {
            val ruleSet = getCurrentRuleset()
            if (allObjectNamesCategoryMap == null || rulesetCachedInNameMap !== ruleSet)
                allObjectNamesCategoryMap = initNamesCategoryMap(ruleSet)
            if (unique.type?.civilopediaLink != null) return unique.type.civilopediaLink
            for (parameter in unique.params + unique.conditionals.flatMap { it.params }) {
                val category = allObjectNamesCategoryMap!![parameter] ?: continue
                return category.name + "/" + parameter
            }
            return ""
        }
```
</Details>

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
OptimizedForDensity: Updated the unique text to be more clear.

core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
SomeTroglodyte: > getUniqueLink()

That was done _long_ before we had the IRulesetObject class hierarchy. Could be done much better nowadays I presume.
Not sure I _can_ like any additional code for links - needs a clear picture which I don't see. Your suggestion _sounds_ a little klugde-y and as if it won't cover all bases, but I may be wrong. The civilopediaText idea at the moment costs nothing except a little json right where the unique itself is, and a revert if we later decide otherwise. And you can give it colors! Any any icon available (if I remember right)!

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -296,6 +299,64 @@ object UnitActions {
...
+            if (tile.getOwner() != null)
+                unit.civInfo.addNotification("We have looted [${pillageYieldNotificationString(toCityPillageYield)}] from a [${improvement.name}] which has been sent to [${closestCity.name}]", tile.position, NotificationIcon.War, tile.getOwner()!!.civName)
+            else
+                unit.civInfo.addNotification("We have looted [${pillageYieldNotificationString(toCityPillageYield)}] from a [${improvement.name}] which has been sent to [${closestCity.name}]", tile.position, NotificationIcon.War)
+        }
+        if (globalPillageYield.values.sum() > 0) {
+            if (tile.getOwner() != null)
+                unit.civInfo.addNotification("We have looted [${pillageYieldNotificationString(globalPillageYield)}] from a [${improvement.name}]", tile.position, NotificationIcon.War, tile.getOwner()!!.civName)
+            else
+                unit.civInfo.addNotification("We have looted [${pillageYieldNotificationString(globalPillageYield)}] from a [${improvement.name}]", tile.position, NotificationIcon.War)
```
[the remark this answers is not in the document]
OptimizedForDensity: I see, thanks for the reference. I have a proof of concept that I'm working on the PR for, although since I'm not as familiar with what your vision is for the notification manager, my attempt focuses on using object oriented approaches for reusing code and moving the notification text to a more central location.

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -694,6 +694,22 @@ class Ruleset {
                     RulesetErrorSeverity.Warning
                 )
             }
+            for (unique in improvement.uniqueObjects) {
+                if (unique.type == UniqueType.PillageYieldRandom || unique.type == UniqueType.PillageYieldFixed) {
+                    val params = Stats.parse(unique.params[0])
```
SomeTroglodyte, line 699: We're in the mod checker. Stats.parse will throw (null pointer no less) if fed garbage. Better Stats.isStats first.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -296,6 +301,66 @@ object UnitActions {
...
+        }
+        for (unique in improvement.getMatchingUniques(UniqueType.PillageYieldFixed)) {
+            for (stat in unique.stats) {
+                pillageYield.add(stat.key, stat.value)
+            }
+        }
+
+        for (stat in pillageYield) {
+            when (stat.key) {
+                Stat.Gold, Stat.Faith, Stat.Science, Stat.Culture -> {
```
SomeTroglodyte, line 326: `in Stat.statsWithCivWideField`

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -296,6 +301,66 @@ object UnitActions {
...
+                    unit.civInfo.addStat(stat.key, stat.value.toInt())
+                    globalPillageYield[stat.key] += stat.value
+                }
+                Stat.Food, Stat.Production -> {
```
SomeTroglodyte, line 330: Can already be the else branch -> addStats is neutral on Happiness

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -296,6 +301,66 @@ object UnitActions {
...
+                    if (closestCity != null) {
+                        closestCity.addStat(stat.key, stat.value.toInt())
+                        toCityPillageYield[stat.key] += stat.value
+                    }
+                }
+                else -> {}
+            }
+        }
+
+        if (toCityPillageYield.values.sum() > 0 && closestCity != null) {
```
SomeTroglodyte, line 340: Stats.isEmpty() is the same as values.sum() > 0 **if** you're sure not to have negative values in there.

tests/src/com/unciv/uniques/GlobalUniquesTests.kt
```diff
@@ -339,4 +341,26 @@ class GlobalUniquesTests {
...
+        civInfo.addGold(-civInfo.gold) // reset gold just to be sure
+
+        val testImprovement = game.createTileImprovement("Pillaging this improvement yields [+20 Gold, +11 Food]")
+        tile.improvement = testImprovement.name
+        val unit = game.addUnit("Warrior", civInfo, tile)
+        unit.currentMovement = 2f
+
+        val pillageAction = UnitActions.getPillageAction(unit)
+        pillageAction?.action?.invoke()
+        Assert.assertTrue(civInfo.gold == 20)
```
SomeTroglodyte, line 362: Could do with a little message? The message could also tell when the getPillageAction went "wrong". Unimportant.

core/src/com/unciv/models/ruleset/Ruleset.kt
OptimizedForDensity: Good call.

tests/src/com/unciv/uniques/GlobalUniquesTests.kt
OptimizedForDensity: Done.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -296,6 +301,66 @@ object UnitActions {
...
+            if (tile.getOwner() != null)
+                unit.civInfo.addNotification(pillagerLootGlobal, tile.position, NotificationIcon.War, tile.getOwner()!!.civName)
+            else
+                unit.civInfo.addNotification(pillagerLootGlobal, tile.position, NotificationIcon.War)
+        }
+    }
+
+    // function that removes the icon from the Stats object since the circular icons all appear the same
+    // delete this and replace above instances with toString() once the text-coloring-affecting-font-icons bug is fixed
+    private fun pillageYieldNotificationString(stats: Stats): String {
```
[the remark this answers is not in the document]
OptimizedForDensity: Good call, at least this way others could use it if they're running into coloration issues.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
OptimizedForDensity: Done.

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -694,6 +694,22 @@ class Ruleset {
...
+                    if (params.values.any { it < 0 }) lines.add(
+                        "${improvement.name} cannot have a negative value for a pillage yield!",
+                        RulesetErrorSeverity.Error
+                    )
+                }
+            }
+            if ((improvement.hasUnique(UniqueType.PillageYieldRandom) || improvement.hasUnique(UniqueType.PillageYieldFixed))
```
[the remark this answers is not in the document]
OptimizedForDensity: I think I see what you're saying. I changed the `hasUnique` function call to use the `StateForConditions.IgnoreConditionals` flag.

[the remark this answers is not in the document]
SomeTroglodyte: Conditionals can be complicated. Just learned a little about them :cat:

core/src/com/unciv/models/ruleset/Ruleset.kt
SomeTroglodyte: But... but... that's an Error! Imagine finding `Pillaging this improvement yields [42 Petunias]` and _not_ nagging about it! Even if they `continue` to fall towards M...

core/src/com/unciv/models/ruleset/Ruleset.kt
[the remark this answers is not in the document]
SomeTroglodyte: Actually, the Unpillagable one wouldn't really work with a conditional anyway, but pillaging - a mod might want to forbid that , say, if below -19 happinyess, and it would work, right? But doesn't hurt.

SomeTroglodyte (approved):

PR #6994, opened 2022-05-28
Update EditorConfig - keep existing credits lists intact

SomeTroglodyte:
Tiny thing - in a recent PR (#6856) autoformat touched a lot of lines, which IMO are fine as they were.

Also - should we keep the continuation indents for if conditions and expression bodies? My impression is in theory I like them in these two cases but continuation still seems to apply in situations where I don't like them - but I haven't kept an example around. Input?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -74,3 +74,6 @@ ij_properties_spaces_around_key_value_delimiter = false
 
 [{*.yaml, *.yml}]
 indent_size = 2
+
+[*.md]
+ij_markdown_force_one_space_after_list_bullet = false


SomeTroglodyte:
> tell which lines

There was only one markdown in there - the credits. It is taking out extra blanks from lines such as `-   Arrows.svg by Intralexical (@will-ca), CC0.` - three blanks after the list dash.

> think they're good

:+1: Yes. This should stabilize in a month if we all give feedback. I vow to pay better attention.

SomeTroglodyte:
OK - in that case, we'll have chaff in PR's touching credits for a while. Would be fine too.
(Anyway I think that flag is bugged, it seems to need `ij_markdown_force_one_space_between_words = false` too for what I indended)

PR #6996, opened 2022-05-29
Territory stolen notification rewording

yairm210:
"the Aztecs has stolen your territory" is just wrong
As for the PR, is due to removing the exclamation mark, ymmv...


SomeTroglodyte (approved):
Still sounds like the scoundrels have absconded with the entirety on someone's territory, not just a few tiles...

Also, the day will come when English.properties: `[civName] has stolen your territory! = Your territory has been stolen by [civName]` will be simpler.. :upside_down_face: 

As for the 0x21: Those languages I understand will be perfectly all right.

SomeTroglodyte:
I think since he's improving the english, and most translators found something much better sounding anyway, so it's fine.

yairm210:
Precisely

yairm210:
Not sure what broke but simplest fix is to redo

PR #7001, opened 2022-05-29
Two new Unit Tests

SomeTroglodyte:
Two new unit tests! Otherwise, nothing to see here. Merge or approve and move on... And for heathen's sake don't look at the changed file count!

<details><summary>Spoiler Warning</summary>

To test, include JVM options in your run config: -Dmonth=12 to see a wintry map or -DeasterEgg=Easter to see the "Easter" easter-eggs ("Samhain" or "Xmas" do stuff too). Otherwise depends on system clock.

image
image
image
image
</details>

Many images are not done carefully. Can be fixed later if this gets any interest. Also, there's likelyhood they'll need replacements with better licenses:
Image sources are a bit dubious, but I found no fitting art on our usual sites. None of the new ones I pulled from should be visited without a good blocker, and their licensing policies are slightly dubious too: Clear on free use but no compatibility to established license templates to be seen. (https://www.freepikcompany.com/legal#nav-freepik, https://pixabay.com/service/license/) - suggestions for sites offering such clipart under CC* are welcome.

Template lines, too, will wait until some bias is shown whether this is desirable.

Building the "mods" in-memory without them appearing in RulesetCache or on disk (or offering them as actual repo based mods) is a conscious decision - github-proficient people will always be able to see this, but within limits I want it to still be a surprise for most. The package choice the same - didn't want a pagage named eastereggs or put this into any of the other existing. ui/utils still too full.

</details>


yairm210:
To be honest playing a game with a map littered with these things looks extremely distracting.
Were it just the main menu I'd have no problem but people spend hours on the map and it can get real old real fast

yairm210:
Is it? I haven't tried - can players not choose these tilesets from options? /because IIRC ImageGetter gets available tilesets by checking images under "/tilesets" and these would qualify

SomeTroglodyte:
> it _is_ just

It is just. Or unjust. Or just a joke. Yes, just just fits as just a description.

Nah did't even have the idea to make these their own tilesets. They map into... Wait a minute. Surely breaks when 5Hex is set as permanent audiovisual mod... Maybe add an alias, but there are others out there I don't know about. Hmmm. Hey @ravignir need the gimps for 640px tiles?

SomeTroglodyte:
But - right now it's go search eggs - outside - time. Though it's not eggs per se.

alexban011:
Personally I think if the graphics would fit with the rest of the graphics I think it would be really cool. 
Considering the time these things would actually be visible I doubt they'd get old (its 1 day/year right?) 

If this was in games and not just main menu it could maybe also replace ruins with easter eggs / gifts, christmas themed music and other stuff like this

yairm210:
Sure let's try it

PR #7009, opened 2022-05-30
Fix EventBus not being able to listen to events with more than one level of inheritance

Azzurite:
I incorrectly assumed that the `superTypes` property included all super types recursively... I even thought I tested it... But apparently I didn't.

### DIFF
--- a/tests/src/com/unciv/logic/event/EventBusTest.kt
+++ b/tests/src/com/unciv/logic/event/EventBusTest.kt
@@ -6,7 +6,8 @@ import org.junit.Test
 import java.lang.ref.WeakReference
 
 class EventBusTest {
-    open class Parent : Event
+    open class Parent2 : Event
+    open class Parent : Parent2()
     class Child : Parent()
 
     @Test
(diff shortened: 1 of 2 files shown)


tests/src/com/unciv/logic/event/EventBusTest.kt
```diff
@@ -6,7 +6,8 @@ import org.junit.Test
 import java.lang.ref.WeakReference
 
 class EventBusTest {
-    open class Parent : Event
+    open class Parent2 : Event
```
SomeTroglodyte, line 9: GrandParent!!

tests/src/com/unciv/logic/event/EventBusTest.kt
```diff
@@ -20,6 +21,17 @@ class EventBusTest {
...
 
+    @Test
+    fun `should receive parent event when child event with two levels of inheritance is sent`() {
+        val events = EventBus.EventReceiver()
+        var callCount = 0
+        events.receive(Parent2::class) { ++callCount }
+
+        EventBus.send(Child())
+
+        assertThat(callCount, `is`(1))
```
SomeTroglodyte, line 32: What's wrong with the Assert.assertTrue(callCount == 1) used in other tests? Is that hamcrest core a new library?

yairm210 (approved):
Looks okay to me. @SomeTroglodyte how insistent are you with regards to changes?

PR #7011, opened 2022-05-30
WIP Translation PTBR

LincolnVS:


SomeTroglodyte:
FYI: The fastlane files added here are redundant, but not harmful. They will be automatically generated/overwritten from the entries you added, too. The idea was that not every translator would understand fastlane but the're familiar with the other system, even if all the `\n` get to be annoying...

LincolnVS:
> FYI: The fastlane files added here are redundant, but not harmful. They will be automatically generated/overwritten from the entries you added, too. The idea was that not every translator would understand fastlane but the're familiar with the other system, even if all the `\n` get to be annoying...

Ah thanks, I got it. But just to confirm, is it okay to leave it that way?

In addition, I plan to finish about 100 more translations today and make the official merge request. After that, is the translation percentage done automatically, or do I have to do it?

SomeTroglodyte:
> okay to leave it that way, do I have to do it

Sim e não

LincolnVS:
It was done, more than 150 new translations and more than 10 corrections of translations that were wrong or did not fit in during the game.

PR #7013, opened 2022-05-30
Allow non-SSL-encrypted HTTP traffic & warn Dropbox users

Azzurite:
To keep multiplayer accessible for most people, adding the possibility to not have to set up a SSL certificate when hosting a server is necessary. It's not like we transfer any sensitive information that way. Android's forcing of SSL by default of course makes sense in a normal app development context, but not in our case.

Then I also adjusted the docs to remove that part and add a few things.

And of course, I added a warning for when you activate multiplayer in the new game screen, explaining the problems with Dropbox.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -15,7 +15,9 @@
             UnusedAttribute: Since we're targeting a range of API levels, it's OK that e.g. appCategory will be ignored by OLD devices
             IconDensities: See https://developer.android.com/training/tv/start/start#banner for the banner attribute, where they recommend supplying only one density.
     -->
+    <!-- Cleartext traffic permission solely so users can host their own server without setting up https -->
     <application
+        android:usesCleartextTraffic="true"
         android:allowBackup="true"
         android:icon="@mipmap/uncivicon"
         android:roundIcon="@mipmap/uncivicon_round"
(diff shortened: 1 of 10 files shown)


core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -103,6 +109,9 @@ class GameOptionsTable(
             {
                 gameParameters.isOnlineMultiplayer = it
                 updatePlayerPickerTable("")
+                if (it) {
```
SomeTroglodyte, line 112: Sounds very pleasing!

android/AndroidManifest.xml
```diff
@@ -15,7 +15,9 @@
             UnusedAttribute: Since we're targeting a range of API levels, it's OK that e.g. appCategory will be ignored by OLD devices
             IconDensities: See https://developer.android.com/training/tv/start/start#banner for the banner attribute, where they recommend supplying only one density.
     -->
+    <!-- Cleartext traffic permission solely so users can host their own server without setting up https -->
     <application
+        android:usesCleartextTraffic="true"
```
SomeTroglodyte, line 20: I thought I tested this and it did nothing. Something to do with minSdk, targetSdk and maybe even older proguard apilevel?

SomeTroglodyte: Then it was AVD's getting crippled networks. Or my routing here. No matter.

yairm210 (approved):

PR #7014, opened 2022-05-30
Notification manager proof of concept

OptimizedForDensity:
This is a quick proof of concept of a notification manager that @Azzurite talked about in #6993 and so what I wrote here is my interpretation of how it _could_ be implemented. This is more of a demonstration than a definitive implementation (especially since I copy and pasted code from #7009 directly into this PR).

The idea is that we can use the new `EventBus` that listens for events and has customizable listeners which can process the events as we need them to in order to handle notifications on the world screen. This `NotificationManager` listens for `Event`s that implement the `INotificationEvent` interface (which has an abstract `createNotification()` function used to created the notifications to all necessary parties). This lets us create dedicated notification-creating code beyond calling `addNotification(...)` that can be reused and inherited in an objected-oriented manner while keeping the gameplay code focuses on gameplay mechanics. Here I've got an example hierarchy of interfaces and classes that help create notifications including supporting code that is reused between notification classes. What do you guys think of this as a way of managing notification code?

### DIFF
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -9,6 +9,7 @@ import com.badlogic.gdx.utils.Align
 import com.unciv.logic.GameInfo
 import com.unciv.logic.GameSaver
 import com.unciv.logic.civilization.PlayerType
+import com.unciv.logic.event.NotificationManager
 import com.unciv.models.metadata.GameSettings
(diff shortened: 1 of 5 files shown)


OptimizedForDensity:
Closing since while I have a greater understanding of how a notification manager like this would be useful, I don't have the time to work on it.

PR #7020, opened 2022-05-31
Friends list

### LINKED ISSUE #6716: QoL - Friends list for remembering player IDs

alexban011:
resolves  #6716 
Draft PR discussion in #6854 

Adds a friends list in the multiplayer screen which lets you remember people's IDs 
If the friends list is not empty, a new button will appear when creating a new game, letting you select a friend from the list.

<details>
 <summary>Video</summary>
https://user-images.githubusercontent.com/21990802/171163956-3d9b0619-808d-4d9f-bb51-da665158704d.mp4
</details>

<details>
 <summary>This is how it looks when vertical</summary>
Screenshot_20220531_142657
</details>


alexban011:
why did it fail here but not on #6854

SomeTroglodyte:
Yup, looks _quite_ well done. :+1:

When viewing the video I also had a few ideas, but that's UI and no reason to not approve - UI can be fiddled later. Structural as in not reusing an existing widget well-matched to a usecase, yes OK. (But - there's some skeletons in that direction too: Several pickers select from a list and are either sorely missing a _visual_ feedback what was just klicked, only adding a description and maybe updating button text - or they roll their own. Needs unification, as does... Well, this one could do with a highlight too, not a biggie) Other UI thing I noticed: Player selection is getting a bit cramped. Could look into more compact buttons so there's still a chance to see 4-5 players at once... Again, not critical.

But I'll want a look at data structures, when I'm not so tired...

alexban011:
> > why did it fail here but not on https://github.com/yairm210/Unciv/pull/6854
> 
> [But, it also failed there?](https://user-images.githubusercontent.com/2777158/171170235-b29e55a4-9996-40f4-82e4-075fe649a04d.png)
> 

Either I'm blind or Github is playing tricks on me :)) will fix it soon when I find out what is wrong. Maybe a merge will fix

SomeTroglodyte:
> Github is playing tricks

... And I restarted the job in the mean time. AndroidLintDebug is not fixed from source code. Just retrying did "make it green".

alexban011:
> ... And I restarted the job in the mean time. AndroidLintDebug is not fixed from source code. Just retrying did "make it green".

Nice, could have sworn it was green in the Draft.

core/src/com/unciv/ui/newgamescreen/PlayerPickerTable.kt
```diff
@@ -241,6 +266,119 @@ class PlayerPickerTable(
...
+        val friendListWithRemovedFriends = friendList.friendList.toMutableList()
+        for (index in gameParameters.players.indices) {
+            val currentFriendId = previousScreen.gameSetupInfo.gameParameters.players[index].playerId
+            friendListWithRemovedFriends.remove(friendList.getFriendWithId(currentFriendId))
+        }
+        return friendListWithRemovedFriends.asSequence()
+    }
+}
+
+private class FriendPickerPopup(
```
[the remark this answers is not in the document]
alexban011: yeah I'm not very proud of the friend picker UI. Its a stripped down version of the Nation Picker. I'll definitely have to change with the PickerPane. More consistent + much better looking

SomeTroglodyte:
It's a github action triggered on every push, so if it fails at random from some external influence... You get the picture.

core/src/com/unciv/ui/multiplayer/ViewFriendsListScreen.kt
```diff
@@ -0,0 +1,92 @@
+package com.unciv.ui.multiplayer
+
+import com.badlogic.gdx.scenes.scene2d.ui.*
+import com.unciv.ui.pickerscreens.PickerScreen
+import com.unciv.ui.popup.Popup
+import com.unciv.ui.utils.*
+import com.unciv.ui.utils.AutoScrollPane as ScrollPane
+
+class ViewFriendsListScreen(previousScreen: BaseScreen) : PickerScreen() {
```
SomeTroglodyte, line 9: The rightSideButton of this Picker is never used, right? Looks a bit dead.

core/src/com/unciv/ui/newgamescreen/FriendTable.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.unciv.Constants
+import com.unciv.ui.multiplayer.FriendList
+import com.unciv.ui.utils.*
+
+class FriendTable(val friend: FriendList.Friend, width: Float, minHeight: Float)
+    : Table() {
+    val innerTable = Table()
+
+    init {
+        val innerColor = Color(1F,1F,1F,1F) //because 0xFFFFFFFF doesn't work for some reason
```
SomeTroglodyte, line 16: I explained the reason, right?

core/src/com/unciv/ui/multiplayer/FriendList.kt
```diff
@@ -0,0 +1,85 @@
...
+        friendList.add(editedFriend)
+        settings.save()
+    }
+
+    fun deleteFriend(friend: Friend) {
+        friendList.remove(friend)
+        settings.save()
+    }
+
+    fun getFriendsList(): MutableList<Friend> {
```
SomeTroglodyte, line 55: Your reference field is public, so any client could just use the field instead of the accessor... Like PlayerPickerTable does. The pattern is - you use accessors to control access, so the field would be private. If you're fine with a public field, no accessor needed? Also, FYI, `fun getFriendsList() = friendList` would compile exactly the same, the only difference being less options when debugging, and in very complex cases the type inference engine can run around in circles, so an explicit type here or there might help. Or it helps the reader... Here, it's simple. Mind - I'm not saying change that, just pointing out tidbits. Ah, another tidbit - the accessor pattern is ingrained in Java, and kotlin uses JVM - so if there weren't the one letter 's' difference, this would get you a complaint that the field getter and this accessor have the same JVM name... Which would disappear the instant you make the field private. Or use `@ JvmName`... Curious, ain't it?

core/src/com/unciv/ui/multiplayer/ViewFriendsListScreen.kt
```diff
@@ -0,0 +1,92 @@
...
+        helpButton.onClick {
+            val helpPopup = Popup(this)
+            helpPopup.addGoodSizedLabel("To add a friend, ask him to send you his player ID.").row()
+            helpPopup.addGoodSizedLabel("Click the 'Add friend' button.").row()
+            helpPopup.addGoodSizedLabel("Insert his player ID and a name for him.").row()
+            helpPopup.addGoodSizedLabel("Then click the 'Add friend' button again.").row()
+            helpPopup.addGoodSizedLabel("").row()
+
+            helpPopup.addGoodSizedLabel("After that you will see him in your friends list.").row()
+            helpPopup.addGoodSizedLabel("").row()
```
SomeTroglodyte, line 37: I don't think an empty label has a different height than an empty Cell - even then a pad would help. So just thingy.row() - or .add().row() if you want to manipulate the Cell before the row() call which changes the chain. Like .add().pad(10f).row() pads the empty cell but .row().pad() finishes the row then sets a default padding for the next row.

But then I would probably have stored all the lines in a raw string and used just one Label... Look for `""".trimIndent()` in the project if you're curious. Not that I _like_ trimIndent..

core/src/com/unciv/ui/multiplayer/ViewFriendsListScreen.kt
```diff
@@ -0,0 +1,92 @@
...
+        tab.y = (stage.height - helpButton.height)
+        stage.addActor(tab)
+
+        val mainTable = Table()
+        mainTable.add(ScrollPane(leftSideTable).apply { setScrollingDisabled(true, false) }).height(stage.height * 2 / 3)
+        mainTable.add(rightSideTable)
+        topTable.add(mainTable).row()
+        scrollPane.setScrollingDisabled(false, true)
+
+        rightSideTable.defaults().uniformX()
```
SomeTroglodyte, line 57: Hey I didn't know that one! Must have forgotten, the Language table also uses it, and that one I inspected long ago. But - here you could chain all the property setters for default cells. And from quick source inspection I doubt uniformX does anything for single-column tables.

core/src/com/unciv/ui/multiplayer/FriendList.kt
```diff
@@ -0,0 +1,85 @@
+package com.unciv.ui.multiplayer
```
SomeTroglodyte, line 1: We were having the "pure" separation of logic, metadata and UI lately... This might be better in metadata after all? maybe logic, there's some - no metadata. Then again, the data doesn't actually _live_ in an instance of this class, it's more of a utility... So logic?

As aside (I'm fine with the design) - some purists would say by the name this should be the actual container, and implement some Collection interface, then an instance of it would 'live' in settings. Just saying. Maybe a Kdoc for the class could clarify. If you want. The similarity to a Collection is still there, so some could argue the method names should follow patterns - add, remove, set. Your parameters and return types differ a little, so I'm not criticizing.

.gitignore
```diff
@@ -126,6 +126,7 @@ Thumbs.db
 SaveFiles/
 android/android-release.apk
 android/assets/GameSettings.json
+android/assets/FriendsList.json
```
SomeTroglodyte, line 129: Outdated?

alexban011:
Changes:
- Color(rgba) replaced with Color.WHITE
- refactored method names
- removed FriendsList.json from .gitignore
- moved FriendsList.kt to logic instead of ui

core/src/com/unciv/ui/multiplayer/EditFriendScreen.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        topTable.add("Player ID".toLabel()).row()
+        val gameIDTable = Table()
+        playerIDTextField.messageText = "Please input a player ID for your friend!".tr()
+        gameIDTable.add(playerIDTextField).pad(10f).width(2*stage.width/3 - pastePlayerIDButton.width)
+        gameIDTable.add(pastePlayerIDButton)
+        topTable.add(gameIDTable).padBottom(30f).row()
+
+        deleteFriendButton.onClick {
+            val askPopup = YesNoPopup("Are you sure you want to this friend?", {
```
Mape6, line 40: The word "delete" is missing in the sentence.
`Are you sure you want to delete this friend?"` would be right.

core/src/com/unciv/ui/multiplayer/AddFriendScreen.kt
```diff
@@ -0,0 +1,71 @@
...
+            when (friendlist.add(friendNameTextField.text, playerIDTextField.text)) {
+                FriendList.ErrorType.NAME -> ToastPopup("Friend name is already in your friends list!", this)
+
+                FriendList.ErrorType.ID -> ToastPopup("Player ID is already in your friends list!", this)
+
+                FriendList.ErrorType.NONAME -> ToastPopup("You have to write a name for your friend!", this)
+
+                FriendList.ErrorType.NOID -> ToastPopup("You have to write an ID for your friend!", this)
+
+                FriendList.ErrorType.YOURSELF -> ToastPopup("You cannot add your user ID in your friend list!", this)
```
Mape6, line 62: `You cannot add your own player ID to your friends list!` seems to be better for me.

core/src/com/unciv/ui/multiplayer/ViewFriendsListScreen.kt
```diff
@@ -0,0 +1,93 @@
...
+            helpPopup.addGoodSizedLabel("Click the 'Add friend' button.").row()
+            helpPopup.addGoodSizedLabel("Insert his player ID and a name for him.").row()
+            helpPopup.addGoodSizedLabel("Then click the 'Add friend' button again.").row()
+            helpPopup.addGoodSizedLabel("").row()
+
+            helpPopup.addGoodSizedLabel("After that you will see him in your friends list.").row()
+            helpPopup.addGoodSizedLabel("").row()
+
+            helpPopup.addGoodSizedLabel("A new button will appear when creating a new").row()
+            helpPopup.addGoodSizedLabel("multiplayer game, which allows you to select your friend.").row()
```
Mape6, line 41: Is it possible to combine both strings with a `\n` linebreak instead?

Mape6 (changes requested):
Very good work! That improves the multiplayer experience a lot.

core/src/com/unciv/ui/multiplayer/EditFriendScreen.kt
alexban011: I can't imagine how I let this happen after reading every word twice. Maybe I deleted it when I cleaned up :))

yairm210:
I'm sure there can be additional changes to this but let's get this out there and see what people say :)
It's not a must have, it's fine if it's not perfect, so lets get feedback!

PR #7026, opened 2022-05-31
Happiness fixes

### LINKED ISSUE #6942: Capital city yields are not really affected by unhappiness penalties.
**Describe the bug**
When your happiness drops below 0 and further on, below 10, you receive nasty penalties. However, these penalties affect all your cities but capital.

**To Reproduce**
1. Drop your happiness below 0 or -10
2. Inspect capital city yields in the overview - they are seemingly affected by the penalty.
3. End your turn
4. Notice the actual amount of production and food added to the pool are as if there were no penalties - for example the capital will grow as normal even though your happiness drops below -20 - whch should completely stop growth.

SomeTroglodyte:
Unit tests! One for each of the global uniques in vanilla, except Strength when below Happiness, which should already be covered.

My approach to fix #6942. Haven't actually found anything specific to the capital, though. Also, _do_ see the discussion there, on the actual limits and open questions how to read the unique wording. This is build on the assumption that "BETWEEN" will by most minds be associated with an inclusive range - I may be prejudiced by SQL. The behaviour was inclusive on one end exclusive on the other, so that needed fixing anyway, but my choice also needs the json to change. Chose not to deprecate/replace because - laziness? Also - if I'm right with wording and expectations most existing mods would be just fine.

Fixes issue not brought up by ravignir: Right after loading calculated values run (ran) on happiness==0 which can be seen on the city empire overview. Does so by calling another update thingy in GameInfo.setTransients. BUT - I think the whole stats update method is suspect and needs a radical change. Event bus maybe, but also a Pull instead of Push model. He who does anything that will invalidate any stat would only _mark_ which portion of stats are stale, the consumer would then pull from some helper with an EnumSet "I need these stats current", and that helper would be the only place needing to know about cascading updates and dependencies. Maybe. Just my two cents. Definitely not for this bugfix.

You may disagree with the tests relying on the actual vanilla json but checking if everything needed for the test is there - could do the main part of the checks with synthetic uniques instead, yes, I used that while debugging to get a clearer complete picture.

The last test is more specifically for the "stats right after loading are wrong" detail - StatPercentBonusCities already has a simple direct test, this one tests the happiness conditionals also working right after a load.

Commits separated into test and fix - the intermediate between them compiles and fails 4 of the tests (rebels are fine). The complete PR should run all tests successfully.


yairm210:
Sorry, this looks much much worse.
Gone are the nice round numbers, replaced by yukky 'between -1 and -9'. I understand the between-ness but this is just ugly.

SomeTroglodyte:
Almost out the door - haven't worked the latest ravignir conversation in anyway. DIY.

Issue #7025, opened 2022-05-31
Pixel unit nation colors un-applied after combat


letstalkaboutdune:
**Platform**
Reported from Win10 Pro 21H1, but occurs on Android as well

**Version**
4.1.5 from Itch Launcher

**Describe the bug**
If units with nation colors in their sprites receive damage during combat, after the red combat animation any nation colors will revert to white until the next turn.

**To Reproduce**
1. Enable a tileset that uses nation colors, such as 5Hex or Hexarealm with the AbsoluteUnits mod
2. Play a round of combat in any game
3. Observe the nation colors reverting to white after combat

**Expected behavior**
Units continue to display nation colors after combat.

**Screenshots**
5Hex example:
![5Hex](https://user-images.githubusercontent.com/56904240/171232144-ab3cf9fb-599f-4231-a693-0517f0315074.png)

Hexarealm + Absolute Units example:
![AbsoluteUnits](https://user-images.githubusercontent.com/56904240/171232168-1a6ce9e0-1186-42a7-a5f4-e1989d3a904e.png)

SomeTroglodyte:
I think it's the "damage visualization" red-flashing code. Got a save of that situation up there?

SomeTroglodyte:
https://user-images.githubusercontent.com/63000004/171291322-37baec28-9ed3-4670-be5f-c1df2861e3f1.mp4

This Caravel no longer loses its China-green stripe on the sail. Disadvantage: Units having RED as their national thingamabob color won't flash at all, at least not visible. Anyone knows of a colorimetric "Distance" function so I can replace RED when getting too near? simply `abs(r-other.r)+abs(g-other.g)+abs(b-other.b)`?

letstalkaboutdune:
"Unbleached," I love it! It looks like you fixed it, thanks for the quick attention to this. 

Units with red nation colors aren't completely red, and have some other color and other non-nation-specific colors in their design as well. I don't know if the red flash is a problem.

letstalkaboutdune:
Spain is probably the reddest - here's a save.

[Spain - 2 turns.txt](https://github.com/yairm210/Unciv/files/8809807/Spain.-.2.turns.txt)]

SomeTroglodyte:
Spain was just fine with red and weird with yellow.

PR #7034, opened 2022-06-01
Improve performance of worldmap panning

Azzurite:
Again, while reviewing this, I'd suggest going through commit-by-commit.

This improves performance of world map panning on my phone by about 100%, from 12 to 24 fps.

This is mainly achieved by simply disabling two things while panning:

1. Pointer enter + exit events. These are simply completely useless while panning, as we only allow panning via pointer, so the only thing that's happening is that a bunch of pointer enter/exit events are fired on the tiles you move your pointer over while panning.
2. `WorldMapHolder.act()`. This means all `act()` of all the tiles, buttons & whatever within `WorldMapHolder` are not being executed, so no animations playing or anything else. This is currently, as far as I can see, only the city button sliding downward/upward when clicked and the spinning circle around unit icons. Of course, this might change in the future, but I think even then it is fine to have stuff temporarily paused while you're panning. Maybe when we add more animating things, this optimization could then only be done if `continuousRendering` is `true` instead of always like now.

Now a performance improvement PR would be nothing without measurements, so here is the call tree for when only pointer enter/exit events are enabled:

image

36.82% of the time spent on firing useless enter/exit events.

Now the `WorldMapHolder.act()` might be surprising, since we already have `ActionlessGroup` for almost everything within there... but believe it or not, simply iterating over all actors within `WorldMapHolder` for large maps is already a decent chunk:

image

As you can see, 18.2% of the total time is spent solely for iterating over the actors. 31.46% total time for `WorldMapHolder.act`.

Now, there's also a third optimization, which basically had 0 effect that I didn't really expect. Before, the `WorldMapHolder` (which is a `ZoomableScrollPane`) always had the size of the maximum scrolled out state, so no matter how much you scrolled in or out, the size of the `WorldMapHolder` stayed the same. This meant that all tiles on maximal zoom out level were always rendered, even if you normally play zoomed in. On the map I tested with, that were around 1800 tiles rendered when fully zoomed out, and around 200-300 on a zoom level that you usually play at. Only rendering 200 instead of 1800 tiles made no measurable performance difference.

I left it and the refactoring I did of the `ZoomableScrollPane` & `WorldMapHolder` & `TileGroupMap`  in because I think it's a really good refactoring :D it basically moves the code within `TileGroupMap`, that is only responsible for getting the `ZoomableScrollPane` to work, to actually be within `ZoomableScrollPane` instead. Now `TileGroupMap` is simply responsible for putting its tiles on the screen, and `ZoomableScrollPane` is responsible for all things zooming and panning. This also moves a bunch of zoom/panning code from `WorldMapHolder` to `ZoomableScrollPane`.

### DIFF
--- a/core/src/com/unciv/ui/UncivStage.kt
+++ b/core/src/com/unciv/ui/UncivStage.kt
@@ -1,34 +1,47 @@
-package com.unciv.ui.crashhandling
+package com.unciv.ui
 
 import com.badlogic.gdx.graphics.g2d.Batch
 import com.badlogic.gdx.scenes.scene2d.Stage
 import com.badlogic.gdx.utils.viewport.Viewport
-import com.unciv.ui.utils.*
+import com.unciv.ui.utils.wrapCrashHandling
+import com.unciv.ui.utils.wrapCrashHandlingUnit
 
 
-/** Stage that safely brings the game to a [CrashScreen] if any event handlers throw an exception or an error that doesn't get otherwise handled. */
-class CrashHandlingStage(viewport: Viewport, batch: Batch) : Stage(viewport, batch) {
+/** Main stage for the game. Safely brings the game to a [CrashScreen] if any event handlers throw an exception or an error that doesn't get otherwise handled. */
+class UncivStage(viewport: Viewport, batch: Batch) : Stage(viewport, batch) {
 
-    override fun draw() = { super.draw() }.wrapCrashHandlingUnit()()
-    override fun act() = { super.act() }.wrapCrashHandlingUnit()()
-    override fun act(delta: Float) = { super.act(delta) }.wrapCrashHandlingUnit()()
+    override fun draw() =
+        { super.draw() }.wrapCrashHandlingUnit()()
 
-    override fun touchDown(screenX: Int, screenY: Int, pointer: Int, button: Int)
-        = { super.touchDown(screenX, screenY, pointer, button) }.wrapCrashHandling()() ?: true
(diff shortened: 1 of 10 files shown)


yairm210 (approved):
👍👍

SomeTroglodyte:
Fyi - "Blood" (see #7025) also uses act() - fine if it pauses or even aborts when panned

SomeTroglodyte:
<details><summary>Zoom INNNNN Screenie</summary>

image
</details>
... not half bad!

yairm210:
I'm going to merge this for this version then

SomeTroglodyte:
> because I left the `scaleX/Y` min values the same (at least I thought)...

There's even a comment where you replaced the limit saying something about equal rights :laughing:
And The difference -of course- can only be seen when "Max zoom out" in options (name!) is >> 2, because before Max zoom _in_ stayed at 2. But - 's fine.

> right-click -> checkout commit

For that I must "see" it - and your link is displayed here as "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." - no branch name. So I don't know where it 'lives' so I can't see it in any git log.. Just FYI, no harm.

SomeTroglodyte:
I was thinking about ZoomableScrollPane line 18, but you're right the comment says the inverse. Thing is, the variables were once named m[ia][nx]ZoomScale to clarify it's going into _scale[XY]_, that is large values mean things rendered larger meaning zoomed **in**. maxZoomOut is large = more zoomed **out**. The maxZoom_Scale_ from before _wasn't_ fed from maxZoomOut, it stayed at 2.0f. But all that is sausage.

PR #7035, opened 2022-06-01
Fix aircraft disappearing when their carrier is teleported

### LINKED ISSUE #7032: Aircraft carriers with planes that are forced out by an open borders agreement ending lose their planes
Version: 4.1.5

As reported by Klayman#1923 on Discord, when you press next turn in the attached game, open borders ends and the aircraft carrier near Neapolis gets pushed out of the territory. It had 4 planes on it before that happened, but afterwards, the planes are gone.

[Save game](https://gist.githubusercontent.com/Azzurite/272c0bdcca5284643dd9aff66159125c/raw/2ce72bbb8da3acae3de0776b2f7db54f3c5879f0/gistfile1.txt)

OptimizedForDensity:
Fixes #7032 by ensuring that transports bring their carried units with them when teleported (by open borders closing for example). All cases of `teleportToClosestMoveableTile()` are handled (in theory), and a new unit test was created to make sure that teleported carriers and their cargo are teleported to the same tile. `teleportToClosestMoveableTile()` will immediately return if called on a transported unit but this function has been given an optional boolean argument that allows for transported units to be teleported just like any other units if the boolean is set to true. By default this argument is false. The below image shows the linked save from #7032 but after a successful teleportation with the red dot being the carrier's location before being ejected.

Successful teleportation

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -414,8 +414,7 @@ class DiplomacyManager() {
 
         if (bordersWereClosed) { // borders were closed, get out!
             for (unit in civInfo.getCivUnits()
-                .filter { it.currentTile.getOwner()?.civName == otherCivName}
-                .filterNot { it.isTransported }.toList()) { // transported units are moved when their transport is moved
+                .filter { it.currentTile.getOwner()?.civName == otherCivName}.toList()) {
                 unit.movement.teleportToClosestMoveableTile()
             }
         }
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -414,7 +414,7 @@ class DiplomacyManager() {
 
         if (bordersWereClosed) { // borders were closed, get out!
             for (unit in civInfo.getCivUnits()
-                .filter { it.currentTile.getOwner()?.civName == otherCivName}.toList()) {
+                .filter { it.currentTile.getOwner()?.civName == otherCivName }.toList()) {
                 unit.movement.teleportToClosestMoveableTile()
             }
         }
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -414,7 +414,8 @@ class DiplomacyManager() {
 
         if (bordersWereClosed) { // borders were closed, get out!
             for (unit in civInfo.getCivUnits()
-                .filter { it.currentTile.getOwner()?.civName == otherCivName }.toList()) {
+                .filter { it.currentTile.getOwner()?.civName == otherCivName}
+                .filterNot { it.isTransported }.toList()) { // transported units are moved when their transport is moved
                 unit.movement.teleportToClosestMoveableTile()
             }
         }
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -414,7 +414,7 @@ class DiplomacyManager() {
 
         if (bordersWereClosed) { // borders were closed, get out!
             for (unit in civInfo.getCivUnits()
-                .filter { it.currentTile.getOwner()?.civName == otherCivName }.toList()) {
+                .filter { it.currentTile.getOwner()?.civName == otherCivName}.toList()) {
```
SomeTroglodyte, line 417: That space _was_ conforming to standards. 😀

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -394,13 +396,23 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
...
             // Cancel sleep or fortification if forcibly displaced - for now, leave movement / auto / explore orders
             if (unit.isSleeping() || unit.isFortified())
                 unit.action = null
             unit.mostRecentMoveType = UnitMovementMemoryType.UnitTeleported
+
+            // bring along the payloads
+            val payloadUnits = origin.getUnits().filter { it.isTransported && unit.canTransport(it) }.toList()
+            for (payload in payloadUnits) {
+                payload.removeFromTile()
+                payload.putInTile(unit.getTile())
```
SomeTroglodyte, line 412: getTile can differ from allowedTile here?

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -370,10 +370,12 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
...
      * This will not use movement points or check for a possible route.
      * It is used e.g. if an enemy city expands its borders, or trades or diplomacy change a unit's
-     * allowed position.
+     * allowed position. Does not move transported units by default but setting
+     * [forceMoveTransportedUnit] to true allows them to be moved.
      * CAN DESTROY THE UNIT.
      */
-    fun teleportToClosestMoveableTile() {
+    fun teleportToClosestMoveableTile(forceMoveTransportedUnit: Boolean = false) {
+        if (unit.isTransported && !forceMoveTransportedUnit) return // handled when carrying unit is teleported
```
SomeTroglodyte, line 378: Seems very strange. The bool gates this early exit for the "planes" but not the pull-along code for the "carrier"? So the flag is only for the case you want to teleport a transported plane but leave the carrier? Would that even work correctly or leave isTransported invalid? I think we do want the "carrier pulls along" to be default, so city expansion or a modded carrier with withdraw chance, for example, would do it too...  OK, I may not understand everything here, but seems like a dead flag, not what the Kdoc seems to describe.

SomeTroglodyte (commented):
With an answer to that "dead parameter" q - approvable, well done

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
OptimizedForDensity: Nope

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
OptimizedForDensity: Yeah that's a good point. I mainly added in case there was a future use that would want to teleport a transported unit separately from its transport but since I don't even know when this case would happen it's best to not implement it.

SomeTroglodyte (approved):

yairm210:
???
If you can't transport it then it shouldn't be upgradable
Probably aircraft should only be upgradable in cities anyway

PR #7041, opened 2022-06-01
Update Catalan.properties

J0anJosep:


heipizhu4:
> there is no need to remove the # Requires Translation text. That gets removed automatically when merged

Actually, I always do the remove because I will be able to use simple maths to tell how many of the translations is added or changed
Like `+5 -12` lines means 2 new lines and 3 changed lines.

PR #2024, opened 2020-02-27
Multiplayer Notification Fix for random Termination

wrov:
This pull request hopefully fixes the problem that the multiplayer turn notifier randomly stops running. The problem seems to have been a reliance on global variables. Because of the Worker architecture of the notifier, the Unciv app context may not exist when it is called by the WorkManager. In that case all global variables may have default or null values.

To remove the dependence on global variables, the turn notifier has been rewritten to store its settings in a Data object that is part of the WorkManager API.

Also all dependence on the .tr() lib has been removed. The .tr() call relies heavily on Unciv.Current being initialized, which is not guaranteed because Android might have unloaded the app from RAM.

As a result, I had to switch the notifier over to Android's native i18n lib, which uses a completely different set of files for its texts. :(

PR #7045, opened 2022-06-01
translated notifications in romanian

alexban011:
is there anything else I need to do ?

### DIFF
--- a/android/res/values-ro/strings.xml
+++ b/android/res/values-ro/strings.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_name">UnCiv</string>
+    <string name="Notify_YourTurn_Short">Unciv - Este rândul tău!</string>
+    <string name="Notify_YourTurn_Long">Prietenii tăi asteaptă după tine in [gameName].</string>
+    <string name="Notify_Error_Short">A aparut o eroare</string>
+    <string name="Notify_Error_Long">Notificator de ture pentru multiplayer oprit.</string>
+    <string name="Notify_Persist_Short">Ultima verificare online la rând:</string>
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
Uh oh. That set of files seems long forgotten.... The fr one being one of the sources of androidLint errors we have... But if you're sure these will be visible somewhere... :laughing: I think I once proposed adding them to our normal lang.properties files then having some auto writer copy them for the next build, but can't remember what became of that.

SomeTroglodyte:
> 6 entries instead of 16

What I meant - but there is a suppression for that specific one. Not really a reason for the intermittent build fails over here with that message...

SomeTroglodyte (approved):

yairm210:
I'm against this - we already have a translation mechanism with automations and tests, no need for external ones

SomeTroglodyte:
> just tested it. It translates the notifications

I believe (yes as opposed to _know_). Yair doesn't. Can you do screenshots?

alexban011:
WhatsApp Image 2022-06-02 at 22 06 00

this is the "Its your turn" notification

SomeTroglodyte:
> s [when this was first introduced](https://github.com/yairm210/Unciv/pull/2024) n

I never read that - good find. Can you also find where I proposed to have them "TranslationFileWriter"ed out of our normal tr files? I don't remember whether I actually coded something back then.. But today it should just follow how fastlane does it.

SomeTroglodyte:
> properly instantiate a Translations class

Which might pull in a ton of dependencies and that WorkManager runner should stay as lightweight as possible? :shrug:

SomeTroglodyte:
No I mean I expect the moment WorkManager instantiates the Translations it will pull in a lot of classes that it didn't need to load before. That xml is lightweight, instantiating Translations is not, and whatever it may pull as dependency... Your call.

PR #7047, opened 2022-06-02
Fix editor zoom in

Azzurite:
Fix the editor map holder being the wrong size (introduced in #7034), causing the center to be at a wrong place & not the whole map being visible.


SomeTroglodyte:
Looks exactly like the patch I showed earlier. Good to know I didn't guess wrong.

PR #7054, opened 2022-06-02
Better friends list UI

alexban011:
Replaced the odd looking friend selection UI with a PickerPane.

https://user-images.githubusercontent.com/21990802/171751497-b980f9a5-2a48-450c-af4c-450c0038690c.mp4

### DIFF
--- a/core/src/com/unciv/logic/multiplayer/FriendList.kt
+++ b/core/src/com/unciv/logic/multiplayer/FriendList.kt
@@ -16,7 +16,7 @@ class FriendList {
         ALREADYINLIST;
     }
 
-    data class Friend(val name: String, val playerID: String) {
+    data class Friend(val name: String, var playerID: String) {
         constructor() : this("", "")
     }
 
@@ -77,6 +77,15 @@ class FriendList {
     fun getFriendWithId(id: String): Friend? {
         for (index in friendList.indices) {
             if (id == friendList[index].playerID) {
+                return friendList[index]
+            }
+        }
+        return null
+    }
+
+    fun getFriendWithName(name: String): Friend? {
+        for (index in friendList.indices) {
+            if (name == friendList[index].name) {
                 return friendList[index]
             }
         }
--- a/core/src/com/unciv/logic/multiplayer/FriendList.kt
+++ b/core/src/com/unciv/logic/multiplayer/FriendList.kt
@@ -16,7 +16,7 @@ class FriendList {
         ALREADYINLIST;
     }
 
-    data class Friend(val name: String, var playerID: String) {
+    data class Friend(val name: String, val playerID: String) {
         constructor() : this("", "")
     }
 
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/multiplayer/FriendList.kt
```diff
@@ -82,4 +82,13 @@ class FriendList {
         }
         return null
     }
+
+    fun getFriendWithName(name: String): Friend? {
```
SomeTroglodyte, line 86: ByName would be a little more idiomatic - I think? 'with' sounds like you're returning more info, like kotlin .withIndex

core/src/com/unciv/logic/multiplayer/FriendList.kt
```diff
@@ -82,4 +82,13 @@ class FriendList {
...
+        for (index in friendList.indices) {
```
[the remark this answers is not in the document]
alexban011: dammit! I always double check the spacing when cleaning up and it looks like I'm still missing some of them :))

core/src/com/unciv/ui/multiplayer/FriendPickerList.kt
```diff
@@ -0,0 +1,56 @@
...
+        friendDisplays[friend.name] = friendDisplay
+        addActor(friendDisplay)
+        children.sort()
+    }
+}
+
+private class FriendDisplay(
+    friend: FriendList.Friend,
+    private val onSelected: (String) -> Unit
+) : Table(), Comparable<FriendDisplay> {
```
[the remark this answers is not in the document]
alexban011: It is used in FriendPickerList - Line 34 to sort the friends, but now that I think of it, it would probably be better to keep the order that they're in when viewing the friends list

core/src/com/unciv/logic/multiplayer/FriendList.kt
```diff
@@ -52,34 +52,25 @@ class FriendList {
...
-        return friendList
-    }
+    fun getFriendsList() = friendList
 
     fun isFriendNameInFriendList(name: String): ErrorType {
-        for (index in friendList.indices) {
-            if (name == friendList[index].name) {
-                return ErrorType.ALREADYINLIST
-            }
+        return if (friendList.firstOrNull { it.name == name } != null ) {
```
SomeTroglodyte, line 58: .. a call to the "new" getFriendByName would likely be inlined, thus giving the exact same result. Just FYI...

SomeTroglodyte (approved):

heipizhu4 (commented):
Is the ‘select friend' translatable? It seems not.

core/src/com/unciv/ui/newgamescreen/PlayerPickerTable.kt
```diff
@@ -276,109 +278,53 @@ class PlayerPickerTable(
...
+        val friendsList = FriendList()
+        val friend = friendsList.getFriendByName(friendName)
+        if (friend != null) {
+            selectedFriendId = friend.playerID
         }
-        player.playerId = selectedFriend?.playerID.toString()
-        close()
-        playerPicker.update()
+        pickerPane.setRightSideButtonEnabled(true)
+        pickerPane.rightSideButton.setText("Select [$friendName]".tr())
```
alexban011, line 325: not sure how to make this translatable. Would "Select [friendName] = " work?

PR #7060, opened 2022-06-03
Update Lithuanian.properties

MindaugasRumsa:


SomeTroglodyte:
W O W. Looks like tremendous work!

SomeTroglodyte:
In case you want to know how in the future - To the right of 'Build and deploy / build' is (but because this is a first-timer it waited until I hit the run button...) a "Details" link. In there, some digging produces:

```
com.unciv.testing.TranslationTests > allTranslationsHaveCorrectPlaceholders STANDARD_OUT
    2022-06-03T20:48:17.506460Z [HeadlessApplication] [TranslationTests] Placeholder `[civName]` not found in `Lithuanian` for entry `I've been informed that my armies have taken tribute from [civName], a city-state under your protection.
    I assure you, this was quite unintentional, and I hope that this does not serve to drive us apart.`
    2022-06-03T20:48:17.552586Z [HeadlessApplication] [TranslationTests] Placeholder `[comment]` not found in `Lithuanian` for entry `Triggers the following global alert: [comment]`
com.unciv.testing.TranslationTests > allTranslationsHaveCorrectPlaceholders FAILED
    java.util.UnknownFormatConversionException: Conversion = '`'
        at java.base/java.util.Formatter.checkText(Formatter.java:2732)
        at java.base/java.util.Formatter.parse(Formatter.java:2718)
        at java.base/java.util.Formatter.format(Formatter.java:2655)
        at java.base/java.util.Formatter.format(Formatter.java:2609)
        at java.base/java.lang.String.format(String.java:2897)
        at com.unciv.utils.LogKt.doLog(Log.kt:176)
        at com.unciv.utils.LogKt.access$doLog(Log.kt:1)
        at com.unciv.utils.Log.debug(Log.kt:60)
        at com.unciv.utils.Log.debug(Log.kt:47)
        at com.unciv.utils.LogKt.debug(Log.kt:157)
        at com.unciv.testing.TranslationTests.allTranslationsHaveCorrectPlaceholders(TranslationTests.kt:99)
```
Now the last one is our problem, The first two small oversights by you. Not bad for such a large submission.

@Azzurite - I see two details that hint our logging system maybe needs some literal escaping? Seems like it tried to interpret - a `"\\n"` became a `"\n"` and that backtick in the test output shouldn't concern it? Didn't look more closely.

MindaugasRumsa:
sorry, i had started translation from 0 at version 4.0.11 (1 a month ago)

android/assets/jsons/translations/Lithuanian.properties
```diff
@@ -2,2124 +2,1400 @@
...
+You demanded tribute from City-States that were under our protection! = Jūs reikalavote duoklės iš mūsų ginamų laisvųjų miestų!
+You sided with a City-State over us = Jūs susitarėte su laisvuoju miestu apeidami mus
+You returned captured units to us = Jūs grąžinote musų dalinius iš nelaisvės
+
+Demands = Reikalavimai
+Please don't settle new cities near us. = Nesikurkite šalia mūsų.
+Very well, we shall look for new lands to settle. = Gerai, mes ieškosime kitų žemių gyvenvietėms.
+We shall do as we please. = Mes elgsimės kaip panorėsim.
+We noticed your new city near our borders, despite your promise. This will have....implications. = Mes aptikome jūsų miestą prie mūsų sienų. Tai neliks... be pasekmių.
+I've been informed that my armies have taken tribute from [civName], a city-state under your protection.\nI assure you, this was quite unintentional, and I hope that this does not serve to drive us apart. = Man pranešė, kad mano kariuomenė ėmė duoklę iš jūsų ginamo laisvojo miesto.\nPatikinu jus, tai buvo nesusipratimas ir tikiuosi kad tai nebus priešastimi mūsų santykių pablogėjimui.
```
SomeTroglodyte, line 199: No need to be sorry. This would be the first one, where "[civName]" just needs to appear on the right as well

android/assets/jsons/translations/Lithuanian.properties
```diff
@@ -2149,5701 +1424,3244 @@ ConditionalsPlacement =
...
+Scholasticism = Skolastika
+Allied City-States provide [stat] equal to [relativeAmount]% of what they produce for themselves = Laisvieji miestai-sąjungininkai duoda [relativeAmount]% nuo jų gaminamo(s): [stat]
+Cultural Diplomacy = Kultūrinė Diplomatija
+[relativeAmount]% resources gifted by City-States = [relativeAmount]% prie resursų, jum dovanojamų iš laisvųjų miestų
+[relativeAmount]% Happiness from luxury resources gifted by City-States = [relativeAmount]% prie laimės nuo retų resursų, gaunamų iš laisvųjų miestų
+Educated Elite = Išsilavinusi Elita
+Allied City-States will occasionally gift Great People = Laisvieji miestai-sąjungininkai kartais dovanos jums Didžiuosius žmones
+Patronage Complete = Patronažas +
+Influence of all other civilizations with all city-states degrades [relativeAmount]% faster = Kitų civilizacijų įtaka visiems laisviesiems miestams mažėja [relativeAmount]% greičiau
+Triggers the following global alert: [param] = Siunčia globalų pranešimą: [param]
```
SomeTroglodyte, line 3375: And this the second, though it looks like it's not your mistake after all.

SomeTroglodyte: Ah, no, you *do* need to leave the generated parameter name in place. Both need to be `[comment]`. I thought both named equal would suffice, but our UniqueType system replaces the parameter name with the "supposed to" name before the test. In-game your line would actually work, but our tests demand you leave the left side before the " = " alone...

android/assets/jsons/translations/Lithuanian.properties
MindaugasRumsa: fixed it

android/assets/jsons/translations/Lithuanian.properties
MindaugasRumsa: changed to "comment"
... in 4.0.11 there was "param"
i thought that checker simply will remove such line and add new one with "requires translation"

in my translation there are some lines witch not exists only in mods,
also lines whose do not exists even in any mods
imho checker must just remove those lines without processing them

SomeTroglodyte: > in 4.0.11-4.0.15 there was "param"

Exactly!

> lines whose do not exists even in any mods

Two versions: The line comes from templates.properties -> technically entirely possible something is left unused and forgotten after a code change. Point them out - separate issue - or hijack that issue dealing with missing translation entries. V2 - not from templates, then must come from the json scan. If _that_ gives unused lines that would be even more curious - and report-worthy. For a translation PR, just ignore and move to more important lines if you want.

SomeTroglodyte:
OK, the reason for the next "red flag" now can't be seen in the log, our logger has a problem with trailing "%" chars at the moment. Did you treat that "Placeholder \`[relativeAmount]\` not found in \`Lithuanian\` for entry \`Damage taken from interception reduced by [relativeAmount]%\`" thing I listed in the paragraph directed @ azzurite? Or you could go look for similar problems in any line ending in `% =`. If you can't find it, we can (options: fix test logging to not try String.format, or tests to escape percents or just set a breakpoint at Log.kt line 176, check out your branch and run the test).

MindaugasRumsa:
tried to fish out all those errors

PR #7062, opened 2022-06-03
spectators can no longer move and attack with units after selecting them

### LINKED ISSUE #4460: Issues with desktop right-click to move
On desktop, you can rightclick after selecting a unit to quickly move or attack with it. However, there are a few bugs that can happen when doing this. Most of these are, because not all necessary checks are made when using this.

- [x] Spectators can move any unit they want by doing this
- [x] Carriers can preform ranged attacks by doing this
- [x] Attacking with an air unit and then rightclicking on any tile will cause an "I can't go there!" exception and remove the air unit from play
- [ ] You can move units while it is not your turn in multiplayer
- [ ] Probably more, but those are the ones I know from the top of my head

alexban011:
No matter what I do Github makes it hard to read so:
I added 
```Kotlin
if (!UncivGame.Current.gameInfo.currentPlayerCiv.isSpectator()) { 
    //code 
}
```
 on the code when you right click after selecting a unit as this caused many issues mentioned in #4460

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
@@ -216,34 +216,36 @@ class WorldMapHolder(
     }
 
     private fun onTileRightClicked(unit: MapUnit, tile: TileInfo) {
-        removeUnitActionOverlay()
-        selectedTile = tile
-        unitMovementPaths.clear()
-        worldScreen.shouldUpdate = true
-
-        if (worldScreen.bottomUnitTable.selectedUnitIsSwapping) {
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
Oh, didn't see that issue when I did #7003.  I think a if () return would be the better formatting.

Linking 4460 with "resolves" is a bit exaggerated - there's more loopholes listed there. Can't investigate the next week or so, but it sounds as if just a few more checks `if () return` might be in order - the difficulty being in finding how non-rightclick ensures the rules and to make sure these checks can "find each other" for the next coder coming along.

alexban011:
i reformatted but a correct fix would probably be better than this

PR #7063, opened 2022-06-03
More rankings & demographics screen icons

OptimizedForDensity:
Only a few of the categories in the rankings/experimental demographics screen have icons (they're `Stats` which have icons added automatically). This PR adds icons for the remaining ones, mostly reused from the overview screen -> stats screen to maintain consistency between menus.

New icons


core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -223,8 +224,12 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
         for (category in RankingType.values()) {
             val headers = Table().apply { defaults().pad(5f) }
-            headers.add(category.name.replace('_', ' ').toLabel()).row()
-            headers.addSeparator().fillX()
+            val textAndIcon = Table().apply { defaults().pad(5f) }
+            val columnImage = getRankingColumnIcon(category)
+            if (columnImage != null) textAndIcon.add(columnImage).center().size(Constants.defaultFontSize.toFloat() * 0.75f).padRight(0f).padTop(0f)
+            textAndIcon.add(category.name.replace('_', ' ').toLabel()).pad(5f).row()
+            headers.add(textAndIcon)
+            headers.addSeparator().size(0f, -1f).fillX()
```
SomeTroglodyte, line 232: :eyes:  that works? What does it do? Zero width, negative height? And fillX is already set by addSeparator?

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -223,8 +224,12 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
 
         for (category in RankingType.values()) {
             val headers = Table().apply { defaults().pad(5f) }
-            headers.add(category.name.replace('_', ' ').toLabel()).row()
-            headers.addSeparator().fillX()
+            val textAndIcon = Table().apply { defaults().pad(5f) }
```
SomeTroglodyte, line 227: No need to defaults if you explicitly set 6 of the 8 contained paddings? Also, doesn't that make overall padding 10f larger than before?

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
```diff
@@ -280,4 +289,17 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
         civGroup.pack()
         return civGroup
     }
+
+    private fun getRankingColumnIcon(type: RankingType): Image? {
```
SomeTroglodyte, line 293: Just FYI: Some may not like it, but that Image getter could be a "property" of the enum... Either as property containing a lambda -
```kotlin
enum class RankingType(val getImage: ()->Image?) {
    Score({ ImageGetter.getImage("OtherIcons/Cultured").apply { color = Color.FIREBRICK } }),
    ...
```
..or as normal method..
```kotlin
enum class RankingType {
    Score {
        override fun getImage() =
                ImageGetter.getImage("OtherIcons/Cultured").apply { color = Color.FIREBRICK }
    },
    ...
    ;
    open fun getImage(): Image? = null
}
```
both then used as `category.getImage()`. No need for a switch statement. Matter of taste.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
OptimizedForDensity: This adds the image with some trial-and-error sizing/padding to get the new icons to match the same size/spacing as the existing Stat icons.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
OptimizedForDensity: I chose the first approach.

core/src/com/unciv/ui/victoryscreen/VictoryScreen.kt
OptimizedForDensity: That's a bit overzealous with padding, plus it led to a visual peculiarity for the demographics screen where the leftmost column's separator wasn't at the same height as the other separators. Removed and simplified.

OptimizedForDensity:
Atlas undeleted.

PR #7064, opened 2022-06-04
Moddable prettier Tutorials - Step 1

SomeTroglodyte:
Concept:
- The "Tutorial" as in object displayable in Civilopedia becomes a RulesetObject. This is overdue and will allow a few simplifications while allowing nicer presentation. And, soon, makes transformation into an actual part of rulesets easy, too. Tutorials could _know_ about the difference between vanilla and g&k, and become moddable.
- The enum actually concerns itself with the tests when such a text should be spammed. Thus it is renamed to [Totoro](https://en.wikipedia.org/wiki/My_Neighbor_Totoro)ialTrigger (sorry, not quite) - _much_ clearer IMHO, and may lead to some day replacing the triggers with uniques bound to python scripts - think scenarios. All that tutorial detection code in WorldScreen needs to find a better home outside UI anyway.

This PR is the bare minimum first step which _still works as before_. Only difference - adding a Tutorial without a corresponding TutorialTrigger is now explicitly allowed.

Side notes:
- dropping "steps" would require to either decide the annoying "1 line, need OK, another line, need OK" goes, or some idea on how to group civilopediaText. Some FormattedLine instance that triggers 'stop rendering here and ask for OK'?
- Interesting how instead of Ruleset.createHashmap the standard library is enough - and more flexible, the ruleset loader could probably save an interim copy or two where it does +=.
- Gdx Array shadowing standard Array - mega annoying, I'm going to `as GdxArray` _all_ unavoidable uses in the future. And where it was - probably unsuspectingly - used in that soon-to-be-redundant Tutorial to Civilopedia code looks cleaner with normal kotlin Arrays too.
- I'll not be listening for at least a week, so - patience.
- That Era change: I was surprised the compiler even allows inheriting the same interface repeatedly. (Would `class Era : RulesetObject(), IHasUniques, IHasUniques, IHasUniques, IHasUniques, IHasUniques` compile too?????) :smile_cat:

### DIFF
--- a/android/assets/jsons/Tutorials.json
+++ b/android/assets/jsons/Tutorials.json
@@ -1,140 +1,243 @@
-{
-  // Each entry s a tutorial, but the tutorial may be spread over separate paragraphs.
-  // Entries starting with a _ will NOT be shown in Civilopedia.
-  //
+[
+  // Each entry is a tutorial, but the tutorial may be spread over separate paragraphs.
 
-  Introduction: [
+  {
+    "name": "Introduction",
+    "steps": [
       "Welcome to Unciv!\nBecause this is a complex game, there are basic tasks to help familiarize you with the game.\nThese are completely optional, and you're welcome to explore the game on your own!"
-  ],
+    ]
+  },
 
   // Civilopedia only, because players said this was too wall-of-text
-  New_Game: [
+  {
+    "name": "New Game",
+    "steps": [
       "Your first mission is to found your capital city.\nThis is actually an important task because your capital city will probably be your most prosperous.\nMany game bonuses apply only to your capital city and it will probably be the center of your empire.",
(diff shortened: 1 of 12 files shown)


SomeTroglodyte:
Yes, random musing - take Ruleset.L235, and a line I would come up with (or did for the Tutorials loader):
```kotlin
            terrains += createHashmap(json().fromJsonFile(Array<Terrain>::class.java, terrainsFile))
            json().fromJsonFile(Array<Terrain>::class.java, terrainsFile).associateByTo(terrains) { it.name }
```
A few chars more source, but v1 copies twice, v2 only once, and no need for a helper. Just standard "learning kotzlin". Nothing to see here, move along, no grisly murder victims, no blood, no random body parts lying around, not even a naked... clochard.

<details><summary>And the other question - compare:</summary>

### Now
image
### Potential Future
image
</details>
.. just a quick mock-up. Less than 20 lines from my stash, a _little_ json change, fetch an image from wikipedia. Oh, and don't look at the jumbled worldscreentopbar, not my fault.

PR #7068, opened 2022-06-04
Update Japanese.properties

itmaru:
Added about 20 lines of Japanese translation.

Didn't know where the 'fastlane_full_description' so good so put one there commented out for now.

### DIFF
--- a/android/assets/jsons/translations/Japanese.properties
+++ b/android/assets/jsons/translations/Japanese.properties
@@ -19,29 +19,7 @@ Fastlane_short_description = 4Xの文明をモチーフとしたターン制の�
 
 # Documentation: https://f-droid.org/en/docs/Build_Metadata_Reference/#Description
 # English to translate: https://github.com/yairm210/Unciv/blob/master/fastlane/metadata/android/en-US/full_description.txt
-Fastlane_full_description =
-#世界一の文明創造ゲームの再実装！オープンソースなので、永遠に無料！
-
-#文明を造り、テクノロジーを研究し、都市を広げ、ライバル文明を倒す！
-
-#リクエスト・バッグなどはこちらにお問い合わせ https://github.com/yairm210/UnCiv/issues
-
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
> Didn't know where the 'fastlane_full_description' so good so put one there commented out for now.

That will, with a long delay, end up on: https://f-droid.org/ja/packages/com.unciv.app/
... but it won't work like you did it. Our translations are all a single "technical line" like `to_translate = translated` - to include several "logical lines" in one translation you need to use `\n` instead of an actual line break.

That little detail must be in our wiki...

heipizhu4:
Fun fact: We've got a little description here in line 1174&1175 in `template.properties`, but it seems that hardly anyone ever noticed that.
```
 # The \n here means: put a newline (enter) here. If this is omitted, the sidebox in the diplomacy overview will become _really_ wide.
 # Feel free to replace it with a space and put it between other words in your translation
```

I wonder if we can put these 2 lines somewhere else.

SomeTroglodyte:
> got a little description here in line 1174&1175

That's specific to that entry, it may have historically been the first one using embedded newlines. I half-expected [this wiki page](https://github.com/yairm210/Unciv/wiki/Translating#pitfalls) to already mention them, but - no.

heipizhu4:
> > got a little description here in line 1174&1175
> 
> That's specific to that entry, it may have historically been the first one using embedded newlines. I half-expected [this wiki page](https://github.com/yairm210/Unciv/wiki/Translating#pitfalls) to already mention them, but - no.

I suppose we've got 2 choices. The first one is to put these 2 lines in the very beginning of the template, or shall we add a line in the wiki? Maybe both.

SomeTroglodyte:
> Maybe both

Wiki and template can _refer_ to it? Just "see wiki" not even a link - suppose xlators are clever enough to follow that? Keep templates simpler? If you'd be so kind...

itmaru:
Thanks for the comments and sorry for the trouble - I've updated this PR with a version that uses \n for line breaks.

PR #7072, opened 2022-06-05
Notification Log

alexban011:
Started working on a notification log. So far it saves and displays notifications in the overview menu. You can click on them and they do the action they would normally do if they were a normal notification.

Next is to separate the notification per turn so you can tell exactly what happened in which turn and probably to give some visual feedback to tell people they can click on the notifications (maybe buttons? I think I'd like to make them look like the normal notifications) It would also be nice to have the notifications in something like the <details> tag, but I couldn't find anything like this in libgdx.

The text also includes square brackets so those would also need to be removed

https://user-images.githubusercontent.com/21990802/172050799-225b149a-c248-430d-a26e-a18eef745cc3.mp4

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -147,6 +147,8 @@ class CivilizationInfo {
     var diplomacy = HashMap<String, DiplomacyManager>()
     var proximity = HashMap<String, Proximity>()
     var notifications = ArrayList<Notification>()
+    var notificationsLog = ArrayList<Notification>()
+    var notifTurnCounter = ArrayList<Int>()
     val popupAlerts = ArrayList<PopupAlert>()
     private var allyCivName: String? = null
     var naturalWonders = ArrayList<String>()
@@ -255,6 +257,7 @@ class CivilizationInfo {
         toReturn.exploredTiles.addAll(gameInfo.tileMap.values.asSequence().map { it.position }.filter { it in exploredTiles })
         toReturn.lastSeenImprovement.putAll(lastSeenImprovement)
         toReturn.notifications.addAll(notifications)
+        toReturn.notificationsLog.addAll(notificationsLog)
         toReturn.citiesCreated = citiesCreated
         toReturn.popupAlerts.addAll(popupAlerts)
         toReturn.tradeRequests.addAll(tradeRequests)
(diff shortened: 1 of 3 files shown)


SomeTroglodyte:
> something like the < details > tag, but I couldn't find anything like this in libgdx

There isn't. ExpanderTab.

> find a royalty free one

Read credits.md

> look like the normal notifications

Just reuse, or better take the expanders (I assume they'll be meant to wrap one turn's worth of notifications) into the scaling, meaning you'd refactor NotificationsScroll to separate the ScrollPane + scaling wrapper from the content? Not easy as Gdx scaling isn't intuitive IMO.

> notifTurnCounter

Ugly. Better wrap one turn's worth of notifications in a class. Its constructor could do all the nextTurn work for you as well. You could even migrate "current" to be just the latest entry in your log - but that would need savegame migration.

Also - candidate for considerable savegame bloat. Configurable maxAge.

alexban011:
> class

what about a 2D array instead? I don't think it would need savegame migration and can be like arr[turn][notification]

alexban011:
This is how it looks at the moment.

https://user-images.githubusercontent.com/21990802/172728599-93f84902-8b44-4e95-928a-21a0c25d8f6e.mp4

> Would be better if place it on pop-up screen than overview screen.
Would it? That would mean adding another button on the screen. I think the overview screen is the better place to put this. I'd like to know more opinions though.

alexban011:
https://user-images.githubusercontent.com/21990802/172733721-140a6763-9f9d-49af-a5b1-6566577b95cb.mp4

alexban011:
https://user-images.githubusercontent.com/21990802/172902278-38311e31-35b2-4ca3-add2-5a1b1ab4ca5e.mp4

This is how it looks so far. I couldn't make it have 2 scroll panes. Tried with autoscroll pane but didn't work. I think I need to disable the scrolling for the entire table but couldn't figure out how to do that. @SomeTroglodyte do you know how I can do that?

touhidurrr:
> This is how it looks so far.

alexban011:
> anchors that takes you to the notiffs

is this doable? Do you happen to know a place in the game where something like this is done? 

> Turn instead of for button text maybe.

I agree, will do

OptimizedForDensity:
Another option would be using a vertical slider to quickly scroll the notifications list. The slider in the game setup screen that controls the number of turns in the game is what I'm talking about, especially how the number pops up once you start scrolling.

alexban011:
took me "only" 6 commits to fix the issue...

PR #7073, opened 2022-06-05
Fix test logging not using format specifiers for arbitrary text

Azzurite:
Should be pretty straightforward. You can't log arbitrary strings that may contain `%` directly, you have to use the `params` for that. So ``debug("%s", "%`")`` is fine, but ``debug("%`")`` is not.

Noticed in #7060, which has failing translations that include a `%` sign.

### DIFF
--- a/tests/src/com/unciv/testing/BasicTests.kt
+++ b/tests/src/com/unciv/testing/BasicTests.kt
@@ -68,7 +68,7 @@ class BasicTests {
         var allObsoletingUnitsHaveUpgrades = true
         for (unit in units) {
             if (unit.obsoleteTech != null && unit.upgradesTo == null && unit.name !="Scout" ) {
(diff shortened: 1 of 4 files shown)


tests/src/com/unciv/testing/BasicTests.kt
```diff
@@ -94,7 +94,7 @@ class BasicTests {
             val ruleset = RulesetCache[baseRuleset.fullName]!!
             val modCheck = ruleset.checkModLinks()
             if (modCheck.isNotOK())
-                debug(modCheck.getErrorText(true))
+                debug("%s", modCheck.getErrorText(true))
```
SomeTroglodyte, line 97: With your change `if (params.isEmpty()) msg else msg.format(*params)` this would have been redundant?

PR #7077, opened 2022-06-06
Refactor: Consistent & correct usage of coroutines

Azzurite:
This changes all concurrency handling to coroutines and makes the API consistent.

The only thing that wasn't handled by coroutines was the GDX GL thread, which still used the `postCrashHandlingRunnable` function everywhere.

This has now been refactored to all be contained within the `Concurrency.kt` file. The comments there should explain everything.

So instead of this 

```kotlin
// ...code unconcerned with threading/coroutines...
launchCrashHandling {
    // ...coroutine code in another thread...
    postCrashHandlingRunnable {
        // ...ui code...
        launchCrashHandling {
             // ...some more threaded code...
        }
    }
}
```

it now looks like this 

```
// ...code unconcerned with threading/coroutines...
Concurrency.run {
    // ...coroutine code in another thread...
    launchOnGLThread {
        // ...ui code...
        launchOnThreadPool {
             // ...some more threaded code...
        }
    }
}
```

(the nesting levels just for simplicity, usually you'd extract some `suspend` methods out of there so it doesn't look like a staircase)

which is designed to be more similar to the way [coroutines work by default](https://kotlinlang.org/docs/coroutines-basics.html), i.e. a `runBlocking` call around `launch` calls. The previous `launchCrashHandling` still came from a place where I didn't understand coroutines well enough.

Also, I never liked  the `crashHandling` suffix on everything. That should be an implementation detail, which it is now.


yairm210:
A lot of conflicts :(

PR #7086, opened 2022-06-07
Update Japanese.properties

itmaru:
Various easy one line translations


itmaru:
Looked like spacing issues between previous commits and recent one - I fixed those (I think?) and remerged so maybe should be all good for now.

PR #7089, opened 2022-06-07
Update ManuallyBlockedMods.json

GeneralWadaling:
Deciv repository has been transferred recently and it has been updated with same standard as Deciv Redux.


GeneralWadaling:
https://github.com/GeneralWadaling/DeCiv

PR #7091, opened 2022-06-07
Fix scout not upgrading through ruins

### LINKED ISSUE #7090: Scout can't upgraded to archer UU through ruins
**Platform**
Android 10, 64bit
**Version**
4.1.7 version
**Describe the bug**
Scout can't upgraded to archer UU(Bowman, slinger etc.) through ruins

**To Reproduce**

```

```
</details>

**Expected behavior**

**Screenshots**

**Additional context**

OptimizedForDensity:
Fixes #7090. There was a missing check to see if the unit has a special upgrade possible so it was affecting scouts upgrading to archers not just unique units like the issue mentions.


yairm210:
Okay, we should refactor the 'specialUpgradesTo' into a proper unique eventually though

PR #7097, opened 2022-06-08
small typo

AxizY:
vertical bar was there instead of L


yairm210:
Nope, that's intentional!
Read carefully and you'll see I was talking about having a *similar name* to the original series

PR #7114, opened 2022-06-09
Refactor:  Change UncivGame.worldScreen and UncivGame.gameInfo to be of nullable type

Azzurite:
This is again a PR in preparation for other stuff that I have in the works, so this is actually a collection of three things, which I again recommend to review commit-by-commit:

* [8cabd80](https://github.com/yairm210/Unciv/pull/7114/commits/8cabd808a8ecf0e1acbe9231718142c0858cda1c): The title refactoring, changing `UncivGame.worldScreen` and `UncivGame.gameInfo` to be of nullable type.
   Having no loaded game is a perfectly fine state our application can be in. `lateinit` is only for variables that are supposed to be initialized immediately, mainly not being initialized in the constructor because some framework does not allow it. This is not the case for the `gameInfo` and `worldScreen` variables: you can open our game, browse the civiliopedia, change the options, manage some multiplayer games or whatever, all without loading a game.
   
   This has a change in `UncivGame` to `override` the parent methods `getScreen` and `setScreen`. The reason for that was that when I changed `worldScreen` to be of nullable type, suddenly the call `setScreen(worldScreen)` did not go to `fun setScreen(BaseScreen)` anymore, but to the base method `fun setScreen(Screen?)`, bypassing our important initialization code when changing screens. You wouldn't believe how much time I had to spend to figure that one out... The override is to prevent that from happening accidentally again.
* [5f25ddb](https://github.com/yairm210/Unciv/pull/7114/commits/5f25ddb5e2d7a01f246ba9ab23b3fcbe52172ecb): Refactor: Make Popups work on `Stage`s instead of `BaseScreen`s
   The reason for this is that each `Actor` has a `getStage()` method, returning the `Stage` it is added to, but neither `Actor` nor `Stage` have a `getScreen()` method. So with this refactoring, it makes it possible to open a `Popup` without knowing the exact `BaseScreen` the `Popup` needs to be attached to, it's enough to have one single `Actor` somewhere.
* [6d85796](https://github.com/yairm210/Unciv/pull/7114/commits/6d857968c199c48ce877c1dba96b0a48abf4c486): Fix "Resume" game loading not fully handling exceptions, pretty simple.


core/src/com/unciv/ui/saves/QuickSave.kt
```diff
@@ -82,13 +85,26 @@ object QuickSave {
...
                 }
             } else {
                 postCrashHandlingRunnable { /// ... and load it into the screen on main thread for GL context
                     try {
                         screen.game.loadGame(savedGame)
                     } catch (oom: OutOfMemoryError) {
                         outOfMemory()
+                    } catch (ex: Exception) {
+                        Log.error("Could not autoload game", ex)
+                        postCrashHandlingRunnable {
```
yairm210, line 104: No need for this since we're already on the main thread

yairm210:
Looks good!

yairm210 (approved):

Issue #1733, opened 2020-01-20
Viewing event history


Whistler11:
**Is your feature request related to a problem? Please describe.**
I use the notifications on the side to decide how much I am going to act in a turn. Sometimes I forget what happened recently, or when I come back to a game, I'd like a refresher. Having a history of them would be great to look back on.

**Describe the solution you'd like**
A tab in the "menu" of "event history" that lets you look back with a little bit more space than what is offered on the side.

**Describe alternatives you've considered**
*make the font of the events larger, so they're clearer to read. This could easily make the screen cluttered and difficult to view the rest of the content, so I don't think this is a solution for everyone. alternatively -> give people an option for the size of the event font. Then people can pick the limits they they are comfortable with.
*have the events as a pop up at the start of every turn. The pop up can be interacted with (and send you to the notification location), dismissed, and recalled from a button that sits where the current event timeline is.

PR #7122, opened 2022-06-10
Notifications Log

### LINKED ISSUE #6536: In-Game Notification Logs
Sometime an Important Notification appears but you miss it. Can't we store them even if just locally?
Screenshot_2022-04-13-03-32-00-83_5b83b50424e447777d9855d9701553df

alexban011:
Created new PR because I messed up the old one too much :) Draft PR: #7072

Added a notification log for keeping track of what happened in the last turns. The age is configurable in the gameplay tab. The reason this setting is per player and not per game is to further decrease save-game bloat. (if someone wants 15 turns worth of notifications but someone else doesn't, its better to just store just as many notifs as the player needs). 

The notifications are stored only if there are any in the turn. So if you have the max age set to 3 turns and you have this:
Turn 0 - 1 notification
Turn 1 - 0 notifications
Turn 2 - 2 notifications
Turn 3 - 0 notifications
Turn 4 - 3 notifications 
You would be able to see the notifications all the way to Turn 0 and the next turn when you get a notification it will remove Turn 0.

I also tested a game created in 4.1.8 and then loaded with this PR and it works. (previous turns are obviously not tracked, but it starts tracking at the current turn)

Note: now it no longer displays "Current turn" if there are no notifications in the current turn
https://user-images.githubusercontent.com/21990802/173048779-169c0571-4188-4b2d-bbd7-21fc9d89c6ea.mp4

closes #7072
resolves  #6536
resolves (maybe) #1733, but I'd need another opinion if this issue is indeed fixed by this before linking it with resolve


core/src/com/unciv/ui/overviewscreen/NotificationsOverviewTable.kt
```diff
@@ -0,0 +1,96 @@
...
+                val ruleset = worldScreen.gameInfo.ruleSet
+                for (icon in notifications[index2].icons.reversed()) {
+                    val image: Actor = when {
+                        ruleset.technologies.containsKey(icon) -> ImageGetter.getTechIcon(icon)
+                        ruleset.nations.containsKey(icon) -> ImageGetter.getNationIndicator(
+                            ruleset.nations[icon]!!,
+                            iconSize
+                        )
+                        ruleset.units.containsKey(icon) -> ImageGetter.getUnitIcon(icon)
+                        else -> ImageGetter.getImage(icon)
```
yairm210, line 84: This looks like the same logic for the notifications in the main screen, so how about unifying them? This could be a function of Notification in fact - getIconImage(ruleset:Ruleset) or something similar

alexban011:
thought merging might fix the issue. Why can't it find the functions? :/

alexban011:
Oh come on how is that a conflict.. those are totally unrelated

core/src/com/unciv/ui/overviewscreen/NotificationsOverviewTable.kt
```diff
@@ -0,0 +1,81 @@
...
+            val turnCounter = notificationLog.lastIndex - index
+            notificationTable.add(notificationsArrayTable(turnCounter.toString(), notificationLog[index].notifications))
+            notificationTable.padTop(20f).row()
+        }
+    }
+
+    private fun notificationsArrayTable(index: String, notifications: ArrayList<Notification>): Table {
+        val turnTable = Table(BaseScreen.skin)
+        if (index != "Current")
+            turnTable.add("Turn $index").row()
```
yairm210, line 55: Not translatable - "Turn [turnNumber]"

core/src/com/unciv/ui/options/GameplayTab.kt
```diff
@@ -35,4 +38,22 @@ fun gameplayTab(
...
+
+    val minimapSlider = UncivSlider(
+        3f, 15f, 1f,
+        initial = settings.notificationsLogMaxTurns.toFloat()
+    ) {
+        val turns = it.toInt()
+        settings.notificationsLogMaxTurns = turns
+        settings.save()
+        if (screen is WorldScreen)
+            screen.shouldUpdate = true
```
yairm210, line 56: Doesn't this only affect the log that isn't in the worldscreen? Why update the worldscreen?

core/src/com/unciv/logic/civilization/Notification.kt
```diff
@@ -48,6 +52,23 @@ open class Notification() {
         this.icons = notificationIcons
         this.action = action
     }
+
+    fun addNotificationIcons(ruleset: Ruleset, iconSize: Float, table: Table) {
+        if (this.icons.isNotEmpty()) {
```
yairm210, line 57: No need for "this", and reverse this check for less indentation

alexban011:
> Where did you take the `Notifications.png` from? Do we have permission to use it?

https://www.flaticon.com/premium-icon/notifications_1436388?term=notification&page=1&position=27&page=1&position=27&related_id=1436388&origin=search

I say in credits that we have a bunch of icons from this website so I took it from here as well

alexban011:
I removed it to keep the credits.md as clean as possible :)

alexban011:
> Ok now this is good :) Very nice work, very useful feature!

thanks! :)

yairm210:
Resolve conflicts

alexban011:
>  impossible to merge while there are conflicts

ohh okk. I'll remove them

touhidurrr:
@alexban011, only 3 to 15 turns? Why not allow unlimited turns?

touhidurrr:
> > only 3 to 15 turns? Why not allow unlimited turns?
> 
> He says why in his PR description.

Yeah, but what if someone wants unlimited retention? Decrease bloat as much as possible? For me 10MB worth of bloat per game means nothing. I don't think for most desktop players also. 

And I doubt that It's going to be 10MB even if you play 1000 turns.

PR #7027, opened 2022-05-31
Rework Policy and Diplomacy buttons

SomeTroglodyte:
Addresses Part 1 of #6989 (the longer I think about it, the more convinced I am @Ouaz is right about points 1 and 2). Looks pretty much exactly as in their mock screenshot.

Based on #6945 -> Therefore Draft.

PR #7130, opened 2022-06-10
Fix positioning of the spectator's fog of war button

OptimizedForDensity:
The fog of war toggle disappeared from the screen recently. The cause lies in the `topBar.y` being 0 until the top bar is updated during `update()`, which causes the fog of war button to be placed below the bottom of the screen as it uses `topBar.y` to set vertical position and is placed before `topBar.y` is properly set. This PR changes it so that `topBar.update(viewingCiv)` is called during the world screen's initialization so that the fog of war button can be placed in the correct spot.

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -154,7 +154,6 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
         techPolicyAndVictoryHolder.add(techButtonHolder)
 
         fogOfWarButton.isVisible = viewingCiv.isSpectator()
-        fogOfWarButton.setPosition(10f, topBar.y - fogOfWarButton.height - 10f)
 
         // Don't show policies until they become relevant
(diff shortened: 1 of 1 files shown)


SomeTroglodyte:
Sorry for that - I believe I saw the need to position that button at an opportune time (meaning when TopBar.height is known efficiently), but it might be I put the actual code to that end in #7027 - or intended to. Thing is, grouped buttons that need to stay together are more efficiently handled as a group than individually, and I vaguely remember being convinced (by code inspection, not by testing) that the "Fog" button is never shown at the same time as most top-left buttons so it fits nicely in that mini-"widget" without needing much effort to position it.

Updating topbar twice is probably not the best answer. Moving the positioning code to where the other top-left buttons are positioned might be better - or go and check that draft. It was fully tested but based on the not-yet-merged TopBar PR, only reason I marked it Draft only. And I went offline shortly after so I couldn't merge+ready it.

I'm still away from my Studio rig, just an internet cafe here, so it'll be a while before I can check. Go ahead and compare that PR if you want, and if you prefer the approach and get it to run nicely with current master feel free to re-PR, all my input here is free for the whole team to use without attribution.

SomeTroglodyte:
> No answer

Was away for a week. No big deal.

PR #7141, opened 2022-06-11
Fix unit being captured two times

Azzurite:
`placeUnitNearTile` already captures any civilian units it places its unit into. But `spawnCapturedUnit` unconditionally tries to capture the unit again. This resulted in capturing a unit that the civ already owned, which shouldn't work. However, `captureCivilianUnit` only checks for the `originalOwner`, not the current owner, and still does the capture anyway. This resulted in the dialog "Do you want to return this captured unit to yourself?" which then finally resulted in an exception. 

Took a while to figure the root cause out. Since `captureCivilianUnit` should not be called for capturing our own unit, I made it throw an `IllegalArgumentException`, which should catch these kind of errors during development, or at least make a bug easier to find.

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -348,9 +348,12 @@ object Battle {
         addedUnit.currentMovement = 0f
         addedUnit.health = 50
         attacker.getCivInfo().addNotification(notification, addedUnit.getTile().position, attacker.getName(), unitName)
-        // Also capture any civilians on the same tile
-        if (tile.civilianUnit != null)
-            captureCivilianUnit(attacker, MapUnitCombatant(tile.civilianUnit!!))
+
+        val civilianUnit = tile.civilianUnit
+        // placeUnitNearTile might not have spawned the unit in exactly this tile, in which case no capture would have happened on this tile. So we need to do that here.
+        if (addedUnit.getTile() != tile && civilianUnit != null) {
+            captureCivilianUnit(attacker, MapUnitCombatant(civilianUnit))
+        }
         return true
     }
 
@@ -559,7 +562,14 @@ object Battle {
         return null
     }
 
+    /**
+     * @throws IllegalArgumentException if the [attacker] and [defender] belong to the same civ.
+     */
     fun captureCivilianUnit(attacker: ICombatant, defender: MapUnitCombatant, checkDefeat: Boolean = true) {
+        if (attacker.getCivInfo() == defender.getCivInfo()) {
+            throw IllegalArgumentException("Can't capture our own unit!")
+        }
+
         // need to save this because if the unit is captured its owner wil be overwritten
         val defenderCiv = defender.getCivInfo()
 


yairm210 (approved):

alexban011:
is this the fix for the issue on discord ?

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -348,9 +348,12 @@ object Battle {
...
         addedUnit.health = 50
         attacker.getCivInfo().addNotification(notification, addedUnit.getTile().position, attacker.getName(), unitName)
-        // Also capture any civilians on the same tile
-        if (tile.civilianUnit != null)
-            captureCivilianUnit(attacker, MapUnitCombatant(tile.civilianUnit!!))
+
+        val civilianUnit = tile.civilianUnit
+        // placeUnitNearTile might not have spawned the unit in exactly this tile, in which case no capture would have happened on this tile. So we need to do that here.
+        if (addedUnit.getTile() != tile && civilianUnit != null) {
+            captureCivilianUnit(attacker, MapUnitCombatant(civilianUnit))
```
JackRainy, line 355: Not sure, when this code will be ever called. Should it be removed at all?
Disregard of where the `placeUnitNearTile` spawns here or there, either the new spawned unit or the attacker will call `moveThroughTile` which does all that civilian-capturing magic, so it will be captured in both cases.

Furthermore, after calling `tryCaptureUnit` (which calls this `spawnCapturedUnit`) there is the next code:
```
        // Should be called after tryCaptureUnit(), as that might spawn a unit on the tile we go to
        if (!captureMilitaryUnitSuccess)
            postBattleMoveToAttackedTile(attacker, defender, attackedTile)
```
which contains some scenarios e.g. _"if we attack e.g. a barbarian in enemy territory that we can't enter, we won't enter it"_ when the civilian must be captured (this code must not be called), if we do not reach the tile.

JackRainy: > The attacker will not call moveThroughTile if it is a ranged unit?

I would expect "no", and I assume the only way to capture the civilian is to move on its tile.

> It will only ever be called by modded units, as Vanilla has no ranged units that can capture a military unit.

I still do not understand why this capture needs to be called even if the ranged unit captures the military one.
If the captured military unit is in the targeted tile - this code is not needed (`placeUnitNearTile` does its work), if it is in the other tile, the civilian in the attacked tile by the ranged unit still must not be captured.

JackRainy: I may agree with your initial suggestion where both the military unit and civilian units are captured and pushed off there. However, I strictly disagree when either military or civilian units end up on the tile where they must not be.
Even assuming in some mods that the ranged unit (with range 2+) civ Y may capture the military unit civ X on the tile where it may not enter, we get into trouble when the captured unit cannot ever leave that tile since all tiles around are impassable.

Let @yairm210 decide on:
1) whether the attacker (either ranged or melee) may capture the military unit on the tile it cannot enter?
2) whether the civilian unit may be captured if the attacker (either ranged or melee) cannot enter that tile?
3) if yes, where the captured military and/or civilian units must end up: on unpassable tile or be pushed off?

yairm210: Agree, those are really good questions (that I don't have a clear answer for) but for a different PR
Personally I think "if you can attack you can capture", and the question is then "can you attack tiles you can't enter"... I think the 'can attack of embark' question basically answered this add "no".
Does this mean melee units will be able to attack without being attacked?
Yes. But I'm okay with that.

PR #7144, opened 2022-06-12
Fix NPE in Nation selection & editor mods popup

Azzurite:
Wrong `stage` variable was being used, `Actor.stage` is null until the Popup is shown. `stageToShowOn` is correct.


SomeTroglodyte:
Strange. MapEditorModsTab wasn't anywhere near any Popup. All Editor Tabs and Subtabs have references to the main MapEditorScreen and use that screen's stage while initializing - unless I tested with breakpoints that the actor's stage was good enough in the circumstances..

PR #7145, opened 2022-06-12
Fix OutOfMemory error when loading game state after already having a game loaded

Azzurite:
Before, when we already had a game running and loaded another game, we used (more than) double the memory because we created a new `WorldScreen` before getting rid of the previous `WorldScreen`.

Now, the previous `WorldScreen` is disposed of before loading the new one. This should mean that any game you can load initially, you can also keep playing.

### DIFF
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -155,10 +155,39 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
     }
 
     /** Loads a game, initializing the state of all important modules. Automatically runs on the appropriate thread. */
-    suspend fun loadGame(newGameInfo: GameInfo): WorldScreen = withThreadPoolContext {
+    suspend fun loadGame(newGameInfo: GameInfo): WorldScreen = withThreadPoolContext toplevel@{
         val prevGameInfo = gameInfo
         gameInfo = newGameInfo
 
+        initializeResources(prevGameInfo, newGameInfo)
+
+        val isLoadingSameGame = worldScreen != null && prevGameInfo != null && prevGameInfo.gameId == newGameInfo.gameId
+        val worldScreenRestoreState = if (isLoadingSameGame) worldScreen!!.getRestoreState() else null
+
+        withGLContext { setScreen(LoadingScreen(getScreen())) }
+
+        worldScreen?.dispose()
+        worldScreen = null // This allows the GC to collect our old WorldScreen, otherwise we keep two WorldScreens in memory.
+
+        return@toplevel withGLContext {
+            val worldScreen = WorldScreen(newGameInfo, newGameInfo.getPlayerToViewAs(), worldScreenRestoreState)
+
+            val moreThanOnePlayer = newGameInfo.civilizations.count { it.playerType == PlayerType.Human } > 1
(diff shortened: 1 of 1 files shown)


yairm210:
I have to say I don't exactly understand which parts are supposed to do what in this change
I hope that this doesn't dispose the current world screen upon "next turn"?

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -604,26 +611,38 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
...
+            mapHolder.scrollY = restoreState.scrollY
+            mapHolder.updateVisualScroll()
         }
 
-        newWorldScreen.selectedCiv = gameInfo.getCivilization(selectedCiv.civName)
-        newWorldScreen.fogOfWar = fogOfWar
-
-        return newWorldScreen
+        selectedCiv = gameInfo.getCivilization(restoreState.selectedCivName)
+        fogOfWar = restoreState.fogOfWar
```
yairm210, line 637: I'm pretty sure these are calculable from the game info, not sure if this convenience is more confusing than it is helpful

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -897,4 +894,32 @@ class WorldScreen(val gameInfo: GameInfo, val viewingCiv:CivilizationInfo) : Bas
...
+}
+
+/** This exists so that no reference to the current world screen remains, so the old world screen can get garbage collected during [UncivGame.loadGame]. */
+private fun startNewScreenJob(gameInfo: GameInfo) {
+    Concurrency.run {
+        val newWorldScreen = try {
+            UncivGame.Current.loadGame(gameInfo)
+        } catch (oom: OutOfMemoryError) {
+            val mainMenu = UncivGame.Current.goToMainMenu()
+            ToastPopup("Not enough memory on phone to load game!", mainMenu)
```
yairm210, line 908: Should be in GL context

core/src/com/unciv/ui/LoadingScreen.kt
```diff
@@ -1,14 +1,30 @@
...
+import com.badlogic.gdx.graphics.g2d.SpriteBatch
 import com.unciv.Constants
+import com.unciv.ui.popup.Popup
 import com.unciv.ui.utils.BaseScreen
-import com.unciv.ui.utils.extensions.center
-import com.unciv.ui.utils.extensions.toLabel
 
+/** A loading screen that creates a screenshot of the current screen and adds a "Loading..." popup on top of that */
 class LoadingScreen : BaseScreen() {
+    val screenshot = Texture(Pixmap.createFromFrameBuffer(0, 0, Gdx.graphics.backBufferWidth, Gdx.graphics.backBufferHeight))
```
yairm210: NEATO

SomeTroglodyte: Why is a screenshot even necessary? The effect looks a bit strange - you get the "Loading..." text first in a _wide_ blue rectangle for a while, then in a _narrow_ blue rectangle for another second before the result pops in.

PR #7147, opened 2022-06-12
Improvements to AI military unit usage

OptimizedForDensity:
The AI military action logic has been changed in a few places and the AI able to better perform actions including attacking and rallying for a defense or attack. Most of the changes are in UnitAutomation.kt, where there are a series of lines that look like `if (trySomeAction(unit)) return`. Once the return statement is hit, the AI stops giving orders to the unit for the turn regardless of how many movement points are left. As a result, the AI would never be able to attack more than once per turn or move after attacking or pillaging, which is a huge benefit to certain promotions and units/uniques. Attacking logic has been changed so that the AI will now attempt to attack as many times as it is allow by promotions (e.g., chu-ko-nus or other units with the logistics promotion). Several of the `trySomeAction(unit)` style functions in this file have also have their return statements changed, usually to `return unit.currentMovement < Constants.minimumMovementEpsilon` so that they can do another action if they have movement points left. The functions affected are
* `tryUpgradeUnit`
* `tryHeadTowardsSiegedCity`
* `tryPillageImprovement`
* `BattleHelper.tryAttackNearbyEnemy`
* `headTowardsEnemyCity`

The first two functions were particularly bad since they would always return true whether or not the action was successful, meaning that units would do nothing while trying to upgrade for dozens of turns or units would endlessly try to travel to a city to defend it without even attempting to attack the besieging army. The AI is also much better with cavalry and tanks and is able to chain together actions in a somewhat logical manner (move into enemy territory -> attack enemy -> pillage improvement to gain health -> retreat outside of enemy borders to heal more).

### DIFF
--- a/core/src/com/unciv/logic/automation/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/UnitAutomation.kt
@@ -474,8 +474,9 @@ object UnitAutomation {
             // move into position far away enough that the bombard doesn't hurt
             if (tileToMoveTo != null) {
                 unit.movement.headTowards(tileToMoveTo)
+                return true
             }
-            return unit.currentMovement < Constants.minimumMovementEpsilon
+            return false
         }
 
         val numberOfUnitsAroundCity = closestReachableEnemyCity.getTilesInDistance(4)
@@ -491,13 +492,13 @@ object UnitAutomation {
 
             if (tileToHeadTo != null) { // no need to worry, keep going as the movement alg. says
                 unit.movement.headTowards(tileToHeadTo)
-                return unit.currentMovement < Constants.minimumMovementEpsilon
             }
+            return true
         }
 
         unit.movement.headTowards(closestReachableEnemyCity) // go for it!
 
-        return unit.currentMovement < Constants.minimumMovementEpsilon
+        return true
     }
 
     fun tryEnterOwnClosestCity(unit: MapUnit): Boolean {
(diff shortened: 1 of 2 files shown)


yairm210:
Awesome!!!

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -243,6 +243,14 @@ object UnitAutomation {
             wander(unit, stayInTerritory = true)
     }
 
+    /** @return true only if the unit has 0 movement left */
+    private fun tryAttacking(unit: MapUnit): Boolean {
+        for (numAttacks in unit.attacksThisTurn until unit.maxAttacksPerTurn()) {
+            if (BattleHelper.tryAttackNearbyEnemy(unit)) return true
```
yairm210, line 249: This should return upon the very first attack, no?

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -342,7 +347,7 @@ object UnitAutomation {
             unit.movement.moveToTile(tileToPillage)
 
         UnitActions.getPillageAction(unit)?.action?.invoke()
-        return true
+        return unit.currentMovement < Constants.minimumMovementEpsilon
```
yairm210, line 350: This should be zero, since we set it to zero if it moves below the epsilon

core/src/com/unciv/logic/automation/UnitAutomation.kt
OptimizedForDensity: I tested with the AI playing as the Chinese since their UU has 2 attacks and they were able to attack twice per turn.

yairm210:
It's fine
I still don't understand why they can multi attack, 'end this function if you managed to attack' is quite indicative.
But even without that fix it's still a big improvement

PR #7153, opened 2022-06-13
Update Lithuanian.properties

MindaugasRumsa51:


android/assets/jsons/translations/Lithuanian.properties
```diff
@@ -590,7 +590,7 @@ Error while refreshing: = Klaida atnaujinimo metu
 Last refresh: [time] minutes ago = Paskutinis atnaujinimas: prieš [time] min.
 Current Turn: = Esamas ėjimas:
 Add Currently Running Game = Pridėti esamą žaidimą
-Paste gameID from clipboard = Įterpti žaidimo ID is duomenų mainų srities
+Paste gameID from clipboard = Įterpti žaidimo ID iš \nduomenų mainų srities
```
alexban011, line 593: there is no \n in the original string

PR #7154, opened 2022-06-13
Fix scroll to wonder in the map editor

Azzurite:
No, I did not want to properly refactor this yet, but the fix is still sound.


SomeTroglodyte:
How is this a fix, replacing Gdx's internal ScrollPane animations with another one?

PR #7155, opened 2022-06-13
Implementation of Teams

alexban011:
Before continuing I thought it might be a good idea to show what I have so far to avoid working on something which won't be implemented. I started by creating team classes for the `Player` objects and then wanted to create `Team` classes for the `CivilizationInfo`s from `GameInfo`, but since the teams only affect so little of the actual gameplay (time and domination victory only) I though the refactoring hell would not be worth it, although I still have this branch but didn't do the refactoring for civTeams. So I ditched that idea and went for this.

<a href="https://gaming.stackexchange.com/questions/251475/full-effects-of-being-in-the-same-team-in-civilization-v">Here is how teams work in Civ 5</a>

Roadmap:
- [ ] When any player on the team researches the tech tree all of the teammates get the technology. If you both research the same tech at the same time, the time for completing it will be halved.
- [x] Tech cost increases by 50% for each additional team member
- [x] You also see the map under fog of war which is cleared by the other player on the team.
- [ ] If any player on the team declares war on someone, the whole team goes into the war.
- [ ] The the points of everyone in the team will be counted together and it shows a team score at the list on the right.
- [ ] You can gift units (both combat & civilian) to your teammates just like you can for city-states normally.
- [x] You have permanent Declaration of Friendship with each other
- [ ] You'll still feel ideological pressure (reduced happiness for adopting different ideologies) from team mates
- [ ] Autocracy tenet Futurism doesn't work against team mates.

So far I have a rough UI which lets you assign teams to players and whenever you research something your teammates also get that tech.

~~I will also add a toggle for the teams at some point~~ looks like civ 5 always has teams enabled so I guess there's no point in doing this

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -37,6 +37,17 @@ class GameInfo {
     var currentTurnStartTime = 0L
     var gameId = UUID.randomUUID().toString() // random string
 
+    fun addTechnologyToTeam(team: String, techName: String) {
+        var teammatesCivs = civilizations - civilizations.filter { it.team != team }
+        for (teammate in teammatesCivs) {
+            if (teammate.tech.techsResearched.find { it == techName } == techName)
+                teammatesCivs -= teammate
+        }
+        for (teammate in teammatesCivs) {
+            teammate.techsResearchedByTeam.add(techName)
+        }
+    }
+
     // Maps a civ to the civ they voted for
     var diplomaticVictoryVotesCast = HashMap<String, String>()
     // Set to false whenever the results still need te be processed
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -62,9 +62,11 @@ class GameInfo {
     fun getTeammatesOfCiv(civInfo: CivilizationInfo): ArrayList<CivilizationInfo> {
         val teammates = ArrayList<CivilizationInfo>()
 
-        for (civ in civilizations) {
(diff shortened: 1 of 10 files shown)


core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -37,6 +37,17 @@ class GameInfo {
     var currentTurnStartTime = 0L
     var gameId = UUID.randomUUID().toString() // random string
 
+    fun addTechnologyToTeam(team: String, techName: String) {
+        var teammatesCivs = civilizations - civilizations.filter { it.team != team }
+        for (teammate in teammatesCivs) {
+            if (teammate.tech.techsResearched.find { it == techName } == techName)
+                teammatesCivs -= teammate
```
alexban011, line 44: maybe I should remove the teammate from Civs after the for? Would this risk a ConcurrentModificationException?

ravignir:
techs should also be more expensive if you play in a team.

OptimizedForDensity:
It would be nice if the tech tree showed techs that your teammates are currently researching since collaboration speeds up research.

alexban011:
yeah thats how I originally started but thought it might be easier and better like this. Then I realized this will need some ugly code to make it work so I stopped. 

I also thought about having TeamManagers in civs but I doubt that would be cleaner than Teams having Civs and I'm not even sure if it would even make anything any better really.

I'm gonna close this PR and will come back with this when I got something worth showing.

alexban011:
> Shouldn't be a major effort to do that

it would require to refactor quite a lot of stuff though. For example every `civilizations.any{ }` needs to check for every civ in every team. Or am I thinking this wrong?

PR #7163, opened 2022-06-14
CityScreen plays a sound when opened

alexban011:
I saw a request on discord to add a bell sound when opening the city screen.
I took that a bit further and made the cityscreen play a sound according to the era the player is in. 

Links for the sounds:
    https://freesound.org/people/VKProduktion/sounds/231537/ (CC0)
    https://freesound.org/people/priesjensen/sounds/482990/ (CC0)
    https://freesound.org/people/dheming/sounds/240981/ (Attribution)
    https://freesound.org/people/_MC5_/sounds/542107/ (Attribution)
    https://freesound.org/people/Hedmarking/sounds/191883/  (CC0)
    https://freesound.org/people/InspectorJ/sounds/385943/  (Attribution)
    https://freesound.org/people/kyles/sounds/451523/ (CC0)
    https://freesound.org/people/bkowshik/sounds/439218/ (CC0)
    https://freesound.org/people/bolkmar/sounds/502896/ (Attribution)
    https://freesound.org/people/LittleRobotSoundFactory/sounds/270404/ (Attribution)
Note: I combined these sounds to get the city sounds

Sounds:
[sounds.zip](https://github.com/yairm210/Unciv/files/8935607/sounds.zip)

Corrupted mod used to test with:
[audioTest.zip](https://github.com/yairm210/Unciv/files/8926641/audioTest.zip)


core/src/com/unciv/ui/cityscreen/CityScreen.kt
```diff
@@ -133,6 +136,21 @@ class CityScreen(
...
+            0 -> SoundPlayer.play(UncivSound.CityAncient)
+            1 -> SoundPlayer.play(UncivSound.CityClassical)
+            2 -> SoundPlayer.play(UncivSound.CityMedieval)
+            3 -> SoundPlayer.play(UncivSound.CityRenaissance)
+            4 -> SoundPlayer.play(UncivSound.CityIndustrial)
+            5 -> SoundPlayer.play(UncivSound.CityModern)
+            6 -> SoundPlayer.play(UncivSound.Click) //atomic
+            7 -> SoundPlayer.play(UncivSound.Click) //information
+            8 -> SoundPlayer.play(UncivSound.Click) //future
+            else -> SoundPlayer.play(UncivSound.Silent)
```
yairm210, line 150: Inherently unmoddable, why not check if a sound file exists with the name of the era, if yes play it, if no then click?

alexban011: can't quite tell how to do this. If I replace `UncivSound.CityAncient`  with `UncivSound("cityAncient")` will it work ?

OptimizedForDensity: Can you just play a sound with something like `UncivSound("city" + city.civInfo.getEra().name)` with a catch for if the sound file doesn't exist (maybe play a default sound like `UncivSound.Click`)?

alexban011: > UncivSound.Click

I only used that to test that its actually playing a sound. The city name button already plays that sound so the default would be Sound.Silent

> UncivSound("city" + city.civInfo.getEra().name)

just did this. But I used the era number instead of the name as the name was giving `Ancient Era` and not just `Ancient` 
But would this be moddable?

alexban011:
I removed the try block as it wasn't gonna give a file not found ex. I guess its already handled somewhere

OptimizedForDensity:
Adding a new volume slider just for entering cities seems strange. Maybe just tying it to sound effects volume could keep things simple.

alexban011:
> Adding a new volume setting just for entering cities seems strange.

I though maybe it'd be good to give players more control over the sounds.

core/src/com/unciv/ui/cityscreen/CityScreen.kt
```diff
@@ -196,6 +201,11 @@ class CityScreen(
         }
     }
 
+    private fun playEnterCitySound() {
+        SoundPlayer.clearCache()
```
alexban011, line 202: how can I only stop this sound instead of stopping them all? Also I doubt clearing the cache is even a good way of stopping the sounds as they would need to be cached again, right?

MindaugasRumsa51: better to move this to the mod

MindaugasRumsa51: https://freesound.org/people/addiofbaddi/sounds/241683/

core/src/com/unciv/ui/cityscreen/CityScreen.kt
SomeTroglodyte: There was a time long long ago when UncivSound was an enum and the attackSound present in json but a dead fish doing nothing. To avoid the mutable enum instance debacle that _was_ happily done elsewhere later, I made it a class allowing runtime-instantiation but keeping source compatibility for the existing "enum constants" which are now explicitly pre-created singletons (enum "constants" are too under the hood).... Just so you can understand better. A new "predefined" might be good if you expect it to be used in dozens of places throughout the code, otherwise don't hesitate to just use the constructor as shown.

> file not found ex. I guess its already handled

Read the sound code. Looks a whopper but most is just caching (Gdx dox on Sounds say "dispose!", reloading _can_ be heard, so the cache has a cleanup, everybody happy). To allow mod sounds it needs to do a folder search, so of course the individual "not found" just moves on - what I did when all folders exhausted I've forgotten - hey I can't check, the class is _gone_ :roll_eyes: Ah there it is, another helpful rename - yup, get does debuglog, cache the "not found" fact, return null; and play does get ?: return. You could extend the "API" of SoundPlayer with a `play(): Boolean` return or with a `isCached(sound: UncivSound) = sound in soundMap` helper or a `play(, fallback: UncivSound? = null)` parameter or probably a few other ways if you need predictable fallbacks.

core/src/com/unciv/ui/cityscreen/CityScreen.kt
SomeTroglodyte: clearCache is public for game shutdown, _not_ normal operations. At the moment you can't stop a sound, just as looping Sounds are no-go: They're _supposed_ to be tiny and short (anything > 1s would be better off streaming, there's Gdx.Music for that). Which is why SoundPlayer doesn't bother exposing the Gdx Sound.loop/pause/resume/stop methods.

If you need stop functionality, extend the helper class - but then you would need to learn why Gdx has two entirely different concepts of what a sound "instance" means - is the Java instance relevant or the long ID? Or are they interchangeable? Inconsistent.

SomeTroglodyte:
"sounds now stop if you exit the city screen" - if they're that long, then the Gdx Sound class is not appropriate.

alexban011:
>  if they're that long, then the Gdx Sound class is not appropriate.

they're at most 10 seconds, which gives enough time to exit and reenter, and thus playing the sound multiple times. I changed this to play music, but I don't know how to make it moddable and how to allow all accepted extensions (at the moment only mp3 works)

Do I need to recreate the music controller, maybe a `CityEnterSoundController` that searches for all folders and does most of what the music controller does? Cause I'm guessing I can't just reuse the music controller, especially not with all those private consts
This class could also be extended to something like `CityEnterAndUnitSelectSoundController` but with a better name :)

alexban011:
Now its using the music controller, which makes the sounds moddable while still being able to stop them. Now this kinda makes me think: what about making these sounds loops instead of being played just once to give the sensation of being in the city non stop, not just when entering... but that would require much more effort than I'm willing to put into this :))

alexban011:
The docs would also need to be updated but I deleted by gh-branch and don't know

SomeTroglodyte:
The music controller was written to play _one_ theme, two only when fading over from one to the next. Looks like what you're now doing is - the Greek's theme is playing, a city is opened, the music fades over to your music. When it ends, the "source" is left changed, so when it looks for new ambient music it won't find the usual but will play anything from the source pool. No, wait, you're instantiating a separate MusicController. So the Greek theme continues playing - but you'll probably get the play anything effect, maybe not audible because they're too short to be faded in. The nuke sound will probably make it sooner or later... (There's a flag just two slot above the one you added - and you mean "or" not "of")

I don't really like having two MusicController instances that don't know about each other - though at the moment there's no usecase for coordination between the 'classic' tracks and your extra track. Wait - there's a killer: the desktop crash issue we had since lwjgl3. The existing MusicController now catches these by being "hooked" into Gdx internals on startup, yours isn't. Look at UncivGame.kt lines 112-116 and HardenGdxAudio.kt.

No idea how to cheaply solve that desktop crash issue. Do without the timer task - but then reusing MusicController would be redundant. Just do _one_ Gdx Audio.newMusic, try-catched, and don't call any methods after play (also try-catched), not even stop was safe by my prejudices. But I believe that would still crash, uncatchable, with - say - an mp3 with an actual bitstream error past the initial buffers the play method would preload.

But if that weren't an issue - don't var musicPath, a constructor val param with default would do; and that remnant cityScreenEnterVolume - just setVolume to soundEffectsVolume instead would do it? This way, players having sound explicitly muted would...

alexban011:
> you'll probably get the play anything effect

I just left the city screen open while writing this comment and it didn't play anything other than the original sound

> and you mean "or" not "of"

Yes I do, the keys are too close together :))

> The existing MusicController now catches these by being "hooked" into Gdx internals on startup, yours isn't. 

gonna have a look at that in a sec

>  Gdx Audio.newMusic

tried that but I don't know how to add support for ogg files and how to stop the music. See https://github.com/yairm210/Unciv/pull/7163/commits/d96d4476694e4b7e6be045958a9cd9d6fea2ec7a

> cityScreenEnterVolume

The sounds are indeed tied to the soundEffectsVolume, and that I had a separate slider for this sound but then removed it and forgot to remove the setting. Don't really know why I used it there, probably cause I saw it and sounded right :))

alexban011:
I tried adding the hooks but it crashes if I leave the city screen

alexban011:
So I've put the functions in their own class to allow me in the near future to add sound for the units as well (when selected) and to make it look cleaner

SomeTroglodyte:
> adding the hooks but it crashes if I leave the city screen

Of course. I'm sure you could quickly guess why by just reading the code or the comments.

alexban011:
> What are you testing on? 

Desktop

But based on the _very long_ comment you left there I guess it doesn't matter that it didn't crash so far. Gonna have to use this.
Would it be wrong to reuse the audioExceptionHelper from UncivGame instead of making another one ?

SomeTroglodyte:
> wrong to reuse the audioExceptionHelper

Yes. Because there can be only one hook under the hood. The only _clean_ way I see is to extend the _one_ MusicController until it can do what you need. Sorry, I can't be bothered to do it myself right now. There may be some not-so-clean way... But you'd have to understand exactly why and how that "hardening" kludge works. It may be fine to reuse Gdx Music, leave the MusicController in peace, but still have it's hooks process buffer refills - which would take care of those exceptions that would otherwise crash past our crashhandling. Add to that to avoid calling _any_ method on a playing instance that could lead to buffer work -> stop might be safe, if called from the main thread. In the end it may be _really_ simple, but I just can't be bothered right now to really think and test it through.

Yes, the Sound class is easier, but there's a reason Gdx discourages their use for anything longer: Afaik the _decoded_ bits are kept in memory as long as the instance lives, and they're always preloaded in their entirety before they can play. And we're still trying to cater for those 256M dalvik heap phones, or aren't we? Then there's the sample rate issue which I don't remember perfectly (meaning an Android device typically has a _native_ sample rate, oem-dependent, and any media with a different one plays fine but with a bunch of harsh sounding logcat warnings - I believe software is doing resampling on-the-fly, which could multiply memory usage). I can't really balance this out from memory to decide how bad maybe 1 Minute of more sounds in-memory would be.

core/src/com/unciv/ui/audio/CitySoundPlayer.kt
```diff
@@ -0,0 +1,62 @@
...
+            playingCitySound = Gdx.audio.newMusic(file)
+        } catch (ex: Throwable) {
+            ex.printStackTrace()
+        }
+
+        try {
+            playingCitySound.play()
+        } catch (ex: Throwable) {
+            ex.printStackTrace()
+        }
```
alexban011, line 52: since this is now "hooked" is there any point in keeping the try blocks?

core/src/com/unciv/ui/audio/CitySoundPlayer.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fun stopCitySound() {
+        try {
+            playingCitySound.stop()
+        } catch (ex: Throwable) {
+            ex.printStackTrace()
+        }
+    }
```
alexban011, line 61: same here

alexban011:
> Also please add the links to 

I will as soon as I get them

> Did he make them himself?

He said he edited them himself. I linked the medieval one.
WLTK sounds are modified Thatched Villagers

alexban011:
> Also I'm sorry to be annoying with the source of these sounds

it's perfectly understandable

> What is "Thatched Villagers"? Where can I find it?
Its the song downloaded by the "download music" button in Options -> SoundTab

alexban011:
Actually I'll change the sounds to make them loop while the city screen is open. I just tested it I think it would be much better from a UX point of view

SomeTroglodyte:
> I corrupted a file and made a mod with it.

Well done, but I'm still not convinced. That exception fired from newMusic, which isn't the problem. The problem is - Gdx is doing a kind of buffered streaming and what happens when the _second_+ buffer(s) needs filling. Whatever problems arise from the first buffer's filling have always been 'catch'-able, because they are done early.

Look at [this source](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L181) - suspicious, right? Thats's where exhausted audio buffers are refilled, and there's no exception handling anywhere if you follow the call chains down - and upwards the next possibility to catch is _outside_ UncivGame, meaning - pointless. What that "hardening" does is redirect the Line 182 `audio.update()` to _our_ code where a try _can_ catch...

So, a good testing "corrupted file" would need to be bad _beyond_ that first buffer filling. I believe - but could be wrong - that means 48k _encoded_ data, [hardcoded](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/audio/OpenALMusic.java#L36) (and _nothing to do with_ the [app config buffer parameters](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3ApplicationConfiguration.java#L49)).

You know what - I still haven't opened Studio for this tirade, but the longer I ponder, the more certain I am that the "main" MusicController's hardened audio update _will_ catch those nasty surprises, at worst leading to "outside city" music changing tracks -> because Gdx.audio _is_ replaced, so your newMusic _is_ placed in the correct stack serviced by the hardened audio update.

alexban011:
> Look at [this source](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/backends/gdx-backend-lwjgl3/src/com/badlogic/gdx/backends/lwjgl3/Lwjgl3Application.java#L181) - suspicious, right?

Unbelievable this hasn't been done in the last 6 years :https://github.com/libgdx/libgdx/commit/448d44b32c33d7c4e9ee05126db7a4aafe1bdcdf#diff-7b1e3aae3e2bda381fef08f608a5b18074128f58c73c0e3c0103458d799734d5 :))

alexban011:
I made the changes @Azzurite 
Wanna have a look at those?

core/src/com/unciv/ui/audio/CitySoundPlayer.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    fun stopCitySound() {
+        try {
+            if (playingCitySound?.isPlaying == true)
+                playingCitySound?.stop()
+        } catch (ex: Throwable) {
+            playingCitySound?.dispose()
+            ex.printStackTrace()
```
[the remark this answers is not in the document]
alexban011: so an Log.debug(ex.printStackTrace()) ?

alexban011:
> When I enter a city in this [problem_game.zip](https://github.com/yairm210/Unciv/files/8936329/problem_game.zip), I just hear a repeated "clicking" sound... weird

Thats very weird. I tried and I get another sound but same effect. I also made a mod for these be "permanent audiovisual" and it plays the sounds correctly..

core/src/com/unciv/ui/audio/CitySoundPlayer.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import com.badlogic.gdx.Files
+import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.audio.Music
+import com.badlogic.gdx.files.FileHandle
+import com.unciv.UncivGame
+import com.unciv.logic.city.CityInfo
+import com.unciv.models.UncivSound
+
+class CitySoundPlayer: MusicController() {
```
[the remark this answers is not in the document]
alexban011: its subclassing it so I can call the functions for the hook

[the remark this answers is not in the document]
alexban011: Should I instead extend the MusicController and make it play 2 sounds at once ? (normal music + these sounds)

core/src/com/unciv/models/ruleset/Era.kt
```diff
@@ -29,6 +29,8 @@ class Era : RulesetObject(), IHasUniques {
     var baseUnitBuyCost = 200
     var embarkDefense = 3
     var startPercent = 0
+    var citySound = ""
+    var cityWLTKSound = ""
```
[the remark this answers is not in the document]
alexban011: I thought they would be needed if they are in the Eras.json
The cityWLTKSound definitely isn't used though, forgot to remove it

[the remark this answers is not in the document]
alexban011: ohh actually it is used in CitySoundPlayer line 47

alexban011:
Ohhhh I see. Its playing those sounds because the mod doesn't have the sounds. Maybe if I set a default sound it will fix this

core/src/com/unciv/UncivGame.kt
```diff
@@ -112,11 +115,19 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
         audioExceptionHelper?.installHooks(
             musicController.getAudioLoopCallback(),
             musicController.getAudioExceptionHandler()
         )
+
+        citySoundController = CitySoundPlayer()
+        audioExceptionHelper2?.installHooks(
+            citySoundController.getAudioLoopCallback(),
+            citySoundController.getAudioExceptionHandler()
+        )
```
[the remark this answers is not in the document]
alexban011: I tested and it doesn't stop the music. Do you mean that the music player won't be able to play music after this one finishes?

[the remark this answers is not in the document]
alexban011: Yeah you're right. I moved everything in the music controller

alexban011:
I think the city sounds should have their own volume. These are completely different than the normal sounds and it would make sense right? Also what if you don't want to hear these but still have the sounds? I think I'll add the volume slider back

android/assets/jsons/Civ V - Vanilla/Eras.json
```diff
@@ -167,11 +177,13 @@
...
         "startingObsoleteWonders": ["Temple of Artemis", "Stonehenge", "The Great Library", "Mausoleum of Halicarnassus", "The Pyramids", "Statue of Zeus",
             "The Great Lighthouse", "Hanging Gardens", "The Oracle", "Great Wall", "Colossus",
-            "Hagia Sophia", "Chichen Itza", "Machu Picchu", "Angkor Wat", "Notre Dame", 
+            "Hagia Sophia", "Chichen Itza", "Machu Picchu", "Angkor Wat", "Notre Dame",
             "Sistine Chapel", "Forbidden Palace", "Himeji Castle", "Taj Mahal", "Porcelain Tower", "Kremlin"],
         "baseUnitBuyCost": 600,
         "embarkDefense": 13,
         "startPercent": 65,
+        "citySound": "cityModern",
+        "cityWLTKSound": "cityWLTKNew",
```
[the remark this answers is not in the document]
alexban011: ohh yes right :))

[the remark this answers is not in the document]
alexban011: ohh but I did..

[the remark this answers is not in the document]
alexban011: ohh but only for G&K

core/src/com/unciv/ui/audio/MusicController.kt
```diff
@@ -476,4 +483,52 @@ class MusicController {
...
+        try {
+            if (playingCitySound?.isPlaying == true)
+                playingCitySound?.stop()
+        } catch (ex: Throwable) {
+            playingCitySound?.dispose()
+            Log.error("Error while stopping city sound: ", ex)
+        }
+    }
+
+    //endregion
```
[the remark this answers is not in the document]
alexban011: ahhhhh 
I reverted that commit back. But if I remove `: MusicController()` from `class CitySoundPlayer` I can't install the hooks without those 2 methods which I don't really know what to put in them.. leave them empty? And actually I shouldn't install the hooks as that would stop the music...

But if I don't install the hooks wouldn't it create the issues mentioned by SomeTroglodyte ?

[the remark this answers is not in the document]
alexban011: > why would it?

to avoid these errors

https://github.com/yairm210/Unciv/blob/9d5b3d4e4a9979d7b52bf0d9580309c672f960ea/desktop/src/com/unciv/app/desktop/HardenGdxAudio.kt#L30

Or is the new audio already installed and that would catch errors which happen in here too?

alexban011:
Hope I got them all :)

alexban011:
> Unfortunately I've found another thing :/ sorry

no need to be sorry. Better to find these before players find them

> Currently, when you exit the screen via Esc

fixing it now

alexban011:
> Yep, this is pretty great :D Good job! Definitely a nice addition.

Thanks! UX is very important, especially for new players, and I think this improves that :)

core/src/com/unciv/ui/audio/CityAmbiencePlayer.kt
```diff
@@ -0,0 +1,63 @@
...
+        } catch (ex: Throwable) {
+            playingCitySound?.dispose()
+            Log.error("Error while playing city sound: ", ex)
+        }
+    }
+
+    fun stop() {
+        try {
+            if (playingCitySound?.isPlaying == true)
+                playingCitySound?.stop()
```
alexban011, line 57: hmm should I dispose() instead of stop() ?

alexban011: well then.. I removed it to keep the code as clean as possible :)

PR #7171, opened 2022-06-15
More shortcuts

doublep:
* `Tab` to perform unit movement: click anywhere on the map, press `Tab` or (as now) click on the shown arrow button.
* When no unit is selected, i.e. when no unit action buttons are shown, use `Z` as a shortcut for "Next unit" (same as `N` or `SPC`). This streamlines cycling through units from keyboard.


SomeTroglodyte (approved):
Simple and useful.. For those in the know. Reminds me there's still a few things there should be some Civilopedia help screen for...

PR #7181, opened 2022-06-16
small fixes

alexban011:
I added a suppress unused parameter on the translation variable and cutout checker to stop displaying that message as it was messing with my feng shui

Additionally I've noticed a couple days ago that I forgot to delete a leftover from the old friend picker UI, so I removed it now

### DIFF
--- a/android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
+++ b/android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
@@ -35,6 +35,7 @@ Sources for Info about current orientation in case need:
         if (activity.requestedOrientation != orientation) activity.requestedOrientation = orientation
     }
 
+    @Suppress("DEPRECATION")
     override fun hasDisplayCutout(): Boolean {
         return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
             activity.windowManager.defaultDisplay.cutout != null
--- a/android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
+++ b/android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
@@ -35,13 +35,15 @@ Sources for Info about current orientation in case need:
         if (activity.requestedOrientation != orientation) activity.requestedOrientation = orientation
     }
 
-    @Suppress("DEPRECATION")
-    override fun hasDisplayCutout(): Boolean {
-        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+    override fun hasDisplayCutout() = when {
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.R ->
+            activity.display?.cutout != null
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q ->
+            @Suppress("DEPRECATION")
             activity.windowManager.defaultDisplay.cutout != null
-        } else {
-             false
-        }
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.P ->
+            activity.window.decorView.rootWindowInsets.displayCutout != null
+        else -> false
     }
 
     override fun toggleDisplayCutout(androidCutout: Boolean) {
--- a/core/src/com/unciv/ui/newgamescreen/PlayerPickerTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/PlayerPickerTable.kt
@@ -50,7 +50,6 @@ class PlayerPickerTable(
 ): Table() {
     val playerListTable = Table()
     val civBlocksWidth = if(blockWidth <= 10f) previousScreen.stage.width / 3 - 5f else blockWidth
-    val friendsBlocksWidth = if(blockWidth <= 10f) previousScreen.stage.width / 5 - 5f else blockWidth / 2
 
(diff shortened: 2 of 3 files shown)


tests/src/com/unciv/testing/TranslationTests.kt
```diff
@@ -171,7 +171,7 @@ class TranslationTests {
 
         var allWordsTranslatedCorrectly = true
         for (translationEntry in translations.values) {
-            for ((language, translation) in translationEntry) {
+            for ((language, @Suppress("UNUSED_PARAMETER") translation) in translationEntry) {
```
SomeTroglodyte, line 174: `for (language, _)` exists for such a case

android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
```diff
@@ -35,6 +35,7 @@ Sources for Info about current orientation in case need:
         if (activity.requestedOrientation != orientation) activity.requestedOrientation = orientation
     }
 
+    @Suppress("DEPRECATION")
```
SomeTroglodyte, line 38: That won't help your Feng Shui much, as the code will FAIL on Pie and below. I may have done something like
```kotlin
    override fun hasDisplayCutout() = when {
        Build.VERSION.SDK_INT >= Build.VERSION_CODES.R ->
            activity.display?.cutout != null
        Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q ->
            @Suppress("DEPRECATION")
            activity.windowManager.defaultDisplay.cutout != null
        Build.VERSION.SDK_INT >= Build.VERSION_CODES.P ->
            true  //todo G00gle tells us cutouts work on Pie but not how to determine whether there are any
        else -> false
    }
```
..and have _loudly_ cursed the API devs (Long.MAX_VALUE)-fold for f(/!"&(!ing that API up so convolutedly. With this, you would enable the cutout toggle on Pie, but it would do nothing if there isn't any? Probably. Or remove the P->true to just not support cutouts on Pie. Note the R branch doesn't do more or better, it just would help the future dev that wishes to raise minApi to 30.

alexban011: > `activity.display?.cutout != null`

the SDK code is broken. That function returns null if the phone is rotated to the right and works in the rest of the cases :/

SomeTroglodyte: P.S.: For Pie, a `ViewCompat.getRootWindowInsets(View(activity))?.displayCutout != null` just might work - or not. Just a wild guess. It would create and right after GC a dummy view for the activity instead of using the actual one, which I have no idea how to get hold of. But WindowInsetsCompat.displayCutout seems to be the correct and only available approach, and ViewCompat.getRootWindowInsets "sounds like" it would be a viable way to get hold of a valid instance of WindowInsetsCompat (the official docs are maximally unhelpful on that - evil intent couldn't have produced a less useful doc).

alexban011: the problem is that `WindowInsetsCompat.displayCutout` always null regardless if you have a cutout or not. 
I tried with `WindowInsetsCompat.CONSUMED.displayCutout != null` but it just returns false

SomeTroglodyte: In that case I defer to your better expertise.

I guess WindowInsetsCompat instances can be obtained from different sources, and one of them may do the cutouts right - most WindowInsetsCompat samples out there get theirs through an override, but that one lives on the ViewCompat _instance_ which I wouldn't know where to place. Android native UI API model requires some serious brain defect to understand. And while my brain _is_ damaged (or I wouldn't waste time here), the damage obviously doesn't meet requirements.

core/src/com/unciv/ui/newgamescreen/PlayerPickerTable.kt
```diff
@@ -50,7 +50,6 @@ class PlayerPickerTable(
 ): Table() {
     val playerListTable = Table()
     val civBlocksWidth = if(blockWidth <= 10f) previousScreen.stage.width / 3 - 5f else blockWidth
```
SomeTroglodyte, line 52: Training until it becomes a no-brainer: if -> blank

android/src/com/unciv/app/PlatformSpecificHelpersAndroid.kt
```diff
@@ -35,12 +35,15 @@ Sources for Info about current orientation in case need:
...
+            activity.display?.cutout != null
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q ->
+            @Suppress("DEPRECATION")
             activity.windowManager.defaultDisplay.cutout != null
-        } else {
-             false
-        }
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.P ->
+            activity.window.decorView.rootWindowInsets.displayCutout != null
+        else -> false
```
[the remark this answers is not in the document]
SomeTroglodyte: Exactly - it's no fun. The whole hardware development is sick to the core. They should certainly be able by now to allow a speaker to project vibrations _through_ the display and a camera to look _through_ it, too - at least for LCD that tech exists. Better yet - leave at least two non-display bevels, you need them anyway for e.g. bike holders etc., and 3mm would be enough nowadays for that crap.

SomeTroglodyte (approved):

PR #7184, opened 2022-06-16
Fix unit action/selection overlay not being closed when performing an action with a new unit

Azzurite:
Discovered in #7042, this was only being done on key press, but it needs to be done on click as well.


SomeTroglodyte:
As I remember right-click also... :speak_no_evil: Oops, forget I said anything...

PR #7186, opened 2022-06-16
Make 'Home Home' shortcut always open capital's city screen

doublep:
This actually work already, but *not always*: you need to wait till scrolling finishes before issuing the second keypress. After this commit this waiting becomes optional. Also it topples some common code of `setCenterPosition()` of the two map holder classes to the common base: `ZoomableScrollPane`.

### DIFF
--- a/core/src/com/unciv/ui/mapeditor/EditorMapHolder.kt
+++ b/core/src/com/unciv/ui/mapeditor/EditorMapHolder.kt
@@ -1,6 +1,5 @@
 package com.unciv.ui.mapeditor
 
-import com.badlogic.gdx.math.Interpolation
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.scenes.scene2d.Action
 import com.badlogic.gdx.scenes.scene2d.EventListener
@@ -9,7 +8,6 @@ import com.badlogic.gdx.scenes.scene2d.InputListener
 import com.badlogic.gdx.scenes.scene2d.Stage
 import com.badlogic.gdx.scenes.scene2d.Touchable
 import com.badlogic.gdx.scenes.scene2d.actions.Actions
-import com.badlogic.gdx.scenes.scene2d.actions.FloatAction
 import com.unciv.UncivGame
 import com.unciv.logic.HexMath
 import com.unciv.logic.map.TileInfo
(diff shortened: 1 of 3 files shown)


SomeTroglodyte:
> not always

Kind of you to notice - I thought nobody used these... And yea I was too lazy to think of a cleverer solution (`scrollingAction != null && actions.any { it == scrollingAction }` - proves my point: clever as in not immediately obvious). From a quick code scan (~3s) I'd say well done. Unifying some holder stuff is overdue - haven't really investigated though whether some intermediate class between ZSP and the end classes would be cleaner. There were other common aspects if I'm remembering right.. Could also make the MainMenu map holder cleaner, at the moment it's reusing and letting most of the class lie disabled. Just musing, not criticism.

Also, can't the 'so we invert it back' bit also be folded into common code? Nah, you'll have considered that and have your reasons.

Lastly - I still don't understand why the maps do their own panning actions when the basic Gdx ScrollPane already has them - are they uglier or have some quirk... Did you get to understand that detail while you did this?

doublep:
> Kind of you to notice - I thought nobody used these... 

Found out about `Home` when doing the shortcut refactoring and noticed from use that it was clunky in the described way. It's actually a nice shortcut, but was a pain to use because of the inconsistent behavior.

> Unifying some holder stuff is overdue - haven't really investigated though whether some intermediate class between ZSP and the end classes would be cleaner.

Certainly possible, but my goal was only `Home` handling ;) I added a little bit of refactoring in the immediately affected functions only, as otherwise I'd have to do the changes I wanted to do twice.

> Also, can't the 'so we invert it back' bit also be folded into common code? Nah, you'll have considered that and have your reasons.

I simply decided to leave these as they are because they involve `tileGroup` not present in the superclass and I also didn't want to split arithmetics between two calls: might create more problems than it would add usefulness.

> Lastly - I still don't understand why the maps do their own panning actions when the basic Gdx ScrollPane already has them - are they uglier or have some quirk... Did you get to understand that detail while you did this?

No, I haven't even looked, was "outside of scope" for me.

doublep:
> what is the "Home Home" shortcut? How do I trigger it?

It's actually `Home` shortcut. It scrolls to your capital or, if the map is already there, it opens capital's city screen. So, in theory, pressing `Home` twice would always get you to the city screen of the capital. However, this didn't quite work that way, because while the map was scrolling, second `Home` would just start scrolling again; it would took you to the city screen only if the map had finished scrolling by the time you pressed the key.

> It doesn't seem to be touched by this PR? I generally don't understand :D

The PR affects return value of `setCenterPosition()` when already scrolling. This was enough to make the shortcut (second press) behave consistently.

SomeTroglodyte:
> that "Home Home" meant the "Home" key on the keyboard

No Dorothy involved!

> second Home would just start scrolling again

... which I always knew and still was too lazy to fix. Thanks!

SomeTroglodyte (approved):

PR #7190, opened 2022-06-17
Fix memory leaks in BaseScreen & gdx.Stage/gdx.TextFields

Azzurite:
**BaseScreen memory leak**

Currently, `BaseScreen` leaks graphics memory by not `dispose()`ing the `SpriteBatch` it creates. We previously did this:

https://github.com/yairm210/Unciv/blob/0533c9eabade88adc3c9b65115288c645663af94/core/src/com/unciv/ui/utils/BaseScreen.kt#L49

but the [documentation of that constructor](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/Stage.java#L107) says: "`batch` Will not be disposed if `Stage.dispose()` is called, handle disposal yourself". We neither called `Stage.dispose()` nor did we ever call `SpriteBatch.dispose()` (for which we didn't even keep a reference).

**TextField/Stage memory leak**

The second memory leak that is fixed here is one more subtle but the one that actually caused this refactoring: 

[This little task](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L309) in `gdx.TextField` creates a new `Thread` that makes the cursor in `TextField`s blink. The `Thread` is only shut down [a few lines above](https://github.com/libgdx/libgdx/blob/bf32f0fbebeab7df0b0f257ea0fd3d9fbc549eff/gdx/src/com/badlogic/gdx/scenes/scene2d/ui/TextField.java#L306) - but only when the keyboard focus changes. So if you don't do `Stage.dispose()` (which removes all actors, also clearing keyboard focus), then this little task never gets stopped and the thread holds a reference to the stage indefinitely.

**Solution**

So I thought, why not just remove the `SpriteBatch` parameter (so that the `Stage` creates a `SpriteBatch` itself and `dispose()`s it) and call `stage.dispose()` when the screen is `dispose()`? And that worked nicely... but we previously `dispose()`d screens in `UncivGame.setScreen`. That means this change breaks all screens that hold a reference to a "previous screen" to go back to - the new screen gets set, the old screen (that we want to switch back to) gets `dispose()`d, clearing all actors on it, thus going back to it just shows an empty screen.

On this old system, there was no way to fix this. So I had to remove this old system of screens that hold a reference to previous screens completely. Now, `UncivGame` handles all screens in a `screenStack`. Instead of `setScreen` calls, there are now only `pushScreen` and `popScreen` and `replaceCurrentScreen` calls. All usages of `setScreen` have been replaced with calls to these new methods. 

I also changed some usages of `resetToWorldScreen` to `popScreen` - for example, when you previously went to the empire overview screen to the "Cities" tab, and entered a city with its button, and you then pressed "Exit city", you arrived back at the world screen. Now, you arrive at the empire overview screen again in the "Cities" tab, even keeping your scroll position - majestic.

### DIFF
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -37,6 +37,7 @@ import com.unciv.utils.concurrency.withGLContext
 import com.unciv.utils.concurrency.withThreadPoolContext
 import com.unciv.utils.debug
 import java.util.*
+import kotlin.collections.ArrayDeque
 
 class UncivGame(parameters: UncivGameParameters) : Game() {
     // we need this secondary constructor because Java code for iOS can't handle Kotlin lambda parameters
@@ -82,8 +83,9 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
     private val wrappedCrashHandlingRender = { super.render() }.wrapCrashHandlingUnit()
     // Stored here because I imagine that might be slightly faster than allocating for a new lambda every time, and the render loop is possibly one of the only places where that could have a significant impact.
 
-
     val translations = Translations()
+
+    val screenStack = ArrayDeque<BaseScreen>()
 
     override fun create() {
         isInitialized = false // this could be on reload, therefore we need to keep setting this to false
@@ -144,8 +146,8 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
                 ImageGetter.ruleset = RulesetCache.getVanillaRuleset() // so that we can enter the map editor without having to load a game first
 
                 when {
-                    settings.isFreshlyCreated -> setScreen(LanguagePickerScreen())
-                    deepLinkedMultiplayerGame == null -> setScreen(MainMenuScreen())
+                    settings.isFreshlyCreated -> pushScreen(LanguagePickerScreen())
+                    deepLinkedMultiplayerGame == null -> pushScreen(MainMenuScreen())
                     else -> tryLoadDeepLinkedGame()
                 }
 
@@ -154,7 +156,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
         }
     }
 
-    /** Loads a game, initializing the state of all important modules. Automatically runs on the appropriate thread. */
+    /** Loads a game, [disposing][BaseScreen.dispose] all other screens. Initializes the state of all important modules. Automatically runs on the appropriate thread. */
     suspend fun loadGame(newGameInfo: GameInfo): WorldScreen = withThreadPoolContext toplevel@{
         val prevGameInfo = gameInfo
         gameInfo = newGameInfo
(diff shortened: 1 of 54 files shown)


core/src/com/unciv/UncivGame.kt
```diff
@@ -212,67 +221,84 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
         loadGame(curGameInfo)
     }
 
-    private data class NewScreens(val screenToShow: BaseScreen, val worldScreen: WorldScreen) {
-        constructor(worldScreen: WorldScreen) : this(worldScreen, worldScreen)
-    }
-
     /**
      * Sets the screen of the game and automatically disposes the old screen as long as it isn't the world screen.
```
SomeTroglodyte, line 225: doc outdated

core/src/com/unciv/UncivGame.kt
```diff
@@ -212,67 +221,84 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
+     *
+     * @return the new screen
      */
-    fun resetToWorldScreen() {
-        setScreen(worldScreen!!)
+    fun popScreen(): BaseScreen? {
+        val oldScreen = screenStack.removeLast()
+        val newScreen = screenStack.lastOrNull()
+        if (newScreen == null) {
+            return null
```
SomeTroglodyte, line 264: Is this an error never to happen or.. is pop last living then push new life allowed? Wouldn't in the latter case it be possible to miss an oldScreen.dispose()?

core/src/com/unciv/UncivGame.kt
```diff
@@ -212,67 +221,84 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
+        }
+        setScreen(newScreen)
+        if (oldScreen !is WorldScreen) { // we want to keep the world screen around, because it's expensive to re-create it
```
SomeTroglodyte, line 267: Sounds like a dangerous pitfall requiring outside code to pay attention. Who would be keeping a viable reference and be able to reuse it? Sounds like that expertise may be better encapsulated close to these routines here, maybe even do a dedicated BaseScreenStack class...

core/src/com/unciv/UncivGame.kt
```diff
@@ -212,67 +221,84 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
+            oldScreen.dispose()
+        }
+        return newScreen
+    }
+
+    /** Replaces the current screen with a new one. Automatically [disposes][BaseScreen.dispose] the old screen as long as it isn't the world screen. */
```
SomeTroglodyte, line 273: 'as long as it isn't the world screen' doesn't seem to be true

core/src/com/unciv/ui/pickerscreens/PromotionPickerScreen.kt
```diff
@@ -112,16 +113,15 @@ class PromotionPickerScreen(val unit: MapUnit) : PickerScreen() {
         displayTutorial(Tutorial.Experience)
     }
 
-    private fun setScrollY(scrollY: Float): PromotionPickerScreen {
+    private fun setScrollY(scrollY: Float) {
         splitPane.pack()    // otherwise scrollPane.maxY == 0
         scrollPane.scrollY = scrollY
         scrollPane.updateVisualScroll()
-        return this
```
SomeTroglodyte, line 119: Looks like you're an abolitionist

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -206,12 +206,7 @@ class DiplomacyScreen(
                 resourcesTable.add(wrapper).padRight(20f)
                 wrapper.addTooltip(name, 18f)
                 wrapper.onClick {
-                    val pedia = CivilopediaScreen(
-                        UncivGame.Current.gameInfo!!.ruleSet,
-                        this,
-                        link = "Resource/$name"
-                    )
-                    UncivGame.Current.setScreen(pedia)
+                    UncivGame.Current.pushScreen(CivilopediaScreen(UncivGame.Current.gameInfo!!.ruleSet, link = "Resource/$name"))
```
SomeTroglodyte, line 209: not your code, I just thought `supplyList.resource.makeLink()` - not really important.

SomeTroglodyte (approved):
All other comments are just food for thought.

I already decided this must be approvable from the first mail alone. The whole concept tastes cleaner than what we had. Will likely need a few fixes, but I think it's a move better done as early as possible.

One more tiny nudge:
- Shouldn't there be some communication between the screen stack and P²'s key handling?

SomeTroglodyte:
> missed while you were gone is this: #7145

Not entirely. I _had_ skimmed the worst of it. But - I'll shup up now, can barely keep up. 2h for ~70 mails; and that's the time budget for this weekend gone. Now where's the 35 Torx so I can leave this dank cave... Or was it the 40 one to fix the handlebar always drifting off-kilter after any non-trivial mountain descent?

PR #7193, opened 2022-06-17
Fix crash when base ruleset contains no strategic resources

Skekdog:
Accidentally did 4 instead of just the one, they all work fine though the latest one is ideal


SomeTroglodyte (approved):

SomeTroglodyte:
Thanks. Counting to two comes _after_ [popping empty the mail "stack"](https://topatoco.com/products/dc-emptyinbox).

PR #7195, opened 2022-06-17
Scroll WorldScreen immediately to the city when exiting it

Azzurite:
Currently there is an animation from your previous point on the map to the current city. However, you likely are already at or very close to the city anyway, and the city screen shows the city tiles also. So when you exit the city screen, it's simply visually scarring to get moved to somewhere else and then back to the city.

This disables this animation, showing the center of the city in the middle of the screen immediately, which feels much more natural.


SomeTroglodyte (approved):

yairm210 (approved):
Conflicts

PR #7198, opened 2022-06-17
Nicer multiplayer settings

Azzurite:
Current:

image

After this PR:

image

### DIFF
--- a/core/src/com/unciv/ui/options/MultiplayerTab.kt
+++ b/core/src/com/unciv/ui/options/MultiplayerTab.kt
@@ -12,8 +12,11 @@ import com.unciv.models.metadata.GameSetting
 import com.unciv.models.metadata.GameSettings
 import com.unciv.models.ruleset.Ruleset
 import com.unciv.models.ruleset.RulesetCache
+import com.unciv.ui.images.ImageGetter
 import com.unciv.ui.popup.Popup
 import com.unciv.ui.utils.BaseScreen
+import com.unciv.ui.utils.extensions.addSeparator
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/options/OptionsPopup.kt
```diff
@@ -215,7 +222,7 @@ open class SettingsSelect<T : Any>(
     }
 
     fun addTo(table: Table) {
-        table.add(label).left()
+        table.add(label).expandX().fillX().left()
```
SomeTroglodyte, line 225: same as growX()

core/src/com/unciv/ui/options/MultiplayerTab.kt
```diff
@@ -274,3 +283,13 @@ private fun createRefreshOptions(unit: ChronoUnit, vararg options: Long): List<S
         SelectItem(duration.format(), duration)
     }
 }
+
+private fun addSelectAsSeparateTable(tab: Table, settingsSelect: SettingsSelect<*>) {
```
SomeTroglodyte, line 287: Sounds like SettingsSelect should have been conceived as Widget from the start :smiling_imp:  - at least some hint on why wrapping is needed instead of, say, colspan?

SomeTroglodyte: Nah. "K.I.S.S." - as long as there's thought behind the decisions and not oversight..

yairm210 (approved):
Definite improvement :)

Issue #7183, opened 2022-06-16
Feature request: Update Multiplayer Server Endpoint on Redirection


touhidurrr:
This behaviour includes:
1. If `host.example.com/isalive` redirects to any url that ends with `/isalive` than replace `multiplayerServerUrl` with part of the url before `/isalive`
    example:
        redirect: `example.com/isalive` -> `example.com/api/isalive`
        new url to replace with: `example.com/api`
2. Similarly replace `multiplayerServerUrl` for `hostname/files/:filename` too

This will allow community server owners to change their hostname easily without negatively affecting players for special circumstances.

PR #7200, opened 2022-06-17
Add Uncivserver.xyz community server in the wiki

Azzurite:
Is this something we want to do? I'd be ok with it :)


touhidurrr:
Lol, the disclaimers looks way to fishy...
Just wanted to say it was default Cloudflare stats with no retention.
And I actually do not collect IP addresses as the source code is on Github.
But I do collect Discord snowflakes for the users of Discord Notifications.

That's all.

touhidurrr:
> These servers may become (temporarily or permanently) unavailable and lose your saves.

Also #7183 might actually be helpful for such cases. Since free services having difficulties keeping them alive happens regularly on the internet. Outrages on free hosting is commonplace.

It would be good if someone looked into the feature request since we have this disclaimer now.

touhidurrr:
> I don't think it's too fishy, I think it's appropriately fishy. People do have to trust you with the things I mentioned there.

I know. Just kidding.
And there should be such disclaimer for any 3rd party service anyways.

Issue #7199, opened 2022-06-17
Explorers doesn’t collect ruin with world wrap on


touhidurrr:
**Platform**
Android 11

**Version**
4.1.11

**Describe the bug**
![Screenshot_2022-06-18-02-42-52-51_5b83b50424e447777d9855d9701553df](https://user-images.githubusercontent.com/46617994/174400145-35bf460c-55c9-4fba-b506-904dcd084db5.jpg)
As you can see the explorer didn’t take the path to the ruins.

**Game Data**
[Korea.json.txt](https://github.com/yairm210/Unciv/files/8931302/Korea.json.txt)

AlexCaesel:
This happens too with the initial soldier?

touhidurrr:
closed because of #7285 & #7314

PR #7205, opened 2022-06-18
Exploring units can take several turns to access a ruin

Skekdog:


OptimizedForDensity:
Tested the fix using the saved data from #7199 and it works regardless of that line of code. In the future, you may want to create separate branches for individual pull requests so that you don't end up trying to commit the same section of code in multiple pull requests.

Skekdog:
> Tested the fix using the saved data from #7199 and it works regardless of that line of code. 
I couldn't figure out how to get past "Waiting for Korea", but in a simple test scenario I set up the change I made helps
Without this change, unit ignores ruins if it can't immediately access them
With this, unit can take some turns to get there

I should probably note that the unit will only go after ruins it can see - not grayed out and not behind fog of war

With this change applied, the unit moves to the ruin
image

Without it, the unit navigates down to the grassland and leaves

Skekdog:
> In the future, you may want to create separate branches for individual pull requests so that you don't end up trying to commit the same section of code in multiple pull requests.

Sorry, new to github

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -48,16 +49,16 @@ object UnitAutomation {
...
+        val tileWithRuinOrEncampment = unit.viewableTiles
             .firstOrNull {
                 (
                     (it.improvement != null && it.getTileImprovement()!!.isAncientRuinsEquivalent())
                     || it.improvement == Constants.barbarianEncampment
                 )
                 && unit.movement.canMoveTo(it)
             } ?: return false
-        unit.movement.moveToTile(tileWithRuinOrEncampment)
+        unit.movement.headTowards(unit.movement.getShortestPath(tileWithRuinOrEncampment,true)[0])
```
OptimizedForDensity, line 61: (reposting this because I think I messed up the previous comment)

This could be simplified with unit.movement.headTowards(tileWithRuinOrEncampment) since that already calls unit.movement.getShortestPath(destination) if it takes more than a turn to reach the ruins (only difference being it doesn't avoid damaging terrain but I don't know how much that would make a difference) (also I don't know why this comment says pending...).

Android Studio also says import com.unciv.models.ruleset.Nation is unused.

Skekdog: > This could be simplified with `unit.movement.headTowards(tileWithRuinOrEncampment)` since that already calls `unit.movement.getShortestPath(destination)`

If that's the case then alright

> Android Studio also says `import com.unciv.models.ruleset.Nation` is unused.

:+1:

Skekdog: I've updated the file to include these fixes

core/src/com/unciv/ui/worldscreen/WorldScreenTopBar.kt
```diff
@@ -158,9 +158,13 @@ class WorldScreenTopBar(val worldScreen: WorldScreen) : Table() {
         }
 
         // in case the icons are configured higher than a label, we add a dummy - height will be measured once before it's updated
-        resourcesWrapper.add(resourceActors[0].icon)
+        if (resourceActors.isNotEmpty()) {
+            resourcesWrapper.add(resourceActors[0].icon)
```
yairm210, line 162: If resource wrapper is only used in this 'if' case, we should move all the initialization if this object to within the if

SomeTroglodyte: Not Skekdog's fault. I probably expected more cross-method reuse.

Skekdog: I've updated WorldScreenTopBar to include this change
resourceTable still has to be initialized because of return resourceTable, and I kept in the resourceTable padding because it looked better and allowed the turnsLabel to not overlap with the civ name

Skekdog:
Just realised that last commit merged a whole bunch of other changes not done by me... I'm really not good at this
Will fix soon

Skekdog:
Moved to #7285

PR #7207, opened 2022-06-18
Various UX improvements

Azzurite:
I separated all of them by commit:

* [Refactor: Format JSON](https://github.com/yairm210/Unciv/pull/7207/commits/a45b68beedc83aff5d9cdc586402688d5bb1d0ac)
   All our other JSONs use 4 spaces
* [Change button hover color to be darker](https://github.com/yairm210/Unciv/pull/7207/commits/35504f04d1e5ccd20a7f4d7a283b3f84dc8e81b4) 
   Previous:
   image
   After, intention is to have better contrast:
   image
* [Refactor: Change disabled buttons to always look disabled, even if they don't have the default style](https://github.com/yairm210/Unciv/pull/7207/commits/e41cd57b42b26d0817864b1b0b1d84684b538e17)
   In preparation for the usage of `TextButtonStyle` instead of `button.color = ...`
* [Make disabled style a little grayer](https://github.com/yairm210/Unciv/pull/7207/commits/cc976ac9af194ef7545f785e8343f14924059006)
   Previous:
   image
   After, intention is to convey being "disabled" a little better:
   image
* [Refactor: Make addCloseButton not start a new row](https://github.com/yairm210/Unciv/pull/7207/commits/9b1df8ed3d9b54523ac8845a0ce6775f00dda730)
   This basically doesn't matter anyway since `addCloseButton` is the last addition to the popup in all places, so an implicit row will be created.
* [Refactor: Remove "...inRow" popup button methods & add button style parameter](https://github.com/yairm210/Unciv/pull/7207/commits/5c37cc268428b4f55882ae23018f43855b5be473)
   `addButton` always adding a row is confusing, since `add` doesn't add a row. So I simply removed all `...inRow` functions and let the users add `.row()` manually. It's only 1 character more, no idea why this helper method was ever necessary.
* [Reorder "ok/cancel" popups to have "ok" on the right side](https://github.com/yairm210/Unciv/pull/7207/commits/f702d8d4158fd18d81bc39c72714df6f7b575219)
   Just look at this [web image search](https://duckduckgo.com/?q=android+dialog&t=ffsb&iar=images&iax=images&ia=images) - all dialogs on Android have the "Yes/Confirm/OK" on the right side.
* [Fix pointer enter/exit events being disabled by default](https://github.com/yairm210/Unciv/pull/7207/commits/d90a13de054f684cbce6271e109c34bdc3059b6a)
   When I implemented this, I initially started with `disablePointerEnterExitEvents` and  then later refactored to `performPointerEnterExitEvents`, but I forgot to change the default value.
* [Fix equalizeLastTwoButtonWidths not setting the cells to be equal size](https://github.com/yairm210/Unciv/pull/7207/commits/225d5b4d8b9832be56cd44ad1d56c3319d68a3a1)
   Previous (buttons weren't centered):
   image
   After (buttons are centered):
   image
* [Change YesNoPopup to ConfirmPopup](https://github.com/yairm210/Unciv/pull/7207/commits/72c8447e41adc8250556bdd6710ecfbca10b49b2)
   Previous:
   image
   After:
   image
* [Add ConfirmPopup to deletion of save files](https://github.com/yairm210/Unciv/pull/7207/commits/7db0a7a797c5c920f04e68b244e6f55013a9d1a9)
   Previously you weren't asked when deleting a save, now you are.

### DIFF
--- a/core/src/com/unciv/ui/popup/Popup.kt
+++ b/core/src/com/unciv/ui/popup/Popup.kt
@@ -9,6 +9,7 @@ import com.badlogic.gdx.scenes.scene2d.ui.Cell
 import com.badlogic.gdx.scenes.scene2d.ui.Label
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.scenes.scene2d.ui.TextButton
+import com.badlogic.gdx.scenes.scene2d.ui.TextButton.TextButtonStyle
 import com.badlogic.gdx.scenes.scene2d.ui.TextField
 import com.badlogic.gdx.utils.Align
 import com.unciv.Constants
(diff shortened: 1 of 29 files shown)


yairm210 (approved):

core/src/com/unciv/ui/popup/Popup.kt
```diff
@@ -109,51 +110,38 @@ open class Popup(
     }
 
     /**
-     * Adds an inline [TextButton].
+     * Adds a [TextButton] and ends the current row.
```
SomeTroglodyte, line 113: "and ends the current row" - nope

core/src/com/unciv/ui/popup/Popup.kt
```diff
@@ -109,51 +110,38 @@ open class Popup(
...
-    /**
-     * Adds a [TextButton] and ends the current row.
-     * @param text The button's caption.
-     * @param key Associate a key with this button's action.
-     * @param action A lambda to be executed when the button is clicked.
-     * @return The new [Cell]
-     */
-    fun addButton(text: String, key: KeyCharAndCode? = null, action: () -> Unit)
-        = addButtonInRow(text, key, action).apply { row() }
     /** @link [addButton] */
```
SomeTroglodyte, line 125: Not from this PR, but that one (@ link) doesn't work on Android Studio, not that I can see. If it did, it would cover only one of the two following overloads, right?

SomeTroglodyte:
Really? **No** _simple_ questions anymore?
image
(And no, the ConfirmPopin API isn't symmetrical enough to allow this un-kludged - not that we need it)

SomeTroglodyte:
> Refactor: Format JSON

Despite the link a "the Skin json" would be clearer to the casual reader.
...So you tackled _that_ beast --- kudos. There's tons of styling done on-the-fly that would be better at home in the Skin, probably most perpetrated by myself...

Thought: The key bindings for yes/no popups now make a tad less sense than before. Might need addressing some day.

SomeTroglodyte (approved):

PR #7209, opened 2022-06-18
Translated strings after generating translation files from Adv Opt.

J0anJosep:
After generating translation files from the Advanced Options Window, show both success strings translated.

One string was already in template.properties. The other one wasn't. I added it manually, but I am not sure this should be done like this.


android/assets/jsons/translations/template.properties
```diff
@@ -710,6 +710,7 @@ HIGHLY EXPERIMENTAL - YOU HAVE BEEN WARNED! =
 Enable portrait orientation = 
 Generate translation files = 
 Translation files are generated successfully. = 
+Fastlane files are generated successfully. =
```
OptimizedForDensity, line 713: This line needs to end with a space after =

J0anJosep: Thank you!

PR #7217, opened 2022-06-19
Update ReligionManager.kt

MindaugasRumsa51:
this function is used in Victory conditions not to determine major religion

### DIFF
--- a/core/src/com/unciv/logic/civilization/ReligionManager.kt
+++ b/core/src/com/unciv/logic/civilization/ReligionManager.kt
@@ -70,7 +70,7 @@ class ReligionManager {
     }
 
     fun isMajorityReligionForCiv(religion: Religion): Boolean {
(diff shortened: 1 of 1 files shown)


MindaugasRumsa51:
was: only 50% cities must have 50% your followers
now: every city in the world must have 50% your followers

MindaugasRumsa51:
maybe i was wrong

MindaugasRumsa51:
this domination condition still unclear: let's calculate: we have 5 cities:  1st=2 citizens 2nd=2 3rd=2 4rth=2 5th=20
only 5th city is my religion and all 20 is mine
20 > than 2+2+2+2. domination, yes? civ6=no
now al first 4 cities are my religion, 5th not mine
domination? NO, civ6=yes
the problem: when there are tons of cities you simply can't control and maybe it is even not possible full domination,
so civ6 decided to make step towards players

PR #7234, opened 2022-06-20
remove already removed tile improvements from lastseenimprovements

MindaugasRumsa51:


MindaugasRumsa51:
closing cause problem disappeared in last version

PR #7239, opened 2022-06-20
Fix AI great people getting stuck trying to build great improvements

OptimizedForDensity:
This PR fixes two cases where AI great people get stuck trying to build great improvements.

The first case happens because the AI doesn't check if they can actually build an improvement on a destination tile, usually because there's an oasis there. This is fixed by only considering tiles where the AI can build the improvement.

The second case occurs because the great person paths to the selected city until it is within 2 turns worth of movement from the city center. After that, the AI paths to the tile to build the improvement. In some cases, usually very large cities that have rough terrain, the AI chooses a tile that is 3 turns worth of movement away from the city center. This causes the great person to get stuck in an endless loop of trying to get to the selected tile then trying to path back to within 2 turns of movement of the city center. This is fixed by adding a condition to skip the path-to-city-center operation if they're already on a tile owned by the chosen city.

### DIFF
--- a/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/SpecificUnitAutomation.kt
@@ -257,14 +257,14 @@ object SpecificUnitAutomation {
             val applicableTiles = city.getWorkableTiles().filter {
                 it.isLand && it.resource == null && !it.isCityCenter()
                         && (unit.currentTile == it || unit.movement.canMoveTo(it))
-                        && !it.containsGreatImprovement()
+                        && !it.containsGreatImprovement() && it.canBuildImprovement(improvement, unit.civInfo)
             }
             if (applicableTiles.none()) continue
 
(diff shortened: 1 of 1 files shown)


yairm210 (approved):

yairm210:
That makes this your 30th PR, I think your skills and dedication show through plentifully - how would you feel about becoming a n official collaborator?


Brief overview of what this means:

- You get Admin powers on the repo because we trust you to not abuse them :)

- We currently require 2 acceptance votes for PRs to be merged - so if you think it's good add an 'accepted' review, if there is one already merge it in. Nothing makes me happier than arguing that people have name a PR and it's been merged without me knowing if its existence :)

- All releases are done via automation, so you can for all practical purposes release a new version. If for example a change opens the game to a new audience (like the M1 via libgdx upgrade), or the latest version had a serious bug and it was patched, you can add a tag to trigger the build and deploy.

That's basically all it entails at the moment. Of course, if you have any other ideas, you're more than welcome :)

PR #7240, opened 2022-06-20
Added civ names under their icon in diplomacy screen

alexban011:
https://user-images.githubusercontent.com/21990802/174677591-6e19c002-2b1c-4369-a0d1-5149affc99b8.mp4

### DIFF
--- a/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
+++ b/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
@@ -1,6 +1,7 @@
 package com.unciv.ui.trade
 
 import com.badlogic.gdx.graphics.Color
+import com.badlogic.gdx.scenes.scene2d.ui.Label
 import com.badlogic.gdx.scenes.scene2d.ui.SplitPane
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.scenes.scene2d.ui.TextButton
@@ -154,6 +155,14 @@ class DiplomacyScreen(
 
             leftSideTable.add(civIndicator).row()
 
+            val civName = Table()
+            val civNameLabel = civ.civName.toLabel()
+            civNameLabel.color = Color.BLACK
+            civName.add(civNameLabel)
+            civName.background = ImageGetter.getRoundedEdgeRectangle(civ.nation.getInnerColor())
+            leftSideTable.add(civName).row()
+
+            civName.onClick { updateRightSide(civ) }
             civIndicator.onClick { updateRightSide(civ) }
         }
 
--- a/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
+++ b/core/src/com/unciv/ui/trade/DiplomacyScreen.kt
@@ -155,14 +155,12 @@ class DiplomacyScreen(
 
             leftSideTable.add(civIndicator).row()
 
-            val civName = Table()
             val civNameLabel = civ.civName.toLabel()
-            civNameLabel.color = Color.BLACK
-            civName.add(civNameLabel)
-            civName.background = ImageGetter.getRoundedEdgeRectangle(civ.nation.getInnerColor())
-            leftSideTable.add(civName).row()
+            civNameLabel.color = civ.nation.getInnerColor()
 
-            civName.onClick { updateRightSide(civ) }
+            leftSideTable.add(civNameLabel).row()
+
+            civNameLabel.onClick { updateRightSide(civ) }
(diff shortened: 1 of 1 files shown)


alexban011:
> They should just use the default text color. 

and then add some padding below ? Cause if not it won't be intuitive which civ the text belongs to

alexban011:
this is with the .color commented. I don't know why it added that padding when I recorded

image

alexban011:
this is with .darken(-1f)

The only one which is kinda harder to read is "Hanoi", but if you look at the original video "Kuala Lumpur" now looks better

https://user-images.githubusercontent.com/21990802/174688732-f18b71f1-8a4d-4850-8df1-4f93e44ca627.mp4

alexban011:
Or should I add the names on the right, unless the game is in portrait mode?

Ouaz:
Make a poll on discord. :D 

I think it makes way more sense to have names on the right for landscape mode, and below for portrait mode (I never played in portrait mode, always in landscape, so I can't really say).

Great UI/QoL enhancement btw.

alexban011:
> the fuck is this :D 

I scrolled through the list of functions and somehow missed the brighten and just went for darken(-1f) :)))

> My preference

should we make a poll as Ouaz suggested ?

My preference would be Colored names, Landscape on the side, portrait below with padding

alexban011:
Correct. Then what about cherry-picking the names which don't look good and set different colors for those? or .brighten() them

Ouaz:
> My preference:
> Default color, Landscape on the side, portrait below with padding

I go with that too. Some people may have problems to read colored names. Even with good eyes, it's sometimes hard to read some colored nation names on the main menu (Hanoï, I say your name).

alexban011:
how much padBottom is that?

alexban011:
Well I got it in this state but it doesn't update unless you exit the screen and enter it back

alexban011:
updated the video in the main message

Ouaz:
On the the right in landscape mode.
Beneath in portrait mode.

Is this screenshot in portrait mode? :)

alexban011:
> Is that supposed to be like this? 

Well no, its not. I'm gonna try to put the name and icon in a table and see if that fixes the SplitPane going over the text

> Is this screenshot in portrait mode?

its during the transition between the two

alexban011:
Converted to draft as it takes a little longer than expected... 

I don't know why its doing this.. settting the leftSideTable in the SplitPane does what I need but obviously it won't scroll anymore and using the leftSideScroll it cuts the text with the SplitPane...

alexban011:
Yeah but the splitpane only starts cutting the text when the window gets close to 4:3 and the splitpane can also be moved if you do play at that weird aspect ratio, but in most situations it'll look good

alexban011:
Apparently Kuala Lumpur was being cut by the SplitPane, so I just placed the names under the icon all the time.

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -1008,4 +1012,7 @@ class DiplomacyScreen(
         super.resize(width, height)
         positionCloseButton()
     }
+    val selectedCiv = selectCiv
+    val selectedTrade = selectTrade
```
yairm210, line 1016: Are these needed?

alexban011: Yes. Without those it can't find them in the function below. I don't really understand why though. It happily accepts viewingCiv but the other 2 it doesn't

alexban011: I made them private now, but if you know how I could get them for the `recreate()` function below, I'll change this

core/src/com/unciv/ui/trade/DiplomacyScreen.kt
```diff
@@ -1008,4 +1012,8 @@ class DiplomacyScreen(
         super.resize(width, height)
         positionCloseButton()
     }
+
+    private val selectedCiv = selectCiv
+    private val selectedTrade = selectTrade
```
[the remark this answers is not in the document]
alexban011: I already tried this, but I still get `Unresolved reference: ____` 
@Azzurite 
@SomeTroglodyte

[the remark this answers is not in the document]
alexban011: Ahh thanks. I feel stupid now haha :))

yairm210 (approved):

PR #7241, opened 2022-06-20
added gameParameter option to disable start bias

alexban011:
added this civ 5 feature which allows disabling the start bias, but the setting doesn't remain turned on after creating a game. 

Not sure why and if this is a problem only for me though. Pretty sure it should keep the setting for the next game as well right?

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -490,6 +490,6 @@ object GameStarter {
                 }
             }
         }
-        return preferredTiles.lastOrNull() ?: freeTiles.random()
+        return preferredTiles.randomOrNull() ?: freeTiles.random()
     }
 }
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -457,6 +459,9 @@ object GameStarter {
         freeTiles: MutableList<TileInfo>,
         startScores: HashMap<TileInfo, Float>
     ): TileInfo {
+        if (gameSetupInfo.gameParameters.noStartBias) {
+            return freeTiles.random()
```
alexban011, line 463: and this makes me wonder.. shouldn't the last return also be .random() instead of .last() to choose a random tile instead of always spawning you on the last tile in the list?

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
```diff
@@ -331,6 +344,12 @@ class MapRegions (val ruleset: Ruleset){
 
         // Next do negative bias ones (ie "Avoid []")
         for (civ in negativeBiasCivs) {
+            // If noStartBias is enabled consider these to be randomCivs
+            if (gameParameters.noStartBias) {
+                randomCivs.addAll(coastBiasCivs)
```
alexban011, line 349: wait wait don't merge! I just noticed I messed up here. Should have been addAll(negativeBiasCivs)

alexban011:
Forgot to change a thing. Now it should be good :)

PR #7242, opened 2022-06-21
Fix translation issues caused by nested brackets and braces

### LINKED ISSUE #7236: Hydro Plant consuming not translates
version 4.1.12-patch1
hydroplant

### LINKED ISSUE #7246: Resources needed in the city screen buildings popup are not translated
image

OptimizedForDensity:
Fixes #7246 and the latest issues in #6131. The issue stems from the fact that braces and brackets within a string can be `"{nested [{inside} {each}] other} {like [this]}".tr()`. Prior to 46b1516159e62840b53abcbe76e3176d0e84e2be, curly braces were processed first then afterwards square brackets first but neither solution is sufficient for the nesting problem. This PR changes it so that the enclosing brackets and braces are processed left to right, `tr()`anslating the contents recursively.

Other changes include a redundant string being removed. "Consumes 1 [resource]" and "Consumes [numberGreaterThan1] [resource]" were coded separately so I merged them into a single string. Testing it reveals no issues that I can tell.

The spaceship victory string was also tweaked.

Images below of previous troublemaking spots (image 3 shows the "Consumes [amount] [resource]" string working for 1 or >1 resources). I checked and all of the open boxes in #6131 ought to be addressed.

<Details>

</Details>

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1175,7 +1175,7 @@ Number of your cities\ndemanding this resource for\n'We Love The King Day' =
 
 [victoryType] Victory = 
 Built [building] = 
-Add all [param] in capital = 
+Add all [comment] in capital = 
 Destroy all players = 
 Capture all capitals = 
 Complete [amount] Policy branches = 
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -86,7 +86,6 @@ Requires a [buildingName] in all cities =
 [buildingName] required: = 
 Requires a [buildingName] in this city = 
 Cannot be built with [buildingName] = 
-Consumes 1 [resource] = 
 Consumes [amount] [resource] = 
 [amount] available = 
 Required tech: [requiredTech] = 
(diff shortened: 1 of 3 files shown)


android/assets/jsons/translations/template.properties
```diff
@@ -1176,7 +1175,7 @@ Number of your cities\ndemanding this resource for\n'We Love The King Day' =
 
 [victoryType] Victory = 
 Built [building] = 
-Add all spaceship parts in capital = 
+Add all [param] in capital =
```
yairm210, line 1178: You sure about this? It doesn't look like you changed any usages of this, and also "Add all [Spaceship part] in capital" is bad English.

OptimizedForDensity: My intention was to make this string the same as the one in `VictoryTypes.json`

https://github.com/yairm210/Unciv/blob/bc9a42e452618552377c0e63a2b828b2b9cf5e7c/android/assets/jsons/Civ%20V%20-%20Gods%20%26%20Kings/VictoryTypes.json#L5-L7

but looking at it again, it seems like this victory is semi-hardcoded (as in changing whatever is inside of the [] would not matter) so the more correct approach would probably be editing `VictoryTypes.json`.

Issue #7249, opened 2022-06-21
AI rarely choose honor


heipizhu4:
**Version**
4.1.12

**Describe the bug**
AI rarely choose honor as their policy. Not sure I should put it in the bug label or the `missing feature from Civ V`.

**To Reproduce**
1. Start a game as spectator
2. Choose relatively more civlizations than usual (so you can test more civs in one time)
3. Pass a few turns
4. Click on the civs one by one and check out their policy
5. You will find nearly all of them choose trandion and liberty.

<details><summary>Saved game</summary>

```
[...]

PR #7252, opened 2022-06-21
Minor reweight of AI policy selection

OptimizedForDensity:
#7249 made me curious to see how the AI picks policies to see why they didn't pick honor. Each nation has a preferred victory type (which defaults to `neutralVictoryType` if their preferred victory is disabled in the game settings) and each policy branch has a priority assigned to it that is associated with each victory type. When picking a policy to unlock, the AI's prioritization is

* pick highest priority policy branch available, choosing randomly if there is a tie and none are in-progress
* if multiple branches tie for highest priority and one or more is partially complete, pick the one with the highest progress, choosing randomly if two or more are tied for progress

The AI weights for vanilla and G&K rulesets are shown below. The AI weights for Tradition and Liberty were either 30 or 0 and the AI weights for Honor were either 10 or 0 depending on victory type, meaning that the AI would always prioritize Tradition and Liberty over Honor. An AI with a neutral preferred victory type would also always choose finishing Tradition and Liberty before considering another policy branch (except for the ideological ones).

This PR cuts down on the weights of Tradition and Liberty so that some AI may consider picking Honor as their first policy branch. Additionally, the weights for an AI with a neutral preferred victory type have been leveled so they will not have a preference towards any specific policies (other than ideological ones but it will pick randomly among those 3 when it reaches the industrial era). Some of the vanilla values have also been changed in case anyone plays with that ruleset.

weights

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Policies.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Policies.json
@@ -3,11 +3,11 @@
         "name": "Tradition",
         "era": "Ancient era",
         "priorities": {
-            "Neutral": 30,
-            "Cultural": 30,
+            "Neutral": 0,
+            "Cultural": 10,
             "Diplomatic": 0,
             "Domination": 0,
-            "Scientific": 30
+            "Scientific": 10
         },
         "uniques": [
             "[+3 Culture] [in capital]",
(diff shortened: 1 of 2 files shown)


OptimizedForDensity:
Forgot I had this PR open :D

I agree, the non-random selection is boring but I may come back to that in a later PR. The science victory weights were tweaked. I agree with dropping honor and boosting rationalism since those two are obvious for science victories. I picked
Tradition: 10
Liberty: 10
Honor: 0
Piety: 0
Patronage: 10
Commerce: 10
Rationalism: 60
Freedom: 50
Autocracy: 40
Order: 50

PR #7255, opened 2022-06-21
Save compatibility handling

### LINKED ISSUE #7248: Prevent people from joining games with old versions
I think we should add the game version used to create the game with in gameInfo and gameInfoPreview and prevent people from joining games using an older version

And then display a simple message like "You need to update the game to x.x.x in order to play in this game"

There isn't any reason you'd want to use an older version

Azzurite:
Resolves #7248

This adds save compatibility handling to Unciv.

I added a new property to `GameInfo`, the `SerializationVersion`. It contains an incrementing `GameInfo` serialization version, as well as the game version the save was created with.

The `GameInfo` serialization version is supposed to only be incremented when actual compatibility issues would occur, i.e. if a change makes it so that a save that is created by the new version can not be loaded at all by a previous version of the game, or it would definitely cause unexpected errors during gameplay. That means that where before the game crashed/failed loading with generic exception message, now there is an incompatible save file message, and where previously crashes happened while playing, now the game isn't loaded and shows the incompatible save file message.

[Here are 3 save files to test with](https://github.com/yairm210/Unciv/files/8952440/SaveFiles.zip). There is `Save_fine`, which was created with a newer version of the game, but the `GameInfo.COMPATIBILITY_VERSION_NUMBER` was not changed, so the save is fine to be loaded. Then there is `Save_compatible`, which is actually loadable by the game, but the `GameInfo.COMPATIBILITY_VERSION_NUMBER` was incremented, thus showing an error while loading. The third save is `Save_incompatible`, which actually has unloadable `GameInfo` data (I changed the `playerType` of the player to `Alien`), which of course also shows an error while loading.

---

There have been a few structural changes made to support this change.

1. `UncivGameParameters` had its `version` parameter removed - the current game version is now found at `UncivGame.VERSION`. This field in the `companion object` of `UncivGame` is updated by our release mechanism, with the `.github/workflows/incrementVersionAndChangelog.js` script.
2. All classes that are serialized as part of `GameInfo` serialization are now implementing the marker interface `IsPartOfGameInfoSerialization` The only purpose for this interface is as a marker for developers to take extra care when changing these classes, as well as provide some documentation. I hope I got all these classes. **If you notice any classes part of serialization but not implementing this interface, please tell me! Or just add it.**
3. All load game exceptions now go through `LoadGameScreen.getLoadExceptionMessage` to display them to the user.


touhidurrr:
.

touhidurrr:
Breaking changes doesn’t always break all games but some.
This just displays an error message in my opinion. Nothing more or less.
As mentioned above this error message doesn’t fix the problem of players not being able to download the latest game from Google Play.

So, don't think it actually does anything. All pain for no gain.

alexban011:
> release the update as an experimental/pre-release

This was created due to games crashing because people joined using an older version on multiplayer and they had the old (unmoddable) game speeds while others had the new ones. In this case it really doesn't matter if you release an experimental version.

alexban011:
> This just displays an error message in my opinion. Nothing more or less.

no. This prevents the save from being broken, prevents people from breaking multiplayer games and additionally it displays a message :D

.github/workflows/incrementVersionAndChangelog.js
```diff
@@ -1,100 +1,127 @@
...
+        per_page: 50
+    });
 
     var commitSummary = "";
-    var ownerToCommits = {}
-    var reachedPreviousVersion = false
-    var nextVersionString = ""
+    var ownerToCommits = {};
+    var reachedPreviousVersion = false;
+    var nextVersionString = "";
```
alexban011, line 31: just asking to gain knowledge.. are these necessary and if so why ?

yairm210:
The popup explaining that the game was generated by a newer version is great, but I'm not sure what happens if it can be loaded.
Sometimes we change only certain parts of the game that may not be present in everyone's savefile, I think if it's loadable then we shouldn't care that it was generated by a more advanced serialization version.

alexban011:
> I think if it's loadable then we shouldn't care that it was generated by a more advanced serialization version.

I don't think so. #7267  is an instance in which the game loaded successfully but still crashed 19 turns later

yairm210:
My proposed solution is simple - try to parse, if you managed then great, only if not - then throw an exception.
Adding this compatibility block effectively means that we'll need to start being really pedantic about constant "one version back" compatibility like @touhidurrr said, which may be what we need anyway.

I'm okay with this solution, but it's like adding a roadblock - it makes things safer at the cost of speed.
At the point we've reached in the project though, that sounds like what we need anyway.

yairm210:
What I'm *not* okay with at the moment is assuming that any change to the serialization class is a breaking change.
Most changes are not a problem - only in rare cases do we change existing items, and it really requires human intervention to see what is and isn't blocking.

Also the gazzilion file changes I don't like at all.


Were this only the "game format version and exception handling" I would approve immediately, but there is too much extra stuff here that I feel is detrimental to the development process.

alexban011:
> try to parse, if you managed then great, only if not - then throw an exception.

This isn't always the case. #7267 is an instance where the game loaded and only failed after 19 turns. So just because it loaded doesn't mean that there won't be problems later

alexban011:
This implementation is light-years ahead of what I suggested in #7248. Like Azzurite said, the message only appears when we "tell" it to, aka when we manually increase the serialization version. Without this, every major change will just make people report issues which are easily fixed by simply updating the game. I think telling people in-game whats wrong is much better (from a UX perspective) than letting them report the issue because the game didn't tell them whats wrong.

Adding something like teams is simply impossible without making sure everyone actually has the feature in their game or without having issues. If a new player joins Unciv together with a friend and they have different versions, if such an issue happens they'd think its a bug. And technically speaking it _is_ a bug. 

To be honest I'm quite surprised we're at <strong>4.1.14</strong> and this feature wasn't needed before. (I'm sure it was needed but maybe the player base wasn't big enough for these issues to be noticed?)

alexban011:
> not everything gets done and stuff stays "needed"

yeah this is actually a pretty good explanation :))

> It's just nicer if you get told why.

well a brand new player doesn't see this as just being nicer. It might simply be the reason he uninstalls the game. But of course, this is an open-source game, maybe it doesn't really matter. It's not like we're being paid to do this, but I still believe UX is very important and this PR definitely improves this. (but actually, most people probably try the game out by themselves which means they won't run into this kind of issues)

yairm210 (approved):
Since all my points have been addressed, nothing more to say :)
Sorry for the confusion

PR #7271, opened 2022-06-22
Fix city ambient sounds not working on Desktop

Azzurite:
This time for realsies. On Desktop, the `Internal` file type is not able to list directories, it always returns an empty array. 

So we now search for the mod files the normal way, and if none are found, then we use the direct internal path to the sound file, which actually works on all platforms.


alexban011:
works

tested on desktop and android. vanilla and modded

yairm210 (approved):

PR #7277, opened 2022-06-23
fixing translatable

alexban011:
fixes "Speeds" and the start year parts of #6131 (start year part requires #7242 to be merged)

the rest should already be handled by #7242.


OptimizedForDensity:
It looks like the untranslated "Start year: [comment]" line can be fixed by changing line 74 of `Speed.kt` to
`yield(FormattedLine("Start year: [" + ("{[${abs(startYear).toInt()}] " + (if (startYear < 0) "BC" else "AD") + "}]").tr()))`
but I think you'll need #7242 for it to work since it has a nested `[{[number] BC/AD}]`

alexban011:
I added that as well.

core/src/com/unciv/models/ruleset/Speed.kt
```diff
@@ -71,7 +71,7 @@ class Speed : RulesetObject() {
         yield(FormattedLine("Golden age length modifier: [${goldenAgeLengthModifier * 100}]%${Fonts.happiness}"))
         yield(FormattedLine("Adjacent city religious pressure: [$religiousPressureAdjacentCity]${Fonts.faith}"))
         yield(FormattedLine("Peace deal duration: [$peaceDealDuration] turns${Fonts.turn}"))
-        yield(FormattedLine("Start year: " + ("[${abs(startYear).toInt()}] " + (if (startYear < 0) "BC" else "AD")).tr()))
+        yield(FormattedLine("Start year: [" + ("{[${abs(startYear).toInt()}] " + (if (startYear < 0) "BC" else "AD") + "}]").tr()))
```
yairm210, line 74: Until the nesting PR is accepted this won't work, so I'm putting this off for now (we could just resolve the terms with .tr() instead of curly braces though so I'm not sure if this trick is needed)

yairm210: I did see that, but if there's something that I approved but did not merge I like to be specific about why not :)

PR #7282, opened 2022-06-24
Fix crash when melee unit captures civilian then tries to attack it

OptimizedForDensity:
A melee unit can target a civilian then move through the civilian to get to the tile it plans to attack from, meaning that they take the unit but won't stop trying to attack it, which causes a crash. This fix adds a check to prevent this crash from happening.

The save to test this is posted in the Discord bugs channel and is too big to be pasted here.


yairm210 (approved):
That is extremely odd, and probably means that we have a bug in getAttackableTiles, but this is a good resolution to the problem as it occurs now

PR #7293, opened 2022-06-24
Add support for era-specific unit sprites

OptimizedForDensity:
Lets modders create unit sprites for each era if they want. These can be combined with civ styles to create regional, era varying sprites.

Syntax is `unit name-style name-era name` or `unit name-era name`, for example an image named `Worker-Medieval era.png` would be the worker sprite for the Medieval era onward unless there is a later sprite for the worker. The effect is demonstrated below where the warrior and worker skins change from the default skin in Ancient era to my beautiful custom sprites once the Classical era is reached.

https://user-images.githubusercontent.com/105244635/175667470-5b36e57f-cbd6-4e3c-87d6-b7f23e83659c.mp4


core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -767,6 +768,20 @@ open class TileGroup(
         }
     }
 
+    private fun ImageAttempter<MapUnit>.tryGetUnitEraSprite(unit: MapUnit, specificUnitIconLocation: String): ImageAttempter<MapUnit> {
+        return this.tryImages(
+            // iterate in reverse order to get the most recent era-specific image
+            (unit.civInfo.getEraNumber() downTo 0).asSequence().map {
+                @Suppress("unused")
+                fun MapUnit.() = if (civInfo.nation.style.isNotEmpty())
```
yairm210, line 776: U wot m8
Why do you need to create a fun for this?
Also why are eras by number not by name?

OptimizedForDensity: > Also why are eras by number not by name?
The logic behind that is so that modders don't have to make so many copies of sprites. If you want the Worker sprite to switch to a guy with a hard hat and a jackhammer starting in the Modern era, you would only have to make a single extra file `Worker-era5.png`. For eras 6, 7, and 8 (Atomic through Future era), they would iterate through the era numbers backwards to find the most recent era where the sprite changed. Going by era name means that the modder would have to copy the identical image for these eras.

yairm210: I understand the "not every era" approach but that would work equally well using the era name, nit the number.
Current era sprites are called by name, and I think the advantage of immediate readability when looking at the sprites (compared to cross referencing with the eras json) vastly outweighs the need to rename sprites of you change an era name.

Because a rename usually indicates that it's supposed to be a different eta, so who knows if it's still relevant?
And what if my mod starts from modern era, I'll need to renumber everything?
I'm not willing to go with numbers.

yairm210: Your identifiers could take precedence over the era name, it alternatively be secondary to it (I think the latter).
But both are good, since they have indicative naming that does not change according to other eras that may exist.

yairm210: https://github.com/yairm210/Unciv/tree/master/android/Images/TileSets/HexaRealm/Tiles

Existing "object-era" naming scheme for tiles

OptimizedForDensity: Alright, I was able to change the loop to use era names while preserving the "use latest available era image" mechanism.

PR #7296, opened 2022-06-24
Make mounted vs city penalties only for attacking

OptimizedForDensity:
According to the Civ 5 wiki, all of the cavalry with a strength penalty against cities should only have the penalty applied when attacking. They should be tankier when attacking as a result.


OptimizedForDensity:
I went ahead and reordered them for better English and to make them consistent with siege weapons which are also `<vs cities> <when attacking>`.

PR #7302, opened 2022-06-25
Arabic.properties

Salman-Aljuhani:


yairm210:
Unfortunately the underlying platform (libgdx) does not support RTL, and especially not connected glyphs like in Arabic, so this cannot enter the game :(

PR #7306, opened 2022-06-25
Fix OutOfMemoryError when loading a game and another is already loaded

Azzurite:
There are two things I found again that essentially hold references to `WorldScreen` while we load one, requiring two times `WorldScreen` in memory.

The first one:

When the status button animations were happening, if you loaded a game right at that moment, the threads responsible for those things held a reference to its object instance, which held a reference to their `Stage` (`Stage.dispose()` only nulls the stage property of immediate children of `root`, not all the children's children), which still contained references to all the tiles and their data, which is what is actually requiring the most memory.

In other words, all the large objects from `WorldScreen` were still held in memory and the GC could not collect them. 

Solved by `cancel`ing the job.

The second one was more fun and is actually imo a bug in libGDX. And solving this actually solves the above issue as well and all future issues with keeping `WorldScreen` references around.

So the first issue is in `MultiplayerHelpers.loadMultiplayerGame`:

https://github.com/yairm210/Unciv/blob/bc9a42e452618552377c0e63a2b828b2b9cf5e7c/core/src/com/unciv/ui/multiplayer/MultiplayerHelpers.kt#L28-L43

The `loadingPopup` is keeping a `WorldScreen.stage` reference throughout the load of the new `WorldScreen`. The theoretical solution there would be to use a new `Popup`. But then I noticed a strange thing when analyzing the heap dump...

`Stage.dispose` seems to be supposed to clear all references to all children... but it doesn't clear the references in `private final Actor[] pointerOverActors = new Actor[20];` and `private @Null Actor mouseOverActor;`. Those were what actually was holding the reference to all the tiles, same as above in the first issue. But when those are properly `null`ed, then we can hold a reference to `WorldScreen` or its `Stage` anywhere we want... the actual children, i.e. all tiles, will be gone, so memory usage of the `WorldScreen` and its `Stage` is negligible. Of course, holding a reference to a `dispose`d `WorldScreen` doesn't make any sense anyway, but that's another topic.

`pointerOverActors` and `mouseOverActor` are private in `Stage`, but a workaround is to call one last `Stage.act()` in the `dispose()` method, since `Stage.act()` is what actually sets these mouse-over fields, and since all children are already removed, sets them to `null`.

Setting these fields to `null` is actually enough to fix the first issue as well, but I feel like it's still correct to cancel these threads when the `WorldScreen` is `dispose`d, as they are doing unnecessary things. The `loadingPopup` I kept, since it displays the error correctly if downloading the game fails and the current `WorldScreen` is kept.


yairm210 (approved):
Hopefully eventually this will be done by libgdx, but until then we can do it ourselves :)

PR #7309, opened 2022-06-25
Fix single pixel gaps in topbar and tutorial

### LINKED ISSUE #7299: there is a gap on the top bar
**Platform**
Android .redmi 5a.lolipop

**Version**
4.1.13 patch 2

**Describe the bug**
At the top of the blue top bar there is a gap whereas in the previous version there was no 

**To Reproduce**

```

```
</details>

**Expected behavior**
I hope in the next version there are no annoying little bugs like this

**Screenshots**

**Additional context**

Skekdog:
Resolves #7299 and also fixes the gap between tutorials and topbar

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -372,7 +372,7 @@ class WorldScreen(
                     .apply { setAlignment(Align.center) }).pad(10f)
             tutorialTaskTable.pack()
             tutorialTaskTable.centerX(stage)
-            tutorialTaskTable.y = topBar.y - tutorialTaskTable.height
+            tutorialTaskTable.y = topBar.y - tutorialTaskTable.height + 1
         }
 
         if (fogOfWar) minimapWrapper.update(selectedCiv)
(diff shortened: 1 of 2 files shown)


OptimizedForDensity:
I tested it and it seems like it works, but I'm curious as to what causes it. If you start a game as a spectator, there isn't a gap but if you click on a civ's units or cities then a gap forms.

alexban011:
> but I'm curious as to what causes it

Gdx is what's causing it 
see <a href="https://github.com/yairm210/Unciv/pull/7027#issuecomment-1157894229">this comment</a>

Skekdog:
> I tested it for the world screen top bar and it seems like it works, but I'm curious as to what causes it. If you start a game as a spectator, there isn't a gap but if you click on a civ's units or cities then a gap forms.

As a spectator I can't seem to have a gap form ever - tested with tutorials and it is fine aswell

Skekdog:
> You're moving topbar, which in Gdx coordinates is already flush with the stage's upper bounds, up by 1 virtual unit, and the tutorialTaskTable up by 2. The relation between these units and actual pixels can vary wildly and is not necessarily near 1:1.

So they aren't just pixels? Oh dear

yairm210:
No one here should be dealing with pixels anyway.
All UI elements talk in coords, and movement should be done in coords as well.

OptimizedForDensity:
No changes have been made in a while and while we're waiting for libGDX to get a fix, I'm going to mark this as a draft PR.

alexban011:
I don't really have a vote here but if it were after me I'd actually close this as won'tfix since its an upstream issue :D

PR #7314, opened 2022-06-26
Fix crash when exploring

Skekdog:
Fix crash reported by @Aditya von Karma#8015 on discord, relating to ruins and barb camps

### DIFF
--- a/core/src/com/unciv/logic/automation/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/UnitAutomation.kt
@@ -48,16 +48,16 @@ object UnitAutomation {
 
     private fun tryGoToRuinAndEncampment(unit: MapUnit): Boolean {
         if (!unit.civInfo.isMajorCiv()) return false // barbs don't have anything to do in ruins
-        val unitDistanceToTiles = unit.movement.getDistanceToTiles()
-        val tileWithRuinOrEncampment = unitDistanceToTiles.keys
+
+        val tileWithRuinOrEncampment = unit.viewableTiles
             .firstOrNull {
                 (
-                    (it.improvement != null && it.getTileImprovement()!!.isAncientRuinsEquivalent())
-                    || it.improvement == Constants.barbarianEncampment
-                )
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -57,6 +57,9 @@ object UnitAutomation {
                         )
                         && unit.movement.canMoveTo(it)
             } ?: return false
+        val firstPath = unit.movement.getShortestPath(tileWithRuinOrEncampment)
+        if (firstPath.isNotEmpty())
+            unit.movement.headTowards(firstPath[0])
         unit.movement.headTowards(tileWithRuinOrEncampment)
```
yairm210, line 63: Did you mean to delete this line?

core/src/com/unciv/logic/automation/UnitAutomation.kt
```diff
@@ -57,6 +57,9 @@ object UnitAutomation {
                         )
                         && unit.movement.canMoveTo(it)
```
yairm210, line 58: I think the correct solution would be to add a `&& unit.movement.canReach(it)` here

That way you solve the 'double path' you have below, and as a bonus if the first viewable and movable ruin happens to not be reachable you can directly head towards the second one:
- First

Skekdog: Done :+1:

yairm210:
I think you misunderstood me, you need both of them :)
Canmoveto first and canreach second

Skekdog:
> I think you misunderstood me, you need both of them :)

It seemed to work fine, but alright

Skekdog:
And do you mean to have both those checks on the same line?

Skekdog:
This is correct now right?

yairm210 (approved):

yairm210:
Yup, canReach is true even if you can approach the time but cannot enter it for whatever reason

PR #7326, opened 2022-06-27
Moddable prettier Tutorials - Step 1 

SomeTroglodyte:
### Reboot of #7064
- TFW support added.
- Officially separating tutorials for Vanilla and G&K would simplify the TFW code even more, not this PR.
- Translation files not included, a TFW run tested to be neutral except for:
- New line "Tutorial = " from UniqueType is unused but not worth the effort to avoid its TFW generation
- Unrelated but now included because it annoyed me too much while testing: The "look at the options" task seemingly didn't see that you obeyed - actually just selecting another unit would show the next one - but you _could_ go options, return, still see the imperative to go options, cycle ad infinitum. There's still a regression here - tutorial tasks do not appear after a resume until some action forces shouldUpdate - not this PR..
- The "\r\n\r\n" code was dead - and fixing irrelevant, that `Regex("\\n{4,}")` code further down should be enough. (Dead because `StringBuilder().run { appendLine(); toList() } == listOf('\n')` is true _even on Windoze_).

## Questions
- _Do_ we want to separate tutorials for Vanilla and G&K? More work when adding or modifying, but could look cleaner - no mention of Religion in Pedia for Vanilla for example. The latter, however, should now be possible to achieve with the corresponding unique alone.. But, the handling for mods would also be simpler? No - mods should already be able to add Pedia-only Tutorials with this, but I haven't tested.
- I think integrating the Tutorial "Tasks" into the system should be done - I would probably just "store" their text in json then create a differentiated TutorialTrigger type for them?
- The actual conditions to trigger "triggered Tutorials" / "Tutorial tasks" could exist e.g. as delegate on the enum or as unique with conditional instead of the enum (e.g. `TutorialTrigger.Unhappiness` -> `UniqueType.TriggersTutorial("Triggers tutorial", UniqueTarget.Tutorial, flags = UniqueFlag.setOfHiddenToUsers)` -> `"Triggers tutorial <when below [0] Happiness>"`)? Opinions?
- In the past I could launch TFW from main menu with two key strokes - thanks to two "regressions" (actually one conscious redesign and one intentional removal) there's now no way to do it with keyboard alone. Am I the only one missing this?

### DIFF
--- a/android/assets/jsons/Tutorials.json
+++ b/android/assets/jsons/Tutorials.json
@@ -1,140 +1,243 @@
-{
-  // Each entry s a tutorial, but the tutorial may be spread over separate paragraphs.
-  // Entries starting with a _ will NOT be shown in Civilopedia.
-  //
+[
+  // Each entry is a tutorial, but the tutorial may be spread over separate paragraphs.
 
-  Introduction: [
+  {
+    "name": "Introduction",
+    "steps": [
       "Welcome to Unciv!\nBecause this is a complex game, there are basic tasks to help familiarize you with the game.\nThese are completely optional, and you're welcome to explore the game on your own!"
-  ],
+    ]
+  },
 
   // Civilopedia only, because players said this was too wall-of-text
-  New_Game: [
+  {
+    "name": "New Game",
+    "steps": [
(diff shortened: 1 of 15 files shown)


yairm210 (approved):
This finally brings tutorials in line with the rest of the json files, as "list of objects", instead of "dictionary". It's worth it if only for that change.

PR #7329, opened 2022-06-29
colorful notifications

alexban011:
Videos are out-of-date

https://user-images.githubusercontent.com/21990802/176780246-647a40f8-6c82-459b-98da-d099ef965871.mp4

image

image


alexban011:
> enum class NotificationColor

that would actually be much cleaner (and maybe even moddable at some point in the future?). I'll convert this to draft and will do that tomorrow. :)

core/src/com/unciv/logic/civilization/Notification.kt
```diff
@@ -36,6 +37,19 @@ object NotificationIcon {
...
+    Default(Color(1f,1f,1f,1f)),
+    Bombard(Color(0.992f, 0.847f, 0.694f, 1f)),
+    Bad(Color(0.996f, 0.752f, 0.698f, 1f)),
+    LessBad(Color(1f, 0.850f, 0.819f, 1f)),
+    Good(Color(0.745f, 1f, 0.698f, 1f)),
+    LessGood(Color(0.843f, 1f, 0.811f, 1f)),
+    WLTK(Color(0.956f, 0.988f, 0.678f, 1f)),
+    Gold(Color(0.976f, 1f, 0.784f, 1f)),
+    Faith(Color(0.839f, 1f, 0.988f, 1f)),
+    Quest(Color(0.733f, 0.945f, 1f, 1f))
```
alexban011, line 50: image

yairm210:
I think we should Discord-poll whether this is better or worse, I'm not sure I like this visually.
Also, if we do this, then 'has been built' notifications should be light brown, as per Production color

alexban011:
>  I'm not sure I like this visually.

I'm pretty sure I can make these colors toggleable in settings for people who don't want them

It gives players a clear idea of what happened and they can more easily tell what happened in a turn without having to read every single notification as most of them aren't that important (_city_ has expanded its border, or _city_ has grown!)

core/src/com/unciv/logic/civilization/Notification.kt
```diff
@@ -36,6 +37,19 @@ object NotificationIcon {
     const val Barbarians = "ImprovementIcons/Barbarian encampment"
 }
 
+enum class NotificationColor(val color: Color) {
+    Default(Color(1f,1f,1f,1f)),
+    Bombard(Color(0.992f, 0.847f, 0.694f, 1f)),
+    Bad(Color(0.996f, 0.752f, 0.698f, 1f)),
+    LessBad(Color(1f, 0.850f, 0.819f, 1f)),
+    Good(Color(0.745f, 1f, 0.698f, 1f)),
+    LessGood(Color(0.843f, 1f, 0.811f, 1f)),
```
OptimizedForDensity, line 46: I feel like these should be reworded to
`VeryBad`
`Bad`
`VeryGood`
`Good`
To me, very good vs good is a little more clear than good vs less good.

alexban011: I will. Didn't really know what to name these so I just went for less___ :)

alexban011:
> I think military related should be different color.

Well if its good its green, if its bad its red. The notification scroll should have as few colors as possible to not turn it into a rainbow mess

alexban011:
> Discord-poll 

@yairm210 could you do it? I can't seem to add pictures or videos in that channel and this poll probably requires showing something like this

notifColors

AdityaMH:
bombard color should red because part of enemy come

OptimizedForDensity:
Maybe a happy medium could be creating the colored notifications framework then letting users select what color certain notifications are? Maybe you want to see at a glance which notifications are brown (which you designated as the "Construction" color and use it for construction started/aborted/finished notifications). This would probably involve an enum for each type of notification then an options popup with a color picker (16 colors preset colors?) that lets the user pair the notification type enum with a color. It could also work well with a separate feature where users can disable certain notifications altogether if they want (just an idea to reuse the same options page/popup).

AdityaMH:
We just need red, brown, crème only. no rainbow.
Red for military thing, brown for building, crème for trade, gold stuff, wltk. and other is white.

AdityaMH:
Or add switch setting to set off (turn off colorful notification), simple (just red, brown, crème, or white), detailed (rainbow) *not recommend yet.

alexban011:
So here's how I colored these:

<pre>
VeryGood:
    Our ___ intercepted and destroyed ___

Good:
    CS can gift great people
    CS assigned you a new quest
    Accepted trade request (not sure if this is important really)
    New ally
    ____ and ___ have signed Peace Treaty
    We have looted ___ from a ___
    Agreed to stop settling near us
    ____ and ___ have signed delcaration of Friendship (other civs also get the "Good" color)
    One time free:
        unit(s),
        unitRuins,
        policy(s),
        greatPerson,
        tech(s)+techRuins,
        revealEntireMap
    One time gain:
        prophet
        pantheon
        trigger voting

VeryBad:
    Our ___ was destroyed by an intercepting __
    Enemy was spotted in/near our territory
    after being hit by our nuke, ___ declared war on us
    A(n) ____ exploded in out territory

Bad:
    Broken pledge to protect
    CS canceled quest because you demanded tribute
    Offered counter offer (it's bad because he didn't acccept what you wanted?)
    Proposed trade no longer relevant
    Trade offer has ended
    _Protector_ is upset that you attacked/demanded...
    _Protector_ is outraged that you destroyed && attacker destroyed ___ you you pledged to protect
    Lost ally
    Cannot provide upkeem-> unit has been disbanded
    Declared war on us (the rest of the civs still get a white notification)
    Denounced us (same as above)
    Relationship with ___ is about to degrade/has degraded
    Refused to stop settling near us
    One of our trades with ___ has been cut short (not sure if this one needs a color though)
    ____ denied trade request
    Barbarians raided and stole gold
    Enemy captured our __
    Your ___ plundered __ from __
    Your citizen are revolting
    Our __ took tile damage
    Enemy has pillaged our ..
    Your territory has been stolen by ___

Construction:
    City changed production
    Unit became obsolete and was removed from queue
    Work started on ___

Plus:
    defendingUnit withdrew from a attackingUnit: Good for defender, Bad for attacker

</pre>

alexban011:
At the moment the notification scroll doesn't updated after changing the setting. Not sure why updating the worldScreen doesn't update the scroll as well

core/src/com/unciv/ui/overviewscreen/NotificationsOverviewTable.kt
```diff
@@ -64,7 +64,10 @@ class NotificationsOverviewTable(
             val label = WrappableLabel(notification.text, labelWidth, Color.BLACK, 20)
 
             notificationTable.add(label)
-            notificationTable.background = ImageGetter.getRoundedEdgeRectangle()
+            notificationTable.background =  if (UncivGame.Current.settings.coloredNotifications)
+                                                ImageGetter.getRoundedEdgeRectangle(notification.color.color)
+                                            else
+                                                ImageGetter.getRoundedEdgeRectangle()
```
alexban011, line 70: @Azzurite It is a setting. The only problem is it doesn't update the notif scroll until "next turn" is pressed

alexban011:
> And I still think it's pretty hard for people that are not reading the code to understand what the colors actually mean.

Is it? I mean there is red for bad stuff, green for good stuff and white for meh stuff :))

I like OptimizedForDenisty's idea with customizable colors and with a couple presets it probably shouldn't be hard to do. Or maybe I should just make these moddable?

core/src/com/unciv/logic/civilization/Notification.kt
```diff
@@ -47,11 +57,18 @@ open class Notification() : IsPartOfGameInfoSerialization {
 
     var icons: ArrayList<String> = ArrayList() // Must be ArrayList and not List so it can be deserialized
     var action: NotificationAction? = null
+    var color: NotificationColor = NotificationColor.Default
```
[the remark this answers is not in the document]
alexban011: > This has the added benefit that you can later change this whenever you want.

I haven't tested this yet as I can't find the time at the moment, but if I do that I think the colors also wouldn't be stored in the save file and this PR wouldn't increase the save file anymore, right?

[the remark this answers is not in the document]
alexban011: I started redoing this, but kinda ran into a problem. Notifications have variables in the names and I don't really know how to store the text without a reference to the variables..

yairm210:
Superceded by 7edfd497

alexban011:
ahh yes thanks.. I tried doing what Azzurite suggested but it turned out to be out of my league since one of the notifications (can't remember which) could be both bad and good depending on the situation.. and defining a color for that was impossible

BUT your implementation is much cleaner.. I didn't thought about simply dividing the notifications like that

alexban011:
but the text is a bit small on the "Huge" screen size

image

AdityaMH:
Rather than this, big round shape with icon and color. and text only appear when press hold to screen (android) or, icon hovered from mouse pointer. more better.
But since new sortable notification added, color thing will be useless.

PR #7333, opened 2022-06-29
fix wrong unique check

MindaugasRumsa51:


yairm210:
I'm pretty sure your wrong about this, we what to be able to give specific units embarkation powers

MindaugasRumsa51:
sorry then

PR #7342, opened 2022-06-30
Generalize production-to-stat conversion uniques

OptimizedForDensity:
Changes a few uniques to be more general
* `UniqueType.EnablesGoldProduction` and `UniqueType.EnablesScienceProduction` are depreciated in favor of `UniqueType.EnablesCivWideStatProduction("Enables conversion of city production to [civWideStat]")` which adds the ability for modders to allow players to convert production to faith or culture in addition to gold and science
* `UniqueType.ProductionToScienceConversionBonus` depreciated in favor of a more flexible `ProductionToCivWideStatConversionBonus("Production to [civWideStat] conversion in cities changed by [relativeAmount]%")`
* `plunderableStat` renamed to `civWideStat` for reuse in the above uniques

Behind the scenes, this also adds a `PerpetualStatConversion` class which inherits `PerpetualConstruction` to reuse code between the production-to-stat conversion objects although I might fold this back into `PerpetualConstruction` since I'm on the fence whether it improves readability.

Stat conversion


yairm210 (approved):

core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
```diff
@@ -444,6 +444,8 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
             is BaseUnit -> UncivSound.Promote
             PerpetualConstruction.gold -> UncivSound.Coin
             PerpetualConstruction.science -> UncivSound.Paper
+            PerpetualConstruction.culture -> UncivSound.Chimes
+            PerpetualConstruction.faith -> UncivSound.Choir
```
yairm210, line 448: I think this is kind of long and reserved for finishing wonders, settling cities etc

OptimizedForDensity: I changed the `UncivSound.Chimes` to `UncivSound.Policy` which is a shorter sound but is still related to culture.

PR #7343, opened 2022-06-30
Unit icon alpha/size control

### LINKED ISSUE #6838: Feature request: Add unit icon size and transparency sliders
**Problem:**

Unit icons take up a large percentage of the tile. This can cause visual clutter and block detail of the unit sprites.

Example2

**Proposed solution:**

Add sliders under Options > Display to allow the user to set transparency (0% to 100%) and size (0% to 100%) of the unit icons.

letstalkaboutdune:
Resolves #6838

Implemented sliders in Options -> Display to give the user control over unit icon opacity and size. Default values are 100% opaque and size of 25, same as the current release.

Demonstration video: https://j.gifs.com/lRmXl1.gif

**Question:** this conflicts with a previous function that makes part of the unit icon 50% transparent when idle. Any ideas on how to implement that differently?


core/src/com/unciv/ui/tilegroups/TileGroupIcons.kt
```diff
@@ -99,13 +100,20 @@ class TileGroupIcons(val tileGroup: TileGroup) {
...
+            /*
+            // HOW TO HANDLE WITH NEW ICON OPACITY CONTROL?
             // Instead of fading out the entire unit with its background, we just fade out its central icon,
             // that way it remains much more visible on the map
             if (!unit.isIdle() && unit.civInfo == viewingCiv) {
                 newImage.unitBaseImage.color.a = 0.5f
                 newImage.actionGroup?.color?.a = 0.5f
             }
+             */
+
```
OptimizedForDensity, line 112: Probably the most straightforward approach would be to move the new code above this section then do
`newImage.unitBaseImage.color.a *= 0.5f`
`newImage.actionGroup?.color?.a *= 0.5f`
Users with fully opaque icons would still see the semi-transparent icons once they finish moving a unit, although I don't know if a user with very transparent icons to begin with could notice a difference.

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -45,6 +45,8 @@ class GameSettings {
 
     var showMinimap: Boolean = true
     var minimapSize: Int = 6    // default corresponds to 15% screen space
+    var unitIconOpacity = 1f // default corresponds to fully opaque
+    var unitIconSize = 25f // default corresponds to maximum size
```
OptimizedForDensity, line 49: This might be better as a percentage slider that has a default of 100% (corresponding to `25f`). The average user probably doesn't care about the size in floats but knows that 200% means double the size and 50% halves it.

letstalkaboutdune: Good point. I can set the max/default value of 25f to 100% and the min to 40% (10 out of 25).

core/src/com/unciv/ui/options/DisplayTab.kt
```diff
@@ -94,6 +98,56 @@ private fun addMinimapSizeSlider(table: Table, settings: GameSettings, selectBox
...
+
+    val minTranslated = "minimum".tr()
+    val maxTranslated = "maximum".tr()
+    val getTipText: (Float) -> String = {
+        when (it) {
+            10f -> minTranslated
+            in 11f..24f -> "%.0f".format(it)
+            else -> maxTranslated
+        }
+    }
```
OptimizedForDensity, line 138: Probably don't need the text to pop up since dragging the slider to one side will intuitively set it to its minimum or maximum.

letstalkaboutdune: Yep, I updated it to go from 40% to 100% in 5% increments.

core/src/com/unciv/models/metadata/GameSettings.kt
letstalkaboutdune: Yep, I updated it to go from 40% to 100% in 5% increments.

android/assets/jsons/translations/template.properties
```diff
@@ -697,6 +697,9 @@ Show pixel units =
 Show pixel improvements = 
 Enable Nuclear Weapons = 
 Experimental Demographics scoreboard = 
+Unit icon opacity = 
+Unit icon size = 
+opaque =
```
OptimizedForDensity, line 702: Unused

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -21,6 +21,7 @@ class UnitGroup(val unit: MapUnit, val size: Float): Group() {
         val background = getBackgroundImageForUnit()
         background.apply {
             this.color = unit.civInfo.nation.getOuterColor()
+            this.color.a = UncivGame.Current.settings.unitIconOpacity;
```
OptimizedForDensity, line 24: Stray semi-colon

OptimizedForDensity:
Otherwise, I copied the code and it worked in my testing and seems to be a good QOL feature (although I never personally found the icons to be too big or in the way).

android/assets/jsons/translations/template.properties
letstalkaboutdune: Oops! Fixed.

core/src/com/unciv/ui/utils/UnitGroup.kt
letstalkaboutdune: Double oops! Also fixed.

letstalkaboutdune:
> The only thing I found was that you can resize the icons to be very, very small, almost too small but that may be screen and eyesight dependent.

I can change the minimum to 50% (increased from 40%). Another thing I can try is to make any selected unit icon revert back to 100% opacity and size (while selected only). This would make sure the selected unit always has good visual focus.

letstalkaboutdune:
I removed the size adjustment slider and only implemented the transparency. The small sizes just looked goofy in my opinion. Also the unit icons now become fully opaque when selected to make sure the user knows what they're doing. 

Example: https://youtu.be/ksS2hq-ji14

alexban011:
Just FYI: you can upload the videos right here. Just drag and drop any video/picture in your message

core/src/com/unciv/ui/options/DisplayTab.kt
```diff
@@ -94,6 +96,25 @@ private fun addMinimapSizeSlider(table: Table, settings: GameSettings, selectBox
...
+
+    val unitIconAlphaSlider = UncivSlider(
+        0f, 1f, 0.1f, initial = settings.unitIconOpacity, getTipText = getTipText
+    ) {
+        settings.unitIconOpacity = it
+        settings.save()
+
+        val worldScreen = UncivGame.Current.getWorldScreenIfActive()
+        if (worldScreen != null)
+            worldScreen.shouldUpdate = true
```
alexban011, line 112: this would be equivalent to `worldscreen?.shouldUpdate =  true`. Just FYI, no need to change this now

letstalkaboutdune:
> Just FYI: you can upload the videos right here. Just drag and drop any video/picture in your message

I tried that, but the file was too big. I need to adjust some settings or something, because a 30s clip at my native resolution was >200MB.

yairm210 (approved):
Looks good to me :)

PR #7351, opened 2022-07-01
Stop automate production setting from automating other players' production in MP

OptimizedForDensity:
The automate production setting has strange behavior in multiplayer games where any human player that has the option enabled will cause the next human player's production to be automated even if they have the option disabled. Now that will no longer be the case, thanks to some reordering in `GameInfo.nextTurn()` and some changed conditions in `CityConstructions.chooseNextConstruction()`.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -243,7 +243,7 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
         var currentPlayerIndex = civilizations.indexOf(thisPlayer)
 
 
-        fun switchTurn() {
+        fun endTurn() {
             thisPlayer.endTurn()
             currentPlayerIndex = (currentPlayerIndex + 1) % civilizations.size
             if (currentPlayerIndex == 0) {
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -283,14 +283,15 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion
...
-            switchTurn()
+            endTurn()
         }
         if (turns == UncivGame.Current.simulateUntilTurnForDebug)
             UncivGame.Current.simulateUntilTurnForDebug = 0
 
         currentTurnStartTime = System.currentTimeMillis()
         currentPlayer = thisPlayer.civName
         currentPlayerCiv = getCivilization(currentPlayer)
+        thisPlayer.startTurn()
```
OptimizedForDensity, line 294: Need to move this function down to below the point where the `currentPlayer` switches to the next human player so that `gameInfo.isUsersTurn()` is looking at the next human rather than the human player who just clicked "Next Turn". This prevents the game from automating the production of the next human player if the one who just ended their turn has the automate production setting checked.

PR #7359, opened 2022-07-02
AI targets enemy units in a smarter way

OptimizedForDensity:
Currently, AI units and cities target enemy units according to lowest health. This is an quick heuristic but it does not account for any bonuses between units (for example, an AI spearman facing a 90 health warrior and 100 health horseman will always target the warrior). With this PR, AI units simulate attacking each enemy unit in the same way a human player would using `BattleDamage.calculateDamageToDefender` to estimate the damage done to enemy units (akin to a human selecting a unit then clicking on each enemy unit to look at the outcome in the battle table). This function accounts for unit modifiers which means the AI will try to account for unit type or promotion advantages when picking attack targets. I don't believe there is a universally best way to target units so I figured a heuristic for the AI to use (which I can change if others feel it could be done in a better way)
* each unit should try to kill an enemy unit this turn, prioritizing by maximum attack strength if it can kill at least 1 this turn
* if the unit cannot kill any other unit this turn, then attack the military unit that it could kill in the least number of theoretical attacks
* if there are no military targets, attack a civilian

Both AI unit target selection and city selection of bombard targets has been changed to follow this rule. Additionally, cities have their target prioritization tweaked to favor non-embarked siege units since those pose a more immediate danger than embarked ones.

### DIFF
--- a/core/src/com/unciv/logic/automation/BattleHelper.kt
+++ b/core/src/com/unciv/logic/automation/BattleHelper.kt
@@ -166,11 +166,35 @@ object BattleHelper {
             enemyTileToAttack = capturableCity // enter it quickly, top priority!
 
         else if (nonCityTilesToAttack.isNotEmpty()) // second priority, units
-            enemyTileToAttack = nonCityTilesToAttack.minByOrNull {
-                Battle.getMapCombatantOfTile(it.tileToAttack)!!.getHealth()
-            }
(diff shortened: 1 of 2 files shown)


yairm210 (approved):
Looks good! I do wonder if maybe in getAttackableTiles we should add the damage done, turns to kill etc, to minimize code duplication, but that's a problem for another time

PR #7362, opened 2022-07-03
Update Spanish.properties

ArchDuque-Pancake:
Improve unique understanding, not even I could really understand xDDD

Btw, there are still the # Requires Translation from last update but they are filled...


yairm210:
prepare version

yairm210:
Version rollout PR created

PR #7369, opened 2022-07-04
Improve AI belief picking

OptimizedForDensity:
A collection of tweaks and bugfixes to AI belief selection. Changes are detailed in comments on specific lines of codes but an overview of the fixes:
* Fix bugged implementation of time-dependent weighting for picking several beliefs. Also extended it to work for mods which shuffle enhancer and founder beliefs (enhancing looks at % of religions enhanced, founding looks at % of religions founded)
* Stop AI from evaluating enhancing beliefs every turn
* Implement an AI evaluation branch for the "stats for every X global followers" unique. As far as I can tell, this was the only belief in base Unciv that was not included in the choose belief automation
* Add missing minus signs that caused discounts to be evaluated unfavorably by the AI
* Scale a few weights based on my personal judgement (open to feedback)


core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
```diff
@@ -99,33 +100,53 @@ object ChooseBeliefsAutomation {
...
         }
         val goodLateModifier = when {
-            amountOfEnhancedReligions < 33 -> 2f
-            amountOfEnhancedReligions < 66 -> 1f
+            gameTimeScalingPercent < 33 -> 2f
+            gameTimeScalingPercent < 66 -> 1f
             else -> 1/2f
-        } 
-        
+        }
```
OptimizedForDensity, line 148: Previously, `amountOfEnhancedReligions` just counted the number of foundable religions which breaks the modifier since the modifier assumed integer percentages (0-100). This meant that the AI always believed it was late game (`goodLateModifier = 2f`). However, due to the `goodEarlyModifier`/`goodLateModifier` being divided further in the function, these modifiers actually had the opposite effect than they were intended. This has been fixed below so the modifiers are multiplied.

core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
```diff
@@ -155,31 +176,36 @@ object ChooseBeliefsAutomation {
...
-                    unique.params[0].toInt() / 5f
+                    unique.params[0].toFloat() / 5f
                 UniqueType.StatsWhenAdoptingReligion, UniqueType.StatsWhenAdoptingReligionSpeed ->
                     unique.stats.values.sum() / 50f
                 UniqueType.RestingPointOfCityStatesFollowingReligionChange ->
-                    unique.params[0].toInt() / 7f
+                    if (civInfo.wantsToFocusOn(Victory.Focus.CityStates))
+                        unique.params[0].toFloat() / 3.5f
+                    else
+                        unique.params[0].toFloat() / 7f
```
OptimizedForDensity, line 186: Add some simple AI personality-esque logic.

core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
```diff
@@ -155,31 +176,36 @@ object ChooseBeliefsAutomation {
...
                 UniqueType.StatsFromGlobalCitiesFollowingReligion ->
-                    50f / unique.stats.values.sum()
+                    unique.stats.values.sum()
```
OptimizedForDensity, line 188: Dividing by the sum of stats causes the AI to evaluate beliefs that give more stats as worse so this has been changed to be just a simple sum.

core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
```diff
@@ -155,31 +176,36 @@ object ChooseBeliefsAutomation {
...
+                UniqueType.StatsFromGlobalFollowers ->
+                    4f * (unique.stats.values.sum() / unique.params[1].toFloat())
```
OptimizedForDensity, line 190: Was missing

core/src/com/unciv/logic/automation/ChooseBeliefsAutomation.kt
```diff
@@ -155,31 +176,36 @@ object ChooseBeliefsAutomation {
...
+                    unique.params[0].toFloat() * goodEarlyModifier / 5f
                 UniqueType.SpreadReligionStrength ->
-                    unique.params[0].toInt() / goodLateModifier
+                    unique.params[0].toFloat() * goodLateModifier / 5f
                 UniqueType.FaithCostOfGreatProphetChange ->
-                    unique.params[0].toInt() / goodLateModifier / 2f    
+                    -unique.params[0].toFloat() * goodLateModifier / 2f
                 UniqueType.BuyBuildingsDiscount, UniqueType.BuyItemsDiscount, UniqueType.BuyUnitsDiscount ->
-                    unique.params[2].toInt() / goodLateModifier
+                    -unique.params[2].toFloat() * goodLateModifier / 5f
```
OptimizedForDensity, line 204: Discounts are written as e.g. `[-33]%` so negative sign needed.

OptimizedForDensity, line 204: Added some subjective scaling weights since these boni tend to be larger numerically compared to the stat-providing uniques (e.g., `[100]%` vs `[+1 Gold] from every global follow`)

PR #7375, opened 2022-07-05
Fix unit icon opacity for idle units

letstalkaboutdune:
This PR handles the unitBaseImage and actionGroup layers of unit icons properly if a unit is idle (out of moves or in a state such as fortified or sleeping), leaving them at 50% opacity when selected. My previous PRs set them to 100% opacity regardless of idle status.

https://user-images.githubusercontent.com/56904240/177252699-39dec492-4ea1-43fc-be05-5400bb077ae2.mp4


yairm210 (approved):
The smaller the change, the easier it is to understand what's going on and therefore to approve :)

PR #7377, opened 2022-07-05
increased mod search results_per_page

### LINKED ISSUE #7321: Feature request: increase the amount of results/page per github search request
I recommend increasing this to the max of 100 to get more repos / result, thus running into less rate limits and faster loading of mods

https://github.com/yairm210/Unciv/blob/8f0e64e02df4578b1287090a1f442996f98ebf11/core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt#L68

alexban011:
resolves #7321

When making the mod categories I kept running into rate-limits and the mods load extremely slow because of this. Sorting and searching don't work until all mods have been loaded and as a side effect this makes the search/sort feature feel like it doesn't work all the time. 

The only indication the user gets are 3 dots but those can only be seen if the user scrolls all the way down...

Technically the limit seems to be 1000 but I don't know if that's really necessary, especially now when we only have less than a third of that

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt
@@ -67,7 +67,7 @@ class ModManagementScreen(
     private val modActionTable = Table().apply { defaults().pad(10f) }
     private val optionsManager = ModManagementOptions(this)
 
-    val amountPerPage = 30
+    val amountPerPage = 100
 
     private var lastSelectedButton: Button? = null
     private var lastSyncMarkedButton: Button? = null


yairm210:
We initially limited results to make it return faster - how long does it take to load 30 results vs 100? (http time, not UI time)

alexban011:
> We initially limited results to make it return faster - how long does it take to load 30 results vs 100? (http time, not UI time)

For 30 I got an average of `882.1666` ms
For 100 I got an average of `2555.5` ms

But for 30 searches it loaded the mods normally and then I left the screen and clicked again on mods and was already running into rate limits. 
It is a noticeable difference but I think its worth it

yairm210:
That's a good way to measure
2 seconds is not short, but it's still acceptable

yairm210 (approved):

PR #7393, opened 2022-07-07
Update Persian_(Pinglish-UN).properties

Herbad:


Herbad:
> Closed because of #7395
> 
> Next time, please try to make a new commit to the same branch (patch-3 in this case) instead of creating a new branch and a new pull request.

😐😐😐😐😐WTF! No one didn't explain what and how should I edit that
Now I have to translate all of the again!!?

PR #7402, opened 2022-07-08
Refactorings & reduce network data usage

Azzurite:
The base branch for these is `multiplayer-data-changes`, not `master`. I have a bunch of serialization changes I want to do, and to not break multiplayer games along several versions, since it will probably take a while to get all my changes through, I thought it'd be best to collect the serialization changes in one branch and then when all is done merge this branch into `master`.

---

This PR is just a bunch of refactorings that I wanted to do which either just make the code nicer or that support the intended feature I want to add (which is having a separate multiplayer data format to not pollute singleplayer games with that data, and to support playing on multiple Unciv servers at the same time). And one refactoring which is also a feature, reduced network data usage.

I suggest reviewing commit-by-commit again and I'll add explanations for the different commits here:

* [Refactor: Rename GameInfoPreview to MultiplayerGameStatus](https://github.com/yairm210/Unciv/pull/xxxx/commits/639f58e86eae09cad5f011b0431ce9ca0fc35f2a): I should have done this and the next commit the other way around, but oh well. Currently, the `GameInfoPreview` is used for multiplayer game status *and* for the preview in the "Load game" screen. Since they need a little bit of separate information, it doesn't make much sense to have them in one class. So here I rename it, and in the [next commit](https://github.com/yairm210/Unciv/pull/xxxx/commits/1d08af47bc357086b584cf07a7af971efec37398) I actually split them up.
* [Refactor: Extract GameInfo interfaces](https://github.com/yairm210/Unciv/pull/xxxx/commits/ca7d10080eba86b4050789d5e379db9fe9635653): This makes methods like `hasLatestTurn` work on both `GameInfo` and `MultiplayerGameStatus`, which is nice polymorphism. It'll also be used later.
* [Refactor: Rename GameInfo.currentPlayer & GameInfo.currentPlayerCiv](https://github.com/yairm210/Unciv/pull/xxxx/commits/13aa860add096f659464c8d11b5c698cab65f379): So this naming is weird and tripped me up previously. `GameInfo.currentPlayer` is a String containing the name of the current civ. `GameInfo.currentPlayerCiv` is of type `Civilization`, containing the current civ. The thing is that these do not contain the actual `Player` class which we already have in the project. Also the civ name has been saved in other classes before, and in almost any of these places the variable is named `civName`. To have it consistent, `currentPlayer` makes much more sense named `currentCivName`. `currentPlayerCiv` can also just be `currentCiv`.
* [Add HasGameTurnData.currentPlayerId](https://github.com/yairm210/Unciv/pull/xxxx/commits/2a7acab6a69b1f1835558337825e4a82321b68b1): We don't techinically need the `civilizations` in the `Multiplayer.GameStatus`. The only place that uses it is the turn check worker, which needs it to figure out if it's our turn in the game. However, the turn checker only needs the playerId from the current civilization and not the whole civilzation, so we can just add this `currentPlayerId` property here to cut down on multiplayer game status size in the following commit.
* [Reduce multiplayer game status serialization size](https://github.com/yairm210/Unciv/pull/xxxx/commits/fd0e391a33ca339db056879dedb661be1cc0b0e9): We're transmitting the `Multiplayer.GameStatus` data very often, the user can configure as low as 15 second interval for all games, and 3 second interval for  the current game. If you assume someone who's playing multiplayer has 5 games running at the same time, that would be 2160 requests per hour. Currently, the GZipped size is around 650 bytes, which totals to around 1371KB or 1.3MB per hour. With this commit, the size reduces to ~130 bytes, which totals to around 274KB per hour. So that's an 80% reduction in transmission size :)
* [Refactor: Add exception log](https://github.com/yairm210/Unciv/pull/xxxx/commits/f7ad362364139f9862dc133eaae10a987abc430d): nothing to say
* [Refactor: Remove "Online" prefix from multiplayer classes](https://github.com/yairm210/Unciv/pull/xxxx/commits/4d539960a478623f3390ee0fe10abb6b59ec6427): We do have local hot-seat multiplayer, which is why I named these classes `OnlineMultiplayer` when I added them. But there isn't actually any code for local multiplayer, only online multiplayer, so `OnlineMultiplayer` is just redundant really.
*  [Refactor: Move MultiplayerGameStatus into Multiplayer & make it a data class](https://github.com/yairm210/Unciv/pull/xxxx/commits/98c36d29dfb56390fb702332e5576803b98fa07d): it's such a small class and so closely coupled with the Multiplayer that it doesn't really warrant a separate file.
* [Refactor: Make tab builder functions not pollute the global scope](https://github.com/yairm210/Unciv/pull/xxxx/commits/524df81649271220f5a00998172f7a5a48286966): in the same vein as the previous commit, there's no reason the IDE needs to be able to autocomplete `aboutTab` anywhere except for within the `OptionsPopup`.
* [Refactor: Move server input into separate file](https://github.com/yairm210/Unciv/pull/xxxx/commits/fab761c942b0713dfd571137c025d169e52b5c9e): This is in preparation of the next feature.


OptimizedForDensity:
Just a heads up, all your commit links say "...yairm210/Unciv/pull/xxxx/..." which I think you intended to replace the xxxx with the PR number.

yairm210 (approved):
There's a big difference between renaming and moving stuff around that doesn't have any effect, and moving stuff that does does - any field within serialization, like the current civ name.
Please try to separate the former, which can be almost auto-approved, from the latter, which requires testing and is potentially game-breaking.

yairm210:
I'm worried that the eventual merge into master will be a gazillion files, this is part of the reason I don't like feature branches in general.
All the non-serialized renaming can be immediately merged into the master branch to minimize the final diff.

Issue #7376, opened 2022-07-05
Barbarian joining Germany should have full HP.


ravignir:

ravignir:
https://user-images.githubusercontent.com/30041793/177304066-ff7e8436-2a90-4ba2-a2ed-a2bb64450889.mp4

alexban011:
I mean.. from what I see in the video they don't get full HP when joining

ravignir:
watch the video again, spearman is attacking, full hp brute joins :D

ravignir:
it is also "bugged", barbs join you and destroy encampment even when they are killed with a ranged attack. The ability should only trigger when melee unit enters an encampment - either as a result of killing an enemy unit or simply moving to empty encampment (if encampment was empty, then ability may still trigger, the last unit at encampment will join you)
```
[...]

PR #7403, opened 2022-07-08
Resolved #7316 - unified unit capture unique mechanics for no double-capture

### LINKED ISSUE #7316: Ottomans privateer has a chance to duplicate barbarian water unit
**Platform**
Android 10
**Version**
4.1.13-patch2
**Describe the bug**
Ottomans privateer has a chance to duplicate barbarian water unit
**To Reproduce**
Screenshot_20220626-230315
Screenshot_20220626-230350

```

```
</details>

**Expected behavior**

**Screenshots**

**Additional context**

yairm210:
Disregarding #7376 for a moment, there are currently 3 unit capture uniques for Germany, Ottomans and Prize Ships.
This PR unifies the functionality to ensue no double-capture occurs and consequentially also unifies capture notifications.

Old uniques were fully deprecated.

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -231,23 +231,28 @@ object Battle {
 
         if (!defender.isDefeated() || defender.unit.isCivilian()) return false
 
-        fun unitCapturedPrizeShipsUnique():Boolean {
+        fun unitCapturedPrizeShipsUnique(): Boolean {
             if (attacker.unit.getMatchingUniques(UniqueType.KillUnitCapture)
                         .none { defender.matchesCategory(it.params[0]) }
             ) return false
 
-            val captureChance = min(0.8f, 0.1f + attacker.getAttackingStrength().toFloat() / defender.getDefendingStrength().toFloat() * 0.4f)
+            val captureChance = min(
+                0.8f,
+                0.1f + attacker.getAttackingStrength().toFloat() / defender.getDefendingStrength()
+                    .toFloat() * 0.4f
+            )
             return Random().nextFloat() <= captureChance
         }
 
-        fun unitCapturedGermanyUnique():Boolean{
-            if (!defender.getCivInfo().isBarbarian() || attackedTile.improvement != Constants.barbarianEncampment)
-                return false
+        fun unitGainFromEncampment(): Boolean {
+            if (!defender.getCivInfo().isBarbarian()) return false
+            if (attackedTile.improvement != Constants.barbarianEncampment) return false
 
             var unitCaptured = false
             // German unique - needs to be checked before we try to move to the enemy tile, since the encampment disappears after we move in
 
-            for (unique in attacker.getCivInfo().getMatchingUniques(UniqueType.GainFromEncampment)) {
+            for (unique in attacker.getCivInfo()
+                .getMatchingUniques(UniqueType.GainFromEncampment)) {
                 attacker.getCivInfo().addGold(unique.params[0].toInt())
                 unitCaptured = true
             }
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -226,21 +222,76 @@ object Battle {
...
+                    unitCaptured = true
+                }
+            }
+            return unitCaptured
+        }
 
-        if (attacker.unit.getMatchingUniques(UniqueType.KillUnitCapture)
-            .none { defender.matchesCategory(it.params[0]) }
-        ) return false
+        val wasUnitCaptured = unitCapturedPrizeShipsUnique() || unitCapturedGermanyUnique() || unitCapturedOttomanUnique()
```
OptimizedForDensity, line 271: The prize ships unique merely captures the unit but the German and Ottoman uniques capture the unit and give gold, so an Ottoman privateer will first try to capture using the privateer unique before trying with the Ottoman unique. This could deprive the player of some gold if the privateer unique capture chance has a successful roll since the Ottoman unique won't have a chance to roll, but this is a gameplay choice and not strictly a code issue.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -226,21 +222,76 @@ object Battle {
...
+        fun unitCapturedPrizeShipsUnique():Boolean {
+            if (attacker.unit.getMatchingUniques(UniqueType.KillUnitCapture)
+                        .none { defender.matchesCategory(it.params[0]) }
+            ) return false
+
+            val captureChance = min(0.8f, 0.1f + attacker.getAttackingStrength().toFloat() / defender.getDefendingStrength().toFloat() * 0.4f)
+            return Random().nextFloat() <= captureChance
+        }
+
+        fun unitCapturedGermanyUnique():Boolean{
```
OptimizedForDensity, line 243: I feel that `unitCapturedGermanyUnique()` should be renamed to something `unitGainFromEncampment` and `unitCapturedOttomanUnique()` renamed to something like `unitGainFromDefeatingUnit` since those are a touch more descriptive.

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -226,21 +222,76 @@ object Battle {
...
-        return spawnCapturedUnit(defender.getName(), attacker, defender.getTile(), "Your [${attacker.getName()}] captured an enemy [${defender.getName()}]!")
+        return spawnCapturedUnit(defender.getName(), attacker, defender.getTile())
+    }
+
+    /** Places a [unitName] unit near [tile] after being attacked by [attacker].
+     * Adds a notification to [attacker]'s civInfo and returns whether the captured unit could be placed */
+    private fun spawnCapturedUnit(unitName: String, attacker: ICombatant, tile: TileInfo): Boolean {
+        val addedUnit = attacker.getCivInfo().placeUnitNearTile(tile.position, unitName) ?: return false
+        addedUnit.currentMovement = 0f
+        addedUnit.health = 50
```
OptimizedForDensity, line 286: `addedUnit.health` as a function argument could solve #7376 in the same PR :)

OptimizedForDensity:
Looks pretty good so far! Also reminds me that "When conquering an encampment..." and "When defeating a [combatantFilter] unit..." could possibly be conditionals but that's for another day.

xlenstra:
> Also reminds me that "When conquering an encampment..." and "When defeating a [combatantFilter] unit..." could possibly be conditionals but that's for another day.

Thought about that before, but the problem is that those aren't restrictions on a unique but triggers for when a triggerable unique should be executed, which is different

core/src/com/unciv/logic/battle/Battle.kt
yairm210: Oh, because of the way that or operators resolve... good catch!
I could separate this into 3 lines but that's not as nice, wonder if there's a better way

android/assets/jsons/translations/template.properties
```diff
@@ -790,8 +790,11 @@ The City-State of [name] has been destroyed! =
 Your [ourUnit] captured an enemy [theirUnit]! = 
 Your [ourUnit] plundered [amount] [Stat] from [theirUnit] = 
 We have captured a barbarian encampment and recovered [goldAmount] gold! = 
-A barbarian [unitType] has joined us! = 
-We have captured an enemy [unitType]! = 
+An enemy [unitType] has joined us! = 
+
+# This might be needed for a rewrite of Germany's unique - see #7376
+# A barbarian [unitType] has joined us! =
```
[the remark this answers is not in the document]
yairm210: Indeed, I means to add the comment and not comment out the line

core/src/com/unciv/logic/battle/Battle.kt
yairm210: That needs to be dealt with separately anyway, since apparently it doesn't actually capture a unit but instead activates when capturing an encampment even if there's no unit... so that'll need to be another PR anyway

OptimizedForDensity (approved):

PR #7404, opened 2022-07-08
Fix policy uniques not being applied after adoption

OptimizedForDensity:
Partially undo commit 3422e161ada01731d1be86ebaf970f124fce6cb1 to fix policy unique objects not being added to the civ's unique list.

Also removes a now-unused `companion object` from `PolicyManager.kt`.


MindaugasRumsa51:
i observing code changes since 4.0.15
but have not noticed removal of these police lines you just added again
in which PR they were removed?

xlenstra (approved):

PR #7406, opened 2022-07-08
Update French.properties

Ouaz:


android/assets/jsons/translations/French.properties
```diff
@@ -422,10 +422,10 @@ Click to copy = Cliquer pour copier
...
 
 Game Speed = Vitesse de la partie
-Quick = Rapide
-Standard = Normale
-Epic = Épique
-Marathon = Marathon
+Quick = (1) Rapide
+Standard = (2) Normale
+Epic = (3) Épique
+Marathon = (4) Marathon
```
OptimizedForDensity, line 428: This is probably better done through code rather than relying on translation files to sort them :) I will take a look at it.

Ouaz: I removed it. Thanks for your help and the PR!

PR #7407, opened 2022-07-08
Add more music triggers

OptimizedForDensity:
https://yairm210.github.io/Unciv/Modders/Images-and-Audio/ notes that there are 4 unimplemented music triggers: civ-specific golden age music, tech-specific music played when that tech is researched, wonder-specific music that is played when that wonder is finished, then the civ theme when the civ is selected in the map editor. These triggers have been implemented to trigger when a human player whose turn it currently is does whatever is needed to cause the music track to play.

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -19,6 +19,8 @@ import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.stats.Stat
 import com.unciv.models.stats.Stats
 import com.unciv.models.translations.tr
+import com.unciv.ui.audio.MusicMood
+import com.unciv.ui.audio.MusicTrackChooserFlags
 import com.unciv.ui.civilopedia.CivilopediaCategories
 import com.unciv.ui.civilopedia.FormattedLine
 import com.unciv.ui.utils.Fonts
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/ui/audio/MusicMood.kt
```diff
@@ -7,6 +7,9 @@ object MusicMood {
     const val Defeat = "Defeat"
     const val Menu = "Menu"
     const val Ambient = "Ambient"
+    const val Golden = "Golden"
+    const val Built = "Built"
```
alexban011, line 11: if this is the `Wonder built` mood I'd think `Wonder` or `WonderBuilt` would be clearer

OptimizedForDensity: Good point.

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -419,6 +421,10 @@ class CityConstructions : IsPartOfGameInfoSerialization {
                 else
                     civ.addNotification("[${construction.name}] has been built in a faraway land", buildingIcon)
             }
+
+            if (cityInfo.civInfo.isPlayerCivilization() && (cityInfo.civInfo.gameInfo.currentPlayerCiv == cityInfo.civInfo)) {
```
yairm210, line 425: Constructions usually complete at the *end* of your turn, but the wonder popup appears at the start of your next turn.
How about we move the music trigger to the wonder popup instead?

core/src/com/unciv/logic/civilization/GoldenAgeManager.kt
```diff
@@ -34,6 +37,9 @@ class GoldenAgeManager : IsPartOfGameInfoSerialization {
         turnsLeftForCurrentGoldenAge += turnsToGoldenAge.toInt()
         civInfo.addNotification("You have entered a Golden Age!", "StatIcons/Happiness")
         civInfo.popupAlerts.add(PopupAlert(AlertType.GoldenAge, ""))
+        if (civInfo.isPlayerCivilization() && (civInfo.gameInfo.currentPlayerCiv == civInfo)) {
```
yairm210, line 40: Here too, golden ages currently start at the end of your turn (see 'endTurn' below), so you'd click 'end turn' and then it would make a sound, which is strange.
This can easily be changed to have them start on the startTurn function with not problems I can think of.

core/src/com/unciv/logic/city/CityConstructions.kt
OptimizedForDensity: Oh good point, plus we won't need the if statement.

yairm210 (approved):

PR #7411, opened 2022-07-09
Add a simple GDX application to help with developing UI components

Azzurite:
This adds a little dev application to develop UI components a little faster. It basically just creates a libGDX and `UncivGame` context and then shows you your UI element. The main advantage is that it starts instantly, so you can make changes and almost immediately see them.

This is not perfect but it's much better than anything else which has been done before, like actually starting the game and navigating until you see your component you're developing.

It's kind of an open question if that should be within the tests module, it should probably be in its own module, but I think it's fine, it's much easier to add one class there than it is to add another project.

### DIFF
--- a/docs/Developers/UI-development.md
+++ b/docs/Developers/UI-development.md
@@ -1,0 +1,29 @@
+# UI Development
+
+Unciv is backed by [GDX's scene2d](https://libgdx.com/wiki/graphics/2d/scene2d/scene2d) for the UI, so check out [their official documentation](https://libgdx.com/wiki/graphics/2d/scene2d/scene2d) for more info about that.
+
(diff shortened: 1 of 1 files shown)


yairm210:
You're leaking Desktop dependencies into Core, which means they'll also get to the compiled Android APK, which means considerably bigger APKs.
Please move this to Desktop.

yairm210:
Oh, it looks like tests is *inside* core for some reason.

I'm not sure why we did that, but as long as you can ensure that APK compiling doesn't include the Desktop dependencies I'm okay with it.

And if it does, we should probably move tests outside of core.

yairm210 (approved):

yairm210:
Please add wiki documentation on how to use this as well

PR #7418, opened 2022-07-10
Fix any civ being able to capture barbarian units

OptimizedForDensity:
The current master branch has a bug where any civ can capture barbarian units. The cause is `listOf(false, false, false).any()` always returns true since `any()` on a list is only checking if the list is non-empty so the `.any()` was changed to `.any { it }` which returns true if any element is true.


yairm210:
Ayyy of course
Rookie mistake

PR #7421, opened 2022-07-10
Fix some graphics issues related to combat

### LINKED ISSUE #7419: Nation colors switched upon kill
**Platform**
Win 10 64-bit

**Version**
4.1.17 (Build 735) from Itch.io launcher

**Describe the bug**
When a melee unit kills an enemy unit and moves into its tile, the attacking unit will erroneously take the nation colors of the defeated unit (temporarily, until the next turn).

**Before kill:**
Before Kill

**After kill:**
After Kill

**To Reproduce**
1. Kill any enemy unit with a melee unit
2. See the error in color handling (note: requires a tileset with nation colors such as Hexarealm+AbsoluteUnits or 5Hex)

**Save game (before kill)**
[Color_test_before_kill.txt](https://github.com/yairm210/Unciv/files/9079586/Color_test_before_kill.txt)

OptimizedForDensity:
Closes #7419. The issue lays in
https://github.com/yairm210/Unciv/blob/cdecb3f52d535b4113d60ed0638d5cc509478e7b/core/src/com/unciv/ui/tilegroups/TileGroup.kt#L719-L722
(and equivalent for civilian units). This variable `pixelMilitaryUnitImageLocation` is the image path for the given military unit in the tile which seemingly prevents the game from reloading the tile image if it's unchanged from its new state `newImageLocation`. If a new image is on the tile, the image is loaded and nation colors applied. However in combat, the unit image can change if, for example, a Warrior kills another Warrior and moves onto its tile. Since the new image location is the same as the old (e.g., both are `path/to/images/Warrior` in the image files), the image isn't reloaded which prevents the nation colors from updating. In the linked issue (using Absolute Units as the mod), the Jaguar and Warrior share the same skin which demonstrates the issue. You can test in the same save with killing the Warrior using the Scout and the issue does not occur. This has been fixed by updating `pixelMilitaryUnitImageLocation` as `"${militaryUnit.civInfo.civName}-${newImageLocation}"` which will ensure nation colors update properly.

This fix led to another issue where now the victorious unit would not flash red due to taking damage. I'm not entirely sure of the cause but I think I narrowed it down to an ordering issue where the flashing red was applied to the pre-update `TileGroup` so when the `TileGroup` updated post-combat, it would remove the flashing image `Actor` and replace it with a non-flashing one. This wasn't happening before because before the above fix, the `TileGroup` is not updated so the flashing image is not unloaded. I think the cause is the same as a separate issue where if you select a melee unit 2 tiles away from an enemy and have it move and attack in the same action, your attacking unit will not flash red; you can test this issue by selecting either Jaguar in the linked save and having it attack Tyre starting from 2 tiles away. Both of these cases have been fixed by updating the `TileGroup` before applying the flashing red.

### DIFF
--- a/core/src/com/unciv/ui/tilegroups/TileGroup.kt
+++ b/core/src/com/unciv/ui/tilegroups/TileGroup.kt
@@ -719,10 +719,10 @@ open class TileGroup(
             newImageLocation = tileSetStrings.getThisUnit() ?: tileSetStrings.fallback?.getThisUnit() ?: ""
         }
 
-        val currentUnitName = if (militaryUnit != null) "${militaryUnit}:" else ""
-        if (pixelMilitaryUnitImageLocation != "$currentUnitName$newImageLocation") {
+        val nationName = if (militaryUnit != null) "${militaryUnit.civInfo.civName}-" else ""
+        if (pixelMilitaryUnitImageLocation != "$nationName$newImageLocation") {
             pixelMilitaryUnitGroup.clear()
-            pixelMilitaryUnitImageLocation = "$currentUnitName$newImageLocation"
+            pixelMilitaryUnitImageLocation = "$nationName$newImageLocation"
 
             if (newImageLocation != "" && ImageGetter.imageExists(newImageLocation)) {
                 val nation = militaryUnit!!.civInfo.nation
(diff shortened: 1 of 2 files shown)


letstalkaboutdune:
Awesome! Thanks for the quick work on this!

yairm210 (approved):

PR #7429, opened 2022-07-11
Disabled religion when there is a disableReligion unique 

alexban011:
this disables religion and hides the checkbox if the vanilla ruleset is selected. 
In civ 5 religion was only implemented in the G&K expansion so I doubt it'd be a good idea to be able to enable it in vanilla

### DIFF
--- a/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
+++ b/core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
@@ -3,7 +3,6 @@ package com.unciv.ui.newgamescreen
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.unciv.UncivGame
 import com.unciv.logic.civilization.CityStateType
-import com.unciv.models.metadata.BaseRuleset
 import com.unciv.models.ruleset.ModOptionsConstants
 import com.unciv.models.ruleset.RulesetCache
 import com.unciv.models.ruleset.unique.UniqueType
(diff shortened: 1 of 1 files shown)


xlenstra:
Imo the entire checkbox should be removed and a few versions from now also the setting

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -71,7 +72,10 @@ class GameOptionsTable(
         checkboxTable.addOneCityChallengeCheckbox()
         checkboxTable.addNuclearWeaponsCheckbox()
         checkboxTable.addIsOnlineMultiplayerCheckbox()
-        checkboxTable.addReligionCheckbox(cityStateSlider)
+        if (gameParameters.baseRuleset != BaseRuleset.Civ_V_Vanilla.fullName)
```
xlenstra, line 75: If you go this way, I'd check whether either the starting era or the mod have a unique disabling religion rather than checking specifically for the vanilla ruleset

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -64,14 +66,16 @@ class GameOptionsTable(
         }).row()
         addVictoryTypeCheckboxes()
 
+        val religionDisabledByRuleset = (ruleset.eras[gameParameters.startingEra]!!.hasUnique(UniqueType.DisablesReligion)
+                || ruleset.modOptions.uniques.contains(ModOptionsConstants.disableReligion))
```
yairm210, line 70: Move this into the addReligionCheckbox, since it's not being used by anything outside of it

yairm210: Also, do we update gameOptionsTable when selecting a starting era? Because if we don't then the first part will be locked onto the default starting era, and won't change with it.

OptimizedForDensity:
For whatever reason, the `update()` line in `Table.addEraSelectBox()` is giving me a stack overflow on the latest repo version.

OptimizedForDensity:
Doesn't stack overflow now, but it still doesn't update when the era changes. I played around with it and commenting out the extra `onChange()` line in `Table.addSelectBox` prevents the stack overflow, meaning you can add that `update()` back into the era select box
```
private fun Table.addSelectBox(text: String, values: Collection<String>, initialState: String, onChange: (newValue: String) -> String?) {
        add(text.toLabel()).left()
        val selectBox = TranslatedSelectBox(values, initialState, BaseScreen.skin)
        selectBox.isDisabled = locked
        selectBox.onChange {
            val changedValue = onChange(selectBox.selected.value)
            if (changedValue != null) selectBox.setSelected(changedValue)
        }
        //onChange(selectBox.selected.value)
        add(selectBox).fillX().row()
    }
```

Then `religionDisabledByRuleset = (ruleset.eras[gameParameters.startingEra]!!.hasUnique(UniqueType.DisablesReligion)
                || ruleset.modOptions.uniques.contains(ModOptionsConstants.disableReligion))` can be added to the top of `getGameOptionsTable()` and it will be set during both load and updates.

That being said, I haven't tested this too rigorously so I don't know if removing that extra `onChange` would have an effect.

OptimizedForDensity:
Closing this since #7606 has removed this checkbox entirely.

PR #7441, opened 2022-07-12
Significantly reduce AI turn time

OptimizedForDensity:
Corrected a large oversight in pathfinding in the function `UnitMovementAlgorithms.getShortestPath()` that results in a large reduction in AI turn time. This function, which is called in both `UnitMovementAlgorithms.headTowards()` and `canReach()`,  tries to get the shortest path between two tiles and does so by getting the furthest tiles the unit can travel to in 1 turn then iterating over them to see how far they can go in a turn from there, and so on until a path is found. This is basically a breadth-first search, but what was missing was a list of all of the tiles the algorithm has visited. Each tile only needs to be visited once and so the function as-is constantly visits old tiles. Adding in this list and modifying `getDistanceToTilesWithinTurn()` (which returns all of the tiles a unit with a given amount of movement can visit if they start from an origin tile) to have an optional argument `tilesToIgnore: HashSet<TileInfo>? = null` lets the function skip certain tiles if they've been visited.

To test this, I simulated 2 games on a large Pangaea map with no world wrap, 12 civs, no city states, deity difficulty, quick speed, and 350 turn limit. I also timed the following functions in `NextTurnAutomation.automateCivMoves()` for the AI to get a sense of how much cumulative time was spent performing each function per game.

Before this PR

<Details>

| Task | Time (milliseconds) |
|---|---|
| adoptPolicy      |                 944.864236 |
| automateCityBombardment      |     807.760365 |
| automateUnits          |        872389.613531 |
| chooseReligiousBeliefs     |        44.507917 |
| chooseTechToResearch      |         20.232754 |
| declareWar             |           340.098728 |
| exchangeLuxuries         |         267.420442 |
| freeUpSpaceResources        |       69.743166 |
| issueRequests          |             7.183230 |
| offerPeaceTreaty       |           376.313843 |
| offerResearchAgreement     |       224.571315 |
| reassignWorkedTiles       |     131759.110484 |
| respondToTradeRequests     |        59.387917 |
| trainSettler                |       18.646751 |
| tryVoteForDiplomaticVictory    |     3.143819 |
| useGold               |           2879.765154 |
Total time to simulate the game: 1010 seconds = 16 minutes, 50 seconds

</Details>

After this PR:

<Details>

| Task | Time (milliseconds) |
|---|---|
| adoptPolicy           |           1208.840178 |
| automateCityBombardment   |        437.996525 |
| automateUnits         |         312751.084442 |
| chooseReligiousBeliefs        |     60.395443 |
| chooseTechToResearch        |       19.824238 |
| declareWar             |           435.276269 |
| exchangeLuxuries         |         190.135690 |
| freeUpSpaceResources       |        62.513019 |
| issueRequests         |              4.775007 |
| offerPeaceTreaty         |         307.989132 |
| offerResearchAgreement    |        121.366432 |
| reassignWorkedTiles       |     158064.741318 |
| respondToTradeRequests    |         81.931555 |
| trainSettler          |             29.130064 |
| tryVoteForDiplomaticVictory    |     2.850685 |
| useGold                |          2259.913244 |
Total time to simulate the game: 476 seconds = just under 8 minutes

</Details>

This is only a sample size of 1 game under 1 set of conditions, but the time spent in `NextTurnAutomation.automateUnits()` decreased by around 65% and the total turn time decreased by about half. `NextTurnAutomation.automateUnits()` also went from being 86% of the total amount of AI turn time to being 66% (for reference, the second largest `reassignWorkedTiles` went from being 13% to 33%).

### DIFF
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -59,14 +59,14 @@ object NextTurnAutomation {
             protectCityStates(civInfo)
             bullyCityStates(civInfo)
         }
-        automateUnits(civInfo)  // this is the most expensive part
+        automateUnits(civInfo)  // very expensive
 
         if (civInfo.isMajorCiv()) {
             // Can only be done now, as the prophet first has to decide to found/enhance a religion
             chooseReligiousBeliefs(civInfo)
         }
 
-        reassignWorkedTiles(civInfo)  // second most expensive
+        reassignWorkedTiles(civInfo)  // very expensive
         trainSettler(civInfo)
         tryVoteForDiplomaticVictory(civInfo)
     }
--- a/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/NextTurnAutomation.kt
@@ -59,14 +59,14 @@ object NextTurnAutomation {
             protectCityStates(civInfo)
             bullyCityStates(civInfo)
         }
-        automateUnits(civInfo)  // very expensive
+        automateUnits(civInfo)  // this is the most expensive part
 
         if (civInfo.isMajorCiv()) {
             // Can only be done now, as the prophet first has to decide to found/enhance a religion
             chooseReligiousBeliefs(civInfo)
         }
 
-        reassignWorkedTiles(civInfo)  // very expensive
+        reassignWorkedTiles(civInfo)  // second most expensive
         trainSettler(civInfo)
         tryVoteForDiplomaticVictory(civInfo)
     }
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -675,6 +675,7 @@ class MapUnit : IsPartOfGameInfoSerialization {
     }
 
     fun useMovementPoints(amount: Float) {
+        movement.resetPathsToTilesWithinTurn()
         turnsFortified = 0
         currentMovement -= amount
         if (currentMovement < 0) currentMovement = 0f
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -675,7 +675,6 @@ class MapUnit : IsPartOfGameInfoSerialization {
     }
 
     fun useMovementPoints(amount: Float) {
-        movement.resetPathsToTilesWithinTurn()
         turnsFortified = 0
         currentMovement -= amount
         if (currentMovement < 0) currentMovement = 0f
--- a/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
+++ b/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
@@ -206,7 +206,7 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
         val newTilesToCheck = ArrayList<TileInfo>()
         val distanceToDestination = HashMap<TileInfo, Float>()
         var considerZoneOfControl = true // only for first distance!
-        val visitedTiles: HashSet<TileInfo> = HashSet()
+        val visitedTiles: HashSet<TileInfo> = hashSetOf(currentTile)
         while (true) {
             if (distance == 2) { // only set this once after distance > 1
                 movementThisTurn = unit.getMaxMovement().toFloat()
@@ -220,15 +220,15 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
                     // Avoid damaging terrain on first pass
                     if (avoidDamagingTerrain && unit.getDamageFromTerrain(reachableTile) > 0)
                         continue
-                    if (reachableTile == destination)
+                    if (reachableTile == destination) {
                         distanceToDestination[tileToCheck] = distanceToTilesThisTurn[reachableTile]!!.totalDistance
-                    else {
+                        break
+                    } else {
                         if (movementTreeParents.containsKey(reachableTile)) continue // We cannot be faster than anything existing...
                         if (!isUnknownTileWeShouldAssumeToBePassable(reachableTile) &&
                                  !canMoveTo(reachableTile)) continue // This is a tile that we can't actually enter - either an intermediary tile containing our unit, or an enemy unit/city
                         movementTreeParents[reachableTile] = tileToCheck
                         newTilesToCheck.add(reachableTile)
-                        visitedTiles.add(reachableTile)
                     }
                 }
             }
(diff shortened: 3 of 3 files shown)


yairm210:
You're missing an important point here.
It's not enough that we already managed to reach a tile, the real question is *in how many turns did we reach it*, and *with how many movement points left over*.
That's why we always check how long it will take us to reach that time from here, before we decide whether to ignore it.

yairm210:
I find this is a common misconception for people who try to reimplement pathfinding in a more efficient way - not only is this an edge graph and not just values-per-nodr, you can't even guarantee that the lowest sum of edges gives you the least turns! It's a real bitch tbh

yairm210:
To properly test correctness, we can have the external function wrap an internal one, and run the internal one twice - once with the new functionality and once without.
If we get different results, log them (a in 3 turns, b in 4)
We can also have a global timer counter to sum total runtime of both versions for comparison

yairm210:
@Azzurite - I'm pretty sure the answer is "no", due to the way that turns are delimitated.
A turn ends when you've expended all your points, but this means the lowest sum of distances is not always the shortest path!
For instance here:
image
Cutting through the hills is 8 movement points total, going around is 7.
But for a unit who has only 1 MP, going through the hills is 4 turns instead of 7; for a unit that has 3 MP, going through the hills is 2 turns instead of 3.
BUT for a unit with 2 MP, it's the same! So it's also 'max mp' dependent!
In fact you can create situations like this:
image

Where for a 2 mp and 1 mp unit the best path is drastically different.

I would love to be proven wrong though! If there's a library that would allow civ's idiosyncrasies, or in fact any solution that gives better-or-equal results, would be good!

yairm210:
> Looking closely at the original code, there is an issue. The first line in the else statement below will cause the for loop to skip over any tiles already checked. Regardless of this PR's suggestions, this line should probably be changed to a .containsKey check then the comparison of the actual costs. Also, the if (reachableTile == destination) branch should have a comparison in case the unit can reach the destination by another route that turn but with more movement left over.

@OptimizedForDensity Nice catch! Yes, we'll never get there in less *turns*, but we will get there with more *movement*! That is a fix we should make!

> Each iteration of the while loop can be thought of as a turn of movement, so the 0th iteration looks at all the tiles the unit can move to on the current turn with its current position and movement. If we can reach an intermediate tile in a turn, why should we evaluate going to this exact same tile on a later iteration? It will never be faster than going to the tile on the first turn they can. 

That's already how it works in the getShortestPath, and would not work in getDistanceToTilesWithinTurn.
Since getDistanceToTilesWithinTurn also saves how long it took us to get somewhere, which helps us decide if we have left over movement points with which to try again.
Like:
image

I am on the left, trying to get to the right. I have 2 MP.
If I take the first "mp left" I encounter, I'll figure that my movement ends once I reach the hill and I can't continue on.
But if I take the longer route, down to the grassland (1 mp) and then take the road (1/3 mp), I still have MP left to make it to the right!

(Unless I entirely misunderstood what you're trying to say)

OptimizedForDensity:
Here's what I mean, I will lay out the text to mimic the actions of the `while` loop in `getShortestPath`: a 2 movement unit starts on the green dot in the lower right and is trying to get to the red dot in the top left. It will initially look in a 2 tile radius of the green dot and would select all of the tiles on the snow ring since those are the furthest tiles it can move on the 0th turn; we want the algorithm to use up the movement fully each iteration since we want the unit to use up its movement each turn. The algorithm did visit the desert tiles but this is discarded in favor of the snow tiles. Therefore, `newTilesToCheck` contains all of the snow tiles and at the end of this `while` loop iteration, `tilesToCheck = newTilesToCheck`. This concludes the 0th iteration, corresponding to the _possible_ move the unit could make on the current turn.

Example1


The next iteration will loop over `tilesToCheck` which is entirely snow tiles. This iteration will check for tiles that it can reach within 2 turns of its starting position on the green dot. Let's say, for example, that the first element in `tilesToCheck` is the blue dot. The algorithm will get all tiles within one turn's movement of this blue dot. My proposal is that we should exclude all of the tiles that we can visit in turn 0 from the movement algorithm in turn 1, the reason being that *if we can reach a tile in turn 0 but choose a path where we end up visiting that same tile in turn 1 we are wasting movement*. The unit could go from the blue dot to the orange dot on turn 1 but why should we do that when we can go to the orange dot on turn 0? I hope this explanation helps.


Example2

Now for the test, I did what you suggested and extracted the old and new `getShortestPath` code then made two new functions `getShortestPathOld` (which is the current repo version) and `getShortestPathNew` (my proposed version which differs from this PR version since I made some fixes). I time both functions side by side, then also output if the length of the determined path differed between the two versions. I ran a simulated game with AI then copy-pasted the output into a csv file. This file contains the last 49612 calls to `getShortestPath` and there was not a single "Worse" message in all of these lines. At the same time, comparing the measured times shows that old `getShortestPath` function spent 173.8 cumulative seconds over these 49612 function calls while the new `getShortestPath` spent 37.4 cumulative seconds. This means that the new function that I'm proposing (the visited tiles list) is better than the old one in computational speed while not sacrificing actual pathfinding performance.

<Details>

```
fun getShortestPath(destination: TileInfo, avoidDamagingTerrain: Boolean = false): List<TileInfo> {
        var old: List<TileInfo>
        val oldTime = measureNanoTime {
            old = getShortestPathOld(destination, avoidDamagingTerrain)
        }

        var new: List<TileInfo>
        val newTime = measureNanoTime {
            new = getShortestPathNew(destination, avoidDamagingTerrain)
        }
        println("Timing $oldTime $newTime")
        if (old.size < new.size)
            println("Worse 0 ${new.size - old.size}") // new is worse than before
        else if (old.size > new.size)
            println("Better ${old.size - new.size} 0") // new is better than before
        return old
    }
```

</Details>

OptimizedForDensity:
> @OptimizedForDensity Nice catch! Yes, we'll never get there in less _turns_, but we will get there with more _movement_! That is a fix we should make!

So I just doubled checked and I think I was confusing myself. `getDistanceToTilesWithinTurn` returns a `LinkedHashMap` so it won't have duplicate tiles therefore the `if (reachableTile == destination)` is executed only once per iteration of the `for (tileToCheck in tilesToCheck)` loop (the `getDistanceToTilesWithinTurn` also makes sure to get the smallest possible `totalDistance` for each tile). Past the `for (tileToCheck in tilesToCheck)` loop there's another section of code
```
// Get the tile from which the distance to the final tile in least -
// this is so that when we finally get there, we'll have as many movement points as possible
                var intermediateTile = distanceToDestination.minByOrNull { it.value }!!.key
```

yairm210 (approved):
THAT is nice! If we could reach this tile on a *previous* turn there is indeed no need to examine it!

PR #7450, opened 2022-07-14
Add multiplayer server per game instead of one global server for all games

Azzurite:
So our current multiplayer architecture is based on that we are FOSS. It should be possible for anyone to host a multiplayer server and have their games played on that. The goal is to not be dependent on one central service like in so many other applications.

The problem with that is that currently, you define one single server that you want to use in your settings, and that's it. What if you have a server with a few games, and another friend has another server with a few games, and you want to play together? Well, better have fun changing the server all the time in the settings, or run multiple copies of the game from different directories. This basically works against our goal of everyone being able to host their own server, since it is encouraging to use one central one to be less annoying.

This PR solves that by attaching the multiplayer server a game is played on to the game itself, and not use the global one from the settings. The global one in the settings is now only a default one for pre-filling the game-specific server.

Now the "Multiplayer -> Add game", "Multiplayer -> Game settings" and "New Game" screens all have an UI element to change the server the game is stored on, which are all the visible changes, everything else is backend stuff to support that.

---

I made this one a draft because this is only the first, most basic implementation to get the feature working. There are a bunch of issues still that I want to address, like for example [the new game screen interface being cut off by default](https://user-images.githubusercontent.com/2777158/178976576-bdb8c415-59b8-46ec-928e-7cd1d19eaa95.png) and some other things:

* [x] Refactor: Change map/event keys from game name to game ID (remember to remove `Multiplayer.changeGameName`)
* [x] [887dcf1](https://github.com/yairm210/Unciv/pull/7450/commits/887dcf1c33df794147c6f8b238060ef0aa55f73d): Better ServerInput layout for NewGameScreen. The server input now simply gets its width from the column width (which already does not fit on smaller resolutions, but I won't fix that here)
* [x] Default server in settings should make clear that it's only the default

and maybe more if I or someone finds other issues.


OptimizedForDensity:
Being able to save servers in the same way you can save friends then select them from a menu when making a new game would be nice.

core/src/com/unciv/logic/multiplayer/storage/SimpleHttp.kt
```diff
@@ -67,4 +80,20 @@ object SimpleHttp {
...
+                acc + cur
+            } else {
+                acc + "/" + cur
+            }
+        }
+        return if (joined.endsWith('/')) {
+            joined.substring(0, joined.length - 1)
+        } else {
+            joined
+        }
```
yairm210, line 97: I think there's a removeSuffix function in string, so this could just be joined.removeSuffix("/")

core/src/com/unciv/logic/multiplayer/storage/UncivServerFileStorage.kt
```diff
@@ -5,48 +5,41 @@ import com.unciv.utils.debug
...
-        SimpleHttp.sendRequest(Net.HttpMethods.DELETE, fileUrl(fileName), "", timeout) {
-                success, result, code ->
-            if (!success) {
-                when (code) {
-                    404 -> throw MultiplayerFileNotFoundException(Exception(result))
-                    else -> throw Exception(result)
-                }
-            }
+    override suspend fun deleteFile(fileName: String) {
+        val (success, error, code) = SimpleHttp.sendRequest(Net.HttpMethods.DELETE, fileUrl(fileName), "", timeout)
```
yairm210, line 30: That is very cool, didn't know you could do that with data classes!

yairm210 (approved):
I'm sure there will be something we've overlooked, but I'm sure we'll sort them out in due time :)

PR #7457, opened 2022-07-15
Add moveTo to template.properties

itanasi:
String is not translatable and used in Unit Overview

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -884,6 +884,7 @@ Ranged strength =
 Bombard strength = 
 Range = 
 Move unit = 
+moveTo = 
 Stop movement = 
 Swap units = 
 Construct improvement = 


MindaugasRumsa51:
i think this is not enough

xlenstra (approved):

xlenstra:
Imo "moveTo" should not be displayed ajd should be replaced with "Moving Towards" or something

alexban011:
where (and when) does this appear? Isn't this a debug feature? Pretty sure we don't need to translate this if it is just for debugging

xlenstra:
iirc its displayed in the unit table for units preforming multi-turn movements

Ouaz:
moveto

By the way, what's the meaning of these coordinates ? Tile -28, -14?

Why not get rid of it (who uses it?) and just use "`Moving`". So we'll have:` Set up, Fortify, Sleep until healed, Sleep, Moving.`

itanasi:
Those are the coordinates of the tile it's moving to

OptimizedForDensity (commented):
This fix doesn't work for me since the game is trying to translate `"moveTo -28,-14".tr()` or whatever the coordinates are, which means that translating is impossible unless the code is modified to put {} around the "moveTo". That would likely break the movement system since it looks for "moveTo" at the beginning of the `action` string so this fix requires more than an extra translation line unfortunately.

Ouaz:
> Those are the coordinates of the tile it's moving to

Yes, that's the part I understood. ^^

What I don't understand is their meaning? Some kind of x-axis, y-axis?

I'm not sure any player use this data, so why not removing it, and just keep "Moving" (and no problem to translate). Or this data is useful, and forget my suggestion. :p

itanasi:
I doubt players understand how to read the coordinates. However, a cool feature would be clicking on it and it zoomed to that tile

alexban011:
> By the way, what's the meaning of these coordinates ? Tile -28, -14?

yeah thats why I thought its just for debugging. The users will never care what the exact coordinates are

PR #7459, opened 2022-07-15
fix settler and one city challenge 

MindaugasRumsa51:
tested ingame

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -80,9 +80,17 @@ object UniqueTriggerActivation {
             OneTimeFreeUnitRuins -> {
                 var unit = civInfo.getEquivalentUnit(unique.params[0])
                 if ( unit.hasUnique(UniqueType.FoundCity) && civInfo.isOneCityChallenger()) {
-                    if (ruleSet.units[Constants.worker] != null) {
-                        unit = civInfo.getEquivalentUnit(Constants.worker)
-                    } else {
+                    var found = false
+                    for (unitForReplace in ruleSet.units.values) {
+                        if(unitForReplace.getMatchingUniques(UniqueType.BuildImprovements)
+                           .any { it.params[0] == "Land" } ) {
+
+                            unit = civInfo.getEquivalentUnit(unitForReplace.name)
+                            found = true
+                            break
+                        }
+                    }
+                    if (!found) {
                         return false
                     }
                 }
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -78,7 +78,11 @@ object UniqueTriggerActivation {
                 return true
             }
             OneTimeFreeUnitRuins -> {
-                val unit = civInfo.getEquivalentUnit(unique.params[0])
+                var unit = civInfo.getEquivalentUnit(unique.params[0])
+                if ( unit.hasUnique(UniqueType.FoundCity) && civInfo.isOneCityChallenger()) {
+                    unit = civInfo.getEquivalentUnit(Constants.worker)
```
yairm210, line 83: Remember that in modded games they may not have a worker unit! Or they could have one with a different name!

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -78,7 +78,15 @@ object UniqueTriggerActivation {
                 return true
             }
             OneTimeFreeUnitRuins -> {
-                val unit = civInfo.getEquivalentUnit(unique.params[0])
+                var unit = civInfo.getEquivalentUnit(unique.params[0])
+                if ( unit.hasUnique(UniqueType.FoundCity) && civInfo.isOneCityChallenger()) {
+                    if (ruleSet.units[Constants.worker] != null) {
+                        unit = civInfo.getEquivalentUnit(Constants.worker)
```
yairm210, line 84: You're still using hardcoded unit names, please search for units by type and unique (like you did to see if it's a settler!)

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -78,7 +78,23 @@ object UniqueTriggerActivation {
...
+                    var found = false
+                    for (unitForReplace in ruleSet.units.values) {
+                        if(unitForReplace.getMatchingUniques(UniqueType.BuildImprovements)
+                           .any { it.params[0] == "Land" } ) {
+
+                            unit = civInfo.getEquivalentUnit(unitForReplace.name)
+                            found = true
+                            break
+                        }
+                    }
```
yairm210, line 92: can be simplified to replacementUnitl = ruleSet.units.values.firstOrNull{it.getMatchingUniques(UniqueType.BuildImprovements)
                           .any { it.params[0] == "Land" }} ?: return false

yairm210 (approved):

MindaugasRumsa51:
question:  in code there are 2 other similar pieces.
for example "AlertPopup.kt" file have line:
captor.placeUnitNearTile(tile.position, Constants.worker)

maybe those need fix too?

OptimizedForDensity:
Nice catch with getting settlers from ruins on OCC. Regarding the other cases of `Constants.worker`, those can be handled in a later PR, this is good for now.

OptimizedForDensity (approved):

PR #7461, opened 2022-07-15
Simplified yields - currently without toggle option

yairm210:
image

Lategame maps are generally so chock-full of info that the whole thing is very messy.
Showing yields exacerbates this to the point of nonsense.

This PR minimizes the noise by removing the stat icons from yields, only showing the stat colors.

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/YieldGroup.kt
+++ b/core/src/com/unciv/ui/cityscreen/YieldGroup.kt
@@ -3,7 +3,9 @@ package com.unciv.ui.cityscreen
 import com.badlogic.gdx.graphics.Color
 import com.badlogic.gdx.scenes.scene2d.ui.HorizontalGroup
 import com.badlogic.gdx.scenes.scene2d.ui.Table
+import com.unciv.models.stats.Stat
 import com.unciv.models.stats.Stats
+import com.unciv.ui.images.IconCircleGroup
 import com.unciv.ui.images.ImageGetter
 import com.unciv.ui.utils.extensions.surroundWithCircle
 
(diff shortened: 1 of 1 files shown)


xlenstra:
Not sure I like this, I feel like if you feel like you already have too much information you can simply disable the yields. If you do want this information, it might as well look good

alexban011:
I'd say to add an option to choose between the old yields and this (but the smaller size, like in the poll).
Or like someone said on discord: Have the yield button toggle  `Off-OldYields-NewYields`

Ouaz:
> Not sure I like this, I feel like if you feel like you already have too much information you can simply disable the yields. If you do want this information, it might as well look good

I think it would be A BIG improvement in city view, where it's a big mess of tile yields, hard to read, when trying to manually assign citizens for example.

yields

alexban011:
> image

ohhh god... my eyes...

what about always having this PR in the city screen and in the worldscreen the discord suggestion I talked about

Ouaz:
Why not just make the yields icon the same size than the small icon already displayed above resources icon -production over strategic and happiness over luxuries - (still with the semi-transparent black outline of this PR)? They are readable both on PC and mobile.

mini

Won't it look good with other icons (i.e. science, gold and faith)?

yairm210:
It seems everyone is saying that what we really need is for the icons to stay the same but just be smaller.
Extremely doable

Ouaz:
Smallest as possible but with the icon still readable for colorblind people.

xlenstra:
> Is this the first feature in the game that might actually only be used by non-colorblind people? I think for everything else the colors are either irrelevant or there is an icon, right?

Nation borders when you haven't found the city yet is another example, but there are very few examples as far as I know

Issue #4674, opened 2021-07-29
Implementation of Diplomacy


ajustsomebody:
I am creating this thread to list out what is implemented in diplomatic relations with other civilizations and what isn't. I really couldnt wait for yairm to reply to my comment therefore i guess he can close this he wants but i have put a lot of time researching these so.

Please point it out if there are problems about any of the below lines, whether it be that such mechanics don't exist in G&K or if it has been implemented or just plain wrong.

(ND) stands for Needs Description, i will be using this often as i write down basic features of a mechanic and not specifics of it. Will come back and fill those blanks in.

Well, there goes my 2 hours.

# Important!
Please, if you are going to implement any of the features below please dont take my words as absolute values in the game as i have marked most of them (ND). Please do your own research as well.

# Embassies

- [ ] For Embassies to be implemented. (G&K)

- Embassies require writing to be exchanged, and you should be able to pay opposing party in a trade request a small smount of gold (ND) for them to exchange embassies to you.

° When you exchange embassies with the other player, the borders of the capital city of that civilization will be implemented, you will not gain permanent vision like in allied city-states however.
° You gain a small amount (ND) of relations with that civilization.
° In case of a denounction, both embassies of those two civs are closed down until you re-open them with an action (ND).
° That action could be done right after the turn you have denounced the civ. If you are at war with that civilization you need to negotiate peace first before that action becomes available.
° Most of the treaties (ND) are only available if you have mutual embassies, with an exception of Declaration of Friendship.
° (ND) I gotta check if you both need mutual embassies to do most of the diplomatic stuff that you need embassies for.
° edit: embassies should give some good relations between civs</details> 

# Relations

- [x] In civ 5 there are 5 diplomatic relation status with civilizations, with those being Neutral, Friendly, Guarded, Hostile and Afraid
Neutral: They are neither friendly, nor hostile. Negotiations will be relatively easy.
Friendly: They are well disposed towards you, and you may hope to take advantage of that in negotiations.
Guarded: They are wary of you and your intentions and, as such, negotiations will be more difficult.
Hostile: They are angry with you, and very willing to declare war on you. Negotiations will be almost impossible.
Afraid: They fear your military superiority or your victory progress. They will often comply with anything you want in negotiations.

-----------------------------------------------------------------

- [x] Global Politics screen, this shows invaluable information about diplomacy of other civs with other civs. example image: https://www.carlsguides.com/strategy/civilization5/gamepictures/diplomacy/politicss.gif

[...]

PR #7469, opened 2022-07-15
Global politics overview screen

alexban011:
resolves part of #4674 

Adds the Global Politics screen from Civ 5

image


alexban011:
Maybe the `Social Policies:` text could be slightly smaller? What do you think?

OptimizedForDensity:
I like it a lot! A few comments:
* Column headers would be nice. I'm not too familiar with Deciv so I don't know if Clan Hideout is a city or wonder or what
* Smaller "Social Policies" text since that label sticks out too much (maybe addressed through a header mentioned above)
* Part of me thinks that the "Politics" tab should be located near the diplomacy tab since they're both geopolitical in nature but this is a very minor critique (or maybe even merged where the player's trade deals are listed on this "Politics" screen)

android/assets/jsons/translations/template.properties
```diff
@@ -1187,6 +1187,12 @@ Number of your cities celebrating\n'We Love The King Day' thanks\nto access to t
 WLTK demand = 
 WLTK- = 
 Number of your cities\ndemanding this resource for\n'We Love The King Day' = 
+Social Policies: = 
+At war with [enemy] = 
+Friends with [civName] = 
+( [numberOfTurns] Turns Left) =
```
OptimizedForDensity, line 1193: I feel like this entry shouldn't be in parentheses. Maybe do something like `( {[numberOfTurns] Turns Left})` in code then have `[numberOfTurns] Turns Left)` as the entry. The braces will handle the translation without needing translators to observe the parentheses and any extra spacing.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -241,6 +242,11 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
     @Transient
     var hasLongCountDisplayUnique = false
 
+    // (CivName, turnWhenThisHappened)
+    var atWarWith = mutableListOf<String>()
```
OptimizedForDensity, line 246: These seem like they should be `@Transient` then `CivilizationInfo.setTransients()` calls `updateGlobalPoliticsInfo()` which sets these variables.

core/src/com/unciv/logic/civilization/WondersInfo.kt
```diff
@@ -0,0 +1,191 @@
...
+import com.unciv.logic.city.CityInfo
+import com.unciv.logic.map.TileInfo
+import com.unciv.models.ruleset.Building
+import com.unciv.models.ruleset.Era
+import com.unciv.models.ruleset.QuestName
+import com.unciv.models.ruleset.unique.UniqueType
+import com.unciv.models.translations.tr
+import com.unciv.ui.civilopedia.CivilopediaCategories
+
+class WondersInfo {
```
OptimizedForDensity, line 14: `WondersInfo` vs `WonderInfo` seems a bit confusing so a name change might be helpful.

android/assets/jsons/translations/template.properties
alexban011: > `[numberOfTurns] Turns Left)`

I assume you meant without the closing bracket right?

OptimizedForDensity: Yes, that was a type

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
alexban011: I set them Transient but I really don't remember what I had in mind with `updateGlobalPoliticsInfo()`. I think I used it at some point and then forgot to remove it afterwards. Should the variables still be transient?

android/assets/jsons/translations/template.properties
alexban011: I updated the picture. Thanks for the suggestions :D

android/assets/jsons/translations/template.properties
```diff
@@ -1187,6 +1187,16 @@ Number of your cities celebrating\n'We Love The King Day' thanks\nto access to t
...
 Number of your cities\ndemanding this resource for\n'We Love The King Day' = 
+Social Policies: = 
+At war with [enemy] = 
+Friends with [civName] = 
+[numberOfTurns] Turns Left = 
+Denounced [otherCiv] = 
+Allied with [civName] = 
+Civilization Info = 
+Social Policies = 
+Wonders =
```
OptimizedForDensity, line 1198: `Wonders` is already an entry in this file, as is `Social policies` (note the lowercase P)

alexban011: fixed :)

android/assets/jsons/translations/template.properties
alexban011: Oh nice why haven't I thought about that :)) thanks!

alexban011:
> Who's at war with who is not shown at all

it is but it only shows that if the declaration happened while having this PR. I couldn't find a place where this data is stored so I made a hashmap to store this.

> [Denounced Russia( 0 turns left)](https://user-images.githubusercontent.com/2777158/179360251-8942fc7f-8910-4ca9-ab91-69a50f5a3bcf.png)

hmmm yep I can reproduce. It was working before so I'll see why it isn't anymore

> Russia( 0 turns left) should be Russia (0 turns left)

Alright, I changed this

> Also I really think this should replace the Diplomacy screen

:( I didn't wanna replace that since I kinda liked how that was done. At first I tried to add this screen to the bottom but then just made another screen. In civ 5 this screen is called "Global Politics", I named it just Politics because I don't think we can have spaces in those names. After deleting the Diplomacy screen, should I keep the `Politics` or `Diplomacy` name?

Ouaz:
Remove the Diplomacy screen? Errr... that's a bit harsh. Nobody will like that. Just ask on discord. ^^

I think the two are complementary. The diagram is **pretty useful** to see relations at a glance. Even if I like this new Politics screen (with additional infos as other civ current era, wonders built, and  policies adopted), I think removing the diagram is a bad idea.

And the diagram has more relation infos that are currently missing in this PR as `Favorable` (dull green), `Afraid` (purple), `Competitor` (dull blue), `Enemy` (yellow) and `Unforgivable` (crimson red).

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
OptimizedForDensity: Yes they should since `CivilizationInfo` inherits the interface `IsPartOfGameInfoSerialization` meaning that adding in new variables will change game saves unless the variables are `@Transient` or `lateinit` or `by lazy`.

Ouaz:
> Yep, I can see exactly what I want at a glance :D And if you select only one civ, I'd say it's basically the same as the table...

Haha, that's an extreme case. But even there, by only looking at the start of the lines, I can see that England is at war with everybody, Almaty is afraid by Egypt, Genoa is enemy with Florence, and Cahokia... oh god... who cares of City-states anyway? XD

> Though what the colors actually mean is never explained anywhere either, so I'm still not sure how useful it really is...

Yeah, I had to compare with the diplomacy screen status to deduct the color meaning the first time. Once done, it's useful (well, at least for me ^^).

alexban011:
> I had to compare with the diplomacy screen status to deduct the color meaning the first time. Once done, it's useful (well, at least for me ^^).

Same for me. I mean if you see them once / a couple times it becomes clear what they mean

alexban011:
I'd love if someone tests this as well. Last time I worked on this it wasn't clearing the old diplomacy screen well, but now it worked. 

Today I worked on windows and when it didn't clear it correctly I worked on linux. Tomorrow I will test it on linux as well even though I don't think there should be differences between the 2 desktops so I don't really know what happened back then.. :)

AdityaMH:
bug can easier found when tested to many player. but need tagged as pre-released.

yairm210 (approved):
Anyone have any objections?

PR #7470, opened 2022-07-16
divide by zero fix (also 'map too small for such amount of players')

MindaugasRumsa51:
i did long term test and stress test of this
for stress test i took my Anroid 5.0 1 GB ram huawei phone
and created TINY map with 58 free cities and max civilizations
i can play without problems

also tested the same TINY with new phone starting future era where at start game gives lot of units. result - map have no free place from units, but you can play, no problems

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
@@ -1389,8 +1389,8 @@ class MapRegions (val ruleset: Ruleset){
 
         // Third add some minor deposits to land tiles
         // Note: In G&K there is a bug where minor deposits are never placed on hills. We're not replicating that.
-        val frequency = (baseMinorDepositFrequency * bonusMultiplier).toInt()
-        val minorDepositsToAdd = (landList.size / frequency) + 1 // I sometimes have division by zero errors on this line
+        val frequency = maxOf((baseMinorDepositFrequency * bonusMultiplier).toInt(),1) 
+        val minorDepositsToAdd = (landList.size / frequency) + 1 
         var minorDepositsAdded = 0
         for (tile in landList) {
             if (tile.resource != null || tileData[tile.position]!!.impacts.containsKey(ImpactType.Strategic))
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
```diff
@@ -1389,8 +1389,8 @@ class MapRegions (val ruleset: Ruleset){
 
         // Third add some minor deposits to land tiles
         // Note: In G&K there is a bug where minor deposits are never placed on hills. We're not replicating that.
-        val frequency = (baseMinorDepositFrequency * bonusMultiplier).toInt()
-        val minorDepositsToAdd = (landList.size / frequency) + 1 // I sometimes have division by zero errors on this line
+        val frequency = maxOf((baseMinorDepositFrequency * bonusMultiplier).toInt(),1) 
+        val minorDepositsToAdd = (landList.size / frequency) + 1
```
alexban011, line 1393: `baseMinorDepositFrequency` is the constant value 30
`bonusMultiplier` can be 1.5, 0.66 or 1
I really can't tell why it would give a divide by zero. Do you have a save-file so I can test?

core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
```diff
@@ -1452,7 +1452,7 @@ class MapRegions (val ruleset: Ruleset){
                 // If there is no matching list, it is because the rule was determined to be impossible and so can be safely skipped
                     ?: continue
                 // Place the resources
-                placeResourcesInTiles((rule.params[0].toFloat() * bonusMultiplier).toInt(), list, listOf(resource), 0 + extraImpact, 2 + extraImpact, false)
+                placeResourcesInTiles(maxOf((rule.params[0].toFloat() * bonusMultiplier).toInt(),1), list, listOf(resource), 0 + extraImpact, 2 + extraImpact, false)
```
alexban011, line 1455: there is no division here so I also can't see why this would give that error

OptimizedForDensity: `placeResourcesInTiles(frequency: Int, ...)` has a line `val amountToPlace = (tileList.size / frequency) + 1`, although it might be better for him to put the fix in the function itself since the function is called in a few places.

MindaugasRumsa51:
sorry
recent testings shows that this bug fixed somewhere else
my info is was 2 months old
so i closing PR

but believe me, fist time i had met this bug, i had met comment about it too,
comment that shows awareness about problem....

xlenstra:
I think that comment might have been mine. Not sure what I was doing at the time as I never had that problem ever since, but I do know that I was able to reproduce it consistently back then.

Issue #7491, opened 2022-07-20
Memory optimizations 3 - Tokyo Drift


yairm210:
Continuing our [tradition](6695), let's get started!

Using a [massive game](https://discord.com/channels/586194543280390151/588357826758574106/999182264422383638) at turn 600, after disabling multiplayer.
As always, numbers are not exact, but proportions are, so we'll be using "memory taken by json" as the constant when dealing with game loading.

First attempt:
![image](https://user-images.githubusercontent.com/8366208/179974398-3107b4b3-b95b-42a1-b83a-c9f7ca4f2c64.png)

A. getImprovementStatsForCity weighs is at 11.4MB/22.5 = about half a JSON, or at about 12% of GameInfo.setTransients() (86MB)

Since the callstack is cityStats.update -> updateTileStats -> tileInfo.getTileStats -> getImprovementStats, we can use tha local unique cache that we already use in tileIngo.getTileStats to save the city uniques for usage in getImprovementStats as well :D

If we add in getTilePercentageStats, that's another 4.5MB we can save by caching citywide uniques

B. getTileStats.matchesTerrainFilter is 2MB of terrain querying, getTilePercentageStats is another 4.5, so that's hopefully about a quarter-JSON we can save by saving the tile uniques in a dedicated uniqueMap - PR #7490

yairm210:
Apart from caching religion uniques, which we should totally do, we're ready to move on to nextTurn!

![image](https://user-images.githubusercontent.com/8366208/180301194-498ada59-86c6-4646-9769-488a103f944d.png)

A. CityInfoReligionManager.getSpreadRange() is a whopping 1/3 of nextTurn memory - what the hell is going on there? Is this another 'just cache religion unqiues already' sign? (Answer: YES)
B. placeUnitNearTile is about a third of startTurn() - most of which goes to updateDetailedCivResources(). Do we check if the unit even consumes any resources before updating everything? If not that's an easy win
C. mapUnit.hasUnique() is taking longer than expected for what *should* be a simple hash lookup - check into that
D. setTerrainFeatures() -> addUniques() is about 1/3 (magic number?) of TileMap.setTransients, we can probably optimize it by checking if there are no terrainFeatures or nat wonders and taking the baseTerrain's uniquemap directly, most tiles are probably like that :)
E. CityConstructions.getStats() iterates over all city buildings, which take city uniques - could maybe benefit from caching if not done yet. Same for CityStats.update
[...]

PR #7492, opened 2022-07-20
40% memory save on tileInfo.getImprovementStats() by caching citywide uniques

yairm210:
Part of  #7491

Separates getImprovementStatsForCity() into its own function and sub-functions for better performance drilldowns 

This is mostly for cityStats.update() but will also cause performance improvements in CityScreen and ImprovementPickerScreen due to the same citywide caching

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -387,7 +387,12 @@ object Automation {
 
         // Improvements are good: less points
         if (tile.improvement != null &&
-            tile.getImprovementStats(tile.getTileImprovement()!!, cityInfo.civInfo, cityInfo).values.sum() > 0f
+            tile.getImprovementStats(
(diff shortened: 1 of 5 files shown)


yairm210:
I've so far only been testing the initial loading of the game - gameInfo.setTransients()

The way I've been doing this is by:
- Connecting my phone to the PC
- Android Studio - "profile android" - Memory - select "Record Java/Kotlin allocations", DO NOT hit record yet
image
- Open load game screen in phone, click on game to load
- NOW hit "record" on PC, and click "load game" on the phone
- Memory allocation will start recording, switch from Table to Visualization

This enhancement speeds up tileInfo.getImprovementStats(), which is used in the following places:
- gameInfo.setTransients (loading game from file, or 'next turn')
- Updating city stats (New citizens, assigning citizens, building buildings, anything that calls CityStats.update())
- CityScreen and ImprovementScreen, as mentioned

PR #7516, opened 2022-07-23

nightingale999:


android/assets/jsons/translations/Traditional_Chinese.properties
```diff
@@ -6413,12 +6405,10 @@ Building = 建築
 Unit = 單位
 UnitType = 職業
 Promotion = 晉升
- # Requires translation!
-Improvement = 
+Improvement = 升級
```
heipizhu4, line 6408: Improvement here means the 'building' you build on tiles with workers, not an upgrade.

heipizhu4 (approved):

PR #7522, opened 2022-07-23
Update Polish.properties

pyotr71:


OptimizedForDensity:
```  Placeholder `[building]` not found in `Polish` for entry `Built [building]` ```

On line 1324, you have `Built [building] = Zbuduj` but the translation needs to have `[building]` in it so either `Built [building] = [building] zbuduj` or `Built [building] = Zbuduj [building]` or whatever the correct translation would be.

PR #7543, opened 2022-07-28
Update Ukrainian.properties - add missing translations

estorski:


estorski:
well, the check didn't work out, but the thing is - the original string contains two placeholders
```
com.unciv.testing.TranslationTests > noTwoPlaceholdersAreTheSame STANDARD_OUT
    Entry {Ukrainian=Наш [attackerName] ([amount]) знищено при перехопленні ворожим підрозділом «[interceptorName]» ([amount])} has the parameter [amount] more than once
    Entry {Ukrainian=Наш [interceptorName] ([amount]) перехопив та знищив ворожий [attackerName] ([amount])} has the parameter [amount] more than once
    Entry {Ukrainian=Наш [attackerName] ([amount]) знищив перехоплюючий [interceptorName] ([amount])} has the parameter [amount] more than once
    Entry {Ukrainian=Наш [interceptorName] ([amount]) було перехоплено та знищено ворожим [attackerName] ([amount])} has the parameter [amount] more than once
    Entry {Ukrainian=Наш [attackerName] ([amount]) було атаковано перехоплюючим [interceptorName] ([amount])} has the parameter [amount] more than once
    Entry {Ukrainian=Наш [interceptorName] ([amount]) перехопив та атакував ворожий [attackerName] ([amount])} has the parameter [amount] more than once

com.unciv.testing.TranslationTests > noTwoPlaceholdersAreTheSame FAILED
    java.lang.AssertionError: This test will only pass when no translation keys have the same parameter twice
```

xlenstra:
Jup, looks like a problem on our side. I'll try to fix this

estorski:
I want to merge this without the conflicting strings. I suspect the originals are going to change anyway a few times more, so will wait until the dust settles.
Thus, I've updated the translations to exclude the changes to the lines in question. Now, it could be merged.

xlenstra:
It should be fixed, so you can re-add your translations if you want to

estorski:
> It should be fixed, so you can re-add your translations if you want to

Yes, thank you. I'll do it with a separate PR

estorski:
> It should be fixed, so you can re-add your translations if you want to

It seems like twice the placeholder (`[amount]`) is still present. I'm curious how it works. Because from what it looks like, the same value is being supplied twice, into two different contexts. From my perspective, it looks like a _logical_ mistake.

JackRainy:
I have picked your changes and updated the conflict strings in #7560

yairm210:
Deconflicted in #7560

PR #7544, opened 2022-07-28
Removed duplicate placeholders

xlenstra:
Removes duplicate placeholders in translation strings introduced in the Air Sweep PR


yairm210:
Maybe we can add a test for this

PR #7547, opened 2022-07-28
Update Dutch.properties

TPunkiee:


TPunkiee:
The original untranslated string contains 2 references to the same variable:
Our [attackerName] ([amount]) was destroyed by an intercepting [interceptorName] ([amount])
Isn't the first occurrence unnecessary?

xlenstra:
Both display the amount of damage done to the attacker and interceptor respectively, so both are necessary. However, the second one should have been `amount2` instead of `amount`. We'll automatically change this next time a version is released, but if you want to translate these strings now you need to change the second occurrence in the untranslated string.

OptimizedForDensity:
Conflicts

PR #7555, opened 2022-07-29
Update Italian.properties

Smashfanful:


Smashfanful:
I checked every edit in the build, how is it possible that nothing seems off yet the build still fails? What did I actually do wrong?

OptimizedForDensity:
The templates are wrong. A bunch of the interception templates have two `[amount]` entries so one needs to be renamed in both the original English and in Italian.

Smashfanful:
Thank you. Just done it.

PR #7564, opened 2022-07-31
Fix erroneous relationship decay notification

oynqr:
Whenever your relationship resting point as well as the relationship
score for city states is at the exact value of 30 or 60, a notification
is shown that the relationship is about to degrade, even though it can
and will not. This fixes said issue by adding checks for the resting
points to getTurnsToRelationshipChange().

A recent example where I encountered this was in a game with many city
states where I had the Aesthetics policy and pledged to protect all of
them, which led to a resting relationship value of 30 with them.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -260,7 +260,7 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
     }
 
     private fun getCityStateInfluenceDegrade(): Float {
-        if (getInfluence() < getCityStateInfluenceRestingPoint())
+        if (getInfluence() <= getCityStateInfluenceRestingPoint())
             return 0f
 
         val decrement = when {
@@ -288,7 +288,7 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
     }
 
     private fun getCityStateInfluenceRecovery(): Float {
-        if (getInfluence() > getCityStateInfluenceRestingPoint())
+        if (getInfluence() >= getCityStateInfluenceRestingPoint())
             return 0f
 
         val increment = 1f  // sic: personality does not matter here


OptimizedForDensity (approved):

OptimizedForDensity (commented):
Whoops, didn't mean to approve yet, that was a misclick. I'm going to double check in game first.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -208,8 +208,8 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
             val dropPerTurn = getCityStateInfluenceDegrade()
             return when {
                 dropPerTurn == 0f -> 0
-                relationshipLevel() >= RelationshipLevel.Ally -> ceil((getInfluence() - 60f) / dropPerTurn).toInt() + 1
-                relationshipLevel() >= RelationshipLevel.Friend -> ceil((getInfluence() - 30f) / dropPerTurn).toInt() + 1
+                relationshipLevel() >= RelationshipLevel.Ally && 60f > getCityStateInfluenceRestingPoint() -> ceil((getInfluence() - 60f) / dropPerTurn).toInt() + 1
```
OptimizedForDensity, line 211: I think these changes should be in the functions `DiplomacyManager.getCityStateInfluenceDegrade()` instead. The function has the code
```
if (getInfluence() < getCityStateInfluenceRestingPoint())
            return 0f
```
that should probably be changed to `<=` then you wouldn't need the additional conditions in this `when` statement since `getCityStateInfluenceDegrade()` would be 0 and therefore the notifications wouldn't appear.

There's also a similar function `getCityStateInfluenceRecovery()` that could have a similar modification but I don't think that's essential for this PR since that specific value isn't used for notifications.

oynqr: Thanks for pointing this out! Applied these changes and am currently testing them a little more.

OptimizedForDensity (approved):
Approved for real this time, nice job!

PR #7566, opened 2022-07-31
Invalid players removed from global quests

SimonCeder:
Solves #7526.
Rechecks each global quest at the start of the turn and removes invalid players, either due to them being dead or being at war with the assigning CS.


yairm210 (approved):

xlenstra:
Just a heads-up, as this is probably the second time I've seen this: we currently try to merge PR's if two collaborators who are not the author have approved of its contents. In this case the change is small and looks fine, but for bigger PR's this does matter more :)

PR #7568, opened 2022-08-01
Checking Korean.properties line-by-line

saud2410:
please don't merge this


saud2410:
Dunno for god's sake what is the problem...

xlenstra:
Problem is on our side; we've added the line "Our [attackerName] ([amount]) was destroyed by an intercepting [interceptorName] ([amount])", but this contains the parameter `amount` twice. If you replace the second `amount` in both the untranslated string and your translation it should work correctly.
Alternatively, you can wait until the next version to translate those lines, then we will have fixed this for all languages also.

PR #7579, opened 2022-08-02
Upload the World_Screen.png

xk730:
because it is Some old.....


Ouaz:
Good work, but I think the arrows are not necessary, as it makes things a bit cluttered. Circled numbers near the corresponding UI objects is enough.

xk730:
> Good work, but I think the arrows are not necessary, as it makes things a bit cluttered. Circled numbers near (or over) the corresponding UI objects is enough.

all right , i'll Try to improve the picture content

Ouaz:
You can keep the frame around UI objects, this is a good idea to make things "divided into parts" and easier to find, when going from the text to the screenshot (and vice-versa).

PR #7581, opened 2022-08-02
fix Inquisitors crash as missionaries were fixed

MindaugasRumsa51:


yairm210:
If these have the same 'get cheapest' functionality, just turn that into a proper function, `getCheapestUnit(Sequence<BaseUnit>)`

yairm210:
Since people are getting crashes from this, I'll merge it, but I still think it should be unified into a function

PR #7587, opened 2022-08-03
Add Damage numbers to Battle Notifications

itanasi:


xlenstra (approved):

yairm210:
Your translation strings don't take into account that only one of the sides may have taken damage, I assume you meant to add that on the defender side? Is there even such a thing as the defender taking zero damage? If there is then there are 4 combos for each battle situation so we'll need to think of something else

itanasi:
Right. To save on number of translations needed. If you deal no damage it will report 0. Obviously Ranged attacks only Attacker does damage. Or I even had the case of a suicide attack where they died before doing any damage.

But I don't think translations handle optional text, so we'd need to explode this out to 4 entries each, which seems excessive. So this was my compromise

yairm210 (approved):

PR #7588, opened 2022-08-03
Minor null safety

itanasi:


xlenstra (approved):

OptimizedForDensity:
Is this to stop the compiler warning for `maxBy`? Because it looks like `maxByOrNull { }!!` would crash in the same case that `maxBy` would crash, just with a different exception.

itanasi:
This is for the tool warning yes.
This would explicitly crash I believe.

OptimizedForDensity (approved):

PR #7577, opened 2022-08-02
Fix Air Sweep Notification Translations

itanasi:
Thanks to Mindaugas Rumša on Discord for pointing them out

PR #7591, opened 2022-08-03
Update template.properties

xk730:


xk730:
emm  think maybe that a little error，so...

xk730:
emm think maybe that a little error，so I modified it
 
”out“>>>”our“

xlenstra:
Duplicate of #7577

PR #7592, opened 2022-08-03
fix #7565: Delete one "Consumes ..." lines of city screen

### LINKED ISSUE #7565: Delete one "Consumes ..." lines of city screen
If a unit or building consumes resource, you can see `Consumes [amount] [resource]` line in the city screen. But if you don't have any of it, you can see the line in red AND white both...?

I think only 1 line is enough for noting it.

What if the line changes its color depending on the resorce?
- while i have resource, only white line is shown.
- while i have no resorce, only red line is shown.

Screenshot_20220731-162353_UnCiv

kasradzenika:
**Replaces error message**

Where: in the `CityConstructionsTable` 

from: `Consumes [amount] $resource`
to: `Need [amount] more $resource`


kasradzenika:
preview:

image

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -615,7 +616,7 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
 
         for ((resource, amount) in getResourceRequirements())
             if (civInfo.getCivResourcesByName()[resource]!! < amount) {
-                rejectionReasons.add(RejectionReason.ConsumesResources.toInstance(resource.getConsumesAmountString(amount)))
+                rejectionReasons.add(RejectionReason.ConsumesResources.toInstance(resource.getNeedMoreAmountString(amount)))
```
OptimizedForDensity, line 619: It looks like you need to change the argument of `getNeedMoreAmountString()` since `amount` is the number of resources you need for the building/unit but `civInfo.getCivResourcesByName()[resource]` is how much the player currently has. If you're trying to build a nuke that requires 2 uranium but only have 1 spare uranium, the text would say "Need 2 more uranium" when you only need 1 more.

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -419,7 +420,7 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
         if (!civInfo.isBarbarian()) { // Barbarians don't need resources
             for ((resource, amount) in getResourceRequirements())
                 if (civInfo.getCivResourcesByName()[resource]!! < amount) {
-                    rejectionReasons.add(RejectionReason.ConsumesResources.toInstance(resource.getConsumesAmountString(amount)))
+                    rejectionReasons.add(RejectionReason.ConsumesResources.toInstance(resource.getNeedMoreAmountString(amount)))
```
OptimizedForDensity, line 423: Same here.

core/src/com/unciv/models/ruleset/Building.kt
xlenstra, line 619: I'd make this `getNeedMoreAmountString(amount - civInfo.getCivResourcesByName()[resource]!!`, so you get the amount of resources you still need extra

OptimizedForDensity (commented):
Also, whenever you add a string that will be displayed on the interface, you'll need to add the same string in [template.properties](https://github.com/yairm210/Unciv/blob/master/android/assets/jsons/translations/template.properties). This way, the string will be added to all the other languages' translation files so translators can translate them. The line you'll need to add will be something like `Need [amount] more [resource] = ` (the names in the square brackets can be anything but something descriptive helps; also note the extra space after the `=`). The strings in this file are grouped together to help translators understand the context to the string since most probably won't be looking through code to see where it appears, so adding the string to the same section as `Consumes [amount] [resource] = ` is optional but helpful.

Overall this is a nice UI/UX change IMO.

core/src/com/unciv/ui/utils/extensions/FormattingExtensions.kt
```diff
@@ -19,6 +19,9 @@ fun Float.toPercent() = 1 + this/100
 /** Convert a [resource name][this] into "Consumes [amount] $resource" string (untranslated) */
 fun String.getConsumesAmountString(amount: Int) = "Consumes [$amount] [$this]"
 
+/** Convert a [resource name][this] into "Need [amount] more $resource" string (untranslated) */
+fun String.getNeedMoreAmountString(amount: Int) = "Need [$amount] more [$this]"
```
xlenstra, line 23: If you make a new string, also add it to the `template.properties` file, so that it can be translated into other langauges

alexban011:
A quick note in case you want to continue contributing:

It's good practice to create a different branch on your repo and work on that one instead of working directly on master. This way you can work on multiple things at once :) 

Also, if a certain PR doesn't get merged you'd have to manually revert the changes on your 'master' which is more time consuming than just deleting a branch.

And I bet there are even more reasons to create different branches for features but these are the ones that came to my mind rn

core/src/com/unciv/ui/utils/extensions/FormattingExtensions.kt
kasradzenika: got it

xlenstra (approved):

yairm210 (approved):

OptimizedForDensity (approved):

PR #7605, opened 2022-08-04
Added a unique for destroying improvements on attack

xlenstra:
As requested in #3242.


yairm210 (approved):
Looks ok, if we had a tile conditional I'd say use that, but we don't

OptimizedForDensity (approved):
Looks good, in the future it might be nice to also have a unique for artillery/bombers or similar units being able to bombard tiles without enemy units on them and destroy improvements (so like a strategic bombing unique to reduce enemy production/tile yield without moving a unit to the tile to pillage it).

PR #7606, opened 2022-08-04
Removes the option for disabling religion

xlenstra:
This PR removes the option in the new game screen to disable religion, but doesn't remove the underlying variable, so games started with this option disabled can still be played without religion.

This is a bit of a controversial change. For me personally the plan was always to remove this option once religion was fully implemented, but due to that having taking over a year, this option might now have become expected for most users.

Reasons for removing this button:
- Civ V also does this.
- When religion is disabled, the special ability of some civilizations just doesn't apply any more. For example, the Celts can't get Faith from forests any more, and Byzantium doesn't have any use for the extra belief when founding a religion. This makes the civilization weaker, and, if they're an AI player, might unexpectedly lower the difficulty of the game.
- It was never meant to be an option. It was first implemented as such when religion was in a very early state to only allow beta tests and warn players that the feature was in heavy development. Later the button marking it as beta was removed, but the option was kept, as it still wasn't fully finished. With #4290 now closed, religion is fully finished, and from this point of view the option is no longer necessary.

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -255,7 +255,6 @@ object GameStarter {
         availableCityStatesNames.addAll( ruleset.nations
             .filter {
                 it.value.isCityState() &&
-                (it.value.cityStateType != CityStateType.Religious || newGameParameters.religionEnabled) &&
                 !it.value.hasUnique(UniqueType.CityStateDeprecated)
             }.keys
             .shuffled()
--- a/core/src/com/unciv/logic/civilization/QuestManager.kt
+++ b/core/src/com/unciv/logic/civilization/QuestManager.kt
@@ -804,7 +804,7 @@ class QuestManager : IsPartOfGameInfoSerialization {
                             && civInfo.gameInfo.getCities().none { it.cityConstructions.isBuilt(building.name) }
                             // Can't be disabled
                             && building.name !in startingEra.startingObsoleteWonders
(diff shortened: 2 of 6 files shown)


yairm210 (approved):
Agree. With the division of Vanilla and G&K rulesets, religion vs no religion should be a ruleset thing, not a settings thing.

yairm210:
Also, please add a deprecation annotation on the "religion enabled" configuration that says that setting this variable was disabled in version 4.2.3, so when we go back to it in a couple of months we can see how old it is and delete it later

OptimizedForDensity (approved):

jmuchemb:
I've just open #8047 and I attached there a save that was played with a custom `Civ V - Gods & Kings - No Religion` ruleset. Because it's likely to happen again in the future, I'd like to share my view about this PR.

At first, I planned to ask to resurrect the option to disable religion. It's just than given my opinions about religion, it's not something that can be fun.

And playing vanilla rules is not a satisfying alternative because there are much more differences that religion.

Then I realized that the option to disable religion indeed had issues. A few changes to the ruleset would be required to keep it balanced and that's what I did. In the form of a small Python script that generate a custom ruleset from `Civ V - Vanilla` & `Civ V - Gods & Kings`.

Slightly related: it may be interesting to have an inheritance mechanism between ruleset to avoid so much duplication e.g. between vanilla & GK.

yairm210:
They are intentionally separated in order to keep each one self-contained, so if you want to create a mod based off of one so you need to do is copy and change
Inheritance between base ruleset mods sounds like a very problematic proposition

PR #7612, opened 2022-08-04
Add free belief unique + refactor a few religion functions

OptimizedForDensity:
A two-in-one commit that was almost big enough to split into two PRs. Consists of the following components
* Removes code duplication regarding counting the number of unused beliefs of different types as well as selecting beliefs for founding/enhancing a religion
* Adds safeguards for uniques adding additional beliefs on founding/enhancing a religion (i.e., make sure there are enough available)
* Refactors adding beliefs through creating pantheons, founding religion, and expanding religion
* Tweaked colors of found pantheon/found religion/enhance religion text in next turn button to reflect the color that the pantheon/founder/enhancer `BeliefType` uses for consistency
* Adds a triggerable unique (`OneTimeFreeBelief("Gain a free [beliefType] belief", UniqueTarget.Triggerable)`) for granting a civ extra beliefs (handles different cases including where a civ gets another belief after founding a religion [termed "Reform Religion"] or gets a second pantheon belief before forming a religion [termed "Expand Pantheon"])

While this is intended to give modders more flexibility in the short term where they can add a ruin reward for another pantheon or add techs/wonders/policies that grant an additional belief, this is also intended to be an easy way to incorporate BNW reformation beliefs when the time for the official Unciv BNW ruleset comes.


core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -84,27 +89,20 @@ class ReligionManager : IsPartOfGameInfoSerialization {
...
+                || (freeBeliefs[BeliefType.Pantheon] != null && freeBeliefs[BeliefType.Pantheon]!! > 0)
     }
 
-    fun isPickablePantheonBelief(belief: Belief): Boolean {
-        if (belief.type != BeliefType.Pantheon) return false
-        return (civInfo.gameInfo.civilizations.none { it.religionManager.religion?.followerBeliefs?.contains(belief.name) == true })
-    }
-
-    fun choosePantheonBelief(belief: Belief) {
-        storedFaith -= faithForPantheon()
```
xlenstra, line 100: This line is removed here, but I can't find where the faith is subtracted instead. The only other place this function is mentioned is inside an `if` statement checking if you're choosing free beliefs, which I'm pretty sure isn't what happens when you found a pantheon

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -84,27 +89,20 @@ class ReligionManager : IsPartOfGameInfoSerialization {
...
+    private fun choosePantheonBelief(belief: Belief) {
```
xlenstra, line 100: I'm not sure this function name still reflects its use, maybe change it to `foundPantheon`. In that case, you might as well change the `belief: Belief` parameter to `religionName: string`, as that is the only thing the `belief` actually used for in this function

core/src/com/unciv/ui/pickerscreens/PantheonPickerScreen.kt
```diff
@@ -29,7 +31,7 @@ class PantheonPickerScreen(
         }
 
         setOKAction("Choose a pantheon") {
-            choosePantheonBelief(selectedPantheon!!)
+            chooseBeliefs(null, null, listOf(selectedPantheon!!))
```
xlenstra, line 34: Seeing these use cases, it might be more readable if you re-order the paramteres of `chooseBeliefs` so the beliefs come first, and the icon name last. If you then give both the `religionName` and `iconName` a default of null, it makes these kinds of functions more readable imo

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -227,36 +226,53 @@ class ReligionManager : IsPartOfGameInfoSerialization {
...
+            for (belief in beliefs) {
+                if (freeBeliefs[belief.type] == null) continue
+                if (religionState == ReligionState.Pantheon && belief.type == BeliefType.Pantheon
+                        && religion!!.followerBeliefs.size == 1 && storedFaith >= faithForPantheon(-1)) {
+                    // paid for the initial pantheon using faith, otherwise this civ is using a free pantheon belief pick
+                    storedFaith -= faithForPantheon(-1) // this civ is included in this calculation now so offset for that
+                    continue
+                }
+                freeBeliefs[belief.type] = max(freeBeliefs[belief.type]!! - 1, 0)
+            }
```
xlenstra, line 264: I'm not entirely certain what's going on here. The decrementing of free beliefs when used is clear, but when exactly are you using the Faith cost? It looks like for that you already need to have a pantheon, and so also buy the second pantheon belief using Faith? Is that what you meant?

core/src/com/unciv/logic/civilization/ReligionManager.kt
```diff
@@ -152,6 +150,13 @@ class ReligionManager : IsPartOfGameInfoSerialization {
         }
     }
 
+    fun numberOfBeliefsAvailable(type: BeliefType): Int {
+        return civInfo.gameInfo.ruleSet.beliefs.values.count {
+            it.type == type
+                    && civInfo.gameInfo.religions.values.none { religion -> it in religion.getBeliefs(type) }
+        }
```
xlenstra, line 159: If you just want to raw number, it might even be faster to count the total number of beliefs of a certain type, and then subtract the number of beliefs each civ has of that type, but that's a minor optimization

core/src/com/unciv/ui/pickerscreens/PantheonPickerScreen.kt
```diff
@@ -17,7 +17,9 @@ class PantheonPickerScreen(
         for (belief in ruleset.beliefs.values) {
             if (belief.type != BeliefType.Pantheon) continue
             val beliefButton = getBeliefButton(belief, withTypeLabel = false)
-            if (choosingCiv.religionManager.isPickablePantheonBelief(belief)) {
+            if (gameInfo.civilizations.none {
+                        it.religionManager.religion?.followerBeliefs?.contains(belief.name) == true
+            }) {
```
xlenstra, line 22: This is in my opinion a bit of a loss in readability. Maybe replace it with a function `anyCivHasBelief(belief: Belief)` that can then also be used in `chooseBeliefOfType` in `NextTurnAutomation.kt` (which does basically the same computation)?

core/src/com/unciv/logic/civilization/ReligionManager.kt
OptimizedForDensity: The idea is that there's an edge case where the player can reach the amount of faith to form a pantheon on the same turn it builds a wonder/completes a tech that gives it a free pantheon pick and we want the player to have both pantheon belief picks. `chooseBeliefs()` is called first when the player/AI forms a pantheon and in this case we want to prioritize spending faith so the player can still choose the free belief. I do agree that these lines are confusing  as is, and with regards to your above comment, I've moved the faith decrease line back to `foundPantheon()` since we'll only be checking if the player has enough faith on the pantheon founding and not on any subsequent free pantheon picks.

xlenstra: Ah I see. Wouldn't it in that case be easier to check at the start of this function (before the `when` statement) if you don't have a religion yet and then set a boolean flag, which can be read in the if statement?

OptimizedForDensity: I'm doing something like that except giving `chooseBeliefs()` an optional boolean flag that indicates whether the player is using the free picks at this time since this case could also happen when a religion is formed/enhanced so just checking the faith amount alone isn't sufficient in the case where you're adding another Founder belief.

xlenstra (approved):

Issue #7610, opened 2022-08-04
Implementation of Espionage - Roadmap


xlenstra:
Similarly to #4290, this issue serves to be a roadmap for implementing espionage, the last major missing feature from G&K. I'm planning on working on this by myself, as I did with religion, but as I had to spend a large part of my holiday working on my thesis I might not have time to finish it before uni starts again, meaning this might develop slower than religion did. I'll put the steps I plan on taking below, which might change over the course of actually implementing espionage. Each of which will probably go in a different PR.

- [x] Create a hidden button to show a button in the game start menu to enable espionage - #7615
- [x] Unique for gaining spies, `civInfo.espionageManager` - #7641
- [x] Screen to send spies to cities - #7665 
- [x] Spies grant vision out of stationed cities - #7762
- [x] Spies take time to move and set up connections - #7776
- [x] Spies sometimes steal techs - #9629 
- [x] Allow rigging city-state elections
- [x] Deal with spies being in cities that are captured or destroyed - #9632 
- [x] Differing difficulty in infiltrating cities #11401
- [x] Buildings, wonders and that policy reducing spy effectiveness #11401
- [x] UI for showing spy effectiveness
- [x] Having a spy in a city may let you see its completed buildings and yields #11396
- [x] Counter-intelligence, spy death & revival #11077
- [x] Spies level up #11077
- [x] Diplomatic effects of having your spies spotted
- [x] City-state coups
- [x] AI for moving spies #10974
- [x] Add civilopedia entry
- [x] Make the hidden button show in the options menu
- [x] Remove the hidden button, but keep the option

yairm210:
I'm sure you don't need me to tell you to focus on your personal life first, since you're already doing so, but I will say (as someone who put way too much time into this) that at the end of the day this is just a game ;)

Good luck with your studies!!

Overfloo1710:
Cuando agregaras minas terrestres y otros tipos de trampas al juego soy alguien que lo juega muy seguido pero ya las partidas se me vuelven muy monótonas agregar cosas como trampas ayudarían a mejorar la experiencia de juego

AesirRaider:
Translation for the comment above this one:

When you add landmines and other types of traps to the game, I am someone who plays it very often but the games are already becoming very monotonous adding things like traps would help improve the gaming experience

AesirRaider:
Is there gonna be a spy agency when espionage is implemented into the game, and, if there is, can you name the agencies you created?

PR #7615, opened 2022-08-05
Added a hidden option for enabling espionage

xlenstra:
This PR adds an option in the debug tab which sets whether or not the 'enable espionage' option is shown when starting a new game. This game option currently does nothing, see #7610.


OptimizedForDensity (approved):
Normally we'd need to change the `GameInfo.CURRENT_COMPATIBILITY_NUMBER` but since this has no features that players should be using in game this is ok for now.

PR #7631, opened 2022-08-06
List which city owns each tile in the city screen UI

OptimizedForDensity:
Adds a way for players to know which city owns each tile. This appears in the bottom right corner of the city screen when a player selects a tile that has an owning city.
TileOwner


AdityaMH:
My notifications would be blocked more by this tile info.
More unreadable and unaccessible notifications.

OptimizedForDensity:
The popup I took a screenshot of only appears when the player is in the city screen so this info won't be blocking notifications when you're on the world screen.

PR #7641, opened 2022-08-09
Basic version of `EspionageManger`, added a unique to gain spies

xlenstra:
This PR adds basic classes for espionage: `EspionageManager` and `Spy`. These will later be extended a lot to include most of the code for espionage. It also adds a triggerable era unique that gives a spy to each player whenever any player enters that era for the first time. See also: #7610.

Edit: Spy names have been added, so this is ready to be reviewed and merged


android/assets/lasterror.txt
```diff
@@ -0,0 +1,20 @@
+java.lang.NullPointerException
```
OptimizedForDensity, line 1: File added by mistake?

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -688,7 +682,8 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     OneTimeRevealSpecificMapTiles("Reveal up to [amount/'all'] [tileFilter] within a [amount] tile radius", UniqueTarget.Ruins),
     OneTimeRevealCrudeMap("From a randomly chosen tile [amount] tiles away from the ruins, reveal tiles up to [amount] tiles away with [amount]% chance", UniqueTarget.Ruins),
     OneTimeTriggerVoting("Triggers voting for the Diplomatic Victory", UniqueTarget.Triggerable),  // used in Building
-    OneTimeGlobalAlert("Triggers the following global alert: [comment]", UniqueTarget.Triggerable), // used in Policy
+    OneTimeGlobalAlert("Triggers the following global alert: [comment]", UniqueTarget.Policy), // used in Policy
+    OneTimeGlobalSpiesWhenEnteringEra("Every major Civilization gains a spy when entering this era", UniqueTarget.Era),
```
OptimizedForDensity, line 686: It might be worth rewording to something like `"Every major Civilization gains a spy as soon as the first civilization enters this era"` since otherwise it sounds like each civ gains a spy as soon as _they_ enter the era rather than what actually happens in code which is all civs gain a spy.

core/src/com/unciv/logic/civilization/TechManager.kt
```diff
@@ -334,7 +318,23 @@ class TechManager : IsPartOfGameInfoSerialization {
...
         }
 
+        val previousEra = civInfo.getEra()
         updateEra()
+        val currentEra = civInfo.getEra()
+        if (previousEra != currentEra) {
+            civInfo.addNotification("You have entered the [$currentEra]!", NotificationIcon.Science)
+            if (civInfo.isMajorCiv()) {
+                for (knownCiv in civInfo.getKnownCivs()) {
+                    knownCiv.addNotification("[${civInfo.civName}] has entered the [$currentEra]!", civInfo.civName, NotificationIcon.Science)
```
OptimizedForDensity, line 328: Looking at this makes me realize that it might be worth adding an "Extra Notifications" option to the game settings where players can opt to get a lot more notifications with extra information like this. I think a more avid Unciv player in a multiplayer game might want to know when this happens but for a casual player this is probably irrelevant (other than for effects that happen once a player does something like advancing to an era where all civs gain a spy since that obviously has an effect on the player).

android/assets/jsons/translations/template.properties
```diff
@@ -811,6 +811,8 @@ Your [ourUnit] captured an enemy [theirUnit]! =
 Your [ourUnit] plundered [amount] [Stat] from [theirUnit] = 
 We have captured a barbarian encampment and recovered [goldAmount] gold! = 
 An enemy [unitType] has joined us! = 
+After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
+We have recruited [spyName] as a spy! =
```
OptimizedForDensity, line 815: Is it worth adding these strings to translation files so early? The espionage option is hidden so only devs would be able to access it but translators would have these strings appear in their files and might feel compelled to translate them (plus since this feature is so experimental these strings might change which makes constantly retranslating them annoying if that happens).

xlenstra: The way I go about this, feature for feature, means that these strings should in theory not change anymore. Additionally, if I don't add them right now I feel like its likely I'll forget to add them later😅

core/src/com/unciv/logic/civilization/TechManager.kt
xlenstra: In this case I've just moved existing code to a different spot so that it actually works again. Additionally this is a notification that is shown in the original, even for unknown civs. It also doesn't happen too often anyway.
In general that might be a good idea though

android/assets/jsons/translations/template.properties
OptimizedForDensity: Yeah that works, looking at these strings made me realize that some translators would see them and might get confused.

core/src/com/unciv/logic/civilization/EspionageManager.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun clone(): Spy {
+        return Spy(name)
+    }
+}
+
+class EspionageManager : IsPartOfGameInfoSerialization {
+
+    var spyCount = 0
+    var spyList = mutableListOf<Spy>()
+    var erasSpyEarnedFor = mutableListOf<String>()
```
OptimizedForDensity, line 15: I feel like this should be a `by lazy` method that looks at various uniques (e.g., English get 2 spies initially compared with the 1 every other civ gets). This change could come later though once more of the espionage system is in place.

android/assets/jsons/translations/template.properties
heipizhu4: Why not add them in annotations? Or no poor translator can achieve 100% translation in a period of time any more [cry]

OptimizedForDensity (approved):

Issue #7178, opened 2022-06-16
Android app hangs on smiley face screen


SomeMichael:
**Platform**
Android 12, Samsung A52 5G

**Version**
4.1.8

**Describe the bug**
Similar behavior as in #2552, but :

**To Reproduce**
1. Open the game
2. Play the game
3. Exit the game
4. Open the game again
Smiley face never vanishes

**Expected behavior**
Game starts

**Additional context**
After forcefully stopping the app in app settings all works fine again

yairm210:
This doesn't give enough context to actually solve the problem :/\

How long between when you close and enter the game? Android differentiates between "kill app" and "pause app"
Does this happen if there are no other apps open? (maybe it's a memory problem?_
Does this happen all the time, sometimes, or was this a one-off?

SomeMichael:
Thanks for the fast response! Here are the requested details

> How long between when you close and enter the game? Android differentiates between "kill app" and "pause app"

It happens immediatly as well as hours later.

> Does this happen if there are no other apps open? (maybe it's a memory problem?

It always happens. 100% reproducible. Just checked it also with all other apps closed
> Does this happen all the time, sometimes, or was this a one-off?

All the time

7eter:
happens for me too sometimes. However after closing the app from the opend app overview and reopening it - everything works fine again.

tortuosit:
Same here. Android 10, Mate 20X. Unciv Portrait mode. Can also close it via Android, start again, then it runs.

tortuosit:
It most of the time happens. And the 2nd try always works (like reported by 7eter)

yairm210:
Hopefully resolved in 4.1.18 :)

AdityaMH:
This only happen, if file from folder com.unciv.app was changed, or deleted.
To solve just restart unciv again.

tortuosit:
Not resolved in 4.1.18 unfortunately.
Started Unciv -> Exit via Android "Back". "Do you want to exit the game" "Exit"
Start Unciv again. Hangs.

tortuosit:
> Hopefully resolved in 4.1.18 :)

Not resolved in 4.1.18 unfortunately.
- Started Unciv -> Exit via Android "Back". "Do you want to exit the game" "Exit"
- Start Unciv again -> Hangs.

SomeMichael:
Sorry, but I have to confirm that.
This bug seams to be a a really nasty one …

----------------------------------------

[...]

PR #7655, opened 2022-08-13
close app completely when clicked 'Exit' button

### LINKED ISSUE #7515: Apps cannot be started in some case
**Platform**
Android 10

**Version**
Github release 4.1.20 (build 739)

**Describe the bug**
You can't start a game

**To Reproduce**
1. Launch a app
2. Create new game
3. Press back button twice and tap Exit button
4. Try to relaunch app
5. You will see initial screen forever

```

```
</details>

**Expected behavior**

**Screenshots**

https://user-images.githubusercontent.com/97716654/180582636-4fbcf3ea-919b-4c03-9d40-8dd1edec48d8.mp4

I'd attach video that reproduce this bug

**Additional context**

CrispyXYZ:
Fixes #7515 . I tested it on my Android phone and the bug recurred. This bug seems to be caused by that the 'Exit' button can not close Unciv completely on Android. call `this.finishAndRemoveTask()` method in overridden `onDetroy()` can fix this problem.

*See also [finishAndRemoveTask()](https://developer.android.com/reference/android/app/Activity.html#finishAndRemoveTask()), [onDestroy()](https://developer.android.com/reference/android/app/Activity#onDestroy())*

### DIFF
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -294,7 +294,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
                 restoreDefault = { musicController.resume() },
                 action = {
                     Gdx.app.exit()
-                    System.exit(0)
+
                 }
             ).open(force = true)
             return null
@@ -399,6 +399,8 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
 
         // On desktop this should only be this one and "DestroyJavaVM"
         logRunningThreads()
+
+        System.exit(0)
     }
 
     private fun logRunningThreads() {
--- a/android/src/com/unciv/app/AndroidLauncher.kt
+++ b/android/src/com/unciv/app/AndroidLauncher.kt
@@ -147,8 +147,8 @@ open class AndroidLauncher : AndroidApplication() {
     }
 
     override fun onDestroy() {
-        super.onDestroy()
-        finishAndRemoveTask();
+        finishAndRemoveTask()
+        if(false) super.onDestroy() // do not remove
     }
 
     override fun onNewIntent(intent: Intent?) {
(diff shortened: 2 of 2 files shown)


CrispyXYZ:
Also fixes #7178

CrispyXYZ:
It was too strange that the bug recurred when i called `super.onDestroy()`.

OptimizedForDensity (approved):
My brain shut off for a second there with the label. This seems to work but I keep getting a "App keeps crashing" pop up but I can't tell if that's from me constantly trying to close something out or if there's a different bug happening.

CrispyXYZ:
Get it! This problem is similar to https://stackoverflow.com/questions/34368616/how-to-exit-a-lib-gdx-application-safely

CrispyXYZ:
I know what i did before. I made Unciv crash when exiting so it could exit completely. Now uses `System.exit(0)` and it'll exit in the right way.

xlenstra:
I don't know enough about the gdx app life cycle to either approve or disapprove this, but have you tested that this doesn't also stop the multiplayer turn checker thread? Even if we completely close the game, that should still always keep running

OptimizedForDensity:
I looked a little more into it and I think what's happening is that the "LoadJSON" job in UncivGame.kt is getting cancelled immediately when the player reloads the app without fully shutting down, and since the main menu is called from within this job, the screen never changes. I don't know what's causing the cancellation but I think fixing this would be a better approach.

core/src/com/unciv/UncivGame.kt
```diff
@@ -294,7 +294,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
                 restoreDefault = { musicController.resume() },
                 action = {
                     Gdx.app.exit()
-
+                    System.exit(0)
```
JackRainy, line 297: Android Studio suggests to replace this line with `exitProcess(0)`.
P.S> I remember I have spent a good potion of time to overcome this bug, but not remember why I've given up.

CrispyXYZ: I have a new plan. call `System.exit(0)` at the end of `dispose()`.

PR #7658, opened 2022-08-13
Removed legacy tileset code

GGGuenni:
This removes a lot of code currently only used by the "Default" tileset. Before I begin, I want to make a proposal to rename the "Default" tileset to "Minimal" or something similar since this tileset is by far no longer our default tileset 😄 

To compensate for the now missing special treatment the tileset had before, I had to add two new TilesetConfig settings. `useSummaryImages `and `tileScales`.

If UseSummaryImages is true, it currently just uses the `NaturalWonder.png` instead of any other natural wonder specific image. I used a general name for it in case someone wants to add more "summary images" to the tileset. This is a more or less direct replacement for `updateNaturalWonderImage`.

TileScales is a new dictionary added to the tileset config to specify the size of specific images since the BaseTerrain hexagon must not be resized for "Default" to work which makes the tileScale setting not usable.

In addition, I finally rewrote the way the useColorAsBaseTerrain setting works. It now adds a hexagon image to every TileGroup no matter what and only applies the base terrain color to that hexagon instead of applying the color to the whole TileGroup. This was the original intention anyway. 
I've also changed the default value of the useColorAsBaseTerrain setting to `false` as it isn't used by any other tileset anyway.


core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -299,17 +301,17 @@ open class TileGroup(
         }
         tileImageIdentifiers = tileBaseImageLocations
 
-        for (image in tileBaseImages.asReversed()) image.remove()
+        for (image in tileBaseImages) image.remove()
         tileBaseImages.clear()
-        for (baseLocation in tileBaseImageLocations.asReversed()) { // reversed because we send each one to back
```
GGGuenni, line 304: I've removed both reverse calls as they were just unnecessary.
The first one: why reverse a list to remove every image from it?
The second one: Instead of reversing the list and adding everything at the front (line 327 (removed)), everything is just added in order now

yairm210 (approved):
Sure, the rename is also a good idea. This tileset hasn't really been touched in years so it's "anything good" add far as I'm concerned

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -241,7 +230,14 @@ open class TileGroup(
 
     private fun getTileBaseImageLocations(viewingCiv: CivilizationInfo?): List<String> {
         if (viewingCiv == null && !showEntireMap) return tileSetStrings.hexagonList
-        if (tileInfo.naturalWonder != null) return listOf(tileSetStrings.orFallback { getTile(tileInfo.naturalWonder!!) })
+
+        val out = if (tileSetStrings.tileSetConfig.useColorAsBaseTerrain)
```
yairm210, line 234: Please rename this val

GGGuenni:
Renaming the tileset only works sort of. If no "Default" tileset is available the options menu creates a "Default" select option anyway which is basically just "FantasyHex". 

Default

Sadly, I couldn't find the code that causes this. 
I did find however that the game defaults to "FantasyHex" if a tileset is no longer available. This should probably be changed to "HexaRealm" or a `DEFAULT_TILESET` constant

UncivGame.kt line 129
```
// If one of the tilesets is no longer available, default back
if (settings.tileSet !in ImageGetter.getAvailableTilesets()) { 
    settings.tileSet = "FantasyHex"
}
```

docs/Modders/Creating-a-custom-tileset.md
```diff
@@ -41,9 +48,19 @@ This is an example of such a config file that I will explain below:
 
 ### useColorAsBaseTerrain
 
-A boolean value ("true" or "false"). Default value: "true"
+A boolean value ("true" or "false"). Default value: "false"
+
+If true, an additional "Hexagon" image is placed below each tile and colored in the corresponding BaseTerrain color. This removes the necessity to add individual BaseTerrain images. This is how the "Default" tileset works.
```
xlenstra, line 53: This still mentions the now renamed 'default' tileset

docs/Modders/Creating-a-custom-tileset.md
```diff
@@ -57,9 +74,27 @@ A color defined with normalized RGBA values. Default value: "{"r":0, "g":0, "b":
 
 Defines the color of the fog of war. The color gets approximated by 60% to allow the colors of the images below to shine through.
 
+### fallbackTileSet
+
+A string value. Default value: "FantasyHex"
```
xlenstra, line 79: This default value was also changed to HexaRealm, right?

docs/Other/Map-related-JSON-files.md
```diff
@@ -119,12 +119,16 @@ A few uniques can be added to ancient ruin effects to modify when they can be ea
...
 | --------- | ---- | -------- | ----- |
-| useColorAsBaseTerrain | Boolean | Default true | ? *WIP* |
-| unexploredTileColor | Color | Default Dark Gray | `{"r":0.25,"g":0.25,"b":0.25,"a":1}` |
-| fogOfWarColor | Color | Default Black | `{"r":0,"g":0,"b":0,"a":1}` |
-| ruleVariants | List | Default empty | see below |
+| [useColorAsBaseTerrain](../Modders/Creating-a-custom-tileset.md#useColorAsBaseTerrain) | Boolean | false | |
+| [useSummaryImages](../Modders/Creating-a-custom-tileset.md#useSummaryImages) | Boolean | false | |
+| [unexploredTileColor](../Modders/Creating-a-custom-tileset.md#unexploredTileColor) | Color | Dark Gray | `{"r":0.25,"g":0.25,"b":0.25,"a":1}` |
+| [fogOfWarColor](../Modders/Creating-a-custom-tileset.md#fogOfWarColor) | Color | Black | `{"r":0,"g":0,"b":0,"a":1}` |
+| [fallbackTileSet](../Modders/Creating-a-custom-tileset.md#fallbackTileSet) | String | "FantasyHex" | null to disable |
```
xlenstra, line 128: Same here

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -649,35 +578,22 @@ open class TileGroup(
...
-                terrainFeatureOverlayImage!!.run {
-                    setSize(30f, 30f)
-                    setColor(1f, 1f, 1f, 0.5f)
-                    center(this@TileGroup)
-                }
-            }
+        val color = when {
+            !isViewable -> Color.WHITE.cpy().lerp(tileSetStrings.tileSetConfig.fogOfWarColor, 0.6f)
+            else -> Color.WHITE.cpy()
         }
```
xlenstra, line 591: when with two cases can just be an if

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -321,18 +321,17 @@ open class TileGroup(
...
             tileBaseImages.add(image)
             baseLayerGroup.addActor(image)
-            setHexagonImageSize(image)
+
+            if (tileSetStrings.tileSetConfig.tileScales.isNotEmpty()) {
+                val scale = tileSetStrings.tileSetConfig.tileScales[tileBaseImageLocations[index].takeLastWhile { it != '/' }]
+                setHexagonImageSize(image, scale)
+            } else {
+                setHexagonImageSize(image)
+            }
```
xlenstra, line 334: Technically you can write this as
```kotlin
val scale = tileSetStrings.tileSetConfig.tileScales[tileBaseImageLocations[index].takeLastWhile { it != '/' }]
setHexagonImageSize(image, scale)
```
So you don't need the if statement, but I'm not sure it improves readability enough to merit this change.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -649,35 +578,22 @@ open class TileGroup(
...
-                terrainFeatureLayerGroup.addActor(terrainFeatureOverlayImage)
-                terrainFeatureOverlayImage!!.run {
-                    setSize(30f, 30f)
-                    setColor(1f, 1f, 1f, 0.5f)
-                    center(this@TileGroup)
-                }
-            }
+        val color = when {
+            !isViewable -> Color.WHITE.cpy().lerp(tileSetStrings.tileSetConfig.fogOfWarColor, 0.6f)
+            else -> Color.WHITE.cpy()
```
xlenstra, line 590: I'm not entirely sure what happens here, doesn't this result in a change in behaviour, where all tiles not currently shown tiles are grey instead of using a greyed out version the base terrain colour?

xlenstra:
When I checked out this branch locally, I couldn't find the 'Minimal' tileset in the options menu tileset picker, only the 'Default' tilese. Selecting this renders everything in FantasyHex

docs/Modders/Creating-a-custom-tileset.md
GGGuenni: I haven't changed the default value for fallbackTileset (TilesetConfig.kt L11) because "Hexarealm" does not contain Unit sprites yet and uses "FantasyHex" as fallbackTileset itself.

GGGuenni:
> When I checked out this branch locally, I couldn't find the 'Minimal' tileset in the options menu tileset picker, only the 'Default' tilese. Selecting this renders everything in FantasyHex

This is because I didn't commit the game.atlas and game.png files to avoid merge conflicts.
Is it still necessary to commit those or do they get regenerated on release?

core/src/com/unciv/ui/tilegroups/TileGroup.kt
GGGuenni: I did change how "useColorAsBaseTerrain" works. It now only colors a base hexagon instead of everything on a tile. That's why I've split the local color variable in the updateTileColor function into "baseTerrainColor" and "color". "baseTerrainColor" is applied only to the first tileBaseImage which if "useColorAsBaseTerrain" is true is a default Hexagon.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
GGGuenni: I wanted to avoid that takeLastWhile is called for every Image (location) even though it's unnecessary

xlenstra:
> > When I checked out this branch locally, I couldn't find the 'Minimal' tileset in the options menu tileset picker, only the 'Default' tilese. Selecting this renders everything in FantasyHex
> 
> This is because I didn't commit the game.atlas and game.png files to avoid merge conflicts. Is it still necessary to commit those or do they get regenerated on release?

I don't think they're automatically regenerated on release (or sometimes even on build), so this might be why it didn't show up. I'll delete those and see if it works then.

xlenstra (approved):

PR #7659, opened 2022-08-13
Ukrainian translations

JackRainy:


android/assets/jsons/translations/Ukrainian.properties
```diff
@@ -810,23 +810,23 @@ A(n) [nukeType] exploded in our territory! = [nukeType] вибухнула на
...
 We have discovered the lost technology of [techName] in the ruins! = У руїнах знайдено загублену технологію [techName]!
 A [unitName] has joined us! = До нас приєднався підрозділ [unitName]!
-An ancient tribe trains our [unitName] in their ways of combat! = Давнє племʼя навчило їхнього способу битви наш підрозділ [unitName]!
+An ancient tribe trains our [unitName] in their ways of combat! = Давнє племʼя навчило їхнього способу битви наш підрозділ «[unitName]»!
 We have found a stash of [amount] gold in the ruins! = У руїнах знайдено скарб золота кількістю [amount]!
 We have found a crudely-drawn map in the ruins! = Ми знайшли в руїнах грубо намальовану мапу!
-[unit] finished exploring. = Підрозділ [unit] завершив розвідку.
-[unit] has no work to do. = Підрозділ [unit] без роботи.
+[unit] finished exploring. = Підрозділ «[unit]» завершив розвідку.
+[unit] has no work to do. = [unit] без роботи.
```
estorski, line 829: Why no quotes here, and inconsistent translation even comparing to the line above?

JackRainy: The primary problem of units: the gender of the unit can be different. 
In the first case we resolve this by referencing it as always-male "підрозділ завершив" and the name/type of the unit goes in quotes. 
In the second case there is no need to resolve that: any gender will be suitable since we have no verb there. For example, Робітник без роботи, Робітниця без роботи, Бо-зна-що без роботи.

estorski: Got it, thanks

PR #7660, opened 2022-08-13
Generalize Great Wall unique

OptimizedForDensity:
Greatly increases flexibility of this unique since even the obsolete tech was hardcoded.
* Change Great Wall unique from `Enemy land units must spend 1 extra movement point when inside your territory (obsolete upon Dynamite)` to `Enemy [Land] units must spend [1] extra movement points when inside your territory <before discovering [Dynamite]>`
* Rename `CivilizationInfo.hasActiveGreatWall` to a more general `CivilizationInfo.hasActiveEnemyMovementPenalty` to help accommodate any future uniques with similar effects
* Rename a few other functions accordingly

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -33,6 +33,7 @@ import com.unciv.models.ruleset.tile.ResourceType
 import com.unciv.models.ruleset.tile.TileResource
 import com.unciv.models.ruleset.unique.StateForConditionals
 import com.unciv.models.ruleset.unique.TemporaryUnique
+import com.unciv.models.ruleset.unique.Unique
 import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.stats.Stat
@@ -106,6 +107,10 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
     /** This is for performance since every movement calculation depends on this, see MapUnit comment */
     @Transient
     var hasActiveEnemyMovementPenalty = false
+
+    /** Same as above variable */
+    @Transient
+    var enemyMovementPenaltyUniques: Sequence<Unique>? = null
 
     @Transient
     var statsForNextTurn = Stats()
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -671,6 +671,18 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
         return diplomacyManager.hasOpenBorders
     }
 
+    fun getEnemyMovementPenalty(enemyUnit: MapUnit): Float {
+        val greatWallUnique = getMatchingUniques(UniqueType.EnemyLandUnitsSpendExtraMovement)
```
xlenstra, line 675: Don't forget to include the `stateForConditionals` as the second parameter, otherwise conditionals cannot be used for this unique

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -34,10 +34,9 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
         val toOwner = to.getOwner()
         val extraCost = if (
             toOwner != null &&
-            to.isLand &&
-            toOwner.hasActiveGreatWall &&
+            toOwner.hasActiveEnemyMovementPenalty &&
             civInfo.isAtWarWith(toOwner)
-        ) 1f else 0f
+        ) toOwner.getEnemyMovementPenalty(unit) else 0f
```
xlenstra, line 39: This calls an additional function from `getMovementCostBetweenAdjacentTiles`, which (as per the comment above the function) should be as time-optimal as possible. Have you measured the impact this change has on a game, both with and without a modded wonder with the new unique?

OptimizedForDensity: I haven't measured it yet but I kept the condition of this `if` statement as close to the original as I could which should hopefully keep performance similar to before without jeopardizing modability. `CivilizationInfo.hasActiveEnemyMovementPenalty` is set once each turn when `GameInfo.setTransients()` is called (or if a civ builds a building that affects `CivilizationInfo.hasActiveEnemyMovementPenalty`). This boolean variable is used in the condition of this `if` statement 2 lines above so we can know at a glance if we should even bother calculating if there is a movement penalty. The idea being that 1 calculation per turn to set a boolean variable for each `civInfo` then checking this boolean variable every time we try to find a path should keep everything as time efficient as possible.

xlenstra: I agree that the case where the condition of the `if` is false is well-optimized, but I'm a bit more worried about the case where it is true. In this case we call another function, which then calls `getMatchingUniques`, which feels like a very costly function

OptimizedForDensity: That's a good point. I've changed it so that there's a new variable that stores the movement penalty uniques in a `Sequence` so we only have to iterate over the sequence rather than call `getMatchingUniques` for every tile that a unit tries to travel to. This cache is updated every time the game updates movement penalty uniques which is at most twice per turn in a base Unciv game (start of turn then if the civ builds the Great Wall) rather than calling `getMatchingUniques` several hundred times each AI turn potentially.

yairm210 (approved):
All in all looks good, can't wait to deprecate the hell out of the old version

PR #7665, opened 2022-08-14
Added a screen to move spies between cities

xlenstra:
This adds a basic screen which gives an overview of current spies and allows players to move them to other cities. Conform the original, this button is placed on the world map, though an argument could be made to instead place it in the overview screens.

<details><summary> Video showcase </summary>

https://user-images.githubusercontent.com/71121390/184554716-50ac2781-c872-49b2-ac1d-60d9e94fb83a.mp4

If the video doesn't work, you can also watch it here: https://youtu.be/Hfo3yiBXpkY

</details>

Currently, the player is only allowed to send spies to cities of major civilizations that don't already have a spy of that civilization. Cities of the civ itself aren't allowed yet as that requires counterintelligence, and cities of city states require rigging city state elections, both of which are not currently planned. As the next thing I'm going to work on is the tech stealing from major civs, this made the most sense.


core/src/com/unciv/ui/worldscreen/TechPolicyDiplomacyButtons.kt
```diff
@@ -56,12 +60,20 @@ class TechPolicyDiplomacyButtons(val worldScreen: WorldScreen) : Table(BaseScree
         diplomacyButtonHolder.onClick {
             game.pushScreen(DiplomacyScreen(viewingCiv))
         }
+        if (game.gameInfo!!.isEspionageEnabled()) {
+            espionageButton.add(ImageGetter.getImage("OtherIcons/Spy_White")).size(30f).pad(15f)
+            espionageButtonHolder.onClick {
+                game.pushScreen(EspionageOverviewScreen(viewingCiv))
+            }
+        }
```
JackRainy, line 68: I have enabled the Espionage in the Debug menu, started a new game with enabled the Espionage option, `game.gameInfo!!.isEspionageEnabled()` is True here, but I still do not see the button.
Where should it be?

xlenstra: It shows up once you have a spy, which happens when anyone enters the Renaissance era, should be next to the diplomacy and policy buttons

JackRainy: Never mind, I've found out that it appears when you have at least 1 spy, and to get it you should enter the Renessance Era by a natural way. I was creating a game in the Informational Era from the very beginning.

core/src/com/unciv/ui/overviewscreen/EspionageOverviewScreen.kt
```diff
@@ -0,0 +1,166 @@
...
+        updateCityList()
+    }
+
+    private fun updateSpyList() {
+        spySelectionTable.clear()
+        spySelectionTable.add("Spy".toLabel()).pad(5f)
+        spySelectionTable.add("Location".toLabel()).pad(5f).row()
+        for (spy in civInfo.espionageManager.spyList) {
+            spySelectionTable.add(spy.name.toLabel()).pad(5f)
+            spySelectionTable.add(spy.getLocationName(civInfo.gameInfo)).pad(5f)
```
JackRainy, line 66: These two columns need to be localized. Didn't you miss `.tr()` here?

core/src/com/unciv/ui/overviewscreen/EspionageOverviewScreen.kt
```diff
@@ -0,0 +1,166 @@
...
+                moveSpyButton.disable()
+                for ((button, city) in moveSpyHereButtons)
+                    // For now, only allow spies to be send to cities of other major civs
+                    // Not own cities as counterintelligence isn't implemented
+                    // Not city-state civs as rigging elections isn't implemented
+                    // Technically, stealing techs from other civs also isn't implemented, but its the first thing I'll add so this makes the most sense to allow.
+                    if (city == null || ( // hideout
+                        city.civInfo.isMajorCiv()
+                        && city.civInfo != civInfo
+                        && (!city.espionage.hasSpyOf(civInfo) || selectedSpy!!.location == city.id)
```
JackRainy, line 82: `(!city.espionage.hasSpyOf(civInfo) || selectedSpy!!.location == city.id)` - it is always true, because `city.espionage.hasSpyOf(civInfo)` does the same check `selectedSpy!!.location == city.id`.
I am not sure `||` is placed correctly here.
Also, `city == null` allows to move a spy to the hideout even if it is already there. Is it expected behaviour?

xlenstra: Not exactly, `!city.espionage.hasSpyOf(civInfo)` means the spy can move to any city not containing a spy, or `selectedSpy!!.location == city.id` to the location the spy is currently stationed (if you don't want to move it even though you selected it)

Spies should always be able to move to the hideout, moving it while its already there effectively cancels the movement

JackRainy: Ok, then I misunderstand the UI.
I would suggest cancelling the movement by pressing the same button "Move" which is highlighted (or renamed to "Cancel") until the "<=" button is pressed.
The possibility to move the spy to the location where it is already present seems to me like a nonsense.

xlenstra:
I've made the requested changes, but for some reasons when I hover over the cancel button _all_ labels on the page get darkened. I'm not sure what the right way is to fix this

JackRainy (approved):
> I've made the requested changes, but for some reasons when I hover over the cancel button _all_ labels on the page get darkened. I'm not sure what the right way is to fix this

Thank you. The fix for the issue is quite simple. You do not need to reassign the label, but use `setText` instead.
Replace, `selectedSpyButton!!.label = "Cancel".toLabel()` with `selectedSpyButton!!.label.setText("Cancel".tr())` and for "Move" (line 165) correspondingly.

Also, I`ve noticed that it is still possible to send a spy to the Hideout even if it is already there. All the rest are good.

P.S> I have a few ideas how to improve/rework the screen further, but probably, I will suggest them in other PRs later at some point.

xlenstra:
> P.S> I have a few ideas how to improve/rework the screen further, but probably, I will suggest them in other PRs later at some point.

By all means feel free to update this screen yourself, UI design definitely isn't one of my strong points.

Issue #7600, opened 2022-08-04
Feature request: Improved option to rename the unit


pyotr71:
Unciv has option to change name of Units, but its active ONLY when Unit achieving promotion:

![promocja](https://user-images.githubusercontent.com/18636457/182833404-5158a4ea-f09e-4f85-9cae-413adbe3cf53.png)

I think it's good idea to add this option permanently.
Where to put the option?
1. On the Overview -> Units screen

OR (maybe better)

2. On the Promotion screen (we get there after clicking promotion symbol)

![units](https://user-images.githubusercontent.com/18636457/182833602-76243bd8-6a90-4a14-8023-7907890e6901.png)

Caballero-Arepa:
I super-agree with this. It has been asked *a lot* on the discord server, and I dont see why not implementing it.
It's such a great qol feature.

huckdogg:
I have a working implementation of the rename feature for all units, even when they have no active promotions and no promotion available.

I changed the unit info box in the main map screen (below) so that clicking the unit name (red box) will take you to the promotion/rename screen. I did this because units with no promotions do not have a link to the promotion/rename page. Clicking the unit icon (yellow box) retains the old behavior of linking to wiki.
![ScreenShot_20220815_1660605113](https://user-images.githubusercontent.com/876926/184736188-6c65c07e-7719-45f6-ad34-21fac2bbbd60.png)

I have also changed the Overview screen for units to have an Edit column with a button that opens the promotion/rename screen.
![ScreenShot_20220815_1660604589](https://user-images.githubusercontent.com/876926/184738118-4cb583e5-e03c-470d-ba4e-e7f64eb9a0c0.png)

However, when the unit name is changed and the promotion screen is closed, the Overview page does not refresh to reflect the unit name change. If anyone has a suggestion on how to refresh the overview, I would appreciate it.

OptimizedForDensity:
[...]

PR #7676, opened 2022-08-18
Improved options to rename units

huckdogg:
This update improves the options to rename units. It allows naming units that have no promotions or cannot be promoted. Unit renames are now allowed, as well as clearing the custom name completely. Units can be renamed from several different game screens (PromotionPickerScreen, WorldScreen from UnitTable, and OverviewScreen for UnitOverviewTable). Each screen has a clickable object that will call UnitRenamePopup, which opens an AskTextPopup for user input, renames the unit, and runs code provided through the lambda by the calling screen.

Resolving Issue #7600

**WorldScreen.kt**
- Set KeyboardListener to ignore keys typed into a TextField

**AskTextPopup.kt**
- Added a row under the Close and OK buttons for the error text. Previously, it displaying on the same row.
- Removed the hardcoded empty string validation. If someone wants to restrict empty strings, they can use the validate parameter

**PromotionPickerScreen.kt**
- Clicking the rename unit button now calls the new class UnitRenamePopup
- The unit instancename (if not null) and unit base name are now displayed above the promotion text at the bottom of the screen. Added a function to build the unit name and promotion description text for display.
- Removed check for null instancename to allow for renaming of unit from this screen
- Removed check for new promotion to allow naming of unit from this screen even if there are no promotions available

**UnitOverviewtable.kt**
- Added a column with blank header after the unit name column. For every unit, this column contains an edit icon that calls UnitRenamePopup.

**UnitRenamePopup.kt**
- New class
- Opens an AskTextPopup that allows the user to rename a unit.
- If the user clears the textbox and clicks OK, the instancename for the unit will be set to null, erasing the custom username. Otherwise, there was no way to clear a custom name.
- Lambda to allow code execution after clicking OK, such as refreshing the calling screen to show the new unit name

**UnitTable.kt**
- Separated the unitIconHolder and unitNameLabel touchable configurations.
- Clicking unitIconHolder now opens the Civilopedia
- Clicking unitNameLabel calls UnitRenamePopup
- Added buildNameLabelText function to create the UnitTable display name for unitNameLabel. This includes the unit health.

**CityScreenCityPickerTable.kt**
- Added a validate for empty string passed for city name, since I removed the hardcoded validation in AskTextPopup

**ReligiousBeliefsPickerScreen.kt**
- Added a validate for empty string passed for religion name, since I removed the hardcoded validation in AskTextPopup

**Known Issue:**
Opening the unit promotion screen from the overview screen, changing unit name in promotion screen, and then closing promotion screen does not update the unit name in the unit overview table until the next load. However, promotion changes do not get updated, either.

ScreenShot_20220818_1660845274
ScreenShot_20220818_1660845360
ScreenShot_20220818_1660845440


core/src/com/unciv/ui/overviewscreen/UnitOverviewTable.kt
```diff
@@ -131,6 +133,7 @@ class UnitOverviewTab(
     private fun Table.updateUnitHeaderTable(): Table {
         defaults().pad(5f)
         add("Name".toLabel())
+        add("".toLabel())
```
xlenstra, line 136: To add an empty field, you can also just do `add()`

xlenstra (approved):

core/src/com/unciv/ui/pickerscreens/UnitRenamePopup.kt
```diff
@@ -0,0 +1,49 @@
...
+    init {
+        AskTextPopup(
+            screen,
+            label = "Choose name for [${unit.baseUnit.name}]",
+            icon = ImageGetter.getUnitIcon(unit.name).surroundWithCircle(80f),
+            defaultText = if (unit.instanceName == null) {
+                unit.baseUnit.name
+            } else {
+                unit.instanceName.toString()
+            },
```
JackRainy, line 34: These lines can be replaced with just one:
```Kotlin
defaultText = unit.instanceName ?: unit.baseUnit.name.tr(),
```
Do not forget that the initial suggestion for the name should be shown in the language the user sees its type, so `.tr()` is needed there.

core/src/com/unciv/ui/pickerscreens/UnitRenamePopup.kt
```diff
@@ -0,0 +1,49 @@
...
+            validate = { it != unit.name },
+            actionOnOk = { userInput ->
+                //If the user inputs an empty string, clear the unit instanceName so the base name is used
+                if (userInput == "") {
+                    unit.instanceName = null
+                } else {
+                    unit.instanceName = userInput
+                }
```
JackRainy, line 42: The minor thing, but these also can be in one line:
``` Kotlin
unit.instanceName = if (userInput == "") null else userInput
```

core/src/com/unciv/ui/worldscreen/unit/UnitTable.kt
```diff
@@ -127,15 +129,20 @@ class UnitTable(val worldScreen: WorldScreen) : Table(){
             if (selectedUnits.size == 1) { //single selected unit
                 separator.isVisible = true
                 val unit = selectedUnit!!
-                var nameLabelText = unit.displayName().tr()
-                if (unit.health < 100) nameLabelText += " (" + unit.health + ")"
+                var nameLabelText = buildNameLabelText(unit)
```
JackRainy, line 132: This `var` can be `val` since it never changes later.

JackRainy (changes requested):
Thank you, quite a great work! Just a few minor things there.

huckdogg:
> Thank you, quite a great work! Just a few minor things there.

Thank you for the kind words and suggestions. I have made the requested changes.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -414,7 +416,13 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
         // City-states can enter ally's territory (the opposite is true anyway even without open borders)
         val newHasOpenBorders = civInfo.getAllyCiv() == otherCivName
-                || trades.flatMap { it.theirOffers }.any { it.name == Constants.openBorders && it.duration > 0 }
+                || trades.flatMap { it.ourOffers }.any { it.name == Constants.openBorders && it.duration > 0 }
+        if (civInfo.civName == "Egypt") {
+            println("updateHasOpenBorders")
+            println(civInfo.civName)
+            println(trades.flatMap { it.theirOffers })
+            println()
+        }
```
JackRainy, line 425: These changes seem to be a debug code, not related to the issue. Also, they change behaviour, please, revert.
FYI: `git add -i` and `git add -p` are really useful tools to avoid commits like this. You control what to include into commit file-by-file and line-by-line.

docs/Modders/uniques.md
```diff
@@ -1890,4 +1887,4 @@ Simple unique parameters are explained by mouseover. Complex parameters are expl
 *[stats]: For example: `+2 Production, +3 Food`. Note that the stat names need to be capitalized!
 *[tech]: The name of any tech.
 *[tileFilter]: Anything that can be used either in an improvementFilter or in a terrainFilter can be used here, plus 'unimproved'
-*[victoryType]: The name of any victory type: 'Neutral', 'Cultural', 'Diplomatic', 'Domination', 'Scientific', 'Time'
+*[victoryType]: The name of any victory type: 'Neutral', 'Cultural', 'Diplomatic', 'Domination', 'Scientific', 'Time'
```
JackRainy, line 1890: Not sure, what has changed here, but probably, this file (same as `DiplomacyManager.kt`) is altered and added by mistake.

JackRainy (changes requested):

docs/Modders/uniques.md
huckdogg: This file keeps getting updated automatically. I'm not familiar with this process, so I'm not sure why. I'll be sure not to push it in the future.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
huckdogg: Thank you for tips. I will use these commands in the future.

xlenstra (approved):

JackRainy (approved):

PR #7681, opened 2022-08-20
JvmArg not work for mac desktop:run.

nacro711072:
Version:
```
kotlinVersion = "1.7.0"
appCodeNumber = 748
appVersion = "4.2.4-patch1"
gradleVersion = 7.2
```

I'm a mac developer. When I build APP with gradle `desktop:run`, i found an error:
截圖 2022-08-20 下午4 13 48

`jvmargs` is not working since the getter of `jvmargs` will new an `ArrayList` on the gradle7.2 :
截圖 2022-08-20 下午4 57 59

### DIFF
--- a/desktop/build.gradle.kts
+++ b/desktop/build.gradle.kts
@@ -19,10 +19,8 @@ sourceSets {
 }
 
 dependencies {
-    if(Os.isFamily(Os.FAMILY_MAC)) {
-        // See https://libgdx.com/news/2021/07/devlog-7-lwjgl3#do-i-need-to-do-anything-else
-        api("com.badlogicgames.gdx:gdx-lwjgl3-glfw-awt-macos:$gdxVersion")
-    }
+    // See https://libgdx.com/news/2021/07/devlog-7-lwjgl3#do-i-need-to-do-anything-else
+    api("com.badlogicgames.gdx:gdx-lwjgl3-glfw-awt-macos:$gdxVersion")
 }
 
 val mainClassName = "com.unciv.app.desktop.DesktopLauncher"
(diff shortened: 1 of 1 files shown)


yairm210:
There's a good chance that one or both of these args are no longer needed since libgdx 1.10.0 - can you check if you can run without them?

nacro711072:
> There's a good chance that one or both of these args are no longer needed since libgdx 1.10.0 - can you check if you can run without them?


https://libgdx.com/news/2021/07/devlog-7-lwjgl3#do-i-need-to-do-anything-else
截圖 2022-08-21 上午11 19 43


After adding package: `com.badlogicgames.gdx:gdx-lwjgl3-glfw-awt-macos:$gdxVersion`, i can run without these args.
But we need support to test on the others platform.

yairm210:
A library is definitely a better alternative to args so everything can work as out-of-the-box as possible :)

If I understand correctly, this will mean that is we build a JAR on Linux (as our current CI works), it won't include that package and therefore will still require the jvm args.
I hope to have time to test that adding this to the regular desktop build has no ill effects, and if so it will be a regular, non-conditional dependency

nacro711072:
> A library is definitely a better alternative to args so everything can work as out-of-the-box as possible :)
> 
> If I understand correctly, this will mean that is we build a JAR on Linux (as our current CI works), it won't include that package and therefore will still require the jvm args. I hope to have time to test that adding this to the regular desktop build has no ill effects, and if so it will be a regular, non-conditional dependency

Today i trying to trace code of this library, and has some conclusion.Maybe we don't need to test for others platform.

Here is my steps:

1. open library, there is only one file `GlfwAWTLoader`
截圖 2022-08-23 上午12 24 58

2. find usage all places
截圖 2022-08-23 上午12 48 01

3. Go to `Lwjgl3Application.java` line 120, which in the method `loadGlfwAwtMacos`
截圖 2022-08-23 上午12 10 23

4. `loadGlfwAwtMacos` is only usage at line 82, and we can find that it's only run on mac.
截圖 2022-08-23 上午12 09 18

As doc mentioned, the extension will be ignored if it not running on mac.
Also, i was running on `window11` today, and it's work perfectly.

Just share and hope you find this information helpful. :)

yairm210 (approved):

PR #7685, opened 2022-08-22
AI land units do not die attacking cities from water

MindaugasRumsa51:


core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -703,6 +703,20 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
...
+        val city = tile.owningCity
+        if (tile.isWater && city !=null &&
+            !unit.civInfo.isPlayerCivilization() &&
+            unit.baseUnit.isLandUnit() &&
+            unit.civInfo.isAtWarWith(tile.getOwner()!!)) {
+
+            //do not die attacking cities from water
+            val distance = getDistance(tile.position, city.location)
+
+            if (distance < 4) {
```
JackRainy, line 716: Why do you use "magic number" 4 here? I may think about the radius of city attack but it is not 4.

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -703,6 +703,20 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
             if (unit.civInfo.isAtWarWith(firstUnit.civInfo))
                 return false
         }
+        
+        val city = tile.owningCity
+        if (tile.isWater && city !=null &&
+            !unit.civInfo.isPlayerCivilization() &&
+            unit.baseUnit.isLandUnit() &&
+            unit.civInfo.isAtWarWith(tile.getOwner()!!)) {
```
JackRainy, line 711: What if the water tile is close to the city, the civ is at war with, but the tile does not belong to this city?
For example: there are 4 tiles in a line: A - B - C - D
A - an enemy city
B - the water tile of the enemy city
C - the friendly water tile with the unit to attack a city A
D - a friendly city
Both cities A and D are able to attack the water tiles B and C here.

JackRainy: Also, the tiles C and D can be even an neutral water (`city == null`). Should this check work in this scenario?

JackRainy:
I am not sure I understand the problem you are trying to solve but this code denies approaching the land units to the cities across the water.
It means the AI will never be able to capture any city on the islands (e.g. the archipelago map), since it will not be able to approach and enter them with the land units.

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
MindaugasRumsa51: so '4' can can somehow changed to city.range

if water tile near city belongs to another city, then ai land units will die

MindaugasRumsa51:
once i saw trireme capturing city..... hmmmm

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
JackRainy: > if water tile near city belongs to another city or neutral waters, then ai land units will die

It means your solution does not solve the problem in full scope, but bring some additional complications, like the AI will be limited in options of capturing the coastal cities (by ships only).
Sorry, that does not sound for me as an improvement. 
Currently, AI uses quite good tactic: bring 4-6 land units, 1 of them will die on the way (that's what you trying to fix) but others will reach the land next turn successfully and will surround the city.

yairm210:
Closing since this solution is flawed and no alternative solution has been offered

PR #7699, opened 2022-08-23
Update Vietnamese.properties

Weltzer:


xlenstra:
Currently, Vietnamese is not supported as a language due to Unciv not being able to render its glyphs/letters. As such, the file of strings to be translated is hopelessly outdated and so these translations will never be shown ingame. I'll not close this as these translations may be useful if we ever get the Vietnamese glyphs showing up, but do keep this in mind

PR #7701, opened 2022-08-23
Fix unit not giftable in one-sided open borders agreement

huckdogg:
Changed isFriendlyTerritory(CivilizationInfo) function to check the active civ's DiplomacyManager against the tileOwner civ, instead of the other way around. This makes sense because we want to see if the tileOwner has opened their borders, not if the requesting civ has opened borders with the tileOwner. This appears to resolve the issue where a unit can't be gifted to the TileOwner if only the tile owner has opened their borders. Issue #7673

I have tested all other calls of the isFriendlyTerritory function, and everything is functioning as expected.


JackRainy (approved):
Approved.
P.S> You may also want to write an unit test for gifts routines at some point.

PR #7714, opened 2022-08-26
Moddable UI Skins [Basics]

GGGuenni:
This PR adds the ability to select a UI Skin from the Options->Display menu. UI Skins currently only support `checkbox`, `checkbox-pressed`, `select-box` (Drop-Down), `select-box-pressed`, `roundedEdgeRectangle` (Buttons) and `rectangleWithOutline` (No idea what it's used for). 
I've also changed how Skins are loaded from mods so they don't automatically overwrite the default Skin but work like tilesets instead.
To give modders the ability to change the padding and stretch areas of their Skins, I've removed the hardcoded NinePatch generation and replaced every skin .png with a [.9.png](https://developer.android.com/studio/write/draw9patch) which already contains the information for padding and stretch areas.

UncivUI

Sadly Libgdx does not support multiple stretch areas which would be extremely useful for skin modding. Fortunately, there is a project called [TenPatch](https://github.com/raeleus/TenPatch) that introduces this functionality to Libgdx. It would be a dependency though and I don't know what you guys think about that.

Up next would be making more of the UI moddable and adding something like SkinStrings (TileSetStrings) and SkinConfig (TilesetConfig) classes.


core/src/com/unciv/ui/options/DisplayTab.kt
```diff
@@ -16,14 +16,14 @@ import com.unciv.ui.utils.WrappableLabel
...
-import com.unciv.ui.worldscreen.WorldScreen
 
 private val resolutionArray = com.badlogic.gdx.utils.Array(arrayOf("750x500", "900x600", "1050x700", "1200x800", "1500x1000"))
 
 fun displayTab(
     optionsPopup: OptionsPopup,
     onResolutionChange: () -> Unit,
-    onTilesetChange: () -> Unit
+    onTilesetChange: () -> Unit,
+    onSkinChange: () -> Unit,
```
yairm210, line 26: Since use the same function for all of these, how about replacing them with a single 'onChange' function?

core/src/com/unciv/ui/options/OptionsPopup.kt
```diff
@@ -147,6 +147,7 @@ class OptionsPopup(
                 }
             }
             withGLContext {
+                BaseScreen.setSkin()
```
yairm210, line 150: Why do we need this?

GGGuenni: A lot of UI elements rely on BaseScreen.skin. When changing skins, these UI elements do not get updated if setSkin is not called and would only update on a restart.

PR #7715, opened 2022-08-26
Add AUR update workflow

Thyrum:
Port the AUR update workflow for [unciv-bin](https://aur.archlinux.org/packages/unciv-bin) from my [aur-packages](https://github.com/Thyrum/aur-packages/blob/master/.github/workflows/unciv-bin.yml) repository to the Unciv github repository.

This does require adding some secrets to the repository and I don't know how I would do this.

Note I haven't been able to properly test this inside the Unciv repository because I lack many of the secrets required (obviously). It should work though because it also works in the aur-packages repository mentioned earlier.

Edit: And my editor automatically removes trailing whitespace, this of course doesn't necessarily have to be added to the repository.

### DIFF
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -1,8 +1,8 @@
 # This is a basic workflow to help you get started with Actions
 
-name: Build and deploy 
+name: Build and deploy
 
-# Controls when the action will run. 
+# Controls when the action will run.
 on:
   # Triggers the workflow on push or pull request events but only for the master branch
   push:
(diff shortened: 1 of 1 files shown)


yairm210:
Please add 'continue on error true' to the package upload as well, so one failure doesn't fail our entire workflow.

Thyrum:
Nice catch, fixed it using a force push.

yairm210:
If this is working as intended, it would be great if you could add a badge with a link to the readme.md love for pi-apps :)

Thyrum:
As stated before the proper secrets to push to the AUR are still missing, how would I go about adding those?

yairm210:
Hm, send them to me at yairm210@hotmail.com and I'll add them in

PR #7721, opened 2022-08-29
delete unnecessary random(randomness.RNG)

lishaoxia1985:


yairm210:
What does this improve?

yairm210:
Why do you think this randomness unnecessary?

lishaoxia1985:
Because this method is to search all tiles in waterTiles and TilesToCheck, so we can always choose the first or last element in every loop, random is unnecessary.

lishaoxia1985:
In fact, the lake tile is not random, when you create water and land tiles, the lake tiles are determined too.

PR #7722, opened 2022-08-29
Update German translation

Mape6:


android/assets/jsons/translations/German.properties
```diff
@@ -360,8 +360,7 @@ Disable starting bias = Keine Startvorgaben
 Raging Barbarians = Wütende Barbaren
 No Ancient Ruins = Keine altertümlichen Ruinen
 No Natural Wonders = Keine Naturwunder
- # Requires translation!
-Allow anyone to spectate = 
+Allow anyone to spectate = Erlaube jedem zu beobachten
```
GGGuenni, line 363: If `You are not allowed to spectate!` is `Du darfst nicht zuschauen!` then this should be `Erlaube jedem zuzuschauen` or `Jedem das Zuschauen erlauben` as this is a checkbox selection

Mape6: You are totally right. I fixed that.

PR #7738, opened 2022-09-03
Fixed a wrongly implemented feature disabling pantheon founding

xlenstra:
Before I had implemented it so that whenever a religion was enhanced, no more pantheons could be founded. However, as I just read in the discord, this is incorrect; you should be able to found pantheons until there are enough to fill all religion slots.


MindaugasRumsa51:
question: how java function "min" is better than kotlin function "minOf"

MindaugasRumsa51:
question2:  it is correct to count all civilizations? maybe only major ones must be counted?

yairm210:
That 'min' is the new Kotlin function as of 1.7

yairm210 (approved):

yairm210:
And we count all religions with a certain state of religion, which should only be true for major civs anyway, since only they can create religions

MindaugasRumsa51:
question3: if "min" is kotlin, then for what purposes file have this:  
import java.lang.Integer.max
import java.lang.Integer.min

Issue #7717, opened 2022-08-28
Unable to load saved game: "cityInfo has not been initialized"


SpacedOutChicken:
Windows 10 64-bit, game version 4.2.6 from Itch

After saving a game with DeCiv Redux enabled, I am unable to load that game. I get this error message: "Unhandled problem, UninitializedPropertyAccessException lateinit property cityInfo has not been initialized". The mod error checker detects no problems with the mod, but saved games are becoming inaccessible.

![Screenshot (149)](https://user-images.githubusercontent.com/58439827/187096129-49a8cb7e-05bf-489a-9084-758e8d1f95b5.png)

Saved game data:
```
[...]

PR #7739, opened 2022-09-03
fix: lateinit property cityInfo has not been initialized issue.

nacro711072:
solve issue: #7717 

1. 
The var `cityInfo` is not initial properly in `CityReligionManager` class. Here is Exception stack:
截圖 2022-09-03 下午6 40 16

And I go to file `CivilizationInfo.kt:847`, we can found that `cityInfo.civInfo = this`(line 485) should be executed before `thingsToFocusOnForVictory = civInfo.getPreferredVictoryTypeObject()?.get...`. Thus, i exchanged the order.
截圖 2022-09-03 下午6 46 06

2. After I finished step 1, it occurred a new exception: The var `cityInfo` is not initial properly in `ReligionManager` class.
截圖 2022-09-03 下午6 50 34

`ReligionManager.isMajorityReligionForCiv(ReligionManager.kt:83)` will use **ALL** `civInfo`

But, `GameInfo.setTransients(GameInfo.kt:494)` will trigger `CivilizationInfo.setTransients` **iteratively**, in this function will init `ReligionManager.cityInfo`

截圖 2022-09-03 下午6 55 29

截圖 2022-09-03 下午6 54 11

Hence, i initialize var `thingsToFocusOnForVictory` after ALL `CivilizationInfo.setTransients` has been invoke. 

Anyone is welcome to suggest prettier approach to solve this issue.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -491,7 +491,12 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
 
         for (religion in religions.values) religion.setTransients(this)
 
-        for (civInfo in civilizations) civInfo.setTransients()
+        civilizations.forEach { civInfo ->
+            civInfo.setTransients()
+        }
+        civilizations.forEach { civInfo ->
+            civInfo.thingsToFocusOnForVictory = civInfo.getPreferredVictoryTypeObject()?.getThingsToFocus(civInfo) ?: setOf()
+        }
 
         convertFortify()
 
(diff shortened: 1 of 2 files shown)


yairm210 (approved):
Good documentation of the problem and why you fixed it like this 👍🏿

core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -491,7 +491,12 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion
 
         for (religion in religions.values) religion.setTransients(this)
 
-        for (civInfo in civilizations) civInfo.setTransients()
+        civilizations.forEach { civInfo ->
+            civInfo.setTransients()
+        }
+        civilizations.forEach { civInfo ->
+            civInfo.thingsToFocusOnForVictory = civInfo.getPreferredVictoryTypeObject()?.getThingsToFocus(civInfo) ?: setOf()
+        }
```
xlenstra, line 499: Why not just use a for loop instead of a forEach?

nacro711072: It's just a personal preference, ha

xlenstra (approved):

PR #7755, opened 2022-09-07
AI: Recognize when no further techs can be researched, even if some techs are blocked

### LINKED ISSUE #7530: Automated science and Mods
when you have mod with two (or more) different technology paths, you will never learn all techs.
problem is here:
ConstructionAutomation.kt  line:
   val allTechsAreResearched = civInfo.tech.getNumberOfTechsResearched() >= civInfo.gameInfo.ruleSet.technologies.size

i would change it to line:
val allTechsAreResearched = (civInfo.tech.era.eraNumber + 1) >= civInfo.gameInfo.ruleSet.eras.size

but i not sure about correctness of such change

yairm210:
We allow mods to block techs for certain civs by policies, other techs, or simply restricted to certain civs
Our construction automation takes into account when all techs have been researched to deprioritize science buildings and production. This *was* only affected when *all* techs were researched, and is now affected when the current civ cannot research any new techs, taking the 'tech restriction' into consideration.

Resolves #7530

### DIFF
--- a/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
@@ -32,7 +32,8 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
     private val militaryUnits = civUnits.count { it.baseUnit.isMilitary() }
     private val workers = civUnits.count { it.hasUniqueToBuildImprovements && it.isCivilian() }.toFloat()
     private val cities = civInfo.cities.size
-    private val allTechsAreResearched = civInfo.tech.getNumberOfTechsResearched() >= civInfo.gameInfo.ruleSet.technologies.size
+    private val allTechsAreResearched = civInfo.gameInfo.ruleSet.technologies.values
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
```diff
@@ -32,7 +32,8 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
     private val militaryUnits = civUnits.count { it.baseUnit.isMilitary() }
     private val workers = civUnits.count { it.hasUniqueToBuildImprovements && it.isCivilian() }.toFloat()
     private val cities = civInfo.cities.size
-    private val allTechsAreResearched = civInfo.tech.getNumberOfTechsResearched() >= civInfo.gameInfo.ruleSet.technologies.size
+    private val allTechsAreResearched = civInfo.gameInfo.ruleSet.technologies.values
+        .all { civInfo.tech.isResearched(it.name) || !civInfo.tech.canBeResearched(it.name)}
```
xlenstra, line 36: Technically `isReached()` is already called in `canBeResearched()`, so adding it isn't necessary, but it does improve readability

PR #7760, opened 2022-09-07
dịch thêm

vvinam:
thêm fanpage và group UnCiv Việt Nam

### DIFF
--- a/android/assets/jsons/translations/Vietnamese.properties
+++ b/android/assets/jsons/translations/Vietnamese.properties
@@ -1,9 +1,10 @@
 
 # Tutorial tasks
-
+#Fanpage UnCiv Việt Nam https://www.facebook.com/101649856020305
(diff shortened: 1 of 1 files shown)


xlenstra:
Do note that the Vietnamese translation is currently not supported, as we are unable to render its glyphs/letters. As such, we also don't update our translation file for Vietnamese, and you're translating strings there were relevant two years ago, and many of which has since changed.
Regardless, as you have made some changes, I'll mark this translation as mergeable, but do note that it will never show in-game as is.

xlenstra:
I've tested it and Vietnamese seems to work fine for me on both kubuntu an android as well, so in my opinion it can be re-enabled. Probably gdx or some other library updated and it's fine for now.
We can try re-enabling Thai for next release and seeing if we get complaints from native speakers, but the spacing really does look horrible.

PR #7772, opened 2022-09-09
Enabled vietnamese

xlenstra:
As tested by both SomeTroglodyte and me in #7760, Vietnamese letters seem to render just fine, so this PR re-enables the language as a choice. Due to the recent translations, ~19% of the file is already translated. This PR also includes the regenerated file so that it can be translated by the Vietnamese community before the next release.

I've tested that rendering Thai glyphs also seems to work, but there are _very_ wide spaces, so I've not enabled that for now.


GGGuenni (approved):

yairm210:
Awesome :)
I wonder what changed? Maybe it was improvements in the LibGDX version
In any case anyways glad to include a new language, but please a flag for it as well

yairm210:
You know, it's been so long I don't really recall why we disabled it and what exactly didn't work ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

PR #7774, opened 2022-09-09
Translate vietnamese

vvinam:
update


xlenstra:
This PR doesn't have any changed lines, did you delete your changes by accident? Anyway, I would advise not adding any more translation until #7772 is merged, as that updates the strings that should be translated.

android/assets/jsons/translations/Vietnamese.properties
```diff
@@ -1,4 +1,4 @@
-
+#update đã dịch tới dòng 3000
```
xlenstra, line 1: Note that this will automatically be removed when we release the next version

xlenstra: Same with the links to facebook below

PR #7776, opened 2022-09-09
Spies now show their current action and take time to move and set up

xlenstra:
This PR continues the implementation of espionage (see #7610) by making spies take time to move to and establish networks in cities. You can also only see with spies after they have established connections in the city, which takes a total of 4 turns.
The espionage overview now also shows the current action the spy is taking, and how long until this action is complete. For unimplemented actions, this amount is 0 and never changes.
<details> <summary> Image: </summary>

image
</details>


android/assets/jsons/translations/template.properties
```diff
@@ -1463,6 +1463,12 @@ Spy Hideout =
 Location = 
 Spy present = 
 Move = 
+None = 
+Moving =
```
yairm210, line 1467: Please add these dynamically in the translationfilewriter by iterating on the spy action enum
There are a couple of other enums we do this with

core/src/com/unciv/logic/civilization/EspionageManager.kt
```diff
@@ -17,15 +32,59 @@ class Spy() : IsPartOfGameInfoSerialization {
...
-        return gameInfo.getCities().firstOrNull { it.id == location }
+    fun setTransients(civInfo: CivilizationInfo) {
+        this.civInfo = civInfo
     }
 
-    fun getLocationName(gameInfo: GameInfo): String {
-        return getLocation(gameInfo)?.name ?: Constants.spyHideout
+    fun endTurn() {
+        --timeTillActionFinish
+        if (timeTillActionFinish == 0)
```
yairm210, line 46: Reverse if to return early

android/assets/jsons/translations/template.properties
xlenstra: Should now be fixed :)

PR #7778, opened 2022-09-10

xk730:


xk730:
Originally, I intended to translate only a little (sesame), but some players urged me in the community... (then I will plant a watermelon......Maybe？)

heipizhu4 (approved):
Nice trans! Just a line not to be sure.

PR #7782, opened 2022-09-11
Can Pillage roads/railroads

itanasi:
Save for testing:
[PillageRoad.txt](https://github.com/yairm210/Unciv/files/9541801/PillageRoad.txt)

Seems pillaging Roads/Railroads gives no gold in Civ5. So maintaining same effect here


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -1206,13 +1206,19 @@ open class TileInfo : IsPartOfGameInfoSerialization {
...
+            roadStatus = RoadStatus.None
             turnsToImprovement = 2
+        } else {
+            if (isLand) {
+                // Setting turnsToImprovement might interfere with UniqueType.CreatesOneImprovement
+                removeCreatesOneImprovementMarker()
+                improvementInProgress = improvement
+                turnsToImprovement = 2
+            }
+            improvement = null
```
xlenstra, line 1220: So if you pillage the same tile twice, the original improvement is fully removed? That differs from civ 5 behaviour iirc

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -306,11 +307,28 @@ object UnitActions {
 
     fun getPillageAction(unit: MapUnit): UnitAction? {
         val tile = unit.currentTile
-        if (unit.isCivilian() || tile.improvement == null || tile.getOwner() == unit.civInfo) return null
+        println(tile.roadStatus)
```
xlenstra, line 310: debug code :)

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -290,7 +291,7 @@ object UnitActions {
             actionList += UnitAction(UnitActionType.Pillage, action = null)
         else actionList += UnitAction(type = UnitActionType.Pillage) {
             if (!worldScreen.hasOpenPopups()) {
-                val pillageText = "Are you sure you want to pillage this [${unit.currentTile.improvement}]?"
+                val pillageText = if(unit.currentTile.improvement == null) "Are you sure you want to pillage this [${unit.currentTile.roadStatus.name}]?" else "Are you sure you want to pillage this [${unit.currentTile.improvement}]?"
```
xlenstra, line 294: This is a _very_ long line, maybe:
```kotlin
val tile = unit.getTile()
val pillagedImprovement = if (tile.improvement == null) tile.roadStatus.name else tile.improvement
val pillageText = Are you sure you want to pillage this [$pillagedImprovement]?
```
Maybe inline the `tile` if you find it too verbose

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -306,11 +307,28 @@ object UnitActions {
...
+                        )
+                    } else {
+                        tile.getOwner()?.addNotification(
+                            "An enemy [${unit.baseUnit.name}] has pillaged our [${tile.improvement}]",
+                            tile.position,
+                            "ImprovementIcons/${tile.improvement!!}",
+                            NotificationIcon.War,
+                            unit.baseUnit.name
+                        )
+                    }
```
xlenstra, line 331: This looks like code duplication; why not add a variable containing the name of the thing that is pillaged and use that instead of adding the same code twice with just one differing variable?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -833,7 +851,7 @@ object UnitActions {
         if (unit.isTransported) return false
         val tileImprovement = tile.getTileImprovement()
         // City ruins, Ancient Ruins, Barbarian Camp, City Center marked in json
-        if (tileImprovement == null || tileImprovement.hasUnique(UniqueType.Unpillagable)) return false
+        if ((tileImprovement == null || tileImprovement.hasUnique(UniqueType.Unpillagable)) && tile.roadStatus == RoadStatus.None) return false
```
xlenstra, line 854: Does this actually work? I'd imagine that if you have a tile with both a road and an improvement, you can now pillage this tile as the road can be pillaged, but as improvements are pillaged first, the improvement is pillaged instead of the road

core/src/com/unciv/logic/map/TileInfo.kt
itanasi: Correct. Need to overhaul the system to handle marking improvements (and routes) as pillaged without removing them. Which is something I started but is a pretty huge overhaul

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: This function is just to T/F if you could pillage on this tile

xlenstra: I understand, but I believe it now returns `true` when the tile has an improvement that cannot be pillaged but it has a road, and in those cases the improvement will be pillaged instead of the road

itanasi: Gotcha. I see it now. Ok

PR #7783, opened 2022-09-11
Solve CS placement clustering

yairm210:
Current code places major civs in the center of regions, and CS states on the edges of regions,
This leads to the unfortunate effect of having large swathes of CS-lands, as far away from the major civs as possible, with ring patterns if the number of Civs is small and the number of CSs is large.

Examples:
image
image
image
image
image
image

ETC

This PR simplifies the CS placement logic to have them scattered randomly across regions, after they have been sorted into regions (that part is untouched), leading to major civs being relatively equidistant from each other, with CSs scattered in later to leave cities scattered in a relatively spaced-out manner, like so (same generation parameters as first of the above images).

2 major civs, 12 CSs:
image
image

And with 4 major civs:
image
image

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapRegions.kt
@@ -987,7 +987,7 @@ class MapRegions (val ruleset: Ruleset){
                 if (!canPlaceMinorCiv(tile)) continue
                 val continent = tile.getContinent()
                 if (continent in uninhabitedContinents) {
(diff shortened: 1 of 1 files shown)


xk730:
CS :OHHHHHH my sense of game participation has been improved !!!

HAAA

xlenstra:
I believe the algorithm we had was coded by SimonCeder, who followed the publicly available Civ 5 code, so this change does create a difference with Civ 5. That being said, the change is in my opinion an improvement.

xlenstra (approved):

PR #7787, opened 2022-09-12
Starting in later eras triggers era uniques in all previous eras

xlenstra:
When we trigger the unique to give spies to civilizations, we only give them to major civilizations that are alive. However, at the start of the game civilizations have no cities, so they are only alive if they have any units. As before this PR the tech of civs was initialized before the units were placed, there would be no units while determining the era of a civ and the subsequent triggering of era uniques. As a result, no civ would get spies. This PR changes this, by first adding the units and only then initializing the tech. As far as I can see, this has no negative side-effects.

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -88,6 +88,11 @@ object GameStarter {
                     }
         }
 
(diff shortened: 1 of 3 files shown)


yairm210:
That rename isn't important enough to block this :)

PR #7790, opened 2022-09-12
Update UnitPromotions.json

Thienanbad:


yairm210:
I understand why you're doing this, so there's a difference in translation between the promotion name and the unit type.
However, such a change would break all existing translations, would remove promotions from all existing units, and in short it's a Huge Mess
Also, please leave a description on PRs so people don't need to guess your intentions

What *specific* problem are you encountering?

Thienanbad:
Because it make confuse for some modder. Some edit besiege some edit siege

Vào 2:09, Th 3, 13 thg 9, 2022 Yair Morgenstern ***@***.***>
đã viết:

yairm210:
There is a reason that these two terms are different, because them mean different things. One is a state, one is an action.
Merging them would make things worse, since languages in which there is no one word that makes sense for both contexts will have to make do with a bad word instead.

Thienanbad:
I have another commits about translation can you merge it?

Vào 2:58, Th 3, 13 thg 9, 2022 Yair Morgenstern ***@***.***>
đã viết:

yairm210:
translation commits are merged on version release

Thienanbad:
One last requeste. Can you create for me a atlas file? I have png and json

Vào 3:01, Th 3, 13 thg 9, 2022 Yair Morgenstern ***@***.***>
đã viết:

xlenstra:
Ask in the discord, there people are more likely to help you

Issue #7781, opened 2022-09-10
Science/Happiness Points from City-States not showing after Loading


CodeKatta:
**Platform**
Android 12 (One UI 4.1)

**Version**
4.2.7 f-droid

**Describe the bug**
If you adopt the policies 'Scholasticism' resp. 'Cultural Diplomacy' (Patronage) the points for 'Science' resp. 'Happiness' you receive from city-states do not show in the green research box (top left) and research tree resp. next to the happy face at the top of the screen **but only after you load a saved game**.
In the 'Stats'-screen the value for 'Happiness' is correct. If you enter a 'City'-screen and exit again or hit 'Next Turn' the correct value is displayed as it should. Wrt 'Science' you will probably notice that you need one fewer turn (or even less for higher techs in the research tree).

As far as I can tell , however, the game engine seems to use the correct values. It's probably just a display issue.

xlenstra:
This is probably a result of the order in which stats for civs are determined. As we probably determine `statsForNextTurn` for major civs first, the stats for city-states still have their default value of 0, which is added by the Scholasticism policy. Then afterwards the city-state stats for next turn are determined. Before the turn ends, we recalculate all those stats to make sure our values are correct, so you should indeed just get the correct amount of stats.

Should be a relatively easy fix (just change the order in which stats are determined when loading a game), but I don't have the time to test that the above is correct and implement it.

nacro711072:
I will take a look.

yairm210:
Solved by #7793

PR #7793, opened 2022-09-13
fix wrong happiness point on resume game if adopt 'Cultural Diplomacy' policy.

nacro711072:
As https://github.com/yairm210/Unciv/issues/7781#issuecomment-1243043600 mentioned, just an order problem.
the resources of all city-state must be updated (`GameInfo.kt:line 510 -> civInfo.updateSightAndResources()`) before calculate the happiness of major civ.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -499,7 +499,12 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
 
         convertFortify()
 
-        for (civInfo in civilizations) {
+        sequence {
+            // update city-state resource first since the happiness of major civ depends on it.
+            // See issue: https://github.com/yairm210/Unciv/issues/7781
+            yieldAll(civilizations.filter { it.isCityState() })
+            yieldAll(civilizations.filter { !it.isCityState(); it.isMajorCiv() })
+        }.forEach { civInfo ->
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -499,7 +499,12 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion
 
         convertFortify()
 
-        for (civInfo in civilizations) {
+        sequence {
+            // update city-state resource first since the happiness of major civ depends on it.
+            // See issue: https://github.com/yairm210/Unciv/issues/7781
+            yieldAll(civilizations.filter { it.isCityState() })
+            yieldAll(civilizations.filter { !it.isCityState(); it.isMajorCiv() })
```
xlenstra, line 506: By definition any civ that is a major civ cannot be a city state, so you can remove the first function call here

core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -499,7 +499,12 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion
...
+        }.forEach { civInfo ->
```
xlenstra, line 507: Why a forEach instead of a for loop?

nacro711072: Sor, my personal habit. 😫

xlenstra (approved):

PR #7799, opened 2022-09-15
Initial Desktop run starts fullscreen

yairm210:
Current Unciv starts as a tiny screen on Desktop upon initial startup.
This was due to the fact that LibGDX, which is where we usually get our screen size from, is not initialized at the stage where we need to configure how it's going to start, which is where we set the initial screen size.

Solved by using regular ol' Java awt to get the initial screen size, so players have an initial fullscreen experience, which is IMO a much better one.
We also save the settings file so this initial size is 'locked in' for future runs.

### DIFF
--- a/core/src/com/unciv/logic/UncivFiles.kt
+++ b/core/src/com/unciv/logic/UncivFiles.kt
@@ -9,14 +9,16 @@ import com.unciv.UncivGame
 import com.unciv.json.fromJsonFile
 import com.unciv.json.json
 import com.unciv.models.metadata.GameSettings
+import com.unciv.models.metadata.WindowState
 import com.unciv.models.metadata.doMigrations
 import com.unciv.models.metadata.isMigrationNecessary
 import com.unciv.ui.saves.Gzip
 import com.unciv.ui.utils.extensions.toNiceString
-import com.unciv.utils.concurrency.Concurrency
 import com.unciv.utils.Log
+import com.unciv.utils.concurrency.Concurrency
 import com.unciv.utils.debug
 import kotlinx.coroutines.Job
+import java.awt.Toolkit
 import java.io.File
 import java.io.Writer
 
@@ -309,7 +311,9 @@ class UncivFiles(
             }
         }
 
-        return settings ?: GameSettings().apply { isFreshlyCreated = true }
+        return settings ?: GameSettings().apply {
+            isFreshlyCreated = true
+        }
     }
 
     fun setGeneralSettings(gameSettings: GameSettings) {
@@ -328,12 +332,22 @@ class UncivFiles(
             // FileHandle is Gdx, but the class and JsonParser are not dependent on app initialization
             // In fact, at this point Gdx.app or Gdx.files are null but this still works.
             val file = FileHandle(base + File.separator + SETTINGS_FILE_NAME)
+
             return if (file.exists())
                 json().fromJsonFile(
                     GameSettings::class.java,
                     file
                 )
-            else GameSettings().apply { isFreshlyCreated = true }
+            else GameSettings().apply { isFreshlyCreated = true
+                // LibGDX not yet configured, use regular java class
+                val screensize = Toolkit.getDefaultToolkit().screenSize
+                windowState = WindowState(
+                    width = screensize.width,
+                    height = screensize.height
+                )
+                file.writeString(json().toJson(this),false)
+            }
+
         }
 
         /** @throws IncompatibleGameInfoVersionException if the [gameData] was created by a version of this game that is incompatible with the current one. */
--- a/core/src/com/unciv/logic/UncivFiles.kt
+++ b/core/src/com/unciv/logic/UncivFiles.kt
@@ -311,9 +311,7 @@ class UncivFiles(
             }
         }
 
-        return settings ?: GameSettings().apply {
-            isFreshlyCreated = true
-        }
+        return settings ?: GameSettings().apply { isFreshlyCreated = true }
     }
 
     fun setGeneralSettings(gameSettings: GameSettings) {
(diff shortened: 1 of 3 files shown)


GGGuenni:
But it still starts in 900x600 resolution, doesn't it? Looks weird as a first impression in full screen I think
The resolution setting has a comment that auto detecting the resolution was a bad idea though so I guess changing it for fullscreen is nothing that's easy to tackle right?

yairm210:
Ah good point, for desktop we should start in a much higher res because if the screen size
It *is* hard to guess which virtual resolutions are good for which size screens, but starting at something in the middle is probably a good idea

core/src/com/unciv/logic/UncivFiles.kt
```diff
@@ -328,12 +330,24 @@ class UncivFiles(
...
+
             return if (file.exists())
                 json().fromJsonFile(
                     GameSettings::class.java,
                     file
                 )
-            else GameSettings().apply { isFreshlyCreated = true }
+            else GameSettings().apply {
+                isFreshlyCreated = true
+                resolution = "1200x800" // By default Desktops should have a higher resolution
```
GGGuenni, line 341: I don't quite get it. Isn't this also called on android when the game is started the first time and not only on desktop?

yairm210: Nope

yairm210: I don't think so at least... Maybe I missed something?

GGGuenni: It does get called apparently 
but running your branch gives me an exception because awt is not available

```Kotlin
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.unciv.app, PID: 20238
    java.lang.NoClassDefFoundError: Failed resolution of: Ljava/awt/Toolkit;
        at com.unciv.logic.UncivFiles$Companion.getSettingsForPlatformLaunchers(UncivFiles.kt:343)
        at com.unciv.app.AndroidLauncher.onCreate(AndroidLauncher.kt:43)
        at android.app.Activity.performCreate(Activity.java:8000)
        at android.app.Activity.performCreate(Activity.java:7984)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1310)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3423)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3602)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2067)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7697)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:952)
     Caused by: java.lang.ClassNotFoundException: Didn't find class "java.awt.Toolkit" on path: DexPathList[[zip file "/data/app/~~ZqDBNMXWb3X93aaU1rrSbg==/com.unciv.app-ustl8wf_tTx0697aZafPyQ==/base.apk"],nativeLibraryDirectories=[/data/app/~~ZqDBNMXWb3X93aaU1rrSbg==/com.unciv.app-ustl8wf_tTx0697aZafPyQ==/lib/arm64, /data/app/~~ZqDBNMXWb3X93aaU1rrSbg==/com.unciv.app-ustl8wf_tTx0697aZafPyQ==/base.apk!/lib/arm64-v8a, /system/lib64, /system/system_ext/lib64, /system/product/lib64]]
        at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:207)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
        at com.unciv.logic.UncivFiles$Companion.getSettingsForPlatformLaunchers(UncivFiles.kt:343) 
        at com.unciv.app.AndroidLauncher.onCreate(AndroidLauncher.kt:43) 
        at android.app.Activity.performCreate(Activity.java:8000) 
        at android.app.Activity.performCreate(Activity.java:7984) 
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1310) 
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3423) 
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3602) 
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85) 
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135) 
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95) 
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2067) 
        at android.os.Handler.dispatchMessage(Handler.java:106) 
        at android.os.Looper.loop(Looper.java:223) 
        at android.app.ActivityThread.main(ActivityThread.java:7697) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:952) 
W/System: A resource failed to call close. 
```

yairm210:
Okay, problem was that even though the AWT functions were not *used* by Android, importing them in the file meant that Kotlin would try to load the lib, fail, and crash
The entire 'get settings before GDX' has now moved to the Desktop project to avoid any and all such interactions

yairm210:
Okay I'm going to have to give this a do over and see how I can accommodate everything I was

PR #7808, opened 2022-09-16

Ranbut:


Ranbut:
I have a question. How I can handle punctuation in the translation? Because part of the text have ? in places of special characters. What is the correct way write these?

PR #7811, opened 2022-09-16
Add Pillaging Roads and Repair Action

itanasi:
Rework Improvements to match Civ5 and support tiles having a pillaged state
Add support for Improvements and Roads to be Pillaged and Repaired
Tiles turn Brown when Pillaged. Workers can Repair and functionality (resources and city connections) is restored immediately (matching Civ5 behavior).
Worker Automation will look for Repairable Tiles. Will also try to avoid Enemy Units more.
Pillaged Resources are updated in yield and resources provided.
NOT REVERSE COMPATIBLE since adding new variables

image

Translations working
image

Applying color filter to tile (brown) and improvements (red)
image

Save to experiment:
[Autosave-Korea-76.txt](https://github.com/yairm210/Unciv/files/9590905/Autosave-Korea-76.txt)


xlenstra:
> NOT REVERSE COMPATIBLE

If this is the case, please update the `CURRENT_COMPATABILITY_NUMBER` variable in the companion object of `GameInfo` (`GameInfo.kt:71`) to enforce this :)

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -352,10 +352,12 @@ object Automation {
         val stats = tile.getTileStats(null, civInfo)
         var rank = rankStatsValue(stats, civInfo)
         if (tile.improvement == null) rank += 0.5f // improvement potential!
+        if (tile.isPillaged()) rank += 0.6f
         if (tile.hasViewableResource(civInfo)) {
             val resource = tile.tileResource
             if (resource.resourceType != ResourceType.Bonus) rank += 1f // for usage
             if (tile.improvement == null) rank += 1f // improvement potential - resources give lots when improved!
+            if (tile.isPillaged()) rank += 1.1f // even better, repair is faster
```
xlenstra, line 360: So this adds the bonus twice, for a total of 1.7f?

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -140,6 +141,29 @@ class WorkerAutomation(
...
+                    return currentTile.startWorkingOnImprovement(
+                        chooseImprovement(unit, currentTile)!!, civInfo, unit
+                    )
+                }
+            }
+            return
+        }
+
+        if (currentTile.isPillaged()) {
+            println("WorkerAutomation: ${unit.label()} -> repairs $currentTile")
```
xlenstra, line 165: Debug output?

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -256,9 +280,11 @@ class WorkerAutomation(
...
                 .filter {
                     (it.civilianUnit == null || it == currentTile)
-                            && tileCanBeImproved(unit, it)
-                            && it.getTilesInDistance(2)
-                            .none { tile -> tile.isCityCenter() && tile.getCity()!!.civInfo.isAtWarWith(civInfo) }
+                            && (tileCanBeImproved(unit, it) || it.isPillaged())
+                            && it.getTilesInDistance(2)  // don't work in range of enemy cities
+                        .none { tile -> tile.isCityCenter() && tile.getCity()!!.civInfo.isAtWarWith(civInfo) }
+                            && it.getTilesInDistance(4)  // don't work in range of enemy units
+                        .none { tile -> tile.militaryUnit != null && unit.civInfo.isAtWarWith(tile.militaryUnit!!.civInfo) && tile.militaryUnit!!.movement.canReach(it) }
```
xlenstra, line 287: `unit.movement.canReach()` is a very costly function, I wouldn't call this for every AI worker; it feels like that could add up to a lot of time in late-game when there are a lot of workers and units.

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -579,14 +579,20 @@ open class TileGroup(
 
     private fun updateTileColor(isViewable: Boolean) {
         val baseTerrainColor = when {
+            isViewable && tileInfo.isPillaged() && tileSetStrings.tileSetConfig.useColorAsBaseTerrain -> tileInfo.getBaseTerrain().getColor().lerp(Color.BROWN, 0.6f)
+            isViewable && tileInfo.isPillaged() -> Color.WHITE.cpy().lerp(Color.BROWN, 0.6f)
```
xlenstra, line 583: I'm not 100% sure I like this, imo pillaged sprites should be implemented via the tileset and not directly in the code. For now it's fine, but it's something that'd be nice to change in a future PR.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -306,14 +310,27 @@ object UnitActions {
...
     fun getPillageAction(unit: MapUnit): UnitAction? {
         val tile = unit.currentTile
-        if (unit.isCivilian() || tile.improvement == null || tile.getOwner() == unit.civInfo) return null
-
+        println("improvement:${tile.improvement}")
+        println("unpillaged improvement:${tile.getUnpillagedImprovement()}")
+        println("road:${tile.roadStatus}")
+        println("unpillaged road:${tile.getUnpillagedRoad()}")
+        if (unit.isCivilian() || (tile.getUnpillagedImprovement() == null && tile.getUnpillagedRoad() == RoadStatus.None) || tile.getOwner() == unit.civInfo) return null
+        println("action")
```
xlenstra, line 318: Debug println?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -482,6 +499,34 @@ object UnitActions {
         )
     }
 
+    private fun addRepairAction(unit: MapUnit, actionList: ArrayList<UnitAction>) {
+        if (!unit.hasUniqueToBuildImprovements) return
+        if (unit.isEmbarked()) return
+        val tile = unit.getTile()
+        if (!(tile.improvementIsPillaged || tile.roadIsPillaged)) return
```
xlenstra, line 506: Don't you have a function `tile.isPillaged()` to check for exactly this?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -482,6 +499,34 @@ object UnitActions {
...
+
+    fun getRepairAction(unit: MapUnit): () -> Unit {
+        return {
+            unit.action = null
+            unit.currentMovement = 0f
+            val tile = unit.currentTile
+            if (tile.improvementIsPillaged)
+                tile.improvementIsPillaged = false
+            else
+                tile.roadIsPillaged = false
```
xlenstra, line 524: Shouldn't repairing take 3 turns instead of being instant?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -831,9 +876,9 @@ object UnitActions {
 
     fun canPillage(unit: MapUnit, tile: TileInfo): Boolean {
         if (unit.isTransported) return false
-        val tileImprovement = tile.getTileImprovement()
+        val tileImprovement = tile.getUnpillagedTileImprovement()
         // City ruins, Ancient Ruins, Barbarian Camp, City Center marked in json
-        if (tileImprovement == null || tileImprovement.hasUnique(UniqueType.Unpillagable)) return false
+        if ((tileImprovement == null || tileImprovement.hasUnique(UniqueType.Unpillagable)) && tile.getUnpillagedRoad() == RoadStatus.None) return false
```
xlenstra, line 881: Doesn't this still have the same problem as last PR? I.e., if you try to pillage a tile with both a road and an unpillagable improvement, you still pillage the improvement?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: I had tried on Civ 5 G&K and it was instant

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: > > NOT REVERSE COMPATIBLE
> 
> If this is the case, please update the `CURRENT_COMPATABILITY_NUMBER` variable in the companion object of `GameInfo` (`GameInfo.kt:71`) to enforce this :)

This is for marking that saves will break if you try to open one with the features in an older version?

itanasi:
> > NOT REVERSE COMPATIBLE
> 
> If this is the case, please update the `CURRENT_COMPATABILITY_NUMBER` variable in the companion object of `GameInfo` (`GameInfo.kt:71`) to enforce this :)

This is for marking if a save will break if opened in an earlier version?

xlenstra:
> > > NOT REVERSE COMPATIBLE
> > 
> > 
> > If this is the case, please update the `CURRENT_COMPATABILITY_NUMBER` variable in the companion object of `GameInfo` (`GameInfo.kt:71`) to enforce this :)
> 
> This is for marking if a save will break if opened in an earlier version?

As far as I understand it, yes

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
xlenstra: What game speed did you try it on? If it was 'quick', I wouldn't be surprised if it instantly repaired it, but maybe on marathon it doesn't

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Yes, similar to the tile.improvement getting 0.5f for needing an improvement, then +1f for if ViewableResource for a 1.5 total

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
itanasi: Ah, supposed to be debug() call instead

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Haha, yup. What I get for merging the Road code on top

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Just checked again. Regular speed is 2 turns

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: tileInfo.setPillaged() handles Unpillagable now

core/src/com/unciv/ui/tilegroups/TileGroup.kt
itanasi: Right. In the future it'd be nice if this was handled purely in the tileset. Although I do like coloring the base tile since Roads are impossible to show.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: K, now 2 turn Repair. Need to figure out the values on all speeds

xlenstra: I'd gamble that its 3 * gameSpeed.modifier, as almost everything is that

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
xlenstra: Ah my bad, missed that.

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Well the code comment claims 3, but Standard Speed G&K is 2. I see the scaling factors in the code but hoping there's some XML values somewhere

itanasi: Nope, tried Vanilla Standard, also 2 turns. Weird

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
itanasi: Removed

yairm210:
There's a lot to like here, but having the Repair action be part of the jsons means that modders can decide to not include it, so your code should always be able to handle the case where the repair action *does not* exist.

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
```diff
@@ -176,8 +176,17 @@
 		"uniques": ["Can be built outside your borders"],
 		"shortcutKey": "."
 	},
+    {
+        "name": "Repair",
+        "terrainsCanBeBuiltOn": ["Land"],
+        "turnsToBuild": 2,
+        "uniques": ["Can be built outside your borders"],
+        "shortcutKey": "R",
```
yairm210, line 184: That's already the shortcut for Road, so if you have a road and a pillaged improvement there's a conflict...

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
```diff
@@ -176,8 +176,17 @@
 		"uniques": ["Can be built outside your borders"],
 		"shortcutKey": "."
 	},
+    {
+        "name": "Repair",
+        "terrainsCanBeBuiltOn": ["Land"],
+        "turnsToBuild": 2,
+        "uniques": ["Can be built outside your borders"],
```
yairm210, line 183: Hmmm I'm not 100% sure about this since this could lead to pillaging and repairing enemy improvements, I'm okay with it for now but we need to keep this in mind

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -256,9 +281,11 @@ class WorkerAutomation(
         val workableTiles = currentTile.getTilesInDistance(4)
                 .filter {
                     (it.civilianUnit == null || it == currentTile)
-                            && tileCanBeImproved(unit, it)
-                            && it.getTilesInDistance(2)
-                            .none { tile -> tile.isCityCenter() && tile.getCity()!!.civInfo.isAtWarWith(civInfo) }
+                            && (tileCanBeImproved(unit, it) || it.isPillaged())
```
yairm210, line 284: Shouldn't "tileCanBeImproved" just automatically return True if the tile is pillaged and the unit can repair pillaged tiles?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -482,6 +495,55 @@ object UnitActions {
...
+        val tile = unit.currentTile
+        if (!tile.isPillaged()) return 0
+        var repairTurns = tile.ruleset.tileImprovements["Repair"]!!.getTurnsToBuild(unit.civInfo, unit)
+        if (tile.improvement != null && tile.improvementIsPillaged) {
+            if (tile.getTileImprovement()!!.getTurnsToBuild(unit.civInfo, unit) < repairTurns)
+                repairTurns = tile.getTileImprovement()!!.getTurnsToBuild(unit.civInfo, unit)
+        } else {
+            if (tile.ruleset.tileImprovements[tile.roadStatus.name]!!.getTurnsToBuild(unit.civInfo, unit) < repairTurns)
+                repairTurns = tile.ruleset.tileImprovements[tile.roadStatus.name]!!.getTurnsToBuild(unit.civInfo, unit)
+        }
```
yairm210, line 508:
```suggestion
        val pillagedImprovement = if(tile.improvementIsPillaged) tile.getTileImprovement()!! else tile.ruleset.tileImprovements[tile.roadStatus.name]!!
        
        val turnsToBuild = pillagedImprovement.getTurnsToBuild(unit.civInfo, unit)
        if (turnsToBuild < repairTurns) repairTurns = turnsToBuild
```

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
itanasi: Hm, I didn't check, but can you pillage neutral improvements?

I can add a check so you can't repair Enemy Improvements (not sure why you'd want to except prepping for a city takeover)

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -482,6 +495,55 @@ object UnitActions {
...
+            title = "${UnitActionType.Repair} - [${turnsToBuild}${Fonts.turn}]",
+            action = getRepairAction(unit).takeIf { couldConstruct }
+        )
+    }
+
+    fun getRepairAction(unit: MapUnit): () -> Unit {
+        return {
+            unit.currentMovement = 0f
+            val tile = unit.currentTile
+            if (getRepairTurns(unit) - 1 == 0) {  // handle instant fix
```
yairm210, line 534: ??? Why this weird phrasing? What's wrong with "if (x == -1)"? Should be equivalent

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
yairm210: There's a "Pillaging this improvement yields [stats]" unique, and if you can repair enemy improvements that can lead to you squatting on an enemy tile with a military unit and a worker and pillage/repairing it every 2 turns

itanasi: OK, I'll drop it then

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
itanasi: tileCanBeImproved seems mostly concerned about building a new improvement and not handling Repair. Fixing the logic to enforce this

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Cleaned up

android/assets/jsons/Civ V - Gods & Kings/TileImprovements.json
itanasi: Fixed

itanasi:
> There's a lot to like here, but having the Repair action be part of the jsons means that modders can decide to not include it, so your code should always be able to handle the case where the repair action _does not_ exist.

OK, created a backup TileImprovement for Repair in TileInfo. I feel like it should get added in Game Ruleset instead? But that feels very hacky and hard to find

xlenstra:
> > There's a lot to like here, but having the Repair action be part of the jsons means that modders can decide to not include it, so your code should always be able to handle the case where the repair action _does not_ exist.
> 
> OK, created a backup TileImprovement for Repair in TileInfo. I feel like it should get added in Game Ruleset instead? But that feels very hacky and hard to find

I'd think that the right way to solve this would not be to include a backup version of the unit action, but rather make it so that if no 'repair' action is in the ruleset, either repairing is just impossible, or pillaged improvements are destroyed immediately. In both cases, making sure nothing weird happens when it isn't included is the main concern. Automatically adding it in the ruleset if it isn't present to me feels like forcing modders to use it, even if they don't want it, and will probably lead to hacky ways in which modders try to disable it (setting the building time to 9999 and then be confused why AI still try to repair, for example).

itanasi:
K, made so if Repair doesn't exist, Pillage go straight to destruction.
I couldn't get a save to update the tile though. gameInfo.setTransients is only run on new games? And the TileInfo.normalizeToRuleset() is only Map Editor?

PR #7821, opened 2022-09-19
Desktop settings initialize to fullscreen, v2

yairm210:
Changes now local to Desktop to avoid affecting Android runs in any way
Redo of #7799


nacro711072 (approved):

GGGuenni (approved):
It's a bit weird on my end because it is not fully centered on the screen. The window is screen-sized but moved up by a few pixels. Don't know why. 
weird
Maybe other Windows users can confirm/deny. 
But it's not that big of a deal

Issue #7826, opened 2022-09-19
Issue running on Mac OSX Monterrey with M1 processors


MRigal:
**Platform**
Mac OSX Monterrey (12.4), with Apple Silicon M1 Pro processor

**Version**
Tried with 4.1.18 and with 4.2.10

**Describe the bug**
The game is impossible to start.

**To Reproduce**
1. Install java bia `brew install java[@11]`
2. Download Unciv.jar
3. Run the command specified at https://yairm210.github.io/Unciv/Other/Installing-on-macOS/

Please note that the behaviour is the same with OpenJDK 18 and 11. I've also tried without the headless option, also without success. Please note that I had it working on an older, Intel-based Mac with openjdk11, see #2551

```
$ /opt/homebrew/opt/openjdk/bin/java --version
openjdk 18.0.2.1 2022-08-18
OpenJDK Runtime Environment Homebrew (build 18.0.2.1+0)
OpenJDK 64-Bit Server VM Homebrew (build 18.0.2.1+0, mixed mode, sharing)
$ /opt/homebrew/opt/openjdk/bin/java -XstartOnFirstThread -Djava.awt.headless=true -jar Unciv.jar
Exception in thread "main" java.awt.HeadlessException
	at java.desktop/sun.awt.HeadlessToolkit.getScreenSize(HeadlessToolkit.java:186)
	at com.unciv.app.desktop.DesktopLauncher.main(DesktopLauncher.kt:55)
$ /opt/homebrew/opt/openjdk/bin/java -XstartOnFirstThread  -jar Unciv.jar
zsh: trace trap  /opt/homebrew/opt/openjdk/bin/java -XstartOnFirstThread -jar Unciv.jar
```

**Expected behavior**
I thought the jar could be used also this way on Mac OSX. I haven't tried yet to compile it.

yairm210:
Oh shoot
That error specifically looks like it may have to do with the screen resize code?
Can I bother you too check 4.2.9?

yairm210:
Also you shouldn't need the -XStartoffirstthread anymore, try running without that? Just a hunch, maybe that's what makes it 'hesdless' and makes the java toolkit not recognize the screen size

MRigal:
But lol `/opt/homebrew/opt/openjdk/bin/java -jar Unciv.jar` just works :-D

It actually works well both with openjdk 11 and openjdk 18. So I think you just need to update the docs for Mac OSX installation :-)

Thanks @yairm210 !!! 💯

[...]

PR #7828, opened 2022-09-20
update doc for mac user.

nacro711072:
#7826

Update doc for mac user. No longer need additional arguments while running jar command.


nacro711072:
On #7681, we added library `com.badlogicgames.gdx:gdx-lwjgl3-glfw-awt-macos` and both of these args (`XstartOnFirstThread` / `java.awt.headless=true`) are no longer needed.

MRigal:
Nice! Faster than me :-)

PR #7833, opened 2022-09-20
Update Installing-on-macOS

MRigal:
As a complement to the simplification of #7828, I propose to ease a bit the installation instructions for mac, proposing to use Homebrew as it is a de-facto standard and as the installation of the official Java 8 may seem scary to some users.

I left both choices to not induce a breaking change, but I'm open to keep only one method.
@yairm210 and @nacro711072  FYI

### DIFF
--- a/docs/Other/Installing-on-macOS.md
+++ b/docs/Other/Installing-on-macOS.md
@@ -4,8 +4,10 @@ There is currently two ways to install Unciv on macOS. It is recommended that yo
 
 ## Installing using JAR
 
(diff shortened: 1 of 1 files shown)


MRigal:
Unfortunately, the markdown is not rendered properly, how it is displayed on Github... The nested list thingy did not work. Any idea how to fix that? https://yairm210.github.io/Unciv/Other/Installing-on-macOS/

yairm210:
Looks like it requires 4 spaces done that's the markdown requirement
https://github.com/Python-Markdown/markdown/issues/3

PR #7834, opened 2022-09-20
Update Simplified_Chinese.properties

xk730:

PR #7845, opened 2022-09-21

heipizhu4:
Small fixes


heipizhu4:
Hopefully won't conflict with #7834 after more commits

PR #7855, opened 2022-09-25
Create Cannot Move Unique

itanasi:
As Requested for mods, a Cannot Move Unique.

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -103,9 +103,6 @@ class MapUnit : IsPartOfGameInfoSerialization {
     val doubleMovementInTerrain = HashMap<String, DoubleMovementTerrainTarget>()
 
     @Transient
-    var cannotMove = false
-
-    @Transient
     var canEnterIceTiles = false
 
     @Transient
@@ -344,7 +341,6 @@ class MapUnit : IsPartOfGameInfoSerialization {
             .minOfOrNull { it.params[0].toFloat() }
 
         //todo: consider parameterizing [terrainFilter] in some of the following:
-        cannotMove = hasUnique(UniqueType.CannotMove)
         canEnterIceTiles = hasUnique(UniqueType.CanEnterIceTiles)
         cannotEnterOceanTiles = hasUnique(UniqueType.CannotEnterOcean, StateForConditionals(civInfo=civInfo, unit=this))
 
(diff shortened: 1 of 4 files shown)


yairm210:
This doesn't need to be cached since it will only be checked very infrequently.
But why not simply set movement to 0? What's the use case?

itanasi:
Having stationary units that can still attack. Such as defensive towers

yairm210:
Fair enough.
I understand *why* you added the transient, but in this specific case I'm not sure it's needed, since you should only have 1 check per unit in the entire turn.
Doesn't really hurt though so cool.

yairm210 (approved):

xlenstra:
I still don't understand why this should be a unique and cannot be done by simply setting the movement of the unit to 0?

zhan313:
> I still don't understand why this should be a unique and cannot be done by simply setting the movement of the unit to 0?

IIRC a unit with 0 Movement can still move 1 tile (or 1 time - if air unit)

yairm210:
As stated, a unit with 0 movement won't be able to attack. They want a stationary unit that can attack.

xlenstra:
> As stated, a unit with 0 movement won't be able to attack. They want a stationary unit that can attack.

Ah, I didn't make that connection, makes sense. But if I'm not mistaken, the unit would still show movement in the Unit Box, right? That might be confusing to users. Wouldn't it then make more sense to make a unique allowing a unit to attack even if it has 0 movement instead? Then the unit would correctly show that it has 0/0 movement, but still be able to attack

xlenstra:
> > I still don't understand why this should be a unique and cannot be done by simply setting the movement of the unit to 0?
> 
> IIRC a unit with 0 Movement can still move 1 tile (or 1 time - if air unit)

Ah, this is actually a good point.

xlenstra (approved):

itanasi:
> > > I still don't understand why this should be a unique and cannot be done by simply setting the movement of the unit to 0?
> > 
> > 
> > IIRC a unit with 0 Movement can still move 1 tile (or 1 time - if air unit)
> 
> Ah, this is actually a good point. Based on this I'll approve this, but I still believe this lacks a bit in explaining the mechanic to the user. I also don't understand why this is a transiently cached unique and not just a unique that is checked in `MapUnit.getMaxMovement()` and returns 0 early if it exists.

Sure, can make that change.

And yes, in the Unit Info they would see Movement Points, but be unable to move. I did check they can still do a Ramged Attack though if someone is in range

itanasi:
So can't set getMaxMovement() to 0 because that borks a lot of other things and doesn't prevent the UI from showing you movement options. Do need to add it to Air Units code otherwise the UI shows you the full movement range even though you can't reach.

xlenstra (approved):

PR #7856, opened 2022-09-25
Change Default Map Generation to Perlin

itanasi:
Switch from Cellular Automata, which with default settings would generally make maps with 60-70% land, to Perlin which is the baseline for most other map generation already. Especially when dealing with Water Level tends to make more intuitive sense


core/src/com/unciv/logic/map/MapParameters.kt
```diff
@@ -128,15 +128,15 @@ object MapShape : IsPartOfGameInfoSerialization {
...
     const val pangaea = "Pangaea"
     const val continents = "Continents"
     const val fourCorners = "Four Corners"
-    const val perlin = "Perlin"
     const val archipelago = "Archipelago"
     const val innerSea = "Inner Sea"
 
     // Cellular automata
-    const val default = "Default"
+    const val cellularAutomata = "Cellular Automata"
```
yairm210, line 139: Please call this something that users can understand - 'cellular automata' isn't a map type, is a map generation type, and users don't know what to expect

PR #7862, opened 2022-09-25
Better desktop window persist

SomeTroglodyte:
Inspired by #7821.

- New settings will start desktop to a fully maximized window on the primary monitor, which will restore to 80% of that monitor size
- Maximized windows will remember they're maximized, which monitor they're maximized on, and their restore position and size across Unciv restarts
- Normal windows will keep size and position including which monitor they're on across Unciv restarts
- Changing monitor setup between Unciv restarts (Unciv starts to a different set of monitors that it saved window state for) will revert gracefully, so at least 33% of the window are remains visible.
- ~~I needed console logging to display only _my_ messages, which was hard to do, so Log.kt is now a little more flexible.~~
- Less yellow/green squigglies than before.
- Tested thoroughly on Mint single-monitor and Win10 multimonitor.

**Note**: `UncivWindowListener` is just renamed from `MultiplayerTurnNotifierDesktop`, since that class essentially was only "the" Lwjgl3WindowListener for the app, and had that name because flashing the taskbar icon was its only purpose. Sorry I didn't manage to get github to track the rename.

## To-Do
- ~~Decide how to leave logging~~
- Decide whether to remove Windows-specific flashing - no idea what the standard `glfwRequestWindowAttention` will actually look like, neither on Windoze nor Mint
- Main menu isn't perfectly centered - not from this PR
- Loading screen banner isn't perfectly centered - likely not from this PR either
- ~~Wayland and other OS where a Window cannot read its position: Anybody able to debug-step into glfwGetWindowPos on Fedora around here?~~


yairm210:
prepare version

yairm210:
Version rollout PR created

yairm210:
I think Kill, tbh
The problem of 'make it fullscreen' was already solved, and I'm not sure what this is actually proposing to add to that

yairm210:
Then by your leave, I think we'll leave this it

PR #7881, opened 2022-10-01
Fix throne typo in start intro part 1

esensar:
Replaces thron with throne in one of the tutorial intro messages


xlenstra (approved):

yairm210:
This is a good fix, but I prefer to do this after merging the other translations so it doesn't cause any conflicts with subsequent new translations :)
So after we create the 'new translations' branch, we'll try to merge this into there. If no conflicts then great, if there are we'll do this fix manually

PR #7886, opened 2022-10-03
fixed no war relation

### LINKED ISSUE #7884: Don't show war relationship in politics tab
The politics tab don't show war relationship though there are war
Screenshot_20221003_032151
Screenshot_20221003_032123

nacro711072:
Fix #7884

The `DiplomacyFlags.DeclaredWar` flag is used to remember the number of turns will expire, and then user can negotiate a peace with enemy. It will be removed from `flagsCountdown` (DiplomacyManager.kt:117) when the time‘s up.


nacro711072:
Duplicated  #7887. More clearly.

PR #7891, opened 2022-10-04
Add Temperature Offset Slider

itanasi:
Adds a new Advanced Map setting to adjust Temperature Offset. At max almost completely removes Snow. At minimum squashes the top/bottom desert into the middle and covers poles in snow.
Extend some of the humidity settings to fully cover range up to 1.0 for extra safety

Examples below are all with same seed

**Example with +0.4 Temp**
Huge
image
Medium
image

**0.0 Temp**
Huge
image
Medium
image

**-0.4 Temp**
Huge
image
Medium
image


MindaugasRumsa51:
can be there planet of snow for modders? for example one of scenes of star wars  
 going on frozen planet of snow

yairm210:
Modders can add their own tile definitions...

itanasi:
As a general heads up, I'm currently thinking if there is a better way to handle the underlying temperature and humidity so that when we increase temperature the deserts would end up taking up more of the space as one might expect, but that is for a future pull

PR #7892, opened 2022-10-05
Update Nations.json

xk730:
Modify the description error about North Korea

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Nations.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Nations.json
@@ -425,7 +425,7 @@
 		"innerColor": [255,0,0],
 		"favoredReligion": "Confucianism",
 		"uniqueName": "Scholars of the Jade Hall",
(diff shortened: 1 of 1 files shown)


xk730:
> Pro:
> 
> * The [code in Building.postBuildEvent](https://github.com/yairm210/Unciv/blob/2a740b9edeea1667d12b049baaee5a48df839ebf/core/src/com/unciv/models/ruleset/Building.kt#L674) does indeed not check for the capital
> 
> Neutral:
> 
> * existing translations lost
> 
> Con:
> 
> * Changes the use of a key without changing [the definition of the key](https://github.com/yairm210/Unciv/blob/2a740b9edeea1667d12b049baaee5a48df839ebf/core/src/com/unciv/models/ruleset/unique/UniqueType.kt#L232)
> * bad english ("Our" capitalized mid-sentence, "anycity" is not a word - just replacing "capital" with "any city" would suffice)
> 
> To-do:
> 
> * Is the behaviour incorrect or the description
> * Redo unique mod-friendlier with conditionals and/or cityFilter

That interesting and important！and As for how to modify, it may be decided by the player after a better solution is proposed.

yairm210:
The correct fix for this is to check for the capital city, as per Civ V, and not to enable this for all cities.
Also, this will not work and is not grammatically correct as mentioned.

PR #7901, opened 2022-10-06
fix untranslated string about politics

nacro711072:
update some politics text. #6131

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1219,6 +1219,7 @@ Number of your cities celebrating\n'We Love The King Day' thanks\nto access to t
 WLTK demand = 
 WLTK- = 
 Number of your cities\ndemanding this resource for\n'We Love The King Day' = 
+Politics = 
 Show global politics = 
 Show diagram = 
 At war with [enemy] = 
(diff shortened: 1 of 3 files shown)


yairm210 (approved):
We don't need .apply when adding onClick since existing function already chains, but that's an existing problem not a new one. Good job!

nacro711072:
> Without digging too deep, this looks well-researched and -done.
> 
> Tiny nag: `String.toTextButton` [exists](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/utils/extensions/Scene2dExtensions.kt#L434). I would prefer it to look more like this:
> 
> ```kotlin
> /** Translate a [String] and make a [TextButton] widget from it */
> fun String.toTextButton() = TextButton(this.tr(), BaseScreen.skin)
> /** Translate a [String] and make a [TextButton] widget from it using a specific [style] */
> fun String.toTextButton(style: TextButtonStyle) = TextButton(this.tr(), style)
> /** Translate a [String] and make a [TextButton] widget from it with click handler [onClickAction] */
> fun String.toTextButton(onClickAction: ()->Unit) = toTextButton().apply { onClick(onClickAction) }
> ```
> 
> ... or so, instead of using a runtime `if` that could be resolved at compile time, plus allowing most of your potential uses here a little readability shortcut like `val diagramButton = "Show diagram".TextButton { updateDiagram() }` ... but hey, as it is, it needs to be used where possible, too.

How about this:
```kotlin
fun String.toTextButton(style: TextButtonStyle? = null, onClickAction: (() -> Unit)? = null): TextButton {
    val text = this.tr()
    val textButton = if (style == null) TextButton(text, BaseScreen.skin) else TextButton(text, style)
    if (onClickAction != null) {
        textButton.onClick { onClickAction() }
    }
    return textButton
} 
```

e.g:
```kotlin
val diagramButton = "Show diagram".toTextButton(onClickAction = ::updateDiagram)
```

> 
> Another: template `[numberOfTurns] Turns Left = ` now even more unused?

Maybe, but not sure definitely.

yairm210:
... since the onclick can be applied globally to any actor we don't need to add it inside the totextbutton function, wr can simply chain it wherever relevant

nacro711072:
> ... since the onclick can be applied globally to any actor we don't need to add it inside the totextbutton function, wr can simply chain it wherever relevant

Got it. Reverted~

PR #7910, opened 2022-10-11
fix citystate declare war on itself

MindaugasRumsa51:


yairm210:
You can't become the ally of a CS you are at war with so this would never happen

yairm210:
Unless you managed to make this happen in since game?

MindaugasRumsa51:
this problem is reported in discord various times
as i understood problem is when you declaring war to your ally city state

yairm210:
So this is a guess, and it's in the wrong place.
This code is when city states *become* your ally.

MindaugasRumsa51:
new report about it says that problem related with NUKE

PR #7964, opened 2022-10-27
fix memory leak due to Pixmap don't be released.

nacro711072:
Ref: https://libgdx.com/wiki/graphics/taking-a-screenshot

All we need to do is to dispose `Pixmap`

### DIFF
--- a/core/src/com/unciv/ui/LoadingScreen.kt
+++ b/core/src/com/unciv/ui/LoadingScreen.kt
@@ -36,7 +36,10 @@ class LoadingScreen(
             for (popup in previousScreen.popups) popup.isVisible = false
             previousScreen.render(Gdx.graphics.getDeltaTime())
         }
-        val screenshot = ScreenshotTexture(Pixmap.createFromFrameBuffer(0, 0, Gdx.graphics.backBufferWidth, Gdx.graphics.backBufferHeight))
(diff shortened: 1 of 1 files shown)


nacro711072:
I play the game with 4 AI and a spectator.

Start:

截圖 2022-10-27 下午4 09 59

After 10 turns:
截圖 2022-10-27 下午4 12 56

Native memory increase to 16X MB

---

After dispose `Pixmap`

Start: 
截圖 2022-10-27 下午4 21 06

After 10 turns:
截圖 2022-10-27 下午4 10 51

yairm210:
Oh nice! I only saw the effect on the graphics, didn't notice there was an effect on the native as well!

nacro711072:
> I remember reading [the Pixmap Jdoc explicitly warning against leaks](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/Pixmap.java#L40), but I didn't 'connect the dots'. (And that text is the tooltip to `Pixmap` right there in our source)
> 
> .. that said, there's only one further Pixmap leak, [this one](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/UncivGame.kt#L151), and it's done only once per run, so not critical.

I don't understand why we need [fontImplementation?.getCharPixmap('S') (UncivGame.kt:151)](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/UncivGame.kt#L151) here to solve ANRs. 🤔 🤔 🤔 

Has any description/issue in detail that can let me take a look on github?

PR #7980, opened 2022-10-31
Added unitset selection to options menu

GGGuenni:
This is a minimal implementation to allow unit images to be selected separately from the tile images.
It raises a question for me, however. Do units need their own config file (only thing I can think of right now would be fallbackUnitSet) or should we at least add a `fallbackUnitSet` variable to the `TilesetConfig`? And in general, would it be wise to split units from tilesets completely (like skins)? That would probably make some mods incompatible.

### DIFF
--- a/core/src/com/unciv/Constants.kt
+++ b/core/src/com/unciv/Constants.kt
@@ -81,6 +81,7 @@ object Constants {
     const val dropboxMultiplayerServer = "Dropbox"
 
     const val defaultTileset = "HexaRealm"
+    const val defaultUnitset = "FantasyHex"
     const val defaultSkin = "Minimal"
 
     /**
--- a/core/src/com/unciv/Constants.kt
+++ b/core/src/com/unciv/Constants.kt
@@ -81,7 +81,7 @@ object Constants {
     const val dropboxMultiplayerServer = "Dropbox"
 
     const val defaultTileset = "HexaRealm"
-    const val defaultUnitset = "FantasyHex"
+    const val defaultUnitset = "HexaRealm"
     const val defaultSkin = "Minimal"
 
     /**
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -38,6 +38,7 @@ class GameSettings {
 
     var turnsBetweenAutosaves = 1
     var tileSet: String = Constants.defaultTileset
+    var unitSet: String = Constants.defaultUnitset
     var skin: String = Constants.defaultSkin
     var showTutorials: Boolean = true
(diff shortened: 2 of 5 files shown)


yairm210:
A. Default unitset should be a hardcoded "tileset", which means "the current tileset"
That way when people change tilesets it will change unitsets automatically

B. What happens of the old unitset is no longer available? For example the mod was deleted? Change back to default

This is a good tile to make AdvancedUnits the default units for HexaRealm :D

yairm210:
"To speak is to take action", so AU will be in Hexarealm, so both tilesets have units :)

GGGuenni:
Regarding B: It already goes back to the defaultUnitset after restart (currently FantasyHex)

Regarding A: So as long as tileset and unitset exist with the same name and both are selected, when changing the tileset the unitset should be switched too? And as soon as another unitset is selected or one does not exist for the selected tileset the unitset should stay as long as non-matching tilesets are selected?

OR should each Tileset have a default Unitset that gets set automatically whenever the Tileset is selected?

Also a bit unrelated: Should the showPixelUnits setting be replaced by a `None` unitset?

yairm210:
Regarding A: The first option, I think
Regarding B: it's possible, do you think that'll make things easier?

GGGuenni:
> The first option, I think

Implemented it with a simple if statement instead of a dedicated `Tileset` selection because It felt weird when switching from `{Tileset: Hexarealm, Unitset: Tileset}` to `{TileSet: ModTilesetWithoutUnitset UnitSet: Hexarealm}`
It's now `{Tileset: Hexarealm, Unitset: Hexarealm}` to `{TileSet: ModTilesetWithoutUnitset UnitSet: Hexarealm}`
or `{Tileset: Hexarealm, Unitset: Hexarealm}` to `{TileSet: FantasyHex UnitSet: FantasyHex}`

> do you think that'll make things easier?

I just think its better UI design

yairm210:
@letstalkaboutdune with this change, we can change the name of your units to "AbsoluteUnits" and still have it be the default unitset.
Are you in favor? Or would you rather it keeps being called "HexaRealm" like the tileset?

letstalkaboutdune:
> @letstalkaboutdune with this change, we can change the name of your units to "AbsoluteUnits" and still have it be the default unitset. Are you in favor? Or would you rather it keeps being called "HexaRealm" like the tileset?

They should be called AbsoluteUnits, since @GeneralWadaling will probably want his unit sprites to be called HexaRealm.

Kuldebar:
I use Ravignir's 5Hex but I like the AbsoluteUnits clarity so this change suits me fine - I have both AbsoluteUnits and AbsoluteUnitsXL installed. (under the assumption that all units will be covered between the two mods - XL was 100% if I recall.

PR #7987, opened 2022-11-04
Update Turkish.properties

rainalert:


yairm210:
Placeholder `[cityName]` not found in `Turkish` for entry `We have looted [amount] from a [improvement] which has been sent to [cityName]`

PR #8000, opened 2022-11-10
AbsoluteUnits - Unique Archers

letstalkaboutdune:
Also fixes bug with Spanish unique embarked sprite.


letstalkaboutdune:
I see that a`ndroid/assets/Tilesets.atlas` and `Tilesets.png` need to get updated as well.  How do I do that?

PR #8005, opened 2022-11-12
Resolve #7817 - don't show unexplored tiles at all

### LINKED ISSUE #7817: Feature request: Minimaps Should Only Encompass Explored Areas
Have the minimap not encompass the whole map and only showing what the player has explored just like most Civ games (Civ V included)
The Image embedded is a demonstration of the feature request, notice how the minimap on the bottom right only shows parts of the map explored by the player? This is what i'm talking about
Demonstration

yairm210:
image

Unexplored tiles do not appear in minimap or main map
Does not include resizing the minimap to what's visible because that is crazy talk and would involve a lot of math and reconstructing the minimap every time the borders expand, maybe later


alexban011:
Looks very good. Sad that we lose the ability to send units in unexplored areas though

yairm210:
I mean it's by definition, if you don't know where is map you don't know to send units there

PR #8015, opened 2022-11-16
Add Repair and Pillaging Roads

itanasi:
Redo pull Request from #7811 because rebasing borked things

Rework Improvements to match Civ5 and support tiles having a pillaged state
Add support for Improvements and Roads to be Pillaged and Repaired
Tiles turn Brown when Pillaged. Workers can Repair and functionality (resources and city connections) is restored immediately (matching Civ5 behavior).
Worker Automation will look for Repairable Tiles. Will also try to avoid Enemy Units more.
Pillaged Resources are updated in yield and resources provided.
NOT REVERSE COMPATIBLE since adding new variables

image

Translations working
image

Applying color filter to tile (brown) and improvements (red)
image

New edit: Repair and Pillage say what will be repaired/pillaged
image

Save to experiment:
[Autosave-Korea-76.txt](https://github.com/yairm210/Unciv/files/9590905/Autosave-Korea-76.txt)


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -205,6 +214,7 @@ open class TileInfo : IsPartOfGameInfoSerialization {
     fun isImpassible() = getLastTerrain().impassable
 
     fun getTileImprovement(): TileImprovement? = if (improvement == null) null else ruleset.tileImprovements[improvement!!]
+    fun getUnpillagedTileImprovement(): TileImprovement? = if (getUnpillagedImprovement() == null) null else ruleset.tileImprovements[improvement!!]
```
yairm210, line 217: imo if getTileImprovement is still actually in use, then *it* should be renamved getPillagedTileImprovement and getUnpillagedTileImprovement renamed  getTileImprovement. The 'default' tile improvement is an unpillaged one.

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -984,11 +995,15 @@ open class TileInfo : IsPartOfGameInfoSerialization {
...
-            lineList += FormattedLine(roadStatus.name, link="Improvement/${roadStatus.name}")
+        if (roadStatus !== RoadStatus.None && !isCityCenter()) {
+            val pillageText = if (roadIsPillaged) " (Pillaged!)" else ""
+            lineList += FormattedLine("[${roadStatus.name}]$pillageText", link = "Improvement/${roadStatus.name}")
+        }
         val shownImprovement = getShownImprovement(viewingCiv)
-        if (shownImprovement != null)
-            lineList += FormattedLine(shownImprovement, link="Improvement/$shownImprovement")
+        if (shownImprovement != null) {
+            val pillageText = if (improvementIsPillaged) " (Pillaged!)" else ""
```
yairm210, line 1004: Also untranslated

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -1195,15 +1210,69 @@ open class TileInfo : IsPartOfGameInfoSerialization {
...
+            } else {
+                // otherwise use pillage/repair systems
+                if (getPillagableImprovement() != null) {
+                    improvementIsPillaged = true
+                } else {
+                    roadIsPillaged = true
+                }
+            }
+        } else {
+            improvement = null
```
yairm210, line 1237: Putting this at the top as 'if (!isLand) { improvement=null; return } would simplify the logic

core/src/com/unciv/ui/tilegroups/TileGroup.kt
```diff
@@ -587,14 +587,20 @@ open class TileGroup(
 
     private fun updateTileColor(isViewable: Boolean) {
         val baseTerrainColor = when {
+            isViewable && tileInfo.isPillaged() && tileSetStrings.tileSetConfig.useColorAsBaseTerrain -> tileInfo.getBaseTerrain().getColor().lerp(Color.BROWN, 0.6f)
+            isViewable && tileInfo.isPillaged() -> Color.WHITE.cpy().lerp(Color.BROWN, 0.6f)
```
yairm210, line 591: ??? Why do you change the entire base terrain color for pillaged improvements? Surely only the improvement icon should be changed?

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -330,7 +343,7 @@ object UnitActions {
         val globalPillageYield = Stats()
         val toCityPillageYield = Stats()
         val closestCity = unit.civInfo.cities.minByOrNull { it.getCenterTile().aerialDistanceTo(tile) }
-        val improvement = tile.ruleset.tileImprovements[tile.improvement]!!
+        val improvement = if (tile.improvement == null) tile.ruleset.tileImprovements[tile.roadStatus.name]!! else tile.ruleset.tileImprovements[tile.improvement]!!
```
yairm210, line 346: See this? THIS line should be a function in TileInfo, and then we can get rid of ALL the if/elses everywhere else!
Just run 'getImprovementToPillage()', and it will either return
- the improvement,
- the road improvement, or
- null

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -811,7 +872,7 @@ object UnitActions {
         if (unit.isFortified() || unit.canFortify() || unit.currentMovement == 0f) return
         // If this unit is working on an improvement, it cannot sleep
         if (unit.currentTile.hasImprovementInProgress()
-            && unit.canBuildImprovement(unit.currentTile.getTileImprovementInProgress()!!)) return
+            && (unit.currentTile.improvementInProgress == Constants.repair || unit.canBuildImprovement(unit.currentTile.getTileImprovementInProgress()!!))) return
```
yairm210, line 875: I think that unit.canBuildImprovement should return 'true' for "repair", and then we don't need this "if/else" here

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -352,10 +352,12 @@ object Automation {
         val stats = tile.getTileStats(null, civInfo)
         var rank = rankStatsValue(stats, civInfo)
         if (tile.improvement == null) rank += 0.5f // improvement potential!
+        if (tile.isPillaged()) rank += 0.6f
```
yairm210, line 355: Remember that this might be a pillaged road. Pillaged roads should not be adding stats

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -984,11 +995,15 @@ open class TileInfo : IsPartOfGameInfoSerialization {
         }
         if (naturalWonder != null)
             lineList += FormattedLine(naturalWonder!!, link="Terrain/$naturalWonder")
-        if (roadStatus !== RoadStatus.None && !isCityCenter())
-            lineList += FormattedLine(roadStatus.name, link="Improvement/${roadStatus.name}")
+        if (roadStatus !== RoadStatus.None && !isCityCenter()) {
+            val pillageText = if (roadIsPillaged) " (Pillaged!)" else ""
```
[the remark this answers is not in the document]
itanasi: No? Definitely in template.properties and just tested. See screenshot above

[improvement] (Pillaged!) =

core/src/com/unciv/ui/tilegroups/TileGroup.kt
itanasi: To make it obvious that the tile has a pillaged improvement/road on it. If the Road is pillaged, coloration is too small. And depending on the improvement sprite could be too indistinct as well. Coloring the whole tile makes it more obvious, similar to Civ5 the entire tile becomes like a wasteland

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Why not? Workers should prioritize pillaged roads as well as improvements

core/src/com/unciv/logic/map/TileInfo.kt
itanasi: Done

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Added function. Will need to review for cleanups

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Already does, so can remove the extra OR here

core/src/com/unciv/logic/map/TileInfo.kt
itanasi: getTileImprovement() is still used to return the tileImprovement when it doesn't matter if it's Pillaged or not. For example checking for Barbarian Camp, Ancient Ruin, and especially for Irreemovable checks or checking if there's an Improvement to Remove.

Some of that can get folded into getPillagedTileImprovement(), but I don't think getPillagedTileImprovement() would be an appropriate name for that functionality since that implies you want to get a tile improvement that's been pillaged to me

core/src/com/unciv/logic/automation/Automation.kt
yairm210: Oh, because this destiny be caught in automate connections...
Good catch! 👍

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -482,6 +495,54 @@ object UnitActions {
...
+    }
+
+    private fun addRepairAction(unit: MapUnit, actionList: ArrayList<UnitAction>) {
+        if (unit.currentTile.ruleset.tileImprovements[Constants.repair] == null) return
+        if (!unit.hasUniqueToBuildImprovements) return
+        if (unit.isEmbarked()) return
+        val tile = unit.getTile()
+        if (!tile.isPillaged()) return
+
+        val couldConstruct = unit.currentMovement > 0
```
yairm210, line 518: If isCityCenter then return, don't show a grayed-out action as current configuration does

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -290,7 +293,9 @@ object UnitActions {
             actionList += UnitAction(UnitActionType.Pillage, action = null)
         else actionList += UnitAction(type = UnitActionType.Pillage) {
             if (!worldScreen.hasOpenPopups()) {
-                val pillageText = "Are you sure you want to pillage this [${unit.currentTile.improvement}]?"
+                val tile = unit.currentTile
+                val pillagedImprovement = if (unit.currentTile.getUnpillagedImprovement() == null) tile.roadStatus.name else tile.improvement
```
yairm210, line 297: Sounds like getImprovementToPillage()

yairm210:
Looks pretty good!

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Hm, a city center should never be able to be Pillaged imo. Adding logic to getFoundCityAction() to prevent this

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Done

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -776,7 +786,8 @@ open class TileInfo : IsPartOfGameInfoSerialization {
     /** Implements [UniqueParameterType.TileFilter][com.unciv.models.ruleset.unique.UniqueParameterType.TileFilter] */
     fun matchesFilter(filter: String, civInfo: CivilizationInfo? = null): Boolean {
         if (matchesTerrainFilter(filter, civInfo)) return true
-        if (improvement != null && ruleset.tileImprovements[improvement]!!.matchesFilter(filter)) return true
+        if (improvement != null && !improvementIsPillaged && ruleset.tileImprovements[improvement]!!.matchesFilter(filter)) return true
+        if (isPillaged() && filter == "pillaged") return true
```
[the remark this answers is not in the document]
itanasi: Yeah, I don't think I'll add it as a filter just yet. Removed

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Cleanups done

itanasi:
OK, code cleanups done. So except for our disagreement on using the color filter on the whole tile ready to be reviewed again

yairm210:
Let's go for it
Be prepared that people may discover other problems :)

Ouaz:
Is it possible to use the Worker icon (the hammer) for the "Repair" civilopedia article?

repair_icon
worker_icon

itanasi:
Yeah I can make an edit later tonight. Also had thought of a few other UI enhancements haha

PR #8026, opened 2022-11-18
AbsoluteUnits - Unique Spearmen

letstalkaboutdune:


yairm210:
Why do you have nation-specific icons that are the same as the general icons? What does this accomplish?

letstalkaboutdune:
> Why do you have nation-specific icons that are the same as the general icons? What does this accomplish?

What it accomplishes is allowing the nation-specific sprite to have different coloring. I do this a few times for the unique units to look exactly how I want and not use the nation color layers.

PR #8055, opened 2022-11-27
Changes to moddable UI

GGGuenni:
+ Added documentation auto-generation for UI elements so whenever a UI element is added it's also added to the docs together with its default value

+ Added UI elements for some PickerScreen backgrounds
  + this includes the `clearColor` setting for SkinConfigs

+ Started using a solution for button states
  + Let's see if modders are fine with it


docs/Modders/Creating-a-UI-skin.md
```diff
@@ -92,32 +118,34 @@ To create a config for your skin you just need to create a new .json file under
...
+            "image": "WorldScreen/TopBar/ResourceTable",
+            "alpha": 0.4
         },
         "WorldScreen/Minimap/Background": {
-          "tint": {"r": 0.2, "g": 0.4, "b": 0.45, "a": 1}
-        },
+            "tint": {"r": 0.2, "g": 0.4, "b": 0.45, "a": 1}
+        }
     }
+}
```
yairm210, line 143: Can we have the colors be regular RGB 0-255 values? 'normalizing' values is not exactly natural when dealing in colors

yairm210 (approved):

docs/Modders/Creating-a-UI-skin.md
GGGuenni: The libgdx Color class does not support integer RGB values for deserialization so I don't think it will be particularly easy to do so 
But yea I don't like working with these float values either

yairm210: I have a helper function fir converting lists of int [1,2,3] to Colors, I use it for all moddable objects

PR #8087, opened 2022-12-05
AI Pillages Neutral Roads only at War

itanasi:
Also track road ownership based on who built the road or who last owned the tile
We could use this to track Road Maintenance properly

Thanks to FiretronP75 for how it works

### DIFF
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -46,7 +46,7 @@ open class TileInfo : IsPartOfGameInfoSerialization {
     fun setOwningCity(city:CityInfo?){
         owningCity = city
         isCityCenterInternal = getCity()?.location == position
-        if (city != null && roadOwner != null)  // only when taking control, otherwise last owner
+        if (city != null)  // only when taking control, otherwise last owner
             roadOwner = city.civInfo
     }
 
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -332,7 +332,7 @@ open class TileInfo : IsPartOfGameInfoSerialization {
         return if (roadOwner != "")
             tileMap.gameInfo.getCivilization(roadOwner)
         else
-            null
+            getOwner()
     }
 
     fun isFriendlyTerritory(civInfo: CivilizationInfo): Boolean {
--- a/core/src/com/unciv/logic/map/TileInfo.kt
+++ b/core/src/com/unciv/logic/map/TileInfo.kt
@@ -47,7 +47,7 @@ open class TileInfo : IsPartOfGameInfoSerialization {
         owningCity = city
         isCityCenterInternal = getCity()?.location == position
         if (city != null)  // only when taking control, otherwise last owner
-            roadOwner = city.civInfo
+            roadOwner = city.civInfo.civName
     }
 
     @Transient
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -104,6 +106,7 @@ open class TileInfo : IsPartOfGameInfoSerialization {
 
     var roadStatus = RoadStatus.None
     var roadIsPillaged = false
+    var roadOwner: CivilizationInfo? = null // either who last built the road or last owner of tile
```
yairm210, line 109: Not good. When serializing, we can't include the whole civilization! Instead we need a transient Civilization and a permanent string value of owningCivName

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -896,7 +896,15 @@ object UnitActions {
         if (!tile.canPillageTile()) return false
         val tileOwner = tile.getOwner()
         // Can't pillage friendly tiles, just like you can't attack them - it's an 'act of war' thing
-        return tileOwner == null || unit.civInfo.isAtWarWith(tileOwner)
+        if (tileOwner == null) {
```
yairm210, line 899: Reverse if for simpler logic

core/src/com/unciv/logic/map/TileInfo.kt
itanasi: Kotlin doesn't seem to have explicit pointer stuff. So is there no smart way for it to store and reference by pointer?

yairm210: Nope, we store the name only and then infer the object when loading the game
Just like with all other objects
There's a lot of interconnectedness in the game and this is a simple way to break it up into bite sized pieces for serialization

itanasi: Ah, that's annoying. K, will check how it's handled elsewhere and copy

itanasi:
Changes done

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
```diff
@@ -896,7 +896,15 @@ object UnitActions {
...
-        return tileOwner == null || unit.civInfo.isAtWarWith(tileOwner)
+        if (tileOwner != null) {
+            return unit.civInfo.isAtWarWith(tileOwner)
+        } else {
+            if (tile.canPillageTile())
+                return true
+            if (tile.canPillageRoad() && tile.roadOwner != "" && unit.civInfo.isAtWarWith(tile.getRoadOwner()!!))
+                return true
+            return false
+        }
```
yairm210, line 907: If we have a 'if return' we don't need the 'else' afterwards, this can be done with no nesting at all :)

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -308,6 +328,13 @@ open class TileInfo : IsPartOfGameInfoSerialization {
         return containingCity.civInfo
     }
 
+    fun getRoadOwner(): CivilizationInfo? {
+        return if (roadOwner != "")
+            tileMap.gameInfo.getCivilization(roadOwner)
+        else
+            null
```
yairm210, line 335: How about returning the civ that owns the tile? That's a good guess

core/src/com/unciv/ui/worldscreen/unit/UnitActions.kt
itanasi: Done

PR #8107, opened 2022-12-10
Units/Buildings can now be given optional Portraits to be displayed instead of flags.

vegeta1k95:
If mod provides an atlas with entries:

`UnitPortraits/*unit name*`
`BuildingPortraits/*building name*`

these portraits will be displayed everywhere (CityView, TechButtons, Civilopedia, etc) instead of flags, except the map itself - on the map units still use flags, as in Civ 5.

Example: my [Civ-5-Icons mod](https://github.com/vegeta1k95/Civ-5-Icons), some unit are given portraits, some don't (so they remain with flag only).
image
image
image

### DIFF
--- a/core/src/com/unciv/ui/images/ImageGetter.kt
+++ b/core/src/com/unciv/ui/images/ImageGetter.kt
@@ -258,7 +258,7 @@ object ImageGetter {
             return if (imageExists(buildingPortraitLocation)) {
                 getImage(buildingPortraitLocation).toGroup(size)
             } else
-                getImage("BuildingIcons/$construction").surroundWithCircle(size)
+                getImage(buildingPortraitLocation).surroundWithCircle(size)
         }
         if (ruleset.units.containsKey(construction)) {
             val unitPortraitLocation = "UnitPortraits/$construction"
--- a/core/src/com/unciv/ui/images/ImageGetter.kt
+++ b/core/src/com/unciv/ui/images/ImageGetter.kt
@@ -252,12 +252,40 @@ object ImageGetter {
         return iconGroup
     }
 
-    fun getConstructionImage(construction: String): Image {
-        if (ruleset.buildings.containsKey(construction)) return getImage("BuildingIcons/$construction")
-        if (ruleset.units.containsKey(construction)) return getUnitIcon(construction)
-        if (construction == "Nothing") return getImage("OtherIcons/Sleep")
-        return getStatIcon(construction)
+    fun getPortraitImage(construction: String, size: Float): Group {
+        if (ruleset.buildings.containsKey(construction)) {
+            return if (buildingPortraitExists(construction)) {
+                val image = getImage("BuildingPortraits/$construction")
+                val group = Group().apply {
+                    setSize(size, size)
+                    image.setSize(size, size)
+                    image.center(this)
+                    image.setOrigin(Align.center)
+                    addActor(image) }
+                group
+            } else
+                getImage("BuildingIcons/$construction").surroundWithCircle(size)
+        }
+        if (ruleset.units.containsKey(construction)) {
+            return if (unitPortraitExists(construction)) {
+                val image = getImage("UnitPortraits/$construction")
+                val group = Group().apply {
+                    setSize(size, size)
+                    image.setSize(size, size)
+                    image.center(this)
+                    image.setOrigin(Align.center)
+                    addActor(image) }
+                group
+            } else
+                getUnitIcon(construction).surroundWithCircle(size)
+        }
+        if (construction == "Nothing")
+            return getImage("OtherIcons/Sleep").surroundWithCircle(size)
+        return getStatIcon(construction).surroundWithCircle(size)
     }
+
+    private fun unitPortraitExists(unitName: String) = imageExists("UnitPortraits/$unitName")
+    private fun buildingPortraitExists(buildingName: String) = imageExists("BuildingPortraits/$buildingName")
 
     fun getPromotionIcon(promotionName: String, size: Float = 30f): Actor {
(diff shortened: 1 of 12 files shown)


yairm210:
Looks cool, but it must be documented in docs/mods somewhere, otherwise how will modders know?

vegeta1k95:
@yairm210 
Added new section to wiki

yairm210:
Not in wiki, that's discrete deprecated and should be removed
Here:
https://github.com/yairm210/Unciv/tree/master/docs/Modders

core/src/com/unciv/ui/images/ImageGetter.kt
```diff
@@ -252,13 +252,41 @@ object ImageGetter {
...
+                group
+            } else
+                getUnitIcon(construction).surroundWithCircle(size)
+        }
+        if (construction == "Nothing")
+            return getImage("OtherIcons/Sleep").surroundWithCircle(size)
+        return getStatIcon(construction).surroundWithCircle(size)
     }
 
+    private fun unitPortraitExists(unitName: String) = imageExists("UnitPortraits/$unitName")
```
yairm210, line 287: The point of the above change was to remove these single-use functions...

yairm210 (changes requested):

vegeta1k95:
@yairm210 
- Generified functions, made extension function
- Removed stuff from index.html
- Added stuff into docs

core/src/com/unciv/ui/images/ImageGetter.kt
```diff
@@ -252,11 +252,24 @@ object ImageGetter {
...
-        if (ruleset.units.containsKey(construction)) return getUnitIcon(construction)
-        if (construction == "Nothing") return getImage("OtherIcons/Sleep")
-        return getStatIcon(construction)
+    fun getPortraitImage(construction: String, size: Float): Group {
+        if (ruleset.buildings.containsKey(construction)) {
+            val buildingPortraitLocation = "BuildingPortraits/$construction"
+            return if (imageExists(buildingPortraitLocation)) {
+                getImage(buildingPortraitLocation).toGroup(size)
+            } else
+                getImage("BuildingIcons/$construction").surroundWithCircle(size)
```
yairm210, line 261:
```suggestion
                getImage(buildingPortraitLocation).surroundWithCircle(size)
```

yairm210 (approved):
Looks good apart from one minor nitpick :)

docs/Modders/Images-and-Audio.md
```diff
@@ -41,6 +41,14 @@ These work best if they are square, between 100x100 and 256x256 pixels, and incl
 
 For example, [here](https://github.com/yairm210/Unciv-leader-portrait-mod-example) is mod showing how to add leader portraits, which can complement the base game.
 
+### Adding Unit and Building Portraits
+
+The base game comes without Unit/Building Portraits (display flags instead by default), but is able to display them in City Screen, Overview Screen, Tech Tree, Civilopedia, etc. A mod can supply these, by adding their images to `/Images/UnitPortraits/` and `/Images/BuildingPortraits`. The file name must correspond exactly with the unit/building name as defined in Units.json and Buildings.json, or they will be ignored.
```
yairm210, line 46:
```suggestion
The base game comes without Unit/Building Portraits (display flags instead by default), but is able to display them in City Screen, Overview Screen, Tech Tree, Civilopedia, etc. A mod can supply these, by adding their images to `/Images/UnitPortraits/` and `/Images/BuildingPortraits/`. The file name must correspond exactly with the unit/building name as defined in Units.json and Buildings.json, or they will be ignored.
```

yairm210:
WAITASECOND
There's no difference between 'buildingIcons' and 'buildingPortraits'
Building icons are already unchanged in-game and are used in the exact same places as 'portraits'
So this change really only needs to include **unit** portraits

docs/Modders/Images-and-Audio.md
yairm210: Actually this doesn't describe what 'flag' vs' portrait' is.

Something like:

The base game uses flat icons, colored to fit the civilization's flag colors, to denote units both on the map and in other UI elements.
A mod can supply "Portraits" - static images that will remain uncolored -  by adding their images to `/Images/UnitPortraits/`, which will be used in all UI elements except for the world map. The file name must correspond exactly with the unit/building name as defined in Units.json and Buildings.json, or they will be ignored.

vegeta1k95:
> WAITASECOND There's no difference between 'buildingIcons' and 'buildingPortraits' Building icons are already unchanged in-game and are used in the exact same places as 'portraits' So this change really only needs to include **unit** portraits

The problem is: building **icons** are monotonic black and surrounded by white circle in code.
If you use proper full RGB **portraits** instead - they are shrinked and put into a white circle, which look bad actually and makes no sense at all.

Example: 
image vs image

In City Screen it looks even worse because icons there are smaller and inner part becomes almost unrecognizable.

The same thing happens for tech icons and resource icons. and even worse, because they are tinted.

yairm210:
Ah I see, good point
In that case just change the docs text and the buildingPortraitLocation and we'll merge this :)

core/src/com/unciv/ui/images/ImageGetter.kt
vegeta1k95: This one is not possible:
`buildingPortraitsLocation` leads to `BuildingPortraits/`
and in else we lead to `BuildingIcons/`

yairm210: Right you are

vegeta1k95:
- Docs fixed
- For buildingPortraitLocation change shouln't be there, see comment on conversation

PR #8109, opened 2022-12-10
Three Continents

FiretronP75:
New generated map Three Continents. Only allows the North or South positions of the 3rd continent because in the East or West positions the 3rd continent is at the equator and would get an advantage of no ice or tundra.


core/src/com/unciv/logic/map/MapParameters.kt
```diff
@@ -131,6 +131,8 @@ object MapType : IsPartOfGameInfoSerialization {
     const val default = "Default"
     const val pangaea = "Pangaea"
     const val continents = "Continents"
+    const val threeContinents = "Three Continents"
+    const val threeContinentsFair = "Three Continents Fair"
```
alexban011, line 135: This might create confusion. Why keep `Three Continents` if one of the continent has a clear advantage? 
I'd say ditch the `Three Continents` and rename the fair one to that, but lets see what other have to say. :)

yairm210: Agree, having an "unfair" mail doesn't sound like a good idea

FiretronP75: Done!

yairm210:
Can you post some pictures of what these Maria look like?

FiretronP75:
3cont-hex-n
3cont-hex-s
3cont-rect-n
3cont-rect-s

yairm210:
There are comments in-code about world wrap for two continents - what does this map look like with world wrap?

yairm210:
But apart from the wrap question, looks good!

FiretronP75:
I copied the world wrap part from two continents or four corners. I don't really understand what that does.

FiretronP75:
OK, I understand what it does now, and tested it, and seems to be working. (world-wrap code)

FiretronP75:
Here's a screen shot (sorry it is 4k)
The world-wrap code makes water on the East and West edge of the continents so they don't touch each other when wrapping.
world wrap

PR #8110, opened 2022-12-10
Two Continents Allow both Lat or Long

FiretronP75:
Rename generated map "Continents" to "Two Continents" and make it choose latitude or longitude land division based on rectangular map dimensions or randomly if equal dimensions or hexagonal.

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
@@ -106,7 +106,12 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
     }
 
     private fun createTwoContinents(tileMap: TileMap) {
-        val isLatitude = randomness.RNG.nextDouble() > 0.5f
+        val isLatitude =
+                if (tileMap.mapParameters.shape === MapShape.hexagonal) randomness.RNG.nextDouble() > 0.5f
+                else if (tileMap.mapParameters.mapSize.height > tileMap.mapParameters.mapSize.width) true
+                else if (tileMap.mapParameters.mapSize.width > tileMap.mapParameters.mapSize.height) false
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
```diff
@@ -141,10 +142,11 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
         return min(0.3, 1.0 - (5.0 * distanceFactor * distanceFactor + randomScale) / 3.0)
     }
 
-    private fun getTwoContinentsTransform(tileInfo: TileInfo, tileMap: TileMap): Double {
+    private fun getTwoContinentsTransform(tileInfo: TileInfo, tileMap: TileMap, isLatitude: Boolean): Double {
```
yairm210, line 145: Rather than having this be random, it should be according to the map dimensions. If width > height split along the middle width, if height > width split along the middle height

FiretronP75: So I coded this, then went to test it, and I can't, because you can only define the map radius not width and height.

FiretronP75: Oh, you can on rectangular just not on hexagonal. I see.

FiretronP75: Done!

yairm210 (approved):

FiretronP75:
2cont-hex-ew
2cont-hex-ns
2cont-tall
2cont-wide

core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
FiretronP75: It is still random when the width and height are the same.

PR #8115, opened 2022-12-11
Raze Special Cities

FiretronP75:
Add new ModOptions uniques: Allow raze capital, Allow raze holy city.


yairm210:
A. Again the 'raze capital' interferes with a victory condition
B. Please use 'if/else's to return false on certain conditions - one big boolean function at the end is much harder to read

FiretronP75:
A - There are already two other things that do this. Do they block the domination victory? Battle.kt 542 (`true` overrides)
```
        if (attackerCiv.isBarbarian() || attackerCiv.isOneCityChallenger()) {
            city.destroyCity(true)
            return
        }

```

FiretronP75:
B - Done!

yairm210:
Now that I think about it we can just check all existing cities that "were original coty" and be done with it
Like, don't go foreach civ, but instead foreach city in all civs

FiretronP75:
Oh you mean just don't worry about the destroyed ones and just keep tracking the still alive capitals as usual? So no further code changes needed?

FiretronP75:
I'm not sure I understand. Happy to do it with further detail.

yairm210:
You know what since this is an "opt in with mod" feature I'm fine with it being in even if I'm not 100% sure of it

PR #8119, opened 2022-12-12
NextTurnActions (pick tech/policy/etc) now can have icons

vegeta1k95:
All current NextTurnActions now can have optional icons displayed along with text on Next Turn button.

This includes:
```
NotificationIcons/PickConstruction
NotificationIcons/PickTech
NotificationIcons/PickPolicy
NotificationIcons/FoundPantheon
NotificationIcons/FoundReligion
NotificationIcons/EnhanceReligion
NotificationIcons/ReformReligion
NotificationIcons/WorldCongressVote
NotificationIcons/MoveAutomatedUnits
NotificationIcons/Working
NotificationIcons/Waiting
NotificationIcons/NextUnit
NotificationIcons/NextTurn
```
All of those could be replaced by mods.
I have added default icons to first three: Construction, Tech and Policy. (But those could be removed and be completely optional also).

It looks like this:
image
image
image


vegeta1k95:
Here is an example with a mod:
image

yairm210:
I think this looks much worse than the existing state
Maybe if you make the images smaller?

In any case this is one if those situations I'll poll the community to decide

vegeta1k95:
@yairm210 
I can remove default ones, leaving the possibility for mods to optionally put those.

vegeta1k95:
@yairm210 
Done. Now no icons by default, by can be added by mods.

yairm210:
I've been trying to narrow down what I don't like and it's that the icons are far too big compared to the text which makes the whole button look weird
What would this look like if the icons were, say, a third of the size?

vegeta1k95:
Here how it looks in 1/3 size:
image
image

vegeta1k95:
So you can merge this one then, it does not contain default icons. But those could be added by mods (as in above screens).
Or do you want me add default ones (small, 1/3) as well?

yairm210:
This is almost fine. We *need to **resize** the images to fit the height of the text in the button, we shouldn't assume that modders know the exact heights, and we want a responsive UI

BUT since there are no default icons, you need to either 
A. add documentation in docs, or 
B. (preferrably) add icons for all existing notifications (just copy-paste the existing icons from the current images - they almost all have obvious ones

```
NotificationIcons/PickConstruction - StatIcons/Production
NotificationIcons/PickTech - StatIcons/Science
NotificationIcons/PickPolicy - StatIcons/Culture
NotificationIcons/FoundPantheon - StatIcons/Faith
NotificationIcons/FoundReligion - StatIcons/Faith
NotificationIcons/EnhanceReligion - StatIcons/Faith
NotificationIcons/ReformReligion - StatIcons/Faith
NotificationIcons/WorldCongressVote - OtherIcons/Politics
NotificationIcons/MoveAutomatedUnits - OtherIcons/ArrowRight
NotificationIcons/Working - EmojiIcons/Turn
NotificationIcons/Waiting - OtherIcons/Loading
NotificationIcons/NextUnit - OtherIcons/ArrowRight
NotificationIcons/NextTurn - EmojiIcons/Turn
```

vegeta1k95:
> (preferrably) add icons for all existing notifications (just copy-paste the existing icons from the current images - they almost all have obvious ones
Done.

Also I switched a little bit an order in which NextTurn actions appear- first choice of Building/Tech/Policy, then iterate Units, as in Civ 5.

core/src/com/unciv/ui/worldscreen/status/NextTurnButton.kt
```diff
@@ -33,13 +39,17 @@ class NextTurnButton(
                nextTurnAction: NextTurnAction? = null) {
         if (nextTurnAction != null) {
             this.nextTurnAction = nextTurnAction
-            setText(nextTurnAction.text.tr())
+            label.setText(nextTurnAction.text.tr())
             label.color = nextTurnAction.color
+            if (nextTurnAction.icon != null && ImageGetter.imageExists(nextTurnAction.icon))
+                iconCell.setActor(ImageGetter.getImage(nextTurnAction.icon))
```
yairm210, line 45: It doesn't look like we're resizing the image to for the text height?

yairm210: (this looks like the last change before merge 😃 )

vegeta1k95: > It doesn't look like we're resizing the image to for the text height?

No, default icons are sized properly to align with text to serve as a reference to a modders.
But there should be a possibility to put different sizes there. I personally prefer big one in my own mods.

yairm210: I see

PR #8128, opened 2022-12-14
Change styles and behaviour of Unit flags as in Civ 5

vegeta1k95:
I have reworked Unit flag system to resemble the Civ 5 style and behaviour.

Now each flag is assembled from the following components:
1) Background 
2) Shadow mask
3) Selection halo

Example, for default flag:
image

Background now colored as in Civ 5 - with a shadow and an outer stroke of second color.

Civilian units now have distinct triangle flags, which really helps with distinguisging among the hordes of units.
Embarked and fortified also had been updated.

Selection marker is now the following: persisten white stroke + (if continious render enabled) a periodic fading halo (fade in/out)

Also: unit flags are dimmed only if units are out of moves and out of attacks (Civ 5 behaviour)! No more dimming of all automated units and fortified units!

Examples:
image
image
image
image
image
image


yairm210:
This clashes heavily with our current "flat" UI style, so I can't accept that
I will accept the dimming part though

vegeta1k95:
Wdym by "flat"?

yairm210:
Our current UI scheme uses single colors - no gradients, shadows, 3d - to give the feeling that the various elements are flat
Your change adds both a gradient and a reflective affect, giving the feeling that the elements are 3-dimentional, and the clash in visual styles is extremely jarring

yairm210:
I don't have a problem with the white dimming as it gets further from the center, if you were wondering

vegeta1k95:
So you have only problem with a shadow mask inside a flag? I can make it optional. Leaving stroke and white selection halo.

yairm210:
I don't want it 'optional but in game' since it ruins the aesthetic
If anyone wants it it should only be through mods
For that we can just use a different circle for the unit, which modders can decide to override with a png of a circle with shadow

yairm210:
We shouldn't need a "unit flag mask" at all if we can just override the flag shape to contain the mask shadow

vegeta1k95:
Here is a version without shadow:
image
image
image

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -4,30 +4,103 @@ import com.badlogic.gdx.graphics.Color
...
+
+        val bgW = size
+        var bgH = size
+
+        if (unit.isFortified())
+            bgH *= 1.1f
+        else if (unit.isEmbarked())
+            bgH *= 1.1f
+        else if (unit.isCivilian())
+            bgH *= 1.2f
```
yairm210, line 40: This is hardcoded and doesn't leave modders much choice to change icons and have everything work
Please calculate according to icon width / icon height
(Also please rename to backgroundWidth and backgroundHeight, we try to keep the code as immediately understandable as possible 😉 )

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -4,30 +4,103 @@ import com.badlogic.gdx.graphics.Color
...
+        flagSelection = getBackgroundSelectionForUnit()
+        flagSelection?.apply {
+            align = Align.center
+            color.a = 0f
+            if (unit.isCivilian() && !unit.isEmbarked())
+                setSize(bgW*2.5f, bgH*2.5f)
+            else
+                setSize(bgW*2f, bgH*2f)
+            center(this@UnitGroup)
+        }
```
yairm210, line 53:
```suggestion
        flagSelection.a = 0f
        flagSelection.setSize(bgW*2.5f, bgH*2.5f)
        flagSelection.center(this)
```

should work. There's no need to use .apply{} here, especially since we use the 'this' internally, doing things explicitly is more readable

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -72,44 +169,39 @@ class UnitGroup(val unit: MapUnit, val size: Float): Group() {
...
         }
 
-        val whiteHalo = getBackgroundImageForUnit()
-        val whiteHaloSize = 30f
-        whiteHalo.setSize(whiteHaloSize, whiteHaloSize)
-        whiteHalo.center(this)
-        addActor(whiteHalo)
-        whiteHalo.toBack()
+        // Unit base icon is faded out only if out of moves
+        if (unit.currentMovement == 0f && !unit.canAttack()) {
```
yairm210, line 182: Units with 0 movement cannot attack by definition
```suggestion
        if (unit.currentMovement == 0f) {
```

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -72,44 +169,39 @@ class UnitGroup(val unit: MapUnit, val size: Float): Group() {
...
+            flagHalo?.color?.a = 0f
+            flagBg?.children?.forEach { it.color.a = 0.5f }
+        } else {
+            unitBaseImage.color.a = 1f
+            flagHalo?.color?.a = 1f
+            flagBg?.children?.forEach { it.color.a = 1f }
+        }
 
+        flagSelection?.apply { color.a = 1f }
+        flagHalo?.apply { color.a = 1f }
```
yairm210, line 193:
```suggestion
        flagSelection.color.a = 1f
        flagHalo.color.a = 1f
```
Why all the complication

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -72,44 +169,39 @@ class UnitGroup(val unit: MapUnit, val size: Float): Group() {
...
+        // Unit base icon is faded out only if out of moves
+        if (unit.currentMovement == 0f && !unit.canAttack()) {
+            unitBaseImage.color.a = 0.5f
+            flagHalo?.color?.a = 0f
+            flagBg?.children?.forEach { it.color.a = 0.5f }
+        } else {
+            unitBaseImage.color.a = 1f
+            flagHalo?.color?.a = 1f
+            flagBg?.children?.forEach { it.color.a = 1f }
+        }
```
yairm210, line 190: Duplicates - we can simpliffy to `alpha = if (unit.currentMovement == 0f) 0.5 else 1` and then set the values only once

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -4,30 +4,103 @@ import com.badlogic.gdx.graphics.Color
...
+
+            val inner = getBackgroundImageForUnit().apply {
+                align = Align.center
+                color = unit.civInfo.nation.getOuterColor()
+                color.a = UncivGame.Current.settings.unitIconOpacity
+                setSize(bgW, bgH)
+            }
+
+            val mask = getBackgroundMaskForUnit()?.apply {
+                align = Align.center
```
yairm210, line 86: We have *double* apply's here, which is even worse. Please flatten to like
 - define group
 - Define group parts
 - Add parts to group

yairm210 (changes requested):
Looks good, but can be vastly simplified

core/src/com/unciv/ui/utils/UnitGroup.kt
vegeta1k95: Done

core/src/com/unciv/ui/utils/UnitGroup.kt
vegeta1k95: Done

vegeta1k95:
- Fully refactored and cleaned, removed all magical constants for unit types. 
- Minor change: action group now uses proper UnitAction icons

yairm210 (approved):

PR #8138, opened 2022-12-15
G&K Neutral Tile Road Maintenance

itanasi:
To align with Civ5 G&K, enabling Roads built outside your territory to cost maintenance.
The uniques for Roads actually are used now. And are parameterize for Stats

Test saves:
Showing 1 Road in territory, 1 outside, costing 2 Main
[RoadMaintTest.txt](https://github.com/yairm210/Unciv/files/10199373/RoadMaintTest.txt)

Several outside of Territory. The Road South of York was originally built by Greece. If you claim the tile, Road Maintenance goes from -3 to -4 since you claim the tile and thus the road.
[RoadMaintTest2.txt](https://github.com/yairm210/Unciv/files/10199375/RoadMaintTest2.txt)

Vanilla game with Road not creating Maintenance cost
[RoadMainVanilla.txt](https://github.com/yairm210/Unciv/files/10199377/RoadMainVanilla.txt)

Making Roads cost Science
image

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -143,6 +143,9 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
     @Transient
     var thingsToFocusOnForVictory = setOf<Victory.Focus>()
 
+    @Transient
+    var neutralRoads = HashSet<Vector2>()
+
     var playerType = PlayerType.AI
 
     /** Used in online multiplayer for human players */
@@ -199,8 +202,6 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
     fun addExploredTiles(tiles:Sequence<Vector2>){
         exploredTiles.addAll(tiles)
     }
-
-    var neutralRoads = HashSet<Vector2>()
 
     var lastSeenImprovement = HashMapVector2<String>()
 
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -200,6 +200,8 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
         exploredTiles.addAll(tiles)
     }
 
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -566,6 +568,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
 
     DefensiveBonus("Gives a defensive bonus of [relativeAmount]%", UniqueTarget.Improvement),
     ImprovementMaintenance("Costs [amount] gold per turn when in your territory", UniqueTarget.Improvement), // Unused
+    ImprovementAllMaintenance("Costs [amount] gold per turn", UniqueTarget.Improvement), // Unused
```
yairm210, line 571: A. We always allow for both old and new unique formats to work for a certain amount of time to allow modders to change existing unqiues
B. If you're replacing a unique, mark it with a @Deprecated so that it's both obvious in the docs and the unique can be automatically converted to the new one

yairm210: Now that I think of it...
If we're changing this unique anyway, why not allow *any* stats? "Costs [stats] per turn"

itanasi: Hookay, done

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -200,6 +200,8 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
         exploredTiles.addAll(tiles)
     }
 
+    var neutralRoads = HashSet<Vector2>()
```
yairm210, line 203: Since this is serialized, this means there are now 2 sources of truth for who owns what - one in the tile and one in the civilizationInfo
To solve this and have one source if truth, this should be transient (not serialized) and built up on setTransients

itanasi: Done

PR #8140, opened 2022-12-15
Flat Earth Hexagonal

FiretronP75:
Add new map shape Flat Earth Hexagonal

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
@@ -512,39 +512,19 @@ class MapGenerator(val ruleset: Ruleset) {
             val humidityRandom = randomness.getPerlinNoise(tile, humiditySeed, scale = scale, nOctaves = 1)
             val humidity = ((humidityRandom + 1.0) / 2.0 + humidityShift).coerceIn(0.0..1.0)
 
-            val randomTemperature = randomness.getPerlinNoise(tile, temperatureSeed, scale = scale, nOctaves = 1)
-            var temperature: Double
-
-            // Flat Earth uses radius from center instead of latitude
-            if (tileMap.mapParameters.shape === MapShape.flatEarth) {
-                val latitudeRatio = abs(tile.latitude) / tileMap.maxLatitude
-                val longitudeRatio = abs(tile.longitude) / tileMap.maxLongitude
-                val radius = sqrt(latitudeRatio.pow(2) + longitudeRatio.pow(2))
-
-                // Radius is in the range of 0.0 to 1.0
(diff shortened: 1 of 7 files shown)


FiretronP75:
FlatEarchArchipelago
FlatEarthFourCorners
FlatEarthPangaea
FlatEarthThreeContinents

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -596,6 +651,146 @@ class MapGenerator(val ruleset: Ruleset) {
...
+
+        // Flat Earth needs a 1 tile wide perimeter of ice/mountain/snow and a 2 radius cluster of ice in the center.
+        if (tileMap.mapParameters.shape === MapShape.flatEarth) {
+
+            // Make sure center tiles and edge tiles are ocean not coast
+            for (tile in tileMap.values) {
+                val isCenterTile = tile.latitude == 0f && tile.longitude == 0f
+                val isEdgeTile = tile.neighbors.count() < 6
+
+                if (isCenterTile || isEdgeTile) {
```
yairm210, line 663: Please flip some if these ifs to reduce nesting...

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -596,6 +651,146 @@ class MapGenerator(val ruleset: Ruleset) {
                     -1f, ruleset.modOptions.constants.spawnIceBelowTemperature,
                     0f, 1f))
             }.toList()
+
+        // Flat Earth needs a 1 tile wide perimeter of ice/mountain/snow and a 2 radius cluster of ice in the center.
+        if (tileMap.mapParameters.shape === MapShape.flatEarth) {
```
yairm210, line 656: You have 150 lines of uninterrupted code here.
Not only does this need its own function, it probably needs subfunctions, maybe its own class

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
FiretronP75: Like `if (!isCenterTile && !isEdgeTile) continue` ?

yairm210: Yup

yairm210:
Apart from minor nitpicks, looks good!
Remember to add the new map type to template.properties so it's translatable

FiretronP75:
Feel free to nitpick all you want. I want to keep the code clean!

FiretronP75:
OK I think it is ready.

yairm210 (approved):

PR #8144, opened 2022-12-15
"Auto Unit Cycle" and "Automated units move on turn start" options implemented (Civ 5)

vegeta1k95:
1) Implemented the default option "Auto Unit Cycle" from Civ 5 (ON by default)
image

If enabled - after unit has done certain actions (like fortify, sleep, spending all movement point, etc.) we automatically switch to the next due unit.

2) Implemented the option "Automated units move on turn start" (default Civ 5 behaviour, but here is optional, default OFF) 

If enabled - instead of manual "Move automated units" button all of the automated units actions will be performed automatically on the start of the next turn (default Civ 5 behaviour)


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -927,7 +928,14 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
         for (city in cities) city.startTurn()  // Most expensive part of startTurn
 
         for (unit in getCivUnits()) unit.startTurn()
-        hasMovedAutomatedUnits = false
+
+        if (playerType == PlayerType.Human && UncivGame.Current.settings.automatedUnitsMoveOnTurnStart) {
+            hasMovedAutomatedUnits = true
+            Concurrency.run("Move automated units") {
+                for (unit in getCivUnits())
```
yairm210, line 935: It's not good to run this in another thread, since the user could manually move units and we'll end up with a conflict

core/src/com/unciv/ui/options/GameplayTab.kt
```diff
@@ -18,6 +18,7 @@ fun gameplayTab(
     val settings = optionsPopup.settings
 
     optionsPopup.addCheckbox(this, "Check for idle units", settings.checkForDueUnits, true) { settings.checkForDueUnits = it }
+    optionsPopup.addCheckbox(this, "Auto Unit Cycle", settings.autoUnitCycle, true) { settings.autoUnitCycle = it }
```
yairm210, line 21: These two texts need to be added to template.properties so they will be translatable

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
vegeta1k95: Do you propose remove threading here or rather add user input blocking measures?

yairm210: Remove threading

vegeta1k95: Done

PR #8147, opened 2022-12-15
Fix minor bug, preventing flag shadows (if added) to appear

vegeta1k95:
Title

### DIFF
--- a/core/src/com/unciv/ui/utils/UnitGroup.kt
+++ b/core/src/com/unciv/ui/utils/UnitGroup.kt
@@ -1,5 +1,6 @@
 package com.unciv.ui.utils
 
+import com.badlogic.gdx.Gdx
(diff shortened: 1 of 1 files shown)


yairm210:
For some unknown reason the flagBg icons are still visible when entering the city screen
Since city screen tiles don't show unit icons by setting 
```kotlin
        unitLayerGroup.isVisible = false
        unitImageLayerGroup.isVisible = false
```
that means that *somehow* the flagBg has left the unitLayerGroup
I have no idea how

image

vegeta1k95:
@yairm210 
This was the case even before my pull request on flags changes (you can check old versions). 
I think I know what is causing it. I'll make a fix and pull request.

yairm210:
Yes this is definitely my bug

PR #8148, opened 2022-12-15
fixed global politics screen showing civ name in non english languages 

### LINKED ISSUE #8141: Feature request: Shortening leader name in overview
Leader name of Overview-Politics is only "name" in English.
But in all other languages, It's "name + nation".
Could you delete the unnecessary nation string? We can see it right below.

<details>
<summary>Details</summary>

English:
overview_english

French:
overview_french
</details>

**Additional context**
One more reason that it's better to be shorten is that it's Civ5 way.
You can see here:
https://youtu.be/5EL0EUmREKE?t=755
It's multi so It shows only player name.
I've checked translated overview also shows only name, but that won't be necessary.

alexban011:
resolves #8141

### DIFF
--- a/core/src/com/unciv/ui/overviewscreen/GlobalPoliticsOverviewTable.kt
+++ b/core/src/com/unciv/ui/overviewscreen/GlobalPoliticsOverviewTable.kt
@@ -125,7 +125,7 @@ class GlobalPoliticsOverviewTable (
 
     private fun getCivInfoTable(civ: CivilizationInfo): Table {
         val civInfoTable = Table(skin)
-        val leaderName = civ.getLeaderDisplayName().removeSuffix(" of " + civ.civName)
+        val leaderName = civ.nation.leaderName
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/ui/overviewscreen/GlobalPoliticsOverviewTable.kt
```diff
@@ -125,7 +125,7 @@ class GlobalPoliticsOverviewTable (
 
     private fun getCivInfoTable(civ: CivilizationInfo): Table {
         val civInfoTable = Table(skin)
-        val leaderName = civ.getLeaderDisplayName().removeSuffix(" of " + civ.civName)
+        val leaderName = civ.nation.leaderName
```
yairm210, line 128: .tr(), this looks like it's currently not translated at all?

alexban011: the line just below this one adds `leaderName.toLabel()` which should translate the string

PR #8149, opened 2022-12-15
Centered Islands

ravignir:
Not sure if the name is most fitting since the results are pretty random. Could as well name it "Mess" :)


alexban011:
Its always a good idea to post screenshots of the maps (hexagonal + rectangular) :D

ravignir:
example

FiretronP75:
Looks great but needs a different name, and then the name in the code should be similar to the name in the GUI. You could use any synonym of random, or use a random human name maybe.

ravignir:
Should it replace default?

yairm210:
What are the advantages/disadvantages of this map generation compared to default?

ravignir:
Unlike default it creates landmass around the center of the map and leaves water on the borders, it is more world wrap friendly. I think land generated by it looks more natural when compared to default. Modifying Water level also gives you more control on whether you prefer single continent (or big ones) or smaller islands. The disadvantage is there is more code.

Here is an example with Water level set to minimum and maximum. New map on the top, default on the bottom.

example

I am still tinkering with it, might also reduce randomness a bit, and probably make tiny 1 tile islands almost impossible to spawn(?).

yairm210:
Sure, signs to me like it can replace the default 
Can you post images of the current default for comparison?

FiretronP75:
Oh, I see why you called it centered islands now. I think Centered Landmasses would be more accurate. Or Watery Edges. But if it can replace Default then it does not need a name.

alexban011:
> Can you post images of the current default for comparison?

He did. Its the 2 images at the bottom of https://github.com/yairm210/Unciv/pull/8149#issuecomment-1356770416

yairm210:
One thing I'm noticing is that there are hardly any lakes

PR #8150, opened 2022-12-15
Fix bug, where Unit flags backgrounds appear on CityScreen tiles

vegeta1k95:
For some reason, there was a code inside TileGroup, which had been forcefully setting visibility of many layers (including unitLayer) to true. Removed.

Was it for some edge-case?

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityTileGroup.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityTileGroup.kt
@@ -1,5 +1,6 @@
 package com.unciv.ui.cityscreen
 
+import com.badlogic.gdx.Gdx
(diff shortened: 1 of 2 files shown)


yairm210 (approved):
How odd
I have no idea what this was for... Perhaps we'll find out later
Good catch!

jmuchemb:
There's a display issue since this commit (I only bisected):
pr8150

Started a game with 4.3.9, moved the warrior to a hill, and I don't see what's behind, but we have the information that e.g. there's a plain on the north.

vegeta1k95:
@jmuchemb Can you load the save using previous version and check if it holds?
@yairm210 Possibly, we should return that part, but remove `unitLayer` from that list.

PR #8153, opened 2022-12-16
Update Filipino.properties

shannaurelle:


saud2410:
Hello, @shannaurelle . Can you re-translate `ConditionalsPlacement =` ?
It's similar to `StartWithCapitalLetter =`, which you translated properly.
Please read comment below from each language file.
> ... The first determines whether the conditionals should be placed before or after the base unique. It should be translated with only the untranslated english word 'before' or 'after', without the quotes. **_Example: In the unique "+20% Strength <for [unitFilter] units>", should the <for [unitFilter] units> be translated before or after the "+20% Strength"?_**

(I know you might not be always available, just wanted to let you know.)

PR #8165, opened 2022-12-17
fixed global politics screen from revealing unknown civs and CityStates

alexban011:
Previously the global politics screen was showing civilizations which weren't known by the viewingPlayer


yairm210:
I'm pretty sure there's a civInfo.knows() function to simplify this

Issue #6373, opened 2022-03-18
Request: (Collapse all / Expand all) toggle button for the city Stats drilldown


Ouaz:
**Is your feature request related to a problem? Please describe.**
As a game progresses, the city stats drilldown column becomes very long, and it is very unconvenient to have to scroll the whole list to check particular items in the middle or at the bottom of the list. Or to close all stats one by one each time a save is loaded, to make easier to pick the desired one afterwards.

**Describe the solution you'd like**
Add a [Collapse all / Expand all] toggle button above the Stats drilldown. This way, it would be easier to access the desired city stats in two clicks.

![collapse_all_button](https://user-images.githubusercontent.com/11946570/159070621-e68eabdc-31fd-450f-a9e2-cd236043d399.png) ![expand_all_button](https://user-images.githubusercontent.com/11946570/159070656-65d3983a-96f3-44fb-9233-e8b879748d9d.png)

**Describe alternatives you've considered**
Make the collapsed Stats drilldown the default display, but:
- The list is short at the beginning of a game, so no need to collapse it.
- Some players may not like that. ^^

Thanks for reading!

SomeTroglodyte:
... and when some are expanded some are collapsed we'd get _two_ buttons? :grin: You know, once I hoped we'd persist the open/close states of these across runs, then your usecase wouldn't weigh so much. But yes I get it...

But now you screenshotted it I notice the little discrepancy thanks to that aggressive Stat iconizing - the expanders without icon stand out a little. Do we have icons for "specialist buildings" as opposed to other buildings?

Ouaz:
> ... and when some are expanded some are collapsed we'd get _two_ buttons? :grin:

Haha, I didn't think of that case. But actually... why not? haha :
![two_buttons_haha](https://user-images.githubusercontent.com/11946570/159142612-312bc2d9-cf42-43a3-a3a5-97ff07702599.png)

So, even if some stats are collapsed, and others expanded, the player chooses what he wants to do. :grin:

Ouaz:
Tested it, works wondefully! Thank you!

PR #8167, opened 2022-12-18
Rework of City Screen: new current buildings list + misc changes

vegeta1k95:
City Screen has been partially reworked:

1) Button "Show drilldown stats" removed. Drilldown stats (buildings) are inserted into right city info panel as expander tabs.
image

2) Categories which were inserted have been given different look (Civ 5 style):
image

Each item shows: icon, name, output stats and specialists slots.
Specialists slots are now showed as filled, if enough of corresponding specialists of that type is active in the city.
image

3) "Sell" building button has been moved to bottom-right tooltip which is called by clicking on a building in the list:
image

Also misc. changes were made:
- Now all tiles in CityScreen show yields, this is default Civ 5 behaviour and is very helpful when considering which tiles to buy next (without it, you need to click manually every tile to see yields)
- Resources icons in CityScreen now are not dimmed.


vegeta1k95:
> Now all tiles in CityScreen show yields, this is default Civ 5 behaviour and is very helpful when considering which tiles to buy next (without it, you need to click manually every tile to see yields)

Okay, I figured out this is the case in Civ 5 only when yields are enabled on a global map. If global map yields are disabled then in City Screen only city tiles show yields, not those outside.

I will implement this in the next commit

yairm210:
We can show all yields of tiles in a 3-tile radius even though that's not what Civ V does, since it helps

yairm210:
Anyway, checked this out myself, looks AMAZING!!!

yairm210:
Wow what an upgrade
I never would have thought to do this myself

yairm210:
Only one thing is not 100% perfect and that's that the specialist-providing buildings should always have the specialists colored, so the user knows what they provide, and not only if they're actually in use

Ouaz:
Interesting improvement... but:

1) you removed the "Show All/Collapse all" toggle button. :(

Then again, we have to scroll the whole list (which can be very long) when wanting to select something at the bottom (or collapse each section one by one)

The button was added by this commit: https://github.com/yairm210/Unciv/pull/6441, based on request https://github.com/yairm210/Unciv/issues/6373

2) So, the detailed stats per city are gone? It allowed to check a policy or belief was applied to a particular city. Now, we just have global stats per city in the City Overview screen.

jmuchemb:
I fail to see how this PR improves things. The main thing I was looking at when clicking on "Show drilldown stats" were stat calculations and I don't see them anymore.

On the other side, the other boxes have almost no value for me. In most cases:
- looking at what I can build is enough to deduce what to build
- I almost only sell buildings when razing a city

core/src/com/unciv/ui/cityscreen/ConstructionInfoTable.kt
```diff
@@ -78,11 +89,36 @@ class ConstructionInfoTable(val cityScreen: CityScreen): Table() {
...
-            val link = (construction as? IRulesetObject)?.makeLink() ?: return
-            if (link.isEmpty()) return
-            touchable = Touchable.enabled
-            onClick {
-                UncivGame.Current.pushScreen(CivilopediaScreen(city.getRuleset(), link = link))
+            // Show sell button if construction is a currently sellable building
+            if (construction is Building && cityConstructions.isBuilt(construction.name)
+                    && construction.isSellable()) {
+                val sellAmount = cityScreen.city.getGoldForSellingBuilding(construction.name)
+                val sellText = "Sell [$sellAmount] " + Fonts.gold
```
Ouaz, line 96: `Sell` This line is not translated in other languages.

sell

yairm210:
I didn't even notice the stat drilldowns weren't there, you're right, that's the reason that panel originally existed in the first place
The stat drilldown expander tabs should be before the building tabs

yairm210:
All right, they're back

PR #8168, opened 2022-12-18
Refactor Cannot Move Unique

### LINKED ISSUE #8166: "Cannot move" unique causing multiple issues - cannot build or upgrade unit with unique
Windows 10, 64-bit, Unciv Version 4.3.9-patch1 from itch.io

If you give a non-air unit the unique "Cannot move" and try to build it, you will get a message saying that the unit has been built, but then the unit will never appear.
If you try to upgrade a unit without that unique into a unit that has the "Cannot move" unique, the upgrade will not happen. The system will act as if it happened, but then the unit will be exactly the same as it was before, and you won't lose any money.
If you give an air unit the "Cannot move" unique, you won't be able to build it at all. The menu will say "No space to place this unit", even when there are no air units currently in that city.

I was trying to build a "Missile Defense System" air unit that acts like an extra bombardment for the city and cannot be moved out of the city under any circumstances, but in the process, I discovered this bug.

I've attached a simple mod that introduces some units with the unique. You can attach this mod to the base game and experiment with it from there.
[StuckAirUnit.zip](https://github.com/yairm210/Unciv/files/10252989/StuckAirUnit.zip)

itanasi:
Fixes #8166

### DIFF
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -477,7 +477,6 @@ class MapUnit : IsPartOfGameInfoSerialization {
     }
 
     fun getMaxMovementForAirUnits(): Int {
-        if (hasUnique(UniqueType.CannotMove)) return getRange()  // also used for marking attack range
         return getRange() * 2
     }
 
@@ -1001,7 +1000,7 @@ class MapUnit : IsPartOfGameInfoSerialization {
         updateVisibleTiles()
     }
 
-    fun putInTile(tile: TileInfo, buildCheck: Boolean = false) {
+    fun putInTile(tile: TileInfo) {
         when {
             !movement.canMoveTo(tile) ->
                 throw Exception("Unit $name at $currentTile can't be put in tile ${tile.position}!")
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -1001,9 +1001,9 @@ class MapUnit : IsPartOfGameInfoSerialization {
         updateVisibleTiles()
     }
 
-    fun putInTile(tile: TileInfo) {
+    fun putInTile(tile: TileInfo, buildCheck: Boolean = false) {
         when {
-            !movement.canMoveTo(tile) ->
+            !movement.canMoveTo(tile, buildCheck = buildCheck) ->
                 throw Exception("Unit $name at $currentTile can't be put in tile ${tile.position}!")
             baseUnit.movesLikeAirUnits() -> tile.airUnits.add(this)
             isCivilian() -> tile.civilianUnit = this
--- a/core/src/com/unciv/logic/map/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/MapUnit.kt
@@ -1003,7 +1003,7 @@ class MapUnit : IsPartOfGameInfoSerialization {
 
     fun putInTile(tile: TileInfo, buildCheck: Boolean = false) {
         when {
-            !movement.canMoveTo(tile, buildCheck = buildCheck) ->
+            !movement.canMoveTo(tile) ->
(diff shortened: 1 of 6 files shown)


yairm210:
I think this is the wring way to go around it.

Since we use 'canMoveTo' as essentially 'canEnter', we should have that function ignore the "cannot move" unique entirely
We can use that unique only when calculating tilesToMoveTo this turn, and everything else should follow

itanasi:
tilesToMoveTo? Can't find such a function.

canMoveTo is used a lot, but that's why it likely needs to be there. This blocks the UI from painting extra tiles. This block unit swapping. This block unit withdrawal or melee post-battle movement. The glitch was that it's also used when spawning in a new unit, and we need an exception for that.

itanasi:
Looking at this again, I really don't see a better solution.

Otherwise we need to cover things like the automation code separately which queries canMoveTo directly and then moveToTile. And handle withdrawal and move after attack separately (shouldn't happen since these should never be melee hopefully? But you never know).

We don't have a function that just returns the valid tiles.

yairm210:
Sure we do, in unit.movement, the function that calculates tiles we can move to in this turn
I'll find it later, but this should not be in "can this unit exist here" but in "can this unit travel here"

itanasi:
Right, getTileToMoveToThisTurn, canReach, getReachableTilesInCurrentTurn, but those functions aren't utilized in the same way as canMoveTo. There's a lot of forced movement that bypasses using it. And canMoveTo is called before canReach as a check, so we're wasting performance

itanasi:
Anyway, will try to disentangle this again

yairm210:
Forced movement is relatively rare, and when it's needed then also units that can't move should be affected, (e.g. expelled from borders) so that sounds right to me

itanasi:
Alright, now moved to all movement code. Also checked for Melee's following, can spawn

core/src/com/unciv/logic/map/MapUnit.kt
```diff
@@ -1001,7 +1001,7 @@ class MapUnit : IsPartOfGameInfoSerialization {
         updateVisibleTiles()
     }
 
-    fun putInTile(tile: TileInfo) {
+    fun putInTile(tile: TileInfo, buildCheck: Boolean = false) {
```
yairm210, line 1004: Remove parameter

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
```diff
@@ -336,9 +336,9 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
         return getDistanceToTiles().containsKey(destination)
     }
 
-    fun getReachableTilesInCurrentTurn(): Sequence<TileInfo> {
+    fun getReachableTilesInCurrentTurn(attacking: Boolean=false): Sequence<TileInfo> {
```
yairm210, line 339: HMMMM
There must be a better solution than adding another input parameter
Is this for attackable tiles? If so we can put the logic there

itanasi: This is specifically so that we can paint the attackable tiles since that code utilizes this to give a subset of tiles to check

yairm210: Then that code should, if the unit cannot move, use neighbors instead
We always prefer to move die to more specific areas rate than more general ones

core/src/com/unciv/logic/map/MapUnit.kt
itanasi: Whoop

core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
itanasi: Done

itanasi:
Done

PR #8174, opened 2022-12-18
WIP: Remember when a city was founded, extra information in city screen

jmuchemb:
city-misc-info

For compatibility with existing saves, "Founded in ???" is displayed when date is unknown.

"Demanded Resource" is only displayed during WLTK.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -214,20 +214,20 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
         return gameParameters.espionageEnabled
     }
 
-    private fun getEquivalentTurn(): Int {
+    private fun getEquivalentTurn(turn: Int): Int {
         val totalTurns = speed.numTotalTurns()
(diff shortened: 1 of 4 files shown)


jmuchemb:
This is related to #8167

It's been a while I have these changes locally and I wanted to finish other things before pushing.

Anyway now, I'm not sure where to display these information. What was great with the "Show drilldown stats" is that we could burry information we rarely look at, but still being able see them with a single click (whereas now, I fear several clicks are needed to expand several boxes we usually kept collapsed).

Maybe an extra box on the right, last position, if it's not too much.

yairm210:
We can put as many boxes in the right side as we want :)
But if the city is demanding a resource that's already directly displayed in the main stat box

jmuchemb:
>  But if the city is demanding a resource that's already directly displayed in the main stat box

During WLTK, I only see the number of remaining turns. I'm unable to find *which* resource.

jmuchemb:
... well, I thought that WLTK would stop if we don't have the demanded resource anymore. I did a test and it looks like I was wrong. So knowing which resource would be useless.

PR #8179, opened 2022-12-19
Make map editor respond to max zoom setting.

FiretronP75:


FiretronP75:
We can finally see a huge map in the editor... and screen shot it too.
ZoomMap6

FiretronP75:
ZoomMap6Rect

PR #8180, opened 2022-12-19
Rework of City Screen: new Great People list + misc changes

vegeta1k95:
1) Stats panels "Production/Food/etc" have been moved under a single big expander "Detailed stats":
image

2) Stats panels "Great Artist/Engineer/etc. points" were removed. Instead there is one category "Great People" now, which looks as follows:
image
It shows a list of GP which have positive GPP yield in this city, along with progress bar of (GPP current / GPP needed).

In my opinion, this UI rework will serve as a good base when we will be moving onto local (city) GP and not global (civ) GP, like it is now.

3) There are new emojis (font icons) for GP/GPP, which can be used in texts/translations
image
This includes: specialists table, where I added +3 GPP yields to each, since specialists produce those.
image

4) Building and Wonders on the right are sorted alphabetically now.
5) City Screen tile yield now work as in Civ5: 
If WorldScreen yields are ON - City Screen yields are displayed everywhere.
If WorldScreen yields are OFF - City Screen yields are displayed only inside boundaries.


yairm210:
If 'detailed stats' gets a subcategory, then so should 'buildings', with the 3 building types under it (change existing buildings to 'other' or something else already translated)

yairm210:
Instead of having these icons as text rendering we could have it as a table of texts and images, taking the existing "great person" images and coloring them golden, and that way it's moddable

yairm210:
I don't think you understood what I meant about the great person points.
But it doesn't matter, we'll deal with that when we get there.

vegeta1k95:
@yairm210 
Big UI update here: completely new "Detailed stats" popup
1) Detailed stats expander has been removed from right menu. Now instead there is a button "Stats" near the top, which opens new detailed stats table popup:
image

2) This popup includes all the information from the previous stat drilldown, but in a better and convenient way.
For instance, top left "+" expands all the inner entries:
image

3) If pressed on the table header with a stat (top row: Prod/Gold/Food/etc) popup truncates all the information except for the relevant stat chosen:
image
One can go back to full view by clicking header again.

4) If pressed on a name entry (first column), table will highlight all row with the same name in all categories:
image
Again, to deselect - just tap on it again.

All of the above can be combined, e.g.:
image

yairm210:
Looks good to me, I'm merging the new stat drilldown in an "experimental" status - we'll let people play around with it, and we'll set up a poll in, say, a week.
If people say that they prefer the old "stats in scroll" (because "I can get to the data I want in one click instead of three" - which is a valid UX point) then we'll go back to that

yairm210:
Since you regenerated the game.png on both PRs they now conflict - could you regenerate this one?
If not I'll deal with it when I can

vegeta1k95:
@yairm210 
I usually do it manually, is there automatic tool to regenerate atlases?

yairm210:
Yes, just delete the game.png, run the game and it'll rebuild :)
It also rebuilds when it sees there's a newer file, but in this case that won't help

vegeta1k95:
I thought it rebuilds atlas only when using desktop version, no? I run project from Android Studio and assembled .apk doesn't contain rebuilt atlas.

yairm210:
Ahhh yes that is the case, I automatically assume that anyone coding is working from a PC...

Ouaz:
1)

`totalTable.add("Base stats".toLabel().apply`

In the new city stats panel, "Base values" should be used (instead of Base stats) as it is already in present `template.properties.`
Currently, "Base stats" is not translated.

2) Could it be possible to keep the stats table header always displayed, even when scrolling down? Because the list can be pretty long, it would be good to always see the column header.

header

vegeta1k95:
1) I'll put that in the next commit
2) It could be done with some work, I'll see what I can about it

PR #8184, opened 2022-12-19
showed unknown civs as `Unknown civilization` in GP and Diplomacy screens

alexban011:
Implements discord suggestion.

Instead of not displaying the text at all, it now displays "At war with an unknown civilization", and such texts.

I apologize for the commits.. I forgot to update the project and change the branch :D Luckily the commits get squashed when merged :)


android/assets/jsons/translations/template.properties
```diff
@@ -262,6 +262,8 @@ Personality =
 Influence = 
 
 Ally: [civilization] with [amount] Influence = 
+Ally: Unknown civilization with [$allyInfluence] Influence =
```
yairm210, line 265: This is unnecessary since we can use the above line with "Unknown civilization" in place of the civ name

android/assets/jsons/translations/template.properties
```diff
@@ -1231,10 +1233,14 @@ Politics =
 Show global politics = 
 Show diagram = 
 At war with [enemy] = 
+At war with an unknown civilization =
```
yairm210, line 1236: Same for all of these

core/src/com/unciv/ui/overviewscreen/GlobalPoliticsOverviewTable.kt
```diff
@@ -174,7 +174,10 @@ class GlobalPoliticsOverviewTable (
         // wars
         for (otherCiv in civ.getKnownCivs()) {
             if(civ.isAtWarWith(otherCiv)) {
-                val warText = "At war with [${otherCiv.civName}]".toLabel()
+                val warText = if (viewingPlayer.knows(otherCiv))
```
yairm210, line 177: Instead of these,how abou a function 'getCivName', that returns the name if known and "Unknown civilization" if unknown
You have 3 usages in this file already

alexban011: Done :) I've also tested the translations

image

core/src/com/unciv/ui/overviewscreen/GlobalPoliticsOverviewTable.kt
```diff
@@ -168,16 +169,21 @@ class GlobalPoliticsOverviewTable (
...
         val politicsTable = Table(skin)
 
         // wars
         for (otherCiv in civ.getKnownCivs()) {
             if(civ.isAtWarWith(otherCiv)) {
-                val warText = if (viewingPlayer.knows(otherCiv))
-                    "At war with [${otherCiv.civName}]".toLabel()
-                else
-                    "At war with an unknown civilization".toLabel()
+                println(getCivName(otherCiv))
```
yairm210, line 185: Remove

yairm210 (approved):

core/src/com/unciv/ui/overviewscreen/GlobalPoliticsOverviewTable.kt
yairm210: I'll do it

alexban011: Oops I can't believe that slipped away after reading the changes at least 2 times :))

PR #8188, opened 2022-12-20
Huge update of Technology Picker screen UI (Civ 5), new original icons + misc changes

vegeta1k95:
1) Main feature: full update of TechPicker UI design, now it looks like exactly as Civ 5 tech tree, e.g:
image
image
image
image

In comparison, Civ 5:
image

2) New original icons for "Remove Forest/Jungle/Marsh/..." improvements, instead of just "Stop" sing
image

3) Perpetual Constructions (Gold, Science, Nothing) now have their own separated icons, for better moddability.
OtherIcons/ConvertGold
OtherIcons/ConvertScience 
OtherIcons/ConverNothing (original one)

4) New icon for uniques, it is used inside a tech tree (Civ 5 style)
image


core/src/com/unciv/ui/images/ImageGetter.kt
```diff
@@ -238,17 +239,21 @@ object ImageGetter {
...
-        return iconGroup.surroundWithThinCircle()
+        return if (withCircle) {
+            val group = icon.surroundWithCircle(size)
+            val improvement = ruleset.tileImprovements[improvementName]
+            if (improvement != null)
+                group.circle.color = getColorFromStats(improvement)
+            group.surroundWithThinCircle()
+        } else {
+            icon.toGroup(size)
+        }
```
yairm210, line 256:
```suggestion
        if (!withCircle) return icon.toGroup(size)
        
        val group = icon.surroundWithCircle(size)
        val improvement = ruleset.tileImprovements[improvementName]
        if (improvement != null)
            group.circle.color = getColorFromStats(improvement)
        return group.surroundWithThinCircle()
```

yairm210:
This PR is big, and has good changes and bad changes.

The good:
- Lines connecting techs
- Icons for "remove X"

The Bad:
- The look of the tech buttons. The block containing what the tech gives just looks *off*.

I'm not sure I'm sold on the new color scheme for completed techs but I'm not against it either.


Are you in our Discord server? I'll ask the community what they think of this

yairm210:
image

Note from the community - era borders can be mistaken for tech requirement lines, make these parts darker or remove the line entirely

vegeta1k95:
I'm on Unciv discord: TrollWarlord

yairm210:
Played around with some of the total conversion mods to see what's up
image

image

If we have more than 5 things that the tech provides we should be ignoring the last ones

yairm210:
But apart from that, looks good!

yairm210:
Since the colored lines have less contrast with the background, they look like they're thinner - maybe we can make the lines in the current path thicker to compensate?

vegeta1k95:
I'll commit fixes with lines and 5+ icons when I come home

yairm210:
No rush, we're all doing this on our spare time :)

yairm210 (approved):
Instead of each time checking if we've reached 5, we could create all icons (performance is negligible) and take only the first 5, that way if we have many icon types we don't need to ensure every one of them is guarded

But apart from that minor nitpick, looks mergable!

vegeta1k95:
@yairm210 
Done.
Now it looks like this:
image

yairm210:
Just one more thing - credits.md should contain credits for all the new images you added

vegeta1k95:
Here you go

PR #8192, opened 2022-12-20
Update wrap water generation of two continents to support tall maps.

FiretronP75:


FiretronP75:
Before:
before
After:
after

PR #8193, opened 2022-12-20
Clean up world wrap water generation

FiretronP75:
Update wrap water generation to only consider longitude as there is no latitude wrap.


FiretronP75:
I did a lot of testing.

PR #8203, opened 2022-12-22
Add new game option: No City Razing.

FiretronP75:


FiretronP75:
This option was on the Civ5 new game advanced setup screen. I am still allowing the override by barbarian and one city challenger.

PR #8208, opened 2022-12-22
Added gradle task of Atlas texture packing for Android build

vegeta1k95:
New gradle task for packing texture atlases for Android builds.
It is basically a copy of desktop ImagePacker but for a build-script.

### DIFF
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -1,5 +1,6 @@
 import com.unciv.build.BuildConfig
 import java.util.*
+import com.unciv.build.AndroidImagePacker
 
 plugins {
     id("com.android.application")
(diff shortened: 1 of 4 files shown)


yairm210:
We *intentionally* leave this out of Android to minimize the apk size, since one of our targets is "minimal size for minimal phones"

Kuldebar:
Baldur's Gate Enhanced has an additional 800mb download after initial playstore install. Chromebook Android gamers are increasing. 

Not trying to be argumentative, but even my 30 dollar android go smartphone has 4gb free ATM but I'd never play games on it.

vegeta1k95:
> We _intentionally_ leave this out of Android to minimize the apk size, since one of our targets is "minimal size for minimal phones"

What do you mean? How is APK size influenced by this?

yairm210:
Ah I see, I misunderstood

yairm210:
> Baldur's Gate Enhanced has an additional 800mb download after initial playstore install. Chromebook Android gamers are increasing.
> 
> Not trying to be argumentative, but even my 30 dollar android go smartphone has 4gb free ATM but I'd never play games on it.

About a year ago I was in this position, working with a phone where I needed to remove an app if I wanted to download a new one. And my phone was relatively good compared to a lot of the ones I see ruining Unciv in the Google Play stats.
Anecdotes are not data either way, but this *is* a real issue.

PR #8211, opened 2022-12-23
Added keyboard shortcut to toggle World Screen UI elements

letstalkaboutdune:
Leaving as a draft since it could use a bit more polish.


android/assets/jsons/translations/template.properties
```diff
@@ -1422,6 +1422,7 @@ Pillaging this improvement yields [stats] =
 Pillaging this improvement yields approximately [stats] = 
 Needs removal of terrain features to be built = 
 Unit type = 
+Toggle UI (World Screen only) =
```
alexban011, line 1425: the check failed because this line needs a space after `=` :)

letstalkaboutdune: That's funny. I know about the trailing space requirement and thought I put it in. =P

letstalkaboutdune:
One thing I notice is that after hitting "Next Turn," all the UI comes back. Looks like there's a lot of important code running after "Next Turn" is clicked and I don't' want to touch it.

letstalkaboutdune:
Sorry for the last 2 commit/reverts, didn't mean for that commit to go in this PR.

yairm210:
I wonder if this should have a settings option, for Android / player who accidentally pressed and don't know what they did

letstalkaboutdune:
Possibly. I didn't think these key combos were possible on Android without an external keyboard.

alexban011:
They're not, which is why this could be an option so Android can also use this feature.

It could be an option which adds a button somewhere on the screen and that button toggles the UI

PR #8215, opened 2022-12-23
Remove "Experimental" warning text

letstalkaboutdune:
The "Experimental" warning text, especially for world wrap, can be removed. They are unnecessary at this point after the features have been working well for many months.

Experimental

### DIFF
--- a/core/src/com/unciv/ui/options/AdvancedTab.kt
+++ b/core/src/com/unciv/ui/options/AdvancedTab.kt
@@ -53,7 +53,7 @@ fun advancedTab(
     addAutosaveTurnsSelectBox(this, settings)
 
     optionsPopup.addCheckbox(
-        this, "{Show experimental world wrap for maps}\n{HIGHLY EXPERIMENTAL - YOU HAVE BEEN WARNED!}",
+        this, "{Show world wrap for maps}",
         settings.showExperimentalWorldWrap
     ) {
         settings.showExperimentalWorldWrap = it
--- a/core/src/com/unciv/ui/options/DisplayTab.kt
+++ b/core/src/com/unciv/ui/options/DisplayTab.kt
@@ -36,7 +36,7 @@ fun displayTab(
     optionsPopup.addCheckbox(this, "Show resources and improvements", settings.showResourcesAndImprovements, true) { settings.showResourcesAndImprovements = it }
     optionsPopup.addCheckbox(this, "Show tutorials", settings.showTutorials, true) { settings.showTutorials = it }
     optionsPopup.addCheckbox(this, "Show pixel improvements", settings.showPixelImprovements, true) { settings.showPixelImprovements = it }
-    optionsPopup.addCheckbox(this, "Experimental Demographics scoreboard", settings.useDemographics, true) { settings.useDemographics = it }
+    optionsPopup.addCheckbox(this, "Show Demographics scoreboard", settings.useDemographics, true) { settings.useDemographics = it }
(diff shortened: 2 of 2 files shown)


yairm210:
If they're no longer experimental then we should be enabling them entirely :)
Remove the checkbox, remove the if, remove the setting

letstalkaboutdune:
> If they're no longer experimental then we should be enabling them entirely :) Remove the checkbox, remove the if, remove the setting

I agree for the world wrap. The thing that scared me was a comment in the code about needing to leave it as an option in case Android users got ANR errors and didn't know why.

For the demographics I actually disagree. The different ways of presenting the information can be helpful at different points in the game, so I'd like to leave the option.

ravignir:
I agree with Yair, there is no point in having two demographics scoreboards... the old one is superior as it simply gives player much more info.

The new one still needs to be improved i.e, you should be able to see the exact score of nations along with all the things that contribute to it. Or maybe move score to separate screen.
dds

letstalkaboutdune:
Fair enough, would be nice. However I have zero interest in taking on that big of a coding and UI design project.

Ouaz:
> I agree with Yair, there is no point in having two demographics scoreboards... the old one is superior as it simply gives player much more info.

Eeer... no.

The alternate demographic scoreboard was implemented because the old one gives **too much** info on how the player performs compared to other civs (basically all values of all civs). https://github.com/yairm210/Unciv/pull/6645

The purpose of the new scoreboard was to give rough estimations, with the best and the worst values, and an average value. What's the point of knowing **exactly** all the data of others civs, at any time? (military strength for example).

Removing the alternate Demographics scoreboard is a big regression, imo.

ravignir:
> > I agree with Yair, there is no point in having two demographics scoreboards... the old one is superior as it simply gives player much more info.
> 
> Eeer... no.
> 
> The alternate demographic scoreboard was implemented because the old one gives **too much** info on how the player performs compared to other civs (basically all values of all civs). #6645
> 
> The purpose of the new scoreboard was to give rough estimations, with the best and the worst values, and an average value. What's the point of knowing **exactly** all the data of others civs, at any time? (military strength for example).
> 
> Removing the alternate Demographics scoreboard is a big regression, imo.
> 
> EDIT:
> 
> "World Wrap" is activated by default now? It means we can't disable it now? Eeeer again... players with low end devices will be happy. -_-' (I have to check on my tablet, but "World wrap" made the scrolling way more laggy for me...)

i agree again. New demographics needed improvement.
I have not said anything about deleting it, but improving it instead (and then remove the old rankings providing too much info ).

letstalkaboutdune:
@yairm210 you just straight up deleted all my Demographics dashboard work? What the hell?!?!?

FiretronP75:
Maybe just rename the demographics option to something more accurate? Or maybe enable it with a mod option rather than completely delete?

yairm210:
@Ouaz The *ability* to create world wrap maps is available to all. It's still an option, default is false, and there's a textual warning about performance

As for Demographics dashboard, I somehow 
completely misunderstood the entire point of the PR, that's on me, reverted

Sorry again 🙏🏿

letstalkaboutdune:
> As for Demographics dashboard, I somehow completely misunderstood the entire point of the PR, that's on me, reverted
> 
> Sorry again 🙏🏿

Thank you for acknowledging and reverting!

Ouaz:
> The ability to create world wrap maps is available to all. It's still an option, default is false, and there's a textual warning about performance

Ah, I didn't see the new "World Wrap" checkbox when creating a new game.

<details>
<summary>My time to be sorry. 🙏🏿</summary>
"This is better than you deserve, but let it not be said that I am an unfair man."
</details>

PR #8218, opened 2022-12-24
Big rework of Policy Picker UI, added new ninepatch RoundedEdgeRectangle

vegeta1k95:
1) Completely new Policy Picker UI:
image
image

2) Branch Starter and Finisher policies descriptions have been moved to branch drowdown info:
image

3) Each policy now have (default added, mods can add/change to their own) an icon displayed, taken from `PolicyBranchIcons/`
image

4) Corrected Rationalism and Autocracy columns/rows layout accordingly to Civ 5
5) Branches are now in Civ 5 "snake" ordering: left to right, then right to left,
6) New shape added to skin - roundedEdgeRectangleSmall, which has smaller corner radius (used everywhere in new Policy UI)


yairm210:
If we're working on the policy UI, let's remove the current restriction for "only 3 rows" - that's a current limiting factor

vegeta1k95:
There is no restrictions now for "only 3 rows" now. Policy trees can be very long and very high

PR #8224, opened 2022-12-24
Corrected Civilizations colors (Civ 5)

vegeta1k95:
Civilizations colors corrected to Civ 5 values


yairm210:
Can you show "after" pictures of Korea in the nation picker screen now? I remember that some of the nations (Korea? and one other?) we slightly altered for text readability in the nation picker screen
It's possible the correct solution is "redo the nation picker UI" but at least let's make sure that this current change doesn't make anything unreadable

vegeta1k95:
image
image

Issue #8099, opened 2022-12-06
Feature request: Update UI for each AI civ that does its turn


hackedpassword:
The further into the game one gets, and especially with more [computer] opponents, hitting Next Turn grinds. Wouldn't it be smarter, and far more lively, to update the field per each opponent turn completion, rather than the aggregate of everything done? Applying this to an auto-spectator feature would be cool to watch too.

btw contributor devs, great work on this game. :D

AdityaMH:
Also you can't see animation in spectators mode

yairm210:
What devilry is this?!

hackedpassword:
The proposed demo is a legit UX improvement.

AdityaMH:
> demo
>  Peek.2023-05-17.21-10.mp4





> The proposed demo is a legit UX improvement.

actually no, it's just bar progress. ~~you still can see your unit or enemy unit disappear with no reason.~~

yairm210:
... And? That's still an improvement, he's right

yairm210:
No I meant black magic, and it's definitely better than the "no info whatsoever" we have now
I assume it updates every time we finish automating a civilization?

AdityaMH:
> Just using one image "currently we have" and create ai moving/attacking cinematic by code. like player battle we have.
revival 2 (j2me), civ 3 mobile (j2me), civ5 mobile low screen (128x128) using that method.

hackedpassword:
Render state change per civ upon New Turn
#3242

AdityaMH:
![Image](https://github.com/user-attachments/assets/8c676308-beba-4059-ae36-779760d6c888)

PR #8234, opened 2022-12-26
Processing Turn for Civ indicator, refactored UncivGame and WorldScreen class

vegeta1k95:
1) I refactrored and demangled some code inside UncivGame, WorldScreen and GameInfo classes, mostly just moving and renaming functions between these, also commented.

2) I have added turn status indicator to a bottom of the screen, which shows which civ is currently doing it's business:
image
image

Those civs, that are met shown with flag and name. Those that are not - as Unknown civilization with "?"


AdityaMH:
I think would be better if Civ indicator placed to "end turn" box.
~~to avoid too much civ5 cloning~~

yairm210:
There is Far Too Much in this one PR. 

What I can see is the following categories:
- Rename from isPlayer to isHuman - ok
- NextTurnButton changes - ok
- Barbarian icon - ok
- isTurnActive rename - I'm against, this makes it less clear what this indicates. The first question that pops up to someone seeing this is "is there such a thing as an inactive turn?" Instead, most of these usages should be replaced with worldScreen.canChangeState
- I agree with aditya that we don't need the next turn widget since we can just update the icon in the next turn button

As for the rest, the 'meat' of the changes, that requires comments inside

Again, this PR is too big. I'll add code comments, but please split this up - even one for 'trivial changes' like renames and one for the bulk will make a difference.

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -715,6 +715,18 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
...
 
+    fun doMoveAutomatedUnits() {
+        for (unit in units)
+            unit.doAction()
+        hasMovedAutomatedUnits = true
+    }
+
+    fun canMoveAutomatedUnits(): Boolean {
+        return !hasMovedAutomatedUnits && units.any {
+            it.currentMovement > Constants.minimumMovementEpsilon
```
yairm210, line 726:
```suggestion
            it.currentMovement > 0f
```

Should never be epsilon since we deal with that in the unit movement

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -715,6 +715,18 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
...
     }
 
+    fun doMoveAutomatedUnits() {
+        for (unit in units)
+            unit.doAction()
+        hasMovedAutomatedUnits = true
+    }
+
+    fun canMoveAutomatedUnits(): Boolean {
+        return !hasMovedAutomatedUnits && units.any {
```
yairm210, line 725: if hasMovedAutomatedUnits return false

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -910,6 +922,30 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
     fun updateViewableTiles() = transients().updateViewableTiles()
     fun updateDetailedCivResources() = transients().updateCivResources()
 
+    fun isAutomatedPlayer(): Boolean {
+        return playerType == PlayerType.AI
```
yairm210, line 926: if playerType == playerType.AI return true

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -910,6 +922,30 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
...
+                // For multiplayer, if there are 3+ players and one is defeated or spectator,
+                // we'll want to skip over their turn
+                || gameInfo.gameParameters.isOnlineMultiplayer
+                && (isDefeated() || isSpectator() && playerId != UncivGame.Current.settings.multiplayer.userId)
+    }
+
+    fun doTurn(): Boolean {
+        if (!isDefeated() || isBarbarian()) {
```
yairm210, line 934: if isDefeated && !isBarbarian return false

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -777,43 +725,45 @@ class WorldScreen(
...
+    }
+
+    private fun doAutomatedMoves() {
+        isTurnActive = false
+        nextTurnButton.disable()
+        Concurrency.run("Move automated units") {
+            viewingCiv.doMoveAutomatedUnits()
+            launchOnGLThread {
+                shouldUpdate = true
+                isTurnActive = true
```
yairm210, line 745: What we *really* need here is to change canChangeState to a field, auto-set to 'false' if not player turn or if spectator.
Then, we can change that value manually, without it impacting other display issues.

core/src/com/unciv/UncivGame.kt
```diff
@@ -172,6 +175,106 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
+
+        val screen = getScreen()
+
+        if (screen is WorldScreen)
+            reloadWorldScreen(replaceOnly = true)
+        else if (screen is MainMenuScreen)
+            replaceCurrentScreen(MainMenuScreen())
+
+        screen?.openOptionsPopup(page)
+    }
```
yairm210, line 276: Sounds like a function for the Options menu, this is extremely specialized, not for UncivGame for sure

core/src/com/unciv/UncivGame.kt
```diff
@@ -172,6 +175,106 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
...
+
+            isTurnProcessing = false
+
+            if (gameInfo != originalGameInfo)
+                return@runOnNonDaemonThreadPool  // while this was turning we loaded another game
+
+            debug("Next turn took %sms", System.currentTimeMillis() - startTime)
+            reloadWorldScreen(gameInfoClone)
+        }
+    }
```
yairm210, line 219: This shouldn't be here either, it's a game-specific function that's only relevant for the worldscreen

yairm210 (changes requested):

core/src/com/unciv/UncivGame.kt
vegeta1k95: I don't agree This function has nothing to do with worldscreen. UncivGame is the class who is basically responsible for main gamedata object (gameInfo) - current + cloned. It is his responsibility to do gameinfo copying, processing and managing worldscreen. WorldScreen itself is just responsible for rendering given data to UI and handling input from user.

We shouldn't conveniently put functions inside of object with other responsibilities

core/src/com/unciv/UncivGame.kt
vegeta1k95: Options are not responsible for WorldScreen's or any screen other than itself. Any logic that involves recreating worldscreen belong to a parent which hold WorldScreen, that is to UncivGame

core/src/com/unciv/UncivGame.kt
yairm210: Screens *are* personally responsible for doing work to bring you to other screens. That is part of their responsibility. UncivGame exists to initialize the first game state, and for convenience functions that affect multiple screens, neither of which this is

core/src/com/unciv/UncivGame.kt
yairm210: In this case I agree, since Options is a cross-screen and game-wide functionality

vegeta1k95:
I will divide this into smaller PRs

AdityaMH:
#8099

AdityaMH:
this actually giving info about where enemy going to or just a gimmick?

AdityaMH:
i mean just useless decoration.

yairm210:
That's what I'm saying, you're wrong
If you have any problems in the actual changes that are made here, okay point it the problematic parts in the code

yairm210:
I don't get why you're commenting on this
Vegeta knows what he's doing and this pr has parts that objectively make the code better
I just don't get the antagonism

PR #8239, opened 2022-12-26
Big rework of City plates

vegeta1k95:
Reworked WorldScreen city plates UI, now they looks like this:

Own cities:
image

Others cities:
image

City-States:
image

More:
image
image


yairm210:
I hate to say it, but that looks much worse than our current one.
The shadows crash horribly with the rest of the UI, the numbers of the left are difficult to read, all in all my instinctual reaction was "GAH"

yairm210:
The little strength symbols on top, and the extra line around the button, are nice
But the rest is a downgrade

Ouaz:
image

It's a detail, but I think the border is too thick. As the plates for cities owned by the player are already different (item produced, growth bar) compared to AI cities, I think the thicker outline is useless.

If possible, a thin black outline like the AI cities, would be better.

vegeta1k95:
This is almost exact copy of Civ 5 nameplates. Side by side:
image vs image

The only thing that differs is that Civ 5 has outlined letters, and I did shadowed (could figure out how to do stroked letters).
(And also font obviously)

vegeta1k95:
I agree that in some cases small letters are badly visible. Would stroked letter help?

Ouaz:
What I like is that the plate have a reduced height, compared to actual ones, therefore less obstructing the view of what is behind (garrisoned units for example).

I cant' say for the readability (without having it on my device screen) like Yairm pointed out, but I agree on the "fancy" things (casted shadows on numbers, "glowing" outlines) that don't blend well with Unciv global UI.

Keep it simple (i.e. kind of "flat") and I think it's a win.

vegeta1k95:
How about this:
Own:
image
image
Enemy:
image
image

image
image
image
image


Removed shadows from labels. 
Made them of Civ color for better readability.
Better button border

vegeta1k95:
Last version: own boundary in color of civ
image
image
image
image
image

vegeta1k95:
We one is better?
Comparison:
1st) 
image
2nd) 
image
3rd)
image

yairm210:
Much better! 3 looks really good!

yairm210:
The big problem is the contrast, since the numbers need to be visible always
Here:

image

It's barely readable because it's white on light yellow
Which is why the nation colors - which are contrasting almost by default - are perfect

vegeta1k95:
New look for plates and new status icons:
image
image
image
image

vegeta1k95:
image
image
image

yairm210:
Looks good except for the icons .. they're all gradiented which makes them 'pop out' on our current UI scheme
Also the city button borders have the same problem

I can see that you really want to make a 3d-looking UI... Maybe start by making a mod with the 3d stuff? So whoever wants can use it for the new icons etc? Because it's really distracting seeing everything one way and then suddenly done elements are 3d-ish

vegeta1k95:
Ok. I will make icons outlined, but solid colored.

vegeta1k95:
image

vegeta1k95:
Final look:
image

yairm210:
The puppet image is new, needs to go to credits.md
Also, did you check other languages to see that the translation is ok?

core/src/com/unciv/ui/utils/extensions/Scene2dExtensions.kt
```diff
@@ -313,7 +313,7 @@ fun Actor.surroundWithCircle(size: Float, resizeActor: Boolean = true, color: Co
     return IconCircleGroup(size, this, resizeActor, color)
 }
 
-fun Actor.surroundWithThinCircle(color: Color=Color.BLACK): IconCircleGroup = surroundWithCircle(width*1.05f, false, color)
+fun Actor.surroundWithThinCircle(color: Color=Color.BLACK): IconCircleGroup = surroundWithCircle(width+2f, false, color)
```
yairm210, line 316: This we might need to pay around with to make sure it looks good for different sizes of circles, but we can deal with that

yairm210: 👍🏿

vegeta1k95:
Translations are good. Credits added

PR #8248, opened 2022-12-28
Add checkbox to Settings AdvancedTab to enable or disable easter eggs

FiretronP75:


FiretronP75:
Because some people don't want someone else's religions shoved in their faces. It is like forcing them to participate in a religion they don't want to participate in. So there needs to be a way to opt out.

PR #8257, opened 2022-12-28
Added WLTKD status icon

vegeta1k95:
Added WLTKD status icon to city-plate.
 Even though it is not in default Civ 5 (but in Civ 5 EUI), I think it is convenient to have it:
image


yairm210:
Credits?

vegeta1k95:
This is from existing icon: Images.Construction/OtherIcons/WTLK 1

Issue #8238, opened 2022-12-26
New policy picker UI doesn't work properly for rulesets with odd number of policy branches


ZGinner:
Android, version 4.3.12 (Build 782)

Found this bug while playing with the Deciv Redux mod. This ruleset have 9 policy branches, which aren't displayed correctly on all screen sizes besides tiny. This isn't a issue for vanilla ruleset with 10 branches and some other modded rulesets with 12 branches.
P.S. Maybe issue isn't in odd number of branches, but in something else, this is just a pattern I found.

1. Use Deciv Redux
2. Click on policy picker with not tiny screen size
3. See error

![Screenshot_2022-12-26-13-29-36-12.jpg](https://user-images.githubusercontent.com/76877062/209567850-9a8409ce-171d-4ae0-8555-ae62755b306c.jpg)

SpacedOutChicken:
I also found this issue in RekMOD, which has an odd number of policies. I did a little tinkering and added a tenth policy to DeCiv, and the problem went away.

yairm210:
@vegeta1k95

PR #8258, opened 2022-12-28
Fix PolicyScreen branches positioning (#8238)

vegeta1k95:
This should fix display of odd number of policy branches (#8238)


vegeta1k95:
image

PR #8280, opened 2022-12-31
Fixed PolicyScreen branches requirements text

### LINKED ISSUE #8274: Required policies being shown as incompatible policies in policy picker screen
Windows 10, 64 bit, and Android 12.

4.3.12 from Play Store and 4.3.14 from itch

DeCiv has a policy (Singularity) that requires three other policies and a building as prerequisites. The trouble is that the policy picker screen is saying the opposite - instead of saying that the policies have to be adopted together, it's saying "Cannot be adopted together".

The policy behaves the way it should, remaining locked until the three prerequisite policies are adopted and the required building is built. It's just the description that's got it backwards.

vegeta1k95:
Fixes #8274

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1439,6 +1439,7 @@ Completed =
 On adoption = 
 On completion =  
 Cannot be adopted together with = 
+Cannot be adopted before = 
 Adopt policy = 
 Adopt free policy = 
 Unlocked at = 
--- a/core/src/com/unciv/models/ruleset/Policy.kt
+++ b/core/src/com/unciv/models/ruleset/Policy.kt
@@ -4,8 +4,10 @@ import com.unciv.models.ruleset.unique.Unique
 import com.unciv.models.ruleset.unique.UniqueFlag
 import com.unciv.models.ruleset.unique.UniqueTarget
 import com.unciv.models.ruleset.unique.UniqueType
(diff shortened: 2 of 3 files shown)


yairm210:
We have a lot if different conditionals, we don't want to face to hardcode each one. Why not just display the entire unique?

vegeta1k95:
The point is to gather all similar conditionals under one sentence, because we don't want repeated boilerplate text, description should be slim:

Instead of:
"Cannot be adopted before X.
Cannot be adopted before Y.
Cannot be adopted before Z.
Cannot be adopted after A.
Cannot be adopted after B.
Cannot be adopted after C."

We want:
"Cannot be adopted before X,Y,Z.
Cannot be adopted after A,B,C"

yairm210:
In that case you should take all conditionals
Unique by conditional type
And "stuff" the comvined results into the original conditional with fillPlaceholders

yairm210:
Alternatively, if you want to work this way, then you should make the 'when' be exhaustive and take into account all conditionals
They usually come in pairs of "before" and "after" so they'll be about half/half

vegeta1k95:
>And "stuff" the comvined results into the original conditional with fillPlaceholders
Did it this way

yairm210:
Looks good!
How does it look for translations?

vegeta1k95:
Done

vegeta1k95:
Additional bugfix for Tiny displays

PR #8296, opened 2023-01-02
Refactored and commented GameInfo.nextTurn()

vegeta1k95:


core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -910,6 +911,23 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
...
+        // Do stuff
+        NextTurnAutomation.automateCivMoves(this)
+
+        // Update barbarian camps
+        if (isBarbarian() && !gameInfo.gameParameters.noBarbarians)
+            gameInfo.barbarians.updateEncampments()
+
+        // Have we won?
+        return victoryManager.hasWon()
+    }
```
yairm210, line 929: "do turn" and "check has won" are two very different things, we shouldn't be putting them into the same function

core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -240,71 +241,87 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion
...
+        fun getNextPlayer() {
+            playerIndex = (playerIndex + 1) % civilizations.size
+            if (playerIndex == 0) {
                 turns++
                 if (UncivGame.Current.simulateUntilTurnForDebug != 0)
                     debug("Starting simulation of turn %s", turns)
             }
-            thisPlayer = civilizations[currentPlayerIndex]
+            player = civilizations[playerIndex]
         }
```
yairm210, line 257: This should be called "setNextPlayer" since it doesn't actually return a value

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
vegeta1k95: Done

PR #8298, opened 2023-01-02
Added various styles for displaying resource icons

vegeta1k95:
Added new option to settings:
image

which changes how resource icons are displayed:

1. Background color coded by yield, res type indicator (current, default)
image
2. Background color coded by type, no indicators 
image
3 Background color coded by type, amount indicator 
image

### DIFF
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -6,7 +6,6 @@ import com.unciv.Constants
 import com.unciv.UncivGame
 import com.unciv.logic.multiplayer.FriendList
 import com.unciv.models.UncivSound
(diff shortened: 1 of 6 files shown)


yairm210:
I think we'll poll this, maybe replace yield coloring entirely
In any case B is entirely superceded by C so even if we see a split vote we'll only need 2 options

yairm210:
We're currently at 24 vs 8 for C, which is a Definite Majority, so let's replace existing UI entirely :)

vegeta1k95:
That means, remove option and implement only C)?

yairm210:
Yup

vegeta1k95:
Done. Now this is default:
image

AdityaMH:
IMG_20230104_140937

PR #8302, opened 2023-01-04
Fix Flat Earth Map Shape Game Saving

FiretronP75:
Use removeTerrainFeatures and addTerrainFeature instead of setTerrainFeatures so serialization will work.


FiretronP75:
Unfortunately, there is no way to fix the old saves. This will only make new games save correctly.

yairm210:
Why would this make a difference? They both use set inside?

FiretronP75:
> Why would this make a difference? They both use set inside?

On the inside, there is a conversion happening, that I was not doing.
`setTerrainFeatures(ArrayList(terrainFeatures).apply { add(terrainFeature) })`

PR #8318, opened 2023-01-06

CrispyXYZ:


adwaitacccp:
The translation is incorrect.

"Demographics" = "人口统计" != "统计" (Greek root: demo- = people, -graph- = write)

CrispyXYZ:
Thanks for your advice!

In fact, it is better to translate into `统计数据` according to the context. (See Screenshot below)

Let's see what Google says:
> **statistical data** relating to the population and particular groups within it.

See #6645 and [Fandom wiki](https://civilization.fandom.com/wiki/Demographics_(Civ5)) for more info.

Screenshot_20230107_182949_com unciv app

adwaitacccp:
Well you are probably right.

PR #8325, opened 2023-01-07
Rework of PromotionPicker UI

vegeta1k95:
1) Big reworks of PromotionPicker UI:
 - Now promotions tree layout is automatically generated, based on the optimally compactified tree form
 - Tree orientation changed from up-down to left-right layout
 - Selected promotion is highlighted as well as it's prerequisites by color and lines
 - Fully compatible with mods
 
 Some examples:
image
image
image
image
image
image

As well as mods:
Deciv Redux:
image
image
Alpha Frontiers:
image
image

2) Fix minor bugs with colors copies.
3) PromotionPicker screen now also opens when "XP" label is pressed in the unit info table.

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
@@ -37,7 +37,7 @@ import java.lang.Integer.max
 import kotlin.math.abs
 import kotlin.math.min
 
-object Colors {
+object PolicyColors {
 
     val policyPickable = colorFromRGB(47,67,92).darken(0.3f)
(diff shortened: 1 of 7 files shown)


yairm210:
The coloring for dependencies is confusing to me, out looks like only direct dependants are highlighted?

vegeta1k95:
> out looks like only direct dependants are highlighted?

Yes. Because mods like Deciv, have circular promotion dependencies A -> B -> C -> A (which is a just a bad design imo), hence nearly whole tree becomes mess with lines and highlights

yairm210:
Looks good all in all
Have you checked this in another language to ensure no translation problems?

vegeta1k95:
Yes, translations should be fine

PR #8335, opened 2023-01-08
Modified Pantheon.png

xk730:
从一道闪电变成了众人想象出的一道闪电
From a flash of lightning to a flash of lightning that everyone imagined


yairm210:
A. I don't think this is a better image
B. Even for good images, we need to add credit in credits.md

PR #8343, opened 2023-01-08
Platform Graphical Asset Update

letstalkaboutdune:
Updated repository graphical assets to match latest versions being used on all storefronts


letstalkaboutdune:
Any idea why this would fail a check? I only added/updated image files.

yairm210:
Also for the banner - since you've changed the size significantly, we'll need to see how that renders on phones

letstalkaboutdune:
> The tests are ok The github preview image is _supposed_ to be a screenshot, that's what I want people to see first thing. I'm not sure what I want for Google Play.

I changed GitHub back to a screenshot, but updated it to the most recent one with all AU sprites. I also resized the banner to roughly match the old one.

I think that Google Play and Itch cover/feature images shouldn't be screenshots, since there are separate areas for screenshots for both of those. Looking at every other app, the featured image contains the logo and non-screenshot artwork.

PR #8354, opened 2023-01-10
AbsoluteUnits - Unique Triremes

letstalkaboutdune:
Unique_Trireme

Also increased banner size (2x each dimension) as the previous looked very small, and fixed a typo in a Steam asset filename.


yairm210:
We display the banner at full size, and on phones - where default is portrait mode - this means the banner will be to wide to fit in the screen.
I'll have to test how this looks and perhaps make changes.

letstalkaboutdune:
> We display the banner at full size, and on phones - where default is portrait mode - this means the banner will be to wide to fit in the screen. I'll have to test how this looks and perhaps make changes.

I thought phone default was landscape mode, and portrait mode had to be enabled with an option. But if we can determine a good resolution, then I'll happily use that. Maybe 2x the dimension is too much.

letstalkaboutdune:
I shrank the banner back down to the original size, because I don't want that issue delaying the release of the new unique sprites.

yairm210:
Please make a new pr including only the bigger banner, for testing 😉

PR #8371, opened 2023-01-13
Prefer to not embark

itanasi:
Slight initial modification of pathfinding to prefer Land tiles as a Land unit all else being equal

### DIFF
--- a/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
+++ b/core/src/com/unciv/logic/map/UnitMovementAlgorithms.kt
@@ -151,7 +151,7 @@ class UnitMovementAlgorithms(val unit: MapUnit) {
         while (tilesToCheck.isNotEmpty()) {
             val updatedTiles = ArrayList<TileInfo>()
             for (tileToCheck in tilesToCheck)
-                for (neighbor in tileToCheck.neighbors) {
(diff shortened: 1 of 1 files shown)


yairm210:
Since pathfinding is notoriously the most expensive part of turn time, please show "next turn" times for a big game, before and after the change

yairm210:
I mean for one thing, if you're not a land unit you don't need sortedByDescending *at all*, that can be an if that saves list reorder

yairm210:
I've been thinking about this the past day, especially "why is still recommends embarked route when there in a min embarked one", and we've been going at this all wrong!
Embarkation is a *multi turn* issue, so we need to be sorting *in the multi turn function* - that is, in getShortestPath!

This can also massively improve performance if we sort by *aerial distance to destination* since now we go through half (on average) of the outer rim of the reached tiles, and sorted we'll go through hardly any!!!

itanasi:
Yup. I just haven't groked getShortestPath yet

yairm210:
Tested this timewise and it's a 6% improvement on average for getShortestPath functions that succeed.
On an 18-19 second "next turn" game, that's only about 1 100ms improvement, which is hardly felt. BUT it solves the current problem as well as a minor boost, so that's a win :)

PR #8376, opened 2023-01-13
removed notificationLog from the game file

### LINKED ISSUE #8363: Feature request: Store non-essential data locally, instead of sending it to the server
The notification log is sent every turn to the server, which means I had to limit the maximum log turns to 5-15 (arbitrary numbers). 

Store the notification log to a separate file which isn't sent to the server. Once the player resigns / the game finishes the file can be deleted.

**Additional context**
I could give this a go but before actually implementing this I'd like to know if I even should try this. That is if you know any disadvantages this could bring.
The advantage is that we can increase the maximum notification log to whatever the user wants. 
The main issue with the notification log pointed by touhidurrr was the 15 turn limit. Storing this sort of data locally without actually bothering to send it over the internet would make it possible to increase the turn limit without increasing the game file size. 

There might be more data that can be just stored locally instead of sending it to the server, but rn I can only think of the notif log.

alexban011:
resolves  #8363

I also increased the max turns to 90 now that it won't clutter save files and the default max number of turns to 15

This deletes the notification log for current saves. 

Should I increase the serialization version? Personally I don't think losing the notif log is that critical, but if people use different versions of the game the notification log will break at every turn.

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
+++ b/core/src/com/unciv/logic/civilization/CivilizationInfo.kt
@@ -1,9 +1,11 @@
 package com.unciv.logic.civilization
 
+import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.math.Vector2
 import com.unciv.Constants
 import com.unciv.UncivGame
 import com.unciv.json.HashMapVector2
+import com.unciv.json.json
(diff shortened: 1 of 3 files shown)


alexban011:
converted to draft since I can't figure out right now how to delete the notification log when deleting a game

PR #8381, opened 2023-01-15
Preparation for Tactical AI Rework: analysis map, domination zones

vegeta1k95:
This is preparatory commit for upcoming changes to tactical AI. 
It does not change anything in current behaviour, but introduces a few things that are going to be useful later.

Ths includes: 

1) New `CityDistanceData` class which stores information about closest city for all tiles. Data is recomputed when founding/razing a city.
Reference: [header](https://github.com/LoneGazebo/Community-Patch-DLL/blob/master/CvGameCoreDLL_Expansion2/CvDistanceMap.h), [source](https://github.com/LoneGazebo/Community-Patch-DLL/blob/master/CvGameCoreDLL_Expansion2/CvDistanceMap.cpp)

2) New `TacticalAnalysisMap` class which computes "zones of domination" based on of players' cities and areas (continents).
Right now it does only a territorial separation (which you can test visually by enabling `debug` inside `TacticalAI` class).
Upcoming: calculate military strength, determine tactical dominance and tactical posture in each zone.
Reference: [header](https://github.com/LoneGazebo/Community-Patch-DLL/blob/master/CvGameCoreDLL_Expansion2/CvTacticalAnalysisMap.h), [source](https://github.com/LoneGazebo/Community-Patch-DLL/blob/master/CvGameCoreDLL_Expansion2/CvTacticalAnalysisMap.cpp)


core/src/com/unciv/logic/CityDistance.kt
```diff
@@ -0,0 +1,141 @@
...
+                return a
+            else if (a == null && b == null)
+                return null
+
+            if (a!!.distance < b!!.distance)
+                return a
+            else if (a.distance > b.distance)
+                return b
+            else {
+                if (a.city.civInfo.isMajorCiv() && b.city.civInfo.isMinorCiv())
```
yairm210, line 27: Unnecessary else leads to unnecessary nesting

core/src/com/unciv/logic/CityDistance.kt
```diff
@@ -0,0 +1,141 @@
...
+            updateDistanceIfLower(owner.civName, thisTile.position, city, distance)
+        }
+    }
+
+    private fun update() {
+
+        // Clear previous info
+        reset()
+
+        for (player in game.civilizations.asSequence()) {
```
yairm210, line 88:
```suggestion
        for (player in game.civilizations) {
```

core/src/com/unciv/logic/CityDistance.kt
```diff
@@ -0,0 +1,141 @@
...
+            if (player.isDefeated())
+                continue
+
+            val isMajor = player.isMajorCiv()
+            if (isMajor)
+                resetPlayer(player.civName)
+
+            // Update distances for every city and every tile
+            for (city in player.cities.asSequence())
+                for (otherTile in game.tileMap.values.asSequence())
```
yairm210, line 100: This goes over the entire map for every city
Sounds to me like a performance point, there are sometimes a hundred cities and ten thousand tiles
We'll need to see how this affects general performance

yairm210: I can think of ways to optimize this

vegeta1k95: The distance recalculation is supposed to be done only when city is founded/razed, so not every turn obviously.

core/src/com/unciv/logic/CityDistance.kt
```diff
@@ -0,0 +1,141 @@
...
+        return if (player != null && player.isMajorCiv())
+            data[player.civName]!![tile.position]?.distance ?: Int.MAX_VALUE
+        else if (majorsOnly)
+            data[IDENTIFIER_MAJOR_CIVS]!![tile.position]?.distance ?: Int.MAX_VALUE
+        else
+            data[IDENTIFIER_ALL_CIVS]!![tile.position]?.distance ?: Int.MAX_VALUE
+
+    }
+
+    fun getClosestCity(tile: TileInfo, player: CivilizationInfo? = null, majorsOnly: Boolean = false) : CityInfo? {
```
yairm210, line 123: Same as above function, we can have just one that returns the citydistancedata directly

core/src/com/unciv/logic/CityDistance.kt
yairm210: In that case it's going to be simply huge
I highly doubt we really need this info for each any every tile in the entire map, it'll be cheaper to calculate when needed

vegeta1k95: This is needed to divide map into tactical zones by cities. Tiles in proximity of the city go to that zone. 
Since city tiles can be bought/expanded/snatched by citadel on each player's turn, we need to update division each player's turn also. We surely don't want to recalculate distances each turn, so that is why they are cached.

yairm210: I understand the tactical zone
But even so, we can agree that e.g. tiles 10 away from the city should not be in the "zone" of a city?
Then already we can massively limit the area we need to initialize

Calculation distance thus limited is also cheap, so we can afford to do it every turn
And not add another huge chunk of data to the save file

vegeta1k95: This is the key piece in the source:
image

yairm210: If we cut that number to say 7 tiles away it drops even more
So instead of 10,000 tiles * number of cities
It drops to 330 initially
And to 180 for 7-tile radius
That's two orders of magnitude for large maps

vegeta1k95: The logic is the following with the current setup:

1) Tiles in radius 2 around the city belong to the city tactical zone independently of the tile owner.
2) Tiles in radius 3-4 around city:
    a) if owned -> belong to owning city zone
    b) if not owned -> belong to closest city by path zone (in original code, I skipped this part already)
3) Tiles in radius >4 cannot belong to this city's zone

The problems start when there is a tile is on a border between two different civ's cities. 
E.g.:  City A -> * -> * X * <- * <- City B
Which city it should belong to? That is why some tiebreaker needed.

vegeta1k95: You will run into overwriting problem if you will just plainly iterate around cities. Some cities will absorb additional tiles by overwriting wihout any tiebraking.

vegeta1k95: > If we cut that number to say 7 tiles away it drops even more So instead of 10,000 tiles * number of cities It drops to 330 initially And to 180 for 7-tile radius That's two orders of magnitude for large maps

You are gonna run into problem when two closely placed cities with 7 radius zones are going to unevenly divide tiles because on overlap there will be no tiebraking - last city is gonna absorb whole intersection.

yairm210: I'm not saying ditch the "closest tile" logic
I'm saying ditch the "iterate entire map"

yairm210: If it's only applicable for 4 tile radius anyway then that's even better!

vegeta1k95: I just copied original algorithm and tested it - it produces good zones division. If you know how to optimize this computing such that it produces good zones division - you are welcome to suggest.

vegeta1k95: The only idea I have atm: calculate and compare distances only for plots in rings of radius up to 4 around cities, all remaining plot will have  undetermined distances.

vegeta1k95:
Seems OK for now, but needs to be tested more thoroughly on various maps

core/src/com/unciv/logic/civilization/CivilizationInfo.kt
```diff
@@ -173,6 +174,9 @@ class CivilizationInfo : IsPartOfGameInfoSerialization {
     private var allyCivName: String? = null
     var naturalWonders = ArrayList<String>()
 
+    /* AI section */
+    val tacticalAI = TacticalAI()
```
yairm210, line 178: Since all parts of this are transient, why not have this transient as well?

yairm210:
Let's give it a go

PR #8388, opened 2023-01-15
Create Arabic.properties

Hisham-12:
just a new file for arabic with template.properties copied over with no additional changes


Hisham-12:
I apologize I missed the bit in https://yairm210.github.io/Unciv/Other/Translating/#starting-out that states that Arabic and Hebrew are not supported.
I have closed it now.

yairm210:
I put that now in the very first paragraph so as not to waste people's time

PR #8393, opened 2023-01-17
Optimization: tiles "isExplored" property delegated to TileInfo class

vegeta1k95:
The field `exploredTiles: HashSet<Vector2>` of `CivilizationInfo` is redundant for the following reasons:
1) Code uses this field only for containment checks: is explored/not explored. The set is never iterated directly.
2) Every time player explores a new tile the whole set has to be recreated. Hence possible performance impact for large maps later in the game.
3) Explored tiles are persistent: once revealed - stays revealed whole game.

This commit suggests instead of `exploredTiles` field in `CivilizationInfo` to use hashset field `exploredBy` in `TileInfo`. This serializable field stores players ids (civ names, for now) who have revealed this tile. 

Pros: 
1) Operation complexity of `tile.isExplored(civ)` is still constant O(1), but can be made even faster because we can use hashing by `Int` IDs instead of `String` civ names or `Vector2` positions.
2) No need to recreate anything upon new tile exploration - just add civ's id to the set. 
3) No concurrent modification errors etc - simple serialization/deserialization.
4) Save games files should not be corrupted

Cons:
1) I could not come up with a backwards compatibility: previous version saved games would have most tiles "unexplored" on the first load.

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -431,7 +431,7 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
                 }
 
         val exploredRevealInfo = exploredRevealTiles
-            .filter { civInfo.hasExplored(it.position) }
+            .filter { civInfo.hasExplored(it) }
             .flatMap { tile ->
                 civInfo.cities.asSequence()
                     .map {
@@ -586,6 +586,10 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
         cityDistances.game = this
 
         guaranteeUnitPromotions()
+
+        for (player in civilizations)
+            for (tile in player.exploredTiles)
+                tileMap[tile].setExplored(player, true)
     }
 
     //endregion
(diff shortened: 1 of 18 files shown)


yairm210:
Sounds good!
To support multiplayer with different versions, we can introduce a temporary period where the data is stored in both places
And then in a couple of versions (like 2 weeks) when we can assume people have a version that contains the new data, we can remove support for the old one

vegeta1k95:
This should do the trick

vegeta1k95:
This should do the trick. Later we can discard it

core/src/com/unciv/logic/map/TileInfo.kt
```diff
@@ -234,6 +237,19 @@ open class TileInfo : IsPartOfGameInfoSerialization {
         return player.viewableTiles.contains(this)
     }
 
+    fun isExplored(player: CivilizationInfo): Boolean {
+        if (UncivGame.Current.viewEntireMapForDebug || player.isSpectator())
+            return true
+        return exploredBy.contains(player.civName) || player.exploredTiles.contains(position)
+    }
+
+    fun setExplored(player: CivilizationInfo, isExplored: Boolean) {
```
yairm210, line 246: Here too we need to save the explored tiles the old way, so that another player doesn't override them for me 
In multiplayer, each player gets the current game and returns the new serialized version. Players on old versions will not even serialize fields that their version doesn't contain!

yairm210:
What I'm missing is a function in gameInfo.setTransients() that foreach civ foreach tile in explored tiles tile.setExplored()
That would currently take effect every turn - so the "source of truth" is the old data and the "value we check in game" is the new data
But this makes compatibility really easy

vegeta1k95:
Now ok?

yairm210:
Looks good
If you haven't yet, can you test this by
- starting a game in older version
- moving to newer version, exploring
- movibg back to older version and cinfirming explored tiles remain?

Code looks like it accommodates this but I think a practical check can help confirm

vegeta1k95:
Tested, seems fine

PR #8394, opened 2023-01-17
Add options to use random number of players and city states.

FiretronP75:


android/assets/jsons/translations/template.properties
```diff
@@ -360,6 +360,8 @@ Three Continents =
 Four Corners = 
 Archipelago = 
 Inner Sea = 
+Random number of Players =
+Random number of City-States =
```
alexban011, line 364: tests fail because you didn't  add a space after `=`

FiretronP75: Thanks, fixed

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
```diff
@@ -98,6 +98,34 @@ class NewGameScreen(
...
         rightSideButton.onClick {
+            if (gameSetupInfo.gameParameters.randomNumberOfCityStates) {
+                val half = gameSetupInfo.gameParameters.numberOfCityStates / 2
+                val otherHalf = gameSetupInfo.gameParameters.numberOfCityStates - half // This is to handle odd numbers
+                gameSetupInfo.gameParameters.numberOfCityStates = half + Random().nextInt(otherHalf)
+            }
+
+            if (gameSetupInfo.gameParameters.randomNumberOfPlayers) {
+                val half = gameSetupInfo.gameParameters.players.size / 2
+                val otherHalf = gameSetupInfo.gameParameters.players.size - half // This is to handle odd numbers
```
alexban011, line 109: If I were you I'd rather let the players decide what "random" should be. 

That is:
 a `min city states` and `min major civs` sliders
and
a `max city states` and `max major civs` sliders

alexban011:
so now this test failed:
```
 com.unciv.uniques.TriggeredUniquesTests > testConditionalTimedUniqueIsTriggerable FAILED
    java.lang.NullPointerException
        at com.unciv.uniques.TriggeredUniquesTests.testConditionalTimedUniqueIsTriggerable(TriggeredUniquesTests.kt:29)
```

But what I noticed is that if I start a game with the random nr of players, I instantly win (probably because it generates 0 opponents) Another reason to let the user decide the min and max values

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
FiretronP75: I was thinking the same thing, started out that way. But it is too complicated because I'd have to manipulate the player table on the right to add and remove players as you slide the max players.

FiretronP75:
> so now this test failed:
> 
> ```
>  com.unciv.uniques.TriggeredUniquesTests > testConditionalTimedUniqueIsTriggerable FAILED
>     java.lang.NullPointerException
>         at com.unciv.uniques.TriggeredUniquesTests.testConditionalTimedUniqueIsTriggerable(TriggeredUniquesTests.kt:29)
> ```
> 
> But what I noticed is that if I start a game with 1 AI and the random nr of players option enabled, I instantly win (probably because it generates 0 opponents) Another reason to let the user decide the min and max values

Do you know why that test would fail?

In my testing, I found I had to work around the Spectator. But yeah I didn't idiot proof having only 2 players. I'll make the min human players + 1 to make sure there is at least one AI.

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -141,6 +143,14 @@ class GameOptionsTable(
             addCheckbox("Disable starting bias", gameParameters.noStartBias)
             { gameParameters.noStartBias = it }
 
+    private fun Table.addRandomPlayersCheckbox() =
+            addCheckbox("Random number of Players", gameParameters.noStartBias)
+            { gameParameters.randomNumberOfPlayers = it }
+
+    private fun Table.addRandomCityStatesCheckbox() =
+            addCheckbox("Random number of City-States", gameParameters.noStartBias)
```
alexban011, line 151: you probably meant `gameParameters.randomNumberOfPlayers` and `gameParameters.randomNumberOfCityStates` instead of `gameParameters.noStartBias`

FiretronP75: LOL

FiretronP75: I wonder why it works despite that.

FiretronP75: Oh, that is the starting value.

FiretronP75: Fixed.

core/src/com/unciv/ui/newgamescreen/NewGameScreen.kt
FiretronP75: There is also the extra validation to make sure the user chooses max larger than min.

FiretronP75: Maybe we could start this way and later you can upgrade it to use 4 sliders.

FiretronP75:
I found there can be a crash if the city states is set to 0 or 1 with random turned on.

alexban011:
> I found there can be a crash if the city states is set to 0 or 1 with random turned on.

alright so that's why it kept crashing for me :)

As for the failing tests you should just ignore them because I saw the last commits make this test fail so you should blame @yairm210 for this :D :D :D

FiretronP75:
> I found there can be a crash if the city states is set to 0 or 1 with random turned on.

Fixed

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
alexban011: > I wonder why it works despite that.

because the second parameter of `addCheckbox()` will just be toggled. The whatever you add in the lambda it will be executed. What actually happened here is that whenever you clicked  the checkbox you toggled the `noStartBias` option as well. So if you want to make this cleaner you can just remove the lambda. (A couple months ago it was necessary to change the option in the lambda, but then someone changed this but didn't change all usages of the function so some usages still have the useless lambda)

yairm210:
As usual, pictures for UI elements

FiretronP75:
Random

FiretronP75:
It randomly selects between half and all of what you defined in the city state slider and the player list.

FiretronP75:
Screen shot of updated random city states. Code swaps min and max if the user accidentally swaps min and max.
CityStateSliders

FiretronP75:
Screen shot of random civs toggle and sliders.
PlayerSliders

FiretronP75:
Both sets of sliders enabled at the same time.
BothSliders

FiretronP75:
I have it all working correctly now.

For players, I decided to keep it simple and rather than add and remove players as you slide the sliders, I add or remove the difference on game start.

FiretronP75:
So for example if you define 3 players, and choose a random range of 4 to 8 players, and it randomly picks 6, then it will add 3 players for you.

And if you define 12 players, and choose a random range of 4 to 8 players, and it randomly picks 7, then it will randomly select 5 AI players to remove. If there are less than 5 to remove (because multiplayer) then it will remove as many as it can.

yairm210:
Let's try it out

alexban011:
Idk if I like this implementation, but hey.. it's not my call :D

If I want to play agains 4 specific civs and have 3-6 random civs, then I wouldn't actually want it to remove what I selected. 

Additionally the original request was this

> We can have the world , our civ, and the resource amount randomly generated however we know how many civs and state there are.
It would be cool if we could have that randomly generated too.

Aka to add more exploration to the game (explore the map to see what you're playing against). This PR (afaik) adds the civs to the player list, which removes the whole random aspect of it. I can randomly click the + button without the need for sliders

FiretronP75:
> If I want to play agains 4 specific civs and have 3-6 random civs, then I wouldn't actually want it to remove what I selected.

That is just a matter of training. User will learn to pick random range of 7-10 to do that.

FiretronP75:
> Additionally the original request was this
> 
> > We can have the world , our civ, and the resource amount randomly generated however we know how many civs and state there are.
> > It would be cool if we could have that randomly generated too.

But also, the OP of that OR gave my implementation idea a 💯 emoji

alexban011:
> That is just a matter of training. User will learn to pick random range of 7-10 to do that.

To me this seems like extra steps make a confusing UX, but ok

alexban011:
> my implementation idea 

But wasn't that idea to remove certain nations from the random pool?

FiretronP75:
> To me this seems like extra steps make a confusing UX, but ok

Well it matches the label at least. If we were only adding to the pool, we'd have to change the label to "Min number of Civilizations to randomly add", not that it would be a problem to do so. I'm just saying the label correctly states what it will do, so I don't think it will be confusing.

alexban011:
> two use cases while yours only supports one use case

Which are ..? 

> not losing the feature

It's already merged. Don't worry I won't remove it. I'm just saying that removing nations you selected seems like a confusing UX, especially for a new user

FiretronP75:
One use case is adding a random number of civs to what you've defined.

The other use case is randomly picking from what you've defined. The OP and I find this important. You do not.

We need this feature for when we want to ensure all the AI players have a bonus related to the map type or theme or scenario.

If I'm doing a water world, I only want AI civs with a water bonus. If I'm doing a Jungle themed custom map, I only want civs meant for Jungles. If I'm doing a Medieval Era mod, then I only want civs that get a bonus during the Medieval Era. I do these setups all the time, and would like to add some randomness to it, so I don't get the exact same water civs each time I do a water world.

alexban011:
> The other use case is randomly picking from what you've defined. The OP and I find this important. You do not.

I do but I can't see any code that would let you define something like that. Did you forget to push it? Cause from what I see you don't define anything other than 2 numbers.

alexban011:
Ohhh now I get it. You select the civs you want and let the randomness remove some of them. Even more confusing and I'm sure a new player will have no idea he can use this feature like this.

I'd have made a "MultiPickerScreen" and the user select the civs he wants to add/rempve to the random pool. I think it would be a lot more clear to the user, but anyways.. maybe some other time :)

PR #8396, opened 2023-01-17
Added option to choose a random amount of major civs

alexban011:
Resolves part of a discord suggestion

The random civs are added to a new variable in the gameParameters so you can't see how many civs were chosen (+ it won't be saved to the lastGameParameters)

This adds a new table which adds additional civs, besides the ones already selected.

https://user-images.githubusercontent.com/21990802/212862333-f7ff51e7-c54a-4a2f-b326-d25315b1b5f2.mp4


yairm210:
Player shouldn't have to click on 'set values', we can listen to events in text with .onChange{}

android/assets/jsons/translations/template.properties
```diff
@@ -591,6 +592,10 @@ You have to write an ID for your friend! =
 You cannot add your own player ID in your friend list! = 
 To add a friend, ask him to send you his player ID.\nClick the 'Add friend' button.\nInsert his player ID and a name for him.\nThen click the 'Add friend' button again.\n\nAfter that you will see him in your friends list.\n\nA new button will appear when creating a new\nmultiplayer game, which allows you to select your friend. = 
 Please input Player ID! = 
+Random civilizations = 
+Set values = 
+Please enter values for maxmimum and minimum number of civilizations! = 
+Please enter only integer values! =
```
yairm210, line 598: This should be part of the textbox, it should only allow characters 0-9, there's a character filter there somewhere

yairm210:
- What if player puts min 10 max 2?
- What if player puts min and max at more major civs than are available? (100)

I think the simplest solution is to have values be a dropdown. We know that there are a limited number of possible values, dropdowns have the double advantages of minimizing bad input and of being easier on phones than inputting text (especially numbers!)

alexban011:
> What if player puts min 10 max 2?

the code already reverses them when creating the map

> What if player puts min and max at more major civs than are available? (100)

yep. Dropdown it is :)

yairm210:
Or slider :D
Whichever you think looks and feels better

alexban011:
Yeah I saw the other comment after writing this one. 

I think a slider would be even easier to use on mobile than a dropdown with 30 values

FiretronP75:
The max slider needs to add and remove rows in the players table, so the user can choose which nations the random number will select from. This is important for making sure you only use nations that fit the map or scenario or mod you are doing.

alexban011:
The original suggestion was to add random civilizations but not to show them. He said that we have random resources and such but he wants to have to explore to find out how many civs there are and how many cs

alexban011:
I don't think showing how many civs there will be is the point of this feature. If you just want a random value you could use random.org and then click the + button that many times. I think "random" should mean you have no clue how many civs there are before actually encountering them

PR #8400, opened 2023-01-17
added option to choose random number of city states

alexban011:
finishes a discord suggestion 

https://user-images.githubusercontent.com/21990802/212981189-9e509b08-6e5f-45fa-9fa2-5c0e7c4eb197.mp4


yairm210:
Also suffers from min > max problem
But much better than choosing major civs! In fact I think this is a good solution to use for major civs as well!

Both this AND major civ randomness should be under a "advanced settings" category, like with map options.

alexban011:
> Both this AND major civ randomness should be under a "advanced settings" category

should the CS randomness be there as well? Cause I think it would be easier to see that slider where the other CS number slider is. If it isn't it would have to remove the slider and the user needs to keep remembering that the CS number can be adjusted from both panels. Maybe I can move the option to the advanced category but keep the sliders in the GameOptionsTable?

As for  the major civ randomness I totally agree to move it to the advanced settings

yairm210:
I think so, this is a pretty niche request.

PR #8426, opened 2023-01-20
Include forked repos in github search.

FiretronP75:


FiretronP75:
> Piques my curiosity: why? And wouldn't forks often be indistinguishable from their masters in our UI? Example where useful?
Think inheritance. The forks extend the original, and get updates from it, while maintaining their extensions to it. For an example see my mods "Upgradable-Scouts-BNW" and "Upgradable-Scouts-Latam" which extend/fork my "Upgradable-Scouts". Other modders are doing this too.

FiretronP75:
> The code, too - that new (to me) searchRequest parameter from Jack's #6718 is under external control but neither sanitized nor URL-encoded. But it encodes the blank it needs (buggy in 6718 fixed in #6725) to `+` while you use `%` - matter of taste, % may be more modern, still inconsistent. Better build the entire `q=` query param with blanks then pass to `URLEncoder.encode`???

I can make any changes you think best.

PR #8430, opened 2023-01-20
Added option to allow players to choose randomly selected civs

### LINKED ISSUE #8399: Feature request: Ban certain nations (from pick or random pool) in online multiplayer
This request is an attempt to enable players to create a more competitive multiplayer experience.

The idea is that in many competitive multiplayer games, there is a "ban phase" where players can choose certain characters, heroes, etc., (in our case nations) to ban from the pick or random pool. I'd like Unciv to implement something like that.

Perhaps the players could agree beforehand and choose a list of as many nations as they want to ban. Or, each player could receive one ban that they could apply to any nation.

In terms of UI, this could be achieved with a "Competitive Mode" tickbox in the new game options, which would then display new buttons for banning. It could also force the Demographics dashboard to be used in game, which limits the amount of ranking information that a player can see for other nations .

alexban011:
resolves #8399

~I mean the actual issue requested the opposite: to ban nations not to choose which you want to use. But I guess an option could be added to switch between blacklisting or whitelisting nations, if you feel like adding yet another option would be beneficial.~
EDIT: I added a blacklist checkbox which lets you remove nations instead of whitelisting them

I'm sorry I use screenshots this time but SimpleScreenRecorder freezes and crashes today.

This is how the option looks like. The `choose nations` button and 'blacklist nations` checkbox only appear when the 'random nations pool' checkbox is checked.

image

This is the nation selection popup. Pretty much the normal nation selection but allows multiple nations to be selected.

randomNation2

I didn't showed this but I had selected Babylon for myself and had 3 other random civilizations, which were chosen out of the 4 I selected.

randomNation3


core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -222,7 +222,22 @@ object GameStarter {
     private fun addCivilizations(newGameParameters: GameParameters, gameInfo: GameInfo, ruleset: Ruleset, existingMap: Boolean) {
         val availableCivNames = Stack<String>()
         // CityState or Spectator civs are not available for Random pick
-        availableCivNames.addAll(ruleset.nations.filter { it.value.isMajorCiv() }.keys.shuffled())
+        if (gameSetupInfo.mapParameters.blacklistRandomNationsPool) {
```
yairm210, line 225: Surely the 'enable' if should come before the 'blacklist' if... that's the higher-level function enabler

yairm210:
Which nations can be contained in random is *not* a map parameter setting. It's probably a nation picker setting, and barring that - a ruleset setting, like 'number of random civs'

alexban011:
I moved them to the GameOptionsTable

alexban011:
Btw do you think the text I used for the checkboxes and the button are clear enough for the user? Do you have any suggestion on better naming?

yairm210:
I don't, if the community has better names we can take that

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -400,3 +440,135 @@ class GameOptionsTable(
     }
 }
 
+private class RandomNationPickerPopup(
```
yairm210, line 443: This is over 100 lines and therefore big enough to be its own file

yairm210 (approved):
Apart from that technicality, looks good!

alexban011:
> Apart from that technicality, looks good!

I also want to add the same thing for the City States, so I'll move it in the next PR when I'll also generalize it so I can reuse it

PR #8446, opened 2023-01-23
Translate the hard-coded notification strings

JackRainy:
A few hard-coded strings were missing `.tr()` call, thus a few notifications were never translated.
Here is the fix.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -72,6 +72,7 @@ Get [unitName] =
 
 Hydro Plant = 
 [buildingName] obsoleted = 
+due to constructing [buildingName] = 
 
 # Diplomacy,Trade,Nations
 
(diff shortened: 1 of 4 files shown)


yairm210:
No! These strings are supposed to be untranslated and enter the notifications untranslated, and are only translated to the user when displayed
This allows both notifications from one user to another, and for the user to change language and have all notifications correctly displayed
Do you have an example where this isn't working?

JackRainy:
> Do you have an example where this isn't working?

Screenshot_20230113-091158048_1

Yes, here is one of the example where it does not work.

yairm210:
So we need to solve that specifically, it's a problematic one to begin with since it has two halves that are joined
Can you attach the saved game?

JackRainy:
> Can you attach the saved game?
Of course, here is a bit artificial saved game a turn _before_ this notification appears. If it does not appear, tries a few times to reload since it depends on the random AI decision.
Also, in a turn you will see another notification about a gained person from the built Wonder. It is also not translated.

Gualdimar:
additionally, the [Stats] part is saved in the selected language and is not translated after changing the language

yairm210:
I understand the problem, and we will fix it, but this is not the solution.

PR #8448, opened 2023-01-23
Modification of the random part of Prize Ship to avoid save scumming

AlatarTheYoung:
Addition of a random seed in the unitCapturedPrizeShipsUnique function in Battle.kt based on the code in the function calculateDamageToDefender in BattleDamage.kt.
The seed is based on :
- The attacker turn
- The defender position

The idea is to prevent reloading the turn in multiplayer to get different results.
With this code, the attacker shall always have the same chance of capture for a specific boat and turn.

That should seriously nerf the privateer issue in multiplayer with a simple and light solution.


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -242,7 +242,8 @@ object Battle {
                 0.1f + attacker.getAttackingStrength().toFloat() / defender.getDefendingStrength()
                     .toFloat() * 0.4f
             )
-            return Random().nextFloat() <= captureChance
+            /** Between 0 and 1.  Defaults to turn and location-based random to avoid save scumming */
+            return Random(attacker.getCivInfo().gameInfo.turns * defender.getTile().position.hashCode().toLong()).nextFloat() <= captureChance
```
yairm210, line 246:
```suggestion
            val random = Random(attacker.getCivInfo().gameInfo.turns * defender.getTile().position.hashCode())
            return random.nextFloat() <= captureChance
```
For readability (line is too long) separate 'what is random' and 'what do we do with it'

AlatarTheYoung: Ok for me, I will try to modify it and run the tests

AlatarTheYoung:
I am on it

AlatarTheYoung:
Compilation checked on my part

PR #8452, opened 2023-01-23
Fix UX of random nations pool && keep advanced tab open when updating the screen

alexban011:
I've created `class MultiNationSelectPopup` which pretty much contains what the `class RandomNationPickerPopup` had, but with an additional boolean parameter to choose if it should show MajorCivs or CityStates. (and also to save them in the appropriate variable in gameParameters)


alexban011:
I also made the advanced tab start opened if either of the random pool options are checked since deselecting them closed the ExpanderTab which was very annoying 


https://user-images.githubusercontent.com/21990802/214097068-36f97f58-f669-465f-9451-2c6fd7714f36.mp4

yairm210:
I'm actually against this, this is too much micromanaging

Also people in Discord seemed to have trouble understanding the random picker, which means that something in the option names needs to change. If even our pro discord players don't intuitively get it then a random player never will

alexban011:
> If even our pro discord players don't intuitively get it then a random player never will

Hmm, do you know how we could make this more intuitive? Maybe a poll with a thread to let people come up with ideas?

Ouaz:
> I'm actually against this, this is too much micromanaging

You think so?

I don't know, if one wants to have only a given type of city-states in a game (e.g. only Militaristic city-states), at least, he has the option to do so.

alexban011:
@yairm210 Ok so I used this PR to resolve the discussion part of #8465 and what we talked on #translations on Discord. (The actual issue was about a crash caused by a mod, but that mod has a billion warnings so I didn't bothered checking what causes the crash. My guess is that the mod should have thrown an error instead of a warning)

I (sort of) fixed the issue which causes the ExpanderTab to start closed whenever the checkboxes are clicked. The problem is that if you deselect all options it still closes the ExpanderTab, but at least this makes it better (UX wise)

Then I added the UX improvement that you can see in this video

https://user-images.githubusercontent.com/21990802/216492986-7350df7c-fd5f-49cd-853c-c2b7004be360.mp4

alexban011:
> I don't know... if one wants to have only a given type of city-states in a game (e.g. only Militaristic city-states), at least, he has the option to do so.

I still agree to this, but that would be another PR, if yairm would agree to this. Being able to choose a specific type of CS was the whole reason for this PR (before the latest changes, of course). Maybe instead of choosing whatever CS you want, it could be a screen which lets you choose what **type** of city states you want. This would be less micromanaging, right?

core/src/com/unciv/ui/newgamescreen/GameOptionsTable.kt
```diff
@@ -498,17 +509,22 @@ private class RandomNationPickerPopup(
         okButton.setPosition(innerTable.width - buttonsOffsetFromEdge, buttonsOffsetFromEdge, Align.bottomRight)
         innerTable.addActor(okButton)
 
+        val switchButton = "OtherIcons/NationSwap".toImageButton(Color.YELLOW)
+        switchButton.onClick { switchAllNations() }
+        switchButton.setPosition(innerTable.width / 2, buttonsOffsetFromEdge + 35, Align.center)
```
alexban011, line 514: the `+ 35` is just hardcoded. I couldn't understand why it wasn't aligning it with the other buttons which just use `buttonsOffsetFromEdge`. I tested on horizontal and vertical screens and both look the same.

Ouaz:
>  Maybe instead of choosing whatever CS you want, it could be a screen which lets you choose what type of city states you want. This would be less micromanaging, right?

Maybe less micromanaging, but less versatile.

With what your proposed first:

For example, one wants 4 city-states (3 cultured and 1 mercantile). One simply checks the civilopedia, then select 3 cultured and 1 mercantile (the name doesn't not matter) in the CS random pool. Not much micromanaging imo.

But that was simply a suggestion of how (or for what) the option could be used, not a request. ^^

PR #8470, opened 2023-01-26
Rework of TileGroup: split into dedicated layers, initial preparation for unit slot mechanics

vegeta1k95:
1) Big refactoring of the code related to `TileGroup` (`CityTileGroup`, `WorldTileGroup`): all the functionality of `TileGroup` has been split among new `TileLayer`s responsible for each piece of layering:
- `TileLayerTerrain` 
- `TileLayerFeatures` - roads
- `TileLayerBorders` 
- `TileLayerMisc` - improvements, resources, yields, arrows, citizens and starting locations
- `TileLayerUnitArt`
- `TileLayerOverlay` - highlight, fog, crosshair
- `TileLayerUnitFlag`
- `TileLayerCityButton`

Each layer handles it's own images/objects and updates accordingly. `TileGroup` now serves mostly to hold these together and schedule updates of each layer when needed.
This part is just a code refactor, should not produce any different behaviour from before (as far as I tested in my games).

2) I started to write `TileLayerUnitArt` and `TileLayerUnitFlag` in a way which will allow for easier change to slot mechanics (Great Admiral stuff), so it's gonna be helpful.

3) Small refactoring of `TileSetCache` class.

**VISUAL CHANGES**

4) Instead of highlighting enemies by red tiles circle, now the enemies cities and units highlighed by red stroke:
image
vs current:
image

5) Big tile yields (6-9) now have amount indicator:
image
image
Numbers 10+ have indicator but number is replaced by `*`


core/src/com/unciv/models/tilesets/TileSet.kt
```diff
@@ -0,0 +1,26 @@
...
+    var fallback: TileSet? = null
+
+    private val configs = HashMap<String, TileSetConfig>()
+
+    fun addConfig(id: String, config: TileSetConfig) {
+        configs[id] = config
+    }
+
+    fun mergeConfig(id: String) {
+        configs[id]?.let { config.updateConfig(it) }
```
yairm210, line 15: Please use val configToMerge = configs[id] :? return

core/src/com/unciv/models/tilesets/TileSet.kt
```diff
@@ -0,0 +1,26 @@
+package com.unciv.models.tilesets
+
+class TileSet(val name: String) {
+
+    var config = TileSetConfig()
+    var fallback: TileSet? = null
+
+    private val configs = HashMap<String, TileSetConfig>()
```
yairm210, line 8: What is the single config and what are these? Maybe more detailed names would help, currently I can't tell what either part is actually for

vegeta1k95: "configs" map contains all configs gathered for this particular tileset (internal + mods), keys are mod-names or DEFAULT for built-in.

"config" is an clear config which serves as a place to merge those from above. This is the actual config we are interested in the game

yairm210:
As usual, there is Too Much in this one PR...
I like the split into layers, makes it much better delineated!
I'm not sure what you're trying to accomplish with the change in the tilesets and configs

But given your prior showing I'm willing to believe that you tested this out enough to ensure that visual mods and tileset mods work properly

yairm210:
So since I only have minor comments you can fix them in your next PR and we can get this going so we can find errors sooner rather than later

PR #8474, opened 2023-01-26
Update Ukrainian.properties

Gualdimar:


Gualdimar:
found many inaccuracies in the translation
WIP

PR #8479, opened 2023-01-27
Add App store Catalan images for android

J0anJosep:
This PR adds localized android images for Catalan, as explained in https://yairm210.github.io/Unciv/Other/Translating/

This should be reviewed, as I couldn't find any related commit doing the same and I don't know if this is right.

Note: I took the images in "/extraImages/Screenshots" as a guide. They were in *.png. Later, I realized that the ones in en-US where *.jpg. If needed, I could convert the format of those files.


yairm210:
Where are the instructions for this? They should be removed.

I'm against translated images, since the UI of the game changes constantly, so if everyone sees the same images we can make sure that everyone sees *relevant* images. Translating images makes this impossible.

J0anJosep:
Ok!
Then you can close the PR.
The instructions are in the link on the first message:
https://yairm210.github.io/Unciv/Other/Translating/
in the section "App store images"

PR #8487, opened 2023-01-28
Optimized BorderedTable, refactored CityButton, new air table for units + misc

vegeta1k95:
1) Rewritten `BorderedTable` class: removed inner Images, override `draw` instead.
Less objects, less bugs, more speed.

2) Chore: Rearranged `CityButton` class and file, separated each plate component in own class: `InfluenceTable`, `DefenceTable`, `CityTable`, etc. No visual changes.

3) New look for air table for units:
imageimageimage

4) Fixes:
- removed bugged selection highlight for Policy buttons in Policy Screen
- fixed wrong alignment for circles of city units
image
- removed redundant "Fortified" action indicator (fortified status is visible via flag shape already)
image to image

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
@@ -28,6 +28,7 @@ import com.unciv.ui.utils.extensions.colorFromRGB
 import com.unciv.ui.utils.extensions.darken
 import com.unciv.ui.utils.extensions.disable
 import com.unciv.ui.utils.extensions.enable
+import com.unciv.ui.utils.extensions.isEnabled
 import com.unciv.ui.utils.extensions.onClick
 import com.unciv.ui.utils.extensions.pad
 import com.unciv.ui.utils.extensions.toGroup
@@ -37,7 +38,7 @@ import java.lang.Integer.max
 import kotlin.math.abs
 import kotlin.math.min
 
-object PolicyColors {
+private object PolicyColors {
 
     val policyPickable = colorFromRGB(47,67,92).darken(0.3f)
     val policyNotPickable =  colorFromRGB(20, 20, 20)
@@ -58,7 +59,7 @@ fun Policy.isPickable() : Boolean {
     val worldScreen = UncivGame.Current.worldScreen ?: return false
     val viewingCiv = worldScreen.viewingCiv
     if (!worldScreen.isPlayersTurn
-            || worldScreen.viewingCiv.isSpectator() // viewingCiv var points to selectedCiv in case of spectator
+            || viewingCiv.isSpectator() // viewingCiv var points to selectedCiv in case of spectator
             || viewingCiv.isDefeated()
             || viewingCiv.policies.isAdopted(this.name)
             || this.policyBranchType == PolicyBranchType.BranchComplete
@@ -71,13 +72,11 @@ fun Policy.isPickable() : Boolean {
 
 class PolicyButton(val policy: Policy, size: Float = 30f) : BorderedTable(
     path = "PolicyScreen/PolicyButton",
-    defaultBorder = BaseScreen.skinStrings.roundedEdgeRectangleSmallShape,
-    defaultInner = BaseScreen.skinStrings.roundedEdgeRectangleSmallShape,
-    borderSize = 2f
+    defaultBgBorder = BaseScreen.skinStrings.roundedEdgeRectangleSmallShape,
+    defaultBgShape = BaseScreen.skinStrings.roundedEdgeRectangleSmallShape,
 ) {
 
     val icon = ImageGetter.getImage("PolicyIcons/" + policy.name)
-    val highlight = ImageGetter.getImage("UnitFlagIcons/UnitFlagSelection")
 
     var isSelected = false
         set(value) {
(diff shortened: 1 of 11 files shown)


core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
```diff
@@ -301,9 +293,8 @@ class PolicyPickerScreen(val worldScreen: WorldScreen, civInfo: Civilization = w
 
         // Main table
         val colorBg = if (branch.isAdopted()) PolicyColors.branchAdopted else PolicyColors.branchNotAdopted
-        val branchGroup = BorderedTable(
-            path="PolicyScree/PolicyBranchBackground",
-            innerColor = colorBg)
+        val branchGroup = BorderedTable(path="PolicyScree/PolicyBranchBackground")
```
OptimizedForDensity, line 296: >"PolicyScree/[...]
Typo but not introduced in this PR.

core/src/com/unciv/ui/tilegroups/layers/TileLayerUnitFlag.kt
```diff
@@ -71,32 +80,28 @@ class TileLayerUnitFlag(tileGroup: TileGroup, size: Float) : TileLayer(tileGroup
     }
 
     private fun getAirUnitTable(unit: MapUnit): Table {
-        val holder = Table()
 
-        val iconColor = unit.civInfo.nation.getInnerColor()
-        val bgColor = unit.civInfo.nation.getOuterColor()
+        val iconColor = unit.civInfo.nation.getOuterColor()  //Color.WHITE.cpy().apply { a = 0.9f }
+        val bgColor = unit.civInfo.nation.getInnerColor() //Color.BLACK.cpy().apply { a = 0.9f }
```
OptimizedForDensity, line 85: Not something that should be put in this PR but it seems to me that there should be a standard class for icons like this on the world screen. Something like `WorldScreenCivilizationIcon(civ: Civilization, shape: [some shape enum like square or circle or diamond or rounded horizontal rectangle {pill shape?} depending on need])` that looks for the `civ`'s main and secondary colors and can be applied to all sorts of image groups.

core/src/com/unciv/ui/tilegroups/CityButton.kt
```diff
@@ -31,25 +30,334 @@ import com.unciv.ui.utils.extensions.colorFromRGB
...
+
+        if (city.isBeingRazed) {
+            val fireImage = ImageGetter.getImage("OtherIcons/Fire")
+            add(fireImage).size(18f).padLeft(2f)
+        }
+
+        if (city.civInfo == viewingCiv && city.isWeLoveTheKingDayActive()) {
+            val wltkdImage = ImageGetter.getImage("OtherIcons/WLTKD")
+            add(wltkdImage).size(18f).padLeft(2f)
+        }
```
OptimizedForDensity, line 186: Lots of `size(18f).padLeft(2f)` that seems like it should be moved to a constant somewhere.

core/src/com/unciv/ui/tilegroups/CityButton.kt
```diff
@@ -31,25 +30,334 @@ import com.unciv.ui.utils.extensions.colorFromRGB
...
+
+        val normalizedInfluence = max(-60f, min(influence, 60f)) / 30f
+
+        val color = when (relationshipLevel) {
+            RelationshipLevel.Unforgivable -> Color.RED
+            RelationshipLevel.Enemy -> Color.ORANGE
+            RelationshipLevel.Afraid -> Color.YELLOW
+            RelationshipLevel.Neutral, RelationshipLevel.Friend -> Color.LIME
+            RelationshipLevel.Ally -> Color.SKY
+            else -> Color.DARK_GRAY
```
OptimizedForDensity, line 57: Looks like `DiplomacyManager.RelationshipLevel` already has these colors coded so they might be worth reusing.

core/src/com/unciv/ui/tilegroups/CityButton.kt
```diff
@@ -31,25 +30,334 @@ import com.unciv.ui.utils.extensions.colorFromRGB
...
+
+        val textColor = city.civInfo.nation.getInnerColor()
+
+        val cityConstructions = city.cityConstructions
+        val cityCurrentConstruction = cityConstructions.getCurrentConstruction()
+
+        val progressTable = Table()
+
+        var percentage = 0f
+        var turns = "∞"
```
OptimizedForDensity, line 325: Reminds me that this character should probably be a constant but that's a random thought and not a critique.

core/src/com/unciv/ui/tilegroups/layers/TileLayerUnitFlag.kt
vegeta1k95: World scree Units icons are visualized already by `UnitGroup` class, which selects shape and color based on type/status/civ.

core/src/com/unciv/ui/tilegroups/CityButton.kt
vegeta1k95: Those colors are different, they are used for diplomacy/leader screen.
E.g. `Ally` color for diplomacy is green, while `Ally` indicator for influence bar should be blue.

core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
vegeta1k95: fixed

OptimizedForDensity:
Do you by chance have a before/after screenshot with a bunch of units in the picture? I'm curious to see how the changes look when a game is played out and the screen is covered with units. Everything looks good otherwise.

vegeta1k95:
Before:
image
After:
image

PR #8488, opened 2023-01-29
Mod Scripting Proof of Concept

OptimizedForDensity:
This is the first of a possible series of pull requests that would implement mod scripting capabilities. In short, modders would be able to write Kotlin script code (.kts files) that gets compiled by the game then called at different points such as triggered uniques or at fixed points (map gen, AI for specific units, start of game, start/end of turn, etc). This code is written exactly as the regular Unciv .kt code which makes it easy to pick up if you're familiar with the code base. I've tested this on both desktop and Android and script execution works on both.

# Demonstration

The following code is a demonstration of scripting capabilities and can be reproduced in this PR branch.

In a file `mods/MyMod/scripts/MyScript.unciv.kts` (arbitrary script name with `.unciv.kts` extension),
```
import com.unciv.logic.city.City
import com.unciv.logic.civilization.Civilization
import com.unciv.logic.civilization.NotificationCategory
import com.unciv.logic.map.tile.Tile
import com.unciv.models.ruleset.unique.Unique
import com.unciv.models.translations.tr

@Suppress("UNUSED_PARAMETER")
fun main(unique: Unique,
         civilization: Civilization,
         city: City?,
         tile: Tile?,
         notification: String?,
         triggerNotificationText: String?
) {
    println("You can see me in the console even though I'm called from a script!")

    civilization.addGold(1000)
    civilization.addNotification("This is a notification generated in a script {${triggerNotificationText!!}}".tr(), NotificationCategory.General)
    civilization.cities.filter { it.civInfo.isMajorCiv() }.forEach {
        it.isBeingRazed = true
    }
}

```

and add a unique to a building like a granary
`"Activates script [mods/MyMod/scripts/MyScript.unciv.kts]"`
and it would result in the following outcome when the granary is built.

Scripting

There is additional error reporting separate from Unciv error reporting to let the user know that a script (and not the game) crashed and what line it crashed. The user can then copy the output to their clipboard to shove in the mod maker's face rather than in an official bug report. If we took the above script and added `val x = 1 / 0` on line 18 (compiler errors are generated separately during compilation so in this case the modder would know that this line would likely cause an error even before running the code), the user would see

ScriptingException

# Technical details

This implementation uses a built-in experimental Kotlin scripting engine that uses the Kotlin-to-JVM compiler to compile Kotlin source code to JVM bytecode. Further details and other example code can be found [here](https://github.com/Kotlin/KEEP/blob/master/proposals/scripting-support.md) which is a mixture of currently-implemented features and a wishlist of future ones. When the script is executed, a virtual machine is started in a separate process and runs the code. We can pass variables to it by reference which is how users can modify the game state. I've set up the system in a way that once the code is compiled, it's written to a .jar file for execution at a later time so users don't have to constantly compile code. At present, the system will place the .jar file in the same directory as the source code and with the same name except for an appended ".jar" extension. Theoretically these JARs could be compiled by mod authors then distributed with other mod files so users can play right away. The libraries needed to *run* the script have a fairly minimal size difference compared to current production files, for instance the difference between this PR's Linux desktop Unciv.jar file being only 3 MiB larger than the current release version of the same build, but the compiler library needed to compile the script (package "kotlin-compiler-embeddable" in Gradle) is a whopping additional 53 MiB which more than doubles the size of the .jar file. This means that the compiler library is optional and does not need to be distributed to everyone to benefit from scripting and the compiler library could be restricted to an optional download or as coded in this PR, only distributed with the desktop build variant and not with Android.

There is one note that I will mention here, Android .jar archives need an additional `classes.dex` file located in the root of the .jar. This can be generated using Android built tools (the D8 JAR that's distributed with Android build tools does exactly this and you can execute it with the following line of code

`java -cp ~/path/to/build-tools-lib/d8.jar com.android.tools.r8.D8 --release --output [outputDirectory] [MyScript.unciv.kts].jar --no-desugaring`

This would need a further step in order to fully automate mod script compilation for distribution or use on Android.

# Security

Mod scripts can run nearly-arbitrary code, and while I've restricted users' ability to import libraries not present in the Unciv runtime, they still have access to Java and Kotlin standard libraries including the ability to manipulate files and open internet connections. We obviously don't want users wrecking other people's systems, but we can implement some features to scan source code to detect certain imports (`java.io.file`, etc.) and reject the script for compilation. That probably wouldn't stop every malicious intent, especially if users can distribute pre-compiled .jars, so an alternative would be requiring everyone who wants to use a script to have to compile locally. For PC users, this might entail using a different Unciv.jar file but Android users might require different treatment like a non-Play Store .apk that would contain the compiler libraries and files so they could play with scripting. This entails further discussion of course.

# Play Store/Steam Permission

Play Store doesn't let apps download certain files like self-executing binaries unless those binaries come from Play Store. This is obviously an issue with modder-created binaries and something that would require further inquiry. The binaries aren't self-executing and can't be run without Unciv but in the eyes of Google this might not make a difference. I don't know Steam's policy either but since we're trying to publish there, this might also be forbidden. This is a pretty powerful mechanic that could really enhance modding capabilities but these are other factors which need to be addressed beyond simple security.

# Todo

As of this PR, scripting is functional although not user friendly. The following is my rough to do list which may or may not be finished by the time this PR gets merged, especially if Google or Steam says otherwise and it'd be too difficult to separate the scripting engine from the regular Unciv engine.

- [ ] A way to notify users that a mod contains scripts (and optionally explicit permission to run either in general or on a mod-by-mod basis)
- [ ] A screen where users can mass compile scripts in a mod (probably the mod's tab in the mod download screen with a "Compile All" button and a second button to compile any non-compiled scripts (as in, ones that the modder is currently working on and wants to make sure they compile before moving on to the next section). Ideally there'd be graphical feedback with this process like a loading bar. The compiling process requires a `suspend` function so we can easily tie it into graphics (or so I think...)
- [ ] Separating static analysis/compiler messages from execution messages in case a user has to compile then execute a script instead of compiling it on a separate screen.


core/src/com/unciv/logic/UncivScript.kt
```diff
@@ -0,0 +1,582 @@
...
+            .filter { it.severity >= minimumReportThreshold }
+            .sortedByDescending { it.severity }
+    }
+}
+
+/**
+ * Inherits [CompiledScriptJarsCache] to modify caching so we can save in the jar manifest then check
+ * Unciv version for any potential incompatibilities.
+ */
+@Suppress("UNCHECKED_CAST")
```
OptimizedForDensity, line 411: Below this point, there's tons of code that is subject to removal. This is an experimental feature within the Kotlin engine with very little documentation so I had to do a lot of testing and debugging to figure out how things worked.

yairm210:
I appreciate you looking into this, but I am **really** against this is so many ways.

- Debugging nightmare
- Operating system permission hell ("Why doesn't this mod work on my new Android version 15? Oh look at that it's a security hazard who would have thought")
- Security as you mentioned (Arbitrary code execution is the mother of all CVEs and that's basically what we're proposing here)
- And finally, if anyone is at the level that they can make something unique for their mod, my grubby hands want it in the main codebase :)

In short: no

In long: ~noooooo~ This feels to me like Chrome native code extensions all over again, and wherever it isn't patched currently it probably will be eventually. The benefits from this are very small (it's not like we don't accept PRs for uniques!) and the risk is very large. 

I feel that this is a bad direction for our modding community as well - one of the things we've been very careful about is making the modding experience available to non-coders, and this is adding another gate to separate the "real mods" with code from the codeless mods, which I feel would be detrimental

PR #8494, opened 2023-01-29
Better World-Wrap

vegeta1k95:
**TL;DR:** Instead of creating mirror copies, move tiles themselves: override `TileGroupMap.draw()` all is needed.

**Explanation:**

Okay, so the logic of the new implementation is stupidly easy: 

1) Instead of creating mirror tiles on the sides, we just check map viewport boundaries and if viewport goes beyond tilemap borders we move tiles from the opposite side. 

That is:
- When camera viewport reaches rightmost tiles, a column of leftmost tiles is moved to the right.
- When camera viewport reaches leftmost tiles, a column of rightmost tiles is moved to the left.
- When camera center reaches map borders, camera skips to opposite side (as before)

Tiles are moved in columns and only when camera viewport exceeds map boundaries, so no unnecessary moves in the middle of the map or bulk moves.

2) For world-wrap we limit zoom such that the camera viewport does not exceed tilemap size - we don't want to render same tiles multiple times.

The code itself is very easy:
Point 1) is implemented inside `TileGroupMap.draw()` (and that is all that is actually needed)
Point 2) is implemented inside `ZoomableScrollPane.zoom()`
the rest of code changes are just cleanup of mirror tiles.

Clickhandlers, actions, map generation, pathfinding and everything else is preserved and works correctly (as far as I tested).

**Testing:** I tested this on my `Google PIxel 4` and the perfromance difference is huge. New implementation can handle huge world-wrap worlds just as fast as usual worlds. This needs more testing on different devices, but I think this will HUGELY improve memory consumption and performance for world-wrap worlds.


vegeta1k95:
@GGGuenni Tag you

OptimizedForDensity:
I just tested this on a tiny pangaea map with human as spectator, 2 AI civs, 1 city state, hexagonal world shape, and when I loaded in I got a ton of flickering on the sides but I don't know if that's specific to my system. Otherwise the rational makes a ton of sense and it seems like a good addition.

vegeta1k95:
> I just tested this on a tiny pangaea map with human as spectator, 2 AI civs, 1 city state, hexagonal world shape, and when I loaded in I got a ton of flickering on the sides but I don't know if that's specific to my system. Otherwise the rationale makes a ton of sense and it seems like a good addition.

Flickering happens when camera viewport is bigger than map, so tiles at sides of the screen rendered (and moved) multiple times. I've had this issue only on Tiny maps with full zoom out. I thought I fixed it, seems like some weird edge case. But it can be fixed

GGGuenni:
There will probably be a few people complaining about the constrained zoom, especially those who have set the max zoom out to 6 but it's a relatively small trade-off compared to the performance boost

> I got a ton of flickering on the sides

On desktop with screen size huge, this is not only visible on tiny but also on small maps so definitely needs to be fixed before merging

vegeta1k95:
> There will probably be a few people complaining about the constrained zoom, especially those who have set the max zoom out to 6

I don't get the idea why would you want to see the same part of the map rendered 2+ times, it is very confusing and I haven't seen this in any strategy game with a world-wrap so far, everywhere viewport is restricted so that you don't render same map parts multiple times.

> On desktop with screen size huge, this is not only visible on tiny but also on small maps so definitely needs to be fixed before merging

Yes, I am working on it.

GGGuenni:
> I don't get the idea why would you want to see the same part of the map rendered 2+ times

I don't think the goal is to be able to see tiles multiple times but to get a better overview of the map by showing more vertical space

vegeta1k95:
@yairm210 @GGGuenni 
Okay, this should have fixed all the flickering on the sides, for all map sizes and screen sizes combinations (at least on my devices). Can you test is out and confirm that it is also OK for you?

yairm210:
Tested locally, looks good!

PR #8506, opened 2023-01-30
Add next-turn-progress bar for growth/production on CityButtons

vegeta1k95:
Add additional coloring to CityButton's growth/construction bars, which denote the percentage of completion on the next turn:
image
image
image

Solid color - current, darkened color - on next turn.


yairm210:
How about starving cities?

vegeta1k95:
> How about starving cities?

Not shown currently for starving cities.

PR #8508, opened 2023-01-31
Performance Improvements to Construction Automation

OptimizedForDensity:
Reduces time to do construction automation by around 10-20% through some caching of the results of `INonPerpetualConstruction.isBuildable()` which is pretty costly (partly because it executes the entire buildability check rather than just returning early since `isBuildable()` will return false if there's even a single rejection reason but I plan to tackle that problem with sequences) as well as the return of sequences to the class since I messed up a few months ago #7504. The caching works because the construction automation operation is performed once per city per turn so there's no worrying about an action being performed that causes the cache to become out of date. Also make a few other tweaks to some logic that I'm willing to change if they're not desirable:

* `@Transient` `uniqueUnits` and `uniqueBuildings` hash sets given to each `Civilization` to make it easier to tell which units/buildings are their unique ones. Previously the game would iterate over the entire units/buildings list in the ruleset to figure out which ones were their uniques.
* Skip space part construction logic if the civ doesn't have the `UniqueType.EnablesConstructionOfSpaceshipParts` unique.
* Rearranged/rewrote some logic in `Automation.chooseMilitaryUnit`.
* AI considers wonders with `UniqueType.EnablesNuclearWeapons` as favorable for military focus rather than looking for a wonder with the name "Manhattan Project".
* Renamed `NextTurnAutomation.reassignWorkedTiles()` to `automateCities()` since this function does a bit more than reassigning worked tiles.
* Slight reordering of `NextTurnAutomation.automateCities()` since the AI will build military units in a city if it's under max health. If this is the case, there's no point in doing the entire construction automation calculation.

I also have at least 2 more PRs related to this one that I will work on so stay tuned for those.


core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -107,7 +107,9 @@ object Automation {
 
     fun tryTrainMilitaryUnit(city: City) {
         if (city.isPuppet) return
-        val chosenUnitName = chooseMilitaryUnit(city)
+        if (city.civInfo.gameInfo.ruleSet.units[city.cityConstructions.currentConstructionFromQueue]?.isMilitary() == true)
```
yairm210, line 110: There's a convenience function in cityConstructions to get the construction

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
```diff
@@ -24,11 +24,17 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
     private val cityInfo = cityConstructions.city
     private val civInfo = cityInfo.civInfo
 
-    private val buildings = cityInfo.getRuleset().buildings.values
+    private val buildings = cityInfo.getRuleset().buildings.values.asSequence()
+
     private val nonWonders = buildings.filterNot { it.isAnyWonder() }
+        .filterNot { buildableBuildings[it.name] == false } // if we already know that this building can't be built here then don't even consider it
```
yairm210, line 30: Surely buildableBuildings will not be initiated at this point yet?

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
```diff
@@ -57,8 +63,14 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
         choices.add(ConstructionChoice(choice, choiceModifier, cityConstructions.getRemainingWork(choice)))
     }
 
-    private fun Collection<INonPerpetualConstruction>.isBuildable(): Collection<INonPerpetualConstruction> {
-        return this.filter { it.isBuildable(cityConstructions) }
+    private fun Sequence<INonPerpetualConstruction>.isBuildable(): Sequence<INonPerpetualConstruction> {
```
yairm210, line 66: This should really be named filterBuildable. I know it was already like this, but if we're already fixing...

core/src/com/unciv/logic/civilization/Civilization.kt
```diff
@@ -185,6 +185,12 @@ class Civilization : IsPartOfGameInfoSerialization {
      */
     val temporaryUniques = ArrayList<TemporaryUnique>()
 
+    /** Easy way to look up a Civilization's unique units and buildings */
+    @Transient
+    val uniqueUnits = hashSetOf<BaseUnit>()
```
yairm210, line 190: Please put these in the cache, we're trying to get rid of transients in the main class :)

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
OptimizedForDensity: That's a sequence so that predicate would be evaluated after the object is initialized but I went ahead and moved `buildableBuildings` above this statement and did the same with the unit equivalent too.

core/src/com/unciv/logic/automation/Automation.kt
OptimizedForDensity: Done.

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
OptimizedForDensity: That's a good idea.

core/src/com/unciv/logic/civilization/Civilization.kt
OptimizedForDensity: Done.

PR #8509, opened 2023-01-31
Notification log fix

### LINKED ISSUE #8501: String is not translated and Notification bubble problem
Platform
Windows and Android

Version
GitHub, 4.4.7, jar(Windows)
Play Store, 4.4.5(Android)

This string is not translated by the game (with any language selected):
The city will not produce anything. 

Notification bubble is not displayed correctly for this string (with any language selected):
An ancient tribe trained us in their ways of combat!

OptimizedForDensity:
Closes #8501 by translating the "This city will produce nothing" text as well as fixing the bugged notification bubble. I also found a separate bug where ruin rewards had a faulty check for "Only available <conditional>" uniques that caused the conditional to never be considered properly. This is fixed as well.

### DIFF
--- a/core/src/com/unciv/ui/overviewscreen/NotificationsOverviewTable.kt
+++ b/core/src/com/unciv/ui/overviewscreen/NotificationsOverviewTable.kt
@@ -88,7 +88,7 @@ class NotificationsOverviewTable(
             for (notification in categoryNotifications) {
                 val notificationTable = Table(BaseScreen.skin)
 
-                val labelWidth = maxEntryWidth * notification.icons.size - 10f
+                val labelWidth = maxEntryWidth - iconSize * notification.icons.size - 10f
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/ui/overviewscreen/NotificationsOverviewTable.kt
```diff
@@ -88,7 +88,7 @@ class NotificationsOverviewTable(
             for (notification in categoryNotifications) {
                 val notificationTable = Table(BaseScreen.skin)
 
-                val labelWidth = maxEntryWidth * notification.icons.size - 10f
+                val labelWidth = maxEntryWidth - iconSize * notification.icons.size - 10f
```
yairm210, line 91: Are you sure you're not missing () here?

OptimizedForDensity: Do you mean parentheses as in for order of operations?

PR #8510, opened 2023-01-31
Add Conditional Stats Costs to Transform

itanasi:


yairm210:
I think maybe we were talking at cross-purposes
What I imagine for this is *modifiers that can be added to any unit action*
Like 'destroys unit', 'consumes resources'... costs health, maybe? Dunno
But these should be things we can add to any unit action, to make it moddable like the triggered uniques

If that sounds too complex to you we can discuss how this can work, I have some ideas :)

Basically:
- UnitActionModifier as a new 'conditional-type' of Unique
- Each unit action can become unavailable based on above
- Each unit action has extra effects based on above
- All the magic happens in the UnitAction class so we don't need to add things in each individual action

itanasi:
I think the main difficulty for me is seeing how you'd list these out for mods. Like, we'd need a Unique that captures both the Action amd the new Cost, covering Unit, UnitType, and perhaps Base Rules. If that was more clear to me maybe I could finagle the underlying systems.

I initially tried to add the conditional just to explicitly listed Unique Actions, but that got bogged down quickly and especially with the Setup Action types (Paradrop and Air Sweep) you need to pass along that cost information to the execution function somehow. Which likely means leaving Transient Data on the Unit object?

Anyways, would need to spend more time design documenting this

PR #8512, opened 2023-01-31
Block repairs in Hostile Tiles

itanasi:
So you can't spam Repair and Pillage while besieging foreign lands


ravignir:
but why?

ravignir:
it is a civ5 mechanic.

ravignir:
it is even more effective in civ5.

https://user-images.githubusercontent.com/30041793/215749546-bf6aa0b8-7cef-4ccf-b019-f85a5708ec01.mp4

itanasi:
Eh just saw some folks seeming unhappy with the tactic. Can drop

PR #8513, opened 2023-01-31
Add back in Random Map

itanasi:


yairm210:
Why?

yairm210:
We removed this because it's unclear what such a map means

yairm210:
I really don't think there's a need for an extra map type called "random" when we now have "generate a random map type"

PR #8514, opened 2023-01-31
Block temperatureShift adjustment unless Easter Egg date

itanasi:
THIS IS OPTIONAL

I noticed the temperature adjustment doesn't have a qualifier to check that it's actually an Easter Egg date, so we are actually always applying the temp shift for each month of the year.

It's kinda a neat Easter Egg some have noticed. But if we want to hide it away... there's this PR.

I didn't qualify the extra Richness since the Main Menu map should probably have some more exotic features anyways


yairm210:
I think it's fine as is
And it is cool, I myself didn't notice this :)

PR #8519, opened 2023-01-31
 fix #8515 - reload world screen after tileset/unitset change from in main menu screen

### LINKED ISSUE #8515: Changing tileset or unitset doesn't take effect until exiting and reopening the app.
**Platform**

Android, 13, Arm 64.

**Version**

I built the game from source code, Version 4.4.7.

**Describe the bug**

When user changes the `Tileset` or `Unitset` from main menu the changes are reflected on main menu screen (in case of tile set) but they do not affect the world screen, user has to exit the app and reopen to see the changes on world screen. 

**To Reproduce**

1. Go to `Options` button in main menu.
2. Go to `Display` section of the `Options`.
3. Change the `Tileset` and/or `Unitset`.
4. Go to world screen/resume the game.
5. Check that the `Tileset` and/or `Unitset` is not changed.

```

```
</details>

**Expected behavior**
`Tileset` and `Unitset` should change in game when user changes them in `Options` menu.

**Screenshots**

**Additional context**

ryg-git:
may fix #8515

### DIFF
--- a/core/src/com/unciv/ui/components/tilegroups/TileSetStrings.kt
+++ b/core/src/com/unciv/ui/components/tilegroups/TileSetStrings.kt
@@ -17,6 +17,8 @@ class TileSetStrings(tileSet: String = UncivGame.Current.settings.tileSet, unitS
 
     // this is so that when we have 100s of TileGroups, they won't all individually come up with all these strings themselves,
     // it gets pretty memory-intensive (10s of MBs which is a lot for lower-end phones)
+    val tileSetName = tileSet
+    val unitSetName = unitSet
     val tileSetLocation = "TileSets/$tileSet/"
     val unitSetLocation = "TileSets/$unitSet/"
     val tileSetConfig = TileSetCache[tileSet]?.config ?: TileSetConfig()
--- a/core/src/com/unciv/ui/screens/mainmenuscreen/MainMenuScreen.kt
+++ b/core/src/com/unciv/ui/screens/mainmenuscreen/MainMenuScreen.kt
@@ -47,6 +47,7 @@ import com.unciv.ui.screens.newgamescreen.NewGameScreen
 import com.unciv.ui.screens.pickerscreens.ModManagementScreen
 import com.unciv.ui.screens.worldscreen.mainmenu.WorldScreenMenuPopup
 import com.unciv.ui.screens.mainmenuscreen.EasterEggRulesets.modifyForEasterEgg
+import com.unciv.ui.screens.worldscreen.WorldScreen
 import com.unciv.utils.concurrency.Concurrency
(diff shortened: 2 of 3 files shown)


alexban011:
I don't think this will work. The game requires a restart in order to rebuild the atlas. I'm not sure how mods are taken into account but my guess is that those textures also need to be baked.

ryg-git:
Maybe on `resetToWorldScreen` (the method called when `Resume` is pressed) we can check if `GameSettings` is changed and `reloadWorldscreen` if it is changes.

alexban011:
No, just reloading the worldscreen won't work.

What I meant is that the textures need to be packed into atlases and then the atlases will be loaded. This is being done at UncivGame.kt:131 in the create() function. This means the game only knows about the new atlas after creating the game. So you'd need to reset the atlases, but if you do so the game will "forget" the old one and everything will become dark.

ryg-git:
If user changes the `Tileset` / `Unitset` when user is on `WorldScreen` by using `WorldScreenMenuPopup` > `Options`, the game (`UncivGame.Current`) is reloading according to [Line 149](https://github.com/yairm210/Unciv/blob/1b22d810c56a065972fd5b7ad93f318bc7eb2a33/core/src/com/unciv/ui/options/OptionsPopup.kt#L148) of `reloadWorldAndOptions()`. This is also mentioned in [this comment](https://github.com/yairm210/Unciv/issues/8515#issuecomment-1410115001).

So I think reloading the `UncivGame` in case user is on `MainMenuScreen` could work. let me know what you think.

alexban011:
If that function would actually reload UncivGame then it might, but that only reloads the worldscreen. It doesn't load the mod atlas. There is no need to reload the worldscreen since that is already being done when selecting the tileset. 

If you put `mod.folderLocation?.let { ImageGetter.loadModAtlases(mod.name, it) }` at the end of the code of the `visualCheckbox` checkbox in `refreshInstalledModActions(mod: Ruleset)` of ModManagementScreen.kt it will load the atlas when selecting the mod as permanent audio visual. From my testing it works fine, but deselecting the option obviously won't remove the atlas. 

I've messed around with this a bit and this code will do what you want, but with the problem mentioned above and the deselecting the permanent audiovisual mod option doesn't update the worldscreen. 

```Kotlin
            if (checked)
                mod.folderLocation?.let { ImageGetter.loadModAtlases(mod.name, it) }
            else {
                    ImageGetter.getAvailableTilesets().minus(mod.name)
                    // UncivGame.Current.reloadWorldscreen() doesn't reload the tileset so I'll let you figure this part out yourself :)
                   // Tbh it doesn't really matter that much since deleting the mod also doesn't replace the worldscreen 
                    }
            }
```

PR #8522, opened 2023-01-31
Fix Faith Healers healing enemies

OptimizedForDensity:
Fixes a bug mentioned on Discord where the Faith Healers belief healed enemy units. ~To do this, I implemented `Ally` and `Enemy` entries `mapUnitFilter`. Ally checks if either `MapUnit.civInfo` or the `otherCiv` argument supplied to `matchesFilter` is a city state and allied to the opposite party (also returns true if `civInfo == otherCiv`. Enemy just checks if `otherCiv` and `civInfo` are at war with each other.~

I will try to implement the new mapUnitFilter types later since I would need to change more uniques than just this one and I'd rather get the bug fix out.

### DIFF
--- a/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
@@ -883,11 +883,7 @@ class MapUnit : IsPartOfGameInfoSerialization {
     }
 
     /** Implements [UniqueParameterType.MapUnitFilter][com.unciv.models.ruleset.unique.UniqueParameterType.MapUnitFilter]
-     * @param otherCiv: The other civilization which may not be the same Civilization that owns this MapUnit.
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/logic/map/mapunit/MapUnit.kt
```diff
@@ -876,6 +876,12 @@ class MapUnit : IsPartOfGameInfoSerialization {
     }
 
 
+    private fun isAlly(otherCiv: Civilization?): Boolean {
```
yairm210, line 879: When would this get null as an input? Sounds to be like it's better off not nullable, OR checking that explicitly first

OptimizedForDensity: Good catch, that was from an earlier commit before I ripped out a bunch of code and forgot to change it back to non-null :D

PR #8538, opened 2023-02-02
Performance optimization: don't draw empty tile layers

vegeta1k95:
**Big rendering optimization:** on each `TileGroup.update()` tile's layers which have nothing to show to the user set `isVisible` to `false`. This means they are skipped over inside of `TileGroupMap.drawChildren()` loop, including all calls to `transform` etc.

Previously all explored tiles had all their layers visible, that means:  `NUM_EXPLORED_TILES_IN_VIEWPORT * 9` calls to draw and transform (9 = 8 layers + 1 group itself for clickability).
Now only layers with visible content are counted, that means e.g.: if tiles have only terrain and nothing else it will be only `NUM_EXPLORED_TILES_IN_VIEWPORT * 2` calls to draw and transform (2 = 1 terrain layer + 1 group itself for clickability).

Overall this hugely reduces (by more than half) a number of draw calls lategame on big maps and increases rendering FPS.

**Testing:** My tests on Google PIxel 4 gave me an improvement around 13-25 FPS on huge fully explored maps (full zoom out), while moving camera around.

### DIFF
--- a/core/src/com/unciv/ui/tilegroups/CityTileGroup.kt
+++ b/core/src/com/unciv/ui/tilegroups/CityTileGroup.kt
@@ -2,6 +2,7 @@ package com.unciv.ui.tilegroups
 
 import com.unciv.UncivGame
 import com.unciv.logic.city.City
+import com.unciv.logic.civilization.Civilization
 import com.unciv.logic.map.tile.Tile
 import com.unciv.ui.images.ImageGetter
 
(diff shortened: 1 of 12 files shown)


yairm210:
Amazing!!! I remember I did something like this for the city button layer on zoom in, which uses scale, but this is a whole other level :D

PR #8549, opened 2023-02-02
Performance optimization: fast and smooth zoom

vegeta1k95:
**Performance optimization:**

This implements custom GestureListener for additionally listening to `pinch` and `pinchStop` input events, and disable `act()` and `hit()` methods of `TileGroupMap` while zooming on touchscreens (Android). 

This result in a very smooth and fast zoom, without FPS throttling (tested on Google Pixel 4).

P.S: With this optimization, performance/rendering difference between usual map and world-wrap is miniscule (if exists at all). Might as well remove "memory intensive" warning from "Start New Game" screen.

### DIFF
--- a/core/src/com/unciv/ui/utils/ZoomGestureListener.kt
+++ b/core/src/com/unciv/ui/utils/ZoomGestureListener.kt
@@ -1,0 +1,80 @@
+package com.unciv.ui.utils
+
+import com.badlogic.gdx.input.GestureDetector
+import com.badlogic.gdx.math.Vector2
+import com.badlogic.gdx.scenes.scene2d.Event
+import com.badlogic.gdx.scenes.scene2d.EventListener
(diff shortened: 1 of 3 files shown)


yairm210:
I'll be honest I don't understand what you did here
What was happening before? And what's happening now?

vegeta1k95:
@yairm210 
Before, nothing was happening.

I added to `WorldMapHolder` a listener to `pinch` (two finger touch, when user starts to zoom) and `pinchStop` (when user stops zoom) input events. 

When `pinch` starts, I disable `TileGroupMap`'s `act` and `hit`, so rendered don't spend time on actions and hit detection when user just zooms.
When pinch stops, I re-enable those.

It is the same logic that already present for panning, but now I did it also for zooming.

yairm210:
😎

PR #8552, opened 2023-02-02
Fix unresponsive policy screen header

### LINKED ISSUE #8547: Feature request: Policy Tree more Clickable
v4.4.7 on Android

It's very difficult to accurately click on the Policy Tree headings to bring down and fold back up the description for Adopting and Completion on mobile with fingers. Seems only the text and button perhaps are clickable? Haven't tested on PC.

Would be nice if the whole bar was clickable. And perhaps thicker for ease on small screens.

Gualdimar:
Partially resolves #8547
Not sure about the header thickness...

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
@@ -3,6 +3,7 @@ package com.unciv.ui.pickerscreens
 import com.badlogic.gdx.graphics.Color
 import com.badlogic.gdx.scenes.scene2d.Actor
 import com.badlogic.gdx.scenes.scene2d.Group
+import com.badlogic.gdx.scenes.scene2d.Touchable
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/ui/pickerscreens/PolicyPickerScreen.kt
```diff
@@ -562,6 +563,8 @@ class PolicyPickerScreen(val worldScreen: WorldScreen, civInfo: Civilization = w
         table.add(expandIcon).minWidth(15f).expandX().left()
         table.add(branch.name.tr().uppercase().toLabel(fontSize = 14).apply { setAlignment(Align.center) }).center()
         table.add(icon).expandX().left().padLeft(5f)
+        
+        table.setTouchable(Touchable.enabled);
```
yairm210, line 567: Remove the ;

Gualdimar: done

Issue #7143, opened 2022-06-12
Feature request: Colapsable Tutorial Tabs


ArchDuque-Pancake:
I decided to install Unciv in my old potato phone again to test some stuff.

When playing, the tutorials are displayed right and all good, but the problen is that they block an important part of the map view, and some of them may take some time until they are done (or until the new player figures out). 
So they stay there blocking the view, and altough that may not be an issue in higher resolutions, it is in lower ones, and in a small phone you cant really increase the resolution or else the buttons will be incredibly hard to press.

So what I propose is for the tutorials to collapse once tapped, when collapsed, only the title will display.
So, you tap the tutorial - collapses and only shows the title (i.e. Move a plane!), and when you tap again it shows the full text.

One example of tutorials that take the most to complete is the "Move a Plane!", as it may take very long to get oil or planes.

As a side note: I noticed that the "World Screen" tutorial is only shown in the Civilopedia, I think it would be very useful to show it in the early tutorials.

1etsp1ay:
I'd rather use the social policy to implement an active tutorial system ... learn
1. mining - going from copper to gold to iron 
2. herb hunting - looking for cotton to coffee to ??? ultimately upeing up say an inn
3. searching for strategic items

That way you can pick the parts which are interesting rather than wade through all the information which otherwise could be in civipedia

RobLoach:
While not expandable, there is some excellent work going on to introduce Civilopedia sub-categories over at https://github.com/yairm210/Unciv/pull/14415 .

PR #8559, opened 2023-02-03
Suggestion: hide tutorial button

Gualdimar:
in response to #7143
tutorial titles still take up a lot of space on the screen. 
my suggestion is to add a button next to the minimap when a tutorial is available. 

screenshots:
<details><summary>no tut available - no button</summary>

</details>
<details><summary>tut is shown- button is bright green</summary>

</details>
<details><summary>tut is hidden- button is dimmed</summary>

</details>
<details><summary>on the tiny screen size, default minimap size</summary>

</details>
<details><summary>on the small screen size, default minimap size</summary>

</details>

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -101,6 +101,8 @@ class WorldScreen(
     var fogOfWar = true
         private set
 
+    var isTutorialShown = true
+    
     /** `true` when it's the player's turn unless he is a spectator*/
     val canChangeState
         get() = isPlayersTurn && !viewingCiv.isSpectator()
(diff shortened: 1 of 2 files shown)


Gualdimar:
need to add: reset the isTutorialShown flag after the tutorial is completed

yairm210:
Nope. If I disabled tutorials this should not take up permanent area on my main screen.
This is a convenience function to an already niche case so I'm not giving it any screen area

Gualdimar:
this button is not a duplicate of the checkbox from the options. if tutorials are turned off, the button won't show.
it is for situations like:
1. tutorials are turned on
2. i`m doing tutorials one by one and got to something like "Meet another civilization!", which will be shown on the screen for an unknown number of turns and obstruct the view
3. i can click the button next to the minimap to free up screen space
4. if I forget what I'm supposed to do, I can press the button again and the tutorial will return to the screen
5. after completing the current tutorial , a new one WILL appear on the screen

this will free up screen space temporarily, without the need to disable tutorials globally

yairm210:
That's the level of complexity that someone in the tutorial stage needs to deal with? I now better understand the situation, I'm still not convinced this is a good change. Maybe have the tutorial panel itself collapsable to an icon when clicked, since that would use the same screen space that's already used

Gualdimar:
like this? 439e380aa5d36e1f1ef6c6af5677e4dc9341add7
<details><summary>screens</summary>


</details>

PR #8560, opened 2023-02-03
UI Double/Multi Click Functionality and Double Click on Tech to Exit Tech Screen

OptimizedForDensity:
Extends existing Unciv functionality to allow for double or multiple clicks to trigger functions separate from single click `onClick()` instances. A premade function `Actor.onDoubleClick()` was added so devs can easily add double click functionality the same way you use `onClick()`. By default, clicking twice within a 250 millisecond window triggers a double click although this can be changed as default and on a button-by-button basis. Additional clicks (with the option for separate sounds) are set up to use an existing `ClickListener` object to prevent multiple click listeners from firing at the same time, and the ability to specify a multi click without a single click is also provided. In the event that many clicks happen within one `tapInterval`, the highest available listener instance with a `tapCount` below the actor's current tap count will be used, and if none exist then nothing will happen.

As a demonstration case, double clicking on a tech button in the tech tree will trigger the existing tech selection function as well as the existing exit screen functionality on the second click which makes it convenient for veteran players to select techs, and this has been tested on both Linux and Android.

Demonstration of double click to change the tech being researched as well as a single click vs double click in the tech screen

https://user-images.githubusercontent.com/105244635/216510528-114bd3a4-c472-4f39-ae8c-a61dde8b1226.mp4

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
+++ b/core/src/com/unciv/ui/pickerscreens/TechPickerScreen.kt
@@ -23,6 +23,7 @@ import com.unciv.ui.utils.extensions.colorFromRGB
 import com.unciv.ui.utils.extensions.darken
 import com.unciv.ui.utils.extensions.disable
 import com.unciv.ui.utils.extensions.onClick
+import com.unciv.ui.utils.extensions.onDoubleClick
 import com.unciv.ui.utils.extensions.toLabel
 import com.unciv.utils.concurrency.Concurrency
 import kotlin.math.abs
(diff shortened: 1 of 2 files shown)


vegeta1k95:
Have you tested it on touchscreen devices? Does it compatible with pan/zoom/longPress functionality?

OptimizedForDensity:
Just now on Android I tested zoom (two fingers spreading apart), pan (two fingers moving together), and long press (one finger holding on the same spot) and none of these triggered a double click which is what I would expect.

yairm210:
Looks good!

PR #8568, opened 2023-02-03
Terrain-Specific Natural Wonder Sprite Support

OptimizedForDensity:
After countless cumulative hours spent looking at a desert Barringer crater in the middle of the tundra I decided to implement support for tileset variants of natural wonders in the same way you can get tileset variants of base terrain + features + improvements combinations. Natural wonders currently replace whatever terrain they're placed on with a specified terrain (Barringer crater replacing the tundra or desert tile it spawns on with a mountain which also lets players build an observatory in a city next to it), so this PR make terrain replacement due to natural wonders an optional feature. This allows for a distinct "Tundra+Barringer Crater" variant in the tileset. The old natural wonder image naming scheme is still supported as a fallback so tilesets won't break.

To support this change, a `"Considered a [terrainFilter] tile"` unique was added that would let certain terrain features qualify as more than 1 type of terrain for purposes like adjacency so Barringer crater would still be able to fulfill the "next to a mountain" requirement for the observatory. It's also a nice QoL feature since this would appear in the Civilopedia and tell players that it's not a bug that they can build a mountain-only observatory next to a crater in a flat piece of tundra or desert. This PR only contains the code changes and none of the artwork/JSON modification to implement the sprites since I'm not the original artist plus we'd have to adjust natural wonder uniques (visibility modifiers, etc.) so there's no net gameplay change.

NaturalWonders

Note that this image shows the tundra providing a single food in the tile which could be removed with the `"Nullifies all other stats this tile provides"` unique while still keeping the natural wonder's bonuses but it also gives modders the option of having Grand Mesa on grasslands give more food due to the grasslands' inherent food production which adds some gameplay variety.


yairm210:
I'm not convinced we even need a "Considered a [tileFilter] tile", since we can just add that tileFilter to the uniques of the tile (e.g. add "Desert to a, I dunno, "Burnt Desert" terrain type)

OptimizedForDensity:
You mean something in the JSON file like this?
```
"uniques": ["Must be adjacent to [0] [Coast] tiles",
			"Must be adjacent to [0] [Grassland] tiles",
			"Must be adjacent to [0] to [2] [Mountain] tiles",
			"Must be adjacent to [0] to [4] [Elevated] tiles",
                        "[Mountain]"],
```

yairm210:
```
"uniques": ["Must be adjacent to [0] [Coast] tiles",
			"Must be adjacent to [0] [Grassland] tiles",
			"Must be adjacent to [0] to [2] [Mountain] tiles",
			"Must be adjacent to [0] to [4] [Elevated] tiles",
                        "Mountain"],
```

yairm210:
No square brackets

OptimizedForDensity:
Just tested without the bracket and it works without an additional unique since there's the `if (terrainUniqueMap.getUniques(filter).any()) return true` line in the filter function.

PR #8591, opened 2023-02-05
Suggestion: Collapsible tutorial tasks

Gualdimar:
in continuation #8559 and #7143
when you click on a tutorial task, it collapses to an icon and frees up screen space. Clicking on the icon again wiil return the tutorial.
<details><summary>Screens</summary>

</details>

### DIFF
--- a/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
+++ b/core/src/com/unciv/ui/worldscreen/WorldScreen.kt
@@ -543,21 +543,22 @@ class WorldScreen(
         val tutorialTask = getCurrentTutorialTask()
         if (tutorialTask == "" || !game.settings.showTutorials || viewingCiv.isDefeated()) {
             tutorialTaskTable.isVisible = false
+            return
+        }
+
+        tutorialTaskTable.isVisible = true
+        if (!UncivGame.Current.isTutorialTaskCollapsed) {
+            tutorialTaskTable.add(tutorialTask.toLabel()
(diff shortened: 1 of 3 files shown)


yairm210:
Looks good, but that is very small - is that even clickable on phones?

Gualdimar:
it's the same size as the green buttons near the minimap, tested on my 6" phone with screen size setting from medium to huge, no problems with hitting.

core/src/com/unciv/ui/worldscreen/WorldScreen.kt
```diff
@@ -548,6 +538,29 @@ class WorldScreen(
         }
     }
 
+    private fun displayTutorialTaskOnUpdate() {
+        tutorialTaskTable.clear()
+        val tutorialTask = getCurrentTutorialTask()
+        if (tutorialTask == "" || !game.settings.showTutorials || viewingCiv.isDefeated()) {
+            tutorialTaskTable.isVisible = false
```
yairm210, line 545: return here, then the rest of the function doesn't need to be in 'else'

yairm210: Previous function had only one level of nesting so this was less of a problem, but the new one has 2 levels. So we can flatten that out to just one.

Gualdimar: good point, fixed

PR #8592, opened 2023-02-05
Fixed resource modifier

### LINKED ISSUE #8585: "Quantity of strategic resources produced by the empire +[relativeAmount]%" not working for values less than 100
itch.io version 4.4.9, Windows 10

The `"Quantity of strategic resources produced by the empire +[relativeAmount]%"` unique doesn't produce anything for values under 100%. Try increasing resources by 25% and instead of turning 4 resources into 5, it stays at 4.

This little mod has a building that increases the strategic resources by 25% for each city. If three or fewer buildings are built, nothing happens, but once four buildings are built, it kicks in, doubling everything.
[StrategicNotIncreasing.zip](https://github.com/yairm210/Unciv/files/10610138/StrategicNotIncreasing.zip)

Gualdimar:
Closes #8585
now a non-integer modifier can be applied

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -167,7 +167,7 @@ class City : IsPartOfGameInfoSerialization {
 
         for (tileInfo in getTiles().filter { it.resource != null }) {
             val resource = tileInfo.tileResource
-            val amount = getTileResourceAmount(tileInfo) * civ.getResourceModifier(resource)
+            val amount = getTileResourceAmount(tileInfo)
             if (amount > 0) cityResources.add(resource, "Tiles", amount)
         }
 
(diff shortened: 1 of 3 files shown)


yairm210:
This is a good question - do percentage extra resources apply at the tile level? At the city level? At the civ level?
The higher up it applies, the more various resources can 'batch together' to give an extra resource

I think the city level, like you have here, is a good middle ground

Gualdimar:
since you mentioned it, unique says "produced by the empire", not "by the tile" or "by the city". 
in my test save civ has 18 Alu + 25% should be 22, but i got only 21
maybe we should refer to the civ5 rules?

yairm210:
In that case all we need is to do the float calculations in the civ-wide getResourceModifier

Gualdimar:
ok, i'll update

PR #8593, opened 2023-02-05
Fixed pillaged improvement defence bonus

Gualdimar:
fixed bug from discord


yairm210:
Wrong solution, since this leaves the other improvement abilities intact for pillaged improvements 
Correct solution is to take *unpillaged* improvement from tile, there's a convenience function for that, something like getUnpillagedImprovement()

Gualdimar:
in this fun only DefensiveBonus is applied

PR #8604, opened 2023-02-06
Experimental: mitigate texture swapping with TextureArraySpriteBatch

vegeta1k95:
This is experimental feature and should be more thoroughly tested for performance/memory consumption/bugs.

Very detailed discussion, along with benchmarks, examples and source of `TextureArraySpriteBatch` is located [here](https://github.com/libgdx/libgdx/issues/5907) and [here](https://github.com/libgdx/libgdx/pull/5914)

**TL;DR Issue:** Default `SpiteBatch` implementation can bind only one texture (atlas page, png) at a time, hence, number of draw flushes proportional to number of texture switches. Instead, new `TextureArraySpriteBatch` stores textures in LRU cache and utilizes improved GL20 shader to which just texture-to-draw index is passed, hence GREATLY reducing texture swapping time.

This is very handy when using mods: their sprites come from different texture (png mod file), so texture swapping may be huge if mod doesn't replace EVERY icon of the corresponding built-in atlas.

**TL;DR  Pros/Cons:**
Pros: texture draw pattern `ABABAB` has greatly reduced draw time: from 100-200ms to 5-7ms
Cons: texture draw pattern `AAAAAA` (no texture swapping) has slightly increased draw time: from 3-5ms to 5-6ms

Possible tweak: use `TextureArraySpriteBatch` only when there are permanent audio-visual mods?


yairm210:
Good summary! Going by pros and cons, I can see that the pros easily win. Half of gameplay on Unciv is with mods, and the game is explicitly geared towards that.
Not to mention that even within the base game, a lot of stuff is divided into different textures, so even for base game players this is probably a win.

yairm210:
I'm in favor of just introducing this and seeing if something breaks, honestly
But that comes with a host of other problems

The "correct" way to introduce this is to put it behind an optional flag in settings-graphics, called "experimental rendering improvements" or something (should be in template.properties as well). 
Then we'll notify in the version notes, add another note directly to people on Discord, and see what feedback we get

vegeta1k95:
@yairm210 
Done. Can rollout this one for test.

PR #8610, opened 2023-02-06
Fixed overview resources tab state check

Gualdimar:
Fixed bug from [discord](https://discord.com/channels/586194543280390151/1072188100018708490)


core/src/com/unciv/ui/overviewscreen/EmpireOverviewCategories.kt
```diff
@@ -47,7 +48,10 @@ enum class EmpireOverviewCategories(
     Resources("StatIcons/Happiness", 'R', Align.topLeft,
         fun (viewingPlayer: Civilization, overviewScreen: EmpireOverviewScreen, persistedData: EmpireOverviewTabPersistableData?)
                 = ResourcesOverviewTab(viewingPlayer, overviewScreen, persistedData),
-        fun (viewingPlayer: Civilization) = viewingPlayer.detailedCivResources.isEmpty().toState()),
+        fun (viewingPlayer: Civilization) = when {
+            viewingPlayer.detailedCivResources.any { it.resource.resourceType != ResourceType.Bonus } -> EmpireOverviewTabState.Normal
+            else -> EmpireOverviewTabState.Disabled
```
yairm210, line 53: The addition of .any{...} looks good, but the if/else should be done with .toState(), like before

Gualdimar: Fixed

PR #8613, opened 2023-02-07
Fixed the display of political relations 

### LINKED ISSUE #8182: Feature request: Politics relation 2-sided
4.3.10 apk, Android

It would be good if relation is shown on both sides in overview-Politics.
I didn't chekced other situation such as friendship or a war between 2 civs without me.
If it's 1-sided like now, player have to scroll up and down.
We have diagram but politics could be more polished.

one-sided

edit: imma just delete details tag :)

Gualdimar:
Closes #8182
Civs did not know about themselves
Now the political relationship is shown 2-sided and is shown when a known civ is in a relation with an unknown civ
<details><summary>screens</summary>

</details>

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -317,7 +317,7 @@ class Civilization : IsPartOfGameInfoSerialization {
 
     /** Returns only undefeated civs, aka the ones we care about */
     fun getKnownCivs() = diplomacy.values.map { it.otherCiv() }.filter { !it.isDefeated() }
-    fun knows(otherCivName: String) = diplomacy.containsKey(otherCivName)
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/logic/civilization/Civilization.kt
```diff
@@ -317,7 +317,7 @@ class Civilization : IsPartOfGameInfoSerialization {
 
     /** Returns only undefeated civs, aka the ones we care about */
     fun getKnownCivs() = diplomacy.values.map { it.otherCiv() }.filter { !it.isDefeated() }
-    fun knows(otherCivName: String) = diplomacy.containsKey(otherCivName)
+    fun knows(otherCivName: String) = (civName == otherCivName || diplomacy.containsKey(otherCivName))
```
yairm210, line 320: I'm not sure what the other implications of this change are, since this is a pretty central function, but the logic in this change is sound. I assume we'll discover the problems as they arise.

PR #8615, opened 2023-02-07
Order indicators in TechScreen like original Civ V

lishaoxia1985:
捕获
By the way, if you update Android Studio to lastest branch (Android Studio Electric Eel | 2022.1.1 Patch 1), com.android.tools.build:gradle:7.4.1 does very well.


yairm210:
See
https://github.com/vegeta1k95/Unciv/commit/d59bcb42c67689efe105b4a428246a32c2b3ec90#commitcomment-99535616

PR #8633, opened 2023-02-09
Improve UX

### LINKED ISSUE #8327: City Selection Building Queue selection and queue order is seemingly erratic to touch pad users
**Platform**
Android, Chromebook also Debian Linux (2 installation separate)

**Version**
Playstore/Flathub 4.3.16

**Describe the bug**
Selecting construction items in the build list should either be more clear: either have it by cliking the + (plus) icon button or a single click, one or the other must go. Also when moving items up or down queue list, it's far too easy to remove an item from the list unintentionally.

**To Reproduce**
Use a touch-pad to manage the building list in city screen, add items to queue, change item order in queue, note output of intended actions.

The PLUS (+) button seems more manageable on Touch-Pad -- but errant click will trigger the other method of input, the two methods walk over the other.

Gualdimar:
Added double click actions to:
Great person picker screen;
Improvement picker screen;
Policy picker screen;
Promotion picker screen;
Load/Save game screen.

Removed adding to/removing from construction queue on second click.
On phones it's quite annoying, and it's very easy to miss a button. Closes #8327

Removed default keyboard focus on load/save screen.
Because of keyboard focus on phones, the keyboard appears on the save screen and takes up half the space. In most cases, the standard name is enough or you need to copy the game to the clipboard.
<details><summary>Keyboard screenshot</summary>

</details>

### DIFF
--- a/core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
+++ b/core/src/com/unciv/ui/cityscreen/CityConstructionsTable.kt
@@ -370,7 +370,7 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
             addToQueueButton.onClick(UncivSound.Silent) {
                 // Since the pickConstructionButton.onClick adds the construction if it's selected,
                 // this effectively adds the construction even if it's unselected
-                addConstructionToQueue(construction, cityScreen.city.cityConstructions)
+                cityScreen.selectConstruction(construction)
             }
             pickConstructionButton.add(addToQueueButton)
         }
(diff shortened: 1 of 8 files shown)


vegeta1k95:
> Removed adding to/removing from construction queue on second click.
> On phones it's quite annoying, and it's very easy to miss a button. Closes #8327

Then we should probably not remove it, but add it as double click. 
Needing to stretch my finger to `+` button across the half of the screen everytime is much more annoying, then accidentally adding to queue sometime.

vegeta1k95:
This whole screen needs big refactoring. By just clicking construction button we update the whole CityScreen essentially, with tiles, stat table, etc. Which is a big waste of time.

So, for time being I suggest keeping "add-by-second-click" functionality. Can remove "removal-by-second-click", it is not used very often anyway.

Gualdimar:
makes sense, added back adding to construction queue on second click

yairm210 (approved):
I can't see anything wrong, but I will be listening for complaints on Discord. We have many players on many devices and what's more comfortable on one may be worse for another.
@vegeta1k95 awaiting your approval as well

vegeta1k95:
Seems fine for me too, can merge

Issue #8575, opened 2023-02-04
The displayed mod size is incorrect


Gualdimar:
The displayed mod size on the mod manager screen does not correspond to reality.
For example, 5Hex Tileset zipped - 6021 Kb, unzipped - 6052 Kb, Unciv shows - 96511 Kb

The first time I saw it, I wasn't sure my phone would handle a mod that was several times the size of the game itself)

vegeta1k95:
AFAIK it shows the size of github repo, because GitHub API returns it as such. Not just source code size.

Gualdimar:
then, imho, it's useless and only confusing for the average player

yairm210:
This looks like a serious problem with the Github API
The question is what do we do with this.

We have 2 options:
- Display incorrect size
- Don't display size at all

I'm honestly not sure which is worse, both are pretty bad options.

I'd like to hear if you think we have other options here.

Gualdimar:
if we want to show the size before downloading, the only thing that came to my mind is to get a list of files with sizes, then sum them all
`https://api.github.com/repos/ravignir/5Hex-Tileset/git/trees/master?recursive=1`

or show the size after downloading...

yairm210:
We can't have an API call for every mod in the list of mods, that's ridiculous
Maybe only when the user clicks on the mod do we try to do that
In any case size after downloading is not the problem we're trying to solve, we want the player to know what he's taking

Gualdimar:
Or the player clicks on a mod in the list, reads the description, decides whether he needs it,
clicks the download button, a confirmation window pops up with text "Do you want to download xxx KB?"
unciv makes an API request only after the download button is clicked

vegeta1k95:
Should be (partially) solved by #8648. Closing for now.

PR #8648, opened 2023-02-10
Mod sizes are updated to proper values upon selection

vegeta1k95:
Upon selection of a mod in `ModScreen` game will request Repo's content via [this API](https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree) (with `recursive` option), calculate total content size and update Repo's size accordingly.
(But before that, there is one more request to obtain SHA of the most recent commit).

If request fails for any reason, size remains unchanged - that of the whole repo.

Pros:
- Solves #8575

Cons: 
- GitHub request rate for unauthorized users is 60 requests/hour. Hence user can parse only ~30 mods sizes per hour properly.
- Requesting content is limited to 100 000 entries with maximum size 7MB for each.  Mod having >100 000 entries is very unprobable, but mods with some singular large files may have truncated size.
- On first selection user will see default size for a second (whole rep size), until precise value is requested, parsed and updated visually.

**Note:** I tried to fetch `.zip`'s size straight from the header of mod archive download URL, but GitHub's server does not provide `Content-Length` attribute in the HTTP responses, so unfortunately could not get size that way.

### DIFF
--- a/core/src/com/unciv/ui/pickerscreens/GitHub.kt
+++ b/core/src/com/unciv/ui/pickerscreens/GitHub.kt
@@ -250,49 +250,18 @@ object Github {
         return null
     }
 
-    class Commit {
-        var sha = ""
-    }
-
-    class Branch {
-        var commit = Commit()
-    }
-
-    class TreeFile {
-        var size: Long = 0L
-    }
-
     class Tree {
+
+        class TreeFile {
+            var size: Long = 0L
+        }
+
         var url: String = ""
         var tree = ArrayList<TreeFile>()
-        var truncated: Boolean = false
-    }
-
-    private fun getBranchTreeSha(link: String): String? {
-        var retries = 2
-        while (retries > 0) {
-            retries--
-            // obey rate limit
-            if (RateLimit.waitForLimit()) return null
-            // try download
-            val inputStream = download(link) {
-                if (it.responseCode == 403 || it.responseCode == 200 && retries == 1) {
-                    // Pass the response headers to the rate limit handler so it can process the rate limit headers
-                    RateLimit.notifyHttpResponse(it)
-                    retries++   // An extra retry so the 403 is ignored in the retry count
-                }
-            } ?: continue
-            return json().fromJson(Branch::class.java, inputStream.bufferedReader().readText()).commit.sha
-        }
-        return null
     }
 
     fun getRepoSize(repo: Repo): Float {
-
-        val sha = getBranchTreeSha("https://api.github.com/repos/${repo.full_name}/branches/${repo.default_branch}")
-            ?: return 0f
-
-        val link = "https://api.github.com/repos/${repo.full_name}/git/trees/$sha?recursive=true"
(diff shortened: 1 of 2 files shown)


Gualdimar:
why do we need a hash when we can use the default branch name? it will return the last commit
`
https://api.github.com/repos/${repo.full_name}/git/trees/${repo.default_branch}?recursive=true
`

vegeta1k95:
I followed their API documentation, which says explicitly SHA1 of a tree. But if that's the case with branch name, then even better, less requests needed.

vegeta1k95:
Also, we can put this whole logic behind download/update button only to minimize a number of requests. @yairm210 What do you think?

yairm210:
Makes sense to me that we store this data on the button so if we go back to it we won't need to recalculate the repo size

vegeta1k95:
@yairm210 It IS stored in the button. I update `Repo.size`

yairm210:
In that case I don't understand your question :/

vegeta1k95:
I am talking about [this](https://github.com/yairm210/Unciv/issues/8575#issuecomment-1419178599) proposition: show size only when user wants to download mods, in confirmation popup or something.

yairm210:
No, I think it's better to display size before, since that's when the user is arriving at the decision whether he wants it it not
As it's currently, if I understand correctly

core/src/com/unciv/ui/pickerscreens/GitHub.kt
```diff
@@ -250,6 +250,74 @@ object Github {
...
+                    retries++   // An extra retry so the 403 is ignored in the retry count
+                }
+            } ?: continue
+            val tree = json().fromJson(Tree::class.java, inputStream.bufferedReader().readText())
+
+            var totalSizeBytes = 0L
+            for (file in tree.tree)
+                totalSizeBytes += file.size
+
+            return totalSizeBytes * 0.000977f
```
yairm210, line 316: This is just unnecessarily confusing. Writing /1024 would be so much clearer.

core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt
```diff
@@ -417,6 +427,29 @@ class ModManagementScreen(
...
+                        launchOnGLThread {
+                            repo.size = repoSize.toInt()
+                            repo.hasUpdatedSize = true
+                            if (selectedMod == repo)
+                                updateModInfo()
+                        }
+                    }
+
+                } catch (e: IOException) {
+                    /* ... */
```
yairm210, line 447: ?

yairm210 (approved):
Looks good otherwise, you can merge after freaking with the nitpicks :)

core/src/com/unciv/ui/pickerscreens/ModManagementScreen.kt
vegeta1k95: If parsing fails for whatever reason, we just catch exception and do nothing

PR #8649, opened 2023-02-10
Improvement to MusicController

sockybob:
Currently there are only one music trigger (Every 10th turn) that plays music based on whether the player is at war with anybody, but it requires you to specify the civ name in prefix. This pr allows optional 'Peace' or 'War' prefix for 'Automatic next-track' music trigger. When player is at war it will choose from tracks without 'Peace' prefix and vice versa, allowing modders to add ambient music that should play only during the war/peace, as well as music that can play independent of that. Existing music mods should work as usual.

### DIFF
--- a/core/src/com/unciv/ui/audio/MusicController.kt
+++ b/core/src/com/unciv/ui/audio/MusicController.kt
@@ -213,8 +213,8 @@ class MusicController {
 					val worldScreen = UncivGame.Current.getWorldScreenIfActive()
                     if (worldScreen == null ||
                             !chooseTrack(
-                                MusicMood.peaceOrWar(!worldScreen.viewingCiv.isAtWar()), 
-                                MusicMood.Ambient, 
+                                prefix = MusicMood.peaceOrWar(!worldScreen.viewingCiv.isAtWar()), 
+                                suffix = MusicMood.Ambient, 
                                 EnumSet.of(MusicTrackChooserFlags.NoMatchingPrefix, MusicTrackChooserFlags.SuffixMustMatch)
                             )
                     ) {
--- a/core/src/com/unciv/ui/audio/MusicController.kt
+++ b/core/src/com/unciv/ui/audio/MusicController.kt
@@ -209,7 +209,17 @@ class MusicController {
             ControllerState.Silence ->
                 if (++ticksOfSilence > silenceLengthInTicks) {
                     ticksOfSilence = 0
-                    chooseTrack()
+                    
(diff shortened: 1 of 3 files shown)


sockybob:
Not sure if the way i implemented it is the best way. I tried to avoid changing/breaking the existing code.

core/src/com/unciv/ui/audio/MusicTrackChooserFlags.kt
```diff
@@ -7,6 +7,8 @@ enum class MusicTrackChooserFlags {
     PrefixMustMatch,
     /** Makes suffix parameter a mandatory match */
     SuffixMustMatch,
+    /** Signalizes that prefix parameter must not match */
+    PrefixMustNotMatch,
```
alexban011, line 11: I think `NoMatchingPrefix` would sound a bit better :)

alexban011:
> Not sure if the way i implemented it is the best way. 

Well if you want devs to review it you should open the PR :)

sockybob:
> Well if you want devs to review it you should open the PR :)

Thanks for the tip

yairm210:
I'll be honest, I don't understand this code very well
@SomeTroglodyte could I bother you with this, please?

sockybob:
Also if we use War/Peace in suffix wouldn't there be possible collisions with other music triggers? There can be civ specific War/Peace tracks so we will need to exclude those somehow?

sockybob:
Another idea is to use directories instead of suffixes and prefixes. Ambient music can go into folder named 'Ambient', ambient war  into 'Ambient\War', civ specific can go into folder with civ name etc.

core/src/com/unciv/ui/audio/MusicController.kt
```diff
@@ -209,7 +209,17 @@ class MusicController {
...
+					val worldScreen = UncivGame.Current.getWorldScreenIfActive()
+                    if (worldScreen == null ||
+                            !chooseTrack(
+                                prefix = MusicMood.peaceOrWar(!worldScreen.viewingCiv.isAtWar()), 
+                                suffix = MusicMood.Ambient, 
+                                EnumSet.of(MusicTrackChooserFlags.NoMatchingPrefix, MusicTrackChooserFlags.SuffixMustMatch)
+                            )
+                    ) {
+                        chooseTrack()
+                    }
```
yairm210, line 222: This is a strange way of doing this. I would assume that chooseTrack() internally would, if no matching tracks exist for the given filters, try and choose a track with no filters

PR #8654, opened 2023-02-11
Blockade mechanics

vegeta1k95:
This PR includes implementation for tile/city blockade mechanics from Civ 5: 

Blocked tile = tile which (this turn) does not produce yields, cannot be worked/locked.

**Mechanics:**
1) Enemy military land unit blocks tile it is on  (includes embarked);
2) Enemy military naval unit blocks tile it is on + adjacent water tiles which do not have friendly/neutral military units on them;
3) If coastal city has **all** adjacent water tiles blocked - city loses it's _harbor_ city connection to the capital;

This includes the following change: if tile was worked/locked and became blocked, city removes lock and reassigns population accordingly.

Looks as follows:
1) Enemy infantry block tiles they are standing on:
image

2) Enemy Destroyer blocks every adjacent tile, except the one where my Carrier is located:
image


AdityaMH:
It's time to see hidden submarine freely.

vegeta1k95:
> It's time to see hidden submarine freely.

That how it was in Civ 5, though it is debatable whether we should implement it the same.
Adding visibility check for units and a matter of one line of code, but it should be decided beforehand.

vegeta1k95:
Added City blockade mechanics

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -429,6 +429,40 @@ open class Tile : IsPartOfGameInfoSerialization {
         return civInfo.cities.firstOrNull { it.isWorked(this) }
     }
 
+    fun isBlockaded(): Boolean {
+        val owner = getOwner() ?: return false
+        val unit = militaryUnit
+
+        // If tile has unit
+        if (unit != null) {
+            return when {
```
yairm210, line 438: Can if/else

core/src/com/unciv/ui/utils/UnitGroup.kt
```diff
@@ -125,7 +126,9 @@ class UnitGroup(val unit: MapUnit, val size: Float): Group() {
         }
 
         if (unit.health < 100) { // add health bar
-            addActor(ImageGetter.getHealthBar(unit.health.toFloat(), 100f, size))
+            val hp = ImageGetter.getHealthBar(unit.health.toFloat(), 100f, size*0.78f)
```
yairm210, line 129: What's the magic 0.78?

yairm210:
Looks good, but since this is an entirely be mechanic it needs a tutorial, which should pop up the first time one of your tiles is blockaded

core/src/com/unciv/ui/utils/UnitGroup.kt
vegeta1k95: This is similar multiplier to FlagBackground `innerMultiplier`. Only 0.1 smaller, so that HP bar width is a bit smaller than flag's inner circle

core/src/com/unciv/logic/map/tile/Tile.kt
vegeta1k95: Don't understand which `if/else` you are talking about here.

yairm210: My mistake, I thought the when had only 2 possible outcomes, not 3

vegeta1k95:
@yairm210 
Added tutorials: 
1) For tile blockade, on CityScreen
2) For city blockade, on WorldScreen

Added city blockade status icon for city plates:
image

Caballero-Arepa:
Amazing!
Cant wait to declare war on my firends xD

> For instance, Civ 5 Vox Populi/Community Patch allows blockade only by visible units (so subs do not blockade).

I dont think so, Germany blockaded Britain in both World Wars with Submarines, and Rusia is doing it in Ukraine, so it is realistic.

PR #8661, opened 2023-02-11
Limit camera movement within explored region

Gualdimar:
A rectangular area is calculated from the two extreme explored points on the map, beyond which the camera cannot go.
The area is recalculated after each new tile is explored and stored for each civilization separately, so they all have their own area.
Since the hex coordinate system is offset by an angle relative to the world one, longitude and latitude were used instead of hex coordinates to calculate the extreme points. The longitude and latitude are then converted to the world coords. This allowed the use of two points on the map, instead of four when using hex coordinates.

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -192,6 +192,9 @@ class Civilization : IsPartOfGameInfoSerialization {
     var citiesCreated = 0
     var exploredTiles = HashSet<Vector2>()
 
+    // Limit camera within explored region
+    var exploredRegion = ExploredRegion()
+
     fun hasExplored(tile: Tile) = tile.isExplored(this)
 
     var lastSeenImprovement = HashMapVector2<String>()
@@ -275,6 +278,7 @@ class Civilization : IsPartOfGameInfoSerialization {
         // Cloning it by-pointer is a horrific move, since the serialization would go over it ANYWAY and still lead to concurrency problems.
         // Cloning it by iterating on the tilemap values may seem ridiculous, but it's a perfectly thread-safe way to go about it, unlike the other solutions.
         toReturn.exploredTiles.addAll(gameInfo.tileMap.values.asSequence().map { it.position }.filter { it in exploredTiles })
+        toReturn.exploredRegion = exploredRegion.clone()
         toReturn.lastSeenImprovement.putAll(lastSeenImprovement)
         toReturn.notifications.addAll(notifications)
         toReturn.notificationsLog.addAll(notificationsLog)
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -247,7 +247,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
      *
      * Sets the returned `WorldScreen` as the only active screen.
      */
-    suspend fun loadGame(newGameInfo: GameInfo): WorldScreen = withThreadPoolContext toplevel@{
+    suspend fun loadGame(newGameInfo: GameInfo, callFromLoadScreen: Boolean = false): WorldScreen = withThreadPoolContext toplevel@{
         val prevGameInfo = gameInfo
         gameInfo = newGameInfo
 
@@ -261,7 +261,7 @@ class UncivGame(parameters: UncivGameParameters) : Game() {
         initializeResources(prevGameInfo, newGameInfo)
 
         val isLoadingSameGame = worldScreen != null && prevGameInfo != null && prevGameInfo.gameId == newGameInfo.gameId
-        val worldScreenRestoreState = if (isLoadingSameGame) worldScreen!!.getRestoreState() else null
+        val worldScreenRestoreState = if (!callFromLoadScreen && isLoadingSameGame) worldScreen!!.getRestoreState() else null
 
         lateinit var loadingScreen: LoadingScreen
 
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -542,6 +542,8 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion 
         }) {
             for (unit in civInfo.units.getCivUnits())
                 unit.updateVisibleTiles(false) // this needs to be done after all the units are assigned to their civs and all other transients are set
+            if(civInfo.playerType == PlayerType.Human)
+                civInfo.exploredRegion.setMapParameters(tileMap.mapParameters.worldWrap, tileMap.mapParameters.mapSize.radius) // Required for the correct calculation of the explored region on world wrap maps
             civInfo.cache.updateSightAndResources() // only run ONCE and not for each unit - this is a huge performance saver!
 
             // Since this depends on the cities of ALL civilizations,
--- a/core/src/com/unciv/logic/civilization/ExploredRegion.kt
+++ b/core/src/com/unciv/logic/civilization/ExploredRegion.kt
@@ -1,0 +1,165 @@
+package com.unciv.logic.civilization
+
+import com.badlogic.gdx.math.Vector2
+import com.unciv.logic.IsPartOfGameInfoSerialization
+import com.unciv.logic.map.HexMath.getLatitude
+import com.unciv.logic.map.HexMath.getLongitude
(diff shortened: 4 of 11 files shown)


vegeta1k95:
This breaks world-wrap continious scrolling.
Create world-wrap world with whole map explored (start at Future Era): scrolling is stuck at map boundaries.

core/src/com/unciv/logic/civilization/Civilization.kt
```diff
@@ -191,6 +196,7 @@ class Civilization : IsPartOfGameInfoSerialization {
     var cities = listOf<City>()
     var citiesCreated = 0
     var exploredTiles = HashSet<Vector2>()
+    var exploredRegion = HashMap<String,Vector2>()
```
vegeta1k95, line 199: Since you made this a non-transient serializable field, there should be a correspondent line in the `Civilization.clone()`

core/src/com/unciv/ui/utils/ZoomableScrollPane.kt
```diff
@@ -71,13 +71,35 @@ open class ZoomableScrollPane(
...
+            val max = centerY-exploredRegion["yWorldVector"]!!.y
+
+            if (result < min) result = min
+            else if (result > max) result = max
+        }
+
+        super.scrollY(result)
         updateCulling()
         onViewportChanged()
     }
```
vegeta1k95, line 105: I think this doesn't belong to `ZoomableScrollPane`, which is an abstract scroll pane object which in principle can contain everything, not only tile map. 

Moreover: CityScreen and MapEditorScreen which contain such scroll pane do not need camera limiting functionality: city screen has map size of like 8 tiles or so and map editor screen is not supposed to restrict vision of a map creator.

I'd propose to move this to `WorldMapHolder`, which is subclass of `ZoomableScrollPane` and already contains a reference to the current `WorldScreen`.

Gualdimar:
> This breaks world-wrap continious scrolling.

hm, world-wrap... i forgot about it. need to test.

Gualdimar:
@vegeta1k95 all your points are valid and can be easily resolved but man, the world wrap... the world wrap just makes things so complicated. I more or less got it to work, but the logic is so confusing, and the number of if statements is so large :exploding_head:
unresolved issues:
1. when 1-2 columns of FoW tiles are left unexplored, if you swing hard, you can jump over the limit and get to the other side
2. when exploring the last 2-4 columns of FoW tiles, it is difficult to determine from which side of the explored area to add them
3. tiles rendered before the edges of the explored area are connected. therefore, so you can see that a few FoW tiles away is a map explored for the other side

Gualdimar:
should now work correctly with world-wrap. better coordinate checking(there should be no overjumps).
new explored tiles are now added from the side closest to the explorer unit or just to the nearest side, if there is no explorer.

the last thing left is tile render over FoW

core/src/com/unciv/logic/civilization/Civilization.kt
```diff
@@ -191,6 +197,7 @@ class Civilization : IsPartOfGameInfoSerialization {
...
```
yairm210, line 200: Since there are only 2 vectors , this should not be a dictionary, treat them by name.
Also, call them 'topright' and 'bottomleft', 'start' and 'end' is very ambiguous

yairm210: Also, *this can be calculated from tilemap data*, so we don't need to persist this at all! All these can be transient!
(Since there are several vectors you can store them all in an object, like 'class exploredRegion')

core/src/com/unciv/ui/utils/ZoomableScrollPane.kt
```diff
@@ -366,4 +366,7 @@ open class ZoomableScrollPane(
     private fun onViewportChanged() {
         onViewportChangedListener?.invoke(maxX, maxY, getViewport())
     }
+
+    open fun checkX(deltaX: Float): Float = scrollX - deltaX
```
yairm210, line 370: These are not good names. What is the intent for these functions?

Gualdimar: These functions check if the coordinates are within the explored region

core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
```diff
@@ -754,4 +754,37 @@ class WorldMapHolder(
...
+            else if ((startX < endX && result < startX) || (endX < startX && result < startX && result > endX) && deltaX > 0)
+                result = startX
+        }
+        return result
+    }
+
+    override fun checkY(deltaY: Float): Float {
+        val exploredRegion = worldScreen.viewingCiv.exploredRegion
+        var result = scrollY + deltaY
+        if(!exploredRegion.isEmpty()) {
```
yairm210, line 780: Reverse this if. Also, under what circumstances is the exploredregion empty? How can that be?

core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
```diff
@@ -754,4 +754,37 @@ class WorldMapHolder(
...
+            val mapCenterY = maxY * 0.5f
+            val startY = mapCenterY-exploredRegion["yWorldVector"]!!.x
```
yairm210, line 782: This is another confusing thing. You're using x,y in the vector for usages that are not, in fact, x and y.
This would be better served by topright/bottomleft vectors for world, OR just use maxWorldY, minWorldY etc in the exploredRegion object

core/src/com/unciv/ui/worldscreen/WorldMapHolder.kt
Gualdimar: If acts as a safeguard if something goes wrong so that the camera doesn't get stuck at the wrong point. I'm not sure I can think through all the possible scenarios.

core/src/com/unciv/logic/civilization/Civilization.kt
```diff
@@ -802,6 +810,80 @@ class Civilization : IsPartOfGameInfoSerialization {
     fun setAllyCiv(newAllyName: String?) { allyCivName = newAllyName }
 
     fun asPreview() = CivilizationInfoPreview(this)
+
+    fun calculateExploredRegion(position: Vector2, explorerPosition: Vector2?) {
```
yairm210, line 814: I hope after the existing comments I gave this will be more readable, but I still think this should definitely have more comments. You seem to be dealing with edge cases for world wrap that are unclear when reading the code, so just saying in lines like 'we passed over the edge of the world' could make things much clearer

yairm210: (also why do you need the explorer position?)

core/src/com/unciv/ui/utils/ZoomableScrollPane.kt
yairm210: So maybe getNewX? If a function is named "check" I expect to receive the result of a check, not the final information

core/src/com/unciv/logic/civilization/Civilization.kt
Gualdimar: ok, I will definitely add comments. I was wondering what you would say about this new feature in general, whether it deserves the right to life)

when exploring the last 2-4 columns of FoW tiles, it is difficult to determine from which side of the explored area to add them, because the new cell can be right in the middle between them. so I use the position of the explorer to determine from which side to expand the region

core/src/com/unciv/logic/civilization/Civilization.kt
Gualdimar: > Also, _this can be calculated from tilemap data_, so we don't need to persist this at all! All these can be transient!

tested it, tiles become explored in a random order, sometimes messing up the explored region.

Gualdimar:
Updated:
- Loading the game from the loading screen resets the current scroll position and zoom. Why: When loading a previously saved game, the camera may have been outside the explored region.
- Moved all logic to a separate class ExploredRegion
- Reduced the number of non-transient properties
- Added comments

Still have a problem with the visibility of the other edge of the map through the fog. I haven't come up with a good solution.

yairm210:
I think that's fine, this solves 95% of the problem :)

yairm210:
Now all we need is 'minimap displays only explored area' to have truly location-agnistic non-world-wrap games :) (world wrap ones would also need to account for "non-centered map when crossing edge")

yairm210 (approved):

PR #8674, opened 2023-02-13
Remove dimming of some icons on unit selection

vegeta1k95:
I think iteration over all tile groups (which may as well be 10 000 tiles) for each unit selection just to dim a few icons is quite unnecessary.


yairm210:
Is this a UI or performance issue? I think the current situation is actually very comfortable as a player, and I'm willing to pay that extra millisecond to have to process less information
What we could do is save whether the last unit selected was military or civilian because usually we do civilians or military in a row, so we can save performance there

vegeta1k95:
I have another idea how to do this. I'll close this for now.

PR #8679, opened 2023-02-14
Desktop: world camera autoscroll, selectable window mode

vegeta1k95:
This contains two improvements for desktop version:

1) When moving mouse to screen boundary, camera will scroll automatically in that direction
2) Option under `Display` tab, which allows to choose screen mode: windowed (default) and fullscreen. Option is visible only on desktop.


Gualdimar:
I think that improvement #1 should also have a switch in the options

vegeta1k95:
Make an autoscroll a Desktop-only option, default off.
image

PR #8684, opened 2023-02-16
Removed zoom limit for world-wrap maps

Gualdimar:
You can now zoom out on world-wrap maps without flickering and with working continuous scrolling

### DIFF
--- a/core/src/com/unciv/ui/components/ZoomableScrollPane.kt
+++ b/core/src/com/unciv/ui/components/ZoomableScrollPane.kt
@@ -45,7 +45,7 @@ open class ZoomableScrollPane(
         this.addListener(zoomListener)
     }
 
-    fun reloadMaxZoom() {
+    open fun reloadMaxZoom() {
 
         maxZoom = UncivGame.Current.settings.maxWorldZoomOut
         minZoom = 1f / maxZoom
--- a/core/src/com/unciv/ui/components/ZoomableScrollPane.kt
+++ b/core/src/com/unciv/ui/components/ZoomableScrollPane.kt
@@ -90,17 +90,6 @@ open class ZoomableScrollPane(
         updatePadding()
         super.sizeChanged()
         updateCulling()
-
-        if (continuousScrollingX) {
-            // For world-wrap we do not allow viewport to become bigger than the map size,
-            // because we don't want to render the same tiles multiple times (they will be
-            // flickering because of movement).
-            // Hence we limit minimal possible zoom to content width + some extra offset.
-            val content = actor
-            if (content != null)
-                minZoom = max((width + 80f) * scaleX / content.width, 1f / UncivGame.Current.settings.maxWorldZoomOut)// add some extra padding offset
-        }
-
     }
 
     private fun updatePadding() {
--- a/core/src/com/unciv/ui/components/tilegroups/TileGroupMap.kt
+++ b/core/src/com/unciv/ui/components/tilegroups/TileGroupMap.kt
@@ -163,9 +163,14 @@ class TileGroupMap<T: TileGroup>(
 
         if (worldWrap) {
 
+            // Prevent flickering when zoomed out so you can see entire map
+            val visibleMapWidth: Float
+            if(mapHolder.width > width) visibleMapWidth = width - groupSize * 1.5f
+            else visibleMapWidth = mapHolder.width
+
             // Where is viewport's boundaries
-            val rightSide = mapHolder.scrollX + mapHolder.width/2f
-            val leftSide = mapHolder.scrollX - mapHolder.width/2f
+            val rightSide = mapHolder.scrollX + visibleMapWidth/2f
+            val leftSide = mapHolder.scrollX - visibleMapWidth/2f
 
             // Have we looked beyond map?
             val diffRight = rightSide - topX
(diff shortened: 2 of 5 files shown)


yairm210:
This looks like a good look to me
I will wait for @vegeta1k95 's thoughts

core/src/com/unciv/ui/tilegroups/TileGroupMap.kt
```diff
@@ -162,9 +162,14 @@ class TileGroupMap<T: TileGroup>(
 
         if (worldWrap) {
 
+            // Prevent flickering when zoomed out
+            val mapWidth: Float
+            if(mapHolder.width > width) mapWidth = width - groupSize * 1.5f
```
yairm210, line 167: How could it not be? What, if the entire map is only 1 tile wide? 😕

Gualdimar: width - TileGroupMap.width or map width, always the same after map generation
mapHolder.width - the width of the visible area in the mapholder, decreases when zooming in, increases when zooming out

Gualdimar: maybe the val name should be changed to something like visibleMapWidth?

yairm210: That would be clearer, yes

yairm210: Because I can totally get how the *visible width* can be bigger than the *total width*
Also maybe add a comment like 'zoomed out so you can see entire map' for clarity

vegeta1k95:
It works good, no flickering. 

But I am not really good with that on maps of size Medium and less we see empty parts on the sides when zooming out **without** advanced zoom option.

Default max zoomout 2x on Small:
image

Sure, tiles are moving properly, but it loses the feeling of "seemless continious planet-like" world.

Maybe we remove zoom limit only when advance zoom used?

Gualdimar:
> Sure, tiles are moving properly, but it loses the feeling of "seamless continious planet-like" world.

when we zoom out on google maps, we see the globe and space on the sides :smile:

vegeta1k95:
> when we zoom out on google maps, we see the globe and space on the sides
Things on a globe don't instanteneous teleport to the other side when rotating, that is the point.

vegeta1k95:
@yairm210 
My proposition would be to keep zoom lock, unless max zoom option is anything different than default.
That way people who want to see more, can do so by adding more zoom.
And people who don't want to see empty space around and enjoy "continious" map scrolling are also satisfied.

What do you think?

Gualdimar:
there were complaints after limiting the zoom. and now, if the player does not like it, he can limit the zoom in the settings. 
for me, seeing the entire map is more important than the feeling of continuous scrolling and world-wrap is more of a convenience feature to reduce scrolling time.

> My proposition would be to keep zoom lock, unless max zoom option is anything different than default.
maybe simply reduce the smallest possible zoom setting from 2 to 1?

vegeta1k95:
> there were complaints after limiting the zoom.
Honestly, I've seen only one on Discord, and it was related to MapEditor screen. Which of course should have no zoom restrictions for map maker's convenience.

> he can limit the zoom in the settings. 
That's the point. You cannot limit zoom below default 2x.
The other point is that zoom magnification is uneven on different map sizes. On Large and Huge 2x zoomout viewport is less than map size. On Medium and less, 2x zoomout viewport is bigger than mapsize,

core/src/com/unciv/ui/tilegroups/TileGroupMap.kt
Gualdimar: Updated

Gualdimar:
Anyway, I`m good, as long as there is an option to zoom out to the full map view

yairm210:
I think for now let's remove the lock on the map editor, since that's something we all agree on
As for the world map, sounds like we need a poll

Gualdimar:
> As for the world map, sounds like we need a poll

Fair enough.

@vegeta1k95 is there any reason why you did the zoom limit check in _ZoomableScrollPane.sizeChanged()_ and not in _WorldScreen.init_?

vegeta1k95:
> is there any reason why you did the zoom limit check in ZoomableScrollPane.sizeChanged() and not in WorldScreen.init?

Zoom limit depends on scrollpane width, if user manually resizes window, limit has to be recalculated.

Gualdimar:
played around with this a bit:
1. after changing the size of the window, the world screen is reinitialized - so this is not a problem
2. reloadMaxZoom() is called only if you open the options screen with a keyboard(ctrl+o) and is not called if you open it through the menu with the mouse. - i think we should reload max zoom in both cases, but it is still possible to make the value calculated once, and not every time after rescaling

vegeta1k95:
If you fix zoom recalculation in options, then yeah, zoom limit can be calculated only once for each worldscreen init. But beware, that it has to be done after scrollpane actor is set

Gualdimar:
Returned zoom limit for the world screen, it is now calculated once on _init_ after _mapHolder.addTiles()_ or after changing options.
Also fixed padding for zoom limiting, if the screen becomes too wide 80 was not enough. Now it is approx 1 row less than the map size.

@yairm210 maybe just add the "Enable zoom limit for word-wrap maps" checkbox to the advanced options tab?

Gualdimar:
> maybe just add the "Enable zoom limit for word-wrap maps" checkbox

or "Limit max zoom to map width" - not only for world-wrap, so you don't need to update the max zoom setting for different map sizes?

yairm210:
Sorry for the conflicts, I just did a big re-org to the file structure of all ui classes
It's been a long time coming and we had to do it sometime, and there's always *something* UI-related under work

All the old classes still exist with their old names, only the location of the files has changed

Gualdimar:
@yairm210 so what about the checkbox in options for enabling/disabling zoom limit?

PR #8701, opened 2023-02-18
Update and rename android/assets/jsons/translations/Simplified_Chines…

Jack1588:
…e.properties to 安卓/资产/jsons/翻译/Simplified_Chinese.properties

stone石头


yairm210:
These files need to remain with their current names or else the game won't be able to find and use them

PR #8702, opened 2023-02-19
Option to disable max zoom limit

Gualdimar:
Added an option on the Advanced tab that allows to disable or enable(default enabled) zoom restrictions by map width.
Also fixed the comment, as there is no more flickerng
image

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -717,7 +717,6 @@ Autoupdate mod uniques =
 Uniques updated! = 
 
 Max zoom out = 
-Limit max zoom out by map width = 
 HIGHLY EXPERIMENTAL - YOU HAVE BEEN WARNED! = 
 Enable portrait orientation = 
 Generate translation files = 
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -717,6 +717,7 @@ Autoupdate mod uniques =
 Uniques updated! = 
 
 Max zoom out = 
+Limit max zoom out by map width = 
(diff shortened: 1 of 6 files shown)


yairm210 (approved):

Gualdimar:
Reverted PR #8608, as it is no longer needed

yairm210:
@SomeTroglodyte I reread you comment 5 times and still don't know what you're referring to

yairm210:
Looking at this again, I don't think that we need both a slider and a checkbox. Can we use the rightmost position of the slider for 'no limit'?

Gualdimar:
Hmm, great idea! I've tested it and tweaked a bit, so for medium and smaller map sizes, the zoom value of 2 is already larger than the map width. And for maps larger than medium, the value of 3 is larger than the map width. 
Therefore, I replaced the checkbox with a check: 
```
val enableZoomLimit = (mapRadius < 21 && maxWorldZoomOut < 3f) || (mapRadius > 20 && maxWorldZoomOut < 4f)
```

yairm210:
Interesting, @vegeta1k95 I assume this is because when worldscreen is generated it at first has all the the tilegroups in their "centered on 0,0" positions, and only when moving the map do they get repositioned

vegeta1k95:
> Interesting, @vegeta1k95 I assume this is because when worldscreen is generated it at first has all the the tilegroups in their "centered on 0,0" positions, and only when moving the map do they get repositioned

Repositioning itself happens on each `draw` call, so should be done without skipping. What probably happens is right after recreation some parameters (like `scrollX`) are not properly initialized yet, so reposition check fails.
If one zooms or panes - parameters update and reposition happens.

I'll check it.

PR #8707, opened 2023-02-20
UI Helper: GUI class

vegeta1k95:
This introduces static UI helper class `GUI` which abstracts (most of) UI methods of `UncivGame`.

This particular PR incapsulates `UncivGame.worldScreen` field, so that it is accessed only via `GUI` interface.

Pros: 
- Unified access to common UI methods through the same interface
- Code is better readable and shorter


yairm210:
I'm okay with this, but we should be minimizing usages of static globals as much as possible, I see this class as a starting point to - as it were - get rid of all the functions it contains :)

PR #8711, opened 2023-02-20
Slider tooltip permanent by default

Gualdimar:
In most cases (actually, except for one: "Brush ([size]):" in the map editor), the value of the slider is not obvious and you need to change it to display a tooltip with the current value.
That's why I made the tooltip permanent by default, but you can change it in the slider's constructor parameters.


yairm210:
Can you show an example of before/after?

Gualdimar:
Before:
image
image

After:
image
image

PR #8715, opened 2023-02-21
Modding: allow mods to supply custom fonts

vegeta1k95:
This PR allows mods to supply custom fonts: `.ttf` files placed inside of `$modFolder/fonts/`.
Fonts will be visible and selectable in the font-list in `Advanced` options (and marked with mod name):

image


android/src/com/unciv/app/FontAndroid.kt
```diff
@@ -0,0 +1,163 @@
...
+                                Typeface.create(fontName, Typeface.NORMAL)
+                            }
+                        }
+                        else
+                        {
+                            Typeface.create(fontName, Typeface.NORMAL)
+                        }
+            }
+
+        }
```
yairm210, line 77: Far too much nesting!
- Reverse initial 'if' to 'if == then return'
- return at the end of 'if' of mod font
- reverse 'if' that includes the build-version

Alternatively, separate the system font into its own function, like `createSystemFont(fontName:String)`

yairm210 (approved):
Looks good it general!

android/src/com/unciv/app/FontAndroid.kt
vegeta1k95: Changed

yairm210:
I'm adding this in since I'm releasing a version, but I do want you to add documentation to the docs - currently this is a 'hidden feature' that modders won't know how to activate

PR #8716, opened 2023-02-21
Added password authentication as server feature

### LINKED ISSUE #8482: Feature request: Password validation in Unciv
Can Unciv send some kind of password Header so that I can do validation for it.

And Users can set passwords per hostname (like uncivserver.xyz). Cause it will be bad if the password of one server gets sent to another one. and people log that.

Suggest thread in Discord: https://canary.discord.com/channels/586194543280390151/1068889614900809768/1068889614900809768

GGGuenni:
Problems not yet solved:
- [x] Add a way to close the auth popup - but where should the button go (Or implement dismissable popups via click on background?)
- [ ] UncivServer does not validate the game access correctly as it doesn't decode the save file
- [ ] (Already existing bug) Playing a multiplayer game alone results in "Waiting for [own civ]"

I've implemented it using ServerFeatures to distinguish servers with specific features easily 

@touhidurrr I've implemented an "/auth" endpoint that is called to authenticate the password (get) and can be used to set a new one (put). All endpoints are called with the new header "Authorization": "Basic [userId base64]:[password base64]" even though some like get("/files/{fileName}") probably don't need it
see 5933ae044c22269b91f4b0706b028058adb5f84a for an example

Resolves #8482


touhidurrr:
So, I guess auth is a first time thing? Consecutive calls of auth will need previous password?

touhidurrr:
I guess the flow will be something like this then:
`userId` in **UncivGames Democracy Bot** => Use Democracy Bot to Manage Password.
`userId` not in **UncivGames Democracy Bot** => Use Unciv Client to Manage Password.

Maybe display errors from the servers too when `auth` is called?

Also thanks for your work!!!

touhidurrr:
Also password should be hashed.

GGGuenni:
> So, I guess /auth is a first time thing? Consecutive calls of auth will need previous password?

As long as no password has been set via `put /auth`, the `get /auth` endpoint will always return with code 200. After a password has been set, all endpoints need to be given the correct `Authorization` header.

> Maybe display errors from the servers too when /auth is called?

Which errors are you thinking of? Currently a failed authentication, no matter if timeout or wrong password, will just return `Authentication failed`

> I guess the flow will be something like this then:

I don't really know what the "UncivGames Democracy Bot" is but if the password is set via an external API call the user will be asked to input the new password as soon as a file upload from unciv is issued. The password will then be saved in the gameSettings so it can be used for any further API calls

GGGuenni:
To let the unciv client know that uncivserver.xyz uses authentication, the /isalive endpoint must return `{authVersion: 1}` instead of its current `true`

Also some images.

Authentication after server status check if a password is already set
unciv_auth_popup_settings

Authentication on nextTurn if the user did not authenticate via status check on the settings page
unciv_auth_popup_worldscreen

This is what the multiplayer settings look like now. Should we also display if a password for the current server is saved in the client? And in addition, ask the user if they want to add a password if none has been set yet when checking the server status?
unciv_auth_settings

touhidurrr:
> To let the unciv client know that uncivserver.xyz uses authentication, the /isalive endpoint must return {authVersion: 1} instead of its current true

Will older version fail for this?
I understand the other requirements and they looks good.

But not hashing mentioned. Also, thanks for the pr again!!

GGGuenni:
> Will older version fail for this?

Any version above 0 will work but I've added it as an integer in case other authentication methods (like Bearer tokens) are implemented in the future. So `{authVersion: 1}` should be used.
If the server returns either `{authVersion: 0}` or any unparsable string (like the current `true`) the server is assumed to not support authentication

> But not hashing mentioned.

I don't really see a real benefit in hashing the password but I will look into it

GGGuenni:
Added an indicator if a password is set for the current server

unciv-not-secured
unciv-already-secured

yairm210:
> ask the user if they want to add a password if none has been set yet when checking the server status?

As a one-time thing, maybe
But that should definitely wait until we've tried this out and ironed out all the problems

yairm210:
- I don't see a reason to hash the password, since we won't be saving it inside the saved game, but only in the server, and connection should be over https anyway
- I think we *should* display the current password to the user - because if I'm moving between phones and I set my password a year ago, I probably don't remember it. And who would care enough to take it from my phone?

GGGuenni:
> I think we should display the current password to the user

I will just set it as the textfield hint if a password is set

> I don't see a reason to hash the password

Only thing left would be the changes to the UncivServer to check the current civ but since uncivserver.xyz runs on a different implementation anyway I wouldn't say it's a must have.
Or should it be included to have a fully functional server ready for use?

GGGuenni:
New hint if password is set
Does not update as soon as password is set. Needs to reopen the settings but I don't think that's a problem

image

yairm210:
Merging this
Expect there to be knockbacks though :)

touhidurrr:
> Merging this
Expect there to be knockbacks though :)

Oh sorry, I was a little busy for work. Since tomorrow is Sunday, I will try to implement this by tomorrow.

oynqr:
I will also implement this for rust_unciv_server, but using secure password hashing, most likely Argon2id. I urge everyone to do the same. Take a look at https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html for some decent recommendations.

The current server-side implementation uses simple string comparison, which could be vulnerable to timing attacks. Combining compression and secrets also makes me a little uneasy.

touhidurrr:
> I will also implement this for rust_unciv_server, but using secure password hashing, most likely Argon2id. I urge everyone to do the same. Take a look at https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html for some decent recommendations.
> 
> The current server-side implementation uses simple string comparison, which could be vulnerable to timing attacks. Combining compression and secrets also makes me a little uneasy.

Good suggestion!! I guess it is better for the hashing to happen of server-side than client-side. I will also implement argon2 then probably.

touhidurrr:
@GGGuenni I am little confused. If the server supports auth, is setting a password required?

GGGuenni:
> If the server supports auth, is setting a password required?

No a password is not required
However, the UncivServer.kt requires an authentication header
So if no password is set the client will send `Basic userId:`


But we could still change it, so that the client does not send an Auth header if no password is set without introducing a second Auth version. At least as long as not too many servers implemented Auth.

Also I did realise yesterday that I've implemented the Basic Auth header a bit incorrectly.
Instead of base64 encoding password and userId separately the `userId:password` string is supposed to be encoded as a whole

touhidurrr:
> Instead of base64 encoding password and userId separately (that's how it's implemented currently) the `userId:password` string is supposed to be encoded as a whole

yeah I noticed that also. can you fix that as a pr before i implement this?

Also I was wondering why do we need `Basic` in the first place. Since we need to parse the game file and get the current players player id anyways, can't it be simply `Bearer` auth with the Base64 encoded password? This would make things simplier.

GGGuenni:
> yeah I noticed that also. can you fix that as a pr before i implement this?

I will make one tomorrow

> Also I was wondering why do we need `Basic` in the first place. Since we need to parse the game file and get the current players player id anyways, can't it be simply `Bearer` auth with the Base64 encoded password? This would make things simpler.

We need the uid for setting the password and I think it's more future proof this way.

> Also maybe check if the password is at least 6 chars long also?

Good Idea

touhidurrr:
Meanwhile my sloppy implementation:
```ts
import * as argon2 from 'argon2';
import fp from 'fastify-plugin';
import { UncivJSON } from '../types/UncivJSON';

// declare types
declare module 'fastify' {
  interface FastifyRequest {
    game?: UncivJSON;
    userId?: string;
  }
}

const authPathRegex = /^\/(auth|files)/;

export default fp(async function (server) {
  server.addHook('preValidation', async function (req, reply) {
    // return on /isalive
    if (authPathRegex.test(req.url)) return;

    let hash = await server.cache.auth.get(req.userId!);
    if (!hash) {
      const dbHash = await server.db.Auth.findOne({ _id: req.userId! });
      if (dbHash) hash = dbHash.hash;
    }

    // if no hash is found, validation done
    if (!hash) return;

    if (req.headers.authorization) {
      const [type, auth] = req.headers.authorization.split(' ');
      if (type === 'Basic') {
        const userPass = Buffer.from(auth, 'base64').toString('utf8');
        const colonIdx = userPass.indexOf(':');
        const userId = userPass.slice(0, colonIdx);
        const password = userPass.slice(colonIdx + 1);

        if (!userId) return reply.code(400).send('400 Bad Request!');

        if (await argon2.verify(hash, password)) {
          req.userId = userId;
          return;
        } else {
          return reply.code(401).send('401 Unauthorized!');
        }
      }
    }

    return reply.code(401).send('401 Unauthorized!');
  });

  console.log('Loaded Auth Plugin!');
});

```

touhidurrr:
Anyways, my sloppy implementation @GGGuenni :
```ts
// src\plugins\Auth.ts
import * as argon2 from 'argon2';
import fp from 'fastify-plugin';
import UncivParser from '../modules/UncivParser';
import type { UncivJSON } from '../types/UncivJSON';

// declare types
declare module 'fastify' {
  interface FastifyRequest {
    game?: UncivJSON;
    userId?: string;
    playerId?: string;
  }
}

export default fp(async function (server) {
  async function getHashWithCache(userId: string) {
    let cachedHash = await server.cache.auth.get(userId);
    if (!cachedHash) {
      const dbHash = await server.db.Auth.findOne({ _id: userId });
      await server.cache.auth.set(userId, dbHash?.hash ?? '');
      return dbHash?.hash ?? null;
    }
    return cachedHash;
  }

  server.addHook('preValidation', async function (req, reply) {
    const isAuthRoute = req.url.startsWith('/auth');
    const isFilesRoute = req.url.startsWith('/files');

    // return if not /auth or /files
    if (isAuthRoute || isFilesRoute) return;

    // if /files, parse game and get playerId
    if (isFilesRoute) {
      const game = UncivParser.parse(req.body as string);
      const { currentPlayer, civilizations } = game;
      const { playerId } = civilizations.find(civ => civ.civName === currentPlayer)!;
      req.game = game;
      req.playerId = playerId;
    }

    if (!req.headers.authorization) {
      // an /auth request without authorization header will fail
      if (isAuthRoute) return reply.code(401).send('401 Unauthorized!');
      else {
        const hash = await getHashWithCache(req.playerId!);
        // if no auth header but hash is found, validation fails
        if (hash) return reply.code(401).send('401 Unauthorized!');
      }
      // not /auth and no hash is found auth header, validation success
      return;
    }

    const [type, auth] = req.headers.authorization.split(' ');
    if (type === 'Basic') {
      const userPass = Buffer.from(auth, 'base64').toString('utf8');
      const colonIdx = userPass.indexOf(':');
      const userId = userPass.slice(0, colonIdx);
      const password = userPass.slice(colonIdx + 1);

      // no userId == error!
      if (!userId || (isFilesRoute && userId !== req.playerId)) {
        return reply.code(401).send('401 Unauthorized!');
      }

      let hash = await getHashWithCache(userId);

      // if no hash is found, validation done
      if (!hash) return;

      if (await argon2.verify(hash, password)) {
        req.userId = userId;
        return;
      } else {
        return reply.code(401).send('401 Unauthorized!');
      }
    }

    // cannot recognize auth type
    return reply.code(400).send('400 Bad Request!');
  });

  console.log('Loaded Auth Plugin!');
});
```
```ts
// src\routes\auth\get.ts
import type { FastifyRequest } from 'fastify/types/request';

const getAuth = async (req: FastifyRequest) => {
  return '200 OK!';
};

export default getAuth;
```
```ts
// // src\routes\auth\put.ts
import * as argon2 from 'argon2';
import type { FastifyRequest } from 'fastify/types/request';

const idRegex = /^[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}$/;

const putAuth = async (req: FastifyRequest<{ Body: string }>) => {
  const { userId, body, server } = req;
  const hash = await argon2.hash(body);

  const isClientId = idRegex.test(userId!);
  await Promise.all([
    server.cache.auth.set(userId!, hash),
    server.db.Auth.updateOne(
      { _id: userId },
      { $set: { hash, type: isClientId ? 'client' : 'discord' } },
      { upsert: true }
    ),
  ]);
  return '200 OK!';
};

export default putAuth;
```

PR #8728, opened 2023-02-22
Android: use best possible device frame rate

vegeta1k95:
Android improvement.

Usually Android devices have adaptable refresh rate: 
- when something going on onscreen, or user touches the screen, refresh rate is automatically set up higher
- when device is idle/nothig going on, refresh rate is set to default 60Hz. 

**Problem:**
But, occasionally (I haven't figured out precisely when) Android device just locks itself on 60Hz screen refresh rate, even if it is capable of more, and doesn't change it independently of touches, animations, etc.
This happens even if phone has 100% battery charge and even if currently plugged to a charger, so very unlikely it is power preserving mechanism.
It happens systemwide in all application which do not **explicitly** request the higher refresh rate.

_This can be tested via developed options FPS toggle._

**Solution:**
This PR allows devices with Android 6(M)+ to lift this restriction and request the highest refresh rate for a game.
So, this means the game will always have:
- Smooth panning and zooming (on 60Hz they seem kind of "laggy")
- Better touch responses.

### DIFF
--- a/android/src/com/unciv/app/AndroidLauncher.kt
+++ b/android/src/com/unciv/app/AndroidLauncher.kt
@@ -3,9 +3,14 @@ package com.unciv.app
 import android.content.Intent
 import android.graphics.Rect
 import android.net.Uri
+import android.opengl.GLSurfaceView
+import android.os.Build
 import android.os.Bundle
+import android.view.Surface
+import android.view.SurfaceHolder
 import android.view.View
 import android.view.ViewTreeObserver
+import androidx.annotation.RequiresApi
 import androidx.core.app.NotificationManagerCompat
 import androidx.work.WorkManager
 import com.badlogic.gdx.Gdx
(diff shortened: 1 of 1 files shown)


vegeta1k95:
Okay, I figured out when refresh lock happens: 
It happens when screen brightness is turned below half (manually or automatically, at night).

I always wondered, why game seems so laggy at night while playing in my bed 😁

yairm210:
Looks good, but this is truly an area where I have no expertise whatsoever so I'm not sure I'm qualified to review
@GGGuenni ?

GGGuenni:
Well, I didn't know this is a thing. [Docs](https://developer.android.com/guide/topics/media/frame-rate) state that using `supportedModes` is not required but this way we always get max refresh rate which is fine. (?)  Honestly, I'm not able to see a difference even between a 75hz and a 144hz monitor so I'm probably also the wrong person to ask 😂 
Is the higher refresh rate noticeable regarding battery life? Android is probably lowering the refresh rate automatically when in battery saving mode right?

> The app does not need to consider the actual supported display refresh rates, which can be obtained by calling [Display.getSupportedModes()](https://developer.android.com/reference/android/view/Display#getSupportedModes()), in order to safely call setFrameRate(). For example, even if the device only supports 60Hz, call setFrameRate() with the frame rate your app prefers. Devices that don't have a better match for the app's frame rate will stay with the current display refresh rate.

vegeta1k95:
> Android is probably lowering the refresh rate automatically when in battery saving mode right?

Android is lowering refresh rate to 60 Hz in the following cases (in order of urgency):

1) Battery save mode
2) Screen brightness below half
3) Screen is idle, nothing happening onscreen

This PR makes game override points 2) and 3): game uses highest refresh rate always, except for battery save mode, in which it is lowered to default 60Hz.

GGGuenni (approved):
Then it's fine I guess

Issue #8625, opened 2023-02-08
Game cannot be saved when choosing what random nations you want with `Epic of Fantasy` enabled


alexban011:
**Platform**
Any

**Version**
4.4.12 + master

**Describe the bug**
The game cannot be saved if you choose the random nations and have Epic of fantasy loaded. This issue seems to only happen with Epic of fantasy. 

**To Reproduce**
1. Create a new game with Epic of Fantasy selected
2. Enable "set available nations for random pool"
3. Populate the list
4. Try to start the game.
5. Optional: remove save() from GameStarter:138 and notice the game loads and crashes only when saving the game.

**Additional context**
Looks like the game is created successfully but it cannot be serialized.

alexban011:
yes. It works fine if that option isn't enabled

Issue #8465, opened 2023-01-24
Crash when using blacklist nations feature w/ Epic of Fantasy mod


letstalkaboutdune:
**Platforms:** Win10x64 and Android 13

**Version:** 4.4.5, installedfrom GitHub.

**Description:** This is related to the blacklist nations from the random pool feature added by @alexban011 in #8430. Only allowing mod nations in the random pool results in a crash. Doing the opposite - blacklisting all base game nations - seems to work.

**Note:** crash seems limited to the `Epic of Fantasy` mod. I tried `Medieval Civilizations` and `Latin American_Civs` and they worked without crashing.

**To Reproduce:**
1.  Open the `Start New Game` screen
2. Enable the extension mod `Epic of Fantasy` that adds new nations
3. Enable the `Random nations pool` option and only select nations from the mod
4. Choose `Random` as the nation for one or more players
5. Click `Start game!`
6. Unciv will crash

Crash report: [Crash_only_mod_civs_allowed.txt](https://github.com/yairm210/Unciv/files/10492435/Crash_only_mod_civs_allowed.txt)

letstalkaboutdune:
Other feedback on this feature:

1. When selecting the `Random nations pool` option for the first time, the entire `Advanced Settings` section collapses. This makes it difficult to use the feature or know if it's working.
2. The settings of `Random nations pool` are not preserved when in a game and selecting the `Start New Game` option, like other settings are.
3. Some column labeling would be very helpful, with some examples below.

In the "positive definition" case of only allowing certain nations to the pool:
![Random_pool_allowed](https://user-images.githubusercontent.com/56904240/214377391-a2db9466-bf3c-4266-b52f-b4852e234301.png)

In the "negative definition" case of only banning certain nations from the pool:
![Random_pool_banned](https://user-images.githubusercontent.com/56904240/214377420-8b36f5d1-d632-48fc-86ff-8af8ac85ae8e.png)

alexban011:
> When selecting the Random nations pool option for the first time, the entire Advanced Settings

This happens because I'm updating that part of the screen. This can be fixed

letstalkaboutdune:
> Yep again, labels and maybe just get rid of the blacklist option and add a button in the popup to toggle between all civs. Might be more intuitive to the users. So the lables would be: `available civs` and `banned civs` and the user would be able to toggle all at once with a button

[...]

PR #8730, opened 2023-02-23
Allow filters in the ModOptions "ToRemove" lists

SomeTroglodyte:
Inspired by #8625 and #8465 mentioning a certain mod which says "not balanced around vanilla civs but around itself" in its readme. Allows the modder to prune those ruleset collections for which a `*ToRemove` already exists with broader filters instead of removing entries individually. 

@AceHank - would that suit you? I tested `nationsToRemove:["Major"]` on your EoF and it works as intended, but it may be a little harsh to preclude mixing your nations with vanilla ones entirely??

Edit: The "*" wildcard idea wasn't that nice, e.g. to clear Nations one would likely be able to leave Spectator and maybe the City States...

Q: Shouldn't matchesFilter by now be an abstract on IRulesetObject or something similar?

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -120,10 +120,13 @@ class Ruleset {
 
     fun add(ruleset: Ruleset) {
         beliefs.putAll(ruleset.beliefs)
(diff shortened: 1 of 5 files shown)


yairm210:
Cool, looks good to me!

PR #8741, opened 2023-02-24
Map exploring disables undo button + ExploredRegion smallfixes

Gualdimar:
Undo button:
- prevent the undo action if a new tile is explored
- fixed comment

Expored region:
- removed redundant boolean assignment
- fixed region x padding for non world wrap maps
- no longer can you go beyond the explored region using the buttons on the keyboard


vegeta1k95:
Looks good to me

PR #8747, opened 2023-02-25
exploredTiles deprecation, step 1 - do not check exploredTiles when checking if a tile is explored

yairm210:
exploredTiles continues to remain updated, to allow 'intermediate' players to not ruin the explored tiles of other players
If no negative feedback comes after this release (4.5.1), then full deprecation will enter 4.5.2


vegeta1k95:
Good, lets hope nothing is broken by this

vegeta1k95 (approved):

PR #8748, opened 2023-02-25
Better control over debug messages

SomeTroglodyte:
Another entirely optional offer...

After a debug session recently that continued to scroll new messages into the console window of Android Studio several minutes _after the process exit_, I thought I should dig this out of the shelf and show it.

Simple: You work on - say - MapUnit code? Add `-DonlyLog=Unit` to your run config right where you put your `-Xmx4096m`, presto lightweight console output.


yairm210:
Sure why not

PR #8763, opened 2023-02-26
Fix broken random nations pool

### LINKED ISSUE #8465: Crash when using blacklist nations feature w/ Epic of Fantasy mod

### LINKED ISSUE #8625: Game cannot be saved when choosing what random nations you want with `Epic of Fantasy` enabled

SomeTroglodyte:
I believe this resolves #8625 and resolves #8465.

Changes:
- Persistence of randomNations as collection of fully serialized Nations was a bad idea, now Strings. No compat code necessary as it was never working, and an emptyList<Nation>() serializes the same as an emptyList<String>().
- Fixed: Random from pool wasn't random at all
- Fixed: Incomplete clone led to options resetting (clone code reordered to field order for easier checking)
- Removed: Unnecessary keepOpen vars
- No reason not to let Advanced Expander state persist (better fix for the problem those keepOpen were afaicg introduced for)
- startsOpen is now "only" a suggestion for the first run of NewGameScreen once per game launch - no longer open by default for rnd #CS or rnd #players as the use of these two can readily be seen with the xpander closed.
- The Advanced and Mods Expanders had different widths - aesthetically imperfect - growX() vs growX() one nesting level deeper. Went the easier way - no growX(), else would have had to review center column -adv map params- too. Also looks better for portrait.
- Made vertical spacing between advanced checkboxes equal to between the mods
- Tweaked vertical spacing between adv and mods slightly - was fine in portrait but a little big in landscape

Fly-by observations, code untouched:
- NewGameScreen.unlockTables and lockTables can probably be removed, likely remainder from removed scenario maps?
- MapParameters clone could benefit from above-mentioned ordering
- MapParameters.numberOfTiles vs MapParameters.getArea - the latter (I did that though I barely remember) makes a distinction for worldwrapped rectangular with odd width - which is problematic anyway as these _must_ drop a column that would otherwise not mesh up with the other side. Should the UI maybe _prevent_ odd width for ww-rect? Otherwise, numberOfTiles only affects scoring while getArea is unused...
- Scrolling of a worldwrapped medium flat earth map is broken. Maybe ww makes no sense for fe anyway?

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -221,11 +221,10 @@ object GameStarter {
 
     private fun addCivilizations(newGameParameters: GameParameters, gameInfo: GameInfo, ruleset: Ruleset, existingMap: Boolean) {
         val availableCivNames = Stack<String>()
-        // CityState or Spectator civs are not available for Random pick
         if (gameSetupInfo.gameParameters.enableRandomNationsPool) {
-            for (nation in gameSetupInfo.gameParameters.randomNations)
-                availableCivNames.add(nation.name)
+            availableCivNames.addAll(gameSetupInfo.gameParameters.randomNations.shuffled())
         } else
+            // CityState or Spectator civs are not available for Random pick
             availableCivNames.addAll(ruleset.nations.filter { it.value.isMajorCiv() }.keys.shuffled())
 
         availableCivNames.removeAll(newGameParameters.players.map { it.chosenCiv }.toSet())
@@ -237,7 +236,7 @@ object GameStarter {
             gameInfo.civilizations.add(barbarianCivilization)
         }
 
-        val civNamesWithStartingLocations = if(existingMap) gameInfo.tileMap.startingLocationsByNation.keys
+        val civNamesWithStartingLocations = if (existingMap) gameInfo.tileMap.startingLocationsByNation.keys
             else emptySet()
         val presetMajors = Stack<String>()
         presetMajors.addAll(availableCivNames.filter { it in civNamesWithStartingLocations })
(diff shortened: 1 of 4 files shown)


yairm210:
Lock - if unused we can remove

The different Area functions are only useful in determining how big a map will be, so we can warn the user if it's too big. Anything else is irrelevant

yairm210:
Ww flat earth - tbh if anyone if choosing that setting that's on them, why would you want that

PR #8773, opened 2023-02-27
Cleaning: platform specifics and UncivGame

vegeta1k95:
This is an attempt to clean `UncivGame` from objects which shouldn't belong there. We want to shorten the amount of global references to `UncivGame`.

Most of platform specific helpers are static (they don't depend on `Game` state), hence were redistributed to other classes where they belong better:

1) `CrashReportSysInfo` removed, it's `.getInfo()` moved to `Log.getSystemInfo()`, since log backend is already platform-dependent.
2) `FontImplementation` moved to `Fonts`, since it is only used there.
3) `CustomFileLocationHelper` refactored:
- renamed to `PlatformSaverLoader` (`AndroidSaverLoader`, `DesktopSaverLoader` respectively);
- less messy callbacks implementation on both Android and Desktop;
- moved to `UncivFiles`, because used only there;

4) `PlatformSpecificHelpers` refactored:
- renamed to `PlatformSpecific`;
- some functions are removed from the interface, because they did nothing or could be set as static variables for the whole JVM lifecycle;
- now it is an interface `UncivGame` has to implement;

5) `UncivGame` is now subclassed on each platform: `AndroidGame` and `DesktopGame`.
- `AndroidGame` implements `PlatformSpecific` functions that are `Activity`-dependent;
- `DesktopGame` handles Discord functionality in form of new `DiscordUpdater` object;

6) Debug variables `viewEntireMapForDebug`, `superchargedForDebug`, `simulateUntilTurnForDebug`
moved to new static global object `DebugUtils`, they were static and game state independet anyway.


yairm210 (approved):
Looks good, with a change this big I'm pretty sure there will be knockbacks but we'll deal with them in time

vegeta1k95:
@yairm210 
I tested out every functionality that this change may have touched on both platforms, except for desktop Discord status (I don't have desktop Discord).
If you can confirm that Discord is doing OK everything should be good then.

PR #8779, opened 2023-03-01
Fixed wrong implementation of Basic auth

GGGuenni:
+ added min length (6) for passwords

Basic auth is [supposed](https://en.wikipedia.org/wiki/Basic_access_authentication) to base64 encode the whole userId:password string instead of encoding userId and password separately

@touhidurrr and @oynqr since you've already started implementing auth

### DIFF
--- a/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
+++ b/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
@@ -216,7 +216,7 @@ class MultiplayerTurnCheckWorker(appContext: Context, workerParams: WorkerParame
                         Pair(USER_ID, settings.userId), Pair(CONFIGURED_DELAY, settings.turnCheckerDelay.seconds),
                         Pair(PERSISTENT_NOTIFICATION_ENABLED, settings.turnCheckerPersistentNotificationEnabled),
                         Pair(FILE_STORAGE, settings.server),
-                        Pair(AUTH_HEADER, "Basic ${Gzip.zip(settings.userId)}:${Gzip.zip(settings.passwords[settings.server] ?: "")}"))
+                        Pair(AUTH_HEADER, "Basic ${Gzip.zip("${settings.userId}:${settings.passwords[settings.server] ?: ""}")}"))
 
                 if (settings.turnCheckerPersistentNotificationEnabled) {
                     showPersistentNotification(applicationContext, "—", settings.turnCheckerDelay)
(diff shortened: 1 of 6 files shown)


oynqr:
This still compresses the string before base64 encoding it, no? It's not supposed to be compressed.

android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
```diff
@@ -216,7 +216,7 @@ class MultiplayerTurnCheckWorker(appContext: Context, workerParams: WorkerParame
                         Pair(USER_ID, settings.userId), Pair(CONFIGURED_DELAY, settings.turnCheckerDelay.seconds),
                         Pair(PERSISTENT_NOTIFICATION_ENABLED, settings.turnCheckerPersistentNotificationEnabled),
                         Pair(FILE_STORAGE, settings.server),
-                        Pair(AUTH_HEADER, "Basic ${Gzip.zip(settings.userId)}:${Gzip.zip(settings.passwords[settings.server] ?: "")}"))
+                        Pair(AUTH_HEADER, "Basic ${Gzip.zip("${settings.userId}:${settings.passwords[settings.server] ?: ""}")}"))
```
yairm210, line 219: This is just confusing to read, please separate the different parts out into different lines.
val password = settings.passwords[settings.server] ?: ""
val prezippedAuthValue = "${settings.userId}:$password"
val zippedAuthValue = Gzip.zip(prezippedAuthValue)
val auth_header = "Basic $zippedAuthValue"

yairm210: Also, please use one unified function - currently implemented in 3 places

touhidurrr:
@GGGuenni is auth version still 1 or 2?

GGGuenni:
Still 1

touhidurrr:
oh, also why is argon2 verify not working? am I getting something wrong here? isn't the put request body the new password?
```ts
import * as argon2 from 'argon2';
import type { FastifyRequest } from 'fastify/types/request';

const idRegex = /^[\da-f]{8}-([\da-f]{4}-){3}[\da-f]{12}$/;

const putAuth = async (req: FastifyRequest<{ Body: string }>) => {
  const { userId, body, server } = req;
  const hash = await argon2.hash(body);

  const isClientId = idRegex.test(userId!);
  await Promise.all([
    server.cache.auth.set(userId!, hash),
    server.db.Auth.updateOne(
      { _id: userId },
      { $set: { hash, type: isClientId ? 'client' : 'discord', timestamp: Date.now() } },
      { upsert: true }
    ),
  ]);
  return '200 OK!';
};

export default putAuth;
```

GGGuenni:
It is but I forgot to remove the compression from the password sorry!
Will make a PR

touhidurrr:
ok

PR #8785, opened 2023-03-02
Android: "Screen Mode" option

vegeta1k95:
Now Android has select box "Screen Mode" similar to Desktop version in Display options:
image

This select box contains various supported display modes for the game: 
- "Default": no preferences on refresh rate, width, height - those will be managed and adapted by system
- Other modes have preferable refresh rate, width, height  - those will be locked and override system behaviour.


yairm210:
The battery drain is noticable, otherwise users would wouldn't notice (and probably Android wouldn't have this feature to begin with)

vegeta1k95:
> Is the benefit really noticeable?

Difference in movement is noticeable for me between 60 and 90 HZ.
But users apparently reported noticeable battery drain too, so decided to make it an option

PR #8789, opened 2023-03-02
Map editor DragPaintListener fix

### LINKED ISSUE #8737: Feature request: Map build UI glitch
In the map build UI when two finger zooming, if one finger is left touching then dragging (to scroll the screen), the last object from Edit is painted/spammed across tiles instead of scrolling. This happens even when switching to View mode where one should be safe from making any edits.

Gualdimar:
Closes #8737


hackedpassword:
Most appreciated!

PR #8790, opened 2023-03-02
Fixed setPassword body is compressed

GGGuenni:
+ Updated server to also use the correct Basic implementation

Everything should be done client-wise once this is merged if you don't have any more issues @oynqr @touhidurrr

### DIFF
--- a/core/src/com/unciv/logic/multiplayer/storage/UncivServerFileStorage.kt
+++ b/core/src/com/unciv/logic/multiplayer/storage/UncivServerFileStorage.kt
@@ -2,7 +2,6 @@ package com.unciv.logic.multiplayer.storage
 
 import com.badlogic.gdx.Net
 import com.badlogic.gdx.utils.Base64Coder
-import com.unciv.ui.screens.savescreens.Gzip
 import com.unciv.utils.debug
 import kotlin.Exception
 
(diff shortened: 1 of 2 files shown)


touhidurrr:
Also this needs 100% rollout I guess for https://github.com/touhidurrr/UncivServer.xyz/pull/5 to be merged.

touhidurrr (approved):

touhidurrr:
@GGGuenni @yairm210 I think we need to use domains or groups as keys. Currently, `uncivserver.xyz` and `replit.uncivserver.xyz` will be considered to be two separate entities.
There is 2 solutions that I can suggest:
1. use domain name as key for the hash
2. ping isalive and collect group  name and use it as a key. something like `{ authVersion: 2, group: 'uncivserver.xyz' }`.

touhidurrr:
In solution 2, the client will keep track of domains under a group and then group password.

GGGuenni:
What are you using the `replit` subdomain for if they both serve the same files?

The only thing I see that might be a problem is setting the password for one changes the password for the other (if the server is implemented that way). But I don't think that's a huge problem. If you change the password and switch the domain you will be asked for a password once and that's it

touhidurrr:
> What are you using the `replit` subdomain for if they both serve the same files?

Basically the database is the same, but the server is different. That subdomain runs on a different virtual server.
Since UncivServer.xyz has a Redis cache mechanism per server, and the server never waits for the game to uploaded to the central db, different servers sees different latencies for different users based on region.

touhidurrr:
@GGGuenni server error message is not being displayed on client on set password failure after https://github.com/touhidurrr/UncivServer.xyz/commit/dd4a2443d3f104cda0acff056141e8d55e70ecbchttps://github.com/touhidurrr/UncivServer.xyz/commit/dd4a2443d3f104cda0acff056141e8d55e70ecbc, can you fix this?

GGGuenni:
Since I can only change future releases (or the current with hotfixes) it does not make any difference if I change something
4.5.2 will work everything below is stuck with `Failed to set password!`

GGGuenni:
Still applies that it will not trigger as 4.5.3+ users will not get the message but are the first ones which could receive it

PR #8795, opened 2023-03-02
Fix scroll-to crossing worldwrapped map seam

SomeTroglodyte:
Currently this PR features:
- Fix Scroll-to triggered by minimap or next-unit crossing worldwrap "seam"
- Keyboard panning crossing worldwrap "seam"
- Mouse-to-edges panning crossing worldwrap "seam"
- Fix mouse panning not respecting the limits imposed by explored area
- Different panning speeds for keyboard and mouse
- Mouse panning "hot zone" wider at top/bottom than left/right - I saw no reason for that
- Made Mouse panning "hot zone" a little wider and made it so it accelerates the nearer you get to the edge. At the edge it's 33% faster than keyboard, feels just about right to me
- Configurable panning speed for keyboard and mouse
- Both panning methods now available in Map editor - Arrows only so WASD stay available for the rest of the UI

I've tested enough by now that my original "why" question feels obsolete. I know of two bugs..
- One: _very_ difficult to repro and also present in master. (Try a very fast flick-pan on desktop near the seam such that you'd cross the seam - if speed and length of mouse-down is in a very narrow window, the map will jump a good distance "away" then animate back quickly to the actually correct new viewport - may be more difficult to produce without the slowdowns by running under debugger)
- Two: Also not caused by this, the limiting of panning to explored area fails as soon as your exploration crosses worldwrap seam - the code was built without that feature in mind. Should be a matter of explicitly allowing min > max meaning "outside the range" and fixing restrictXY functions accordingly - not this PR, and would need the coder to be actually awake at their keyboard...

~~One to-do left:~~
- [x] Test thoroughly with worldwrap but only partially explored map, especially both with starting location near seam and far from it

Done, happy enough with the result.

<hr/>
<details><summary>-- Original Post --</summary>

Important: Someone please explain why this works!

Currently some world map scrolling does not work as expected on worldwrapped maps. Affects minimap scrolling and auto-next-unit, maybe more.

<details><summary>A little demo</summary>

A little small but zoom in to see where the "camera" is and where I clicked next:
image

Video what happened - the pan "stuck to the seam" and didn't reach the intended destination - in real life it's much clearer that there was a "change of direction":

https://user-images.githubusercontent.com/63000004/222562035-f716482d-464e-4161-b05a-f6b3abf0a1e3.mp4

<br/>
<br/>
Video same thing with this patch:

https://user-images.githubusercontent.com/63000004/222562162-71865fc3-a134-4725-9473-499f27cb47da.mp4

</details>

Now I'm sure I had seen cases where the pan "took the much longer way" - destination is e.g. 10 tiles eastward but 50 tiles westward - remember talking about worldwrapped maps - and scrollTo decides to go the 50 tiles - but I couldn't repro.

This will surprisingly _not_ cause a **non**-wordwrapped map to pan over a non-existing seam - it will pan correctly over the whole map to move from one edge to the other. Honestly, I am not sure why this simple patch has no need for the ZoomableScrollPane to [_know_](https://torment.fandom.com/wiki/Dak%27kon) whether the map is wrapped or not....

Small aside: Last time I played a while ago I'm sure there was a function to re-center on a selected unit or city after scrolling away - click on name? I think it's a pity the name click now goes exclusively to rename - that was fine when reserved to within a city screen or as reward for reaching a promotion...
</details>


Gualdimar:
> Important: Someone please explain why this works!

This is because "teleportation" to the other side of the map is only implemented in the FlickScrollListener : ActorGestureListener() and when the Scroll pane scrolls to it bypasses the ActorGestureListener().

Your line adds "teleportation":
`if (newX < 0f) newX += maxX else if (newX > maxX) newX -= maxX`



> was a function to re-center on a selected unit or city after scrolling away

Works if you click on the table itself

Gualdimar:
Thanks to you, I figured out that world-wrap doesn't work if you scroll with the buttons on the keyboard😁
Can you check if your PR fixes this as well?

Gualdimar:
I think all direct assignments to variables scrollX and scrollY should be replaced by two functions, for example: scroll(deltax, deltaY) and scrollTo(x,y). Then all the logic will be in one place. 
But I don't have enough free time to do it now.

> did you know that keyboard panning could double speed by pressing both Up and W etc simultaneously?

Not sure if this is a bug or a feature😅

Gualdimar:
> * Two: Also not caused by this, the limiting of panning to explored area fails as soon as your exploration crosses worldwrap seam - the code was built without that feature in mind. Should be a matter of explicitly allowing min > max meaning "outside the range" and fixing restrictXY functions accordingly - not this PR, and would need the coder to be actually awake at their keyboard...

Can you provide a more detailed description? So I know how to reproduce and fix it.

Gualdimar:
Oh, I think I get it, although I'm not completely sure. Try to move this `if` right before `return result` in restrictX():

'''
        if (continuousScrollingX) {
            if (result < 0f) result += maxX
            else if (result > maxX) result -= maxX
        }
'''

Teleportation should be done after the restriction, not before.

Gualdimar:
After working with the minimap recently, I realized that rectangular maps use height and width, not radius. I had missed this when implementing the explored region. So, for rectangular maps with a world wrap, restriction may not work correctly near the edge of the map. I'll fix it when I get to my workstation.
Wonder if this was the cause of this bug🤔

Gualdimar:
> Oh, I think I get it, although I'm not completely sure. Try to move this `if` right before `return result` in restrictX():
> 
> ```
>         if (continuousScrollingX) {
>             if (result < 0f) result += maxX
>             else if (result > maxX) result -= maxX
>         }
> ```
> 
> Teleportation should be done after the restriction, not before.

This change will disable teleportation if exploredRegion.shouldRestrictX() == true, but teleportation should be done after the restriction. Otherwise, if the edge of the explored region is on the edge of the map, a hard swipe with the mouse or finger will "jump over" the restriction...

Gualdimar:
> in normal hex coords

Since the hex coordinate system is offset by an angle relative to the world one, longitude and latitude were used instead of hex coordinates to calculate the extreme points. The longitude and latitude are then converted to the stage coords.



> does it rely on newly tested tile alwasy being adjacent to the so-far recorded area? Is that guaranteed?

If tile is beyond the region - yes.

> After all, moving onto a hill can suddenly see mountains at least 3 tiles away, can it not?

In such case the region will be extended by 3 tiles in that direction) Same works for city-states reveal, after encountering their units.

> I think we need a debug option to paint the hex coords directly onto the map

Great idea! Along with stage coords.

core/src/com/unciv/ui/components/ZoomableScrollPane.kt
```diff
@@ -261,23 +264,16 @@ open class ZoomableScrollPane(
...
-
-            if (posY <= 6f) {
-                scrollY -= 3f
-            } else if (posY >= stage.viewport.screenHeight - 6f) {
-                scrollY += 3f
+            fun getPanDelta(pos: Int, max: Int): Float {
+                if (pos < 8) return (pos - 8) / 6f  // negative!
+                if (pos > max - 8) return (8 - max + pos) / 6f
+                return 0f
             }
```
yairm210, line 271: Please use clearer names and add comments.

Currently, I do not understand the logic here:
What is the 6f about? Why would the *delta* include the current position instead of just moving us around by a set "
max - padding" amount? I confuse

core/src/com/unciv/ui/components/ZoomableScrollPane.kt
```diff
@@ -333,15 +334,17 @@ open class ZoomableScrollPane(
...
         private val originalScrollX = zoomableScrollPane.scrollX
         private val originalScrollY = zoomableScrollPane.scrollY
 
-        override fun update(percent: Float) {
-            zoomableScrollPane.scrollX = zoomableScrollPane.scrollingTo!!.x * percent + originalScrollX * (1 - percent)
-            zoomableScrollPane.scrollY = zoomableScrollPane.scrollingTo!!.y * percent + originalScrollY * (1 - percent)
-            zoomableScrollPane.updateVisualScroll()
+        override fun update(percent: Float) = zoomableScrollPane.run {
+            var newX = scrollingTo!!.x * percent + originalScrollX * (1 - percent)
+            if (newX < 0f) newX += maxX else if (newX > maxX) newX -= maxX
```
yairm210, line 344: Add comments e.g. // moving right through wraparound border // moving left through wraparound border

Gualdimar:
> So, for rectangular maps with a world wrap, restriction may not work correctly near the edge of the map.

Fixed in #8826. Tested with this PR and could not catch this bug, but there are never enough tests)
There are other problems with this PR, as I said:
- map editor world wrap not working
- teleportation breaks restriction. Suggestion:
```
override fun restrictX(deltaX: Float): Float {
        val exploredRegion = worldScreen.viewingCiv.exploredRegion
        var result = scrollX - deltaX

        if (exploredRegion.shouldRecalculateCoords()) exploredRegion.calculateStageCoords(maxX, maxY)

        if (exploredRegion.shouldRestrictX()) {
            val leftX = exploredRegion.getLeftX()
            val rightX = exploredRegion.getRightX()

            if (deltaX < 0 && scrollX <= rightX && result > rightX)
                result = rightX
            else if (deltaX > 0 && scrollX >= leftX && result < leftX)
                result = leftX
        }

        if (continuousScrollingX) {
            if (result < 0f) result += maxX
            else if (result > maxX) result -= maxX
        }

        return result
    }
```

PR #8802, opened 2023-03-03
Custom key bindings

SomeTroglodyte:
#### Allows the user to configure which key is bound to which action...
image

Side benefit: An excuse keys aren't internationalized - like tschermanz needing to hit "Z"(z-es) instead of "Y"(yes) or "J"(ja) - or François wishing "O" to say "Oui" - now one can fix that themselves.

#### To do:
- Bind all appropriate keys through this - now I have only implemented a **few samples** - chore
- Better UI (now: works if you know the english, untranslated representation) - create widget that translates an actual keypress into our internal representation - definitely a separate PR
- Translatability of the key representations (just a grind, had to draw a line somewhere)

#### Tech speak
- Read the commit messages, I've separated stuff cleanly (except one outdated WorldScreen comment not in the lint commit)
- I consciously avoided changing Paul's work - but I split his KeyShortcutDispatcher into its own source file, prerequisite to understand the interactions.
- ~~Can't show a screenshot - at the moment Studio crashes when compiling, internal error - wanted to show this for inspection nevertheless, and see if the docker boxes here can still compile.~~
- Side effect of fixing said compiler error - moved SettingsSelect to own class file, OptionsPopup now has the blue round icon back as it should
- Will provoke new "translation required"'s through new TFW loop -> only ~~kr..~~ German included

Update: same compile error here, back to basics - I was sure the exact same sources already ran fine - oh well.
Update: The compiler crash was caused by following an inspection suggestion on Timo's code... Needs a few experiments. ***VERY*** interesting. The Kdoc of `by` is slightly wrong.


yairm210:
Urggh I really don't want to deal with this.
I can just imagine the endless requests we'll get to add more and more of the keybindings to be customizable.

I'm willing to okay this, because I know that this is a _personal_ problem I have, and I shouldn't let this negatively impact the game - it's not just my game, it's yours as well. But still, I have to say - Big Oof.

yairm210:
I'm not sure we *have* a lot of actual real multistep projects. Most of the work now is small standalone issues, like this one

yairm210:
This is... not actually working for me.
image

It still going into empire overview on pressing E though

yairm210:
Also it accepts long strings and not just clicks, and *also* it doesn't seem to register when there are keybinding conflicts

yairm210:
I'm disabling this for the next version since it's practically unusable

yairm210:
(Also "Keys" should probably be before "Advanced")

yairm210:
I disabled it in the current version, not urgent

PR #8804, opened 2023-03-03
Make City center minimum tile yields moddable

SomeTroglodyte:
You know, the 2 Food, 1 Production minimum that's currently hardcoded.

Also does the small reduction of duplicate code in TileStatFunctions I hinted elsewhere. That one changes behaviour only if a mod manages to get TerrrainFeatures on a Natural Wonder tile - and then in a IMHO manner closer to expectations.

Question: I was tempted to define a reusable Stats.ZERO akin to Vector2.ZERO - but that would have the same problems with mutability. One could subclass to make largely compatible MutableStats / ImmutableStats variants - major refactoring... Still - opinions?


yairm210:
Would this mean that all mods will now not have minimum yields for city centers?

yairm210:
For now I would say - city centers by default give what they currently give, unless there's an explicit unique for "no stats", like [+0 Gold] (never tried that but it should parse as 'no stats')

core/src/com/unciv/logic/map/tile/TileStatFunctions.kt
```diff
@@ -76,14 +59,16 @@ class TileStatFunctions(val tile: Tile) {
...
+                val ensureMinUnique = improvement
+                    .getMatchingUniques(UniqueType.EnsureMinimumStats, stateForConditionals)
+                    .firstOrNull()
+                if (ensureMinUnique != null) minimumStats = ensureMinUnique.stats
+            }
         }
 
-        for ((stat, value) in stats)
-            if (value < 0f) stats[stat] = 0f
+        stats.coerceAtLeast(minimumStats)  // Minimum 0 or as defined by City center
```
yairm210, line 71: Since this is the only place this is actually used, I don't think we need a special function for that. We can do it here, looping over key/value pairs in the stats, in 3 lines of code. "for stat,value in stats.hashmap(); minimumValue = minimumStats[stat]; if value<minimumValue stats[stat]=minimumValue"

yairm210: I would still go with B 
If we see a third usage then I would go with C

yairm210: "Three's a crowd"

yairm210: You're right, got mixed up

PR #8816, opened 2023-03-05
Belarusian Locale for better sorting

SomeTroglodyte:
Nice flag by the way.
reference: https://en.wikipedia.org/wiki/ISO_3166-2:BY
Also ensures fastlane, once included, will end up in the proper folder

@monodevx  - see, we do take care of the details :grinning:

### DIFF
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -155,6 +155,7 @@ class GameSettings {
 
 enum class LocaleCode(var language: String, var country: String) {
     Arabic("ar", "IQ"),
+    Belarusian("be", "BY"),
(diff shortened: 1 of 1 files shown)


monodevx:
> Nice flag by the way. reference: https://en.wikipedia.org/wiki/ISO_3166-2:BY Also ensures fastlane, once included, will end up in the proper folder
> 
> @monodevx - see, we do take care of the details 😀

thanks, so much °-°

yairm210:
I merge translations only when I'm about to release a version, so it could take a couple of days

monodevx:
Thanks, for answer, Yairm.

Good day to you.

8 сакавіка 2023 г. 12:32:49 GMT+03:00, Yair Morgenstern ***@***.***> піша:

PR #8818, opened 2023-03-05
Debug option to paint coords on map tiles

SomeTroglodyte:
See [this comment](https://github.com/yairm210/Unciv/pull/8795#issuecomment-1454925108)

<details><summary>Sample</summary>

image
</details>


core/src/com/unciv/utils/Debug.kt
```diff
@@ -5,9 +5,12 @@ object DebugUtils {
     /**
      * This exists so that when debugging we can see the entire map.
      * Remember to turn this to false before commit and upload!
+     * Or use the "secret" debug page of the options popup instead.
      */
     var VISIBLE_MAP: Boolean = false
 
+    var PAINT_COORDS: Boolean = false
```
yairm210, line 12: Instead of "paint" how about "show coords on map"? This name is unnecessarily confusing

PR #8822, opened 2023-03-05
Android: selectable orientation

vegeta1k95:
This adds a selectable orientation on android: 
- Landscape (fixed, default)
- Portrait (fixed)
- Dynamic (changes according to system settings)
image

Also, few more code cleanups for platform launchers.


yairm210:
A. I'm slightly confused that you added 'landscape' as orientation. I'm sure it's required, I just don't understand why :)
B. Missing translations
C. Instead of "Dynamic" which could mean "we change it when we want", how about "system default" or "device setting" or something?

AdityaMH:
> C. Instead of "Dynamic" which could mean "we change it when we want", how about "system default" or "device setting" or something?

Reversed Landscape

AdityaMH:
> This adds a selectable orientation on android:
> 
> * Landscape (fixed, default)
> * Portrait (fixed)
> * Dynamic (changes according to system settings)

Wait, if someone who turn-off auto-rotate would be like this:
Landscape = 90° degree screen
Portait = Portrait
Dynamic = Portait

vegeta1k95:
> A. I'm slightly confused that you added 'landscape' as orientation. I'm sure it's required, I just don't understand why :) 

So that user can return again to default fixed Landscape mode, if he is in Portrait mode or Auto mode.
The last two override manifest behavior, so there should be a way to revert.

> B. Missing translations 

Added.

> C. Instead of "Dynamic" which could mean "we change it when we want", how about "system default" or "device setting" or something?

Renamed to `Auto (sensor adjusted)`. Now this mode ignores system Auto-Rotate setting and rotates screen nevertheless according to sensors.

image

vegeta1k95:
@AdityaMH 
Fixed in last commit, see above, now `Auto` option ignores system Auto-rotate setting and rotates screen anyway.

yairm210:
Going over this again it becomes clear that `class DesktopScreenMode` should *definitely* be an Enum, like the Android modes are in this PR...

PR #8839, opened 2023-03-06
Extended CI with various extra tests, added Java 17 support in unit tests (fixes #8835)

### LINKED ISSUE #8835: Feature request: add Java 17 support for unittests and code checks as well (the game seems to works with it)
Currently, running the code checks and the unit tests on the code base requires Java 11. Using Java 17 causes the runs to fail. Specifically, the two CI steps are affected:
- `./gradlew check`
- `./gradlew --no-build-cache cleanTest test tests:test`

Note that the game and the server seems to work, so this is not too high priority. Additionally, Java 11 is the LTS version; it gets updates until at least late 2024 (some distributions even longer).

You can verify this by e.g. adding a matrix for the CI jobs:

```yaml
jobs:
  build:
    strategy:
      matrix:
        java_version:
          - 11
          - 17

    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3

      - uses: actions/setup-java@v3
        with:
          cache: 'gradle'
          distribution: temurin
          java-version: ${{ matrix.java_version }}
```

CrsiX:
This new CI steps are executed for every new commit (and tag) in this repository, therefore it provides continuous testing of the source code. The Windows tests are stripped down to not build Android. The output of the CI jobs is uploaded as artifacts and could be tested on end user devices, if a user wants to test bleeding-edge software from a specific commit.

Additionally, Java 17 support has been added to the unittests and code checks by upgrading various dependencies.

Fixes #8835


CrsiX:
Indeed, that's my experience as well. To be honest, the artifact upload is nice to have, but since it fails so often, I think it would be wise to just mark it as allow-to-fail.
There's no overloading of the system, but instead inefficient use of resources. I could split up the jobs to further increase their efficiency (say, run Android builds on Ubuntu only, while also extracting test steps into its own job). This should speed up execution by a few minutes due to increased parallelism (I haven't checked yet).

CrsiX:
In my test case, this change resulted in significant speed increase if cache is available (less than two minutes run time) or est. 3 minutes speed increase if there's no cache available (max of around 10 minutes per job).
Additionally, I've marked the artifact upload as allowed to fail :)

yairm210:
Looks good
We can then reserve the deploy for tags only, so the deploy jobs don't appear in every pr

PR #8840, opened 2023-03-07
Move some Industrial sprites to Modern

Caballero-Arepa:
I moved some of the sprites of AU from the Industrial era to the Modern, as these cars are clearly unfit to be so early in the industrial.
   At the start of the era you aren't even close to unlocking the Steam Engine, not to mention Combustion.

This also partially correlates with HexaRealm units who update on modern era and not on Industrial.

Hopefully I didn't mess anything up xD


letstalkaboutdune:
I don't want this change. Industrial era is what Civ 5 uses as the turning point between early game and late game for purposes of unit variations. It also marks a turning point for handling of religion, Great Prophets, and Great People in general.

yairm210:
Denied

PR #8844, opened 2023-03-07
Add Replay feature in VictoryScreen

WhoIsJohannes:
I'm saving for each tile whenever ownership changes in a map from turn (int) to ownerCivName. I'm using a map from string to string for serialization as I was running into some issues with directly serializing a TreeMap and I didn't want to write a custom serializer.

The UI is quite rudimentary, but it does the job. I copied a lot of code from the Minimap. I'm not happy with the duplication, but I didn't see an easy way to make the code reusable as it's quite entangled with the Tilemap. I also had to add two things that I didn't understand how they work and had to change the logic:

- Updating border colors if the city was captured
- Removing the city circle if the city is razed

For some reason these two things work for the minimap, but I don't know why. At least I can't see it in the code and it didn't work for me in the Replay map, so I added it.


yairm210:
Pics please :)

WhoIsJohannes:
Thanks for looking into the code. I uploaded a screenshot here

https://drive.google.com/drive/folders/1lRm0dMhc8UYis4OZ-HtSRR2tj3APyQUg

I also tried to do a screen recording, but failed even after wasting half an hour and now I'm too angry to keep trying, so I leave it to your imagination. As you can see from the code it will advance the turn every 100ms and just update the year on the top along with the map.

yairm210:
Any particular reason the hashmap is keyed by string, and not by int? Turns are always int

WhoIsJohannes:
It's because of serialization. TreeMap is not handled well and on
deserialization they come out as string. It sucks, but because of type
erasure for generics in Java this is actually possible. You end up with a
map that is still typed as having int keys, but it's actually strings in
there. There's already some special serialization logic for this in the
code for Vector2d IIRC, but I didn't want to go that route and felt it's
easier to just have storage and logic objects.

Il mer 8 mar 2023, 11:54 Yair Morgenstern ***@***.***> ha
scritto:

Gualdimar:
Looks interesting, but missing translation.
And I suggest making the map size dependent on the screen size.
Here is a video with a replay of 250 turns
https://user-images.githubusercontent.com/1225948/223699280-03e4c80e-1409-42df-9632-24120d08df6e.mp4

yairm210:
I'm fine with strings as keys, not a huge issue for me ;)

yairm210:
I think that at least for the tiles, you could use MinimapTile and not have to define an extra class for the replay tiles

WhoIsJohannes:
Thanks for all the feedback and thanks for creating the recording!

> And I suggest making the map size dependent on the screen size.

I think this is already the case. It happens in `calcTileSize` in `ReplayMap.kt`. It references the dimensions of the world screen though, which probably ends up being accidentally correct. I am really not familiar at all with this UI framework and would appreciate any pointers on how to do this better if necessary.

> I think that at least for the tiles, you could use MinimapTile and not have to define an extra class for the replay tiles

I think so, too, but then we need to slightly change the API of the MinimapTile to accept an (optional?) turn? It could then decide whether to use the current state or the history state. We'd also still have to figure out how these currently work:

- Removal of the city circle from the minimap
- Recoloring of borders once a city is captured

Would you have any idea where this logic is? I couldn't find it. This is the other diff currently since I'm doing that "manually" in my ReplaymapTile impl.

Thanks!

WhoIsJohannes:
I'll put the translations on my TODO list. I guess everything I need can be found here?

https://yairm210.github.io/Unciv/Developers/Translations%2C-mods%2C-and-modding-freedom-in-Open-Source/

Gualdimar:
> I think this is already the case. It happens in `calcTileSize` in `ReplayMap.kt`

Oh, true. Then I suggest to increase the size of the map from 22% to 27%:
<details><summary>22%</summary>


</details>

<details><summary>27%</summary>


</details>



> I'll put the translations on my TODO list. I guess everything I need can be found here?

Moslty. The translation is done in helpers (toTextButton(), toLable(), etc.) or with .tr()
For "Repeat", you just need to add a line to template.properties.
There is already a translation for "[year] BC/AD", you just need to format the text in the appropriate way and call .tr(), for example:
`("[" + abs(year).toString() + "]" + (if (year < 0) " BC" else " AD")).tr()`
and don't forget about the Mayan year)

yairm210:
I phrased that wrong
I meant "inherit from minimap tile", that way you can add what you want on top of that

If it makes sense to you, you could inherit from the minimap as well to add the replay functionality

The more copied functions we can avoid, the better :)

WhoIsJohannes:
I don't think inheritance works well. MinimapTile reads some values from the Tile and neighboring tile while the other needs to read it from the history.

I could create a base class for both of them, but would prefer composition over inheritance.

 I can extract common logic into a helper that is then called from both implementations. What do you think? What would be a good location for the helper?

yairm210:
Ui/components
https://github.com/yairm210/Unciv/tree/master/core/src/com/unciv/ui/components

yairm210:
I'm okay with this

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapTileUtil.kt
```diff
@@ -0,0 +1,28 @@
...
+            tileLayer.addActor(image)
+
+            // keeps track of the current top/bottom/left/rightmost tiles to size and position the minimap correctly
+            topX = max(topX, image.x + tileSize)
+            topY = max(topY, image.y + tileSize)
+            bottomX = min(bottomX, image.x)
+            bottomY = min(bottomY, image.y)
+        }
+
+        return Rectangle2D.Float(topX, topY, topX-bottomX, topY-bottomY)
```
Gualdimar, line 26: I suggest replacing with `return Rectangle2D.Float(bottomX , bottomY, topX-bottomX, topY-bottomY)`, because topX/Y is not used, unlike bottomX/Y

Gualdimar:
Rectangular maps now bigger, but hexagonal maps don't fit🙃
<details><summary>Rectangular</summary>


</details>
<details><summary>Hex</summary>


</details>

Gualdimar:
Also there is a problem WorldScreenTopBar, after closing Replay screen, after the next worldScreen.update()

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapTileUtil.kt
WhoIsJohannes: I understand, however if I read the code right, topX and topY are always greater than bottomX and bottomY respectively, so if I do what you suggest, I'd end up with the lower left corner as the anchor and negative widths and heights just to avoid some computation on the receiver. I think the extra computation is the lesser evil. Let me know if you feel strongly about this. Another option would be to use a custom rectangle that is just defined by two points rather than one point and width and height, but I didn't want to reinvent the wheel for such a basic data structure.

core/src/com/unciv/logic/map/tile/TileHistory.kt
```diff
@@ -0,0 +1,64 @@
...
+import kotlin.collections.HashMap
+
+/**
+ * Records events throughout the game related to a tile.
+ *
+ * Used for end of game replay.
+ */
+open class TileHistory : IsPartOfGameInfoSerialization {
+
+    class TileHistoryState : IsPartOfGameInfoSerialization {
```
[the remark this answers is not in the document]
WhoIsJohannes: I think I had this initially and it complained about serialization. I think I was missing the default values since it complained about the lack of a default constructor. Just tried again and it works now with default values. Thanks for pointing this out. Done!

core/src/com/unciv/logic/map/tile/TileHistory.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    fun recordRelinquishOwnership(tile: Tile) {
+        historyForSerialization[tile.tileMap.gameInfo.turns.toString()] =
+                TileHistoryState()
+        historyForLookup = null
+    }
+
+    fun getState(turn: Int): TileHistoryState {
+        historyForLookup = historyForLookup
```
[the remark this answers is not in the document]
WhoIsJohannes: I agree with you. However I don't really have a callback on when this class gets intialized from the serialization / gets serialized, do I? If so I could hook into those to then set the string fields from the actual working fields.

core/src/com/unciv/logic/city/managers/CityInfoConquestFunctions.kt
```diff
@@ -324,6 +324,12 @@ class CityInfoConquestFunctions(val city: City){
             // Update proximity rankings
             civ.updateProximity(oldCiv,
                 oldCiv.updateProximity(civ))
+
+            // Update history
+            city.tiles.map { vector2 -> civ.gameInfo.tileMap[vector2] }
```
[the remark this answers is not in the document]
WhoIsJohannes: Done

WhoIsJohannes:
Yes, I didn't want to touch the NonStringKeyMapSerializer and thought this was the path of least resistance...

WhoIsJohannes:
> Rectangular maps now bigger, but hexagonal maps don't fit

This should now work.

> Also there is a problem with WorldScreenTopBar, after closing the Replay screen, after the next worldScreen.init()(next turn or resize)

I could repro this. Open the Replay, go to next turn. Top bar looks odd. I found the culprit. It's the padding bottom on the label. If I remove it, it doesn't happen. I tried applying a padding top to the element below (the ReplayMap) and thought that might solve it, but it still happens. Then I thought for some weird reason "".toLabel() ends up refering to the same object from two different places (there's actually something like this created in the top bar), but changing my label to initially be "asldfadsh".toLabel() doesn't solve the problem either. So I'm in the unpleasant situation of being able to solve the problem, but not understanding why it occured in the first place. Would appreciate if anybody has an idea what could be going on here.

So apart from potentially making the serialization better with @SomeTroglodyte 's help (I haven't had the time to look at this yet), I think I addressed everything else.

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapTileUtil.kt
Gualdimar: topX and topY are not used anywhere except for computing width and height inside function itself. Using a topX/Y as returning arguments, in my opinion, only adds one additional math operation to calculate bottomX/Y, which are used in shifting, at least. So what`s the point?
And Unciv already has 3 different coord layers with different anchor points: hex positioning, hex drawing, stage itself and center, top left, bottom left(maybe another order, I don't remember exactly). 
Anyway, I am not insisting on anything, but only sharing my observations and suggestions. The decision is up to you and @yairm210

WhoIsJohannes: If we return a point and two lengths (width and height), I agree with you. It would be better to return Point(bottomX, bottomY), width & height, but IMO in the context of a rectangle one expects that you can do otherCornerPoint(xCoord+width, yCoord+height) and the only way to achieve this would be to have negative widths and heights, which I personally find odd. It seems this is supported, but the behavior isn't even documented (https://bugs.openjdk.org/browse/JDK-6800505).

We could introduce a custom wrapper class and not use the rectangle notion. Do you think that would be better or do you have other ideas how to solve this?

WhoIsJohannes:
> .. see here : [WhoIsJohannes#1](https://github.com/WhoIsJohannes/Unciv/pull/1)

Thanks, I now looked through this. Some questions:

- Why did you introduce an interface ITileHistoryState? Is that necessary for the serialization to work?
- Extending LinkedHashMap in TileHistory feels to me like it exposes too much implementation detail. It also makes it easy to use the class in the wrong way. I felt it was nice before that you were forced to go through the setters and getters. Would the serialization still work if we keep the map as a field of TileHistory? Do you feel strongly about this change?
-  How do you want to proceed with this? Do you want me to work your changes into my pull request or do you want to build on top of my code later?

Thanks!

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapTileUtil.kt
Gualdimar: I'm not sure I understand. bottomX and bottomY are points on the same plane but in a different location. And we can calculate topX and topY with the same formula (bottomX+width, bottomY+height), if we ever need them. 
Am I missing something?

WhoIsJohannes:
> I'd say a case for the debugger and deep digging into what-is. The "secret" debug switch to turn on most of Gdx Scene2D debug decorations from the Options Popup might also help to see what's what. Did it like that recently for that Diplomacy/Trade/Repeat issue... https://github.com/yairm210/Unciv/pull/8808 - that was half an hour of inspection and Alt-F8 to reliably find out and understand the cause.

Sorry, I can't find it. What switch in the options are you referring to? Also how did you use the debugger to help with this?

WhoIsJohannes:
After already hitting a merge conflict with the Minimap I'm getting increasingly worried about those :D. It's otherwise not urgent though. I'll wait for you to finish your prototype and then let's agree on whether to go that route or stick with this, it's probably worth it since it's hard to change later. Thanks for spending time on this! No worries if you're busy on the weekend - I'm too and usually can only squeeze in an hour or so in the evening anyways.

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapTileUtil.kt
Gualdimar:
```
Rectangle2D.Float(float x, float y, float w, float h)
Constructs and initializes a Rectangle2D from the specified float coordinates.
```

Negative will be its coordinates` x = bottomX` and `y = bottomY`, width will remain unchanged `w = topX-bottomX`, as well as height ` h = topY-bottomY`

e.g we have these parameters:
```
topX = 20
topY = 30
botomX =  -10
botomY= -20
width = 20 - -10 = 30
heigth = 30 - -20 = 50
```
You suggest:
`return (20, 30, 30, 50)`
I suggest:
`return (-10, -20, 30, 50)`

core/src/com/unciv/ui/screens/victoryscreen/VictoryScreen.kt
```diff
@@ -187,6 +197,43 @@ class VictoryScreen(val worldScreen: WorldScreen) : PickerScreen() {
...
+            object : Timer.Task() {
+                override fun run() {
+                    updateReplayTable(yearLabel, replayMap, nextTurn++)
+                }
+            }, 0.0f,
+            // A game of 600 rounds will take one minute.
+            0.1f,
+            // End at the last turn.
+            finalTurn
+        )
```
Gualdimar, line 219: The timer does not stop when you change tabs or close the victory screen

Gualdimar:
> If I remove it, it doesn't happen.

Something to think about. The following two constructs(which are basicle the same) do not cause the bug, but do apply padding:
```
replayTable.add(yearLabel).apply { padBottom(5f) }.row()
replayTable.add(yearLabel).padBottom(5f).row()
```

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapTileUtil.kt
WhoIsJohannes: Okay, I'm not going to die on this hill. Done! :)

core/src/com/unciv/ui/screens/victoryscreen/VictoryScreen.kt
WhoIsJohannes: I added code to cancel the Timer.Task if the Replay is opened again and also when the VictoryScreen is closed. I didn't find an elegant solution though to do it when the tab is switched. I tried to add onClick for the whole tab bar, but unfortunately that gets executed after the ReplayMap is initialized unfortunately, so the task will be canceled immediately. Any idea how to do it nicer than add it to every other onClick handler? I think it might be acceptable to have it live on in the scope of the VictoryScreen once since it is also finite. WDYT?

Gualdimar: You can call `replayTimer?.cancel()` in each `setSomeTable()` fun, where `contentsTable.clear()` is called

yairm210:
I'm going to merge this now, as it's definitely "good enough" and an improvement on the current state
Future improvements can come in future PRs, we don't need to get everything 100% on the first run

WhoIsJohannes:
Maybe let's wait for whether we can improve on the serialization until
Monday.

Il sab 11 mar 2023, 19:08 Yair Morgenstern ***@***.***> ha
scritto:

yairm210:
OK, next version then, hopefully

WhoIsJohannes:
Also now that I think about it, we should probably also somehow handle
saves that start recording halfway through the game. The replay would look
weird.

Il sab 11 mar 2023, 19:13 Yair Morgenstern ***@***.***> ha
scritto:

yairm210:
Sure, as far as I'm concerned this can go in when you tell me you think it's good :)

WhoIsJohannes:
Thanks! Are you aware of anything in flight that touches similar files?

Il sab 11 mar 2023, 19:18 Yair Morgenstern ***@***.***> ha
scritto:

yairm210:
Nothing that I know of

core/src/com/unciv/ui/screens/victoryscreen/VictoryScreen.kt
WhoIsJohannes: Done

WhoIsJohannes:
@SomeTroglodyte I wasn't notified and now I'm really glad I waited. This is so awesome, thanks a ton! I already integrated your code for the serialization with the difference that now we don't need to hold two maps anymore. We can just use a TreeMap and that's it or am I missing something?

Also funny we thought about the halway thing independently, but I think you solved it much better than I would have. I would have just errored out, but it's actually better to just start at the turn where we started recording. That is great. I'll go ahead and work through your diff to integrate it into this pull request and then I think this is good to go.

WhoIsJohannes:
I'm excited @SomeTroglodyte . Your code is super elegant - works like a charm with two changes:

- The new field in GameInfo needs to be handled in clone, otherwise it will be set anew every turn.
- We need to adapt how many times the timer task runs.

Here is a game that I started recording halfway through:

https://pastebin.com/p6S9rcbM

Would appreciate if you could all do one final round of testing. From my end I think everything is fine now.

WhoIsJohannes:
A TreeMap is just an extension of a Hash map. Its API is just more
powerful, but the serialized form of both would be the same.

Thanks again for the solution to started games. You're exactly right.
Finding the right places to hook things in would have taken me forever.

Il dom 12 mar 2023, 00:42 SomeTroglodyte ***@***.***> ha
scritto:

WhoIsJohannes:
> Repeated UncivGame.Current.gameInfo? in VictoryScreen - you sure that can differ from what the class caches locally as this.gameInfo!!?

Missed it. Thanks for pointing that out. Done!

> Pastebin - no longer available.

https://drive.google.com/file/d/11tpZy7vmkn6PtzPCfDGxqxSEyND6czxo/view?usp=sharing

Let's try with drive, sorry about that.

WhoIsJohannes:
One more question @SomeTroglodyte : If we decide later to add new fields to the history for whatever reason (e.g. let's say we also want to show when a wonder was built or when war was declared etc.), how difficult do you think it'll be to update the custom serialization in a backwards-compatible way? Is there anything we could do already to prepare for that? I was thinking about maybe wrapping the existing fields a bit better or do you think it's not necessary?

WhoIsJohannes:
Nice! Are you going to add the slider after this is integrated? That'd be
super cool!

Il dom 12 mar 2023, 14:43 SomeTroglodyte ***@***.***> ha
scritto:

WhoIsJohannes:
@yairm210 Would appreciate if you could merge this now. I think everything is clear now and other stuff can be follow-up.

@SomeTroglodyte thanks for sharing

PR #8847, opened 2023-03-08
Update french translations.

HydrolienF:


Ouaz:
`Auto (en fonction des capteurs)`

I did not test, but I think it's a bit too long for the orientation option scroll list, and may break the line or extend the list width too much.

`Auto (capteurs)` is enough I think.

```
Paysage (fixe)
Portrait (fixe)
Auto (capteurs)
```

Otherwise, all's good.

HydrolienF:
Ok, I have change it.

PR #8853, opened 2023-03-09
Update German translation

Mape6:


Mape6:
> There is a tiny error in line 1476: "No yields = Keine Eträge"
> 
> It should be "No yields = Keine Erträge"

Thanks for the review. Fixed it.

Issue #8864, opened 2023-03-11
Future Tech is researched too fast


SomeTroglodyte:
... because `techsInProgress` is never cleared for it - or anything. Simple enough. Last progress for all researched techs is also needlessy but harmlessly kept in saves.

### Steps to repro:
* The attached save is a construct with `overflowScience:6365`, the number the tech tree displays as required for the next Future Tech.
* Hit next turn
* Check tech button turns to research, and numbers in tech tree
* The required beakers for the next Future Tech has not increased.
* The "account" has kept (overflowScience + income)
* Each next turn will immediately complete another Future Tech.

### Questions:
* Shoudn't completing a Future Tech increase the cost for the next one? Did Civ5 increase here? I think Civ4 did.
* More moddability? Would implementing increased-cost formulae including allowing separate counters for multiple repeatable techs be desirable?

<details><summary>Saved game</summary>

```
[...]

PR #8866, opened 2023-03-11
Future tech moddable

SomeTroglodyte:
In case the answer to the second question in #8864 is yes...

If it's no (which I'd fully understand) then the ["fix"](https://github.com/yairm210/Unciv/commit/27833dfb0a2e7ae14959e9a29a2d3e8fdfcdd7fa), ["lint"](https://github.com/yairm210/Unciv/commit/0455e5891e3663d54e11d141b9b45e0c47d8e276) and ["kotlinify"](https://github.com/yairm210/Unciv/commit/66a982a9c68b1d1cd5494dfba9255cb300b7ac81) commits might still be interesting.

The [Move "who knows" decoration](https://github.com/yairm210/Unciv/commit/add3db26a147e8c7017707c5ac82e1a54276aa27)  commit looks like <details><summary>this:</summary>

image
</details>
... and has the side effect Future Tech gets a Star icon less (compensating the additional one for the score unique), and the text won't appear on the picker screen's bottom pane.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Techs.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Techs.json
@@ -648,8 +648,9 @@
 			{
 				"name": "Future Tech",
 				"row": 5,
-				"prerequisites": ["Globalization","Particle Physics", "Nuclear Fusion", "Nanotechnology", "Stealth"],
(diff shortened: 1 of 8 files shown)


yairm210:
This looks like a lot of added complexity that no one seems to be asking for
Sorry, but no

yairm210:
Fixed and linting are always welcome

PR #8880, opened 2023-03-13
Change serializer to make migrateBarbarianCamps unnecessary

SomeTroglodyte:
I horsed around with NonStringKeyMapSerializer and had a version that still did the job but was _much_ simpler - but it relied on `out` variance and needed one `@UnsafeVariance` - not pretty, so I ditched it.

But one aspect remained: that 'old' code tried to deal with Gdx Json creating invalid objects in memory by - creating the same invalid objects in memory. Just so the conversion String to original key type can be done later... That irked me - why replicate buggy code one on one just so one can graft a check and alternate path onto it?

So - see diff.
Effectively we're exchanging one `@Suppress("UNCHECKED_CAST")` for another - with the difference that before, we _knew_ the cast was illegal, while now we have _checked_ the cast is valid.

Note this _will_ still read "old" saves that would have run the migrateBarbarianCamps loop just fine.

### DIFF
--- a/core/src/com/unciv/json/NonStringKeyMapSerializer.kt
+++ b/core/src/com/unciv/json/NonStringKeyMapSerializer.kt
@@ -1,5 +1,6 @@
 package com.unciv.json
 
+import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.utils.Json
 import com.badlogic.gdx.utils.Json.Serializer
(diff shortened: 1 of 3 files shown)


yairm210:
Alternatively we could get rid of support for the old way.
It's been long enough, and I'm always suspicious of changing the serializer, it's almost always more trouble than it's worth

PR #8895, opened 2023-03-14

pdt012:


pdt012:
The current translation ("个人ID") actually means "user ID", not "gameID".

PR #8903, opened 2023-03-14
World wrap scrolling fix

Gualdimar:
Fixes the scrolling teleportation problem from PR #8795
Works with panning, keyboard scrolling, scrollTo() and in the map editor.


Gualdimar:
> there may have been a major and very satisfying "Eureka" behind it

Something like this happened when my eye accidentally caught on to the fact that scrollX is actually derived from a setter and a getter)

PR #8904, opened 2023-03-14
Barbary galleys are little less suicidal

Gualdimar:
Not sure if this is the right solution, but it fixes the case from my last save in #8551.
In my case `BattleDamage.calculateDamageToAttacker()` always returns 99, unit.getDamageFromTerrain() - 0 and `unit.health` is 100. But the galley actually takes 100 damage and dies.
This PR fixes the case, but now the galley come comes close to the city, where I can bombard it down to 7 HP on my turn. It's not suicide, but it doesn't do any harm to me, so what's the point?


yairm210:
Calculate damage to attacker contains some randomness, so it makes sense that it will be off by a bit, and this fix won't actually solve the problem
If you want to make absolutely sure they don't suicide you'll need to tell calculatedamage to do max damage (1)
This doesn't bother me that much though, suicidal barbarians makes sense thematically

Issue #7479, opened 2022-07-18
Copyright of texts


GrimPixel:
The texts of every introduction, “May the blessings of heaven be upon you...” are protected by copyright. I think they are better removed. No one read that anyway.

ArchDuque-Pancake:
I actually do read them :(

letstalkaboutdune:
I volunteer to help with this. They don't need to be an essay, just 2-3 sentences at most. Perhaps most important is an equivalent for the part repeated for every civ: 

_"Can you build a civilization that will stand the test of time?"_

Maybe something like: 

_"Will you forge an empire that lasts through the ages of history?"_

alexban011:
Can they copyright such texts ? (Or to be more specific: Did they actually claim ownership over those texts?)

yairm210:
Any text they write is automatically copyrighted as part of the larger work. Whether they claim ownership might be interesting from a "will they do something" perspective (answer: no, they've been cool with worse things), but from a "can they do something" perspective the answer is "yes", and we want to be clean :)

letstalkaboutdune:
I think it would be pretty manageable to get new text in there, especially if it follows a template. Something like:

"Oh <adjective> <name of Civ leader>, you <famous accomplishment during their reign>. But <describe the fall of their empire>. Now you have a chance to re-write history. Will you forge an empire that lasts through the ages?"

saud2410:
I think not every civ would have "call-ups-downs-ask" intro. There can be call-downs-ups-ask pattern too. tho it's a matter of sequence.

PR #8905, opened 2023-03-14
Update Nations.json

Smashfanful:


yairm210:
These texts are copyrighted, as we have stated in #7479, and therefore were removed
If you want to add *new texts* we're more than willing, but they need to be ones you came up with yourself

Issue #8877, opened 2023-03-12
Show whether city has a garrison in city overview


WhoIsJohannes:
Smaller improvement I'd like to work on, just wanted to get quick feedback. I haven't found any good way to see which cities have a garrison. This is relevant for some culture improvements and technologies (?). Would it be okay to add another column for this in the city table? Maybe it could show the icon of the unit that is garrisoned there.

WhoIsJohannes:
Ah, damn it. I just finished this and only realized when trying to create the pull request that you had already done this.

https://github.com/yairm210/Unciv/pull/8945/commits/9450f9860410de1ce2a378c0d71efe28aaf5293a

Anyways, I think there's a bug in your code. You should use this:

private fun City.getGarrison(): MapUnit? {
        return this.getCenterTile().getUnits()
            .firstOrNull() { it.civ == this.civ && it.canGarrison() }
    }

Otherwise there will also be planes showing up I guess.

WhoIsJohannes:
Do you mind doing it if you already identified 4 places? I'd only know the
one in the city overview table?

Is there anything else you're currently working on, so we don't do the same
thing twice again? 😅

Il ven 17 mar 2023, 23:03 SomeTroglodyte ***@***.***> ha
scritto:

PR #8906, opened 2023-03-15
Show garrison in City overview

SomeTroglodyte:
#8877 sounded like a good idea, so I tried out how it may look - and thought "why not".
I suggest reading commits separately.


yairm210:
Pics?

PR #8919, opened 2023-03-16
Improved Minimap colors

Caballero-Arepa:
This PR is intended to improve the minimap colors, and consecuently the minimal tileset, as the current one is very very palid and dead.

New

I based this colors on cartography tendencies, to make it look more like a proper biome map, which is what it is by being the minimap.

Overall improves readability of the minimap, and I see no downsides to improving the colors.
   We of course can change around some colors, or saturation etc, it shall be discussed, but for my taste these are good.

Comparacion of the new (left) and old (right)
Comparisong

For now, I have a few possible changes for some tiles:
- For Coast: If you want it Darker: 87, 139, 199 or Lighter: 59, 137, 227
- For Snow, a less bright one: 229, 229, 255

This is the "V2" that I showed on the [discord](https://discord.com/channels/586194543280390151/1084699936353955900).


AdityaMH:
Can you try this Inca in desert tile or India in grassland tile?

yairm210:
Following the discussion there I think this is a good update

Gualdimar:
> Can you try this Inca in desert tile or India in grassland tile?

image

PR #8920, opened 2023-03-16
Key bindings Step 4 - translation generation

SomeTroglodyte:
I was tempted to make some of this prettier, but refrained.

...
* TranslationFileWriter outputs an entry for every key name known to Gdx except the few we actively excluded
* It won't count these for completion percentages or decorate with "requires"
* For my convenience - but maybe you like it too - one can skip mods from tr generation by holding left-shift (can take a lot of time)
* A stacktrace that should have gone through Timo's helpers
* German included - probable conflict with @Mape6 's stuff, no biggie, we'll fix as needed

Note: On my box a run takes about 5 to 10 times as long as what is logged on the console as "Translation writer took 23ms for Tutorials" and so on - yup, curious where that goes, but haven't yet looked


Gualdimar:
Why such a complication? Why can't we just stick to the most common qwerty layout (which is printed on every keyboard anyway)?

Gualdimar:
@SomeTroglodyte, you gave examples of layouts, not real keyboards. Real keyboards usually have a qwerty layout and a local one
de:
image
ua:
image
ch:
image

Gualdimar:
Serial keyboards have a fixed number(which is based on the qwerty layout) of buttons and if all localized characters do not fit, they are typed by combinations (usually right alt+something).
For example, in your picture of the ua layout, the letter next to the left shift is never a separate key on real keyboards. It is typed with the combination rightAlt+U 

> That topmost one looks interesting - model and maker?

the first one I came across on Google by searching for "de keyboard"

Gualdimar:
> And before you ask why then capture keyDown and not keyTyped

I wasn't going to ask that)
I'm just asking why we need to localize button names if almost all keyboards have Latin letters printed next to the local ones?
I`m sticking to the KISS principle.

Can you give examples of games that use localized characters rather than Latin ones in the keyboard bindings?

In addition, the tutorial says that you can only assign direct buttons and ctrl+button combinations, but the translation list also contains the @ symbol, which is printed with the shift+button combination. So, where is the truth?

yairm210:
I'm woth Gual on this one, since
- the entire web and all computing are based on English characters
- Microsoft Word and Google docs, general use products used by hundreds of millions, don't have different shortcut keys for different languages
- No one playing Uncuv had ever expressed the need for such things

So I'm calling this a YAGNI unless proven otherwise

Gualdimar:
> We can still roll back... Was fun coding it either way.

Personally, I prefer to use the mouse, but it's nice to be able to customize the controls for yourself, right?
But I think it should be something simple. In the original civ5, players edit the xml file tol alter bindings and are happy with it😁

PR #8925, opened 2023-03-16
Trade ending notification

Gualdimar:
Added two new notifications:
[tradeOffer] from [otherCivName] will end in [3] turns
[tradeOffer] from [otherCivName] will end on the next turn

\+ We have lost alliance with [nation]. will not be displayed if a city-stae is captured


Gualdimar:
> Minor improvement: "will end next turn" to minimize text, it's a long notification anyway

Done

PR #8927, opened 2023-03-16
Fixed the positioning of popups after changing the screen orientation on Android

### LINKED ISSUE #8443: Game doesn't align popups correctly after rotating the screen
**Platform**
Android 13

**Version**
Google Play, version 4.3.12

**Describe the bug**
If the game has been rotated to Portrait mode and then to landscape, popups like Exit game are not positioned correctly, sometimes even completely outside the screen.

**To Reproduce**
1. Rotate to Portrait mode
2. Rotate back to Landscape mode
3. Press Back button
4. The popup is in the left upper corner or even out of the screen.

**Expected behavior**
Popups are in the middle of the screen

**Additional context**
1. Maybe it has already been reported, then sorry
2. Happens on both small and large resolutions
3. Happens also the other way around, when turning from landscape to portrait. Then the popups are in the bottom right corner.

Gualdimar:
Resolves #8443
I'm not very good with android stuff, but no one has paid attention to this problem for a long time, so here we are...
After changing the orientation, `UncivStage()` is reinitialized and has the correct `lastKnownVisibleArea`, but `onGlobalLayout()` call messes up the `lastKnownVisibleArea` because it operates on the old `stage` dimensions from the previous screen orientation.

### DIFF
--- a/android/src/com/unciv/app/AndroidGame.kt
+++ b/android/src/com/unciv/app/AndroidGame.kt
@@ -1,5 +1,6 @@
 package com.unciv.app
 
+import android.app.Activity
 import android.content.Intent
 import android.graphics.Rect
 import android.net.Uri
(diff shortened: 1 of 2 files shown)


Gualdimar:
And why do we need to `addScreenObscuredListener()`?
I tried to comment on it and everything is positioned where it should be...

PR #8956, opened 2023-03-18
Add city size (population) to TradeOffer.kt

WhoIsJohannes:
https://drive.google.com/file/d/1hCLStCXklrd-dALtoDZAuNBeWgwkt64h/view?usp=sharing

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -291,7 +291,6 @@ Declare war on [nation] =
 Luxury resources = 
 Strategic resources = 
 Owned by you: [amountOwned] = 
-[cityName] ([population]) = 
 Non-existent city = 
 
 # Nation picker
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -291,6 +291,7 @@ Declare war on [nation] =
 Luxury resources = 
 Strategic resources = 
 Owned by you: [amountOwned] = 
+[cityName] ([population]) = 
 Non-existent city = 
 
 # Nation picker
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/logic/trade/TradeOffer.kt
```diff
@@ -38,7 +38,11 @@ data class TradeOffer(val name: String, val type: TradeType, var amount: Int = 1
         var offerText = when(type){
             TradeType.WarDeclaration -> "Declare war on [$name]"
             TradeType.Introduction -> "Introduction to [$name]"
-            TradeType.City -> UncivGame.Current.gameInfo!!.getCities().firstOrNull{ it.id == name }?.name ?: "Non-existent city"
+            TradeType.City -> {
+                val city =
+                        UncivGame.Current.gameInfo!!.getCities().firstOrNull { it.id == name }
+                if (city != null) city.name + " (" + city.population.population + ")" else "Non-existent city"
```
yairm210, line 44: You need to either.tr() city name or wrap in curly brackets because it's not going to be translated like this

WhoIsJohannes: Done

WhoIsJohannes:
Thanks! Can you check again? I don't think a new template is overkill.

Am Sa., 18. März 2023 um 22:01 Uhr schrieb SomeTroglodyte <
***@***.***>:

android/assets/jsons/translations/template.properties
```diff
@@ -291,6 +291,7 @@ Declare war on [nation] =
 Luxury resources = 
 Strategic resources = 
 Owned by you: [amountOwned] = 
+[cityName] ([population]) =
```
Gualdimar, line 294: There is nothing to translate in this line. There is no point to add this to the templates.

WhoIsJohannes: Hm... The idea was to handle RTL languages and maybe languages that would put the parentheses on the other side for other reasons (or not use parentheses maybe?)

yairm210: I'm in favor of simpler,  we can always complicate in the future if necessary

WhoIsJohannes:
Okay, how about now?

PR #8958, opened 2023-03-18
Map scroll speed

SomeTroglodyte:
Next-most useful thing from #8795 redone:
- Another slider controls what before was the hard "6f" in the WASD listener
- Map mouse panning in map editor
- Map mouse panning tries to compensate for framerate and to be same speed as WASD
- I didn't change the sensitivity which is different for x and y - I doubt that's useful, but still not touched
- Ergo a little thing from the old PR is missing - mouse edge panning accelerating the nearer one gets to the edge. Likely nobody would notice anyway.
- I took care not to invert any signs (as in numeric value sign, not road sign) in relation to the old code
- Researched language - panning is the proper term used in any GIS or maps api out there. I _almost_ relabeled that checkbox, but - your call, should we?
- @Mape6  promptly misunderstood the option, so I fixed that and added a comment for other xlators

### DIFF
--- a/android/assets/jsons/translations/German.properties
+++ b/android/assets/jsons/translations/German.properties
@@ -686,7 +686,8 @@ Landscape (fixed) = Quer (fixiert)
 Portrait (fixed) = Hochkant (fixiert)
 Auto (sensor adjusted) = Auto (durch Sensor justiert)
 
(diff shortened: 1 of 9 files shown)


yairm210:
How does this mesh with scrolling around world wrap?

PR #8962, opened 2023-03-19
Generic Widget/Provider framework for sortable grids

SomeTroglodyte:
... and applied as City overview reorg.

Functionally no change, though it should improve performance marginally as less objects keep being collected and rebuilt.
That said, saving even more GC stress would now be easier to do - presentation is more decoupled from content-specific knowledge, it's easier to see where there's potential.

It's a bit too specific about the header containing icons, but I'll get there once -if- I move the units overview to this. The vertical resources overview is also a candidate, and I'm sure there are zero to three candidates I haven't thought about.


core/src/com/unciv/ui/components/ISortableGridContentProvider.kt
```diff
@@ -0,0 +1,65 @@
...
+ * - [isVisible] can hide a column
+ * - [align], [fillX], [expandX], [equalizeHeight] control geometry
+ * - [getComparator] or [getEntryValue] control sorting, [defaultDescending] the initial order
+ * - [headerTip] and [getHeaderIcon] define how the header row looks
+ * - [getEntryValue] or [getEntryActor] define what the cells display
+ * - [getEntryValue] or [getTotalsActor] define what the totals row displays
+ * @param IT The item type - what defines the row
+ * @param ST The parent Screen Type - passed to [getEntryActor] for potential OnClick calls
+ */
+interface ISortableGridContentProvider<IT, ST> {
```
yairm210, line 17: If this is supposed to be a screen type then we should harden the interface to only allow screen types

core/src/com/unciv/ui/components/SortableGrid.kt
```diff
@@ -0,0 +1,212 @@
...
+                icon.center(this)
+                if (column.headerTip.isNotEmpty())
+                    icon.addTooltip(column.headerTip, 18f, tipAlign = Align.center)
+                this.addActor(icon)
+            }
+        }
+
+        /** Show or remove the sort symbol.
+         * @param showSort null removes the symbol, `false` shows an up arrow, `true` a down arrow */
+        fun setSortState(showSort: Boolean?) {
```
yairm210, line 200: Instead of a fake trinary sort, we should use an enum with 3 values

core/src/com/unciv/ui/components/SortableGrid.kt
```diff
@@ -0,0 +1,212 @@
...
+    /** Size for header icons - if you set this too low, there is a chance that the tables will be misaligned */
+    private val iconSize: Float = 50f,
+    /** vertical padding for all Cells */
+    paddingVert: Float = 5f,
+    /** horizontal padding for all Cells */
+    paddingHorz: Float = 8f,
+    /** When `true`, the header row isn't part of the widget but delivered through [getHeader] */
+    private val separateHeader: Boolean = false,
+    /** Called after every update - during init and re-sort */
+    private val updateCallback: ((header: Table, details: Table, totals: Table) -> Unit)? = null
```
yairm210, line 41: You usually over-generalize things, so I'll say ahead of time that any of these that is currently *not* an input parameter - shouldn't be. We'll add them if needed.

core/src/com/unciv/ui/screens/overviewscreen/CityOverviewTab.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    private val grid = SortableGrid(
+        columns = CityOverviewTabColumn.values().asIterable(),
+        data = viewingPlayer.cities,
+        parentScreen = overviewScreen,
+        sortState = persistableData,
+        iconSize = 50f,  //if you set this too low, there is a chance that the tables will be misaligned
+        paddingVert = 5f,
+        paddingHorz = 8f,
+        separateHeader = false
```
yairm210, line 30: Same as above - remove anything that isn't currently passed as input

PR #8973, opened 2023-03-20
Desktop starting size fix

### LINKED ISSUE #8931: Desktop: Saved window size has almost zero usefulness
**Platform**
Both a Linux, in my case Mint 20.3 cinnamon, and a Windowz, in my case 10 21H2

**Version**
Pretty fresh from master, 828 = 4.5.5

**Steps to reproduce**
- Set a desktop Unciv (both OS) to Windowed mode
- Resize to something recognizable, smaller than the screen, but different from 640x480
- Leave the game cleanly

- Restart the game on Linux
    -> will be windowed+maximized filling the screen (excluding taskbar), saved **size is lost** (windowmanager "restore" will not restore it)
    Understandable: `config.setMaximized(true)` in DesktopLauncher.main and method used nowhere else
- Restart it on Windoz 10
    -> will have the size you left it at, but revert to 640x480 the moment you move it. Yes, **size _jumps_** the moment you drag the title bar.
    `(Gdx.graphics as Lwjgl3Graphics).window.config.windowWidth`/windowHeight contains that 640x480 value before and after the setWindowedMode call, but patching it in-memory just before the first drag has no effect. However, the same steps on Linux show the same values there.

**Quick and dirty**
- Comment out the setMaximized call and after loading the settings, call setWindowedMode(settings.windowState.width, settings.windowState.height) _on the config_ we're preparing there. Yes there is a difference between Graphics.setWindowedMode and Lwjgl3ApplicationConfiguration.setWindowedMode.
- Repeat steps above
- Both seem to react identically and more or less fine. The stored size amounts to something on Linux and no size jumps on Doze.
- The values passed to config.setWindowedMode in launcher.main will turn up in that debug watch formula above
- So does that mean Graphics.setWindowedMode fails to pass the new values down to the Lwjgl3Window but bypasses it and _does_ pass the size to the GLFW window the Lwjgl3Window is proxying...??? :see_no_evil: 

**Suggestion**
Ditch windowState _or_ ditch setMaximized(true) (_or_ read the complete solution over in closed PR's :speak_no_evil: )

SomeTroglodyte:
Closes #8931

First commit is the band-aid already talked about.
Second commit fixes window _still_  jumping back to the "starting" size in situations like (but not limited to) switching windowed/borderless, and I couldn't do so without doing that TODO in there... @vegeta1k95 - I'm sure you could find a better way.
Hard to see due to rewrite as Enum, but that's just two new details in the DesktopDisplay file: Bubbling the knowledge whether saving size makes sense up to the UI, and doing the same setWindowSize _to the config_ as the launcher does through reflection.

TLDR - a kludge-o-rama band-aid, but works better than before. Tested so far on 3 Platforms - doze,ix, AVD (12L).

### DIFF
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -1,6 +1,6 @@
 package com.unciv.models.metadata
 
-import com.badlogic.gdx.Application
+import com.badlogic.gdx.Application.ApplicationType
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.utils.Base64Coder
 import com.unciv.Constants
@@ -11,6 +11,7 @@ import com.unciv.ui.components.FontFamilyData
 import com.unciv.ui.components.Fonts
 import com.unciv.ui.components.KeyboardBindings
 import com.unciv.ui.screens.overviewscreen.EmpireOverviewCategories
+import com.unciv.utils.Display
 import com.unciv.utils.ScreenOrientation
 import java.text.Collator
 import java.time.Duration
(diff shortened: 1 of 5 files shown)


yairm210:
I'm not entirely sure about this since I don't feel I understand the problem space well enough, but at this point solving the problem takes priority. We can find a better way later.

PR #8986, opened 2023-03-21
Key bindings - unit actions, world screen and grouped UI

SomeTroglodyte:
#### New Features
* Unit Actions, all of the ones in the UnitActionType enum
* World screen all existing, two that were removed but still present in comments only (Ctrl-R, Ctrl-Y), plus two for symmetry (<sub><sup>yes  bindings with no default key work properly - you can now toggle all of the little green aliens glued to the minimap with the kbd if you whish and go and give them a binding</sup></sub>)
* UI grouped by Expanders, all sorted by translated labels
* Complete german i18n to prove to myself it's all there - ~~one exception I keep forgetting but will remember to template one of these days.~~

#### Screenies
<details><summary>Unit actions (English)</summary>

image
</details>
<details><summary>Worldscreen (German)</summary>

image
</details>

#### Invitation for discussion
* Every single unit action has its own binding - easier to code. Alternative: One binding per what has the same key - e.g. all great person special actions together. Not so clear for Paradrop and Pillage. The structures allow this easily too, but just regex'ing the enum over and then checking whether the labels would hit the existing translations - well, it _was_ easier this way.
* It seems the ability to explicitly remove a binding that has a default key - so no key would trigger it - might be useful? From code that would work but there's no UI and I have no easy idea - another trick-shot like the double-click? Nah. Right-click pop-in menu? Maybe, could be a page of code or three.

#### Coming Soon
* Conflict detection - already works (per category, active for world screen only atm - avoids the issue how unit actions are perfectly fine with dupes)
* Research into simpler "API" - instead of just adding overloads to Paul's actor-bound KeyShortcut extension functions, _maybe_ there is something much simpler
* Get the tooltips in order so all reflect the actual bound key (see Empire Overview) - if possible, semiautomatic addToolTip removing many explicit calls

#### Notes
* **Sorry** for the catastrophic-looking commit history - local merge was a mess, it offered code that was long gone from _both_ local branch and the master to merge in - repeatedly. I'm praying. Anyway, this messed up the original plan to present this as three separate PR's.
* I stumbled on "please help us sort these" in templates and processed it, on the way sorting the 'advanced' section as it is on screen plus moving some lines to the tabs they actually are on (all still options popup).


Gualdimar:
> One binding per what has the same key

I think it will be better this way. One binding for "Construct something", one for "Special ability", one for "Fortify/Sleep", and so on.
Or there can be a checkbox like "Group actions"

> It seems the ability to explicitly remove a binding that has a default key

Maybe Esc or Backspace while hovering?

PR #8992, opened 2023-03-21
Fix MiscLayer not respecting fog of war for spectator

### LINKED ISSUE #8836: TileLayerMisc visible after FoW toggle
In spectator or debug modes, if you turn on the fog of war, the misc layer remains visible until world screen reload

SomeTroglodyte:
_Please_ evaluate carefully
(Fix #8836)

* repeated addActor(yields) should have been a resource leak
* yieldsInitialized always in sync with isVisible not helping any
* Cache improvement actor
* Call reset even when all layers will be set invisible - see 'fixme' culling comment - possibly related?
* Misc layer's reset was a bit incomplete
* listOf malloc then off to GC every update()? Nah.

Remember - I was not too familiar with that code, so feel free to laugh or ask for motivation where I forgot to mention it above.

Passes following tests: @Gualdimar 's observation, Pedia, City Screen - not much more tested. Unit tests of course, but are there even any for visuals? No.


yairm210:
It's very hard for me to parse what's going on here with so many changes at once

As far as I'm concerned this needs to pass the following:
- Spectator with and without fog, also spectator between turns
- Non-spectator civ with visible, unexplored, and explored-non-visible tiles
- Map editor

Once it passes all of those we can merge it

yairm210:
- viewingCiv null is for map editor, civilopedia and main screen
- isForceVisible is a *debug tool* to allow you to see tiles that in-game you shoudn't - it's an override
- showResourcesAndImprovements is a user setting

In terms of hierarchy:
- showResourcesAndImprovements, as a user setting, overrides everything - if false do not show
- isForceVisible, as a debug tool, is next - if true, show. 
- You could say that isForceVisible and viewingCiv==null serve the same function of 'show me everything' and that is true, you could (in the tile) have a single val that combines them (showEverything = isForceVisible || viewingCiv==null) - but they're still both necessary

yairm210:
IIRC spectators also see everything using viewingCiv==null if fog of war is off, and if that's *not* the case - then it should be.

PR #8998, opened 2023-03-22
Link fastlane images to the more current ones under extraImages

SomeTroglodyte:
This would be optimal to keep the fastlane images in sync with the ones under extraImages without needing a manual copy PR every time.

## This is an experiment
I created this on a Windows workspace (normal mklink not bash ln), pushed and then pulled onto a Linux box. Symlinks are reproduced as symlinks and their paths are even backlash/normal slash translated. Nice!

**but**

That could mean every windows user who wants to change these links will need to
- turn on `git config core.symlink true`
- turn on developer mode in their 'doze settings
.. or else they will clone the images themselves, which is fine as long as they don't include them in commits.

Also, git**hub** seems to not know they're symlinks, but git does.

Opinions? Can somebody with a fork confirm pulling this branch results in proper symlinks?


yairm210:
I'm in general very wary of adding symlinks since they are, in the end, platform-dependant. 
And as fashionable as it is to hate on the Doge there are a lot of devs - and good ones! - that work with it
So given the choice of 'sync between 2 images that are already synced by automation, and make a very niche part more confusing for some devs' - both risk and reward are very low, but I'll stick with the current state
It's a good idea though! And it's not your fault - hopefully within a decade Windows will work on the Linux kernel and such problems will be behind us...

yairm210:
There is automation for only the worldScreen, right, forgot about the other screens
As for the docker containers, the hard part is not taking the picture, it's getting the game to the state where you would want to take a picture.
Since they're VNC-ready, I briefly entertained the idea of using VNC commands to set mouse x/y and click... or use keyboard shortcuts... nothing came of it

PR #9011, opened 2023-03-23
More Spectator/Barbarians fixes related to income

SomeTroglodyte:
OK, here my branch where I analyzed that "Spectator" game. Inspect and take apart at your leisure.

- Spectator gets relevant notifications -> better let the normal code for notif log history turnaround run before the early exit
- Defated Civs still _do_ run endTurn. I'll leave the startTurn - obviously should go there too, but that's an ass-umption, I didn't test.
- Barbarians now get 0 income - no --gold, no golden ages. Likely un-noticeable unless you read the saved json, so more defense-in-depth kind of thing.
- That CityStateFunctions test - no-brainer. !isCityState includes major, barbs, and speccie... Unnecessary.
- The spectator doesn't need to see his own stats in overview - like +N happiness for seeing all natural wonders
- The razing-first: Before had two filters, both materializing as list 'cuz source not a seq, then yieldall into a seq builder, then tolist - one simple sort is enough, and does only the one copy we need.
- The disband for debt code had a few flaws, and one 'I'm not sure about' - open to discussion.
    - The for loop never took into account any change due to disband - `1 until N` is a range constructor, so N is never re-evaluated.
    - `1 until (gold/-100)` executes its first iteration when gold falls to -200 or below - upper end exclusive. So the guarding `if` should also test <= -200
    - maintenance should be recalculated
    - the sequence is built, then used, then that `-=` operator... Someone was thinking in Set's, or the seq was patched in later - this wraps the seq into another one avoiding the entry - sequences! To take an element out, you need to complicate the instructions it represents. Another malloc.
    - ..and then it was discarded anyway
    - Now something very debatable: That `first()` was as good as nondeterministic - if playing without reloads, order of creation, more or less, probably. Load - rebuilt from tilemap, tile enumeration order. Disbanding the cheapest first - or better the GDR that is more likely to cancel the debt??? Or plonk the original back in as in that comment?

- ~~I'll check that failing test soon~~ - it tested an incomplete situation conflicting with the isDefeated test now in there

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -505,7 +505,7 @@ class Civilization : IsPartOfGameInfoSerialization {
      */
     fun isDefeated() = when {
         isBarbarian() || isSpectator() -> false     // Barbarians and voyeurs can't lose
-        hasEverOwnedOriginalCapital == true -> cities.isEmpty()
+        hasEverOwnedOriginalCapital -> cities.isEmpty()
         else -> units.getCivUnits().none()
     }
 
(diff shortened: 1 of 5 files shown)


yairm210:
I don't think either way is really better

PR #9021, opened 2023-03-24
Update Spanish.properties

Hello200000:
Mini update


Caballero-Arepa:
I dont like this changes

PR #9025, opened 2023-03-24
Add "Cannot annex cities" unique

Skekdog:
Follows the same rules of Venice in Civ5 (notably meaning cities can be razed whilst puppets).
I'll do the purchase in puppeted cities unique later (or someone else can).

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -7,7 +7,7 @@
 # If the first word in a sentence starts with a capital in your language, 
 # put the english word 'true' behind the '=', otherwise 'false'.
 # Don't translate these words to your language, only put 'true' or 'false'.
-StartWithCapitalLetter = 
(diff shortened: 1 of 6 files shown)


Skekdog:
>Still, the number suggests you did not allow your IDE to respect our `.editorconfig`...

Hmmm
I thought I had done that - maybe something changed when I nuked the old fork

As for everything else, I'll keep it in mind. I can't make any changes at the moment (and not for the next few days), but I will do so when possible. Also I thought I had completely nuked the fork on github and local copies, yet the music change still remained. I'll definitely be looking at the diff github provides now...

Thanks for letting me know about this - and how do I go about nuking a fork properly?
Apparently deleting the local copy and the entire repository wasn't enough to prevent an old music change creeping in

Skekdog:
Re-installing windows would surely be enough right? I am due a re-install soon anyway

PR #9026, opened 2023-03-24
City-state music plays when first meeting them

Skekdog:
War music plays when meeting the war pop-up plays. I think it's better like that.


Skekdog:
I believe that this PR can still be merged, right? Unless that would cause even more problems
And thanks for the tips, though I'm sure that was what I did. I think there are still some leftovers from when I attempted to nuke my fork.

PR #9027, opened 2023-03-24
Replace Rankings in VictoryScreen with Charts

WhoIsJohannes:
I'm running out of time to figure out the UI on the VictoryScreen, but I think this reached a state, where I'd like to get a review going.

Looks like this now:
https://drive.google.com/file/d/1rXjAkeModWmNNw_Bc4jat4vczYBIq8D4/view?usp=share_link

I spent a lot of effort making the size of the chart match whatever you wish it to be not using negative x and y coordinates etc in the hope to be able to align it nicely on the victory screen, but no matter what I do, it always seems to come out in a weird position. Would appreciate some help with that.


WhoIsJohannes:
Chat GPT pointed me to it... Just paste the code and ask how it can be
simplified...

Il ven 24 mar 2023, 23:15 SomeTroglodyte ***@***.***> ha
scritto:

WhoIsJohannes:
Thanks for the initial feedback. I'll have to take some time to read
carefully through what you wrote since I don't think I understood
everything right away. Maybe I'll ask Chat GPT to translate xD

> Bundling chart and civGroup

I did this since I expect these two to communicate with each other in the
next iteration. I.e. you click on a civ and that civ disappears from the
chart etc. Do you think that kind of behavior would still easily possible
if we move the two things into separate Widgets?

> VictoryScreen refactor

How about this: Let's merge in the standlone LineChart.kt and the CivGroup
refactor and then you pick it up and use it in whatever way you like for
the new VicotyScreen? I don't have any strong opinions and it seems you
already have it planned out in your head (and the necessary skills to do it
nicely), so I really don't want to get in your way.



Am Sa., 25. März 2023 um 00:21 Uhr schrieb SomeTroglodyte <
***@***.***>:

WhoIsJohannes:
> Coding style notes, same getLineChartData - policy in accordance with kotlinlang.org/docs/coding-conventions is - for over forEach unless clear added value, like end of complex linq-style cascade. And I'm sure that entire thing could be written as a pretty readable map/flatten/fold combo.

Simplified it a bit further

> Not what you asked. OK, how to actually run this... Oh, you're replacing the rankings page? Oh nice - I see. What's what there and what's nested where, not obvious. In term of Actors I mean... My very first Idea - that would be simpler if my start to clean up VictoryScreen were through first including the not yet written followup. TabbedPager is pretty good isolating the contents from the surroundings of itself. You could even use one for the which-stat selection, though it's geared to indeoendent pages, it's doable having just one char instance passed around... Back on track..

I synced past your change. It displays nicer now. Still on top of each other, but at least both are now centered.

> First thing: The legend would be better out of the LineChart widget. One, flexibilty, two, encapsulation, three, taking a complex Widget prepared for inclusion into a Layout and then only calling its draw() directly feels very wrong. Just do it outside and use normal containers. E.g. the bottom label of the PickerPane could be replaced with a HorizontalGroup with wrap on, could easily hold ~12+ civ buttons. Scale(0.5) and it may be enough.

I agree with you. Do you mind if we move on with this for now? I'm running out of time on this one. I'd really like to move to the AI improvements and my wife is already getting angry at me for spending so much time "on this stupid game"...

> First thing (yes insert before the other first): LineChart implements Layout but does not really participate. [...]

I still don't really understand what you're saying here. It is probably due to my limited understanding of how the UI works in GDX. I added some overrides, but I doubt it's sufficient for what you had in mind?

Do you mind if we merge this (after some more in-depth review) minus the changes in VictoryScreen and you take over?

WhoIsJohannes:
Happy to review, thanks for working on it, but will be Monday earliest.

Il sab 1 apr 2023, 12:34 SomeTroglodyte ***@***.***> ha
scritto:

WhoIsJohannes:
Took a quick look. It's a lot of code - probably whoever wrote the code
before is a better candidate, but some thought:

> VictoryScreenReorg
<https://github.com/SomeTroglodyte/Unciv/tree/VictoryScreenReorg>

Nice! I only skimmed it, but it all looks like clear improvements to me.

> VictoryReplaySlider
<https://github.com/SomeTroglodyte/Unciv/tree/VictoryReplaySlider>

Unfortunately don't know how to see a diff to the previous one, but in
general, I like the slider - thanks for improving on this!

> VictoryChartsJohannes
<https://github.com/SomeTroglodyte/Unciv/tree/VictoryChartsJohannes>

> WhoIsJohannes's code drew all the CivGroups on the right ...

LGTM, I like it. Do you have a screenshot for that? Btw I found a bug where
if the civ is defeated the CivGroup would no longer show its color. Is that
bug fixed by any chance already now or still there?

Btw I think I want to add negative values later on. I want to reserve 10%
below the x axis for it and just have a simpler condensed scale for it.
Maybe use powers of ten and then just squeeze whatever values are there
into it and keep the positive scale the same. E.g. it could be between 0
and -100 in the bottom 10% of the chart and 0 to 300 or whatever in the
remaining 90%. I think it wouldn't complicate the code too much and would
fit most use cases nicely.

> private fun getLastTurnDataPoints(): MutableMap<Civilization, DataPoint> {

I think there's another bug for defeated Civs here. But is the score even
shown anymore now? Just the ordering was off or even the last value was off
previously I think.

Btw, can we remove that alternative demographics view? It seems inferior to
me as usually I need to see the value for some particular civ and if they
don't happen to be best or worst I won't see it. If this isn't in active
development or widely used, seems unnecessary to carry it forward? Maybe a
good time to prune it?

> VictoryScreenChart
<https://github.com/SomeTroglodyte/Unciv/tree/VictoryScreenChart>

> private const val maxVertices = 5000

If the game has 1000 rounds, you might end up with a lot more lines. Would
that be a problem?

> calcGranularity

I need a manual for this algorithm xD. The all negative case seems more
theoretic. Do you think it would ever happen in a real game?

I see that you already added the highlighting feature. That's super cool.

All in all, you have my blessing whichever way this takes. I'm not married
to my code and if you took the time to rewrite it or I could at least
somewhat inspire this really cool feature, it is more than enough for me.
I'm actually really grateful you took the time to productionize this
better. Really looking forward to see this in the game.

Cheers
Johannes




Am So., 2. Apr. 2023 um 19:03 Uhr schrieb SomeTroglodyte <
***@***.***>:

WhoIsJohannes:
> Not a bug, made so on purpose by whoever originally did that CivGroup
thing. On the contrary - *your* code reveals too much by showing actual
nation line colors for undiscovered Civs, which could ease guessing who it
is.

1. Yes, I should say "a bug in this context" since if the civ is defeated
you won't see which color it has from this screen.
2. Good point, maybe we should use a grey color for all civs that aren't
encountered yet by the current player.
3. Using two colors sgtm, but it complicates the drawing unfortunately.

> Bug? Didn't look too closely, but first impression said "that number
could be right". Numbers are in the CivGroups as before - on the left,
outside LineChart.

I think it shows the latest stat before it was defated since that's the
last data point for that civ, but we should probably show 0 or nothing.

> That's what I avoided to get you to just run these branches

Will do once you get your first reorg in to look at a smaller diff

> No, please, go over as much as you can, 1 - still your baby or your
trigger, 2 - too much too fast, needs review and thorough triage. 4 xeyes
see 8 times as much as 1 🤓

Happy to collaborate on it, just let me know how you'd like to do it, so we
avoid doing the same work twice. Let's get your reorg in first and then let
me have another look at the diff. Or maybe we also merge the LineChart
class as it is as a baseline without using it? Not sure it makes sense
though.



Am Mo., 3. Apr. 2023 um 00:13 Uhr schrieb SomeTroglodyte <
***@***.***>:

PR #9035, opened 2023-03-25
Unit overview force column, close promote exploit

SomeTroglodyte:
Individual commits:
* [Close exploit allowing promotion with 0 movement](https://github.com/yairm210/Unciv/commit/565bd002dd65f49913932eb37cc42f3c09533b24)
        Caused by the picker first doing the right check and disabling the right side button, but soon after on the actual selection of a promotion - poof. So, one could promote after attacking or with zero movement via unit overview or via unittable click, even if blocked as unit action...

* [Replace Unit Overview health column with Force](https://github.com/yairm210/Unciv/commit/2ffde754db46e1fca577084d2a2ce4606a79610b)
        I think ages ago we hinted that health is redundant since there is already the health _bar_ on the  left
        Force was readily available and may be useful if/once we make that grid sortable
        I would prefer some abstraction instead of a raw number - get max over all units on the map, percentage to color gradient, map to a list of icons (wet towel to gorilla 🐁🐀🐍🐈🐒🐐🐅🦏🦍🐘) - potentially expensive and far more work than this. Not doing such a thing without some guidance what would be attractive/useful here.

* [Linting](https://github.com/yairm210/Unciv/commit/10f5dbd82bed117b6fee2b8a8c9f5306f4f1cada)
        May actually improve promotion column in rare cases - trigger "hey why detour over floats when ints do" -> I noticed it wasn't including the promote star in the layout formula...

For laughs: Every single emoji up there is completely different in edit mode from here.
image

* Note: **AI can still promote with 0 movement points** if I read that code correctly (`UnitAutomation.tryUpgradeUnit` doesn't check, but perhaps the check is upsstream)

### DIFF
--- a/core/src/com/unciv/ui/screens/overviewscreen/UnitOverviewTable.kt
+++ b/core/src/com/unciv/ui/screens/overviewscreen/UnitOverviewTable.kt
@@ -34,7 +34,6 @@ import com.unciv.ui.screens.pickerscreens.PromotionPickerScreen
 import com.unciv.ui.screens.pickerscreens.UnitRenamePopup
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsUpgrade
 import kotlin.math.abs
-import kotlin.math.ceil
 
 /**
  * Supplies the Unit sub-table for the Empire Overview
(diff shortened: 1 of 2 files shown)


yairm210:
We don't use the term 'force' in Unciv, we use Strength, and it has an icon itself.
Please remove the extra column from this PR, there's too much going on here at once.

PR #9043, opened 2023-03-26
A renamed unit shows that new name in can promote notifications

SomeTroglodyte:
image
..."But I renamed that guy"...
image

I hesitated between displayName and shortDisplayname, potential multiple passes through tr() and the tidbit we don't want the chosen name to be translated... Never going to be really safe.


Caballero-Arepa:
Nice, my worry is that it will be too long, but I like nontheless

Could it be done in other cases as well? Like when an unit dies.

PR #9051, opened 2023-03-27
Untinted embedded symbols - Notifications and Citystate diplomacy

SomeTroglodyte:
<details><summary>Visuals</summary>

### Notifications (world, not history) category "Production"
Current:
image

This PR:
image

Alternative:
image

### Citystate friend/ally boni:
Current:
image

This PR:
image

Alternative:
image

</details>

The entire "trick" here is that `override fun layout`...

The whole approach is - debatable. It works just fine and is - given the constraints I started from - not a bad architecture. But - maybe think bigger? Turning on that Gdx markup flag globally is as far as I know per se not even destructive, and haven't noticed a performace impact - and I mean not as this PR does but running everything with it ON...

### DIFF
--- a/core/src/com/unciv/ui/components/ColorMarkupLabel.kt
+++ b/core/src/com/unciv/ui/components/ColorMarkupLabel.kt
@@ -1,0 +1,107 @@
+@file:Suppress("unused")
+
+package com.unciv.ui.components
+
+import com.badlogic.gdx.graphics.Color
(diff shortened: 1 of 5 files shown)


yairm210:
How does this look for other languages?

yairm210:
Looks good in general, and since it solves a problem we aren't otherwise solving I'm in favor as long as it doesn't somehow ruin the translations :)
It *shouldn't* because the colors are applied after translation, but I still would rather have confirmation first

PR #9087, opened 2023-04-01
Music player controls

SomeTroglodyte:
<details>

Before
image

After
image

Same on OptionsPopup.
(Hello "colspan" - no, the real reason I did this was - I painted that Pause.png for something else...)
</details>

Notes
- no credit for the pause png, that was 20 seconds tops from a blank canvas in gimp. Add one anyway?
- most of this just removes code duplication..


Caballero-Arepa:
Very nice, very thanks

PR #9096, opened 2023-04-03
Don't do anything crazy if military is inferior and never give away m…

WhoIsJohannes:
…ore than 10% (ceil) of your cities in any case.

Also left one TODO in the code, would love some input on that.


WhoIsJohannes:
Okay, I made the code more readable to return a positive number and adding it to the sum rather than returning a negative number and subtracting it. So no change, but more readable. IMO this whole confusion makes me even more suspicious this was never intended like this, but anyways, I also added a comment.

yairm210 (approved):
Looks okay
I'm not certain about the /10, since if you have 10 cities then maybe giving away 2 is feasible, but we can always play around with it more later

WhoIsJohannes:
Agree - let's do max 20% as I wouldn't know otherwise where the cutoff is.

PR #9099, opened 2023-04-03
Show a highlight for the tile that seems most suitable to found a cit…

WhoIsJohannes:
…y (can be turned off in settings)

Also wondering if we should increase the search radius a bit (currently 6) to make AI not found cities that close to each other?

https://drive.google.com/file/d/1ghrZ5x_6x1YHit7pbKBFe6mKUs1-Me6-/view?usp=sharing
https://drive.google.com/file/d/1UfTpyNk4f6TaEXGk0dLCUL83B0s8BIV0/view?usp=sharing
https://drive.google.com/file/d/1pGThgAd_jwCPthALoj4MbOvONJ84GMBK/view?usp=sharing

Still not sure how many such locations we should show or whether there should be a cutoff on the quality (e.g. min score).

Some more considerations:

- The AI is currently cheating by knowing the entire map for this calculation
- It is a little bit of an exploit. You might see a tile as recommended where you have no knowledge of its surroundings, but the fact that it shows up as recommended location would give you a clue that there must be some luxury around there etc.
- The recommended tiles might change as you move the settler (since you're discovering the map as you go) which might be confusing for the user
- I'm not sure whether the overlay plays nice with all sort of tiles, it might overlay some tiles that have heavy resources (e.g. 3 food + 1 production or similar)


yairm210:
- AI needs all the help it can get, not a problem for me
- for players, we can only check tiles we know of
- As usual, pics

WhoIsJohannes:
There's some screenshots linked although with only one suggestion.

Il lun 3 apr 2023, 12:23 Yair Morgenstern ***@***.***> ha
scritto:

WhoIsJohannes:
> for players, we can only check tiles we know of

Done

Screenshot from 2023-04-03 16-06-14
Screenshot from 2023-04-03 16-06-26
Screenshot from 2023-04-03 16-06-36

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -0,0 +1,135 @@
...
+            ) // Restrict vision when far from home to avoid death marches
+
+            val possibleCityLocations = unit.getTile().getTilesInDistance(range)
+                .filter { it.isExplored(unit.civ) || unit.civ.isAI() }
+                .filter {
+                    val tileOwner = it.getOwner()
+                    it.isLand && !it.isImpassible() && (tileOwner == null || tileOwner == unit.civ) // don't allow settler to settle inside other civ's territory
+                            && (unit.currentTile == it || unit.movement.canMoveTo(it))
+                            && it !in tilesNearCities
+                }.toList()
```
yairm210, line 60: why tolist if you're going to .map it in a second anyway?

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -0,0 +1,135 @@
...
+
+            // This is to improve performance - instead of ranking each tile in the area up to 19 times, do it once.
+            val nearbyTileRankings = getNearbyTileRankings(unit.getTile(), unit.civ)
+
+            val distanceFromHome = if (unit.civ.cities.isEmpty()) 0
+            else unit.civ.cities.minOf { it.getCenterTile().aerialDistanceTo(unit.getTile()) }
+            val range = max(
+                1,
+                min(5, 8 - distanceFromHome)
+            ) // Restrict vision when far from home to avoid death marches
```
yairm210, line 51: how about 
```suggestion
            val range = (8 - distanceFromHome).coerceIn(1,5) // Restrict vision when far from home to avoid death marches
```

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -0,0 +1,135 @@
...
+
+            val possibleCityLocations = unit.getTile().getTilesInDistance(range)
+                .filter { it.isExplored(unit.civ) || unit.civ.isAI() }
```
yairm210, line 54: That's the fourth time you've used this, maybe make this an internal function? Like `fun isKnownLocation(tile:Tile) = tile.isExplored(unit.civ) || unit.civ.isAI() `

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -0,0 +1,135 @@
...
+            )
+        }
+
+        private fun getNearbyTileRankings(
+            tile: Tile,
+            civ: Civilization
+        ): Map<Tile, Float> {
+            return tile.getTilesInDistance(7)
+                .filter { it.isExplored(civ) || civ.isAI() }
+                .associateBy({ it }, { Automation.rankTile(it, civ) })
```
yairm210, line 96: We should already have the tile rankings at the top, as nearbyTileRankings, for performance reasons (we don't want to have to calculate them for each tile we're considering as city center)

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
WhoIsJohannes: Done

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
WhoIsJohannes: Done

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
WhoIsJohannes: Done

WhoIsJohannes:
@yairm210 This one is also waiting for review.

PR #9101, opened 2023-04-03
Prevent MapEditor 'pinch zoom' painting tiles

### LINKED ISSUE #3435: Map editor - Zooming often places a tile
**Describe the bug**
After two finger pinch zooming in the map editor, a tile is placed when lifting fingers.

**To Reproduce**
1. Go to 'Map editor'
2. Make a new map
3. Make it empty
4. Choose tile brush that will stand out
5. Zoom in and out via pinch gesture

SomeTroglodyte:
Fixes #3435
* The actual fix is just the one line in MapEditorEditTab.
* I tried and _couldn't_ get the other candidate culprit listener (dragpaint) to fire with pinching, so it's unchanged - tests on a hardware phone.
* I also integrated the zoom performance thing the other two mapholders got. As duplicated code. Managing that in a nicer way would pretty much mean an intermediate class `MapHolder` in the hierarchy and a thorough review what other stuff would belong there (as ZoomableScrollPane shouldn't know it contains a map - it shouldn't even specialize in Group imho). So, not here.

Asides:
* Let's think about more package reorg - components should get a 'widgets' subpackage methinks??
* Unit tests - it's lateinit and nullability again. Couldn't we introduce rules like 'All our uses of lateinit require _either_ a good documentation why it's _guaranteed_ to be initialized when the constructor is done _or_ it's private, strictly forbidden to be leaked outside the class, and all accesses are documented why the lateinit must be initialized by then - or guarded with ::field.isInitialized - and then nullability would be a cleaner design..... .... .. . ?? We have Tile that has a TileMap lateinit even though loose tiles not on a map are used both in editor and unit tests, and we have maps not part of a game, perfectly valid, and each map has a GameInfo lateinit - feels like an Anti-Pattern. As it is, several unit tests will need fiddling - and will e.g. show that the roadowner design was just fine - under the __assumption_ every tile is on a map and every map is part of a game......

### DIFF
--- a/core/src/com/unciv/ui/screens/mapeditorscreen/EditorMapHolder.kt
+++ b/core/src/com/unciv/ui/screens/mapeditorscreen/EditorMapHolder.kt
@@ -17,6 +17,7 @@ import com.unciv.ui.components.tilegroups.TileSetStrings
 import com.unciv.ui.screens.basescreen.BaseScreen
 import com.unciv.ui.components.ZoomableScrollPane
 import com.unciv.ui.components.extensions.onClick
+import com.unciv.ui.screens.basescreen.UncivStage
 
 
 /**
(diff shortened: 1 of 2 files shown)


yairm210:
The test problems were problems with the test itself setting up stuff in a way that only *happened* to work, not a problem with the lateinit or the classes themselves

PR #9104, opened 2023-04-03
Change map resource controls presentation

### LINKED ISSUE #5839: Map - Map editor vs "new game" map
**Platform**
Windows 10

**Version**
3.18.11 patch-1

**Describe the bug**
Maps created in the map editor have different amount of (luxury) resources than the ones created through "start new game". See those two examples, both set to default settings, no mods. The only 2 options I changed are "Resource settings" to "Abundant" and the bar of "Resource richness" all the way to 50 (max).
This bug probably started appearing when the "Resource Setting" & regions were introduced.
Notice the massive difference in the amount of luxuries between the map editor (image 2) and the one created through the "start new game".  My hunch is that the "start new game" maps somehow defaults back to the standard settings when creating the map.

image
image

SomeTroglodyte:
Closes #5839 
Just a band-aid, a clean solution would entail far more map editor work - including actual exact control over resourceAmount per tile...

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -379,6 +379,7 @@ Sparse =
 Abundant = 
 Strategic Balance = 
 Legendary Start = 
+This is used for painting resources, not in map generator steps: = 
 
 Advanced Settings = 
 RNG Seed = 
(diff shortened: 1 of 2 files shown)


Ouaz:
painting

^^

Happens on Android and Windows (4.6.0).

PR #9108, opened 2023-04-03
Notifications can be hidden

SomeTroglodyte:
Inspired by @Gualdimar , took less than an hour for a working POC and days to polish. Time to get it out here and see if it's to anyone's taste.

<details><summary>Demo clip</summary>

https://user-images.githubusercontent.com/63000004/229607399-626964b3-49f2-424b-833d-dbef3e6f3785.mp4

</details>

#### Hints on the clip
- With the reduced resolution for the screencap a little hard to see - the Categories all have lines, all now the same total width independent of category name length
- You can still, as before, only grab the pane's children. That means once the last piece of the longest line is flinged off to the right you've got nothing to grab, so the button appears, a click animates the pane fully back in.
- The fades are asymmetric on purpose. In - zen. Out - "aye sir!" - I like it that way.
- It's possible to have some portion of a notif still "sticking out" but behind the top or bottom widgets - tough luck, no button. Detecting the situation left as future exercise - in the mean time just resize the window, change orientation, play until another notification pops in, or until next turn. Open to simple ideas. The options UI might also help mitigate.

#### Other Notes
- The detail that I reintroduced the "notifications continue behind the floating UI elements" - optional, can be extricated. The pro is, as before, that cutting off a notification "oval" at the ScrollPane's clip boundary looks jagged, and a tiny little bit that it's easier to see in some situations that there _is_ more. Yes very much a matter of taste. OK, having the scp always have screen height makes a few things simpler too.
- Incomplete as in - needs some user preference. Postponed intentionally. Coarse idea: "World screen notifications: [Fixed/Hideable/Disabled]" with Fixed behaving as before, Disabled just that and the Button redirects to Overview Notifications history instead, and Hideable like this PR. Maybe let user also choose whether a change in notification content is allowed to reopen the pane (like it is doing here) or not, tinting the button instead???
- Also closes the minuscule imperfection where the categories had a smaller click/grab area than the detail lines
- I know I forgot one padding and some comments on which padding does what - next time

### DIFF
--- a/core/src/com/unciv/logic/civilization/Notification.kt
+++ b/core/src/com/unciv/logic/civilization/Notification.kt
@@ -70,17 +70,18 @@ open class Notification() : IsPartOfGameInfoSerialization {
         this.category = category.name
     }
 
-    fun addNotificationIcons(ruleset: Ruleset, iconSize: Float, table: Table) {
+    fun addNotificationIconsTo(table: Table, ruleset: Ruleset, iconSize: Float) {
         if (icons.isEmpty()) return
         for (icon in icons.reversed()) {
             val image: Actor = when {
-                ruleset.technologies.containsKey(icon) -> ImageGetter.getTechIconPortrait(icon, iconSize)
-                ruleset.nations.containsKey(icon) -> ImageGetter.getNationPortrait(
-                    ruleset.nations[icon]!!,
-                    iconSize
-                )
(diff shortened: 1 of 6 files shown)


yairm210:
Looks really good!!

AdityaMH:
Then separate by category. So you can focus to battle notification only or domestic (production, unit, promtion, pillaging) only.

yairm210:
Sure, I'm in favor
Worst case we'll have to undo

PR #9120, opened 2023-04-04
Show city size in CityScreenCityPickerTable.kt

WhoIsJohannes:
I need this quite often when deciding whether to build something such as "Library" (get 1 research for each 2 population -> "Oh, wait, what's the population of this city?")

Longer term I want to check how hard it would be to show the effects of something more directly e.g. for the social policy picker screen. E.g. "Gain one happiness for each city with a garrison" or "Empire produces double the amount of resources" or "1 extra happiness for each walls, castle, arsenal etc.", would be nice to see directly to how much happiness, resources that translates to make a better decision.

Screenshot from 2023-04-04 21-47-44
Screenshot from 2023-04-04 21-47-49


WhoIsJohannes:
Ah, you mean in below the stats? I never noticed it to be honest. I still think it's nice to have it more prominent, but will let the boss decide.

PR #9121, opened 2023-04-04
Victory screen score charts

SomeTroglodyte:
Please attribute to @WhoIsJohannes entirely, I just adapted it into the VictoryScreen.

There's still a lot to iron out, and it's remotely possible the little conceptual flaw (heavy draw override instead of doing the heavy lifting in layout only when needed) may be noticeable on phones, but I'd advocate letting people try it out.

<details><summary>Current screenies</summary>

image

Zoom-in:
image
Yes that's some off-by-one we'll have to hunt later - green is the human, active player.

</details>

#### Open questions:
* When to show -> a simple matter of one isHidden override. I was extremely generous here.
* Whether to drop other offerings instead - at the moment, with the debug finger acrobatics, you can see all six possible "pages" side by side...
* Whether to guard the page with some "warning: experimental" switch before release


WhoIsJohannes:
This is very exciting. Thanks a ton. I'll take a closer look later and will
probably add the negative values later.

You don't need to do the thing with attribution honestly. I don't care and
you probably spent more time on this than me by now. I'm just happy to see
this awesome game become even better all while I can learn some kotlin and
do something that is fun.

Il mar 4 apr 2023, 22:00 SomeTroglodyte ***@***.***> ha
scritto:

PR #9125, opened 2023-04-05
Great people automation

WhoIsJohannes:
Automate great scientists, great engineers and great merchants.

- Great Scientists: Will hurry research at a probability of 50%
- Great Engineer: At a probability of 50%: Will walk to a city at a maximum of 5 turns away having a population of at least 3 and having a wonder that can be hurried and that needs at least another 5 turns to be completed normally. Once the great engineer has arrived in that city, it will hurry the most expensive such wonder. If it's not possible to reach that or any such city, the great engineer will fall back to constructing a manufactory or starting a golden age.
- Great Merchant: At a probability of 50%: Will walk to the closest city state that player isn't at war with and has already discovered that is reachable within a maximum of 10 turns. Once there, it will conduct a trade mission. Similar fallbacks apply as for Great Engineer.

Further changes:

- Civilian units will also walk unaccompanied if there are no enemy military units within a radius of 5 around them.
- If a great person can't use their special ability or place a special building, they may start a golden age.


core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -288,14 +301,108 @@ object SpecificUnitAutomation {
...
                 if (unit.currentTile.isPillaged())
                     UnitActions.getRepairAction(unit).invoke()
                 else
-                    UnitActions.getImprovementConstructionActions(unit, unit.currentTile).firstOrNull()?.action?.invoke()
-            return
+                    UnitActions.getImprovementConstructionActions(unit, unit.currentTile)
+                        .firstOrNull()?.action?.invoke()
+                return true
+            }
+            if (unitTileBeforeMovement == unit.currentTile) {
```
yairm210, line 314: Simplify to "return a != b"

core/src/com/unciv/logic/map/mapunit/UnitMovement.kt
```diff
@@ -726,6 +726,9 @@ class UnitMovement(val unit: MapUnit) {
             if (!unitSpecificAllowOcean && unit.cache.cannotEnterOceanTiles) return false
         }
 
+        if (unit.getMatchingUniques(UniqueType.CanTradeWithCityStateForGoldAndInfluence)
```
yairm210, line 729: Does HasUnique consider state for conditionals? If so that's what you should be using

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -758,4 +758,8 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
         }
         return false
     }
+
+    fun canBeHurried() : Boolean {
```
yairm210, line 762: This doesn't need to be a function in the building, we don't need a boolean for every unique a building can have

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -262,16 +263,63 @@ object UnitAutomation {
...
 
-        // ToDo: automation of great people skills (may speed up construction, provides a science boost, etc.)
+        // TODO: The AI tends to have a lot of great generals. Maybe there should be a cutoff
+        //  (depending on number of cities) and after that they should just be used to start golden
+        //  ages?
 
         return // The AI doesn't know how to handle unknown civilian units
     }
 
+    private fun tryStartGoldenAge(unit: MapUnit) {
```
yairm210, line 318: Do, or do not
There is no "try"
(This function can't fail)

WhoIsJohannes:
I did this to not aim for wonders that can't be hurried. I didn't really analyze it, because it might also change in the future. I thought this condition would always make sense. Do you think it's problematic for this?

I added quite a lot of comments to the code that hopefully makes it readable, but mostly it's for great scientists (sometimes hurry research), great engineers (sometimes hurry wonders) and great merchants (sometimes conduct trade missions). In addition if nothing works and no improvement can be placed, the AI will now also use great people to start a golden age. Does that help?

WhoIsJohannes:
Sorry, was out over Easter. Just addressed / replied to all comments. Thanks for the review!

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
yairm210: My good man
All the functions that return booleans, above, don't have the "try" prefix
And this one does
That should be reversed

WhoIsJohannes:
@yairm210 Would appreciate a review on this one, I updated the description to be more detailed.

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -288,14 +301,105 @@ object SpecificUnitAutomation {
...
+     * states. */
+    fun conductTradeMission(unit: MapUnit): Boolean {
+        val closestCityStateTile =
+                unit.civ.gameInfo.civilizations
+                    .filter {
+                        !unit.civ.isAtWarWith(it) && it.isCityState() && it.cities.isNotEmpty() && unit.civ.hasExplored(
+                            it.cities[0].getCenterTile()
+                        )
+                    }
+                    .flatMap { it.cities[0].getTiles() }
```
yairm210, line 331: flatmap {} and then filter by hasExplored(), more readable

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -288,14 +301,105 @@ object SpecificUnitAutomation {
...
+                    .filter { unit.movement.canReach(it) && unit.movement.getShortestPath(it).size <= 10 }
```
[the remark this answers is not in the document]
yairm210: In fact you run this *thrice*, once in the minByOrNull! 
This should be reduced to once only, you can run minByOrNull() and then takeIf

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -288,14 +301,105 @@ object SpecificUnitAutomation {
...
+                    .minByOrNull { unit.movement.getShortestPath(it).size }
+                    ?: return false
+
+        val conductTradeMissionAction = UnitActions.getUnitActions(unit)
+            .filter { it.type == UnitActionType.ConductTradeMission }
+            .firstOrNull()
```
yairm210, line 338: firstOrNull accepts a filter, can be unified

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -288,14 +301,105 @@ object SpecificUnitAutomation {
...
+            // additional complexity for questionable gain.
+            (unit.movement.canMoveTo(city.getCenterTile()) || unit.currentTile == city.getCenterTile())
+                && unit.movement.getShortestPath(city.getCenterTile()).size <= 5
+                // Don't speed up construction in small cities. There's a risk the great
+                // engineer can't get it done entirely and then it takes forever for the small
+                // city to finish the rest.
+                && city.population.population >= 3
+                && getWonderThatWouldBenefitFromBeingSpedUp(city) != null
+        }.minByOrNull { unit.movement.getShortestPath(it.getCenterTile()).size }
+        if (nearbyCityWithAvailableWonders != null) {
```
yairm210, line 372: Reverse if for `if (... == null) return false`

WhoIsJohannes:
Let's discuss the last comment first. I don't agree with this. It makes
sense to get a gold boost sometimes even earlier in the game, same thing
for science. It can be nice to be ahead for a short time. IMO gold boost
from trade mission isn't relevant towards the end of the game, since you
usually have plenty anyways. Great scientist I typically always use to
boost research rather than building academies for my own play.

How about we change the probabilities? We can do 30% instead of 50%?

You didn't mention great engineers. Are you okay with that logic? IMO
getting wonders especially early in the game can be make or break, so I
think it makes sense to use some of them even earlier in the game.

Should we go with your suggestion: How would you determine how close to the
end of the game we are? Would that be scientific completion of the leading
player or something like that?

Il gio 13 apr 2023, 14:59 Yair Morgenstern ***@***.***> ha
scritto:

yairm210:
Since this seems to be a point of contention, I thing we should ask players on Discord how they choose to utilize their great people.
It's entirely possible we'll get a lot of different answers, but even the worst of those would be better than just by coin flip, so it's a win either way

WhoIsJohannes:
Resolved all comments.

PR #9128, opened 2023-04-06
Who wants another Easter Egg?

SomeTroglodyte:
... Also fixes two minor unrelated flaws:
* Turning easter eggs off in the settings would still show you "that stuff" on the appropriate days - by opening Civilopedia from the main menu.
* ColoredMarkupLabel wasn't used with wrap on yet, but it trips - hard - like normal Gdx Label does too, but it's harder to work around from outside the class. This thoroughly fixes that using reflection.

... Notes
* No spoilers
* Testers just add `-DeasterEgg=Eyes` to the run config
* The Label and wrap thing - I realize WrappableLabel has a completely different approach to the same issue, and does offer one feature more - which is done way too expensively (should call GlyphLayout directly for measurements) - merits a rework / merge, but not today.
* What is "The Label and wrap thing" - I believe Layout.getPrefWidth defaults to returning 0 because it's meant to mean "I have no idea" (but no proper JavaDoc). Label hard-exits with 0 immediately when wrap is on (Line 243). Table, however, interprets that literally (Table.java starting at line 826) as "I want to be infinitesimally skinny" -> Almost all glyphs end up arranged vertically...

### DIFF
--- a/core/src/com/unciv/ui/components/ColorMarkupLabel.kt
+++ b/core/src/com/unciv/ui/components/ColorMarkupLabel.kt
@@ -34,6 +34,10 @@ class ColorMarkupLabel private constructor(
                 fontSize: Int = Constants.defaultFontSize)
         : this (fontSize, prepareText(text, textColor, symbolColor))
 
+    /** Only if wrap was turned on, this is the prefWidth before.
+     *  Used for getMaxWidth as better estimate than the default 0. */
+    private var unwrappedPrefWidth = 0f
+
     init {
         if (fontSize != Constants.defaultFontSize) {
             val labelStyle = LabelStyle(style) // clone otherwise all default-styled Labels affected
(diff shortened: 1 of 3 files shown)


yairm210:
What the hell is this all about

Ouaz:
I would say something happening in Halloween maybe. XD

_*big_eyes*_

yairm210:
I have to say "no thank you" to this.
If we're going to add more easter eggs, they'll be for other holidays around the world - we're now in Ramadan, for example, and there's loads of holidays worldwide.
Since Unciv is about different cultures, we should also be multicultural with our easter eggs (or Pesach Afikomans as it may be)

yairm210:
I'm fine with having only Gregorian calendar holidays. But let's liven it up - Día de los Muertos! Diwali! St Patrick's! Qingming! There's [this lovely guy](https://en.m.wikipedia.org/wiki/G%C3%A4vle_goat)!

yairm210:
Even small things like turning the wheat symbol to a beer mug for Oktober, not all easter eggs need to be bombastic

PR #9130, opened 2023-04-06
Carrier right-click attack fix

SomeTroglodyte:
See comments in #4460 - is this the right thing to do? Improves things in my tests, doesn't smell like it could regress anything..

With the remaining 2 points in that issue being multiplayer and a fuzzy comment - close to closing that? Someone test multi? I think right click is solid from gameplay, there may be a little inconvenience moving units into own cities via r-cl, though I haven't really analyzed that yet, not even well enough to give a better description.


yairm210:
Looks ok, if there is something wrong with it it's clever enough to hide from me

Issue #9135, opened 2023-04-07
`Never appears as a Barbarian unit` does not work for Unit type


marek22k:
UnCiv Version: 4.5.16-patch1

Hello,

I am creating a mod and I noticed that the Unique `Never appears as a Barbarian unit` does not work for Unit type. Units of this type still appear as barbarian. Someone else on Discord could observe the same thing.

PR #9137, opened 2023-04-08
IHasUniques redesign, UnitType uniques, PerpetualStatConversion conditionals

SomeTroglodyte:
I'm only ~33% convinced these changes (see three commits separately) should go in as they are. Runs and seems to do a few nextTurns correctly, but not in-depth tested.

Key insights
* Due to `HiddenFromCivilopedia` being a UniqueType, any uniqueObjects or uniqueMap initialization "must not" access a ruleset. This hampers the idea to simply let BaseUnit load the uniques of its UnitType.
* One approach to that led to this idea of doing the late-caching without any Lazy-type delegate.
* Those ideas I played with in #9135 are bull, they might work fine in other contexts. But in Unciv, these lazies are fragile, you can get such code to run but produce wrong results due to them tripping in the wrong order.

Decisions
* As mentioned, what I did to PerpetualStatConversion would allow a mod to tie one to a condition that can become unfulfilled during gameplay - like you're allowed to convert prod to gold only if there's a Trading post near - improvement goes, you can still keep that in the queue, just not remove it and put it back... Same for the building conditional and selling the building... Maybe it's preferrable to document that such combos are intentionally unsupported?
* Instead of allowing the can't be a barbarian unique on UnitTypes, we could also opt for documented non-support.

Interesting tidbit: I found a formula to test whether a lazy is initialized without tripping it. Passes inspection and works in Debugger evaluation. Defined as function, however, it trips the compiler so hard you need to invalidate cache + kill all gradle daemons to recover... `fun isTypeInitialized() = ((::type).run { isAccessible = true; getDelegate() } as Lazy<*>).isInitialized()` - Fun for the whole family!

### DIFF
--- a/core/src/com/unciv/models/ruleset/IConstruction.kt
+++ b/core/src/com/unciv/models/ruleset/IConstruction.kt
@@ -224,10 +224,12 @@ open class PerpetualStatConversion(val stat: Stat) :
     fun getConversionRate(city: City) : Int = (1/city.cityStats.getStatConversionRate(stat)).roundToInt()
 
     override fun isBuildable(cityConstructions: CityConstructions): Boolean {
-        if (stat == Stat.Faith && !cityConstructions.city.civ.gameInfo.isReligionEnabled())
+        val city = cityConstructions.city
+        if (stat == Stat.Faith && !city.civ.gameInfo.isReligionEnabled())
             return false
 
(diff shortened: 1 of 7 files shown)


yairm210:
Please separate different aspects to different PRs. You write a lot and it's hard enough as it is for me to review it all, and mixing separate issues into the same PR makes it impossible for my smol brain

PR #9147, opened 2023-04-09
Resource stockpiles!

yairm210:
This is Step 1 of enabling stockpiled resources, as discussed in Discord @SeventhM

For the anti-Discordians amogus, Stockpiled resources means that if you have a mine that generates 2 iron, it would do so per turn, so after 3 turns you'd have 6
And then if, say, a unit requires 6 iron to be built, only then could you start the construction

This is the other staple of the genre compared to 'resource level' of Civ V, and it's used in Endless Legend and other games

Civ VI actually does both - Vanilla Civ VI uses resource levels like Civ V, but Gathering Storm does resource stockpiling

This PR *only* contains the bare minimum required to use Strategic Resources as a global counter, to replace shenanigans .

### How stockpiled resources work

You can increase the amount per-turn with 'provides' and 'consumes' uniques like with regular resources, BUT this is stockpiled instead. As I said, it's currently just a global counter, only useful for counting turns or something (Lonely Cave generates 1 Insight/turn, we need a total of 100 Things to research Enlightenment)

You can trigger gaining/losing resources with the new `"Instantly consumes [amount] [resource]"` / `"Instantly provides [amount] [resource]" uniques

You can add a one-time cost to units/buildings with the "Costs [amount] [resource]" unique. This cost is taken *when starting work on the construction*, so you can decide to use the resources from several places at once - but once the first starts building, the others won't have enough resources left and will therefore not be able to start

Roadmap:

- [x] Allow constructing buildings/units with stockpiled resources
- [ ] Allow trading of stockpiled resources

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -963,6 +963,8 @@ You may choose a free Policy =
 You may choose [amount] free Policies = 
 You gain the [policy] Policy = 
 You enter a Golden Age = 
+You have gained [amount] [resourceName] = 
+You have lost [amount] [resourceName] = 
 
 ## Trigger causes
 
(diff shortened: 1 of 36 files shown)


yairm210:
Pics!

image

image

image

yairm210:
image

I think we're ready :D

PR #9154, opened 2023-04-09
Make new "notifications" setting translatable

Ouaz:
Related to #9148 

Before:

notif_setting

After:

notif_setting_translated


Ouaz:
I think it would be better, yes.

I'd say:
```
- Visible
- Hidden
- Permanent
- Disabled
```

PS: that's a very nice QoL improvement, by the way (I just put a smiley on the PR a few days ago, rather than posting a message just to say "hey, that's really good")
So, I say it here. ^^

PR #9162, opened 2023-04-10
Fix map editor resource label wrapped to about 1 char width

### LINKED ISSUE #9157: Map generator screen UI's layout is bad
On my desktop, the text bellow is diplayed in a very narrow area:
This is used for painting resources, not in map generator steps:

Version 4.6.0 build 845

Képernyőkép erről: 2023-04-10 13-42-21

SomeTroglodyte:
Fixes #9157 
The touches to ColorMarkupLabel are preventive and a little optimization - it will need to be joined with WrappableLabel one of these days anyway, and I believe _that_ one actually needs a rewrite of all the Gdx.Label layout routines...


yairm210:
It seems that the color markup no longer works though
image

yairm210:
Hm no it doesn't seem to be from this commit

Issue #7792, opened 2022-09-13
AI has an obsession


Caballero-Arepa:
We have found that the AI has absurd obsesion with certain unit.

The case happened when doing an AI only match with all latam civs, and for some reason, some reason, Venezuela just straight up only and exclusively built "Mixed Militias".

Just that, it didnt build any other unit, and even well deep into the industrial era, it still did not made any other unit.

The obsesion was so big, that Ven. had a large amount of this explorers, and subsequely, got conquered as easy as butter.

![Screenshot_20220912-214726_Gallery.png](https://user-images.githubusercontent.com/78449553/189796637-52fdd789-1d32-4cac-b494-7dffc6990763.png)

This seems to not be the only case, from what I heard, Atlas from Deciv also has an obsesion with the particular unit and shows the same behaviour.

Unfortunately, I dont have anymore that save file, but this happens nearly every single game.
If needed, I may work to get another save file.

Ps: This extraordinary conduct also causes more lag, with such much units on the ground, the bot couldn't leave them quiet, so it made a bunch of lag.

RealBamboolord:
It probably has something to do with the units being able to build tile improvements

PR #9164, opened 2023-04-10
Fix issue #7792

MioBestWaifu:
The issue was caused by Latin-Civs Mixed Militias being considered workers in ConstructionAutomation.addWorkerChoice() but not beign counted as one in ConstructionAutomation.workers. Now, any unit that can build improvements counts as a worker in it. In addition, the AI simulates military units that can found city, build improvements or are religious as civilians in peacetime and as soldiers in wartime.

### DIFF
--- a/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
@@ -38,7 +38,7 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
 
     private val civUnits = civInfo.units.getCivUnits()
     private val militaryUnits = civUnits.count { it.baseUnit.isMilitary() }
-    private val workers = civUnits.count { it.cache.hasUniqueToBuildImprovements && it.isCivilian() }.toFloat()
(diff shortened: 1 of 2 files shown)


MioBestWaifu:
> 

Thanks for the hints, its my first time contributing to a real git project so i dont know how to properly issue pull requests. I tested the Latin-civs mixed militia and a custom soldier-settler in ai-only scenarios, as far as i know there is no other unit like that in the vanilla game, and it worked fine in about 20 simulations or so.  automateCivilianUnit() may fail to find a task if the worker has nothing to build or the AI cant find a place to settle, but i think its possible to make the unit do something else in that case. I did not check if isAtWar() is overly expensive, will look at it later. The civilian tasks take higher priority because the AI will mostly build those units for their unique purpose and not as regular soldiers, especially builders and settlers as these are limited in the code. 

If the changes in UnitAutomation generate further bugs or are undesirible they may be discarded as making the AI check if the "worker" is a civilian in addWorkerChoice() will also solve the bug in the original issue.

yairm210:
IsAtWar is not expensive, it should be the the first part of the || - but that's not a big deal
AutomateCivilianUnit can't fail

This looks good!

PR #9167, opened 2023-04-11
Arbitrary images in font!!!

yairm210:
This was quite annoying due to GL stuff, getting the black background and the sprite things to work was very fiddly
Luckily 90% of the final solution was provided [here](https://stackoverflow.com/questions/44767681/drawing-text-onto-a-pixmap-libgdx) so big kudos for [Abhishek Aryan](https://stackoverflow.com/users/3445320/abhishek-aryan) may his memory be blessed

image

image

And we can now do this to anything - buildings and promotions come to mind... ;)

We'll probably need to add lines in the tech tree descriptions because they're already quite full

### DIFF
--- a/core/src/com/unciv/models/ruleset/nation/Nation.kt
+++ b/core/src/com/unciv/models/ruleset/nation/Nation.kt
@@ -191,6 +191,7 @@ class Nation : RulesetObject() {
                 building.hasUnique(UniqueType.HiddenFromCivilopedia) -> continue
                 !religionEnabled && building.hasUnique(UniqueType.HiddenWithoutReligion) -> continue
             }
+            yield(FormattedLine(separator = true))
             yield(FormattedLine("{${building.name}} -", link=building.makeLink()))
             if (building.replaces != null && ruleset.buildings.containsKey(building.replaces!!)) {
                 val originalBuilding = ruleset.buildings[building.replaces!!]!!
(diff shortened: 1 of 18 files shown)


yairm210:
Regarding performance - the pixmaps themselves are only generated on-demand, so it's not like we frontload 100 buildings, but when you enter the civilopedia and all the names appear in a row then it will. I did notice a slight lag when entering the Civilopedia page for resources/units for the first time, but in my opinion totally worth it
It does mean that in places we already have the image, it will appear twice, so the question is should we get rid of one of the images.
Since I'm planning on image-izing almost all the basic game objects, my answer is currently "eventually yes"

image

yairm210:
Not to harp on but hot damn looks good
image

yairm210:
image

Much better, but now that notifications contain symbols we need to fix them up as well

yairm210:
image

Side bonus :)

yairm210:
Almost all the starting bonuses are double-icon'd as well so that's something we'll need to deal with

So far things left for this:
- Remove double icons on Civilopedia
- Remove double icons when choosing Civ

We can do this by either removing the old icons or the new ones. I prefer removing the new because the old are bigger and more visible, but maybe we can pick-and-choose.
This means that tr() needs to accept an optional boolean parameter for 'hideIcons', simple enough

yairm210:
Actually, no...
The nation picker is currently unhealthily split up because of those extra icons, so removing them would make the whole thing more readable, except for the base icons of 'unique unit' and 'unique building'. So we still need that override

yairm210:
Not true! We need the color markup label

PR #9172, opened 2023-04-11
Show replay after 50 turns, but only the discovered part of it.

WhoIsJohannes:
We don't really need to wait until the end of the game to show the replay if we just show the discovered portion of the map.
Screenshot from 2023-04-11 21-44-13
Screenshot from 2023-04-11 21-53-09

### DIFF
--- a/core/src/com/unciv/ui/screens/victoryscreen/ReplayMap.kt
+++ b/core/src/com/unciv/ui/screens/victoryscreen/ReplayMap.kt
@@ -3,6 +3,7 @@ package com.unciv.ui.screens.victoryscreen
 import com.badlogic.gdx.graphics.g2d.Batch
 import com.badlogic.gdx.scenes.scene2d.Group
 import com.unciv.UncivGame
+import com.unciv.logic.civilization.Civilization
 import com.unciv.logic.map.TileMap
 import com.unciv.ui.screens.worldscreen.minimap.MinimapTile
 import com.unciv.ui.screens.worldscreen.minimap.MinimapTileUtil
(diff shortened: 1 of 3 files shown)


WhoIsJohannes:
How difficult do you think it would be to make it behave like the minimap (fill to only the discovered part)? That would be even nicer IMO.

WhoIsJohannes:
Can you elaborate how exploredRegion would help? I just checked it out and
I'm not sure I can follow your idea or at least it doesn't seem simple to
me.

Am Di., 11. Apr. 2023 um 23:06 Uhr schrieb SomeTroglodyte <
***@***.***>:

WhoIsJohannes:
@yairm210 Please review. Not planning any further changes here.

Gualdimar:
I have not tested, but I assume that this breaks the "I don't know where I am in the world" approach. Seems you can roughly determine where your position is relative to the center of the map.

> How difficult do you think it would be to make it behave like the minimap (fill to only the discovered part)? That would be even nicer IMO.

You can check out the current implementation of the minimap: 
* the size of the explored region can be obtained from `civInfo.exploredRegion.getHeight()` and `civInfo.exploredRegion.getWidth()`. This should be the size of the replay map(the tile size is calculated accordingly).
* you can check if the tile is in the region with `civInfo.exploredRegion.isPositionInRegion(tileInfo.position)`

It's not that difficult to implement. I don't have time to do this now, but I can help by explaining the unclear parts)

WhoIsJohannes:
It's true that this can be revealing the position on the map. I'm also a bit tight on time. Lots of trips right now and little sleep, so might also take me a while to get it done. Let me know in case you pick it up, so we don't duplicate efforts.

PR #9181, opened 2023-04-13
Show highlight for settlers where they can found and not found cities.

WhoIsJohannes:
Playing around following a suggestion on the Discord that I kind of liked. The idea of showing both possible and impossible locations (rather than just possible ones) is that I only show them at a radius of 5 and I think by showing both this becomes obvious. If you only show possible ones, it might be confusing whether that tile 6 away is just not possible or not part of the highlight.

I'm not entirely happy with the UI. It feels a bit too intrusive, maybe someone has a better idea? Also I don't like that it hides the movement range of the settler. Any ideas on that?

I made it configurable at least, but would like to hear some feedback. I'm a bit torn on this change myself.

Screenshot from 2023-04-13 21-50-11
Screenshot from 2023-04-13 21-50-27

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -730,6 +730,7 @@ Show resources and improvements =
 Show tile yields = 
 Show unit movement arrows = 
 Show suggested city locations for units that can found cities = 
+Show possible city locations for units that can found cities = 
 Show pixel units = 
 Show pixel improvements = 
 Unit icon opacity = 
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -9,22 +9,73 @@ import com.unciv.models.ruleset.tile.ResourceType
 import com.unciv.models.ruleset.tile.TileResource
 
 class CityLocationTileRanker {
+    enum class IntendedUse {
+        AI_CITY_FOUNDING,
+        HUMAN_CITY_LOCATION_SUGGESTION,
+    }
+
     companion object {
-        fun getBestTilesToFoundCity(unit: MapUnit): Sequence<Pair<Tile, Float>> {
+        fun getBestTilesToFoundCity(unit: MapUnit, intendedUse: IntendedUse): Sequence<Pair<Tile, Float>> {
+            val radius = when(intendedUse) {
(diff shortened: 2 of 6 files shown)


AdityaMH:
This PR not merged, but why i can see this feature in lastest patch?

WhoIsJohannes:
This is different. The one that is live shows good City Locations with an
icon (three houses). This one highlights all tiles in a radius around the
settler in red or green.

Il ven 14 apr 2023, 12:10 AdityaMH ***@***.***> ha scritto:

yairm210:
image

This is what led to the current Discord poll
I really don't like the polkadot view, so why not copy from the best

WhoIsJohannes:
So are you planning to add support for whole tile highlight for airplanes?
If so, I'll wait for that?

Am So., 16. Apr. 2023 um 18:11 Uhr schrieb Yair Morgenstern <
***@***.***>:

yairm210:
Done
Also, since you can color tiles in white, it doesn't have to be green/red only - you can have white for 'you can but why should you', and green for 'this is a good place to settle' :)

WhoIsJohannes:
I would like to stick with green and red and the icon and then hopefully we
can still see the movement circles. Or do you have another idea for a color
scheme that odesn't use white? Maybe light green and darker green for good
spots? I'm worried that would be too difficult to differentiate though or
cover the underlying tile too much.

Am Mo., 17. Apr. 2023 um 08:03 Uhr schrieb Yair Morgenstern <
***@***.***>:

WhoIsJohannes:
Done, it looks like this now. Let me know what you think. I think it's still a bit too aggressive. Wonder if we should hide it before the first city is founded as it's not that useful then? It's a bit inconsistent though and might be even more confusing. Not sure.

Screenshot from 2023-04-17 10-47-31
Screenshot from 2023-04-17 10-47-47
Screenshot from 2023-04-17 10-49-17
Screenshot from 2023-04-17 10-49-43

yairm210:
:/

Nope, sorry, both versions are visually unappealing and detract more than they add, I'm going to have to give this a pass

yairm210:
If you were on Discord you would know that it's been polled out, I'm removing it next version

WhoIsJohannes:
I agree on this with you, wasn't happy with it myself. Good call.

Il lun 17 apr 2023, 19:12 Yair Morgenstern ***@***.***> ha
scritto:

Issue #7387, opened 2022-07-06
New "Notification" tab and "repetitive" notifications on smartphones.


the-Arioch:
New build introduced "notification" tabs. That can be good, because it is muc hlarger than "OSD"/"HUD" notifications rendered over map.

Frankly, the OSD notifications are so small that is it really hard to tap on the specific one on 5.5" phone. You would inevitably mis-aim and then the loop starts anew. Very annoying.

And there are notifications like "you have found 5 coal resources" that are*supposed* ot be tapped at one and again.

So, the new tab (where the notifications arew twice the size) could be handy, but... it is closed after use.
One more good feature there is that there is history, and you can for example click "N resources X" notification from 2-3 turns ago.

But repetitive clicks are still very hard.

Propose: when a repetitive action notificaiton is clicked in the new Notification tab - it get "reposted" to the OSD map, and it gets there double(triple?) the size to be tappable on physicalyl small screens (bind to real, physical screen size?). And they are afloat until the player taps elsewhere (implicitly closing that "reposted" notification). Or maybe it could be "pinned" to prevent implicit closing. Or maybe there should only be explicit "close me" button.

alexban011:
> but... it is closed after use

Not much we can do here. Maybe someone knows a way to remember the scroll position and then automatically scroll to the last position. I guess that would make it easier ?

TheEXAM:
Since the main issue is how hard it is to re-tap a notification on a phone, I think it would just be simpler to have notifications slide a bit toward the center of the screen when tapped so they stand out, therefore making it easier to tap by pressing on the protruding part of it.

The notification can return to its normal position once the player presses anywhere away from said notif, such as on the map or any other button.

the-Arioch:
won't help. Hard part is to reliably aim at the needed notification among dozen other notifications, when their real height seems to be 0.5 mm or 1 mm at most, much less than my non-transparent fingers.

Repositioning a list of tiny notifications would not address it.

TheEXAM:
Then have notifications move away from a tapped notification above/below them, thereby making the tapped notification stand out with a wide gap. It should be easier to tap something if there's no other notification beside it to obstruct.

PR #9182, opened 2023-04-13
Notifications can be "selected"

SomeTroglodyte:
This time I picked "let's solve #7387". Let's dub this the "Fat finger" update.

#### Questions first:
* Wording of the new option - click? select? tap?
* Revived old notifications don't have their text changed, they just get the papyrus look via background tint - alt: " (past)" or somesuch, as transmitting the actual turn number would be a little bit more than I was willing to do in this PR. Ideas?
* All Notification classes demand a WorldScreen - tempted to redesign such that e.g. tech is happy with any BaseScreen so the notif log in overview can pop the tech screen on tap and let it go back to overview on close... Comments?

#### Features
* "Scroll position preservation" on updates a bit nicer as X is now preserved relative to screen edge instead of the moving left edge of the ScrollPane
* Tapping a notification that is not fully in view scrolls it to the center of the un-covered area (a 55:45 decision over scrolling it _just_ far enough)
* Tapping a notification from history (preceding turns) "revives" that notification into the on-worldscreen list, tinted yellowish (Hue 50 / Saturation 33% / Luminance 99%) - disappears as soon as you do practically anything else. Eases reusing an old "Coal discovered" notif a little.
* Once you turn on the new display option, newly "selected" notifications are visibly enlarged (and the whole list nudged a bit to the left) to ease hitting them repeatedly. 50% + some extra pad, still way smaller than a button, but should help a lot. If it's not enough, all factors are in the const block so tweaking is easy.
* Location notifications get their own "cursor" - they won't reset their cycle whenever you pan the map away. Some might dislike this, as they won't _ever_ reset that cycle except on reload, but for "resource discovered" it's IMHO definitely a plus.

#### No Screenshotz!
<details><summary>All right, a video instead</summary>

First I'll "pick" an "attention Skirmisher" notification that's hidden behind the minimap by the portion peeking out -> it jumps to center of the notifscroll (animating that instead should be trivial, but I had had enough by the time I had the idea).
Next I'll show using a previous-turn notification, and last turning on the enlarge option:

https://user-images.githubusercontent.com/63000004/231885352-6990811e-fcab-4c53-a945-75f6ba70d122.mp4

</details>


AdityaMH:
I thought is about marked unread/readed (clicked) notification.

yairm210:
I agree with this functionality, I disagree with it being a setting
We have too many settings as it is, this is something that should be enabled by default

AdityaMH:
I agree with yairm. But why not do it both?
Apply all functionally (large text, colored, ~~and unread/read sysyem~~) when clicked?

yairm210:
Okay

Gualdimar:
Is it just me, or does the right third of the screen look like a mess? All elements are of different sizes and overlap each other. The tile info table is particularly eye-catching (maybe increase its padding?). 

Also, there may be some mistaps on the minimap buttons, which will scroll the world and notfications(can be annoying).

When the notification increases, the width of the entire block changes, so the section headers move to the left, and when the notification decreases, the headers move back to the right. It looks a bit jarring.

the-Arioch:
Just watched the video. No actual experience. 

> want off on my desktop, but on on a phone

Just though the same. Perhaps the default behavior could be bound to input methods (mouse VS touch screen) when it could be told one from another?..

The "fat opaque fingers" were the primary cause...

Just some more thoughts.

1. Could this enlarging somehow, in some cases, lead to map center (where, I think, the notification anchor point is scrolled to) being covered? I mean, just to think about, maybe this enlargement should also offset the scrolling a bit?

2. Maybe finger-scrolling the enlarged notification up/down could select (and enlarge) prior/next notification?

the-Arioch:
This latter option would also provide for canceling the kids-aimed tap:

1. I tap a notification and it grows.
2. I see at this moment that some other notification was pressed, not the intended one.
3. Instead of removing the finger (which would commit the tap like releasing mouse button would commit a click) I swipe up or down selecting another notification. Then I would be able so scroll mor or to tap an already enlarged, easier to aim at one.

the-Arioch:
finger-scrolling the enlarged notification up/down could select
>
> Go ahead try implementing that with the existing libs. Or better not, I
> don't wish you any harm.
>

Can there be some transparent yet mouse-capturing control, that could be
placed over the notifications?

>

PR #9185, opened 2023-04-13
Adding CityFilter uniques

SeventhM:
Adds the following uniques:
"[stats] from every [tileFilter/specialist/buildingFilter] [cityFilter]"
"[relativeAmount]% [stat] from every [buildingFilter] [cityFilter]"
"[relativeAmount]% [stat] Yield from every [tileFilter/buildingFilter] [cityFilter]"
"[relativeAmount]% Yield from every [tileFilter/buildingFilter] [cityFilter]"

----------------------------------------------

Currently, there is no conditional regarding cityFilters generically for objects for stat increases or percent increases. Even if there was one, there's no current implantation available for buildings for `[relativeAmount]%` [stat] Yield from every [tileFilter/buildingFilter] `[cityFilter].` This is because all current building percentage uniques explicitly only give percentage stats to city, and none allow for percentage increases of specific buildings.

Currently, the closest unique to any of these is `StatsFromTiles`, aka `[stats] from [tileFilter] tiles [cityFilter]`, as it's the only generic stats or percentage stats unique with a cityFilter. However, this unique does not support buildings or specialists. And this does not fix the current lack of a percentage stats unique with a cityFilter

As such, there are 3 primary reasons for this PR

1. Clarity to mod developers at what certain uniques do. E.g., `StatPercentFromObject` currently function very differently if it targets a tile vs if it targets. This information is not particularly obvious to mod developers or players until it's seen in game. This is particularly egregious with `AllStatsPercentFromObject`, aka `[relativeAmount]% Yield from every [tileFilter/buildingFilter]`, as it implies a yield increase despite not functioning as such at all when targeting buildings
2. As stated before, it is currently not possible to increase the stats of a particular building by a percentage rather than a city. While Civ 5 base does not use this feature, such a feature would be particularly beneficial towards mod developers. This is especially true for any mods wishing to create a similar gameplay experience to Brave New World despite that current lack of the tourism system. The mod I picked up previously used culture increases as analogous towards gaining tourism, to which hotels, airports, and the national visitor center in bnw would need a percent increase towards just specific buildings like specifically wonder to fit within that paradigm shift. What makes this particularly bad is this uniques we do have, `StatPercentFromObject` and `AllStatsPercentFromObject` is possible to do in other means using the conditional combination of `[relativeAmount]% [stat] [cityFilter] <in cities with a [buildingFilter]>` and should be a larger source of scrutiny
3. Currently, it is not possible to use something like `StatsFromObject` on a non wonder building as it will stack from each case of the building. There's currently no way to avoid this from happening, particularly with specialist stat boosts. None of the alternatives work in this scenario to limit the boost to just that city. In general, there should be a larger use of cityFilters for uniques like this, or we should drop many of the uniques that use [cityFilter]s in favor of turning into a conditional


SeventhM:
I realized this after the fact, but the way I implemented `"[relativeAmount]% [stat] from every [tileFilter/buildingFilter] [cityFilter]"` means that it doesn't do anything with the tileFilter. I'll be removing it tomorrow when I'm at a comptuer. Also, the implementation of the percent yield filters needs to be moved to their own functions

SeventhM:
Ok, Fixed the error I had and am updating the commit message with the reasoning. Because of an odd workflow I'm using, I ended up just resetting the whole file back to what is currently up to date to remove the linting.

yairm210:
There is too much lacking from this PR for me to be able to accept it as-is.
All the existing places where there is automation do not treat the new uniques, and we also need to deprecate old uniques when they are superceded


"[relativeAmount]% [stat] from every [tileFilter/buildingFilter] [cityFilter]" - the buildingFilter is understandable, but the tileFilter is complicated because there's a big difference between *give double food from Farms and "give double food from Farm *tiles*" - I assume you mean the former? But tileFilter also includes e.g. Plains, so "[+100]% [Food] from [Plains]" would mean double the yield of only the terrain bonuses?

Basically this needs to be split up into different PRs. Part 1 I think I understand and I'll make my own PR to see if that answers your needs

> Currently, it is not possible to use something like StatsFromObject on a non wonder building as it will stack from each case of the building. There's currently no way to avoid this from happening, particularly with specialist stat boosts

I don't understand your use case - what does it even mean for you to add stats to a building/specialist but have them 'not stack'?

yairm210:
I'm going to try to add a CityFilter conditional - with changes that have been done to the translation, this should *probably* be supported and should address many of these issues and more

yairm210:
Okay, so that can't work because of a mismatch between that and another unique :(
But what we *can* do is add the conditionals of cityFilter in gradually as their own Conditional uniques, and then eventually we can cover them all
That'll be a huge project but one worth doing - cityFilter currently doesn't accept any parameters in it, and with Conditionals it could :)
There are a couple of things that will be a big hassle to replace but it should cover the 90%

What city filters do you need specifically?

SeventhM:
> > Currently, it is not possible to use something like StatsFromObject on a non wonder building as it will stack from each case of the building. There's currently no way to avoid this from happening, particularly with specialist stat boosts
> 
> I don't understand your use case - what does it even mean for you to add stats to a building/specialist but have them 'not stack'?

The current primary use case I've ran into was attempting to add stats to specifically artist specialists. Previously, I've tried doing so on hotels, but it ends up adding the stats to all artists in all cities rather than just in the one city. The secondary issue is using a building to add stats to another building. Currently you can do `[stats] [cityFilter] <in cities with a [buildingFilter]> `, but this has readability issues (`[+5 Gold] [in this city] <in cities with a [Monument]>` reads worse than `[+5 Gold] from every [Monument] [in this city]`) and it does not work as intended for uniques (`[+5 Gold] [in this city] <in cities with a [Test Unique]>` would not give 5 gold to every test unique build, just 5 gold once to the city).



> What city filters do you need specifically?

`in this city` is _**sorely**_  needed. Beyond that, `in capital`, `in annexed cities`, `in puppetted cities`, `in all cities with a garrison`, and maybe `in holy cities` would likely be nice to have for mod developers. 


> "[relativeAmount]% [stat] from every [tileFilter/buildingFilter] [cityFilter]" - the buildingFilter is understandable, but the tileFilter is complicated because there's a big difference between *give double food from Farms and "give double food from Farm _tiles_" - I assume you mean the former? But tileFilter also includes e.g. Plains, so "[+100]% [Food] from [Plains]" would mean double the yield of only the terrain bonuses?

2 Notes here
1. Forgot to edit that. `[relativeAmount]% [stat] from every [tileFilter/buildingFilter] [cityFilter]` only applies to buildingFilters as I realized it didn't actually do anything for tiles. I assume you mean "[relativeAmount]% [stat] _Yield_ from every [tileFilter/buildingFilter] [cityFilter]"? The distinction being the one yield one increases the stats of the object, while the non yield one behaves as what we have currently: giving a percent bonus to the city itself. This may need a rewording 
2. To your question, I believe the former? Not going to lie, I'm most following the implementation currently provided by already existing uniques.  Currently, that's how it works with the uniques we already have that use tileFilter, so I believe only increasing the terrain bonus with `"[+100]% [Food] from [Plains]"` is "intentional

yairm210:
New version releasing now has "in this city" and "in other cities" conditionals.
This should work with a lot of uniques, but because of technical mumbojumbo some uniques are checked against the *city* and some against the *civ*, any unique we want to attach these conditionals onto has to be checked against the *city*, so tell me if you find any uniques that these don't work on

SeventhM:
Copy that. I'm going to close this and make a new PR to just add in `[relativeAmount]% [stat] Yield from every [tileFilter/buildingFilter]`

PR #9213, opened 2023-04-17
Disallow buying tiles in annexed cities in resistance

SomeTroglodyte:
I'm 99% sure the rules here should be "No!".

But say - wasn't there some discussion months ago that the price progression for _cultural_ expansion and _bought_ expansion should be independent? Buying X tiles shouldn't increase next culture cost and vice versa? I can't find the discussion. Would be one extra counter like free promotions or free policies are working right now, and would be fine for a migration phase to just start it at 0 for updating mid-game...


yairm210:
There was such a discussion, don't know where.
Personally I don't see great value in doing so

PR #9226, opened 2023-04-19
Little documentation cleanup

SomeTroglodyte:
The latest update forgot that markup needs an empty line to end a bulleted list, which tripped my triggers. Also sounded like "exclude SaveGames to skip indexing mods"...
gitignore just a little reordered

Question: Some of the recommendations - shouldn't our editorconfig already cover these, like continuation indent and trailing spaces, and shouldn't we document that support for project-level editorconfig must be (or stay, dunno the default) turned on?


yairm210:
If you can find a way to autoapply those then sure, I haven't delved into it

PR #9230, opened 2023-04-19
"can be promoted" notification only when it's actually new

SomeTroglodyte:
The motivation was originally that I saw three identical notifications for the same unit in one turn - couldn't repro or find suspects (likely simply being attacked between turns + Blitz + already tons of XP), but I had this idea of limiting the notification to when the unit couldn't already promote before the XP gain. If we want reminders, a "hey you forgot to promote" in startTurn might be better?

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -523,43 +523,46 @@ object Battle {
 
     // XP!
     private fun addXp(thisCombatant: ICombatant, amount: Int, otherCombatant: ICombatant) {
-        var baseXP = amount
         if (thisCombatant !is MapUnitCombatant) return
-        val modConstants = thisCombatant.unit.civ.gameInfo.ruleset.modOptions.constants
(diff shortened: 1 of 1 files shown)


letstalkaboutdune:
Does this also fix the game taunting you when your recently deceased units could have been promoted?

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -523,43 +523,46 @@ object Battle {
...
+        if (!otherIsBarbarian && civ.isMajorCiv()) { // Can't get great generals from Barbarians
+            val greatGeneralPointsModifier = thisCombatant
+                .getMatchingUniques(UniqueType.GreatPersonEarnedFaster, stateForConditionals, true)
+                .filter { unique ->
+                    val unitName = unique.params[0]
+                    // From the unique we know this unit exists
+                    val unit = civ.gameInfo.ruleset.units[unitName]!!
+                    unit.uniques.contains("Great Person - [War]")
+                }
+                .fold (1f) { acc, unique -> acc +  unique.params[1].toFloat() / 100 }
```
yairm210, line 558: Please use "sumOf" instead of "fold", it's much clearer

PR #9234, opened 2023-04-20
Show a preview of custom maps on new game screen

### LINKED ISSUE #7313: Minimap on custom map selection
if map have hard name to remember,
Or download maps from mods can be annoying. To choose and start game one per one to see whole map.
Tak berjudul19_20220626152240

SomeTroglodyte:
Goal: Close #7313 

I failed. As in, I didn't manage to understand the coordinate systems properly to fix ReplayMap's size, so I left its calculations as they were, under certain conditions it draws far enough out of its bounds to need scrolling, same as before.

image

Can't do video from this box, but they load concurrently and fade in/out nicely.


WhoIsJohannes:
I'm happy to take a look, but I'm traveling until Saturday without my
laptop, so it'll have to wait.

Il gio 20 apr 2023, 20:14 SomeTroglodyte ***@***.***> ha
scritto:

WhoIsJohannes:
Try this instead in ReplayMap.kt:

```
 private fun calcTileSize(): Float {
        val height = viewingCiv?.exploredRegion?.getHeight() ?: tileMap.mapParameters.mapSize.height
        val width = viewingCiv?.exploredRegion?.getWidth() ?: tileMap.mapParameters.mapSize.width
        return min(
            replayMapHeight / height / 0.75f, // hex height;width = 0.75 * d -> d = height;width / 0.75
            replayMapWidth / width / 0.75f
        )
    }
```

This was some copy pasta that I didn't clean up properly before. Not sure what the calculation is about honestly for the width. This seems simpler and works. Also not sure what the 0.5f padding was. It's not here, is it? Now the ReplayMap in the VictoryScreen Replay looks a bit small. Could you play around with the numbers there to size it correctly again?

All that said: I wonder if it would be better to use Minimap instead of ReplayMap? Or should we rename the ReplayMap? Why did you decide to go that route in the first place and not go with the minimap?

PR #9236, opened 2023-04-20
Fixing some typos in the German translation

Knirps1:


Knirps1:
I saw some of those typos while playing and then went through the file with hunspell (a spellchecker).

I agree, „Biebel” is really cute.

As for new or old orthography. I don't have a strong opinion. I would just prefer to choose one way to write a word and stick with it. I assume you refer to daß, muß and Fluß. All of those have more occurrences with ss. So we should either change them or go with new orthography.

I don't know, if you are able to edit. But I'm fine with changing anything.

Issue #9136, opened 2023-04-07
Unexpected prompt for server password


Chris1712:
**Platform**
linux x64 client (or android client), linux x64 multiplayer server

**Version**
Server 4.5.16-patch1, client 4.5.16-patch1

**Describe the bug**
A multiplayer game I've been playing for around a month no longer works - at the end of my turn I'm prompted to enter a server password, but one hasn't been configured. An empty password doesn't work, and the  Furthermore the `/isAlive` endpoint for the server I'm using returns `{authVersion: 0}` so I'm not sure why my client believes a password is needed.

If I've read https://github.com/yairm210/Unciv/pull/8716 correctly I don't think currently the server supports setting a password. I've found that reverting the server to 4.4.19, predating that work, resolves the problem even with the latest clients.

**To Reproduce**
1. Attempt to finish a turn as a player in a multiplayer game.
2. See the prompt for password, regardless of whether one has been set.

I can supply a savegame/work on a minimal recreation if that'd be helpful. The server doesn't appear to have created any other config files where passwords could be stored.

**Expected behavior**
No prompt to supply a password at the end of the turn.

**Screenshots**
The prompt I see:
![image](https://user-images.githubusercontent.com/8271330/230660146-0f4f395a-116b-47f5-a223-8301937450c3.png)

The multiplayer settings (I've got no ability to configure a password):
![image](https://user-images.githubusercontent.com/8271330/230660458-d5d6ecfb-93c6-4a6b-bd98-682a84bd1e9f.png)

yairm210:
@GGGuenni @touhidurrr Any ideas?

GGGuenni:
> I've found that reverting the server to 4.4.19, predating that work, resolves the problem even with the latest clients.

That's *super weird* especially if the server does return authVersion 0 in both versions

So I've got no clue what could be the problem right know

GGGuenni:
Quick look at the code doesn't look suspicious to me
The worldscreen only pops the Auth popup if a MultiplayerAuthException is thrown https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/screens/worldscreen/WorldScreen.kt#L571-L580

The exception is only thrown if a 401 is returned

Default value is still 0

So everything seems fine

Chris1712:
Thanks for taking a look @GGGuenni. I'll do some more investigation myself and see if I can figure anything out, I'll report back.

touhidurrr:
[...]

PR #9261, opened 2023-04-23
Don't return 401 when auth is disabled

Chris1712:
See https://github.com/yairm210/Unciv/issues/9136
The `extractAuth` function returns null when server auth is disabled, but `validateGameAccess` ought to return true for that case. Fortunately the logic is implemented in the `validateAuth` function so we can call that instead.

### DIFF
--- a/server/src/com/unciv/app/server/UncivServer.kt
+++ b/server/src/com/unciv/app/server/UncivServer.kt
@@ -78,13 +78,7 @@ private class UncivServerRunner : CliktCommand() {
         if (!file.exists())
             return true
 
-        // Extract the user id and password from the auth string
-        val (userId, password) = extractAuth(authString) ?: return false
-
-        if (authMap[userId] == null || authMap[userId] == password)
-            return true
-
-        return false
+        return validateAuth(authString)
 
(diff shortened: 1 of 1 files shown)


touhidurrr:
@Chris1712 can you kindly explain how the current auth works? I am pretty confused. I want to implement it before moving to v2.

Chris1712:
@touhidurrr I've not delved into it very deeply, I've just selfishly fixed a bug that affected me ;)

That said, my guesses:

1. Looks like it's base64 yeah: https://github.com/yairm210/Unciv/blob/b3166c4922b82ad51c83ff11a31e403bced9039c/core/src/com/unciv/models/metadata/GameSettings.kt#L240
2. This seems to be the string directly, no base64 encoding here: https://github.com/yairm210/Unciv/blob/b3166c4922b82ad51c83ff11a31e403bced9039c/core/src/com/unciv/logic/multiplayer/storage/UncivServerFileStorage.kt#L79

touhidurrr:
@Chris1712,  I get some weird characters when reading password from that base64. are you surr that's not a bug??

PR #9263, opened 2023-04-23
Don't leak other civs in the game through LineChart colors.

WhoIsJohannes:
Also wondering if we should change the bahvior of "knows" to return true for foo.knows(foo).

Screenshot from 2023-04-23 20-30-00
Screenshot from 2023-04-23 20-30-07
Screenshot from 2023-04-23 20-30-54
Screenshot from 2023-04-23 20-30-58


yairm210:
Last time we tried that it needed things up, we could add an override... Or maybe knowsOrIs()...

PR #9265, opened 2023-04-23
Performance improvements

WhoIsJohannes:
- Putting some canMoveTo later in evaluation
- Don't try fog busting if all tiles within a radius of 5 are already visible (ie. not in fog of war). I expect this to be relevant especially in the late game where most of the map is visible and there's a lot of military.
- Cache some values in getDistanceToTilesWithinTurn

I saw these methods as hotspots using VisualVM on a game that I spectated at round 50 when clicking Next Round twice.


yairm210:
Did you see any difference in performance in getDistanceToTilesWithinTurn?

yairm210:
Good catch with the fog btw!

WhoIsJohannes:
I think getDistanceToTilesWithinTurn will be more efficient since those two calls are relatively expensive and happen in a deeply nested loop. However I think the benefit would be greater if this would be cached across units for the entire turn.

yairm210:
Movement cost can be different for different units, so we can't cache it between units
#blame_civ_v

PR #9267, opened 2023-04-24
Do not consider Barbarian strength when deciding if to go to war

WhoIsJohannes:


WhoIsJohannes:
When calculating the sum of force of enemies at war with, don't consider barbs (currently they're added to the sum, after this change they are not). So if you're not at war with anyone (except barbs), now the sum will be 0, before it was the force of barbs.

PR #9271, opened 2023-04-24
Performance improvements

WhoIsJohannes:
Please see commit history for details.


core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -285,21 +285,18 @@ class WorkerAutomation(
...
                         .none { tile -> tile.militaryUnit != null && tile.militaryUnit!!.civ.isAtWarWith(civInfo)}
                 }
                 .sortedByDescending { getPriority(it) }
 
-        // the tile needs to be actually reachable - more difficult than it seems,
-        // which is why we DON'T calculate this for every possible tile in the radius,
-        // but only for the tile that's about to be chosen.
-        val selectedTile = workableTiles.firstOrNull { unit.movement.canReach(it) }
+        // These are the expensive calculations (tileCanBeImproved, canReach), so we only apply these filters after everything else it done.
+        val selectedTile = workableTiles.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
```
yairm210, line 297: tileCanBeImproved is cheaper than canReach, please reverse &&

core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
```diff
@@ -160,11 +160,15 @@ class CityPopulationManager : IsPartOfGameInfoSerialization {
...
+        val rankTileForCityWorkCache = mutableMapOf<Tile, Float>() // Create the tile cache
+
         for (i in 1..getFreePopulation()) {
             //evaluate tiles
             val (bestTile, valueBestTile) = tilesToEvaluate
                     .filterNot { it.providesYield() }
-                    .associateWith { Automation.rankTileForCityWork(it, city, cityStats, localUniqueCache) }
+                    .associateWith { rankTileForCityWorkCache.getOrPut(it) {
+                        Automation.rankTileForCityWork(it, city, cityStats, localUniqueCache)
+                    } }
```
yairm210, line 171: Uncacheable, since rank tile *for city work* depends on how much the city needs food - and the more population we assign to food tiles, the less they become important, so this caching actually disrupts

WhoIsJohannes: Will revert... Thanks for catching. Just FTR this is super heavy, too and ends up in the updateStats thing I mentioned in the other report again:

Screenshot from 2023-04-24 22-33-01

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
yairm210: Wow! Surprising
Data wins again 👍🏿

core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
yairm210: Can you drill down more into the function to see why? Chances are there's stuff we can cache in that function if it's heavy

PR #9274, opened 2023-04-24
Update French.properties

Ouaz:
Small translation improvements:
- Remove a few past tenses which are unnecessary in French (in the context, past participle alone is enough, especially in notifications)
- More precise translation for "the defence of [cityName]", to refer to the unit garrisoned in the attacked city ("la défense" doesn't sound well)


yairm210:
prepare version

Issue #9134, opened 2023-04-07
City production to Faith does not work


marek22k:
**Platform**
Android

**Version**
4.5.16-patch1

**Describe the bug**
The following unique does not work as a policy: `Enables conversion of city production to [Faith] <in cities with a [Stonehenge]>`

**To Reproduce**
Add the policy and go to the product menu in a city. Faith is not displayed.

**Expected behavior**
You can produce Faith.

**Additional context**
I am the second one who encountered this problem. I talked to a person in Discord who could observe the same behavior.

marek22k:
https://paste.i2pd.xyz/?7637f3c844fd01f0#EfAsdU1HLyPA3N6FaGUu8thRqBpXbD5jWMSHu4fEq3N3
(I can also upload the ZIP file somewhere else)

Install mod and then start a new game with the mod and the starting age "Modern". Then complete the policy "Holy World" and "Holy City".

marek22k:
How did you manage that? I have used the unique `Enables conversion of city production to [Faith] <in cities with a [Stonehenge]>"`, founded a pantheon and completed the policy, but can't produce a fai
![Screenshot at 2023-04-08 00-10-29](https://user-images.githubusercontent.com/46723399/230685355-92d1bfa7-7f1f-4d6b-b0f6-be518b8c0ad9.png)
th.
![Screenshot at 2023-04-08 00-10-18](https://user-images.githubusercontent.com/46723399/230685351-82ebfc99-6b9f-49cb-aa40-fc1a4
![Screenshot at 2023-04-08 00-11-48](https://user-images.githubusercontent.com/46723399/230685358-5e958140-2fe5-494d-a12f-419a38b2bae5.png)
fabef99.png)

![Screenshot at 2023-04-08 00-10-05](https://user-images.githubusercontent.com/46723399/230685350-e927b3fb-04c1-4933-8f5a-fee79a8b304b.png)

marek22k:
> Do we want a ~silly~ unique like "Enables conversion of city production to [Culture] <within [2] tiles of a [Landmark]>" to work as well?

A fun mod may allow themselves something like that :-) If it's not too inconvenient for you, I personally would like it.

marek22k:
![Screenshot_20230424-145810_1](https://user-images.githubusercontent.com/46723399/234003443-02b28d04-dda0-4a58-a1c1-404443f201ec.png)
I think there is still a issue with the icon.

marek22k:
[...]

PR #9275, opened 2023-04-24
Modders can use culture/faith conversion without own assets

SomeTroglodyte:
Marek22k in #9134 feels like the icons needed for a mod using UniqueType.EnablesCivWideStatProduction should already be available - I you feel they're right, this does it. If not, just close.


yairm210:
Mmmm yeah sure why not

PR #9278, opened 2023-04-25
Visual clue a load game from clipboard is underway

SomeTroglodyte:
I keep wondering, while loading saves from issues, whether I did even hit the button or not - hence this.

Reusing the "Working..." string (now an 11th time - should we move it to Constants?), and like elsewhere showing it on the button itself. There's more places that could use more feedback like this in save/load, but I limited this to what actually irks me.


yairm210:
Looks good!
And yes, a reused string sounds like it should be a constant :)

PR #9282, opened 2023-04-25
Update Japanese.properties

namazulake:


namazulake:
I have resent pull request with correct branch today. Please see newer request.

PR #9284, opened 2023-04-25
Cache civ uniques while refreshing world map tiles.

WhoIsJohannes:
This should have a pretty significant performance improvement of about 10% next round time.

Screenshot from 2023-04-25 19-26-44

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -301,8 +301,23 @@ class LocalUniqueCache(val cache:Boolean = true) {
     // This stores sequences *that iterate directly on a list* - that is, pre-resolved
     private val keyToUniques = HashMap<String, Sequence<Unique>>()
 
+    fun forCityGetMatchingUniques(city: City, uniqueType: UniqueType) : Sequence<Unique> {
+        return forCityGetMatchingUniques(city, uniqueType, StateForConditionals(city.civ, city))
+    }
+    fun forCityGetMatchingUniques(city: City, uniqueType: UniqueType, stateForConditionals: StateForConditionals) : Sequence<Unique> {
+        return get("city-${city.id}-${uniqueType.name}-${stateForConditionals}", city.getMatchingUniques(uniqueType, stateForConditionals))
+    }
+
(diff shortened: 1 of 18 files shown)


WhoIsJohannes:
Hm... I see yield icons flipping back and forth... 1 food one round, 2 food the other round and so on. What am I missing here? Nothing should change while the world map is refreshed right?

WhoIsJohannes:
I found it... A lot of the code using the cache assumed that it is only reusing the cache across a single city, so it didn't matter that the key didn't include the city. But now I'm using the cache across several cities, so that doesn't work anymore. I fixed it now, but it feels ugly. I will check if there's a way to generate the cache key differently, so that this problem can't happen in the first place.

WhoIsJohannes:
Done - I think this is ready for review now.

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -301,8 +301,23 @@ class LocalUniqueCache(val cache:Boolean = true) {
...
     private val keyToUniques = HashMap<String, Sequence<Unique>>()
 
+    fun forCityGetMatchingUniques(city: City, uniqueType: UniqueType) : Sequence<Unique> {
+        return forCityGetMatchingUniques(city, uniqueType, StateForConditionals(city.civ, city))
+    }
+    fun forCityGetMatchingUniques(city: City, uniqueType: UniqueType, stateForConditionals: StateForConditionals) : Sequence<Unique> {
+        return get("city-${city.id}-${uniqueType.name}-${stateForConditionals}", city.getMatchingUniques(uniqueType, stateForConditionals))
+    }
+
+    fun forCivGetMatchingUniques(civ: Civilization, uniqueType: UniqueType) : Sequence<Unique> {
```
yairm210, line 311: state for conditionals should be an optional parameter defaulting to this and then we don't need 2 functions
Same for city functions

yairm210:
Looks great! 👍🏿 
Love the generalizing of unique cache for city/civ!

WhoIsJohannes:
Ah, good catch. For some reason I only had a boolean in mind. Do you mind
sending a PR? Thanks for investigating!

Il mer 26 apr 2023, 10:52 SomeTroglodyte ***@***.***> ha
scritto:

PR #9287, opened 2023-04-26
Allow to reveal explored resources from a city's demanding resources in `CityOverviewTab`

chr56:
## Summary
Allow to reveal explored resources from a city's demanding resources in `CityOverviewTab`.

## Description
Sometimes, A city you owned demands a new Luxury Resource. You want to reveal this resource in map (The notification like `[n] sources of [resourceName] revealed, e.g. near [cityName]`). And, you can make this notification showed by two ways currently: A new Technology has researched with a new resource that can be discovered and showed in map; Or clicking the icon image in `ResourcesOverviewTab`.
Now, I hope to display this notification directly, without jumping to `ResourcesOverviewTab`.
So, I added the callback of demanding resource image to show the notification directly.


chr56:
d47f33b8459bb90d3a9a7f1ae46f87693dea2c9b: suggestion adopted

Squash these commits before merging. :)

Issue #9205, opened 2023-04-17
Feature request: Clearer friendlier Battle strength comparison


SomeTroglodyte:
@Ouaz presented this in an otherwise unrelated [issue](https://github.com/yairm210/Unciv/issues/9186) [discussion](https://github.com/yairm210/Unciv/issues/9186#issuecomment-1510495186) - and suggestion three is way cool. _Quote_:

***Currently:**
![current](https://user-images.githubusercontent.com/11946570/232338777-72958d8b-2a1b-459f-af29-6a731171acc3.png)

**1st suggestion: HP numbers on/off setting** (simplest fix ^^)
A setting in Display > Interface:  `[ ] Show HP numbers in battle table`. If unchecked, no numbers, so people with OCD are happy again.
<img src="https://user-images.githubusercontent.com/11946570/232338841-8da980e9-d04c-4a2e-bd3c-5cc5ccd29c27.png">

**2nd suggestion: Detailed data**
As people finds  `65 -> 35-20` is not a logic way to write the HP (despite the bar above showing exactly that), detailed data is used instead:
![detailed_hp](https://user-images.githubusercontent.com/11946570/232343249-8a5e6242-0f2d-4984-9a6c-a57e43ed6b7a.png)

**3rd suggestion: Big Bang overhaul** (designed by @letstalkaboutdune )
![big_bang](https://user-images.githubusercontent.com/11946570/232341913-5d2b1316-6bd1-46a9-8a55-e8ba059f135b.png)
- Vertical HP bars for direct comparison.
- Simplified text bonus for better readability and more compact battle table.
- Bonus texts are linked to their corresponding Civilopedia articles if the player needs more information.

The 1st suggestion would make my day, if the other ones are too much a hassle to implement. :smiley: 

--- _End-Quote_. 

It was also mentioned that the current version is extremely jarring to certain minorities, and if that can be alleviated with something as nice as that, I'd say yes please. (no guarantee I'll take the bait myself though, so if anyone wants to take a shot, "assign yourself" up there top right)

itanasi:
I don't think it was intentional. It was a mock up and likely just swapped which started full.

letstalkaboutdune:
It's been quite a while, but I'm pretty sure it's just a mockup. No meaning to which side is full vs. not, just an example.

yairm210:
We had this discussion endlessly on Discord, we did a gazillion polls.
People want to see the numbers.
[...]

PR #9291, opened 2023-04-26
New HealthBar Widget

SomeTroglodyte:
Preparation for #9205, so far mostly invisible.

A new Widget written bottom-up, mostly oriented on the Gdx ProgressBar, painting its thing directly from Drawables instead of nesting higher level Widgets. Reduced chance of visual artifacts due to rounding.

What you _do_ see is that the city production progress bars in construction queue / offers are the same as the one on the city button, three-part with next turn prognosis.

The PR is organized in one commit per goal, so you can see the various other implementations of the same thing being replaced one by one.

### Open questions
* The influence bar for city states under the CityButton is left alone, it has the pretty quirk of showing a "gap" in the bar where the relation level changes. I haven't yet had a good elegant idea how to integrate that feature, so that's left for later.
* It uses a Style approach for attributes that usually don't change during a Widget lifetime, like Gdx.Label or TextButton - and there's only a few different ones Unciv needs, _and_ the Widget _can_ share them. But - where to store them? In the skin (and I mean Gdx Skin not the moddability stuff) like the Gdx Widgets do? Easy, works, partially tested, but I think that's not very readable. Or - enum instances. Also partially tested and works easily enough. Or a Map-based cache, or static val's in an object, or or or - So for the moment the Styles are created on the fly. I request ***input*** on which approach would be welcomed best.

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTableHelpers.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTableHelpers.kt
@@ -13,6 +13,7 @@ import com.unciv.UncivGame
 import com.unciv.logic.battle.ICombatant
 import com.unciv.logic.battle.MapUnitCombatant
 import com.unciv.logic.map.HexMath
+import com.unciv.ui.components.HealthBar
 import com.unciv.ui.components.tilegroups.TileSetStrings
 import com.unciv.ui.images.ImageGetter
 import com.unciv.ui.screens.worldscreen.WorldScreen


yairm210:
Using draw() is very clever, but 'clever' isn't what we're going for. We're going for 'easily changable and maintainable', and in that aspect this PR is a big step back.
I want the N+1 developer to be able to edit this UI element using the knowledge that is needed to edit the rest of the UI elements - namely Tables, Groups, Labels and Images. For this reason I have been avoiding using other LibGDX objects like HorizontalGroup, because I want us to retain the minimal set of UI tools.

There are places that achieving what we want is simply not possible given the constraints of those tools, in which case the development concerns take second place to the functional concerns. This is not one of those places.

yairm210:
The Label solution is one that we'll can't solve by simply stacking the existing components together, which is why I'm fine south there being new components

letstalkaboutdune:
> Using draw() is very clever, but 'clever' isn't what we're going for. We're going for 'easily changable and maintainable', and in that aspect this PR is a big step back. I want the N+1 developer to be able to edit this UI element using the knowledge that is needed to edit the rest of the UI elements - namely Tables, Groups, Labels and Images. For this reason I have been avoiding using other LibGDX objects like HorizontalGroup, because I want us to retain the minimal set of UI tools.
> 
> There are places that achieving what we want is simply not possible given the constraints of those tools, in which case the development concerns take second place to the functional concerns. This is not one of those places.

I'm extremely disappointed by this. Seems like shooting yourself in the foot.

letstalkaboutdune:
You know what? My sprites are too hard for novice artists to draw. I'm gonna revoke access to AU for Unciv. Hell, let's just go back to the Minimal tileset and only icons.

letstalkaboutdune:
Imagine being SomeTrog, working hard to implement this and does a great job. Then it all gets dumped in the garbage because it's too "clever." What a slap in the face.

yairm210:
@letstalkaboutdune i think that saying "there's a simpler solution to this exact problem" is a legitimate statement.
Coding is not like art, where each nee person creates an image from scratch. It's built gradually and collaboratively. Once you have a solution that has a high bar of understanding, you're effectively blocking other contributors from working on it.

This is why e.g. the music section went untouched for years. It was complex in a way that anyone looking at it said "nope I'm not touching this".

I think your disappointment is misplaced, and that you're focusing on the wrong thing. Continual improvement is *more important*. It's like, just because someone worked hard on something, therefore we should make the game harder to work on going forward when an alternative path exists? The mind boggles. That's like saying that if someone took twoce as long as you did to create images that were worse, we should replace yours because we should respect the effort put in.

yairm210:
As the repo owner, I have the fun job of setting the bar, and that includes making sure that we can keep improving in the future, something that contributors worry about less.

All this to say, "I respectfully disagree". It's only a slap in the face if you think only about yourself and not about what's good for the project.

letstalkaboutdune:
I'm sorry but that's completely bogus. 

Clever solutions are implemented **_all the time_** in technical work to extract more features and functionality out of tools than what was originally intended. And you know what people do when they implement clever solutions? They **_fucking document them_** so others can understand, they move on with the cool new feature, and everyone's happy. Setting the bar for an implementation so low so that the lowest common denominator can understand it would mean no progress. This is even true with open source projects - just look at emulators like Dolphin who implement lots of advanced rendering techniques and per-game hacks in order to achieve their goals.

Each person creates an image in art from scratch? In what universe is that true? Do you pay any attention at all to how modders work on their mods? That's horribly inefficient and bad practice in art as well as coding, especially in large projects of similar art like a sprite set. Modularity and reuse is king.

@SomeTroglodyte, I'm sorry your work was thrown away like this. I really appreciate you working on this. Can I buy you a beer/coffee/pizza? If you have a Paypal or somewhere I can throw you some digital dollars, please let me know.

yairm210:
Again you seen to be missing the point.
Clever solutions are implemented all the time *when necessary*.
Here the complexity is *unnecessary*.

I think that "reduce complexity where possible" is a simple concept that even the most complex open source projects would agree with. I really don't understand what's gotten your goat about this.

yairm210:
Complexity is a *real cost* for systems support. It's easy for you to advocate adding in complexity, because all you see are the benefits of the feature and not the costs of maintenance. If you were on the maintaining side of this project I think you would have a different perspective.

letstalkaboutdune:
> Again you seen to be missing the point. Clever solutions are implemented all the time _when necessary_. Here the complexity is _unnecessary_.
> 
> I think that "reduce complexity where possible" is a simple concept that even the most complex open source projects would agree with. I really don't understand what's gotten your goat about this.

You're right, how could I possibly be upset by this? It's not like I've ever advocated for this type of BattleTable rework and other UI improvements, debated it endlessly, made mockups, and even taken a break from Discord because I got so frustrated with how UI development was going. Completely shocking.


>That's like saying that if someone took twice as long as you did to create images that were worse, we should replace yours because we should respect the effort put in.

No, that's like if a professional animator discovered Unciv, got excited, made their own beautiful and fully-animated unit sprite set, submitted it to the base game, and you rejected it because it "didn't use freely available tools" or "was too difficult for modders to replicate." Which is actually what I believe you, (or possibly others, some people definitely said it) about AU becoming the default before changing their mind. The truth is that raising the bar... raises the bar.

But whatever, you don't like it. I'm not going to change your mind by bitching here. I wasn't planning on even monitoring or discussing this or any other UI discussion because your mindset boggles my mind in a bad way, but I had a momentary lapse of willpower.

PR #9295, opened 2023-04-27
MapEditor: allow to retract generated steps

chr56:
# Summary
Frequently, you want to undo changes in `Partial (Stepped) Map Generation" (or created map)  in Map Editor, or remove all tiles/TileFeature/TileImprovement of a certain type quickly.

Now, this PR provides a way (currently not perfect) to retract some large changes in MapEditor

# Changes

This PR adds:

- `MapRegression`: like `MapGenerator`, but does the opposite — retract changes
- `MapEditorRetractTab`: A tab in Map Editor, which looks like subtab `MapEditorGenerateStepsTab` in `MapEditorGenerateTab`, provides UI of retracting changes.

## ScreenShot
Screenshotpng

## Known Bugs

- [ ] If user does not retract steps in order of reverted step in normal Generator Steps, may cause some glitchy tile against Terrain Ruleset.(For example, if retrack coast lines before rare terrains, atolls might be left on oceans)
- [ ] `retractNaturalWonders` may not working perfectly.
- [ ] Can not fully revert changes. Since some steps are unretractable, we can only guess what the previous tile might be.
- [ ] Some strings might need to be "refined" and "improved". (So, not "String.tr()"ed)

## Test

**Almost Good**

- Only Vanilla & K&G, no mod is tested.
- Tested on Desktop, no Android.
- Tests look fine for me excepting bugs mentioned above.

# Other Infomation

**Please review my codes carefully.** 
I am new to GDX, and not familiar with this project's code (My first large PR); ~I don't trust and believe myself as well.~

(This should be a draft.)

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapRegression.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapRegression.kt
@@ -1,0 +1,252 @@
+package com.unciv.logic.map.mapgenerator
+
+import com.unciv.Constants
+import com.unciv.logic.map.TileMap
+import com.unciv.logic.map.tile.Tile
+import com.unciv.models.ruleset.Ruleset
+import com.unciv.models.ruleset.tile.TerrainType
+import com.unciv.models.ruleset.unique.UniqueType
+import com.unciv.ui.screens.mapeditorscreen.MapGeneratorSteps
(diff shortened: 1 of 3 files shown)


yairm210:
Been long enough I'm considering this stale, you can reopen if you want

Issue #9098, opened 2023-04-03
Performance review


yairm210:
In the past we've documented a lot of memory improvements a la #7491, #6695, #5919 but we've been less consistent with CPU performance, because those are usually one-offs so the increase is gradual

So looks like it's about time!

I took a relatively large game and recorded traces for 1 minute.

![image](https://user-images.githubusercontent.com/8366208/229450207-5c7537a5-c519-478d-a9b1-17894513251f.png)

What do we see here?

![image](https://user-images.githubusercontent.com/8366208/229450915-5bd23310-841d-4068-bc73-568f4ce46f9f.png)

That should be a boolean. And it is, but that boolean is being retrieved by a function that issues a map lookup - the last terrain only changes if the terrain features change, and we already have those caches and a function for updating, so we can shove 'last terrain' in there as well - 9c121086

![image](https://user-images.githubusercontent.com/8366208/229451158-294157be-b562-4c00-a986-75e17ba85c02.png)

That's getting called a lot, the civ's tiles stay the same most of the time so we don't need to go looking for them (we can cache), and this also means if units move within the territory we can just not update the viewable tiles at all - f50b88c9

![image](https://user-images.githubusercontent.com/8366208/229452163-022bf911-a66a-4305-a0a8-1c338049a01b.png)

Why is *that* not a cached boolean? It is in the type, BUT - getting the type from the base unit is another lookup! One more for caching - f0ee25dc

![image](https://user-images.githubusercontent.com/8366208/229452419-93da0698-514b-4464-9c2a-c401ee6e25b8.png)

Worker automation is thinking a lot about what the stats for the improvement are - that's because it's not using a unique cache - 3db03a78

![image](https://user-images.githubusercontent.com/8366208/229452786-ec5c6f9a-6da0-4a9e-9f10-ebab93584c10.png)

Also not using a unique cache - ad299a8a

![image](https://user-images.githubusercontent.com/8366208/229452939-d2b96dcd-9cb5-4b5c-906d-b71bd042747d.png)

Same - TODO

![image](https://user-images.githubusercontent.com/8366208/229453109-fa1c22c9-1cb1-4295-9b03-dda08d2445b8.png)

What the hell? TODO

yairm210:
![image](https://user-images.githubusercontent.com/8366208/229454951-03d298e2-27c0-40cd-ba51-98f0ff9ff251.png)

Caused by isBuildable, caused by getRejectionReasons.
At least part of the problem is this:
[...]

PR #9296, opened 2023-04-27
Various performance improvements

WhoIsJohannes:
Please see comments on the individual commits. I tried to explain everything there with screenshots from the profiler.

### DIFF
--- a/core/src/com/unciv/logic/map/tile/Tile.kt
+++ b/core/src/com/unciv/logic/map/tile/Tile.kt
@@ -232,7 +232,7 @@ open class Tile : IsPartOfGameInfoSerialization {
     }
 
     fun isExplored(player: Civilization): Boolean {
-        if (DebugUtils.VISIBLE_MAP || player.isSpectator())
+        if (DebugUtils.VISIBLE_MAP || player.civName == Constants.spectator)
             return true
         return exploredBy.contains(player.civName)
     }
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -557,13 +557,22 @@ object UnitAutomation {
                 .firstOrNull { unit.movement.canReach(it.getCenterTile()) }
 
         if (closestReachableEnemyCity != null) {
-            return headTowardsEnemyCity(unit, closestReachableEnemyCity.getCenterTile())
+            return headTowardsEnemyCity(
+                unit,
+                closestReachableEnemyCity.getCenterTile(),
+                // This should be cached after the `canReach` call above.
+                unit.movement.getShortestPath(closestReachableEnemyCity.getCenterTile())
+            )
         }
         return false
     }
 
 
-    private fun headTowardsEnemyCity(unit: MapUnit, closestReachableEnemyCity: Tile): Boolean {
+    private fun headTowardsEnemyCity(
+        unit: MapUnit,
+        closestReachableEnemyCity: Tile,
+        shortestPath: List<Tile>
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -232,7 +232,7 @@ open class Tile : IsPartOfGameInfoSerialization {
     }
 
     fun isExplored(player: Civilization): Boolean {
-        if (DebugUtils.VISIBLE_MAP || player.isSpectator())
+        if (DebugUtils.VISIBLE_MAP || player.civName == Constants.spectator)
```
yairm210, line 235: This is actually worse performance wise, no?

WhoIsJohannes:
It's called like 10 million times just in one next turn evaluation for me.
I also suspect the lazy and just objects being dereferenced / passed around.

Il gio 27 apr 2023, 17:50 SomeTroglodyte ***@***.***> ha
scritto:

yairm210:
@SomeTroglodyte This is still a much smaller performance hit than string comparisons - that's why those lazies exist in the first place, as a result of performance debugging (#9098)

Doing `val x = y` won't work with json-unpacked classes, since they are initially empty and are only filled on json loading.
Do you have any suggestion for an alternative?

WhoIsJohannes:
Thanks for all the brain power put into this. Have you guys seen my
screenshots? It's definitely faster this way, believe it or not 😋.
Unfortunately they got lost a bit, only through clicking on some comment
icon somewhere will bring them up since I put them before sending the PR.

Il sab 29 apr 2023, 21:30 SomeTroglodyte ***@***.***> ha
scritto:

yairm210:
There's actually a simple solution for this use case - since the isSpectator is in the *civ* class and not the *nation* class, we can use a var bool in the civ class, and set it in setTransients, this avoiding the problem entirely :D

yairm210:
This is another one of those 'too much going on' PRs, if you split it into different PRs then we can merge the simple fixes and argue about the contentious ones :)

yairm210:
Relevant to the lazy evaluation (search for "negative latency"):
https://apenwarr.ca/log/20230415



Tldr of the rest of the post:
- The hard part of science is generating testable hypotheses, not the actual testing
- Understanding the underlying components of a system *does not* mean you understand the system
- Engineering is "getting stuff done passably", not perfectly

- To get things done efficiently, to do real engineering, sometimes you need to *guess stuff* (hypothesis) and then test it, and that requires *insight* or system design understanding. The open question is, will AIs be able to do that

WhoIsJohannes:
How about this: I revert the lazy part and we figure it out separately. Is
that the only part that we're discussing at this point?

Il dom 30 apr 2023, 17:34 SomeTroglodyte ***@***.***> ha
scritto:

yairm210:
I'm also not 100% sure that the movement improvements don't change anything but there's so much going on here that I find it hard to review

yairm210:
I am willing to go the other way though - let's merge this and then see what happens/if we can improve on it 😜

WhoIsJohannes:
Thanks! I was hoping the individual commits would help, but I get it. I'm
always a bit worried building PRs on top of each other because I don't want
to get into merging hell... And I somehow need to run everything together
locally. Anyways, for now I'm done with performance I think and I'm not
planning to do much any time soon. We're travelling for a long time and
I'll have to start my real job after that. Also there isn't so much right
now I really think needs doing (other than AI, but that is not something I
could do quickly).

Il lun 1 mag 2023, 06:35 Yair Morgenstern ***@***.***> ha
scritto:

yairm210:
Good luck!

PR #9306, opened 2023-05-02
Allow closing a Popup by clicking outside its area

### LINKED ISSUE #9300: UX: Close popups on click on non-popup area
If a user opens a popup like the one shown below, it doesn't close it by clicking on the area around it (i.e., non-popup area). It would be better user experience if that would happen, I think.

Demo-Popup

Maybe it's also a nice idea to add a switch to the `Popup` class that determines whether this close handling should be enabled for a given popup, so that there could be a difference between required dialogs and optional ones.

SomeTroglodyte:
Closes #9300

Needs more testing - and the careful evaluation why we haven't done this before. There was some discussion already ages ago I vaguely remember...

Rationale regarding the linting of unused methods: One is an API offer (but may likely go when all code close to the screen only speaks KeyBinding), the other was seemingly made redundant by the KeyShortcut project, and said it had a specific use.

### DIFF
--- a/core/src/com/unciv/ui/components/TabbedPager.kt
+++ b/core/src/com/unciv/ui/components/TabbedPager.kt
@@ -621,6 +621,8 @@ open class TabbedPager(
                 addOKButton {
                     if (passEntry.text.hashCode() == secretHashCode) unlockAction() else lockAction()
                 }
+                clickBehindToClose = true
+                onCloseCallback = lockAction
                 this.keyboardFocus = passEntry
             }
         }
--- a/core/src/com/unciv/ui/popups/Popup.kt
+++ b/core/src/com/unciv/ui/popups/Popup.kt
@@ -3,6 +3,7 @@ package com.unciv.ui.popups
 import com.badlogic.gdx.graphics.Color
 import com.badlogic.gdx.math.Rectangle
 import com.badlogic.gdx.scenes.scene2d.Actor
+import com.badlogic.gdx.scenes.scene2d.InputEvent
 import com.badlogic.gdx.scenes.scene2d.Stage
 import com.badlogic.gdx.scenes.scene2d.Touchable
 import com.badlogic.gdx.scenes.scene2d.ui.Button
@@ -12,6 +13,7 @@ import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.scenes.scene2d.ui.TextButton
 import com.badlogic.gdx.scenes.scene2d.ui.TextButton.TextButtonStyle
 import com.badlogic.gdx.scenes.scene2d.ui.TextField
+import com.badlogic.gdx.scenes.scene2d.utils.ClickListener
 import com.badlogic.gdx.utils.Align
 import com.unciv.Constants
 import com.unciv.logic.event.EventBus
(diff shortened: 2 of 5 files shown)


CrsiX:
From my perspective, the changes look good. It works on my machine (Linux) and my phone in almost all cases and also with multiple popups stacked above each other. There's just one case where it's not working; but since that case is a popup I've created, I will review it for myself before further discussion.

Additionally, shouldn't I also be able to close any popup with `ESC` key just the same way?

----

That mechanic doesn't work on the `NationPickerPopup`, since it has no `Close` button. This was your intention when filtering for close buttons, but is that intention what a user would expect? I mean, I see the point: it's easy to filter whether a popup _should_ be allowed to be closed by that button's existence, e.g. filtering out `ToastPopup`s. But I guess adding that mechanic there would be user-friendlier. Patch:

```
diff --git a/core/src/com/unciv/ui/screens/newgamescreen/PlayerPickerTable.kt b/core/src/com/unciv/ui/screens/newgamescreen/PlayerPickerTable.kt
index b667a8b86..7b091dafb 100644
--- a/core/src/com/unciv/ui/screens/newgamescreen/PlayerPickerTable.kt
+++ b/core/src/com/unciv/ui/screens/newgamescreen/PlayerPickerTable.kt
@@ -471,6 +471,7 @@ class NationPickerPopup(
         closeButton.keyShortcuts.add(KeyCharAndCode.BACK)
         closeButton.setPosition(buttonsOffsetFromEdge, buttonsOffsetFromEdge, Align.bottomLeft)
         innerTable.addActor(closeButton)
+        hasCloseButton = true
 
         val okButton = "OtherIcons/Checkmark".toImageButton(Color.LIME)
         okButton.onClick { returnSelected() }
```

See below image, there's plenty of space on the right and left side of the popup, that could be clicked. IMHO, I would expect it to close that popup as well:
NationPickerPopup

CrsiX:
> Curious cat here

It was in my `RegisterLoginPopup`'s `askConfirmUsage`. See [here](https://github.com/hopfenspace/Unciv/blob/a7dd017b13f0dab116c3efeb9045e02392fa3ee5/core/src/com/unciv/ui/popups/RegisterLoginPopup.kt#L119) for code. It contained the `addButton(Constants.OK)`, but when replacing with `addCloseButton`, it gave me some null pointers. It's now fixed, even though not committed yet.

> What, that's not working?

Actually, I might have mixed it up with the `RegisterLoginPopup` from above. Doing the tests now again, it simply works™.

> Found an exotic candidate that's also better with custom enabler. Question - would anyone ever expect a click-behind to mean apply/affirmative instead of discard/negative???

I don't think so. I mean, I didn't accepted some change, so why would it be applying instead of discarding?

----

Looking at your update, maybe it's also good to add a public method like `enableCloseByContextClick` (or, please come up with a better name), that just sets `clickBehindToClose = true`. It would allow to use that feature for local popups, that are no subclasses, i.e. in cases like `val popup = Popup(); ... ; popup.open()`.

CrsiX:
> No, after a little consideration, I'll leave the clickBehindToClose as is.

Fine for me. It's documented in the current form and as part of the public API of `Popup`, is it?

> Excellent input, @CrsiX , muito obrigado.

Thank you, you're welcome :)

CrsiX:
> Huh - that works with no side effects?

:D

> block.invoke() === block();;; and cell.actor.keyShortcuts.add(KeyCharAndCode.ESC) should still be redundant if you do the same for BACK anyway

Well, I probably need to take a second look at something, thanks for the comment :)

yairm210:
Looks good!

PR #9314, opened 2023-05-02
Add `May not annex cities` unique

Skekdog:
Following behaviour of Civ5 Venice. Do tell if I've messed something up, though for once it does look all clear.


yairm210:
I thought Venice couldn't puppet, either?

Skekdog:
> I thought Venice couldn't puppet, either?

No, I took a look and the rules were that Venice can puppet and raze without annexing

Skekdog:
It does turn out that Venice is unable to raze cities after choosing to create a puppet, though, which I will add a commit to correct. (And apparently 3 tanks can't defeat a prehistorical settlement). Pardon the video quality

https://user-images.githubusercontent.com/47750815/235954384-eb8a75d0-cfd5-47bb-8b94-5808dc47741c.mp4

Skekdog:
> Summary - will the AI be able to handle this and can I imagine modders attaching unit- or city-based conditionals...

I'll take a look tomorrow, thanks for the review.

core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
```diff
@@ -146,24 +147,25 @@ class AlertPopup(
...
-            addAnnexOption {
+            val mayAnnex = !conqueringCiv.getMatchingUniques(UniqueType.MayNotAnnexCities).any()
+            addAnnexOption(mayAnnex = mayAnnex) {
                 city.puppetCity(conqueringCiv)
                 city.annexCity()
                 worldScreen.shouldUpdate = true
                 close()
             }
-            addSeparator()
+                addSeparator()
```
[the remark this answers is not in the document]
Skekdog: Yeah I'm not sure when that got there, my bad

core/src/com/unciv/logic/city/managers/CityInfoConquestFunctions.kt
```diff
@@ -315,6 +315,13 @@ class CityInfoConquestFunctions(val city: City){
 
             if (civ.gameInfo.isReligionEnabled()) religion.removeUnknownPantheons()
 
+            if (newCiv.getMatchingUniques(UniqueType.MayNotAnnexCities).any()) {
```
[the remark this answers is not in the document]
Skekdog: I was not aware of hasUnique. As for Conditionals, I was thinking of implementing them sometime later (as I don't really see where they might be used) though I suppose it would be better to do it all at once.

As for AI - well, I completely forgot about that. My bad.

Skekdog:
So in Civ5, Venice is stuck with puppet cities if they lose their capital... This seems pretty dumb (although Venice can faith/gold purchase in puppets) so I think we should annex one of their puppets - not matching civ5 but this wouldn't be the first difference

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
```diff
@@ -259,7 +259,8 @@ class CityScreen(
             addWltkIcon("OtherIcons/WLTK 1") { color = Color.FIREBRICK }.padRight(10f)
         }
 
-        if (city.isPuppet) {
+        val canAnnex = !city.civ.getMatchingUniques(UniqueType.MayNotAnnexCities).any()
```
yairm210, line 262: here we can use hasUnique

core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
```diff
@@ -239,13 +241,13 @@ class AlertPopup(
                 close()
             }
         } else {
-            addAnnexOption {
+            addAnnexOption(marryingCiv.getMatchingUniques(UniqueType.MayNotAnnexCities).any()) {
```
yairm210, line 244: and here

core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
```diff
@@ -239,13 +241,13 @@ class AlertPopup(
...
                 city.annexCity()
                 close()
             }
             addSeparator()
 
-            addPuppetOption {
+            addPuppetOption(marryingCiv.getMatchingUniques(UniqueType.MayNotAnnexCities).any()) {
```
yairm210, line 250: and here

yairm210 (commented):
All in all, looks pretty good, I only have this minor point to make

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
Skekdog: 👍
They've all been replaced locally, am yet to commit

yairm210: Reminder :)

Skekdog: Oops turns out I didn't replace them locally

PR #9322, opened 2023-05-04
ci: added Detekt code analysis

alexban011:
~So before actually setting up the config file I think I should get some input/feedback. [Detekt](https://github.com/detekt/detekt) is a code analysis tool very similar to CodeSonar. Right now the actual report generation doesn't work for some reason (I haven't looked into that though) so the warnings are only shown in the actions page. HTML/MD generation should be more effective than reading through 1400 lines :)~

~The config is almost unchanged. I only disabled MaximumLineLength and MagicNumber which my guess is that they're not needed and got the report from 7400 warnings to 1400. The config still needs work, but again, I'd like some input before I keep going.~

The above text is no longer relevant. Read this instead: https://github.com/yairm210/Unciv/pull/9322#issuecomment-1540802410

### DIFF
--- a/.github/workflows/detekt.yml
+++ b/.github/workflows/detekt.yml
@@ -1,0 +1,35 @@
+name: Detekt
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  detekt:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Set up JDK
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+
+    - name: Setup Detekt
+      id: setup_detekt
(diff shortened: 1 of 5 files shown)


alexban011:
The CI fails at the moment because I kept de default of 0 max issues. We would most likely set that to something big so it doesn't instantly fail the test

alexban011:
> I guess it's this

Yes

alexban011:
[reports.zip](https://github.com/yairm210/Unciv/files/11401754/reports.zip)

alexban011:
This is the last report as of [f1b6a9e](https://github.com/yairm210/Unciv/pull/9322/commits/f1b6a9e830d25767bcd908f3eb951f95ff14fbb7): [reports.zip](https://github.com/yairm210/Unciv/files/11435583/reports.zip)

yairm210:
If we can make it error on the slecific warnings that you have already fixed, to avoid them in the future, that would be amazing 😍

alexban011:
> If we can make it error on the slecific warnings that you have already fixed, to avoid them in the future, that would be amazing heart_eyes

Well we can set a severity level on each specific warning: https://detekt.dev/docs/introduction/reporting#severity

PR #9327, opened 2023-05-05
Prevent some state changing actions during next-turn

SomeTroglodyte:
* Right-click to move is possible while waiting for next turn
* ..and upgrading units through the overview units tab.

Of course, these changes are undone once processing finishes, but still, let's fix them.


core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
```diff
@@ -206,12 +207,10 @@ class WorldMapHolder(
     }
 
     private fun onTileRightClicked(unit: MapUnit, tile: Tile) {
-        if (UncivGame.Current.gameInfo!!.getCurrentPlayerCivilization().isSpectator()) {
-            return
-        }
         removeUnitActionOverlay()
         selectedTile = tile
         unitMovementPaths.clear()
+        if (!GUI.isAllowedChangeState()) return
```
yairm210, line 213: This should definitely use the worldscreen and not the static - we're trying to avoid that as much as possible

yairm210: Reminder :)

PR #9364, opened 2023-05-10
Fixed a couple warnings reported by code analysis

alexban011:
Reviewing commit-by-commit is strongly recommended


yairm210:
Looks good!
There are several places where we can add clarity by reducing linecount, but that's for another day ;)

core/src/com/unciv/ui/screens/overviewscreen/WonderOverviewTable.kt
```diff
@@ -44,7 +44,8 @@ class WonderOverviewTab(
 
         top()
         defaults().pad(10f).align(Align.center)
-        (1..5).forEach { _ -> add() }  // dummies so equalizeColumns can work because the first grid cell is colspan(5)
+        for (i in 1..5)
```
yairm210, line 47: HAH so your linter agrees with me that for > forEach! #validated

alexban011: Yep. So does this: https://sites.google.com/a/athaydes.com/renato-athaydes/posts/kotlinshiddencosts-benchmarks

`The above graph shows that using the forEach function for a Range, as the author predicted, should absolutely be avoided. It performs 300% slower than a simple for-loop!`

Issue #9367, opened 2023-05-10
4.6.10-p2 can no longer load games from <=4.6.9


SomeTroglodyte:
Example: `barbarians:{camps:{class:com.unciv.json.HashMapVector2,entries:[[{x:-1,y:15},{class:com.unciv.logic.Encampment,position:{x:-1,y:15}}],[{x:-3,y:-19},{class:com.unciv.logic.Encampment,position:{x:-3,y:-19}}]]}},` - but that class was moved to package `com.unciv.logic.automation.civilization`, I believe in #9364

I _could_ rewrite the custom deserializer so it accepts both formats, but at my current performance that would take a few days...

Effect: 
![image](https://github.com/yairm210/Unciv/assets/63000004/18306fb7-8e90-4b20-a69b-6ce812d96417)

The wording is quite misleading - that version marked the introduction of `GameInfoSerializationVersion`, and was not the version my game was saved with...

PR #9370, opened 2023-05-10
Fix 4.6.10 no longer loading older games

SomeTroglodyte:
For #9367 the quickest and dirtiest hotfix


yairm210:
Why do we even need a hashmap when serializing? Each encampment contains the location anyway

We can have the hashmap as a transient field if we really need it (maybe not!) and save the rest of this things as a normal list

PR #9379, opened 2023-05-12
Multi-server preparations: Save server info in game, save server functionality in specific class

yairm210:
@touhidurrr @CrsiX 
To allow a single player to be in games from several servers at once, some background preparation is required

This PR saves the URL of the game server in the game parameters on which the game was created, and moves server-specific functionality (check liveness, authenticate) to the server level

This PR does *not* allow players to 'juggle' between different servers.

The work required to get there is:
- UI work to allow users to handle a 'list' of known servers (add/remove/set default)
- UI work to pick a server when starting a new multiplayer game
- Updating and checking several servers in parallel on game startup

### DIFF
--- a/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
+++ b/android/src/com/unciv/app/MultiplayerTurnCheckWorker.kt
@@ -29,7 +29,7 @@ import com.badlogic.gdx.backends.android.DefaultAndroidFiles
 import com.unciv.logic.GameInfo
 import com.unciv.logic.files.UncivFiles
 import com.unciv.logic.multiplayer.storage.FileStorageRateLimitReached
-import com.unciv.logic.multiplayer.storage.OnlineMultiplayerFiles
+import com.unciv.logic.multiplayer.storage.OnlineMultiplayerServer
 import com.unciv.models.metadata.GameSettingsMultiplayer
 import kotlinx.coroutines.runBlocking
 import java.io.FileNotFoundException
@@ -307,7 +307,7 @@ class MultiplayerTurnCheckWorker(appContext: Context, workerParams: WorkerParame
 
                 try {
                     Log.d(LOG_TAG, "doWork download $gameId")
-                    val gamePreview = OnlineMultiplayerFiles(fileStorage, mapOf("Authorization" to authHeader)).tryDownloadGamePreview(gameId)
+                    val gamePreview = OnlineMultiplayerServer(fileStorage, mapOf("Authorization" to authHeader)).tryDownloadGamePreview(gameId)
                     Log.d(LOG_TAG, "doWork download $gameId done")
                     val currentTurnPlayer = gamePreview.getCivilization(gamePreview.currentPlayer)
 
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -181,7 +181,7 @@ open class UncivGame(val isConsoleMode: Boolean = false) : Game(), PlatformSpeci
         Concurrency.run {
             // Check if the server is available in case the feature set has changed
             try {
-                onlineMultiplayer.checkServerStatus()
+                onlineMultiplayer.multiplayerServer.checkServerStatus()
             } catch (ex: Exception) {
                 debug("Couldn't connect to server: " + ex.message)
             }
--- a/core/src/com/unciv/logic/multiplayer/OnlineMultiplayer.kt
+++ b/core/src/com/unciv/logic/multiplayer/OnlineMultiplayer.kt
@@ -3,7 +3,6 @@ package com.unciv.logic.multiplayer
 import com.badlogic.gdx.files.FileHandle
 import com.unciv.Constants
 import com.unciv.UncivGame
-import com.unciv.json.json
 import com.unciv.logic.GameInfo
 import com.unciv.logic.GameInfoPreview
 import com.unciv.logic.civilization.NotificationCategory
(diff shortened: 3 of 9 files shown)


CrsiX:
TL;DR: Looks fine :)

This is a good step forward to address multi-server settings. However, I think most users will usually stick to one server when they have it configured, if the UI is not perfectly accessible. Nevertheless, a good start.

I want to point out that in my upcoming APIv2 implementation, storing the server URL in the `OnlineMultiplayerServer` is not necessary anymore, since the game lists are stored on the server -- so, if you play at server A, you will only see games stored on server A; and when playing on server B, you will only see games stored on server B. By querying the server for the list of games, the user doesn't need to store any of the details locally.
Of course, for APIv1 games, this is fine and allows for greater flexibility.

Regarding the three additional tasks:

1. I would place this server list in a more "unpopular" place, like the multiplayer settings tab. I wouldn't like to see the server list every time the user wants to create a new multiplayer game or access an existing one.
2. This should only be necessary when there's no configured default server. See point 1: changing via settings menu. So, at the very first game start, you may ask for a multiplayer server, but stick to the decision later on.
3. Well, when we already change the "ping" behavior (`isalive`), then we could also address a little privacy issue: if the game starts, it currently phones "home" to the configured server. I would like that to happen _only_ if I play a multiplayer game / access any multiplayer-related resources. Additionally, only the configured default server should be queried; the other servers should be queried only when accessed (e.g. click on "check connectivity"). But well, this is more text for an issue than a PR :D

yairm210:
A. If we support multiple servers, we can query all of them even in V2
B. Maybe I didn't explain myself well. The management of servers will be in options - multiplayer, as it is currently. The *usage* of servers will be when starting a new game. And there we absolutely need to have a dropdown. It's no worse than the other dozen configurables that people usually leave at the default. Not allowing that defeats the whole purpose of multi-server.

CrsiX:
> If we support multiple servers, we can query all of them even in V2

Yes, but note that V2 servers _require_ the use of username/password accounts. Meaning that a user might need to register before accessing a server never played on before when selecting it in that drop-down menu (of course, that won't be much of a problem, I guess).

> The _usage_ of servers will be when starting a new game.

Oh, I indeed misunderstood your point. Then, I agree with you. :)

touhidurrr:
So, what if one server has multiple URLs?

CrsiX:
There's no way you could tell that they are the same server. Even if they share the same domain name, IP address and you can use the same accounts, you can't be sure. Therefore, you always need to differentiate between them, i.e. treat them as two different servers.
The only exception I can think of would be `http://example.com` and `https://example.com`: the protocol scheme shouldn't matter.

touhidurrr:
> There's no way you could tell that they are the same server. Even if they share the same domain name, IP address and you can use the same accounts, you can't be sure. Therefore, you always need to differentiate between them, i.e. treat them as two different servers. The only exception I can think of would be `http://example.com` and `https://example.com`: the protocol scheme shouldn't matter.

But it happens to be that I use tons of servers that share the same content but has different endpoints and sync with each other in real-time. At present there are 2 such servers running, the main Heroku server and the backup Replit server. `replit.uncivserver.xyz` is the replit server hostname when any other hostnames like `uncivserver.xyz` and `*.uncivserver.xyz` will serve games from the heroku server. There is also an endpoint that uses different SSL certificates for SSL failures.

As such, since you want servers to be selectable from a global list, I would want these lists to be nested if possible. And different servers using a unique server-id maybe? like `UncivServer,xyz` for any `uncivserver.xyz` urls?
I don't see why not as I saw issues making users unable to access some servers in many times. Chinese players even tell me that my server is blocked by the great firewall for some unknown reason. Which is hilarious but again a real-life scenario.

So, it would be highly requested by me to allow multiple URLs.

touhidurrr:
Also, another question, @yairm210 @CrsiX. Are game previews deprecated on v2 or not? Personally, I would like them to disappear in v2 as they don't feel necessary with WebSockets.

CrsiX:
> Are game previews deprecated on v2 or not? 

They are. We're using a `data_id` which is a simple integer incremented every time someone uploads a new game state. If your local `data_id` is lower than the one of the server, you need to re-fetch the game. Simple and efficient, besides the automatic updates via WebSocket.

> So, it would be highly requested by me to allow multiple URLs.

I think I misunderstood you. Essentially, you want to have a single server with multiple, equal URLs. Say, `a.example.com` and `b.example.com`, which both serve the exact same content, do you? Then, how exactly would you choose the URL you want to connect to, if you select `example.com` as your game server? Stick to the base URL first, but fallback to try all the other available subdomains? This probably needs some _sane_ mechanism, but I can't think of a good one besides trying them out one by one or letting the user choose. However, if you let the user choose, then they could simply choose to use `a.example.com` instead of `example.com` in the first place, I think. I'm not sure about that matter at the moment.

> Chinese players even tell me that my server is blocked by the great firewall for some unknown reason.

Well, this will probably always be a problem. For example, our V2 server is hosted in Germany, probably also blocked. Maybe we need to host it in China to make it available for those players in the future?

touhidurrr:
> Then, how exactly would you choose the URL you want to connect to, if you select example.com as your game server?

@CrsiX the plan is to let the users choose.
Nested option is the idea. Example:
```
UncivServer,xyz <- title
A Popular Multiplayer Server for Unciv <- description
Choose Server: <- appears if a server has multiple urls
1. uncivserver.xyz <- url 1
   the main server. choose it for best performance. <- description 1
2. replit.uncivserver.xyz <- url 1
   the backup server. choose it if the main server is not working. degraded performance. <- description 1
```
Something like this

CrsiX:
Understood, thanks for the details :)

Well, I guess that requires primarily UI changes. In the logic, a single server URL will be passed around, no matter how it got there. Only for some UI element to choose that URL, this changes become relevant.

touhidurrr:
> Well, I guess that requires primarily UI changes.

@CrsiX Also the game would require to recognise `replit.uncivserver.xyz` and `uncivserver.xyz` for example as the part of the same server.

PR #9380, opened 2023-05-12
ci: improved detekt configuration and ci

alexban011:
it still needs some work but it's getting there

~And as for [this comment](https://github.com/yairm210/Unciv/pull/9322#issuecomment-1541598909) I'm still looking into this. It can be done by creating a custom rule, but I hope I'll find an easier solution~

For review read this: https://github.com/yairm210/Unciv/pull/9380#issuecomment-1546631397

### DIFF
--- a/.github/workflows/detektAnalysis.yml
+++ b/.github/workflows/detektAnalysis.yml
@@ -32,4 +32,6 @@ jobs:
         detekt_version: '1.23.0-RC3'
 
     - name: Run Detekt
-      run: detekt-cli --auto-correct --parallel --config detekt/config/detekt.yml
+      run: detekt-cli
+      with:
+          args: --parallel --config detekt/config/detekt.yml
--- a/.github/workflows/detektAnalysis.yml
+++ b/.github/workflows/detektAnalysis.yml
@@ -31,7 +31,9 @@ jobs:
       with:
         detekt_version: '1.23.0-RC3'
 
-    - name: Run Detekt
-      run: detekt-cli
-      with:
-          args: --parallel --config detekt/config/detekt.yml
+    - name: Detekt errors
+      run: detekt-cli --parallel --config detekt/config/detekt-errors.yml
+
+    - name: Detekt warnings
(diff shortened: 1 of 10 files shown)


alexban011:
> Thought about battling the (35 instances) abysmally vile and evil `import java.util.*`? 

Yes. The WildcardImports rule excludes java.util.* from the check by default. Tomorrow I'll remove the exclude and it should report those as well

alexban011:
I'm not quite sure how the .editorconfig works, but I think it'll complement. Since it's used during writing of the code it should try to prevent these right? In this case Detekt will only report them in case they slip through

alexban011:
Ok so it looks like detekt doesn't currently have a system to make it fail on specific warnings and instead treats all warnings as errors. The maintainer suggested a workaround where we have 2 config files.

The first config file has 1000 maxIssues: currently there are around 480 so this one won't fail. 
The second config file (detekt-errors.yml) is pretty much a copy of the first one but:
1. maxIssues is set to 0, so if any issue is detected the analysis will fail and report the issues.
2. everything is disabled besides a couple of the `style` warnings, most notably WildcardImports (including `java.util.*`). Whenever we want to "ban" other issues, we can enable them in the detekt-errors.yml file and disable them from detekt-warnings.yml (so they're not being reported twice)

Then the CI runs detekt-cli twice, one for the errors and one for the rest of the warnings

alexban011:
> /home/runner/work/Unciv/Unciv/core/src/com/unciv/logic/GameInfo.kt:39:1: java.util.* is a wildcard import. Replace it with fully qualified imports. [WildcardImport]

Very interesting. It looks like it sees a star import but thee isn't any..? Maybe .editorconfig and detekt don't complement each other after all

alexban011:
Thanks! I guess I should have thought about doing a rebase before blaming .editorconfig :))

alexban011:
>  I for my part avoid rebase like the plague except for local branches that have never been pushed

I agree but I didn't want the CI to fail so I rebased just to show that the detekt CI actually works as intended :)

PR #9382, opened 2023-05-12
Combat java.util.star-import

SomeTroglodyte:
Evil-be-gone. In a way pure linting, but since it replaces quite a few direct java library uses with their kotlin counterparts the result is theoretically not 100% identical. Still, I say if kotlin has a Random we shouldn't bypass it, same for ArrayDeque or the collection typealiases. Additional benefit - we can clearly see and search on which specific java api's we are actually depending on, just in case we ever need to avoid that.


yairm210:
I'm not sure what you mean by the indent use :/
In any case this PR is good, we can deal with that if necessary in another PR

yairm210:
Oh
Doesn't make a difference to me either way either
So we can leave as is

PR #9387, opened 2023-05-14
Animate battle damage numbers

### LINKED ISSUE #9353: Attack animations
https://user-images.githubusercontent.com/63000004/236915684-9905d1f5-d448-47f0-8334-3afb7e3e8fce.mp4

Did that Carrier just attack? Nope, the Bomber on it did. Which is also why its HP doesn't go down despite the red flash.

Questions...
* I suppose in `BattleTableHelpers.WorldScreen.battleAnimation.getMapActorsForCombatant` changing [line 100/101](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTableHelpers.kt#L100) to `} else if (!combatant.isAirUnit()) {` would fix that - but is that the right approach?
* [`val slot = if (combatant.isCivilian()) 0 else 1`](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTableHelpers.kt#L102) - class TileLayerUnitArt makes no guarantee which slot is used for what -> fragile code merits at least a few comments?
* [`private fun getAttackAnimationLocation`](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTableHelpers.kt#L71) - `unitSpecificAttackAnimationLocation` and `unitTypeAttackAnimationLocation` use the exact same formula. Is the first one meant to use `attacker.unit.name` instead of `attacker.getUnitType().name`?
* Are there any such attack animations in existence to test?
* We won't be getting actual Unit art :small_airplane: for air units anytime soon, right? Animating one in a nice "stuka run" curve might be macabre :ghost: phun... :zany_face: 
* Any reason why [`}.mapTo(arrayListOf()) { it to it.color.cpy() }.toMap()`](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTableHelpers.kt#L111) and not `}.associateWith { it.color.cpy() }`?

SomeTroglodyte:
Closes #9353 - <details><summary>clip</summary>

https://github.com/yairm210/Unciv/assets/63000004/fb58fb0c-0e73-4d7d-9340-e1dae78a6efc
</details>

I've left some tasks untouched - unify right-click code with BattleTable attack, or scaling the labels to world zoom. There's unrelated linting and comments, side effects of exploring possible approaches earls on, e.g. inconsistencies between UncivTooltip and UncivSlider - I checked what I did for these to see what's reusable and that constant was just too inconsistent. The attack animation question I've decided by guess, as I still do not know of a mod with images to test. And the issue with both damage numbers playing at the same location I've dealt with the simplest way possible - if that would be the case, the number for the dead defender is omitted.

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -83,14 +83,16 @@ object Battle {
     /**
      * This is meant to be called only after all prerequisite checks have been done.
      */
-    fun attackOrNuke(attacker: ICombatant, attackableTile: AttackableTile) {
-        if (attacker is MapUnitCombatant && attacker.unit.baseUnit.isNuclearWeapon())
+    fun attackOrNuke(attacker: ICombatant, attackableTile: AttackableTile): DamageDealt {
+        return if (attacker is MapUnitCombatant && attacker.unit.baseUnit.isNuclearWeapon()) {
             NUKE(attacker, attackableTile.tileToAttack)
-        else
+            DamageDealt.None
+        } else {
             attack(attacker, getMapCombatantOfTile(attackableTile.tileToAttack)!!)
+        }
     }
 
-    fun attack(attacker: ICombatant, defender: ICombatant) {
+    fun attack(attacker: ICombatant, defender: ICombatant): DamageDealt {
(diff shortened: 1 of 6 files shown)


yairm210:
YEAH

PR #9395, opened 2023-05-15
Happiness change from bought buildings can reapply citizen focus

### LINKED ISSUE #9238: Cities start to starve while hapiness go up
**Platform**
Windows 11 64bit

**Version**
Version 4.6.4-patch2 (Build 854)

**Describe the bug**
At the start of a turn, the hapiness is negative (-12) but no city is starving. At the end of the turn, after just buying a courthouse (in a city that has just been bought), 7 cities are starving while the hapiness is positive (+3).

**To Reproduce**
1. Buy a city.
2. On the turn you acquire the city, you must have a negative hapiness.
3. Buy a courthouse in the recently bought city.
4. See that the cities are now starving.
Note: the next turn, the cities are still starving.

**Expected behavior**
The cities should not starve.

**Screenshots**

<details><summary> Screenshot </summary>

image

</details>

SomeTroglodyte:
Closes #9238 - I cant believe I overlooked this branch sitting around for so long.

I still have the feeling this could/should be optimized to mitigate the extra perfomance cost, but I can't think of a straighforward way.

Also makes the Courthouse match the "Happiness" building filter which it didnt before - for mods a potential breaking change. What do you think?

### DIFF
--- a/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
@@ -314,7 +314,7 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
 
     private fun addHappinessBuildingChoice() {
         val happinessBuilding = nonWonders
-            .filter { (it.isStatRelated(Stat.Happiness)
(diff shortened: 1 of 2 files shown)


yairm210:
I think that's fine

PR #9401, opened 2023-05-17
Fixed more warnings

alexban011:
Last report: [reports.zip](https://github.com/yairm210/Unciv/files/11502648/reports.zip)


yairm210:
Conflicts

core/src/com/unciv/logic/BackwardCompatibility.kt
```diff
@@ -156,9 +156,8 @@ object BackwardCompatibility {
     }
 
     /** Move max XP from barbarians to new home */
-    @Suppress("DEPRECATION")
-    fun ModOptions.updateDeprecations() {
-    }
+    @Suppress("DEPRECATION", "EmptyFunctionBlock")
+    fun ModOptions.updateDeprecations() { }
```
alexban011, line 160: Why exactly are we calling this function?

tests/src/com/unciv/dev/FasterUIDevelopment.kt
```diff
@@ -41,6 +41,7 @@ object FasterUIDevelopment {
             actor = "This could be your UI element in development!".toLabel()
         }
 
+        @Suppress("EmptyFunctionBlock")
         fun afterAdd() {
```
alexban011, line 45: same here

core/src/com/unciv/ui/screens/victoryscreen/ReplayMap.kt
```diff
@@ -47,6 +47,7 @@ abstract class IndependentMiniMap(
     protected open fun includeTileFilter(tile: Tile): Boolean = true
 
     private fun createReplayMap(tileSize: Float): List<MinimapTile> {
+        @Suppress("EmptyFunctionBlock")
         val doNothing = fun(){}
```
alexban011, line 51: same here. MinimapTile(tile, tileSize, doNothing) could just be instantiated without doNothing right?

core/src/com/unciv/logic/civilization/ExploredRegion.kt
```diff
@@ -12,7 +12,7 @@ import com.unciv.ui.components.tilegroups.TileGroupMap
 import kotlin.math.abs
 import kotlin.math.sqrt
 
-class ExploredRegion () : IsPartOfGameInfoSerialization {
```
alexban011, line 15: is the default costructor neccessary for serialization? Should I suppress the warning?

alexban011:
And finally:
> Disabled TooGenericExceptionCaught && TooGenericExceptionThrown

Should I disable these? There are quite a few cases where we want to do this (like X threw an exception -> fall back to whatever) so the code would be full of @Suppress annotations

core/src/com/unciv/logic/BackwardCompatibility.kt
yairm210: This is so if there's stuff to deprecate the devs can add the stages here. Currently there are none, but this is legit

tests/src/com/unciv/dev/FasterUIDevelopment.kt
yairm210: This is a dev helper tool, this is legit

yairm210:
And yes, keep those disabled

All in all excellent work!

core/src/com/unciv/logic/civilization/ExploredRegion.kt
alexban011: thanks!

PR #9409, opened 2023-05-19
Next-Turn Progressbar

SomeTroglodyte:
See demo in #8099 - a little more polished.

Skin-Moddable feature largely untested, as I'm not good at painting Ninepatches - might work, or maybe not - I suspect Table does not respect the content area size of a background Ninepatch properly. Granularity is still cities, and thus it looks a little jerky on turn 0 when tons of cities are founded increasing max - now less obvious as I fade the thing in slowly. (That could also have been alleviated with the dropped HealthBar simply by turning on its animation feature, which would have had an effect akin to blurring an image)

I'm packing for a trip, so I might not respond for a week or three.


yairm210:
I don't like the fact that we've introduced a static, I think we can do the same by passing in the updater class to the nextTurn function

yairm210:
I see... I didn't consider that the number of cities increasing or decreasing messes with the advancement

PR #8686, opened 2023-02-17
Fixed unit upgrade check

Gualdimar:
Fixed bug from [discord](https://discord.com/channels/586194543280390151/588357826758574106/1075505021745111050)
For some reaseon, in the previous form of the _getRejectionReasons_ function, the _civ_ object inside the _sequence_ curly brackets had the number of resources that was before calling unit.civ.units.removeUnit(unit), resulting in insufficient resources.
Now it works correctly:

PR #9420, opened 2023-05-21
Resolve To-do in Unit upgrade resource requirements

SomeTroglodyte:
#### Untested
It runs several NextTurns fine, and I found no obvious issues in Overview or similar screens. **But** the specific case where I have used up all my irons and want to upgrade an iron-consuming unit to another iron-consuming unit - I haven't got a save for that, and am struggling to provoke one (Someone help out???).

So - please don't merge but read the gist of the diffs and let your brain process. I think the many eliminated Elvises and non-null assertions are a good cause to consider.

#### Motivation
From reading history I revisited #8686 and then tried to implement what I commented there. On the way I found that leaving those Requirements in a HashMap<String,Int> resulted in enough helper functions to justify changing the type to Counter<String>.

Reading Counter, I found that get override that never returns null but its signature says otherwise - uh, kotlin has no problem with changing return type nullability in an override.

All in separate commits for readability.

As for the linting-change to Civilization.getMatchingUniques: Explicit return type helps the IDE save resources. Of course type can be inferred, and compilation works and is no slower, but the IDE (at least for me) often chokes on this one and flags it, only to un-flag it a quarter of an hour later... And - consistency with getTriggeredUniques which aready had explicit return type.

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -279,9 +279,9 @@ object Automation {
             for (city in civInfo.cities) {
                 val otherConstruction = city.cityConstructions.getCurrentConstruction()
                 if (otherConstruction is Building)
-                    futureForBuildings += otherConstruction.getResourceRequirementsPerTurn()[resource] ?: 0
+                    futureForBuildings += otherConstruction.getResourceRequirementsPerTurn()[resource]
                 else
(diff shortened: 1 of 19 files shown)


yairm210:
Thanks!

yairm210:
Worst case scenario, we figure out it isn't and we revert
That's why working on Unciv rather than anything else, because the consequences are so minimal we can get away with anything ;)

PR #9427, opened 2023-05-22
Game saves are regular json, not minimalized json

### LINKED ISSUE #9414: Feature request: valid JSON save files
I wrote multiplayer unciv server and found that Unciv save files contains broken json **https://javadoc.io/static/com.badlogicgames.gdx/gdx/1.11.0/com/badlogic/gdx/utils/JsonWriter.OutputType.html#minimal**, which can't be ready by something except libgdx. Can be this fixed or there is a reason why it is done?

yairm210:
Resolves #9414 - json parser can handle both types, so backwards compatible 100%
@touhidurrr If we save games as gzipped *regular* json instead of what we have now, will that affect the server in any way?

### DIFF
--- a/core/src/com/unciv/json/UncivJson.kt
+++ b/core/src/com/unciv/json/UncivJson.kt
@@ -3,6 +3,7 @@ package com.unciv.json
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.utils.Json
+import com.badlogic.gdx.utils.JsonWriter
 import com.badlogic.gdx.utils.SerializationException
 import com.unciv.logic.civilization.CivRankingHistory
 import com.unciv.logic.map.tile.TileHistory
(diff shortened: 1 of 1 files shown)


yairm210:
Oh oops, put other commits in this branch instead of elsewhere, I'll fix that

yairm210:
It was not, I unfortunately deleted this branch accidentally so it was auto-closed
But that is something I do a lot, because close-open doesn't require me to enter a different page. Laziness Will Out.

touhidurrr:
> @touhidurrr If we save games as gzipped regular json instead of what we have now, will that affect the server in any way?

What do you mean by gzipped regular json? Without Base64?
The answer is no, I would have to deactivate base64 parsing to make the discord notifications work but thats all.

yairm210:
No, still gzipped,still base64. Just the json changes to standard json.
If it's good for you then away we go :)

PR #9430, opened 2023-05-22
Fix ConcurrentModificationException when puppeting a city

SomeTroglodyte:
Smallest possible fix for #9429, tested just once from a reliable repro save.

How do you feel about Counter.coerceAtMost or the chances to find similar Counter uses using code inspection alone?


yairm210:
I'm against, because the max number of specialists can change often, AND I'm not sure what serialization would do to an extra field in the map...

Issue #9328, opened 2023-05-05
worker will not replace improvement to enable strategy resources


ninjatao:
**Platform**
Android 13, 64bit

**Version**
4.6.8

**Describe the bug**
Worker automation: worker will not replace improvement to enable strategy resources.

**To Reproduce**
1. Let a worker build a farm on tile with Iron
2. Automate the worker
3. Worker will ignore the tile where it should build a mine, whether automatedWorkersReplaceImprovements is on or off.

**Expected behavior**
Worker should build a mine (replacing the farm).

**Additional context**
Not tested but one quick guess:
tileCanBeImproved() in WorkerAutomation.kt, from line 333:

        } else if (!tile.containsGreatImprovement() && tile.hasViewableResource(civInfo)
            && tile.tileResource.isImprovedBy(tile.improvement!!)   **(<- maybe this should have '!' ahead?)**
            && (chooseImprovement(unit, tile) // if the chosen improvement is not null and buildable
                .let { it != null && tile.improvementFunctions.canBuildImprovement(it, civInfo) && unit.canBuildImprovement(it, tile)}))
            return true
        return false // couldn't find anything to construct here

SeventhM:
Closed as Solved as per #9449

PR #9449, opened 2023-05-24
Fix bug worker will not replace improvement to enable strategy resour…

janarvid:
…ces #9328

- Created a unit test to reproduce the bug
- Fixed the bug using the suggested fix in the issue
- Verified that the test worked
- All other tests also ran successfully

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -333,7 +333,7 @@ class WorkerAutomation(
             val chosenImprovement = chooseImprovement(unit, tile)
             if (chosenImprovement != null && tile.improvementFunctions.canBuildImprovement(chosenImprovement, civInfo) && unit.canBuildImprovement(chosenImprovement, tile)) return true
         } else if (!tile.containsGreatImprovement() && tile.hasViewableResource(civInfo)
-            && tile.tileResource.isImprovedBy(tile.improvement!!)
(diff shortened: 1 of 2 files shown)


janarvid:
Thanks for checking @SomeTroglodyte, I'll comment your suggestions and fix later.

janarvid:
Hi, I've now updated the PR to make the test logic more clear and have also added some comments.

1. Yes, you're right about the ! precedence.  I've now removed the parenthesis.
2. I've tried to make the test logic more clear and have also made some comments.
3. Well ... 'sut' is an abbreviation for 'System Under Test' and is commonly used in unit tests.  Since this is not clear for everyone I've changed it to 'workerAutomation'
4. Yes, completely agree with you on the code duplicate problem, I didn't want to make too many changes in my first PR.  I guess many of these functions like 'createCity' and 'addUnit' are/will be used in many tests.  Also objects like ruleset, gameInfo and uncivGame are used in many tests.  I think the best way to do this is to create a new base test class, put these and others in it and then let tests extend this new base class.  But as you say this is for another PR. 
5. Fixed

Hopefully this PR will pass now ...

yairm210:
Looks good, I'm assuming that since Trog looked at this and had no further improvements that it's OK with him as well :)

yairm210:
Had I seen a problem I would have said, but I don't

yairm210:
That's an interesting question - did greeks place coins on eyes, or on eye *sockets*? Was it Obol per Eyeball?

PR #9451, opened 2023-05-25
Fix mouse wheel on NotificationsScroll

SomeTroglodyte:
Mostly a test whether I could recover from a bad experiment: Android Studio Hedgehog preview. Rollback success - I can push again.

The NotificationsScroll since a few versions ago can be dragged/flicked in all 4 directions - and I forgot the mouse wheel, where libGdx has such a moronic default. I mean, a look at the hardware should tell you, if there's a choice, scroll vertically not horizontally... But Gdx chose to prefer horizontal if content exceeds ScrollPane area on both axes.


yairm210:
Please add a comment on this override explaining why it is needed

PR #9455, opened 2023-05-25
All notifications from Overview are now temporary

SomeTroglodyte:
... and make the "Unimproved" numbers clickable.

Verbose commit history.

The first might not please some people, as by their intended use they disappear once you stop using one, but - it keeps them out of save games, and sidesteps the loophole of repeated use. Mesa likah bettah.

That latter is a _very_ old idea, now quite simple. Like you said before, the problem is, how are people gonna know? Color cyan maybe, emulating web pages, only without the hover underline / pointer change? Civilopedia?

As for the links from city demand / resource icon -
* I still don't like that those go over the entire world (OK, discovered only, but still feels too much. But - it's potential strategic info for the maniac aggressor for who to assault next - or for the muesli-guzzling peacenik for who to befriend, so I've got no valid idea how to trim it down)
* The "e.g. near city" is usually way inappropriate - yes it names the closest of _my_ cities but that can be half a world away, naming a competitor city would often be nicer if it's discovered, but that would be quite a different PR.
* Another use of existing statics that might perhaps be avoided, but it was already there, and I'm improving it a little - the NotificationsOverviewTable _keeping_ a WorldScreen reference around was a bit too much...

Also raises the question how to improve NotificationsOverviewTable - one endless list is not the only option. Turns as expanders (:pensive:)? As subtabs (:upside_down_face:)? Add a floating SelectBox for the turns operating as kinda bookmarks (:thinking:)? Keyboard digits jump to current-X (:keyboard:)? Keyboard Arrows/PgUp/PgDown (:racehorse:)? Format the turns headers to stand out a little more? All of the above?


yairm210:
Thank you for splitting this up into different commits, I would never have been able to parse this if not on a commit-by-commit basis

- Resources more than 5 tiles away from a city should not be considered 'near' it - solves the 'half a world away' problem
- Turns as expanders - possible. As subtabs - absolutely not, we don't want 100 subtabs. Selectbox is also out since then you can only see one turn at a time which is not what you want if you're history-scrolling

PR #9456, opened 2023-05-25
Fix UniqueType.RoadMaintenance being ignored

### LINKED ISSUE #9453: The unique "[-50]% maintenance on road & railroads" doesn't work properly
**Platform**
Windows 11 64bit, Android

**Version**
4.6.14-patch1(PC) ; 4.6.13-patch1(Android)

**Describe the bug**
Ruleset: Civ5 - GnK
Mods: Difficulty_Sandbox(for test purpose), my simple nation mod, some other irrelevant mods

I made a simple nation mod which passed the Locate mod errors check. The mod works fine on everything except for the unique "[-50]% maintenance on road & railroads", which is the same unique as Inca.
Base Road Maintenance is 1 gold per road, right? After I built 2 roads using my mod nation, the transportation upkeep shows -2 instead of -1. Please check the screenshots section.

**To Reproduce**
1. build 2 roads on Non-Hill tiles using the nation Inca
2. check the transportation upkeep to see if it's -2 or -1

**Expected behavior**
Under the unique "[-50]% maintenance on road & railroads", the transportation upkeep of 2 roads should be -1.

**Screenshots**

**Additional context**

SomeTroglodyte:
Fixes #9453 - yes the actual fix is just 7 characters.

...but the questions posed there are still valid. This contains an answer to the first one by reasoning only: If a mod ever declared, say, changing maintenance costs over eras, by listing several maintenance uniques with mutually exclusive conditionals, that would now work and would have failed before (by applying all as long as any fulfills a condition)... Right??

### DIFF
--- a/core/src/com/unciv/logic/civilization/transients/CivInfoStatsForNextTurn.kt
+++ b/core/src/com/unciv/logic/civilization/transients/CivInfoStatsForNextTurn.kt
@@ -7,6 +7,7 @@ import com.unciv.logic.civilization.diplomacy.RelationshipLevel
 import com.unciv.logic.map.tile.RoadStatus
 import com.unciv.models.ruleset.Policy
 import com.unciv.models.ruleset.tile.ResourceType
+import com.unciv.models.ruleset.tile.TileImprovement
 import com.unciv.models.ruleset.unique.StateForConditionals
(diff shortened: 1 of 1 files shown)


yairm210:
I don't think it would have failed before, since the conditionals should still have filtered it out. Maybe I'm missing something?

PR #9459, opened 2023-05-25
More reapply CityFocus on yield changes

### LINKED ISSUE #9425: Buying a courthouse don't change the citizen focus
**Platform**
Windows 11 (64bit)

**Version**
Version 4.6.13 (Build 869)

**Describe the bug**
When buying a courthouse in a city, the hapiness go back to -10 but next turn their are cities that starve. (It's the same bug as https://github.com/yairm210/Unciv/issues/9189)

**To Reproduce**
1. Conquer a city.
2. Wait until the resistance end.
3. Buy a courthouse (while the hapiness is less than -10).
4. See that some cities are starving while they shouldn't.

SomeTroglodyte:
I've played a whole game with this code, and noticed no evil things... So, please evaluate carefully and tell me what you think

However, _strictly_ testing for correctness isn't trivial.

Key decision was to wrap the "Hey we need to reevaluate CitizenFocus" method such that it first determines whether a player is actually triggering the change and is watching, or if it's maybe next-turn automation; and based on that chooses the expensive actual evaluation or to simply sets the flag to do it on the next startTurn. So AI buying tiles _will_ pass through the method, but it'll be cheap... However, its name could be clearer, as the "Deferred" part comes potentially after the decision, but I didn't want to give it a monster name like "reassignPopulationImmediatelyOrDeferredBasedOnUIState"... It's right on top of the diff in logic/city/City.

Closes #9425, see discussion there. It's a nice surprise the easy solution to that "Courthouse in A means B needs to go back to farming" thing, the coroutine wrapper, works neatly and just fine, and I couldn't break it by continuing gameplay without giving it time...

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -1,6 +1,7 @@
 package com.unciv.logic.city
 
 import com.badlogic.gdx.math.Vector2
+import com.unciv.GUI
 import com.unciv.logic.IsPartOfGameInfoSerialization
 import com.unciv.logic.city.managers.CityEspionageManager
 import com.unciv.logic.city.managers.CityExpansionManager
@@ -403,6 +404,10 @@ class City : IsPartOfGameInfoSerialization {
         reassignPopulation(resetLocked = true)
     }
 
+    /** Apply worked tiles optimization (aka CityFocus) - Expensive!
+     *
+     *  If the next City.startTurn is soon enough, then use [reassignPopulationDeferred] instead.
+     */
     fun reassignPopulation(resetLocked: Boolean = false) {
         if (resetLocked) {
             workedTiles = hashSetOf()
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/logic/city/City.kt
```diff
@@ -412,9 +417,20 @@ class City : IsPartOfGameInfoSerialization {
...
     }
 
+    /** Apply worked tiles optimization (aka CityFocus) -
+     *  immediately for a human player whoes turn it is (interactive),
+     *  or deferred to the next startTurn while nextTurn is running (for AI)
+     *  @see reassignPopulation
+     */
+    fun reassignPopulationDeferred() {
+        // TODO - is this the best (or even correct) way to detect "interactive" UI calls?
+        if (GUI.isMyTurn() && GUI.getViewingPlayer() == civ) reassignPopulation()
```
yairm210, line 431: Absolutely not - this information is in the *game data*, so we can check city.civ.gameInfo for this info without addressing the GUI at all

yairm210:
Looks good!

core/src/com/unciv/logic/city/City.kt
yairm210: Why would you want to dever stuff on start turn and end turn? Performance reasons? Reassign population is not very heavy IIRC

yairm210:
I would still prefer to not use GUI if possible, I thought your comment indicated that you were using it for performance reasons (to not update if we're in the middle of ending a turn)

PR #9461, opened 2023-05-26
Map editor update - concurrency, resource amounts, file double-click

SomeTroglodyte:
Started out because I found those thread() and Gdx.app.postRunnable calls which needed migration to our Concurrency module.

Minor UX improvements
- non-strategic resources having an amount (can still be loaded from existing map files, but will be autofixed by replacing the resource with itself, or amount can be adjusted down to 0 manually)
- newly placed strategic resources will, as before, have an amount determined by the major/minor deposit random and the abundance setting. But clicking again will increase amount by one each click - how to tell users? Let them discover by accident for now.
- amounts can be manually adjusted on the View tab (yes I know, it's no longer pure, but that was the place where a single tile is selected)
- Doubleclick on file names, as load/save game already supports
- Minor layout tweaks

Note: The patch to TileLayerMisc is needed otherwise a changing amount won't be displayed. It basically informs the local icon cache to discard on amount changes. Not needed for normal play, so is that extra overhead avoidable there? No idea.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -501,6 +501,7 @@ Map copy and paste =
 Position: [param] = 
 Starting location(s): [param] = 
 Continent: [param] ([amount] tiles) = 
+Resource abundance = 
 Change map to fit selected ruleset? = 
 Area: [amount] tiles, [amount2] continents/islands = 
 Area: [amount] tiles, [amount2]% water, [amount3] continents/islands = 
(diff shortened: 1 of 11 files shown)


Caballero-Arepa:
Amazing! Thank you very much.

hackedpassword:
Every single time I go to load a map, I double tap the map, knowing it won't load. Now it will, and it works great!! The rest are great additions too, thank you! Looking forward to more map editor upgrades!

hackedpassword:
Hope! LOL

PR #9467, opened 2023-05-27
Update civ resources when a researched tech reveals a resource

### LINKED ISSUE #9462: Tech gained during turn does not update strategic resources
image

This was done by hurrying Atomic Theory using a Great Scientist, though I suppose it can also be done via scienceOverflow from normal research. The Manufactory was already placed on the Uranium tile by luck...

I guess this could be solved with a big fat sequence chain in the tech gaining code - if interactive, scan all strategic resources, filter enabling tech = gained tech, cross join all tiles, filter tile resource = enabled resource, filter tile owner = viewing civ, map city, distinct, then for each update tile yields?

SomeTroglodyte:
Closes #9462 - pretty brute force, but seems good enough.


yairm210:
Looks good, this is better than how I would have done it since you took the 'multiple resources from single tech only cause one update' into account :)

PR #9472, opened 2023-05-28
Advanced map parameters rework + UI to copy and paste them

### LINKED ISSUE #8108: Feature request: Copy Paste Map Generator Settings
After playing around with map generator settings, it is hard to remember the interesting combos you found.

A button to copy to clipboard and paste from clipboard would allow you to save various setting combos to a text file (or share with a friend). This feature is available for the generated map. I'm requesting it for the Advanced Settings slider values, not the map.

SomeTroglodyte:
** Careful ** - some dangers lurk, see below

Closes #8108 
Experimental "Menu" Widget as one-off SelectBox subclass.
Moving fields with backward compatible deserialization.

#### To consider
* SelectBox is still an ugly beast, not very customization-friendly. Read the hack code - too much? Try the UX - imperfect as it is - opinion?
* The major "urgh" is that SelectBox insists on having a selection plus other limitations, so my "Menu title" needs to go in the dropdown, I just hide it - gives a blank space. Unavoidable without rewriting major parts of Gdx List or the SelectBox itself.
* But - extensible. Allowing a few "settings save slots" in GameOptions, accessible from that menu, would be easy.
* I left proxies for the old "locality" of those settings - since each is only accessed three to maybe a max of six times, changing all accesses and dropping the proxies would actually be less code in a few more files - preference?
* Even if this is overall a "not worth it" - that Serializer holds two important lessons learned, which we should remember:
        * How to override one of the read or write methods without actually changing anything
        * That this "merge" style approach actually is benign and works as simple as it looks
        * Oh, and three, custom serialization could be the basis to a transparent conversion of those object{constants} collections to proper enums - could be pretty straightforward, by allowing the deserializer to read _either_ name _or_ label
* Some of this is my AOS[^1], like having the same field order everywhere - sorry
* Verbose commit history

[^1]: almost obsessive syndrome


yairm210:
Sorry, too many PRs in succession. How about some pics?

PR #9481, opened 2023-05-29
Victory screen moddable Spaceship picture

### LINKED ISSUE #9477: How about... Some visuals to that scientific victory?
Spaceship-1

Same question - this would be mod only, that sample comes from http://fr.tintin.com/, so not under an open license.

SomeTroglodyte:
Closes #9477 - see visuals there.


yairm210:
I'm actually against this - do not like.
The so-called "spaceship" victory needs to be abstracted to a "multi part victory", I know I haven't gotten around to doing that but I still don't want to lock-in any more spaceship-specific parts.

PR #9482, opened 2023-05-29
Unit Overview: Improving a tile is also "what the unit is doing"

### LINKED ISSUE #9476: How about... "What is that Worker doing??"
image

:yum: Yes please or no bleargh :vomiting_face: ?

I mean, yes, it's already there, but so is the rest of the action column (and health) - tiny on the unit icon itself...

SomeTroglodyte:
Closes #9476


yairm210:
So I can't see what automated units are doing?

yairm210:
Thanks for improving the char-to-image readability!

Issue #9488, opened 2023-05-30
How about: Map painting canvas madness


SomeTroglodyte:
<details><summary>Actual screenshots</summary>

![image](https://github.com/yairm210/Unciv/assets/63000004/460da16d-5208-43cc-9886-7c38f00ea4dc)
![image](https://github.com/yairm210/Unciv/assets/63000004/74694d38-8aa3-4aa6-8034-0ba5d1b86c8a)

</details>

There was another open issue requesting something like this, can anybody link it?

Caballero-Arepa:
Holy cow... this would be absolutely time saving

alexban011:
> ours crashes on X11

won't the file picker used in the LoadGameScreen.kt work for this? I haven't looked too much into how that works, but it looks like that works for me. 

Anyway I do think that #9490 looks much better than whatever this 1999 AD stuff is supposed to look like:
![image](https://github.com/yairm210/Unciv/assets/21990802/ce3778be-d531-4267-abc3-477346efafee)

yairm210:
I recall that someone already made a "create map automagically from file based on pixel colors to base terrain colors", I'm not sure which is better - that one cuts out a lot of work, and this one is better for terrain features...

Issue #9439, opened 2023-05-23
Unable to convert value to required type: true


marek22k:
**Platform**
Operating system type, major version, architecture (64/32 bit).
Parrot OS (Debian 11-based)

**Version**
4.6.13 (Build 869)

**Describe the bug**
```
$./Unciv 
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
2023-05-23T08:31:35.384783Z [threadpool-daemon-1] [OnlineMultiplayerServer$checkServerStatus] [ERROR] https://uncivserver.xyz does not support server feature set | com.badlogic.gdx.utils.SerializationException: Unable to convert value to required type: true (com.unciv.logic.multiplayer.ServerFeatureSet)
	at com.badlogic.gdx.utils.Json.readValue(Json.java:1173)
	at com.badlogic.gdx.utils.Json.fromJson(Json.java:834)
	at com.unciv.logic.multiplayer.storage.OnlineMultiplayerServer$checkServerStatus$1.invoke(OnlineMultiplayerServer.kt:53)
	at com.unciv.logic.multiplayer.storage.OnlineMultiplayerServer$checkServerStatus$1.invoke(OnlineMultiplayerServer.kt:49)
	at com.unciv.logic.multiplayer.storage.SimpleHttp.sendRequest(SimpleHttp.kt:62)
	at com.unciv.logic.multiplayer.storage.SimpleHttp.sendGetRequest(SimpleHttp.kt:21)
	at com.unciv.logic.multiplayer.storage.SimpleHttp.sendGetRequest$default(SimpleHttp.kt:20)
	at com.unciv.logic.multiplayer.storage.OnlineMultiplayerServer.checkServerStatus(OnlineMultiplayerServer.kt:49)
	at com.unciv.UncivGame$create$1.invokeSuspend(UncivGame.kt:184)
	at com.unciv.UncivGame$create$1.invoke(UncivGame.kt)
	at com.unciv.UncivGame$create$1.invoke(UncivGame.kt)
	at com.unciv.utils.ConcurrencyKt$launchCrashHandling$1.invokeSuspend(Concurrency.kt:87)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
[...]

PR #9490, opened 2023-05-30
Replacement PlatformSaverLoader for Linux X11 systems

SomeTroglodyte:
Since #9488 needs a working file chooser, and the problem with java awt JFileChooser or FileDialog crashing the game on X11 Linux still persists (#9439), I finally caved and wrote one.

<details><summary>Demo</summary>

https://github.com/yairm210/Unciv/assets/63000004/411bd205-aa6a-4efe-8e6b-515512bf55e4

</details>

I took that incomplete sample I found, had it autotranslated to kotlin, then fixed what was missing and rewrote until I liked it. Therefore the API of the new Popup is oriented on the demo and ease of integration. The requirements of PlatformSaverLoader are a bit different, so there's pure a glue wrapper. The old Android instance is not touched, nor is desktop unless the os name is Linux and the graphic device is X11.

Note - the thing _is_ tested on Android and works fine, but of course since java.nio can't load most files on Android, this chooser won't see any files outside the two allowed places either. You can navigate almost everywhere, but most folders will appear empty even if they aren't. The lie is google's, not java's or ours.

### DIFF
--- a/core/src/com/unciv/logic/files/FileChooser.kt
+++ b/core/src/com/unciv/logic/files/FileChooser.kt
@@ -1,0 +1,294 @@
+package com.unciv.logic.files
+
+import com.badlogic.gdx.Files
+import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.files.FileHandle
+import com.badlogic.gdx.graphics.Color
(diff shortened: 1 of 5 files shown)


yairm210:
Definitely better than the nothing we have now

PR #9493, opened 2023-05-31
Allow image overlay and changing world wrap in map editor

### LINKED ISSUE #9488: How about: Map painting canvas madness

SomeTroglodyte:
Closes #9488 

Since then - 
* We now have a File picker, and this uses it cross-platform, not PlatformSaverLoader because complications, see todo in source [^1]
* I can't get an image to stay where it belongs in world wrap mode, so I turned 180° and turn world wrap off as soon as you load an overlay
* I hear the mad mob calling for a lynching, so - you can now turn it back on, or turn it on after generating without - if the tile collection allows - which is another story [^2]

I hesitate to do screens or video - needs to be seen on a big screen. Get the community maps mod, [search for a high resolution image](https://www.surfertoday.com/environment/maps-of-the-world), put it where the game can reach it, load the mod's 160x80 Earth map, overlay the image, then play see-saw with the alpha slider.

[^1]: PlatformSaverLoader would be required to get any file on Android not in the usual locations (/data/data/_package_/files or /sdcard/Android/data/_package_/files), BUT its API only allows loading text applying encoding. But we ideally want a Gdx FileHandle allowing for a _managed_ Texture, one that knows its source so it can recover from GL context loss. Best you gan get on droid is a stream, but then you would have to make a PixMap backed Texture managed which it isn't by default........ ..... ...

[^2]: Our maps still lie. TileMap constructors will (hex) or can (rect) omit a column from the map for world wrap to get a clean seam, but won't record the change in mapSize. Iow, hex maps always have different tile layout in ww than non-ww, while rect ww will get a column chopped off when ordered as odd width... Or, there's simply maps you can't convert without manipulating the tile array or getting a cute-looking sawtooth seam...

### DIFF
--- a/android/assets/Skin.json
+++ b/android/assets/Skin.json
@@ -208,7 +208,8 @@
             "font": "button",
             "fontColor": "color",
             "downFontColor": "pressed",
-            "overFontColor": "highlight"
+            "overFontColor": "highlight",
+            "disabledFontColor": "gray"
         }
     },
     "com.badlogic.gdx.scenes.scene2d.ui.Label$LabelStyle": {
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -514,6 +514,14 @@ River generation failed! =
 Please don't use step 'Landmass' with map type 'Empty', create a new empty map instead. = 
 This map has errors: = 
 The incompatible elements have been removed. = 
+Current map: World Wrap = 
+Overlay image = 
+Click to choose a file = 
+Choose an image = 
+Overlay transparency: = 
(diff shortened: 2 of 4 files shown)


yairm210:
Overlay feels to me like it requires a different class. MapScreenOverlayFunctions perhaps.
That's a lot of code for a very specific use-case

yairm210:
No, that looks worse, you're right

yairm210:
prepare version

yairm210:
Version rollout PR created

Issue #8817, opened 2023-03-05
Feature request: multiplayer game with server API and more features


CrsiX:
Hi, we really like the game but the multiplayer functionality just sucks, sorry. (And the server is pretty ... hacky. Maybe a plain FTP server would have been a better choice).

Therefore, we would like to put our hands on a new multiplayer service. We essentially build our own server (in Rust) while also hooking the relevant client functionality. Additionally, we integrate feature requests such as the following:
- #8476 (yes)
- #8063 (definitively)
- #7745 (pretty far ahead the roadmap, but a nice idea)
- Also, we want to completely overhaul the API, like using WebSockets/HTTP2 instead of polling on intervals and much more intuitive user experience.

However, we would like some help on integrating some of the stuff in the client app. For example, we build the server and the API and all of the network-related code, while someone else could jump in to tweak the user interface to make use of the newly available features (e.g. chatting or username/password instead of UUIDs).

Side question: how many people are currently using your public multiplayer service? A rough approximation (20 a day or 20k a day)?

yairm210:
That's a question for @touhidurrr, he built the server at https://github.com/touhidurrr/UncivServer.xyz
Given the '3M requests per day' I assume we're talking more in the 20K range

I have no problem with overhauling the API (current API was basically a MVP for migrating away from Dropbox, hence the FTP-ness)
*But* we need to let existing users retain their current games.

This means one of the following:
- Changing the implementation *together with* uncivserver.xyz
- Adding an alternate implementation *alongside* the existing one

I would obviously prefer the former but @touhidurrr also has his own free time restrictions.

@GGGuenni has been leading the recent changes for passwords &c so he's more relevant to this than I am
As I have stated many times in the past, this is super outside of the roadmap for the project - but if you have people willing to work on it from both sides, then why not

Starting with the basics - 'create multiplayer game and allow others to join'.
We'll want to start with a really stripped-down POC:
- One player creates the game with set number of players, game options, etc
- 
- All initial players (except the owner?) are AIs. Each player who 'joins' the game effectively replaces one of the AIs, so we can start the game at any point with the same amount of players
- Uploads to server (GameSetupInfo, including a new field: "creator ID" for the person who created and can start this game), which registers this as a 'lobby game
- Other players can retrieve 'lobby games' and ask to join.
- When a player joins: 
  - If there is an AI he can replace - does so
  - If not: error message to user
- Creator gets updates of new players joining the game and can decide at any point to 'start the game'
[...]

PR #9497, opened 2023-06-01
Multiplayer APIv2

CrsiX:
This may be a larger PR and I'm open to suggestions. Please read on.

### Highlights

- in-game chats
- lobbies & invites
- friend system & friend chats
- event notifications
- lazy loading of data
- it works™

### What it does

This PR introduces a new library (`ktor`) and sub-package that speaks with our [new multiplayer server](https://runciv.hopfenspace.org), as well as a lot of new UI elements that make use of those new API features, bundled with some new events to integrate the WebSocket smoothly. It already comes with a new Android turn checker as well. For more details, see #8817.

We have carefully checked the new experience. It works pretty well, except for the list of [bugs and other issues](https://github.com/hopfenspace/Unciv/issues) I created at our fork. Most importantly, I don't want to break backwards-compatibility with APIv0 and APIv1 (that's how I started to call the Dropbox and all the Dropbox-replacement servers, respectively). That means: at the game start (as well as on server change), the server is queried for the `/isalive` endpoint to determine if it's an "old-style" server. If not (response code `400`), it is checked for `/api/version`. Only if the `apiVersion` can be matched, you will see the new features. Otherwise, everything should stay the way it currently is: players would not see any difference if they don't use a new APIv2-compatible server.

Note, however, that the servers are not compatible for a specific game. At the moment, you can't create a game in APIv1 and continue playing in APIv2 (or vice versa), and there are no plans to allow that, yet. That means: a party needs to decide on a server to play on beforehand.

### API

I bundled all the details in `com.unciv.logic.multiplayer.apiv2`. It includes the main class `APIv2`, which holds a lot of other stuff besides the core functionality, which resides in `APIv2Wrapper`. Accessing functionality is possible like so: `UncivGame.Current.onlineMultiplayer.api.auth.login` to use the login endpoint, for example. There's also `ApiV2FileStorageEmulator` which enabled in-place usage of the new functionality whenever the current APIv0/APIv1 would be used (e.g. saving/loading games).

The "game preview" mechanism is not used at all. The server holds a counter `dataID`, which is incremented every time some player uploads a new game. Checking the value of that counter is enough to detect changed game states.

### UX

There are a bunch of new screens, buttons, tables and other elements around. Some important screens:

- A new `LobbyBrowserScreen` when you click on "multiplayer" in the main menu, which provides a list of your current games and all open lobbies of the current server, as well as buttons to create lobbies, manage friends, ...
- A new `MultiplayerGameScreen` which you see when you click on the multiplayer button of the `WorldScreen`, providing the chat, player operations, access to the friends and recent games
- A new `LobbyScreen` which replaced the `NewGameScreen` for our purposes here, it also holds a chat window

The user is usually notified by popups about incoming events (e.g. new chat message, lobby invitation, a friendship request, ...).

### Missing things

- Most importantly, *any* kind of persistence: Whenever the code needs some values, the server is queried for them (except if it's just the current operation, then the values are piggy-backed as long as needed). This means you probably see requests whenever some window is opened, because the data is loaded lazy. It doesn't bother or block the UI, though. At the moment, I didn't bother, because a) the server is fast and b) the data is very small, only a few hundred bytes at most usually.
- Android turn check re-connect fixes, since it's pretty hard to keep the WebSocket alive and the data in sync. I've already spent hours tweaking that.
- A few buttons and other elements will just popup with "not implemented yet". Of course, those things will be addressed in the future.
- Joining a lobby by invitation & one-to-one in-game chats (all-player chat exists, though).

### Take note

- Changing the server or checking the connection will drop the instance of `OnlineMultiplayer` now, so that everything is cleaned up properly.
- The `NewGameScreen` will use horizontally aligned picker buttons instead of vertically aligned as before (provides much more screen space).
- There are no localized strings _anywhere_, because a) this is a draft and b) I wasn't sure how to do it properly.
- Kotlin builds take seemingly longer now, or is my machine just slow? I thought about splitting all that `apiv2` package stuff out, but that didn't work out easily.
- I think I didn't change any existing data structures, so that everything should just continue working. The only exception is the data format of the uploads to APIv2 servers, which is "full" JSON, but since the PR 9427 is merged, this is no specialty anymore.

cc @SomeTroglodyte @touhidurrr 

If you checkout and build, I suggest the following patch (it provides better debugging experience when testing on multiple devices):
<details>

```git
diff --git a/android/AndroidManifest.xml b/android/AndroidManifest.xml
index e0ef891c2..5cc931a0a 100644
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -21,7 +21,7 @@
         android:allowBackup="true"
         android:icon="@mipmap/uncivicon"
         android:roundIcon="@mipmap/uncivicon_round"
-        android:label="@string/app_name"
+        android:label="Unciv V2"
         android:isGame="true"
         android:largeHeap="true"
         android:appCategory="game"
diff --git a/android/build.gradle.kts b/android/build.gradle.kts
index cfd212cfd..f35a9c2c9 100644
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -26,7 +26,7 @@ android {
         resources.excludes += "DebugProbesKt.bin"
     }
     defaultConfig {
-        applicationId = "com.unciv.app"
+        applicationId = "com.unciv.app.apiv2"
         minSdk = 21
         targetSdk = 32 // See #5044
         versionCode = BuildConfig.appCodeNumber
diff --git a/android/src/com/unciv/app/AndroidLogBackend.kt b/android/src/com/unciv/app/AndroidLogBackend.kt
index ed32784cc..0efba78c2 100644
--- a/android/src/com/unciv/app/AndroidLogBackend.kt
+++ b/android/src/com/unciv/app/AndroidLogBackend.kt
@@ -18,7 +18,7 @@ class AndroidLogBackend : LogBackend {
     }
 
     override fun isRelease(): Boolean {
-        return !BuildConfig.DEBUG
+        return false // !BuildConfig.DEBUG
     }
 
     override fun getSystemInfo(): String {
diff --git a/core/src/com/unciv/Constants.kt b/core/src/com/unciv/Constants.kt
index 977c70e80..fe9a2f6c2 100644
--- a/core/src/com/unciv/Constants.kt
+++ b/core/src/com/unciv/Constants.kt
@@ -87,6 +87,7 @@ object Constants {
 
     const val dropboxMultiplayerServer = "Dropbox"
     const val uncivXyzServer = "https://uncivserver.xyz"
+    const val runcivServer = "https://runciv.hopfenspace.org"
 
     const val defaultTileset = "HexaRealm"
     const val defaultUnitset = "AbsoluteUnits"
diff --git a/core/src/com/unciv/models/metadata/GameSettings.kt b/core/src/com/unciv/models/metadata/GameSettings.kt
index eafb8e4f8..3be4b6f59 100644
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -225,7 +225,7 @@ class GameSettingsMultiplayer {
     var passwords = mutableMapOf<String, String>()
     @Suppress("unused")  // @GGuenni knows what he intended with this field
     var userName: String = ""
-    var server = Constants.uncivXyzServer
+    var server = Constants.runcivServer
     var friendList: MutableList<FriendList.Friend> = mutableListOf()
     var turnCheckerEnabled = true
     var turnCheckerPersistentNotificationEnabled = true
diff --git a/core/src/com/unciv/ui/crashhandling/CrashScreen.kt b/core/src/com/unciv/ui/crashhandling/CrashScreen.kt
index 58cdf49ba..6aeedf1cc 100644
--- a/core/src/com/unciv/ui/crashhandling/CrashScreen.kt
+++ b/core/src/com/unciv/ui/crashhandling/CrashScreen.kt
@@ -181,7 +181,7 @@ class CrashScreen(val exception: Throwable): BaseScreen() {
         )
             .onClick {
                 if (copied) {
-                    Gdx.net.openURI("https://github.com/yairm210/Unciv/issues")
+                    Gdx.net.openURI("https://github.com/hopfenspace/Unciv/issues")
                 } else {
                     ToastPopup(
                         "Please copy the error report first.",
diff --git a/desktop/src/com/unciv/app/desktop/DesktopLogBackend.kt b/desktop/src/com/unciv/app/desktop/DesktopLogBackend.kt
index 122c167ef..6cbbe09e2 100644
--- a/desktop/src/com/unciv/app/desktop/DesktopLogBackend.kt
+++ b/desktop/src/com/unciv/app/desktop/DesktopLogBackend.kt
@@ -11,7 +11,7 @@ class DesktopLogBackend : DefaultLogBackend() {
             && System.getProperty("kotlinx.coroutines.debug") == null
 
     override fun isRelease(): Boolean {
-        return release
+        return false
     }
 
     override fun getSystemInfo(): String {
```

</details>


touhidurrr:
@CrsiX any test version for your **it works™**?

touhidurrr:
Also, I would like the client to send the server some notifications if someone has won, lost, or resigned a game in v2.
Also, you are requesting info for all open games I see, But how do you know when a game has been finished? In an users game list, all of his games played ever will be shown if this isn't resolved properly.
Also, I would like to implement a rating system.

For all of this reasons, I think this would be a very important feature for v2.

touhidurrr:
Also, I don't like the naming convention in many cases as it does not clarify what they do.
For example:
image
This should be `JoinLobbyResponse` as it Lobby might have many functions. Which makes what response this is confusing. Similarly:
image


image
Not sure what `ChatFull` does here at all. Not from name or description. Similarly:
image

Many documentation and naming scheme problems like this.

Also, for clarification, where is the WebSocket docs? Or, schemas = ws docs as I assumed?

CrsiX:
> @CrsiX any test version for your it works™? Want to see how it looks like.

Just compiled for [Android](https://nas.hopfenspace.org/public/apk/Unciv-APIv2-debug.apk) and [Desktop](https://nas.hopfenspace.org/public/apk/Unciv-APIv2-debug.jar) with the above patch applied now. I would suggest you also compile it locally.

Assuming you use the desktop:
1. Download the .jar
2. Copy it to a second folder
3. Start two Unciv instances in each folder separately
4. Play together with yourself to test the functionality

> For all of this reasons, I think this would be a very important feature for v2.

Please add these suggestions to the issue and not this PR, thanks :)

> But how do you know when a game has been finished? 

I think you got a point here... I may have overlooked something. Indeed, I need to check that out, thanks for spotting. [New issue](https://github.com/hopfenspace/Unciv/issues/75)

> Many documentation and naming scheme problems like this.

That's right. I know what the API is doing, since I have implemented that stuff, but of course, it needs clarification and extension. It's not easily understandable just by reading it, yet. However, this is out of scope for that PR as well, please head over to the server repo [here](https://github.com/hopfenspace/runciv) for these suggestions.

> Also, for clarification, where is the WebSocket docs? Or, schemas = ws docs as I assumed?

Same issue here. At the moment WS docs are only in code: [source](https://github.com/hopfenspace/runciv/blob/ca1ad4297c8bf5f77a9a475f9f7bd996e306032a/src/chan/ws_manager_chan.rs#L73).

touhidurrr:
> > Many documentation and naming scheme problems like this.
> 
> That's right. I know what the API is doing, since I have implemented that stuff, but of course, it needs clarification and extension. It's not easily understandable just by reading it, yet. However, this is out of scope for that PR as well, please head over to the server repo [here](https://github.com/hopfenspace/runciv) for these suggestions.
> 

In Multiplayer v1, there were only 3 API endpoints initially. Later 2 more endpoints were added for auth.
For comparison, in https://runciv.hopfenspace.org/docs/#/ there are **33** endpoints in total.
In https://github.com/hopfenspace/runciv/blob/ca1ad4297c8bf5f77a9a475f9f7bd996e306032a/src/chan/ws_manager_chan.rs#L73 there are **15+** ws messages.
All in all it's **~50** new types of communication between the client and the server. That's, **10x** more complexity added to the table.

For all these reasons, I don't think it is ideal in any way to merge this pr to Unciv without completed documentation. I don't understand why are you treating it as just another feature request. It is clearly not that. Your statement:
> However, this is out of scope for that PR as well, please head over to the server repo [here](https://github.com/hopfenspace/runciv) for these suggestions.
does not make any sense to me.

The necessity of documentation is clearly not out of scope for this pr. It should be a requirement for this pr specifically. We already have a requirement for documenting new Mod features. Multiplayer v2 is not less complex than that. So, a situation where only you understand does not seem ideal to me.

I think this pr needs proper documentation before merging. It already has 6k+ lines and adding some docs for oblivious people like me isn't going to hurt much.

touhidurrr (changes requested):
Add docs!

CrsiX:
> For comparison, in https://runciv.hopfenspace.org/docs/#/ there are 33 endpoints in total.

To be fair, we also provide a _lot_ more functionality. It needs to have a place somewhere ;)

> I don't think it is ideal in any way to merge this pr to Unciv without completed documentation

You got an absolute right point there. That's one reason I marked this PR as a draft at the moment: it is to collect your ideas and suggestions. I've forwarded your request to the server repo.

> I think this pr needs proper documentation before merging. It already has 6k+ lines and adding some docs for oblivious people like me isn't going to hurt much.

Okay, so the aforementioned request for docs was about the API, but that is about the client code, I assume. I already added docs to some of the core classes and methods used in the `apiv2` package, but I will also have a look at other places and extend it. Also, I will add a section about it in the `docs/` folder, which looks like a good place to hold general overview of what it's doing (even though the docs there might not target the overly technical person, but it would be a good starting point for developers & other interested people).

CrsiX:
@SomeTroglodyte Did you have a look into what I've produced here already? I would like to fix some remaining little quirks and get a working and merge-able state that can be reviewed. I know it's a lot of stuff, that's why I will add docs later that day or the day after. Addressing the merge conflicts would only be the very last step to avoid doing it twenty times.

@yairm210 Do you have any objections against this PR in general?

CrsiX:
> Oh my, you know I don't do multiplayer, so my motivation is... <sub>low</sub>. Add to that 70 files changed in 140 commits, and I get a headache :hammer: 

Would it be okay if I try splitting up the bunch of changes into smaller diffs for reviewing? Like, one PR adding the new API stuff, one for integration into existing OnlineMultiplayer functionality, one adding and changing UI and one for the Android stuff? (It would still be thousands of changed lines, though.)
Because, I want to produce meaningful & useful code that is actually used later, so I would be happy to help reduce your workload when attempting to merge it. Of course, something like 8000 lines of changes is huge, so if you have any advise for me, I would be happy to help.

> By the way, my version of fun nowadays is "did you know ...

I think I don't quite get what you want to say here... Could you please clarify that paragraph? :) 

----


Additionally, you can address me in the future if multiplayer-related problems occur.

CrsiX:
> Of course that would help us all, just not you

I am very aware of that. But the thing is, that _I_ want it to get merged, so I am the one to go the extra mile... I mean, you "could" accept such a huge PR, but I understand if it's nothing you want to review throughout. So, either I split it up as good as possible or I don't. I would be fine with both. Which path should I choose? ^^

> Really? Just kotlin architecture curiosities

Thanks, understood. IMHO, those object singletons should be used sparely, shouldn't they? Thinks like the `UncivGame.Current` are fine, but having instances of the other things seems much cleaner to me.

CrsiX:
> I am prejudiced against all online games

Sorry about that, some of them are really great. Any maybe Unciv becomes a little great in that matter, too. That's why I'm doing it :)

Also, thank you for the detailed executions. That clarifies what I'm about to do in the next few days (or weeks, I can't tell, I'm sometimes also pretty slow on coding work). The first PR of adding dependencies & ground work should be comparatively easily to produce, as you already pointed out.

For everything else, well, we will see how painful it becomes :)

----

Regarding the Kotlin stuff. I'm actually new to Kotlin, starting when doing this PR work, so I'm not too into it. I got the essence of it, thanks :)
(One may argue that my work should be inspected much more closely, then. But I became used to it rather quickly. After all, it's not entirely different from other languages. It's not like comparing Haskell to Assembly.)

CrsiX:
> So that other PR will move ApiV2 forward for realsies, and you keep updating this one to keep the voyage's "destination" current? OK...

Something like this. I think the easiest & most effective way is to keep my track while periodically merging current master and fixing the issues bubbling up during merge. Yesterday, I had a merge with more than 350 affected files, with 12 files having non-trivial conflicts.

So, ideally, I keep up with that branch (`dev` on my fork) and split its functionality out in smaller PRs to this repo. And when most of them are done, the diff should become small, so that either merging this PR becomes feasible or applying the diff directly is not a too huge headache.

Of course, this also introduces some problems. Like synchronization of changes and so on. But that's why I'm keeping `dev` updated.

> Just noticed: Coroutine use that doesn't go through our "Concurrency" module?

I think I always used `Concurrency.run*` functionalities to start new coroutines, could you point me where you've been looking? I favor integration about something different, especially since handling concurrency is not so trivial sometimes. Was it in the Android work, possibly? Searching...

CrsiX:
Android: Since I'm using `CoroutineWorker` for Android turn checks, the `startWork` method creates coroutines in another `CoroutineScope` than what we usually use. I've set that scope to our `override val coroutineContext = Dispatcher.DAEMON`, is that enough to fix it? If not, I will take a second deeper look.

CrsiX:
> Flows

I've seen them in the `init` of `OnlineMultiplayer`. Not sure about their functionality, so I didn't touch them.

> didn't check, sorry

No problem, it's valuable input and makes me re-think stuff :) 

---

Regarding the crash-handling on Android. My V2 `doWork` is basically completely wrapped in try-catch, because I don't want things to break. Worst case would some background check error which leads to a crash screen and interrupts a game -- worst user experience.
At the moment, I just dump any errors to the logs, where _I_ can read them. Of course, normal players wouldn't, but that's not the relevant part at the moment. :D 
That being said, I **did** have some crashes during my tests, that came from deep inside some libraries (once libGDX, once even the Android / Java runtime), but even though they break my turn checker, they shouldn't break the game itself. But since they weren't reproducible, I simply dumped them to the end of the backlog.

CrsiX:
a) Did you use my builds / your own build from my branch? (Just to be sure)
b) What exactly did you do? (So that I can reproduce it)
c) Can you provide debug logs? (So I can confirm it)
d) Are you using https://runciv.hopfenspace.org as the server or something different? I know of problems where you are only using http:// instead, which doesn't work properly yet.

I would like to help investigate and fix any v1 problems as soon as possible :)

touhidurrr:
@CrsiX, I was trying out your [v5.0-alpha2](https://github.com/hopfenspace/Unciv/releases/tag/v5.0-alpha2) release with the server address:
```
https://uncivserver.xyz
```

CrsiX:
I wasn't able to reproduce until now, I have opened up https://github.com/hopfenspace/Unciv/issues/82 to keep track of your issue. Thanks for reporting :)

CrsiX:
I think you found a bug which I already closed in a later version (1 day after the release you tested). See https://github.com/hopfenspace/Unciv/issues/57 for the original. So, in my current `dev`, this is fixed, but not in the release you've just tested, because it was slightly older.
Anyway, thanks for spotting.

CrsiX:
I'm preparing PR 2.

> Point me to some *specific* design decision to evaluate, then maybe I'll actually look.

Please give me your 5 cents on the way the new API is accessed, functionality is used and replaced when the server URL changes.

1. Access is usually done via `UncivGame.Current.onlineMultiplayer.api` where `api` is the current `ApiV2` instance of the current server. Accessing it will always do a quick check if it has been initialized and is compatible; if not it raises an error.
2. Compatibility is checked by `UncivGame.Current.onlineMultiplayer.isInitialized() && UncivGame.Current.onlineMultiplayer.apiVersion == ApiVersion.APIv2`, which is a `lateinit` created by a specific `initialize` coroutine. Therefore, there's the helper `isInitialized`, to avoid errors by looking up that value.
3. All calls to `UncivGame.Current.onlineMultiplayer.api` are usually guarded by a check that the API is compatible before access to avoid the aforementioned error. Here, "usually" means: if the caller, e.g. UI element, is only accessible after such a check, like `LobbyBrowserScreen`, then there are no further checks -- I don't expect the server URL to change _while_ such a screen is active, even though it could be the case if someone enforces it by `refreshOnlineMultiplayer`, see below.
4. The default functionality used by the interface `FileStorage` has already been added, so that's no problem. What will be added next is a more advanced check _which_ file storage "backend" should be used. I thought about making them all `suspend`able, since this enforces the use of a coroutine and avoids the `@Suppress("RedundantSuspendModifier")`... At least on my branch the IDE tells me I can remove it.
5. Replacing the server URL now **needs** a replacement of the `OnlineMultiplayer` instance (so there will be a helper for that: `refreshOnlineMultiplayer` [here](https://github.com/hopfenspace/Unciv/blob/50ada3b27b232de3edec37a2fa9cf2121224cbbb/core/src/com/unciv/UncivGame.kt#LL556C13-L556C37)) -- maybe not the best way to do it, but throwing it away is the simplest way to make sure that there are no properties (e.g. `apiVersion` or so) that lie around somewhere. Also note that `OnlineMultiplayer` now is `dispose`-able for proper cleanup.
6. Since I've named my property `api`, I currently limit the future extensibility a bit for something like APIv3, if that is a problem :) -- Also, is it perfectly clear what it is just by the name, `onlineMultiplayer.api`?

CrsiX:
> `FileStorage`

Honestly, I thought about throwing it away at first, since it's also only exposed to the `OnlineMultiplayer`, afaict. But anyways, I keep it around, since it does its job, and I first and foremost don't want to break APIv0 & APIv1.

> (3) guards - good Kdoc should be enough.

Well, I see that this is still a problem, since I didn't have _that much_ of docs (yet). The `ApiVersion` thing is good documented, but besides that, it's not much docs.

> I imagine your monster as big black box with plugs & knobs on the outside

It's not thaaaaat scary, I promise 🙃 

> init ... then you can't guarantee a synchronous caller anything

Exactly, that's why I have `initialize` which is a separate suspending method which does all that stuff. It should complete before first use of the `OnlineMultiplayer` API functionality, but since it does a network call (or, multiple of them), it may take anything between 10ms and 5000ms, or even timeout altogether. So, I'm not yet really guarded against such problems, since my dev machine & its network is fast.
Also, I didn't want to only allow suspending calls, since there are a bunch of simpler methods which could complete earlier. Especially when we introduce persistence (or at least a useful cache), certain lookups will become much faster, e.g. what friends/games/lobbies/chats are there and are they up-to-date? I made `awaitInitialized` to guard this, but it's pretty ugly with an active sleep loop, didn't have a better idea back then. I need to get something like observer pattern / wait-notify / a simple channel.

> "but" - I see the direct use of `kotlinx.coroutines.runBlocking`

Uhm, well, the reason is probably typing `runBlocking` and letting the IDE auto-import it, using the `kotlinx` version first? I check on that.

CrsiX:
> Also, is "to be used in `main` functions and in tests." (right in the function's Kdoc) relevant? Or that there is an equal-named wrapper in com.unciv.utils.Concurrency?

Of course. But... I'm not quite sure about bridging suspending and non-suspending code easily. The fact that in my current branch _all but two_ calls to `runBlocking` are written by me made me pause too now. Thinking: wouldn't it be better to integrate those suspending calls, e.g. via callbacks or other mechanisms? I've just realized that I've probably over-used them (esp. in `LobbyPlayerList`, that class is a mess). So it popped up on my task list now.

CrsiX:
I'm curious. The `runOnGLThread` uses the `Dispatcher.GL` with `Gdx.app.postRunnable`. I'm using the `runOnGLThread` for most modifications to the UI (and where I don't, I have crashes sometimes -- remember that `LobbyPlayerList`, a huge mess). As far as the documentation is concerned, the GL context is single-threaded.

1. Is this still the case in Unciv? (I don't see anything telling otherwise, but just to be sure.)
2. Can I _always_ assume that `runOnGLThread` will be synchronized, i.e. it won't require mutexing to avoid concurrent access to variables, if all accesses are done on the GL thread? (As long as it's not `dispose`d.)

I can't finally answer those two questions just by reading docs & code. Thanks :)

CrsiX:
If we put our trust into StackOverflow answers [here](https://stackoverflow.com/a/25396622), then it's also single-threaded on Android. I will just try implementing it that way and see if it works or not.

yairm210:
GL is single threaded always because that's how G
OpenGL works

SeventhM:
If you say you're still working on this, I'll trust you and leave this to not stale (unless Yairm) objects to the label here as well). If noting else, someone could grab most of the work and probably implement it

CrsiX:
It's fine if this PR gets closed by the bot if you ask me. The work I have done is still there; and since it was unreviewable due to size back then, I would need to break it down into smaller components that can be integrated one by one -- this is the tricky part. And since I haven't touched it in half a year, it will be a mess for me just as for anyone else (but if anyone actually wants to touch it, feel free) :D

touhidurrr:
> This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.

Lol. Whats the purpose of will not stale tag then?

SeventhM:
Huh... I bet the will not stale tag is only applying to issues and not PRs

PR #9505, opened 2023-06-02
More efficient use of the charts space

JackRainy:
| Before | After |
| - | - |
| Screenshot 2023-06-03 01-18-52 | Screenshot 2023-06-03 01-19-57 |
| Screenshot 2023-06-03 01-18-23 | Screenshot 2023-06-03 01-02-01 |
| Screenshot 2023-06-03 01-18-08 | Screenshot 2023-06-03 00-59-42 |


JackRainy:
No, I definitely have no time for a new control / widget. There is still a war in Ukraine. :(

However, I am going to make this chart a bit more readable. For example, this change makes the Y axis more compact, and I am going to introduce bounds to limit the X axis as well.
See those loooong sleeves in the early turns? They are absolutely useless and unreadable. I would like either to zoom in to them, or trim them and focus on the later turns.
That's on my desk right now.

PR #9514, opened 2023-06-04
Upgraded music player popup

SomeTroglodyte:
This is the spruced up music popup shown in #9513, and the branch is _based_ on that one, so read only the last commit.


yairm210:
Self-conflict

yairm210:
Nope, let's go with this and if we think of anything better we can improve in the future

PR #9516, opened 2023-06-04
Mod checks against sellable, missing or multiple Palace(s)

SomeTroglodyte:
Might help a little against cases such as in #9510


SeventhM:
I think this PR goes a little too far. The primary issue here is the game not being able to handle not having a capital. The mod I'm working on is based on a mod in Civ 5, and Civ 5 handles this more elegantly. Instead it doesn't allow you to sell your capital unless you already have another capital elsewhere. And yes, in that mod it is 100% intentional that you're allowed 2 capitals, and it's 100% intentional that they can be sold, namely because Civ 5 only allows one of them to be considered the capital and leaves the other one as just generically a palace

SeventhM:
Sure, but if the need is to create better safeguards against the possibility of not having a capital, the better solution might be to actually still go ahead and address the `getCapital()!!` calls or at least the function itself, not risking breaking functionally in mods that quite frankly, would likely be possible in civ 5 proper. There's plenty of ways that it could probably be addressed (from disallowing selling your last capital to just fixing the function calls to be safer without a capital (, but disallowing selling capitals altogether seems far more likely to break functionality than address issues

PR #9522, opened 2023-06-04
Make ExpanderTab "expand" properly

SomeTroglodyte:
Warning:
* Quirk: The distance between those expanders in the clip is larger with this PR than in master, puzzling.
* Quirk: The lists within the construction Expanders in CityScreen sometimes fail to grow their children cells fully.
* Still needs some testing - helpers welcome.

Nears completion and looks pretty darn cool, but that's all it is - a cooler look&feel. Will need to do a video of how the dynamic content in WorldScreenMusicPopup's history now _animates_ when a new entry comes in - came almost free of charge with a little sideways thinking.

<details><summary>Clip</summary>

https://github.com/yairm210/Unciv/assets/63000004/7a487c0a-099c-438b-921f-7b5ab117dfab

</details>


yairm210:
Reverted this after users reported problems with the city construction table
I checked it and there were indeed serious problems
image

image

Especially problematic on phones
I preferred to revert because I don't know what your schedule is like to fix this problem - I'm *all for* the change, I just want it non-breaking :)

Issue #9534, opened 2023-06-07
Feature request: add more topics


GrimPixel:
**Is your feature request related to a problem? Please describe.**
There are few topics to distinguish the use of mods. It is hard to see what the mod does simply by their names

**Describe the solution you'd like**
Add more topics, like “unciv-nations”, “unciv-religions”, “unciv-buildings”, “unciv-improvements”, “unciv-units”.

yairm210:
@SomeTroglodyte 
They won't appear as filters and as texts in the button:
![image](https://github.com/yairm210/Unciv/assets/8366208/cbe90925-0adc-43af-879f-4c664a697c72)

I think this is a good idea, as well as replacing "fun" with "mechanics" since that's what it's actually being used for practically speaking

PR #9542, opened 2023-06-07
Make mod categories curated in Json

SomeTroglodyte:
UI change:
* You'll know why when mod manager 'filter and sort' ignores you
* e.g. "The Great Unciv Rework" can now be found by both its categories

Behind the scenes change:
* Changes in categories are made in ModCategories.json,
* ... optionally after using the "Update mod categories" button in Options (hide in Debug tab? hide without Ctrl-Shift?)
* So new categories in use in the wild are prepared for us - to let users use them in-app we'll need to
      * Edit the label if necessary
      * Decide on sort position and move if necessary
      * Add the label to templates - optional: Next PR feeds that json to xtlfw
      * Remove the hidden flag
* Updated the wiki a little. Did I describe what "Fun" was meant for OK?

Commit separation not perfect, but still helps reading I hope - because 1st draft built on, not undone

<details><summary>discarded original comment</summary>

This is **really** more of a dumb idea, but is what produced the screenshot in #9534 ... Should we unexpectedly want it. Little problem: Where should translation templates come from? Bigger problem: We should find an idea how to correct people missing the supported topic by just one character... Pity making a topic curated is such a hassle.

What we could actually do is - github query function and enum to normal class as in this PR, then make it a non-ruleset json that fills the Category collection, and make the query to a tool button in options we can use to merge new categories into the json, where we curate it, add display name, mark undesired... Opinion?

Reminder: com.unciv.ui.screens.pickerscreens is completely wrong for object Github - whereto?
</details>

### DIFF
--- a/android/assets/jsons/ModCategories.json
+++ b/android/assets/jsons/ModCategories.json
@@ -1,0 +1,50 @@
+[
+{
+	"label": "All mods",
+	"topic": "unciv-mod",
+	"createDate": "2020-08-26T11:35:39Z",
+	"modifyDate": "2023-06-08T05:39:47Z"
+},
+{
+	"label": "Rulesets",
+	"topic": "unciv-mod-rulesets",
+	"createDate": "2022-07-23T06:41:09Z",
+	"modifyDate": "2022-11-22T01:58:08Z"
+},
+{
+	"label": "Expansions",
+	"topic": "unciv-mod-expansions",
+	"createDate": "2022-07-23T03:55:05Z",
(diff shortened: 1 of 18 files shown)


yairm210:
No thank you, having a bajillion tags in the dropdown is not on my list of "nonsense I need to put up with"
Especially since a zero-star mod can put "unciv-rudeword" and have it appear for everyone, whereas if he put it in the mod name you'd have to do a lot of scrolling for that
I'd rather hardcode 

I will go for the background changes if they improve existing, though

Issue #7514, opened 2022-07-22
Request: Trigger (existing) notifications for particular events


Ouaz:
1) The notification "`Your city [cityName] was converted to [religionName]!`" is **only** triggered when a city is converted to a religion after an unit action executed by the player (e.g. Great Prophet spreading religion, whatever is the religion).

This notification is not triggered when:
- The player's city is converted to a foreign religion by an AI religious unit (e.g. foreign Great Prophet)
- The player's city is converted to any religion by natural religion spread

Could it possible to also trigger the notification "`Your city [cityName] was converted to [religionName]!`" in these cases, because currently, it's easy to miss that some of my cities have been "sneakily" converted?

Note: The notifications "`You gained [Stats] as your religion was spread to [cityName]`" and "`You gained [Stats] as your religion was spread to an unknown city`" exist in the stringtables but are never triggered in-game, even when converting a foreign city to the player's religion with a Great Prophet, or when the player's religion is naturally spread to a foreign city.

2) The notification "`[cityName] has been disconnected from your capital!`" is only triggered when a road/railroad is removed (by mistake most of the time) by the player. If a road/railroad is **pillaged by an enemy**, the notification is not triggered (possibly because the road is considered to be still "existing" but needs to be repaired), so it's easy to miss that a city is not connected anymore to the capital (other than deducing the gold/turn decrease is due to a tile having been pillaged, and that tile was containing a road connecting two cities)

So, could it possible to also trigger the notification "`[cityName] has been disconnected from your capital!`" when a road connecting two cities is pillaged by an enemy (or a tile, containing any improvement and a road, is pillaged)?

Thanks for reading.

itanasi:
This sounds like something the Event Bus would be helpful for!

yairm210:
Good catch!
The problem here is that the game doesn't *remember* what cities were connected to capital last turn, so it doesn't know what changed... To implement this we'll need to add another field to the saved game data

Ouaz:
> The notification "Your city [cityName] was converted to [religionName]!" is only triggered when a city is converted to a religion after an unit action executed by the player (e.g. Great Prophet spreading religion, whatever is the religion).
> This notification is not triggered when: The player's city is converted to any religion by natural religion spread

I think this one has been fixed a few months ago, but I have to check.

PR #9545, opened 2023-06-08
Improve connected to / disconnected from Capital handling

SomeTroglodyte:
Issue #7514 "disconnected" part. I think that could then be closed.

<details><summary>Skrynshutt</summary>

image

</details>

...That screenshot has been made, however, with massive manipulation of the AI: tryPillageImprovement changed to prefer current tile, and an if (enemy territory) tryPillageImprovement call near the start of automateUnitMoves, otherwise I could play 30 turns of that war with 4-6 mongols on that road and they never disconnected me. So...

#### Questions
* That defensive design for backward compatibility - a much simpler approach would re-throw "your city X has been connected to your capital" on the first load after updating for each connected city, once (or one "disconnected" for each unconnected = worse). That would also be what this could migrate to after a grace period - enum to Boolean, and don't differentiate between 'certain' and 'unsure' for the connected last turn tests.
* Would that 'prefer current tile' bit make sense for the general codebase? My code see below
* Non-barbs only ever call tryPillageImprovement from tryHealUnit: Make sense to make those health thresholds moddable? ModOptions/Civilization?
* `UniqueType.NoMovementToPillage` has no influence on AI - should it have? For units with that ability that prefer current tile change would likely make a lot of sense, and maybe they should see if pillaging is an advantage even outside of tryHealUnit?

<details><summary>tryPillageImprovement</summary>

```kotlin
    fun tryPillageImprovement(unit: MapUnit): Boolean {
        if (unit.isCivilian()) return false

        fun Tile.canUnitPillageHere() = (
            canPillageTileImprovement()
                || (canPillageRoad() && getRoadOwner() != null && unit.civ.isAtWarWith(getRoadOwner()!!))
            ) && UnitActions.canPillage(unit, this)

        if (!unit.getTile().canUnitPillageHere()) {
            val unitDistanceToTiles = unit.movement.getDistanceToTiles()
            val tilesThatCanWalkToAndThenPillage = unitDistanceToTiles
                .filter { it.value.totalDistance > 0f && it.value.totalDistance < unit.currentMovement }.keys
                .filter { unit.movement.canMoveTo(it) && it.canUnitPillageHere() }

            if (tilesThatCanWalkToAndThenPillage.isEmpty()) return false
            val tileToPillage = tilesThatCanWalkToAndThenPillage.maxByOrNull { it.getDefensiveBonus() }!!
            if (unit.getTile() != tileToPillage)
                unit.movement.moveToTile(tileToPillage)
        }

        UnitActionsPillage.getPillageAction(unit)?.action?.invoke()
        return unit.currentMovement == 0f
    }
```
</details>

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -101,6 +101,13 @@ class City : IsPartOfGameInfoSerialization {
 
     internal var flagsCountdown = HashMap<String, Int>()
 
+    /** Persisted connected-to-capital (by any medium) to allow "disconnected" notifications after loading */
+    // Unknown only exists to support older saves, so those do not generate spurions connected/disconnected messages.
+    // The other names are chosen so serialization is compatible with a Boolean to allow easy replacement in the future.
+    @Suppress("EnumEntryName")
+    enum class ConnectedToCapitalStatus { Unknown, `false`, `true` }
+    var connectedToCapitalStatus = ConnectedToCapitalStatus.Unknown
+
     fun hasDiplomaticMarriage(): Boolean = foundingCiv == ""
 
     //region pure functions
@@ -129,6 +136,7 @@ class City : IsPartOfGameInfoSerialization {
         toReturn.cityAIFocus = cityAIFocus
         toReturn.avoidGrowth = avoidGrowth
         toReturn.manualSpecialists = manualSpecialists
+        toReturn.connectedToCapitalStatus = connectedToCapitalStatus
         return toReturn
     }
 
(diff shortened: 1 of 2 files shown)


yairm210:
This looks good, but I'm not sure about the enum names being the same as booleans, if it works it works but I want to make sure that it indeed does

Have you tested what happens on
- Loading old save 
- Loading new save
To see that the serialization doesn't cause problems?
How does this object look in-text?

yairm210:
- moddable healing for pillage - yes, sounds like a modOptions unique with civ modifiers (units gain [relativeAmount] health from pillaging)

- nomovement unique affects AI - yes, should

- prefer current tile - not sure what this affects, can't think of a good tactical reason to do so

alexban011:
I'm looking into the recent linting issues that have built up and this PR adds these enum names (`whatever`) which aren't in the kotlin convention. Is there a specific reason as to why you did this  @SomeTroglodyte ? 

Aka should I fix by renaming these to remove the ` symbol or should I add the symbol to the regex pattern?

PR #9552, opened 2023-06-08
Refactor input-related components into own package

SomeTroglodyte:
I promise, not a single functional change.

The branch was named with a further-reaching intention, but after some review I decided to do this first. The separation suddenly makes it crystal clear that ActorAttachments.activationActions and OnClickListener.clickFunctions cover exactly the same usecase, and those listeners need deduplication before any improvements...


yairm210:
Code restructure with no functional change is almost always net positive :)
Thanks!

PR #9553, opened 2023-06-09
Nation picker gets an Icon View, keyboard selection, and fixed sort

### LINKED ISSUE #9526: Feature request: List civilizations in a table
It requires a lot of scrolling to find a specific civilization.

List icons in a table.

SomeTroglodyte:
Little Nation picker upcrate to close #9526, see screenshot there

Untested with mods containing nations where the untranslated json names start with non-ascii characters - ~too lazy to download that - what was it called? - that one tripping the atlas loader with non-UTF-8 Javas~ Leader Mission has no such nations.

But that showed that choosing the first letter of the second word as key shortcut (O for The Ottomans) - if it contains a space at all - isn't necessarily best - "Burkina Faso". Since there's round robin I could register all word beginnings instead?


AdityaMH:
Sometimes i want to see leader portrait in civilization selection list and in world screen instead civ icon. 🤣

AdityaMH:
Sometimes i want to see leader portrait in civilization selection list and in world screen instead civ icon. 🤣

core/src/com/unciv/ui/screens/newgamescreen/NationPickerPopup.kt
```diff
@@ -0,0 +1,269 @@
...
+        val nationSequence = (part1 + part2)
+            .sortedWith(
+                compareBy<NationIterationElement> { it.sort }
+                .thenBy(UncivGame.Current.settings.getCollatorFromLocale()) { it.translatedName }
+            )
+
+        var selectInfo: SelectInfo? = null
+        var currentY = 0f
+        var currentX = 0f
+        for ((nation, _, translatedName) in nationSequence) {
```
yairm210, line 172: Separate this out into another function

core/src/com/unciv/ui/screens/newgamescreen/NationPickerPopup.kt
```diff
@@ -0,0 +1,269 @@
...
+        innerTable.addActor(okButton)
+
+        val switchViewButton = "OtherIcons/NationSwap".toImageButton(Color.ROYAL)
+        switchViewButton.onActivation {
+            listMode = if (listMode == NationPickerListMode.Icons)
+                    NationPickerListMode.List
+                else NationPickerListMode.Icons
+            settings.nationPickerListMode = listMode
+            updateNationListTable()
+        }
```
yairm210, line 108: separate into another function, perhaps? 'addActionIcons'?

yairm210:
All in all this looks great!

core/src/com/unciv/ui/screens/newgamescreen/NationPickerPopup.kt
yairm210: I was thinking of all 3 icons blocks in one function

PR #9559, opened 2023-06-10
Improve Alert Popup scrolling

SomeTroglodyte:
See #9513 - never intended that. Should have checked. Thing is, distribution into "scrolls if need be" and "static" is automatic in `Scrollability.WithoutButtons` mode as long as the actual buttons are added with Popup's button methods... And AlertPopup bypassed them entirely. So they went to the upper half, where defaults().fillX() is in effect - ugly.


yairm210:
I'm not sure which is better, both look good to me
I'll poll this on Discord

Ouaz:
That's what I call a big close button. Not jarring, I think it's ok.

yairm210:
Current state of the poll:
IMG_20230612_072305

Let's wait another day

yairm210:
image

Looks like the wide button is what people want

yairm210:
Not sure if to close this, or if there are other parts of this PR that you still want to use

PR #9571, opened 2023-06-12
Resolves #9398 - load new multiplayer state when within a game but not on world screen

### LINKED ISSUE #9398: player's turn is not updated
This bug occur only in multiplayer mode

An player its using Linux and other player its using, desktop with Linux

Bug description:
The game is in a looping and the turn is never passed.

How to play:
1 - it's player 1's turn;
2 - player 2 leaves the city window open or the technology window;
3 - player 1 passes while player 2 has the window open.

yairm210:
I first tried checking if this screen was in the screenstack, but that's no good - since when you 'go to main menu' from a game, it doesn't remove that game from the stack, instead adding the main menu on. This is for faster reload on 'resume game' but it does mean that it's not strictly hierarchical.

I tried making it so, but encountered problems on resume game.

This seems the simplest fix, but I do wonder if there's a better way, hence the PR

### DIFF
--- a/core/src/com/unciv/ui/screens/overviewscreen/GlobalPoliticsOverviewTable.kt
+++ b/core/src/com/unciv/ui/screens/overviewscreen/GlobalPoliticsOverviewTable.kt
@@ -99,22 +99,18 @@ class GlobalPoliticsOverviewTable (
 
             // civ image
             add(ImageGetter.getNationPortrait(civ.nation, 100f)).pad(20f)
-
             addSeparatorVertical(Color.GRAY)
 
             // info about civ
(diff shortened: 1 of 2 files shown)


yairm210:
I think that our problems began in the first place when we decided that worldscreen -> mainmenu was a valid *addition to the stack*
Instead, mainmenu should *override the stack* - we can save the worldscreen on the side in order to access it on resume, but it shouldn't be the *parent* of the main menu

yairm210:
I think that restructuring the stack so that returning to "main menu" unloads world screen is the correct way forward

yairm210:
Since this isn't going anywhere...

PR #9575, opened 2023-06-12
Update Polish.properties

Lesiakower:


yairm210:
Was dealt with (badly 😢)  last week

PR #9589, opened 2023-06-14
Multiplayer v2: networking stack, dependencies & API definition

CrsiX:
This is part 1 of the original PR https://github.com/yairm210/Unciv/pull/9497 which has been split up. It adds the new networking dependency, the networking stack and the API definitions. It does not touch relevant other files (the game, the UI, the online multiplayer logic), but _just_ adds new content for networking.

The user experience will not be changed, because the code added here is _not_ used at the moment. This was proposed by @SomeTroglodyte to reduce the amount of new & changed code. After this PR, I will prepare the next PR which will integrate the networking stack in the current game & multiplayer logic. Without that upcoming PR, a user will not notice any difference (except for a slightly bigger APK/JAR file).

Referencing https://github.com/yairm210/Unciv/issues/8817.

(Do you see **significantly** increased build times with the new dependencies for second and later builds?)


CrsiX:
I see a bunch of `WildcardImport` errors in the CI run for this pull request. Should I address them before merging?

CrsiX:
> B: investigate best practices

Apparently, even the ktor tutorials use wildcard imports in their examples. Like [this snippet](https://github.com/ktorio/ktor-documentation/blob/main/codeSnippets/snippets/client-websockets/src/main/kotlin/com/example/Application.kt) referenced [here](https://ktor.io/docs/websocket-client.html#example), or practically any other snippet in that repository.
They are tutorials, though. Looking for some non-tutorial projects, [this](https://github.com/papsign/Ktor-OpenAPI-Generator) and [this](https://github.com/mathias21/KtorEasy) are using wildcards, apparently, where appropriate (number of imports > 1).

alexban011:
> the ktor tutorials use wildcard imports in their examples

usually examples are there to more or less show you how to do X, not the best/cleanest way to do that. Using wildcard imports is generally a bad idea and should be avoided where possible. 

> I'd probably tend to configure detekt to accept.

Unfortunately this isn't just a config option and requires creating a custom rule based on the `WildcardImports` rule and extending it to specifically avoid whatever we want, then disabling the current WildcardImports rule and enabling our own. Currently I really don't have time to look into how exactly this would be done, so for now I'd say lets not use wildcards at all. Doesn't Idea have a system which automatically replaces wildcard imports with qualified imports? The fix shouldn't take that long since from what I see there are only 9 files with this issue

CrsiX:
I don't know if that `.editorconfig` works... But changing the project-specific settings through the Android Studio Settings Editor worked. Mine look like this now:
A


However, at least from my perspective, wildcard imports sometimes could be useful to avoid things like this: 

```java
import kotlin.collections.List
import kotlin.collections.MutableMap
import kotlin.collections.filter
import kotlin.collections.getOrNull
import kotlin.collections.joinToString
import kotlin.collections.listOf
import kotlin.collections.map
import kotlin.collections.mutableMapOf
import kotlin.collections.plus
import kotlin.collections.random
import kotlin.collections.set
```

Anyways, I will just reformat the imports of the new files. And at least on my machine, they will always use single imports now.

yairm210:
You're adding in a ktor implementation *to the base game*. And not just for Desktop, for Android as well.\
If this significantly increases the size of APKs and AABs, I'm going to have to give that a hard pass. One of our stated goals in this project is to minimize the package size, and here you're adding in *a lot* for what appears to be just json parsing purposes.

yairm210:
Previous APK size: 19.8 MB
New APK size: 21.8 MB
We're adding 10% of the APK size here, which is not great, but if that's what's needed to implement the new Multiplayer then I'm okay with that.

core/src/com/unciv/logic/multiplayer/apiv2/ApiV2Wrapper.kt
```diff
@@ -0,0 +1,255 @@
...
+
+/**
+ * APIv2 exception class that is compatible with [UncivShowableException]
+ */
+class ApiException(val error: ApiErrorResponse) : UncivShowableException(lookupErrorMessage(error.statusCode))
+
+/**
+ * Convert an API status code to a string that can be translated and shown to users
+ */
+private fun lookupErrorMessage(statusCode: ApiStatusCode): String {
```
yairm210, line 225: Since every status code needs a user-readable string, why not just put these values in the `enum class ApiStatusCode`?

yairm210: (This is not a merge blocker I intent to merge at the next version release)

CrsiX: You mean by `override fun toString(): String`? Yes, that would be much cleaner.

CrsiX:
> Previous APK size: 19.8 MB
> New APK size: 21.8 MB
> We're adding 10% of the APK size here

I know about that problem. 1.4MB come from the compiled ktor library alone. Even though JSON parsing could be done by another lib, WebSocket handling can not (easily, that is). I haven't found another WebSocket implementation besides ktor (also under the aspects documentation & quality). Since serialization "only" adds 13 KB, I'm fine with that.

Also, since ktor provides a full HTTP stack as well, one could think about stripping out the existing HTTP stack in the future. (I'm not exactly proposing this here, this is off-topic.)

alexban011:
> wildcard imports sometimes could be useful to avoid things like this

Yes but hiding those is the IDE's job, not ours. As long as you don't code using Notepad/equivalent you should be fine :D

CrsiX:
> No, enums are classes and enum values are singleton instances. So, look at...

Okay, so instead of a simple int for an enum, we could also store the required int along a default message text in English ("default" because translation is handled somewhere else). Something like this, I suppose (it's a little worse for overview, but definitively better for usability and doesn't use `toString` or `lookupErrorMessage` helper):

```diff
diff --git a/core/src/com/unciv/logic/multiplayer/apiv2/ResponseStructs.kt b/core/src/com/unciv/logic/multiplayer/apiv2/ResponseStructs.kt
index 751b34abb..50453dd5a 100644
--- a/core/src/com/unciv/logic/multiplayer/apiv2/ResponseStructs.kt
+++ b/core/src/com/unciv/logic/multiplayer/apiv2/ResponseStructs.kt
@@ -48,17 +48,17 @@ data class ApiErrorResponse(
  * by the client. Error codes in the range of 2000..3000 represent server errors.
  */
 @Serializable(with = ApiStatusCodeSerializer::class)
-enum class ApiStatusCode(val value: Int) {
-    Unauthenticated(1000),
-    NotFound(1001),
-    InvalidContentType(1002),
-    InvalidJson(1003),
-    PayloadOverflow(1004),
+enum class ApiStatusCode(val value: Int, val message: String) {
+    Unauthenticated(1000, "You are not logged in. Please login first."),
+    NotFound(1001, "The operation couldn't be completed, since the resource was not found."),
+    InvalidContentType(1002, "The media content type was invalid. Please report this as a bug."),
+    InvalidJson(1003, "The server didn't understand the sent data. Please report this as a bug."),
+    PayloadOverflow(1004, "The amount of data sent to the server was too large. Please report this as a bug."),
```

---

> My mails say you had the IDE exploding the kotlin standard library into explicit imports

Yes, I messed something up. The IDE auto-removed them, so don't worry. (Therefore I edited my previous post, but you still spotted it :D )

PR #9600, opened 2023-06-15
Fix translation problems due to nested brackets in getDifferences

SomeTroglodyte:
#6131 - one item, I always forget we did _not_ buy a good nested bracket matching parser, sorry.
The other item - I'd rather hide the button, but where? Debug is too clumsy if you'll run this more than monthly, or is it?
Also, some cleanup I forgot from that replaced upgrade tooltip


yairm210:
Hide what button...?

Issue #9583, opened 2023-06-13
Sudden Crash


A1y0sh4:
**Platform:** Desktop
**Version:** 4.6.18-patch1 (Build 876)
**Rulesets:** [Civ V - Gods & Kings, Community Civs, Thin Bubbly Borders, The Great Unciv Rework, Civ V - Vanilla, Only Mod Civilizations]
**Last Screen:** `com.unciv.ui.screens.worldscreen.WorldScreen`

--------------------------------

OS: Windows 10
Java: Oracle Corporation 25.232-b09
	Max Memory: 910 MB

--------------------------------

**Message:**
```
java.lang.NullPointerException
	at com.unciv.logic.civilization.Civilization.getEquivalentBuilding(Civilization.kt:497)
	at com.unciv.logic.automation.civilization.ReligionAutomation$tryBuyAnyReligiousBuilding$buildingToBePurchased$1.invoke(ReligionAutomation.kt:93)
	at com.unciv.logic.automation.civilization.ReligionAutomation$tryBuyAnyReligiousBuilding$buildingToBePurchased$1.invoke(ReligionAutomation.kt:93)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:210)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:210)
	at kotlin.sequences.FilteringSequence$iterator$1.calcNext(Sequences.kt:170)
	at kotlin.sequences.FilteringSequence$iterator$1.hasNext(Sequences.kt:194)
	at kotlin.sequences.FilteringSequence$iterator$1.calcNext(Sequences.kt:169)
	at kotlin.sequences.FilteringSequence$iterator$1.hasNext(Sequences.kt:194)
	at com.unciv.logic.automation.civilization.ReligionAutomation.tryBuyAnyReligiousBuilding(ReligionAutomation.kt:431)
	at com.unciv.logic.automation.civilization.ReligionAutomation.spendFaithOnReligion(ReligionAutomation.kt:73)
	at com.unciv.logic.automation.civilization.NextTurnAutomation.automateCivMoves(NextTurnAutomation.kt:68)
	at com.unciv.logic.civilization.managers.TurnManager.automateTurn(TurnManager.kt:321)
	at com.unciv.logic.GameInfo.nextTurn(GameInfo.kt:336)
[...]

PR #9601, opened 2023-06-15
Fix conditionals for trigger upon declaring friendship running twice

SomeTroglodyte:
[See here](https://github.com/yairm210/Unciv/issues/9583#issuecomment-1593842368)


yairm210:
Interesting
Wonder if getTriggeredUniques in general should always return non-conditional-filtered results 🤔

yairm210:
All notification text is a "nice to have", but nice to haves can also be tested, why not

PR #9605, opened 2023-06-16
Notifications architectural update

SomeTroglodyte:
Invisible. Forward compatibility 100%, backward - if you load a game saved with this in an older Unciv, you lose all notification actions, but not the notifications themselves.

Goal: More flexibility for notification clicks, for example "you build the pyramids" - first click: locate city, second: open civilopedia entry.

Notes
* I have a few in mind that could do with improved feedback, but maybe ask on Discord - or you have ideas as well?
* The goal means LocationAction changed and now has only 1 location, while the execute call round-robin moved to the Notification level.
* But - you can still build a Notification using LocationAction(location1, location2,...) due to magic. That does not return one instance, but a collection of them that the upstream consumers also accept.
* The NotificationAction subclass system was extensible and still is, maybe more so - but new ones now need an extra step, because I wanted pretty json: Just add to NotificationActionsDeserializer.
* Incidentally, the pretty json is a bit more compact, but it also made writing the backward compatibility easier - clear picture of how the json should be helps.


yairm210:
The compatibility issue makes this a no-go for me. People play multiplayer games with several people using different versions.

What we can do is:
- Add a *reader* for the new format, so we can read both new and old formats
- Wait for that reader to spread out to the playerbase (say 2 weeks?)
- Then add a writer to start writing the notifications in the new format

yairm210:
@SomeTroglodyte 
Bringing this up - it's time to remove compatibility for old notifications, and simplify the reader.

Hopefully we can get it down to just overriding the serialization for the *action* and not for the entire *notification* now that that's settled

PR #9608, opened 2023-06-16
Update Nations.json

Smashfanful:
Added civilizations intros


xlenstra:
Based on my local copy of Civ V, these seem to be almost copied verbatim. Don't think we can add these because of copyright reasons

yairm210:
@Smashfanful please stop trying to add these in...

PR #9611, opened 2023-06-17
Reordered the region of other global uniques and wrote a few unique tests

xlenstra:
I'm back, at least for a bit while I'm taking a break from life.

This PR basically reorders a large number of the uniques within the enum so that similar uniques are more closely together. I've also written a few unit tests for some of the uniques


yairm210:
:D welcome back!

xlenstra:
> // Todo: Sign should not be part of the unique placeholder - that comment on CostsResources is yours? Ummm. Sign is not part of the placeholder? What am I missing?

I did write that but for a different unique, must have forgot to move that with the unique at some point. I'll see if I can find again where it belongs

> Came across it because - City construction buttons - question aside: You think there's a case for putting a deterministic ordering on the displayed resource requirements? At the moment, it's randomized, but with G&K there's only ever one...

Sorting them by the order they appear in the `ruleset.tileResources` hashmap shouldn't be too difficult and might allow for some consistency in mods that do allow for that

PR #9616, opened 2023-06-18
Support building GraalVM native image

ACGNnsj:
I've tested it on Windows. In principle, it would also work for Linux or MacOS.
Based on my testing, the native application runs with a significantly smaller memory footprint and possibly with faster startup times.
Given the necessity of GraalVM, I didn't add Gradle tasks to build native image.
I've published a release to my own fork.


yairm210:
I'm missing the use case here - who is this *for*, exactly?

yairm210:
Closing as 'not relevant'

PR #9621, opened 2023-06-18
Allowing zero cost buildings and improvements

SeventhM:
This PR changes the default cost of buildings to -1 and the default time to build improvements to -1. The attempted goal is to explicitly allow 0 cost buildings and improvements that build in 0 turns. This is already handled by the game correctly (1 cost buildings can cost 0 on quick speed and zero cost improvements are rounded up to 1) and should have no real serious implications. This is primarily to allow for zero cost buildings and improvements on Epic or Marathon speeds


core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
```diff
@@ -27,7 +27,7 @@ class TileImprovement : RulesetStatsObject() {
     override fun getUniqueTarget() = UniqueTarget.Improvement
     val shortcutKey: Char? = null
     // This is the base cost. A cost of 0 means created instead of buildable.
```
xlenstra, line 29: This comment is now outdated

SeventhM:
> Edit: Ouch: `override var cost: Int = 0`

Ughh, I see what happened. Something I changed on my end but forgot on the commit. This is what  get for not doing this as I probably should

> for units, the default cost was meant to stay at 0? OK, but might confuse new coders.

Yes and no. For units, the default value changes nothing as we don't pull that from tech data automatically. Therefore, there's no interference between the expected default and a mod setting to 0. Even if I changed the default to -1, I can't find a single check that this actually becomes relevant to a mod creator. Hence why I didn't bother to change it

PR #9622, opened 2023-06-19
Move on water

SeventhM:
This PR adds the unique `"May travel on Water tiles without embarking"` as previously discussed on discord a while back. I have it set up to still get attacking penalties for attacking over a coast, though I'm not sure if it should actually have those penalties


SeventhM:
Admittedly, looking back the implementation of this one had a bit of sloppier testing than it probably should given that it seemed like a fairly simplistic set of changes to be done. That said, I probably won't be able to do really much testing until tomorrow due to not having access to the device I usually edit on


> From a superficial reading, I get the impression this can't work. For one, `if(!cache.canMoveOnWater) return false` is not only missing a blank, I think it's inverted.

That... It's inverted. Whoops, thanks. I'm not sure I understand what you mean by it missing a blank


> For two, the new, independent block in `getAttackModifiers` - shouldn't it test your flag somehow?

It... Does? `&& attacker.unit.cache.canMoveOnWater` should catch it

> And since it's setting the same modifier as the test above, and if both were to take the `true` branch, the second would overwrite - so an `else` might clarify that you expect both conditions can never be true at the same time - which I think that inversion breaks.

I'm not sure what you mean here. Assuming I fixed the I fixed the inversion in `isEmbark()`, there's no way for them to override each other. Perhaps I could put the new check in the prior one `if(attacker.unit.isEmbarked() || (the check) && !attacker.unit.hasUnique(UniqueType.AttackAcrossCoast))` to save on logic, but if anything, maybe the better argument actually to merge my check with the naval check further down given they both check for the same stuff and set the same modifier. 

> Three - isn't setting (dis)embarkment movement costs to 1 sufficient for the intended usecase?

No. The intended use case is specifically for the unit to not count as an embarked unit on water, and to use all of it's stats as normal. Setting the embarkation and disembarkation to 1 still leaves the attack penalty for being embarked at all and the inability to give sight bonuses unless it's a sight bonus for embarked units. So it must actually go through all of these changes. Although, I do need to add in the check to disembarking as I thought it previously checked if the unit was embarked, not if the unit was on water

> implicit expectations "could" be tested in RulesetValidator - like you must never set this for an air unit or other such madness. What happens if one slaps it onto a civilian? What if... Say, the MM gave it withdraw too? How will that stack?

Putting it on an air unit wouldn't do anything, but this is a fair thing to do. When I'll get the chance I'll add it. Putting it on a civilian unit shouldn't matter, and should be up to the developer of that mod. Embarked units can already build improvements, naval and embarked units can already activate abilities, I can't imagine slapping this onto a civilian unit would do anything. Giving it withdraw too would be weird, but likely also fine and up to the map maker's discretion. Though, I haven't checked if withdraw allows a unit to go on water. That seems like something the mod maker should be aware of, not so much an implementation problem

> with ~30 or so calls to isEmbarked, how can we be certain that no code operates on assumptions liki - if we're on water and not embarked and not transported, we _know_ it's not a land unit...?

You can check search them, I guess. I just did a check just now through GitHub (again, not on the device with Android studio, so can't be as through. The only example I can think of at all that's actually relevant is in logic/map/tile/Tiles where naval units can block yields from nearby tiles but not typically land units. The loss of the embark check means it checks as a naval unit, which may or may not be intentional

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -119,11 +119,11 @@ object BattleDamage {
                 modifiers["Landing"] = -50
 
             // Land Melee Unit attacking to Water
-            if (attacker.unit.type.isLandUnit() && !attacker.unit.isEmbarked() && attacker.isMelee() && defender.getTile().isWater
+            if (attacker.unit.type.isLandUnit() && !attacker.getTile().isWater && attacker.isMelee() && defender.getTile().isWater
```
yairm210, line 122: Non-embarked unit shouldn't get a penalty attacking water

core/src/com/unciv/logic/battle/BattleDamage.kt
```diff
@@ -119,11 +119,11 @@ object BattleDamage {
...
                     && !attacker.unit.hasUnique(UniqueType.AttackAcrossCoast))
                 modifiers["Boarding"] = -50
-            // Naval Unit Melee attacking to Land (not City) unit
-            if (attacker.unit.type.isWaterUnit() && attacker.isMelee() && !defender.getTile().isWater
+            // Melee Unit on water attacking to Land (not City) unit
+            if (!attacker.unit.type.isAirUnit() && attacker.isMelee() && attacker.getTile().isWater && !defender.getTile().isWater
```
yairm210, line 126: Or when attacking from water

yairm210:
General gist looks good

core/src/com/unciv/logic/battle/BattleDamage.kt
SeventhM: Both of these cases were to keep consistency with the code already here. I don't remember how Civ 5 handles Lekmod's helicopters, or even if it was added as a dll in Civ 5 (for reference, in Lekmod did something like this for helicopters a long while back). And given we already have a unique to remove this penalty, this seems like something that isn't as big of a deal as I may seem. In my eyes, that's more of a design question for us to ask (or poll on discord) than an actual problem

SeventhM:
> Should we redefine vanilla Helicopters with this right away? They indeed feel very wrong as they are now.

Weirdly enough, not unless we poll users on it first. Civ 5 has helicopters embark like any other unit for some reason, so redefining helicopters with this would be a gameplay difference

yairm210 (approved):

PR #9629, opened 2023-06-19
Spies now occasionally steal technologies

xlenstra:
Continues the development of #7610 after a short break of a few months. This PR allows spies to steal technologies from the civilizations they are spying on every so often


android/assets/jsons/translations/template.properties
```diff
@@ -923,6 +923,10 @@ We have captured a barbarian encampment and recovered [goldAmount] gold! =
 An enemy [unitType] has joined us! = 
 After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
 We have recruited [spyName] as a spy! = 
+A Spy from [civilization] stole the Technology [techName] from [cityName]! =
```
yairm210, line 926: Let's be consistent - should Spy be capitalized or not?

android/assets/jsons/translations/template.properties
```diff
@@ -923,6 +923,10 @@ We have captured a barbarian encampment and recovered [goldAmount] gold! =
 An enemy [unitType] has joined us! = 
 After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
 We have recruited [spyName] as a spy! = 
+A Spy from [civilization] stole the Technology [techName] from [cityName]! = 
+An unidentifeid spy stole the Technology [techName] from [cityName]! =
```
yairm210, line 927: Spelling error in 'unidentified'

core/src/com/unciv/logic/civilization/managers/EspionageManager.kt
```diff
@@ -4,11 +4,15 @@ import com.unciv.Constants
...
+import com.unciv.logic.civilization.NotificationCategory
+import com.unciv.logic.civilization.NotificationIcon
+import kotlin.random.Random
 
 enum class SpyAction(val stringName: String) {
     None("None"),
     Moving("Moving"),
     EstablishNetwork("Establishing Network"),
+    Surveillance("Surveiying City"),
     StealingTech("Stealing Tech"),
```
yairm210, line 16: Spelling :)
"Surveying"

xlenstra:
> Stealing tech goes faster the more science your victim produces? Ouch, but OK, if that's how it goes.

Yes, but they than also have probably researched more technologies, which, at least to my understanding, also makes it take more points to steal a tech so gaining points faster shouldn't have that much of an effect

> ... and val spyCount get() = spyList.size removing the two writes would work too ...

That would work for now, but at some point in the future spies will be able to die when they're caught spying and then the values of these values would diverge. Probably. Depends on implementation and I'm not sure how that would go. I'll change it for now as its a minor change and if necessary I can change it back later.

> CityEspionageManager

I've added that in the past, probably with some idea what it would be used for. Can't remember that now, it's been too long. For now I'm inclined to keep it in until more of espionage is done as I can imagine it at some point becoming useful (probably around counterespionage, not sure). If it turns out it isn't useful, we can always remove it later

core/src/com/unciv/models/Spy.kt
```diff
@@ -0,0 +1,164 @@
...
+}
+
+
+class Spy() : IsPartOfGameInfoSerialization {
+    // `location == null` means that the spy is in its hideout
+    var location: String? = null
+    lateinit var name: String
+    var action = SpyAction.None
+        private set
+    var timeTillActionFinish = 0
```
yairm210, line 30: Instead of 'time' llease use 'turns' for consistency
'turnsRemainingToAction' or somesuch

yairm210:
General question: what happens if a spy is in a city and that city gets destroyed? Or conquered?

xlenstra:
> General question: what happens if a spy is in a city and that city gets destroyed? Or conquered?

Guess I have something new to add to #7610

yairm210 (approved):

PR #9632, opened 2023-06-20
Spies in cities that are captured or destroyed now go to the hideout

xlenstra:
As noticed in #9629. Might have minor conflicts with #9629.

### DIFF
--- a/core/src/com/unciv/logic/city/managers/CityEspionageManager.kt
+++ b/core/src/com/unciv/logic/city/managers/CityEspionageManager.kt
@@ -3,6 +3,9 @@ package com.unciv.logic.city.managers
 import com.unciv.logic.IsPartOfGameInfoSerialization
 import com.unciv.logic.city.City
 import com.unciv.logic.civilization.Civilization
+import com.unciv.logic.civilization.NotificationCategory
+import com.unciv.logic.civilization.NotificationIcon
+import com.unciv.logic.civilization.managers.Spy
 
 class CityEspionageManager : IsPartOfGameInfoSerialization{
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/logic/city/managers/CityEspionageManager.kt
```diff
@@ -20,4 +23,22 @@ class CityEspionageManager : IsPartOfGameInfoSerialization{
         return civInfo.espionageManager.spyList.any { it.location == city.id }
     }
 
+    private fun getAllStationedSpies(): List<Spy> {
+        return city.civ.gameInfo.civilizations.flatMap { civ ->
+            civ.espionageManager.spyList.filter { it.location == city.id }
+        }
+    }
+
+    fun removeAllPresentSpies(reason: String) {
```
yairm210, line 32: Please make this an enum, not hardcoded strings

xlenstra: Will do

core/src/com/unciv/logic/city/managers/CityEspionageManager.kt
```diff
@@ -7,6 +7,12 @@ import com.unciv.logic.civilization.NotificationCategory
 import com.unciv.logic.civilization.NotificationIcon
 import com.unciv.logic.civilization.managers.Spy
 
+enum class SpyFleeReason {
```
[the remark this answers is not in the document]
xlenstra: Tbh, this feels like overkill to me.

PR #9639, opened 2023-06-21
More unit unique tests

xlenstra:
Added some unit tests for unhappiness uniques. Also came across some weird behaviour while doing so and wrote up a large comment explaining why this is weird. If someone who is more familiar with the code for determining AI happiness could look at it, it would be much appreciated.

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Nations.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Nations.json
@@ -358,7 +358,7 @@
 
 		"favoredReligion": "Hinduism",
 		"uniqueName": "Population Growth",
-		"uniques": ["Unhappiness from number of Cities doubled", "[-50]% Unhappiness from [Population] [in all cities]"],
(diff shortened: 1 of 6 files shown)


SeventhM:
> However, if we were playing on chieftain difficulty, we would get back 0.6 twice and the modifier would be 0.36. Thus, in general there seems to be something wrong here

According to the Civ wiki, and civfanatics forums, this is almost correct. It does get any multipliers that players get from chieftain difficulty multiplied to the actual difficulty bonuses the ais gets from the selected difficulty. Your conclusion is wrong, however. While `unhappinessModifier` is indeed 0.6, it is multiplied by `aiUnhappinessModifier` which is 1 on chieftain, same as on Prince difficulty 

Fair warning by the way. If the wiki is to be believed, this whole section **_needs fixing_** once we start development on brave new world. According to it the AI no longer plays on Chieftain difficulty in Brave New World, but rather its own difficulty

PR #9644, opened 2023-06-23
Update Polish.properties + question

Lesiakower:
A note regarding line 1683: Why is it written [spyname] instead of [spyName] like before?


Lesiakower:
> Sorry for the confusion - led you to forget the placeholder in the line above

Thanks, I fixed it.

PR #9646, opened 2023-06-23
Update German.properties

Knirps1:


alexban011:
Placeholder `[spyname]` not found in `German` for entry `Due to the chaos ensuing in [cityName], your spy [spyname] has fled back to our hideout.`

Knirps1:
Thanks, this should be fixed now.

PR #9655, opened 2023-06-24
Promotion picker allowing picking chains in one go

SomeTroglodyte:
My take to close #9261 

* Also fixes the promotion star not disappearing from Units overview when "used up".
* If you have enough XP to get more than one promotion, all reachable promotions are colored accordingly
* If you have enough XP to get a promotion "down the prerequisites chain" and the path is unambiguous, you can do so in one click-accept or doubleclick
* If you have enough XP to get a promotion "down the prerequisites chain" and the path is ambiguous, the bottom label says so
* If you do NOT have enough XP to get a promotion "down the prerequisites chain" but the path is unambiguous, the entire path is shown
* Ordering is hopefully better - both json and alphabet would put Cover below Ambush, but the old picker reversed them - I can't even explain why. Should even work tolerably without any row/column info, which is why a certain monster in the Ruleset mixer has been removed. Those no-longer-really-coordinates aren't used anywhere else as far as I know.
* One difference is - certain unit types had Logistics displayed near the _last_ path leading to it, due to a certain strange pruning of the predecessor/successor collections, while this will _always_ put down stuff at the earliest appropriate position - meaning upwards.
* The main reason for the extensive reorganization and rewrite of the last Author's code is - I needed that to even have a chance to get my head around the problem.
* Small aside - Colour defaults only use standard constructors, not our extensions, because the libGdx IDEA plugin then can show them in the gutter beside the source line.
* This time, commit history is my actual progress, not logically organized.

The open question for input on colours in that issue is postponed by having all colours in the Skin, so changing is easy even for noncoders.


yairm210:
Pics please?

android/assets/jsons/translations/German.properties
```diff
@@ -1695,6 +1695,7 @@ Dogfighting III = Kurvenkampf III
 Choose name for [unitName] = Wähle Namen für [unitName]
 [unitFilter] units gain the [promotion] promotion = [unitFilter] Einheiten erhalten die [promotion] Beförderung
 Requires = Benötigt
+- Path to [promotion] is ambiguous = (Der Weg zu [promotion] ist noch nicht klar!)
```
yairm210, line 1698: This doesn't make sense to me. If you want it in brackets then bracket it, if you want it with a separator separate it. Adding the meta-formatting to the translation line is a surefire way to cause confution

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -308,13 +285,13 @@ class Ruleset {
...
-                topRow += 1
-                promotion.row = topRow
-            }
+//         var topRow = unitPromotions.values.filter { it.column == 0 }.maxOfOrNull { it.row } ?: -1
+//         for (promotion in unitPromotions.values)
+//             if (promotion.row == -1){
+//                 promotion.column = 0
+//                 topRow += 1
+//                 promotion.row = topRow
+//             }
```
yairm210, line 294: ?

core/src/com/unciv/ui/screens/pickerscreens/promotion/PromotionPickerScreen.kt
yairm210, line 1: Since you both moved and changed it, I can't tell what was changed

core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
```diff
@@ -0,0 +1,183 @@
...
+
+        init {
+            val splitName = Promotion.getBaseNameAndLevel(promotion.name)
+            this.level = splitName.level
+            this.baseName = splitName.basePromotionName
+        }
+    }
+
+    init {
+        val collator = GUI.getSettings().getCollatorFromLocale()
```
yairm210, line 62: Why are you using this to sort the best path? This is a UI issue only, no?

PR #9660, opened 2023-06-24
Zoom in/out of the history charts

JackRainy:
The `LineChart` control has been totally rewritten.
The key improvements are:
* No objects are created during the `draw()` - less memory usage
* The drawing happens in the same batch as other controls - improved performance
* The drawing happens by native 2D libGdx controls, not using 3D shape renderer
* There is an autoscaling based on the data to be displayed (e.g. from 100 to 200, not permanently from 0 to X)
* The negative axis is not limited by just 1 step

And of course, the primary change: zoom in and out by click on the certain area of the chart.

https://github.com/yairm210/Unciv/assets/27405436/ba73619c-c88a-4be8-8b03-864c5c42e0d2

### DIFF
--- a/core/src/com/unciv/ui/components/LineChart.kt
+++ b/core/src/com/unciv/ui/components/LineChart.kt
@@ -3,6 +3,7 @@ package com.unciv.ui.components
 import com.badlogic.gdx.graphics.Color
 import com.badlogic.gdx.graphics.g2d.Batch
 import com.badlogic.gdx.graphics.glutils.ShapeRenderer
+import com.badlogic.gdx.math.MathUtils.lerp
 import com.badlogic.gdx.math.Matrix4
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.scenes.scene2d.ui.Label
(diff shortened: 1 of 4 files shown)


JackRainy:
> Like I said from the beginning.

Yep, I didn't want to spend time on creating the new control, but at some point it was unavoidable :)

core/src/com/unciv/ui/components/LineChart.kt
```diff
@@ -78,36 +116,47 @@ class LineChart(
...
         return (ceil(value / oneWithZeros) * oneWithZeros).toInt()
     }
 
+    /**
+     *  Returns the previous number of power 10, with maximal step <= 100.
+     *  Examples: 0 => 0, -3 => -10, 97 => 90, 567 => 500, 123321 => 123300
+     */
+    private fun getPrevNumberDivisibleByPowOfTen(value: Int): Int {
+        if (value == 0) return 0
+        val numberOfDigits = max(2 , min(ceil(log10(abs(value).toDouble())).toInt(), 3))
```
yairm210, line 133: Instead of max, min use .coerce - it's much more readable

yairm210 (approved):
Amazing work!

PR #9662, opened 2023-06-24
Update French.properties

Ouaz:


Ouaz:
Erf, I did the same error than other translators! ^^ Fixing it.

PR #9665, opened 2023-06-24
Fixing workers dying in mountains bug for Carthage

### LINKED ISSUE #9654: Workers commit suicide when playing Carthage with lots of mountains.
**Platform**
Android 5.1
Security Patch Level 2015-11-01
Kernel Version 3.10.72
Build Number HY U804MC_20201127_V1.1.00
Not sure how to determine the architecture. I think it's arm64.

**Version**
Unciv Version 4.6.1-patch Build 866 from GitHub

**Describe the bug**
When playing a game as Carthage on a map with a lot of mountains, automated workers frequently climb the mountains and kill themselves with rule damage.

**To Reproduce**
Start a new game with map elevation set to 0.80 and the Human player set to Carthage.
Build a civilization in a mountainous area and build plenty of workers. 
Set the workers to Automated.
After the first General is obtained, watch as your workers all quickly commit suicide by climbing mountains and gaining 50 tile damage.
</details>

**Expected behavior**
I expect the automated workers to not cross mountains without a direct order from the player, since crossing mountains on their own kills them.

random271:
Fixes #9654

Simple fix, checks if there are other tiles beside mountains; if not, moves the worker to a mountain tile.


core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -279,6 +279,13 @@ class WorkerAutomation(
                 }
                 .sortedByDescending { getPriority(it) }
 
+        // Carthage can move through mountains, special case
+        val mountainWorkableTiles = workableTiles.filter{tile -> tile.isMountain()}
```
SeventhM, line 283: I feel like this can be made more generic to include modded terrains that deals damage. Perhaps check if the tile has the unique to damage units?

random271: Haven't thought about that, now it should take that into consideration.

random271:
Is something like this better? Thanks for the tips!

random271:
Android Studio yup. I don't have any experience with Kotlin so nuances of the language are new to me. I assumed count is constant (should have checked) and I've seen lateinit used for things which aren't supposed to be null and well, initialized later. Regarding firstOrNull, yeah absolutely, you can just  workableTiles =  workableTilesNotDealingDamage and pass that on, but I don't really think there is that much of difference except for code style, this way possibly being faster?

Sorry for the inconvenience, it's past midnight and my brain is also shutting down slowly but I found the project interesting and wanted to contribute, and today I had some spare time so I started small.

yairm210:
Github is giving me an error when I try to run this workflow, it's a Github problem
But that means that for now this will have to wait

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -279,9 +279,17 @@ class WorkerAutomation(
...
         // These are the expensive calculations (tileCanBeImproved, canReach), so we only apply these filters after everything else it done.
-        val selectedTile =
-            workableTiles.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
+        if (workableTilesNotDealingDamage.count() > 0)
+            selectedTile = workableTilesNotDealingDamage.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
+                ?: return currentTile
+        // else if all available tiles are damage dealing tiles, move to the damage dealing tile
+        else
+            selectedTile = workableTiles.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
                 ?: return currentTile
```
yairm210, line 293:
```suggestion
            selectedTile = workableTilesNotDealingDamage.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
        // else if all available tiles are damage dealing tiles, move to the damage dealing tile
            ?: selectedTile = workableTiles.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
                ?: return currentTile
```

Alternatively, you can forgo the workableTilesNotDealingDamage and just do:
```kotlin
workableTiles.sortedByDescending {  tile -> unit.getDamageFromTerrain(tile) <= 0 }.firstOrNull { unit.movement.canReach(it) && (tileCanBeImproved(unit, it) || it.isPillaged()) }
```

yairm210 (approved):

Issue #9659, opened 2023-06-24
crash unciv


iiitomo:
**Platform**
LineageOS

**Version**
Latest from Fdroid

**Describe the bug**
Crash

**To Reproduce**
First crash for me.

**Platform:** Android
**Version:** 4.7.0-patch1 (Build 879)
**Rulesets:** [Civ V - Vanilla, Civ V - Gods & Kings]
**Last Screen:** `com.unciv.ui.screens.worldscreen.WorldScreen`

--------------------------------

Device Model: Redmi Note 7
API Level: 29

--------------------------------

**Message:**
```
java.util.NoSuchElementException: Collection contains no element matching the predicate.
	at com.unciv.logic.automation.unit.SpecificUnitAutomation.speedupWonderConstruction(SpecificUnitAutomation.kt:753)
	at com.unciv.logic.automation.unit.UnitAutomation.automateCivilianUnit(UnitAutomation.kt:320)
	at com.unciv.logic.automation.unit.UnitAutomation.automateUnitMoves(UnitAutomation.kt:155)
	at com.unciv.logic.map.mapunit.MapUnit.doAction(MapUnit.kt:492)
	at com.unciv.ui.screens.worldscreen.status.NextTurnButton$getNextTurnAction$13$1.invokeSuspend(NextTurnButton.kt:167)
	at com.unciv.ui.screens.worldscreen.status.NextTurnButton$getNextTurnAction$13$1.invoke(Unknown Source:8)
	at com.unciv.ui.screens.worldscreen.status.NextTurnButton$getNextTurnAction$13$1.invoke(Unknown Source:4)
	at com.unciv.utils.ConcurrencyKt$launchCrashHandling$1.invokeSuspend(Concurrency.kt:87)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at com.unciv.utils.CrashHandlingDispatcher$dispatch$1.invoke(Concurrency.kt:173)
	at com.unciv.utils.CrashHandlingDispatcher$dispatch$1.invoke(Concurrency.kt:173)
	at com.unciv.ui.crashhandling.CrashHandlingExtensionsKt$wrapCrashHandling$1.invoke(CrashHandlingExtensions.kt:17)
[...]

PR #9671, opened 2023-06-25
More defensive coding of speedupWonderConstruction

SomeTroglodyte:
This might solve #9659 - though I haven't got any games this could be tested on realistically.


yairm210:
Best defensive coding is good offensive coding

yairm210:
My coding for example is extremely offensive

PR #9673, opened 2023-06-26
Check for trigger uniques when starting and recaluating population

SeventhM:
This PR does the following
- Checks tech uniques for starting techs
- Checks nation and global uniques when starting (allows for just having a trigger immediately as a game starts)
- Reassign population upon getting a tech or getting a policy
- Recalculate resources always after getting a tech. Techs can give resources, so it may not be correct to assume that no new visible resources means no need to recalculate

I'm sure there's a better way to add the global checks than adding the checks to `determineStartingUnitsAndLocations` but I'm drawing blanks for now. I'm also sure the uniques that give a free unit should probably allow for placing a unit at a location and not just a city, but I'll leave that for another time

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -19,6 +19,8 @@ import com.unciv.models.ruleset.Ruleset
 import com.unciv.models.ruleset.RulesetCache
 import com.unciv.models.ruleset.unique.StateForConditionals
 import com.unciv.models.ruleset.unique.UniqueType
+import com.unciv.models.ruleset.unique.UniqueType.UniqueTarget
+import com.unciv.models.ruleset.unique.UniqueTriggerActivation
 import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.stats.Stats
 import com.unciv.models.translations.equalsPlaceholderText
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -169,21 +169,14 @@ object GameStarter {
     private fun addCivTechs(gameInfo: GameInfo, ruleset: Ruleset, gameSetupInfo: GameSetupInfo) {
         for (civInfo in gameInfo.civilizations.filter { !it.isBarbarian() }) {
 
-        //Catch uniques from techs that should've triggered already
-            for(tech in civInfo.tech.techsResearched) {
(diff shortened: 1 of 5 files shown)


SeventhM:
Interesting. Didn't expect a null pointer there. I'm now actually away from android studio to figure that out, but I'll look into it tomorrow

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -167,6 +169,18 @@ object GameStarter {
     private fun addCivTechs(gameInfo: GameInfo, ruleset: Ruleset, gameSetupInfo: GameSetupInfo) {
         for (civInfo in gameInfo.civilizations.filter { !it.isBarbarian() }) {
 
+        //Catch uniques from techs that should've triggered already
```
yairm210, line 172: Where are we adding these 'untriggered techs' in the first place, if not through addTechnology?

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -418,6 +432,15 @@ object GameStarter {
         for (civ in gameInfo.civilizations.filter { !it.isBarbarian() && !it.isSpectator() }) {
             val startingLocation = startingLocations[civ]!!
 
+            //Trigger any global uniques that should triggered.
+            //We may need the starting location for some uniques, which is why we're doing it now
+            for (unique in ruleset.globalUniques.uniqueObjects)
```
yairm210, line 437: Unify these into one for loop - for unique in a.asSequence() + b.asSequence()
Since we want to treat them the same way

core/src/com/unciv/logic/GameStarter.kt
SeventhM: They're added in `addCivilizations`, line 353 (or 367 of this PR). I did first immediately think to move those added techs to addCivTechs to allow for using addTechnology but then that adds in the notification that you've reached that era. It probably is safer to find a way to move it to here anyways, but then we either need to accept that it shows we made it to the ancient era on game start or unless we rework how addTechnology checks that

SeventhM:
> Or even maybe

I think I like this over giving addTechnology a parameter. I'll still give addTechnology a parameter for showing it's research notification (in case it comes up elsewhere as well), but it seems almost disturbing to me that the only reason it fails once instead of twice is you don't have a unit test for adding a policy. And since this is the common point of failure, no point in making the same fix in multiple places

yairm210 (approved):

PR #9690, opened 2023-06-28
Minimal support for Alpha Frontier-like Workers

SomeTroglodyte:
Commit messages tell a lot.

* The "[Add a chance](https://github.com/yairm210/Unciv/commit/ed2afa4b1c8a7db89ca647fe3f03b494073031fd)" one is half of the point of this. It's minimal as it only tries the water AI when land AI reached a medium level fail. Affects only the "workboats" unique.
* The second "half" just removes blockers that kept findTileToWork from working on water when a Worker has "Can build [Water] improvements on tiles" - [later commit](https://github.com/yairm210/Unciv/pull/9690/commits/04ca43deba1764a6524b8da0397fe10fe6dd3e33).

What would really help would be a fuzzy logic so improve land could return a best "quality" and create by sacrifice on water could do the same, both are compared the better ones win. Better result for more CPU. Concept could be extended to large parts of automation...

~~UNTESTED so far~~

Tests were done, by issue OP as well. See below. Looks like sizable improvement for little work (and little risk for G&K players).

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
@@ -6,7 +6,6 @@ import com.unciv.logic.battle.Battle
 import com.unciv.logic.battle.GreatGeneralImplementation
 import com.unciv.logic.battle.MapUnitCombatant
 import com.unciv.logic.city.City
-import com.unciv.logic.civilization.Civilization
 import com.unciv.logic.civilization.diplomacy.DiplomaticModifiers
 import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.Tile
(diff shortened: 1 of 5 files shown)


SeventhM:
Hmm... something tells me eventually we may need to deprecate the workboat unique. Oh well, for now we can work around it I guess

Heads up: I don't remember if `Can build [Water] improvements on tiles` functions correctly (I think I remember testing it in the past and it not working, but I could be wrong), but I do I know that if you have water improvement with a turnsToBuild, you can actually use that unique just fine. Perhaps try removing the checks for if the tile is land in `tileCanBeImproved`

hackedpassword:
> ... Also raises the question what would happen if some mod had a carthage-style "can climb mountains" 

Was reading through and this caught my attention. Mod {the great unciv rework} has 2 natural wonders [granting climber abilities](https://github.com/A1y0sh4/The-Great-Unciv-Rework/blob/6991c88c716a64951575a4389e9788edce4d6b3c/jsons/Terrains.json#L636C12-L636C12). My [in-development map] (https://github.com/hackedpassword/NextgenMaps-Labs/blob/main/maps/Castles%20v7) takes advantage of these mountains as designed gameplay. fyi

PR #9699, opened 2023-06-29
Check for triggers conditions

SeventhM:
Fixes triggers with trigger conditions triggering on game start

### DIFF
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -428,7 +428,9 @@ object GameStarter {
 
             //Trigger any global or nation uniques that should triggered.
             //We may need the starting location for some uniques, which is why we're doing it now
-            for (unique in ruleset.globalUniques.uniqueObjects + civ.nation.uniqueObjects)
+            val startingTriggers=(ruleset.globalUniques.uniqueObjects + civ.nation.uniqueObjects)
+                .filter { !it.conditionals.any { it.type!!.targetTypes.any { it.canAcceptUniqueTarget(UniqueTarget.TriggerCondition) } } }
+            for (unique in startingTriggers)
                 if(unique.isTriggerable)
                     UniqueTriggerActivation.triggerCivwideUnique(unique,civ, tile = startingLocation)
         }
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -428,7 +428,9 @@ object GameStarter {
 
             //Trigger any global or nation uniques that should triggered.
             //We may need the starting location for some uniques, which is why we're doing it now
-            for (unique in ruleset.globalUniques.uniqueObjects + civ.nation.uniqueObjects)
+            val startingTriggers=(ruleset.globalUniques.uniqueObjects + civ.nation.uniqueObjects)
+                .filter { !it.conditionals.any { it.type!!.targetTypes.any { it.canAcceptUniqueTarget(UniqueTarget.TriggerCondition) } } }
```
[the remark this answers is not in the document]
SeventhM: > !any == none?

Yes, though I didn't realize that none was an option there instead of !any

As for the explanation: I'm not sure the best way to add it as a comment. Maybe changing the names could help there, but a written explanation seems more appropriate in this situation as (imo) it's somewhat straightforwards once you're aware of how it functions. The trigger conditionals are supposed to be called when they come up, and the other conditionals are checked in `unique.conditionsApply`

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -428,9 +428,14 @@ object GameStarter {
 
             //Trigger any global or nation uniques that should triggered.
             //We may need the starting location for some uniques, which is why we're doing it now
-            for (unique in ruleset.globalUniques.uniqueObjects + civ.nation.uniqueObjects)
+            val startingTriggers = (ruleset.globalUniques.uniqueObjects + civ.nation.uniqueObjects)
+                //We don't want uniques with trigger conditionals. Those are called when they are relevant
+                .filter { unique -> unique.conditionals.none {
+                    conditional -> conditional.type!!.targetTypes.any {
+                        it.canAcceptUniqueTarget(UniqueTarget.TriggerCondition) } } }
```
yairm210, line 435: I think there are several places in code where we make this check, should probably be a function on the unique class

SeventhM: Sure I can do that. While I'm at it, I can fix the issue of unitTriggerConditions not being excluded in the areas where it previously did the check

yairm210 (approved):

Issue #9648, opened 2023-06-23
Feature request: Allow AI workers in mods like Alpha Frontier to get both land and water resources


hackedpassword:
Creating this on the coattails of the 4.7.3 update with the new unique "May travel on Water tiles without embarking".

Alpha Frontier's workers are designed to be both land and water worker units. Takes a little getting used to, when it clicks the strategy becomes apparent and desirable. Definitely mod-only feature that would never be accepted in vanilla.

The AI currently ignores all water resources, so playing against the AI is lopsided.

hackedpassword:
Adding related:
#7713
maybe #6509

hackedpassword:
I'll try to produce the saves in the scenarios you described. I've also reached out to AF's designer to review this issue.

hackedpassword:
In mod [NextgenMaps-Labs](https://github.com/hackedpassword/NextgenMaps-Labs) there are now two maps:
[AF_trogtest_9648](https://github.com/hackedpassword/NextgenMaps-Labs/blob/main/maps/AF_trogtest_9648) - source test map
[9648 full upgrades mid war](https://github.com/hackedpassword/NextgenMaps-Labs/blob/main/maps/9648%20full%20upgrades%20mid%20war) - mid way into war

Game is set to easy mode. Tiles/resources explicitly built for this scenario. In the war, several workers are embarked in positions that I think align to the requirements you gave. See how it works, hopefully this is what you were envisioning.

hackedpassword:
> Does that prove anything?

Looks like some water resources were improved, so it assuredly does prove something.

Is it known when constructors went for the water resources? Being that they sacrifice themselves, with caution I say, it would be useful to know what their rationale was (as in relativity) for deciding when to split from land duties. This might/prob is out of scope for the current test, but it is an important one to consider.

I'm thinking about it more and see that AI players need to make those decisions when to commit constructors to water improvements. If a human player hits automate, what is their intent? Land or water? 

A possible formula might include a running total of constructor bots on auto, every [5th] built triggers the bot fleet to assess best water resource available, closest bot deployed for the improvement. Now you have 4 bots on auto, perhaps another 5 on manual, then another 5 toggled to auto and one is deployed. Those numbers kinda suck whatever. 

Another possible formula could be after X turns and Y improvements made, per bot, that bot is deployed for the water. This sounds simpler. 

[...]

PR #9700, opened 2023-06-29
Automation constants

SomeTroglodyte:
Toying with #9648 I got the idea (again) that a moddable way to tune a lot of the hardcoded constants used in automation might be good. As a first step, just collect, name and document. Not meant to be final, so line lengths, wording, and names were not paid _too_ much attention to. I limited myself to scan just two of the AI files...

What do you think - moddability-worthy? Global or maybe even civ-specific? Or both, default and override option, usually null?

Accidentally based on the other #9648 PR - so, better decide that one first. The new commits are all named "Automation review - *" at the bottom.

Some actual code changes I am confident are transparent, but
- Storing an IntRange once - does that save many instance creations or take a away an optimization opportunity from the compiler
- Removing the Random call from headTowardsEnemyCity - expectedDamagePerTurn - calculateDamageToDefender: Seems obvious that for a "how soon can we conquer" decision taking the statistical average is fine, but - Does that reduce the chance that this randomness can break the AI out of a stalemate?

Testing limited to - it passes a next-turn

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -22,9 +22,40 @@ import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsPillage
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsUpgrade
 
 object UnitAutomation {
-
+    /** Range to search for enemy military units, then collect their reachable tiles to form a set of tiles for civilians to avoid */
+    private const val civilianEnemyUnitSearchRange = 5
+    /** Range to try for exploration, from the current position */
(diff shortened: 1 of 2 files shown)


yairm210:
My opinion on this is that some of these are really used in several different places - like 'min health to explore' - and there it makes sense to unify the declaration
But most of these are single-use, and extracting them into constants does nothing except add an extra layer of unnecessary abstraction

yairm210:
I don't see AI moddability being a thing, though
I mean, if they have a better AI, I want that *in the code*
If they have a better AI for their specific mod, I want that *in the code* after we figure out what separates their mod from the baseline
I don't see a use case where we would say "well actually only for this mod I want AI health threshold before attacking to be 70" or something. I just don't see it.

PR #9701, opened 2023-06-30
Fix linting warnings

alexban011:
this is draft because I'm still waiting for an answer regarding one of the warnings


yairm210:
Is there anything here besides a rename of NotificationIcon to NotificationIcons? Because if not I'll just do that rename and we can close this PR 🤔

alexban011:
Yeah.. but you can just revert 9475af5052b9d1819c87e3e4422ad5b246f118eb and maybe just rename the file instead of the actual class. This should be better, I just don't have time to do anything at the moment

alexban011:
So apparently I did find myself some time to fix this.

I reverted what SomeTroglodyte already fixed/explained in the mentioned PR (and apparently I also reverted another linting issue I resolved but whatever I'll do that another time) and know you can merge this @yairm210 if everything is oki doki

PR #9706, opened 2023-06-30
On City Raze, previous owner doesn't pay Road Maintenance

### LINKED ISSUE #9661: Roads that are not owned cost maintenance
Since some versions, neutral tile roads suddenly have maintenance costs. The bug is that the game doesn't properly calculate if road tiles are actually owned by a player. If you invade a city to raze it, after the razing is done the remaining neutral tile roads still require maintenance. I don't think they should. It is a major inconvenience to manually get rid of those random road tiles.

itanasi:
Fixed #9661


yairm210 (approved):

itanasi:
Well I stared at the cpp last night and concluded that yes this is the correct behavior. Tested it on a save by Razing a City and my maintenance went down next turn by the number of roads in the area.

Certainly could mod the previous "you always own this Road forever" behavior back in

PR #9709, opened 2023-06-30
Better support for lacking a capital

SeventhM:
This PR aims to better avoid crashes due to players lacking a capital for whatever reason. This is partly to better support scenarios where a player sells their capital, but also to fix a major issue with a large chunk of unit tests. I am aware of the previous arguments made in #9516 and #6889. However upon second glance, the issues are actually much deeper than those previous PRs made it out to be. Any scenario that required a city to do something was changed to instead use that civ's first city, though I'm not sure that's the best way to handle it

________
Full explanation here: Currently, there are many cases where buildings didn't trigger uniques, update connections (for harbors), add free buildings, or update city health. This oversight is good news for the unit tests as these checks are largely unnecessary there. However, it is a problem for mods, especially if a player starts in a later era and would have a building that is supposed to do one of these things. Unfortunately, updating these things, specifically city connections, has unforeseen consequences
Under normal conditions, cities do not check if they are connected to the capital unless a new turn has started or a building has finished being built. Since cities by default do not count as connected to the capital, `CityStats.getStatsFromTradeRoute()` will not crash trying to check the capital's population count (in an effort to update the citystats, line 499 calling line 460 calling line 103 in `logic.city.CityStats`). Fixing the prior issue of buildings not updating cities properly means that it will be treated as connected when you remove a city, and thanks to even more broken capital checks, it wouldn't properly clear itself as connected if you sell your capital, leading to a crash
This is where we have a problem in the unit tests. Many of the unit tests immediately removes the capital in order to use a capital that gives no stats. Previously, this only worked because settler buildings didn't update the city properly. If I attempt to fix that (as I was planning to do) many of the tests breaks. Meaning that whether we allow players to not have a capital or not, it is still required to at the very least fix the city connection code in order to allow settler buildings to update cities properly without breaking the unit tests. And since this has come up so many times by now, and given some mods wants to use this as a feature, it seems far more worthwhile to supports players not having a capital than continue saying that the game assumes there's always a capital

This PR does not fully fix for the AI not having a capital, primarily because that's not possible in any capacity that I can tell. Checking for getCapital checks that primarily affects the AI could be done, I just don't know if that's something we want to support

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -213,7 +213,7 @@ object Automation {
 
         // If we have vision of our entire starting continent (ish) we are not afraid
         civInfo.gameInfo.tileMap.assignContinents(TileMap.AssignContinentsMode.Ensure)
-        val startingContinent = civInfo.getCapital()!!.getCenterTile().getContinent()
+        val startingContinent = (civInfo.getCapital() ?: civInfo.cities.firstOrNull())!!.getCenterTile().getContinent()
         val startingContinentSize = civInfo.gameInfo.tileMap.continentSizes[startingContinent]
         if (startingContinentSize != null && startingContinentSize < civInfo.viewableTiles.size * multiplier)
             return false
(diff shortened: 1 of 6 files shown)


yairm210:
Since you seem to have "get capital OR the first city" in so many places, I have an alternative suggestion. Put that logic *inside the getCapital() function* so that it always returns a city if there is no capital but there are cities. We can gate this behind an optional parameter (returnAnyCityIfNoCapital:Boolean=false) if you see that this causes other problems

PR #9714, opened 2023-07-01
Added title gain from finishing a policy tree

random271:
Just a general idea about the implementation, added 2 classes - Leader and PolicyTitle + data for each title and for each leader and an updater for leader name upon completing one policy tree


yairm210:
Denied out of the box 
If you want there to be titles, these need to be uniques 

As for male-female, those can *also* be text-only uniques that you add to civs, and you decide with title unique to trigger using nationFilter conditionals
Like "Grants the title [King] <for [Male leader]>"

random271:
Got it, I'll try to implement it that way.

PR #9729, opened 2023-07-03
Fix civilopedia gold cost

Skekdog:
Previous formula applied exponent in the wrong place leading to wildly incorrect results

Fixes as reported at https://discord.com/channels/586194543280390151/1124750820735455232 (but uses Standard speed)

### DIFF
--- a/core/src/com/unciv/models/ruleset/Building.kt
+++ b/core/src/com/unciv/models/ruleset/Building.kt
@@ -240,9 +240,7 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
         if (cost > 0) {
             val stats = mutableListOf("$cost${Fonts.production}")
             if (canBePurchasedWithStat(null, Stat.Gold)) {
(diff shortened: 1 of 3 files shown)


Skekdog:
Just realised cost for buildings is also wrong, might as well fix it here

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -241,10 +241,10 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
             val stats = mutableListOf("$cost${Fonts.production}")
             if (canBePurchasedWithStat(null, Stat.Gold)) {
                 // We need what INonPerpetualConstruction.getBaseGoldCost calculates but without any game- or civ-specific modifiers
-                val buyCost = (30.0 * cost.toFloat().pow(0.75f) * hurryCostModifier.toPercent()).toInt() / 10 * 10
+                val buyCost = ((30.0 * cost.toFloat()).pow(0.75) * hurryCostModifier.toPercent() / 10).toInt() * 10
```
yairm210, line 244: Sounds like this should be a unified function in IConstruction

PR #9731, opened 2023-07-03
Fix buildings/units not triggering and golden age stat updates

SeventhM:
Part 2 to #9673
PR does the following
- Moves code from `building.postBuildEvent` to `cityConstruction.addBuilding`. This fixes buildings not updating or triggering as intended when added via tests, when founding a city, as a free building (i.e. free buildings that add more free builds didn't function before), when replacing previous buildings (i.e., when capturing a city), when liberating a city, or when moving a capital
- Adds code so units can trigger when added. It also updates the `Unique.hasTriggerConditional` code to include unit action modifiers (I know they technically aren't the same, so now the function name is inaccurate, but the intended use case seems the same)
- Updates city stats upon getting a golden age. Stats such as culture generation is visible on the mapscreen, so this needs to update immediately if an artist is used


SeventhM:
Strange. Those unit tests was literally the whole point of #9709. I'm pretty sure there was no issues on my end. I guess I'll need to investigate it tomorrow

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -479,16 +480,72 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
+
+        for (unique in civ.getTriggeredUniques(UniqueType.TriggerUponConstructingBuilding, StateForConditionals(civ, city)))
+            if (unique.conditionals.any {it.type == UniqueType.TriggerUponConstructingBuilding && building.matchesFilter(it.params[0])})
+                UniqueTriggerActivation.triggerCivwideUnique(unique, civ, city, triggerNotificationText = triggerNotificationText)
+
+        for (unique in civ.getTriggeredUniques(UniqueType.TriggerUponConstructingBuildingCityFilter, StateForConditionals(civ, city)))
+            if (unique.conditionals.any {it.type == UniqueType.TriggerUponConstructingBuildingCityFilter
+                    && building.matchesFilter(it.params[0])
+                    && city.matchesFilter(it.params[1])})
+                UniqueTriggerActivation.triggerCivwideUnique(unique, civ, city, triggerNotificationText = triggerNotificationText)
```
yairm210, line 514: This is starting to be a big function, I feel that this part of it is significant enough to extract to a separate function (triggerNotificationsForNewBuilding), YMMV

core/src/com/unciv/logic/civilization/managers/UnitManager.kt
```diff
@@ -71,9 +71,13 @@ class UnitManager(val civInfo:Civilization) {
         val unit = civInfo.gameInfo.tileMap.placeUnitNearTile(location, unitName, civInfo)
 
         if (unit != null) {
+            val triggerNotificationText = "due to gaining a [${unit.name}] "
```
yairm210, line 74: This is a new notification text so it needs to be added to template.properties (below "due to constructing [buildingName]")

yairm210:
All in all looks good :)

yairm210 (approved):

PR #9754, opened 2023-07-05
Added notification for destruction of tile improvements [via unit ability]

random271:
Case when it happens #9750


random271:
Thought this would be quick, studio is 50/50 on whether my laptop will freeze or not so I am saving on the electricity bills :)

yairm210:
Oh I just disabled those inspections I think

PR #9756, opened 2023-07-05
Little Diplomatic Victory makeover

SomeTroglodyte:
### Notes
* Warning: "Abstain" handling is a rule change. No sources - can't find any that still deal with G&K, they all limit themselves to BNW. If you think that wrong, the other changes are a little interspersed in the commit history, but separating could still be easily done.
* Any input on wording or punctuation of new text (see templates) welcome - better get them right now than patch later.
* Q: Should OtherIcons/Stop for the Abstain button be changed? Give it its own path for moddability? Better default icon suggestions?
* I skipped the issue of DiplomaticVoteResultScreen revealing un-met civs and not respecting randomNumberOfPlayers/randomNumberOfCityStates - pretty sure it should reveal (as it does now) if a leader was elected, but not otherwise - then going over the other source of sorted civs - but I'm out of steam for now, and lacking test games for that situation. Likelihood the UN builder wins and already knows all is pretty high after all. On an edited map with a wide mountain range partitioning the world maybe...

image
image

Too lazy to do a "before" comparison.

</details>

### Changes
* UN owner now counts as 2 votes - one checkbox in #4697
* Player can no longer vote for civs he never met, just as AI can't (so DiplomaticVotePickerScreen no longer leaks info you're not entitled to)
* Player can choose to abstain
* AI's abstain instead of voting for an unforgivable enemy, and will randomly abstain for "forgivable" enemies - I see their vote for the bad guy as picking the lesser evil
* DiplomaticVoteResultScreen improved, and with additional info if vote was indecisive (e.g. Back/ESC/Space now closes it. The # votes column is a bit verbose - but should we ever go BNW - then it would be more useful)
* One more turn handling better - currently, you get the vote result, then the victory screen, then "you won, wanna see the victory screen"
* Better sorting...

### DIFF
--- a/android/assets/jsons/translations/German.properties
+++ b/android/assets/jsons/translations/German.properties
@@ -1428,6 +1428,7 @@ Choose a civ to vote for = Wähle eine Zivilisation, für die du abstimmen möch
 Choose who should become the world leader and win a Diplomatic Victory! = Wähle, wer der Anführer der Welt werden und somit den Diplomatiesieg erhalten soll
 Voted for = hat gestimmt für
 Vote for [civilizationName] = Abstimmen für [civilizationName]
+Abstain = Enthalten
 Continue = Fortfahren
 Abstained = Enthalten
 Vote for World Leader = Stimme für den Anführer der Welt ab
--- a/android/assets/jsons/translations/German.properties
+++ b/android/assets/jsons/translations/German.properties
@@ -1426,11 +1426,19 @@ Worst = Schlechtester
 Turns until the next\ndiplomacy victory vote: [amount] = Runden bis zur nächsten Abstimmung\nüber den Diplomatiesieg: [amount]
 Choose a civ to vote for = Wähle eine Zivilisation, für die du abstimmen möchtest
 Choose who should become the world leader and win a Diplomatic Victory! = Wähle, wer der Anführer der Welt werden und somit den Diplomatiesieg erhalten soll
-Voted for = Abgestimmt für
+Voted for = hat gestimmt für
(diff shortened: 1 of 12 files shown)


Ouaz:
I see no `Abstain = ` entry added in template.properties. :D

android/assets/jsons/translations/German.properties
```diff
@@ -1426,11 +1426,20 @@ Worst = Schlechtester
...
 Choose who should become the world leader and win a Diplomatic Victory! = Wähle, wer der Anführer der Welt werden und somit den Diplomatiesieg erhalten soll
-Voted for = Abgestimmt für
+Voted for = hat gestimmt für
 Vote for [civilizationName] = Abstimmen für [civilizationName]
+Abstain = Enthalten
 Continue = Fortfahren
 Abstained = Enthalten
 Vote for World Leader = Stimme für den Anführer der Welt ab
+[number] votes = [number] Stimmen
+[number] vote = [number] Stimme
```
yairm210, line 1436: If this is always going to be 1 we don't need the brackets
Alternatively we could do "[amount] vote(s)"

yairm210 (approved):

android/assets/jsons/translations/German.properties
yairm210: ...they can use a non-breaking space character, that should technically work, interesting

yairm210: No you're right, hm
I got nothing
But no one has yet requested that, so I don't think we need to solve that problem if it's not a problem :)

PR #9763, opened 2023-07-06
Treat remaining untyped Uniques in default rulesets

### LINKED ISSUE #9761: RFC: Recent "Fixed India's 'double unhappiness' unique" should have been caught by a unit test
Ref: [Fixed India's 'double unhappiness' unique](https://github.com/yairm210/Unciv/commit/19bf15558d039edbd82c0470e32c218c376fe728)

* There's only a few untyped uniques left. Two are easy. I intend to fix them - but that raises the question of UniqueMap being keyed on strings... The commented move to UniqueType key - we can't, right, or mods using untyped uniques as keys to be matched in other uniques filters would break??? Or not - a filter remains a string and won't use that map anyway. So, let's think about that map's key move soon.
* "Who knows what the future holds?" - I would tend to simply move that to civilopediaText, right?
* "Units ending their turn on [Mountain] tiles take [50] damage" - Carthage - yes that does _nothing_. And it'd _better_ do nothing or Carthage's units would receive 100 damage on mountains...  I would tend to simply move that to civilopediaText too - but as is (already translated), or better in some form like "remember, all units, however they manage to enter them, take damage on mountains"? Or - revive uniqueText?
* "Aircraft" on UnitTypes - used as filter for carrier uniques. Hmmm... I'd type this one in the enum and comment it's only used as filter, and has no actual implementation. Right?

Once these are done we could step up BasicTests.baseRulesetHasNoBugs with `modCheckerAllowUntypedUniques = false` - except it's broken by tryFixUnknownUniques = false. With that fixed, and raising the error's level to WarningOptionsOnly from OK, the unit test **would have caught that India bug**. Tested.

But - there are also a whole bunch of specific tests (allUnitsUniquesHaveTheirUniqueTypes, allTerrainRelatedUniquesHaveTheirUniqueTypes,... ) - more than half of all RulesetObject types are covered, but not Nation. Ditch them all as a fixed RulesetValidator will catch them all as well?

SomeTroglodyte:
...And make existing unit test catch them. Closes #9761


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -476,6 +476,8 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     CarryAirUnits("Can carry [amount] [mapUnitFilter] units", UniqueTarget.Unit),
     CarryExtraAirUnits("Can carry [amount] extra [mapUnitFilter] units", UniqueTarget.Unit),
     CannotBeCarriedBy("Cannot be carried by [mapUnitFilter] units", UniqueTarget.Unit),
+    // Used only as mapUnitFilter marker for CarryAirUnits and CarryExtraAirUnits, has no actual implementation
```
yairm210, line 479: In that case this shouldn't be a unique here then

android/assets/jsons/Civ V - Gods & Kings/Techs.json
```diff
@@ -650,8 +650,9 @@
 				"name": "Future Tech",
 				"row": 5,
 				"prerequisites": ["Globalization","Particle Physics", "Nuclear Fusion", "Nanotechnology", "Stealth"],
-				"uniques": ["Who knows what the future holds?", "Can be continually researched"],
-				"quote": "'I think we agree, the past is over.' - George W. Bush"
+				"uniques": ["Can be continually researched"],
+				"quote": "'I think we agree, the past is over.' - George W. Bush",
+                "civilopediaText": [{"text": "Who knows what the future holds?"}]
```
yairm210, line 655: Will this be displayed in the tech tree? I think it won't, maybe it should 🤔

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: Why replace? Keep it as is, but not as a typed unique. Working as intended

yairm210: We can have this one untyped, because it's working as intended - it's not harcoded referenced anywhere, it's a *filtering* unique :D

yairm210: That last one

yairm210 (approved):
Looks good (to answer your question)

PR #9764, opened 2023-07-06
Fix issues from gaining free beliefs

SeventhM:
- Fixes `Gain a free [Any] belief` causing an infinite loop (it wasn't decrementing the amount of "any" belief you got, instead decrementing the belief type of the one you gained. You would get stuck choosing every belief until there is no more)
- Fixes beliefs chosen when using a prophet reducing the amount of free beliefs you got. Previously, it would count as using a free belief if you were choosing beliefs post prophet use, regardless of if you actually were choosing free beliefs or not. That would decrement the amount of free beliefs when it shouldn't

____
Leaving this as a draft initially as there's still a few things I'm not sure is ideal
1. If you get multiple pantheons upon founding a pantheon, you have to choose them one at a time. The picker screen as is doesn't allow for picking more than one. Not surer if it's out of scope to fix that here, but it seems unnecessarily slow to get multiple pantheons
2. It seems weird to me that free beliefs are chosen separately than prophet beliefs, especially given how Byzantium works. I'm aware doing things like that is technically how Civ 5 works, but it seems unnecessarily tedious and it comes with a few side effects. Namely, if you just picked free beliefs when you got them, even when you get them from founding/enhancing, you wouldn't actually need the extra step of tracking which came from "Any" and which didn't. We can just clear all of the free beliefs and call it a day
3. Speaking of tracking where a belief is being chosen, I'm almost certain the helper class in NextTurnAutomation (which I copied from the religion screen) is misplaced. Not sure where it should be

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -646,8 +646,7 @@ object NextTurnAutomation {
         val chosenPantheon = chooseBeliefOfType(civInfo, BeliefType.Pantheon)
             ?: return // panic!
         civInfo.religionManager.chooseBeliefs(
-            listOf(chosenPantheon.belief!!),
-            emptyList(),
+            listOf(chosenPantheon),
             useFreeBeliefs = civInfo.religionManager.usingFreeBeliefs()
         )
     }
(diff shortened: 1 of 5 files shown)


SeventhM:
@SomeTroglodyte I'm confused as to why this was closed

SeventhM:
Ok, Reverting some of the changes outside of the religion manager if we're having free beliefs immediately be available upon founding/expanding. This still doesn't fix the ui issues I have with the pantheon picker (again, probably outside of scope), but it's good enough for me to be willing to have it merged

Known issue that I may look into as a separate PR: we need to limit the amount of free beliefs by the amount of available beliefs a bit more often than we currently do. It is possible (and seemingly was always possible) to have more free beliefs than there are available beliefs after a different civ founds. It doesn't seem worth fixing that in this PR, but it is something I noticed

yairm210 (approved):

PR #9774, opened 2023-07-09
City-level resources

yairm210:
Allows designating resources as city-level resources (think: Civ VI artifact slots/power, Civ IV 'health', etc)

What's really cool is that resources from tiles work as expected, which means you can make city-specific resource chains all the way from base tiles :D

City-level resources are...
- Not displayed in civ top bar
- Not considered civ-level resources (for trade, non-city conditionals, etc)
- Are explicitly treated in city conditionals
- ARE considered for constructing buildings requiring resource
- ARE NOT considered for construction units requiring resource

Example (Civ6 mod, Districts):
image

@SeventhM Since you're the primary "modder and dev" I would like your feedback on this

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -254,6 +254,14 @@ class City : IsPartOfGameInfoSerialization {
         return cityResources
     }
 
+    fun getResourceAmount(resourceName:String): Int {
+        val resource = getRuleset().tileResources[resourceName] ?: return 0
+
+        if (resource.hasUnique(UniqueType.CityResource))
+            return getCityResources().firstOrNull { it.resource == resource }?.amount ?: 0
+        return civ.getCivResourcesByName()[resourceName]!!
+    }
+
     private fun getTileResourceAmount(tile: Tile): Int {
         if (tile.resource == null) return 0
         if (!tile.providesResources(civ)) return 0
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -258,7 +258,7 @@ class City : IsPartOfGameInfoSerialization {
         val resource = getRuleset().tileResources[resourceName] ?: return 0
 
         if (resource.hasUnique(UniqueType.CityResource))
-            return getCityResources().firstOrNull { it.resource == resource }?.amount ?: 0
+            return getCityResources().filter { it.resource == resource }.sumOf { it.amount }
         return civ.getCivResourcesByName()[resourceName]!!
     }
 
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -254,12 +254,14 @@ class City : IsPartOfGameInfoSerialization {
         return cityResources
     }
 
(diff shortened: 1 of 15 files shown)


SeventhM:
Sorta looking at it at a quick glance rn, will probably look with further detail later, but

> * ARE NOT considered for construction units requiring resource

Is there a way to make it considered for constructing units? I could see a modder trybto use this as a more fine grained way to limit how many of a type of unit you get per city

yairm210:
That sounds way out of scope - that would mean we need to track what unit came from what city, and deal with situations like what happens if the city is taken over/destroyed

core/src/com/unciv/logic/city/City.kt
```diff
@@ -254,6 +254,14 @@ class City : IsPartOfGameInfoSerialization {
         return cityResources
     }
 
+    fun getResourceAmount(resourceName:String): Int {
+        val resource = getRuleset().tileResources[resourceName] ?: return 0
+
+        if (resource.hasUnique(UniqueType.CityResource))
+            return getCityResources().firstOrNull { it.resource == resource }?.amount ?: 0
```
SeventhM, line 261: If you're getting resources from multiple sources (ie, both improvements and buildings), this only finds one of these sources. You'd want the sum of that resource, not the first listing of it

core/src/com/unciv/ui/screens/cityscreen/CityStatsTable.kt
```diff
@@ -178,6 +178,15 @@ class CityStatsTable(private val cityScreen: CityScreen): Table() {
             tableWithIcons.add("In resistance for another [${city.getFlag(CityFlags.Resistance)}] turns".toLabel()).row()
         }
 
+        val resourceTable = Table()
+        for (resourceSupply in city.getCityResources())
```
SeventhM, line 182: Likewise, you want the sum here too. Otherwise, it shows each source of the resource separately

SeventhM:
Looks fine enough to me at first look over. If anything, I wish there was some way to view where each source of the resource is coming from (even if it's just "this is from buildings, that from improvements") in the city view, but I'm not sure on how the ui should look on that one. Probably a question to ask the discord

core/src/com/unciv/ui/screens/cityscreen/CityStatsTable.kt
```diff
@@ -178,6 +180,18 @@ class CityStatsTable(private val cityScreen: CityScreen): Table() {
...
+
+        val resourceCounter = Counter<TileResource>()
+        for (resourceSupply in city.getCityResources()) resourceCounter.add(resourceSupply.resource, resourceSupply.amount)
+        for ((resource, amount) in resourceCounter)
+            if (resource.hasUnique(UniqueType.CityResource)){
+                resourceTable.add(amount.toLabel())
+                resourceTable.add(ImageGetter.getResourcePortrait(resource.name, 20f))
+                    .padRight(5f)
+                }
+        tableWithIcons.add(resourceTable)
```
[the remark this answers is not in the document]
yairm210: The copy is the flattening of different sources, as pointed out by Seventh above
Agree with the "don't add if empty" :)

PR #9790, opened 2023-07-11
Policy cost modconstants

Dimensiondog:
* docs updated
* basepolicycost
* numpolicyadoptedmultiplier
* policycostexponent

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/PolicyManager.kt
@@ -124,8 +124,7 @@ class PolicyManager : IsPartOfGameInfoSerialization {
     // from https://forums.civfanatics.com/threads/the-number-crunching-thread.389702/
     // round down to nearest 5
     fun getCultureNeededForNextPolicy(): Int {
-        var policyCultureCost = civInfo.gameInfo.ruleset.modOptions.constants.policyBaseCost + (numberOfAdoptedPolicies * civInfo.gameInfo.ruleset.modOptions.constants.numPoliciesCostMultiplier).toDouble().pow(civInfo.gameInfo.ruleset.modOptions.constants.policyCostExponent)
-        // https://civilization.fandom.com/wiki/Map_(Civ5)
+var policyCultureCost = civInfo.gameInfo.ruleset.modOptions.constants.run { policyBaseCost + (numberOfAdoptedPolicies * numPoliciesCostMultiplier).toDouble().pow(policyCostExponent) }        // https://civilization.fandom.com/wiki/Map_(Civ5)
         val worldSizeModifier = with(civInfo.gameInfo.tileMap.mapParameters.mapSize) {
             when {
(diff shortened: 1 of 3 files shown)


Dimensiondog:
Yes I couldn't get studio to work so I just worked on github.

Dimensiondog:
Fixed modconstant name and took your advice and made more concise.

Dimensiondog:
> "couldn't get studio to work"? Shouldn't be hard. On apt-based Linuxes [my old notes](https://github.com/SomeTroglodyte/Scratchpad/blob/master/Unciv-notes/Getting-Started.md) might help.

No im just on an older computer and don't have enough ram

yairm210:
This actually runs counter to a long term plan of mine - moddable calculations - so while I don't have a problem with the concept itself, I'll have to say no to this

Dimensiondog:
> This actually runs counter to a long term plan of mine - moddable calculations - so while I don't have a problem with the concept itself, I'll have to say no to this

what is moddable calculations?

yairm210:
Right you're not in Discord, I'll error here what I wrote there


Whoof, where to start... 


Currently, there are a lot of calculations (policy cost and unit upkeep costs are good examples) that hardcoded depend on certain other variables.
Your suggestion is to make the individual elements moddable, `policyBaseCost + (numberOfAdoptedPolicies * numPoliciesCostMultiplier).toDouble().pow(policyCostExponent)` 
But there's also a second part to this calculation, `(policyCultureCost * (1 + cityModifier)).roundToInt()` where cityModifier is also dependant on map size.

That's a heckuva lot of different variables. But what if I don't want policy cost to depend on number of adopted policies? What if I don't want extra cities to affect me? What if instead I want to, in fact have a *completely different calculation*?

---

My long-term plan is to make modders be able to specify these *entirely different* calculations.
Imagine if you will that you could define variables in a json file, like 'policyCost': 'policyBaseCost + (numberOfAdoptedPolicies * numPoliciesCostMultiplier)^policyCostExponent ' - and Unciv would evaluate that string to a number. You could then use any building block numbers that appear in any other calculation to mix things up. Want number of techs to increase policy cost? Number of wonders to decrease it? Whatever! You can use a list of *base game-provided numbers* to define *modded values*.

But wait there's more! If we already have calculated, named numbers, we can use them as inputs for `[amount]` parameters as well!

---

Of course, this will all require **a lot** of work. But we've done even more work than this when defining the filters, and conditionals, and even the uniques themselves.

I would rather not add small pieces to modify what I feel are pretty hardcoded calculations, when I intend to eventually forgo those entirely.

yairm210:
I guess the real question is how non-coder-user-friendly we want this to be

Dimensiondog:
I could probably make moddable constants where it was a json with the formula and modders could just put the formula any formula that would work if put straight into unciv code.
For example:
"policycostformula": 25 + (numberOfAdoptedPolicies * 6).toDouble().pow(1.7)

yairm210:
In any case, this should be closed

PR #9793, opened 2023-07-11
UniqueTarget compliance test

### LINKED ISSUE #9773: the Pantheon of Craftsmen does not work
There is a bug that has been going on for a long time, which means that the Pantheon of Craftsmen does not work. Further, the entry "<in cities with at least [amount] [populationFilter]>" is not valid. Screenshot below.
H5AW FJ8K 82I_UV60TP(`G
M09``AE8PE08WV203S{YG)E_tmb

SomeTroglodyte:
Triggered by and closes #9773

[Unit test run after "Implement" commit](https://github.com/yairm210/Unciv/actions/runs/5519973727/jobs/10065961785?pr=9793)
[Unit test run after "Fix wrong Stats target" commit](https://github.com/yairm210/Unciv/actions/runs/5520082476/jobs/10066211328?pr=9793)
[Unit test run after "Fix two Pantheon uniques" commit](https://github.com/yairm210/Unciv/actions/runs/5520193114/jobs/10066474016?pr=9793)

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTarget.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTarget.kt
@@ -15,8 +15,6 @@ import com.unciv.models.ruleset.unit.BaseUnit
  * @param documentationString Copied to uniques.md by `UniqueDocsWriter`
  * @param inheritsFrom means that all such uniques are acceptable as well. For example, all Global uniques are acceptable for Nations, Eras, etc.
  * @param isModifier Marks types that are allowed _exclusively_ in a "<conditional>" style modifier, as standalone it is applicable to nothing.
- *
- * //todo implement "other half" of isModifier -> conditionals and triggers should always have it
  */
 enum class UniqueTarget(
     val documentationString:String = "",


core/src/com/unciv/models/ruleset/unique/UniqueTarget.kt
```diff
@@ -0,0 +1,113 @@
...
+        inheritsFrom = Global, classes = listOf(Belief::class),
+        additionalFilter = { belief -> (belief as Belief).type.isFounder } ),
+    FollowerBelief("Uniques for Pantheon and Follower type beliefs, that will apply to each city where the religion is the majority religion",
+        classes = listOf(Belief::class),
+        additionalFilter = { belief -> (belief as Belief).type.isFollower }),
+
+    // City-specific
+    Building(inheritsFrom = Global, classes = listOf(com.unciv.models.ruleset.Building::class),
+        additionalFilter = { building -> !(building as com.unciv.models.ruleset.Building).isAnyWonder() }),
+    Wonder(inheritsFrom = Building, classes = listOf(com.unciv.models.ruleset.Building::class),
```
yairm210, line 62: There should be no wonder uniques. Sounds like a different unit test huh

core/src/com/unciv/models/ruleset/unique/UniqueTarget.kt
```diff
@@ -0,0 +1,113 @@
...
+        additionalFilter = { building -> (building as com.unciv.models.ruleset.Building).isAnyWonder() }),
+
+    // Unit-specific
+    Unit("Uniques that can be added to units, unit types, or promotions"
+        , inheritsFrom = UnitTriggerable, classes = listOf(BaseUnit::class)),
+    UnitType(inheritsFrom = Unit, classes = listOf(com.unciv.models.ruleset.unit.UnitType::class)),
+    Promotion(inheritsFrom = Unit, classes = listOf(com.unciv.models.ruleset.unit.Promotion::class)),
```
yairm210, line 69: Same for these - all 3 of these should be able to get 'Unit' promotions.

core/src/com/unciv/models/ruleset/unique/UniqueTarget.kt
yairm210: And since it inherits from building anyway, do we need the 'classes'?

yairm210:
Nah just wondering

yairm210:
Okay this has been open for a while and every time it's something else, let's break it down

- unique type numbering, no problem
- json fixes, no problem
- belief type changes, no problem

But I do have a problem with the unique target knowing the classes, that is unacceptable

And now I finally know what the correct solution is

Every ihasuniques *already has* a UniqueTarget field, we can use that to determine compliance

If you want to split the PR in 2 for the 'no problem' stuff to be merged now, that's preferable to me, but not mandatory

PR #9795, opened 2023-07-12
Update Czech.properties

jahodovykolac:


jahodovykolac:
I forgot the brackets, but there was a mistake anyway, let me fix it real quick

PR #9797, opened 2023-07-13
Overhaul NUKE code to behave closer to original

### LINKED ISSUE #9779: Roads created by cities can get pillaged by nukes
**Platform**
Android 13

**Version**
Google Play, 4.7.7, build 889

**Describe the bug**
Roads created by city center can get pillaged by nukes. However, they can't get repaired, but also don't appear as pillaged in the infobox at the right. The tile on the other hand is tinted red, and units take normal movement points.

**Expected behavior**
The roads don't get pillaged or they can be repaired.

**Screenshots**
Screenshot_20230710_104437_Unciv

SomeTroglodyte:
Fixes #9779 - [This comment is relevant](https://github.com/yairm210/Unciv/issues/9779#issuecomment-1632581142).
Reference: original code #4211, we're still pretty close to that. 

### Rule changes:
* Nuke Level >= 3, Non-Original Capital >=5 pop: Before - set to 1 pop; After: Destroyed
* Population loss random distribution: Before - uniform over 30/60 to 80 range; After: triangular
* Damage to units Nuke Level 1: Before - uniform 40-100; After: triangular 30-110 as per Civ5 - diagram below
* Ground zero no longer ensures unit destruction, instead it is treated as if Nuke Level 2 with possible damage mitigation - as per Civ5 sources
* Missing resource left as damage mitigation, Bomb Shelter now mitigates garrison damage too, even if the city is destroyed by the nuke - as per Civ5 source
* Civilians "only" get destroyed if the predicted damage (including level, ground zero, random, and mitigations) is >= 60 - as per Civ5 - but pre-existing damage added in - unlike Civ5 which has no civilian damage at all afaik. Rare and can affect Carthage and modded only. I was very tempted to convert the rule to projecting the damage distribution such that resulting death probability for 100 health units is the same, but - more difficult to calculate, and would be significantly more powerful (long-term).
* Pillages improvements only if adding Fallout - as per Civ5: Normal 50% dice on land/non-ground-zero/non-impassable/no-city will either Pillage+Fallout or nothing. Modified - with DestroyableByNukesChance features their probability is used instead. Modified again - Water and Impassable tiles treated same as Land for chance calculations and pillaging, then the actual Fallout is skipped (mods can have destructible Atolls or Mountain improvements - the Fallout skip on Mountains should be more complicated in that case, but I opted for a simpler compromise). This also keeps Unciv nukes affecting water improvements - Civ5 ones don't.
* 100% chance for Fallout on ground zero (unless there's a surviving city there) - side effect: Forests on ground zero always destroyed - seems more aligned with Civ5 where I see no feature removal from nukes at all.
	
### Not changed - but different from Civ5 (according to how I read the sources)
* Nuke Level 0 on Cities: Civ5 - same pop loss as level 1, ours: none
* Civ5 doesn't treat Nuke Level > 2 on Original Capitals any differently than Level 2, we maximize population loss
* Civ5 has no interception for nukes - we do
* Civ5 never applies tile effects on city tiles even if the nuke destroys the city - we do
* Civ5 never touches Routes (as it calls roads internally - "roads" "exist" only for display) - we do
* Civ5 ignores "missing resource" - and certainly does not allow nuking barbarians the full strength despite them having no resources
* In Civ5, having tiles or units belonging to an unknown civ in the blast radius forbids the attack entirely, we instead ignore the test whether declaring war would be allowed for them
* Nuke Level 0 on units: Civ5 - same as level 2 (bug), ours: 20-50 linear distribution as before
* Civ5 has code protecting tiles with a top terrain feature having nuke immunity - such an immunity can be defined in XML or LUA, on units promotions buildings and features - but I'm pretty sure Civ5 G&K has none - can be simulated with a DestroyableByNukesChance(0) - this PR has an extra test for that (ground zero overrides the chance unless the chance says zero).

### Question
* Parameterize the hardcoded constants - where? Not ModConstants, more like a Unique on the nuke?

### Draft while...
* Actually test it

<details><summary>Histogram unit damage distribution</summary>

20k sample size, 50 buckets:
image

Note - damage reduction comes before the capping to 100, so in that case the histogram would be symmetrical.
</details>


yairm210:
Sorry, I don't have the headspace to answer these questions :(

yairm210 (approved):

PR #9806, opened 2023-07-15
Avoid built buildings

SeventhM:
PR attempts to removes almost every call to cityConstructions.builtBuildings. Most other systems does not use it (it uses builtBuildingObjects in some way or form) and it opens the way for possible other bugs. The primary places where it was left was in setTransients, the add/remove functions, the clone function, and any places where the other PR of mine already removed the intraction. The only place where I'm not sure builtBuildingObjects is necessarily better is for the wonder overview tab, but everything else should be consistent with how it looks like it should be used

There's the function for `changeBuildingNameIfNotInRuleset`. From what I can tell, it's currently only used for the Hanse, it was done 2 years ago (past deprecation period likely), and seems imo more likely to cause issues in mods than it is to help (given it was done 2 years ago). This PR removes the call to the function for now

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -185,7 +185,7 @@ class CityConstructions : IsPartOfGameInfoSerialization {
 
     fun getCurrentConstruction(): IConstruction = getConstruction(currentConstructionFromQueue)
 
-    fun isAllBuilt(builingList: List<String>): Boolean {
+    fun isAllBuilt(buildingList: List<String>): Boolean {
         for(building in buildingList)
         {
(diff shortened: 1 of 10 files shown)


core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -185,7 +185,16 @@ class CityConstructions : IsPartOfGameInfoSerialization {
 
     fun getCurrentConstruction(): IConstruction = getConstruction(currentConstructionFromQueue)
 
-    fun isBuilt(buildingName: String): Boolean = builtBuildings.contains(buildingName)
+    fun isAllBuilt(buildingList: List<String>): Boolean {
```
yairm210, line 188: = buildingsList.all {isBuilt(it)}

PR #9809, opened 2023-07-17
Prevent deserialization problem with Espionage

SomeTroglodyte:
There was an issue where I commented on this, but can't find it.
With current master, no save can be reloaded that has a non-empty erasSpyEarnedFor.

* Too much comment?
* Sniffing into Espionage - I would be tempted to allow unique quotes and portraits per individual Spy, at least as Mod. Imagine a mugshot taken from wikipedia plus the quote "A harlot? Yes, but a traitoress, never! - Phrase attributed to margaretha Zelle during the trial" for "Mata Hari"...
    Civ5 itself seem not to have such a feature... Yes/No, and if Yes - data schema? Extra Json, mapping decorations per Spy name / deprecate spyNames in favour of an array of SpyMetadata objects / custom deserializer that reads both? Display on-click in Espionage screen or on getting a Spy as AlertPopup like the Tech/Wonder quotes?

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -56,6 +56,13 @@ import java.util.UUID
  * When you change the structure of any class with this interface in a way which makes it impossible
  * to load the new saves from an older game version, increment [CURRENT_COMPATIBILITY_NUMBER]! And don't forget
  * to add backwards compatibility for the previous format.
+ *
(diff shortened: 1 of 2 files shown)


yairm210:
Yeah you're commenting why you make the change, but the next dev who comes will be like 'why is it taking about mutableSetOf if it doesn't appear anywhere?"

yairm210 (approved):

PR #9811, opened 2023-07-18
A few more useful notification actions

SomeTroglodyte:
... by going to Phase IV for the Notifications migration ...

And miscellaneous - see commit messages


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -581,7 +583,12 @@ object Battle {
         }
 
         if (!thisCombatant.isDefeated() && !unitCouldAlreadyPromote && promotions.canBePromoted())
-            civ.addNotification("[${thisCombatant.unit.displayName()}] can be promoted!",thisCombatant.getTile().position, NotificationCategory.Units, thisCombatant.unit.name)
+            thisCombatant.run {
```
yairm210, line 586: I really don't see the need to put this in a 'run' scope, it would be clearer without...

yairm210 (approved):

PR #9824, opened 2023-07-20
NotificationAction compatibility patch

SomeTroglodyte:
... see #9811 comments


yairm210:
I'm releasing a patch version now, I'll add this in :)

PR #9828, opened 2023-07-21
Key bindings CityScreen

SomeTroglodyte:
See commit history

Side effect: Reordering queue with taps/clicks keeps selected entry in view (done 'cuz overdue and when moving with keys it's more poke-you-in-the obvious); and it's faster with the mouse too (skips the fat DTO build - the delay felt really bad with the keys)

Questions:
* I assigned keys to open/close expanders, then tried to scroll them into view - that has the little problem with the animation, as at the time it acts the geometry is not final. Not pretty but mostly harmless - except when Units are open and one hits O for the "Other" category, then the imperfection is noticeable. Unsolvable afaik atm, but probably straightforward with the outstanding ExpanderTab animation update (which simmers here due to not finding a game to repro those broken layouts). So - ignore and leave it in or limit usefulness of those keys and deactivate that part...
* When to remove the "Secret page" hiding flag

Refrained from touching:
* Selection in available lists doesn't scroll-into-view, may feel a little inconsistent
* Invalid Queues still possible (e.g. Gold moved up with never-reached constructions after it)
* Trying to make keyboard operation without any mouse/ff help feasible

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -9,7 +9,7 @@ import com.unciv.logic.civilization.AlertType
 import com.unciv.logic.civilization.NotificationCategory
 import com.unciv.logic.civilization.NotificationIcon
 import com.unciv.logic.civilization.PopupAlert
-import com.unciv.logic.map.mapunit.MapUnit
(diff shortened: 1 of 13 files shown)


yairm210:
I would rather limit usefulness of keys rather than have a janky user experience :)

PR #9832, opened 2023-07-22
more Persian UN Translation

ilia-abbasi:
I added some translations as much as I could.
The last commit is the translations and fixes from line 5000 till the end.


ilia-abbasi:
Fixed the typos which were mentioned in the failed check.

PR #9834, opened 2023-07-23
New unique: May purchase other civs tiles

Dimensiondog:
Draft as I still haven't figured out how to build unciv to test and also unsure how to make sure a player cannot buy their own tiles.


yairm210:
This seems ridiculously overpowered

core/src/com/unciv/logic/city/managers/CityExpansionManager.kt
```diff
@@ -58,7 +58,13 @@ class CityExpansionManager : IsPartOfGameInfoSerialization {
     fun canBuyTile(tile: Tile): Boolean {
         return when {
             city.isPuppet || city.isBeingRazed -> false
-            tile.getOwner() != null -> false
+            tile.getOwner() != null -> {
```
yairm210, line 61: If X true else false is exactly like just saying "X"

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -136,6 +136,8 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     /// Border growth
     BorderGrowthPercentage("[relativeAmount]% Culture cost of natural border growth [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
     TileCostPercentage("[relativeAmount]% Gold cost of acquiring tiles [cityFilter]", UniqueTarget.FollowerBelief, UniqueTarget.Global),
+    CanPurchaseOtherCivTiles("May purchase other Civs Tiles", UniqueTarget.FollowerBelief, UniqueTarget.Global),
```
yairm210, line 139: As a *follower belief* means that this needs to take the city unique into account, which isn't what's happening now

Dimensiondog: > As a _follower belief_ means that this needs to take the city unique into account, which isn't what's happening now

Isn't the check for if a city can purchase a tile city based?

core/src/com/unciv/logic/city/managers/CityExpansionManager.kt
Dimensiondog: Im confused by this

yairm210: "If (<condition>) true else false" is exactly the same as just writing the condition

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: The uncivInfo.hasUnique you wrote should check if the *city* has the unique

Dimensiondog:
Changed the code hope it works better now.

alexban011:
```
e: file:///home/runner/work/Unciv/Unciv/core/src/com/unciv/logic/city/managers/CityExpansionManager.kt:61:32 Unresolved reference: civ
e: file:///home/runner/work/Unciv/Unciv/core/src/com/unciv/logic/city/managers/CityExpansionManager.kt:61:40 Unresolved reference: uncivInfo
```

Well for the first error you need to call `city.civ` instead of just `civ`

For the second error I also replaced that `uncivInfo` with `city.civ` but you DO NEED to test yourself if it actually works

yairm210:
No, you need to use the *city* to get the uniques, if you want this to be a follower belief
As stated above

PR #9835, opened 2023-07-23

AutumnPizazz:


android/assets/jsons/translations/Simplified_Chinese.properties
```diff
@@ -302,6 +302,7 @@ National ability = 民族特性
 [firstValue] vs [secondValue] = [firstValue]对[secondValue]
 Gained = 获得
 Lost = 失去
+vs = 对战
```
yairm210, line 305: This is something we'll be to add on our side, is currently not translated at all - that's on me

AutumnPizazz: Okay.

Issue #9836, opened 2023-07-23
AI does not replace forts


tuvus:
**Platform**
Windows
Android

**Version**
4.7.9-patch 1

**Describe the bug**
Workers will not go out of their way to replace fort improvements with more beneficial improvements. There may be a few cases when they do while standing on the tile though.

Players can "cheese" AI factions with open borders by placing forts over all of their resources and tiles, significantly weakening the AI's economy.

**To Reproduce**
Either test it out on your own map by building forts on another AI civilization's tiles or create a new map designed to quickly spot the error.

1. Create a new custom map of tiny size. The land should all be grassland and pain forts over the entire map.
2. Play the map with only you and another civilization on industrial age to spawn with a few workers.
3. Explore the map to find the other civilization. They aren't replacing the forts with farms even though they should be.

**Expected behavior**
AI should replace forts when the advantage of using other improvements outweighs the fort.

PR #9837, opened 2023-07-23
Fixed automated workers not replacing forts for AI civ.

tuvus:
This fixes the issues described in #9836. Workers will now replace forts on their land. They will also immediately cancel in-progress forts (although it needs a little more testing to confirm that the new improvement did not lose a turn of construction).

To prevent automated workers from removing player Civ forts, automated workers don't replace forts for the player Civ.

Note: From what I have seen, the AI does not seem to build forts before this change. I doubt this changed it in any way, but it could reduce future commits to improving new fort placement by the AI.

Also Note: There were a few lines in WorkerAutomation.kt from lines 136-152 that referenced the old tile (currentTile) instead of the new tile (reachedTile) the worker moved to that I changed. Logically the previous version didn't make sense to me. This does change non-fort-related behaviors.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -162,13 +162,6 @@ class WorkerAutomation(
             return
         }
 
-        //If we have reached a fort tile that is in progress and shouldn't be there, cancel it.
-        if (currentTile.improvementInProgress == Constants.fort
-                && !evaluateFortSuroundings(currentTile, false)) {
-            debug("Replacing fort in progress with new improvement")
-            currentTile.stopWorkingOnImprovement()
-        }
-
         if (currentTile.improvementInProgress == null && tileCanBeImproved(unit, currentTile)) {
             debug("WorkerAutomation: ${unit.label()} -> start improving $currentTile")
             return currentTile.startWorkingOnImprovement(chooseImprovement(unit, currentTile)!!, civInfo, unit)
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -556,8 +556,8 @@ class WorkerAutomation(
      * @return Yes the location is good for a Fort here
      */
     fun evaluateFortPlacement(tile: Tile, isCitadel: Boolean): Boolean {
-        return evaluateFortSuroundings(tile,isCitadel)
-            && tile.improvement != Constants.fort // don't build fort if it is already here
+        return tile.improvement != Constants.fort // don't build fort if it is already here
+            && evaluateFortSuroundings(tile,isCitadel)
     }
 
     private fun hasWorkableSeaResource(tile: Tile, civInfo: Civilization): Boolean =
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -129,20 +129,27 @@ class WorkerAutomation(
             debug("WorkerAutomation: %s -> head towards %s", unit.label(), tileToWork)
             val reachedTile = unit.movement.headTowards(tileToWork)
             if (reachedTile != currentTile) unit.doAction() // otherwise, we get a situation where the worker is automated, so it tries to move but doesn't, then tries to automate, then move, etc, forever. Stack overflow exception!
+
+            //If we have reached a fort tile that is in progress and shouldn't be there, cancel it.
+            if (reachedTile == tileToWork && (reachedTile.improvementInProgress == Constants.fort && !evaluateFortSuroundings(currentTile, false))) {
```
yairm210, line 134: We don't need the ectra parentheses here :)

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -155,6 +162,13 @@ class WorkerAutomation(
             return
         }
 
+        //If we have reached a fort tile that is in progress and shouldn't be there, cancel it.
+        if ((currentTile.improvementInProgress == Constants.fort
```
yairm210, line 166: Double parentheses again

yairm210: Why do we need this twice in the same function? Looks the same as line 134

yairm210 (changes requested):
Overall looks good! And welcome 🤗

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
tuvus: The lines 134-154 don't make sense now that I look at them more closely. Here is my interpretation of what it was trying to do, but I don't agree with what I perceive to be the goal of it.
If a worker moves from a tile that does not have any improvements scheduled (mostly from running away) and ends up moving to it's destination tile without improving anything on the tile. Then the worker will improve the tile that it came from.
This might not be desired behavior because manually controlled workers cannot repeat this functionality.

Line 134 has reachedTile == tileToWork which doesn't make sense to me and the inconsistent use of currentTile and reachedTile in 135 and 136 are confusing me.

If we want to keep this functionality we will need to remove the changes I made replacing currentTile with reachedTile. However I recommend reverting the replacements and fixing this in another commit because it is not relevant to the purpose of this commit.

yairm210: Between the first and second batches there is no worker movement going on.
I think we can keep the first (line 135) so that other improvements can be scheduled, and delete the second (line 166) because it will always be false

tuvus: Agreed, it seems to be working without line 164-170, logically it makes sense as well that it is unnecessary.

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -538,6 +549,17 @@ class WorkerAutomation(
...
     }
 
+    /**
+     * Do we want to build a Fort [here][tile] considering surroundings?
+     *
+     * @param  isCitadel Controls within borders check - true also allows 1 tile outside borders
+     * @return Yes the location is good for a Fort here
+     */
+    fun evaluateFortPlacement(tile: Tile, isCitadel: Boolean): Boolean {
+        return evaluateFortSuroundings(tile,isCitadel)
```
yairm210, line 559: How about we flip this if? String comparison is much cheaper than the evaluation

tuvus: Done!

yairm210 (approved):
Looks good, will enter next version :)

PR #9848, opened 2023-07-26
Update SDK Platforms text. Update photos

itanasi:


yairm210:
Thanks for this!
The link should remain 'docs'

itanasi:
If you're sure. The images don't show up on the github.io page though

yairm210:
You're right, I didn't notice that :/
So let's try your change, it can't be worse than what we already have
Sorry for the hassle 😔

itanasi:
No idea. I'm just basing this off of this guide/suggestion
https://tomcam.github.io/least-github-pages/adding-images-github-pages-site.html

yairm210:
I'll merge as is and we can test other configurations later

PR #9850, opened 2023-07-26
Simple Originating-Mod display in Civilopedia

SomeTroglodyte:
r

~~Forgot the i10n template. Any suggestions to wording before I add that?~~ Patched. Also patched for easter eggs - they don't have origins, intentionally.


yairm210:
Looks good! This was one of the things I thought of when introducing that field but then later I forgot about it completely 😅

yairm210 (approved):

PR #9856, opened 2023-07-29
Fix promotions being unavailable when it can be reached a slower way

SeventhM:
Previously if a promotion could require 2 promotions from branch A but 1 branch B and the branch B also lead to an earlier promotion from branch A, it erroneously consider that a loop. Solved by copying the logic from the ruleset Validator (going child up to parent, rather than parent to child and passing in the hashset rather than using a global one)

### DIFF
--- a/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
+++ b/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
@@ -89,9 +89,9 @@ internal class PromotionTree(val unit: MapUnit) {
 
         // Fill parent/child relations, ignoring prerequisites not in possiblePromotions
         for (node in nodes.values) {
+            if (detectLoop(node)) continue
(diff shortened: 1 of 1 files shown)


yairm210:
Can you write a test for this? This is exactly the kind of thing where thinking through it is harder than writing a test to ensure correctness :)

SeventhM:
Sure. While I'm at it, I'll add in a few other things I'm going to PR for anyways

SeventhM:
Actually, double checking how this works... I don't know if I can make a test for this that well. I can look into it further tomorrow, but this all uses an internal class that edits privately. I can remove it from being an internal class, but it will tank far more looking into before I add in the test for this specifically

yairm210:
@SeventhM Are you ready to merge this? Or do you want to keep working on it?

yairm210 (approved):

yairm210:
Oh I forgot to add this to the previous release :|
Welp

PR #9857, opened 2023-07-29
Fix free buildings from other buildings not showing up correctly

SeventhM:
Alt title: Rearrange order of actions when adding buildings


SeventhM:
Btw, the explanation: you need the uniques to actually be registered correctly before many of the `hasUnique` functions will see them. This is the case when adding free buildings and for updating resources. This seemingly does not seem to be the case for city connections, but city connections would need to be updated first anyways to avoid crashes when missing a capital

yairm210:
I'm pretty sure that 'update resources' should be called every time we update uniques, since it an always affect it
Sounds like a different PR though

PR #9860, opened 2023-07-30
Added tests to ensure that resources from buildings behave as expected

yairm210:
@SeventhM 
This contains
A. Tests for the use-cases of resources and building-granting that I understood from your PRs
B. A change that the unique update now triggers the resource update always, since new uniques could always mean more/less resources

If there are any other cases I missed we can add them in, but I want to get into the habit of writing tests to cover these edge cases so we know what behavior is expected in the first place :)


SeventhM:
I... Actually was going to pr this myself once I finished building a test for promotions. So there's a number of tests you're missing here that I would've covered
- Testing if getting resources from tiles works... At all (yes we don't have a test for this currently)
- Testing having an improvement that provides resources directly and an improvement that consumes resources 
- Tests for the similar unique `"Quantity of strategic resources produced by the empire +[relativeAmount]%"`

Also, I'm not sure why this is put in it's own file when similar tests are done in `globaluniquetests`

yairm210:
I found that file to be a little too full (600 lines) and this seemed like a smaller set of tests that could be separated from it

tests/src/com/unciv/uniques/ResourceTests.kt
```diff
@@ -0,0 +1,108 @@
...
+        city.cityConstructions.addBuilding(consumesCoal.name)
+        Assert.assertTrue(civInfo.getCivResourcesByName()["Coal"] == -1)
+    }
+
+    @Test
+    fun testResourceProductionAndConsumptionModifierDoesNotAffectConsumption() {
+        val consumesCoal = game.createBuilding("Consumes [1] [Coal]", "Provides [1] [Coal]",
+            "Double quantity of [Coal] produced")
+        city.cityConstructions.addBuilding(consumesCoal.name)
+        Assert.assertTrue(civInfo.getCivResourcesByName()["Coal"] == 1)
```
SeventhM, line 34: Can this be split into at least two buildings (one consumes, one provides, the double quantity can be on either building) and at least two separate tests (one for the provide building, and one for the consume building) so it's more obvious what's going on?

SeventhM:
Also now that I'm reading my implantation that I was working on, while I think this is a nitpick, I think maybe you should probably either use `civInfo.getResourceAmount(resource)` or `city.getResourceAmount(resource)` for consistency for how specific known resources are checked elsewhere in the code instead of `getCivResourcesByName()[resource]`. Minimum, using the city check could actually add in tests for city resources as well

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -311,6 +311,9 @@ open class Tile : IsPartOfGameInfoSerialization {
     fun changeImprovement(improvementStr: String?) {
         improvementIsPillaged = false
         improvement = improvementStr
+
+        if (resource != null && owningCity != null)
+            owningCity!!.civ.cache.updateCivResources()
```
SeventhM, line 316: If the responsibility on updating resources is here and not whatever is changing the improvement, then a reminder that you need to update resources if the improvement has provides/consumes a resource

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -931,6 +934,8 @@ open class Tile : IsPartOfGameInfoSerialization {
         }
 
         owningCity?.reassignPopulationDeferred()
+        if (resource != null && owningCity != null)
+            owningCity!!.civ.cache.updateCivResources()
```
SeventhM, line 938: Same story here. If the resource is within a city, you should just update resources in case the improvement consumes/provides resources

core/src/com/unciv/logic/map/tile/Tile.kt
yairm210: It's actually worse than that... because the *previous* improvement could have been providing a resource and now we replaced it...
I think I'll just update resources anyway instead of if/elsing all these conditions, we can always improve runtime later if we see it's mature optimization :)

PR #9864, opened 2023-07-31
Update German.properties

Knirps1:
Fixed a few mistakes and added more translations


Knirps1:
Thanks for the input, the "tritts du..." is explained in the comment lines above. It is the second part of a sentence.

PR #9866, opened 2023-07-31
Additional unit tests

SeventhM:
Added unit tests for the following:
- Moving resources along with a city (secretly a check for #9822)
- Promotions that can be reached from multiple directions (https://github.com/yairm210/Unciv/pull/9856#issuecomment-1656919939)

@yairm210 I still have the tests I was working on/finished for resources. You want me to just add #9860 to this PR, or leave that to you?

### DIFF
--- a/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
+++ b/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
@@ -10,9 +10,9 @@ import com.unciv.models.translations.tr
 class PromotionTree(val unit: MapUnit) {
     /** Ordered set of Promotions to show - by Json column/row and translated name */
     // Not using SortedSet - that uses needlessly complex implementations that remember the comparator
-    val possiblePromotions: LinkedHashSet<Promotion>
+    lateinit var possiblePromotions: LinkedHashSet<Promotion>
     /** Ordered map, key is the Promotion name, same order as [possiblePromotions] */
-    private val nodes: LinkedHashMap<String, PromotionNode>
+    private lateinit var nodes: LinkedHashMap<String, PromotionNode>
 
     class PromotionNode(
         val promotion: Promotion,
--- a/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
+++ b/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
@@ -7,7 +7,7 @@ import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.models.ruleset.unit.Promotion
 import com.unciv.models.translations.tr
 
-internal class PromotionTree(val unit: MapUnit) {
+class PromotionTree(val unit: MapUnit) {
     /** Ordered set of Promotions to show - by Json column/row and translated name */
     // Not using SortedSet - that uses needlessly complex implementations that remember the comparator
     val possiblePromotions: LinkedHashSet<Promotion>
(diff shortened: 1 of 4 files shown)


SeventhM:
Strange. I swear I set the promotions up the same as they are in the mod that breaks (deciv). It should've failed the promotion test

SeventhM:
Oh great, this doesn't trip because it breaks elsewhere in a way that deciv doesn't... ugh, so apparently how it breaks depends on the order of the promotions being checked. Which means we need to double check that promotions actually have the correct parents/children, not simply check if it's allowed to be purchased. Ugh

SeventhM:
Is it possible to convert this to a draft for now? The act of even creating a generic test for promotions is confusing me more than expected. Currently unless I figure out what's going wrong here I'm... not even sure we can create such a test currently

Here's my current notes:
1. Unrelated, but PromotionTree needs to be moved out of UI and needs a function to update itself so I'm not forced to redefine it each time I need to read it. That would minimum allow me to switch my variables to values
2. Whether or not thinks breaks before #9856 is seemingly down to the order of the prerequisite promotions. This means to make a test for it correctly would be to make identical tests with the promotions in different orders. This could be done, but I wonder if there's a way to write a test that isn't testing the exact same thing twice, just with stuff reordered

yairm210:
@SomeTroglodyte I know you're not fond of them, but the UI has only improved since they took over. I remember what it used to be like and boy howdy it's not even close

SeventhM:
Ok, did some more editing, and I think I got the test more in a state that I like. Instead of testing its original test case (a promotion set up with two distances to the same promotion), I've changed the scope of the test to just generically testing if PromotionTree is behaving as expected. Now I'm just left with 2 issues:
1. I don't like that I'm digging into PromtionTree the way that I am. I'm this close to just moving PromotionTree now to logic.map.mapunit and forcing you to access it via UnitPromtions. Less digging into the class and a more intuitive system in exchange for maybe some performance (it would need to be reset every time the unit either gets a promotion or anytime the transients are set rather than just with the UI/AI accesses them)
2. The Test as is is now a bit long

SeventhM:
New notes from digging into PromotionTree
1. Unfortunately, I think it doesn't make sense to tie the tree to the mapUnit. It can't be updated correctly whenever you set up transients due to its reliance on the unit information (which is unavailable because the unit is not initialized). And even if you could, it doesn't update often enough to matter. I.e., If you get a new tech via a Great Scientist effect, you do not update the necessary unit transients, which matters if a promotion is only available after getting a tech. In effect, despite me testing moving PromotionTree into the unit's information, I usually effectively rebuilt the PromotionTree anyways upon accessing it
2. Interestingly, the AI does not use this code at all. It only upgrades one at a time. This has a few consequences
A. The AI's code is functionally different than the player code. This feels like something that should be avoided in case it comes up in the future
B. Currently the AI picks a promotion at random. If we were to make the AI consider things a bit more, That first point may become even more relevant given that the advantage of Promotion Tree is being able to follow where the promotions lead to much easier
C. That said, given the first point and given an assumption that we don't have the AI check PromotionTree to get the list of promotion it can upgrade to, PromotionTree should actually stay in the UI section, despite only being logic. Very Interesting
4. That said, PromotionTree is missing some quality of life features. I will probably move its init code to its own function, and have init just call that function. I will probably add a function to just get a node, reachable or not (in case it comes up in the future). Stuff like that. Even if it's somewhat functionally the same, I still think it's better if we update PromotionTree via an init function rather than via a reassignment

PR #9870, opened 2023-08-01
Allow free unit triggers for any location

SeventhM:
Intent: allowing free unit triggers from places such as the start of a game, from unit triggers (free unit when killing) ~~and from finishing an improvement~~

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -53,7 +53,7 @@ object UniqueTriggerActivation {
             UniqueType.OneTimeFreeUnit -> {
                 val unitName = unique.params[0]
                 val unit = ruleSet.units[unitName]
-                if (chosenCity == null
+                if ((chosenCity == null && tile == null)
                         || unit == null
                         || unit.hasUnique(UniqueType.FoundCity) && civInfo.isOneCityChallenger())
                     return false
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/civilization/managers/UnitManager.kt
```diff
@@ -76,7 +76,8 @@ class UnitManager(val civInfo:Civilization) {
                 if (!unique.hasTriggerConditional())
                     UniqueTriggerActivation.triggerUnitwideUnique(unique, unit, triggerNotificationText = triggerNotificationText)
             for (unique in civInfo.getTriggeredUniques(UniqueType.TriggerUponGainingUnit))
-                if (unit.matchesFilter(unique.params[0]))
+                if (unique.conditionals.any { it.isOfType(UniqueType.TriggerUponGainingUnit) && 
+                        unit.matchesFilter(unique.params[0]) })
```
yairm210, line 80: WELP that's my bad

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -63,7 +63,9 @@ object UniqueTriggerActivation {
                 if (limit!=null && limit <= civInfo.units.getCivUnits().count { it.name==unitName })
                     return false
 
-                val placedUnit = civInfo.units.addUnit(unitName, chosenCity) ?: return false
+                val placedUnit = if (city != null || tile == null)
```
yairm210, line 66: Sorry what? If city is null and tile is null then add next to chosen city? Something here is off

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -94,7 +96,8 @@ object UniqueTriggerActivation {
 
                 val tilesUnitsWerePlacedOn: MutableList<Vector2> = mutableListOf()
                 repeat(actualAmount) {
-                    val placedUnit = civInfo.units.addUnit(unitName, chosenCity)
+                    val placedUnit = if (city != null || tile == null) civInfo.units.addUnit(unitName, chosenCity)
```
yairm210, line 99: Same as above

yairm210 (approved):

PR #9872, opened 2023-08-01
Fixed AI not removing fallout

### LINKED ISSUE #9871: Workers don't remove Fallout
**Platform**
Windows
Android

**Version**
4.7.11

**Describe the bug**
Workers won't clear fallout or build any buildings on fallout except for roads.

**To Reproduce**
1. Create a tiny map with only grassland and add fallout everywhere with floodfill.
2. Start on future era with one AI and the player as spectator.
3. The AI will build roads but won't remove the fallout and build farms.

**Expected behavior**
AI should remove fallout.

tuvus:
Fixes  #9871

Workers now treat fallout just like repairing an improvement. There might also be an option to alter the chooseImprovement() function  on line 384 to clear the fallout as well. Although I think the bug is not coming from the WorkerAutomation file because I reverted the file to the one in 4.7.0 and commented out unrelated errors and it still didn't work.

Note: I deleted unit.currentMovement > 0 on line 141, this makes the worker show it's improvement once it arrives even if it doesn't have enough movement points to work on it this turn. This is so that the worker shows it's action even when the improvement only takes one turn to complete.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -143,9 +143,14 @@ class WorkerAutomation(
                     debug("WorkerAutomation: ${unit.label()} -> repairs $reachedTile")
                     UnitActions.getRepairAction(unit).invoke()
                     return
+                } else if (reachedTile.terrainFeatures.contains("Fallout")) {
+                    debug("WorkerAutomation: ${unit.label()} -> removes fallout $reachedTile")
+                    reachedTile.improvementInProgress = Constants.remove + "Fallout"
+                    reachedTile.turnsToImprovement = reachedTile.ruleset.tileImprovements["Remove Fallout"]!!.getTurnsToBuild(unit.civ, unit)
+                    return
                 }
                 if (reachedTile.improvementInProgress == null && reachedTile.isLand
-                        && tileCanBeImproved(unit, reachedTile)
+                    && tileCanBeImproved(unit, reachedTile)
                 ) {
                     debug("WorkerAutomation: ${unit.label()} -> start improving $reachedTile")
                     return reachedTile.startWorkingOnImprovement(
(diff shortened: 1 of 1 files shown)


AutumnPizazz:
I feel that such a modification can only fix one type of situation caused by nuclear radiation, but cannot solve the problem of the new mod that generates unproductive artificial terrain, or the problem of the mod that changes the name of the nuclear radiation. QAQ @tuvus

tuvus:
I agree, but I couldn't find what change caused the problem in chooseImprovement or tileCanBeImproved so I implemented a different change instead.

AutumnPizazz:
Your answer is too advanced for me. I haven't studied Java systematically, so my contributions are limited to the ports of the given modules. (இωஇ )
Leave the checking and fixing to other programmers. ค(TㅅT)ค

yairm210:
We are *definitely* not going to reference improvements by specific name :D 
Instead of "fallout" we should be checking the uniques if improvement.hasUnique("Nullifies all other stats this tile provides") or similar

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -138,14 +138,20 @@ class WorkerAutomation(
 
             // If there's move still left, perform action
             // Unit may stop due to Enemy Unit within walking range during doAction() call
-            if (unit.currentMovement > 0 && reachedTile == tileToWork) {
```
yairm210, line 141: Why did you remove this? This is a necessary line, workers should not start removing things if they have 0 movement left this turn

yairm210:
And in general, fallout removal should be done via the "tileToImprove", "tileCanBeImproved" system, where the improvement is "Remove Fallout" - this works fine with other removal options, and I'm pretty sure I've seen it happen with fallout
If not, that's what we should be debugging - chooseImprovement:436 is what should say "This tile should get a remove fallout improvement"

Issue #9475, opened 2023-05-29
A unrecoverable error message.


FtFoxyYT:
Hi so uh I have spent minutes to get almost every single unciv mod. And then a error message has appeared and it says to remove some mods but honestly I have no idea how to do that because the error message is blocking the way. If this is fixable please tell me I just wanna play some unciv.

FtFoxyYT:
**Platform:** Android
**Version:** 4.6.15 (Build 872)
[...]

PR #9881, opened 2023-08-03
Ruleset validator: Tilesets

SomeTroglodyte:
As threatened in #9475 - proactive defense to ignore Tileset config jsons that throw, plus display of a variety of possible Tileset mod mistakes in "Locate mod errors".

The mod is fine! I broke it just for this screenshot! Really!
image

Remember we had to defend against this, it's still faulty:
image

And having that Civ6 Tileset mod as broken as in the first screenie, but running master - no action, just start it:
image

</details>

This has one lazy shortcut - the unit test "baseRulesetHasNoBugs" might fail with this code, if or when we should reorg the atlases so Tileset textures are in packs not named "Tilesets". Won't affect mod checking.


yairm210 (approved):

yairm210:
Looks okay to me, I guess we'll find out ☺️

PR #9888, opened 2023-08-05
feat: added auto next turn

### LINKED ISSUE #2375: Hope to have an automatic next round function
It is hoped that an automatic next round function will appear. You can select the switch in the setting (it is off by default). When the next round is automatic, click the button of the next round to cancel the automatic next round.:-)

### LINKED ISSUE #9450: Feature request: Behold... Auto next turn !! Automatically press `Next Turn` for `X` turns !!!
Sometimes in Multiplayer, there is nothing to do except pressing next turns. Auto next turn might be a good option in these scenarios to improve the speed of gameplay.
The idea is auto next turns would be a tick mark below the next turn option that will automatically press next turn when the game updates. If there is nothing to do except next turn and automate.

alexban011:
Resolves #9450
Resolves #2375

Currently it requires `check for idle units` to be disabled. Should it automatically ignore those ? Can you think of other instances in which the game shouldn't go to the next turn besides new tech and new policy? 

Plus it should actually check if you have a city placed instead of `turns != 0` and it should also have a limit, as mentioned in #2375 but I'll do that after I get a response for the questions above :)

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -763,6 +763,7 @@ When disabled, saves battery life but certain animations will be suspended =
 Gameplay = 
 Check for idle units = 
 Auto Unit Cycle = 
+Auto next turn = 
 Move units with a single tap = 
 Auto-assign city production = 
 Auto-build roads = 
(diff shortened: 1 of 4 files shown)


AdityaMH:
I'm disagree this except can block next turn if have event new tech, pick policy event, enemy nearby, finishing my unit moving order, and a tracking (an animation-like) where my unit and and enemy per turn (since invisible barbarians stealing workers and make workers gone nowhere.).

alexban011:
I don't currently have time, so I'll close it and reopen it when I do

PR #9889, opened 2023-08-05
Architecture: NextTurnAction as enum

SomeTroglodyte:
Done as short puzzle inspired by how #9888 tests whether "Next Turn" actually does "Next Turn"...

Tested and it works, but no net improvement. 9888 could do a nextTurnAction == NextTurnAction.NextTurn with this, much cleaner. One little change - FoundPantheon and EnhancePantheon now look for separate icons of these names, not the same one. Could easily be forced back with another override.

Anecdote: Tried a UI/Logic agnostic architecture approach first, an interface, the enum carrying only static properties but implementing the interface, and an "instance class" that could carry dynamic stuff like actions with closures or context-dependent labels... With an elegant factory mechanism on the enum instances for the latter... Worked too, but somehow this idea looked cleaner. Sorry, forgot to commit that interim version.

### DIFF
--- a/core/src/com/unciv/ui/components/tilegroups/CityButton.kt
+++ b/core/src/com/unciv/ui/components/tilegroups/CityButton.kt
@@ -12,19 +12,19 @@ import com.badlogic.gdx.utils.Align
 import com.unciv.GUI
 import com.unciv.logic.battle.CityCombatant
 import com.unciv.logic.city.City
-import com.unciv.models.ruleset.INonPerpetualConstruction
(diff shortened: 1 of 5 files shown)


yairm210:
Separating the action (a logic component) from the button (a UI component) is definitely the way to go! :)

yairm210:
My worry is that the enum still references the worldscreen, instead of referencing the gameinfo, which feels off to me

yairm210 (approved):

PR #9898, opened 2023-08-08
Ruleset object reorganization

SomeTroglodyte:
Read commit titles!

#### Background

In a not to lucid moment I wondered why there's mod-wide Uniques both in ModOptions and GlobalUniques - now I know, but on the way some more headscratching led to this.

Side effects: None as far as I can see - except both ModOptions and GlobalUniques are now identifiable as Unique source objects. ModOptions previously had null there, GlobalUniques an empty string.

A little surprising that this works - both the "ModOptions crashes and burns" and "I bet there's a way of initializing these without having to override it everywhere..." comments went overboard.

For "initializing without override" - no it's still an override but a much shorter one, thus much less challenging to actually inline the initialization only once somehow... And implementation duplication is gone.
As for "burns" - nope, I tried, couldn't repro. Changed the Aliens mod just for that test, downloads and starts. I am convinced the problems were simply due to the old approach forgetting to mark the fields transient. Or maybe better compiler intelligence from kotlin version.

#### To-do:
There's still a bunch of "impure" files in the Ruleset package (those Studio displays the ".kt" ending for). Some can likely be made "pure" by refactoring minor helper classes to be nested ones. Containers of top-level methods only might stay that way. Some might be cleanest to split into tiny <10 line files - after all, we already have INamed like that. I simply stopped midway, because some refactors will touch tons of imports, and right now in this PR the changes (by commit) are still readable.

#### More to-do:
No reason to treat ModOptions Uniques specially. Ergo, remove ModOptionsConstants, make them UniqueTypes instead. Then - Validate modOptions! Then - use that uniques list for declarative mod compatibility. Such as...
```kotlin
    ///////////////////////////////////////////// region 95 Mod META /////////////////////////////////////////////
    ModDisablesReligion("Disable religion", UniqueTarget.ModOptions),
    ...
    ModCompatibleWith("Mod is compatible with [comment]", UniqueTarget.ModOptions),
    ModIncompatibleWith("Mod is not compatible with [comment]", UniqueTarget.ModOptions),
    ModIsAudioVisual("Mod is suitable as permanent audiovisual mod", UniqueTarget.ModOptions),
    ModIsOnlyAudioVisual("Mod is only suitable as permanent audiovisual mod", UniqueTarget.ModOptions),

    // endregion
```

### DIFF
--- a/core/src/com/unciv/models/ruleset/GlobalUniques.kt
+++ b/core/src/com/unciv/models/ruleset/GlobalUniques.kt
@@ -5,6 +5,8 @@ import com.unciv.models.ruleset.unique.UniqueTarget
 import com.unciv.models.ruleset.unique.UniqueType
 
 class GlobalUniques: RulesetObject() {
+    override var name = "GlobalUniques"
+
     override fun getUniqueTarget() = UniqueTarget.Global
     override fun makeLink() = "" // No own category on Civilopedia screen
 
--- a/core/src/com/unciv/models/ruleset/ModOptions.kt
+++ b/core/src/com/unciv/models/ruleset/ModOptions.kt
@@ -1,0 +1,46 @@
+package com.unciv.models.ruleset
+
+import com.unciv.models.ModConstants
+import com.unciv.models.ruleset.unique.IHasUniques
+import com.unciv.models.ruleset.unique.Unique
+import com.unciv.models.ruleset.unique.UniqueMap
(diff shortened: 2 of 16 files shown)


yairm210:
Had you not split it up into 5 different commits this woods have been impossible for me to parse
As it is though, looks good

yairm210:
Absolutely agree that modOptions uniques should be typed

PR #9900, opened 2023-08-10
Add Defensive pact

tuvus:
Adds on to #4674

Defense pacts are now available as trade offer to civilizations with a Declaration of Friendship already signed. Moved Research Pacts out of the diplomacy screen and into the trade screen like how it is in Civ 5.  There is also a new Civilopedia entry on Defensive Pacts.

When a defensive pact is signed each member joins the other member's current and future defensive wars. When a Civ declares a war, they then end all Defensive Pacts that they have.

AI's will only consider Defensive Pacts when they are Allies with the other Civ (>= 80). Other than that I have not added any kind of logic for them to evaluate weather a Defensive Pact is helpful for them or not. This should be added in the future but I feel like we need to refactor some of the hasMotivation strength comparing evaluations from NextTurnAutomation into Diplomatic functions or some other class so we can reuse the code for it.

I mostly copied the Declaration of Friendship style for consistency, but there might be some bugs with the relationship modifiers.
The current relationship modifiers are just fillers and will probably be changed.


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -801,6 +804,29 @@ object NextTurnAutomation {
...
+            .filter {
+                civInfo.diplomacyFunctions.canSignDefensivePactWith(it)
+                    && !civInfo.getDiplomacyManager(it).hasFlag(DiplomacyFlags.DeclinedDefensivePact)
+                    && civInfo.getDiplomacyManager(it).relationshipIgnoreAfraid() == RelationshipLevel.Ally
+            }
+            .sortedByDescending { it.stats.statsForNextTurn.science }
+
+        for (otherCiv in canSignDefensivePactCiv) {
+            // Default setting is 1, this will be changed according to different civ.
+            if ((1..10).random() > 1) continue
```
yairm210, line 820: So 10% chance? Random.nextInt(10) is the way to do this, unless range has its own random() then we're building a list here for no reason

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyFunctions.kt
```diff
@@ -113,6 +112,22 @@ class DiplomacyFunctions(val civInfo: Civilization){
                 ).toInt()
     }
 
+    fun canSignDefensivePact(): Boolean {
+        if (!civInfo.isMajorCiv()) return false
+        if (!civInfo.hasUnique(UniqueType.EnablesDefensivePacts)) return false
+        if (civInfo.gameInfo.ruleset.technologies.values
```
yairm210, line 118: This is a remnant from research agreement that is not relevant

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -714,12 +731,26 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
                 civInfo.addNotification("[${offer.name}] from [$otherCivName] has ended",
                     NotificationCategory.Trade, otherCivName, NotificationIcon.Trade)
+
+        // If we attacked, then we need to end all of our defensive pacts acording to Civ 5
+        if (isOffensiveWar) {
+            civInfo.diplomacy.values.filter { it.otherCiv() != otherCiv() // So that we don't cancel the defensive pact before negative modifiers are added
+                && it.diplomaticStatus == DiplomaticStatus.DefensivePact }.forEach {
+                // Trades with defensive pact are now invalid
+                val defensivePactOffer = it.trades.firstOrNull { trade -> trade.ourOffers.any { offer -> offer.name == Constants.defensivePact} }
+                it.trades.remove(defensivePactOffer)
```
yairm210, line 747: Needs to be cancelled for both sides. See example of removeUntenebleTrades.
Also we need notifications for both sides

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -929,6 +982,56 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+        otherCivDiplomacy().setModifier(DiplomaticModifiers.DefensivePact, 30f)
+        diplomaticStatus = DiplomaticStatus.DefensivePact;
+        otherCivDiplomacy().diplomaticStatus = DiplomaticStatus.DefensivePact;
+
+        for (thirdCiv in getCommonKnownCivs().filter { it.isMajorCiv() }) {
+            thirdCiv.addNotification("[${civInfo.civName}] and [$otherCivName] have signed the Defensive Pact!",
+                NotificationCategory.Diplomacy, civInfo.civName, NotificationIcon.Diplomacy, otherCivName)
+            thirdCiv.getDiplomacyManager(civInfo).setDefensivePactBasedModifier()
+        }
+        civInfo.isAtWar()
```
yairm210, line 997: ?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -929,6 +982,56 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+            UniqueTriggerActivation.triggerCivwideUnique(unique, otherCiv())
+    }
+
+    private fun setDefensivePactBasedModifier() {
+        removeModifier(DiplomaticModifiers.SignedDefensivePactWithOurAllies)
+        removeModifier(DiplomaticModifiers.SignedDefensivePactWithOurEnemies)
+        for (thirdCiv in getCommonKnownCivs()
+            .filter { it.getDiplomacyManager(civInfo).hasFlag(DiplomacyFlags.DefensivePact) }) {
+            //Note: These modifiers are additive to the freindship modifiers
+            when (otherCiv().getDiplomacyManager(thirdCiv).relationshipIgnoreAfraid()) {
```
yairm210, line 1024: When should be for the amount, since the action is the same for all skins except 0
Would result in cleaner code

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -929,6 +982,56 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+            thirdCiv.addNotification("[${civInfo.civName}] and [$otherCivName] have signed the Defensive Pact!",
+                NotificationCategory.Diplomacy, civInfo.civName, NotificationIcon.Diplomacy, otherCivName)
+            thirdCiv.getDiplomacyManager(civInfo).setDefensivePactBasedModifier()
+        }
+        civInfo.isAtWar()
+        //Have them join our wars
+        for (newWar in civInfo.diplomacy.values.filter { it.diplomaticStatus == DiplomaticStatus.DefensiveWar && !it.otherCiv().isDefeated()
+            && !otherCivDiplomacy().civInfo.isAtWarWith(it.otherCiv())
+        }) {
+                otherCiv().diplomacy.values.find { it.otherCiv() == newWar.otherCiv() }?.declareWar(true)
```
yairm210, line 1002: GetDiplomacyManager

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -929,6 +982,56 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+        diplomaticStatus = DiplomaticStatus.DefensivePact;
+        otherCivDiplomacy().diplomaticStatus = DiplomaticStatus.DefensivePact;
+
+        for (thirdCiv in getCommonKnownCivs().filter { it.isMajorCiv() }) {
+            thirdCiv.addNotification("[${civInfo.civName}] and [$otherCivName] have signed the Defensive Pact!",
+                NotificationCategory.Diplomacy, civInfo.civName, NotificationIcon.Diplomacy, otherCivName)
+            thirdCiv.getDiplomacyManager(civInfo).setDefensivePactBasedModifier()
+        }
+        civInfo.isAtWar()
+        //Have them join our wars
```
yairm210, line 998: Are you SURE that signing a defensive pact immediately causes way?

core/src/com/unciv/logic/civilization/diplomacy/DiplomaticStatus.kt
```diff
@@ -5,5 +5,7 @@ import com.unciv.logic.IsPartOfGameInfoSerialization
 enum class DiplomaticStatus : IsPartOfGameInfoSerialization {
     Peace,
     Protector,  //city state's diplomacy for major civ can be marked as Protector, not vice versa.
-    War
+    DefensivePact,
+    OffensiveWar,
```
yairm210, line 9: We have to retain War, since older games use it and this is a breaking change

yairm210:
In order to not break multiplayer games where only one side has updated, we need to do the following:
- Add one small PR that address OffensiveWar, DefensiveWar, and DefensivePact, with the isAtWar function changes. Also include defensive pact enum values for diplomatic UI
- Version N+1 will start accepting that status as valid, even though it's not in use
- Only then can we start replacing War with offensive/ defensive
- That way, if version N+2 says "this is a X" then version N+1 - which doesn't know what that means mechanically - will still accept that as a valid value and won't break the parsing

yairm210:
And more than anything:
GOOD JOB!
This is a BIG change that people have been asking for!

But we need to be careful not to ruin people's experience when we try to improve it ;)

I may even suggest splitting this into 3 versions for safety, for now an initial one for the enum values is an absolute must

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
yairm210: Ok cool, then not a problem 👍

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -240,10 +246,10 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
 
     /** Same as [relationshipLevel] but omits the distinction Neutral/Afraid, which can be _much_ cheaper */
     fun relationshipIgnoreAfraid(): RelationshipLevel {
-        if (civInfo.isHuman() && otherCiv().isHuman())
+        if (false && civInfo.isHuman() && otherCiv().isHuman())
```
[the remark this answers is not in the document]
tuvus: Oops, I added some extra conditions to make debugging easier and just forgot to take them out again.

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
tuvus: The line above it in offerDeclarationOfFriendship() at 780 has: if ((1..10).random() <= 5), so I just copied that. Should I replace both with Random.nextInt(10)?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
tuvus: I think I misinterpreted an article saying "you will immediately be pulled into war" as joining previous defensive wars. I will revert this change.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
tuvus: setFriendshipBasedModifier() uses this way of deciding the modifier. The problem is that there is a difference between DiplomaticModifiers.SignedDefensivePactWithOurEnemies and DiplomaticModifiers.SignedDefensivePactWithOurAllies. The revised code would require two when statements unless there is another way of doing it. 

`
val relationshipLevel =  otherCiv().getDiplomacyManager(thirdCiv).relationshipIgnoreAfraid()
            addModifier(when (relationshipLevel) {
                RelationshipLevel.Unforgivable, RelationshipLevel.Enemy -> DiplomaticModifiers.SignedDefensivePactWithOurEnemies
                RelationshipLevel.Friend, RelationshipLevel.Ally -> DiplomaticModifiers.SignedDefensivePactWithOurAllies
                else -> DiplomaticModifiers.SignedDefensivePactWithOurAllies }, when (relationshipLevel) {
                RelationshipLevel.Unforgivable -> -30f
                RelationshipLevel.Enemy -> -15f
                RelationshipLevel.Friend ->  5f
                RelationshipLevel.Ally -> 15f
                else -> {0f}
            })
`

If you want me to implement this change, I would probably change setFriendshipBasedModifier() as well.

core/src/com/unciv/logic/civilization/diplomacy/DiplomaticStatus.kt
tuvus: Should we replace all current Wars with DefensiveWars in some function once the game is loaded or add it to IsAtWar()?
If we really want to, we can remove OffensiveWar and DefensiveWar since signing a defensive pact no longer affects wars started in the past.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyFunctions.kt
yairm210: No, it's relevant to the research agreement, since that gives you techs that you can research
It's not relevant to defensive pact because that nothing to do with techs

And yes, that's the check :)

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
yairm210: I agree, having the 'when' there set a variable would my much cleaner

core/src/com/unciv/logic/civilization/diplomacy/DiplomaticStatus.kt
yairm210: Is it doesn't affect it, then yes, let's revert this. It will save us a huge headache

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
tuvus: I will leave it out of this pull request because it is consistent with offerResearchAgreement() and offerDeclarationOfFriendship. We can change those in a different pull request if we want.

tuvus:
Okay, I think I have resolved all the changes that need to be made. One last thing that I want to ask: Can we confirm the value of the modifiers, I have 10 for signing a defensive pact automatically and around 5 for signing with their ally, up to -15 for signing with their enemy. Breaking a defensive pact gives a -20 to the civ and -10 to all other Civs. This is in addition to the -40 and -20 from breaking the declaration of friendship.

Finally, should signing a Defensive Pact extend the duration of the declaration of friendship to the duration of the defensive pact? Because the declaration of friendship can only be alone it the duration can be modified without affecting any other trade.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -591,6 +597,9 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
                         if (!otherCivDiplomacy().hasFlag(DiplomacyFlags.ResearchAgreement))
                             scienceFromResearchAgreement()
                     }
+                    DiplomacyFlags.DefensivePact.name -> {
```
yairm210, line 600: What is this for?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -714,12 +729,34 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+                it.otherCivDiplomacy().trades.remove(theirDefensivePactOffer)
+                it.removeFlag(DiplomacyFlags.DefensivePact)
+                it.otherCivDiplomacy().removeFlag(DiplomacyFlags.DefensivePact)
+                it.diplomaticStatus = DiplomaticStatus.Peace
+                it.otherCivDiplomacy().diplomaticStatus = DiplomaticStatus.Peace
+                for (civ in getCommonKnownCivs().filter { civ -> civ.isMajorCiv() }) {
+                    civ.addNotification("[${civInfo.civName}] canceled their Defensive Pact with [${it.otherCivName}]!",
+                        NotificationCategory.Diplomacy, civInfo.civName, NotificationIcon.Diplomacy, it.otherCivName)
+                }
+                civInfo.addNotification("We have canceled our Defensive Pact with [${it.otherCivName}]!",
```
yairm210, line 756: What about the other side? Looks like they don't get a notification, only for us and 3rd parties

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -737,8 +774,16 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
         removeFlag(DiplomacyFlags.BorderConflict)
 
 
-        // Go through city state allies.
         if (!civInfo.isCityState()) {
+            // Go through defensive pact allies.
+            if (!isOffensiveWar) {
+                for (defensivePact in civInfo.diplomacy.values.filter { it.diplomaticStatus == DiplomaticStatus.DefensivePact && !it.otherCiv().isDefeated()
+                    && !otherCivDiplomacy().civInfo.isAtWarWith(it.otherCiv())}) {
```
yairm210, line 781: This check doesn't make sense to me on first trading, it takes a while to understand what's going on, which to me indicates that readability is an issue. We need to call the parties by meaningful names

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -737,8 +774,16 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+                    otherCiv().diplomacy.values.find { it.otherCiv() == defensivePact.otherCiv()}?.declareWar(true)
```
yairm210, line 782: GetDiplomacyManager

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -919,16 +978,65 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
                 .filter { it.getDiplomacyManager(civInfo).hasFlag(DiplomacyFlags.DeclarationOfFriendship) }) {
-            when (otherCiv().getDiplomacyManager(thirdCiv).relationshipIgnoreAfraid()) {
-                RelationshipLevel.Unforgivable -> addModifier(DiplomaticModifiers.DeclaredFriendshipWithOurEnemies, -15f)
-                RelationshipLevel.Enemy -> addModifier(DiplomaticModifiers.DeclaredFriendshipWithOurEnemies, -5f)
-                RelationshipLevel.Friend -> addModifier(DiplomaticModifiers.DeclaredFriendshipWithOurAllies, 5f)
-                RelationshipLevel.Ally -> addModifier(DiplomaticModifiers.DeclaredFriendshipWithOurAllies, 15f)
-                else -> {}
-            }
+            val relationshipLevel =  otherCiv().getDiplomacyManager(thirdCiv).relationshipIgnoreAfraid()
+            addModifier(when (relationshipLevel) {
```
yairm210, line 982: Split this into 2 'when's that set 2 variables, with a final call toaddModifier to use those 2 variables.
This is not readable

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -919,16 +978,65 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+        // Ignore contitionals as triggerCivwideUnique will check again, and that would break
+        // UniqueType.ConditionalChance - 25% declared chance would work as 6% actual chance
+        for (unique in civInfo.getTriggeredUniques(UniqueType.TriggerUponSigningDefensivePact, StateForConditionals.IgnoreConditionals))
+            UniqueTriggerActivation.triggerCivwideUnique(unique, civInfo)
+        for (unique in otherCiv().getTriggeredUniques(UniqueType.TriggerUponSigningDefensivePact, StateForConditionals.IgnoreConditionals))
+            UniqueTriggerActivation.triggerCivwideUnique(unique, otherCiv())
+    }
+
+    private fun setDefensivePactBasedModifier() {
+        removeModifier(DiplomaticModifiers.SignedDefensivePactWithOurAllies)
```
yairm210, line 1020: Why are we removing these modifiers? If I had a defensive pact with your enemy, then with your sky, why should only the last one count?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -919,16 +978,65 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+        removeModifier(DiplomaticModifiers.SignedDefensivePactWithOurEnemies)
+        for (thirdCiv in getCommonKnownCivs()
+            .filter { it.getDiplomacyManager(civInfo).hasFlag(DiplomacyFlags.DefensivePact) }) {
+            //Note: These modifiers are additive to the friendship modifiers
+            val relationshipLevel =  otherCiv().getDiplomacyManager(thirdCiv).relationshipIgnoreAfraid()
+            addModifier(when (relationshipLevel) {
```
yairm210, line 1026: Here too, add variables for the 'when' results

core/src/com/unciv/logic/trade/Trade.kt
```diff
@@ -60,6 +60,9 @@ class TradeRequest : IsPartOfGameInfoSerialization {
             diplomacyManager.setFlag(DiplomacyFlags.DeclinedLuxExchange,20)
         if (trade.ourOffers.any { it.name == Constants.researchAgreement })
             diplomacyManager.setFlag(DiplomacyFlags.DeclinedResearchAgreement,20)
+        if (trade.ourOffers.any { it.name == Constants.defensivePact })
+            diplomacyManager.setFlag(DiplomacyFlags.DeclinedDefensivePact,5)
```
yairm210, line 64: I think 10 turns between defensive pact offers makes sense

core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
```diff
@@ -791,8 +764,8 @@ class DiplomacyScreen(
         val tradeButton = "Trade".toTextButton()
         tradeButton.onClick {
             setTrade(otherCiv).apply {
-                tradeLogic.ourAvailableOffers.apply { remove(firstOrNull { it.type == TradeType.Treaty }) }
-                tradeLogic.theirAvailableOffers.apply { remove(firstOrNull { it.type == TradeType.Treaty }) }
+//                 tradeLogic.ourAvailableOffers.apply { remove(firstOrNull { it.type == TradeType.Treaty }) }
```
yairm210, line 767: What's this about?

core/src/com/unciv/ui/screens/diplomacyscreen/OffersListScroll.kt
```diff
@@ -103,11 +103,13 @@ class OffersListScroll(
                 }
 
                 val amountPerClick =
-                        if (offer.type == Gold) 50
-                        else 1
+                    when (offer.type) {
+                        Gold -> 50
+                        Treaty -> Int.MAX_VALUE
```
yairm210, line 108: Why...?

tests/src/com/unciv/logic/map/UnitMovementTests.kt
```diff
@@ -199,7 +199,7 @@ class UnitMovementTests {
 
         city.hasJustBeenConquered = true
         civInfo.diplomacy[otherCiv.civName] = DiplomacyManager(otherCiv, otherCiv.civName)
-        civInfo.getDiplomacyManager(otherCiv).diplomaticStatus = DiplomaticStatus.War
+        civInfo.getDiplomacyManager(otherCiv).diplomaticStatus == DiplomaticStatus.War
```
yairm210, line 202: Sounds like a mistake

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
tuvus: The trade creates the flag, which counts down to zero and then cancels the defensive pact by setting the DiplomaticStatus back to peace. Previously the diplomaticStatus was set to DefensivePact.

yairm210: Gotcha

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
tuvus: I experimented with it, but getCommonKnownCivs() actually returns the other civ as well. It doesn't, however return the Civ that this DiplomacyManager belongs to. Logically it doesn't make sense for GetCommonKnownCivs() to return the other Civ as well and not the current civ but that is how it functions aparently.

yairm210: That's very strange, I'm pretty sure there's a test for that
In any case that's something we need to fix

yairm210: No, I'm certain there are tests for that
I'll recheck this to make sure

tuvus: I figured it out. I saw the notification from the trade. There was no notification from it. I'll change it to not send a trade-end notification.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
tuvus: I kept the defensive pact behavior consistent with the declaration of friendship behavior. This is the same logic as in setFriendshipBasedModifier().

core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
tuvus: These two lines of code are very old, and I think they were meant to delete the research agreement button on the trade screen since it was offered through the diplomacy screen. Now that there is no longer a research button on the diplomacy screen, and we want the research button to show up, it is no longer needed.

core/src/com/unciv/ui/screens/diplomacyscreen/OffersListScroll.kt
tuvus: The research agreement has a gold value according to its cost. In order to move the entire research agreement from available offers to offers we need to have this at Int.Max_VALUE.

tests/src/com/unciv/logic/map/UnitMovementTests.kt
tuvus: Good catch! This was from when I reverted isAtWar() back.

yairm210:
I think this is good to go :)
Please note that there are bound to be mistakes that we did not catch, so you'll need to be ready to catch them
More playtesting, especially with hotseat multiplayer where you can be 2 sides *and* agressor, will help to mitigate some of that

yairm210 (approved):

tuvus:
To simplify the defensive pact and declaration of friendship, should declaring a defensive pact extend the declaration of friendship to the end of the defensive pact? It doesn't make much sense how they can have a defensive pact without a declaration of friendship.

yairm210:
I personally don't care either way, so you know how it was in Civ V?

tuvus:
I don't know about the DLC, but in the base game, a defensive pact does not require a declaration of friendship. I thought that a defensive pact did require a declaration of friendship because I read it was much harder to get, of course, this is without DLC.

yairm210:
So let's not make it continue

yairm210:
One problem I can forsee is that if some players update multiplayer and some don't, then the ones that do will create defensive pacts, but when the turn shifts to the other player - their code doesn't know that logic yet so it won't activate.

To mitigate that, I'm going to put an artificial restriction on the defensive pact creation points (automation and UI), so we get one or two versions where "if there was a pact I would know how to handle it, but no pacts can be created". 
We can then remove that restriction in, say, 2 versions, so in the meantime people have had time to update and everyone gets defensive pacts that work for all players :)

tuvus:
That is a good idea. I have more modifications to the defensive pact, including tests. Should I submit another pull request when they are ready?

yairm210:
Yes :)

tuvus:
I read here that defensive pacts require a declaration of friendship in the Brave New World expansion. Should I try to make it more complicated and make it so that it doesn't require a DOF in the base game, or should we just use the more refined version?

https://forums.civfanatics.com/resources/diplomacy-interacting-with-civs-and-positive-negative-diplomacy-modifiers-bnw.25611/

yairm210:
I think requiring DOF makes sense in general, I don't think anyone will care that Vanilla is not exactly Vanilla since it makes less sense

Also - released Defensive Pact ability in 4.7.17 :D

PR #9906, opened 2023-08-12
Add "Improvement" as an improvement filter

SeventhM:
Note: It's set to non road improvements for now thanks to roads being hardcoded and the current expected usecase of current roads. Should probably change once roads are genericized

### DIFF
--- a/core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
+++ b/core/src/com/unciv/models/ruleset/tile/TileImprovement.kt
@@ -26,7 +26,7 @@ class TileImprovement : RulesetStatsObject() {
     override fun getUniqueTarget() = UniqueTarget.Improvement
     val shortcutKey: Char? = null
     // This is the base cost. A cost of 0 means created instead of buildable.
-    val turnsToBuild: Int = -1
+    var turnsToBuild: Int = -1
 
 
     fun getTurnsToBuild(civInfo: Civilization, unit: MapUnit): Int {
@@ -91,6 +91,7 @@ class TileImprovement : RulesetStatsObject() {
             "All" -> true
             "All Road" -> isRoad()
             "Great Improvement", "Great" -> isGreatImprovement()
+            "Improvement" -> !isRoad()
             in uniqueMap -> true
             else -> false
         }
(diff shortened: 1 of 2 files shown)


yairm210:
Usually this is in the form of "X improvement(s)" so "improvement improvement" white just look weird. Why not call it "non-road" improvement? Much clearer

SeventhM:
tileFilter also checks improvement. Usually that is in the form of "X tiles". While "improvement improvement" looks weird, "non road tiles" is actively misleading. There's already precedence in weirdly repeating the word "improvement" with "Great Improvement". At best, "Non-Road Improvement" is best, at worst, it doesn't matter with the current expected use case

yairm210:
If that's the use case, then this should be added to the tileFilter, and not the improvementFilter

SeventhM:
Unfortunately, also no. Uniques like `StatsFromTiles`, aka `"[stats] from [tileFilter] tiles [cityFilter]"`, also exists and would need to check specifically improvementFilter, or else count the additional stats as part of the tile, rather than as part of the improvement. This is relevant for uniques such as `StatPercentFromObject`, aka `"[relativeAmount]% [stat] from every [tileFilter/buildingFilter]"` which actually cares about which object we increase the stats to, rather than the tile as a whole

yairm210:
Hmmmm
I think it makes things ugly but I understand why

Please
- Add a comment describing the use cases this is for
- add a note in TileFilter docs (yes, not improvementFilter, modders shouldn't need to know that we do ugly things behind the scenes) so modders know that this is a valid option

SeventhM:
I actually double checked the matches filter function, and actually, it's probably safe to just have this be all improvements, rather than just non road ones. Changing that now

yairm210:
Hmmm putting it directly near 'unimproved' makes me think, how about 'improved tiles' instead of 'Improvement tiles', as the counterpart to 'unimproved tiles'?

SeventhM:
Again, technically that's misleading. The point of this filter is to generically target any improvement without targeting the underlying tile (especially for stat increases). "improved" implies we're targeting the tile itself, not the improvement, which misses the point of the pr. That could be argued to be an additional filter to add, but not a replacement

yairm210 (approved):

PR #9913, opened 2023-08-15
Nuke blast simulation no longer shows invisible units.

### LINKED ISSUE #9912: Nuke blast simulation UI shows non visible units
**Describe the bug**
Units have target icons while simulating a nuke blast even though they are invisible.

**To Reproduce**
1. Start a future game with other Civs and uranium
2. Build a nuke
3. Click on another civ's land without vision and see the target icons pop up without being visible.

**Expected behavior**
Nuke blast UI does not show invisible units.

tuvus:
Fixes #9912

A target Icon is only added to units with a blast radius if the tile is visible and contains a visible unit or if the tile contains a city and has been explored.

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
@@ -695,7 +695,8 @@ class WorldMapHolder(
             val attackableTiles: List<AttackableTile> =
                 if (nukeBlastRadius >= 0)
                     selectedTile!!.getTilesInDistance(nukeBlastRadius)
-                        .filter { it.getFirstUnit() != null }
+                        // Should not display invisible submarine units even if the tile is visible.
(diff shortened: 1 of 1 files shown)


yairm210 (approved):

core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
```diff
@@ -695,7 +695,8 @@ class WorldMapHolder(
             val attackableTiles: List<AttackableTile> =
                 if (nukeBlastRadius >= 0)
                     selectedTile!!.getTilesInDistance(nukeBlastRadius)
-                        .filter { it.getFirstUnit() != null }
+                        // Should not display invisible submarine units even if the tile is visible.
+                        .filter { targetTile -> (targetTile.isVisible(unit.civ) && targetTile.getUnits().any { !it.isInvisible(unit.civ) }) || (targetTile.isCityCenter() && unit.civ.hasExplored(targetTile)) }
```
alexban011, line 699: personally I think this line would look way more readable if each condition was on it's own line

Issue #9319, opened 2023-05-03
Map/dev starting locations


hackedpassword:
Originally posted on discord. 

"As I work on another map, it's clear that setting starting locations (SL) is needed for random spawns." 

Suggestion for the map editor: 1) new entity for general spawn SLs for random spawns 2) civ general SL 3) CS general SL 4) No-spawn SL that occludes tiles from any random SL ; If the user selects to have more AI players than SLs available, overflow players default to random tile SL (current behavior) excluding no-spawn.

It was also suggested that SL classes, like by water, in desert, by mountain, etc are entities able to apply to tiles also. This could be useful for specific or semi random placement in specific areas, like say only on specific shorelines or near certain mountains.

Caballero-Arepa:
I support this.

hackedpassword:
@tuvus Great feature, been super useful! Doesn't update the starting position count, no biggie. Really appreciate the addition!

hackedpassword:
I guess what you're saying is the top count iterates per type, versus total entity count. K, expectations updated.

PR #9920, opened 2023-08-16
General Starting locations in map editor

tuvus:
Fixes part of #9319

This adds "Any CIv starting location" to the list of starting locations that can be placed in the map editor. It uses the Spectator nation and Icon as a placeholder. Nations that need to be placed without a starting location specific to their nation will use these "Any Civ" starting locations before performing a find tile operation to find a new one.


core/src/com/unciv/ui/screens/mapeditorscreen/tabs/MapEditorEditSubTabs.kt
```diff
@@ -281,18 +281,24 @@ class MapEditorEditStartsTab(
...
+        nationsToAdd.addAll(nations)
         add(
             MarkupRenderer.render(
-            getNations(),
+            nationsToAdd,
             iconDisplay = FormattedLine.IconDisplay.NoLink
         ) {
             UncivGame.Current.musicController.chooseTrack(it, MusicMood.Theme, MusicTrackChooserFlags.setSpecific)
-            editTab.setBrush(BrushHandlerType.Direct, it, "Nation/$it") { tile ->
+            editTab.setBrush(BrushHandlerType.Direct, when (it) {Constants.spectator -> "Any Civ" else -> it}, "Nation/$it") { tile ->
```
yairm210, line 301: If instead of when

core/src/com/unciv/ui/screens/mapeditorscreen/tabs/MapEditorEditSubTabs.kt
```diff
@@ -281,18 +281,24 @@ class MapEditorEditStartsTab(
...
         add(eraser.render(0f).apply { onClick {
-            editTab.setBrush(BrushHandlerType.Direct, "Remove starting locations", eraserIcon, true) { tile ->
+            editTab.setBrush(BrushHandlerType.Direct, "Remove", eraserIcon, true) { tile ->
                 tile.tileMap.removeStartingLocations(tile.position)
             }
         } }).padBottom(0f).row()
-
+        // Create the nation list with the spectator nation included
+        // We use Nation/Spectator because it hasn't been used yet and we need an icon within the Nation.
+        val anyCiv = FormattedLine("Any Civ starting location", Constants.spectator, "Nation/Spectator", size = 24)
```
yairm210, line 290: User displayed text needs to be added to template.properties

yairm210: So it can be translated

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -585,8 +586,14 @@ object GameStarter {
...
             val startingLocation = if (presetStartingLocation != null) presetStartingLocation
             else {
-                if (freeTiles.isEmpty()) break // we failed to get all the starting tiles with this minimum distance
-                getOneStartingLocation(civ, tileMap, freeTiles, startScores)
+                val presetAnyStartingLocation = tileMap.startingLocationsByNation[Constants.spectator]?.randomOrNull()
+                if (presetAnyStartingLocation != null) { // Since we don't have any nation specific starting locations, pick one of the free ones to use
+                    tileMap.startingLocationsByNation[Constants.spectator]?.remove(presetAnyStartingLocation)
+                    presetAnyStartingLocation
+                } else {
+                    if (freeTiles.isEmpty()) break // We failed to get all the starting tiles with this minimum distance
```
yairm210, line 594: Too much nesting, we need to give a way to better organize this tree of conditions

core/src/com/unciv/logic/GameStarter.kt
```diff
@@ -580,21 +581,36 @@ object GameStarter {
...
+        var startingLocation = tileMap.startingLocationsByNation[civ.civName]?.randomOrNull()
+        if (startingLocation == null) {
+            startingLocation = tileMap.startingLocationsByNation[Constants.spectator]?.randomOrNull()
+            if (startingLocation != null) {
+                tileMap.startingLocationsByNation[Constants.spectator]?.remove(startingLocation)
+            }
+        }
+        if (startingLocation == null && freeTiles.isNotEmpty())
+            startingLocation = getOneStartingLocation(civ, tileMap, freeTiles, startScores)
+        // If startingLocation is null we failed to get all the starting tiles with this minimum distance
```
yairm210, line 611: Won't this throw a type cast error? If failure is an option, and it seems it is, this should return null, so our type result should be 'Tile?'

Issue #9907, opened 2023-08-13
This is a complex bug that causes the game to crash when destroying cities in multiplayer mode


AutumnPizazz:
**Platform**
Both Android and Windows11

**Version**
4.7.13

**Describe the bug**
As the title.
There's some questioned saves. ( I change it as txt and it could be none '.xxx'' file. )
These two save files require my mod "The Coexistence in harmony".

[Autosave-Nation_Qin-24.txt](https://github.com/yairm210/Unciv/files/12329471/Autosave-Nation_Qin-24.txt)
[Autosave-Nation_Greece-14.txt](https://github.com/yairm210/Unciv/files/12329473/Autosave-Nation_Greece-14.txt)

Based on the tests conducted, this issue rarely occurs (or I haven't observed it) when not using any mods. However, it tends to occur more frequently after applying certain mods, although occasionally the problem does not manifest. 

**Screenshots**
![image](https://github.com/yairm210/Unciv/assets/114050365/e691dd27-7188-4b4f-a5e3-348f7f7a0e0c)
![image](https://github.com/yairm210/Unciv/assets/114050365/aa103595-8660-44a2-bfeb-a8397e4a78a5)

AutumnPizazz:
Your response is very profound, to the extent that I may not fully understand your meaning (due to my limited English proficiency). However, I can provide some information to the best of my ability.

AutumnPizazz:
The names "the Coexistence in harmony - xxx" and "... - music" that you mentioned indeed appeared in my previous mod. The actual meaning of "xxx" or "聆彸唳" is "beta version." However, these two mods are only locally isolated and saved in a non-testing version of Unciv, which means they are not used for testing purposes.

AutumnPizazz:
Regarding the issue of non-UTF-8 encoding, what I need to inform you is that I have been using VScode for editing files like JSON and Properties. In the editing interface, I can see that the encoding format is indeed UTF-8.
![image](https://github.com/yairm210/Unciv/assets/114050365/57653fc8-1560-4aef-af76-57e8ce7254c2)

AutumnPizazz:
>Try to open your saves linked above in your VS - either some content will be wrong
Indeed, there is this issue.
![image](https://github.com/yairm210/Unciv/assets/114050365/3a646fff-5fae-4535-8bf7-e0c1b8be938f)

AutumnPizazz:
[...]

PR #9927, opened 2023-08-17
Consistent UTF-8 where possible

SomeTroglodyte:
Thinking about #9907 I reviewed all code I could find involving encodings.
* Enforce UTF-8 where it defaults to Java's """system default"""
* Uniform source of what UTF-8 is: Charsets.UTF_8 everywhere, both where a Charset is needed and where Gdx needs a String.
* Output Java's """system default""" with the system info (so it's only shown with crashes for now)

It's interesting how diverse the responsibilities I found were: Some places need no change because a kotlin-intrinsic stream wrapper enforces UTF-8 explicitly, some work because Gdx JsonReader.parse does so (but other Gdx code doesn't, specifically all FileHandle methods, only where the Json-specifics control the Charset does Gdx "do the right thing"),...

Testing so far limited to compiles and unit tests clean..

### DIFF
--- a/core/src/com/unciv/logic/GameInfo.kt
+++ b/core/src/com/unciv/logic/GameInfo.kt
@@ -43,6 +43,7 @@ import com.unciv.ui.screens.savescreens.Gzip
 import com.unciv.ui.screens.worldscreen.status.NextTurnProgress
 import com.unciv.utils.DebugUtils
 import com.unciv.utils.debug
+import java.nio.charset.Charset
 import java.security.MessageDigest
 import java.util.UUID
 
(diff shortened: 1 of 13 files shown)


yairm210 (approved):

core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -43,6 +43,7 @@ import com.unciv.ui.screens.savescreens.Gzip
 import com.unciv.ui.screens.worldscreen.status.NextTurnProgress
 import com.unciv.utils.DebugUtils
 import com.unciv.utils.debug
+import java.nio.charset.Charset
```
yairm210, line 46: Unnecessary?

yairm210:
I'm fine with that. Anything that brand because of this change probably shouldn't have worked in the first place

Issue #5751, opened 2021-12-05
Improved Research Agreement Feedback


itanasi:
**Is your feature request related to a problem? Please describe.**
Some are frustrated when Research Agreements aren't available. Some feedback as to why (not enough gold or no tech) could help.

**Describe the solution you'd like**
Block Research Agreement but give a similar to buildings Rejection Text as to why you can't propose one in a trade. Such as other Civ doesn't have Education, you or they don't have enough Gold on hand.

**Describe alternatives you've considered**
Yes you can check the opponent's current gold easily in the Trade menu. But not as obvious what Research Agreement's gold requirement is.

**Additional context**
See Alpha Horus comments in Discord
https://discordapp.com/channels/586194543280390151/653651863832363053/916736502627725323

itanasi:
Still valid

PR #9933, opened 2023-08-17
Research agreements with gold offer

tuvus:
Right now, research agreements can't be offered if either of the civilizations doesn't have enough gold to pay for it. This commit allows the research agreement to be offered if the Civ that does have enough money can also pay for the other Civ's cost. If the research agreement can't be payed at all then the button will be disabled but still visible.

Note: There are still some tricks to allow your Civ to get negative gold. Once the research agreement is offered, you can use any amount of the gold you want, once accepted you will then go to negative gold. In base game Civilization, the research agreement or trade would be canceled.

Also note: This commit does not account for gold being offered by both sides at the same because I think that needs to be fixed later on.

### DIFF
--- a/core/src/com/unciv/ui/screens/diplomacyscreen/OfferColumnsTable.kt
+++ b/core/src/com/unciv/ui/screens/diplomacyscreen/OfferColumnsTable.kt
@@ -18,10 +18,12 @@ import com.unciv.ui.screens.basescreen.BaseScreen
 class OfferColumnsTable(
     private val tradeLogic: TradeLogic,
     private val screen: DiplomacyScreen,
-    private val onChange: () -> Unit
+    private val onChange: () -> Unit,
+    private val ourCiv: Civilization,
+    private val theirCiv: Civilization
 ): Table(BaseScreen.skin) {
 
-    private fun addOffer(offer: TradeOffer, offerList: TradeOffersList, correspondingOfferList: TradeOffersList) {
+    fun addOffer(offer: TradeOffer, offerList: TradeOffersList, correspondingOfferList: TradeOffersList) {
         offerList.add(offer.copy())
         if (offer.type == TradeType.Treaty) correspondingOfferList.add(offer.copy())
         onChange()
@@ -104,10 +106,10 @@ class OfferColumnsTable(
             .removeAll(Constants.tradable)
         val theirUntradables = tradeLogic.otherCivilization.getCivResourcesWithOriginsForTrade()
             .removeAll(Constants.tradable)
-        ourAvailableOffersTable.update(ourFilteredOffers, tradeLogic.theirAvailableOffers, ourUntradables)
-        ourOffersTable.update(tradeLogic.currentTrade.ourOffers, tradeLogic.theirAvailableOffers)
-        theirOffersTable.update(tradeLogic.currentTrade.theirOffers, tradeLogic.ourAvailableOffers)
-        theirAvailableOffersTable.update(theirFilteredOffers, tradeLogic.ourAvailableOffers, theirUntradables)
+        ourAvailableOffersTable.update(ourFilteredOffers, tradeLogic.theirAvailableOffers, ourUntradables, ourCiv, theirCiv)
+        ourOffersTable.update(tradeLogic.currentTrade.ourOffers, tradeLogic.theirAvailableOffers, ourCiv = ourCiv, theirCiv = theirCiv)
+        theirOffersTable.update(tradeLogic.currentTrade.theirOffers, tradeLogic.ourAvailableOffers, ourCiv = ourCiv, theirCiv = theirCiv)
+        theirAvailableOffersTable.update(theirFilteredOffers, tradeLogic.ourAvailableOffers, theirUntradables, ourCiv, theirCiv)
     }
 
     private fun openGoldSelectionPopup(offer: TradeOffer, ourOffers: TradeOffersList, maxGold: Int) {
--- a/core/src/com/unciv/ui/screens/diplomacyscreen/OfferColumnsTable.kt
+++ b/core/src/com/unciv/ui/screens/diplomacyscreen/OfferColumnsTable.kt
@@ -18,9 +18,9 @@ import com.unciv.ui.screens.basescreen.BaseScreen
 class OfferColumnsTable(
     private val tradeLogic: TradeLogic,
     private val screen: DiplomacyScreen,
-    private val onChange: () -> Unit,
     private val ourCiv: Civilization,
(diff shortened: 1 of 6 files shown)


yairm210:
I'm not sure this is Civ V behavior - is it?

tuvus:
I know that it isn't base civ V behavior. I don't know about the expansions. But I think it follows the logic that the civilizations looking to form a research pact must have the requirements at the time of signing it. If you want to create a research pact with an AI or a player who just doesn't keep their gold until your turn, you can still then offer it to them and get the benefits. This follows the intentions of Civ V more than the actual implementation like other parts of Unciv do.

core/src/com/unciv/ui/screens/diplomacyscreen/OfferColumnsTable.kt
```diff
@@ -18,10 +18,12 @@ import com.unciv.ui.screens.basescreen.BaseScreen
 class OfferColumnsTable(
     private val tradeLogic: TradeLogic,
     private val screen: DiplomacyScreen,
-    private val onChange: () -> Unit
+    private val onChange: () -> Unit,
```
yairm210, line 21: The OnChange should be the last parameter of the function, since it's a deferred action

core/src/com/unciv/ui/screens/diplomacyscreen/TradeTable.kt
```diff
@@ -49,6 +51,28 @@ class TradeTable(
...
             }
+            // If there is a research agreement trade, make sure both civilizations should be able to pay for it.
+            // If not lets add an extra gold offer to satisfy this.
+            // There must be enough gold to add to the offer to satisfy this, otherwise the research agreement button would be disabled
+            if (tradeLogic.currentTrade.ourOffers.any { it.name == Constants.researchAgreement}) {
+                val researchCost = currentPlayerCiv.diplomacyFunctions.getResearchAgreementCost()
+                val currentPlayerOfferedGold = tradeLogic.currentTrade.ourOffers.firstOrNull { it.type == TradeType.Gold }?.amount ?: 0
+                val otherCivOfferedGold = tradeLogic.currentTrade.theirOffers.firstOrNull { it.type == TradeType.Gold }?.amount ?: 0
+                // Check if we require more gold from them
+                if (currentPlayerCiv.gold + otherCivOfferedGold < researchCost) {
```
yairm210, line 62: Here I see that the sum of both player's gold needs to be the research cost, whereas above in OffersListScroll it's twice the research cost - seems inconsistent?

core/src/com/unciv/ui/screens/diplomacyscreen/TradeTable.kt
```diff
@@ -49,6 +51,28 @@ class TradeTable(
...
+                    offerColumnsTable.addOffer(
+                        tradeLogic.theirAvailableOffers.first { it.type == TradeType.Gold }
```
yairm210, line 64: Please split this to a separate variable, for readability
In line 71 as well

core/src/com/unciv/ui/screens/diplomacyscreen/OffersListScroll.kt
```diff
@@ -109,7 +112,9 @@ class OffersListScroll(
                         else -> 1
                     }
 
-                if (offer.isTradable() && offer.name != Constants.peaceTreaty) { // can't disable peace treaty!
+                if (offer.isTradable() && offer.name != Constants.peaceTreaty // can't disable peace treaty!
+                    && (offer.name != Constants.researchAgreement // If we have a research agreement make sure the total gold of both Civs is higher than the total cost
+                        || (ourCiv.gold + theirCiv.gold > ourCiv.diplomacyFunctions.getResearchAgreementCost() * 2))) {
```
yairm210, line 117: Seems to me this is the wrong place to make this check anyway - we should be making it when populating the offersToDisplay, NOT filtering it out later. That way, we won't have to send the 'ourCiv' and 'theirCiv' parameters to this function at all

core/src/com/unciv/ui/screens/diplomacyscreen/TradeTable.kt
tuvus: In OffersListScroll we are checking if both civilization's gold combined can pay for the total research cost. Here in TradeTable we are checking how much gold one civilization needs to give the other. We can assume that both of them combined can pay for it since otherwise the button would not be pressable.

Just a note: The AI will only offer the research agreement when both sides can pay for their half of the agreement.

core/src/com/unciv/ui/screens/diplomacyscreen/OffersListScroll.kt
tuvus: The only filtering that is being done to the research agreement is at tradeLogic.getAvailableOffers(). In OffersListScroll we are only disabling and enabling it based on if both Civs have enough combined gold or not. If they don't have enough gold we should still display the research agreement button but disable it. The only place to disable a tradeButton is here in offersListScroll.
In Civ V the player can see the research pact and it's cost even if it is not affordable.

I do agree that passing ourCiv and theirCiv is a bit much though.

tuvus:
Shouldn't canSignResearchAgreementsWith() take the max of both Civ's research costs instead of just the current one?
From https://forums.civfanatics.com/resources/research-agreements-bnw.25568/
"RA Costs Modern Era RA with an AI three eras behind​ Research agreements have a gold cost that increases based upon the highest era reached among the participants"

yairm210:
Maybe it should! This isn't an area of Civ V that I investigated, definitely possible we're misaligned :)

yairm210 (approved):
Looks good to me :)
What would happen if I already offered them gold, but then I need to offer them more to offest the cost of the agreement? Do the two gold offers merge properly, or do we end up with 2 different offers of gold? If the latter, we should fix that

tuvus:
There will only be one offer of gold and they do merge properly. However, as noted above, I don't take any gold offer on the other side into account. So for example, you can give enough gold to fund their side of the research agreement and ask for the remaining amount of their gold in the same trade. This is because both sides can offer gold and GoldPerTurn at once. I decided that this has nothing to do with the research pact and is outside of the scope of this change.

I also want to point out that the research pact should cancel if the player offers it and then spends the gold that would go to it on something else(This is Civ V behavior). Another option is that we could immediately deduct the gold and then give it back when the trade offer is canceled. This is probably preferable because you could sign multiple research pacts in the same turn and go negative because they only use the gold when the trade is accepted.

tuvus:
Solves part of #5751.

PR #9941, opened 2023-08-18
Add option to Enable/Disable unit upgrades for automated units

jlmcdonnell:
This PR adds an option to enable or disable unit upgrades for player units that are automated. The default state is enabled, to maintain the existing game behavior.

### Rationale
When units that are automated have the ability to upgrade, they will attempt to do so given that the resources requirements are satisfied. This may not be desirable to players who had intended to use those resources for other purposes.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -1,6 +1,7 @@
 package com.unciv.logic.automation.unit
 
 import com.unciv.Constants
+import com.unciv.UncivGame
 import com.unciv.logic.automation.Automation
 import com.unciv.logic.automation.civilization.NextTurnAutomation
 import com.unciv.logic.battle.Battle
@@ -16,6 +17,7 @@ import com.unciv.logic.civilization.managers.ReligionState
 import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.Tile
 import com.unciv.models.UnitActionType
+import com.unciv.models.metadata.GameSettings
 import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActions
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsPillage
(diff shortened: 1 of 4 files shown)


yairm210:
Good catch!
I'm not sure anyone actually wants automated promotions on their units, I'll Discord poll this, sounds like something we can disable entirely

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -127,6 +129,8 @@ object UnitAutomation {
     }
 
     internal fun tryUpgradeUnit(unit: MapUnit): Boolean {
+        val isPlayer = unit.civ.isCurrentPlayer()
```
yairm210, line 132: This is not the right check, this could prevent AI as well and is less indicative of what we're actually trying to avoid :)
What we want it to check unit.civ.isHuman()

yairm210:
My mistake, I see that you deal here with unit *upgrades*, not promotions.
But the same *is* true for promotions, see https://github.com/yairm210/Unciv/blob/72a7f4df571deb27b4333ee22dd5f9506c375197/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt#L158

yairm210:
So regarding user feedback:
A. We will keep the setting toggle :)
B. We will probably add a similar setting for promotions

core/src/com/unciv/ui/popups/options/GameplayTab.kt
```diff
@@ -42,6 +42,11 @@ fun gameplayTab(
         "Automated units move on turn start",
         settings.automatedUnitsMoveOnTurnStart, true
     ) { settings.automatedUnitsMoveOnTurnStart = it }
+    optionsPopup.addCheckbox(
+        this,
+        "Automated units can upgrade",
```
yairm210, line 47: This text needs to be added to template.properties for translation

huckdogg: Does this line now need the tr() function for translation ?

        "Automated units can upgrade".tr(),

yairm210 (approved):

PR #9942, opened 2023-08-18
ImprovementPicker screen displays tile owner civ and city

huckdogg:
The ImprovementPicker screen now shows the owner civilization and city of the selected tile. Also shows if there is no current owner.

I wasn't exactly sure how to handle the variance in text with the translation file. If there is a better way, please let me know.

ScreenShot_20230818_1692395630
Tile is owned by current player

ScreenShot_20230818_1692395975
Tile has no owner

ScreenShot_20230818_1692395990
Tile is owned by another civilization


core/src/com/unciv/ui/screens/pickerscreens/ImprovementPickerScreen.kt
```diff
@@ -191,6 +191,16 @@ class ImprovementPickerScreen(
             regularImprovements.row()
         }
 
+        val tileOwnerText = if (tile.getOwner() == null) {
+            "Tile owned by no one"
```
yairm210, line 195: "Unowned tile"

huckdogg: I changed the wording for this message.

huckdogg:
I have changed the city name text to a button when the current tile is owned by the player civ (see the updated screenshot in the main post). The button links to the city overview screen, with the current tile selected. Closing the city overview returns to the ImprovementPicker screen.

PR #9951, opened 2023-08-22
Bombard to the north

### LINKED ISSUE #9445: UI prevents city bombard against enemy units in tiles directly above the city
**Platform**
Windows 10 / Android

**Version**
4.6.12-patch1 (github)

**Describe the bug**
When an enemy unit is directly above a city and the city has a garrisoned unit I am unable to bombard that enemy because when I click the city the garrisoned unit is showing in front of and covering the enemy unit. It appears the hitbox of the garrisoned unit completely overlaps the enemy unit so there is no way to select the enemy unit to bombard.

**To Reproduce**
1. Have an enemy unit in the tile above the city
2. Have a garrisoned unit in the city
3. On the map, click on the city name
4. Try to bombard the unit above the city. The garrisoned unit covers the enemy so it can't be clicked on to be bombarded.

</details>

**Expected behavior**
The enemy unit should be able to be clicked on to bombard.

**Screenshots**

Here the destroyer is directly above the city and I have a canon garrisoned in the city:
image

Clicking on the city then causes the garrisoned unit to cover the destroyer so I can't click on the destroyer to bombard it:
image

**Additional context**
I tried adjusting the screen size but it doesn't help. I have 5hex tileset and RekMOD enabled.

A similar situation occurs when there are 2 units garrisoned in the city where I am unable to bombard the enemy in the above right and the above left tiles of the city.

### LINKED ISSUE #9760: Cannot bombard tile directly above city
**Platform**
Windows 11

**Version**
4.7.7

**Describe the bug**
When i choose the city and there are some warrior, then button appears above which is ok. But enemy which is standing under the city. 1 Cell up. I can't damage. Because of huge warrior button are so big so i unable to damage him and always pusing it. Why this isn't fixed?

**To Reproduce**
1. Enemy 1 cell up the city
2. Make some warrior stay in city
3. Try to make city shoot to the enemy. You will see that you will not be able to choose enemy to shoot because there would be a button to chose warrior that stays in the city

</details>

**Expected behavior**
Make button smaller on PC's so player would be abble to access all the cells

**Screenshots**
Screenshot

**Additional context**
Make some additional setting for pc to make this button smaller than on Mobile. Unable to play and it's annoying

SomeTroglodyte:
~~Sure there's an open issue for this - _at least_ one. Please someone find them for me?~~
Closes #9760 closes #9445

There was a workaround described somewhere, in this testing game I just was unable to find it, *despite* having understood it a while back... so... there's something to the complaints.

This is a "just a little" approach: It tweaks the size and displacement of the culpable unit overlay by -20%. Also, since when the garrisoned unit has no movement left I fould you can barely see _why_ you can't select the victim up there, as it's much clearer than the overlaid garrison, I tweaked one alpha a tiny tad up...

***BEFORE***
Selected, garrison has MP // same with debug rects (cover entire north hex) // Selected, garrison has no moves left
image
image
image

***AFTER***
Selected, garrison has MP // demo I managed to hit the barb up there // Selected, garrison has no moves left
image
image
image

</details>

I'm sure this is not the only possible approach (like - nothing of that touchable except the immermost unit circle), and maybe it's not even the cleanest for the intended goal. But - Q&D.

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
@@ -333,12 +333,6 @@ class WorldMapHolder(
              * so that and that alone will be relegated to the concurrent thread.
              */
 
-            /** LibGdx sometimes has these weird errors when you try to edit the UI layout from 2 separate threads.
-             * And so, all UI editing will be done on the main thread.
-             * The only "heavy lifting" that needs to be done is getting the turns to get there,
-             * so that and that alone will be relegated to the concurrent thread.
-             */
-
             val unitToTurnsToTile = HashMap<MapUnit, Int>()
             for (unit in selectedUnits) {
                 val shortestPath = ArrayList<Tile>()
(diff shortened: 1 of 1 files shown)


yairm210:
A. This looks like a good solution, I tried many things and nothing really worked :)
B. Have you tested how easy it is to select the unit in the city, on a phone? 48 *should* be enough to be clickable, but I want to make sure

yairm210 (approved):

PR #9955, opened 2023-08-22
"Jump To Destination" Unit Action Button for units that are moving

huckdogg:
When a unit is moving to position that will take multiple turns, a new button called "Jump to Destination" is displayed in the unit actions button list. Clicking the button will move the map view to the destination tile and cause the white circle at the tile to flash.

image

This commit also fixes the situation where the player is unable to open the Show More unit actions list while the unit is moving and out of moves for the turn. Previously, clicking Show More in this scenario would jump to the next active unit.


yairm210:
- Need icon credit
- 'Jump to destination' sounds like the *unit* will be jumping to its destination - how about 'show unit destination'? That way it's clear that no action is being taken
- Agree that it's getting a bit crowded...
- Maybe we can highlight the destination tile with blue/white, so that if you have units that are close to their destination you won't need to click on that since you can just see it clearly? Not sure if suited to the same PR though

huckdogg:
I changed the wording to "Show unit destionation", added a new icon with credit, and moved the button to the Show More menu. I didn't adjust the Show Destination behavior at this point. Clicking the button always centers the screen on the destination tile.

PR #9961, opened 2023-08-23
City construct menu

SomeTroglodyte:
In the works...

image


yairm210:
How do you open this exactly?

Ouaz:
Really nice addition.

There's a little bug though (or maybe it's intended?).
When using the buttons in the construction queue (delete, up/down arrows), it opens the construct menu each time (without right-clicking) after a slight delay.

construct_buttons

PR #9965, opened 2023-08-24
Add tests for city class and introduce small refactor

Framonti:
This change is not affiliated to any GitHub issue

Introduction of some tests for the most complex methods of class City.
Introduction of refactoring by extracting methods to help readability of one of aforementioned methods

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -202,19 +202,39 @@ class City : IsPartOfGameInfoSerialization {
     fun getCityResources(): ResourceSupplyList {
         val cityResources = ResourceSupplyList()
 
-        val resourceModifer = HashMap<String,Float>()
+        val resourceModifer = HashMap<String, Float>()
         for (resource in civ.gameInfo.ruleset.tileResources.values)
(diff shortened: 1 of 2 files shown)


Framonti:
Hi @SomeTroglodyte ,
It may absolutely be possible that some tests were already covering parts of the code (probably indirectly, I checked usages of the methods under tests before writing them and found no references in tests). I'm not really familiar yet with the codebase, so I must have missed them. 

For example, there is the following test 
```
@Test
    fun testConsumesResourceUnique() {
        val consumesCoal = game.createBuilding("Consumes [1] [Coal]")
        city.cityConstructions.addBuilding(consumesCoal.name)
        Assert.assertTrue(civInfo.getCivResourcesByName()["Coal"] == -1)
    }
```
Which does look pretty similar to one of the test I wrote, but it's not quite the same.

No problems about waiting before merging 👍

Framonti:
I understand. 
I can see how some tests may have looked similar, no worries. I double-checked the existing tests. There exist some that are similar, as the example I showed before. But no tests directly checked the behaviour of City class methods.

Thanks for your code review👍

PR #9967, opened 2023-08-24
AI now move units closer to enemy first in wartime

tuvus:
Units tend to be clumped up in a large ball when they need to move into enemy territory to attack. This code prioritizes moving the units in front first so the AI can move them quicker to their target with fewer inefficiencies.

If you take a medium or large map with nukes disabled, set yourself as the spectator and add 6 AI civs, then let them play out. Once they get a huge death ball, they are very inefficient with fighting. Therefore I created this first change to help fix this.

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -20,6 +20,7 @@ import com.unciv.logic.civilization.diplomacy.DiplomaticStatus
 import com.unciv.logic.civilization.diplomacy.RelationshipLevel
 import com.unciv.logic.civilization.managers.ReligionState
 import com.unciv.logic.map.BFS
+import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.Tile
 import com.unciv.logic.trade.Trade
 import com.unciv.logic.trade.TradeEvaluation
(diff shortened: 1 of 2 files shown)


tuvus:
I think this is only temporary. I am planning on making more AI changes, and having the distance to the closest enemy unit stored in each unit is very valuable. So I might end up attaching it to the mapUnit itself. There is much to change for the AI.

yairm210:
I think we can consider TacticalAI as an experiment that never got off the ground, it's not really in use but you can think of it as inspiration

tuvus:
I run many games with AI only at a later era so that they can build up into a death ball and I can test the AI. Playing a game to that point would take too long and sometimes I need multiple saves to test. The AI doesn't perform well and tends to leave cities unharmed even when it is surrounded by units (I am working on that right now). So most of changes will probably go into improving the spectator features and the AI.

Screenshot_20230824-143059_Unciv

As you can see in this screenshot, Siam has a very large army but it takes forever for it to get anywhere because the units in the back have nowhere to go. The save is around 4.7.12 I think.

Also, I would like to suggest the option of allowing players to use the automated features for half or more of their turn. For example, if Siam here were a player, turns would take like 20 minutes to complete. This is way too long, so instead of moving every unit, the player does their most important moves and decisions (For example tech, policies or nukes) first and presses button(s) to automate some or all of the rest. From what I have looked into the code it seems like this would be "relatively" simple to do.

tuvus:
Here's another one, this time with modifications to the war AI. I'm looking at the unit behavior.

CivBigArmyWithAI

PR #9968, opened 2023-08-24
Ai nuke improvement

tuvus:
This commit completely changes the Nuking AI. Previously, the Nuke AI would only attack using the SpecificUnitAutomation.automateBomber function since it didn't filter out Nukes.
The Nuke AI prioritizes tiles based on cities, units, and tile improvements. By default, the AI does not launch its nuke unless the value of the tiles hit is high.

The AI is not allowed to nuke tiles that are not visible, and it will only account for visible units.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
@@ -494,6 +494,7 @@ object SpecificUnitAutomation {
     }
 
     fun automateNukes(unit: MapUnit) {
+        if (!unit.civ.isAtWar()) return
         // We should *Almost* never want to nuke our own city, so don't consider it
         val tilesInRange = unit.currentTile.getTilesInDistanceRange(2..unit.getRange())
         var highestTileNukeValue = 0
@@ -513,7 +514,7 @@ object SpecificUnitAutomation {
 
     /**
      * Ranks the tile to nuke based off of all tiles in it's blast radius
-     * By default the value is -400 to prevent inefficient nuking.
+     * By default the value is -500 to prevent inefficient nuking.
      */
     fun getNukeLocationValue(nuke: MapUnit, tile: Tile): Int {
         val civ = nuke.civ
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -632,8 +632,8 @@ object UnitAutomation {
             .map { BattleDamage.calculateDamageToDefender(MapUnitCombatant(it), cityCombatant) }
             .sum() // City heals 20 per turn
 
-        if (expectedDamagePerTurn < city.health && // If we can take immediately, go for it
-                (expectedDamagePerTurn <= 20 || city.health / (expectedDamagePerTurn-20) > 5)){ // otherwise check if we can take within a couple of turns
+        if (expectedDamagePerTurn < city.health * 4 && // If we can take immediately, go for it
+                (expectedDamagePerTurn <= 5 || city.health / (Math.max(1, expectedDamagePerTurn-20)) > 5)){ // otherwise check if we can take within a couple of turns
```
yairm210, line 636: This is incorrect - the city heals 20 per turn, that's the basis for this calculation. If you deal 1 per turn for 5 turns, you won't get the city- it's healing faster than your dealing damage!

yairm210: But I agree that we should have put it in a named val cityHealingPerTurn for context clarity

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -493,27 +494,78 @@ object SpecificUnitAutomation {
...
+            }
+            // Never nuke our own Civ, don't nuke single enemy civs as well
+            if (targetTile.isCityCenter() 
+                && !(targetTile.getCity()!!.health <= 50f 
+                    && targetTile.neighbors.any {it.militaryUnit?.civ == civ})) // Prefer not to nuke cities that we are about to take
+                explosionValue += evaluateCivValue(targetTile.getCity()?.civ!!, -100000, 250)
+            else if (targetTile.owningCity != null) {
+                val owningCiv = targetTile.owningCity?.civ!!
+                // If there is a tile to add fallout to there is a 50% chance it will get fallout
+                if (!(tile.isWater || tile.isImpassible() || targetTile.terrainFeatures.any { it == "Fallout" }))
```
yairm210, line 557: Is fallout parametrized in the nuke unique? I think it is, in which means this should not be a hardcoded "fallout" string

tuvus: I am not sure, lines 905 and 906 in Battle.kt use "Fallout". Fallout uses UniqueType.NullifyYields though which might have been used to abstract it from the effect.

yairm210 (approved):

PR #9970, opened 2023-08-24
Fix Permanent Audiovisual toggle and start on declarative Mod Compatibility

SomeTroglodyte:
See commit comments.
* The first fixes an actual bug (whodunnit? some troglodyte I don't know)
* Couldn't stop there - I wish for a world where that option is only there when it's actually meaningful, but that cannot be reliably inferred from current data. So, the mod author should help...

Notes....
* Declarative "Permanent Audiovisual" compatibility is complete and works fine for me. However, it's still "if in doubt guess" - if this approach is welcomed, then we could remove that part once all relevant mods have their uniques.
* Mod-Mod incompatibility is intentionally incomplete, as in waiting for approval. No downsides from merging afaik, but the to-do's would be:
        * Incompatibility extension / base mod works, ext/ext not implemented. Could reject turning the second one on, or toast, or flip the other off, or...
        * I'd also imagine a "Requires" - implementation could be a little more complicated when that references an ext mod, could be pretty straightforward for base ones
        * Modcheck UI could use that to avoid a bunch of meaningless comparisons
        * Display these uniques somewhere? where?
        * I'd also imagine a voluntary "Compatible" promise - could be valuable if we also introduced a user option like "strict/lax", in the former case only combos one of the parties promises works would be allowed...
* All my mods have those options since today, if you want to test.
* Visuals - screenies - not very helpful as it's so subtle. e.g. "Modern Joke Religions" simply disappears from the list if you switch G&K to Vanilla, won't enter a new game, and doesn't participate in the modcheck run by the newgame screen. Once you switch back to G&K it's back and even kept selection... My larger music mods are gone from the per-game-selectable mods too, because I said so (wiki should recommend to use that only if mod auth is sure nobody wants to force their friend to turn that mod on via a save). So mods with, say, music themes for specific nations from other mods should pick the other unique.
* Documentation - where should more details go? How to reach the modders? If the concept is good, we should fix the missing UniqueType's text before implementing, modders could start on them early?

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
@@ -510,6 +510,16 @@ enum class UniqueParameterType(
         override fun getTranslationWriterStringsForOutput() = knownValues
     },
 
+    /** Mod declarative compatibility: Behaves like [Unknown], but makes for nicer auto-generated documentation. */
+    ModName("modFilter", "DeCiv Redux", """A Mod name, case-sensitive _or_ a simple wildcard filter beginning and ending in an Asterisk, case-insensitive""", "Mod name filter") {
+        override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
+            UniqueType.UniqueComplianceErrorSeverity? =
+            if ('-' !in parameterText && ('*' !in parameterText || parameterText.matches(Regex("""^\*[^*]+\*$""")))) null
+            else UniqueType.UniqueComplianceErrorSeverity.RulesetInvariant
+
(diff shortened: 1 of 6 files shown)


yairm210 (approved):

yairm210:
Modding-  there's a doc specifically for audiovisual mods that I'm pretty sure *you* wrote, that sounds like the natural place
I hope the "can be" does not (currently) prevent mods without it from during the way they did up till now? We need a transitory period if we're going to change behavior for existing mods

yairm210:
Actually looking through it again I can't think of a mod that would be affected in this way, and if there *are* I would probably rather know about it directly, so... Unsure what the correct way forward should be here...

PR #9973, opened 2023-08-26
Refactor: Remove unnecessary ruleset checks

SeventhM:
This PR primarily removes many cases of explicitly passing around the names of IConstructions. The goal is to address many, many, many situations where a function will check the ruleset multiple times despite knowing that it is in the ruleset. This also includes avoiding `getConstruction` when necessary as not only does it check the ruleset, but it can sometimes give a building if a unit and building share the same name... which is a problem if we expected a unit

This PR also includes a number of smaller fixes, as well as a number of other notes:
1. Many functions have been split into the main one and one that still look up a unit/building as a string when necessary. This is usually in situations where it normally wouldn't know what it's looking for (such as objects that need deserializing)
    - Note that removeBuilding is one of the few exceptions that doesn't just look up the object and then immediately call the function that does assume we have an object. The reasoning is remove building is currently functioning in an... interesting way. There is a function in logic.BackwardsCompatibility that does genuinely want to remove the building via a string and unit tests that also does the same. I think this should probably be adjusted (why are we removing buildings if we don't know it's there), but I decided not to do that here
2. In cityConstructions.addBuilding, I moved updating road connections to updateUniques and moved updateUniques again. On one hand, this is to help [avoid confusion](https://github.com/yairm210/Unciv/issues/9926). On the other, this does have real world side effects. For example, in Rekmod, I'm counting a building via a resource in order to do a trigger (btw, we need a conditional for this), and currently, I need one less than the amount of buildings to function correctly. This PR would change this functionality to actually counting the resource on the building itself as well for the trigger
3. Similar to removeBuilding, I find it weird that in theory you can sell a building you don't have
4. changeImprovement has been changed to always update resources, since improvements can have resources themselves and we're already always updating stats. I'm not sure why the qualifier was there in the first place. I simply noticed this when I was editing down the tests
5. https://github.com/yairm210/Unciv/blob/066c698117c15a2cc22282656b95585972c266c4/core/src/com/unciv/logic/battle/Battle.kt#L736C68-L736C68 This line should probably be less strict than just "Land". A mod like Rekmod actually doesn't have any units valid under this as I switched this from "Land" to "Any" in order to build water improvements

### DIFF
--- a/core/src/com/unciv/logic/BackwardCompatibility.kt
+++ b/core/src/com/unciv/logic/BackwardCompatibility.kt
@@ -68,7 +68,7 @@ object BackwardCompatibility {
 
             for (building in city.cityConstructions.getBuiltBuildings()) {
                 if (!ruleset.buildings.containsKey(building.name))
-                    city.cityConstructions.removeBuilding(building.name)
+                    city.cityConstructions.removeBuilding(building)
             }
 
             fun isInvalidConstruction(construction: String) =
--- a/core/src/com/unciv/logic/GameStarter.kt
+++ b/core/src/com/unciv/logic/GameStarter.kt
@@ -454,7 +454,7 @@ object GameStarter {
         ruleset: Ruleset,
         eraUnitReplacement: String,
         settlerLikeUnits: Map<String, BaseUnit>
-    ): String? {
+    ): BaseUnit? {
         var unit = unitParam // We want to change it and this is the easiest way to do so
         if (unit == Constants.eraSpecificUnit) unit = eraUnitReplacement
         if (unit == Constants.settler && Constants.settler !in ruleset.units) {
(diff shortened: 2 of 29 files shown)


SeventhM:
Also, I mentioned this on Discord, but will mention this here to get @SomeTroglodyte's opinion on this: I previously noticed these optimizations when working on unit tests for other things. However, it seemed to me at the time that it wasn't as ideal thanks to requiring the name to pasted around guarantees other contributors can't edit baseUnits or Buildings mid code as easily without breaking things. Obviously, at least some of the optimizations in this PR are just strictly better than what was before (why are we calling getConstruction when we already have a building that we *just* got from the ruleset? Why is there a spot where this leads to us mapping something twice for no reason?)  but I'm still not sure what the opinion on defensively avoiding people assuming what the already have is in the ruleset vs just optimizing away these extra checks

SeventhM:
> Will conflict with #9961 as I do some of the same in there.. Which means basically approval to this PR (without even looking at the diff).

Comparing the diff, I'm not actually certain that's true. It looks like we're doing this in separate places from each other (I'm mostly concerned with addUnit, addBuilding, removeBuilding, placeUnitNearTile, and purchaseConstruction, and basically any side effects of editing calls to those functions, not with the enqueued functions)

Though if I got approval from you and Yairm said to PR any optimization I noticed via this, I'm set this as ready then

SeventhM:
> 5 - Nope? That just means capturing civilians will work differently on Land, not that you can't do Improvement-builders on Water.

No, it means that capturing settlers won't give workers in my ruleset because there's no worker who specifically builds land improvements

yairm210:
I approve of this and an ready to merge, @SomeTroglodyte any objections? Just so if there are conflicts, we can get to them early

yairm210:
Since no objections have been made, I'll merge it now

PR #9979, opened 2023-08-27
Improved AI chooseAttackTarget

tuvus:
This commit restructures the AI's chooseAttackTarget to assign a value to each tile that can be attacked and choose the best one. This style should be easier for future developers to tweak and edit. It may also be helpful later when we want to say chooseAttackTarget with value > 100 or else heal or some more complex military action instead of having a series of if statements.

Military units are now more inclined to attack Cities by default. The more allied units around the city, the higher the priority. Siege units now almost fully prioritize cities (as that is what they are meant to do).


core/src/com/unciv/logic/automation/unit/BattleHelper.kt
```diff
@@ -52,64 +54,101 @@ object BattleHelper {
...
+            // Associate enemy units with number of hits from this unit to kill them
+            val attacksToKill = (militaryUnit.health.toFloat() / 
+                BattleDamage.calculateDamageToDefender(MapUnitCombatant(attacker), MapUnitCombatant(militaryUnit))).coerceAtLeast(1f)
+            // We can kill them in this turn
+            if (attacksToKill <= 1) attackValue += 100
+            // On average, this should take around 3 turns, so -30
+            else attackValue -= (attacksToKill * 10).toInt()
+        } else {
+            val civilianUnit = attackTile.tileToAttack.civilianUnit
+            if (civilianUnit != null) {
```
yairm210, line 135: Please babe this 'else if', to reduce nesting - you can set the val inside the else if

core/src/com/unciv/logic/automation/unit/BattleHelper.kt
```diff
@@ -52,64 +54,101 @@ object BattleHelper {
...
+                attackValue = 0
+                attackValue += 50
+                // Only melee units should really attack/capture civilian units, ranged units take more than one turn
+                if (attacker.baseUnit.isMelee()) {
+                    if (civilianUnit.isGreatPerson()) {
+                        attackValue += 50
+                        // This is really good if we can kill a great general
+                        if (civilianUnit.isGreatPersonOfType("Great General")) attackValue += 150
```
yairm210, line 143: In general, all great people should be prioritized

core/src/com/unciv/logic/automation/unit/BattleHelper.kt
```diff
@@ -52,64 +54,101 @@ object BattleHelper {
...
+                    }
+                    if (civilianUnit.hasUnique(UniqueType.FoundCity)) attackValue += 30
+                }
+            }
         }
+        // Prioritise closer units as they are generally more threatening to this unit
+        attackValue += (attackTile.movementLeftAfterMovingToAttackTile * 5).toInt()
```
yairm210, line 150: Only true for military units, no? That way both sides of the 'if' can return immediately, as well - better for readability

core/src/com/unciv/logic/automation/unit/BattleHelper.kt
```diff
@@ -52,64 +54,101 @@ object BattleHelper {
...
+    /**
+     * Returns a value which represents the attacker's motivation to attack a unit.
+     */
+    private fun getUnitAttackValue(attacker: MapUnit, attackTile: AttackableTile): Int {
+        // Base attack value, there is nothing there...
+        var attackValue = Int.MIN_VALUE
+        // Prioritize attacking military
+        val militaryUnit = attackTile.tileToAttack.militaryUnit
+        if (militaryUnit != null) {
+            attackValue = 0
```
yairm210, line 124: =150 directly

core/src/com/unciv/logic/automation/unit/BattleHelper.kt
```diff
@@ -52,64 +54,101 @@ object BattleHelper {
...
 
+    /**
+     * Choses the best target in attackableEnemies, this could be a city or a unit.
+     */
     private fun chooseAttackTarget(unit: MapUnit, attackableEnemies: List<AttackableTile>): AttackableTile? {
-        val cityTilesToAttack = attackableEnemies.filter { it.tileToAttack.isCityCenter() }
-        val nonCityTilesToAttack = attackableEnemies.filter { !it.tileToAttack.isCityCenter() }
-
+        var highestAttackValue = 0
+        var attackTile: AttackableTile? = null
```
yairm210, line 62: This whole thing feels like maxBy {}

core/src/com/unciv/logic/automation/unit/BattleHelper.kt
tuvus: Do you mean 'UniqueType.GreatGeneralProvidesDoubleCombatBonus'? It doesn't sound right though, other than that, I can't find any other Unique to match it.

core/src/com/unciv/logic/automation/unit/BattleHelper.kt
tuvus: I don't think so. The idea is, that if the Unit has to move less to attack, it is closer to its friends and therefore, less vulnerable.

trentonk23: NEEDS THIS

yairm210: Ah good point

yairm210 (approved):

PR #9986, opened 2023-08-27
Fix promotions being available when they shouldn't

### LINKED ISSUE #9982: "Only available <for units with [promotion]>" not working
Windows 10, itch.io v. 4.7.17-patch1

I designed a promotion with the prerequisite of Drill I but with the unique of `"Only available <for units with [Shock I]>"`. The idea was to make a promotion that requires both previous promotions as prerequisites. It's not working; units with Drill I can access the promotion even without getting Shock I.
[PromotionTest.zip](https://github.com/yairm210/Unciv/files/12448942/PromotionTest.zip)

SeventhM:
Closes #9982

### DIFF
--- a/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
+++ b/core/src/com/unciv/ui/screens/pickerscreens/PromotionTree.kt
@@ -142,7 +142,7 @@ class PromotionTree(val unit: MapUnit) {
                         // else: Already reached, but a worse way - overwrite fully
                     }
                     child.depth = depth + 1
-                    child.distanceToAdopted = distance
+                    if (!child.unreachable) child.distanceToAdopted = distance // Only edit this if the node could possibly be reached
                     child.pathIsAmbiguous = node.pathIsAmbiguous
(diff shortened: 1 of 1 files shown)


SeventhM:
On second thought, I think this approach is probably wrong. The problem currently is if we avoid actually calculating unreachable nodes, the depth might become inaccurate which might affect something (Idk). I'm not at a computer right now so I can't test, but honestly, setting the distance to adopt to the maximum value ahead of time seems honestly unnecessary. So does making sure we don't iterate through there. So setting stuff to unreachable seems almost better done \*after\* we iterate through the nodes

Btw, I am aware that technically, you can just have `canBuyUpTo` return if the node is unreadable. I'm actually not doing it that way partially because I haven't double checked everything here yet and to keep consistency for how we normally check these things

SeventhM:
Ok, nevermond my nevermind, this likely takes the path of unreachable nodes occasionally as perfered, which is not ideal. Probably a good reason why I shouldn't edit code without a computer to test at

yairm210 (approved):

PR #9987, opened 2023-08-28
Allow buildings to require population, Allow buildings to use condtionals

### LINKED ISSUE #9977: "Requires at least [amount] population" not working on buildings
Windowss 10, version 4.7.17-patch1

The `"Requires at least [amount] population"` unique isn't applying to buildings. You can build the buildings even when the population is too low. I made a little test mod with a building that should have required three population, and it came as an option to build when the population of the city was only one.
[TestBuilding.zip](https://github.com/yairm210/Unciv/files/12447218/TestBuilding.zip)

SeventhM:
Closes #9977

I do find it a bit weird that the way Units and buildings check their uniques is ever so slightly different. Also, the unit to require population technically does state it works for buildings, so this PR just makes this work as expected


yairm210:
If there's enough common functionality to warrant a unified function in INonPerpetualConstruction then I'm for it

PR #9992, opened 2023-08-28
Refactor BattleDamage object and test it

Framonti:
This PR is not affiliated with any GitHub issue

* Small refactors for object BattleDamage by introducing helper functions and moving magic constant numbers to dedicated file
* Introduce tests covering most common damage modifiers cases

I suggest looking at the PR commit-by-commit

### DIFF
--- a/core/src/com/unciv/logic/battle/BattleConstants.kt
+++ b/core/src/com/unciv/logic/battle/BattleConstants.kt
@@ -1,0 +1,15 @@
+package com.unciv.logic.battle
+
+object BattleConstants {
+    //https://www.carlsguides.com/strategy/civilization5/war/combatbonuses.php
+    const val LANDING_MALUS = -50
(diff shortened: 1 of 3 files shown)


yairm210:
To me, an Enum is things that are logically groupable - as in, you will iterate on them, or have a function that accepts any of that group. That way they share a context.

These values share no context whatsoever, they're just a bunch of consts that are used in different places

Framonti:
Hi @SomeTroglodyte , thanks for the typo catch!
I've fixed it

Framonti:
Are there any more changes to be done, or can we merge this PR?

yairm210:
Will be merged on the next version release, that's what approval means

Framonti:
Makes sense, I'm used to having a develop branch but here we're not using a gitflow like that, thanks for let me know.

Issue #9591, opened 2023-06-15
Feature RFC: Civilopedia Search


SomeTroglodyte:
https://github.com/yairm210/Unciv/assets/63000004/bf5a61d8-fd79-469d-b5c8-66a9be6cd569

[For now, an imperfect test-of-concept](https://drop.picasoft.net/r/VkfpLu7S#Wnjqv1KDWdmiQ4OrUPJQgaKs0JnUaPmCfba6tjzHkgk=). 

Should search current *combined* ruleset by default, forgot the dropdown entry. Expander because the new one animates nicer (from bottom) while results come in from the background job - should replace w/standard pane. It's fulltext on the translated pedia texts, but I haven't really dug into whether there are loopholes it could miss. Also didn't know where to stash a UI for the question "I'm looking at Sentry Tower - which mod is it from?" Filtering mod by mod finds the answer, but...

yairm210:
Definitely interesting! I'm not convinced that the dropdown is actually necessary since when you have a ruleset they you're using that entire ruleset

As for "source for objects", we could do this in the Ruleset creation - have each object contain an extra string for "sourceRuleset", and populate it when we read the files to create the objects
Then we can perhaps display that at the bottom of each Civilopedia item, just a small label of "source: {modName}"

CrsiX:
I like that. I would space the results a little more, I guess one could miss-tap easily (at least in that view on smaller Android phones).
Also, do you have a minimum number of chars? Searching for `e` would probably yield a million results, basically the whole civilopedia, maybe only search for something like three or more characters?

hackedpassword:
> I like that. I would space the results a little more

You know, trying to select from drop down menus is already a game of Operation...

@SomeTroglodyte Awesome. I like the concept of being able to search per mod, since what I'm looking for is almost certainly targeted to the mod in context.

Would you be able to attach the associated icon to each item?

hackedpassword:
The explanation is appreciated @SomeTroglodyte. From the UX/UI side I think "why does it have to be so complicated, Just stick an imagine on the line!" but the rational knows-just-enough-software-dev-be-dangerous side says "stay out of the way and let real devs solve resource puzzles you don't even know where to start with!" Quite the internal conflict. Funny thing is, I showed up to this thread for exactly this, to hear more about the internal mechanics of the game. One day maybe it'll make coherent sense.

yairm210:
@SomeTroglodyte Bump - would gladly have this in the game

PR #9997, opened 2023-08-29
Pedia Search

SomeTroglodyte:
Mostly unchanged code from what produced the clip in #9591.

<details><summary>Another clip of the buttons</summary>

https://github.com/yairm210/Unciv/assets/63000004/1bd4c7c9-c9bd-4515-84e2-92ff28c85216

</details>
... That's something I couldn't really decide. Leave that move+replace of the Close button in (it is in this PR), or revert entirely and postpone the decision how and where and whether at all to make close positioning consistent, or - just replace the text close with this icon one ... (Hey wow even though peek is long abandoned the 'show mouse' option now works with the latest Mint upgrade)


core/src/com/unciv/models/ruleset/Belief.kt
```diff
@@ -23,7 +23,7 @@ class Belief() : RulesetObject() {
     override fun makeLink() = "Belief/$name"
     override fun getCivilopediaTextHeader() = FormattedLine(name, icon = makeLink(), header = 2, color = if (type == BeliefType.None) "#e34a2b" else "")
     override fun getSortGroup(ruleset: Ruleset) = type.ordinal
-    override fun getIconName() = if (type == BeliefType.None) "Religion" else type.name
+    //override fun getIconName() = if (type == BeliefType.None) "Religion" else type.name -> No, ImageGetter does that
```
yairm210, line 26: Delete?

core/src/com/unciv/models/ruleset/Speed.kt
```diff
@@ -56,7 +56,7 @@ class Speed : RulesetObject(), IsPartOfGameInfoSerialization {
 
     override fun getUniqueTarget(): UniqueTarget = UniqueTarget.Speed
 
-    override fun makeLink(): String = "GameSpeed/$name"
+    override fun makeLink(): String = "Speed/$name"
```
yairm210, line 59: I have a suggestion - since these are linked to the Enum of the page, we can add a 'fun `makeLink(CivilopediaCategory:Category)` in ICivilopedia and use that, instead of this manual string stuff.

But probably a different PR.

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaCategories.kt
```diff
@@ -115,87 +118,105 @@ enum class CivilopediaCategories (
     val hide: Boolean,      // Omitted on CivilopediaScreen
     val getImage: ((name: String, size: Float) -> Actor?)?,
     val key: KeyCharAndCode = KeyCharAndCode.UNKNOWN,
-    val headerIcon: String
-    ) {
+    val headerIcon: String,
+    val getCategoryIterator: (ruleset: Ruleset, tutorialController: TutorialController) -> Collection<ICivilopediaText>
```
yairm210, line 122: tutorialController is only used in one of these, can't we just pass in the TutorialController value *not* as part of the lambda signature? That way they'll all be Ruleset -> result, without all these _'s everywhere

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaSearchPopup.kt
```diff
@@ -0,0 +1,178 @@
...
+
+    private var searchJob: Job? = null
+    private var checkLine: (String) -> Boolean = { _ -> false }
+
+    init {
+        searchText.maxLength = 100
+
+        add("{Search text}:".toLabel())
+        add(searchText).growX().row()
+        add("Mod filter:".toLabel())
```
yairm210, line 50: translation values. Also, why the difference? Either both should get the : as part of the original text, or neither

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaSearchPopup.kt
```diff
@@ -0,0 +1,178 @@
...
+            resultExpander.innerTable.clear()
+        } else {
+            resultExpander = ExpanderTab("Results") {}
+            resultCell.setActor(resultExpander)
+            resultExpander.innerTable.defaults().left().growX().pad(2f)
+        }
+
+        searchJob = Concurrency.run("PediaSearch") {
+            searchLoop()
+        }.apply {
```
yairm210, line 102: Take this out of apply to reduce nesting - this is bad enough as it is

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaSearchPopup.kt
```diff
@@ -0,0 +1,178 @@
...
+                ToastPopup("That's an invalid regular expression.", pediaScreen, 4000).open(true)
+                searchButton.enable()
+                return
+            }
+        } else {
+            val words = text.split(' ').toSet()
+            checkLine = { line -> words.all { line.contains(it, ignoreCase = true) } }
+        }
+
+        ruleset = modSelect.selectedRuleset() ?: return
```
yairm210, line 90: You re-enable is on line 82, but not here - intentional since there is no ruleset?

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaSearchPopup.kt
```diff
@@ -0,0 +1,178 @@
...
+
+        if (::resultExpander.isInitialized) {
+            resultExpander.innerTable.clear()
+        } else {
+            resultExpander = ExpanderTab("Results") {}
+            resultCell.setActor(resultExpander)
+            resultExpander.innerTable.defaults().left().growX().pad(2f)
```
yairm210, line 97: .left() AND .growX()?! I am baffled

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaSearchPopup.kt
```diff
@@ -0,0 +1,178 @@
...
+
+        searchJob = Concurrency.run("PediaSearch") {
+            searchLoop()
+        }.apply {
+            invokeOnCompletion {
+                searchJob = null
+                Concurrency.runOnGLThread {
+                    if (resultExpander.innerTable.cells.isEmpty)
+                        resultExpander.innerTable.add(
+                            FormattedLine("Nothing found!", color = "#f53", header = 3, centered = true)
```
yairm210, line 108: Separate this to a val please.

core/src/com/unciv/ui/screens/civilopediascreen/CivilopediaSearchPopup.kt
```diff
@@ -0,0 +1,178 @@
...
+            if (!isActive) break
+            if (category.hide) continue
+            if (!ruleset.modOptions.isBaseRuleset && category == CivilopediaCategories.Tutorial) continue
+            for (entry in category.getCategoryIterator(ruleset, tutorialController)) {
+                if (!isActive) break
+                if (entry !is INamed) continue
+                if (!ruleset.modOptions.isBaseRuleset) {
+                    val sort = entry.getSortGroup(ruleset)
+                    if (category == CivilopediaCategories.UnitType && sort < 2) continue
+                    if (category == CivilopediaCategories.Belief && sort == 0) continue
```
yairm210, line 127: Magic voodoo numbers - entirely unhelpful, can't tell what's going on

SeventhM:
> Did I change that code inadvertently here

No, I added the update to the ressources back in a PR I did, and then it was removed [here](https://github.com/yairm210/Unciv/commit/41fa29f50c7d5fed8cad46cebe45ce4cbf2c6631).

> The comment clearly says "DO NOT pass when simulating"

Comment refers to changing the improvement secretly in `ui.screens.mapeditorscreen
.TileInfoNormalizer`, not anything to do with unit tests. I kinda question doing things this way tbh (there's some weird implications from checking the civ rather than just passing in a boolean that explictly says we're simulating, but that's probably better saved for another PR/commit)

yairm210:
I'm fine with passing in a boolean for simulating as well, this was just the easiest solution at the time. I didn't take into account how many things need to be done on real improvement creation vs simulation...

core/src/com/unciv/models/ruleset/Belief.kt
yairm210: Please delete this line now that we've seen the comment :)

android/assets/jsons/translations/template.properties
```diff
@@ -1600,6 +1600,13 @@ Toggle UI (World Screen only) =
 Overrides yields from underlying terrain = 
 No yields = 
 Mod: [modname] = 
+Search text: = 
+That's an invalid regular expression. =
```
yairm210, line 1604:
```suggestion
Invalid regular expression = 
```
Just as informative, and shorter

core/src/com/unciv/models/ruleset/Belief.kt
yairm210: Reminder

yairm210 (approved):

yairm210:
Please remove the commented out line, though

PR #10012, opened 2023-08-31
Fix: "Only available" not working properly for religions or transforming/upgrading units

SeventhM:
Tbh, I'm not as much of a fan of this implementation. "Only available" needs special rules so often that I almost feel like maybe the better solution is to actually address this in `conditionsApplies` rather than everytime it comes up

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -41,6 +41,7 @@ import com.unciv.models.ruleset.PolicyBranch
 import com.unciv.models.ruleset.Victory
 import com.unciv.models.ruleset.tech.Technology
 import com.unciv.models.ruleset.tile.ResourceType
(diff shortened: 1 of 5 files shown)


yairm210:
We definitely need some centralized function for dealing with Only available uniques, I'm just not sure where would be the correct place to put it...

SeventhM:
On my personal build, I tried out putting whether or not a unique is always ignored as a parameter for UniqueType, and putting a check for that variable in getMatchingUniques. It's not entirely clean (why would we store a whole variable for just one unique? Also there's still crazy situations such as #9987 where we don't use getMatchingUnique for I believe proformance reasons where specific fixes are still necessary), but it does cut the need to remember and import StateForConditionals.IgnoreConditionals every single time (which... do we even use that for \*anything\* else)

yairm210 (approved):
In any case, this did solve a problem we currently have, we can clean it up later

yairm210:
And yes, we use the ignore conditionals for a lot of things, it's super useful!

PR #10015, opened 2023-08-31
Add unique for a promotion to be free

SeventhM:


yairm210:
Please add AI logic to utilize free promotions, if any exist

SeventhM:
AI promotion logic in general is pure random (there's no actual logic there yet). You want me to just prioritize the free ones first?

yairm210:
That's exactly what I was assuming for
But since older is free then a second will be available, we should change the 'if promotion is available' to 'while promotion is available' so the AI will be able to select both on that turn

SeventhM:
There are situations where that doesn't quite work. Namely, if I make a promotion not be available if you get a certain free promotion (such as providing choice of potential upgrades, but only 1), it should rebuild the list of available promotions each time. Putting the while loop on the check of whether it can promote in the first place is better imo

yairm210 (approved):

Issue #10024, opened 2023-09-02
Defensive Pact Cancel Double Notification


tuvus:
**Version**
4.7.18

**Describe the bug**
When a civ declares war and has a defensive pact with another civ to cancel, the other civ will receive two separate notifications.

**Expected behavior**
The civ should only receive one notification from the canceled defensive pact.

PR #10025, opened 2023-09-02
Defensive pact notification fix

tuvus:
Solves #10024

This commit removes the double notification for a civ when someone else cancels their defensive pact.
This commit also adds a message to the player when declaring war that their defensive pacts will cancel and informs them of the Civs who will join the defenders in the war.


tuvus:
Can someone double-check that I added the translations correctly and that they will actually translate?

yairm210:
This will not translate, you need to wrap the individual text elements to translate in curly brackets
Otherwise it looks at the whole text and goes "I don't have that key in my translations dictionary, nothing to do here"

tuvus:
So instead of "\n An unknown civilization will also join them in the war", I should do "\n {An unknown civilization will also join them in the war}"?

If so, do I also need to put it around "Declare war on [${otherCiv.civName}]?" because I am now adding more text to it?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -784,8 +784,6 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
             }
             civInfo.addNotification("We have canceled our Defensive Pact with [${it.otherCivName}]!",
                 NotificationCategory.Diplomacy, NotificationIcon.Diplomacy, it.otherCivName)
-            it.otherCiv().addNotification("[${civInfo.civName}] has canceled our Defensive Pact with us!",
-                NotificationCategory.Diplomacy, civInfo.civName, NotificationIcon.Diplomacy)
```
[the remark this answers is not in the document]
tuvus: I decided to drop the "us" notification because it would require more translation and complexity. I would need to filter it out of the getCommonCivs(), and it's inconsistent because we don't use it for the other notifications.

Side note: Does anyone care that signing a DoF sends two notifications to each Civ? This tends to clutter the screen a bit.

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -765,25 +765,29 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+                val defensivePactOffer = diploManager.trades
+                    .firstOrNull { trade -> trade.ourOffers.any { offer -> offer.name == Constants.defensivePact } }
+                diploManager.trades.remove(defensivePactOffer)
+                val theirDefensivePactOffer = diploManager.otherCivDiplomacy().trades
+                    .firstOrNull { trade -> trade.ourOffers.any { offer -> offer.name == Constants.defensivePact } }
+                diploManager.otherCivDiplomacy().trades.remove(theirDefensivePactOffer)
+                diploManager.removeFlag(DiplomacyFlags.DefensivePact)
+                diploManager.otherCivDiplomacy().removeFlag(DiplomacyFlags.DefensivePact)
+                diploManager.diplomaticStatus = DiplomaticStatus.Peace
+                diploManager.otherCivDiplomacy().diplomaticStatus = DiplomaticStatus.Peace
```
yairm210, line 783: This whole thing is just double code for the same thing.
Either:
A. do `for (manager in listOf(diploManager, diploManager.otherCivDiplomacy())`
B. Create a function `invalidateDefensivePact(diploManager:DiplomacyManager)`

The goal is to increase clarity and ensure we don't miss anything for either side

tuvus: If that is fine, I will continue refactoring this code in a separate branch. I have another branch with defensive pact tests that are very helpful for testing for regression, but they haven't been pushed to the main branch yet.

core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
```diff
@@ -872,8 +872,37 @@ class DiplomacyScreen(
...
+            // Go through and find all of the defensive pact chains and add them to the list
+            for (i in 0 until  otherCivDefensivePactList.count())
+                otherCivDefensivePactList.addAll(otherCivDefensivePactList[i].diplomacy.values
+                    .filter { !otherCivDefensivePactList.contains(it.otherCiv()) 
+                        && it.otherCiv() != viewingCiv && it.otherCiv() != otherCiv
+                        && !it.otherCiv().isAtWarWith(viewingCiv) }
+                    .map { it.otherCiv() })
+            for (thirdDefensivePactCiv in otherCivDefensivePactList) {
+                defensivePactJoinText += if (viewingCiv.knows(thirdDefensivePactCiv))
+                    "\n [${thirdDefensivePactCiv.civName}] will also join them in the war"
```
yairm210, line 890: Reminder that these are not yet wrapped in curly brackets, this will not translate

tuvus: I'm still working on it. Do you have any idea on how I can do the loop that increases in size as it is looped through? Everything seems to either throw an error or shortcut the loop to not include newly added elements. A standard for (int i = 0; i < othercivDefensivePactList.count; i++) would work perfectly, but is this acceptable?

tuvus:
I think it should be translated now.

> Read `getDiplomaticVictoryVoteBreakdown` for an example.

This was really helpful and the style looks much cleaner than before.

PR #10041, opened 2023-09-03
Fix cities getting the resource list of other cities

### LINKED ISSUE #10038: "Provides [amount] [resource]" providing far too many resources
Windows 10, Itch.io version 4.8.0 and Android 13, Play Store version 4.7.19

When you build a building with the `"Provides [amount] [resource]"` unique, it gives too many resources. Specifically, instead of giving you that amount of resources from one city, it gives you that amount from all cities.

To reproduce:
1. Build at least two cities
2. Build a Recycling Center in just one city
3. Instead of providing 2 Aluminum, the center will provide 2 per city you own

In games like DeCiv where you have multiple buildings that provide a resource, the result is to produce hundreds or even thousands of excess strategic resource.

SeventhM:
Fixes #10038
Note: Because of this change, local-style beliefs can no longer provide resources at all. On one hand, this is fine as it would erroneously claim it came from a building and resources from follower beliefs applying to the cities themselves may or may not be confusing (there was a whole discussion on discord wondering how it actually worked). On the other, maybe beliefs should be counted separately as well

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -264,7 +264,7 @@ class City : IsPartOfGameInfoSerialization {
     }
 
     private fun getCityResourcesFromUniqueBuildings(cityResources: ResourceSupplyList, resourceModifer: HashMap<String, Float>) {
-        for (unique in getMatchingUniques(UniqueType.ProvidesResources, StateForConditionals(civ, this))) { // E.G "Provides [1] [Iron]"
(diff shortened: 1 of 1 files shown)


yairm210 (approved):

yairm210:
If we want to enable local style uniques, we can add to that city.getLocalMatchingUniques

yairm210:
Actually, we COULD hardcode that specific unique to be counted as a local unique! Interesting, not sure it's three best solution though

Issue #9980, opened 2023-08-27
Feature request: Disable constructions


hackedpassword:
I tire of fighting with my own cities not to build certain things, and sometimes they sneak it through anyway. Then for say uranium for example, hey where's that resource? It was here a minute ago? Someone prob built a reactor now I have to audit 15 cities and find the perpetrator. Can we denounce our own cities?? Let's do that.

So the proposed feature here is to long press the + (plus) on a construction, any unit, building, wonder, other, whatever. That item is then sent to another section like "Disabled", looking like any other section in construction, still with a plus like any other. If the player hits the plus on the disabled item, pops it back where it should be. Meanwhile, anything in the disabled section is not factored when a city is choosing constructions on its own.

![Screenshot_20230827_105856](https://github.com/yairm210/Unciv/assets/119975434/96884f7a-b63b-49f3-8e43-bf7cc1ff0874)

Skekdog:
You can disable auto city construction in settings

hackedpassword:
I don't want to disable construction, I want to disable certain construction.

hackedpassword:
Is that a hint? Ouch!

hackedpassword:
> 😈 just perpetrating your joke again
LOL you know I got you. 
I'll check that jar out later maybe tonight at best, pressing priorities this day. 

Implementation sounds tricky. I'm guessing then that the lists shown on the construction screen are generated with some inferred data structure magic rather than say a stored list of available construction. Makes sense, consequentially making more sense why a separate blacklist would have to be held. 

Is there a structure for save games, some local per installation config that can associate data to individual saves? Hmm that wouldn't be optimal either when uploading saves. 

I feel this idea may be proven technically impractical seeing the issues involved @SomeTroglodyte.

hackedpassword:
My review of your .jar - a new in-city menu that adds refreshing functionality!

1. I love it. This is a feature that belongs, flows naturally and provides expected gameplay UX.
2. Works as advertised* - I'm tempted to migrate to this .jar for regular games.
3. * As implemented, anything added to the disabled menu is global for all cities. That's a frustration.
4. Disabling all buildings and wonders leaving units enabled will ignore units for auto-build next turn.
5. See #1. Being able to bench a build is way better than I first imagined. It's good.

I hope others give it a try! I'm certain not everyone will feel the same but since you have to long-press/right-click the menu, there's no UI change otherwise.

hackedpassword:
[...]

PR #10061, opened 2023-09-05
Global Constructions Blacklist

SomeTroglodyte:
For #9980, I think there's a screenshot there?

Expands the city construction context menu by Disable/Enable. 'Disabled' adds the construction name to a global blacklist in settings, and henceforth automation for human players won't build them, and the available constructions offers parks them in a separate category expander - meaning they're not in the way but can still be built manually.


yairm210:
No pics

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -68,6 +68,7 @@ class GameSettings {
     var skin: String = Constants.defaultSkin
     var showTutorials: Boolean = true
     var autoAssignCityProduction: Boolean = true
+    var dontAutoAssignConstructions = HashSet<String>()
```
yairm210, line 71: disabledAutoAssignConstructions, perhaps?
"dontAutoAssignConstructions" sounds like a boolean, easily confused with above "autoAssignCityProduction"

core/src/com/unciv/ui/popups/CityScreenConstructionMenu.kt
```diff
@@ -100,6 +110,20 @@ class CityScreenConstructionMenu(
...
+
+    private fun canDisable() = constructionName !in dontAutoAssignConstructions &&
+        constructionName != PerpetualConstruction.idle.name
+    private fun disableEntry() {
+        dontAutoAssignConstructions.add(constructionName)
+        settings.save()
+    }
+
+    private fun canEnable() = constructionName in dontAutoAssignConstructions &&
+        constructionName != PerpetualConstruction.idle.name
```
yairm210, line 124: We have the construction itself, we can compare it to idle directly, not via names
Also idle should never be in the disabled constructions, and if somehow it IS then we want to be able to enable it, this check seems unnecessary

core/src/com/unciv/ui/components/input/KeyboardBinding.kt
```diff
@@ -136,6 +136,7 @@ enum class KeyboardBinding(
     BuildWonders(Category.CityScreen, "Buildable Wonders", 'w'),
     BuildNationalWonders(Category.CityScreen, "Buildable National Wonders", 'n'),
     BuildOther(Category.CityScreen, "Other Constructions", 'o'),
+    BuildDisabled(Category.CityScreen, "Disabled Constructions", KeyCharAndCode.ctrl('h')),
```
yairm210, line 139: Not translated

core/src/com/unciv/ui/popups/CityScreenConstructionMenu.kt
```diff
@@ -62,6 +68,10 @@ class CityScreenConstructionMenu(
             table.add(getButton("Add or move to the top in all cities", KeyboardBinding.AddConstructionAllTop, ::addAllQueuesTop)).row()
         if (canRemoveAllQueues())
             table.add(getButton("Remove from the queue in all cities", KeyboardBinding.RemoveConstructionAll, ::removeAllQueues)).row()
+        if (canDisable())
+            table.add(getButton("Disable", KeyboardBinding.BuildDisabled, ::disableEntry)).row()
+        if (canEnable())
+            table.add(getButton("Enable", KeyboardBinding.BuildDisabled, ::enableEntry)).row()
```
yairm210, line 74: Not sure if these are translated, maybe they are?

core/src/com/unciv/ui/components/input/KeyboardBinding.kt
yairm210: Nifty

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
```diff
@@ -24,9 +25,14 @@ class ConstructionAutomation(val cityConstructions: CityConstructions){
     private val city = cityConstructions.city
     private val civInfo = city.civ
 
+    private val dontAutoAssignConstructions: Set<String> =
```
yairm210, line 28: Change this name as well

yairm210 (approved):

Issue #10070, opened 2023-09-06
Unimplemented AI Diplomacy


tuvus:
The AI does not currently engage in all diplomatic aspects of the game. This shows most notably that the player will never receive a diplomatic notification between the AI other than war and peace.

Here is a list of Diplomatic functions that the AI does not currently offer.
* Declaration of friendship (DoF)
* Mutual open borders agreement (except for a trade counter offer)
* Denouncing a civ
* Asking another civ to join war or offering to join a friend's war

The DoF logic was disabled in #1808 because the AI forced the other Civ to sign instead of offering it. Adding this functionality back  in is dangerous because the AI is likely to get stuck into an infinite friendship loop and never declare war.

Issue #4668, opened 2021-07-28
AI diplomacy ideas


jahodovykolac:
(More can come when I or somebody comes up with something)

- [x] Do AIs have relationships? It doesn't seem like it: A Civ spams me with nukes - nothing happens, nobody declares war. I use two nukes - I get -50 points which will probably result in war.
- [x] AIs could make trade offers with cities, when they're rich to expand or in situations like this:
![Screenshot_20210728-221515_UnCiv](https://user-images.githubusercontent.com/79529669/127389779-a3102925-4aea-4c3f-9592-619c23a2bab2.jpg)
(Here America or others would buy Smolensk and/or Yekaterinburg, or Russia would sell these - they're split away from mainland)
![Screenshot_20210728-221530_UnCiv](https://user-images.githubusercontent.com/79529669/127390119-a5ee690d-eec5-44e4-84bc-5a137b8d37d0.jpg)
(Here Siam would buy Smolensk and/or Orenburg to connect their cities)
- [ ] Allied Civs would gift units (to help strenghten me) instead of declaring war on me when I have really bad army.
- [ ] AI civs can declare friendship betwean each other (maybe they can but I haven't seen it for the whole time playing Unciv)
- [ ] AIs don't always reject asking for not settling near us (they always did to me - now I do the same for them hehe)
- [ ] AIs should tolerate doing something in self-defense, e.g. using nukes when the enemy just spams you with them.

logicminimal:
the diplomacy might be the least implemented feature in the game, time will fix this

yairm210:
A. Yes, AI have relationship levels between each other. Yes, it's affected by things like nuclear weapons.
B. Interesting idea, do PEOPLE usually buy/sell cities?
C. Would YOU gift a unit to a weaker civ you're allied with? I wouldn't. Therefore I won't make the AI do it.

logicminimal:
> A. Yes, AI have relationship levels between each other. Yes, it's affected by things like nuclear weapons.
> B. Interesting idea, do PEOPLE usually buy/sell cities?
> C. Would YOU gift a unit to a weaker civ you're allied with? I wouldn't. Therefore I won't make the AI do it.

out of context but how much of the diplomacy do you think si implemented in unciv compared to civ 5 yairm?

yairm210:
Definitely most of it.
It's not that Unciv is so behind, as much as it is that people expect more diplomacy options than appear in Civ V...

yairm210:
[...]

Issue #1808, opened 2020-01-30
Declaration of friendship


r3versi:
Declaration of friendship should be a request made by one civ and e eventually accepted by the other civ.

lyrjie:
I suggest we disable it asap completely until it's reworked, a lot of people are saying the game is unplayable in the current state

PR #10071, opened 2023-09-06
AI diplomatic actions rework

tuvus:
Here is the start of my rework of the unimplemented AI diplomatic actions posted in #10070.

This reactivates the offerDeclarationOfFriendship function and adds a new way of deciding if it should be offered or not. The new wantsToSignDeclarationOfFrienship function factors:
* Based on influence with the other Civ
* How many known Civs it currently has a DoF with
* How many know Civs there are vs. all the Civs in the game
* How many dead Civs there are
* How motivated they are to attack this Civ

The AI now offers mutual open borders. They do this to friendly nations that they aren't likely to attack and if the other nation is not at war with any of its friends.

I felt the need to combine these features together because they need to be balanced together, as stated in the issue, the AI might send too many DoFs if there are too many positive modifiers.


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -768,23 +770,78 @@ object NextTurnAutomation {
...
+            // Goes from 0 to -50 as more civs die
+            // this is meant to prevent the game from stalemating when a group of friends
+            // conquers all oposition
+            motivation -= deadCivs / allCivs * 50
+        }
+
+        // Wait to declare frienships until more civs
+        // Goes from -30 to 0 when we know 75% of allCivs
+        val civsToKnow = .75f * allAliveCivs
+        motivation -= min(((civsToKnow - knownCivs) / civsToKnow * 30f).toInt(), 0)
```
yairm210, line 812: This should *probably* be max(...,  0)
Instead of min/max, please use coerce, for readability - it tends to make these issues trivial

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -768,23 +770,78 @@ object NextTurnAutomation {
...
+            }
         }
     }
 
+    private fun wantsToSignDeclarationOfFrienship(civInfo: Civilization, otherCiv: Civilization): Boolean {
+        // Shortcut, if it is below favorable then don't consider it
+        if (civInfo.getDiplomacyManager(otherCiv).isRelationshipLevelLT(RelationshipLevel.Favorable)) return false
+        
+        val numOfFriends = civInfo.diplomacy.count { it.value.hasFlag(DiplomacyFlags.DeclarationOfFriendship) }
+        val knownCivs = civInfo.getKnownCivs().filter { it.isMajorCiv() && it.isAlive() }.count()
```
yairm210, line 790: replace .filter{...}.count() with .count{...}

tuvus:
Solves a part of issue #4668 that was not fully resolved.

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
tuvus: Yes, I think I changed it to be positive and then -= without changing the min.

tuvus:
TradeLogic.acceptTrade() processed treaties from both sides, even though it only applied to signing a research pact. signPeaceTreaty and signDefensivePact were both being called from both sides even though they handled both sides of the treaty. So to remove the double notifications, I made it only one sided.

core/src/com/unciv/logic/trade/TradeLogic.kt
```diff
@@ -155,8 +159,6 @@ class TradeLogic(val ourCivilization:Civilization, val otherCivilization: Civili
 
         for (offer in currentTrade.theirOffers.filter { it.type == TradeType.Treaty })
             transferTrade(otherCivilization, ourCivilization, offer)
-        for (offer in currentTrade.ourOffers.filter { it.type == TradeType.Treaty })
```
yairm210, line 158: ??? Evo is this removed?

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -816,19 +871,26 @@ object NextTurnAutomation {
...
+                tradeLogic.currentTrade.theirOffers.add(TradeOffer(Constants.defensivePact, TradeType.Treaty))
+    
+                otherCiv.tradeRequests.add(TradeRequest(civInfo.civName, tradeLogic.currentTrade.reverse()))
+            }
         }
     }
+    
+    fun wantsToSignDefensivePact(civInfo: Civilization, otherCiv: Civilization): Boolean {
+        if (civInfo.getDiplomacyManager(otherCiv).isRelationshipLevelLT(RelationshipLevel.Ally)) return false
+        val friendlyCivs = civInfo.getDiplomacyManager(otherCiv).getCommonKnownCivs().filter { it.getDiplomacyManager(civInfo).isRelationshipLevelGE(RelationshipLevel.Friend) }
```
yairm210, line 890: Please change this to "for civ in getcommonk owncivs, if a &b, return false"
This double iteration obfuscates what we're actually trying to find

yairm210:
This PR is trying to do *at least* 3 different unrelated things:
- Show diplomacy notifications to spectator - Approved
- Defensive pact AI changes - Tentatively approved
- Change how treaties are handled in trades - I'll need a detailed explanation for this, sounds convoluted and unnecessary 
Please split this into different PRs

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
tuvus: I wasn't sure if I wanted to add more conditions to the defensive pact later. But I think at some point we have to push it out and tweak it later.

tuvus:
Okay, I will move the trade change to a different PR. I think there is a lot of tweaking still to do regarding the DoFs, Defensive Pacts, and open borders. I mostly tested these with only AI because, actually, playing the game takes a very long time for every change.
If anyone has the time I would recommend that other people run it for themselves before merging it to the general public since it has a high chance of messing things up (as happened in the past with #1808).

This PR is not critical to any bugs in the game, so we can take our time and test it more if need be.

tuvus:
Looking through the source code from expansion 2, there doesn't seem to be much logic in the way of AI deciding on defensive pacts. It seems like an area they didn't value putting in the effort into developing it further (I agree with them), so I will go on my own way to develop this further. The same goes for DOF logic.

In general, I think Civ V diplomacy AI seems to be an area in which we should want to branch off. 

Unless anyone has any other recommendations, I think these changes are in a stable state and not game-breaking in any obvious way. So I think they are ready to be released. If anyone wants to test it, I would recommend playing on a medium map with around 8 AI as a spectator.

Issue #10055, opened 2023-09-05
Brackets within brackets causing problems (bracket in promotion name)


SpacedOutChicken:
I tried creating a nation with a `"[Military] units gain the [[Jaguar] ability] promotion"` unique, and it failed to start. Having those brackets inside the brackets is causing problems. Is there any way to fix this?

yairm210:
Sounds similar to the curly bracket nesting change we did way back

Issue #10054, opened 2023-09-05
"This Promotion is free" not accessible without excess XP


SpacedOutChicken:
Windows 10, Itch.io version 4.8.1

A promotion with `"This Promotion is free"` can't be selected unless you have enough XP to select a promotion. The free promotion can be selected without costing XP, so that's all good, but it only comes up as an option when you have XP to spend.

This mod contains a little free promotion, compatible with the base game. Since it's free, can it be made accessible at any time, even without enough XP for a regular promotion?
[FreePromotionTest.zip](https://github.com/yairm210/Unciv/files/12518191/FreePromotionTest.zip)

SeventhM:
I'mma pretend I didn't see this for now. That said imo
1. Premature optimization, but checking for that is likely non trivial. We would have to build a list of promotions each time instead of lazily checking if we have enough xp. This is especially true for players, rather than the AI. Again, probably premature optimization, but imo it is worth at least mentioning
2. I think we'd at least need a conditional for if a unit can promote/if a unit has x amount of xp in case this behavior is actually wanted

SeventhM:
Setting as "will not stale for now as a known issue, though, I do want to give my perspective here: I don't believe that a unique that doesn't cost xp/doesn't increase the price of later promotions should also inherently als o have the promotion not need xp to be gotten. These seems like different uniques that should exist. Perhaps the problem here is wording

One of these days I might get around to this... maybe. Thinking about it now, maybe this might be something that should be 3 uniques, not just 2. Idk

PR #10072, opened 2023-09-06
Unify placeholder parsing for unique parameters

yairm210:
Current code has 2 methods of parsing placeholder parameters (say that 3 times with peas in your mouth)
This PR unifies them so we always use the method that gives correct results for nested brackets, so placeholder params can contain other square brackets
Solves #10055


SeventhM:
> Solves #10054

I believe you mean #10055

yairm210:
Correct!

PR #10076, opened 2023-09-06
Universal nested brackets parser demo

SomeTroglodyte:
This could support arbitrarily nested [], {} and <> for `tr()`, possibly in one go... For now just a POC-level toy. Wrote it from the gut  without consulting that existing routine or stackoverflow&co, to ensure cleaner kotlin.

Note - the no-allow requirement is only met when using the `parse{}: Unit` overload, not when using the `parse(): List<ResultEntry>` one, as ResultEntries are not pooled. I do expect the emit-based use will be the more practical. Pooling the stack entries will only be worth it when one BracketsParser instance is hammered with tons of parse calls, and maybe not even then. Did not do the instrumentation or measurements, just wanted to see which pooling provider might be useable how easily. Since Gdx is already linked, why not. Supporting the Disposable interface is on the other hand almost certainly not worth it - either you keep using it or you let one go out of scope. Voluntarily releasing resources for an instance that won't go collectible immediately won't happen.

P.S.: I also ran the entire test suite shoehorning the new `String.getPlaceholderParameters()` in instead of the current one (still longer name) via `import as`, without otherwise altering the existing tests -> succeeds.

### DIFF
--- a/tests/src/com/unciv/logic/TranslationTests.kt
+++ b/tests/src/com/unciv/logic/TranslationTests.kt
@@ -4,6 +4,7 @@ package com.unciv.logic
 import com.badlogic.gdx.Gdx
 import com.unciv.Constants
 import com.unciv.UncivGame
+import com.unciv.models.metadata.BaseRuleset
 import com.unciv.models.metadata.GameSettings
 import com.unciv.models.ruleset.Ruleset
 import com.unciv.models.ruleset.RulesetCache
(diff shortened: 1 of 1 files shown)


yairm210:
Looks like you're having fun with this 😉

I'm fine with not understanding how it works, as long as the tests ensure it fits indeed work 😁

yairm210:
Okay, seems to me you've caught the 2 types of errors I can think of
- unclosed brackets
- invalid closing of other scope

I can see how recursive translation would apply here, I can see how nested unique modifiers would work, no problems I can see :)

Issue #10078, opened 2023-09-07
Politics screen does not show Defensive Pacts


tuvus:
I forgot to change the politics screen to show defensive pacts on the table and diagram.

AutumnPizazz:
Speaking of the defense agreement, I hope to add that during the declaration of friendship, both sides cannot declare war each other.

tuvus:
This has been talked about a lot before, but I don't think it will change. Is there a specific reason that you want this?
Some related issues are #6767 #3755.

AutumnPizazz:
My negligence. I thought of  Friendship  declaration in the context of Civ VI. This will add a little more interesting element to the PVP.

AutumnPizazz:
To put it in a negative way, perhaps we can add an option in the start screen to determine whether this game allows declaration of war during the friendship declaration period?

tuvus:
The AI currently almost never cancels a DoF (Aka not renew) without a DoW. We would have to write much more detailed code to combat this. Otherwise it is likely to have permenent peace. So I don't know how playable this would be. Especially with the AI offering DoFs soon.

AutumnPizazz:
I don't understand the code side of things, so my proposal may not be practical.
I think maybe it can be done by "only for human players"?

The-Talik:
Has this been fixed?  I don't see any way to initiate a defensive pact from the diplomacy screen.

tuvus:
This issue references the politics screen. In order to offer a defensive pact through the diplomacy screen you need to have a declaration of friendship and go into the trades screen and offer a defensive pact. So yes it has been fixed. 
Did that answer your question?

The-Talik:
Hm.  I definitely don't have a defensive pact option in the trades screen.  Is there any other reason why it would not show up?  I do have a signed declaration of friendship, and confirmed I don't already have a def pact with them.

tuvus:
Could you send some screen shots or the game file?

The-Talik:
![Screenshot_20240418-172600](https://github.com/yairm210/Unciv/assets/1664272/f5447602-91b7-4095-b70f-c0128540ad3f)
![Screenshot_20240418-172616](https://github.com/yairm210/Unciv/assets/1664272/0242af6c-b463-461c-a3e3-c0c746aff5bf)

tuvus:
What version are you on?

The-Talik:
4.11.4 (Build 987)  Google Play version.

tuvus:
[...]

PR #10079, opened 2023-09-07
Fixed politics tab not showing defensive pacts

tuvus:
Solves #10078

If a Civ has a defensive pact, it replaces the DoF on the politics table and is highlighted as Cyan. (I think I got the translation right this time.)
Defensive pacts and city-state ally relations now show as Cyan in the politics diagram as well.

Please request a different color if you think there is a more suitable one. I just think Cyan looks good.

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1402,6 +1402,7 @@ Politics =
 Show global politics = 
 Show diagram = 
 At war with [enemy] = 
+Defensive pact with [civName] = 
 Friends with [civName] = 
 an unknown civilization = 
 [numberOfTurns] Turns Left = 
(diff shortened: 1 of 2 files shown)


yairm210:
Please add before/after pictures for visual changes
Not just got this PR, but in general

tuvus:
Before
---
BeforePoliticsTable
BeforePoliticsDiagram
After
---
AfterPoliticsTable
AfterPoliticsDiagram

Every "Friends with" in the before picture also has a defensive pact which isn't shown.
(The AI is signing too many defensive pacts, which I think I need to fix.)

yairm210 (approved):

tuvus:
I personally can never tell who is allied with a city-state on the diagram when it is green. So, making it blue shows the one civ who is allied with it. Actually, now that I think about it, does the city-state use influence or their opinion? Each city-state also has a diplomacyManager with every other civ and hence an opinion. I don't think the diagram changes this for city-states.

PR #10086, opened 2023-09-08
test and refactor diplomacy manager

Framonti:
This PR is not affiliated to any GitHub issue

* Small refactor of DiplomacyManager: move magic numbers to dedicated constant file, simplify boolean expressions, extract helper methods to improve readability
* Tests for DiplomacyManager: add tests covering most base cases
* UnitMovement: simplify boolean expression

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -218,12 +218,10 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
             compareRelationshipLevel(level, 1)
     /** @see compareRelationshipLevel */
     fun isRelationshipLevelLE(level: RelationshipLevel) =
-            if (level == RelationshipLevel.Ally) true
-            else compareRelationshipLevel(level + 1, -1)
+        level == RelationshipLevel.Ally || compareRelationshipLevel(level + 1, -1)
     /** @see compareRelationshipLevel */
     fun isRelationshipLevelGE(level: RelationshipLevel) =
-            if (level == RelationshipLevel.Unforgivable) true
-            else compareRelationshipLevel(level + -1, 1)
+        level == RelationshipLevel.Unforgivable || compareRelationshipLevel(level + -1, 1)
 
     /** Careful: Cheap unless this is a CityState and influence is in 0 until 30,
      *  where the distinction Neutral/Afraid gets expensive.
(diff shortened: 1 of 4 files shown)


Framonti:
Hi, 
I'd like to ask a couple of questions 
1. Is there any reason for initializing the diplomaticStatus to War rather than Peace? `var diplomaticStatus = DiplomaticStatus.War`
2. DiplomacyManager manages diplomacy both for Major civs and City states. Is there any reason I'm not aware of (e.g., performance) for having a single class instead of using inheritance to split it?

From my naive understanding, the following structure could be better:
* Parent class DiplomacyManager, possibly abstract/interface. Implements all functions common to both major civ and city states
* Two children class, MajorCivDiplomacyManager and CityStateDiplomacyManager: they each implement/override their specific logic.

Benefits: smaller, more testable and maintainable classes. E.g., all logic regarding influence only on City state manager.

Thanks in advance for your answers

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -588,46 +577,71 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+            }
+
+            DiplomacyFlags.DefensivePact.name -> {
+                diplomaticStatus = DiplomaticStatus.Peace
+            }
+            // This is confusingly named - in fact, the civ that has the flag set is the MAJOR civ
+            DiplomacyFlags.ProvideMilitaryUnit.name -> {
+                // Do not unset the flag - they may return soon, and we'll continue from that point on
+                if (civInfo.cities.isEmpty() || otherCiv().cities.isEmpty())
+                    return true
```
yairm210, line 615: The name of this function didn't indicate what the return value is, so add a docstring on the function

yairm210: Actually, this check should be *ouside* this function.
This function should be "what do we do when flag expires", the question of "should this flag expire" is a different question which sucks not be dealt with in this function

core/src/com/unciv/logic/map/mapunit/UnitMovement.kt
```diff
@@ -741,12 +741,11 @@ class UnitMovement(val unit: MapUnit) {
...
 
     // Can a paratrooper land at this tile?
-    fun canParadropOn(destination: Tile): Boolean {
+    private fun canParadropOn(destination: Tile): Boolean {
         if (unit.cache.cannotMove) return false
         // Can only move to land tiles within range that are visible and not impassible
         // Based on some testing done in the base game
-        if (!destination.isLand || destination.isImpassible() || !unit.civ.viewableTiles.contains(destination)) return false
-        return true
+        return destination.isLand && !destination.isImpassible() && unit.civ.viewableTiles.contains(destination)
```
yairm210, line 748: The correct refactor would be to separate each false condition to a different if/else check, for debuggability

yairm210:
DiplomacyManager is a single class for deserialization

I have no problem with refactoring the functions for A vs B to be in different files, or classes (diplomacyCityStateFunctions) but the class needs to remain single

yairm210:
Also, these PRs are getting ridiculously long, please split this up to 2 PRs - one for testing functionality, which we'll merge first, and a second for the refactor, which we'll test according to the new tests.
This PR as-is does not guarantee 1:1 functionality with existing code - I trust you verified this, but regarding how we add these into the code, tests should come first

PR #10088, opened 2023-09-09
Defensive pact Tests

tuvus:
Part of #9900
Added tests to the defensive pact, which I made a while ago.


tuvus:
Whoops, it seems that since I used the same branch, it still keeps the old commits since they were squashed together.

PR #10091, opened 2023-09-09
Fix free units with a build limit not spawning

SeventhM:
I feel like this might be handling things incorrectly. Shouldn't we check if we're above a build limit in placeUnitNearTile instead of the unique? It's not like we even need to check in the unique since it can already handle if the unit doesn't spawn


yairm210 (approved):
This looks good, I feel this is a general question for unit placement but having this change is definitely better than not having it

yairm210:
Regarding the unit limits, we should probably add it in every place that creates a unit, for user clarity

For example, "transform to limited unit" should be disabled, "purchase/buildx should be disabled, upgrade should be disabled, etc
Otherwise it looks like a bug - "I clicked upgrade and it didn't work"

PR #10098, opened 2023-09-11
Update Italian.properties

Smashfanful:


yairm210:
Will conflict with #10097

TommasoPetrolito:
> Will conflict with #10097

I have changed #10097 (only edition on line 134 is still there so that no conflict should happen now)

PR #10101, opened 2023-09-11
Removed double notifications and processing of treaties when traded

tuvus:
Originally a part of #10071

Treaties are now only processed from one side of the trade. 

Due to the nature of a treaty they must be on each side of the trade. The previous code processed the treaties on both sides of the trade. However, DiplomacyManager.makePeace() calls makePeaceOneSide() on both sides, as well as signDefensivePact(). Calling makePeace() and signDefensivePact() twice did not cause any logical problems, but it does cause a double notification to appear, which is redundant and takes up notification space.

DoubleNotification

This change processes each treaty from only one side arbitrarily. The only difference with these is which Civ appears first on the notifications. The research agreement was the only thing that needed to be processed from both sides, which I fixed.

### DIFF
--- a/core/src/com/unciv/logic/trade/TradeLogic.kt
+++ b/core/src/com/unciv/logic/trade/TradeLogic.kt
@@ -157,6 +157,7 @@ class TradeLogic(val ourCivilization:Civilization, val otherCivilization: Civili
         for (offer in currentTrade.ourOffers.filterNot { it.type == TradeType.Treaty })
             transferTrade(ourCivilization, otherCivilization, offer)
 
+        // Transfter of treaties should only be done from one side to avoid double signing and notifying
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/trade/TradeLogic.kt
```diff
@@ -155,8 +159,6 @@ class TradeLogic(val ourCivilization:Civilization, val otherCivilization: Civili
 
         for (offer in currentTrade.theirOffers.filter { it.type == TradeType.Treaty })
```
yairm210, line 160: Please add a documentation note here, if I was reading this for the first time I would be confused why it's not mirrored

yairm210 (approved):

PR #10104, opened 2023-09-11
Japanese translation

s-ariga:
Hi yairm210/Unciv,
This is the first time PRing to Unciv.
I have translated some strings to Japanese.
Please consider merging them.

Thank you!


Ouaz:
2 typos that make fail the checks:

Line 1952 : `[terranFilter]` instead of `[terrainFilter]`

Line 2004 : `[buindingFilter]` instead of `[buildingFilter]`

Issue #10139, opened 2023-09-17
Returning a worker does not give notification


tuvus:
**Version**
4.8.4

**Describe the bug**
When another player returns your captured worker, you don't get a notification saying that your worker has been returned.

**To Reproduce**
1. Begin a two-player hotseat game on tiny map with barbarians and no AI
2. Build workers and have them get captured
3. Rescue the worker of the other Civ
4. The other Civ does not receive a notification about their rescued worker

**Expected behavior**
Players should receive a notification when their worker was rescued and returned by someone else.

PR #10140, opened 2023-09-17
Added return worker notification

tuvus:
Solves  #10139

This adds a notification to the Civ who receives their captured worker after it was rescued from another civ.
Two things that I would like to note are that the notification shows the tile where the worker was before it was captured. But what if that tile is outside of the viewing constraint of the Civ?
Also, I couldn't find any code for the AI choosing to return the worker.

Settler return


tuvus:
That's a great idea! There is no barbarian tutorial. We might want to add one. 
I also thought the loophole was weird. But I think instead of disabling it, it would be better to send the unit to the closest unit the Civ owns. Then we have better support for a Civ without a city for any mods.

Issue #10107, opened 2023-09-12
Free buildings stay free when their "Source" is gone


SomeTroglodyte:
**Version**
#10094 branch

**Describe the bug**
Conquering and pupetting the city with CN Tower will:
* Not affect already built Broadcast Towers for the conqueror
* Give other conqueror cities a free Broadcast Tower immediately
* Selling built Broadcast Towers will give you a free one _next turn_
* Victim cities will all keep their free Broadcast Towers, 

**To Reproduce**
1. Load attached game (yes heavily cheated to get the required situation)
2. Select one of the McInf's around Carthage and attack
3. Check Mecca: Non-free BT
4. Check Medina: Free BT
5. Next-turn
6. Check cities: Free BT

<details><summary>Saved game</summary>

```
[...]

PR #10142, opened 2023-09-17
Apply conditionals for free buildings to the destination city instead of the originating city

SeventhM:
What I was originally going to write:

Intended situation: You have a building with the unique `Gain a free [Building 1] [in all cities] <in cities without a [Building 2]>`
Currently: If the building "Building 2" exists in the originating city, this would not work. This will instead only fix itself at the start of the next turn after you calculate production
This PR: If the originating city has the building, it will give a "Building 1" to every city instead of the originating city. This is to cover a situation missed by the removal of "in all other cities", while providing more options to mod makers
_____
It turns out there's a few more issues with how we handle free buildings
1. Before the optimizations that lead to removing "in all other cities", this used `getLocalUniques()`. Switching this line to `getMatchingUniques()` makes no sense as later we already cover that with `city.civ.getMatchingUniques`. It should be switch to reading both the builtBuildingsUniqueMap and the uniques of the city's religion to keep consistency with old functionality
2. The prior point means that the second section, the one deleted in this initial PR, is redundant and non functional. `getMatchingUniques` already includes all of  `city.civ.getMatchingUniques`. The only benefit of that call is to label buildings that are already built as free. This initial PR copies that functionality into the first initial check, so it's kinda redundant now
3. Because of these changes, it seems to make more sense in my eyes if this is a function that civ.cache does, not cityConstructions. This... has consquences. As mentioned prior, we currently give free buildings immedaitely ***after*** handling production calculation. Handling free buildings by the civ instead basically either can only be done before we calculate production or after we also handle WLTK and resource demands
4. Unless I'm missing something, `freeBuildingsProvidedFromThisCity` seems poorly handled and barely functioning here. Assuming pre this PR, It claims to include a list of every building that this city is giving to other cities and more or less says "If I got a free building from elsewhere (including another city), I guess I created it". However, If I read correctly, this just doesn't function correctly with `CityConquestFunctions.removeBuildingsOnMoveToCiv()`. There it removes buildings from other cities that came from those other cities rather than removing buildings this city noted as coming from this city
5. As previously noted, we update free buildings at the start of a turn. This and when we get new buildings are the ***__only__*** times we update the free buildings. This is a problem if this unique exists on, say, a tech or policy. Between this, reassignPopulationDeferred, updateCivResources, and updateCitiesConnectedToCapital, I'm starting to think we should just have a `civ.cache.updateCities` that includes all of these functions that we commonly need to call after doing nearly anything

Leaving as a draft until I get a confirmation of how the best handle these issues


SeventhM:
Side note on point 4: If #10107 is supposed to \*not\* delete the building, the whole function of `CityConquestFunctions.removeBuildingsOnMoveToCiv()` seems questionable as it would actually be going against what is established as intentional behavior

SeventhM:
Current changes in the PR:
1. Adding free buildings was moved to CivConstructions
2. Adding free buildings now also function when we try to add other free buildings
3. Have the unique function as a triggerable, like the other free building uniques

Main problems left is
1. This continues the trend of the free buildings not functioning correctly (cities no longer state they are giving out a free building). This is maybe fine if #10107 is functioning as intended
2. Currently we're treating free buildings as both a triggerable and a passive unique. I'm not sure if we should

SeventhM:
No response in a while. This should be close enough to fine to be worth switching to be ready. Again, main discussion is on #10107. Otherwise, the main functional change is the move to having free buildings of this manner behave like other free buildings

Issue #7777, opened 2022-09-10
Feature request: auto open border


Thienanbad:
When a civ is destroyed and we liberate ít, it will auto open border so our army won't be kicked out and we don't have to spend few turns to make trade offer after liberated

Caballero-Arepa:
I agree

yairm210:
This is a redo of #2947 - since that one is old I'll close it

Thienanbad:
> This is a redo of #2947 - since that one is old I'll close it

Can you accept this featured, it is necessary when liberate a nation it help our units not to be transported to enemy land

yairm210:
I don't understand what you want
I didn't close this issue, so it's accepted
If you're asking if I'm planning on working on it, no, I have other priorities

Thienanbad:
Then when you planning to do it?

Vào 23:25, Th 5, 6 thg 10, 2022 Yair Morgenstern ***@***.***>
đã viết:

Thienanbad:
Oke man, I will wait. I hope you will do it soon. It's very important
mechanic for the game.

Vào 2:21, Th 6, 7 thg 10, 2022 Yair Morgenstern ***@***.***>
đã viết:

SeventhM:
Closed as Solved as per #10150

PR #10150, opened 2023-09-18
Liberate civ open borders fix

tuvus:
Solves #7777

This commit adds an open borders trade from the liberated Civ. The liberated Civ does not get open borders from the conquering Civ (to give the conquering civ more choice and power).

This change is not standard Civ V behavior based on online searching. It is, however, a quality-of-life improvement/game refinement.


tuvus:
Before
LiberateNoOpenBorders
After
LiberateOpenBorders

yairm210 (approved):

yairm210:
Ravignir says this is Civ V behavior

Issue #10152, opened 2023-09-19
Feature request: json schema for modders


yairm210:
Vscode can support json schemas: https://dev.to/brpaz/how-to-create-your-own-auto-completion-for-json-and-yaml-files-on-vs-code-with-the-help-of-json-schema-k1i 

As do jetbrains, in some of their products: https://www.jetbrains.com/help/idea/json.html

So the idea is to enable modder productivity using this tool.

And json schema can also contain documentation...

But let's do this in baby steps!
- Define basic "building json" schema (where do we save it? Docs/modders/schemas?)
- Instructions for how to set up the schema in vscode (free), and Android Studio (indicated as possible here: https://stackoverflow.com/questions/57408912/reset-json-schema-validation)
- See if we can generate that json schema from the class - not a niche request, there are probably libraries for that
- Definite other json file schemas - automated or not, based on previous step

Once we've done all that, then we can think about subsequent steps ;)

Caballero-Arepa:
Interesting

yairm210:
https://github.com/Kotlin/kotlinx.serialization/blob/1fe9ec1d07125c12a196d5a49959c400ea593cdf/examples/example-visitors/src/main/kotlin/App.kt

Looks like Kotlinx does the hard parts, just need to check if that works as-is or of we need to add yukky `@Serializable` annotations.
Even if so it's a small price to pay

yairm210:
https://github.com/Ricky12Awesome/json-schema-serialization

Looks like this adds support for other json schema limits like int and float ranges, etc. Neat stuff

PR #10158, opened 2023-09-21
Buildings json schema

yairm210:
POC for #10152 

image

Example for warning (nested - array of strings):
image

Honestly though, kind of underwhelming at the moment

Instructions for use:

- Add schema mapping
  - Search "json schema mappings" in doubletap
  - Click '+', direct schema to file at `docs/modding/schemas/buildings.json` for local file, or `https://raw.githubusercontent.com/yairm210/Unciv/9b63b329f112d6edb4280243498cc0e358e61d8d/docs/modding/schemas/buildings.json` for an example external URL (this is what modders will be doing)
  - Add a file path pattern: `*/Buildings.json`
 
image

- You can turn on 'show json schema' setting the same way to see what schema applies to a file, on bottom bar
image

Still doesn't manage to show me there's a bad field though:
image

even though I manually added `
    "additionalProperties": false` to the json schema, sounded like a jetbrains implementation issue - but then I see [this exists](https://youtrack.jetbrains.com/issue/WEB-51501) so now I don't know what to think.

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -7,7 +7,6 @@
     <inspection_tool class="JsonPathEvaluateUnknownKey" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JsonPathUnknownFunction" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JsonPathUnknownOperator" enabled="false" level="WARNING" enabled_by_default="false" />
-    <inspection_tool class="JsonSchemaCompliance" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JsonSchemaDeprecation" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
     <inspection_tool class="JsonSchemaRefReference" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JsonStandardCompliance" enabled="false" level="ERROR" enabled_by_default="false" />
--- a/android/assets/jsons/Civ V - Gods & Kings/Buildings.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Buildings.json
@@ -15,12 +15,16 @@
 	},
 	{
 		"name": "Monument",
+        "fds": 2,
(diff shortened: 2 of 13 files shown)


yairm210:
image

Hahaa

yairm210:
image

Hmmm that may be overkill

yairm210:
Managed to get the field thing working, but it requires manual intervention in the json schemal file, ideally the library would do this but it don't and it's not maintained... I don't think I want to be the maintainer of yet another repo at the moment, first let's implement what we can with what we got

Oh and I forgot to mention that the 'pure kotlinx' solution is simply not, for 3 years now, for what I can only describe as "reasons"
https://github.com/Kotlin/kotlinx.serialization/issues/1115

After digging in to the Jetbrains pipeline, it's... problematic, let's say. The biggest factor imo is the disconnect between the build pipelines in TeamCity and the actual PRs in Github, which are for some reason not linked. So their build tree shows that most of their projects have some failing pipelines in them, and yet no one cares.

I would assume that in a couple of years they'll get their act together and realize that all green is worth dropping PRs on, even if that means that the currenlty failing  pipelines will be entirely removed.
That's my feeling from looking at the kotlin compiler tests anyway.

yairm210:
Just for kicks:

We can even add a regex to the unique pattern to catch unclosed brackets

For example, unclosed `<>` catcher:

```json
      "uniques": {
        "type": "array",
        "items": {
          "type": "string",
            "pattern": "^[^><]*(\\<[^>]*\\>[^><]*)*$"
        }
      },
```

Not super informative:

image

but it does show!

yairm210:
After digging into this a bit more I understand that the autogenerated ones just won't cut it for other reasons.

For example: We have a lot of Stats in use, we can make that explicit and referenceable.
Same with civilopedia text.

The "this is a nullable" fields - modders shouldn't care about the internals, if it defaults to null or empty string, only if it's required or not.

 And then we can add handwritten definitions as well


So the experiment was a success, but my conclusion is that there is so much good here that the autogeneration doesn't cut it

yairm210:
Okay, VScode is the winner, it's really good at this!
https://yairm210.github.io/Unciv/Modders/Type-checking/ is go, this PR has served its purpose

PR #10167, opened 2023-09-22
Fix Nuke Notification

tuvus:
This commit does the following changes:
1. If the nuke doesn't land, the war declaration message back to the active player talks about an attempt instead the explosion
Note: I haven't put a notification to other Civs for this. We might want to change this.
2. Nuke explosion notifications occur above the notifications of the nuke-hit units.
3. Added nuke detonation notification to all known civilizations
4. Added nuke detonation notification from an unknown civilization. (Historically accurate)
5. Modified/rearranged symbols on the previous notification

According to my searches online, this is not the default Civ V behavior, but I think it is a needed change.
I previously posted on Discord to ask what others thought about it. I haven't taken the time to learn the unique system yet to make a "notifysOnAttack" unique.
https://discord.com/channels/586194543280390151/1144025738832527420

NukeNotificationOfWar
NukeFromUnknownCiv
NukeNotificationAboveAttacks


tuvus:
> "[nukeType] has been detonated from ..." might sound better with "by"
"[nukeType] exploded in our territory" - the third icon - is that the aggressor? Looks strange having an icon there not named in the message.

Both of these were done

>Or - what if we haven't met them yet and shouldn't know their identity?

Shouldn't we declare war and meet them first? That isn't the case right now, though.

>"A nuke has attacked ... " - often it was indirect, so "has hit" might sound better?

This message is a part of postBattleNotifications() and it doesn't seem like I should add an extra conditional just to nukes.

PR #10178, opened 2023-09-24
AI Open Borders Offer fix

### LINKED ISSUE #10177: AI offers Open Borders even when relevant tech is not researched
**Platform**
Android 13

**Version**
4.8.7 GitHub

**Describe the bug**
AI offers Open Borders even when Civil Service is not researched. In the saved game, Mongolia offers Open Borders, despite only being Classical Era (Civil Service is Medieval)

**To Reproduce**
Play a game against AIs on a low difficulty for a few turns

GitHub is crashing when I try to add save file, will hopefully add it in comment

tuvus:
Fixes  #10177

This commit does a few things.

1. Adds a technology requirement for the AI to offer open borders
2. Increases multiple Declined- flags from 10 to 20. AIs can process these offers without a problem, so reducing their Declined flags times relative to players might be helpful since it is partly a quality-of-life feature.
3. Increased the AI valuing not offering/accepting open borders if it is a suitable war target

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -859,7 +859,7 @@ object NextTurnAutomation {
         // Don't accept if they are at war with our friends, they might use our land to attack them
         if (civInfo.diplomacy.values.any { it.isRelationshipLevelGE(RelationshipLevel.Friend) && it.otherCiv().isAtWarWith(otherCiv)})
             return false
-        if (hasAtLeastMotivationToAttack(civInfo, otherCiv, civInfo.getDiplomacyManager(otherCiv).opinionOfOtherCiv().toInt()) > 0)
+        if (hasAtLeastMotivationToAttack(civInfo, otherCiv, (civInfo.getDiplomacyManager(otherCiv).opinionOfOtherCiv()/ 2 - 10).toInt()) >= 0)
             return false
         return true
     }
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -857,7 +859,7 @@ object NextTurnAutomation {
         // Don't accept if they are at war with our friends, they might use our land to attack them
         if (civInfo.diplomacy.values.any { it.isRelationshipLevelGE(RelationshipLevel.Friend) && it.otherCiv().isAtWarWith(otherCiv)})
             return false
-        if (hasAtLeastMotivationToAttack(civInfo, otherCiv, civInfo.getDiplomacyManager(otherCiv).opinionOfOtherCiv().toInt()) > 0)
+        if (hasAtLeastMotivationToAttack(civInfo, otherCiv, (civInfo.getDiplomacyManager(otherCiv).opinionOfOtherCiv()/ 2 - 10).toInt()) >= 0)
```
yairm210, line 862: How about moving the 10 to the right side of the equation, so it's clear which part is configurable?

tuvus: That sounds good on paper. But opinionOfOtherCiv shortcuts if it is guaranteed to be below the atLeast value. Therefore it would return early and not have the effect that I want.

yairm210 (approved):

Issue #10175, opened 2023-09-23
Feature request: Civilopedia sprites


hackedpassword:
So there I was, looking at some civilopedia page, when suddenly..

out of nowhere..

a realization appears! HP 20 ATT 8

I reach for my Blade of Reckoning - critical hit! Fountains of GP rain from the sky!

I'll let the images speak for themselves. :)

![Civilopedia_panzer](https://github.com/yairm210/Unciv/assets/119975434/803013cd-711f-4b0b-af7e-27a55cbbc4aa)

![Civilopedia_crossbowman](https://github.com/yairm210/Unciv/assets/119975434/f43767ed-76eb-48ce-8a8b-59d6b2ee654d)

![Civilopedia_GDR_small](https://github.com/yairm210/Unciv/assets/119975434/0d35d300-8f9d-4520-bc63-56a921dade2c)

![Civilopedia_GDR_full](https://github.com/yairm210/Unciv/assets/119975434/905af18a-f9d3-4857-aaec-852a0f9ef399)

![Civilopedia_GDR_designspec](https://github.com/yairm210/Unciv/assets/119975434/6e856d43-7334-40e9-a9d1-436b7718f009)

You get the idea. Level up.

Caballero-Arepa:
You can already do that, look at Alpha Frontier mod...

BUT, I absolutely agree that it should be BIG and to the side, not tiny and at the top

hackedpassword:
Alpha Frontier, always leading the pack. So much respect. @carriontrooper

hackedpassword:
It's not so much the aesthetics, it's the presentation.

The whole reason this post even exists is because I was looking in the civopedia, and found myself trying to correlate what whatever was to the game asset. There was a void here. A void that I have noticed since I first started playing this game but couldn't put my finger on.

This falls under important UI enhancements. I ask the idea be considered for its representative value. Going to follow up on AF's mini implementation in the intermediary.

Caballero-Arepa:
I mean, for FH to be that big, its not as pretty as remaining small, but for AU or HR, big is better.

But yeah, I understand.


What I would do if I had the power, would be to:

We have the side margins, let's say they are 10% each.

The remaining 80% of the horizontal space is divided 50% for text, 30% for the scaled image. No overlaping of the text over the image.

Custom made civilopediaText would still strech as it currently does and pushing down the image.

On top of that, I would add the possibility to place the custom text at the top or at the bottom.

[...]

PR #10187, opened 2023-09-26
Pedia pixel units

SomeTroglodyte:
Partial simplified #10175
See commit history
Screenshots contained in issue comments apply, except they didn't have the cropping code yet, so this now has better centering/padding than shown there.

Big Questions:
* Is that "crop off transparent borders" part too over the top? Too little gain for that much complexity?
* Where should that slider go? Which "Display" tab section? I chose "UI". Or Advanced tab?

Open ideas:
* Era/style variants - but how to design new extensions into the already loaded FormattedLine API. I had a click-to-cycle kludge working and removed it - it was just texture region cache lookups, no actual steering from BaseUnitDescriptions. But a slideshow using fade actions _would_ look cool...
* Moddable control over what civlopediaText goes above, what below the generated lines. Waiting for the mega-simple idea where to add that metadata.

@letstalkaboutdune - AbsoluteUnits look good enough with this, without your art I would have ditched the experiment. Only Gatling Gun isn't centered nicely - and gimp confirms some almost-transparent pixels on the left are not entirely transparent... Perfect proof that crop code works correctly 😸 
@carriontrooper - Thanks for part of the inspiration

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -748,6 +748,8 @@ Reset =
 Show zoom buttons in world screen = 
 Experimental Demographics scoreboard = 
 
+Size of Unitset art in Civilopedia = 
+
 ### Visual Hints subgroup
 
 Visual Hints = 
--- a/core/src/com/unciv/models/metadata/GameSettings.kt
+++ b/core/src/com/unciv/models/metadata/GameSettings.kt
@@ -134,6 +134,9 @@ class GameSettings {
     enum class NationPickerListMode { Icons, List }
     var nationPickerListMode = NationPickerListMode.List
 
+    /** Size of automatic display of UnitSet art in Civilopedia - 0 to disable */
+    var pediaUnitArtSize = 0f
+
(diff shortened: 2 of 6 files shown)


Ouaz:
Can I suggest to put the unit sprite at the end of the article?

Because if at the top:

1) a lot of space is used to display the sprite, moreover if set to a big size.

2) I think most players want a quick access to an unit data when opening an article in the civilopedia. Having to scroll a bit each time because there's the unit sprite frame at the top would be annoying after a while.

That's why I think to have it at the bottom makes more sense, centered (and the size can be set freely, without breaking the article layout), with or without the separators (like on the 3rd screenshot).

And, optional, maybe add "Current unitset sprite".

Can't do a mockup, I'm on the phone ^^

yairm210:
Whaaat why does the code analysis check stay even after I remove the Codacy connection on both ends, this sounds like a Github caching issue rather than a Codacy one
I'll merge this ignoring Codacy

yairm210 (approved):

PR #10191, opened 2023-09-27
Update Italian.properties

Smashfanful:


yairm210:
Placeholder `[nukeType]` not found in `Italian` for entry `After an attempted attack by our [nukeType], [civName] has declared war on us!`

PR #10224, opened 2023-10-03
fix golden age points computation to consider negative happiness

Framonti:
As discussed on Discord, this PR aligns the Golden age points calculation to Civ formula.


core/src/com/unciv/logic/civilization/managers/GoldenAgeManager.kt
```diff
@@ -58,7 +59,8 @@ class GoldenAgeManager : IsPartOfGameInfoSerialization {
     }
 
     fun endTurn(happiness: Int) {
-        if (happiness > 0 && !isGoldenAge()) storedHappiness += happiness
+        if (!isGoldenAge())
+            storedHappiness = max(0, storedHappiness + happiness)
```
yairm210, line 63: Instead of max use .coerceAtLeast(0)

PR #10233, opened 2023-10-04
Align ruleset icons in text to font metrics

SomeTroglodyte:
### Scope
This is about the ruleset icons that are made part of the font, but without taking the surronding font's typographic metrics into account. With defaults, they just seem slightly too big and slightly out of whack, depending on system.

### Approach
* Make metrics available cross-platform. Since understanding is vastly different between AWT and Android Paint, define our own.
      * All measurements are positive distances, not directions, as floats
      * Leading is declared to be part of the character bounding box and at the top
      * Character bounding boxes are meant to be placed contiguously
      * Ascent and Descent as positive numbers as per common understanding: Recommended / Average height of capital letters or descenders respectively.
      * Spacing below the descenders is optional and included in height
      * Baseline is at (leading + ascent) from the top of the bounding box
      * This is close to how AWT thinks, except AWT wilfully hides fractional parts that are present in the underlying sun classes - even reflection can't access those, and AWT has no provision for bottom spacing
      * Android measures everything as direction from the baseline down, so it will translate
* Size the actors to ascent - goal is to match a 'M' or 'X' or 'C' visually
* Make the box we draw in to get a Pixmap use metrics height and width proportional to actor width after above fitting
* Place actor between leading and baseline (per definition that's ascent height)
* Nudge down if it's a Portrait with border: Those lie about their dimensions, they paint their border outside their bounds
* Since now we need to scale the Actor, and ours - unlike a standard Gdx Image - can't be scaled by setting size - use isTransform
* Since isTransform is now on, delegate the pixel-copy upside down flipping to Gdx
* Treat normal character getGlyph and "symbol" font codepoints accordingly to the character bounding box defined above
* Add a little kerning to a ruleset icon when it's round and followed by the most open uppercase letters - really visible for Arabia
* Since the nation icon in the left floating button part of the topbar was done as auto-text-icon it would now scale with the font scale setting. I disliked that look therefore took the nation icon out of the text (second commit)
* That looked misaligned even though Scene2D-debug showed it to be perfectly aligned -> compensate label centering for descenders (fourth coimmit)
* Each icon-as-char get-Pixmap forced a reload of the entire atlas png it's in - dumb consumePixmap design - but I did a nondestructive read-only accessor for the pedia unit art -> reuse that as optimization

### Screenshots
<details><summary>Visuals</summary>

What it's about:
image

How this PR scales them with the font scale setting as clip (look bottom right behind the options popup):
https://github.com/yairm210/Unciv/assets/63000004/4103426e-90b2-4657-a6d6-bc6c258b1cd7

Or, the badly designed Jokerman font (paints way above its declared ascender size) at 1.5 scale:
Before
image
After
image

</details>


yairm210:
Each of these commits is its own PR, there is a LOT going on here, I'll need time to parse this
First comment - instead of hardcoded pixel values, use percentages of the font size (not 22 but size/2, not 5 but size/10, etc)

core/src/com/unciv/ui/components/Fonts.kt
```diff
@@ -116,12 +149,30 @@ class NativeBitmapFontData(
...
 
+            if (ch.code >= UNUSED_CHARACTER_CODES_START) {
+                // This is a Ruleset object icon - first avoid "glue"'ing them to the next char..
+                glyph.xadvance += (glyph.width / 27f).roundToInt()  // 2px for 55px width
+                if (assumeRoundIcon) {
+                    // Now, if we guessed it's round, do some kerning, only for the most conspicuous combos
+                    // Using a flat value is not perfect, but better than nothing.
+                    // Will look ugly for very unusual Fonts - should we check and do this only for default fonts?
+                    val kern = -(glyph.width / 18f).roundToInt()  // 3px for 55px width
+                    glyph.setKerning('A'.code, kern)
```
yairm210, line 169: This looks positively arcane

core/src/com/unciv/ui/components/Fonts.kt
```diff
@@ -273,10 +332,55 @@ object Fonts {
...
+        val actorWidth = actorHeight * actor.width / actor.height
+        val boxWidth = ceil(actorWidth).toInt()
+        // Nudge down by the border size if it's a Portrait having one, so the "core" sits on the baseline
+        val border = (actor as? Portrait)?.borderSize ?: 0f
+        // Scale to desired font dimensions - modifying the actor this way is OK as the decisions are
+        // the same each repetition, and size in the Group case or aspect ratio otherwise is preserved
+        if (actor is Group) {
+            // We can't just actor.setSize - a Group won't scale its children that way
+            actor.isTransform = true
+            val scale = actorWidth / actor.width
```
yairm210, line 368: We need to use other names - desiredActorWidth perhaps

core/src/com/unciv/ui/components/Fonts.kt
```diff
@@ -273,10 +332,55 @@ object Fonts {
...
+            actor.setScale(scale, -scale)
+            // Now the Actor is scaled, we need to position it at the baseline, Y from top of the box
+            actor.setPosition(0f, metrics.leading + metrics.ascent + border * scale + 1)
+        } else {
+            // Assume it's an Image obeying Actor size, but needing explicit Y flipping
+            // place actor from top of bounding box down
+            // (don't think the Gdx (Y is upwards) way - due to the transformMatrix below)
+            actor.setPosition(0f, metrics.leading + border)
+            actor.setSize(boxWidth.toFloat(), actorHeight)
```
yairm210, line 377: Surely actorWidth, not boxWidth?

core/src/com/unciv/ui/components/Fonts.kt
```diff
@@ -116,12 +149,30 @@ class NativeBitmapFontData(
             glyph.height = charPixmap.height
             glyph.xadvance = glyph.width
 
+            // Check alpha of two pixels to guess this is a round icon: transparent in the corner but not near center
+            val assumeRoundIcon = (charPixmap.getPixel(22, 22) and 255) > 0 && (charPixmap.getPixel(5, 5) and 255) == 0
```
yairm210, line 153: These are the hardcoded values I meant

yairm210:
Yes, definitely split into a separate FontIcons class

core/src/com/unciv/ui/components/Fonts.kt
```diff
@@ -273,10 +374,55 @@ object Fonts {
...
+        val boxHeight = ceil(metrics.height).toInt()
+        // Empiric slight size reduction - "correctly calculated" they just look a bit too big
+        val scaledActorHeight = metrics.ascent * 0.93f
+        val scaledActorWidth = scaledActorHeight * actor.width / actor.height
+        val boxWidth = ceil(scaledActorWidth).toInt()
+        // Nudge down by the border size if it's a Portrait having one, so the "core" sits on the baseline
+        val border = (actor as? Portrait)?.borderSize ?: 0f
+        // Scale to desired font dimensions - modifying the actor this way is OK as the decisions are
+        // the same each repetition, and size in the Group case or aspect ratio otherwise is preserved
+        if (actor is Group) {
```
yairm210, line 407: Split into subfunction, scaleAndPositionActor, with anything we can contain in that function

tests/src/com/unciv/dev/FasterUIDevelopment.kt
```diff
@@ -146,23 +147,23 @@ class FontDesktop : FontImplementation {
...
             width = height
         }
+
         val bi = BufferedImage(width, height, BufferedImage.TYPE_4BYTE_ABGR)
         val g = bi.createGraphics()
         g.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON)
         g.font = font
         g.color = java.awt.Color.WHITE
-        g.drawString(char.toString(), 0, metric.ascent)
+        g.drawString(char.toString(), 0, metric.leading + metric.ascent)
```
yairm210, line 165: Please make leading plus ascent function of metric, like distanceFromTopToBaseline()

This is hard enough to read as it is without having to hold a picture of the various font height parameters in your head

desktop/src/com/unciv/app/desktop/DesktopFont.kt
```diff
@@ -62,17 +63,20 @@ class DesktopFont : FontImplementation {
...
             width = height
         }
+
         val bi = BufferedImage(width, height, BufferedImage.TYPE_4BYTE_ABGR)
         val g = bi.createGraphics()
         g.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON)
         g.font = font
         g.color = Color.WHITE
-        g.drawString(char.toString(), 0, metric.ascent)
+        g.drawString(char.toString(), 0, metric.leading + metric.ascent)
```
yairm210, line 78: Reuse here

core/src/com/unciv/ui/components/Fonts.kt
```diff
@@ -273,10 +374,55 @@ object Fonts {
...
+     *  the total height as given by the font's metrics, and width scaled to maintain aspect ratio.
+     */
     fun getPixmapFromActor(actor: Actor): Pixmap {
+        // We want our - mostly circular - icon to match a typical large uppercase letter in height
+        // The drawing bounding box should have room, however for the font's leading and descent
+        val metrics = fontImplementation.getMetrics()
+        val boxHeight = ceil(metrics.height).toInt()
+        // Empiric slight size reduction - "correctly calculated" they just look a bit too big
+        val scaledActorHeight = metrics.ascent * 0.93f
+        val scaledActorWidth = scaledActorHeight * actor.width / actor.height
```
yairm210, line 401: Clearer as "actor.width * (scaledActorHeight / actor.height)"

core/src/com/unciv/ui/screens/worldscreen/topbar/WorldScreenTopBar.kt
```diff
@@ -190,20 +192,31 @@ class WorldScreenTopBar(internal val worldScreen: WorldScreen) : Table() {
...
         private val menuButton = ImageGetter.getImage("OtherIcons/MenuIcon")
 
         init {
+            // vertically align the Nation name by ascender height without descender:
+            //  Normal vertical centering uses the entire font height, but that looks off here because there's
+            //  few descenders in the typical Nation name. So we calculate an estimate of the descender height
+            //  in world coordinates (25 is the Label font size set below), then, since the cells themselves
+            //  have no default padding, we remove that much padding from the top of this entire Table, and
+            //  give the Label that much top padding in return. Approximated since we're ignoring 'leading'.
+            val descenderHeight = Fonts.fontImplementation.getMetrics().run { descent / height } * 25f
```
yairm210, line 208: Please add `val fontSize = 25` above, and use `* fontSize` here and `.setFontSize(fontSize)` below

yairm210 (approved):
Next step is definitely to split font icons functions from fonts.kt

PR #10242, opened 2023-10-04
Some religion stuff

### LINKED ISSUE #10235: Missionaries no longer behave as intended
Vanilla (likely irrelevant: Save created with 4.8.10 on Android, mostly default settings - my Modern Joke Religions mod is in there, but that one has only Religions and Beliefs, no changes to Units or anything else)
* A missionary no longer dissoves into a puff of smoke when their Faith is used up (their spread religion actions are exhausted).
* A missionary retains movement after spreading d... f... ahem, _Religion_ - minus one MP instead of stopping for the turn like before.
* When that "action costs 1 movement" is _intended_ for a mod unit, then the unit should not be marked as non-due, that is, auto-select next unit should stay with the unit.

SomeTroglodyte:
Uh, I shouldn't have simply left all of this in the (just barely renamed) 'Debug' branch, and could use more testing.

But - it's such a nice surprise I need to show it.

Closes #10235 

* Includes conditionals for the Missionary (and Prophet) to restore proper behaviour - suboptimal display for now
* Removes the Religious units from Vanilla (works a few turns - this is where I say could use more testing)
* Simplifies Civilopedia access to Tutorials and thereby enables hiding Tutorials selectively
* Hides Religion tutorials from Civilopedia - duh

The ease with which that simplification went is facepalm level. I suspect that stale 'Tutorials2' branch I got in my local git may have gone there too.... But I'm too much of a coward to check-out and re-read the diffs. I had that vision before the long pause and I even vaguely remember a working demo where the 'guidance' tutorials - both float-on-top and popup - had all the power of MarkupRender...

And that @xlenstra comment - I loved it and it pains me a little taking it out, but it will be preserved in git history. The 'signed by' style is a hommage...

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Units.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Units.json
@@ -1636,7 +1636,7 @@
 		"unitType": "Civilian",
 		"uniques": [
             "Can instantly construct a [Holy site] improvement <by consuming this unit> <if it hasn't used other actions yet>",
-            "Can Spread Religion <[4] times>",
+            "Can Spread Religion <for [2] movement> <[4] times> <after which this unit is consumed>",
 			"Removes other religions when spreading religion",
             "May found a religion <by consuming this unit> <if it hasn't used other actions yet>",
             "May enhance a religion <by consuming this unit>  <if it hasn't used other actions yet>",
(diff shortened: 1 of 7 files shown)


yairm210 (approved):

yairm210:
We will probably replace with a "for all movement" modifier once we do that but for now 👍

PR #10245, opened 2023-10-05
AI now uses free tech points

### LINKED ISSUE #10244: AI doesn't choose free tech
While working on the AutoPlay feature I realized that the free tech points that the AI get are never used. The functionality should come from specifically NextTurnAutomation.chooseTechToResearch but there isen't any logic for it there. 
I double checked, and the AI does use free policies and does choose the free great person from completing tradition.

I'll probably add the functionality of the AI choosing the most expensive tech as their free tech.

tuvus:
Fixes #10244

Added extra free tech loop. The AI chooses the most expensive tech first.

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -540,7 +540,7 @@ object NextTurnAutomation {
                 } else {
                     //Choose randomly between cheapest and second cheapest group
                     val techsAdvanced = costs[1]
-                    techsAdvanced.toCollection(cheapestTechs.toMutableList()).random()
+                    (cheapestTechs + techsAdvanced).random()
                 }
 
             civInfo.tech.techsToResearch.add(techToResearch.name)
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -514,6 +514,16 @@ object NextTurnAutomation {
     }
 
     private fun chooseTechToResearch(civInfo: Civilization) {
+        while(civInfo.tech.freeTechs > 0) {
+            val researchableTechs = civInfo.gameInfo.ruleset.technologies.values
+                .filter { civInfo.tech.canBeResearched(it.name) }
+            val techsGroups = researchableTechs.groupBy { it.cost }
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -514,6 +514,16 @@ object NextTurnAutomation {
     }
 
     private fun chooseTechToResearch(civInfo: Civilization) {
+        while(civInfo.tech.freeTechs > 0) {
+            val researchableTechs = civInfo.gameInfo.ruleset.technologies.values
```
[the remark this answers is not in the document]
tuvus: I changed the .toSortedMap to a list of lists since costs[0] causes an error because the key is the actual research value. The previous code used techsGroups[costs[1]]!!, which is the same as using regular integer indexing but needs a list of keys as well.

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -514,23 +514,33 @@ object NextTurnAutomation {
...
             //Do not consider advanced techs if only one tech left in cheapest group
             val techToResearch: Technology =
                 if (cheapestTechs.size == 1 || costs.size == 1) {
                     cheapestTechs.random()
                 } else {
                     //Choose randomly between cheapest and second cheapest group
-                    val techsAdvanced = techsGroups[costs[1]]!!
-                    (cheapestTechs + techsAdvanced).random()
+                    val techsAdvanced = costs[1]
+                    techsAdvanced.toCollection(cheapestTechs.toMutableList()).random()
```
yairm210, line 543: This is a little confusing as a way to add 2 lists together - if you can find a clearer way?
I think that "+" works on lists, so just (a + b).random()

tuvus: That's what was originally there but SomeTroglodyte recommended using toMutableSet() before I switched it to a list.
> Oh my - that `+` in existing code operates on Lists - never optimal...
I don't know enough about how Kotlin implements these collections to know which one is better. The collections will be really small, meaning performance isn't much of an issue, so I think the + operation looks better.

yairm210: 😓 
Personally I prefer readability where performance is not an issue, and I feel this is one of those cases, but I'm good with either, your choice

PR #10262, opened 2023-10-07
Refactor the "loading image" from Multiplayer into a reusable Widget

SomeTroglodyte:
... Yes I had the urge to reuse it, and it was the trigger to that package reorg

Thing is - can' test it completely. The Widget itself is tested in FasterUIDevelopment and just fine, the in-game one renders, but of course I get no events, so... Testers?

Implementation notes:
* Using the new kotlin time stuff, as I found extensive use of Java Duration and such a challenge must be answered. Still - needs that OptIn annotation, and Studio often is blind to it and complains. Repair IDE twice fixes that...
* Original had Gdx Actions mixed with Coroutines - 'nah', an Action can do delays on the fly. After some XP it boiled down to one really surprisingly simple line.


yairm210:
If you're going to genericize this, then get rid of references to multiplayer in the loading image

yairm210 (approved):

PR #10299, opened 2023-10-09
Fix auto assign production not working after a building is built

### LINKED ISSUE #10258: Auto-assign city production option no longer working
**Platform**
Android

**Version**
4.8.10 (Build 920) from Play Store

**Describe the bug**
The option doesn't appear to work anymore.  Even when the option is on, once city build queues are empty you get the Pick Construction prompt instead of a building or unit automatically being added to the Current construction.

**Additional context**
It still works occasionally, but I think only for your capital city?  I haven't tested exhaustively.

SeventhM:
Fixes #10258

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -524,8 +524,6 @@ class CityConstructions : IsPartOfGameInfoSerialization {
 
         updateUniques()
 
-        validateConstructionQueue()
-
         /** Support for [UniqueType.CreatesOneImprovement] */
         applyCreateOneImprovement(building)
 
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -387,6 +387,7 @@ class CityConstructions : IsPartOfGameInfoSerialization {
             if (construction.isBuildable(this))
                 constructionQueue.add(constructionName)
         }
+        chooseNextConstruction()
     }
 
     private fun validateInProgressConstructions() {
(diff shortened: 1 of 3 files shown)


SeventhM:
Huh, didn't expect errors there. Especially since I sorta just set it up in my dev environment and then played a full game for the menes

SeventhM:
Yeah, but I'm usually only on a computer for a small period of time, and I didn't expect that to trigger anything so I didn't actually check my notifications until later

SeventhM:
> I'd say addBuilding kicking off AI automation does smell anyway

I'm not sure I agree
1. Adding a building should 100% validate the queue. While other things might effect the via "Only Available" uniques, gaining a building, especially a building that adds other buildings, should confirm that the queue is correct
2. Anytime we clear the top item in the queue, we should have the AI choose a new building to build. We currently already do this naturally in the `removeFromQueue` function. What makes `validateConstructionQueue` special is it doesn't technically remove items from the queue. It instead nukes the queue, then rebuilds it. As such, It probably should have the AI choose a new building to build

It's... fine... if addBuilding doesn't directly validate the queue and thus start off the AI. But it means that anything that does add buildings needs to make sure to do both of these things, which unnecessarily means we're repeating the code anyways. Even worse, if we're running on the assumption that `validateConstructionQueue` should naturally restart the AI, then it doesn't matter if you remove validating from addBuilding. You still fail on any test that happens to run `constructionComplete`

> lets _almost_ all unit tests succeed again, but - is that best?

We're currently missing tests for next turn logic. If we made a test that stuff can be built across turns, then we would still need a fix for this, again because of `constructionComplete` removing the current construction from the queue if we finish building it. There's not much point in dancing around the issue

Side note, a more pressing concern may be more apparent in how a number of functions that do call `chooseNextConstruction` pass around a "automate" parameter. Perhaps the real problem is that we aren't explicitly checking if it makes sense to choose a new construction every time we validate the queue. But even then, I'm not sure that reasonably address our concerns either

SeventhM:
I guess that's fair. Let me double check if everything that adds buildings also validates, then. As stated however, the patch in OnlineMultiplayer is still necessary for when it will come up later

yairm210:
Please note I will not be available to release patches for this if something goes wrong, so please be on top of this
To release a patch version all you need to do is add a tag like '4.8.13-patch1' to a commit and everything else is automatic
You can also change the VERSION in UncivGame.kt if you want but that's only reflected to users so it's fine if you don't

PR #10311, opened 2023-10-12
Added five new uniques

PLynx01:
This is my first programming pull request (I made some translatins two years ago). It consists of three new uniques:

1. <when above [amount] gold>
2. <when below [amount] gold>
3. Remove [buildingFilter] in [cityFilter]

I believe that the 1 will be useful for in-game activities involving stockpiling gold and the 3 will be useful for mods adding "abstaract buildings" (that's how i call it). These can be empire-wide edicts and city-related decisions that can be revoked, unlike social policies.

Unit tests were passed
I also updated the documentation to cover the new uniques

### DIFF
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -581,6 +581,23 @@ class CityConstructions : IsPartOfGameInfoSerialization {
         updateUniques()
     }
 
+    fun removeBuildings(buildings: Set<Building>) {
+        val newBuildings = ArrayList<Building>(builtBuildingObjects.size - buildings.size)
+        builtBuildingObjects.asSequence().filter { it !in buildings }.toCollection(newBuildings)
+
+        builtBuildingObjects = newBuildings
+
+        val buildingNames = buildings.map {
+            it.name
+        }.toSet()
+
+        builtBuildings.removeIf {
+            it in buildingNames
+        }
+
+        updateUniques()
+    }
+
     fun updateUniques(onLoadGame:Boolean = false) {
         builtBuildingUniqueMap.clear()
         for (building in getBuiltBuildings())
--- a/core/src/com/unciv/logic/city/CityConstructions.kt
+++ b/core/src/com/unciv/logic/city/CityConstructions.kt
@@ -582,20 +582,11 @@ class CityConstructions : IsPartOfGameInfoSerialization {
     }
 
     fun removeBuildings(buildings: Set<Building>) {
-        val newBuildings = ArrayList<Building>(builtBuildingObjects.size - buildings.size)
-        builtBuildingObjects.asSequence().filter { it !in buildings }.toCollection(newBuildings)
-
-        builtBuildingObjects = newBuildings
-
-        val buildingNames = buildings.map {
-            it.name
-        }.toSet()
-
+        val buildingsToRemove = buildings.map { it.name }.toSet()
         builtBuildings.removeIf {
-            it in buildingNames
+            it in buildingsToRemove
         }
-
-        updateUniques()
+        setTransients()
     }
 
     fun updateUniques(onLoadGame:Boolean = false) {
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -130,6 +130,7 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     /// Building Maintenance
     GainFreeBuildings("Gain a free [buildingName] [cityFilter]", UniqueTarget.Global, UniqueTarget.Triggerable),
     BuildingMaintenance("[relativeAmount]% maintenance cost for buildings [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
+    RemoveBuilding("Remove [buildingFilter] in [cityFilter]", UniqueTarget.Global, UniqueTarget.Triggerable),
 
     /// Border growth
     BorderGrowthPercentage("[relativeAmount]% Culture cost of natural border growth [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -634,6 +634,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     ConditionalWhenAboveAmountResource("when above [amount] [resource]", UniqueTarget.Conditional),
     ConditionalWhenBelowAmountResource("when below [amount] [resource]", UniqueTarget.Conditional),
 
+    ConditionalWhenAboveAmountGold("when above [amount] gold", UniqueTarget.Conditional),
+    ConditionalWhenBelowAmountGold("when below [amount] gold", UniqueTarget.Conditional),
+
     /////// city conditionals
     ConditionalInThisCity("in this city", UniqueTarget.Conditional),
     ConditionalCityWithBuilding("in cities with a [buildingFilter]", UniqueTarget.Conditional),
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -635,8 +635,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     ConditionalWhenAboveAmountResource("when above [amount] [resource]", UniqueTarget.Conditional),
     ConditionalWhenBelowAmountResource("when below [amount] [resource]", UniqueTarget.Conditional),
 
-    ConditionalWhenAboveAmountGold("when above [amount] gold", UniqueTarget.Conditional),
-    ConditionalWhenBelowAmountGold("when below [amount] gold", UniqueTarget.Conditional),
+    // These two functions support only stockpileable resources (Gold, Faith, Culture, Science)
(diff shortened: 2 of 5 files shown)


SeventhM:
I might look at this more tomorrow, but immediate thoughts
1. `ConditionalWhenAboveAmountGold`: As stated, I 100% see this being requested to work with any stat, so it might as well be generalized as such. Probably most notable for the faith stat which often is kept for a while like gold
2. These sort of things (prices, targets and immediate stats) tend to be modified by game speed. I would consider doing that here
3. As Trog said, avoid removing stuff from the list your itterating. For Remove building here, I'm pretty sure we commonly actually get a list of buildings to remove first (with, say `getBuiltBuildings.filter{ it.matchesFilter() }`) and then remove them from there
4. I'm not sure if modders will or won't agree with `matchesFilter` here. For adding buildings, at least,  this sort of thing usually expects a building's exact name, not "all buildings of this category". Something to consider
5. If you are using `matchesFilter`, it seems more reasonable to have the text be "Remove [buildingFilter] buildings [cityFilter]" or something similar

PLynx01:
> Sounds like someone has fun learning. Good thing. Welcome!
> 
> Unordered thoughts:
> "I also updated the documentation to cover the new uniques" - I hope you didn't edit that by hand, that file is generated entirely automatically. The result of a generation run, in a typical setup, will then be offered by Studio as normal part of commits. But if you don't (include in commits) - some next release cycle will do it anyway.
> 
> `ConditionalWhenAboveAmountGold` and buddy look straightforward, not much that can go wrong. But - think of what modding can use as "currency" for buying units/buildings: not only Gold. So parameterizing the tested stat here might be more resilient for the future. Yup, will anybody want to condition-test on the "stored" happiness, the one used for golden ages? Or the stored culture, the 'account' you "buy" policies with? Doesn't sound likely. But still. If you want to go there, look at `Civilization.getStatReserve`.  Problem with that is getting the modders to understand what exactly they're testing (as dox are purely generated). Also, some might complain they want the conditional on a city-scoped Unique and want to test Production or Food - which have 'accounts' on the city level........
> 
> As for RemoveBuilding - I don't have all info on context, but some things...
> * Is its use clear enough to modders - that it's a one-time thing while 'provides building' isn't? Maybe it is...
> * Reason why Cite-States are exempt?
> * I don't like the extra code for "in this city" - but that's where I "don't have all info". :shrug: 
> * "Unit tests were passed" - sounds like a hint to actually build tests checking these new thingies :grin: 
> * loop + remove: Danger of concurrent modification exceptions, which are a bit of a bogeyman around here. Now there _is_ a provision in the standard libraries for exactly that combo, but will this code propery use it? Likely not. getBuiltBuildings is a Sequence, and I don't think their Iterator has that capability. If you're curious, there's likely tons of documentation - I myself would dig into library sources instead. Short: Modification while iterating needs to use the methods offered _by the iterator_ - and you don't have a reference to it with a simple for loop. ArrayList (as example) alone offers three iterator implementations -at least- and the ArrayList.iterator() method returns the one that offers a remove. If you call ArrayList.listIterator instead, you get a different iterator that supports replacing the element currently pointed at or inserting at the cursor, but no remove - and the spliterator is even more esoteric. [^1]
> 
> Short: Don't alter a collection you're iterating over without extra precautions. Not ever. And the extra precautions are - either use copies or those things mentioned above that are a bit mysterious and therefore less readable to your next coder buddy coming along...
> 
> Look for our withItem and withoutItem extension methods. They might apply or not, but even if not, helps to know why they exist and how they solve the task.
> 
> [^1]: As for esoteric: There actually is a `CowIterator` in Java, and it has **very much** to do with `ConcurrentModificationException`s. Moo!

Firstly, I didn't know that documentation is automatically edited. I will remember this for the next contributions.

Secondly, I think that only Gold and Faith can be freely stockpiled and spent at any time. I also believe that the first unique added by me would be useful for some kind of mod adding "Economic Victory", for example Accumulate 100000 gold to win the game.

I can include also city-states, it is easy to change that.

When it comes to iterating across city buildings, is there a better method that I should use? Give me a short code snippet if you can.

Remember that it is my first programming pull request, so the it's difficult not to make any mistakes.

PLynx01:
> All untested!

Could you give me some advice how to test the new functions?

PLynx01:
I also have another idea: An unique that will make a building disappear after being built. It is intended to be used alongside other uniques. As I mentioned earlier, it would be particularly useful in case of so-called abstract buildings. These can be:

- Edicts and decisions (both nation-wide and city-related, like these in Paradox strategy games)
- Commands that can be triggered from the city (defined by modders)

What do you think about it?

SeventhM:
> Secondly, I think that only Gold and Faith can be freely stockpiled and spent at any time

Sure, but it really doesn't hurt to allow it for any given stat. You never know what a modder comes up with as a mechanic

PLynx01:
> Sure, but it really doesn't hurt to allow it for any given stat. You never know what a modder comes up with as a mechanic

I added support for more stockpileable stats. Now this unique suppports

- Gold
- Faith
- Culture
- Science

TODO: Add a game speed-adjusted version.

PLynx01:
>     2. These sort of things (prices, targets and immediate stats) tend to be modified by game speed. I would consider doing that here

I added these uniques you mentioned @SeventhM

PLynx01:
@SomeTroglodyte 
What are the requirements to merge my PR into game?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -581,6 +581,23 @@ class CityConstructions : IsPartOfGameInfoSerialization {
         updateUniques()
     }
 
+    fun removeBuildings(buildings: Set<Building>) {
+        val newBuildings = ArrayList<Building>(builtBuildingObjects.size - buildings.size)
```
SeventhM, line 585: This assumes every building passed into this function is in the builtBuildingsObjects, which means if the city actually doesn't have the building, the empty ArrayList might be smaller than expected. Giving this a starting size seems unnecessary here (and is largely ignored), having it in the first place seems unnecessary and misleading

SeventhM:
> should we doubt that further

Honestly, that seems fine enough. If anything, I'd argue it's odd they don't both change in ta similar way (either both using a remove in place function or both reassign the variable). ~~Especially since it took me way too long to figure out what the collection line did~~


Note, the "in this city" cityFilter now doesn't work because of how it normally expects to filter it. That fix for it likely needs to be added back into the UniqueTriggerActivation code

core/src/com/unciv/logic/city/CityConstructions.kt
SeventhM: I guess. Even still, it assumes that it'll always be called in a situation where we know the buildings are actually there. None of the other lines in this function really care about that, though, so as baseline I almost feels like it should intuitively either just not bother with an initial capacity (consistent with how we handle it anyways. Premature optimization...) or actually check how many buildings that will be left over when we're done and set the initial capacity correctly

SeventhM:
> Not sure what you mean

"in this city" is treated the same as "in all cities". We filter it via classifying it as a local unique. Since this is done in UniqueTriggerActivation and not immediately from a `getMatchingUniques`, we do need to actually give a special case to it. The special casing was removed in https://github.com/yairm210/Unciv/pull/10311/commits/0f056c05d246f5a350ca660f8887c2ef0049dcb1

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -634,6 +635,14 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     ConditionalWhenAboveAmountResource("when above [amount] [resource]", UniqueTarget.Conditional),
     ConditionalWhenBelowAmountResource("when below [amount] [resource]", UniqueTarget.Conditional),
 
+    // These two conditionals support only stockpileable resources (Gold, Faith, Culture, Science)
+    ConditionalWhenAboveAmountStat("has more than [amount] [stat]", UniqueTarget.Conditional),
```
yairm210, line 639: This doesn't parse in an English sentence - please rename to 'when above', 'when below'

yairm210: Use the existing conditional for resource, checking if the param is a known resource. If yes, according to resource, if not, according to stat, if the unique has a stat

yairm210: The game speed modified version should take stat only, not resource

PLynx01:
> Use the existing conditional for resource, checking if the param is a known resource. If yes, according to resource, if not, according to stat, if the unique has a stat

@yairm210 Is this what you asked for?

[https://github.com/yairm210/Unciv/pull/10311/commits/f6e54d5933bc2988d1a999d3a4d0494c12e46be8](url)

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
PLynx01: I've made some changes requested by you. Can you review them?

yairm210:
Yes, looks good, please address final issues in the linked commit

PLynx01:
> Yes, looks good, please address final issues in the linked commit

Is everything ok?

PLynx01:
@yairm210 
Can you merge my PR now?

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -581,6 +581,23 @@ class CityConstructions : IsPartOfGameInfoSerialization {
         updateUniques()
     }
 
+    fun removeBuildings(buildings: Set<Building>) {
```
yairm210, line 584: This is too complex - filter out builtBuildings, and then call setTransients
Should be 3 lines of code total
- convert to set of names
- removeif
- setTransients

PLynx01:
@yairm210 
I've implemented your suggestions again. Is my PR ready to be merged?

yairm210 (approved):

Issue #10176, opened 2023-09-24
New Feature: AutoPlay


tuvus:
**Intro and relavence**
There has been a lot of talk of some some sort of auto next turn feature. The specifics range from a simple end turn without any movements to doing part or the entire turn. Basically, it will need a lot of customizability and options.

My interest is that I need a solution for quickly testing the AI so I have decided to work on implementing this feature.  I don't actually need the full feature but I have decided to do it anyway (Lets hope I can actually finish it).

**End Goal:**
The AutoPlay feature will allow the player do 4 main things:
- Automate certain parts of a turn individually, for example only military or civilian units so the player can focus on the important parts of their turn.
- Automate their turn but go out of automation when the player need handle some major event like tech, diplomacy, policy or religion.
- Completely automate their turn, using the full AI functionality.
- Automate their turn for X turns, while in multiplayer. A player's turn might be completely automated and skipped until they have some important notification, if they turn it on in the settings. The automation would be run on the other player's phone and if it reaches a condition that the automating player set it stops and sends the last save (discarding the automation for that turn). This might help speed up the early game or a stalemate late game.

Because of these use cases each player need to be able to have their own automation settings for their civ attached to the save file. The player might also want to have different settings per game.

<hr>

**Demonstration**
To demonstrate this, I wrote AutoPlayHack. This is a quick demonstration of how the AutoPlay would work.
https://github.com/tuvus/Unciv/tree/AutoPlayHack
<details>
Here is a quick options list. Note that if fullAI is pressed, the other buttons disappear.

![AutoPlayOptions](https://github.com/yairm210/Unciv/assets/7538725/531bd5a5-0c2e-4a0a-aa79-6f31d3694dfe)

AutoPlay button is right next to the next turn button. I have put a temporary symbol on it. Clicking the AutoPlay button or hovering over it while AutoPlay is active deactivates it. 

![AutoPlayButton](https://github.com/yairm210/Unciv/assets/7538725/48b43c81-98b8-46a5-a187-aa44793d917b)

Clicking on the AutoPlay button opens up a menu with the options shown. The Civ's Military, Civilian and Economy can be AutoPlayed independently here.

![AutoPlayPopup](https://github.com/yairm210/Unciv/assets/7538725/c14f2b43-48ee-4ca2-803d-8be977e61867)

</details>

I am working on this on a different branch.
The initial idea and way of handling calling nextTurn() comes from alexban011 in #9888.

Related issues:
#9888
#2375
#9450

[...]

PR #10315, opened 2023-10-13
Initial AutoPlay implementation

tuvus:
Part of  #10176

This pull request adds the basic AutoPlay features of auto-playing part of the turn, the entire turn and over multiple turns in single-player. This pull request is meant to be only the beginning of implementing the full AutoPlay features. 
Please read the Civilopedia AutoPlay picture to get more details.

Some things to note are: 
- After clicking any of the partial turn actions in the popup menu (such as AutoPlay Military), the nextTurn button shows as disabled even though it isn't really. Doing any sort of click will make it look back to normal again.
- The AutoPlay button is meant to be to between the multiplayer and nextTurn button but will appear on the end if the AutoPlay button is disabled then re-enabled in the options.

- **Why are there fewer options than originally planned?:** Adding conditional statements for if the AI should do some specific action (automate diplomacy, economy, etc.) based on the user AutoPlay settings greatly reduces readability. I felt like it would require a bit more work to organize it than I originally expected. There are a lot of random if(!isHuman()) clauses, and overlap with the regular player automation settings. So I decided to separate the more complex side features from the main feature.

- **Why is multiplayer multiturn support not implemented yet?:** Multiturn AutoPlay in multiplayer is a difficult concept, especially with Multiplayer API v2 being in progress. Adding in the multiplayer multiturn AutoPlay adds in more complexity which I think we should only commit to if the AutoPlay works out.

AutoplayCivilopedia
AutoPlayPartialOptions
AutoPlayButton
AutoPlayMenuePopup

</details> 

If anyone has any tweaks or recommendations, I am more than open to hearing them. Maybe we should rename AutoPlay to something else?
Once again, could someone check that I have the translations right? Is the Civilopedia auto-translated?


AdityaMH:
Viewing AI compete each other as spectator mode make me boring, and now bring feature make me bored even more.
Also if me using this feature, make me more harder to see reason why my worker disappear or my city being razed  by who.

*too lazy scrolling and reading 100+ notification in one minutes/turn.

yairm210:
I have a literal war going on so I don't have the headspace to review this now
Maybe @SeventhM and @SomeTroglodyte can help

tuvus:
Hmm, I am getting a lot of unresolved reference Errors, and I am not sure why. Did my Kotlin error-checking version update or something? My master branch succeeds at building/running but still has these errors.

Any quick tips would be appreciated. (and yes, I have tried googling it)

tuvus:
Closer. I tried deleting all my local files and re-downloading them but that also didn't work. And then I tried deleting and undoing MainMenuScreen (which was causing GameInfo to have an unresolved reference) and apparently, that worked. So then I re-merged an older AutoPlay branch with master. It's starting to look better but there still are a few more errors.

tuvus:
BAM! Now let's hope nothing else is screwed up.

The thing is, when I switched from my AutoPlay branch to the master the master got the error as well. So deleting it and re-adding it must have fixed git in some way.

yairm210 (approved):
I'm willing to give this a shot but I am kind of worried, sissy case we can always revert this

tuvus:
There will surely be some minor problems at least. The button is hidden behind a checkbox, so most people might not even notice it.

tuvus:
I also have an extra feature where starting the multiturn AutoPlay autosave so that it can be canceled and the previous turn can be replayed. The feature won't be included in the initial version since it needs more testing.
This feature is the beginning of adding conditional logic for if happens that the user specified they want to manually look over and play.

tuvus:
Does anyone know why UnitMovement.ky has `if (unit.civ.isHuman()) return listOf()`? I haven't looked into it very much yet but if anyone knows the reason why, it might require a change for AutoPlay.

tuvus:
Thanks! I think we can use the cache with AutoPlay. We will just have to clear the cache every time we start a partial AutoPlay.

> That's why I limited the caching to AI units, the AI will only control 1 unit at a time and once it's done with that unit it won't ever move the unit again for the turn
Oof, I had planned on violating this precondition in future updates. I'll have to be a little more careful then.

PR #10339, opened 2023-10-23
Add missing translations to Ukrainian

kostia1st:


yairm210:
Placeholder `[unit]` not found in `Ukrainian` for entry `An enemy [unit] has destroyed our tile improvement [improvement]`

PR #10348, opened 2023-10-24
Exclude GameSettings.json from Git

alberts8:
Fixed the path of the GameSettings.json file in the gitignore.


alberts8:
> settings should appear right there under android/assets

Ok then something on my side is causing the settings file to be in the root directory. I close this pull request and redo my development setup.

PR #10355, opened 2023-10-25
Tighten RulesetValidator for Terrain

SomeTroglodyte:
This adds a warning for the exploit documented in [HybridTileTech.md](https://github.com/hackedpassword/Z2/blob/main/HybridTileTech.md). Sorry, @hackedpassword, but at least you get a Warning only and a honorable mention in a source comment.

What happens with that exploit is - you get a tile where neither isLand nor isWater return true. Consequences for pathfinding *seem* nondestructive, but that unintentional un-planned-for situation is a time bomb - a regression trigger waiting to happen.

The idea in that Mod, however, should get a solution some day. "Bridges and Canals", if and when implemented in some form, may also solve some current problems - ships passing cities on hills take 2 movement for the step iirc, and so on. Something that allows ships and land units to share tile movement domains, both being able to cross without embarkation, if possible constructible (think panama canal), and optionally requiring a road to function as bridge? Something in a supported and unit-tested fashion.

### DIFF
--- a/core/src/com/unciv/models/ruleset/validation/RulesetValidator.kt
+++ b/core/src/com/unciv/models/ruleset/validation/RulesetValidator.kt
@@ -328,9 +328,23 @@ class RulesetValidator(val ruleset: Ruleset) {
         if (ruleset.terrains.values.none { it.type == TerrainType.Land && !it.impassable })
             lines += "No passable land terrains exist!"
         for (terrain in ruleset.terrains.values) {
-            for (baseTerrain in terrain.occursOn)
-                if (!ruleset.terrains.containsKey(baseTerrain))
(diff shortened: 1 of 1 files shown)


yairm210 (approved):

hackedpassword:
Well, let's at least set the record straight on what you're calling an exploit.

https://github.com/hackedpassword/Z2/blob/main/HybridTileTech.md

Not denying the technicality, but also concerned about this _feature_ being removed, which Z2 was born from and relies upon. See Z2 Terrains.json: Bridges, CoastTile, OceanTile

I can understand this tech makes you coders nervous, hoping you'll find a way to make it supportable.

yairm210:
!
Are you okay with elaborating a bit about your health status?

hackedpassword:
> current

I know exactly what you mean, and I have even more appreciation for your work.

PR #10361, opened 2023-10-26
Automated units stay automated after upgrade

### LINKED ISSUE #10323: Feature request: Automated units stay automated after upgrading
Currently when automated units auto-upgrade this breaks automation and you need to re-automate every unit that upgraded. This normally isn't much of a bother but in the late game for example when tons of units upgrade to mech infantry it's a bit of a pain to automate them all again, especially since you need to click more and then click automate, making it two clicks/presses for each unit. 

It would be nice to have an option to toggle "Automated Units Stay Automated After Upgrading" in the options menu so people could turn it on or off to their liking.

SomeTroglodyte:
... is this the way to close #10323 ?


yairm210 (approved):
Looks like it to me

Issue #10362, opened 2023-10-27
AI surrounds cities with cavalry/transports without attacking


tuvus:
Previously, while running AI-only simulations looking at their diplomacy, I have noticed that sometimes the state of the game reaches a stalemate. This tends to happen when a civ surrounds the last city of another civ with weaker units that don't want to attack the city. As a result, the weak civ survives, and every civ ends up declaring war on them, causing every other civ to become allies. This problem most likely comes from #9979.

This also occurs with military transports blocking naval units from attacking a city.

<details><summary>Edit: Cavalry not attacking city pictures</summary>

![CalvalryNoAttackCity1](https://github.com/yairm210/Unciv/assets/7538725/81bf84a8-82ae-4d57-922c-70804b700263)
![CalvalryNoAttackCity2](https://github.com/yairm210/Unciv/assets/7538725/0341811d-7ff6-4c19-a507-6d25dfba7a68)

</details> 

I have developed two separate changes that seem to help prevent this situation from occuring although a few more tweaks might be required.

PR #10364, opened 2023-10-27
AI units swap-retreat

tuvus:
Part of #10362

When an AI has a big group of units I often see that melee units tend to fortify in front of an enemy city when they are lower on hp since they can't move anywhere to heal. This commit addresses that issue by allowing those units to swap with the healthier melee units that are waiting on them.

Previously I implemented the MapUnit.getDistanceToClosestEnemyUnit() to prioritize updating units closer to enemy units first. Now I have changed this function to store its value in a cache. I also changed the unit update priority to include the unit's health, since more damaged units need to be swapped first.

The main addition is the tryRetreat() function. If a unit is below 50 health, it will check all tiles that the unit can move to and see if it should swap with a friendly unit. This function makes use of the cached getDistanceToClosestEnemy() value for a quick comparison.

tryRetreat() does not do anything with the non-swappable tiles yet. If we do want to include the non-swappable tiles, I would suggest changing getDistanceToClosestEnemy() to be assigned only to a tile instead of a unit.

One last thing to note, I changed canMoveTo() to also include a swap option and canUnitSwapToReachableTile() to use the canMoveTo() swap option so that it doesn't remove and place units back again. For context, I was getting some rare errors with the tryRetreat() function and this solved it.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -195,7 +195,7 @@ object UnitAutomation {
 
         if (tryHeadTowardsOurSiegedCity(unit)) return
 
-        if (unit.health < 50 && (tryRetreat(unit) || tryHealUnit(unit))) return // do nothing but heal
+        if (unit.health < 50 && (trySwapRetreat(unit) || tryHealUnit(unit))) return // do nothing but heal
 
         // if a embarked melee unit can land and attack next turn, do not attack from water.
         if (BattleHelper.tryDisembarkUnitToAttackPosition(unit)) return
@@ -254,7 +254,7 @@ object UnitAutomation {
         return true
     }
     
-    private fun tryRetreat(unit: MapUnit): Boolean {
+    private fun trySwapRetreat(unit: MapUnit): Boolean {
         if (!unit.civ.isAtWar()) return false
         // Precondition: This must be a military unit
         if (unit.isCivilian()) return false
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -267,7 +267,7 @@ object UnitAutomation {
         for (swapTile in swapableTiles) {
             val otherUnit = swapTile.militaryUnit!!
             if (otherUnit.health > 80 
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -253,6 +253,38 @@ object UnitAutomation {
         unit.movement.headTowards(encampmentToHeadTowards)
         return true
     }
+    
+    private fun tryRetreat(unit: MapUnit): Boolean {
```
yairm210, line 257: TrySwapRetreat, since this is only for swap

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -253,6 +253,38 @@ object UnitAutomation {
...
+        if (unit.baseUnit.isAirUnit()) {
+            return false
+        }
+
+        val unitDistanceToTiles = unit.movement.getDistanceToTiles()
+        val swapableTiles = unitDistanceToTiles.keys.filter { it.militaryUnit != null && it.militaryUnit!!.owner == unit.owner}.reversed()
+        for (swapTile in swapableTiles) {
+            val otherUnit = swapTile.militaryUnit!!
+            if (otherUnit.health > 80 
+                && unit.getDistanceToEnemyUnit(6, false) < otherUnit.getDistanceToEnemyUnit(6,false)) {
```
yairm210, line 273: Unit distance is the same for all swappable tiles, put it outside the for loop

tuvus: How necessary is that since getDistanceToEnemyUnit caches the result?

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
tuvus: Okay, I only included swapping because it can be done without much of a performance impact. If we want to change it to include non-swap tiles, we should move getDistanceToClosestEnemy() to be associated with the tiles instead of the units.

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
yairm210: If that's the case, not necessary ☺️

PR #10372, opened 2023-10-28
Documentation to help modders avoid that big question mark in cities

SomeTroglodyte:


yairm210:
Counterpoint - we ourselves could copy the default city image so modders don't need to!

yairm210:
I say let's get rid of it, we can always resurrect it later if we find it has a purpose

PR #10376, opened 2023-10-28
No ruins undo

### LINKED ISSUE #10373: Ruin reward rollback
**Platform**
Windows, Android

**Version**
4.8.13 (Build 923)

**Describe the bug**
You can change the reward you receive from a ruin by Ctrl-Z unit's move.

**To Reproduce**
1. Find ruin
2. Get the opportunity to roll back the movement. (I don't know, it works weird, sometimes it appears, sometimes it doesn't)
3. Roll back move Unit on ruin.

I'm showing how it works in a multiplayer game. But this can be reproduced in single player as well.
Unciv_AjHEZCWTfU

UPD. Attached save
Save game 
[Bug_ruin.txt](https://github.com/yairm210/Unciv/files/13196795/Bug_ruin.txt)

SomeTroglodyte:
Closes #10373

But first I encapsulated the entire Undo thing, consistent interface (not a literal one though, just the fourth variant of kludge for the lack of partial classes in kotlin), we could plug in a N-level stack without any outside change.

### DIFF
--- a/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
@@ -2,6 +2,7 @@ package com.unciv.logic.map.mapunit
 
 import com.badlogic.gdx.math.Vector2
 import com.unciv.Constants
+import com.unciv.GUI
 import com.unciv.logic.IsPartOfGameInfoSerialization
 import com.unciv.logic.automation.unit.UnitAutomation
 import com.unciv.logic.battle.BattleUnitCapture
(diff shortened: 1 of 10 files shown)


yairm210 (approved):

core/src/com/unciv/logic/map/mapunit/MapUnit.kt
```diff
@@ -750,6 +751,7 @@ class MapUnit : IsPartOfGameInfoSerialization {
     private fun getAncientRuinBonus(tile: Tile) {
         tile.removeImprovement()
         civ.ruinsManager.selectNextRuinsReward(this)
+        GUI.clearUndoCheckpoints()
```
yairm210, line 754: Do we need this anymore with the other changes in place? I think we don't

yairm210:
I'm in favor of the UndoHander, I would rather forgo the blocking of the undo because it's unclear to the player why ("every unit move I can undo except for this one, BUG!") and the problem it solves is... not really solved, just made harder to do

PR #10391, opened 2023-10-30
Unit test for spurious placeholders in translations that are not present in their key

### LINKED ISSUE #10369: City names in field improvements are displayed incorrectly in the German translation
**Platform**
Linux Mint 21.2, 4.8.14, 64bit.

**Version**
From Flathub, Build 924

**Describe the bug**
When I create a field improvement, the menu shows "- \[city name](dir)".

**To Reproduce**
1. start a game
2. have the game in German
2. choose a worker and select Build Field Improvement
3. the "code" instead of the city name is displayed

**Expected behavior**
Instead of this code, the corresponding city name should be displayed and a "link" to the city.

**Screenshots**
grafik

**Additional context**
Happens in all multiplayer games
Tested in single player and multiplayer

Another problem that might have something to do with it: The icon for guilds is not displayed in German (as shown in the screenshot).

SomeTroglodyte:
## Will fail checks intentionally
... as this [does not fix all occurrences](https://github.com/yairm210/Unciv/issues/10369#issuecomment-1783817192) the test is built to flag.

Closes #10369 

So what do you think - ignore problem & discard, ignore bad checks, remove flagged translations, try to fix translations (e.g. using different typographical brackets where they seem to be meant as punctuation)...?

Note: That test container class has a lot of code duplication, I didn't even try to reduce that, so the new test is just a minimally altered copypasta of the other one.

### DIFF
--- a/tests/src/com/unciv/logic/TranslationTests.kt
+++ b/tests/src/com/unciv/logic/TranslationTests.kt
@@ -18,12 +18,12 @@ import com.unciv.models.translations.tr
 import com.unciv.testing.GdxTestRunner
 import com.unciv.utils.Log
 import com.unciv.utils.debug
+import java.io.OutputStream
+import java.io.PrintStream
 import org.junit.Assert
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
-import java.io.OutputStream
-import java.io.PrintStream
 
 @RunWith(GdxTestRunner::class)
 class TranslationTests {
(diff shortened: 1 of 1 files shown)


yairm210:
This looks like a good thing, but I can't just merge a test that fails, because then I lose the ability to know if something has gone wrong elsewhere

SeventhM:
Perhaps make a separate issue for this, to make it more obvious what the problem is?

PR #10394, opened 2023-10-30
Allow modders to hide individual Uniques from Civilopedia

SomeTroglodyte:
Big diff, but little "core" change.
* The key functionality is the new UniqueType.ConditionalHideUniqueFromUsers, a function Unique.isHiddenToUsers centrally combining it with UniqueFlag.HiddenToUsers (which is exclusively under our control), and use thereof in central helpers. Only TranslationFileWriter is still allowed to query HiddenToUsers directly.
* The bulk is centralizing all unique display handling for getDescription and getCivilopediaTextLines into a set of helpers - which became needed after reviewing current code diversity and discovering some display code ignored HiddenToUsers
* As such, the unification fixes a few marginal bugs with negligible visual impact and negligible probability of activating

### Key question:
Some of the diversity found became helper parameters, to preserve existing behaviour as close as possible. That these parameters exist helps reviewing which differences exist where. But they also raise the question whether unifying behaviour could make them obsolete: Sort sometimes by json order sometimes alphabetically? Why? Could it be nicer or acceptable to color _all_ occurrences of ConsumesResources reddish instead of doing so only on Buildings and Units?

### DIFF
--- a/core/src/com/unciv/models/ruleset/Belief.kt
+++ b/core/src/com/unciv/models/ruleset/Belief.kt
@@ -2,9 +2,9 @@ package com.unciv.models.ruleset
 
 import com.unciv.Constants
 import com.unciv.UncivGame
-import com.unciv.models.ruleset.unique.UniqueFlag
 import com.unciv.models.ruleset.unique.UniqueTarget
 import com.unciv.models.translations.tr
+import com.unciv.ui.objectdescriptions.uniquesToCivilopediaTextLines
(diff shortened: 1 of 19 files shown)


yairm210:
Consumes resources always red - acceptable

yairm210:
Sort: tbh I don't really care, I think that alphabetically is generally good everywhere

yairm210 (approved):

PR #10418, opened 2023-11-03
Reorg SpecialistAllocationTable

SomeTroglodyte:
I remember that it already irked me last year that the +/- buttons for allocation Engineers always seemed to jump around behind my back... And I always had a little difficulty identifying which line they are.

I'll update this desc w/screens&all later  (...and fix one comment typo), need a break.


yairm210 (approved):
Pics please

PR #10422, opened 2023-11-03
Translation game on Rusyn language (rue)

KotDivuar:
Translation game on rusyn language. I am create a file in package of translations. Sorry, i`m new github-user, find this  for me was hard. Thank you


KotDivuar:
This is first version of translation, not everything is done yet, but it will be

KotDivuar:
> So, for an entirely new language we'll eventually need three more things - see #8801
> 
> TheLocale/Sorting collation thing may be a puzzle, the ISO 639-3 code 'rue' is established, but the Java libraries will return the default Collator for Locales instantiated with that language code. I just tested. So we would likely have to lie and feed it the ukrainian language to get proper cyrillic sorting?

Yes, i used "ґ" and " ї" in translate, we may use ukrainian code, if it need. Certainly possible

yairm210:
Placeholder `[civName]` not found in `Rusyn` for entry `Peace with [civName]?`
    Placeholder `[cityState]` not found in `Rusyn` for entry `[civName] is upset that you demanded tribute from [cityState], whom they have pledged to protect!`
    Placeholder `[promotion]` not found in `Rusyn` for entry `[unitFilter] units gain the [promotion] promotion`

KotDivuar:
Wow, need to fix this placeholder-problem?

PR #10424, opened 2023-11-03
Rivers... Moddable Stats and Civilopedia

### LINKED ISSUE #7596: Add river on civilopedia
**Please consider adding river and its description on civilopedia - terrain.**

All terrains are documented except river.

Yeah, they spawn between tiles, not on them.

But i think... it would be good to have it.

River also consists map.

And iirc we don't have explanation about movement cost of crossing river; expending all cost.

SomeTroglodyte:
Closes #7596
Sorry, several only loosely related features, see commit messages.

Gist: To make river tile stats moddable I had the idea to use a virtual TerrainFeature, similar to how the Repair improvement is a placeholder. It is never added anywhere in GameInfo, but called upon to get Stats. Works surprisingly well. Once it existed - use as place to explain rivers in Civilopedia. Works surprisingly well, too. Then it -of course- appeared in map editor, and to fix that I introduced another Unique allowing to remove some other hardcoding there... And then I couldn't resist doing a Civilopedia link on the current brush display, I wanted to when I redid the map editor but back then I believe there was no screen stack yet.

<details><summary>Screenshots (not final words)</summary>

image

That entry/header image is moddable as "Grassland+River":
image
</details>

#### Before merging
Triple-check my poor wording! Here's the new text for convenience:
<details>

#### Lakes
Lakes provide fresh water to shore tiles, changing yields depending on technological progress and allowing farming where it would otherwise not be possible (similar to Rivers and Oases).
#### Oasis
Oases provide fresh water to adjacent tiles, changing yields depending on technological progress and allowing farming where it would otherwise not be possible (similar to Rivers and Lakes).
#### River
Rivers exist on tile edges, not as terrain feature per se.
Nevertheless, the tiles on both sides of a river edge benefit, but these benefits do not stack.
One benefit is additional Gold yield, the other is that the tile has access to fresh water, changing yields depending on technological progress and allowing farming where it would otherwise not be possible (similar to Oases and Lakes).
Additionally, some buildings can only be built on river tiles, and they might have significance for certain religions.
Movement across rivers takes all remaining movement points of a unit unless there is a bridge.
When being attacked from across a river, a defender usually gets a significant bonus.

</details>

#### Open stuff
* Decide - should the virtual Terrain be called "River" or "Rivers"?
* Decide - should that virtual Terrain object be cached on a higher level than each Tile?
* Worth it to cache the actual Stats if the virtual Terrain's Stats can be determined to be static and not condition-bound?
* Document the modding option to use a custom pedia icon for the River entry

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Nations.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Nations.json
@@ -1,5 +1,6 @@
 [
 	//Spectator
+    //Not "Excluded from map editor" as it serves as placeholder for "Any Civ" starting locations
 	{
 		"name": "Spectator",
 		"outerColor": [255,255,255],
@@ -1348,7 +1349,7 @@
 		"outerColor": [0, 0, 0],
 		"innerColor": [211,180,113],
 		"cities": ["Dublin"],
-		"uniques": ["Will not be displayed in Civilopedia", "Will not be chosen for new games"]
+		"uniques": ["Will not be displayed in Civilopedia", "Will not be chosen for new games", "Excluded from map editor"]
 	},
 	{
 		"name": "Edinburgh",
@@ -1361,7 +1362,7 @@
 		"outerColor": [0, 0, 0],
 		"innerColor": [0,102,102],
 		"cities": ["Edinburgh"],
-		"uniques": ["Will not be displayed in Civilopedia", "Will not be chosen for new games"]
+		"uniques": ["Will not be displayed in Civilopedia", "Will not be chosen for new games", "Excluded from map editor"]
 	},
 	{
 		"name": "M'Banza-Kongo",
@@ -1444,7 +1445,7 @@
 		"name": "Barbarians",
 		"outerColor": [0,0,0],
         "innerColor": [185,12,12],
-		"uniques": ["Can only heal by pillaging"]
+		"uniques": ["Can only heal by pillaging", "Excluded from map editor"]
 	}
 
 ]
--- a/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
@@ -113,7 +113,13 @@
 		"RGB": [0, 171, 169],
 		"uniques": ["Fresh water",
 					"Considered [Food] when determining start locations",
-					"Considered [Desirable] when determining start locations"]
+					"Considered [Desirable] when determining start locations"],
+        "civilopediaText": [
+            {"text": "Lakes provide fresh water to shore tiles, changing yields depending on technological progress and allowing farming where it would otherwise not be possible (similar to Rivers and Oases)."},
+            {"text": "Rivers", "link":"Terrain/River"},
+            {"text": "Oasis", "link":"Terrain/Oasis"},
+            {"text": "Farm", "link":"Improvement/Farm"},
+        ]
 	},
 	{
 		"name": "Mountain",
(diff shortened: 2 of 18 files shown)


hackedpassword:
The other day I actually deleted a sidenote comment for this exact feature proposal. Yes!

> Decide - should the virtual Terrain be called "River" or "Rivers"?

I'd recommend singular tense. If a river touches a tile on one edge then continues away, that tile only has a 'river', or, if it one river path touching any number of tile edges, or, even if it were two paths touching the same tile, each is still a 'river'.

SeventhM:
> Lakes provide fresh water to shore tiles

Shouldn't this have the same wording as oasis and Rivers? After all, the fresh water mechanic is defined by the "Fresh Water" unique, not by any mechanic in specific. Speaking of... I kinda feel like this river should have the fresh water unique. Maybe a better wording is "[] provide fresh water to adjacent land tiles"

> Nevertheless, the tiles on both sides of a river edge benefit,

"Nevertheless" seems unnecessary here imo. Could just be me, though. Something like "Tiles on both sides gain its benefits. These benefits do not stack."


> changing yields depending on technological progress...

> Additionally, some buildings can only be built on river tiles...

> One benefit is additional Gold yield...

Seems unnecessary. If a building says it needs fresh water, I think it'll be obvious enough water that means. More notably, mentioning the gold benefit in text might trip up mod devs who copy-paste the terrain and change the benefit but don't read the civilopediaText

> a defender usually gets a significant bonus.

Maybe say what the exact bonus is here?



> virtual Terrain be called "River" or "Rivers"

"River". Consistent with existing mods, reads better in uniques ("... from every [Rivers] tile" is plural seemingly out of nowhere)



> Worth it to cache the actual Stats if the virtual Terrain's Stats can be determined to be static and not condition-bound?

I want to say yes, but then you will have to check uniques anyways regardless if it's cached. So I'm not sure there would be a point in doing so unless we can show it's not premature optimization

yairm210:
Definitely "River" as singular

SeventhM:
> and won't support _any_ UniqueType (yet?) except Stats

Out of scope, maybe, but I 100% expect other modders will request other terrain uniques works with rivers. Kinda the natural evolution imo for getting rivers to be modable. So probably the Tile.allTerrain approach is necessary (I would need to double check to see if that's the right approach though). ~~Also I need to fix an aspect about NullifyYields, that said~~ It's probably not that big a deal for it to layer on top of everything else in terrains. I'd need to double check, though. That could get complicated


> would get a hint on what further effects exist

> I for one trust our modders pretty far.

Sure, I guess. But from a modder perspective it's still a bit of extra work for modders that doesn't feel necessary imo. It says what uniques it gives already, it gets auto linked to, and the fresh water unique does a lot of heavy lifting. I can't see a mod really needing the extra information conveyed necessarily. And that's not even discussing BNW fountain of youth where you'd need to repeat the blurb again (since it counts as Fresh Water). Saying it gives fresh water and it's stats~~/uniques~~ to nearby tiles just seems like good enough imo

> Changing the name of the virtual placeholder could be done without affecting the tileFilters

Sure, but then that affects auto links and it might give a false impression to modders. From a modder perspective, it seems better to pretend it's an actual terrain to keep functionality with everything else 

> Also, wondering about the "M"

Uhh, it's the "Seventh" that's out of place here, actually. I typically go with the username "Mcompany", and brand my stuff with the letter M. I just am too lazy to figure out the email/password to my other GitHub account and use a "seven" branding for secondaries

hackedpassword:
> That River is not a real river, never exists on the map, and won't support any UniqueType (yet?) except Stats - how to bring that across to modders?

A commented object in Terrains.json (both vanilla and G&K) showing the current limits of river modability (sp?).

Since River is different than a tile, as I interpret it, no simple feat to merge into uniques. Would it be useful to partition river unique modifiers by tagging a handling logic like:

"[+1 Food] in [Grassland] tiles \<as a river\>",  // freestyled this

.. then the tile handler would absorb external modifiers passed to it for each touching tile and move on as if uniques as applied to that tile were part of its uniques processing. Or something like that.

SeventhM:
I'm almost willing to say the extra attempts to say "(similar to...)" is unnecessary from the same point of view of not listing the gold increase (btw, something I was just reminded of: did you know the gold was removed in bnw?). Otherwise, for now, unless someone else has something to say, looks fine to me

SeventhM (approved):

PR #10428, opened 2023-11-04
Improved settler AI

tuvus:
How often have you seen the AI settle their cities one tile away from the coast or river? I have, and I decided to make a change to it. What I didn't expect was the CityLocationTileRanker code to search for _every_ tile in a distance of 6 of _every_ city just to add them to a list to exclude from ranking. So, I decided that I had to rewrite the entire CityLocationTileRanker class to fix this and its other inefficiencies.

AISettleNotOnCoast

</details> 

Changes:
- AI now settles closer to rivers and coasts.
- AI now ranks tiles to settle not only by their stats but also all the resource types, freshwater etc.
- AI also ranks tiles to settle based on how close it is to the city, how big the city is and if it is their own city.
- The AI no longer settles the first settler in place. It might, after all, be one tile from a coast. Instead, it tries to settle within three turns.
- The settle location highlights now account for every tile even if the player can't see them or their resources to keep them consistent with the AI. If we are going to give the players a hint, it should be a good one.
- Players should be freer to search for a better location to settle now since the AI will likely spend a few turns to search as well.

In conclusion, the new code should be more dynamic and logical and should also be easier to edit and change the weights.
I can confirm that I have done extensive testing.

Any thoughts?

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -18,7 +18,7 @@ object CityLocationTileRanker {
         else unit.civ.cities.minOf { it.getCenterTile().aerialDistanceTo(unit.getTile()) }
         val range = (8 - distanceFromHome).coerceIn(1, 5) // Restrict vision when far from home to avoid death marches
         val nearbyCities = unit.civ.gameInfo.getCities()
-            .filter { it.getCenterTile().aerialDistanceTo(unit.getTile()) <= 4 + range }
+            .filter { it.getCenterTile().aerialDistanceTo(unit.getTile()) <= 3 + range }
 
         val possibleCityLocations = unit.getTile().getTilesInDistance(range)
             .filter { canSettleTile(it, unit.civ, nearbyCities) && (unit.currentTile == it || unit.movement.canMoveTo(it)) }
@@ -42,10 +42,10 @@ object CityLocationTileRanker {
         val modConstants = civ.gameInfo.ruleset.modOptions.constants
         // The AI is allowed to cheat and act like it knows the whole map.
         if (!(tile.isExplored(civ) || civ.isAI())) return false
-        if (!tile.isLand) return false
+        if (!tile.isLand || tile.isImpassible()) return false
         if (!(tile.getOwner() == null || tile.getOwner() == civ)) return false
         if (nearbyCities.any {
-                it.getCenterTile().aerialDistanceTo(tile) <
+                it.getCenterTile().aerialDistanceTo(tile) <=
                     if (tile.getContinent() == it.getCenterTile().getContinent()) modConstants.minimalCityDistance
                     else modConstants.minimalCityDistanceOnDifferentContinents
             }) return false
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -10,7 +10,10 @@ import com.unciv.models.ruleset.unique.LocalUniqueCache
 import com.unciv.models.ruleset.unique.UniqueType
 
 object CityLocationTileRanker {
-    fun getBestTilesToFoundCity(unit: MapUnit): Sequence<Pair<Tile, Float>> {
+    /**
+     * Returns a hashmap of tiles to their ranking plus the a the highest value tile and its value
+     */
+    fun getBestTilesToFoundCity(unit: MapUnit): Pair<HashMap<Tile, Float>, Pair<Tile?, Float>> {
         val distanceFromHome = if (unit.civ.cities.isEmpty()) 0
         else unit.civ.cities.minOf { it.getCenterTile().aerialDistanceTo(unit.getTile()) }
         val range = (8 - distanceFromHome).coerceIn(1, 5) // Restrict vision when far from home to avoid death marches
@@ -21,8 +24,18 @@ object CityLocationTileRanker {
             .filter { canSettleTile(it, unit.civ, nearbyCities) && (unit.currentTile == it || unit.movement.canMoveTo(it)) }
         val uniqueCache = LocalUniqueCache()
         val baseTileMap = HashMap<Tile, Float>()
-        return possibleCityLocations.map {
-            it to rankTileToSettle(it, unit.civ, nearbyCities, baseTileMap, uniqueCache)}.sortedByDescending { it.second }
+        val tileRankMap = HashMap<Tile,Float>()
+        var maxValueTile: Tile? = null
+        var maxValueRank: Float = 0f
+        for (tile in possibleCityLocations) {
+            val tileValue = rankTileToSettle(tile, unit.civ, nearbyCities, baseTileMap, uniqueCache)
+            if (tileValue > maxValueRank) {
+                maxValueTile = tile
+                maxValueRank = tileValue
+            }
+            tileRankMap[tile] = tileValue
+        }
+        return Pair(tileRankMap, Pair(maxValueTile, maxValueRank))
     }
 
     private fun canSettleTile(tile: Tile, civ: Civilization, nearbyCities: Sequence<City>): Boolean {
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -18,7 +18,7 @@ object CityLocationTileRanker {
             .filter { it.getCenterTile().aerialDistanceTo(unit.getTile()) > 3 + range }
 
         val possibleCityLocations = unit.getTile().getTilesInDistance(range)
-            .filter { canSettleTile(it, unit.civ, nearbyCities) && unit.movement.canMoveTo(it) }
+            .filter { canSettleTile(it, unit.civ, nearbyCities) && (unit.currentTile == it || unit.movement.canMoveTo(it)) }
         val uniqueCache = LocalUniqueCache()
         val baseTileMap = HashMap<Tile, Float>()
         return possibleCityLocations.map {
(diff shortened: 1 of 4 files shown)


yairm210:
AI *should* settle in-place if the map specifically places them there, that's why that code was there in the first place

"Near coast" is not always desirable, especially for the first city - more water tiles means less tiles that can generate production, meaning less building in the city overall, so there's a balancing act here.
If you already have cities nearby that are coastal, then this becomes even less important.

Players *should not* be given hints according to tiles they don't see! The reason we cheat with the AI is so that the AI has a slightly better chance, since overall human players are much better than the AI.

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -1,143 +1,141 @@
...
+        var maxValueRank: Float = 0f
+        for (tile in possibleCityLocations) {
+            val tileValue = rankTileToSettle(tile, unit.civ, nearbyCities, baseTileMap, uniqueCache)
+            if (tileValue > maxValueRank) {
+                maxValueTile = tile
+                maxValueRank = tileValue
             }
+            tileRankMap[tile] = tileValue
+        }
+        return Pair(tileRankMap, Pair(maxValueTile, maxValueRank))
```
yairm210, line 40: Sounds like we just need a localized data class, instead of a triplet? Would make things clearer on both ends for the next person reading this

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -1,143 +1,141 @@
...
-                    )
+    private fun canSettleTile(tile: Tile, civ: Civilization, nearbyCities: Sequence<City>): Boolean {
+        val modConstants = civ.gameInfo.ruleset.modOptions.constants
+        if (!tile.isLand || tile.isImpassible()) return false
+        if (tile.getOwner() != null && tile.getOwner() != civ) return false
+        if (nearbyCities.any {
+                it.getCenterTile().aerialDistanceTo(tile) <=
+                    if (tile.getContinent() == it.getCenterTile().getContinent()) modConstants.minimalCityDistance
+                    else modConstants.minimalCityDistanceOnDifferentContinents
+            }) return false
```
yairm210, line 51: Turn this into a for loop, please, then we can have the distance as a val and everything will be clearer

for city in nearby cities
distance = if () a else b
if distancetotile < distance return false

core/src/com/unciv/ui/screens/worldscreen/WorldMapHolder.kt
```diff
@@ -712,13 +712,11 @@ class WorldMapHolder(
...
-                && UncivGame.Current.settings.showSettlersSuggestedCityLocations
-        ) {
-            CityLocationTileRanker.getBestTilesToFoundCity(unit).map { it.first }
-                .filter { it.isExplored(unit.civ) }.take(3).forEach {
-                    tileGroups[it]!!.layerOverlay.showGoodCityLocationIndicator()
-                }
-
+                && UncivGame.Current.settings.showSettlersSuggestedCityLocations) {
+            CityLocationTileRanker.getBestTilesToFoundCity(unit).first.asSequence()
+                .sortedByDescending { it.value }.filter { it.key.isExplored(unit.civ) }.take(3).forEach {
```
yairm210, line 717: Filter first, sort later? I don't think it really matters, but if we're already changing this...

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -97,29 +97,69 @@ object SpecificUnitAutomation {
...
+        // If we have gone more than 3 turns without founding a city lets search a wider area
+        // TODO: Figure out a way to not use turns as it might not be the best metric, what if we are a civ starting in the middle of a game?
+        val rangeToSearch = if (unit.civ.cities.isEmpty() && unit.civ.gameInfo.turns < 4) (3 - unit.civ.gameInfo.turns).coerceAtLeast(1) else null
+        
+        // It's possible that we'll see a tile "over the sea" that's better than the tiles close by, but that's not a reason to abandon the close tiles!
+        // Also this lead to some routing problems, see https://github.com/yairm210/Unciv/issues/3653
+        val bestTilesToFoundCityReturn = CityLocationTileRanker.getBestTilesToFoundCity(unit, rangeToSearch)
+        val bestTilesToFoundCity = bestTilesToFoundCityReturn.first
+        val bestTile = bestTilesToFoundCityReturn.second.first
+        val bestTileRank = bestTilesToFoundCityReturn.second.second
```
yairm210, line 112: This is what I meant about a new data class. Just putting this in a class with named fields would make this instantly less esoteric

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -97,29 +97,69 @@ object SpecificUnitAutomation {
...
+                        bestTilesToFoundCity[it.getTile()]!!
+                    else -1f
+                }
+                if (bestSettlerInRange == unit && foundCityAction?.action != null) {
+                    foundCityAction.action.invoke()
+                    return
+                }
+                // Since this settler is not in the best location, lets assume the best settler will found their city where they are
+                if (bestSettlerInRange != null)
+                    bestTilesToFoundCity.filter { it.key.aerialDistanceTo(bestSettlerInRange.getTile()) > 4 }
```
yairm210, line 133: This doesn't do anything, you need to assign it back to the bestTilesToFoundCity :)

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
yairm210: I mean we won't even need these as separate vals, we can just reference A.B directly and it will be readable

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -1,143 +1,141 @@
...
+            var locationSpecificTileValue = 0f
+            // Don't settle near but not on the coast
+            if (rankTile.isCoastalTile() && !onCoast) locationSpecificTileValue -= 10
+            // Apply the effect of having a lighthouse, since we can probably assume that we will build it
+            if (onCoast && rankTile.isOcean) locationSpecificTileValue += 1
+            // Check if there are any new unique luxury resources
+            if (rankTile.resource != null && rankTile.tileResource.resourceType == ResourceType.Luxury
+                && !(civ.hasResource(rankTile.resource!!) || newUniqueLuxuryResources.contains(rankTile.resource))) {
+                locationSpecificTileValue += 10
+                newUniqueLuxuryResources.add(rankTile.resource!!)
```
yairm210, line 102: This should be moved outside of this function

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -1,143 +1,141 @@
...
+            distanceToCityModifier *= when {
+                city.population.population >= 12 -> 4f
+                city.population.population >= 8 -> 3f
+                city.population.population >= 3 -> 2f
+                else -> 1f
+            }
+            // It is worse to settle cities near our own compare to near another civ
+            // Do not settle near our capital unless really necessary
+            // Having a strong capital is esential to constructing wonders
+            if (city.civ == civ) distanceToCityModifier *= if (city.isCapital()) 5 else 2
```
yairm210, line 84: Sounds like a separate function, `getDistanceToCityModifier(newCityTile, city)`

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
yairm210, line 40: Should be a data class, instead of a triplet - Would make things clearer on both ends for the next person reading this

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
tuvus: The only reason I was doing returning the best tile separately was for performance, I guess. If we compute the best value while ranking the tiles it should take much less time than taking maxOf() or something like that. 
But yes, I could store the best tile and get the best tile rank that way. Would a Pair of the map and best tile still be too much clutter?

tuvus:
> AI _should_ settle in-place if the map specifically places them there, that's why that code was there in the first place
>
> "Near coast" is not always desirable, especially for the first city - more water tiles means less tiles that can generate production, meaning less building in the city overall, so there's a balancing act here. If you already have cities nearby that are coastal, then this becomes even less important.

Okay, I will revert to this since we can always change it later.

> Players _should not_ be given hints according to tiles they don't see! The reason we cheat with the AI is so that the AI has a slightly better chance, since overall human players are much better than the AI.

Just to clarify, the hints only show on explored tiles, although the hints account for unexplored tiles and invisible. This is already an advantage for the AI.
In my opinion, I have never found these hints useful, although this might change with valuing the tiles differently. I feel like if we are going to give the player some sort of hint, they should be able to trust it to not be a poor decision for the later game. I am fine either way. It is unlikely that it will ever be able to match a human decision.

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
yairm210: I think that a one-line data class is just so easy to make that there's no reason to use pairs at all :)

tuvus:
Is the tab between lines not preferred? I like using the up and down arrow keys to navigate through a code block. If the whitespace isn't there then the cursor will be moved all the way to the left. Then to write a line, I would either have to press tab a few times or backspace and enter, unless there is some other shortcut that I don't know.

yairm210 (approved):
LGTM

tuvus:
@yairm210 
Should I revert the change, making the players receive hints based on unexplored tiles and invisible resources, as discussed above?

yairm210:
Yes, I think so

yairm210:
But not a blocker for me

yairm210:
We can add extra improvements later, I'd rather get this version out :)

PR #10431, opened 2023-11-04
Add three new uniques

PLynx01:
I would like to add three new conditionals. These are

<is controlled by Human>
<is controlled by AI>
Sell [buildingFilter] [cityFilter]

The first two are very simple. They will be used by modders to make features restricted to Human players or AI.
The last is the variation of removeBuilding unique was added by me. The difference is that SellBuilding gives money for removed buildings. 

I hope the implementation would be quicker than in case of my before PR.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -664,6 +664,29 @@ object UniqueTriggerActivation {
                 return true
             }
 
+            UniqueType.SellBuilding -> {
+
+                val applicableCities = civInfo.cities.asSequence().filter {
+                    it.matchesFilter(unique.params[1])
+                }
+
+                for (applicableCity in applicableCities) {
+                    var buildingsToSell = applicableCity.cityConstructions.getBuiltBuildings().filter {
+                        it.matchesFilter(unique.params[0])
+                    }.toSet()
+
+                    buildingsToSell = buildingsToSell.filter {
+                        it.isSellable()
+                    }.toSet()
+
+                    for (building in buildingsToSell) {
+                        applicableCity.sellBuilding(building)
+                    }
+                }
+
+                return true
+            }
+
             else -> {}
         }
         return false
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -671,13 +671,11 @@ object UniqueTriggerActivation {
                 }
 
                 for (applicableCity in applicableCities) {
-                    var buildingsToSell = applicableCity.cityConstructions.getBuiltBuildings().filter {
+                    val buildingsToSell = applicableCity.cityConstructions.getBuiltBuildings().filter {
                         it.matchesFilter(unique.params[0])
-                    }.toSet()
-
-                    buildingsToSell = buildingsToSell.filter {
+                    }.filter {
                         it.isSellable()
-                    }.toSet()
+                    }
 
                     for (building in buildingsToSell) {
                         applicableCity.sellBuilding(building)
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -672,9 +672,7 @@ object UniqueTriggerActivation {
 
                 for (applicableCity in applicableCities) {
                     val buildingsToSell = applicableCity.cityConstructions.getBuiltBuildings().filter {
-                        it.matchesFilter(unique.params[0])
-                    }.filter {
-                        it.isSellable()
+                        it.matchesFilter(unique.params[0]) && it.isSellable()
                     }
 
                     for (building in buildingsToSell) {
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -598,6 +599,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
 
 
     /////// civ conditionals
+    ConditionalIsHuman("is controlled by human", UniqueTarget.Conditional),
+    ConditionalIsAI("is controlled by AI", UniqueTarget.Conditional),
```
yairm210, line 603: This is not how conditionals are phrased - they need to start with "when ..."

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -664,6 +664,29 @@ object UniqueTriggerActivation {
...
 
+            UniqueType.SellBuilding -> {
+
+                val applicableCities = civInfo.cities.asSequence().filter {
+                    it.matchesFilter(unique.params[1])
+                }
+
+                for (applicableCity in applicableCities) {
+                    var buildingsToSell = applicableCity.cityConstructions.getBuiltBuildings().filter {
+                        it.matchesFilter(unique.params[0])
```
yairm210, line 675: Why not apply both filters here? Filter for isSellable as well
And you don't need to .toSet() this at all

PLynx01:
I've applied your recommendations. Can you merge now?

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
yairm210: I think you misunderstood me... .filter { A && B }

PLynx01: I've updated this function according to your remark. Is this PR ready to be merged?

SeventhM:
Question: why is human/AI checks conditionals and not nationFilters?

yairm210:
@SeventhM GOOD POINT!

yairm210:
Nation filters should be "AI" and "Human"

PLynx01:
> Nation filters should be "AI" and "Human"

So do I need to modify the nationFilter? I mean, add human and AI?

PLynx01:
> Nation filters should be "AI" and "Human"

So, how the mentioned uniques should look like?

Option A
(refers to current nation)
<when controlled by human>
<when controlled by AI>

Option B
<when [nationFilter] controlled by human>
<when [nationFilter] controlled by AI>

Option A, B, both or maybe something else?
Give me the text of the proposed uniques if you can.

yairm210:
Not as a unique, but as an option for nationFilter
So nationFilter should accept "AI" and "Human" as acceptable filters and filter them

PLynx01:
> Not as a unique, but as an option for nationFilter So nationFilter should accept "AI" and "Human" as acceptable filters and filter them

In that case, we don't need new unique. We need to add AI and Human options to this:
<for [nationFilter]>

PLynx01:
@yairm210 
Is this what you asked for?

13e1871

PLynx01:
@yairm210 

Is my PR ready to be merged? Or do you have some remarks about it.

yairm210:
This raises a good point - AI vs Human is not actually *nation-dependant*, it's actually *civ dependant*. We currently don't have a CivFilter to contain both Nation filters and other filters, makes me wonder what other "civ but not nation" filters we're missing out on...
We'll need to meditate on this

PLynx01:
> This raises a good point - AI vs Human is not actually *nation-dependant*, it's actually *civ dependant*. We currently don't have a CivFilter to contain both Nation filters and other filters, makes me wonder what other "civ but not nation" filters we're missing out on...
> We'll need to meditate on this 

Can't we just make two simple Human/AI uniques as I stated in beginning of the thread?

What is the reason to integrate them into <for [nationFilter]> ?

Please explain me

yairm210:
So they can be used anywhere that we have unit filters or nation filters

PLynx01:
One more question:
What is the difference between Nation and Civilization in game?

Can these two be used interchangeably?

SeventhM:
> What is the difference between Nation and Civilization in game

In the code, a "nation" is the base civ and its uniques, while a "civilization" is more or less a collection that includes the player and all of its mechanics. In game... I think so far it's more or less interchangeable. I don't think we actually have anything where information on the nation itself matters besides I guess the civilopedia

SeventhM:
> only one use

mapUnitFilter and tileFilter also uses nationFilter

PLynx01:
So, what's our decision?

1. Altering the nationFilter to accept Human/AI 

2. Separate two uniques for Human/AI players

3. Other solution (please reply)

yairm210:
I Implemented solution 1

PLynx01:
> I Implemented solution 1 

If the Human/AI filter is implemented, should we add the Sell Building(s) unique mentioned earlier?

yairm210:
Sell building can be added as a unique

PLynx01:
> Sell building can be added as a unique

So what should I do to implement this?

PLynx01:
The discussion about SellBuildings has been moved to #10514

PR #10443, opened 2023-11-07
Add gain stat modified by speed and improvement speed by filter uniques

SeventhM:
Adds "Gain [amount] [stat/resource] (modified by game speed)"
Adds "[relativeAmount]% [improvementFilter] improvement construction time"
For the deprecation message, do I put in the current version (4.8.19)?

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -125,7 +125,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     RoadsConnectAcrossRivers("Roads connect tiles across rivers", UniqueTarget.Global),
     RoadMaintenance("[relativeAmount]% maintenance on road & railroads", UniqueTarget.Global),
     NoImprovementMaintenanceInSpecificTiles("No Maintenance costs for improvements in [tileFilter] tiles", UniqueTarget.Global),
+    @Deprecated("As of",ReplaceWith("[relativeAmount]% [All] improvement construction time"))
     TileImprovementTime("[relativeAmount]% tile improvement construction time", UniqueTarget.Global, UniqueTarget.Unit),
+    SpecificImprovementTime("[relativeAmount]% [improvementFilter] improvement construction time", UniqueTarget.Global, UniqueTarget.Unit),
 
     /// Building Maintenance
     GainFreeBuildings("Gain a free [buildingName] [cityFilter]", UniqueTarget.Global, UniqueTarget.Triggerable),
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -705,7 +707,8 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     OneTimeConsumeResources("Instantly consumes [amount] [stockpiledResource]", UniqueTarget.Triggerable),
     OneTimeProvideResources("Instantly provides [amount] [stockpiledResource]", UniqueTarget.Triggerable),
 
-    OneTimeGainStat("Gain [amount] [stat/resource]", UniqueTarget.Triggerable),
```
yairm210, line 708: Please don't override :)

SeventhM: The only change there is the text. This is already replaced by `OneTimeProvideResources`, and the trigger already longer checked resources

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -125,7 +125,9 @@ enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags:
     RoadsConnectAcrossRivers("Roads connect tiles across rivers", UniqueTarget.Global),
     RoadMaintenance("[relativeAmount]% maintenance on road & railroads", UniqueTarget.Global),
     NoImprovementMaintenanceInSpecificTiles("No Maintenance costs for improvements in [tileFilter] tiles", UniqueTarget.Global),
+    @Deprecated("As of",ReplaceWith("[relativeAmount]% [All] improvement construction time"))
     TileImprovementTime("[relativeAmount]% tile improvement construction time", UniqueTarget.Global, UniqueTarget.Unit),
+    SpecificImprovementTime("[relativeAmount]% [improvementFilter] improvement construction time", UniqueTarget.Global, UniqueTarget.Unit),
```
yairm210, line 130: Is this necessary? Sounds to me like maybe what we actually need is an improvementFilter conditional

If we are going this route though, let's reword this to "[relativeAmount]% construction time for [improvementFilter] improvements"

yairm210: Also lower-case "all" should be an option, because the current uniques make me wince when I see an upper-case All in the middle of a sentence...

SeventhM: > Sounds to me like maybe what we actually need is an improvementFilter conditional

Maybe, but unlike, say, a cit6Filter conditional, I'm not sure how often that'll come up. Off the top of my head, this is probably one of the only uniques that don't already support improvements as a tileFilter and that'll make sense for a conditional to apply to. The only other immediate alternative I can think of is a maintenance reducer for improvements

I can go ahead and reword it though

SeventhM: > Also lower-case "all" should be an option

Probably something more to add to the list of available  improvement filters (and other filters for that matter), not entirely in scope here

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
SeventhM: > forget it.

Yeah, it looks weird, but that's just github's diff engine combining stuff unnecessarily

yairm210 (approved):
The "All" issue is indeed out of scope for this

PR #10446, opened 2023-11-07
Added Human and AI filters, separated civFilter from nationFilter

yairm210:
As discussed in #10443, we separate between filtering on the *nation* - the object independent of the current game - to the *civ* - the instance of that nation in the specific game

Only one usage of nation filtering not via civ filtering exists, but I want to keep the option open

This adds "Human" and "AI" filters at the Civ level, since it is game-dependent

@SeventhM

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -492,8 +492,8 @@ class Civilization : IsPartOfGameInfoSerialization {
 
     fun matchesFilter(filter: String): Boolean {
         return when (filter){
-            "Human" -> isHuman()
-            "AI" -> isAI()
+            "Human player" -> isHuman()
+            "AI player" -> isAI()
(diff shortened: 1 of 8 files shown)


SeventhM (approved):

SeventhM:
> whether the matching Unique is on the Civ's policies or on its Nation

I'm... Confused. There's currently no real field for civ related untyped uniques. civFilter in this sense is to nationFilter what mapUnitFilter is to baseUnitFilter. Just as nationFilter does look through policies, civFilter wouldn't either. The point of the extra layer is to distinguish game reliant states (such as players vs AIs or teams if such a system was added) from json reliant states (such as untyped uniques or the sort)

Speaking of... Now that I think about it, shouldn't allies/enemies/friends be a civFilter in this analogy? That's also reliant on game state

yairm210:
Actually this should be *Human Player* and *AI Player* to distinguish this from a unit type

PR #10448, opened 2023-11-08
Fix bug: the GlobalPoliticsTable disappears

lishaoxia1985:
1. The bug is that: the default Politics of Overview screen shows diagram, and then you touch the button "Show global politics" in Politics of Overview screen, the GlobalPoliticsTable disappears (Maybe sometimes the position of GlobalPoliticsTable changes, when the default Politics of Overview screen shows diagram it's in the center, when the default Politics of Overview screen shows global politics it's on the top)
2. Now we replace the TextButton "City-States" with CheckBox

PR #10451, opened 2023-11-08
(UI) Politics overview diagram: Add legend popup

SomeTroglodyte:
#### Check out #10448 first

<details><summary>Scrnsht</summary>

image

... access by clicking/tappin right in the middle.
</details>

* The legend actually "lies" about China having a Defensive Pact with 7 of the 8 City-states, due to #10079 "lying" with the line color - @tuvus please explain (again?), and should we adapt the legend (needing much more text and a new template string)?
* File desperately needs splitting. Should be a separate, refactor only PR. Input - the two "halves" offer themselves (now editor regions) - but they're tightly integrated with the fixed/scrollable split of the parents, so as self-contained Widgets might not work out (kotlin, get partial classes!). Standalone top-level functions? object as Namespace-like container? The now two sub-Widgets are no puzzler - own files...
* Something is wrong with AnimatedMenuPopup's centering, but I am too mind-weak to analyze that right now. Noticeable only on lower "Display size" settings - as seen above. And it's not only that one helper not doing what its name says...

<sub><sup>The easter egg is a little clearer on an actual screen than up there. alpha = 0.15 seems just right, 0.2 was too obvious.</sup></sub>


tuvus:
> The legend actually "lies" about China having a Defensive Pact with 7 of the 8 City-states, due to https://github.com/yairm210/Unciv/pull/10079 "lying" with the line color - @tuvus please explain (again?), and should we adapt the legend (needing much more text and a new template string)?

What do you mean by lying? The Cyan line between two major Civs signifies a defensive pact. The Cyan line between a major Civ and a city-state signifies that the city-state is allied with the major Civ. 

Quote from #10079:

> I personally can never tell who is allied with a city-state on the diagram when it is green. So, making it [cyan] shows the one civ who is allied with it. Actually, now that I think about it, does the city-state use influence or their opinion? Each city-state also has a diplomacyManager with every other civ and hence an opinion. I don't think the diagram changes this for city-states.

The lines shown previous to #10079 from a city-state to any civ only reflected their diplomatic relations. However, city-states never use their diplomatic relations. Even if a major civ is allied to the city-state, the city-state will have a regular green line to it's ally. So I used the cyan line color to signify a defensive pact and alliance to a city state. From a war standpoint, they are functionally equivalent to what will happen if you declare war on a Civ with cyan line.

If you want to change either of these colors, go ahead.

yairm210:
Please have the text for the l we legend be white, done of those are barely legible

yairm210:
Background image is far too large, please cut it down to at least a quarter of current size

Ouaz:
> Also, I'd like input on the pix in the other post (also behind a 'details', mayhap too well hidden)

I'll go for B, even if the color difference is barely noticeable, as Ally and Defensive Pact is not the same thing at all.

Btw, long time ago, we suggested a rework of the line colors (with Cabalero I think), because as you said _"I see there was an intended progression of the level colors"_, but the progression is so "light", that it's a bit hard to actually make the difference between `Competitor/Neutral` and `Favorable/Friend/Ally` (in the end, I go check directly the relation in the Diplomacy screen to be sure).

I can't find the Github discussion anymore (and it disappeared from my notification list, weird), but I kept the mockup:
image

The discussion has ended in a stale state, because of no agreements (_"No, I prefer this color over this one"_, _"I want pink for friend"_, _"Brown for this or that is the worst idea ever"_, etc. :rofl: )

tuvus:
Just so you know, an ally is not the same as being allied with a city-state or having a defensive pact. It means that the two Civs have a relationship of greater than 80.

Also, why make Friend and Favorable very different colors if there is very little practical difference in the game?
Maybe make Afraid yellow and Competitor blue? 

Otherwise, I think this coloring scheme might better describe the negative side of relations. For positive relations, I think the current version is a little better.

Ouaz:
> Also, why make Friend and Favorable very different colors if there is very little practical difference in the game?

Mainly because being "Friend" with a city-state grants the resource bonuses. I don't remember which relationship value range triggers "Favorable" ? (15-30 ?)
Having current colors so close for Favorable/Friend makes hard to distinguish the relationship on the diagram. The idea is to make "Favorable" color stands out more (it grants no bonus, so it requires to work the relationship with a given city-state, to reach "Friend" and have the bright green instead)

> Maybe make Afraid yellow and Competitor blue?

The idea is a kind of spectrum progression, based on relationship numeric values:

- Negative relationship: warm colors  (`yellow < orange < red < purple` = `competitor < enemy < war < unforgivable`)

- In-between: neutral colors (`grey < soft brown` = `neutral < afraid`) - soft brown making somewhat the link with the following yellow color -

- Positive relationship:  blue/green colors (`dark blue < bright green < turquoise` = `favorable < friend < ally` ). Green for friend because it's the most common relationship when playing "normally" (not being a warmonger lol) and therefore making other relationships stand out more.

Competitor in blue or Afraid in yellow would break this progression.

The current color scheme from Competitor to Ally has too much similar colors (almost the same tone for me ^^).
image

Ouaz:
> Decide how to treat "Ally=GREEN on DiplomacyScreen left side and CityButton but CYAN on Overview": Unify or keep, if keep then as another field on RelationshipLevel: convert color to buttonColor and lineColor?

Unify by keeping CYAN everywhere. ^^ GREEN is friendship (more common relationship).

yairm210:
Argh
The 'yellow star on cyan' burns my eyes
Please add a black star behind the yellow one to give it an outline

yairm210:
Also probably a black outline to the color circle in general

yairm210:
@I'm just going to merge this and we can argue colors later.

yairm210 (approved):

Ouaz:
Haha, another UI improvement well hidden (one has to think to click the diagram itself ^^)

PR #10471, opened 2023-11-10
(UX, minor) Change "took tile damage" notification to **select** the poor victim

SomeTroglodyte:
.. so one can better go about saving it


yairm210:
If there are 2 units in the tile, this may select the wrong one
If the unit took tile damage and was later killed, the time may be empty, and will cause a crash

Issue #10130, opened 2023-09-14
AI appears to ignore defensive pacts when attacking. 


SpacedOutChicken:
Android 13, Play Store version 4.8.4

(No saved game yet, will try to get you one)

I had a force of about 60000 when the AI attacked me with a force of about 11000. I couldn't figure out why they would do this, but then I noticed that they also declared war with my ally, who only had a force of about 3000, but who had a defensive pact with me. I believe that the AI failed to take that pact into account when declaring war with the weaker side, causing them to start a fight with me that they couldn't win.

tuvus:
Yes, the AI doesn't currently take defensive pact allies into account. From what I have searched on Civ V, their AI also doesn't take defensive pacts into account. (Not that this is preferable behaviour)

I am planning on adding this functionality later, but it takes a while to tune the AI not to value defensive pacts too much (or else it will get stuck not attacking). For a reference, it is taking a while to tune the AI to declare friendship and defensive pacts already.

SpacedOutChicken:
Also, sorry for not adding the "bug" label when I posted this. For some reason, it didn't get added, and now I can't assign it. Can someone fix that?

tuvus:
Only Yairm can change the labels, I think.

tuvus:
I don't know where I can get the source code. I have searched on GitHub and found something but I am unsure if it is accurate or has the expansions. My reference comes from players complaining about it, so it isn't very reliable. It could also depend on the expations.

yairm210:
Solved by linked PR

PR #10342, opened 2023-10-24
AI diplomacy balancing

tuvus:
This commit changes how the AI values signing friendship, defensive pacts, and declaring war.

From the simulation I have done on the previous versions, the AI declares peace and defensive pacts too often since being able to offer them. This results in an undesirable stalemate were everyone gangs up on one civilization.

<details><summary>Picture of stalemate</summary>

AIDiplomacyStalemate

</details> 

Using AutoPlay, I ran an AI-only simulation (prince difficulty) to see how the AI behaves on larger maps to work out changes. Usually with 6-13 AI.

Here is a list of changes that I have made to the AI's decisions:
DOF = Declaration of Friendship, DP = Defensive Pact
- DOF & DP & Open borders: Each of these now has a declined_X_ flag applied to the DiplomacyManager to prevent repetitive calculations.
- DOF & DP: The more wars the AI has, the more they want to sign a DOF & DP
- DOF: AI wants to ally with Civ's less as the other civ has more friends
- DP: AI does not want to sign DP with other Civs that are mad at their allies
- War: AI now ranks each civ that it has enough motivation to attack based on score and chooses the highest one
- War Motivation: AI now takes into account the other Civ's DP allies
- War Motivation: AI now takes into account relative score, production and 
- War Motivation: AI is less likely to declare war on a Civ that they have a DP with
- War Motivation: AI is more likely to attack Civs based on their wonders
- War Motivation: AI is more likely to attack a Civ if they are at war with their ally

Of course, more balancing will be required and most numbers are arbitrary. So, any suggestions and recommendations are more than welcome.

Note: I used the AutoPlay feature and merged parts of it into this branch. After finishing the simulations, I tried to revert the commits again. I don't think anything has passed through but something could have.

Also, could someone check that I did the score, production, and tech modifiers correctly and logically?

PR #10476, opened 2023-11-11
AI diplomacy balancing

tuvus:
The previous AI diplomacy balancing branch was having some problems merging with AutoPlay on some of my other branches, so I made a new clean branch with all of its changes and without its history.
See original PR #10342


tuvus:
Fixes #10130

PR #10482, opened 2023-11-12
Force atlas loader to use UTF-8

SomeTroglodyte:
### Don't merge

I believe this to work, but have no environment + data available at the moment to test. That, and our far-eastern friends with the weird µSquishy system encodings have learned to avoid the consequences by now.

So - this is more of a "behold and despair", then forget kind of thing, it is probably better to leave that loader vulnerable. But if in need, this is the way, only possible since Gdx 1.12.

### A look back

A texture atlas is a bunch of single png files combined, which is good for 3D chips, as they can load only so many textures at once, but can easily only paint parts of them. The .atlas files are text files that carry information which original is where. Thus, packing copies the original file names into that text file. Next consequence: reader and packer must be in agreement on the encoding of that text - but as it is, nobody controls that encoding but the OS (when reading using a certain API, packing has better control). And Windows still uses weird non-Unicode encodings by default - as "codepace for non-unicode apps" (huh? we _are_ unicode?) but here's the thing - Java queries that setting and reuses it as its own default. With Java 18, they decided otherwise, fixing the mismatch. You can force Windows to behave - but the setting is hard to find and declares itself beta. Or you can tell the individual file reading functions to use a specific encoding - which was inaccessible for the atlas loader. That has changed with Gdx 1.12, it now uses a different reader, still without telling it what encoding to use - but that one is overridable. Uff, now I'm confused. What a mess.

### DIFF
--- a/core/src/com/unciv/ui/images/ImageGetter.kt
+++ b/core/src/com/unciv/ui/images/ImageGetter.kt
@@ -37,6 +37,7 @@ import com.unciv.ui.components.extensions.toLabel
 import com.unciv.ui.components.fonts.FontRulesetIcons
 import com.unciv.ui.screens.basescreen.BaseScreen
 import com.unciv.utils.debug
+import java.io.BufferedReader
 import kotlin.math.atan2
 import kotlin.math.max
 import kotlin.math.min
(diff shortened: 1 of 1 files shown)


yairm210:
Please don't open PRs and then say "ignore this don't merge". What am I supposed to do with this, then...?

yairm210:
Ok
Then I think if anyone creates non-utf8 atlases we can define that as a Their Problem

Issue #10497, opened 2023-11-13
Feature request: better contrast between resource logos


Cwpute:
**Is your feature request related to a problem? Please describe.**
![unciv icons](https://github.com/yairm210/Unciv/assets/80604512/d2823107-84c5-48a7-8b10-2096bacebabc)
I am colourblind, and sometimes i wish i could distinguish the resource logos better, especially between Gold, Population, and Food. They all have icons which makes them distinct, and of course you end up infering which is which depending on the terrain and building on top, but still, for players discovering the game, quick visualisation, or more surprizing scenarios, having better coulour differenciation will help.

**Describe the solution you'd like**
I fiddled a bit with the colours of the logos. First i went for Food logo, but it didn't work well.
Quickly i switched to working on the Population logo:
![unciv icons test2](https://github.com/yairm210/Unciv/assets/80604512/a2466d80-00f0-456d-ba0f-7c8aac11de65)
… that's right, a simple colour swap between the Population icon and its background.
If this one changes, it means you'd have to also change the "empty hex logo" when nobody is exploiting a tile. I'd argue no logo at all would also work, or the same one with transparency applied. But extra work still.

I realized midway through that what mattered the most was differenciating Food and Gold directly on the map. So i tried my hand on the Gold logo:
![unciv icons test4](https://github.com/yairm210/Unciv/assets/80604512/08928d52-9ffe-4e6a-ac35-5b98f8d8437a)
It still is a swap between background and icon, but i put more contrast between the two, which i think strikes a good balance.

**Describe alternatives you've considered**
Just for a test, i combined two variations of the Population and Gold logos, and i think it looks good too.
![unciv icons test5](https://github.com/yairm210/Unciv/assets/80604512/e42581d8-7fba-42c6-b7ef-9b07d55d7db4)

**Additional context**
There are many colourblind variations, i personally don't see red as vivid as most describe it, and i sometimes mix up light green and yellow (hence this suggestion).
If anyone else having colourblindness can give their own input on the matter, you're welcome!

Cwpute:
Doesn't it fall within the accessibility features rather than mods ? What makes it more fit for a mod than an included base feature ?
And even without talking about accessibility, wouldn't it profit to everybody having better contrasted logos ?

yairm210:
The middle gold icon looks very bad, but the second version of the gold/population look good
Can you provide those as files so we can poll this on Discord?

Cwpute:
[...]

PR #10505, opened 2023-11-14
Map overlay toggle buttons rework

### LINKED ISSUE #10491: Tiny UI regression: Map overlay toggle buttons
Colour discrepancies I'm sure were uniform last time I toyed with these:
image

Finding a breaking change not worth it - #6659 (ages ago) moved the code, so git blame is a bit harder than normal. Throw in the opportunity for some better moddability and redoing these consistently seems the way to go.

See: `com.unciv.ui.screens.worldscreen.minimap.MapOverlayToggleButton` and 4x instantiation in MinimapHolder. Strange architecture - why is MapOverlayToggleButton not a Widget but has a factory for one? Why does it have a lazy that is _immediatly_ triggered in init? Did I do that???

Will keep warm for a bored (southern hemisphere 🤪) winter day.

SomeTroglodyte:
Close #10491 , #10497 partially taken into account as hint to use the higher contrast ratio of the old ones.

* Changes the MapOverlayToggleButton class from a 'kinda' factory to an actual Gdx Actor
* Removes re-coloring kludges
* Feeds it new, dedicated (and thus separately moddable) icons
* Icons redone from source (as in, svg dl from tnp then import into oversized canvas only scaling down for export).
* Includes "source" gimps for easy resolution change
* Uses the unused Footprints I found in credits for "Movement arrows" toggle, but the source gimp offers both in its layers, so switching is easy. Non-use could be the result of discussions, didn't search, but I quite like them.

2-frame gif:
Toggles


yairm210:
Pic looks good!

PR #10510, opened 2023-11-16
Test/add tests for cityexpansionmanager

Framonti:
Tests for CityExpansionManagers + mini update of precondition checking to improve error message.

Discussion point: on Discord, someone opened a suggestion/bug about not allowing City States to buy tiles with gold, as it's not their behaviour in Civ V. 

We can fix it by simply updating canBuyTitle as such
```kotlin
fun canBuyTile(tile: Tile): Boolean {
        return when {
            city.isPuppet || city.isBeingRazed -> false
            tile.getOwner() != null -> false
            city.isInResistance() -> false
            city.civ.isCityState() -> false  // new condition
            tile !in city.tilesInRange -> false
            else -> tile.neighbors.any { it.getCity() == city }
        }
    }

```
I can add this change in this PR if we agree on it without much troubles.
Let me know

### DIFF
--- a/core/src/com/unciv/logic/city/managers/CityExpansionManager.kt
+++ b/core/src/com/unciv/logic/city/managers/CityExpansionManager.kt
@@ -171,7 +171,7 @@ class CityExpansionManager : IsPartOfGameInfoSerialization {
      * @param tile The tile to take over
      */
     fun takeOwnership(tile: Tile) {
-        if (tile.isCityCenter()) throw Exception("What?!")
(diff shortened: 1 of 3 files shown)


yairm210:
If in Civ V city states can't buy tiles, let's go with that 👍🏿

yairm210 (approved):

Framonti:
I invited the discord user who raised the question to continue the discussion here on GitHub. In the meantime, I won't change anything.

PR #10523, opened 2023-11-19
Add a TriggerUponDiscoveringTile unique type

karmaBonfire:
These changes add a `TriggerUponDiscoveringTile` unique of target `UnitTriggerCondition`. It allows to define actions that occur when a unit discovers a previously fogged tile and triggers for each such tile.

It is a modding-related feature since AFAIK there are no relevant mechanics in Civ5.

I have tested following uniques:
 - `This Unit gains [5] XP <upon discovering a [Water] tile>`
 - `This Unit gains [5] XP <with [50]% chance> <upon discovering a [Water] tile>`

This is my first Unciv PR and my first Kotlin PR - reviews are welcome and encouraged!

### DIFF
--- a/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
@@ -21,6 +21,7 @@ import com.unciv.models.ruleset.unique.StateForConditionals
 import com.unciv.models.ruleset.unique.Unique
 import com.unciv.models.ruleset.unique.UniqueMap
 import com.unciv.models.ruleset.unique.UniqueType
+import com.unciv.models.ruleset.unique.UniqueTriggerActivation
 import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.ruleset.unit.UnitType
 import com.unciv.models.stats.Stats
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/map/mapunit/MapUnit.kt
```diff
@@ -343,8 +344,23 @@ class MapUnit : IsPartOfGameInfoSerialization {
...
+            }
+            for (tile in newlyExploredTiles) {
+                // Include tile in the state for correct RNG seeding
+                val state = StateForConditionals(civInfo=civ, unit=this, tile=tile);
+                for (unique in getTriggeredUniques(UniqueType.TriggerUponDiscoveringTile, state)) {
+                    if (unique.conditionals.any { it.type == UniqueType.TriggerUponDiscoveringTile
+                        && tile.matchesFilter(it.params[0], civ) })
+                        UniqueTriggerActivation.triggerUnitwideUnique(unique, this)
+                }
+            }
```
yairm210, line 360:
```suggestion
            if (newlyExploredTiles.isNotEmpty){
                val unfilteredTriggeredUniques = getTriggeredUniques(UniqueType.TriggerUponDiscoveringTile, StateForConditionals.IgnoreConditionals)
                for (tile in newlyExploredTiles) {
                    // Include tile in the state for correct RNG seeding
                    val state = StateForConditionals(civInfo=civ, unit=this, tile=tile);
                    for (unique in unfilteredConditionals) {
                        if (unique.conditionals.any { it.type == UniqueType.TriggerUponDiscoveringTile
                            && tile.matchesFilter(it.params[0], civ) }
                                && unique.conditionalsApply(state))
                            UniqueTriggerActivation.triggerUnitwideUnique(unique, this)
                    }
                }
            }
```

I think this is what SomeTrog was aiming for, tbh my first thought was also to get triggerables once

yairm210 (approved):
All in all VERY impressive first PR and a fantastic flavor for unique explorer types!

core/src/com/unciv/logic/map/mapunit/MapUnit.kt
yairm210: You're right, it needs a .toList() :)

karmaBonfire:
@SomeTroglodyte @yairm210 Thanks, I really appreciate your feedback! Glad you like this feature. I just wanted to make a mod but found out there's no triggers I need, so I decided to dive deeper and implement it :)

core/src/com/unciv/logic/map/mapunit/MapUnit.kt
karmaBonfire: That's fair, given that this trigger is pretty niche, the game would calculate newlyExploredTiles far more often than needed. I've implemented a fix with both your suggestions, thanks again for your comments!

karmaBonfire:
> We have no way of telling modders this conditional will only work on units, right?

Well, this trigger is listed in UnitTriggerCondition category, so I suppose it's clear enough it only works for units

yairm210:
Unciv is currently a good *emergent storytelling* tool but a very, VERY bad *structured narrative* tool... I think anyone who wants something like that is best served going to another game

PR #10526, opened 2023-11-19
[Translation] Add back "general" unit types

Ouaz:
[This commit](https://github.com/yairm210/Unciv/commit/dc6413d707d19189a16efd80c938d89be0f1a2c6#diff-68ce11ca325ca50972b37fa8eefa996aa5da37939ebde573ee9a1f293ae2c598R106) seems to have removed these 5 entries from all translation files (see [FR rollout](https://github.com/yairm210/Unciv/commit/2fd7779395f57b3818f27f8bbb0bdcbd44e9d25c#diff-6e923159c026fd23144f896837e0f75f0d2ab4dc76c1a70bb539dbdee9512527L2336) ), making them untranslated in-game :
- `Military = `
- `Great Person = `
- `Nuclear Weapon = `
- `relevant = `
- `City =`

These entries are used in Civilopedia articles ("Military" in many articles, other ones only in a few):

"Military" everywhere:
military

"Sweden" nation article:
greatperson

"Alhambra" wonder article:
relevant

"Battering Ram" unit article:
city

Adding them back to the translation file fixes the problem:
image
Etc.

Note: I'm not sure adding them to template.properties is the correct way to fix the problem (as other unit type entries are not listed there). But it works. ^^


yairm210:
Ayyy damn that's my bad for not noticing translation diffs
But they did need to move though

Ouaz:
Should I open an issue or we keep this as a reminder?

yairm210:
Since I understand that touching code may be a daunting thing, even if you're told exactly what to do, if you'd rather we put this line in instead of you just say the word :)

yairm210:
And @SomeTroglodyte You did the research for this to find out *why* and *what to do* so you deserve credit as well ;)

Ouaz:
I could have notice earlier, but `City = ` wasn't added back with this fix.

AFAIK, it is only used in the "Battering Ram" civilopedia article.

city

It should be added in "Map unit filters" as there's a `Non-City = ` entry there.

city

PR #10527, opened 2023-11-19
[code quality] Reorg, clean up and comment a few things

SomeTroglodyte:
I promise I've paid utmost attention to introduce zero functional changes - one exception. Descriptive commit messages.

The one change is: `DesktopGame.notifyTurnStarted` now has an effect on non-Windows, too, simply passing off to `GLFW.glfwRequestWindowAttention` as intended by the framework. Actually, the entire ex-MultiplayerTurnNotifierDesktop file could go and be replaced by 
```kotlin
    override fun notifyTurnStarted() {
        (Gdx.graphics as? Lwjgl3Graphics)?.window?.flash()
    }
```
... (But then it is possible it would flash only once on Windows, instead of our thrice - left to upstream discretion)... but I left the listener in, will be useful in the future.

### DIFF
--- a/desktop/src/com/unciv/app/desktop/UncivWindowListener.kt
+++ b/desktop/src/com/unciv/app/desktop/UncivWindowListener.kt
@@ -10,7 +10,7 @@ import com.sun.jna.platform.win32.WinUser
 import com.unciv.utils.Log
 import org.lwjgl.glfw.GLFWNativeWin32
 
-class MultiplayerTurnNotifierDesktop: Lwjgl3WindowAdapter() {
+class UncivWindowListener: Lwjgl3WindowAdapter() {
     companion object {
         val user32: User32? = try {
             if (System.getProperty("os.name")?.contains("Windows") == true) {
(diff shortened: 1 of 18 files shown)


desktop/src/com/unciv/app/desktop/UncivWindowListener.kt
```diff
@@ -39,27 +37,57 @@ class MultiplayerTurnNotifierDesktop: Lwjgl3WindowAdapter() {
...
         } catch (e: Throwable) {
             /** try to ignore even if we get an [Error], just log it */
             Log.error("Error while notifying the user of their turn", e)
         }
     }
+
+    private companion object {
+        /** Marshals JNA access to the Windows User32 API through [com.sun.jna.platform.win32.User32]
+         *
+         *  (which, by the way, says "Incomplete implementation to support demos.")
```
yairm210, line 80: lol
We have a saying, "nothing is as permanent as temporary"

PR #10539, opened 2023-11-21
Add a setting to forbid closing popups by clicking behind them

karmaBonfire:
These changes add a setting which, when enabled, prevents popups from being closed by clicking outside of them. Default value is `false` - this will keep original behavior. This feature was requested in [this Discord thread](https://discord.com/channels/586194543280390151/1176287339870425240) and I agreed it would be nice to have.

I've tested this change on a "meeting a city-state" popup.

<details>
<summary>A screenshot of "Display" setting tab</summary>

image

</details>

I feel that `forbidPopupClickBehindToClose` is a bit verbose, but I can't think of a better name.

### DIFF
--- a/core/src/com/unciv/ui/popups/options/DisplayTab.kt
+++ b/core/src/com/unciv/ui/popups/options/DisplayTab.kt
@@ -71,6 +71,7 @@ fun displayTab(
     addResetTutorials(this, settings)
     optionsPopup.addCheckbox(this, "Show zoom buttons in world screen", settings.showZoomButtons, true) { settings.showZoomButtons = it }
     optionsPopup.addCheckbox(this, "Experimental Demographics scoreboard", settings.useDemographics, true) { settings.useDemographics = it }
+    optionsPopup.addCheckbox(this, "Never close popups by clicking outside", settings.forbidPopupClickBehindToClose, true) { settings.forbidPopupClickBehindToClose = it }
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/ui/popups/options/DisplayTab.kt
```diff
@@ -71,6 +71,7 @@ fun displayTab(
     addResetTutorials(this, settings)
     optionsPopup.addCheckbox(this, "Show zoom buttons in world screen", settings.showZoomButtons, true) { settings.showZoomButtons = it }
     optionsPopup.addCheckbox(this, "Experimental Demographics scoreboard", settings.useDemographics, true) { settings.useDemographics = it }
+    optionsPopup.addCheckbox(this, "Never close popups by clicking outside", settings.forbidPopupClickBehindToClose, false) { settings.forbidPopupClickBehindToClose = it }
```
yairm210, line 74: This text needs to be in template.properties, for translating

yairm210:
Almost perfect, good job!

core/src/com/unciv/ui/popups/options/DisplayTab.kt
karmaBonfire: Done!

yairm210:
@SomeTroglodyte for context, most players like it

IMG_20231121_192625

yairm210 (approved):

PR #10540, opened 2023-11-21
Initial scenario/dev console, with 1 command :D

yairm210:
For a long time I've been thinking about how we can bring back scenarios, because they're such a cool storytelling tool, but they *suck* because every tiny thing that we want to manually edit we need to make UI for which is hella annoying!

INSTEAD
WE CAN DO THE OLDEN GOLDEN DEBUG PATH
WHICH IS DEV CONSOLE

Clicking ` will open up the dev console, only on single-player games
Which will allow you to cheat.

More importantly, it will allow you to change *specific things* in game *without* extra UI!
And since we already have a map, we can click on tiles, so we can `units add settler babylon` without having to specify tile x/y!
The world awaits! 🤩


yairm210:
WOW stuff is SO MUCH FASTER when you don't need UI
Damn I hate UI
Maybe shouldn't have made a game that's entirely visual, huh

yairm210:
add-remove-tile
console_commands
unit-add-remove

yairm210:
No, what will-ca offered was *running outside code interoperable with the game*
I'm offering an in-game CLI which is much, much smaller and simpler than that. No loops no magic (for now...?)

yairm210:
cure world peace and bring all known diseases

hackedpassword:
Fantastic yairm210. I saw that use of context, phenomenal.

yairm210:
Clean, concise, comfortable, composable

PR #10541, opened 2023-11-21
Add a map import tool able to read "Battle for Wesnoth" maps

SomeTroglodyte:
... sic. Time for another easter egg? Maybe, but it works.

<details><summary>Screenshots (outdated, see below)</summary>

### Where it's hidden:
image

### Example, from Wesnoth's "World Conquest 3p" generator, whatever that is:
Wesnoth map editor:
image

Imports to:
image

### Another example
Wesnoth map editor:
image

Imports to:
image

</details>

These are outdated because I just fixed a coordinate system mismatch bug and am too lazy to re-do the screenies. So those still show columns shifted up/down alternatingly.

There's room for improvement, especially in the tile conversion mappings. They're external in json, but not moddable (difficulty: merging makes little sense, mod selection could only be easy via "perm audiovis" and then which one if several have the file? Better add another file picker when there's demand)

### DIFF
--- a/core/src/com/unciv/ui/screens/mapeditorscreen/MapEditorWesnothImporter.kt
+++ b/core/src/com/unciv/ui/screens/mapeditorscreen/MapEditorWesnothImporter.kt
@@ -1,0 +1,157 @@
+package com.unciv.ui.screens.mapeditorscreen
+
+import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.files.FileHandle
+import com.unciv.Constants
+import com.unciv.json.json
+import com.unciv.logic.UncivShowableException
+import com.unciv.logic.files.FileChooser
+import com.unciv.logic.map.HexMath
+import com.unciv.logic.map.MapShape
+import com.unciv.logic.map.MapSizeNew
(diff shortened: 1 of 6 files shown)


yairm210 (approved):
Sounds cool! Open Source collab ;)

core/src/com/unciv/ui/screens/mapeditorscreen/MapEditorWesnothImporter.kt
```diff
@@ -0,0 +1,156 @@
...
+        }
+        allObjects.firstOrNull { it is TileImprovement }?.apply { improvement = name }
+        if (start == null) return
+        map.addStartingLocation(Constants.spectator, this)
+    }
+
+    private fun translateTerrainWML(code: String?) = sequence {
+        if (code == null) return@sequence
+        translationCodes[code]?.also { yieldAll(it) }
+        if (code.length >= 3) // kotlin slice is unsafe, unlike python slice
```
yairm210, line 149: I refuse to believe there isn't a better way to do this. What are you trying to accomplish?

PR #10559, opened 2023-11-23
Fixed Bosnian

SomeTroglodyte:
See commits individually
Closes #10557 

image

... added trikkyness:
- trailing blanks disappeared
- lines out of sync by one around the `#Trades` comment - manually fixed

... Well, do we want 3%?

PR #10561, opened 2023-11-24
"Fixed" Rusyn language

SomeTroglodyte:
Similar treatment (but keeping history), less work than #10559 - but would be a pity to waste at 67%?
Closes #10422 (and has 1 more commit from OP than that PR - their other branch looked plausible)
Attribution please entirely to @KotDivuar
Failing test solved pragmatically by deleting where not obvious

The matter with an appropriate Collator when the Java libraries just don't have a match had a little surprise - fastlane files. No idea whether F-Droid will actually accept that /rus/ folder...

image


KotDivuar:
Wow, it's wonderful! Thank you very much for add and fix! Unfortunately, I've had a few too busy weeks to get back to work for translate. But now I'm ready again. 

As I understand it, the rue-file now is not readable, since in my version of the game (updated, android) a new language has not appeared? Or did I misunderstand

KotDivuar:
Thank you

PR #10562, opened 2023-11-24
Allow mods to specify multiple prerequisite techs for a unit

dHannasch:
### TLDR

This branch makes it possible for mods to specify, in the JSON, that a unit requires multiple technologies.

This was tricky in Civilization V --- see e.g. https://forums.civfanatics.com/threads/producing-unit-condition-multiple-techs.389964
 --- but Unciv's unified data model seems to make it relatively straightforward.

This lists a unit in the Civilopedia for *both* its required technologies, which seems most natural to me.

### The Problem

Mods sometimes want to have units that depend on multiple technologies. For example, you might have two parallel tech tracks, one for weapons and one for mobility. The player can then build units with the best weapons and mobility they have researched.

But technically, that's not possible. The state-of-the-art workaround is to create a dummy zero-cost fake tech to own the unit, but that results in ugly spaghetti-snarl tech trees that are more difficult for the player to read and navigate through links. More commonly, modders *don't* do that, and instead resort to various improvised bidirectional kludges, resulting in warnings.

HorsemanObsoletesAtTechRifling

The goal, then, is to allow a mod to specify, in the JSON,, e.g.
> "name": "Chariot Archer",
> "requiredTechs": ["Animal Husbandry", "Archery"]

and have it Just Work.

requiredTechsCivilopedia

### The Data Model

This branch is *technically* fully functional in the sense that it allows "requiredTechs" to be specified in the JSON, but obviously this is only half the job: to be useful in practice, I have to give the same treatment to "obsoleteTech". But I wanted to check in and get your thoughts whenever you have a moment.

In particular, while the JSON needs to allow either a list or a single string for backwards compatibility, I am proposing to make the internal representation *always* a set of strings rather than a single string, even though the overwhelming supermajority of mods will never use this functionality. I think this prevents introducing additional complication to the codebase. I'd make an analogy to
>     var promotions = HashSet<String>()

The overwhelming majority of units get zero free promotions. Of the units who do get free promotions, the overwhelming majority get only one, such as Artillery getting Indirect Fire. Nevertheless the promotions are always handled as a set. Instead of sometimes having null and sometimes having a string and sometimes having a set of strings, we always have a set of strings. That makes most things much simpler to handle, since the set can naturally have size zero or one.

Since "requiredTech" is used in a lot of places, this branch has to touch a lot of files...but as you can see, we can often leave those checks looking *less* complicated, rather than more, even though we're handling more possibilities.

### The Ugly

So, we can have a single data model that seamlessly handles all cases, and use it everywhere. Buuuut...obviously the *actual extant JSON* for all the currently-existing mods has a single string for "requiredTech", not a list. To retain backwards compatibility, at *some* point that single string needs to be converted into a set of size one. What I'm proposing is to do that at the earliest possible step, when reading the JSON itself. That...technically works...but as you can see, it's by far the trickiest and ugliest part of the code.

There's actually two different ways to do that. My bias is very much toward the first --- admittedly, partly because I'm not actually sure how to do the second.

The first option is to call `json.readFields(this, jsonData)` first, building the BaseUnit *wrong*, and then fix it up after the fact. `json.readFields(this, jsonData)` ignores fields that it doesn't recognize; if it's expecting "requiredTechs" and it sees "requiredTech" then it just passes that by. A secondary conditional can then set "requiredTechs" to a set of size one.

The second option is to handle all of the children symmetrically. The big problem there is that every single field in https://yairm210.github.io/Unciv/Modders/Mod-file-structure/4-Unit-related-JSON-files/#unitsjson has to be individually listed and handled. I don't know any way to make a linter or anything check that all the fields are in fact handled. If you later add an additional field to the unit schema and don't also add it to the custom serializer, it'll result in a nastily hard-to-find bug. I don't love the first approach, but I think it's better than the second approach. Nevertheless, I included a big in BaseUnit.kt showing how the second approach could work (but it's incomplete, since I didn't actually figure out how to handle all the fields that way; of course, as the file currently stands, `json.readFields(this, jsonData)` is still called so everything in the "when" except "requiredTech" is redundant and unnecessary).


yairm210:
Multiple techs are possible with "only available when <[tech] is researched>" unique which is why no data model change was done so far

In general we prefer to move to uniques for all ew features due to simplicity composability etc

If there's something that "only available when <[tech] is researched>" does *not* provide I have no problem adding a unique that looks like "Requires [tech] to be researched", but changing single field to multifield is not the way
This is also how we enabled multiple resources

yairm210:
Steps to success:
- A function on the unit/building called "getRequiredTechs()", returns list[string], change usages of requiredTech to use that
- Add unique, add implementation to getRequiredTechs only

dHannasch:
Oh, huh. Wait, so if you were doing https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/models/ruleset/tech/Technology.kt#L13 from scratch today, you'd have no `var prerequisites = HashSet<String>()` and implement the relationships among techs with Uniques?

Wait, actually, sorry, I don't think I understand. Or rather, certainly once the required techs are in the uniques, then we can have a function BaseUnit.getRequiredTechs() that fetches them from there (or for that matter Technology.getRequiredTechs). But in terms of the Units.json, I'm...very unclear whether you're saying you don't want "requiredTechs" in the Units.json or not. When you mention uniques are better for composability (...composability of what? of Uniques?), are you talking about the internal representation, or the JSON?

Running with the assumption that you're on board with having "requiredTechs" in the Units.json (though I'm not actually confident of that), at some point they have to get parsed out of there and forwarded to the uniques, right? At what point are you picturing that happening? During the JSON parse?

In terms of obsoleteTech, are you picturing "Only available <before discovering [Chivalry]>"?

Hmm, should "upgradesTo" also be a Unique? Should "unitType" be a Unique? You mentioned composability; are you talking about things like having an Archery unique which in turn owns "Accuracy I promotion allowed", "Accuracy II promotion allowed", etc?

yairm210:
I'm saying I don't want requiredTechs in the json
Yes, if I was reimplementing objects today, almost everything would we in uniques, barring required fields (name) and description-related fields
This is line the tenth field we've needed to rethink since the start, and if they were uniques it would be barely an inconvenience

I'm not *deprecating* any existing fields, I can love with the duality, most mods don't need more that 1 tech or resource per object. But I'm not saying any *new* fields.

So why are uniques so great? They enable:
- Conditional-izing, and possible future other modifiers
- Descriptions, translations, built-in
- Depreciation, built-in
- Validation, built-i 

When I started, none of these existed, everything was ad hoc. 
Now we have systems for everything

SeventhM:
An aside, and this 100% is premature optimization, but doing something like putting it in a field does make it easier to cache the required techs... Hmmm

That gives me an unrelated idea... Currently, there's an issue with cached values for map units (such as whether it can move through terrain easier)... Why don't we cache the uniques themselves? I doubt something like `unit.mapTerrainCostUniques.any { it.conditionalApplies() } &&` would be significantly more expensive and it's probably uncommon enough that there's duplicates

SeventhM:
Oh, also, your op uses the mod Deciv as an example. While the idea is nice for other mods, this is not the actual issue in regards to that mod's warnings. Many of the warnings are for units on the same tech as the thing their replacing, or genuinely is for weird tech tree design decisions that may be should have been rethought. While a unique to cover units requiring multiple techs is probably an important thing to support, your example does not follow correctly

SeventhM:
> Already do - but for movement,

No, we cache flags for uniques, not the subset of uniques itself. Flags don't update... Wait, nevermind, Idk why I assumed `UnitMovement.moveToTile` didn't call `unit.updateUniques`. I'm probably thinking of tech/policy unlocks that doesn't call `unit.updateUniques`

dHannasch:
Okay, required techs are done with Uniques now. Everything *seems* to work, though I'm nervous about logic errors in the `requiredTechs()` function cropping up in edge cases. A HashSet is *really hard to screw up*, a search through Uniques for matching conditionals is easy to screw up.

archerOnlyAvailable
archeryStillShowsArcher

I didn't hear any approval for the `json.readFields`-then-fix-up method, but I didn't hear anything against it either, so I deleted the other approach. Parsing out requiredTech is simplified using @SomeTroglodyte's `getString` method. Admittedly, this is slightly less efficient, since it'll have to search through the entire linked list for each of {requiredTech, obsoleteTech, requiredResource, upgradesTo}, but I don't think efficiency of the JSON parsing will ever matter, since it's only done once. (By contrast, I'm not sure how many times `requiredTechs()` might be called.)

Giving the same treatment to obsoleteTech looks fairly straightforward. requiredResource doesn't look too bad, with the caveat that in that case it's very important to make sure to remove all traces of the old non-Unique-based method since it would be very bad to accidentally double resource consumption in some edge case. upgradesTo needs a new Unique. (I mean, and then of course to actually make it do something, replace `var upgradesTo` with `fun upgradesTo()` where `fun upgradesTo()` references the Unique.)

dHannasch:
> Huh? I thought no json changes thus no implements Json.Serializable? Too tired for qualified inspection.

I...don't think I understand the question? In both the old version and the new version, literally the entire point of customizing the JSON parsing was to *not* change the JSON (despite the internals changing).

dHannasch:
I changed all references to `obsoleteTech` to point to a function that checks for "Only available <before discovering [$obsoleteTech]>" Uniques. The JSON remains the same, by way of the custom JSON parser checking for "obsoleteTech" and converting it to a "Only available <before discovering [$obsoleteTech]>" Unique.

archerOnlyAvailable

core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
```diff
@@ -37,9 +43,89 @@ interface IHasUniques : INamed {
     fun getMatchingUniques(uniqueType: UniqueType, stateForConditionals: StateForConditionals? = null) =
         getMatchingUniques(uniqueType.placeholderText, stateForConditionals)
 
+    fun requiredTechs(): Sequence<String> {
+        val uniquesForWhenThisIsAvailable: Sequence<Unique> = getMatchingUniques(UniqueType.OnlyAvailableWhen, StateForConditionals.IgnoreConditionals)
+        if (uniquesForWhenThisIsAvailable.none() && uniques.any{ it.contains("Only available <after discovering [") })
```
SeventhM, line 48: Wait... What? There's a simpler way to find conditionals. Uniques have a `conditionals` field so you can just do ``&& uniquesForAvailable.conditionals.any { it.isOfType(UniqueType.ConditionalTech }``

SeventhM:
Also I hate to be rude, but either I'm really dense or the added code in IHasUniques looks unreadable. Some of this string catenation for the exceptions should minimum be split into more lines imo, if not cleaned up entirely

Why are we throwing again? Either I'm reading correctly and that code should never run, or I'm reading incorrectly, and either something really needs to go bad for the code to run

yairm210:
I appreciate the work that went into this, but this PR is now clearly in Too Much Going On territory.

You're trying to:
- Change the way we handle required tech
- Change the way we handle obsolete tech
- Change json serialization (why...?)

Let's break this down into different PRs


PR 1: A function on the unit/building called "getRequiredTechs()", returns list[string], change usages of requiredTech to use that
This should be *the current requiredTech* PLUS "only available <when X>"

The bulk of the work is moving from X.requiredTech to x.getRequiredTechs()


PR 2: A function on the unit/building called "getObsoletingTechs()", returns list[string], change usages of obsoleteTech to use that
Same as above


We should *not* be changing json parsing, at all, instead we should retain the 'single value' field and use that as the basis for the functions above, as we do in getResourceRequirementsPerTurn, for example

SeventhM:
Also, I think you missed my point regarding IHasUniques

Generally speaking, we usually should never have a need to look at the text of a unique. Most values are stored in their own places, and the process of reading the uniques themselves (as well as RulesetValidator.kt or UniqueValidator.kt) should pick up any errors from the ruleset, not the objects themselves

Slightly inaccurate versions of uniques are valid and can have merit if a player wants to simply display information. Idk why they would use the conditional style of text for it, but it is possible. If these things wasn't valid, they should be caught by UniqueValidator. If they wasn't, they should be treated as valid. They should never throw an exception in this case


Slightly irrelevant question: why is this all being done in IHasUniques, and not, say, IConstruction or BaseUnit, or Building? While there are concerns I've brought up in the past about unnecessary code duplication between Building and BaseUnit, I'd imagine the place to centralize this is IConstruction, not IHasUniques. It seems like a bit beyond necessary (and probably less flexible) to move things this far from where they started

dHannasch:
> Let's break this down into different PRs

I mean, sure, yes, but I was sort of looking for feedback on the overall approach.

And, I mean, you definitely gave that. So, thank you. But I have no idea what it means in practice.

> the current requiredTech PLUS "only available "

In retrospect, I guess that's a possible interpretation of what you said. I must admit I did not think of that as an option. And that can certainly be done. But that sort of seems like it implies things about what's happening elsewhere in the codebase?

> change usages of requiredTech to use that
> The bulk of the work is moving from X.requiredTech to x.getRequiredTechs()

If we're retaining `requiredTech`, and we're getting rid of the usages of `requiredTech` and replacing them with function calls, then we're doing that to...all of them? Some of them?

In your desired end-state, is `requiredTech` a stub that's never used outside the function that intermingles it with the Uniques? Or are there two parallel systems threaded through the infrastructure, with some places still using `requiredTech`?

You mention `getResourceRequirementsPerTurn()`; `requiredResource` is used in rather a lot of places besides the internals of `getResourceRequirementsPerTurn()`. Presumably you're picturing something similar for `requiredTech`, but I don't know how you decided where `requiredResource` should be used versus where `getResourceRequirementsPerTurn()` should be used.

Are you okay with a comment on `requiredTech` warning future developers not to use it for anything? Or on `requiredResource`, for that matter? I thought that was your intent when you originally rejected changing the type from a single String to a set like `promotions`; that you didn't want any new development to further develop that field, even if it wasn't increasing the number of places in the codebase referencing that field. But maybe I was wrong about your original objection.

SeventhM:
> You mention `getResourceRequirementsPerTurn()`; `requiredResource` is used in rather a lot of places besides the internals of `getResourceRequirementsPerTurn()`. Presumably you're picturing something similar for `requiredTech`, but I don't know how you decided where `requiredResource` should be used versus where `getResourceRequirementsPerTurn()` should be used.

I edited this up recently, so while I'm not at a computer to double check this, I can give an answer:

The field, `requiredResource`, is not directly used in any part of the code base except in places where it needs a description alongside other uniques. It is instead used in `getResourceRequirementsPerTurn` to make it easier for any function to pick up the information from IConstruction

The function, `requiredResource(String, StateForConditionals)` is a simplified helper function for `getResourceRewuirementsPerTurn` for anything that needs the resource requirements of an object, but does not care how many of each resource is needed. I forget where this is actually used (again, not at a computer where Studio makes searching easier), but it is helpful to not need to write out the logic itself each time you just want the simplified information

dHannasch:
> Slightly inaccurate versions of uniques are valid and can have merit if a player wants to simply display information. Idk why they would use the conditional style of text for it, but it is possible.

Ohhhh. I wanted to have a sanity check to ensure that the logic would never silently fail to register a prerequisite (that it would fail *loudly* instead), but if "Only available <after discovering [ I'm not even going to close the square brackets}" would technically be a valid Unique, then maybe no such sanity check is possible.

SeventhM:
> I'm not even going to close the square brackets

Would crash when reading already. And if it didn't, it would fail to pass UniqueValidator unless you closed the bracket and filled the parameter with something valid. Such a crash is already controlled for

dHannasch:
> places where it needs a description alongside other uniques

You're speaking of e.g. https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/objectdescriptions/BaseUnitDescriptions.kt#L108 and https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/ui/objectdescriptions/BuildingDescriptions.kt#L179, right?

yairm210:
Hm
We should probably unify that as well
For another day

SeventhM:
> You're speaking of

Yes

dHannasch:
> Slightly irrelevant question: why is this all being done in IHasUniques, and not, say, IConstruction or BaseUnit, or Building? While there are concerns I've brought up in the past about unnecessary code duplication between Building and BaseUnit, I'd imagine the place to centralize this is IConstruction, not IHasUniques.

OnlyAvailableWhen can apply to lots of things besides units and buildings.

>     OnlyAvailableWhen("Only available", UniqueTarget.Unit, UniqueTarget.Building, UniqueTarget.Improvement, UniqueTarget.Policy, UniqueTarget.Tech, UniqueTarget.Promotion, UniqueTarget.Ruins, UniqueTarget.FollowerBelief, UniqueTarget.FounderBelief),

I'm not planning to do this, like, *now*, but given that OnlyAvailableWhen is the way to do tech prerequisites, I was sort of low-key thinking of IHasUniques as the place to start centralizing everything that might ever want tech prerequisites. E.g., there's also TileImprovement.techRequired.

https://github.com/dHannasch/Unciv/blob/resolve-warning-horseman-obsoletes-at-tech-rifling/core/src/com/unciv/ui/objectdescriptions/TechnologyDescriptions.kt
We can replace the variable reference with a function call not only for units and buildings but for tile improvements and whatever else.

I figured might as well put the capability in IHasUniques. I mean, even if we don't want to unify everything immediately, that's ultimately where we'd like to be living, right?

> if you were doing https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/models/ruleset/tech/Technology.kt#L13 from scratch today, you'd have no var prerequisites = HashSet<String>() and implement the relationships among techs with Uniques?
> Yes, if I was reimplementing objects today, almost everything would we in uniques, barring required fields (name) and description-related fields

SeventhM:
> and soon all untyped mod Uniques can move to red flags

Red seems harsh for a non crash situation, even if we get 100% full support for comment uniques. Knowing from discord, or even just how often Deciv gets brought up, going from options only warning to just a warning seems good enough. Also, since this discussion is of a conditional, which Idk if it can support conditionals themselves, upping to red would still be covering situations that could potentially have a valid usecase



> I figured might as well put the capability in IHasUniques. I mean, even if we don't want to unify everything immediately, that's ultimately where we'd like to be living, right?

I'mma let someone else give their opinion on this, but my two sense is while this is valid, I'm not sure how sensible it is. E.g., not everything has an era field like policies, not everything has an obsoletes field etc. Imo, this should minimum be split into at least 2 different PRs: one to simply migrate to using uniques and combining information from the legacy fields, and one more focused on unifying behavior where it makes sense after the fact

core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -39,14 +39,18 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
...
+    // https://yairm210.github.io/Unciv/Developers/Translations%2C-mods%2C-and-modding-freedom-in-Open-Source#filters
     var requiredResource: String? = null
 
     override fun getUniqueTarget() = UniqueTarget.Unit
 
     var replacementTextForUniques = ""
     var promotions = HashSet<String>()
-    var obsoleteTech: String? = null
+    // Warning: future development should not increase the role of obsoleteTech, and should reduce it when possible.
+    // The Unique "Only available before discovering [$obsoleteTech]" accomplishes the same task.
```
SeventhM, line 52: Not sure this is true. There is mechanic baggage regarding obsoleting objects, at least in the form of production switching to what units upgrade into

This assumption is unnecessary. `UniqueType.ObsoleteWith` covers obsoleting functionality. If you're migrating fields to uniques, that UniqueType should be included in the discussion

SeventhM:
> but given that OnlyAvailableWhen is the way to do tech prerequisites

Actually... Hmm...

Should it? I feel like that would have enough baggage, especially given the system of conditions would make it harder to check if it should be valid. Kinda feel like if a mod dev wants it to show as requiring a tech, we should just have a unique for this to be explicit, not implicit

SeventhM:
> > You should know by now

You should know by now I often forget where I set my sense of humor down at. I know it's around here somewhere...

dHannasch:
> not everything has an era field like policies, not everything has an obsoletes field etc.

Is your point that not everything has tech prerequisites?

Currently only buildings, units, tile improvements, and technologies have tech prerequisites. But, I mean, there's no reason in principle that policies and promotions couldn't have individual tech prerequisites. (In ye olden dayes, policies *did* have individual tech prerequisites, see e.g. https://civilization.fandom.com/wiki/Democracy_(advance)_(Civ1) https://civilization.fandom.com/wiki/Democracy_(technology)_(Civ4).) As I understand it, OnlyAvailableWhen already supports them for the actual functionality. If IHasUniques has the capability to list out tech prerequisites, then we'd be a better position to allow for such tech prerequisites. I haven't looked into it, but I *think* it would be as simple as adding another line to TechnologyDescriptions. If OnlyAvailableWhen already covers the actual functionality, that would be the only thing needed to make tech prereqs for policies just as seamlessly transparent as for units and buildings.

dHannasch:
> the system of conditions would make it harder to check if it should be valid

Can you elaborate? What does this mean? What would be different if there were a specific unique OnlyAvailableWhenTechIsDiscovered?

SeventhM:
> As I understand it, OnlyAvailableWhen already supports them for the actual functionality

Ok, sure, but does this cover some of the baggage that the previous fields cover? Is there truly a benefit to including `OnlyAvailableWhen` with a resource conditional as if it's equivalent to `ConsumesResource`? Tbh, Idk, that's why I'm deferring judgement to someone else and just throwing my opinion in the ring. But it seems to me that `OnlyAvailableWith` isn't as flexible as a dedicated unique, and it relies on modders making implicit assumptions on how things work and not explicitly giving tools to modders. And it seems to me that it is possible we come across some baggage that makes it not be worth it to center around IHasUniques

> What would be different if there were a specific unique OnlyAvailableWhenTechIsDiscovered?

`OnlyAvaliableWhen` cannot check its conditionals without being gotten via `IgnoreConditionals`. A dedicated unique can. E.g., 
```Only available <after founding a religion> <before discovering [Theology]>```
and
```Requires the [Theology] tech <before founding a religion>```
has lingual differences and implies the intent may be different. However, the former cannot be filtered correctly unless we add extra tricks to the conditionals system or unless we manually check `conditionalApplies` while the latter can. And you can't change the former to match the latter in wording because
```Only available <before founding a religion> <after discovering [Theology]>```
is functionally different

dHannasch:
> including `OnlyAvailableWhen` with a resource conditional as if it's equivalent to `ConsumesResource`

I...don't think anyone suggested anything remotely similar to that? Did I miss something? We already have `ConsumesResources` as its own Unique. And `ConsumesResources` has nothing to do with tech prerequisites.

What does that have to do with tech prerequisites or IConstruction or IHasUniques?

The thing we're talking about is whether the function to fetch tech prerequisites should be added to IConstruction or added to IHasUniques, right?

dHannasch:
> Only available <after founding a religion> <before discovering [Theology]>

As far as having an extra Unique, though, I think I see what you're getting at. Yes, it would be more flexible to allow different boolean combinations of conditions. As I understand it, all the conditionals in OnlyAvailableWhen are implicitly ANDed, right? They must all be satisfied. There's no way for a modder to say that something should be available if *either* Theology is discovered *or* a religion is founded.

SeventhM:
> I...don't think anyone suggested anything remotely similar to that?

Using `Only available <before discovering [tech]>` over `Obsolete with [tech]` or `Only available <after discovering [tech]>` over a dedicated unique is similar to using `Only available <with [resource]>`. There's some baggage I forgot about (so maybe... wrong example), but[ I genuinely have used `ConsumesResources` intentionally over `OnlyAvalableWhen`](https://github.com/ravignir/RekMOD/blob/877df4502cc7b2cc33ac2e53eb626a4e0287f6bc/jsons/Buildings.json#L520C8-L520C8) to have something show in the construction menu despite not actually consuming the resource

> The thing we're talking about is whether the function to fetch tech prerequisites should be added to IConstruction or added to IHasUniques, right?

To be fair, there 2 discussions here: One regarding dedicated uniques, and one regarding where logic should be centralized. Again, these two discussions should be separate PRs

yairm210:
Closing as Done :)

PR #10569, opened 2023-11-24
(UX) Friendlier Language choosers

SomeTroglodyte:
Idea:
- Tweak Language Pickers to scroll the selected one into view when appropriate
- Allow selection with letter keys, by english name's first letter, round-robin, so e.g. 'R' cycles Russian and Romanian...

... 90% success, but:
* Thanks to reloadWorldAndOptions, the round-robin on the Options page is limited to the first two
* Scroll-to-view on the one-time LanguagePicker compromises: Too lazy to search why the Layout of LanguagePickerScreen has the individual LanguageTable's local coordinate system mismatch that of the ScrollPane - centering made sure visibility is achieved anyway, if a little north of the center
* Far more code than I imagined

### DIFF
--- a/core/src/com/unciv/ui/components/extensions/Scene2dExtensions.kt
+++ b/core/src/com/unciv/ui/components/extensions/Scene2dExtensions.kt
@@ -19,6 +19,7 @@ import com.badlogic.gdx.scenes.scene2d.ui.CheckBox
 import com.badlogic.gdx.scenes.scene2d.ui.Image
 import com.badlogic.gdx.scenes.scene2d.ui.ImageButton
 import com.badlogic.gdx.scenes.scene2d.ui.Label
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/ui/components/extensions/Scene2dExtensions.kt
```diff
@@ -130,6 +131,11 @@ fun Actor.getAscendant(predicate: (Actor) -> Boolean): Actor? {
     return null
 }
 
+/** Gets the nearest parent of this actor that is a [T], or null if none of its parents is of that type. */
+inline fun <reified T> Actor.getAscendant(): T? {
```
yairm210, line 135: Nice! Actually saw a use of reified in production code last week for exactly this kind of behavior - 'cast as T' - so it's nice to see it again here

yairm210 (approved):

PR #10575, opened 2023-11-25
Comment unique

SomeTroglodyte:
Sic... Allows non-filtering pass-through text in `uniques` - as long as it doesn't break the brackets parser.
Since pass-through is the entire point, No need to bother translators with that line, all included.

Adds extra documentation capability to inject comments into uniques.md - which, being entirely generated, could do with hints of this kind in other Uniques, too. I'm not sure the format agrees with that preprocessor you're using, untested!

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -9,7 +9,12 @@ import com.unciv.models.translations.getPlaceholderText
 // I didn't put this in a companion object because APPARENTLY doing that means you can't use it in the init function.
 private val numberRegex = Regex("\\d+$") // Any number of trailing digits
 
-enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags: List<UniqueFlag> = emptyList()) {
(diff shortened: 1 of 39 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -9,7 +9,12 @@ import com.unciv.models.translations.getPlaceholderText
 // I didn't put this in a companion object because APPARENTLY doing that means you can't use it in the init function.
 private val numberRegex = Regex("\\d+$") // Any number of trailing digits
 
-enum class UniqueType(val text: String, vararg targets: UniqueTarget, val flags: List<UniqueFlag> = emptyList()) {
+enum class UniqueType(
+    val text: String,
+    vararg targets: UniqueTarget,
+    val flags: List<UniqueFlag> = emptyList(),
+    val docDescription: String? = null
```
yairm210, line 16: Hmm, I see myself adding lots of these, every time there's a modder question about a speciific unique that's unanswered in current docs, good work!

yairm210 (approved):

PR #10580, opened 2023-11-25
Wesnoth map import polished up

SomeTroglodyte:
This got lost somehow.. Was ready shortly after the merge of the initial version.
* Some rules tweaks to better match some Wesnoth maps
* Fallback base terrain un-hardcoded
* Allow for entire spec as defined, especially scenario markers that would prevent import before
* Code better commented and a certain simplification


yairm210 (approved):

yairm210:
release patch

PR #10582, opened 2023-11-26
(CQ) Linting and comments on TileSetStrings and ImageAttempter

SomeTroglodyte:
... and I need to stop and get this out before going further. (Edit: not functional changes, CQ tag just invented for code quality/readability. Even the run/invoke swap is neutral - invoke means more direct, run may or may not be optimized away)

### Big (or tiny) Questions
- ImageAttempter chaining could do with a hard think: Can it be done with less memory allocations, but without changing the signature? Allocating a new one on each failing step seems wasteful.
- tryGetUnitImageLocation could/should use _one_ ImageAttempter chain.
- tryGetUnitImageLocation - baseUnit.replaces - isn't that too limited? That means if it's nation-specific and there's no specific image, images of the replaced unit will go in but without all the era and style alternatives?
- The fallback mechanism could use ImageAttempter chaining instead of Elvis for extensibility - actually Elvis is more performant at the moment
- No synergy between getUnitImageLocation and getOwnedTileImageLocation?
- getUnitImageLocation - using full 4- or 5- letter true/false in the embarked part of the key - could save a few bytes using 0/1

### DIFF
--- a/core/src/com/unciv/ui/components/tilegroups/TileSetStrings.kt
+++ b/core/src/com/unciv/ui/components/tilegroups/TileSetStrings.kt
@@ -179,7 +179,7 @@ class TileSetStrings(
         )
         // if in cache return that
         val currentImageMapping = imageParamsToImageLocation[imageKey]
-        if (currentImageMapping!=null) return currentImageMapping
+        if (currentImageMapping != null) return currentImageMapping
 
         val imageLocation = tryGetUnitImageLocation(unit)
             ?: fallback?.tryGetUnitImageLocation(unit)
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/components/tilegroups/TileSetStrings.kt
```diff
@@ -90,18 +106,18 @@ class TileSetStrings(tileSet: String = UncivGame.Current.settings.tileSet, unitS
     fun orFallback(image: String, fallbackImage: TileSetStrings.() -> String): String {
         return if (fallback == null || ImageGetter.imageExists(image))
             image
-        else fallback!!.run(fallbackImage)
+        else fallbackImage.invoke(fallback!!)
```
yairm210, line 109: Not sure about this, why this change?
And the onFallback below

yairm210: Fair enough 👍🏿

yairm210: I honestly do not recall, it's possible that the tile and unit fallbacks need redoing

yairm210 (approved):

Issue #10583, opened 2023-11-26
CityFounder places then removes the City center improvement?


SomeTroglodyte:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

current master from source

### Describe the bug

Trying a mod I missed a City center -
[this](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/logic/city/managers/CityFounder.kt#L18) places one
[and this](https://github.com/yairm210/Unciv/blob/master/core/src/com/unciv/logic/city/managers/CityFounder.kt#L59) removes it right away

... what has changed? Is the improvement supposed to come later from some trigger, so is L17+18 obsolete? If so, the first autoAssignPopulation is operating under false circumstances.

### Steps to Reproduce

New game, reset to defaults, set base ruleset to "The Star Wars Special Edition", start, found.

### Expected Behavior

The big Question Mark graphic for the city center - nope, just base terrain

### Screenshots

![image](https://github.com/yairm210/Unciv/assets/63000004/f52111d8-680b-47dc-b179-5489f9c4aaf9)

### Operating System

Linux

### Additional Information

.. haven't founded a city in other recent gameplay/testing, so no idea if this is a recent overall regression or not - yet - but I need to pause so I'll leave this reminder.

SeventhM:
Oof, wrong order. I see what happened. Before, it changed the improvement after setting up the city (it's just an invisible city for a little bit). https://github.com/yairm210/Unciv/commit/c75e7f9407ec26b9be05810b2f6376d3ea1df87e accidentally moved it to before the city is set up

Best case: line 59-60 is obsolete and should be removed

Likely case: the improvement change needs to happen later

SeventhM:
> Yet still not done with this

You sure?

Even on a mod with a big question mark, running patch 2 on my phone works just fine. It's possible you have a mod that didn't have a "city center" improvement, but I don't see how things would have been different before or after the changes

SeventhM:
> The other one

Again, this one, I'm not sure what you are referring to or what could've changed to affect this. Post hotfix, vanilla and mods with vanilla eras have functioning HexaRealm sprites, just like before the commit. And issues with the big question mark are the same as it's ever been

PR #10584, opened 2023-11-26
Fix founding cities removing city center tile improvement

SeventhM:
Not at computer, so no tests done. Simple enough hotfix, though

Fun fact: we are assigning population 3 separate times now. Not sure that's intentional

### DIFF
--- a/core/src/com/unciv/logic/city/managers/CityFounder.kt
+++ b/core/src/com/unciv/logic/city/managers/CityFounder.kt
@@ -14,8 +14,6 @@ import com.unciv.models.ruleset.unique.UniqueType
 
 class CityFounder {
     fun foundCity(civInfo: Civilization, cityLocation: Vector2): City {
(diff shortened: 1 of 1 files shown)


SeventhM:
Oh, and closes #10583. This is immediately noticable (was reported on discord before Trog noticed), so leaving it open temporarily

SeventhM:
> addStartingBuildings and expansion.reset, and with autoAssignPopulation

Wait a second... expansion.reset is more or less 6 autoAssign calls by itself... Nope, I'm not testing this right now. Find me some other day, this nonsense is out of scope. This PR is a hotfix and nothing more. These discussions are irrelevant 

> there must have been a reason...

Fun fact: TestGame.addCity \*doesn't\* set the tile to a city center. Surprised we haven't accidentally found this out

yairm210 (approved):

PR #10585, opened 2023-11-26
Wrap INonPerpetualConstruction.requiredTech into function

dHannasch:
This is Step Zero, not changing any functionality. The function in this branch is a stub function that doesn't pull from the Uniques.

This doesn't touch the more complicated usages yet since those should use the new era functions that this branch does not include. So this doesn't remove all references to `requiredTech`, but it does put a deprecated warning on it to discourage adding new usages.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
@@ -51,7 +51,7 @@ class CityStateFunctions(val civInfo: Civilization) {
 
         // Unique unit for militaristic city-states
         if (uniqueTypes.contains(UniqueType.CityStateMilitaryUnits)) {
-            val possibleUnits = ruleset.units.values.filter { it.requiredTech != null
+            val possibleUnits = ruleset.units.values.filter { it.requiredTechs().any()
                 && ruleset.eras[ruleset.technologies[it.requiredTech!!]!!.era()]!!.eraNumber > ruleset.eras[startingEra]!!.eraNumber // Not from the start era or before
                 && it.uniqueTo != null && it.uniqueTo in unusedMajorCivs // Must be from a major civ not in the game
                 && ruleset.unitTypes[it.unitType]!!.isLandUnit() && ( it.strength > 0 || it.rangedStrength > 0 ) } // Must be a land military unit
--- a/core/src/com/unciv/models/ruleset/Building.kt
+++ b/core/src/com/unciv/models/ruleset/Building.kt
@@ -22,6 +22,7 @@ import com.unciv.ui.objectdescriptions.BuildingDescriptions
 
 class Building : RulesetStatsObject(), INonPerpetualConstruction {
 
+    @Deprecated
     override var requiredTech: String? = null
     override var cost: Int = -1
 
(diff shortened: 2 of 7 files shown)


SeventhM:
Request: I still think extra required techs should be from a new `UniqueType.RequiredTech` unique. So let's wait on Depreciation messages (assuming that's the right approach, which I doubt it is tbh), let's wait to add that unique

yairm210 (approved):
Looks good!

yairm210:
@SeventhM I agree, because then we can have *conditional dependency*
"Requires [tech] to be researched <if policy not adopted>", is something that "only available when" cannot mimic

yairm210:
Since this is a requirement for the next stages, I'll merge it now so we can start taking about the next step

dHannasch:
I agree that e.g. "This policy cannot be adopted unless *either* Theology is discovered *or* a religion is founded" is a useful capability. But I don't think making a special Unique just for requiring a tech is the right way to do that. Because unless I'm missing something, there's nothing special about requiring a tech. You could just as easily want "This policy cannot be adopted unless *either* Stonehenge is built *or* a religion is founded." I think the right way to allow that is to allow disjunctions in conditionals *period*, not specifically for tech requirements.

(As for how to do that, this is totally spitballing, but there could be a special Disjunction conditional that has a sequence of sub-conditionals. I haven't looked into how hard that would be to do.)

SeventhM:
> The cost comes _cumulatively_

You're trying to make this big point about number of uniques and bloat, but I don't see where I'm supposed to care. While you've been trying to avoid extra uniques here, 10591 and 10597 opened trying to add more uniques. #3242 is constantly requesting more uniques or slight variants of existing uniques. On the discord we're constantly having new things that mods want uniques for. And that's not even talking stuff we're missing for espionage or for Brave New World to function, or a trigger conditional for building improvements. The list of uniques I've cataloged that I have to add probably goes over an additional 10, and is nothing to say that others figures out more to add

At what point is this a moot point? At what point do you stop, take a step back, and realize we've already hit that point of "the list is extremely long" and realize that nobody cares. If uniques are forgotten, it'll probably be remembered as someone on the discord starts experimenting on the game (as there's no shortage of experimental mods that have little interest in function). If there was a poll on discord that showed people thought the list of uniques was too long, maybe I'd think you're onto something. But frankly that's just not the case

> Because there's nothing special about `Only available <after [Theology] is researched>`.

Let me make the argument as crystal clear as I can: I don't care about if there's something special about a `RequiredTech` unique. My primary argument is about feature parity. You're saying that `Only available <after [Theology] is researched>` can replace a dedicated unique? Fine, then find a way to do so. Don't like my prior example? Then let's do
```
Requires the [Theology] tech <before founding a religion> <for [civFilter]>
```
There's both an "and" (the 2 conditionals in the one unique) and what you read as an "or" (For this civ, it's available either after founding or after getting the tech). Can you get ``OnlyAvailableWhen`` to do that? That's not a unreasonable question, it's just simply asking, "if ``OnlyAvailableWhen`` truly can replace dedicated uniques (assuming there's no mechanical baggage, which... you actually might've found by accident in #10595), then it should replace it fully, not halfway". That would be the reasonable assumption for such a claim, afterall.

> Why not `Only available <if [Shrine] is constructed>`

This exists already in the form of `RequiresBuildingInSomeCities` when param[1] == 1. You have the wrong conditional. You're looking for `Only available <in cities with a [Shrine]>`

> UniqueType.ConditionalDisjunction

On further examination, doesn't this not work unless you change getMatchingUniques to look for this conditional first? Or unless the end function is explicitly checking for this conditional? I'm pretty sure if you'd tried to use this in an actual function, it'll be a lot more complicated than you think

PR #10586, opened 2023-11-26
Fixed AI exploration

tuvus:
Fixes a tiny bug in  #10481.

The previous code checks for if there is an enemy unit at a distance 3 instead of if there isn't an enemy unit at a distance of 3 when looking for a good tile to explore.

The discussion about the issue is on discord.


tuvus:
Yes, this is exactly it.

tuvus:
There are no enemies within a distance of 3 to all tiles it wants to explore, so it has finished exploring with the bugged code.

tuvus:
FYI, automated units can still fog-bust, but they won't explore.

PR #10587, opened 2023-11-26
Add iHasUniques.requiredTechs

dHannasch:
This doesn't touch any usages, just drops all the techs from the Uniques into the function.

This shouldn't cause any problems, because https://github.com/yairm210/Unciv/pull/10585 didn't change any of the more complicated usages where dropping in the techs from the Uniques would cause problems.


core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
```diff
@@ -42,4 +42,15 @@ interface IHasUniques : INamed {
 
     fun hasUnique(uniqueType: UniqueType, stateForConditionals: StateForConditionals? = null) =
         getMatchingUniques(uniqueType.placeholderText, stateForConditionals).any()
+
+    fun techsRequiredByUniques(): Sequence<String> {
+        val uniquesForWhenThisIsAvailable: Sequence<Unique> = getMatchingUniques(UniqueType.OnlyAvailableWhen, StateForConditionals.IgnoreConditionals)
```
yairm210, line 47: I think this would look more readable as a chain, eg a.b{}.c{}
What do you think?

dHannasch: Better?

I want to keep the listing of the Uniques separate, since there's some fiddly technical detail happening there (fiddly to me, anyway) in handling the StateForConditionals, but the rest looks simple as a chain.

PR #10591, opened 2023-11-26
Stats per Stat unique

PLynx01:
Syntax:
[stats] per [amount] [stat]

Example: 
[+1 Gold] per [100] [Gold]

In that case, if player has 1000 gold, he/she gets additonal 10 gold per turn. If the number of stockpiled gold is 5000, the bonus is 50 gold per turn. If the player has less than 100 gold, there is no bonus.

Why?
This unique can be used by modders to give players the additional bonus for accumulating a large amount of specified stat. The most obvious example is the compound interest.

If you have any suggestions, please share your thoughts in comments

### DIFF
--- a/core/src/com/unciv/logic/civilization/transients/CivInfoStatsForNextTurn.kt
+++ b/core/src/com/unciv/logic/civilization/transients/CivInfoStatsForNextTurn.kt
@@ -311,6 +311,11 @@ class CivInfoStatsForNextTurn(val civInfo: Civilization) {
             if (unique.sourceObjectType != UniqueTarget.Building && unique.sourceObjectType != UniqueTarget.Wonder)
                 statMap.add(unique.sourceObjectType!!.name, unique.stats)
 
+        for (unique in civInfo.getMatchingUniques(UniqueType.StatsPerStat)) {
+            val amount = civInfo.getStatReserve(Stat.valueOf(unique.params[2])) / unique.params[1].toInt()
+            statMap.add("Stats", unique.stats.times(amount))
+        }
+
         val statsPerNaturalWonder = Stats(happiness = 1f)
 
         for (unique in civInfo.getMatchingUniques(UniqueType.StatsFromNaturalWonders))
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -28,7 +28,7 @@ enum class UniqueType(
     StatsFromSpecialist("[stats] from every specialist [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
     StatsPerPopulation("[stats] per [amount] population [cityFilter]", UniqueTarget.Global, UniqueTarget.FollowerBelief),
     StatsPerPolicies("[stats] per [amount] social policies adopted", UniqueTarget.Global),
-    StatsPerStat("[stats] per [amount] [stat]", UniqueTarget.Global),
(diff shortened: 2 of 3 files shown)


dHannasch:
> StatPercentBonus should already cover the usecase?

I don't think so? I mean..."[+1]% [Gold]" is a confusingly ambiguous and I'm not sure from the description what it does. But from inspecting the code, it adds 1% to each turn's income of Gold.

dHannasch:
@PLynx01 Might I suggest adding the words "at end of turn" or "at start of turn" or similar? That's probably what most people would assume anyway, but less ambiguity is better.

PLynx01:
> @PLynx01 Might I suggest adding the words "at end of turn" or "at start of turn" or similar? That's probably what most people would assume anyway, but less ambiguity is better.

I don't know which of these two to choose, because it's the bonus is actually added **between** turns.

SeventhM:
> StatPercentBonus should already cover the usecase?

Weirdly, technically no. That affects amount of producing resource, not amount of stockpiled resource

As for what's implemented, hmm... What about something like ``[relativeAmount]% [Stat] for current total [civWideStat]``

PLynx01:
> As for what's implemented, hmm... What about something like ``[relativeAmount]% [Stat] for current total [civWideStat]``?

As I said earlier, my idea is about compounding the stockpiled stats.

I think the percentage from stockpiled resources is more understandable than "[Stats] per [amount] [Stat]" unique.

Which should we implement?
Let the @yairm210 decide.

yairm210:
I am against percentages in general, where we can have not-percentages.
We need to specify that this is for *civ-wide stats* because this looks like something that can e.g. double food in a city.
The bonus is given at end-of-turn but I don't think it needs specifying.

A good question is how do we tell the player that this enters the civ-wide stockpile and not a per-city one? Do we have any existing uniques that do something like that?

PLynx01:
> I am against percentages in general, where we can have not-percentages.
> We need to specify that this is for *civ-wide stats* because this looks like something that can e.g. double food in a city.
> The bonus is given at end-of-turn but I don't think it needs specifying.
> 
> A good question is how do we tell the player that this enters the civ-wide stockpile and not a per-city one? Do we have any existing uniques that do something like that?

What about this:
[Stats] per every [amount] stockpiled [stat] civ-wide.

As far as I know, the following stats are stockpileable civ-wide:

- Gold
- Culture
- Science
- Faith
- Happiness

I believe that in most cases the Gold will be used as a last argument in the unique above.

yairm210:
> > I am against percentages in general, where we can have not-percentages.
> > We need to specify that this is for _civ-wide stats_ because this looks like something that can e.g. double food in a city.
> > The bonus is given at end-of-turn but I don't think it needs specifying.
> > A good question is how do we tell the player that this enters the civ-wide stockpile and not a per-city one? Do we have any existing uniques that do something like that?
> 
> What about this: [Stats] per every [amount] stockpiled [stat] civ-wide.
> 
> As far as I know, the following stats are stockpileable civ-wide:
> 
> * Gold
> * Culture
> * Science
> * Faith
> * Happiness
> 
> I believe that in most cases the Gold will be used as a last argument in the unique above.

That looks good, remember that the [stat] should be a [civWideStat] instead

PLynx01:
@yairm210 
Is this PR ready to merge?

yairm210:
Please *stop* asking that every time.
I can see when changes are made, it's just pushy and annoying.

yairm210 (approved):

PLynx01:
> Please *stop* asking that every time.
> I can see when changes are made, it's just pushy and annoying.

OK, if you don't want it, I'll stop doing that.

Issue #10602, opened 2023-11-27
Game crash on treaty end notification


Cwpute:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.9.0

### Describe the bug

https://notebin.de/?2c5bf494dd924444#5whViwJiKsczdXwtuuN1C567j4HAxHLC5yMN8M75auU5

### Steps to Reproduce

1. Online multiplayer game running for 207 turns
2. I end my turn
3. Waiting for the other player's turn, browse the notifications
4. Tap on a notification for a free-circulation treaty that just ended
5. crash

### Expected Behavior

shouldn't crash

### Link to save file

https://notebin.de/?adc635e45fa352fb#8piu6YAXUJAMPoEZWq2Qe22XiMYKXL8CSjWuH7oQaaBW

### Operating System

Android

Cwpute:
>  the diplomacy manager disappeared due to a civ dying or somesuch.

Quite likely, as we had just defeated China, some 3 to 5 turns ago.

>  was playing such an old Unciv version

So this was likely caused by an unupdated client on their side right ? might be the case, as i suggested we don't update our clients, in the fear it might break our game. But it seems this is not the case.
If you have protected online games against corruption between different versions, you might want to advertize it a little bit :) as its a nice feuture !

Cwpute:
I tried louding the save from the crash report in the first pastebin to test it myself and i can reproduce the crash.
When i load the game, i'm playing as Russia, it is my turn, and this is the notification i tap: 
![crash1](https://github.com/yairm210/Unciv/assets/80604512/6af81e7d-b9d1-4b71-8925-1ca492452e0c)
It is the notification related to "free-roaming" treaty (the one that allows you to roam another civs territory) that just ended.
Tapping it crashes the game: 
![Screenshot_20231128-191429](https://github.com/yairm210/Unciv/assets/80604512/2db1d076-8e6b-4955-a058-269f8d7a1aa4)
Here the crash report i got from that renewed test, for good measure: https://notebin.de/?a0ea274634645b41#4duy9qAx5c7RdA8mtWZ9jBr987AS2b5PmgL8hCDaeAJo

PR #10610, opened 2023-11-28
(Tool) Tiny helper to load MP saves (from issues for debugging) as hotseat

SomeTroglodyte:
... as button on the Debug tab. This is inspired by #10602 making me want such a tool for the maybe sixth time over a year, but at the end of the day - worth it? You throw the dice, I'm 50:50.


yairm210:
I don't see a reason to hide the debug tab behind a password, that's the main limiting factor for why I never use it.
Please un-password that. If anyone is hardcore enough to want to use that, they should deserve to.

yairm210 (approved):

PR #10617, opened 2023-11-29
Fix uniques being affected by city relevant uniques after a unit is already built

### LINKED ISSUE #10603: Unable to upgrade some units.

### Game Version

4.9.1

### Describe the bug

Units requiring a building cannot be upgraded, even when the required technology is unlocked and the necessary building is constructed. The bug only affects the Alpha Frontier mod. 

### Steps to Reproduce

1. Create a game with Alpha Frontier
2. Train a Rover Explorer
3. Research “Robotics” technology and build a Rover Storage. 
4. Upgrade the already existing Rover Explorer : the upgrade button does not exist. 
5. Note that you can still train Combat Rovers, but existing Rover Explorers cannot improve. Also note that the bug seems to affect all units that require a building, almost half of the units. 

### Expected Behavior

The button “Upgrade to Combat Rover” should appear 

### Screenshots

Screenshot_20231127_211618_Unciv
bfc09c)

Screenshot_20231114_114443_Unciv

### Link to save file

_No response_

### Operating System

Android

### Additional Information

Bug already reported on discord and also spotted by other people

SeventhM:
Closes #10603
Couple reasonings here:
1. For non air units, city relevant uniques don't make a whole lot of sense. Even if a unit is garrisoned, it's rare that it's reasonable that a unit is expected to be affected by city based uniques. Also, for non air garrisoned uniques, it make have inconsistent results just moving a unit away from the city that is unexpected
2. Changing checkOnCiv to be similar is mostly unnecessary. It's rare for a unique to not pass in a civ intentionally, and this lets us find bugs more easily. Changing checkOnCiv also would be inconsistent with the many conditionals that use the Civ to find the ruleset

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -171,7 +171,7 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
 
         /** Helper to simplify conditional tests requiring a City */
         fun checkOnCity(predicate: (City.() -> Boolean)): Boolean {
-            if (state.city == null) return false
+            if (state.city == null) return true
             return state.city.predicate()
         }
 


yairm210:
This immediately feels wrong, can you provide an example of such a unique?
Maybe some types should default to true and others to false? Maybe if the unit is done condition?

But for example if you have "+1 strength in coastal cities" and the unit is in a coastal city, it should (thanks to recent changes) get +1, but if it's entirely outside a city it should also get that? But if it's in a city that isn't coastal it shouldn't? Sounds very wrong

yairm210:
Basically it comes down to "should unit uniques be able to ha e city conditionals" and I think the answer is definitely yes

SeventhM:
> But for example if you have "+1 strength in coastal cities" and the unit is in a coastal city, it should (thanks to recent changes) get +1, but if it's entirely outside a city it should also get that? But if it's in a city that isn't coastal it shouldn't? Sounds very wrong

`City center` is a tile improvement that covers most of those cases. The only time one of the city based uniques comes up usually is `in cities with a [buildingName]`. There you can make an argument regarding it (especially for air units), but I doubt most modders have that in mind

Even if you wanted to do that, returning false is still the wrong approach. The right approach is using a `relevantCity` value and checking if ``tile.improvement == city center && city == null``

SeventhM:
I can close this, as it seems we won't go with this approach. That said, passing the city through StateForConditionals requires the unit in question to know there's a city wherever it is. At that point that's just a worse version of what I already said: have conditionsApplies have a `relevantCity` value similar to the `relevantTile` or `relevantUnit` we already do. That's find for the unit if it's on a city, but leaves too many questions if it's not

> If a unit can only be built in a specific cities, then it should also only be _upgradable to_ in those cities

I don't see why. And clearly, in practice, the modder in question didn't either. Upgrading isn't necessarily something someone would expect to be city dependent. At least not through the more generic OnlyAvailableWhen unique. This is far more a question of when should we expect a result out of city condtionals. For something like OnlyAvailableWhen, I can't see other modders necessarily agreeing with the notion that it should only function to allow upgrading if the unit is in a qualifying city. For something like the Strength unique, perhaps you should expect it to be false when not over a city

yairm210:
We can change the behavior when upgrading, to ignore 'only available when' conditions (or a subset), but I personally don't understand it. If a unit is only available under certain conditions, why should you be able to upgrade to it *not* under those conditions? It's just like required tech and resource

Absolutely agree regarding relevantCity, that seems like the way forward 👍🏿

SeventhM:
> It's just like required tech and resource

Again, like those situations, it's a difference between explicit effects and implicit effects. Sure, probably `OnlyAvailableWhen` is meant to be hard limit, but it's so generic that there's ambiguity about where. Should it be a hard limit on upgrading, or just in cities? Do we expect a player to expect it comes up in a given situation (such as for upgrading)? Does it even make sense to cover these situations? What if a modder wanted a unit to be upgradable before it's buildable?

Sure, we have explicit uniques to cover a lot of those situations (such as the unbuildable unique) but it helps to be able to use uniques that aren't so vague, and have explicit expectations on when OnlyAvailable matters

PR #10620, opened 2023-11-29
Fix Unit rename popup offering up the icon's virtual codepoint

SomeTroglodyte:
image

Gets saved in game json, then on a later start could theoretically have the possibility that the codepoint will be used by a different icon... They're dynamic after all, but luckily no element of the generating code has hash-based ordering.

Simple enough, though there's implications:
* The instanceName is translated later for display - IMHO wrong. But with all the double-tr happening it could get a bit difficult to ensure the input stays unchanged. At least this seems clearer than the translation issues with renaming cities.
* Edit: Forgot about Babylon - City rename popup now got same fix.
* The validating test compares translated with untranslated - huh? Fix or drop test or...?


yairm210 (approved):
Not sure about the test
Apart from the icon, this is correct behavior

Issue #10628, opened 2023-11-30
error


user-145580717118795958879:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Game Version

4.9.3

### Describe the bug

https://wwpk.lanzouy.com/iw3CZ1geb2lg

### Steps to Reproduce

Tap the space.

### Expected Behavior

Proceed to the next round.

### Operating System

Android

yairm210:
**Platform:** Desktop
**Version:** 4.9.3 (Build 938)
**Rulesets:** [Civ V - Vanilla, Civ V - Gods & Kings]
**Last Screen:** `com.unciv.ui.screens.worldscreen.WorldScreen`

--------------------------------

OS: Windows 11
	Windows 10 Home China Version 22H2 (Build 25330)
Java: Eclipse Adoptium Temurin-11.0.20+8
	Max Memory: 1024 MB
System default encoding: GBK


--------------------------------


**Message:**
```
java.lang.NullPointerException
	at com.unciv.GUI.getWorldScreen(GUI.kt:45)
	at com.unciv.ui.screens.worldscreen.unit.actions.UnitActions.addSwapAction(UnitActions.kt:110)
	at com.unciv.ui.screens.worldscreen.unit.actions.UnitActions.getAdditionalActions(UnitActions.kt:98)
	at com.unciv.ui.screens.worldscreen.unit.actions.UnitActions.invokeUnitAction(UnitActions.kt:27)
	at com.unciv.logic.automation.unit.CivilianUnitAutomation.automateCivilianUnit(CivilianUnitAutomation.kt:82)
	at com.unciv.logic.automation.unit.UnitAutomation.automateUnitMoves(UnitAutomation.kt:154)
	at com.unciv.logic.automation.civilization.NextTurnAutomation.automateUnits(NextTurnAutomation.kt:339)
	at com.unciv.logic.automation.civilization.NextTurnAutomation.automateCivMoves(NextTurnAutomation.kt:68)
	at com.unciv.logic.civilization.managers.TurnManager.automateTurn(TurnManager.kt:330)
	at com.unciv.logic.GameInfo.nextTurn(GameInfo.kt:376)
	at com.unciv.ui.screens.worldscreen.WorldScreen$nextTurn$1.invokeSuspend(WorldScreen.kt:619)
	at com.unciv.ui.screens.worldscreen.WorldScreen$nextTurn$1.invoke(WorldScreen.kt)
[...]

PR #10632, opened 2023-11-30
RFC: Direction for converting unit actions to map

yairm210:
Problem pointed out by @SomeTroglodyte in #10628 is that to invoke a single unit action, we currently generate *all* possible unit actions just to choose one
This PR is a possible solution for this problem - we convert the list of 

```kotlin
        UnitActionsFromUniques.addParadropAction(unit, actionList)
        UnitActionsFromUniques.addAirSweepAction(unit, actionList)
        UnitActionsFromUniques.addSetupAction(unit, actionList)
```

to a dictionary of "this unit action is generated from this function", keeping in mind there may be multiple actions of a single type

This mapping can be used in both the "get all actions" function, and for single actions.

This is a better solution than the enum, because 
A. we can do this change gradually, one function at a time
B. We can decide where the generation function for each type of action goes - no massive rewrite required

After we have all actions mapped out, we can revisit and see if it makes sense to convert to an Enum - I personally don't see any advantage, but we'll see when we get there

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
@@ -23,19 +23,26 @@ object UnitActions {
 
     /** Returns whether the action was invoked */
     fun invokeUnitAction(unit: MapUnit, unitActionType: UnitActionType): Boolean {
-        val unitAction = getNormalActions(unit).firstOrNull { it.type == unitActionType }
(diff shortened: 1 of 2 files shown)


yairm210:
Another problem with invokeUnitAction as-is, is that an AI unit could potentially have 2 actions of a single type: the first it's unable to act upon due to constraints, the second it can act upon.
Current function will find the *first* one, which is in "display-only" mode so to speak, try to activate it - nothing will happen - and it will return.
Solution should be for the "first" functions to contain a filter.
But that's A. another PR and B. not a situation we've encountered yet

PR #10637, opened 2023-11-30
Fix Carthage civilopedia article

Ouaz:
A text line was weirdly displayed on top of the article:

<details>
<summary>Before:</summary>

</details>

<details>
<summary>After:</summary>

</details>

I'm not sure this is the correct way to fix this (moving the info directly into the related unique), but rather than opening an issue, I created a PR directly.

Edit: Ah yes, it has to match `LandUnitsCrossTerrainAfterUnitGained`, so the fix is stupid. *slow learner*

Sorry!

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Nations.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Nations.json
@@ -895,7 +895,7 @@
 		"innerColor": [81, 0, 137],
 		"favoredReligion": "Islam",
 		"uniqueName": "Phoenician Heritage",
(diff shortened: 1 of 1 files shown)


Ouaz:
Thanks, I'll try the two and choose the best looking one.

Ouaz:
Comment works (and is correctly translated).
<details>
<summary>Comment fix screenshot</summary>

</details>

> Just link the civilopediaText line to something, I believe that moves it next to other linked lines

I tried, but it moves the line at the very end of the article.
<details>
<summary>Link fix (rejected)</summary>

</details>

The comment looks better anyway (as it's just an info, not a bonus from the Civ ability).

PR #10645, opened 2023-12-02
Fixed trading with city-state through notifications

tuvus:
There was a bug reported on Discord where a player found that after signing peace with a city-state and clicking on the "Peace treaty will end in 3 turns", the NotificationAction will open up a trade with the city-state.

This PR simply changes the Diplomacy action to force showTrade to be false if either the current civ or the other civ is a city-state. This allows us to blindly pass a true showTrade parameter without having to worry if the other civ is a city-state or not.


yairm210:
Does feel like a bit of a patch, but not bad enough that I'm against

PR #10650, opened 2023-12-03
Don't list Unique requiring a tech in the Civilopedia for that tech

dHannasch:
~~Don't look at this before https://github.com/yairm210/Unciv/pull/10649 is resolved!~~

This branch is branched off of https://github.com/yairm210/Unciv/pull/10649. The really fiddly bits are in https://github.com/yairm210/Unciv/pull/10649, and those are best handled separately.

After https://github.com/yairm210/Unciv/pull/10649 is merged, it should become clear that this is...well, still a tad odd. This passes Unique-filter functions when fetching the short descriptions of units and buildings for the description of the Technology that enables them. For whatever reason, the unit and building descriptions both accept filter functions for their uniques, but the expected signatures are inverse of each other. There's no way to quite get around the confusingness of that, but this mitigates that by using very explicit named functions `technology::uniqueIsRequirementForThisTech()` and `technology::uniqueIsNotRequirementForThisTech()`.

The upshot is that in the description of each tech, any requirement *for that tech* is not listed as a Unique (because it's implicit in the fact that the unit/building is listed at all), but the other required techs are listed normally.
image
image
Of course, the main description still lists all required techs.
image


core/src/com/unciv/ui/objectdescriptions/BaseUnitDescriptions.kt
```diff
@@ -25,15 +25,15 @@ import com.unciv.ui.screens.civilopediascreen.MarkupRenderer
 object BaseUnitDescriptions {
 
     /** Generate short description as comma-separated string for Technology description "Units enabled" and GreatPersonPickerScreen */
-    fun getShortDescription(baseUnit: BaseUnit): String {
+    fun getShortDescription(baseUnit: BaseUnit, exclude: Unique.() -> Boolean = {false}): String {
```
yairm210, line 28: instead of 'exclude' please call it 'uniqueExclusionFilter' or somesuch

Same for the rest of these

yairm210:
Sorry, only now noticed I missed this

core/src/com/unciv/ui/objectdescriptions/BaseUnitDescriptions.kt
yairm210: Reminder

yairm210 (approved):

PR #10595, opened 2023-11-27
Add iHasUniques.techColumn()

dHannasch:
~~Don't look at this before https://github.com/yairm210/Unciv/pull/10593 is resolved!~~

This branch is branched off of https://github.com/yairm210/Unciv/pull/10593. This will be much harder to read with https://github.com/yairm210/Unciv/pull/10593 changes intermixed. After https://github.com/yairm210/Unciv/pull/10593 is merged, this should be much more readable.

(This should, however, be orthogonal to https://github.com/yairm210/Unciv/pull/10594.)

This adds the convenience function `techColumn()` and replaces all references to `ruleset.technologies[building.requiredTech]?.column` with calls to `techColumn()`.

PR #10651, opened 2023-12-03
Remove last uses of requiredTech, integrate requiredTech into Uniques

dHannasch:
Don't look at this before https://github.com/yairm210/Unciv/pull/10650 is resolved!

...well, technically they're orthogonal, but this greatly increases use of existing functionality that I think https://github.com/yairm210/Unciv/pull/10650 improves.

Background: There are just a couple of remaining uses of the `@Deprecated requiredTech: String` field, now that https://github.com/yairm210/Unciv/pull/10595 is merged. ~~(Last except for https://github.com/yairm210/Unciv/pull/10595, but I got the impression https://github.com/yairm210/Unciv/pull/10595 was approved-in-principle with just some minor revisions.)~~

In every other case, there was functionality for `requiredTech` that was not active for techs required via Uniques, and so it was a simple matter of dropping in the `requiredTechs()` function for every use of `requiredTech`. We didn't really do that, because in many cases it would have added complications to already-complicated bits of code, so in many cases we did some simplifications to avoid that. But still.

The last couple of uses of `requiredTech` are in the Civilopedia. Here the "problem", which in a sense is not really a problem, is that you *already* have a fully-functional system for displaying Uniques.

(With one caveat: I think it's a little *too* functional in some cases, displaying Uniques when they don't really need to be displayed. But https://github.com/yairm210/Unciv/pull/10650 fixes that.)

So we don't need to pull the additional required techs from the Uniques and work them into the existing code that currently uses `requiredTech`. These last few lines of code really do apply to `requiredTech` and *only* to `requiredTech`, because the techs required via Uniques are already handled.

If we integrate `requiredTech` with the Uniques, then the "problem" immediately evaporates: we can simply *delete* these lines of code, and replace them with *nothing*. The infrastructure that you already built for handling Uniques already handles everything that those lines are doing.

But if `requiredTech` is kept separately from the Uniques, then these additional external uses of `requiredTech` will have to persist indefinitely.

This branch integrates `requiredTech` with the Uniques and removes the last uses of `requiredTech` outside the `legacyRequiredTechs()` function.

You previously rejected something similar to this branch when it was hacking the JSON parsing, but I figured I'd try again just in case your underlying objection was literally to messing with the JSON parsing itself. This instead intercepts the uniques at the moment of converting them from Strings to Unique objects, and splices in an additional Unique if `requiredTech` is set. Same basic principle, but without messing with the JSON parsing (because indeed JSON parsing is very fiddly and it's good to avoid messing with that if possible).

This makes the descriptions look like:
image

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
+++ b/core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
@@ -17,8 +17,12 @@ interface IHasUniques : INamed {
     val uniqueMap: Map<String, List<Unique>>
 
     fun uniqueObjectsProvider(): List<Unique> {
-        if (uniques.isEmpty()) return emptyList()
-        return uniques.map { Unique(it, getUniqueTarget(), name) } + legacyRequiredTechsAsUniques().toList()
+        var uniqueObjectList: List<Unique>
+        if (uniques.isEmpty())
+            uniqueObjectList = emptyList()
+        else
+            uniqueObjectList = uniques.map { Unique(it, getUniqueTarget(), name) }
+        return uniqueObjectList + legacyRequiredTechsAsUniques().toList()
     }
     fun uniqueMapProvider(): UniqueMap {
         val newUniqueMap = UniqueMap()
--- a/core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
+++ b/core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
@@ -59,6 +59,10 @@ interface IHasUniques : INamed {
 
     fun legacyRequiredTechs(): Sequence<String> = sequenceOf()
 
+    fun legacyRequiredTechsAsUniques(): Sequence<Unique> = legacyRequiredTechs()
+            .map{ "Only available <after discovering [$it]>" }
+            .map{ Unique(it, getUniqueTarget(), name) }
+
     fun requiredTechs(): Sequence<String> = legacyRequiredTechs() + techsRequiredByUniques()
 
     fun requiredTechnologies(ruleset: Ruleset): Sequence<Technology> =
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/IHasUniques.kt
```diff
@@ -17,8 +17,12 @@ interface IHasUniques : INamed {
...
 
     fun uniqueObjectsProvider(): List<Unique> {
-        if (uniques.isEmpty()) return emptyList()
-        return uniques.map { Unique(it, getUniqueTarget(), name) }
+        var uniqueObjectList: List<Unique>
+        if (uniques.isEmpty())
+            uniqueObjectList = emptyList()
+        else
+            uniqueObjectList = uniques.map { Unique(it, getUniqueTarget(), name) }
+        return uniqueObjectList + legacyRequiredTechsAsUniques().toList()
```
yairm210, line 25: I see no reason to do this, and a bunch of ways this could mess with us later
If we need to handle tech uniques differently, this is not the way

core/src/com/unciv/ui/objectdescriptions/BuildingDescriptions.kt
```diff
@@ -169,9 +169,6 @@ object BuildingDescriptions {
             textList += FormattedLine(stats.joinToString("/", "{Cost}: "))
         }
 
-        if (requiredTech != null)
-            textList += FormattedLine("Required tech: [$requiredTech]",
-                link="Technology/$requiredTech")
```
yairm210, line 174: What we *should* be doing here is 

for tech in requiredTechs
textList += "Required tech: [{tech}]"

yairm210:
Is this stale?

PR #10656, opened 2023-12-03
Dev Console: Lint, crash fix, add Stat

SomeTroglodyte:
Oops, bungled separating the crash fix into a separate commit: It's the two `params[2]` to `params[0]` in unit a/r promotion.
That was all I set out to do, but since there were a few missing horizontal whitespaces...

This also prepares a bit for leaving the console open after a successful command and display some helpful message - like `civ add india faith 100` -> "India now has 999 stored Faith" or somesuch. Opinion?


yairm210:
I don't think that responses are required, however we CAN have "information commands" - like `civ get india faith` or `civ get india stats` or somesuch

yairm210 (approved):

PR #10664, opened 2023-12-03
Add BaseUnit.automaticallyUpgradedInProductionToUnitByTech()

dHannasch:
As discussed in https://github.com/yairm210/Unciv/pull/10654, I want to split apart the effect of "unit is automatically upgraded if in production when tech is discovered" from the effect of "unit is no longer buildable after tech is discovered".

This changes all the usages of `techsThatObsoleteThis()` when what is meant is that the unit should automatically upgrade if in production.

This should not change any behavior. but this isn't quite a simple rename like https://github.com/yairm210/Unciv/pull/10663. This idea of the unit being automatically upgraded by a tech is wrapped in the new function `BaseUnit.automaticallyUpgradedInProductionToUnitByTech()`, which the `TechManager` now calls. In addition, the two-step map in `TechManager.obsoleteOldUnits()` is collapsed to a one-step map.

This does not add any functionality to separate the two concepts, though. This is just reorganization to support separating them.


core/src/com/unciv/models/ruleset/unit/BaseUnit.kt
```diff
@@ -227,6 +227,14 @@ class BaseUnit : RulesetObject(), INonPerpetualConstruction {
         return true
     }
 
+    // This returns the name of the unit this tech upgrades this unit to,
+    // or null if there is no automatic upgrade at that tech.
+    fun automaticallyUpgradedInProductionToUnitByTech(techName: String): String? {
+        if (obsoleteTech != null && obsoleteTech == techName)
```
yairm210, line 233: It's weird that you have 'techs that obsolete this' but you're still using the obsoleteTech to perform this function, shouldn't we *not* be using that anymore?

yairm210: My point here still stands

tests/src/com/unciv/testing/BasicTests.kt
```diff
@@ -63,12 +63,13 @@ class BasicTests {
 
     // If there's a unit that obsoletes with no upgrade then when it obsoletes
 // and we try to work on its upgrade, we'll get an exception - see techManager
+    // But...Scout obsoletes at Scientific Theory with no upgrade...?
```
yairm210, line 66: I had no idea this was hardcoded, we should probably add a unique for this, this is generally not supposed to happen

yairm210 (approved):

PR #10690, opened 2023-12-09
Add the stationed unit conditional unique

PLynx01:
Requested by hackedpassword: [https://github.com/yairm210/Unciv/issues/3242#issuecomment-1826373779](url)

This is the conditional unique which is active for sleeping, fortified or healing units.

If there is something wrong with this PR, please tell me in the comments.


yairm210:
There's no explanation what "stationed" is., unclear for users

This should probably be split into 3 conditionals, one for each action type

PR #10693, opened 2023-12-10
Better Frigate and Ship of the Line sound

tuvus:
I thought the frigate attack sound was a little weak, so I combined the tank shot with the cannon shot to create a new sound for the frigate and ship of the line. The new sound is a volley of 8 cannons (from the tank shot) firing and crashing into another ship (from the cannon shot). The sound is relatively long at 5.5 seconds but doesn't cut off like some of the other sounds.

The two sounds were merged together with audacity and I did check the licensing.


tuvus:
What I have noticed in the weapon sounds is that they cut off too early. They need to have more of a fade out so that it doesn't sound like a cheap audio clip that we took of the internet. So I might work on that. (This is more of an easy side project until I have more time to plan out the worker AI rework more)

PR #10695, opened 2023-12-10
Add new map types

remdu:
Add the map types 

- Fractal
    named like this because it aimed to be similar to Civ 5 Fractal
    similar to Perlin but with different parameters and less land at the edges of the map so that there is almost always a possibility 
    of circumnavigation.
- Lakes
    fully covered in land except for lakes, basically an inverted archipelago map
- Small continents
    basically archipelago but zoomed in and with less land at the edges.
    
Also, for Fractal and small continents, the map is regenerated with lower water level by 0.01 if there is >70% water, similarly to what Pangaea did. For Pangaea, we also add a lower water level by 0.01 for each retry.

edit : 
Fractal, large map
image
Fractal, tiny map
image
Fractal, medium hexagonal
image

Lakes, large map
image
Lakes, tiny map
image

Small continents, large map
image
Small continents, medium map
image
Small continents, medium hexagonal 
image


yairm210:
You can't expect me to add map types without putting pictures here is created maps :)

core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
```diff
@@ -312,6 +357,49 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
         return 1.0 - (5.0 * shouldBeWater*shouldBeWater + randomScale) / 3.0
     }
 
+    private fun getOceanEdgesTransform(tile: Tile, tileMap: TileMap): Double {
+        val a = tileMap.maxLongitude
```
yairm210, line 361: These are confusing names. maxX and maxY, perhaps?

core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
```diff
@@ -312,6 +357,49 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
...
+        val b = tileMap.maxLatitude
+        val x = tile.longitude
+        val y = tile.latitude
+
+        var elevationOffset = 0.0
+
+        val xdistanceratio = abs(x)/a
+        val ydistanceratio = abs(y)/b
+        if (tileMap.mapParameters.shape === MapShape.hexagonal || tileMap.mapParameters.shape === MapShape.flatEarth) {
+            val startdropoff = 0.8
```
yairm210, line 371: What are these measurements? Long/lat? Percentages? Please add meaningful names.
You can't have elevationOffest and a uncategorized offset, either

core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
```diff
@@ -312,6 +357,49 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
...
+            val xdrsquared = xdistanceratio*xdistanceratio
+            val ydrsquared = ydistanceratio*ydistanceratio
+            val distancefromcenter = sqrt(xdrsquared+ydrsquared)
+            var offset = 0.0
+            if (distancefromcenter > startdropoff) {
+                offset = (distancefromcenter - startdropoff)*(1/(1.0-startdropoff))
```
yairm210, line 377: This should be split into named variables

remdu:
I updated with pictures, made readability changes as requested. 
@SomeTroglodyte I don't change existing maps here aside from pangaea where i just do 
`waterThreshold -= 0.01
`every loop iteration

PR #10713, opened 2023-12-11
Removing possible peace exploit

MioBestWaifu:
I was running simulations and analyzing AI behavior for an unrelated reason and found that the AI may accept peace against a disarmed enemy. This is caused by an integer overflow in TradeEvaluation.evaluatePeaceCostForThem() making it return a negative number when it should return positive, which can happen if the evaluator is thousands of times more powerful than the enemy, the more likely scenario beign if the enemy's army is wiped out (theirCombatStrength = 1) and the evaluator force (ourCombatStrength) >= 14655. This is rare, happening in only 6 out of 34 AI-only games, always in turn 360 or higher, but a human player going for a victory other than domination could abuse this possibility by disbanding their army and suing for peace as soon as possible, optimizing their time and resources.


yairm210:
This is *truly* fascinating 🤩

yairm210 (approved):

PR #10716, opened 2023-12-11
Allow arbitrary victory types for the purposes of policies

SeventhM:
This \*should\* have little effect on anything except for any mods giving civs non-standard victory types. In theory, this is not how we should do this; policy choice likely should be separated from victory preference for the most amount of options (hell, Civ 5 doesn't even seem to have victory preferences from my lazy quick glances). In practice, this should be harmless even if victory type was separated from policy preference

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/PolicyManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/PolicyManager.kt
@@ -42,7 +42,7 @@ class PolicyManager : IsPartOfGameInfoSerialization {
         get() {
             val value = HashMap<PolicyBranch, Int>()
             for (branch in branches) {
-                value[branch] = branch.priorities[civInfo.getPreferredVictoryType()] ?: 0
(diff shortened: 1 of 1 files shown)


yairm210:
Ideally, civ.preferredVictoryType should be the one that should be able to return arbitrary ones.

But that would mean modelling what 'preferred victory' actually entails, and allowing *those* AI parameters to be added to arbitrary victory types, which is A Bit Much, and something that if we want to, we can do gradually.

yairm210 (approved):

SeventhM:
> But that would mean modelling what 'preferred victory' actually entails, and allowing _those_ AI parameters to be added to arbitrary victory types, which is A Bit Much, and something that if we want to, we can do gradually.

The lazy naive approach I almost want to consider is adding an `AI.json` file or something of the sort. Having complex AI preferences in Nation just makes that file even more bloated (it's already one of the larger files in some mods) and we already have somewhat a template to piggy back off of in Victory.Fpcus

PR #10717, opened 2023-12-12
Implementing basic civ personalities in diplomacy

MioBestWaifu:
**Edited this description to reflect the actual changes**

This PR is a part of #4674. It allows the definition of diplomatic biases affecting the acceptability of trade, declaration of war, and proposal of defensive pact, open borders and declaration of friendship, to be done in a "personality" property added to to Nation. It also adds the frame for implementing biases towards city-states and building decisions in some other PR.

**Summary of changes and implementation:**
- Personality class: the class to hold the various bias hashmaps and accessor methods. It is in models/ruleset/nation for lack of a better place. The hashmaps shall contain real numbers indicating how a given bias shall affect its domain relative to the average.
- MotivationToAttackAutomation class: its previous method hasAtLeastMotivationToAttack was removed and three others created in its place: 
  - motivationToDeclareWar (without atLeast) returns the overall motivation of the war and avoids the BFS, it is used in determining if the ai wants open borders, defensive pact and DoF because the BFS is not very useful for either. It takes into account basically the same things hasAtLeastMotivationToAttack did, only in different proportions. 
  - motivationToDeclareWar (with atLeast) calls the above and check if there is a path because in this case the BFS is useful. It is only called for actual war declaration.
  - motivationToStayInWar is a simplified version of motivationToDeclareWar as it only takes winnability into account, while with the previous use of hasAtLeastMotivationToAttack AI would factor a number of other things as well.
- Open borders and defensive pact biases implementation: done by simply modifying the minimum roll in the relevant methods of DiplomacyAutomation. In the case of defensive pact i found the AI always wants to sign a pact if all conditions are met sot changing actual desirability was not suitable, thats why i went with the random roll option instead.
- Declaration of friendship implementation: changed nothing in the calculations of wantsToSignDeclarationOfFriendship, made it compare the motivation value to the minimum provided by Personality 
- Trade implementation: not suitable to be implemented for inter-AI trade because the AI almost always makes a even offer or an offer in some specified ratio (plot in image below, Y axis is their offers / our offers, X axis is percentage), so this bias is only used if the offer comes from a human, and even then its a direct multiplication os the value of the bias due to lack of data. 
TradePlot

- War implementation: changed MotivationToAttackAutomation as described above and made the atLeast value for motivationToDeclareWar come from Personality

**Impacts and results:**
When biases are not set, found little to no changes to proposal of defensive pact, friendship and open borders and acceptability of trade. War declaration may increase or decrease up to 10% realtive to before depending on game parameters. Adjusting the default minimum value for war in the 17-25 range should lead to small but significant changes in the rate of war. 

When open borders, DoF, and defensive pact biases are set, they led to the expected outcome. Note that increasing DoF bias has a limited impact because there are just so many friends the AI can make.  

While trade bias was not restricted to offer from humans, increasing it to 1.5 led to insignificant increases in accept rate (91% vs 89%), but decreasing it to 0.5 led to a massive decrease in accept rate (2% vs 74%), which is way it was restricted. 

Setting war bias also works, even if it does not appear so if you simply count war declarations for say, Rome, across X games with and without bias. Reducing the bias does lead to a close to expected decrease in war declaration, but increasing does not. This is chiefly for two reasons: Rome can only be in one war at a time and Rome has no clue it is a trigger-happy civ and does not research, pick policies, build and settle accordingly, leading it to make less than optimal war decisions and getting beaten more often than it should. So, until the AI is improved or until building and other biases are implemented it is strongly recommended to not set war bias higher than 1. 

As for perfomance, i forgot to make precise readings, but i guess there sould be some improvement because is it commented in the code thad BFS is expensive and there should be less of it now. Also, increasing war bias does lead to a negative perfomance impact due to more of said BFS and more war in general.


SeventhM:
> Use Nuke/Gandhi=12??

It's a reference to a maybe, possibly, possibly not bug in an earlier civ game where supposedly Gandhi's friendliness could underflow and make Gandhi a bit nuclear happy. I remember hearing the story might've been embellished by Firaxis a bit, so Idk how much of the story is true

MioBestWaifu:
Also, should there be some RNG in AI personality?

core/src/com/unciv/models/ruleset/nation/Personality.kt
```diff
@@ -0,0 +1,91 @@
...
+        )
+    }
+
+    private fun baseCityStates():HashMap<String, Float>{
+        return hashMapOf(
+            "protect" to 1f,
+            "conquer" to 1f,
+            "ally" to 1f,
+            "bully" to 1f,
+            "interact" to 1f
```
yairm210, line 48: I don't like these magic strings. Each of these sections has *specific* values that are actually meaningful, a set of specific strings is best encapsulated in an Enum.

Even worse - you *require* these! e.g. below, diplomacy["war"]!! will crash the game if the modder forgot to add a war value!

What you actually need here, then, is not a hashmap at all.
It's classes. DiplomacyPersonality, ConstructionPersonality, CityStateInteractionPersonality, with *fields* for each of these values, with defaults. That way, if someone wants to only override the 'bully' personality (whatever that means) then he just needs to create a json with 1 value, rather than everything, and mods are future-safe as well, allowing us to add new values without ruining older mods.

core/src/com/unciv/models/ruleset/nation/Personality.kt
```diff
@@ -0,0 +1,91 @@
...
+        var bias = 100 - diplomacy["war"]!! * 20
+        return when {
+            //The attack function behaves diffently across its domain
+            bias >= 89f ->(7.97 * bias - 686.33).toFloat()
+            bias == 80f-> 21f //Standard
+            bias >=78f->(2.77 * bias - 199.47).toFloat()
+            bias >=22f-> (0.59 * bias - 31).toFloat()
+            bias >=11f -> (1.13 * bias - 47).toFloat()
+            bias > 0f -> (2.76 * bias - 65).toFloat()
+            else -> -500f
```
yairm210, line 62: What are these magic numbers about...?

MioBestWaifu: They are the functions best describing how motivationToDeclareWar behaves, to find the best minimum motivation for a given bias. Could be simplified but results would be less accurate. Plot of motivationToDeclareWar results is below
WarPlot

core/src/com/unciv/models/ruleset/nation/Personality.kt
```diff
@@ -0,0 +1,91 @@
...
+        //Standard is DoF will be asked about 20% of possibilities
+        var bias = 100 - diplomacy["friendship"]!! * 20
+        return when {
+            //Again, function behaves differently across its domain
+            bias >= 82f -> (3.77 * bias - 312.16).toFloat()
+            bias == 80f-> 0f //Standard
+            bias >= 50f-> (0.91 * bias - 72.77).toFloat()
+            bias >= 12 -> (3.36 * bias - 194.49).toFloat()
+            bias > 0f -> (6.11 * bias - 229.37).toFloat()
+            else -> -500f
```
yairm210, line 88: If we're already letting the modder control AI, an easier way would be to let him control either a flat value that can be added to motivation, OR a multiplier for motivation, This function is not only unclear to read, it's also unclear to the modder what his change will actually *do*.

core/src/com/unciv/models/ruleset/nation/Personality.kt
MioBestWaifu: OK, will do it.

yairm210:
Why on earth did you get rid of the motivation map? That's a great tool for developers to understand *why* the AI made the decision it did, by seeing what the value components were.

core/src/com/unciv/models/ruleset/nation/Personality.kt
yairm210: Plot according to what? Where does the blue line come from?

core/src/com/unciv/logic/trade/TradeEvaluation.kt
```diff
@@ -80,6 +80,10 @@ class TradeEvaluation {
         // If we're making a peace treaty, don't try to up the bargain for people you don't like.
         // Leads to spartan behaviour where you demand more, the more you hate the enemy...unhelpful
         if (trade.ourOffers.none { it.name == Constants.peaceTreaty || it.name == Constants.researchAgreement}) {
+            if (tradePartner.isHuman() && trade.ourOffers.none { it.name == Constants.defensivePact}){
```
yairm210, line 83: Even if our partner is AI we should not be affecting peace evaluations.

core/src/com/unciv/logic/automation/civilization/MotivationToAttackAutomation.kt
```diff
@@ -160,95 +237,6 @@ object MotivationToAttackAutomation {
...
-        val combatStrengthRatio = ourCombatStrength / theirCombatStrength
-        val combatStrengthModifier = when {
-            combatStrengthRatio > 3f -> 30
-            combatStrengthRatio > 2.5f -> 25
-            combatStrengthRatio > 2f -> 20
-            combatStrengthRatio > 1.5f -> 10
-            else -> 0
-        }
-        return combatStrengthModifier
-    }
```
yairm210, line 250: These all look like things we should be taking into account?

MioBestWaifu: They are taken into account, here: [powerAdvantageScore](https://github.com/MioBestWaifu/Unciv/blob/449c7bb43d16c7c07f545778c5e18114febc0bf6/core/src/com/unciv/logic/automation/civilization/MotivationToAttackAutomation.kt#L152). Except for defensive pact, forgot about it, will fix that

yairm210: OK, fair enough

MioBestWaifu:
Closed this because, after thinking more about it, this simply does not seem like a good way of making civs different from each other, this whole functionality would be close to useless. War bias would only be half-functional because a warmonger civ does not know it should rush trebuchet, a pacifist civ does not know it should keep a stash of cash in case of war or any other insurance policy.  Similarly, setting a high build wonder priority would not necessarily lead to the AI picking the tradition policy line and so on. There could also be settable biases for all of those but it seems like a terrible idea. It is proprably better to make the AI customizable in some higher-level way, like setting "behaviour: [conquer all, prioritize siege weapons]" or something like that.

SeventhM:
I'm... Not sure I agree with your sentiment. Between my statement [here](https://github.com/yairm210/Unciv/pull/10716#issuecomment-1852715579) and Victory.Focus already doing similar stuff, it seems obvious to me that a way to set priorities behind what's hardcoded (and perhaps allow for aggressive culture AIs) ideal. Obviously it probably will have flaws in exactly how to process those biased, but I kinda want to still point at how Victory.Focus does it say that is at least doable. I almost rather not have it so that we only have higher level options, but I could be on the minority on that one

I think the biggest concern you seem to have is whether the AIs make the right option for individual actions. And yeah, I think that's close to useless to use biases for. But i doubt modders would use it for or expect that, and I would expect this to just be something the AI devs (I feel like @ ing Tuvus, but I won't for now) could easily fix up eventually

What I will say is possibly that the larger concern is the amount of bloat this may add to the Nation file. Hence why my comment in the other PR was that it should probably be a separate file entirely

MioBestWaifu:
> I'm... Not sure I agree with your sentiment. Between my statement [here](https://github.com/yairm210/Unciv/pull/10716#issuecomment-1852715579) and Victory.Focus already doing similar stuff, it seems obvious to me that a way to set priorities behind what's hardcoded (and perhaps allow for aggressive culture AIs) ideal. Obviously it probably will have flaws in exactly how to process those biased, but I kinda want to still point at how Victory.Focus does it say that is at least doable. I almost rather not have it so that we only have higher level options, but I could be on the minority on that one
> 
> I think the biggest concern you seem to have is whether the AIs make the right option for individual actions. And yeah, I think that's close to useless to use biases for. But i doubt modders would use it for or expect that, and I would expect this to just be something the AI devs (I feel like @ ing Tuvus, but I won't for now) could easily fix up eventually
> 
> What I will say is possibly that the larger concern is the amount of bloat this may add to the Nation file. Hence why my comment in the other PR was that it should probably be a separate file entirely

The way victory focus affects decision-making does look like a good template to personalities. Will submit a proposal using the same mechanisms later.

PR #10720, opened 2023-12-12
Add damage dealing unit trigger

SeventhM:
Note the different between this and OneTimeUnitHeal is
A. It's more clear and less hacky
B. `UniqueType.HealingEffectsDoubled`
~~C. Niche gimmicks of a unit being alive until end turn is called, if I remember correct~~

Is there a reason why healing allows for negative hp anyways?


yairm210:
Not really
Both damage and heal should be restricted to positive amounts, I'm pretty sure we have a parameter type for that - positiveAmount perhaps?

yairm210 (approved):

PR #10728, opened 2023-12-13
Fix stats from tileFilter unique not working on improvements

SeventhM:
I totally haven't noticed that the Fresh Water line below this on it now redundant. Also, the "non-fresh water" line \*should\* be redundant, but accidentally isn't a tileFilter. I could probably fix that here tbh, it would be functionally unnoticeable, but the question would be should we keep "non-fresh water" for compatibility reasons or should we expect people to update to "non-[Fresh Water]"


yairm210:
I'd be happy leave it now, but add a rulesetvalidator deprecation warning for future-proofing, and then later when it stops working they'll also be able to know why

yairm210 (approved):

PR #10737, opened 2023-12-14
Allow battle calculations when not your turn

soggerr:
Allow BattleTable display when it is not the viewing players turn.
This allows for more planning while waiting for other players.


soggerr:
> _**more**_?
lol, to each their own I guess, just don't click on the map then when it isn't your turn ;) I think this will definitely speed up games by letting human players do more planning while the other player plays their turn.

> if(
done

yairm210 (approved):

PR #10738, opened 2023-12-14
Show average damage in battle calculations

soggerr:
Change BattleTable to calulate average damage and show alongside the min and max for easier calculations when planning to attack with multiple units such as when sieging a city

Unfortunately for min-maxers the randomness is still turn dependent for a given defending unit, so the damage randomness will not average out across multiple attacks on a given turn but they will all be more or all be less depending on the luck that turn.

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTable.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTable.kt
@@ -204,9 +204,11 @@ class BattleTable(val worldScreen: WorldScreen) : Table() {
 
         val maxDamageToDefender = BattleDamage.calculateDamageToDefender(attacker, defender, tileToAttackFrom, 1f)
         val minDamageToDefender = BattleDamage.calculateDamageToDefender(attacker, defender, tileToAttackFrom, 0f)
+        val avgDamageToDefender = arrayOf(maxDamageToDefender, minDamageToDefender).average().roundToInt()
 
         val maxDamageToAttacker = BattleDamage.calculateDamageToAttacker(attacker, defender, tileToAttackFrom, 1f)
         val minDamageToAttacker = BattleDamage.calculateDamageToAttacker(attacker, defender, tileToAttackFrom, 0f)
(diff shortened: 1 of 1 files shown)


soggerr:
The problem I am trying to solve is: how do I plan killing a strong unit or city that requires many multiple attacks? I need some way to tell how much damage each attack will do so I can add them up in my head to see if it is enough.

Here is screenshot:
image

core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTable.kt
```diff
@@ -204,9 +204,11 @@ class BattleTable(val worldScreen: WorldScreen) : Table() {
 
         val maxDamageToDefender = BattleDamage.calculateDamageToDefender(attacker, defender, tileToAttackFrom, 1f)
         val minDamageToDefender = BattleDamage.calculateDamageToDefender(attacker, defender, tileToAttackFrom, 0f)
+        val avgDamageToDefender = maxDamageToDefender - ((maxDamageToDefender - minDamageToDefender)/2.0).roundToInt()
```
yairm210, line 207: Maybe math had an avg function?

yairm210: Nope , doesn't look like it

yairm210 (approved):

core/src/com/unciv/ui/screens/worldscreen/bottombar/BattleTable.kt
soggerr: As far as I am aware there is no functionality for average or mean calculation in java.lang.Math. Does unciv import any other math libraries that I should use?

soggerr: Yes, I understand. I was just wondering if you are using any math libs elsewhere to do averages you could suggest. I just uploaded a new patch switching it to use the collections average if you would prefer it to use functions.

PR #10739, opened 2023-12-14
Show tied civs and player score in city state competitions

soggerr:
Show all leading civilizations in City State global competitions when there is a tie for the lead. Also, show the viewing civ's score, so a player can see how they compare to leaders without having to keep track/calculate in their head.

### DIFF
--- a/android/assets/jsons/translations/French.properties
+++ b/android/assets/jsons/translations/French.properties
@@ -172,8 +172,8 @@ Ally = Allié
 
 [questName] (+[influenceAmount] influence) = [questName] (+[influenceAmount] influence)
 [remainingTurns] turns remaining = [remainingTurns] tours restants
-Current leader is [civInfo] with [amount] [stat] generated. = [civInfo] est actuellement en tête et a généré [amount] [stat].
-Current leader is [civInfo] with [amount] Technologies discovered. = [civInfo] est actuellement en tête avec [amount] Technologies découvertes.
+Current leader(s) = Le/Les Leader(s)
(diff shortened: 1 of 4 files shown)


soggerr:
I don't know how compatible a variable length list of civ names is compatible with translation system?

soggerr:
> variable number of placeholders - nope

yeah, that's what I figured. Ok, I will reformat based on your suggestions.

yairm210:
For all visual changes, please attach screenshots :)

soggerr:
Ok, changed around strings to try to solve translation issue. The string composition is messy and will be fragile, but if translation logic can't handle variable length components or combined entries, then I guess we have to do manual translations like this? 

Quest with one leader, viewing civ is trailing:
image

Quest with multiple leaders, including viewing civ:
image

Translations working:
image

Global quests with no scores (invest):
image

Quest when there is no leader yet (new quest):
image

soggerr:
hmmm, the tr() import is there and it is building on my machine, not sure why Gradle build is failing. Any suggestion to point the Gradle compiler at the correct tr()?

soggerr:
hmmm, must have missed something handling commits across branches for the pull requests. I will have a look when I am done work or maybe tomorrow to dix the build, my bad. If I am doing another pass, I will use some more formal string formatting to make it look nicer to.

I agree that having all this string manipulation and translation work in logic doesn't make a lot of sense. Not sure what the standard is here on what level of abstraction to pass to the UI, but since CityStateDiplomacyTable is already directly using objects like AssignedQuest, is it ok to just move then logic from getScoreStringForGlobalQuest into the UI layer and make getScoreForQuest public?

Issue #10725, opened 2023-12-13
RFC: What can we learn from Shattered


SomeTroglodyte:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

Since #10710 reminded me that I did read parts of [Shattered Pixel's](https://github.com/00-Evan/shattered-pixel-dungeon) code recently, specifically looking for how it deals with certain Android display things, I'd like to share my notes:

* Disable Compass and Accelerometer on Android (for performance?)
* Overrides Android Backup support (limiting which files are saved)
* Detects metered networks on Android (mod mgr?)
* Has detection for Android isInMultiWindowMode and adjusts its screen size accordingly
* Desktop: Minimum Window 720x400 - a little more generous than ours
* Desktop: setWindowIcon gets a choice of icons up to 256x256
* Desktop: True maximized persistence (did I rewrite and push that branch? If I did, we got that too)
* Desktop: Special case for MacOS and fullscreen
* Uses XDG_DATA_HOME, %AppData% or Mac 'Library/Application Support/' for its storage (DesktopLauncher.java L144+)
* Has in-game "cloud" services for communicating announcements and update-available
* Utilizes org.lwjgl.util.tinyfd.TinyFileDialogs for crash messages (would that filepicker work instead of mine?)
* Uses gdx.utils.SharedLibraryLoader for OS detection (we do System.getproperty directly instead iirc) - but needs to fix it for FreeBSD
* Has moved to Gradle 8.3!
* Shattered does *not* use Gdx ParticleEffect for its beams and spells, but maybe we could for WLTKD...

These range from highly desirable to 'whatever', so any input about which of those ideas we could pursue for Unciv too would be welcome.

### Desired Solution

opinions

### Alternative Approaches

n/a

PR #10740, opened 2023-12-14
Bump Gradle to 8.5 and kotlin to 1.9.20

SomeTroglodyte:
#### An Experiment! Implications unclear.

This is what I ended up with after following the second comment in #10725 - it runs all targets, from gradlew commands or from Studio run configurations. Still, somehow the requirement to remove all those `sourceCompatibility` declarations doesn't spark confidence...


SeventhM:
I'm not as knowledgeable of the different versions, so I'm almost definitely missing something... But is there a reason to move to 8.1.2 instead of 8.1.4 or 8.2.0?

SeventhM:
I keep up to date on Studio kinda just because (I'm torn on the new UI tbh, some of the stuff moved around in ways that I don't like, but it's not the worst UI change I've seen), and it gave me a warning for 8.1.2 that 8.1.4 was newer, so ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯. For the record, this pr worked fine on my dev build, even when going to 8.1.4. Haven't tried 8.2.0 yet, but I'm pretty sure it'll work too

SeventhM:
Oh hey
> Configure project :
AGPBI: {"kind":"warning","text":"The option setting 'android.defaults.buildfeatures.buildconfig=true' is deprecated.\nThe current default is 'false'.\nIt will be removed in version 9.0 of the Android Gradle plugin.\nYou can resolve this warning in Android Studio via `Refactor` > `Migrate BuildConfig to Gradle Build Files`","sources":[{}]}

SeventhM:
Am messing around and did an inspection
> 'Enum.values()' is recommended to be replaced by 'Enum.entries' since 1.9

SeventhM:
Not sure if it's related, but I've recently been running into a new set of problems: It'll work fine for whatever session I'm currently working with, but if I move my environment to a different computer, I run into issues with `.idea` that force me to delete pretty much the whole folder. Otherwise, Studio starts complaining about an illegal character somewhere. Idk if it's a issue with how Studio is saving configurations, or something else entirely, but it is annoying

yairm210:
To minimize risk, I added the changes that I could that don't require upgrading

I also upgraded the java version in the deploy so this should pass

SeventhM:
> How?

Studio is on a flash drive for me. Also, I lazily haven't integrated my environment with git. I instead either manually updated stuff or redownload everything. I probably could get hit working with the flash drive, but ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

yairm210:
Lazy loading - developer edition 👍🏿

PR #10741, opened 2023-12-14
Coast spread algorithm

remdu:
This is a change to the algorithm to spawn coast tiles. 
Before, the line
```
val coastLength =
                    max(1, randomness.RNG.nextInt(max(1, map.mapParameters.maxCoastExtension)))
```
was wrong because nextInt takes values up to the input parameter (exclusive) so there was always 1 less real max coast extension than we chose. Also, with high values of max coast extension, coast could spawn unconnected to land or other coast tile, which is a behavior that does not exist in civ 5 afaik, and can be misleading (discovery of coast could mislead you into thinking you've found another continent)
So I changed the algorithm so that it instead spread coast extension by extension, with a random factor for tiles that are at least 2 tiles away from land. I expected it to also be faster on high max coast extension because we don't need to look up a large neighborhood, but when measuring the average, it actually seem slower by a few milliseconds. I'm a beginner in kotlin so I may be missing something.

I also change the default MaxCoastExtension to 3 instead of 2, because in civ 5 it seems to be at least 3. But this is a big change to all map types so I don't know if it's the right thing to do.

Anyway, here are a few pictures of coast from this algorithm : 
Small Continents, Huge
image
Pangaea, Huge
image
Fractal, Huge
image


core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -277,12 +277,26 @@ class MapGenerator(val ruleset: Ruleset, private val coroutineScope: CoroutineSc
 
         //Coasts
         if (ruleset.terrains.containsKey(Constants.coast)) {
-            for (tile in map.values.filter { it.baseTerrain == Constants.ocean }) {
-                val coastLength =
-                    max(1, randomness.RNG.nextInt(max(1, map.mapParameters.maxCoastExtension)))
-                if (tile.getTilesInDistance(coastLength).any { it.isLand }) {
-                    tile.baseTerrain = Constants.coast
-                    tile.setTransients()
+            for (i in 1..map.mapParameters.maxCoastExtension) {
```
yairm210, line 280: This should be a separate function :)

remdu:
Like this ?

remdu:
@SomeTroglodyte I use nextBoolean() now
for your second point, i changed the forEach to a for loop
For your third point, there is no reason for toCoast to have multiple copies of the same tile since we loop through each tile once. I checked anyway with using a Set() in parallel and the set had the same size as the mutableList.
For your fourth point, I changed it back to 2.

remdu:
I'm not sure I understand your proposition well.
I have also my own idea of how to change the algorithm, by adding a "Even coast" toggle. It could do coast like this for Max Coast extension = 2
image
The coast is more evenly spread, with only rare case where one coast tile is only connected to 1 coast tile, and also rarer "holes" in the coast. This is achieved by simply moving a break statement in the algorithm. I think being able to choose between "branchy" coast or "even" coast could be good.

But this is probably a talk for a new issue for future changes to coast placement ?

yairm210 (approved):

PR #10745, opened 2023-12-15
New Citizen Focus Options

### LINKED ISSUE #10501: Citizen Management auto reset
Hello,

I just discovered Unciv and I am impressed of the game.

I think there is a problem in citizen management (or maybe there is something I don't understand) 
When i manually manage my citizens, the following turn everything is reset and I have to redo everything. 

DOes anyone have the same problem ? Is it a bug or a parameter I didnt see ?

itanasi:
Add Gold Growth and Production Growth Focus
New Manual Focus doesn't reallocate population

Fixes #10501

v3 of compact table
image


core/src/com/unciv/logic/city/CityFocus.kt
```diff
@@ -32,20 +32,23 @@ enum class CityFocus(
...
-    ProductionGrowthFocus("Production Growth Focus", false) {
+    ProductionGrowthFocus("Production Growth Focus", true) {
         override fun getStatMultiplier(stat: Stat) = when (stat) {
             Stat.Production, Stat.Food -> 2f
             else -> 1f
         }
     },
     FaithFocus("[${Stat.Faith.name}] Focus", true, Stat.Faith),
     HappinessFocus("[${Stat.Happiness.name}] Focus", false, Stat.Happiness),
+    Manual("Manual Focus", true, null) {
```
yairm210, line 49: Why is tech enabled true?

itanasi: You mean Table? That's the flag that lets the option show up in the UI.

itanasi:
Suggestion for how to reduce the selection table? Switching to a 3-row is possible, but use symbols instead of text perhaps. A dropdown feels clunky.

I also realized I could show the new options in a more efficient manner.

itanasi:
Correct, clicking on it will reassign. But I tested as improvements and population show up they don't get rebalanced. Hm, so fix it so that selecting Manual doesn't trigger the rebalance. Got it.

Could also change the title. It's not really a Focus. Just... Manual?

As for the symbols yeah that's part of what I want to do better.

As for saving a row I think having them soan both columns helps indicate their difference from the selectable choices below. Although Avoid Growth could be a Radio Button for extra clarity, but that would break the visual language imo

itanasi:
image

itanasi:
Bit more compact
image

itanasi:
Is there an easy way to have just the stat icon in a string,

itanasi:
v3, maximally compact
image

yairm210 (approved):

android/assets/jsons/translations/template.properties
```diff
@@ -1256,9 +1256,10 @@ Unlock =
 Move to city = 
 Reset Citizens = 
 Citizen Management = 
+Citizen Focus = 
 Avoid Growth = 
-Default Focus = 
-[stat] Focus = 
+# Default =
```
yairm210, line 1261: Should this be a comment...?

core/src/com/unciv/logic/city/CityFocus.kt
```diff
@@ -24,28 +26,31 @@ enum class CityFocus(
...
-    FoodFocus("[${Stat.Food.name}] Focus", true, Stat.Food),
-    ProductionFocus("[${Stat.Production.name}] Focus", true, Stat.Production),
-    GoldFocus("[${Stat.Gold.name}] Focus", true, Stat.Gold),
-    ScienceFocus("[${Stat.Science.name}] Focus", true, Stat.Science),
-    CultureFocus("[${Stat.Culture.name}] Focus", true, Stat.Culture),
-    GoldGrowthFocus("Gold Growth Focus", false) {
+    Manual("Manual", true, null) {
+        override fun getStatMultiplier(stat: Stat) = 1f
+    },
+    FoodFocus("[${Stat.Food.character}]", true, Stat.Food),
```
yairm210, line 35: Why are all these wrapped in square brackets? There should be no translation going on here anyway

android/assets/jsons/translations/template.properties
itanasi: That was intentional since Default already gets translated elsewhere. I can elaborate more

core/src/com/unciv/logic/city/CityFocus.kt
itanasi: Good point

android/assets/jsons/translations/template.properties
yairm210: Then this should be removed, not commented

core/src/com/unciv/logic/GameInfo.kt
```diff
@@ -91,7 +91,7 @@ class GameInfo : IsPartOfGameInfoSerialization, HasGameInfoSerializationVersion
     companion object {
         /** The current compatibility version of [GameInfo]. This number is incremented whenever changes are made to the save file structure that guarantee that
          * previous versions of the game will not be able to load or play a game normally. */
-        const val CURRENT_COMPATIBILITY_NUMBER = 3
+        const val CURRENT_COMPATIBILITY_NUMBER = 4
```
alexban011, line 94: are you sure this breaks compatibility & backwards compatibility can't be implemented in a feasible way? 

Back when Azurite implemented this it was only to be used partly for stuff like preventing people from using an older version of the game compared to other players even if the save file still loads but would then break the save after serializing

itanasi:
Old saves won't load if anyone uses the new Manual Focus in the Save File. So if someone starts a MP across versions they could abruptly break.

itanasi:
We could add the Manual enum datatype and not enable it for a few versions.

yairm210:
Damn I missed that more discussion was done here

In general the solution we have for enums changing is saving the value *as a string* and not as an enum
So that's probably a change we need to do prior to this one, I'll revert and return this

yairm210:
Sorry for the mess

yairm210:
@itanasi if that looks good to you, then we can merge it and then reopen this (or perhaps revert the revert?)

itanasi:
Not at a work station. We can let this propagate so that when we add Manual no saves break

Issue #10743, opened 2023-12-14
Crash on selecting mod


krijnsent:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.9.8

### Describe the bug

Game crash in Start Game screen

### Steps to Reproduce

Click "start new game"
Select installed mod "The great unciv rework"
Select the Extension mod "Brave new world"
Game crashes

### Link to save file

**Platform:** Desktop
**Version:** 4.9.8 (Build 943)
**Rulesets:** [Civ V - Vanilla, RekMOD, Brave New World, Community Maps, DeCiv Redux, 3rd and 4th Unique Component, Civ6 mod, Barbarian xp farm, Alpha Frontier, Outlaws of the Wastes, Civ V - Gods & Kings, The Great Unciv Rework, 5Hex Tileset]
**Last Screen:** `com.unciv.ui.screens.newgamescreen.NewGameScreen`

--------------------------------

OS: Windows 11
	Windows 10 Home Version 22H2 (Build 22621)
Java: Eclipse Adoptium Temurin-11.0.20+8
	Max Memory: 1024 MB
System default encoding: windows-1252

--------------------------------

**Message:**
```
java.lang.RuntimeException: Actor: CheckBox: Brave New World
	at com.badlogic.gdx.scenes.scene2d.Actor.notify(Actor.java:192)
	at com.badlogic.gdx.scenes.scene2d.Actor.fire(Actor.java:152)
	at com.badlogic.gdx.scenes.scene2d.ui.Button.setChecked(Button.java:125)
	at com.badlogic.gdx.scenes.scene2d.ui.Button$1.clicked(Button.java:93)
	at com.badlogic.gdx.scenes.scene2d.utils.ClickListener.touchUp(ClickListener.java:88)
	at com.badlogic.gdx.scenes.scene2d.InputListener.handle(InputListener.java:71)
	at com.badlogic.gdx.scenes.scene2d.Stage.touchUp(Stage.java:354)
	at com.unciv.ui.screens.basescreen.UncivStage.access$touchUp$s80204510(UncivStage.kt:17)
	at com.unciv.ui.screens.basescreen.UncivStage$touchUp$1.invoke(UncivStage.kt:87)
	at com.unciv.ui.screens.basescreen.UncivStage$touchUp$1.invoke(UncivStage.kt:87)
[...]

PR #10751, opened 2023-12-15
Mod declarative compatibility - a little more

SomeTroglodyte:
Triggered by #10743 
* Very impotrant: Fexed a hurrblah tpyo i n a phfunktshun nahme - sorry cutlery lovers.
* Allow mods to specify not only incompatibilities but requirements - as mod name filter only so far. But something like Brave New World which is a simple extension Mod maybe never meant to run on TGUR could say "I'll run on G&K and nothing else"... If they wished it so, and if there's still a maintainer to do ModOptions.
* Both requirement and incompatibility declarations applied dynamically to checkboxes in new game screen.
* Cleaned up some outdated comments.
* Possibly fixed the annoying bug where one was asked to affirm an outdated modcheck error list?
* Moved all pertaining code into a specialized file.

#### Open questions/missing stuff
* I'd like to have mod filters that test object existence or other mod properties. Like - I want to run only on base rulesets that allow Religion, or - I'm fine as long as the base ruleset has a Great Archaeologist. But how without introducing a new syntax for modFilter? Something conditionals-like, but I can't get a mental picture of how it could/should look like.
* ~Visuals missing: later, I tested this on a box that doesn't run `peek` so I'll need to re-do the test changes on the other box~
* Testing - I did some, but honestly this would need some more pairs of eyes than my one-and-a-pirates-patch.
* The ModCompatibility object now lives alongside RulesetValidator in 'models' even though it's more close to 'logic'. Suggestions welcome.
* Undecided - deselectIncompatibleMods or disableIncompatibleMods - which is the friendlier UX... I tend to disable.

<details><summary>Demo as clip</summary>

I've **_declared_** Aggtelek to require "`*Civ 5 -*`" and Aliens to require G&K and be incompatible with Aggtelek - first see how below the entries change when switching base above, then in the second half I'll demo the mutual exclusion of these two mods...

https://github.com/yairm210/Unciv/assets/63000004/5a6d4f6a-8e3a-48dc-b393-ba0a8613c7cb

(Oh wow peek records audio? You can hear the fade-over to the Aliens theme...)
</details>


core/src/com/unciv/models/ruleset/validation/ModCompatibility.kt
```diff
@@ -0,0 +1,119 @@
...
+     *  - Extension mod is incompatible with any _other_ **selected** extension mod -> Nope
+     *  - Extension mod has no ModRequires unique -> OK
+     *  - For each ModRequires: Not([baseRuleset] meets filter OR any other **selected** extension mod meets filter) -> Nope
+     *  - All ModRequires tested -> OK
+     */
+    fun meetsAllRequirements(mod: Ruleset, baseRuleset: Ruleset, selectedExtensionMods: Iterable<Ruleset>): Boolean {
+        val otherSelectedExtensionMods = selectedExtensionMods.filterNot { it == mod }.toList()
+        if (otherSelectedExtensionMods.any { isIncompatible(mod, it) }) return false
+
+        for (unique in mod.modOptions.getMatchingUniques(UniqueType.ModRequires)) {
```
yairm210, line 112: Looks exactly like the check in meets base requirements, extract to function?

yairm210 (approved):

PR #10753, opened 2023-12-15
Fix: key bindings lost on quickly entering and leaving options

SomeTroglodyte:
Or more precise - see commit message - but that may be too unintelligible (what a wonderfully funny word) for the changelog.

Note: I had a branch once where there was a right-click menu with a "reset to defaults" for category/all... Where did that go? You don't remember anything in that direction, right?
Another note: Needs another tweak - sorting is including the symbol codepoints again. Some other time.


yairm210 (approved):

Lesiakower:
> I had a branch once where there was a right-click menu with a "reset to defaults" for category/all... Where did that go?

Yeah, if it doesn't come back I will have to get rid of my old muscle memory :'(

PR #10758, opened 2023-12-16
Stop on Path Blocked

itanasi:
If a unit is trying to move to a tile, and doesn't move this turn, likely path is blocked and we should stop the unit so that it can get new orders


yairm210 (approved):
Does this unit then enter the idle units?

itanasi:
> Does this unit then enter the idle units?

Not on this turn, but next turn it will

PR #10759, opened 2023-12-16
Add the remove unit promotion unique

PLynx01:
Some players and modders asked for this feature, so I decided to implement it to the game. 

I believe it will be useful for modders to make the unit effects that can be removed under certain conditions or unit actions.

If you have some remarks, please share your thoughts in the comments. Constructive feedback welcome

### DIFF
--- a/core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
+++ b/core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
@@ -97,10 +97,9 @@ class UnitPromotions : IsPartOfGameInfoSerialization {
 
         if (getPromotions().contains(promotion)) {
             promotions.remove(promotionName)
+            unit.updateUniques()
+            unit.updateVisibleTiles()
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
```diff
@@ -91,6 +91,18 @@ class UnitPromotions : IsPartOfGameInfoSerialization {
...
 
+    fun removePromotion(promotionName: String) {
+        val ruleset = unit.civ.gameInfo.ruleset
+        val promotion = ruleset.unitPromotions[promotionName]!!
+
+        if (getPromotions().contains(promotion)) {
+            promotions.remove(promotionName)
+        }
+
+        unit.updateUniques()
```
yairm210, line 102: These should be within the if as well
Alternatively, if false return

yairm210 (approved):

PR #10760, opened 2023-12-16
Add unique check for connect road automation

willjallen:
Units that can't normally build roads in mods have access to the connect road feature. This PR fixes this.


yairm210 (approved):

SeventhM:
Hmm... Is there a way to use MapUnit.canBuildImprovement instead, in case we find a way to generize roads?

willjallen:
In `canBuildImprovement`:
```kotlin
if (tile.improvementInProgress == Constants.repair && buildImprovementUniques.any()) return true
```
This would cause the automation button to show up in situations where it shouldn't.

SeventhM:
Huh... Never noticed that... Hmmm... That kinda reads as a bug in its own right

PR #10777, opened 2023-12-18
Fix any tech researched clearing all units from queues

SomeTroglodyte:


yairm210:
prepare patch

yairm210:
release patch

yairm210:
Technically we could associateBy and then filterNotNull (or remove by predicate?) but right now we're looking for a fast solution and if it works it works

PR #10782, opened 2023-12-19
Randomize seed checkbox

remdu:
So I noticed something annoying when using the map editor.
When you manually modify the seed and click create, it will work and use the input seed if it is the first time since you opened the map editor, but otherwise it will generate a new random seed just before generating the map. This is because there is code for generating anew when doing partial map generation that also applies to new (MapGeneratorSteps.All) maps. That makes it annoying when you want to use a seed you like.
I tested different solutions

1. Avoid reseed when doing new full generation. This allows to manually choose the seed, but you always need to choose the seed, which is annoying.
2. Same as 1 but add a "Randomize seed" button. This works well but you can't generate new maps every time you click "create" anymore. You have to click "Randomize seed" every time.
3. Reseed after the map is generated rather than before. This allows to manually put your own seed, but when you find by accident a map you like, you can't save its seed anymore.
4. The method I think is best. Same as 1 but add a "Randomize seed" checkbox. This solves all of these problems.

I've tested it on desktop and in an emulator running Android 11.0 and it works.

Here's a screenshot of how the new checkbox looks : 
image

### DIFF
--- a/core/src/com/unciv/ui/screens/mapeditorscreen/tabs/MapEditorGenerateTab.kt
+++ b/core/src/com/unciv/ui/screens/mapeditorscreen/tabs/MapEditorGenerateTab.kt
@@ -66,7 +66,10 @@ class MapEditorGenerateTab(
     }
 
     private fun generate(step: MapGeneratorSteps) {
-        if (step <= MapGeneratorSteps.Landmass && step in seedUsedForStep) {
+        if (step == MapGeneratorSteps.All && newTab.mapParametersTable.randomizeSeed) {
+            newTab.mapParametersTable.reseed()
+        }
+        if (step == MapGeneratorSteps.Landmass && step in seedUsedForStep) {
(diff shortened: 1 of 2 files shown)


remdu:
@SomeTroglodyte
- Running the same map generation step/all-steps twice in succession should always yield new results - otherwise "huh, nothing happened".

That point already works and still works with my changes.
- Entering a known seed then hitting generate all should give exactly the same map (type and size and params assumed identical) as playing that seed from new game

This part doesn't work currently. You can get the right map if you just opened the map editor. But if you've already generated a map, you can't specify the seed anymore, or rather it's going to be reseeded right before map generation anyway. That's what I fix with my PR
- Entering a known seed then hitting all single generation steps once in sequence should still give exactly the same map

This part doesn't work currently. And my PR doesn't fix it. I tried and you can get the same landmass, but already at the "raise mountains and hills" step you get different mountains from what you get when you generate the whole map. I don't know where this difference come from atm. Needs investigation

PR #10784, opened 2023-12-20
Improve connect road feature (RFC)

willjallen:
Now that I've played actual games with this feature, i've noticed a few annoying things. This PR is to clean up the documentation, readability and hopefully address some of these issues

- Workers should wait for the next tile in the path to clear instead of skipping over it. We should also raise an alert that the path is blocked

image
image

This can cause the resulting road to have gaps in it. Not cool. Additionally, I think the worker should have the road icon instead of the generic automated icon to more clearly distinguish it.

- Movement optimal vs directness

image

This issue has been raised previously and I'd like some input on this. Here the pathfinding algorithm is preferring the path connected by existing roads as opposed to a more direct path. I've found this to be somewhat annoying when I want to make specific road paths in wartime scenarios or otherwise. I think we should lean more towards player intent as opposed to optimality. Doing this probably requires some clever balancing between absolute distance and movement cost.

- Dangerous tiles

The current implementation does not make use of the set of dangerous tiles passed to it. In practice, the logic which makes the worker run away is already implemented higher up in the automated worker code. This typically causes the worker to move off the path and cancel the automation anyway, but I can forsee some obscure bugs produced as a consequence. These two pieces should be more elegantly connected.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -125,20 +125,15 @@ class WorkerAutomation(
 
 
     /**
-     * Automate the process of connecting a road between two points.
-     * Current thoughts:
-     * Will be a special case of MapUnit.automated property
-     * Unit has new attributes startTile endTile
-     * - We will progress towards the end path sequentially, taking absolute least distance w/o regard for movement cost
-         * - Cancel upon risk of capture
-         * - Cancel upon blocked
-     * - End automation upon finish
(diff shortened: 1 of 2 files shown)


tuvus:
> Movement optimal vs directness

MST is optimal in the placement of the minimum number of roads, but I don't think that is always what we want here. For and extreme example, sometimes as a player I want to build railroads that are two-wide so that many units can efficiently move past. 

Are you able to make a road reuse slider so we can make it customizable? 
In the early game players will want to build roads to connect them to cities wanting to build the least amount of roads possible (and MST), however, in the later game, the directness should be favored due to income and better unit movement.

PR #10791, opened 2023-12-22
Another proposal for civ personalities

MioBestWaifu:
This a proposal to implement personalities trying to avoid the pitfalls of #10717, namely that it was too convoluted and defined a behaviour the AI could not execute correctly. So, this one tries to be simpler and start from the fundamental block of policy rather than more complex decisions.

Instead of defining possibly dozens of numbers in Nations that may or may not mean what they seem, you define an AutomationProfile in a separate JSON, listing what to increase and what to decrease, with properties defined in a Profile class for each major area. Those modifiers operate in the very same way as victory focus does but with half the strength, serving as tiebreakers but mostly not overriding the victory focus. 

Civs receive the profile that has the name indicated in a profileName property in Nations, or has the same name as the civ, or is called Default, in that order. This allows for a single profile to be applied to multiple civs and for mods to manipulate all civs with a single profile. 

This PR only implements policy, but, if accepted, research and buildings will be done in the very same way. 

Left the readings from JSON commented because what it should be called or wether it should be a separate file at all is probaly to be decided by someone who knows better.

### DIFF
--- a/core/src/com/unciv/logic/automation/profile/AutomationProfile.kt
+++ b/core/src/com/unciv/logic/automation/profile/AutomationProfile.kt
@@ -2,4 +2,11 @@ package com.unciv.logic.automation.profile
 
 class AutomationProfile {
     val policy = PolicyProfile()
+
+    fun clone():AutomationProfile{
+        val toReturn = AutomationProfile()
+        toReturn.policy.increase = policy.increase.clone() as ArrayList<String>
+        toReturn.policy.decrease = policy.decrease.clone() as ArrayList<String>
+        return toReturn
+    }
 }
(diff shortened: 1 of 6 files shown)


SeventhM:
Immediate things throwing me off
1. If the intent is to be primarily dealt with in the ruleset, why is this in `logic.auotmation` and not `models.ruleset.automation`, `models.ruleset.nation`, or even just `models.ruleset`?
2. As a modder facing file, and given the idea is to eventually add more AI personality stuff to this file, it seems to me that the name of "Personality.kt"/"Personality.json" is fine. "Profile.json" seems misleading and "AutomatedProfile.json" seems like a mouthful for a simple idea
3. I'm not sure if it makes sense that Civilization has a transient profile and not Nation, given this information doesn't change between turns or even between games, unless the idea is to, like Civ 5, add a level of randomness to the AI's personalities

PR #10794, opened 2023-12-22
Save city focus as string, to allow changes in city focuses without breaking games

yairm210:
Prerequisite for #10745 since that changes city focuses

Checked with existing games with focus, focuses work 100% :)


yairm210:
ACTUALLY this should DEFINITELY be merged from this version, because if this doesn't exist in version n, then when n+1 comes out and it contains new string values, then version n would break on it!
So this needs to be in a version *before* the actual change

yairm210:
@SomeTroglodyte yes, adding is easy, getting rid of old values is harder, this solution is a one-time solution after which you don't need to think about it which is why long-term it's better than enums even if we're only adding

PR #10797, opened 2023-12-22
Add <every [amount] turns> conditional unique

PLynx01:
I would like to add an another unique. It has been discussed on Discord recently, so I decided to implement it myself. Due to unique system limitations, it is based on turn divisibility by the amount parameter.

If you have some remarks, share them in the comments.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -764,7 +764,7 @@ enum class UniqueType(
     TriggerUponConstructingBuildingCityFilter("upon constructing [buildingFilter] [cityFilter]", UniqueTarget.TriggerCondition),
     TriggerUponGainingUnit("upon gaining a [baseUnitFilter] unit", UniqueTarget.TriggerCondition),
     TriggerUponTurnEnd("upon turn end", UniqueTarget.TriggerCondition),
-    TriggerEveryTurns("every [amount] turns", UniqueTarget.TriggerCondition),
+    TriggerEveryTurns("every [positiveAmount] turns", UniqueTarget.TriggerCondition),
 
     TriggerUponFoundingPantheon("upon founding a Pantheon", UniqueTarget.TriggerCondition),
     TriggerUponFoundingReligion("upon founding a Religion", UniqueTarget.TriggerCondition),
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -764,6 +764,7 @@ enum class UniqueType(
     TriggerUponConstructingBuildingCityFilter("upon constructing [buildingFilter] [cityFilter]", UniqueTarget.TriggerCondition),
     TriggerUponGainingUnit("upon gaining a [baseUnitFilter] unit", UniqueTarget.TriggerCondition),
     TriggerUponTurnEnd("upon turn end", UniqueTarget.TriggerCondition),
+    TriggerEveryTurns("every [amount] turns", UniqueTarget.TriggerCondition),
```
yairm210, line 767: Should be positiveAmount

yairm210:
I think we should add this as a general conditional, <every [n] turn>, to allow for many effects that will change over time 😁
And then 'trigger every nth tirn' with be <upon turn end> <every [n]th turn>
Problem is 2nd, 3rd, don't mesh well with that - need better text that accommodate them

yairm210:
We don't need 'every turn' since I'm talking about a conditional, that would be equal to always true
"when turn divides by [positiveAmount]" is technically correct but... not nice

SeventhM:
> <upon turn end> <every [n]th turn>

Even as a general conditional, it seems to me like "<every [n] turns>" makes more sense than wrestling with reading `1th`, `2th`, `3th`

yairm210:
Sure, let's go with `every [positiveAmount] turns` as a conditional
We can rephrase later if we find something better

PLynx01:
@yairm210 

Can you have a look here? This PR is not merged yet. Could you tell me what is the reason?

yairm210:
As I said, I want this to be a *conditional* not a special trigger
It could then be attached to other uniques, to depend on the turn we're on to activate effects.
For example an observatorygiving double science boosts every 5 turns or whatnot. Units that are weak most of the time but sometimes become strong, or vice versa.

PLynx01:
> As I said, I want this to be a *conditional* not a special trigger
> It could then be attached to other uniques, to depend on the turn we're on to activate effects.
> For example an observatorygiving double science boosts every 5 turns or whatnot. Units that are weak most of the time but sometimes become strong, or vice versa. 

I made it as TriggerCondition. Why it should be a Conditional?

yairm210:
I don't know how many times I can say the same thing, my man.
I want a conditional we can add to *any* unique to limit it to specific turns. TriggerUniques are a very specific thing that only applies to triggered one-time effects.

PLynx01:
Here is the Conditional (not TriggerCondition) version of this unique #10871

SeventhM:
Actually, @yairm210 on second thought, couldn't this PR do "<every [positiveAmount] turns> <for [positiveAmount] turns>" to get the same effect?

PR #10798, opened 2023-12-22
Update Romanian.properties

Tomkode:
I've added translations for some of the phrases.


Tomkode (commented):
I wrote [civBame] instead of [civName]. Now it should pass the tests.

PR #10824, opened 2023-12-28
Trigger 'capture city' as a unit trigger

yairm210:
This will default back to civ trigger if the unique is not a unit trigger, so there should be 100% backwards compatibility


SeventhM (approved):

SeventhM:
For the record, I'm half wondering if the better solution is for `triggerCivwideUnique` to always call `triggerUnitwideUnique` if there's a relevantUnit

PR #10827, opened 2023-12-28
New Citizen Focus Options - again

yairm210:
Redo of #10745 

@itanasi we can just remove the compatibility increment and then merge it :)


itanasi:
Well, that was eagle eyed of you.

Done. And found a spot where accidentally doing Food*6 weight in Automation?

PR #10828, opened 2023-12-29
Add unique to allow for generalized great generals

SeventhM:
Couple of notes
1. The unique is somewhat intentionally unintuitive. The unique is not based on the Great Unit itself, but rather for  the unit that is attacking. That is to say "Can be earned through combat <for [Land] units>" is only valid if a land unit is attacking and "Can be earned through combat <vs [Water] units>" is only valid vs water units
2. I'm pretty sure that despite the name, most of our great people stuff functions all the same for non great people. As such, I decided to forgo checks for a great person here for consistency
3. In Battle.kt, it checked for a Great Person with the "War" label. This is inconsistent with how it was treated in GreatPersonManager. There, it was hard coded to strictly work with a "Great General" unit. That is to say, a fake "General Great" with the unique could still speed up the gain of a Great General despite having a different name. With how things are rearranged, this is no longer possible
4. `UniqueType.GreatPersonEarnedFaster.params[0]` uses "greatPerson", which I believe is an alias for "unitName". However, given neither the combat system of generating great people or the passive way (via buildings/specialists) is exclusive to great people, I've instead decided to treat it as an alias for "unitFilter" here, as it might be generally expected
5. Similar to #9614, I've added an entry in `BackwardsCompatibility` to migrate the great points over. A neat part of how I did things is that a ruleset with no unique based defined Great General and just a unit and a ruleset with a basic Great General unit with the unique should function identically once migrated

### DIFF
--- a/core/src/com/unciv/logic/BackwardCompatibility.kt
+++ b/core/src/com/unciv/logic/BackwardCompatibility.kt
@@ -39,12 +39,12 @@ object BackwardCompatibility {
 
     fun GameInfo.migrateGreatGeneralPools() {
         for (civ in civilizations) civ.greatPeople.run {
-            if (pointsForNextGreatGeneral >= pointsForNextGreatGeneralCounter["Great Genreral"]) {
+            if (pointsForNextGreatGeneral >= pointsForNextGreatGeneralCounter["Great General"]) {
                 pointsForNextGreatGeneralCounter["Great General"] = pointsForNextGreatGeneral
             } else pointsForNextGreatGeneral = pointsForNextGreatGeneralCounter["Great General"]
 
 
-            if (greatGeneralPoints >= greatGeneralPointsCounter["Great Genreral"]) {
+            if (greatGeneralPoints >= greatGeneralPointsCounter["Great General"]) {
                 greatGeneralPointsCounter["Great General"] = greatGeneralPoints
             } else greatGeneralPoints = greatGeneralPointsCounter["Great General"]
         }
--- a/core/src/com/unciv/logic/BackwardCompatibility.kt
+++ b/core/src/com/unciv/logic/BackwardCompatibility.kt
@@ -37,6 +37,19 @@ object BackwardCompatibility {
         removeTechAndPolicies()
     }
 
+    fun GameInfo.migrateGreatGeneralPools() {
+        for (civ in civilizations) civ.greatPeople.run {
+            if (pointsForNextGreatGeneral >= pointsForNextGreatGeneralCounter["Great Genreral"]) {
+                pointsForNextGreatGeneralCounter["Great General"] = pointsForNextGreatGeneral
+            } else pointsForNextGreatGeneral = pointsForNextGreatGeneralCounter["Great General"]
+
+
+            if (greatGeneralPoints >= greatGeneralPointsCounter["Great Genreral"]) {
+                greatGeneralPointsCounter["Great General"] = greatGeneralPoints
+            } else greatGeneralPoints = greatGeneralPointsCounter["Great General"]
+        }
+    }
+
     private fun GameInfo.removeUnitsAndPromotions() {
         for (tile in tileMap.values) {
             for (unit in tile.getUnits().toList()) {
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -496,6 +496,8 @@ object Battle {
                     }.toMutableList()
             // For compatibility with older rulesets
             if (greatGeneralUnits.isEmpty() && civ.gameInfo.ruleset.units["Great General"] != null)
+                && civ.gameInfo.ruleset.units.values.none { 
+                    it.hasUnique(UniqueType.GreatPersonFromCombat, StateForConditionals.IgnoreConditionals) })
                 greatGeneralUnits += civ.gameInfo.ruleset.units["Great General"]!!
 
             for (unit in greatGeneralUnits) {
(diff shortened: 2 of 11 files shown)


core/src/com/unciv/logic/BackwardCompatibility.kt
```diff
@@ -37,6 +37,19 @@ object BackwardCompatibility {
         removeTechAndPolicies()
     }
 
+    fun GameInfo.migrateGreatGeneralPools() {
+        for (civ in civilizations) civ.greatPeople.run {
```
yairm210, line 41: Please add vals for oldName and newName

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -482,19 +483,33 @@ object Battle {
...
+            val greatGeneralUnits = civ.gameInfo.ruleset.units.values
+                    .filter { it.hasUnique(UniqueType.GreatPersonFromCombat, stateForConditionals) }
+                    // Check if the unit is allowed for the Civ, ignoring build constrants
+                    .filter { !it.getRejectionReasons(civ).any { reason ->
+                        reason.type != RejectionReasonType.Unbuildable &&
+                        reason.type != RejectionReasonType.CannotBeBuilt &&
+                        reason.type != RejectionReasonType.CannotBeBuiltUnhappiness &&
+                        // Allow Generals even if not allowed via tech
+                        reason.type != RejectionReasonType.RequiresTech &&
+                        reason.type != RejectionReasonType.Obsoleted }
```
yairm210, line 495: Should be something like `val ignorableRejectionReasons = setOf(...)`
And then this should be `reason,type !in ignorableRejectionReasons`

core/src/com/unciv/logic/civilization/managers/GreatPersonManager.kt
```diff
@@ -19,8 +19,10 @@ class GreatPersonManager : IsPartOfGameInfoSerialization {
     /** Base points, without speed modifier */
     var pointsForNextGreatPersonCounter = Counter<String>()  // Initial values assigned in getPointsRequiredForGreatPerson as needed
     var pointsForNextGreatGeneral = 200
+    var pointsForNextGreatGeneralCounter = Counter<String>() // Initial values assigned when needed
 
     var greatPersonPointsCounter = Counter<String>()
+    var greatGeneralPointsCounter = Counter<String>()
```
yairm210, line 25: Why do we need this as a separate counter? We can just use the regular greatPersonPointsCounter
The same unit shoudn't be able to be gained in both ways

core/src/com/unciv/logic/civilization/managers/GreatPersonManager.kt
```diff
@@ -33,8 +35,8 @@ class GreatPersonManager : IsPartOfGameInfoSerialization {
         toReturn.freeGreatPeople = freeGreatPeople
         toReturn.greatPersonPointsCounter = greatPersonPointsCounter.clone()
         toReturn.pointsForNextGreatPersonCounter = pointsForNextGreatPersonCounter.clone()
-        toReturn.pointsForNextGreatGeneral = pointsForNextGreatGeneral
-        toReturn.greatGeneralPoints = greatGeneralPoints
```
yairm210, line 37: Please retain this for the intermediate versions, so that in multiplayer users with older version don't lose their progress
Actually in multiplayer the newer players would lose their progress, so perhaps having one version where we add the new counters and persist them...?

SeventhM:
> no need for another Counter.

It is possible to generate generals through both methods. This situation kinda forces generals to need a separate counter to the gpp counter, or else you run into issues when needing to increase the value after spawning a general. You also need this to be a counter in order to allow more than one general (probably the easiest way to enable Admirals with little effort

core/src/com/unciv/logic/BackwardCompatibility.kt
SeventhM: I'm somwewhat confused... the vals are to simplify the data, but they'd only be used in the if statement, no? The assignment still needs the actual members

SeventhM (commented):
Though, maybe the more worthwhile thing to do is to actually check the uniqueMap instead of relying on StateForConditionals so I don't even need this import

core/src/com/unciv/logic/BackwardCompatibility.kt
yairm210: No, I mean
`val oldName = "Great Genreral"
val newName = "Great General"`

SeventhM: Oh, I didn't realize I accidentally wrote a typo. There no old/new name, that's just me accidentally typing an extra r

yairm210: All the more reason to make this a val ;)

yairm210 (approved):
I still don't like the hardcoded strings all over the migration but it's not enough for me to refuse this.
I do want @SomeTroglodyte's opinion before merging.

SeventhM:
Sorry, started hopping online a bit late today. Also, I won't have access to a computer today, so I probably couldn't do much besides add the val

The way I see it, adding a val sounds fine enough (I could do it in a few seconds probably, even not on a computer with Studio), but it seems like overkill for what's essentially a local constant for just 7 (really 6) lines. Putting it in `Constants` more reasonable since it's also used as the default in `Battle` if there's no units with the unique, but that'll be overkill when we remove the migration code later. Unless the concern is stuff like typos, I'd expect a constant for such a small amount of code to be covered with comments, but the name seems obvious enough, so at that point, if clarity is the concern, kdocs seems like the better alternative

PR #10829, opened 2023-12-29
AI Builds Courthouse immediately

tuvus:
This PR doesn't really address any problems. Since the AI will only buy buildings earlier in the turn than annexing a city, the AI will at least wait to buy a courthouse in the next turn, if at all. So this PR adds an extra buy courthouse immediately after annexing behavior.

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -369,7 +369,7 @@ object NextTurnAutomation {
                     && !city.isInResistance() && !civInfo.hasUnique(UniqueType.MayNotAnnexCities)
             ) {
                 city.annexCity()
-                
+
                 // Immediately buy a courthouse or equivalent if we can
                 if (city.cityStats.hasExtraAnnexUnhappiness() && civInfo.getHappiness() < 35) {
                     val courthouseBuilding = city.cityConstructions.getBuildableBuildings()
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -369,6 +369,16 @@ object NextTurnAutomation {
                     && !city.isInResistance() && !civInfo.hasUnique(UniqueType.MayNotAnnexCities)
             ) {
                 city.annexCity()
+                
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -369,6 +369,16 @@ object NextTurnAutomation {
...
                 city.annexCity()
+
+                // Immediately buy a courthouse or equivalent if we can
+                if (city.cityStats.hasExtraAnnexUnhappiness() && civInfo.getHappiness() < 35) {
+                    val courthouseBuilding = city.cityConstructions.getBuildableBuildings()
+                        .filter {it.hasUnique(UniqueType.RemoveAnnexUnhappiness)}
+                        .filter { it.getStatBuyCost(city, Stat.Gold)!! >= civInfo.gold }
+                        .minByOrNull { it.getStatBuyCost(city, Stat.Gold)!! }
+                    if (courthouseBuilding != null)
+                        city.cityConstructions.purchaseConstruction(courthouseBuilding, city.cityConstructions.constructionQueue.indexOf(courthouseBuilding.name), false)
```
yairm210, line 380: The proper place for this is in useGoldAutomation
Current gold automation is *before* city automation, so that will use the gold on other things and here you'll end up with an empty treasury
Also, intiutively, that's where we do the purchase automations, so...

tuvus: We purchase unit upgrades in the AutomateUnits() method so this is not a new thing to do here. The AI does not necessarily use all its gold in useGoldAutomation (ex `civ.gold / 3 >= building cost`) so buying the courthouse here can save the turn of extra unhappiness. Also for some reason, the use gold for building automation sorts the cities from biggest population to smallest population which would not prioritize the courthouse. 
However, sorting the cities this way might not be a good idea in the first place since bigger cities can build buildings faster which means the gold is not well spent.

I think this is an argument of readability over AI behavior improvement, which can go either way.

yairm210: If we want specific behavior for courthouses, that should also be in GoldAutomation
That should also be where we purchase unit upgrades, I was unaware that there was a rogue element but it's not what we want.
We want the decision-making to be explicit in one place, so that then we can decide what is more important (for example, courthouses > unit upgrades > buying other buildings)

PR #10839, opened 2023-12-30
Fix conditionals being ignored for some triggers

SeventhM:


SeventhM:
Huh, that... complicates things. Timing uniques naturally don't need to follow conditionals, but that also means they also can fail conditionsApplies checks if thrown everywhere. But these conditionsApplies checks are a necessity (e.g., "This Unit takes [200] damage <for [Human player] units>" don't work as intended without it)


Hmm..., going to have to think up a different PR to fix it, because it's definitely out of scope here

PR #10840, opened 2023-12-31
Avoid Growth Prevents New Population

itanasi:
As it says on the tin


SeventhM:
I'm not sure I agree with the intent. This is a feature of civ 4, but not civ 5

PR #10841, opened 2023-12-31
Show carried over production from obsolete units in city queue when you get a 'Free Tech'

soggerr:
Without this, when you get a mid-turn free tech like Oxford University, the units in city queues are updated but the hammers aren't carried over from the obsolete unit until the end of turn. This means that to the player it shows 0 progress for all the new units in queue and appears like the progress from obsolete units was lost which can be disconcerting and makes strategizing during that turn difficult.

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/TechManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/TechManager.kt
@@ -369,12 +369,17 @@ class TechManager : IsPartOfGameInfoSerialization {
                 .toMutableList()
         }
 
-        // As long as TurnManager does cities after tech, we don't need to clean up
-        // inProgressConstructions - CityConstructions.validateInProgressConstructions does it.
-
         // Add notifications for obsolete units/constructions
         for ((unit, cities) in unitUpgrades) {
             if (cities.isEmpty()) continue
+
+            //The validation check happens again while processing start and end of turn,
+            //but for mid-turn free tech picks like Oxford University, it should happen immediately
+            //so the hammers from the obsolete unit are guaranteed to go to the upgraded unit
+            //and players don't think they lost all their production mid turn
+            for(city in cities)
+                city.cityConstructions.validateInProgressConstructions()
+
             val locationAction = LocationAction(cities.asSequence().map { it.location })
             val cityText = if (cities.size == 1) "[${cities.first().name}]"
                 else "[${cities.size}] cities"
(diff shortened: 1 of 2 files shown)


soggerr:
validateInProgressConstructions does everything needed, and there is no consequence to run it more than once in a turn when this occurs. It doesn't seem to me like exposing it public should be a concern and keeps all that upgrade logic in one place.

core/src/com/unciv/logic/civilization/managers/TechManager.kt
```diff
@@ -369,12 +369,17 @@ class TechManager : IsPartOfGameInfoSerialization {
...
         // Add notifications for obsolete units/constructions
         for ((unit, cities) in unitUpgrades) {
             if (cities.isEmpty()) continue
+
+            //The validation check happens again while processing start and end of turn,
+            //but for mid-turn free tech picks like Oxford University, it should happen immediately
+            //so the hammers from the obsolete unit are guaranteed to go to the upgraded unit
+            //and players don't think they lost all their production mid turn
+            for(c in cities)
+                c.cityConstructions.validateInProgressConstructions()
```
yairm210, line 381:
```suggestion
            for (city in cities)
                city.cityConstructions.validateInProgressConstructions()
```

yairm210: This should go after line 369, since that's where we deal with rebuilding the construction queue it's also where we should validate it

soggerr: I thought about adding it to that loop, but that loop is over every single city. I mean the wasted checks aren't really going to affect performance much because the number of cities and depth of inProgressConstructions is probably never going to be too large, but it felt inefficient?

I thought about putting it in the transform function, but it is possible a single city could have two obsolete units from the same tech so it would require replacing validateInProgressConstructions with a new function that handles one unit replacement at a time (That replacement seems more logical if we can assume that tech is the only way to obsolete a unit, then we could stop calling validate function every start and end of turn and just only handle it when tech upgrades occur).

I am up for any of the options, you choose and I will change it:

1. leave where it is (I will fix the loop variable name)
2. move call to line 369 and run it on every city
3. rewrite validateInProgressConstructions to only happen on tech upgrades and not on every turn (probably replace it with a new function something like 'handleObsoleteConstruction(oldConstruction, newConstruction)' that does the queue replacement and inProgressConstruction transfer all at once)

yairm210: I see, this is fine

yairm210 (approved):

soggerr:
I force push the first one to make the matching branch in my fork, the force push on the change was just a mistake reusing the same command from bash history. In this case the only difference is it forces a new hash.

PR #10847, opened 2023-12-31
Worker remove feature far away from cities fix

tuvus:
Fixes part of #10776
This PR fixes the case where the worker AI wants to remove a marsh or forest far away from its borders for no reason. Workers now only remove marshes or forests if they are on a tile neighboring an owned tile and will be in the range of the city.

Note: The one-time production bonus still needs to be taken into account.


tuvus:
I am not sure how this error relates to the changes that I have made. I did not change any other file and it says the test will only pass when game.png exists. I only got this error on my local machine after I merged master into this branch.

Issue #10858, opened 2024-01-03
Existing translations using square brackets that do not match a placeholder


SomeTroglodyte:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

from source

### Describe the bug

As mentioned in #10391, we have a few cases of translations using square brackets that do not match a placeholder:

Extra placeholder `[새 게임]` in `Korean` for entry `To create a multiplayer game, check the 'multiplayer' toggle in the New Game screen, and for each human player insert that player's user ID.`
Extra placeholder `[amount]` in `Indonesian` for entry `An enemy [unit] was destroyed while attacking [cityName]`
Extra placeholder `[amount2]` in `Indonesian` for entry `An enemy [unit] was destroyed while attacking [cityName]`
Extra placeholder `[relAmount]` in `Vietnamese` for entry `Population loss from nuclear attacks [relativeAmount]% [cityFilter]`
Extra placeholder `[...]` in `Hungarian` for entry `'The art of war teaches us to rely not on the likelihood of the enemy's not attacking, but rather on the fact that we have made our position unassailable.' - Sun Tzu`
Extra placeholder `[...]` in `Hungarian` for entry `'The ancient Oracle said that I was the wisest of all the Greeks. It is because I alone, of all the Greeks, know that I know nothing' - Socrates`
Extra placeholder `[...]` in `Hungarian` for entry `'Architecture has recorded the great ideas of the human race. Not only every religious symbol, but every human thought has its page in that vast book.' - Victor Hugo`
Extra placeholder `[...]` in `Hungarian` for entry `'Bushido is realized in the presence of death. This means choosing death whenever there is a choice between life and death. There is no other reasoning.' - Yamamoto Tsunetomo`
Extra placeholder `[tvenkinio]` in `Lithuanian` for entry `'Don't clap too hard - it's a very old building.' - John Osbourne`

Some of these are mistakes we would like to catch with a unit test, others seem to be intentional stylistic choices - and those would also fail such a test, so this issue is meant to see if we can't do the "good" ones differently.

As far as I can tell, the interesting lines were done by - 
#6158 : @saud2410 
#8915 : @balazsszalab 
 ... would you mind replacing these square brackets with different symbols? Everything but `[]`, `<>`, `{}` or `«»` should work.
#7060 : ghosted account - we'll need to fix this ourselves

The lines looking like mistakes were last touched by -
#7644 @Kensvin28 
 ... please fix?
[...]

PR #10860, opened 2024-01-03
Update Korean.properties

saud2410:
Fixing #10858


saud2410:
@SomeTroglodyte I would readily follow suggestion if it's related to development in any way.

#6158 was 2022 so I'm not sure why i used `[ ]` but probably i thought it's more visible than `' '`. The reason is 1) my translation is undisciplined. 2) `' '` feels lack, at least `" "` is more comfortable. Hangeul, our character system have many short lines. And we're kinda used to see Chinese character which have tons of short lines similar to comma. `[ ]` looks similar to `【 】` which have been used as a distinctive quotation mark even among Chinese characters.

yairm210:
prepare version

yairm210:
Version rollout PR created

PR #10863, opened 2024-01-03
Unit test against unmatched placeholders in a translation

SomeTroglodyte:
Forbid free use of square brackets in translations in exchange for more mistake protection - once the existing violations are all fixed ...we're getting there soon. Replaces #10391.


yairm210:
YESS

PR #10866, opened 2024-01-03
City-stationed unit icons get a circular touchable area

SomeTroglodyte:
Started as pure experiment: Could **maybe** help with the "can't bombard enemy to the north" problem. Emphasis "maybe" as in Ink Dots. What do you think?
image

Bombarding to the south, SE or SW can be tricky too, as we know, but improving that with rounded clickable areas might not be that easy. E.g. the CityButton:
image
... "cutting off" those corners would not help much (and would need more code). I would need to see an actual situation to look for opportunities, and I've forgotten which issues dealt with that, or am too lazy to search for them atm...


yairm210:
This looks better than the current situation even if it doesn't solve the bombard problem

yairm210 (approved):

PR #10624, opened 2023-11-29
Change the icons from some units/buildings

SeventhM:
Note: I don't have any of the attribution for any of these. These icons have been in Rekmod since before I started editing it. Maybe ask @ravignir on that one

PR #10867, opened 2024-01-03
Change the icons from some units/buildings

SeventhM:
Thanks to not feeling like trying to fix the conflicts (likely would need to fiddle with the command line to fix, something that I haven't set up for well yet), I'll just reopen #10624 as a separate PR

@yairm210 @SomeTroglodyte Neither of you have still answered my question [here](https://github.com/yairm210/Unciv/pull/10624#issuecomment-1837640442) regarding whether or not [k4zoo's Grand Temple trace](https://github.com/k4zoo/Civ5ExpansionMod/blob/master/Images/BuildingIcons/Grand%20Temple.png) is fine to add to this PR or not, given it is a trace of Civ 5's image, but not really a direct copy like the other traces


yairm210:
A quick search yielded this:
https://www.reddit.com/r/AdobeIllustrator/comments/170yhxi/image_tracing_photo_copyright_questions/

So I'd err on the side of caution

Issue #10868, opened 2024-01-03
Stat on tile unique doubling on improvement tiles (again)


Caballero-Arepa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.9.15

### Describe the bug

This bug was supposedly fixed in the update 4.9.13 or around that, but it is still here with us.

### Steps to Reproduce

1. To reproduce, you can use [this version of the Latam mod](https://github.com/Caballero-Arepa/Latin-American_Civs/releases/tag/Honduras), because I'll add a restriction to only improved tile to the unique, which will fic the issue for me.
2. Start a new game with Honduras civ, and settle next to a Bananas tile.

### Screenshots

![image](https://github.com/yairm210/Unciv/assets/78449553/bfe84a83-2a04-476d-b8a9-7cab8afcc8ee)
![image](https://github.com/yairm210/Unciv/assets/78449553/ee9f82dc-1c92-4885-bff5-ce73d9ae6ffa)
![image](https://github.com/yairm210/Unciv/assets/78449553/17d25a7e-5175-4efc-b1de-9fd79048d7be)
(Unique)
![image](https://github.com/yairm210/Unciv/assets/78449553/372b7616-d5ac-4a11-91cc-44deb11ff7e6)

<details><summary>### Link to save file</summary>

```
[...]

PR #10878, opened 2024-01-04
Add terrain filters for resources, any terrain, or "improved"

SeventhM:


yairm210 (approved):

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -495,7 +496,7 @@ open class Tile : IsPartOfGameInfoSerialization {
     /** Implements [UniqueParameterType.TerrainFilter][com.unciv.models.ruleset.unique.UniqueParameterType.TerrainFilter] */
     fun matchesSingleTerrainFilter(filter: String, observingCiv: Civilization? = null): Boolean {
         return when (filter) {
-            "All" -> true
+            "All", "Terrain" -> true
```
yairm210, line 499: Under what situations would you use "Terrain"?

SeventhM: https://github.com/yairm210/Unciv/issues/10868#issuecomment-1877771043

PR #10880, opened 2024-01-05
Spectator now receives important diplomatic notifications again

tuvus:
Fixes a complaint posted in Discord. The problem came from changing Civilization.getKnownCivs() to not include spectators.


yairm210 (approved):
Good! It's important to distinguish other nations that we can interact with vs non-interactable 👍🏿

PR #10889, opened 2024-01-07
Add game year conditionals to complement turn number ones

SomeTroglodyte:
Follow-up to #10871, thanks to @PLynx01 
...did I get the `ConditionalTimedUnique` doc right?

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Unique.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Unique.kt
@@ -228,9 +228,15 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
             UniqueType.ConditionalBeforeYears -> checkOnCiv { gameInfo.getYear(0) < condition.params[0].toInt() }
             UniqueType.ConditionalAfterYears -> checkOnCiv { gameInfo.getYear(0) >= condition.params[0].toInt() }
             UniqueType.ConditionalEveryYears -> checkOnCiv {
+                // turns to years is complex. Therefore, we cannot expect a certain multiple of the period parameter wo be hit exactly by the game year.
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -224,6 +224,14 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
             UniqueType.ConditionalEveryTurns -> checkOnCiv { gameInfo.turns % condition.params[0].toInt() == 0}
             UniqueType.ConditionalBeforeTurns -> checkOnCiv { gameInfo.turns < condition.params[0].toInt() }
             UniqueType.ConditionalAfterTurns -> checkOnCiv { gameInfo.turns >= condition.params[0].toInt() }
+            UniqueType.ConditionalEveryTurns -> checkOnCiv { gameInfo.turns % condition.params[0].toInt() == 0 }
+            UniqueType.ConditionalBeforeYears -> checkOnCiv { gameInfo.getYear(0) < condition.params[0].toInt() }
+            UniqueType.ConditionalAfterYears -> checkOnCiv { gameInfo.getYear(0) >= condition.params[0].toInt() }
+            UniqueType.ConditionalEveryYears -> checkOnCiv {
+                val period = condition.params[0].toInt()
+                gameInfo.getEquivalentTurn() == 0 || // This will also avoid calling gameInfo.getYear(-1) when turns==0
+                    gameInfo.getYear(0) / period > gameInfo.getYear(-1) / period
```
yairm210, line 233: .........
Year -1
That's not even defined, so using it to test for anything sounds like a bad idea

yairm210: What are you actually trying to do here? I confuse

PR #10894, opened 2024-01-08
Allow map editor to generate smaller than "tiny" Pangaea maps

SomeTroglodyte:
https://github.com/yairm210/Unciv/commit/b9299ac50dbef13004f990b9050c60ddb10fd93f broke an edge case: Enter a very small map size, and it becomes impossible to get continent size >= 25, resulting in an infinite loop.

This picture was not possible without this fix:
<details>

image
</details>

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
@@ -153,9 +153,9 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
 
     private fun createPangaea(tileMap: TileMap) {
         val largeContinentThreshold = (tileMap.values.size / 4).coerceAtMost(25)
-        var maxRetries = 200  // A bit much but enough to reach waterThreshold -1
+        var retryCount = 200  // A bit much but when relevant (tiny map) an iteration is relatively cheap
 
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
```diff
@@ -152,7 +152,10 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
     }
 
     private fun createPangaea(tileMap: TileMap) {
-        do {
+        val largeContinentThreshold = (tileMap.values.size / 4).coerceAtMost(25)
+        var maxRetries = 200  // A bit much but enough to reach waterThreshold -1
```
yairm210, line 156: Even 100 should be enough for that, no?

core/src/com/unciv/logic/map/mapgenerator/MapLandmassGenerator.kt
```diff
@@ -152,7 +152,10 @@ class MapLandmassGenerator(val ruleset: Ruleset, val randomness: MapGenerationRa
...
+
+        while(--maxRetries >= 0) {
```
yairm210, line 158: Don't use a parameter called "max retries" to save the number of retries left :(

yairm210 (approved):

PR #10900, opened 2024-01-08
Display mod release versions

SomeTroglodyte:
First off: This is far from perfect, but getting it better is futile without some fundamental redesign. Thus - maybe merge instead of throwing it away, then address the shortcomings later.

Inspiration came from ##10868, and this is a follow-up to #10881.

#### Summary
- Mods with releases will now download the latest release instead (10881 did so only when feeding the repo base url to the "download from url" button)
- There's a Version line on the right-hand info pane, fed from these releases
- Modders not doing releases now can include any "version" string in their ModOptions if they want to
- Little fix: Updated disappears after deleting a mod with such a marker
- No visible effect: Completely unrelated "cleanup" of ModConstants - read commit header. But - mods will start losing their empty modConstants additions that our rewrite serializes, and the authors often re-publish. In other words, most modOptions out there have the string `,"constants":{"unitUpgradeCost":{}}` - entirely no-op - in their json. I read so many of them during testing that I looked into why and fixed it... Almost unvoluntarily.

<details><summary>Likely still applicable screenshot</summary>

Look to the right side at the Version: and Updated: rows, then compare with the [latest-release publish date here](https://api.github.com/repos/Caballero-Arepa/Latin-American_Civs/releases/latest), and with the repo's [last push date here](https://api.github.com/repos/Caballero-Arepa/Latin-American_Civs).
image
</details>

#### Problems
- Fetching release info can't be done within the main repo query - no fields. Doing it later is one roundtrip per repo.
- This PR does the extra per-repo query when the center button is clicked. That means the repo data before uses a different date - pudhed_at, while after it will use and save the release published_at. That means it is most likely an up-to-date installed mod will be marked "Updated" until it's clicked - then the Update marker disappears once the roundtrip is done.
- Doing the latest-release query before populating the center column would be far too slow and hit the rate limit much earlier.
- The late-fix approach used here is unable to update the bottom label to include the release notes - same design problem that led to these major "need to replace listeners as their closures carry stale cache data" things already there, wouldn't want to add another. So - Release.body is inactive.
- Most mod releases do not include specific release assets, relying on the "source zip" offer instead. That is what this PR supports. [A counter-example would be this](https://github.com/Jiangeshi/Leader-Mission/releases/tag/%E7%89%88%E6%9C%AC) - this PR would not properly handle that release by using the asset, but would use the source zip like for other mods. I guess. Depends on the REST api after all, what it uses for zipball_url...
- There is a problem with previews/avatars being displayed twice, shortly, but selecting another then going back it will be OK - didn't quite get around to debug. May be it's not even caused by this.
- Still needs (not a new problem) to catch and clean up partially done mod download/installs that throw somewhere underway. Leftover folders and zips will get cleaned once one tries to download the exact same mod, due to temp names being hashed from the mod, but that's not really tidy.
- Did I leave a todo comment to enable parsing a github 'tree/tree_sha' format URL, which would enable fetching of any specific commit state of a mod? If I forgot - noted here, would not cost much.
- Minor thing I forgot: getVersion should omit the tag if the release name _ends_ with the tag, not only if they're equal - Alpha Frontier uses such a scheme.

#### Solutions (future)
* GraphQL allows what the repo search API lacks: Get repo data including latest release in one roundtrip. Major redesign.
* The UI could do with a different ModUIData concept - and reusable button concept. ModUIData being either "installed" or "online" was a bad idea, and making it immutable was worse. One per mod learning all it can, successively, could work better. left and center buttons could share a cache but still be polymorphic - interface. Major redesign.

<details><summary>GraphQL query that supports what we need</summary>

```
query {
  topic(name: "unciv-mod") {
    repositories(orderBy:{field:STARGAZERS, direction:DESC}, first:20) {
      totalCount
      pageInfo {
        endCursor
        hasNextPage
      }
      nodes {
        name
        nameWithOwner
        owner {
          login
          avatarUrl
        }
        description
        pushedAt
        latestRelease {
          name
          tagName
          author{login}
          publishedAt
          description
          resourcePath
          url
          releaseAssets(first:1) {
            nodes {
              name
              downloadUrl
            }
          }
        }
        stargazerCount
        defaultBranchRef{name}
        repositoryTopics(first:10) {
          nodes{topic{name}}
        }
      }
    }
  }
}
```

- can be tested in [Graph_i_QL](https://docs.github.com/en/graphql/overview/explorer).
- the `endCursor` returned previously would go into an `after:` parameter for a next page
- Needs new parser classes and a different ratelimit implementation likely, at least the API has a separate limiter.
- Haven't really dug deeper, but it's likely this API will "spaywall" demand an auth token 😠 
</details>

### DIFF
--- a/core/src/com/unciv/logic/github/GithubAPI.kt
+++ b/core/src/com/unciv/logic/github/GithubAPI.kt
@@ -330,7 +330,9 @@ object GithubAPI {
         owner = release.author
         html_url = release.html_url  // Used as open on Github link
         if (full_name.isEmpty()) full_name = "${owner.login}/${repoName.folderNameToRepoName()}"  // safety - not seen it happen
-        // We could copy `description = release.body` here to get the release notice into the picker bottom pane - problem: It's markdown?
+        // We could copy `description = release.body` here to get the release notice into the picker bottom pane
+        // problems: It's markdown and would look bad? also, must be updated before the buttons are created or we'll need another updateInstalledModUIData for online
+        // description = description + "\n\n" + release.body
         return this
     }
     //endregion
--- a/core/src/com/unciv/ui/screens/modmanager/ModManagementScreen.kt
+++ b/core/src/com/unciv/ui/screens/modmanager/ModManagementScreen.kt
@@ -399,7 +399,7 @@ class ModManagementScreen private constructor(
         syncOnlineSelected(repo.name, button)
         showModDescription(repo.name)
 
-        //todo bug: this date fetch is too late for a correct "Updated" marker
+        //todo bug: this date fetch is too late for a correct "Updated" marker - the unMarkUpdatedMod is a temp kludge
         if (!repo.hasCheckedRelease) {
             repo.hasCheckedRelease = true
             Concurrency.run("GitHubLatestRelease") {
@@ -409,6 +409,10 @@ class ModManagementScreen private constructor(
                     if (repo.release_tag.isNotEmpty() && (selectedMod === repo)) {
                         launchOnGLThread {
                             modActionTable.update(repo) // Updates three things - version, date and "Open Github page" link
+                            installedModInfo[repo.name]?.ruleset?.modOptions?.also {
+                                if (repo.pushed_at == it.lastUpdated)
+                                    unMarkUpdatedMod(repo.name)
+                            }
                         }
                     }
                 } catch (ignore: IOException) {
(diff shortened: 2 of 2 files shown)


yairm210:
This is really a discussion we should have with the mod makers on Discord, what is more comfortable for them.

yairm210:
Yeah, I'm going to go with 'no' for this.
Adds an extra layer of complexity that isn't really necessary, requires more upkeep from modders

PR #10917, opened 2024-01-11
Worker remove feature AI fix

tuvus:
This PR fixes part of #10776, specifically regarding the value of removing a terrain feature so that a better improvement can be built. Also adds one test!

When determining the value of removing a feature, we need to take into account the improvement that can be built afterwards. This change does that and hopefully doesn't include any bugs for mods. The test that was added specifically fails in the version before this change.


tuvus:
Also, I have noticed that the AI is very picky with its improvements now and will replace markets with farms by removing the forest first. So just be aware of this behavior change. It might need to be tweaked again later.

yairm210 (approved):

PR #10918, opened 2024-01-11
AI worker road priority rework

tuvus:
This PR is the main follow-up to #10776 and changes the approach that we use to build roads. The previous road-building AI is the reason for the indecisiveness and lack of overall work being done by the workers. So what exactly does this PR change?

- Before a worker does any logic on choosing which tile to improve, it searches for nearby cities that need to be connected to the road network. This is then cached for the rest of the workers and all tiles nearby that we would like to build a road on are marked in the `tilesOfRoadsToConnectCities` hashmap.
- Now that we have a way to check if we want to build a road on a tile close to the worker efficiently, `getImprovementRanking` can easily tell us how much we should value building the road there based on city size and how many roads need to be built to complete the connection. This allows `findTileToWork` to include desirable roads in its nearby search, moving the essential road building code out of `tryConnectCities`.
- `tryConnectCities` is now only used to go to a farther away road building location since `findTileToWork` has a higher priority than it. It isn't very efficient, it could be better if we can do a `unit.movement.distanceToTiles` combined with the `BFS.step` function since after each step we can check the new tile is in `tilesOfRoadsToConnectCities`.

The result is that the workers do a lot more work. They don't wander around aimlessly like they did before and actually sit down to build improvements. However, they now don't build roads unless they really have to. Causing the road network to be either non-existant or patchy. This, however, is a fix for another time since I am not sure if these changes broke that or if it was always broken and the bad AI hid it.

<details><summary>Example</summary>
The AI here has 5 workers, and most of them are off-screen (it needs to build more). The road connection is very funky because the cities are connected via port and, according to the AI don't need to be connected via road. The workers in this game have been constantly working, which is very good. This is sort of an extreme example since most cities can be connected via sea. Bigger maps with more inland will likely have better results.

NewWorkerRoadAI
NewWorkerRoadAI

</details>

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -280,7 +280,7 @@ class WorkerAutomation(
      */
     private fun getRoadConnectionBetweenCities(unit: MapUnit, city: City): List<Tile> {
         if (city in roadsToConnectCitiesCache) return roadsToConnectCitiesCache[city]!!
-        
+
         val isCandidateTilePredicate: (Tile) -> Boolean = { it.isLand && unit.movement.canPassThrough(it) }
         val toConnectTile = city.getCenterTile()
         val bfs: BFS = bfsCache[toConnectTile.position] ?:
@@ -291,7 +291,7 @@ class WorkerAutomation(
             )
             bfsCache[toConnectTile.position] = this@apply
         }
-        val cityTilesToSeek = HashSet(tilesOfConnectedCities.sortedBy { it.aerialDistanceTo(city.getCenterTile()) })
+        val cityTilesToSeek = HashSet(tilesOfConnectedCities)
 
         var nextTile = bfs.nextStep()
         do {
@@ -309,7 +309,7 @@ class WorkerAutomation(
             if (bfs.hasEnded()) break // No tiles left
             nextTile = bfs.nextStep()
         } while (nextTile != null)
-        
+
         roadsToConnectCitiesCache[city] = listOf()
         return roadsToConnectCitiesCache[city]!!
     }
@@ -327,7 +327,7 @@ class WorkerAutomation(
             return
         }
         val tileToWork = findTileToWork(unit, dangerousTiles)
-        
+
         if (tileToWork != currentTile) {
             debug("WorkerAutomation: %s -> head towards %s", unit.label(), tileToWork)
             val reachedTile = unit.movement.headTowards(tileToWork)
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -294,21 +294,20 @@ class WorkerAutomation(
         val cityTilesToSeek = HashSet(tilesOfConnectedCities)
 
         var nextTile = bfs.nextStep()
-        do {
+        while (nextTile != null) {
             if (nextTile in cityTilesToSeek) {
                 // We have a winner!
-                val cityTile = nextTile!!
+                val cityTile = nextTile
                 val pathToCity = bfs.getPathTo(cityTile)
-                roadsToConnectCitiesCache[city] = pathToCity.toList()
+                roadsToConnectCitiesCache[city] = pathToCity.toList().filter { it.roadStatus != bestRoadAvailable }
                 for (tile in pathToCity) { 
                     if (tile !in tilesOfRoadsToConnectCities)
                         tilesOfRoadsToConnectCities[tile] = city
                 }
                 return roadsToConnectCitiesCache[city]!!
             }
-            if (bfs.hasEnded()) break // No tiles left
             nextTile = bfs.nextStep()
-        } while (nextTile != null)
+        } 
 
         roadsToConnectCitiesCache[city] = listOf()
         return roadsToConnectCitiesCache[city]!!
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -267,25 +273,61 @@ class WorkerAutomation(
...
+            maxSize = HexMath.getNumberOfTilesInHexagon(
+                WorkerAutomationConst.maxBfsReachPadding +
+                    tilesOfConnectedCities.minOf { it.aerialDistanceTo(toConnectTile) }
+            )
+            bfsCache[toConnectTile.position] = this@apply
+        }
+        val cityTilesToSeek = HashSet(tilesOfConnectedCities.sortedBy { it.aerialDistanceTo(city.getCenterTile()) })
+
+        var nextTile = bfs.nextStep()
+        do {
```
yairm210, line 297: Why do/while instead of regular while?

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -373,84 +412,59 @@ class WorkerAutomation(
...
             // Cities that are too far away make the canReach() calculations devastatingly long
             it.getCenterTile().aerialDistanceTo(unit.getTile()) < 20
         }
-        if (candidateCities.none()) return false // do nothing.
+        if (candidateCities.none()) return listOf() // do nothing.
 
-        val isCandidateTilePredicate: (Tile) -> Boolean = { it.isLand && unit.movement.canPassThrough(it) }
+        // Search through ALL candidate cities to build the cache
+        for (toConnectCity in candidateCities) {
+            getRoadConnectionBetweenCities(unit, toConnectCity).filter { it.getUnpillagedRoad() < bestRoadAvailable }
```
yairm210, line 428: What's with the filter? It doesn't actually do anything

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -105,6 +105,12 @@ class WorkerAutomation(
         result
     }
 
+    /** Cache of roads to connect cities each turn */
+    private val roadsToConnectCitiesCache: HashMap<City, List<Tile>> = HashMap()
+    
+    /** Hashmap of all cached tiles in each list in [roadsToConnectCitiesCache] */
+    private val tilesOfRoadsToConnectCities: HashMap<Tile, City> = HashMap()
```
yairm210, line 112: Surely a tile between two cities could link to two cities. In that case do you sort by closest? That's what I gather from 294 and if so that's fine

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
tuvus: Copy and paste. I don't know, I would have to look into it more.

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
tuvus: Actually there is no reason to sort in the first place since we are now converting line 294 to a hashmap instead of a list.
As for your question, I don't exactly understand.

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
tuvus: That is a good question. We also don't seem to need BFS.hasEnded()

tuvus:
I fixed the major road-building bug stated in the original message. It turned out to be in chooseImprovement(). Apparently the function did not like to build roads on resource tiles.
I'll run the AI a little more, but it is looking very good.

And I encountered another crash while writing this...

tuvus:
Fixed it!

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -648,7 +668,21 @@ class WorkerAutomation(
...
+        // Add the value of roads if we want to build it here
+        if (improvement.isRoad() && bestRoadAvailable.improvement(ruleSet) == improvement
+            && tile in tilesOfRoadsToConnectCities) {
+            var value = 1f
+            val city = tilesOfRoadsToConnectCities[tile]!!
+            if (civInfo.stats.statsForNextTurn.gold >= 20)
+                // Bigger cities have a higher priority to connect
+                value += (city.population.population - 3) * .3f
+            // Higher priority if we are closer to connecting the city
+            value += 5 - roadsToConnectCitiesCache[city]!!.size
```
yairm210, line 681: So there will be negative ranking if the list is large? Will the worker do anything that's negative? If not they this road will be perpetually unbuilt

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -267,25 +273,60 @@ class WorkerAutomation(
...
+                for (tile in pathToCity) { 
+                    if (tile !in tilesOfRoadsToConnectCities)
+                        tilesOfRoadsToConnectCities[tile] = city
+                }
+                return roadsToConnectCitiesCache[city]!!
+            }
+            nextTile = bfs.nextStep()
+        } 
 
+        roadsToConnectCitiesCache[city] = listOf()
```
yairm210, line 312: So this function always returns an empty list...?

tuvus: No, if it gets to that line, then there is no road that can connect the city. If there is a road to connect the city, it will be found and returned at line 307 in the bfs.

yairm210 (approved):

PR #10714, opened 2023-12-11
Great Person Points - Rounding changes, Breakdown UI

SomeTroglodyte:
... and since it irked me for ages, I secretly snuck in.... :speak_no_evil: another minor rule change... :speak_no_evil: :speak_no_evil: :speak_no_evil:  (:angel: :smiling_imp: shut up, no spoilers!)  <sub><sup>And I reduced that unit icon a little and moved it off the stuff it was overlapping :see_no_evil: :hear_no_evil: :speak_no_evil: </sup></sub>

Closes #10671 , screenies there still apply. To get that Popup, click/tap anywhere in that GPP's row (do we need to communicate that new option somehow??)

PR #10924, opened 2024-01-12
Don't prematurely round Great person point boni and GPP breakdown UI

### LINKED ISSUE #10671: GPP boost malfunctioning

### Version/OS

4.9.5 apk
Android 13

### Description

Great Person generation boost becomes weird when specialist and wonder is combined. See case D.

A. 1 slot from Workshop(3) = 3 pt
B. 1 slot(3) + Angkor Wat(1) = 4 pt
C. 1 slot(3) + Garden(+25%) = 3 pt
D. ( 1 slot + 1 wonder ) + Garden = 4 pt 🤨

### Steps

1. Download & load savefile. It requires [Difficulty_Sandbox mod](https://github.com/paontv/Difficulty_Sandbox).
2. Enter city screen to see there's 1 engineer specialist(3 pt), Great Mosque of Djenne(1 pt) and Garden is available.
3. Buy Garden to see if great person point changes.
4. Go to next turn to see accumulated eng point rise from 6 to 10, not to 11, meaning it's not a visual but calculation bug.

### Expected Behavior

Math time. ***4 × 1.25 = 5***. But i'm getting only 4 pt. What's the point of Garden?

### Savefile

[Arabia_-_2_turns.zip](https://github.com/yairm210/Unciv/files/13560933/Arabia_-_2_turns.zip)

Screenshot_20231205_212610_Unciv

### Additional Information

Actually Garden seems to work when the value is big enough. When my city have 14 pt then got extra +25%, 3 pt was added thus it became 17 pt. I guess instead of `rnd[ ( specialist + wonder ) × 1.25 ]` current calculation is `rnd( spc × 1.25) + rnd( wnd × 1.25)`.

SomeTroglodyte:
Redo of #10714, more or less from scratch. Math the same. UI more colourful. Re-closes #10671.

<details>

Clicking a specific GPP row towards the left side:
image

Clicking in the big icons column to the right gives the all-gpp breakdown instead:
image

</details>

Got half a mind to do the missing "Specialists" Civilopedia article here so that line can be linked as well - too tired now. Or am I just missing it? Where are Specialists really best explained in Pedia?


core/src/com/unciv/logic/city/GreatPersonPointsBreakdown.kt
```diff
@@ -0,0 +1,157 @@
...
+    /** Represents any source of Great Person Point or GPP percentage boni */
+    data class Entry(
+        /** Simple label for the source of these points */
+        val source: String,
+        /** If true, [counter] represents a percentage bonus */
+        val isPercentage: Boolean,
+        /** In case we want to show the breakdown with decorations and/or Civilopedia linking */
+        val pediaLink: String?,
+        /** For display only - affects all GP */
+        val isAllGPP: Boolean = false,
```
yairm210, line 27: What does this actually mean, though? Very unclear

yairm210:
I'm getting a very icky feeling from this PR and it's hard for me to pin down in words why that is, but there are a couple of obvious factors.
The overarching feeling is that there is too much *unnecessary complexity* here.

Let's try and break it down:
- You have an iterator and size function that return the data to make this an iterable - why? Why an iterable and not, say, an ArrayList<Entry> which would be clearer and less custom code?
- You put two _different_ data types (GPP and bonuses - yes, NOT boni, I know [you're](https://www.reddit.com/r/Diablo/comments/37u65v/can_we_please_stop_calling_bonuses_boni/) [German](https://forum.paradoxplaza.com/forum/threads/its-bonuses-not-boni.1046478/) but the code is English, sorry) in the same iterable, and separate between them using a boolean - this is the equivalent of a factory putting two different products on the same assembly line and then sorting them out at a later stage, it's just bad design (Factorio nonwithstanding)
- The constructor has two wildly different code paths for onlyAllGPPBoni (bad name!) true and false, but this is badly reflected in the code

I think that by fixing these points it would be much more readable

core/src/com/unciv/ui/screens/cityscreen/GreatPersonPointsBreakdownPopup.kt
```diff
@@ -0,0 +1,57 @@
...
+        val ruleset = game.gameInfo!!.ruleset
+        add(MarkupRenderer.render(lines) {
+            game.pushScreen(CivilopediaScreen(ruleset, link = it))
+        })
+
+        addCloseButton()
+        open(true)
+    }
+
+    private fun GreatPersonPointsBreakdown.Entry.toString(isPercentage: Boolean) =
```
yairm210, line 44: Why not call the other tostring...???

core/src/com/unciv/logic/city/GreatPersonPointsBreakdown.kt
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        // Apply percent bonuses
+        for (key in result.keys.filter { it in bonuses }) {
+            result.add(key, result[key] * bonuses[key] / 100)
+        }
+
+        // Round fixed-point to integers, toSet() because a result of 0 will remove the entry (-99% bonus in a certain Mod)
+        for (key in result.keys.toSet())
+            result[key] = (result[key] * 10 + 5) / fixedPointFactor / 10
```
yairm210, line 148: You sure that the round is down on half? Like 10.5 is rounded to 10?

SeventhM:
> * Should we implement the [+100%](http://modiki.civfanatics.com/index.php/Civ5_GlobalDefines) bonus Civ5 has for a golden age?

Maybe, but I think someone should confirm in game first

> This patch:

`"[-25]% [Culture] [in puppeted cities]"` is possible btw, both as is and in conditional form (`"[-25]% [Culture] <in [Puppeted] cities>"`)

Ouaz:
Nice work!

> Got half a mind to do the missing "Specialists" Civilopedia article here so that line can be linked as well - too tired now. Or am I just missing it? Where are Specialists really best explained in Pedia?

In a new article, I guess. ^^

PR #10928, opened 2024-01-13
Fix Github Actions pipelines

yairm210:


SeventhM:
Ahhhh, that explains things

> You mean

Looks like they linked https://github.com/actions/runner-images/issues/8952 which did warn ahead of time, just not in a way we'd catch. Well, we were considering updating our stuff anyways so I guess all this does is increase our timeline

PR #10930, opened 2024-01-14
Fix ModOptions unique parameter types not checked and "uniquetype" ModOptionsConstants

SomeTroglodyte:
... and read other commit messages too. ~Closes 10432.~

### Questions
- As I only minimally renamed the ex-ModOptionsConstants entries, there's now `UniqueType.DisableReligion` and `UniqueType.DisablesReligion`, no problem for the game or modders but ugly for us. Renaming the ModOptions one to ModDisableReligion would be my first idea - but then - rename all the others too?
- UniqueFlags - was always meant as Set, then EnumSet is _still_ not kotlin-ified and thus needs some ugly syntax... Could probably hide the EnumSet implementation detail outside the class, but couldn't decide whether that's desirable. Implementation as Set would be a little more expensive under the hood - class instances instead of bitmask. Also, "doesn't support any conditionals at all" is a very simplistic one - might die in the future should be decide to have a `UniqueType`x`UniqueType` Base Unique vs. Conditional interoperability matrix....
- Automatic parsing into translation templates - which of the _meta_ unique texts do we want to keep out of that?

### ~Example~ (deleted as outdated)

Edit: Suppression reverted.


yairm210:
I like 90% of this but specifically not the suppression unique
I would have done it as a unique modifier that you can attach to a unique that is not user-visible and makes the modchecker not display non-error warnings - that way the *source* and the *suppression of warning* appear in the same place

yairm210:
Let's as the modder amogus. @SeventhM what say you?

SeventhM:
Not at a computer so not completely looking at it right now, but even avoiding the question of a unique modifier (which this probably also should be) it seems odd to me that this is a modOptions unique and not a nation/unit/building/belief/improvement unique. I'm sure while ignoring most, say, unit upgrade warnings, is nice, I not sure all warnings should be that easy to hide off screen. I know it's easier as a modOptions unique to check all of them at once, but it's probably better for the modder to do it at least near the object where it's relevant

core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
```diff
@@ -579,9 +579,14 @@ enum class UniqueParameterType(
     /** Mod declarative compatibility: Define Mod relations by their name. */
     ModName("modFilter", "DeCiv Redux", """A Mod name, case-sensitive _or_ a simple wildcard filter beginning and ending in an Asterisk, case-insensitive""", "Mod name filter") {
         override fun getErrorSeverity(parameterText: String, ruleset: Ruleset):
-            UniqueType.UniqueParameterErrorSeverity? =
-            if ('-' !in parameterText && ('*' !in parameterText || parameterText.matches(Regex("""^\*[^*]+\*$""")))) null
-            else UniqueType.UniqueParameterErrorSeverity.RulesetInvariant
+            UniqueType.UniqueParameterErrorSeverity? = when {
+                BaseRuleset.values().any { it.fullName == parameterText } -> null  // Only Builtin ruleset names can contain '-'
+                parameterText == "*Civ V -*" || parameterText == "*Civ V - *" -> null  // Wildcard filter for builtin
+                '-' in parameterText -> UniqueType.UniqueParameterErrorSeverity.RulesetInvariant
```
SeventhM, line 585: Didn't know this was a thing. What's the issue with dashes in the mod's name?

SeventhM:
I'll need to see the code on this branch, but I almost want to say that for non modOptions suppressions, perhaps the object can be passed in specifically at the point of adding to RulesetError list

SeventhM:
> but that's a major refactor

Perhaps it's just me (I've had plenty of PRs that's just kinda refractors touching half the code base), but if it's just 111 lines solely because that's how many spots you're passing in the object and not actually majorly changing how the code functions, it's not that big a deal. I mean, 100 lines kinda is a pretty small refactor anyways

SeventhM:
> Somehow it will make suppression too easy

I mean, if we give a suppression option, it'll be "too easy" regardless. Probably the best case scenario to avoid that is specifically making sure it's as localized as possible to be as annoying as possible. Still, the whole point is to give a simple way to suppress warnings so imo it's not that big a deal

I do wonder on that branch if `add()` should probably be the spot where you check if it should be suppressed, not after the fact as it is currently

yairm210 (approved):

PR #10932, opened 2024-01-14
[ModCategories] unciv-mod-scenarios

hachchch:
I think mod creators want a scenario mod category.


SeventhM:
Closing for not following documentation. I'm perfectly fine with this (seems like an obvious thing to request), but we should let our GitHub actions handle most of the work for us

PR #10944, opened 2024-01-15
Workers now build forts

tuvus:
This PR works on adding back in fort building AI for workers. I will probably post some other questions regarding the desired behavior.

Is the picture below unreasonable for the fort-building AI? The AI has plenty of workers because I changed their priority and forts have no maintenance cost. They also aren't losing stats for building the forts on improbable tiles.

AIFortBuilding

</details>

Is it really better for a fort to be built on an owned tile vs a neutral tile? Is it bad for a fort to be built adjacent to a city?

FYI, we are only looking at fort building here and whether or not the AI will use the forts. That's a different question entirely.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -369,11 +369,11 @@ object UnitAutomation {
         val tilesThatCanWalkToAndThenPillage = unitDistanceToTiles
             .filter { it.value.totalDistance < unit.currentMovement }.keys
             .filter { unit.movement.canMoveTo(it) && UnitActionsPillage.canPillage(unit, it)
-                    && ((it.canPillageTileImprovement() && !it.getTileImprovement()!!.hasUnique(UniqueType.DefensiveBonus))
+                    && (it.canPillageTileImprovement()
                     || (it.canPillageRoad() && it.getRoadOwner() != null && unit.civ.isAtWarWith(it.getRoadOwner()!!))) }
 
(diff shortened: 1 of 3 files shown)


yairm210:
I don't know what should be considered OK fort building practices 🤔 
I think we need players to say

tuvus:
Okay, I will ask discord then.

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -369,8 +369,8 @@ object UnitAutomation {
         val tilesThatCanWalkToAndThenPillage = unitDistanceToTiles
             .filter { it.value.totalDistance < unit.currentMovement }.keys
             .filter { unit.movement.canMoveTo(it) && UnitActionsPillage.canPillage(unit, it)
-                    && (it.canPillageTileImprovement()
-                    || (it.canPillageRoad() && it.getRoadOwner() != null && unit.civ.isAtWarWith(it.getRoadOwner()!!)))}
+                    && ((it.canPillageTileImprovement() && !it.getTileImprovement()!!.hasUnique(UniqueType.DefensiveBonus))
```
SeventhM, line 372: Units probably should still pillage stat providing forts and/or forts that that damages them

tuvus:
The code is not perfect. However, the AI will build forts and doesn't seem to build too many of them. Currently, they will only build forts in their own land. Like all things, this will take some more tweaking some other time. Frankly the workers building forts is only a visual effect for the players right now since the unit AI doesn't use them enough. Therefore its probably better to work on some other area.

Issue #10951, opened 2024-01-16
RFC: UX: UnitActions on small displays


SomeTroglodyte:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

Unit action buttons on cramped displays overlap fow/diplo/policy/undo/tech. Now exasperated due to build-road-to.

### Desired Solution

Dynamic "paging", flexible, not limited to "show more"/"back" each having a static list of candidate actions. Number of buttons per "page" measures the screen.

### Alternative Approaches

Here's the catch - it's working to a point, but still, with too cramped screens, it MUST fail, as often there is not enough room left between the other WorldScreen elements for even three such buttons.

<details><summary>A series of screenies</summary>

### A perfectly OK situation:
![image](https://github.com/yairm210/Unciv/assets/63000004/d0c72ce3-23af-46ca-b980-5a7fac4b7a6a)
### Less, still OK:
![image](https://github.com/yairm210/Unciv/assets/63000004/5538d08e-2708-47f8-a701-e03208005899)
### Tiny display setting (still landscape!):
![image](https://github.com/yairm210/Unciv/assets/63000004/f4eb1a1d-18a8-4145-9527-73981235edd9)
### Small display setting, and a Worker has more to do:
![image](https://github.com/yairm210/Unciv/assets/63000004/2bdbffc4-8c9d-41a6-91ac-67aaa298c656)

## WIP branch
* "A perfectly OK situation" is unchanged.
### The "Less" from above now looks like this:
![image](https://github.com/yairm210/Unciv/assets/63000004/8a39a095-38e3-4758-bdb7-d61b1284fe82)
* "Tiny display setting" is unchanged as I defined 4 to be the minimum max-buttons-per-page
### Same situation as last, with the Worker:
![image](https://github.com/yairm210/Unciv/assets/63000004/ee3d1024-55a5-4ca7-bede-dbac98c4855c)

</details>

### Questions:
- `@`coders - am I allowed to reorg WorldScreen.update to move UnitActionTable update far enough down that the measurements it needs are done? Below `fogOfWarButton.setPosition`?
[...]

PR #10955, opened 2024-01-16
Unit actions rework

SomeTroglodyte:
The WIP behind #10951


yairm210:
This PR is too big for my brain to process, even piece by piece.
Can you extract the minor changes to a different PR so we can merge that, and this one will be cleaner and contain only the major changes?

PR #10962, opened 2024-01-18
Improved Console autocomplete

SomeTroglodyte:
Needs "just try it" - or read the commit novels. Reading diffs also best done per commit.

Motivation to even look at the code: `unit addp<tab>gr<tab>` - not too helpful, need to backspace to choose the `II` level, and replace spaces. Now it is as quick as can be: `unit addp<tab>gr<tab>i` actually gives you 'Great Generals II'. That was a relatively minor patch, but there's a few more QOL improvements here like - `unit addp<Tab>dr<Tab>` repeatedly will add all three Drill levels one after another.


yairm210:
Looks fine except for normalizeToRuleset - we don't want that, we want to allow scenario editors to do things that are 'unsupported' - placing a mine in an ocean or whatever

yairm210 (approved):

PR #10974, opened 2024-01-20
Espionage automation

tuvus:
This PR adds in a new EspionageAutomation class which sends spies out to steal tech for the AI. The AI prioritizes sending spies to cities with the largest population to steal tech from since they are most likely to have the most science generation.
Of course, this code will need to be changed once more is added on to espionage.


core/src/com/unciv/models/Spy.kt
```diff
@@ -56,6 +56,8 @@ class Spy() : IsPartOfGameInfoSerialization {
     }
 
     fun endTurn() {
+        // If the city
+        if (getLocation() == null) moveTo(null)
```
yairm210, line 60: ???

tuvus: It was kind of a hotfix for when a city was captured that the spy was in. I changed the behaviour to be when the city is captured instead of when the spy updates now. (Now that I read up on it, this is civ V behavior)

PR #10979, opened 2024-01-21
Assume the relevant city for triggered uniques

SeventhM:
Also added in some fixes to `OneTimeAmountFreeUnits` to match the single unit variant of it. Uh... if there's no tile and no city, shouldn't it place the unit at `units.last()` and not `units.first()` so it's less likely to run out of space?

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -122,12 +122,12 @@ object UniqueTriggerActivation {
                     val placedUnit = when {
                         // Set unit at city if there's an explict city or if there's no tile to set at
                         relevantCity != null || (tile == null && civInfo.cities.isNotEmpty()) ->
-                            civInfo.units.addUnit(civUnit, chosenCity) ?: return false
+                            civInfo.units.addUnit(civUnit, chosenCity)
                         // Else set the unit at the given tile
-                        tile != null -> civInfo.units.placeUnitNearTile(tile.position, civUnit) ?: return false
+                        tile != null -> civInfo.units.placeUnitNearTile(tile.position, civUnit)
                         // Else set unit unit near other units if we have no cities
                         civInfo.units.getCivUnits().any() ->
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -41,16 +41,20 @@ object UniqueTriggerActivation {
         notification: String? = null,
         triggerNotificationText: String? = null
     ): Boolean {
+
+        val relevantCity by lazy {
+            city?: tile?.getCity()
```
yairm210, line 46: Since this is definitely in use at line 55, we can put it at 54 and have it not be lazy

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -115,12 +119,21 @@ object UniqueTriggerActivation {
 
                 val tilesUnitsWerePlacedOn: MutableList<Vector2> = mutableListOf()
                 repeat(actualAmount) {
-                    val placedUnit = if (city != null || tile == null) civInfo.units.addUnit(unit, chosenCity)
-                        else civInfo.units.placeUnitNearTile(tile.position, unit)
+                    val placedUnit = when {
```
yairm210, line 122: Hm duplocated code, maybe extract to small function? Not a sticking point just a thought

yairm210 (approved):
Looks good! Suggestions are not blockers ;)

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
SeventhM: Not sure where it being lazy is relevant

Actually, huh. On line 55 it's redundant. `conditionslsApply` has the same relevant city check. We still need to do it twice so where it's placed I doubt matters

~~Also, why are we checking conditions applies here? I know I just PR'es to check this at the source of the trigger rather than in here... Which apparently was redundant for non unit here... But the check here still feels off. Must just be me, Idk~~

PR #10988, opened 2024-01-22
Unit actions "paging" new architecture

### LINKED ISSUE #10906: Sleep action missing when Sleep until healed available

### Game Version

from source

### Describe the bug

Up to (I believe) 4.9.6, a wounded water unit in own territory could choose either 'Sleep until healed' or 'Sleep' - the latter on the 'second page'. Gone. Still OK for units that fortify instead.

### Steps to Reproduce

Sic

### Screenshots

_No response_

### Link to save file

_No response_

### Operating System

Linux

### Additional Information

From https://github.com/yairm210/Unciv/commit/fad06b3aec2d04abb14e9a8ab98b961fc2966180 - the old code used `if (isDamaged || !showingAdditionalActions) actionList += UnitAction(UnitActionType.Sleep,...` which means it's always possible, but if damaged moved to the 'additional actions'.

SomeTroglodyte:
Continuation to #10966, still not reaching #10955 - but doing some things better along the way.

_Postponed_ UX features:
* Dynamic paging to avoid buttons overlapping on cramped displays
* Keyboard shortcuts available for out-of-page actions
* Avoid distributing very few actions to pages
* Vague idea: Ask ImprovementPicker if we happen to select a unit with just one enabled Worker action - and then offer it directly. Legionaries - they need to 'i'-'r' now while they had the road building offered directly ages ago, or am I mistaken? Maybe not the best example as Legion can build Forts too, so it would be _two_ extra UnitActions - but when made intelligently it could help a few mods nicely.

_Included_ features:
* Sleep for wounded units is back (closes #10906)
* Sleep/Fortify use the exact same rules
* Wounded units can switch Sleep/Fortify freely with '* until healed' back and forth (<sub><sup>only marred a little by auto next-unit stealing focus away.</sup></sub>)

_Postponed_ Architecture changes:
* Need to see if the map's getter function can't be switched to the `suspend fun SequenceScope<UnitAction>.addSomethingActions` signature which I believe to be much more efficient than the `sequence { yieldAll(otherSequence) }` construct - at the moment avoided as extension receiver and object-as-namespace are a bit excluding each other...

Toolset note:
* Somehow Android Studio lately has taken to auto-adding imports one cannot configure or turn off in any way, and that the compiler does not actually need - here it force-imports all public methods (or fields too if there were any) of UnitActions into the UnitActions file - complete bullshit in most cases, but I've given up on killing them. In other places it's worse - you can even get warnings from those unnecessary imports that would be fine without. Anybody observed that too, or even found a cure?

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
@@ -27,9 +27,6 @@ import com.unciv.ui.screens.worldscreen.unit.actions.UnitActions.invokeUnitActio
  *  API for unit tests: [getGiftAction]
  */
 object UnitActions {
-    // Todo invokeUnitAction fallback might be more efficient if it could ask getUnitActions to skip the actionTypeToFunctions-mapped ones
-    //      - or enumerate them last, as in that case we can expect that to never be reached.
-    //      But - ordering in the UI? Maybe make defaultPage a float and use fractions for sorting?
 
     /**
      *  Get an instance of [UnitAction] of the [unitActionType] type for [unit] and execute its [action][UnitAction.action], if enabled.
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
```diff
@@ -73,22 +93,47 @@ object UnitActions {
...
+            // Sleep moves to second page if current action is SleepUntilHealed or if unit is wounded and it's not already the current action
+            if (unit.isSleepingUntilHealed() || unit.health < 100 && !(unit.isSleeping() && !unit.isActionUntilHealed())) 1 else 0
+        },
+        UnitActionType.SleepUntilHealed to { unit ->
+            // SleepUntilHealed only moves to the second page if Sleep is the current action
+            if (unit.isSleeping() && !unit.isActionUntilHealed()) 1 else 0
+        },
+        UnitActionType.Explore to { unit ->
+            if (unit.isCivilian()) 1 else 0
+        },
```
yairm210, line 129: This really feels like something that should be defined where the default page is...?
Probably fine as is

yairm210 (approved):
Okay! I feel I don't understand why the pageaction buckets is at it is, but I assume you're prepping for when a third page will be optional

All in all 👍🏿

Issue #7589, opened 2022-08-03
Feature request: AI utilizes Air Sweep


itanasi:
Didn't code up the AI to utilize Air Sweep. Difficult to strategize Air Sweeps to help cover Bombers with current framework

Stub here for someone to build that into the AI.

PR #10991, opened 2024-01-23
Air unit automation improvement

tuvus:
Before this PR a few essential things were missing from the air unit automation. First is that fighters didn't air sweep, and second is that fighters and bombers didn't account for low health, and third is that missiles (like nukes) used bomber automation instead of missile automation.

Changes:
- Added fighter air sweep logic. Targets the closest tile with an enemy unit (kind of arbitrary, mostly, we need to air sweep in the right direction).
- Added fighter air supremacy logic. When to air sweep, heal and attack ground units. (based on the number of our fighters and their planes).
- Changed air unit priority to: fighters -> nukes -> bombers -> missiles. (Was random)
- Bombers and fighters are more aware of their health.
- Units that self-destruct on attack don't take into account how much damage is dealt to them.

As usual more tweaks are probably needed later. I don't promise there won't be any bugs (logical or otherwise). Testing every edge case would take way too much time and effort.

### DIFF
--- a/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
@@ -486,11 +486,6 @@ class MapUnit : IsPartOfGameInfoSerialization {
             damageFactor *= 1f - unique.params[0].toFloat() / 100f
         return damageFactor
     }
-    
-    fun airSweepDamagePercentBonus(): Int {
-        return getMatchingUniques(UniqueType.StrengthWhenAirsweep)
-            .sumOf { it.params[0].toInt() }
-    }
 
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/logic/map/mapunit/MapUnit.kt
```diff
@@ -486,6 +486,11 @@ class MapUnit : IsPartOfGameInfoSerialization {
             damageFactor *= 1f - unique.params[0].toFloat() / 100f
         return damageFactor
     }
+    
+    fun airSweepDamagePercentBonus(): Int {
```
yairm210, line 490: This is only used in one place, it really doesn't need to be in a centralized place like this

yairm210 (approved):

tuvus:
Solves #7589

PR #10995, opened 2024-01-24
Reorganize IConstruction.kt

### LINKED ISSUE #10975: "Requires a [buildingFilter] in at least [amount] cities" not highlighting buildings that can't be built

### Game Version

4.10.0

### Describe the bug

The `Requires a [buildingFilter] in at least [amount] cities` unique is not highlighting the building that can't be built until conditions are met. When you look at the list of available buildings, any building with that condition looks like it should be buildable, without any conditions attached. The game won't let you select it, but it makes it appear as if it is selectable, which is misleading.

### Steps to Reproduce

1. Start DeCiv Redux game as the Atlas nation, in the Rebuilding era or later
2. Found city, then open city to see available buildings
3. Medical Network will appear to be buildable, unlike other national wonders that are clearly marked as not buildable

### Screenshots

Screenshot (162)

### Link to save file

_No response_

### Operating System

Windows

### Additional Information

_No response_

SomeTroglodyte:
A pure refactor without logic changes - except it also fixes #10975, can be considered alternatively to #10994 or independently after it.

Raises many questions I intentionally left unanswered, some as `todo` comments, but there's more:
* The direct `rejectionPrecedence` field is more hassle to use when a new one needs to be _inserted_ into the ordering - unless we adopt the old Basic line numbering convention and e.g. start with 10 in multiples of 10??? Another option would be to change the field into a bool and demand the enum instances ordered by display precedence...
* No mapping of the conditionals applied to OnlyAvailableWhen to RejectionReasons exists at all - none I can see?
* `Building.getRejectionReasons` has both `@Suppress("NON_EXHAUSTIVE_WHEN")` and `else -> {}` - one of them should go
* `Building.getRejectionReasons` checks `conditionalsApply` for `UniqueType.OnlyAvailableWhen` **twice** - with distinct but equivalent syntaxes
* `RejectionReasonType.RequiresBuildingInSomeCity.toInstance("Apollo project not built!")` should be able to be removed by now?

### DIFF
--- a/core/src/com/unciv/models/ruleset/construction/RejectionReason.kt
+++ b/core/src/com/unciv/models/ruleset/construction/RejectionReason.kt
@@ -9,6 +9,7 @@ class RejectionReason(
 
     // Used in Battle.addXp - true means it's _not_ blocking a unit from being awarded for Great General points
     // Used in UnitUpgradeManager.canUpgrade - true means the `ignoreRequirements` parameter will filter these away
+    // true means it's limited by tech, policy, or era in some manner
     fun techPolicyEraWonderRequirements(): Boolean = type.isTechPolicyEraWonderRequirement
 
     // Used in CityConstructions.validateInProgressConstructions - true means remove+refund or replace with upgrade + transfer points
@@ -18,6 +19,7 @@ class RejectionReason(
     fun isImportantRejection(): Boolean = type.rejectionPrecedence >= 0
 
     // Used in Battle.addXp - true means it's _not_ blocking a unit from being awarded for Great General points
+    // true means the reason is specific to the city construction screen. Contains only the ones for units, not buildings 
     fun isConstructionRejection(): Boolean = type.isConstructionRejection
 
     /** Smaller numbers have the highest precedence (unless it's -1 in which case [isImportantRejection] returns `false`) */


SeventhM:
Idk if it's just me, but it seems strange to split `RejectionReasonType` into a different file than `RejectionReason` as the latter basically is totally empty besides checking booleand of the former

core/src/com/unciv/models/ruleset/construction/RejectionReason.kt
```diff
@@ -0,0 +1,26 @@
...
+    // Used in UnitUpgradeManager.canUpgrade - true means the `ignoreRequirements` parameter will filter these away
+    fun techPolicyEraWonderRequirements(): Boolean = type.isTechPolicyEraWonderRequirement
+
+    // Used in CityConstructions.validateInProgressConstructions - true means remove+refund or replace with upgrade + transfer points
+    fun hasAReasonToBeRemovedFromQueue(): Boolean = type.isReasonToDefinitivelyRemoveFromQueue
+
+    // Used in CityConstructionsTable.getConstructionButtonDTOs to filter reasons from being displayed on the button
+    fun isImportantRejection(): Boolean = type.rejectionPrecedence >= 0
+
+    // Used in Battle.addXp - true means it's _not_ blocking a unit from being awarded for Great General points
```
SeventhM, line 20: Comment is somewhat inaccurate. It is also intended to be used in #10947. Comment left in RejectionReasonType is more accurate: these are intended to be reasons specifically relevant to construction and are to be ignored in cases where units are created outside of the build tab. Useful for quickly checking both whether the unit is replaced and whether you have the correct tech in masse without adding a whole extra set of logic. Note that `.map{ civ.getEquivalentUnit(it) }` is not sufficient, you actually have to do something like `.filter{ civ.getEquivalentUnit(it) == it }` to cover this same use case here which is more expensive than the check that `getRejectionReasons` uses, likely less expensive than checking the entirety of rejection reasons, but the entirety of rejection reasons also gives a free max unit built check, so it's kinda a wash

SeventhM:
> * There already are <10 line files in this project

Not sure how this is an argument against my point. My point is there's only, like, 10 meaningful lines in `RejectionReason`, almost all of it is just `fun x: Boolean = type.x`. Which kinda leads to my point of

> * 1 file 1 class/iface/whatever is standard, some have tons if tiny classes

Most of these are (more or less) self contained. Yeah, `Stat` is a relatively small file, but it doesn't really rely on too many other classes to make sense. Sure, there's a bunch of one off classes in ui, but they all handle their own code, they don't exclusively handle code of other classes. `RejectionReason` by itself doesn't make too much sense. That's why it seems at bit odd on its own and not in the same file as `RejectionReasonType`

PR #11000, opened 2024-01-25
Fix build errors

SeventhM:
Side note: In-Game, these don't actually show as errors, just warnings, so that may be a bug in its own right
Also not sure why especially the city state one is an error when you just now deprecated it. If we simply set the deprecation level back to the default (Warning), the only part of this PR that's necessary is fixing Mongolia and fixing the deprecation message

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Nations.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Nations.json
@@ -653,7 +653,7 @@
 		"innerColor": [255,120,0],
 		"favoredReligion": "Tengriism",
 		"uniqueName": "Mongol Terror",
-		"uniques": ["+30% Strength when fighting City-State units and cities", "[+1] Movement <for [Mounted] units>"],
+		"uniques": ["[+30%]% Strength <vs [City-States]>", "[+1] Movement <for [Mounted] units>"],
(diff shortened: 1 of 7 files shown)


SeventhM:
> state.theirCombatant?.matchesCategory(condition.params[0]) == true

You're out of date https://github.com/yairm210/Unciv/commit/414a11a2d6025a398839d3e2e4edb7853feebbcd

> = category == "City" || category == "All"

https://github.com/yairm210/Unciv/commit/21581b43a0eaa3557fa026465f5d7bcb259dc20d

yairm210:
Wow I did a really botched job

yairm210:
Yeah the city state was also supposed to be a warning,  I leave it on warning for a minor version which is 20 versions which is like 2.5 months for good measure

SeventhM:
> we're still breaking

I mean, there's a probably a pretty good list we're breaking by putting ImprovementTime to Error tbh, let alone the conditional vs city-states Actually, annoyingly because this was done just after a release and doesn't function in a release yet, I can't update Rekmod's Mongolia to account for it until after its already broken

yairm210:
I'll re-add it 
I'm usually the one cleaning up after others, not use to others cleaning my messes 😓

PR #11011, opened 2024-01-26
Unify unit and civ triggers

SeventhM:
This PR aims to simplify unique triggers by calling a central function instead of a different function based on if it's a unit or civ. This aims to do a few things

1. Make it easier to include triggers from units with triggers from other sources. Main benefits are seen in the simplification seen in `TileInfoImprovementFunctions`
2. Make it simpler when trying to edit a unique to include/not include a unit
3. Make the formatting closer to StateForConditionals (possibly we can unify the uses here

Extras added in
1. Overrides so that `triggerUnique(city.civ, city)` can simplify down to `triggerUnique(city)` and the same for units
2. StateForConditionals does not support constructors apparently. So conditionalsApplies was edited to assume the civ from the unit/city to allow for the same simplification
3. `upon founding a city` now triggers the uniques found on the unit itself

Main notes here:
1. Out of scope, but is there a better way to try to trigger uniques off of units? Particularly because we have a blindspot in that gaining techs or policies or whatever doesn't trigger uniques off of units
2. `triggerUnitwideUniques` was privated, but tbh everything in it should be moved to the main `triggerUniques` function. It can do so pretty safely now anyways
3. That `conditionalsApplies` check in `triggerUniques` still looks extremely out of place to me

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -31,11 +31,32 @@ import kotlin.random.Random
 
 // Buildings, techs, policies, ancient ruins and promotions can have 'triggered' effects
 object UniqueTriggerActivation {
+
+    fun triggerUnique(
+        unique: Unique,
+        city: City,
+        notification: String? = null,
+        triggerNotificationText: String? = null
+    ): Boolean {
+        return triggerUnique(unique, city.civ, city,
+            notification = notification, triggerNotificationText = triggerNotificationText)
+    }
+    fun triggerUnique(
+        unique: Unique,
+        unit: MapUnit,
+        notification: String? = null,
+        triggerNotificationText: String? = null
+    ): Boolean {
+        return triggerUnique(unique, unit.civ, unit =  unit, tile = unit.currentTile,
+            notification = notification, triggerNotificationText = triggerNotificationText)
+    }
+
     /** @return whether an action was successfully performed */
-    fun triggerCivwideUnique(
+    fun triggerUnique(
         unique: Unique,
         civInfo: Civilization,
         city: City? = null,
+        unit: MapUnit? = null,
         tile: Tile? = city?.getCenterTile(),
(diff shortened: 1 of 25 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -31,11 +31,32 @@ import kotlin.random.Random
...
+    }
+
     /** @return whether an action was successfully performed */
-    fun triggerCivwideUnique(
+    fun triggerUnique(
         unique: Unique,
         civInfo: Civilization,
         city: City? = null,
+        unit: MapUnit? = null,
         tile: Tile? = city?.getCenterTile(),
```
yairm210, line 60:
```suggestion
        tile: Tile? = city?.getCenterTile() ?: unit?.currentTile,
```

yairm210 (approved):
Looks good!

yairm210:
We need the conditionalApplies, because when on new tech / policy / building we try to trigger each of their uniques.
So it makes sense to me that checking the conditionals should be done in one place, and not for each of them individually. 
I'm open to other alternatives, this was just the practical choice

SeventhM:
> I'm open to other alternatives, this was just the practical choice

Most scenarios, intentionally or otherwise, we've already done the StateForConditionals check. In fact I recently PR'ed to do the check every time before triggering, so now we've gone from "most of the time" to "we always". The only way for doing the check in `triggerUniques` to make sense is in case a relevant situation changes (one trigger only happens before researching a tech, the other researches the tech for free), which I could go either way on supporting

SeventhM:
> What do you mean? Why not?

Huh. C++ brain thought constructors used the syntax of `ClassName (){}`

SeventhM:
> convert back to public is fine too

Considering I'm probably actually just going to merge everything from `TriggerUnitWideUniques` into `TriggerUniques`, I'm hard against this as even a suggestion. Actually, I already merged the two functions in my dev environment with practically zero repercussions, so I'll just commit that in

PR #11012, opened 2024-01-26
Ranged units capture civilian

tuvus:
This PR addresses the problem of ranged units, such as cities and archers, shooting captured workers and settlers instead of capturing them.

- Cities no longer shoot capturable civilian units
- Ranged units only shoot uncapturable civilian units when necessary. Otherwise, they try to capture the civilian unit by moving to them instead.


yairm210 (approved):

tuvus:
Can civilians ever be military units? If so, this code might have a problem.

SeventhM:
> Can civilians ever be military units? If so, this code might have a problem.

No. It's hardcoded that civilian= non military and military= unit with any strength

PR #11013, opened 2024-01-26
Minor Mod manager fix, lints and dox

SomeTroglodyte:
Excerpts from #10900 worth preserving - see commit messages

### DIFF
--- a/core/src/com/unciv/models/ModConstants.kt
+++ b/core/src/com/unciv/models/ModConstants.kt
@@ -1,5 +1,16 @@
 package com.unciv.models
 
+/** Used as a member of [ModOptions][com.unciv.models.ruleset.ModOptions] for moddable "constants" - factors in formulae and such.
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/models/ModConstants.kt
```diff
@@ -71,32 +83,37 @@ class ModConstants {
...
+        return equalsReflected(other)
+    }
+
+    override fun hashCode(): Int {
+        var result = 0
+        for (field in this::class.java.declaredFields) {
+            result = result * 31 + field.get(this).hashCode()
+        }
+        return result
+    }
```
yairm210, line 105: I see this is the same as the Vector2 hash implementation, I wonder if there's a builtin "multihash" function that accepts an iterable of hashes and returns one...

yairm210 (approved):

PR #11014, opened 2024-01-26
Avoid modifying the ruleset object and unique stats

yairm210:
Solves "infinite nat wonder stats" reported on Discord


yairm210:
release patch

PR #11030, opened 2024-01-28
ThreatManager improvement

tuvus:
This PR is a rework of `ThreatManager` that improves upon #10481.
There are two functions that `ThreatManager` currently offers, `getDistanceToClosestEnemy()` and `getEnemyTilesInDistance()`. Previously it was optimized mostly to `getDistanceToClosestEnem()`, this PR, however, reduces the optimization to `ClosestEnemy` in favor of caching more information that is used to optimize `EnemyTiles`.

This PR is more of a checkpoint for future plans to expand the ThreatManager system. 

My AutoPlay Testing has not found any bugs yet (surprising! Oh wait, I forgot to run the unit tests... thank you, unit testing, I have to mess up at least one thing).

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
@@ -89,18 +89,14 @@ class ThreatManager(val civInfo: Civilization) {
 
         if (tileData != null && tileData.distanceSearched >= maxDist) {
             // Add all tiles that we have previously found
-            var index = 0
-            while (tileDataTilesWithEnemies.size > index) {
-                val tileWithDistance = tileDataTilesWithEnemies[index]
-                if (tileWithDistance.second > maxDist)
-                    return tilesWithEnemies
-                if (doesTileHaveMilitaryEnemy(tileWithDistance.first)) {
+            val tilesWithEnemiesIterator = tileDataTilesWithEnemies.listIterator()
+            for (tileWithDistance in tilesWithEnemiesIterator) {
+                // Check if the next tile is out of our search range, if so lets stop here
+                if (tileWithDistance.second > maxDist) return tilesWithEnemies
+                // Check if the threat on the tile is still present
+                if (doesTileHaveMilitaryEnemy(tileWithDistance.first))
                     tilesWithEnemies.add(tileWithDistance.first)
-                } else {
-                    tileDataTilesWithEnemies.removeAt(index)
-                    continue
-                }
-                index++
+                else tilesWithEnemiesIterator.remove()
             }
         }
 
--- a/core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
@@ -83,10 +83,7 @@ class ThreatManager(val civInfo: Civilization) {
     fun getTilesWithEnemyUnitsInDistance(tile: Tile, maxDist: Int): MutableList<Tile> {
         val tileData = distanceToClosestEnemyTiles[tile]
 
-        // Shortcut, we don't need to search for anything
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
```diff
@@ -34,69 +34,97 @@ class ThreatManager(val civInfo: Civilization) {
...
+                    return tilesWithEnemies
+                if (doesTileHaveMilitaryEnemy(tileWithDistance.first)) {
+                    tilesWithEnemies.add(tileWithDistance.first)
+                } else {
+                    tileDataTilesWithEnemies.removeAt(index)
+                    continue
+                }
+                index++
+            }
+        }
```
yairm210, line 105: This feels needlessly confusing and I'm having a hard time trying to formulate what is wrong.

This part should be *converting the tileData to a list of tiles within range that have enemies*, right? 
So
A. In what case would the cached tileDataTilesWithEnemies *not* have an enemy? If we got rid of it with a previous unit? That should be explicit, this is very roundabout!
B. What we actually care about *here* is the tilesWithEnemies!

So I would formulate this as:

```kotlin
if (...)
 return tileDataTilesWithEnemies.filter {it.second <= maxDist}
.filter { doesTileHaveMilitaryEnemy(tileWithDistance.first }
```

If you want to update the tileDataTilesWithEnemies to *remove* those tiles that we determined no longer have units, that's fine, you can do that as a separate step, but not at expense of readability.

core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
```diff
@@ -34,69 +34,97 @@ class ThreatManager(val civInfo: Civilization) {
...
+
+        // Shortcut, we don't need to search for anything more
         if (tileData != null && maxDist <= tileData.distanceSearched)
```
yairm210, line 108: Doesn't this mean that if tileData != null you will always return...?

core/src/com/unciv/logic/civilization/managers/ThreatManager.kt
tuvus: Do you find the version with the mutableListIterator to be more readable? Functional programming doesn't seem to be efficient with having to deal with two things at once (which goes against the whole point of functional programming). We could loop through it twice but then we also won't be using the sorted property of the list.

tuvus: For context, I plan on running this section of code indirectly a lot more in the future. There are also some other performance improvements that I am investigating will make the code bigger and less readable. If someone wants to make a change in this area of the code, they should probably have a much better understanding of how it works compared to code in other parts of Unciv. If we need to quickly change a feature we probably shouldn't be looking to modify this code, especially if it will be used in many places.

Ideally, each military unit would call a variant of `getTilesWithEnemiesInDistance(4)` or `gTWED(4)` (60 tiles to search) to look around the surrounding area to see if they should continue attacking, hold or retreat. Preferably also on the tiles melee units want to attack move to as well to make sure they are safe enough.
If we want to check if a city is being surrounded by neutral units, we might also want to call `gTWED(6)` (126 tiles to search) on cities to find those units.
Air units are calling `gTWED(5, 8, 20)` (90, 216, 1260 tiles to search) every turn to attack.
Air units attack before all ground units, which is great because it would overshadow the city check that they are in and we could use the information gathered from the larger checks to speed up the smaller checks around it.

An example of where ending the search early is very helpful: a stealth bomber calls `gTWED(20)` and finds some enemies, then we call `gTWED(6)` of smaller for any reason. The original range of 20 call has 1134 tiles that the smaller call does not overlap with. If we don't end the search early, each smaller call would have to search through all previously found enemies on the additional 1134 tiles (note that each tile must either have a city or be visible to count).

yairm210 (approved):

PR #11034, opened 2024-01-28
Restrict happiness change updates only to relevant thresholds

yairm210:
As discussed - we currently update on every happiness change, we can limit this to only update all cities if we pass a happiness threshold that is relevant to *some* unique

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -70,8 +70,24 @@ class Ruleset {
     var victories = LinkedHashMap<String, Victory>()
     var cityStateTypes = LinkedHashMap<String, CityStateType>()
 
-    val greatGeneralUnits by lazy { 
+    val greatGeneralUnits by lazy {
         units.values.filter { it.hasUnique(UniqueType.GreatPersonFromCombat, StateForConditionals.IgnoreConditionals) }
+    }
+
+    /** Contains all happiness levels that moving *from* them, to one *below* them, can change uniques that apply */
+    val allHappinessLevelsThatAffectUniques by lazy {
+        sequence {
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -70,10 +70,27 @@ class Ruleset {
...
 
-    val greatGeneralUnits by lazy { 
+    val greatGeneralUnits by lazy {
         units.values.filter { it.hasUnique(UniqueType.GreatPersonFromCombat, StateForConditionals.IgnoreConditionals) }
     }
 
+    /** Contains all happiness levels that moving *from* them, to one *below* them, can change uniques that apply */
+    val allHappinessLevelsThatAffectUniques by lazy {
+        sequence {
+            for (rulesetObject in (units.values + terrains.values + buildings.values + technologies.values + eras.values
```
SeventhM, line 80: Something tells me we should have a helper function/lazy for this in case it comes up again that we're looking through all objects. Half surprised we don't already have one

tuvus:
Hey, at least there were no semicolons lying around!

SeventhM:
Out of scope, I guess, but on second thought I'm surprised by the lack of generalization. There shouldn't be anything stopping us from having conditionals for per turn yields of other stats. In which case, perhaps this should be changed at some point to check for updating on any change. Just a thought

Besides the obvious need for a helper function/lazy/`allRulesetObjects` and Trog's point about `until` vs `..` (over my head for someone who's sent zero effort looking into the difference, looks fine enough to me

SeventhM:
I don't remember if 0 happiness is considered growth stunted in civ 5, but I'm pretty sure it shouldn't be. If the rules change results in that, then maybe we need to fix the unique in the rulesets and tell all the other rulesets to fix... Nah, not quite a fan imo

PR #11056, opened 2024-01-31
Players can't move spies when it is not their turn

tuvus:
Minor PR fix, does as stated and was tested.


core/src/com/unciv/ui/screens/overviewscreen/EspionageOverviewScreen.kt
```diff
@@ -105,7 +105,7 @@ class EspionageOverviewScreen(val civInfo: Civilization, val worldScreen: WorldS
                         )
                 }
             }
-            if (worldScreen.viewingCiv != civInfo) {
+            if (worldScreen.viewingCiv != civInfo || !worldScreen.viewingCiv.isCurrentPlayer()) {
```
yairm210, line 108:
```suggestion
            if (!worldScreen.canChangeState) {
```

PR #11057, opened 2024-02-01
Added unit escorting formation!!!

tuvus:
This PR adds the escort formation feature of Civ 5 (I think, I haven't tested it on Civ V). The basic concept is that you can link two units together, one civilian one military and they will move together and stuff. I put some placeholder icons because I needed something and didn't want to include them yet. Civilopedia entry will need to be added later.

Escorting is stored as a boolean on each mapunit. A unit is considered to be escorting if their escorting boolean is true and they have their unit complement to escort with. Starting and stopping escorting clears the movement cache because escorting units have their movement depend on the other unit's movement capabilities.

A lot of the Movement.kt methods now have an extra `isEscorting = true` boolean for the first call. In the case that `isEscorting` is true and the unit is actually escorting these methods will call themselves on the other escort unit with `isEscorting = false` to confirm that the other unit can make the same action.

<details><summary>Pictures showing UI</summary>
Unlinked

Escort1

Horseman linked, showing 2 movement points compared to its usual 4.

Escort2

Worker linked.

Escort3

</details> 

This PR includes tests as well, yay! On a base level, the PR isn't very dangerous, if there are any serious errors, the player can just not use it. I did not include it in any AI logic yet for this reason.

Also, I did have a merge conflict, as usual, I pressed a few buttons to make the red stuff go away. Hopefully, it didn't screw anything up.


core/src/com/unciv/logic/map/mapunit/MapUnit.kt
```diff
@@ -231,13 +234,14 @@ class MapUnit : IsPartOfGameInfoSerialization {
...
 
-    fun isIdle(): Boolean {
+    fun isIdle(escorting: Boolean = true): Boolean {
         if (currentMovement == 0f) return false
         val tile = getTile()
         if (tile.improvementInProgress != null &&
             canBuildImprovement(tile.getTileImprovementInProgress()!!) &&
             !tile.isMarkedForCreatesOneImprovement()
         ) return false
+        if (isEscorting() && escorting && !getOtherEscortUnit()!!.isIdle(false)) return false
```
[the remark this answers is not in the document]
tuvus: > isEscorting() && escorting is double-checking as I read the functions below

This is not true. The local escorting variable determines if it should call itself again on the other escorting unit if it is escorting. I should definitely rename it.
`getOtherEscortUnit()` uses `isEscorting()` inside of it. We can call `getOtherEscortUnit()` even if the unit is not escorting yet. See UnitActions.kt line `180`.
I should definitely write more java docs, I was just wondering how much comments I should write.

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
```diff
@@ -158,11 +158,41 @@ object UnitActions {
...
+        val worldScreen = GUI.getWorldScreen()
+        val selectedUnits = worldScreen.bottomUnitTable.selectedUnits
+        if (selectedUnits.size == 2) {
+            val tile = selectedUnits.first().getTile()
+            if (selectedUnits.last().getTile() != tile) return
+            if (selectedUnits.any { it.baseUnit.movesLikeAirUnits() }) return
+        } else if (selectedUnits.size == 1) {
+        } else {
+            return
+        }
```
yairm210, line 179:
```suggestion
        } else if (selectedUnits.size != 1) return
```

yairm210: Although I'm not sure what exactly you're getting at here, what is escort with 1 unit?

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
```diff
@@ -158,11 +158,41 @@ object UnitActions {
...
 
+    private suspend fun SequenceScope<UnitAction>.addEscortAction(unit: MapUnit) {
+        // Air units cannot escort
+        if (unit.baseUnit.movesLikeAirUnits()) return
+
+        val worldScreen = GUI.getWorldScreen()
+        val selectedUnits = worldScreen.bottomUnitTable.selectedUnits
+        if (selectedUnits.size == 2) {
+            val tile = selectedUnits.first().getTile()
+            if (selectedUnits.last().getTile() != tile) return
```
yairm210, line 174: This really confuses me.

So there can be 3 different units? The one passed in and 2 selected?
Or is the unit always one of the selected?

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
tuvus: We usually only have one unit selected. We need to get the other unit on the tile to get the two units. If we select two units with multi-select, this code allows us to still put them into a formation.

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActions.kt
tuvus: One of the two selected units is the original unit passed in. When we have two units we need to confirm that they are both on the same tile so that we aren't telling the original unit to start a formation with some other random unit. I will add a comment.

tuvus:
What is the difference between `distanceToTiles` and `getDistanceToTilesWithinTurn`? It seems like `getDistanceToTilesWithinTurn` should be made private. Otherwise, we would need to add the escort rule to it as well.

core/src/com/unciv/logic/map/mapunit/MapUnit.kt
```diff
@@ -854,6 +879,24 @@ class MapUnit : IsPartOfGameInfoSerialization {
...
+        } else {
+            escorting = false
+        }
+        movement.clearPathfindingCache()
+    }
+
+    fun stopEscorting() {
+        if (getOtherEscortUnit() != null) {
+            getOtherEscortUnit()!!.escorting = false
+        }
```
yairm210, line 895: Can this be one-lined?
`getOtherEscortUnit()?.escorting = false`
or does that not work?

core/src/com/unciv/logic/map/mapunit/movement/UnitMovement.kt
```diff
@@ -238,8 +238,14 @@ class UnitMovement(val unit: MapUnit) {
      * @return The tile that we reached this turn
      */
     fun headTowards(destination: Tile): Tile {
+        val escortUnit = if (unit.getOtherEscortUnit() != null && unit.escorting)
+            unit.getOtherEscortUnit() else null
```
yairm210, line 242:
```suggestion
        val escortUnit = if (unit.escorting) unit.getOtherEscortUnit() else null
```
since if getOtherEscortUnit is null, escortUnit will be null anyway, no need to have a separate check

core/src/com/unciv/logic/map/mapunit/movement/UnitMovement.kt
```diff
@@ -807,4 +854,13 @@ class PathsToTilesWithinTurn : LinkedHashMap<Tile, UnitMovement.ParentTileAndTot
...
         return reversePathList.reversed()
     }
+    
+    fun removeAllTilesNotInSet(otherPathsToTiles: PathsToTilesWithinTurn) {
+        val iterator = this.iterator()
+        while(iterator.hasNext()) {
+            if (!otherPathsToTiles.contains(iterator.next().key)) {
+                iterator.remove()
+            }
+        }
```
yairm210, line 864: how about .removeAll {  }, we don't need a special function for this

tuvus: I used .removeIf {} and removed that method entirely since it had only one use.

yairm210 (approved):

PR #11061, opened 2024-02-01
Yet Another Health Bar

SomeTroglodyte:
Since all the work gone into #9205 and #9291 is a bit of a pity to just throw away, I've written another (two) HealthBar Widget from scratch. This PR is meant to check architecture welcome-ness, and meant to be run as FasterUIDevelopment like this:
<details><summary>Run configuration</summary>

```xml
<component name="ProjectRunConfigurationManager">
  <configuration default="false" name="FasterUIDevelopment" type="Application" factoryName="Application">
    <option name="MAIN_CLASS_NAME" value="com.unciv.dev.FasterUIDevelopment" />
    <module name="Unciv.tests.test" />
    <option name="WORKING_DIRECTORY" value="$ProjectFileDir$/android/assets" />
    <method v="2">
      <option name="Make" enabled="true" />
    </method>
  </configuration>
</component>
```
</details>
... as it contains baggage that would need removal for production use.

I've intentionally lef myself forget the last one and redone this before pulling up the old code to get it into the tester - astounding how much ended up similar. But also - some API signatures of the old one were bad ideas, the new surface is better. Nevertheless, from heap efficiency, performance, and maintainability, I'd still prefer the classic Widget(layout/draw) of V0.

<details><summary>Screen-cap video</summary>

https://github.com/yairm210/Unciv/assets/63000004/e5e26a43-c929-40f8-9597-0cd267c64319

Just shows all of them do their job. Turn off that setRound(false) line and the Table-based one glitches out:

https://github.com/yairm210/Unciv/assets/63000004/29dd0f26-66fa-40b7-a5db-3d96635af299

</details>

Compares 4 variants :
* Current `ImageGetter.getHealthBar` - this is static and has no support for changing the 'health' value, thus the tester will not animate it, but discard and replace when necessary.
* This new implementation based on Table. Would fail miserably with ugly visual glitches - pixels drawing one-off, but for that `setRound(false)` call. Incomplete, to be able to replace all instances the old one was able to replace it would still need a few additions, one of which a layout nightmare that was straightforward in...
* ... the old implementation that worked same as Gdx's own widgets by drawing itself directly, copied unchanged from the old pr - and no hassle to integrate into the tester.
* A simple as-close-as-possible to Gdx standards implementation based on ProgressBar (basis of Slider). Impossible to get some of the enhanced features of the others into that, it's limited to that two-segment display, just for comparison.

### DIFF
--- a/core/src/com/unciv/ui/components/widgets/HealthBar.kt
+++ b/core/src/com/unciv/ui/components/widgets/HealthBar.kt
@@ -49,6 +49,7 @@ class HealthBar(
 
         background = BaseScreen.skinStrings.getUiBackground(style.backgroundPath, tintColor = style.backgroundTint)
         pad(style.borderSize)
+        setRound(false)
 
         for (i in segments.indices) {
             val segment = segments[i]
@@ -56,15 +57,8 @@ class HealthBar(
                 throw IllegalArgumentException("HealthBar segment with dynamicColor must leave color at WHITE")
 
             val backgroundPath = "General/HealthBar" + (segment.backgroundPath ?.let { "/$it" } ?: "Segment")
-            if (BaseScreen.skinStrings.hasUiBackground(backgroundPath)) {
-                val background = BaseScreen.skinStrings.getUiBackground(backgroundPath, tintColor = segment.color)
-                add(BackgroundActor(background, Align.center))
-            } else {
-                // This is a kludge to reduce visual impact of Table quirks - shouldn't be necessary, but using BackgroundActor will paint cells into the padding
-                val image = ImageGetter.getWhiteDot()
-                image.color = segment.color
-                add(image)
-            }
+            val background = BaseScreen.skinStrings.getUiBackground(backgroundPath, tintColor = segment.color)
+            add(BackgroundActor(background, Align.center))
 
             if (vertical) row()
         }


yairm210:
You've convinced me that both the new table and V0 versions are better than current.
I'm not entirely convinced that for maintainability the table is worse than V0, but since you've stuck around this long it looks like you'll probably be around if/when we encounter bugs, so I'm fine with that.

So clean up the unnecessary code and we'll review this :)

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+        /** @see pad */
+        var padHorizontal: Float = 0f
+
+        /** At layout time, there must be enough colors. One per segment, except the last one can be omitted and will mean Color.CLEAR */
+        var colors: Array<Color> = arrayOf(Color.GREEN)
+
+        var minBarWidth: Float = 0f
+        var minBarHeight: Float = 0f
+
+        var animateDuration: Float = 0f
```
yairm210, line 69: durationSeconds please - I hate having to guess seconds or milliseconds 
Same for the rest

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+    operator fun get(index: Int) = values[index]
+    operator fun set(index: Int, value: Float) {
+        values[index] = value.coerceIn(minValue, maxValue)
+        invalidate()
+    }
+
+    /** Shortcut to first value */
+    var value: Float
+        get() = values[0]
+        set(value) { values[0] = value }
```
yairm210, line 125: What are these about?

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+
+    /** Set [style] value [flashingSegment][HealthBarStyle.flashingSegment] **before** turning this on */
+    var allowAnimations = false
+        set(value) {
+            updateAnimations(value)
+            field = value
+        }
+
+    /** Special case for the UnitGroup - one segment colored by value and 1f black border */
+    fun setupForUnitOrCity(current: Int) {
```
yairm210, line 135: Current health?

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+        style.background = BaseScreen.skinStrings.getUiBackground("General/HealthBar", tintColor = Color.BLACK)
+        style.padVertical = 1f
+        style.padHorizontal = 1.2f
+        values[0] = current.toFloat()
+        style.colors[0] = when {
+            current <= 33 -> Color.RED
+            current <= 66 -> Color.ORANGE
+            else -> Color.GREEN
```
yairm210, line 143: For cities that have >100 health, this is just wrong - we should be looking at *percentage of current health out of maximum*

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+    var style: HealthBarStyle = HealthBarStyle(isVertical)
+) : Widget() {
+    /**
+     *  A reusable Widget to draw progress bars, health bars and the like.
+     *  Convenience constructor using Int limits, and default constructor.
+     *
+     *  Note that a default instance without any further configuration will draw nothing and have zero pref size.
+     */
+    constructor(
+        minValue: Int = 0,
```
yairm210, line 45: Let's not even allow minValue other than zero - that's nonsense for a healthbar.

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    @JvmName("setValuesVarArg")
+    fun setValues(vararg newValues: Float) = setValues(newValues)
+
+    fun setValues(vararg newValues: Int) {
+        val floatValues = FloatArray(newValues.size)
+        for (i in newValues.indices) floatValues[i] = newValues[i].toFloat()
+        setValues(floatValues)
+    }
```
yairm210, line 168: Part of clarifying values is clarifying this as well. Maybe values should be a separate class with all the functions it has.

core/src/com/unciv/ui/components/widgets/HealthBar.kt
```diff
@@ -0,0 +1,288 @@
...
+        val across = (
+            when {
+                vertical && width > 0f -> width
+                vertical && prefWidth > 0f -> prefWidth
+                vertical -> 10f
+                height > 0f -> height
+                prefHeight > 0f -> prefHeight
+                else -> 10f
+            } - style.borderSize * 2
+        ).coerceAtLeast(0f)
```
yairm210, line 110: ????

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+
+    override fun layout() {
+        values.sort()
+        if (style.fontSize == 0) return
+
+        val labelText = (values.asSequence() + maxValue)
+            .joinToString("/") { "%.0f".format(it) }
+        val label = this.label ?:
+        Label(null, Label.LabelStyle(Fonts.font, Color.WHITE))
+            .apply { this@HealthBarV0.label = this }
```
yairm210, line 182: If you're using 2 differnt "this" in an apply, that probably shouldn't be an apply...
What's wrong with label = Label()????????

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
```diff
@@ -0,0 +1,317 @@
...
+            newValues.copyInto(this, 0, 0, size.coerceAtMost(newValues.size))
+        }
+
+        override fun act(delta: Float): Boolean {
+            val complete = super.act(delta)
+            val value = this.value
+            for (i in values.indices) {
+                values[i] = (1f - value) * startValues[i] + value * endValues[i]
+            }
+            return complete
```
yairm210, line 252: More of The Problem

core/src/com/unciv/ui/components/widgets/HealthBarV0.kt
yairm210: That sounds like an entirely different set of problems - that bar extends *from the center* towards the other side
These are two very different problem spaces

PR #11066, opened 2024-02-02
Avoid crash from city combantants

SeventhM:
Genuinely doubt anyone will see the relevant crash given I only started editing the StatForConditionals on my personal build

~~Also, kinda annoying I can't do `city = ourCombatant as? CityCombatant?.city` simply because `city` is not actually an accessible value~~ edit: ignore this, i'm silly


SeventhM:
> Operator precedence

Fing goddammit. I more or less tried a ton of times and didn't even realize I needed those parentheses

yairm210 (approved):

PR #11077, opened 2024-02-03
Further Espionage Implementation

tuvus:
This PR adds most of what was missing in the base structure of espionage excluding buildings, policies and religions:

- Spies now have ranks, they start at 1 and as of now the limit is 3. The limit should probably be added in a global unique to be customizable.
- Spies can now be killed when failing to steal tech, or when rigging elections with an allied spy also rigging elections. The chance of death depends on the rank of each spy. If the spy is killed, the defending spy will level up and the action will fail.
- When a spy dies it revives after 5 turns as a new rank 1 spy with a new name. (This is how civ V does it)
- Spies now steal tech 33% faster for each extra rank.
- Spies can now rig elections, in my implementation, they get around 10 influence every 10 turns. Each bonus spy rank increases the influence gained by 3. This is not how Civ V does it, they use city-state elections and rigging elections can decrease the city-states influence with other civs. However I didn't want to spend much more time searching for little benefit. So I made my own implementation.
- The AI has some spy behavior, it is mostly random though and will need to be improved in a different commit.
- The location of a spy is no longer stored as a string and instead is stored as a vector 2. This will allow for much easier debugging in the future and makes the value a little more concrete. **As a side effect, all current espionage games will become incompatible with all future espionage games.** My expectation is that players shouldn't have been playing with espionage since it was hidden behind the debug screen.

Both of these are AI viewed as a spectator.

EspionageScreen1
EspionageScreen2

</details> 

And yes, I will add translations soon. (Done)

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1731,18 +1731,22 @@ After an unknown civilization entered the [eraName], we have recruited [spyName]
 We have recruited [spyName] as a spy! = 
 Your spy [spyName] has leveled up! = 
 Your spy [spyName] cannot steal any more techs from [civName] as we've already researched all the technology they know! = 
+
 # Stealing Technology defending civ
 An unidentified spy stole the Technology [techName] from [cityName]! = 
 A spy from [civName] stole the Technology [techName] from [cityName]! = 
 A spy from [civName] tried to steal our Technology was found and killed in [cityName]! = 
 A spy from [civName] tried to steal our Technology was found and killed in [cityName] by [spyName]! = 
+
 # Stealing Technology offending civ
 Your spy [spyName] stole the Technology [techName] from [cityName]! = 
 Your spy [spyName] was killed trying to steal Technology in [cityName]! = 
+
 # Rigging elections
 A spy from [civName] tried to rig elections and was found and killed in [cityName] by [spyName]! = 
 Your spy [spyName] was killed trying to rig the election in [cityName]! = 
 Your spy successfully rigged the election in [cityName]! = 
+
 # Spy fleeing city
 After the city of [cityName] was destroyed, your spy [spyName] has fled back to our hideout. = 
 After the city of [cityName] was conquered, your spy [spyName] has fled back to our hideout. = 
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1729,14 +1729,24 @@ Move =
 
 After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
 We have recruited [spyName] as a spy! = 
-A spy from [civilization] stole the Technology [techName] from [cityName]! = 
+Your spy [spyName] has leveled up! = 
+Your spy [spyName] cannot steal any more techs from [civName] as we've already researched all the technology they know! = 
+# Stealing Technology defending civ
 An unidentified spy stole the Technology [techName] from [cityName]! = 
-Your spy [name] stole the Technology [techName] from [cityName]! = 
-Your spy [name] cannot steal any more techs from [civilization] as we've already researched all the technology they know! = 
-
+A spy from [civName] stole the Technology [techName] from [cityName]! = 
+A spy from [civName] tried to steal our Technology was found and killed in [cityName]! = 
+A spy from [civName] tried to steal our Technology was found and killed in [cityName] by [spyName]! = 
+# Stealing Technology offending civ
+Your spy [spyName] stole the Technology [techName] from [cityName]! = 
+Your spy [spyName] was killed trying to steal Technology in [cityName]! = 
+# Rigging elections
+A spy from [civName] tried to rig elections and was found and killed in [cityName] by [spyName]! = 
+Your spy [spyName] was killed trying to rig the election in [cityName]! = 
+Your spy successfully rigged the election in [cityName]! = 
+# Spy fleeing city
 After the city of [cityName] was destroyed, your spy [spyName] has fled back to our hideout. = 
 After the city of [cityName] was conquered, your spy [spyName] has fled back to our hideout. = 
-Due to the chaos ensuing in [cityName], your spy [spyname] has fled back to our hideout. = 
+Due to the chaos ensuing in [cityName], your spy [spyName] has fled back to our hideout. = 
 
 # Promotions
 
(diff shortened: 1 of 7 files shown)


tuvus:
If I separated them into different PRs and still wanted to work on them at the same time then I would have to have merge conflicts. AFAIK merging the PRs together wouldn't work as well since we are squashing each PR into one commit.

core/src/com/unciv/logic/automation/unit/EspionageAutomation.kt
```diff
@@ -1,33 +1,83 @@
...
+            // There is nothing for our spy to do, put it in a random city
+            spy.moveTo(civInfo.gameInfo.getCities().filter { it.getCenterTile().isVisible(civInfo) && spy.canMoveTo(it) }.toList().randomOrNull())
         }
     }
+
+    /**
+     * Moves the spy to a city that we can steal a tech from
+     */
+    fun automateSpyStealTech(spy: Spy): Boolean {
+        if (civsToStealFrom.isNotEmpty()) {
```
yairm210, line 58: Fail faster - switch the if to "if X.empty() return false" for zero nesting

core/src/com/unciv/logic/automation/unit/EspionageAutomation.kt
```diff
@@ -1,33 +1,83 @@
...
+                if (automateSpyStealTech(spy)) continue
+                if (automateSpyRigElection(spy)) continue
+                if(automateSpyCounterInteligence(spy)) continue
             }
-            if (spy.action == SpyAction.None) {
-                spy.moveTo(civInfo.getKnownCivs().filter { otherCiv -> otherCiv.isMajorCiv() && otherCiv.cities.any { it.getCenterTile().isVisible(civInfo) }}
-                    .toList().randomOrNull()?.cities?.filter { it.getCenterTile().isVisible(civInfo) }?.randomOrNull())
-            }
+            // There is nothing for our spy to do, put it in a random city
+            spy.moveTo(civInfo.gameInfo.getCities().filter { it.getCenterTile().isVisible(civInfo) && spy.canMoveTo(it) }.toList().randomOrNull())
```
yairm210, line 50: Separate the citiy to move to, to a val

core/src/com/unciv/logic/automation/unit/EspionageAutomation.kt
```diff
@@ -1,33 +1,83 @@
...
-            if (spy.isDoingWork()) continue
-            if (civsToStealFrom.isNotEmpty()) {
-                // We want to move the spy to the city with the highest science generation
-                // Players can't usually figure this out so lets do highest population instead
-                spy.moveTo(getCivsToStealFromSorted.first().cities.filter { it.getCenterTile().isVisible(civInfo) }.maxByOrNull { it.population.population })
+    fun automateSpies() {
+        val spies = civInfo.espionageManager.spyList
+        val spiesDoingCounterIntelligence = spies.count { it.action == SpyAction.CounterIntelligence }
+        val spiesToMove = spies.filter { it.isAlive() && 
+            (!it.isDoingWork() || (it.action == SpyAction.CounterIntelligence && spiesDoingCounterIntelligence > 2))}
```
yairm210, line 28: This is a complex boolean operation, we can make it slightly simpler by replacing the && with .filter{...},filter{...}

core/src/com/unciv/logic/automation/unit/EspionageAutomation.kt
```diff
@@ -1,33 +1,83 @@
...
+            return true
+        }
+        return false
+    }
+
+    /**
+     * Moves the spy to a random city-state
+     */
+    private fun automateSpyRigElection(spy: Spy): Boolean {
+        val potentialCities = cityStatesToRig.flatMap { it.cities }.filter { !it.isBeingRazed && spy.canMoveTo(it) }
```
yairm210, line 71: Just because we know them, doesn't mean we know the location of all their cities - can we move to a city we don't know of yet?

core/src/com/unciv/logic/automation/unit/EspionageAutomation.kt
tuvus: Actually, I can remove the second counter-intelligence clause. I had it in there before I set a time limit for how long counter-intelligence counts as work for.

core/src/com/unciv/logic/automation/unit/EspionageAutomation.kt
tuvus: Spy.canMoveTo(city) checks if we can see the city center tile. It cleans up a lot of the code here.

ravignir:
>Spies can now be killed when failing to steal tech, or when **rigging elections** with an allied spy also rigging elections. The chance of death depends on the rank of each spy. If the spy is killed, the defending spy will level up and the action will fail.

did you mean staging a coup?

tuvus:
No, from what I read, staging a coup was a different option. I haven't implemented that yet. The rigging elections code I wrote was just an implementation and wasn't going for what Civ V did.

ravignir:
> No, from what I read, staging a coup was a different option. I haven't implemented that yet. The rigging elections code I wrote was just an implementation and wasn't going for what Civ V did.

failing to rig elections does not kill a spy in civ5.

tuvus:
I haven't played it myself so I don't know and I didn't want to scour through the DLL for the one feature. So I just made my own simple implementation to push the espionage development further.

android/assets/jsons/translations/template.properties
```diff
@@ -1729,14 +1729,24 @@ Move =
...
-A spy from [civilization] stole the Technology [techName] from [cityName]! = 
+Your spy [spyName] has leveled up! = 
+Your spy [spyName] cannot steal any more techs from [civName] as we've already researched all the technology they know! = 
+# Stealing Technology defending civ
 An unidentified spy stole the Technology [techName] from [cityName]! = 
-Your spy [name] stole the Technology [techName] from [cityName]! = 
-Your spy [name] cannot steal any more techs from [civilization] as we've already researched all the technology they know! = 
-
+A spy from [civName] stole the Technology [techName] from [cityName]! = 
+A spy from [civName] tried to steal our Technology was found and killed in [cityName]! =
```
yairm210, line 1737: Badly phrased

android/assets/jsons/translations/template.properties
```diff
@@ -1729,14 +1729,24 @@ Move =
 
 After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
 We have recruited [spyName] as a spy! = 
-A spy from [civilization] stole the Technology [techName] from [cityName]! = 
+Your spy [spyName] has leveled up! = 
+Your spy [spyName] cannot steal any more techs from [civName] as we've already researched all the technology they know! = 
+# Stealing Technology defending civ
```
yairm210, line 1734: Each of these sections should have a blank line before the title

yairm210 (approved):
Just requires some template cleanup and we're good to go

PR #11079, opened 2024-02-04
Update Spanish.properties

Caballero-Arepa:
I honestly forgot that I did translations and that the long awaited intro texts are here...


yairm210:
Placeholder `[positiveAmount]` not found in `Spanish` for entry `Can trade civilization introductions for [positiveAmount] Gold`

Issue #11060, opened 2024-02-01
Adding unit conditionals to tile damage unit causes it to stop working entirely


SpacedOutChicken:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.10.5

### Describe the bug

In this mod, water (lava) tiles are supposed to deal damage to units that walk on them. I added the conditional of `<for [All] units>` to the `"Units ending their turn on this terrain take [1] damage"` unique, and it stopped working. If I remove the conditional, it works again, but with that conditional added, it does damage to no units at all.

### Steps to Reproduce

[ConditionalTileDamage.zip](https://github.com/yairm210/Unciv/files/14126796/ConditionalTileDamage.zip)
Run that mod, and park a Basic Drone on top of a Lava tile, and it will do no tile damage to the unit.

### Operating System

Windows

### Additional Information

I'm hoping to be able to target the unique to only specific units, but as it is now, adding any kind of conditional to the unique just shuts it down.

RealBamboolord:
I know this has been requested before, so afaik this isn't a bug report but rather a feature request (I'm no dev though, so I can't tell for sure)

SpacedOutChicken:
It *seems* like this should already work based on how the conditionals are written, but if this is actually a feature request, I can close this and make a comment in the modding request thread instead.

RealBamboolord:
That's probably the better option. Anyways, I support this feature request

yairm210:
After looking into this, this is limited by caching.
We have all kinds of caching going on for making movement faster, and this is one of them - since "make units avoid tiles they will be damaged by" was requested, and we need to retain speed.
I'm sure it's possible to retain much of that speed by having smart caching for "potential problem" vs "definite problem" but it's not trivial.

yairm210:
For now I'll add this to the list of uniques that suffer this same fate...

SpacedOutChicken:
I do appreciate the feature of having units automatically try to avoid tile damage, so thank you for keeping that. I wonder how this will affect my mod. Oh well, that's my problem.

PR #11080, opened 2024-02-04
Simplify tile damage code

SeventhM:
The main bulk of the code is to remove the extra caching for non standard tile damage. This is for a few main reasons
1. The unique in question `Units ending their turn on this terrain take [amount] damage` both implies it only works for terrains and is only set as a terrain unique. If any mods used this before, they likely would be warned off against it by the mod checker
2. The code around it is sketchy in the first place. Despite being a civ level cache (shouldn't this be in `CivInfoTransientCache`?) the check in MapUnit check only the unit's uniques. Once again, this implies that this never worked in this context before in the first place, unless some mad mod repeating the unique on both the civ and the unit... despite mod checker warnings
3. While they don't play nicely with the caching we have, we have alternatives to what something like #11060 wants to do, the most obvious being `This Unit takes [positiveAmount] damage <in [tileFilter] tiles> <upon turn end>` on the unit itself or at least a request for a better worded unique

Main thing about this PR I want to ask about before I go around editing the whole codebase: the function has been simplified down to more or less just `tile.allTerrains.sumOf { it.damagePerTurn }`, which reads to be more of a function in Tile itself. I don't know if I should move it now or leave this stub of a function in case we want add more support along these lines, particularly to  satisfy #11060

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -126,10 +126,6 @@ class Civilization : IsPartOfGameInfoSerialization {
     var passThroughImpassableUnlocked = false   // Cached Boolean equal to passableImpassables.isNotEmpty()
 
     @Transient
-    var nonStandardTerrainDamage = false
-
-
-    @Transient
     var thingsToFocusOnForVictory = setOf<Victory.Focus>()
 
     @Transient
@@ -701,9 +697,6 @@ class Civilization : IsPartOfGameInfoSerialization {
                     city.workedTiles.remove(workedTile)
 
         passThroughImpassableUnlocked = passableImpassables.isNotEmpty()
-        // Cache whether this civ gets nonstandard terrain damage for performance reasons.
-        nonStandardTerrainDamage = getMatchingUniques(UniqueType.DamagesContainingUnits)
-            .any { gameInfo.ruleset.terrains[it.params[0]]!!.damagePerTurn != it.params[1].toInt() }
 
         hasLongCountDisplayUnique = hasUnique(UniqueType.MayanCalendarDisplay)
 
--- a/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
+++ b/core/src/com/unciv/logic/map/mapunit/MapUnit.kt
@@ -485,14 +485,6 @@ class MapUnit : IsPartOfGameInfoSerialization {
     }
 
     fun getDamageFromTerrain(tile: Tile = currentTile): Int {
-        if (civ.nonStandardTerrainDamage) {
-            for (unique in getMatchingUniques(UniqueType.DamagesContainingUnits)) {
-                if (unique.params[0] in tile.allTerrains.map { it.name }) {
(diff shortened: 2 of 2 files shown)


SeventhM:
> * This satisfies the "Carthage won't slow down the entire game" wish,

Uhhh... That's not the unique Carthage has
1. Carthage's unique is `Units ending their turn on [Mountain] tiles take [50] damage`. Note the use of the word "mountain" instead of "this terrain". Actually, this isn't even a real unique. It's a comment
2. If this did supposedly trigger when someone played Carthage, still bad news: that would not have actually done anything . Like I said, despite caching for the Civ (in the wrong place I might add), the implementation actually checked the \*unit's\* uniques, meaning it would never trigger in the first place. The only reason it would still take 50 damage is because of the use of an early return rather than an if else

Though it does read to me now that something must've gone wrong and someone forgot what it originally was like. Because the implementation (despite incorrectly checking the unit's uniques) does actually read like Carthage's unique and would probably crash if taken seriously (there's no such thing as a params[1] in the current unique). I guess someone edited the unique's text at some point

SeventhM:
https://github.com/yairm210/Unciv/commit/8eb24ac273a44460d82e19f31c869e7cae3da7bb
Ah, I see. getMatchingUniques used to always get civ uniques, unlike currently where by default it doesn't. Also yep, the Carthage version was removed somewhere (I don't feel like finding the commit for it this second, maybe later) and got confused for the general purpose one. Not that it matters because the Carthage one was redundant. Good to know

SeventhM:
> * Yes, comments are redundant

Actually, note that until https://github.com/yairm210/Unciv/commit/086867731d05f2ba143c3e1b43df284ab09f3d7f, Carthage's unique did actually function as a separate unique. When I say "it is redundant", my point is less "this is useful for explaining but doesn't actually do anything", but rather "this leads to a completely different code path for the exact same result". Sure, back then there wasn't a meaningful was to write comment or filtering uniques, but I'd still argue that the code was kept up to standards as the time passed. It's kinda telling we don't have an issue asking it to be fixed while mods like Deciv started quietly dropping even the comment

> * No value attached to that 'should', not discouraging alternative implementations

I am discouraging alternative implementations. The whole point of using plain text for uniques is that they're descriptive and self explanatory. If we want it attached to Civs, perhaps it we should bring back the original Carthage unique and fix the code so that we actually check the civ's uniques and not the unit's. If we're putting it on units, then perhaps we should make a unique that's actually sensible in units. If you want there to be multiple interpretations, then perhaps we shouldn't use placeholder text that implies otherwise. `Units take [amount] damage at the end of their turn` is far more flexible than what we currently have

Which kinda leads to my point. Worst case, the current code \*is\* bad. We have a cache for a nonfunctional branch of code, which means even if the amount of memory/time saved is miniscule, this is a very obvious optimization to delete the code path entirely.  Best case, we put in some actually replacement to make the code actually function. But the solution 100% shouldn't be to force the wrong unique in the wrong location and say "it looks close enough". Especially for a unique that no modder (or at least none out of the vast majority of modders, ***including*** the mad ones) currently is using in this context

yairm210 (approved):

PR #11081, opened 2024-02-04
Split Strat Balance and Legend Start into its own checkbox

SeventhM:
~~This totally isn't a selfish attempt to add in Lekmod features.~~ The legendary start/strategic balance options aren't particularly related to the amount of resources available. Imo, besides UI clutter, I don't see why the options needs to be exclusive

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/mapregions/LuxuryResourcePlacementLogic.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/mapregions/LuxuryResourcePlacementLogic.kt
@@ -239,7 +239,8 @@ object LuxuryResourcePlacementLogic {
         for (region in regions) {
             val tilesToCheck = tileMap[region.startPosition!!].getTilesInDistanceRange(1..2)
             val candidateLuxuries = randomLuxuries.shuffled().toMutableList()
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/logic/map/mapgenerator/mapregions/LuxuryResourcePlacementLogic.kt
```diff
@@ -239,7 +239,8 @@ object LuxuryResourcePlacementLogic {
         for (region in regions) {
             val tilesToCheck = tileMap[region.startPosition!!].getTilesInDistanceRange(1..2)
             val candidateLuxuries = randomLuxuries.shuffled().toMutableList()
-            if (tileMap.mapParameters.mapResources != MapResources.strategicBalance)
+            if (tileMap.mapParameters.mapResources != MapResources.strategicBalance &&
```
yairm210, line 242: mapResources can also be strategic balance/legendary start? Do we want to support both types, or are we deprecating the old one? If the first, I don't understand why. If the second, we should add deprecation warnings on those mapresource vals, for warning.

SeventhM: I'd say the second

PR #11087, opened 2024-02-06
docs: fix light/dark mode tooltip in the wiki

alexban011:
I just ~~stole~~ borrowed your mkdocs.yml config and noticed the tooltip for the light/dark mode button are reversed. 

When in light mode it said switch to light mode which was probably a mistake. :D

Same for dark mode


yairm210:
Glad you're finding it useful!
And thanks for the fix :)

PR #11095, opened 2024-02-06
Patch AlertPopup to correctly close when it cannot find a required asset

### LINKED ISSUE #11090: UnCiv freezes after liberation of worker barbarians on foreign land without open borders

### Game Version

4.10.6 (build 965)

### Describe the bug

Hello,

I have a scout, which has the ability to enter all fields without open borders due to a promotion. Now I wanted to free a foreign barbarian worker unit. However, this resulted in UnCiv not responding. I have attached a video about this.

Unfortunately I could not find the game file with the problem, but I could create another one which has the same problem.
After a restart it sometimes worked, but sometimes not.

### Steps to Reproduce

1. Have scout unit with corresponding property
2. Free a barbarian worker unit on foreign land3.

### Screenshots

https://github.com/yairm210/Unciv/assets/46723399/7d42adc0-c142-4e8f-969c-c6e032f1a310
(Compressed, as GitHub does not allow uploads over 10 MB. [Original file](https://upload.disroot.org/r/chxj_Bd8#yDdEh3QIEd/6jT5BwaRnWnK0KVXaW+gltknJHBav73U=))

### Link to save file

[Logiker (191 Runden).json](https://github.com/yairm210/Unciv/files/14180076/Logiker.191.Runden.json)

### Operating System

Android

### Additional Information

Also 4.10.7 (build 966) is also affected.

SomeTroglodyte:
One way to fix #11090, though not the only one possible, and leaves questions open.
Compare the result of capturing that settler pre-patch as shown in the issue with:
image

* So the Settler was thrown off the tile into the sea where it has a right to exist
* It also was not converted to a Worker as it should when captured (but I've seen the code preparing that by looking for what a Worker _is_ running while debugging)
* Could we detect that teleport and change the PopupAlert's value? (this hardening would still have independent merit)
* Should we instead look for unique identifiers for units, like cities have, which would not only be a better `value`to pass here (but breaking save compat), but also in some other places - like upgrading from unit overview, selection in some cases,...
* Could also have been done by making Popup.open open and overriding it, and could also have been done without flag - at that point, 'no cells' could also be tested. I threw the dice and did this instead.
* Entirely untreated: That observation that capture code (actually, UnitManager.placeUnitNearTile) would run UniqueTriggerActivation.triggerUnique on _all_ uniques of a created Worker, just not the ones having a trigger conditional... Or why that function doesn't save 76 indentation spaces and a line by `?: return null`

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
@@ -40,6 +40,10 @@ import java.util.EnumSet
  * [Popup] communicating events other than trade offers to the player.
  * (e.g. First Contact, Wonder built, Tech researched,...)
  *
+ * **Opens itself at the end of instantiation!**
+ *
+ * (In rare cases, it chooses not to: Mods making a RecapturedCivilian not find the unit as it was illegal and removed after the actual capture)
+ *
  * Called in [WorldScreen].update, which pulls them from viewingCiv.popupAlerts.
  *
  * @param worldScreen The parent screen
@@ -73,6 +77,8 @@ class AlertPopup(
         addCloseButton(text, KeyboardBinding.NextTurnAlternate, null)
 
     init {
+        var skipThisAlert = false
+
         // This makes the buttons fill up available width. See comments in #9559.
         // To implement a middle ground, I would either simply replace growX() with minWidth(240f) or so,
         // or replace the Popup.equalizeLastTwoButtonWidths() function with something intelligent not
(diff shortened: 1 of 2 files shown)


SeventhM:
Don't feel like downloading the concoction of mods that original issue had. Though, this seems simple enough that I shouldn't need to. Would take me significantly more time to dig through the popup code for me to be certain the right solution here (I'm aware of how most of `logic` and `models` subsections of the code works, but not so much ui, automation, or especially the platform specific stuff)

> * It also was not converted to a Worker as it should when captured

Weird. Do you know who was the original owner?

Also, there's a lot a weirdness imo in `captureOrConvertToWorker`.
1. We destroy before even checking if there's a unit to replace it with. not weird in itself, but should we notify the user about this when we do run into this situation?
2. `workerTypeUnit` can hit units that shouldn't be available. Depending on a mod's structure, you can actually hit a civ specific unit. `PlaceUnitNearTile` intentionally doesn't address either. Probably should add in an ``&& it.getRejectionReasons(capturingCiv).any { reason -> !reason.isConstructionRejection() }``
3. `capturedBy()` only moves the unit while converting a unit to a worker creates a new unit. Inconsistency? Relevant thing here is triggers and a few extras that creating a new unit does if I remember correctly. Also, haven't checked how this affects movement

> * Could we detect that teleport and change the PopupAlert's value?

#11096 should fix it. Perhaps `captureOrConvertToWorker` should also return the unit to make it extra easy

> * would run UniqueTriggerActivation.triggerUnique on _all_ uniques of a created Worker

Discussion in #11090. Like I said, only real thing to add is an check for whether the unique is triggerable. Better for performance, I guess, otherwise, not really the biggest deal

yairm210 (approved):

PR #11100, opened 2024-02-08
Add unique for increasing price every time it's built

SeventhM:
Also extends the unique for cost increasing for number of cities to units, as there's no real harm to it

Side note... Why exactly do we not have a version to pass in the city? That seems like it'll be more useful, at least here (maybe not for unit gold prices) than the civ itself for the sake of StateForConditionals

### DIFF
--- a/core/src/com/unciv/logic/civilization/CivConstructions.kt
+++ b/core/src/com/unciv/logic/civilization/CivConstructions.kt
@@ -21,6 +21,9 @@ class CivConstructions : IsPartOfGameInfoSerialization {
     /** Maps construction names to the amount of times bought */
     val boughtItemsWithIncreasingPrice: Counter<String> = Counter()
 
+    /** Maps construction names to the amount of times bought */
+    val builtItemsWithIncreasingPrice: Counter<String> = Counter()
+
     /** Maps cities by id to a set of all free buildings by name they contain.
      *  The building name is the Nation-specific equivalent if available.
      *  Sources: [UniqueType.FreeStatBuildings] **and** [UniqueType.FreeSpecificBuildings]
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/logic/civilization/CivConstructions.kt
```diff
@@ -21,6 +21,9 @@ class CivConstructions : IsPartOfGameInfoSerialization {
     /** Maps construction names to the amount of times bought */
     val boughtItemsWithIncreasingPrice: Counter<String> = Counter()
 
+    /** Maps construction names to the amount of times bought */
```
yairm210, line 24: "Built"

core/src/com/unciv/models/ruleset/unit/BaseUnitCost.kt
```diff
@@ -11,6 +11,13 @@ class BaseUnitCost(val baseUnit: BaseUnit) {
 
     fun getProductionCost(civInfo: Civilization): Int {
         var productionCost = baseUnit.cost.toFloat()
+
+        for (unique in baseUnit.getMatchingUniques(UniqueType.CostIncreasesPerCity, StateForConditionals(civInfo)))
```
yairm210, line 15: Sounds to me like we really need a shared function in a parent class

SeventhM: You mean like `IConstruction.getProductionCost`? Shared function exists, but is blank rn

yairm210: Looking over this again the shared code is like 7 lines in total, I think it's fine to leave as-is.

yairm210 (approved):

PR #11101, opened 2024-02-08
Conditional for building in amount of cities

SeventhM:


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -641,6 +641,8 @@ enum class UniqueType(
     ConditionalAfterGeneratingGreatProphet("after generating a Great Prophet", UniqueTarget.Conditional),
 
     ConditionalBuildingBuilt("if [buildingFilter] is constructed", UniqueTarget.Conditional),
+    ConditionalBuildingBuiltAll("if [buildingFilter] is constructed in all [cityFilter] cities", UniqueTarget.Conditional),
+    ConditionalBuildingBuiltAmount("if [buildingFilter] is constructed in [positiveAmount] [cityFilter] cities", UniqueTarget.Conditional),
```
yairm210, line 645: "in at least", so people don't think they can make different effects for 1,2,3 etc

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -165,6 +165,12 @@ object Conditionals {
 
             UniqueType.ConditionalBuildingBuilt ->
                 checkOnCiv { cities.any { it.cityConstructions.containsBuildingOrEquivalent(condition.params[0]) } }
+            UniqueType.ConditionalBuildingBuiltAll ->
+                checkOnCiv { cities.all { it.cityConstructions.containsBuildingOrEquivalent(condition.params[0])
```
yairm210, line 169: Noooononono
The filter for the cities *must* be applied before the .all{} otherwise irrelevant cities cause failure

SeventhM:
> Is Replacing `RequiresBuildingInAllCities` and `RequiresBuildingInSomeCities` with `OnlyAvailable` plus these here planned?

.........Yes, totally was the plan. Tbh, had other uses, but that should just be deprecated here, shouldn't it?

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
SeventhM: Was going to try to be funny and say
```kotlin
            checkOnCiv { cities.none { !(!it.cityConstructions.containsBuildingOrEquivalent(condition.params[0])
                    && it.matchesFilter(condition.params[1])) } }
```
is faster and lazier, but, no, looks ugly

SeventhM:
Also, Idk if should be a separate PR, but why doesn't this also target units?

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -282,8 +282,10 @@ enum class UniqueType(
 
     CostIncreasesPerCity("Cost increases by [amount] per owned city", UniqueTarget.Building),
 
+    @Deprecated("as of 4.10.17", ReplaceWith("Only available <if [buildingFilter] is constructed in all [All] cities>"))
```
yairm210, line 285: "all all cities" looks SO bad, this is a blocker for me. 
We can either have a non-cityfilter version, or find a nicer looking cityfilter, not this.

SeventhM: I'll be fine with a non cityFilter version (though this does still seem to me like a spot where a cityFilter would be wanted. There's some awkwardness with city filters/conditionals and other conditionals/triggers in general). But that'll have to wait a few hours so I can sleep

SeventhM: Side note: there's a general question I have on whether these functions should include an `isPuppet` check or not. Seems weird not to, given that's how the current "requires in all cities" unique works and you can't control puppets, but I can see the argument otherwise as well. On the bright side, it cleans up the text a bit (I probably should've set the depreciation message to "in all [non-[Puppet] cities`

Side-side note: should this even use the newer cityFilter filters if cityFilter is kept? Or should it use the old ones. New ones are probably easier to work with, but they do miss out on some options. Thoughts to consider

yairm210: The new ones - if we're missing options we can fill them in :)

SeventhM: How about ”in all among [filter] cities”? Changes line to read as ”in all among [All] cities”

SeventhM: I feel like intuitively, global checks should probably be a separate unique, but I can be willing to back down on that front

yairm210:
I think adding the "among" makes it worse.
Since we have the non-puppet cities "saving" us here, we can just say that we don't advocate using "all" anywhere and call it a day.
In other news, it's bothered me for a while that "All" is capitalized, I'll add a non-capitalized version for All filters... I mean *all* filters

SeventhM:
> I think adding the "among" makes it worse.

Well, I'm not quite sure what the wording here should be then. Pretty sure the word "all" (or "every" or similar) should be here so it's clear it's different from the conditional of it being in at least one city. `is constructed in [non-[Puppet]] cities` is somewhat vague

yairm210:
Sorry for being unclear
Let's go with "all", since the exclusion of puppets means we should never have "all all" appear as text :)

SeventhM:
Now that I look at the other one, `is constructed in at least [3] [Your] cities` doesn't read right. `is constructed in at least [3] of [Your] cities`? Hmm.. might go with that. Could also help the other `is constructed in all of [All] cities`?

SeventhM:
Why does `non-[Puppet]` not trigger the test, but `non-[Puppeted]` does? That's weird. I'm pretty sure the registered filter is "Puppeted"

Oh, and `CityFilter.isKnownValue` should include `CivFilter.isKnownValue` now

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -255,9 +256,45 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
...
+                                    yield(RejectionReasonType.RequiresBuildingInSomeCities.toInstance(
+                                        "Requires a [$building] in at least [$amount] cities" +
+                                            " ($numberOfCities/$numberOfCities)"))
+                                }
+                            }
+                            UniqueType.ConditionalBuildingBuiltAll -> {
+                                val building = civ.getEquivalentBuilding(conditional.params[0]).name
+                                val cityFilter = conditional.params[1]
+                                if(civ.cities.any { it.matchesFilter(cityFilter)
+                                    !it.isPuppet && !it.cityConstructions.containsBuildingOrEquivalent(building)
```
yairm210, line 285: I'm missing an && between the matchesFilter and the !it.isPuppet - what am I missing here?

core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -255,9 +256,45 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
...
+                                }) {
+                                    displayable = true
+                                    yield(RejectionReasonType.RequiresBuildingInAllCities.toInstance(
+                                        "Requires a [${building}] in all cities"))
+                                }
+                            }
+                            else -> { shouldYield = true}
+                        }
+                    }
+                    if (!displayable && shouldYield)
```
yairm210, line 295: Why do we need both of these? What are the two conditions we're trying to avoid? I only see one, if displayable is false then shouldYield will always be true, no?

core/src/com/unciv/ui/objectdescriptions/BuildingDescriptions.kt
```diff
@@ -93,12 +80,38 @@ object BuildingDescriptions {
...
+                                    || !it.matchesFilter(conditional.params[1])
+                            }
+                            // Could be red. But IMO that should be done by enabling GDX's ColorMarkupLanguage globally instead of adding a separate label.
+                            if (missingCities.isNotEmpty()) translatedLines += "\n" +
+                                "[${city.civ.getEquivalentBuilding(unique.params[0])}] required:".tr() +
+                                " " + missingCities.joinToString(", ") { it.name.tr(hideIcons = true) }
+                            // Can't nest square bracket placeholders inside curlies, and don't see any way to define wildcard placeholders. So run translation explicitly on base text.
+                        }
+                    }
+            }
```
yairm210, line 114: This really feels like a separate function

core/src/com/unciv/models/ruleset/Building.kt
SeventhM: No, if all conditionals are valid, it shouldn't yield the rejection. It's ugly, yes, but not sure the better solution

core/src/com/unciv/models/ruleset/Building.kt
SeventhM: That's... Huh... No warning. I guess kotlin assumes a semicolon here. Weird

The `!it.isPuppet` is now redundant anyways... But huh

core/src/com/unciv/ui/objectdescriptions/BuildingDescriptions.kt
SeventhM: Not sure what the fix is for the "text should be red" comment. Tbh, put basically no effort into looking for the solution, though. `FormattedLine`? Nah, needs to be a string. Oh... huh, actually, this just doesn't support color because of that fact. Interesting

core/src/com/unciv/models/ruleset/Building.kt
SeventhM: I just realized I'm dumb here. If we reach the else condition in here, we should just immediately yield and not need these bools at all

core/src/com/unciv/ui/objectdescriptions/BuildingDescriptions.kt
SeventhM: Future me's problem. We're still missing a few stuff I the description of differences for replacement units such as tech differences, great person points, or specialists. I'm willing to call that as not relevant to this pr but something to note for later

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -283,8 +283,10 @@ enum class UniqueType(
     CostIncreasesPerCity("Cost increases by [amount] per owned city", UniqueTarget.Building, UniqueTarget.Unit),
     CostIncreasesWhenBuilt("Cost increases by [amount] when built", UniqueTarget.Building, UniqueTarget.Unit),
 
+    @Deprecated("as of 4.10.17", ReplaceWith("Only available <if [buildingFilter] is constructed in all [non-[Puppeted]] cities>"))
     RequiresBuildingInAllCities("Requires a [buildingFilter] in all cities", UniqueTarget.Building),
-    RequiresBuildingInSomeCities("Requires a [buildingFilter] in at least [amount] cities", UniqueTarget.Building),
+    @Deprecated("as of 4.10.17", ReplaceWith("Only available <if [buildingFilter] is constructed in at least [positiveAmount] [All] cities>"))
```
yairm210, line 288: HMMM
Another bad one
We can start with replacing only the "all" unique for now, since that seems fine
Or we can find a solution for this as well 
I mean, we could go with "non-puppeted" for this as well, perhaps...? But that does change the meaning, and makes it more restrictive, rather than less...

yairm210: "of your"? "owned"? 
"of your" sounds like the best *English*, but I wonder if other languages will have difficulty..."owned" sounds like the correct solution for a language-foible-agnostic solution

yairm210:
I honestly didn't expect this to not enter the next version, but sometimes things are more complicated than they first appear

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
SeventhM: Well that's goes back to my comment [above](https://github.com/yairm210/Unciv/pull/11101#issuecomment-1937380146). "Of [cityFilter] cities" reads best for some filters, less so for others. Kinda split on it myself tbh

yairm210 (approved):

Issue #11105, opened 2024-02-08
Building on a system with only Java 17 installed


SomeTroglodyte:
### Problem Description

Our current project files only build on systems with a Java 11 installed - as in, installed system-wide and detectable through path and/or JAVA_HOME. This box here has only Java 17 and I refuse to pollute it with such disgusting old smelly old cheese like a Java 11 (joking of course)...

### Related Issue Links

PR #10740

### Desired Solution

Throw the dice?

### Alternative Approaches

* Document the requirement for Java 11 preinstalled
* Give jvmToolchain the required config it needs to auto-download a jre on the fly
* Change build scripts (and packaged jre's!) to pure Java 17 (and see to activating the Z allocator!)

I like none of them, not entirely.

### Additional Context

I still got build issues since gradle 8. Don't want to install a Java 11, and changing build scripts around without committing them is - not too hard, but fails from time to time. I think there's an outstanding issue with gradle that is meant to go into 8.5 but still isn't ("boohoo kotlin and java targets are not equal" and compilation aborts). Also, build scripts that work in one branch suddenly fail in another, reindexing orgies sometimes help somtimes not.

**But** - I've learned the cross-compile capability seems gone from gradle or at least it's undocumented (their documentation is all but unreadable): jvmToolchain(17) + targetCompatibility = JavaVersion.VERSION_11 produces binaries (if it compiles at all) that will **not** run on a jre 11, they're marked for 17. sourceCompatibility changes nothing. I believe targetCompatibility has been changed to be ignored for kotlin sources, might still work for pure java sources. I'm currently debugging only on pure 17 builds, with the entire java-specific blocks (in 3 kts scripts - core desktop testing) removed, so far no build has failed, only leaving the targetCompatibility statements in leads to the mentioned compile and indexing fails.

Note the android builds seem to control target jre level through an entirely different mechanism (AGP probably working around the gradle bugs) and are not affected, cross-compile to java8-compatible binaries works in this case.

yairm210:
I see 
That *is* a problem, I wonder if we can use something that's not toolchain to specify target language version...

PR #11111, opened 2024-02-09
Compile on jdk 17 FOR jre 11, includes gradle upgrade

yairm210:
Credit @SomeTroglodyte 

Part of #11105


yairm210:
And #11111! What luck, I get a *checks notes* chance to debug java compilation version compatibility, whatever that means

PR #11115, opened 2024-02-10
Autoplay menu cleanup

SomeTroglodyte:
Since I recently used that thing for the first time, I noticed things...
* Slightly off the screen at the top (unless cramped display moved the button down)[^1]
* All key bindings the same - and "Up" somehow was inappropriate
* Kept the "I might not actually want to open" code needlessly

This is why I looked at the code:
image
Better:
image
Bindings (another two are down in the WorldScreen category):
image

</details>

[^1]: Due to YAGNI, the component's clamp-to-stay-within-the-stage was only done for the unit upgrade menu which always sits close to bottom left...


android/assets/jsons/translations/French.properties
```diff
@@ -2568,6 +2568,7 @@ You returned captured units to us = Vous nous avez rendu des unités capturées
 
 Main Menu = Menu principal
 World Screen = Écran de jeu
+Auto Play Menu = Menu Jeu Auto
```
yairm210, line 2571: Why is this "Auto Play" when everything else is AutoPlay? Let's be consistent
I know it's not your fault, but if you're already at it...

yairm210: We can probably just replace all English text and keep existing translations

yairm210 (approved):

PR #11119, opened 2024-02-10
Workers wake up on tile expansion

tuvus:
It does as stated. This allows players to pre-emptively put their workers on tiles that the city may expand to in order to construct the improvement there.
Should the city look for workers on tiles around the tile that has been expanded to?


yairm210 (approved):

yairm210:
Why would the city want to do that...? What case are we talking about?

PR #11126, opened 2024-02-11
proper vs cityFilter conditional

yairm210:
- combatantFilter was a mistake - almost all places we use it have "[combatantFilter] units" which gives the game away.
- The only time we'll have a single effect for both cities and units is if we're using a civFilter - so let's just say so directly
- The cityFilter that I shoved into the combatant was not the right place - this should be part of the "vs cities" conditional soyou can say which cities, just like with units

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -1,6 +1,7 @@
 package com.unciv.models.ruleset.unique
 
 import com.unciv.logic.GameInfo
+import com.unciv.logic.battle.CityCombatant
 import com.unciv.logic.battle.CombatAction
 import com.unciv.logic.battle.MapUnitCombatant
 import com.unciv.logic.city.City
@@ -188,8 +189,9 @@ object Conditionals {
             UniqueType.ConditionalWhenGarrisoned ->
                 checkOnCity { getCenterTile().militaryUnit?.canGarrison() == true }
 
-            UniqueType.ConditionalVsCity -> state.theirCombatant?.matchesFilter("City") == true
-            UniqueType.ConditionalVsUnits,  UniqueType.ConditionalVsCombatant -> state.theirCombatant?.matchesFilter(condition.params[0]) == true
+            UniqueType.ConditionalVsCity -> state.theirCombatant?.let {
+                it is CityCombatant && it.city.matchesFilter(condition.params[0]) } == true
+            UniqueType.ConditionalVsUnits,  UniqueType.ConditionalVsCiv -> state.theirCombatant?.matchesFilter(condition.params[0]) == true
             UniqueType.ConditionalOurUnit, UniqueType.ConditionalOurUnitOnUnit ->
                 relevantUnit?.matchesFilter(condition.params[0]) == true
             UniqueType.ConditionalUnitWithPromotion -> relevantUnit?.promotions?.promotions?.contains(condition.params[0]) == true
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -351,7 +351,7 @@ enum class UniqueType(
     FlankAttackBonus("[relativeAmount]% to Flank Attack bonuses", UniqueTarget.Unit, UniqueTarget.Global),
     @Deprecated("as of 4.10.3", ReplaceWith("[+30]% Strength <vs [City-States]>"))
     StrengthBonusVsCityStates("+30% Strength when fighting City-State units and cities", UniqueTarget.Global),
-    StrengthForAdjacentEnemies("[relativeAmount]% Strength for enemy [combatantFilter] units in adjacent [tileFilter] tiles", UniqueTarget.Unit),
+    StrengthForAdjacentEnemies("[relativeAmount]% Strength for enemy [mapUnitFilter] units in adjacent [tileFilter] tiles", UniqueTarget.Unit),
     StrengthWhenStacked("[relativeAmount]% Strength when stacked with [mapUnitFilter]", UniqueTarget.Unit),  // candidate for conditional!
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -188,8 +189,9 @@ object Conditionals {
             UniqueType.ConditionalWhenGarrisoned ->
                 checkOnCity { getCenterTile().militaryUnit?.canGarrison() == true }
 
-            UniqueType.ConditionalVsCity -> state.theirCombatant?.matchesFilter("City") == true
-            UniqueType.ConditionalVsUnits,  UniqueType.ConditionalVsCombatant -> state.theirCombatant?.matchesFilter(condition.params[0]) == true
+            UniqueType.ConditionalVsCity -> state.theirCombatant?.let {
+                it is CityCombatant && it.city.matchesFilter(condition.params[0]) } == true
+            UniqueType.ConditionalVsUnits,  UniqueType.ConditionalVsCiv -> state.theirCombatant?.matchesFilter(condition.params[0]) == true
```
SeventhM, line 194: 1. `VsUnits` should also have an `it is MapUnitCombatant` check
2. I don't really like having `vs civFilter` counting on combatant.matchesFilter. If anything, we should be using `combatant.getCivInfo.matchesFilter`

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -367,7 +367,7 @@ enum class UniqueType(
     StatsWhenSpreading("When spreading religion to a city, gain [amount] times the amount of followers of other religions as [stat]", UniqueTarget.Unit, UniqueTarget.Global),
 
     // Attack restrictions
-    CanOnlyAttackUnits("Can only attack [combatantFilter] units", UniqueTarget.Unit),
+    CanOnlyAttackUnits("Can only attack [mapUnitFilter] units", UniqueTarget.Unit),
```
SeventhM, line 370: Haven't checked, but I'm concerned if this affects any uniques that used this to limit to attacking cities. This seems like the only direct way to spell it out

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -429,7 +429,7 @@ enum class UniqueType(
     UnitUpgradeCost("[relativeAmount]% Gold cost of upgrading", UniqueTarget.Unit, UniqueTarget.Global),
 
     // Gains from battle
-    DamageUnitsPlunder("Earn [amount]% of the damage done to [combatantFilter] units as [civWideStat]", UniqueTarget.Unit, UniqueTarget.Global),
+    DamageUnitsPlunder("Earn [amount]% of the damage done to [mapUnitFilter] units as [civWideStat]", UniqueTarget.Unit, UniqueTarget.Global),
```
SeventhM, line 432: Doubly so here, as I think (but haven't checked) this actually affects some units in the built in rulesets

SeventhM:
On second thought, `vs [units]` or `vs [cities]` would probably be nice syntax wise and reduce the amount of in use uniques, rather than increase. I almost wonder if this is going in the opposite direction as it should: perhaps we should remove "units" from the text of the other combatantFilter uniques, rather than swap to `mapUnitFilter` here

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
yairm210: I see, we can dispose of the ICombatant.matchesFilter entirely
Correct!

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: `Cannot attack <vs [all] cities>` ?

yairm210: I see that `cannot attack` currently doesn't respect conditionals, sounds like a precondition

SeventhM: > `Cannot attack <vs [all] cities>` ?

No, that's `Cannot attack <vs [non-[all]] cities>` which reads clunky or `Cannot attack <vs [all] units>` which is kinda roundabout

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -672,9 +672,13 @@ enum class UniqueType(
     ConditionalOurUnitOnUnit("when [mapUnitFilter]", UniqueTarget.Conditional), // Same but for the unit itself
     ConditionalUnitWithPromotion("for units with [promotion]", UniqueTarget.Conditional),
     ConditionalUnitWithoutPromotion("for units without [promotion]", UniqueTarget.Conditional),
-    ConditionalVsCity("vs cities", UniqueTarget.Conditional),
+
+    ConditionalVsCity("vs [cityFilter] cities", UniqueTarget.Conditional),
+    @Deprecated("as of 4.10.9", ReplaceWith("vs [all] cities"))
+    ConditionalVsCityOld("vs cities", UniqueTarget.Conditional),
```
SeventhM, line 678: Just realized, you forgot to leave behind a temporary conditional logic for the older legacy version

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: I really didn't want to go this route, but we *could* have nested filters.
Just like "non-[]", we can have "[cityFilter] cities" and "[mapUnitFilter] units" be legit combatantFilters
This would mean changing a bunch of actual uniques as well though

SeventhM: Side irrelevant question: should empty string be added to Constants.all? Allows for `vs [non-[]] cities`

yairm210:
My conclusion is that *most places where we have unit combatant we can also have city combatants*, so we should ditch this approach and go for one where combatantFilter has 3 acceptable components:
- "[mapUnitFilter] units"
- "[cityFilter] cities"
- civFilter

PR #11128, opened 2024-02-13
Framework stuff: Preparation for another Warning-Suppression branch

SomeTroglodyte:
This is everything from my Suppression branches not directly suppression-related. See commit messages. Should have no visible effect.

The now unused field+parameter will allow `RulesetErrorList.add` to do the entire suppression filtering, allowing attaching the suppression as separate unique or as modifier to the triggering unique, while still allowing suppression of warnings about non-IHasUniques contexts.

Note: Basing the existing duplicate suppression on hashing can speed things up a little bit. Using a `Map<String, RulesetError>` (found no better alternative - a LinkedHashSet will _not_ do, and it's a map under the hood anyway, and `Map<RulesetError, Severity>` would be a tad more complicated):
<details><summary>timings</summary>

Modcheck with Mods (random collection from analyzing issues): 
```
[5Hex Tileset, AFX, Additional Music Various, Aggtelek, Aliens have crashed on Earth, BabylonUI, Barbarian xp farm, Barbarians but Better, Better Pantheons, Civ V - Gods & Kings, Civ V - Vanilla, Civ V Leader portraits, Civilopedia River Image, DeCiv Redux, Difficulties, Dreams of Steam, Extra Buildings, Extra Promotions, Extra Units, Extra Wonders, Future Expansion 2022 09 18, Future Expansion Continued, Great Farmer, Higher quality builtin sounds, IceBreaker, Logicians, Modern Joke Religions, More Luxuries, Neuroleaders, Policy Colors, Policy mod, Resource Recyclers, Smoky Skies Redux, Warfare Expanded Lite, Yet Another Resource Mod]
```
... switching from base none to G&K and back, discarding first 4 results:
- no base: 442ms, 422ms, 396ms
- Base = G&K: 501ms, 434ms, 450ms

map-based RulesetErrorList:
- no base: 322ms, 286ms, 313ms
- Base = G&K: 398ms, 408ms, 358ms

</details>
... is that worth it? If yes, in case I lose it - on top of this:
<details><summary>patch</summary>

```patch
Subject: [PATCH] Map-based RulesetErrorList
---
Index: core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt
IDEA additional info:
<+>UTF-8
diff --git a/core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt b/core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt
--- a/core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt	
+++ b/core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt	
@@ -26,7 +26,9 @@
  */
 class RulesetErrorList(
     ruleset: Ruleset? = null
-) : ArrayList<RulesetError>() {
+) : Collection<RulesetError> {
+    private val map = linkedMapOf<String, RulesetError>()
+
     /** Add an [element], preventing duplicates (in which case the highest severity wins).
      *
      *  [sourceObject] is for future use and should be the originating object. When it is not known or not a [IHasUniques], pass `null`.
@@ -43,28 +45,28 @@
     fun add(sourceObject: IHasUniques?, text: String, errorSeverityToReport: RulesetErrorSeverity = RulesetErrorSeverity.Error) =
         add(sourceObject, RulesetError(text, errorSeverityToReport))
 
-    @Deprecated("No adding without explicit source object", ReplaceWith("add(sourceObject, element)"))
-    override fun add(element: RulesetError) = super.add(element)
-
     /** Add all [elements] with duplicate check, but without suppression check */
-    override fun addAll(elements: Collection<RulesetError>): Boolean {
+    fun addAll(elements: Collection<RulesetError>): Boolean {
         var result = false
         for (element in elements)
             if (addWithDuplicateCheck(element)) result = true
         return result
     }
 
-    private fun addWithDuplicateCheck(element: RulesetError) =
-        removeLowerSeverityDuplicate(element) && super.add(element)
+    private fun addWithDuplicateCheck(element: RulesetError): Boolean {
+        if (!removeLowerSeverityDuplicate(element)) return false
+        map[element.text] = element
+        return true
+    }
 
     /** @return `true` if the element is not present, or it was removed due to having a lower severity */
     private fun removeLowerSeverityDuplicate(element: RulesetError): Boolean {
         // Suppress duplicates due to the double run of some checks for invariant/specific,
         // Without changing collection type or making RulesetError obey the equality contract
-        val existing = firstOrNull { it.text == element.text }
+        val existing = map[element.text]
             ?: return true
         if (existing.errorSeverityToReport >= element.errorSeverityToReport) return false
-        remove(existing)
+        map.remove(existing.text)
         return true
     }
 
@@ -92,6 +94,28 @@
                         it.text.replace('<','〈').replace('>','〉')
                 }
 
+    //region Collection interface
+    override val size: Int
+        get() = map.size
+
+    override fun isEmpty() = map.isEmpty()
+
+    override fun iterator() = map.values.iterator()
+
+    override fun containsAll(elements: Collection<RulesetError>): Boolean = throw UnsupportedOperationException()
+
+    override fun contains(element: RulesetError) = map.containsKey(element.text)
+
+    fun <R: Comparable<R>> sortByDescending(selector: (RulesetError) -> R) {
+        @Suppress("UNCHECKED_CAST") val copy = map.clone() as LinkedHashMap<String, RulesetError>
+        map.clear()
+        for (select in copy.values.map(selector).distinct().sortedDescending())
+            for ((key, error) in copy)
+                if (selector(error) == select)
+                   map[key] = error
+    }
+    //endregion
+
     companion object {
         private val empty = RulesetErrorList()
 
```
</details>

~.... Ugh, no, wait, this is missing the tidbit that a few UniqueValidator methods like checkUntypedUnique return Lists, while they should return RulesetErrorList ....~ ...done, last two commits.


core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt
```diff
@@ -60,4 +91,21 @@ class RulesetErrorList : ArrayList<RulesetError>() {
...
+        private val empty = RulesetErrorList()
+
+        fun of(ruleset: Ruleset? = null, sourceObject: IHasUniques? = null, vararg errors: RulesetError): RulesetErrorList {
+            if (errors.isEmpty()) return empty
+            val result = RulesetErrorList(ruleset)
+            for (error in errors)
+                result.add(sourceObject, error)
+            return result
+        }
+        fun of(ruleset: Ruleset? = null, sourceObject: IHasUniques? = null, text: String, severity: RulesetErrorSeverity = RulesetErrorSeverity.Error): RulesetErrorList {
```
yairm210, line 105: Putting the only required parameter as the third parameter instead of the first feels like an antipattern

core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt
```diff
@@ -60,4 +91,21 @@ class RulesetErrorList : ArrayList<RulesetError>() {
                         // out of place. Prevent via kludge:
                         it.text.replace('<','〈').replace('>','〉')
                 }
+
+    companion object {
+        private val empty = RulesetErrorList()
+
+        fun of(ruleset: Ruleset? = null, sourceObject: IHasUniques? = null, vararg errors: RulesetError): RulesetErrorList {
+            if (errors.isEmpty()) return empty
```
yairm210, line 99: This is a *Mutable* instance, meaning if someone adds to it, from then on `of()` will no longer be empty. Definite future bug.

core/src/com/unciv/models/ruleset/validation/RulesetErrorList.kt
```diff
@@ -11,30 +13,59 @@ enum class RulesetErrorSeverity(val color: Color) {
...
+        return addWithDuplicateCheck(element)
     }
 
-    fun add(text: String, errorSeverityToReport: RulesetErrorSeverity) {
-        add(RulesetError(text, errorSeverityToReport))
+    /** Shortcut: Add a new [RulesetError] built from [text] and [errorSeverityToReport].
+     *
+     *  [sourceObject] is for future use and should be the originating object. When it is not known or not a [IHasUniques], pass `null`.
+     */
+    fun add(sourceObject: IHasUniques?, text: String, errorSeverityToReport: RulesetErrorSeverity = RulesetErrorSeverity.Error) =
```
yairm210, line 43: Here too - the text should definitely be first

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -109,7 +109,6 @@ object Conditionals {
         return when (condition.type) {
             // These are 'what to do' and not 'when to do' conditionals
             UniqueType.ConditionalTimedUnique -> true
```
yairm210, line 111: This reveals that also ConditionalTimedUnique should be a metamodifier!

yairm210 (approved):

PR #11133, opened 2024-02-14
Fix translation placeholder mismatches failing unit tests

SomeTroglodyte:
Just an offer in case it saves work...


yairm210:
YES! Thank you!

Issue #10908, opened 2024-01-09
Weird crash when creating Huge vertical map


Caballero-Arepa:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Game Version

4.9.16

### Describe the bug

It hasn't happened ever before, and I'm on PC which is even weirder, with which I've generated maps larger than 57x88.

### Steps to Reproduce

I go to the Map Editor and want to make a new map, a huge map, but vertical instead of horizontal.
I click it, and it crashes, but only with that size or above. I can get perfectly fine a Huge map from the pre-set sizes, but when I use the Huge dimensions with the Custom size option, it crashes.

### Link to save file

<details><summary>Details</summary>
<p>

**Platform:** Desktop
**Version:** 4.9.16 (Build 952)
**Rulesets:** [New Building Icons, Policy mod, RekMOD, Iron and Blood, Vaalmaseda RP game, Belle Cartographie, 3rd and 4th Unique Component, Latam Scenarios, UncivRpGame, New minimal, Latin American_Civs, Star Wars for Alpha Frontier, Community Civs, Inverted World, Difficulty_Sandbox, More Fauna And Flora, Civ V - Vanilla, FantasyHex UI, Countryball Units, Community Civs Extra, Civ6 Tileset, Alpha Frontier, Naciones pa Unciv ES, Tactical unit set, Civ V - Gods & Kings, RyteakoRpGame]
**Last Screen:** `com.unciv.ui.screens.mapeditorscreen.MapEditorScreen`

--------------------------------

OS: Windows 10
Java: Oracle Corporation 25.232-b09
	Max Memory: 910 MB
System default encoding: windows-1252

--------------------------------

**Message:**
```
java.lang.IllegalStateException: TileMap.setTransients called on existing tileMatrix of different size
	at com.unciv.logic.map.TileMap.setTransients(TileMap.kt:469)
	at com.unciv.ui.screens.mapeditorscreen.MapEditorScreen.newMapHolder(MapEditorScreen.kt:175)
	at com.unciv.ui.screens.mapeditorscreen.MapEditorScreen.loadMap(MapEditorScreen.kt:216)
	at com.unciv.ui.screens.mapeditorscreen.tabs.MapEditorGenerateTab.generate$freshMapCompleted(MapEditorGenerateTab.kt:94)
[...]

PR #11144, opened 2024-02-16
Fix creating 105x90 rectangular no-wrap maps

SomeTroglodyte:
seen [on disc :vomiting_face: rd](https://discord.com/channels/586194543280390151/1195754723542835282).
Also reported earlier by at least @Caballero-Arepa - issue ... hmmm .... #10908 this one. I'm sure it was the same underlying problem.

And no, no increased presence of me on d*. I just reacted to an insulting mail by Dirtcord itself.

After all, repro turned out easy: Reset to defaults, rect, custom, 105x90, that check trips, with leftX a -71 and a recalculated rightX = +70, so the check looks for 142 in the 143..145 range. Or looked after this PR.

### DIFF
--- a/core/src/com/unciv/logic/map/TileMap.kt
+++ b/core/src/com/unciv/logic/map/TileMap.kt
@@ -465,8 +465,8 @@ class TileMap(initialCapacity: Int = 10) : IsPartOfGameInfoSerialization {
             }
         } else {
             // Yes the map generator calls this repeatedly, and we don't want to end up with an oversized tileMatrix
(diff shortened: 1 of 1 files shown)


Caballero-Arepa:
Oh! wow. For me, the problem went away by itself. Thankfully it was an easy fix 😉

> seen on disc 🤮 rd

> And no, no increased presence of me on d*. I just reacted to an insulting mail by Dirtcord itself.

XDD why do you hate discord so much? I think it alright, and way better than Whats🤮p

yairm210 (approved):

PR #11148, opened 2024-02-17
Fix ships unable to enter single-tile lakes via a city

SomeTroglodyte:
Fix [ships unable to enter single-tile lakes via a city](https://discord.com/channels/586194543280390151/1183053078220513312), the brute-force way.

@tuvus - some more elegant solution? Is that shortcut even worth it anymore if the "but cities are passable if they're mine" test complicates it? Any thoughts already invested in the "tiles should be able to allow or disallow more than one domain" idea? There's still modders wishing for canals and bridges...


SeventhM:
Rule of thumb: links to discord channels like this tend to feel out of place outside of discord. Don't recommend it

tuvus:
Why not a call to canPassThrough()? It seems like you are just copying a very small piece of it. I would also account for other conditions as well.

tuvus:
I though I read somewhere that cities on hills should take only one movement to move into. Even with land units. I can't seem to find any issues though or on discord.

PR #11152, opened 2024-02-18
Terraforming!

yairm210:
The long *long* awaited feature
Tested by adding 
 "Create a [Forest] on this tile <for [0] movement>", "Turn this tile into a [Ocean] <for [0] movement>"


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -755,6 +755,9 @@ enum class UniqueType(
     SkipPromotion("Doing so will consume this opportunity to choose a Promotion", UniqueTarget.Promotion),
     FreePromotion("This Promotion is free", UniqueTarget.Promotion),
 
+    OneTimeCreateTerrainFeatureImprovement("Create a [terrainFeature/improvementName] on this tile", UniqueTarget.UnitTriggerable),
```
SeventhM, line 758: Makes `ConstructImprovementInstantly` redundant? Tbh, I think some of the unit actions should probably become triggers anyways. Probably should deprecate?

yairm210: To do that, we'll need to add a way to limit this construction according to the improvement's 'can be constructed on' rules

SeventhM: To be fair, I'm not sure this unique implies it should ignore limitations

SeventhM:
Why are we adding a new parameter in the first place and not using TerrainName? Especially if we want to support adding in nat wonders. I kinda feel like it should either be one or the other: Either allow Nat Wonders or just go the full distance and allowing adding terrain wholesale. You can already change the base terrain with some of the natural wonders anyways

yairm210:
Good point regarding cities!

PR #11155, opened 2024-02-19
External links: Right-click and some housecleaning

### LINKED ISSUE #11129: Feature request: Give the ability to 'copy GitHub' URL instead of 'Open Github page'
### Before creating

### Problem Description

Right now, it's opening the wrong browser.
It would be nice if it copies the URL instead, or small text box that selects the entire text to copy.

### Related Issue Links

_No response_

### Desired Solution

Copy button

### Alternative Approaches

Text box of URL that auto selects all text of the URL

### Additional Context

_No response_

SomeTroglodyte:
Closes #11129

- Right-click _or_ longpress works, as usual
- The mod manager "go to repo" button confirms with a toast. The support in MarkupRenderer covers more cases, but can't toast - it's Stage-agnostic
- Additionally a short hint in UI tips
- Additionally a wiki link on the About tab
- Placing the base url's in Constants - ugh, but Github or GithubAPI don't fit well either. Central is good not only in the unlikely event of a move, but also to _find_ the uses.

### DIFF
--- a/core/src/com/unciv/Constants.kt
+++ b/core/src/com/unciv/Constants.kt
@@ -106,4 +106,10 @@ object Constants {
 
     const val defaultFontSize = 18
     const val headingFontSize = 24
+
+    /** URL to the root of the Unciv repository, including trailing slash */
+    // Note: Should the project move, this covers external links, but not comments e.g. mentioning issues
+    const val uncivRepoURL = "https://github.com/yairm210/Unciv/"
+    /** URL to the wiki, including trailing slash */
+    const val wikiURL = "https://yairm210.github.io/Unciv/"
 }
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/ui/screens/civilopediascreen/MarkupRenderer.kt
```diff
@@ -51,10 +53,14 @@ object MarkupRenderer {
                 actor.onClick {
                     linkAction(line.link)
                 }
-            else if (line.linkType == FormattedLine.LinkType.External)
+            else if (line.linkType == FormattedLine.LinkType.External) {
                 actor.onClick {
                     Gdx.net.openURI(line.link)
                 }
+                actor.onRightClick {
+                    Gdx.app.clipboard.contents = line.link
```
yairm210, line 61: What about a popup here?

yairm210 (approved):

Issue #11125, opened 2024-02-11
Feature request: UI -- escape button is either 1) 'X' or 2) 'Close' and it can be anywhere on the screen


PathosEthosLogos:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

To address 'before creating' checklist, I don't know if they're from these games because I don't have them and I never played them.

The 'escape' button is inconsistent in multiple ways. Here are few patterns that I found:

1) 'Close' or 'X'  
2) Button placement is bottom-left, top-right, etc. -- everywhere
3) Most minor: colour scheme for these buttons should be more uniform

### Desired Solution

I feel that there should be a discussion on UI philosophy, apply the changes in the code, then have the result of the discussion written down as bullet points in README or similar for contributors to follow in the future. I also feel that the UI/UX for desktop play is wee bit clumsy due to keeping UI/UX functions for mobile use. I feel this is a different (and bigger) topic though.

### Alternative Approaches

Just do the work on coding it, send pull request, then get rejected :sad

Issue #4551, opened 2021-07-17
Align positioning of Close Buttons


mnbpdx:
Hey there folks! Love the app, hoping to contribute, looking for a good first thing to work on.

### Is your feature request related to a problem? Please describe.

I've noticed that the position of the `Close` buttons on various screens isn't consistent. 

- On the Technology and Policy screens, the `Close` button is in the **bottom left**.
- On the Overview screen, its in the **top right** inside the scrolling table
- On the Diplomacy screen, its in the **top left**.

This is totally not a big deal, but seemed like it might be an easy first issue for me to tackle, if folks were interested in rearranging those.

### Describe the solution you'd like

Right now, it seems like the obvious choice would be moving all of those to the **bottom left**, to align with the Technology and Policy screens.

Not sure how many screens this would affect, or if it would be necessary/wanted at all. Let me know what y'all think.

### Additional context

Here's the current state of the Tech screen, which I'm proposing every other screen align to:
<img width="1792" alt="Untitled" src="https://user-images.githubusercontent.com/19321892/126046600-f62d2de2-faf7-43be-a833-7e572c27895f.png">

yairm210:
Moving this in the diplomacy screen shouldn't be a problem, but in the overview (and civilopedia) the buttons are in a different context, of "choosing a tab". Moving only the button doesn't make much sense, and I'm not sure about moving the entire team picker, I feel it's more intuitive for the tab picker to be in the top than on the bottom.
@SomeTroglodyte thoughts on tab picker location?

SomeTroglodyte:
I've noticed the discrepancy, too, long ago. Maybe a motivation to push keyboard support - always `Escape`... :wink:  Still, I haven't yet gotten either a clever intuition or taken the time to properly think about it. One argument would be - desktop users would have a bias towards top right from other apps, 'droid apps I can compare tend to have a comparable arrow-left in the top left if at all...

Also to consider: Special case "floating big X" used by Player picker; Close in the "tab switchers" is already a special case 'cuz it doesn't scroll with the rest (if I'm not mistaken),... Also, the bottom right corner seems to me to be the most unbalanced corner, as it widely diverges from being mostly dead space on many screens to being the center of attention... No idea how to harmonize all that. Maybe no close button anywhere? Desktop = ESC / Droid = builtin Back? Nah, would block the road to ios...

Maybe lower left everywhere is the cheapest option, and make it a floating button on diplo pedia and overview alike?

alextfish:
[...]

PR #11169, opened 2024-02-21
Change EmpireOverviewScreen closing to use the same button and UX as Civilopedia

SomeTroglodyte:
This reduces the number of different "close" UX to - three???[^1] Top right "X", bottom left "Close", and Popups bottom center. Getting to two would already be enough - standard desktop patterns.

There was an issue somewhere - or two... #11125, #4551. This is not a comprehensive solution - but Overview is an easy small step. PickerScreens is harder to envision - a floating X top right will surely cover important stuff in some of them...

Centralizing the `getCloseButton` factory as the first commit does - I'm not convinced that was the best decision. One - it's not an extension but in a file named extensions. Two - the uses are varied enough that this much parametrization was needed. Three - the NationPicker and GameOptions versions were already using local dedicated factories.. Votes?

Note: Extending the TabbedPager API like done here - I originally had it symmetic, allowing a nonscrolling extra on the left too, but that broke Table layout. Couldn't find the cause on short notice - but TabbedPager also misbehaves in cramped portrait display conditions, so it needs an analysis somtime anyway.

[^1]: I believe that's the last of the "top left" ones??? Or am I missing one?

### DIFF
--- a/core/src/com/unciv/ui/components/widgets/TabbedPager.kt
+++ b/core/src/com/unciv/ui/components/widgets/TabbedPager.kt
@@ -86,6 +86,7 @@ open class TabbedPager(
     private val header = Table(BaseScreen.skin)
     val headerScroll = LinkedScrollPane(horizontalOnly = true, header)
     protected var headerHeight = 0f
+    private val headerDecorationRightCell: Cell<Actor?>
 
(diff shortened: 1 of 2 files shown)


yairm210:
A UI PR needs a pic

yairm210:
Code looks fine

Issue #11167, opened 2024-02-21
Effect "<for [amount] turns> <upon being defeated>" not working


SpacedOutChicken:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.10.11

### Describe the bug

I attempted to give a unit a temporary civ-wide effect that kicks in when the unit gets defeated: `"[+100]% Strength <for [Military] units> <for [50] turns> <upon being defeated>"` It isn't working. After the unit is defeated, the strength boost fails to take effect.

### Steps to Reproduce

This mod contains a simple unit that has the specified effect. 
[TestUponDefeat.zip](https://github.com/yairm210/Unciv/files/14353690/TestUponDefeat.zip)

### Operating System

Windows

SpacedOutChicken:
In that case, is there a functional way of doing what I'm trying to do? I'd love to be able to kick-start other temporary effects by expending units. Imagine something like `"[+2 Science] [in all cities] <for [10] turns> <by consuming this unit>"`. Can that be done with existing uniques? If not, I'll make it a modding request.

SpacedOutChicken:
Sorry for not including a saved game, but the setup is easy enough. You build the Sacrificial Lamb unit, which has a strength of 1 and the specified unique, and then you send the Sacrificial Lamb out to get killed by a Barbarian unit. I can attach a saved game later if you need one.

SpacedOutChicken:
> Aside: If we get that working it will be quite probably cumulative - lose 20 units in one turn -> have +2000% bonus.
Sounds like fun to me. It's not likely to come up in what I'm working on, but it would be interesting to have an OP faction that just gets stronger and stronger as wars progress...

SpacedOutChicken:
I swear that I'm not proposing these things to be difficult. I'm just trying to come up with something cool.

SeventhM:
I would probably need to read this further to fully understand what Trog is saying (kinda busy rn), but something tells me Trog is over complicating things
`<for [amount] turns>` currently only adds a unique to the Civ's global uniques. If the unique for the unit doesn't check global uniques, it won't seem to work. Maybe the solution is to allow someway to allow units themselves to have temporary uniques? Sounds complicated but potentially doable, would need to check later and see if it creates any weirdness regarding assuming where a unique comes from

SeventhM:
> Imagine something like `"[+2 Science] [in all cities] <for [10] turns> <by consuming this unit>"`

I believe this already works. Would need to double check

[...]

PR #11188, opened 2024-02-23
Unit actions from Uniques shouldn't be active at zero movement

SomeTroglodyte:
Consider carefully...
* Once #11181 is merged, unit uniques such as the _other_ example[^1] in #11167 will work (though still be defamed by mod checker) - even for freshly bought units, or units that have exhausted their movement this turn. In contrast, Great People don't have that luxury.
* So I propose _any_ action supplied through getTriggerUniqueActions should "grey out" and show as "not available right now" when currentMovement==0

...

* Thus this approach - doing the test earlier, e.g. in canUse, would hide the action button entirely.
* Thought about looking for a "for [x] movement" modifier and testing against that x - decided against. Gut feeling.
* I looked at that getTriggerFunction that is just compared to null then discarded and evaluated _again_ shortly after in triggerUnique - can't imagine the context between those two points in time changing, so not doing the work twice seemed a good thing. You might think otherwise - throw of the dice to me. One more reference passed as closure into that action lambda than before, cheap enough.
* Same with using a temp one-shot anonymous factory function just for its "shout-out" ability. I find it clear enough, but others may not. Can comment or unwrap the hard way.

[^1]: `"[+2 Science] [in all cities] <for [10] turns> <by consuming this unit>"`

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsFromUniques.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsFromUniques.kt
@@ -199,12 +199,15 @@ object UnitActionsFromUniques {
             }
             val title = UnitActionModifiers.actionTextWithSideEffects(baseTitle, unique, unit)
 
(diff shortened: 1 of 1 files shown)


yairm210:
Makes sense to me

PR #11190, opened 2024-02-24
New automated translations for czech language

jpodivin:
I've run the untranslated texts through [facebook/nllb-200-distilled-600M](https://huggingface.co/facebook/nllb-200-distilled-600M) model, keeping quotes and wonder descriptions untranslated, because there was a higher probability of things going wrong. I've also skipped strings with variables, for the same reason.

Still, there are cases when the model just couldn't get the terminology right. While the translations are correct in principle, they may seem odd to the native speaker. I know, because I'm one of them.

I've used following script to produce translations, with some simple adjustments it can be used for other languages.

```

from transformers import pipeline

SOURCE_PTH = './android/assets/jsons/translations/Czech.properties'

TARGET_PTH = SOURCE_PTH + '.new'

QUOTE_DELIM = "#################### Lines from Buildings from Civ V - Vanilla ####################"
translator = pipeline('translation', model="facebook/nllb-200-distilled-600M")


def strip_special(line):

    return line


def translate(line):
    new_line = translator(
        strip_special(line),
        src_lang='eng_Latn', tgt_lang='ces_Latn')[0]['translation_text']
    return new_line


def has_var(line):
    return '[' in line


translate_next = False

with open(SOURCE_PTH) as f_in:
    while True:
        line = f_in.readline()
        if not line:
            break
        split = line.split('=')
        text = split[0]
        if 'Requires translation' in line:
            translate_next = True  # Translate next line
            continue
        elif QUOTE_DELIM in line:  # Stop translationn
            break
        elif translate_next:
            translate_next = False
            if line != '\n' and not line.startswith('#') and not split[1].strip() and not has_var(line):
                translation = translate(text)
                print(f"Translated '{text}' to '{translation}'!")
                line = f"{text} = {translation}\n"
            else:
                with open(TARGET_PTH, mode='a') as f_out:
                    f_out.write(" # Requires translation!\n")

        with open(TARGET_PTH, mode='a') as f_out:
            f_out.write(line)
    with open(TARGET_PTH, mode='a') as f_out:
        f_out.write(line)
        f_out.write(f_in.read())

```

### DIFF
--- a/android/assets/jsons/translations/Czech.properties
+++ b/android/assets/jsons/translations/Czech.properties
@@ -121,8 +121,7 @@ Declare war = Vyhlásit válku
 Declare war on [civName]? = Vyhlásit válku civilizaci [civName]?
  # Requires translation!
 [civName] will also join them in the war = 
- # Requires translation!
-An unknown civilization will also join them in the war = 
+An unknown civilization will also join them in the war  = V válce se k nim připojí i neznámá civilizace .
  # Requires translation!
 This will cancel your defensive pact with [civName] = 
 Go to on map = Ukázat na mapě
@@ -271,8 +270,7 @@ Sum: = Součet:
 
 Trade = Obchodovat
 Offer trade = Nabídnout
- # Requires translation!
-They'll decide on their turn = 
+They'll decide on their turn  = Oni se rozhodnou na vlastní řadu.
 Retract offer = Zrušit nabídku
 What do you have in mind? = Co máte na mysli?
 Our items = Naše obchodní artikly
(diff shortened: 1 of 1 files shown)


android/assets/jsons/translations/Czech.properties
```diff
@@ -3083,169 +2771,94 @@ Our mutual military struggle brings us closer together. = Naše společné vále
...
+Music Player  = Hrací nástroj 
+Developer Console  = Konzola vývojářů 
+Empire Overview Trades  = Empire Overview Trades 
+Empire Overview Units  = Empire Overview Units 
+Empire Overview Politics  = Empire Overview Politika 
+Social Policies  = Sociální politiky 
+Technology Tree  = Technologické stromky 
+Empire Overview Notifications  = Empire Overview oznámení 
+Empire Overview Stats  = Empire Overview Stats 
+Empire Overview Resources  = Empire Overview Resources
```
yairm210, line 2808: These are straight-up untranslated, and not the only ones

yairm210:
It's fine to use this as the *base* for translations, but these do require manual approval by a native speaker. If you are one, please go through and validate the translations

PR #11192, opened 2024-02-25
Better military unit retreat

tuvus:
This PR changes the functionality and name of trySwapRetreat to tryRetreat. 

TryRetreat uses `ThreatManager.getDistanceToClosestEnemy()` to find which tiles are suitable to retreat to. Previously trySwapRetreat only checked tiles with units to swap to. Now tryRetreat checks tiles that don't have military units on them as well.

Checking the `distanceToClosestEnemy()` on a tile is expensive so it tries to sort the tiles by a cheaper `getArielDistanceTo(closestCity)` first. If there is no close city, then it will try a cheaper getDistanceToClosestEnemy to filter out the tiles that are very close to an enemy.

Added a calvary unit tactic: If a unit has successfully attacked and is at low health afterward it will try to spend the remaining movement points to run away.

Overall, the performance impact should be relatively light compared to the AI decision improvement.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -213,7 +213,7 @@ object UnitAutomation {
 
         if (tryUpgradeUnit(unit)) return
 
-        if (unit.health < 50 && (trySwapRetreat(unit) || tryHealUnit(unit))) return // do nothing but heal
+        if (unit.health < 50 && (tryRetreat(unit) || tryHealUnit(unit))) return // do nothing but heal
 
         // If there are no enemies nearby and we can heal here, wait until we are at full health
         if (unit.health < 100 && canUnitHealInTurnsOnCurrentTile(unit,2, 4)) return
@@ -262,6 +262,8 @@ object UnitAutomation {
     private fun tryAttacking(unit: MapUnit): Boolean {
         repeat(unit.maxAttacksPerTurn() - unit.attacksThisTurn) {
             if (BattleHelper.tryAttackNearbyEnemy(unit)) return true
+            // Calvary style tctic, attack and then retreat
+            if (unit.health < 50 && tryRetreat(unit)) return true
         }
         return false
     }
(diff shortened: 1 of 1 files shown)


yairm210:
Haven't forgotten, waiting for a time I'll have enough headspace to properly review 😓

tuvus:
No problem!

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -276,34 +278,56 @@ object UnitAutomation {
...
+
+            if (retreatTile.militaryUnit == null) {
+                // See if we can retreat to the tile
+                if (unit.movement.canMoveTo(retreatTile)) {
+                    unit.movement.moveToTile(retreatTile)
                     return true
                 }
+            } else if (retreatTile.militaryUnit!!.civ == unit.civ) {
+                // The tile is taken, lets see if we want to swap retreat to it
+                val otherUnit = retreatTile.militaryUnit!!
```
yairm210, line 318: This can be moved above the if/else (without the !! at the end) so we can check nullability on it and use if not null

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -276,34 +278,56 @@ object UnitAutomation {
...
+            unitDistanceToTiles.asSequence().map { it.key }.sortedBy { it.aerialDistanceTo(closestCity.getCenterTile()) }
+        } else {
+            // Rare case, what if we don't have a city nearby?
+            // Lets favor the tiles that don't have enemies close by
+            // Ideally we should check in a greater radius but might get way too expensive
+            unitDistanceToTiles.asSequence().map { it.key }.sortedByDescending { unit.civ.threatManager.getDistanceToClosestEnemyUnit(it, 3, false) }
+        }
+
+        // Lets check all tiles and swap with the first one
+        for (retreatTile in sortedTilesToRetreatTo) {
```
yairm210, line 305: Line 306 can be lifted out of the for loop

yairm210:
What is up with the files changed?! I think you need to merge master or something, that is very strange

tuvus:
I don't think that worked, I will try to reset to an older version and merge again.

tuvus:
Hmm, I might have to force push in this case. Otherwise there is no way for me to rewrite the history on github to solve the problem.

tuvus:
It should be fixed now with the force push.

yairm210 (approved):

PR #11195, opened 2024-02-25
Military unit healing improvement

tuvus:
This PR works on improving the AI military unit's healing decisions.
It does have a merge conflict with #11192, so I will have to handle that once one of these gets merged.

Here is a list of the changes
- Units will prioritize healing over accompanying a great person or heading to a city under siege. If the unit is damaged, they aren't going to be too helpful anyway.
- Military units try to pillage while attacking. The implementation is a little funky. They retroactively pillage tiles that they moved through before attacking.
- UnitAutomation.tryHeal() now tries to pillage multiple tiles if able to heal. It will only try to pillage non-road tiles as well. I think this means that the civ AI will not pillage roads at all now.
- Barbarians now pillage more
- Units will stay on their tile to heal instead of moving to a different tile if they can heal safely to full in 2 turns

There is no mentionable performance impact (tryHeal() could actually be faster on average)

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -212,12 +212,12 @@ object UnitAutomation {
 
         if (tryUpgradeUnit(unit)) return
 
+        if (unit.health < 50 && (trySwapRetreat(unit) || tryHealUnit(unit))) return // do nothing but heal
+
         // Accompany settlers
         if (tryAccompanySettlerOrGreatPerson(unit)) return
 
         if (tryHeadTowardsOurSiegedCity(unit)) return
-
-        if (unit.health < 50 && (trySwapRetreat(unit) || tryHealUnit(unit))) return // do nothing but heal
 
         // if a embarked melee unit can land and attack next turn, do not attack from water.
         if (BattleHelper.tryDisembarkUnitToAttackPosition(unit)) return
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -318,7 +318,8 @@ object UnitAutomation {
 
         // Try pillage improvements until healed
         while(tryPillageImprovement(unit, false)) {
-            if (unit.currentMovement == 0f || unit.health == 100) return true
+            // If we are fully healed and can still do things, lets keep on going by returning false
+            if (unit.currentMovement == 0f || unit.health == 100) return unit.currentMovement == 0f
         }
 
         val unitDistanceToTiles = unit.movement.getDistanceToTiles()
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -313,10 +316,17 @@ object UnitAutomation {
         if (unit.baseUnit.isRanged() && unit.hasUnique(UniqueType.HealsEvenAfterAction))
             return false // will heal anyway, and attacks don't hurt
 
-        if (tryPillageImprovement(unit)) return true
+        // Try pillage improvements until healed
+        while(tryPillageImprovement(unit, false)) {
+            if (unit.currentMovement == 0f || unit.health == 100) return true
```
yairm210, line 321: If I managed to heal to 100 and I still have movement, why should I return? I should keep going!

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -74,6 +76,21 @@ object Battle {
             attacker.unit.action = UnitActionType.SetUp.value
             attacker.unit.useMovementPoints(1f)
         }
+
+        if (tryHealPillage) {
+            // Now lets retroactively see if we can pillage any improvement on the path improvement to heal
+            // While still being able to attack
+            val pillageIterator = tilesMovedThrough.toMutableList().listIterator()
```
yairm210, line 83: There is *really* no need for an iterator here, works fine for 'just a list' and less confusing that way (list.first(), list.remove())

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -365,6 +375,17 @@ object UnitAutomation {
...
 
+    /**
+     * @return true if the tile is safe and the unit can heal to full within [turns]
+     */
+    private fun canUnitHealInTurnsOnCurrentTile(unit: MapUnit, turns: Int, noEnemyDistance: Int = 3): Boolean {
+        // Check if we are not in a safe city and there is an enemy nearby this isn't a good tile to heal on
+        if (!(unit.getTile().isCityCenter() && unit.getTile().getCity()!!.health > 50) 
+            && unit.civ.threatManager.getDistanceToClosestEnemyUnit(unit.getTile(), noEnemyDistance) <= noEnemyDistance) return false
+        return (!unit.hasUnique(UniqueType.HealsEvenAfterAction)
+            && (100 - unit.health) / turns <= unit.rankTileForHealing(unit.getTile()))
```
yairm210, line 386:
```suggestion
        if (unit.hasUnique(UniqueType.HealsEvenAfterAction) return false
        val healthRequiredPerTurn = (100 - unit.health) / turns
        return healthRequiredPerTurn <= unit.rankTileForHealing(unit.getTile())
```

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
tuvus: Good point! That definitely makes sense.

PR #11213, opened 2024-02-27
Add MovedToNewCapital buiding unique

PLynx01:
This building-specific unique indicates that building will be relocated to the new capital, when the capital is moved to new city.

I think it will be useful for buildings that must be coupled with the capital city, especially the abstract buildings like edicts, modifiers, laws or country missions (like those from Europa Universalis IV)

This code is untested, so I don't know whether it works or not. If you have any objections, tell me in the comments. Constructive feedback welcome.

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -865,6 +865,22 @@ class Civilization : IsPartOfGameInfoSerialization {
             // move new capital
             city.cityConstructions.addBuilding(city.capitalCityIndicator())
             city.isBeingRazed = false // stop razing the new capital if it was being razed
+
+            // move the buildings with MovedToNewCapital unique
+            if (oldCapital != null) {
+                // Get the Sequence of the buildings to move
+                val buildingsToMove = oldCapital.cityConstructions.getBuiltBuildings().filter {
+                    it.hasUnique(UniqueType.MovedToNewCapital)
+                }
+
+                // Remove the buildings from old capital
+                oldCapital.cityConstructions.removeBuildings(buildingsToMove.toSet())
+
+                // Add the buildings to new capital
+                buildingsToMove.forEach {
+                    city.cityConstructions.addBuilding(it)
+                }
+            }
         }
         oldCapital?.cityConstructions?.removeBuilding(oldCapital.capitalCityIndicator())
     }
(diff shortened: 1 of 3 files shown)


PLynx01:
> Umm, yes, won't work, from a short look without testing
> 
> <details><summary>spoilers</summary>
> 
> <sub><sup>What, you thought I'd give the suspense away so easily????</sup></sub>
>  :smiling_imp: :rofl: :smiling_imp: 
> </details>

Really? Let @yairm210 see this PR. I am waiting for his remarks.

core/src/com/unciv/logic/civilization/Civilization.kt
```diff
@@ -865,6 +865,22 @@ class Civilization : IsPartOfGameInfoSerialization {
             // move new capital
             city.cityConstructions.addBuilding(city.capitalCityIndicator())
             city.isBeingRazed = false // stop razing the new capital if it was being razed
+
+            // move the buildings with MovedToNewCapital unique
+            if (oldCapital != null) {
+                // Get the Sequence of the buildings to move
+                val buildingsToMove = oldCapital.cityConstructions.getBuiltBuildings().filter {
+                    it.hasUnique(UniqueType.MovedToNewCapital)
+                }
```
yairm210, line 874: The .toSet() needs to be *here* - otherwise, once we come to "remove the buildings", we evaluate this a second time and see that hey, there are no such buildings in the old capital, look at that!

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -305,6 +305,7 @@ enum class UniqueType(
     Unsellable("Unsellable", UniqueTarget.Building),
     ObsoleteWith("Obsolete with [tech]", UniqueTarget.Building, UniqueTarget.Resource, UniqueTarget.Improvement),
     IndicatesCapital("Indicates the capital city", UniqueTarget.Building),
+    MovedToNewCapital("Moved to the new capital", UniqueTarget.Building),
```
yairm210, line 308: Unclear phrasing - "Moved" can be passive, or past-tense. Better to use "Moves", and under what condition - so all in all "Moves to new capital when capital changes".

core/src/com/unciv/logic/civilization/Civilization.kt
yairm210: Oh I see SomeTrog already mentioned this

yairm210 (approved):

PR #11230, opened 2024-03-01
Fix tileset sanity check for builtin rulesets when run from jar

### LINKED ISSUE #11228: Gods and Kings giving an error in the mod checker

### Game Version

4.10.14

### Describe the bug

The mod checker has started putting out this error in the checker:

> The Mod tileset config folder contains no json files
> Mod has no configuration for tileset graphics: FantasyHex, Minimal, HexaRealm

I'm not sure what it's about, but it's got me worried.

### Steps to Reproduce

Checking the mods will expose this issue.

### Screenshots

_No response_

### Link to save file

_No response_

### Operating System

Windows

### Additional Information

_No response_

SomeTroglodyte:
Fixes #11228, the "large" way.

Just bailing from that `if (ruleset.folderLocation == null && this::class.java.package?.specificationVersion != null)` would be fine too - if we trust the released builtin ruleset must have passed the full test in the release workflow... Whaddayouthink?

### DIFF
--- a/core/src/com/unciv/models/ruleset/validation/RulesetValidator.kt
+++ b/core/src/com/unciv/models/ruleset/validation/RulesetValidator.kt
@@ -1,6 +1,7 @@
 package com.unciv.models.ruleset.validation
 
 import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.files.FileHandle
 import com.badlogic.gdx.graphics.Color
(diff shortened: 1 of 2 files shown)


yairm210:
A. looks like a lot of duplication?
B. What is the difference between the two? That the JAR only checks mod tilesets and not base ruleset ones? If so that's a 👍🏿  from me, but it should be clearer from the function names imo

PR #11231, opened 2024-03-02
Fix conditionals and triggers being valid for cities on non city tiles

SeventhM:
Another "Idk what the title should be". Also, seems weird that `getCity` doesn't exactly match what it's intended to do (seems like it should be "getOwningCity"), but the amount of places where this mistake was made seems to be low if at all

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -33,8 +33,9 @@ object Conditionals {
         }
 
         val relevantCity by lazy {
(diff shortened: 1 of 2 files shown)


yairm210:
This was actually intentional! Result of a long discussion on Discord, basically we decided that city conditionals should apply on all tiles belonging to that city, limiting to city center can be done via the fact that there's a city center improvement there

SeventhM:
> This was actually intentional!

Hmm.... I... guess... Seems non obvious, though. This definitely should have proper documentation imo to relay this information

PR #11233, opened 2024-03-02
Fix Civ resources being counted multiple times

SeventhM:
Note: No matter what, either you're checking the unique's sources and trying to find what city it came from that way or you're checking each city twice for building uniques. Unless you can find a way past one of those two scenarios, there's no real valuing in continuing over uniques here that claim to be from Buildings or Wonders

Also......... Wtf, why is Civ based city resources counted from coming from buildings. That's makes it very difficult to understand why it's breaking. That alone is a reason to remove the continue in CivInfoTransients. Even if it didn't add a new entry to the civ for every resource, we would be better off trying to find a way to give uniques a better trail to find where it came from than relying on that as a hack

### DIFF
--- a/core/src/com/unciv/logic/city/CityResources.kt
+++ b/core/src/com/unciv/logic/city/CityResources.kt
@@ -82,11 +82,11 @@ object CityResources {
     }
 
     private fun getCityResourcesFromCiv(city: City, cityResources: ResourceSupplyList, resourceModifer: HashMap<String, Float>) {
-        // This includes the uniques from buildings, from this and all other cities
-        for (unique in city.civ.getMatchingUniques(UniqueType.ProvidesResources, StateForConditionals(city.civ, city))) { // E.G "Provides [1] [Iron]"
+        val stateForConditionals = StateForConditionals(city)
+        for (unique in city.getMatchingUniques(UniqueType.ProvidesResources, stateForConditionals)) { // E.G "Provides [1] [Iron]"
             val resource = city.getRuleset().tileResources[unique.params[1]]
                 ?: continue
-            if (!resource.hasUnique(UniqueType.CityResource, StateForConditionals(city))) continue
+            if (!resource.hasUnique(UniqueType.CityResource, stateForConditionals)) continue
             cityResources.add(
                 resource, "Buildings",
                 (unique.params[0].toFloat() * resourceModifer[resource.name]!!).toInt()
--- a/core/src/com/unciv/logic/city/CityResources.kt
+++ b/core/src/com/unciv/logic/city/CityResources.kt
@@ -88,7 +88,7 @@ object CityResources {
                 ?: continue
             if (!resource.hasUnique(UniqueType.CityResource, stateForConditionals)) continue
             cityResources.add(
-                resource, "Buildings",
+                resource, unique.getSourceNameForUser(),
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/city/CityResources.kt
```diff
@@ -82,12 +82,13 @@ object CityResources {
...
-        // This includes the uniques from buildings, from this and all other cities
-        for (unique in city.civ.getMatchingUniques(UniqueType.ProvidesResources, StateForConditionals(city.civ, city))) { // E.G "Provides [1] [Iron]"
+        val stateForConditionals = StateForConditionals(city)
+        for (unique in city.getMatchingUniques(UniqueType.ProvidesResources, stateForConditionals)) { // E.G "Provides [1] [Iron]"
             val resource = city.getRuleset().tileResources[unique.params[1]]
                 ?: continue
+            if (!resource.hasUnique(UniqueType.CityResource, stateForConditionals)) continue
             cityResources.add(
-                resource, "Buildings",
+                resource, unique.getSourceNameForUser(),
```
yairm210, line 91: WELP
That's 100% my bad

yairm210:
I think I understand the root cause of this problem.
getCityResources is a bad name.

It can have 2 meanings:
- The resources *provided by* this city (as used by updateCivResources)
- The resources *available to* this city (as used by getResourceAmount)

Along the way this shifted from B to A
So the correct solution would be to have 2 functions - getResourcesProvidedByCity, and getCityResourcesAvailableToCity.

The former would be the current function minus the 'get from civ', and the latter would be just a call to the former plus the 'getCityResourcesFromCiv'

yairm210:
It's not as simple as I thought... If we want to get building uniques to register in other cities, but not register twice in the city they originate from, we need to play around with things, it's a bit of a pain

yairm210:
My take in #11236

SeventhM:
> So the correct solution would be to have 2 functions - getResourcesProvidedByCity, and getCityResourcesAvailableToCity.

Hmmm... I guess. Here's my full take on the actual issues at hand, maybe I could be corrected on some of them
1. There's a question of who's responsible for reporting Civ resources. Currently (before either PR), each city is responsible for reporting their resources. Obviously, the implementation is a bit off, but this suggestion would fix it. My main issue (which is most notable in the unit test) is just... I'm not sure where this is useful for. Unless the goal is to show the user which city each resource is coming from (which I'm not dismissing, btw, could be helpful) this information is never seen to the end user, so it seems weird at least for the test to assume the city is responsible for reporting Civ resources and not the Civ. Maybe this information is for devs debugging, or maybe this layout is for optimization, but minimum, I feel like the tests should still assume the Civ is responsible regardless unless it is a city resource
2. Currently, the unique is being used as if it reads `Provides [amount] [resource] [in all cities]`. For Civ resources, this is usually incorrect, hence the PR. Even for city resources, this seems error prone, though fine if we expect modders to only use it responsibly. But it does hint that this should probably be a different unique, perhaps something added to UniqueType.StatsPerCity or such.
3. Not sure what it is, but something feels off about the idea of `getCityResourcesAvailableToCity`. Not sure if the issue is the obvious question of "why is civ resources cached and city ones aren't" or if the issue is somewhere else. It sounds fine enough on paper, I just need more time thinking it over

PR #11236, opened 2024-03-02
Fix double-count of civ-wide resources, while allowing propagation of city-wide resources via uniques

yairm210:
Next step is turning this object into a class - it is (always* city-specific, and that way we can move the city and resourceModifiers into fields to avoid generating them in different places and passing around


core/src/com/unciv/logic/city/CityResources.kt
```diff
@@ -8,20 +8,43 @@ import com.unciv.models.ruleset.unique.UniqueType
...
+        // This way we get them once, but it is ugly, I welcome other ideas :/
+        getCityResourcesFromCiv(city, cityResources, resourceModifers)
+
+        return cityResources
+    }
 
-        getResourcesFromTiles(city, resourceModifer, cityResources)
 
-        getResourceFromUniqueImprovedTiles(city, cityResources, resourceModifer)
+    private fun getResourcesGeneratedByCityNotIncludingBuildings(city: City, resourceModifers: HashMap<String, Float>): ResourceSupplyList {
```
SeventhM, line 40: Why not including buildings? You can check buildings of just this city with `city.getMatchingLocalOnlyUniques`

Side note, probably accident of mine, but `city.getMatchingLocalOnlyUniques` and `getLocalMatchingUniques` reads as identical. One needs removing

core/src/com/unciv/logic/city/CityResources.kt
```diff
@@ -33,25 +56,36 @@ object CityResources {
         return cityResources
     }
 
+    private fun addCityResourcesFromUniqueBuildings(city: City, cityResources: ResourceSupplyList, resourceModifer: HashMap<String, Float>) {
+        for (unique in city.getMatchingUniquesWithNonLocalEffects(UniqueType.ProvidesResources, StateForConditionals(city.civ, city))) { // E.G "Provides [1] [Iron]"
```
SeventhM, line 60: This excludes follower religion and "In this city" uniques. Use `city.getMatchingLocalOnlyUniques`

core/src/com/unciv/logic/city/CityResources.kt
```diff
@@ -81,14 +115,14 @@ object CityResources {
...
 
-    private fun getCityResourcesFromCiv(city: City, cityResources: ResourceSupplyList, resourceModifer: HashMap<String, Float>) {
+    private fun getCityResourcesFromCiv(city: City, cityResources: ResourceSupplyList, resourceModifers: HashMap<String, Float>) {
         // This includes the uniques from buildings, from this and all other cities
-        for (unique in city.civ.getMatchingUniques(UniqueType.ProvidesResources, StateForConditionals(city.civ, city))) { // E.G "Provides [1] [Iron]"
+        for (unique in city.getMatchingUniques(UniqueType.ProvidesResources, StateForConditionals(city.civ, city))) { // E.G "Provides [1] [Iron]"
             val resource = city.getRuleset().tileResources[unique.params[1]]
                 ?: continue
             cityResources.add(
                 resource, "Buildings",
```
SeventhM, line 124: Forgot to fix the origin name here

core/src/com/unciv/logic/city/CityResources.kt
SeventhM: Ah, nevermind. I see what's going on. Still, using local only uniques here is ideal. The later calls can easily check the uniques for local effects to avoid doubling

SeventhM:
Here's a potential suggestion:
- Add a boolean to `city.getMatchingUniques` to ignore civ uniques (default to true)
- Change `addCityResourcesFromUniqueBuildings` to use `city.getMatchingUniques`, civ uniques == false
- In Civilization, where you have the line to continue on Buildings and wonders, also add in follower beliefs. You keep forgeting about those

There shouldn't be a situation where `getCityResourcesAvailableToCity` and `getResourcesGeneratedByCity` are reliant on each other, so this shouldn't double. Though I could be wrong, though.

Beyond that, this... should be... fine... well, something that bothers me is that `getResourceAmount` checks for the city resource unique. I'm not sure if that's an optimization thing or not. If not, it feels like either it should be done elsewhere, or more likely, it's unnecessary because `getCityResourcesAvailableToCity` accidentally already filters for that for redundancy.... wait... no... uh... huh... 

`getCityResourcesAvailableToCity` misses resources from tiles from other cities. Fine for city resources, problem for non city resources. The check for if it's a city resource should be both here and in `getResourceAmount`

core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
```diff
@@ -677,7 +677,7 @@ class CityStateFunctions(val civInfo: Civilization) {
     fun getCityStateResourcesForAlly() = ResourceSupplyList().apply {
         for (city in civInfo.cities) {
             // IGNORE the fact that they consume their own resources - #4769
-            addPositiveByResource(city.getCityResources(), Constants.cityStates)
+            addPositiveByResource(city.getResourcesGeneratedByCity(), Constants.cityStates)
```
SeventhM, line 680: Why is this not Civ resources? City states can capture unrazeable cities. Mods may give a city state a resource

core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
```diff
@@ -677,7 +677,7 @@ class CityStateFunctions(val civInfo: Civilization) {
...
         }
```
SeventhM, line 681: Why is this not civ resources? Mods can give city states resources to trade

yairm210:
> In Civilization, where you have the line to continue on Buildings and wonders, also add in follower beliefs. You keep forgeting about those

Follower beliefs are by definition only city-wide, you don't get them from GetMatchingUniques at all

core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
yairm210: Good point!!

yairm210: I remember why we did this, it was to avoid the fact that resources are consumed - this means kind of rebuilding the getCivResources... I'll add a todo here, out of scope for this PR

yairm210:
Re:  city.getMatchingUniques without civInfo, that's not what we want.
Again we have a naming problem - there's a difference between *uniques available to this city* and *uniques generated by this city that are available globally*, and what we want here is the latter. getMatchingUniques is meant to be the former.

yairm210:
In order to fix the next version I'm going to merge this now, the tests have convinced me that it's safe.
If there are further required changes we can add them in later.

SeventhM:
> Follower beliefs are by definition only city-wide, you don't get them from GetMatchingUniques at all

You actually do get them from city.getMatchingUniques. and the fact that they are defined city wide is kinda the point I'm trying to get across

> Again we have a naming problem

I don't think so, I think you're missing what I'm seeing or that we both see two very different end results:
1. `addCityResourcesGeneratedFromUniqueBuildings` uses `getMatchingUniquesWithNonLocalEffects` I forget why I added it (probably some technical limitations somewhere else), but this is not what you want. This explicitly only gets uniques from buildings that don't have `<in this city>` (see my comment in the other PR about the weird implied cityFilter you've added to this unique) and it explicitly avoids follower beliefs this means
2. As of this PR follower beliefs cannot give resources. You would need to actually be able to do some form of city.getMatchingUniques. This is why I say we need a version that ignores civ Uniques, as that would otherwise be identical to the other suggestion of `getMatchingUniquesWithNonLocalEffects` but adding follower beliefs
3. There are some areas where `getCityResourcesAvailableToCity` isn't exactly working as intended (see my comment about missing civ resources from tiles). This is most evident for WLTKD. I need to grab a computer and see if we have a test for that

Issue #10533, opened 2023-11-20
UI: Display option for Explore button


hackedpassword:
My bad @SomeTroglodyte you're right UI issue not modding. I'll delete the post if it's way too off 3242 topic, just drop a hint here.

-----

A `Display` option allowing toggling of the ever present Explore button.

Personally, after a few early eras there's no use for it. There are exceptions.

This is a constant source of UI grief for me. In later eras, it doesn't even make sense to be onscreen, _especially after revealing the map_. Should be auto-removed at that point regardless. :/

_Originally posted by @hackedpassword in https://github.com/yairm210/Unciv/issues/3242#issuecomment-1819833384_

-----

Really, was only trying to seed a suggestion, since we're here...

> branch

Seems awfully sophisticated for an if-then checkbox option render. Like creating a swiss army knife for a toothpick. Though those features sound interesting. Anyway, this issue bugs me and I want to crush that button into digital dust. Since I recently [dabbled with android studio](https://github.com/hackedpassword/atlas-regen) that button is now on my radar of destruction buahahaha

yairm210:
How did you open this issue?

hackedpassword:
> How did you open this issue?

I came from 3242, and I think I stayed in the github app, but exactly how I can't recall. Looked over at chrome cause that works better for issues, used that for another issue. I'm not sure, but I do see the reasoning for you asking. Strange. I'll let you know if I figure it out.

yairm210:
https://github.com/orgs/community/discussions/23034
Known issue for *checks watch* 4 years, so I'm sure they'll be dealing with it aaaany day now

hackedpassword:
> ... that one, right?

That looks right. Sidestepped the tagging/assignment process, had no idea. Now we know and knowing is half the battle - hmm half seems like a disproportionate percentile

hackedpassword:
> but I'd like swap to be reachable easier...

Since there's a unique that puts the automate button front and top, wouldn't it make sense to say:
"Swap is a primary action",
"Explore is a secondary action",

yairm210:
I'm fine with having explore be a secondary action after map reveal
But swap as a primary... I'll need to meditate on that

yairm210:
That would look awful on horizontal

yairm210:
[...]

PR #11243, opened 2024-03-03
Better Unit Actions Sorting

tuvus:
Related to #10533
Some of the UnitActions aren't displayed in an order that makes much sense. 
For example, sleep, connect road, and automate are in front of wait and swap units when they are used less often and are multiturn actions. Explore tends to be one of the most visible buttons but can't be used in the middle/late game. (see "before pictures" below)

As a fix, this PR adds in a new frequency sorting system. Each action is given a "use frequency", which is an int, usually between 0 and 100 of how the actions should be ordered. 
A higher value means more often and that it should be on an earlier page. 100 is very frequently used, 50 is somewhat frequently used, less than 25 is press one time for multi-turn movement. A Rare case is > 100 if a button is something like add in capital, promote or something that we need to inform the player that it is an option.

<details><summary>Here is the general order of actions from highest to lowest:</summary>
<p>

- Promotion
- Upgrading
- ReligionActions, One time uniques
- Construction
- Setup
- AirSweep
- Wait
- SwapUnits
- SleepUntilHealed/FortifyUntilHealed
- Sleep/Fortify
- ConnectRoad
- Automate
- Explore
- Gift
- Disband

</p>
</details> 

<details><summary>BeforePictures</summary>

ButtonSortingBefore1
ButtonSortingBefore2
ButtonSortingBefore3
ButtonSortingBefore4
ButtonSortingBefore5

</details> 

<details><summary>AfterPictures</summary>

ButtonSorting1
ButtonSorting2
ButtonSorting3

</details> 

There also is a `getActionDefaultPage()` system. However, it isn't very comprehensive, doesn't allow for sorting within a page, and may force a new page to be created even if it is the only action on it. I have left this in for now, but we might want to completely replace it.

### DIFF
--- a/core/src/com/unciv/models/UnitAction.kt
+++ b/core/src/com/unciv/models/UnitAction.kt
@@ -79,7 +79,7 @@ class UpgradeUnitAction(
     val goldCostOfUpgrade: Int,
     val newResourceRequirements: Counter<String>,
     action: (() -> Unit)?
-) : UnitAction(UnitActionType.Upgrade, 10, title, action = action)
+) : UnitAction(UnitActionType.Upgrade, 120, title, action = action)
 
 /** Unit Actions - generic enum with static properties
  *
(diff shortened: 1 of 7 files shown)


tuvus:
I was also thinking of doing a float. Then we can push things in between other actions better.

yairm210 (approved):

PR #11244, opened 2024-03-03
Skip next unit button

tuvus:
In the late game, it gets kind of annoying having to either press wait on each unit or press the next unit button for every unit to end a turn. 
This PR fixes this by adding a right click menu (or long press) where the player can skip pressing the next turn button. I also added a move automated units button so the player can still move their automated units and then end their turn. The move automated units button is only shown if there are units to automate.

NextTurnMenu

</details>

### DIFF
--- a/core/src/com/unciv/ui/components/input/KeyboardBinding.kt
+++ b/core/src/com/unciv/ui/components/input/KeyboardBinding.kt
@@ -142,6 +142,10 @@ enum class KeyboardBinding(
     AutoPlayMenuCivilians(Category.AutoPlayMenu, "AutoPlay Civilians Once", 'c'),
     AutoPlayMenuEconomy(Category.AutoPlayMenu, "AutoPlay Economy Once", 'e'),
 
+    // NextTurnMenu
+    NextTurnMenuNextTurn(Category.NextTurnMenu, "Next Turn", 'n'),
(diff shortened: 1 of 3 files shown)


tuvus:
The nextTurn button can become disabled when pressing the move automated units with the right click. This forces the user to make a click on the worldscreen before it enables again. This also occurs when doing some autoplay right clicks. Does anyone know a workaround?

yairm210:
This is exactly why there's an option for automate units, that way you don't need to select each one. Turn that in, and every turn is 2 clicks

tuvus:
Are you talking about automate units on turn start? It is not necessarily a replacement. A player might not want to commit to turn the option on for all of their games. This also allows them to make changes before the automated units are moved.

yairm210:
I'm fine with this, since it's an alt-menu - but if this starts affecting the main button that's a no :)

yairm210:
@tuvus signs like what WoldS reen.shouldUpdate is meant to solve, set it to true when you're done with the button

yairm210 (approved):

tuvus:
> Also, that Popup has a closing animation, and can call stuff either right at the close click or at the end of the animation - init { afterCloseCallback = { worldScreen.shouldUpdate = true } } should do it, but you do your testing.

That works! I will create a new PR since this is already merged and I will also add it to AutoPlayMenu.

PR #11249, opened 2024-03-04
Add configurable natural wonder discovery stat bonuses

PLynx01:
Some time ago, I noticed that the natural wonder discovery stat bonus was hard-coded at 100 Gold (or 500 when discovered as first). I've changed that and from now, modders can configure not only the amount of bonus, but also type of granted stat. It can be Gold, Science, Culture or Faith (these are civ-wide stats)

### DIFF
--- a/core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
+++ b/core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
@@ -262,17 +262,26 @@ class CivInfoTransientCache(val civInfo: Civilization) {
                 }
             }
 
+            var naturalWonder: String? = null
 
             if (!statsGained.isEmpty()) {
+                naturalWonder = tile.naturalWonder!!
+            }
+
+            if (!statsGained.isEmpty() && naturalWonder != null) {
                 civInfo.addStats(statsGained)
-                civInfo.addNotification("We have received [${statsGained}] for discovering [${tile.naturalWonder}]",
+                civInfo.addNotification("We have received [${statsGained}] for discovering [${naturalWonder}]",
                     Notification.NotificationCategory.General, statsGained.toString()
                     )
             }
 
             if (goldGained > 0) {
+                naturalWonder = tile.naturalWonder
+            }
+
+            if (goldGained > 0 && naturalWonder != null) {
                 civInfo.addGold(goldGained)
-                civInfo.addNotification("We have received [$goldGained] Gold for discovering [${tile.naturalWonder}]",
+                civInfo.addNotification("We have received [$goldGained] Gold for discovering [${naturalWonder}]",
                     Notification.NotificationCategory.General, NotificationIcon.Gold)
             }
 
--- a/core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
+++ b/core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
@@ -18,6 +18,7 @@ import com.unciv.models.ruleset.unique.UniqueTarget
 import com.unciv.models.ruleset.unique.UniqueTriggerActivation
 import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.models.ruleset.unit.BaseUnit
+import com.unciv.models.stats.Stat
 import com.unciv.utils.DebugUtils
 
 /** CivInfo class was getting too crowded */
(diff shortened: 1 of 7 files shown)


PLynx01:
I don't know what went wrong with code checks. Please explain me.

SeventhM:
> Aren't deprecated UniqueTypes moved to the other region

Technically yes, though I'm usually pretty lazy about it myself

I have two major issues with this PR:
1. The notification can be simplified by just checking `relevantStat.name`
2. This PR seems redundant. `Gain [amount] [stat] (modified by game speed) <upon discovering a Natural Wonder>` Already covers this usecase. Seems the better option is just to deprecate the old unique and make zero changes to the code beyond that

yairm210:
Good generalization¡

core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
```diff
@@ -215,23 +217,43 @@ class CivInfoTransientCache(val civInfo: Civilization) {
...
+                    relevantStat = Stat.safeValueOf(unique.params[1])
+                    val bonusAmount = unique.params[0].toInt()
+
+                    statGained += bonusAmount
+                }
             }
 
-            if (civInfo.hasUnique(UniqueType.GoldWhenDiscoveringNaturalWonder)) {
-                goldGained += if (discoveredNaturalWonders.contains(tile.naturalWonder!!)) 100 else 500
+            if (civInfo.hasUnique(UniqueType.StatBonusWhenDiscoveringNaturalWonder)) {
```
yairm210, line 236: Unnecessary if, since the for loop loops over all such uniques anyeay

core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
```diff
@@ -215,23 +217,43 @@ class CivInfoTransientCache(val civInfo: Civilization) {
...
+                    val firstDiscoveredBonus = unique.params[2].toInt()
+
+                    statGained += if (discoveredNaturalWonders.contains(tile.naturalWonder!!))
+                        normalBonus
+                    else
+                        firstDiscoveredBonus
+                }
             }
 
-            if (goldGained > 0) {
```
yairm210, line 230: Do not replace existing unique implementation! This will break all current mods!
Instead add implementation for new unique alongside the old one

core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
```diff
@@ -215,23 +217,43 @@ class CivInfoTransientCache(val civInfo: Civilization) {
...
+                    statGained += bonusAmount
+                }
             }
 
-            if (civInfo.hasUnique(UniqueType.GoldWhenDiscoveringNaturalWonder)) {
-                goldGained += if (discoveredNaturalWonders.contains(tile.naturalWonder!!)) 100 else 500
+            if (civInfo.hasUnique(UniqueType.StatBonusWhenDiscoveringNaturalWonder)) {
+                for (unique in civInfo.getMatchingUniques(UniqueType.StatBonusWhenDiscoveringNaturalWonder)) {
+
+                    if (relevantStat == null)
```
yairm210, line 239: ???
Makes no sense to base of of the *wonder's* stat when we have a stat encoded in the unique

yairm210:
There's an assumption here that each nat wonder only gives one stat, that will break really fast when modders start adding multiple uniques.
Instead we should assume we can gain many stats from a nat wonder

core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
yairm210: Tbh I didn't even read your comments... There are Too Many PRs so I'm just trying to go over all of them...
Your PRs for example are gnarly. I don't have the headspace for the rivers right now...

yairm210:
Much better! Still removes the current implementation, please return it so modders can gradually migrate

core/src/com/unciv/logic/civilization/transients/CivInfoTransientCache.kt
```diff
@@ -215,23 +217,63 @@ class CivInfoTransientCache(val civInfo: Civilization) {
...
+
+                for (unique in tile.getTerrainMatchingUniques(UniqueType.GoldWhenDiscoveringNaturalWonder)) {
+                    goldGained += 500
+                }
+            }
+
+
+            if (!statsGained.isEmpty()) {
+                civInfo.addStats(statsGained)
+                civInfo.addNotification("We have received [${statsGained}] for discovering [${tile.naturalWonder}]",
```
yairm210, line 268: I see that this new notification text does not yet exist.
You need to add it to template.properties, preferably right under "We have received [goldAmount] Gold for discovering [naturalWonder]"

Format should be:
```
We have received [statsGained] for discovering [naturalWonder] = 
```

yairm210: Since it's going to be deprecated in a month or so anyway, I don't really see a point. Allow current mods to keep using existing translations.

yairm210: Yes, keep both
We'll delete the old ones when the old unique is actually deprecated

yairm210 (approved):
Looks good!

I assume you checked that the old uniques still work?

PR #11250, opened 2024-03-04
Add Conditionals for comparing Stats and Resources 

PLynx01:
I've made 4 new conditional uniques for comparing Stats and Resources numbers.

- The first one is true when stat or resource has an exact number.
- The second one is true when two of these have equal amount.
- The third and fourth are very similar, because they both compare stat and resource with each other, and return true when first or second is bigger.

The last three uniques support all 4 ways of comparison:
1. Resource - Resource
2. Stat - Stat
3. Stat - Resource
4. Resource - Stat

I believe these additions will give modders a range of possibilities as Resources can be used as custom variables

If you have some suggestions, please tell me in comments below. Constructive feedback welcome.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -83,6 +83,42 @@ object Conditionals {
             return compare(relevantCiv!!.getEraNumber(), era.eraNumber)
         }
 
+        /** Helper for comparing the Resource and Stat numbers **/
+        fun compareResourceOrStatAmount(compare: (first: Int, second: Int) -> Boolean): Boolean {
+            if (gameInfo == null) return false
+            val first = condition.params[0]
+            val second = condition.params[1]
+
+            // Compare Resource with Resource
+            if (checkOnResourceName(first) && checkOnResourceName(second)) {
+                return compare(getResourceAmount(first), getResourceAmount(second))
+            }
+            // Compare Stat with Stat
+            else if (Stat.isStat(first) && Stat.isStat(second)) {
+                val firstStat = Stat.safeValueOf(first) ?: return false
+                val secondStat = Stat.safeValueOf(second) ?: return false
+
+                return compare(relevantCiv!!.getStatReserve(firstStat), relevantCiv!!.getStatReserve(secondStat))
+            }
+            // Compare Stat with Resource
+            else if (Stat.isStat(first) && checkOnResourceName(second)) {
+                val stat = Stat.safeValueOf(first) ?: return false
+                val resourceAmount = getResourceAmount(second)
+
+                return compare(relevantCiv!!.getStatReserve(stat), resourceAmount)
+            }
+            // Compare Resource with Stat
+            else if (checkOnResourceName(first) && Stat.isStat(second)) {
+                val resourceAmount = getResourceAmount(first)
+                val stat = Stat.safeValueOf(second) ?: return false
+
+                return compare(resourceAmount, relevantCiv!!.getStatReserve(stat))
+            }
+            else {
+                return false
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -78,6 +83,42 @@ object Conditionals {
...
         }
 
+        /** Helper for comparing the Resource and Stat numbers **/
+        fun compareResourceOrStatAmount(compare: (first: Int, second: Int) -> Boolean): Boolean {
+            if (gameInfo == null) return false
+            val first = condition.params[0]
+            val second = condition.params[1]
+
+            // Compare Resource with Resource
+            if (checkOnResourceName(first) && checkOnResourceName(second)) {
```
yairm210, line 93: Entirely wedding way to go around this.
You should check each element indicidually to get number A and B.
Then compare.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -660,14 +660,22 @@ enum class UniqueType(
...
-    // Supports also stockpileable resources (Gold, Faith, Culture, Science)
+    // Supports also stockpileable stats (Gold, Faith, Culture, Science)
     ConditionalWhenAboveAmountStatResource("when above [amount] [stat/resource]", UniqueTarget.Conditional),
     ConditionalWhenBelowAmountStatResource("when below [amount] [stat/resource]", UniqueTarget.Conditional),
 
     // The game speed-adjusted versions of above
     ConditionalWhenAboveAmountStatResourceSpeed("when above [amount] [stat/resource] (modified by game speed)", UniqueTarget.Conditional),
     ConditionalWhenBelowAmountStatResourceSpeed("when below [amount] [stat/resource] (modified by game speed)", UniqueTarget.Conditional),
 
+    // For exact number of resources or stats
```
yairm210, line 671: These don't fit into uniques! They should start with "when"

SeventhM: Why not start with "with"? Seems like the better option with this wording

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
PLynx01: Done. Check it out.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: Mulling it over I agree that "with" sounds better in this case

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
woo1127: > Done. Check it out.

Below are my suggestion to optimize the code, not tested yet, however

```kotlin
        fun getStatOrResourceAmount(param: String): Int? {
            return if (gameInfo!!.ruleset.tileResources.containsKey(param)) getResourceAmount(param)
            else if (Stat.isStat(param)) {
                if (relevantCiv != null)
                    relevantCity?.getStatReserve(Stat.safeValueOf(param)!!)
                relevantCiv?.getStatReserve(Stat.safeValueOf(param)!!)
            }
            else null
        }

        /** Helper for comparing the Resource and Stat numbers **/
        fun compareResourceOrStatAmount(compare: (first: Int, second: Int) -> Boolean): Boolean {
            if (gameInfo == null) return false
            return compare(
                getStatOrResourceAmount(condition.params[0]) ?: return false,
                getStatOrResourceAmount(condition.params[1]) ?: return false
            )
        }
```

PLynx01:
I close this PR, because I made a new one:
#11280

PR #11252, opened 2024-03-05
Allow resources from follower beliefs

SeventhM:
All changes as per suggested [previously](https://github.com/yairm210/Unciv/pull/11236#issuecomment-1974932098)

There's still the awkwardness that I think should need some discussion of city resources implicitly functioning as `Provides [amount] [resource] [in all cities]` rather than what the unique actually says

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -454,24 +454,22 @@ class City : IsPartOfGameInfoSerialization {
     // Finds matching uniques provided from both local and non-local sources.
     fun getMatchingUniques(
         uniqueType: UniqueType,
-        stateForConditionals: StateForConditionals = StateForConditionals(civ, this)
+        stateForConditionals: StateForConditionals = StateForConditionals(this),
+        civUniques: Boolean = true
     ): Sequence<Unique> {
-        return civ.getMatchingUniques(uniqueType, stateForConditionals) +
+        return if (civUniques)
+            civ.getMatchingUniques(uniqueType, stateForConditionals) +
                 getLocalMatchingUniques(uniqueType, stateForConditionals)
-    }
-
-    // Uniques special to this city
-    fun getLocalMatchingUniques(uniqueType: UniqueType, stateForConditionals: StateForConditionals = StateForConditionals(civ, this)): Sequence<Unique> {
-        return (
-            cityConstructions.builtBuildingUniqueMap.getUniques(uniqueType).filter { it.isLocalEffect }
+        else (
+            cityConstructions.builtBuildingUniqueMap.getUniques(uniqueType)
                 + religion.getUniques().filter { it.type == uniqueType }
             ).filter {
                 !it.isTimedTriggerable && it.conditionalsApply(stateForConditionals)
             }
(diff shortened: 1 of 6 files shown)


yairm210:
I don't think this is required at all.
Looking at the call stack:
- getCityResourcesAvailableToCity
- getCityResourcesFromCiv 
- city.getMatchingUniques 
- getLocalMatchingUniques
- religion.getUniques().filter { it.type == uniqueType }

I would claim, not having actually checked a game, that follower beliefs are already included.

Do you have an example game that contradicts this?

SeventhM:
[Autosave.txt](https://github.com/yairm210/Unciv/files/14548113/Autosave.txt)
G&K ruleset. Mod is the below. Can't build a farm as is. Can't make a swordsman as is. Away from computer to confirm this does work on this PR tbh, but pretty sure it does
<details>
  <summary>Global Uniques</summary>
  
  ```xml
  {
    "name": "Global uniques",
    "uniques": [
        "Free [Great Prophet] appears"
    ]
}
  ```
  
</details>
<details>
  <summary>Beliefs</summary>
  
  ```xml
  [
    {
        "name": "Test Belief",
        "type": "Follower",
        "uniques": ["Provides [1] [Iron]"]
    },
]
  ```
  
</details>
<details>
  <summary>Tile Improvement</summary>
  
  ```xml
  [
    {
        "name": "Farm",
        "terrainsCanBeBuiltOn": ["Plains","Grassland","Desert","Flood plains"],
        "food": 1,
        "turnsToBuild": 7,
        "techRequired": "Agriculture",
        "uniques": [
            "Can also be built on tiles adjacent to fresh water <in [Land] tiles>",
            "[+1 Food] from [Fresh water] tiles <after discovering [Civil Service]>",
            "[+1 Food] from [non-fresh water] tiles <after discovering [Fertilizer]>",
            "Can only be built to improve a resource <in [Strategic resource] tiles>",
            "Can only be built to improve a resource <in [Bonus resource] tiles>",
            "Can only be built to improve a resource <in [Luxury resource] tiles>",
            "Pillaging this improvement yields approximately [+18 Gold]"
            "Consumes [1] [Iron]"
        ],
        "shortcutKey": "F"
    },
]
  ```
  
</details>

> Looking at the call stack:

Incorrect. Resources when building constructions/for conditionals uses `getAvailableResourceAmount`. For non city resources, this uses `city.civ.getResourceAmount(resourceName)`. Even if this used `getCityResourcesAvailableToCity` for constructions/conditionals, you it still doesn't work for tile improvements

yairm210:
Ahhh there's a nice distinction here!
Follower beliefs *apply to* specific cities
You claim their effects should not be *limited in scope* to those cities, in the case of resources at least.
I'll need to think about this, if we allow these uniques from followers to affect ither cities then the whole concept of "applies only to city X" falls apart, I'd say this is something unique to resources or disallow it entirely

yairm210:
Managed to replicate your problem with a test, required change is minimal - "resources we grant to civ" should address *all local* uniques, and not just *uniques we provide to other cities*.
#11286

yairm210 (approved):

PR #11253, opened 2024-03-05
More city filters: Resisting, being razed

SomeTroglodyte:
Just an idea - what if some mod wished to alleviate the strict rule that my Oxford University just disappears from construction queues the moment I start winning a war, no matter what I do to conquered cities... Also, more symmetry.

The short form should fit into "<if [whatnot] is constructed in all [filter] cities>" - and I found no form of "being razed" that does. At least "Razing" doesn't get worse with the "non" modifier. I'd like "Burning" but that's not quite intuitive. Better ideas?

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
@@ -255,6 +255,8 @@ enum class UniqueParameterType(
             "in foreign cities", "Foreign",
             "in annexed cities", "Annexed",
             "in puppeted cities", "Puppeted",
+            "in resisting cities", "Resisting",
+            "in cities being razed", "Razing",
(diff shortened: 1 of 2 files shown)


yairm210:
Mmmm English is hard :/
No better alternative to offer :(

core/src/com/unciv/models/ruleset/unique/UniqueParameterType.kt
```diff
@@ -255,6 +255,8 @@ enum class UniqueParameterType(
             "in foreign cities", "Foreign",
             "in annexed cities", "Annexed",
             "in puppeted cities", "Puppeted",
+            "in resisting cities", "Resisting",
```
yairm210, line 258: Do we use this list to populate translation files? If not we probably should

yairm210 (approved):

PR #11254, opened 2024-03-05
Allow AI to consider production bonuses when building

SeventhM:
Yes. I know, I know. This break AI gold buying decisions. Am deciding to split that to its own PR, if for no other reason than patch notes, but also AIs really shouldn't be using production considerations for gold buying, and that may or may not be more involved


core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -267,6 +267,10 @@ class CityConstructions : IsPartOfGameInfoSerialization {
         if (workLeft <= productionOverflow) // if we already have stored up enough production to finish it directly
             return 1 // we'll finish this next turn
 
+        return ceil((workLeft-productionOverflow) / productionForConstruction(constructionName).toDouble()).toInt()
```
yairm210, line 270: Sounds be separated to vals, for readability

yairm210 (approved):

PR #11255, opened 2024-03-05
Add the unfiltered version of new era unique

PLynx01:
I've added the unfiltered version of new era unique. It may seem too simple, but it has some use cases:

1. It is intended to replace the multiple filtered new era uniques in mods, thus shortening their unique lists
2. It will be used for Polish Civilization's unique ability in future base BNW ruleset or in other mods, because in the original Civ 5 BNW Poland receives a free social policy on entering the new era. (see the link below)

[https://civilization.fandom.com/wiki/Civilizations_(Civ5)](https://civilization.fandom.com/wiki/Civilizations_(Civ5))

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/TechManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/TechManager.kt
@@ -465,7 +465,7 @@ class TechManager : IsPartOfGameInfoSerialization {
                 UniqueTriggerActivation.triggerUnique(
                     unique,
                     civInfo,
-                    triggerNotificationText = "due to entering the new era")
+                    triggerNotificationText = "due to entering the [${currentEra.name}]")
         }
     }
 
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -780,7 +780,7 @@ enum class UniqueType(
     @Deprecated("as of 4.10.15", ReplaceWith("upon discovering [tech] technology"))
     TriggerUponResearchOld("upon discovering [tech]", UniqueTarget.TriggerCondition),
     TriggerUponEnteringEra("upon entering the [era]", UniqueTarget.TriggerCondition),
-    TriggerUponEnteringEraUnfiltered("upon entering the new era", UniqueTarget.TriggerCondition),
+    TriggerUponEnteringEraUnfiltered("upon entering a new era", UniqueTarget.TriggerCondition),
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -780,6 +780,7 @@ enum class UniqueType(
     @Deprecated("as of 4.10.15", ReplaceWith("upon discovering [tech] technology"))
     TriggerUponResearchOld("upon discovering [tech]", UniqueTarget.TriggerCondition),
     TriggerUponEnteringEra("upon entering the [era]", UniqueTarget.TriggerCondition),
+    TriggerUponEnteringEraUnfiltered("upon entering the new era", UniqueTarget.TriggerCondition),
```
yairm210, line 783: *A* new era

core/src/com/unciv/logic/civilization/managers/TechManager.kt
```diff
@@ -459,6 +459,13 @@ class TechManager : IsPartOfGameInfoSerialization {
                             civInfo,
                             triggerNotificationText = "due to entering the [$eraName]"
                         )
+
+            // The unfiltered version
+            for (unique in civInfo.getTriggeredUniques(UniqueType.TriggerUponEnteringEraUnfiltered))
+                UniqueTriggerActivation.triggerUnique(
+                    unique,
+                    civInfo,
+                    triggerNotificationText = "due to entering the new era")
```
yairm210, line 468: Reuse the existing text, above lone 460 - more informative, AND translators don't need to add a line

PLynx01:
@yairm210 
Maybe I should remade this as a era filter, like "any era" or "new era" ? 

Because you want to have as few uniques as possible that do as much as possible?

yairm210:
Hm good point
Eras currently look like "Future era" so this would look bad anyway ("future era era"?)
We can have an era filter I guess, since eras have their own uniques
So "Upon entering [eraFilterOrName]" where eraFilter is either an era name or of format "[EraFilter] era"?
Would require more scaffolding

PLynx01:
> Hm good point
> Eras currently look like "Future era" so this would look bad anyway ("future era era"?)
> We can have an era filter I guess, since eras have their own uniques
> So "Upon entering [eraFilterOrName]" where eraFilter is either an era name or of format "[EraFilter] era"?
> Would require more scaffolding

So, what's your final opinion on that matter?

yairm210 (approved):

PR #11256, opened 2024-03-05
River terraform

SomeTroglodyte:
Extension to #11152.

Currently, using "River" as parameter will not be flagged on validation, but will never be active - occursOn empty. Which is good, actually placing that feature would be bad.

Caveat: convertTerrains - RiverGenerator runs it once after the river is done, on the collected neighbors. You get double conversion only if two rivers by chance get close enough. Snow-Tundra-Plains, was it? Well, since this will run stepwise one edge at a time, you'll get that more often. No idea how to mitigate.

This image (from [this comment](https://github.com/yairm210/Unciv/pull/11152#issuecomment-1954188127)) was done with not quite the same code, but should still apply:
<details><summary>anim gif</summary>

Untitled

</details>

... and I still have no idea how this Unique can be made to take several turns to "build" ...

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
@@ -220,20 +220,23 @@ class MapGenerator(val ruleset: Ruleset, private val coroutineScope: CoroutineSc
         debug("MapGenerator.%s took %s.%sms", text, delta/1000000L, (delta/10000L).rem(100))
     }
 
-    fun convertTerrains(tiles: Iterable<Tile>) {
-        for (tile in tiles) {
-            val conversionUnique =
-                tile.getBaseTerrain().getMatchingUniques(UniqueType.ChangesTerrain)
-                    .firstOrNull { tile.isAdjacentTo(it.params[1]) }
-                    ?: continue
-            val terrain = ruleset.terrains[conversionUnique.params[0]] ?: continue
+    fun convertTerrains(tiles: Iterable<Tile>) = Helpers.convertTerrains(ruleset, tiles)
+    object Helpers {
+        fun convertTerrains(ruleset: Ruleset, tiles: Iterable<Tile>) {
+            for (tile in tiles) {
+                val conversionUnique =
+                    tile.getBaseTerrain().getMatchingUniques(UniqueType.ChangesTerrain)
+                        .firstOrNull { tile.isAdjacentTo(it.params[1]) }
+                        ?: continue
+                val terrain = ruleset.terrains[conversionUnique.params[0]] ?: continue
 
-            if (terrain.type == TerrainType.TerrainFeature) {
-                if (!terrain.occursOn.contains(tile.lastTerrain.name)) continue
-                tile.addTerrainFeature(terrain.name)
-            } else
-                tile.baseTerrain = terrain.name
-            tile.setTerrainTransients()
+                if (terrain.type == TerrainType.TerrainFeature) {
+                    if (!terrain.occursOn.contains(tile.lastTerrain.name)) continue
+                    tile.addTerrainFeature(terrain.name)
+                } else
+                    tile.baseTerrain = terrain.name
+                tile.setTerrainTransients()
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
```diff
@@ -189,4 +190,105 @@ class RiverGenerator(
...
+                if (neighbor.priority > maxPriority) maxPriority = neighbor.priority
+            }
+
+            // Finally - choose
+            val choice = viableNeighbors.filter { it.priority == maxPriority }.random()
+            tile.setConnectedByRiver(choice.otherTile, choice.clockPosition)
+            return true
+        }
+
+        private fun Tile.setConnectedByRiver(otherTile: Tile, clockPosition: Int) {
```
yairm210, line 279: If we know the neighbor, we can get the clock position from that - having two inputs implies cases where the position is not the relative neighbor position?

core/src/com/unciv/logic/map/mapgenerator/MapGenerator.kt
```diff
@@ -220,20 +220,23 @@ class MapGenerator(val ruleset: Ruleset, private val coroutineScope: CoroutineSc
...
+    object Helpers {
+        fun convertTerrains(ruleset: Ruleset, tiles: Iterable<Tile>) {
+            for (tile in tiles) {
+                val conversionUnique =
+                    tile.getBaseTerrain().getMatchingUniques(UniqueType.ChangesTerrain)
+                        .firstOrNull { tile.isAdjacentTo(it.params[1]) }
+                        ?: continue
+                val terrain = ruleset.terrains[conversionUnique.params[0]] ?: continue
+
+                if (terrain.type == TerrainType.TerrainFeature) {
```
yairm210, line 233: Reverse this - so we have if-else-if instead of if-if-else

core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
```diff
@@ -189,4 +190,89 @@ class RiverGenerator(
...
+            if (viableNeighbors.all { it.isConnectedByRiver }) return false // cannot place another river
+
+            // Greatly encourage connecting to sea unless the tile already has a river to sea, in which case slightly discourage another one
+            val edgeToSeaPriority = if (viableNeighbors.none { it.isConnectedByRiver && it.edgeLeadsToSea }) 9 else -1
+
+            // Assign priorities
+            var maxPriority = Int.MIN_VALUE
+            for (neighbor in viableNeighbors) {
+                neighbor.priority = when {
+                    neighbor.isConnectedByRiver -> -9 // ensures this isn't chosen
```
yairm210, line 257: What if this is the only option?

core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
```diff
@@ -189,4 +190,89 @@ class RiverGenerator(
...
+                if (!otherTile.isLand) continue // No rivers bordering water
+                viableNeighbors += NeighborData(clockPosition, otherTile)
+            }
+            if (viableNeighbors.all { it.isConnectedByRiver }) return false // cannot place another river
+
+            // Greatly encourage connecting to sea unless the tile already has a river to sea, in which case slightly discourage another one
+            val edgeToSeaPriority = if (viableNeighbors.none { it.isConnectedByRiver && it.edgeLeadsToSea }) 9 else -1
+
+            // Assign priorities
+            var maxPriority = Int.MIN_VALUE
```
yairm210, line 254: This sounds like groupBy.maxBy

core/src/com/unciv/logic/map/mapgenerator/RiverGenerator.kt
```diff
@@ -189,4 +190,89 @@ class RiverGenerator(
...
+                    verticesFormYCount = sequence {
+                        yield(leftSharedNeighbor?.run { isConnectedByRiver(tile) && isConnectedByRiver(otherTile) })
+                        yield(rightSharedNeighbor?.run { isConnectedByRiver(tile) && isConnectedByRiver(otherTile) })
+                    }.count { it == true }
+                }
+            }
+
+            // Collect data
+            val viableNeighbors = mutableListOf<NeighborData>() // includes those we already have a river edge with - need the stats
+            for (clockPosition in 2..12 step 2) {
```
yairm210, line 242: If you get clock position inside neighborData, this can just be tile.neighbors

yairm210 (approved):

PR #11260, opened 2024-03-05
Add the Conditional Unique for checking speed

PLynx01:
As I said on Discord, I've posted a pull request adding the conditional unique for checking the game speed, alongside with proper parameter type. This unique is simple and straightforward, but I don't know whether the wording is correct.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueType.kt
@@ -634,7 +634,7 @@ enum class UniqueType(
     ConditionalStartingFromEra("starting from the [era]", UniqueTarget.Conditional),
     ConditionalIfStartingInEra("if starting in the [era]", UniqueTarget.Conditional),
 
-    ConditionalSpeed("when speed is [speed]", UniqueTarget.Conditional),
+    ConditionalSpeed("on [speed] game speed", UniqueTarget.Conditional),
 
     ConditionalFirstCivToResearch("if no other Civilization has researched this", UniqueTarget.Conditional),
     ConditionalTech("after discovering [tech]", UniqueTarget.Conditional),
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -634,6 +634,8 @@ enum class UniqueType(
     ConditionalStartingFromEra("starting from the [era]", UniqueTarget.Conditional),
     ConditionalIfStartingInEra("if starting in the [era]", UniqueTarget.Conditional),
 
+    ConditionalSpeed("when speed is [speed]", UniqueTarget.Conditional),
+
```
yairm210, line 638: When *game* speed is []

yairm210:
All in all looks good!
Compact and combinable, just what we want from uniques

SeventhM:
I swear something feels off about it in terms of wording, but that's just me. ~~Would a native speaking real write the speed as a noun like that? Something like "for [speed] game speed" sounds a bit robotic I guess, but Idk feels a bit better. Eh, probably just nitpicking~~

yairm210:
If you can think of a different phrasing, we can do that :)
If not, I'm in favor of merging as-is - we can always change later, and this is a classic case of autoupdate to the rescue - modders will only need to click the button

SeventhM:
By default I'm thinking something like `for the [speed] game speed` or `on [speed] game speed` or at the very very least `when the game speed is [speed]`

yairm210:
That last one was my suggestion :)
"on [] game speed" sounds better to me though, nice! 👍🏿

SeventhM:
> That last one was my suggestion 

Key thing was missing the word "the"

SeventhM:
Mergeable... Translation?

yairm210:
Incorrectly marked, will change the text

PR #11262, opened 2024-03-05
Add year-related conditional uniques

PLynx01:
We know it should be added a long time ago, there was a try to implement them, but it failed. Maybe this attempt will succeed. I added the four year-related conditionals alongside with year unique parameter and two conditional helper functions. 

If you have some suggestions, please share them in the comments. Constructive feedback welcome.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -106,6 +106,31 @@ object Conditionals {
             return compare(relevantCiv!!.getStatReserve(stat), limit * gameSpeedModifier)
         }
 
+        fun isYearValid(year: String): Boolean {
+            return try {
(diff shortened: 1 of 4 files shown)


yairm210:
Since you already have int comparison uniques pending we can use that for all "intables"
We want to have as few uniques as possible, that do as much as possible

PLynx01:
> Since you already have int comparison uniques pending we can use that for all "intables"
> We want to have as few uniques as possible, that do as much as possible

So, in that case, should I make an int-comparing unique that accepts Stats, Resources and Years?

PLynx01:
> We want to have as few uniques as possible, that do as much as possible

If you really want this, why don't we make, for example the comparison filters with following values?

- equal
- different
- larger
- smaller
- equal or smaller
- equal or bigger

They can reduce the number of uniques significantly, without reducing their possibilities.

PR #11273, opened 2024-03-07
Update Czech.properties

jahodovykolac:
For this translation I have one question - what exactly is Wesnoth?

And for any other Czech translators, I have noticed an inconsistency in translating unit upgrade (sometimes vylepšení, sometimes přestavba), which is extra conflicting with the word transform now being a thing. I suggest upgrade should be translated as vylepšení, but don't want to change stuff without other opinion

### DIFF
--- a/android/assets/jsons/translations/Czech.properties
+++ b/android/assets/jsons/translations/Czech.properties
@@ -1160,7 +1160,7 @@ Escort formation =
 Stop Escort formation = 
 Paradrop = Výsadek
  # Requires translation!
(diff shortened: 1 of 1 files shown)


jahodovykolac:
> A prominent open source game - and a feature in our map editor to load maps created in their map editor.

Thanks, I'll do the lines with that

> Me niid tu kan hav?

Me bran tu tird, worts tu seim

yairm210 (approved):

PR #11274, opened 2024-03-07
Expand CanOnlyBeBuiltInCertainCities to include Units and convert to use Conditionals

itanasi:
Rework CanOnlyBeBuiltInCertainCities to act more like OnlyAvailable. Expand to cover Units as well as Buildings. Takes chain of Conditionals.

Now allows Transform and Upgrade Actions to target units that can be restricted by features like specific buildings instead of being blocked by OnlyAvailable.


SeventhM:
I thought the discussion went the opposite way around: only available stays here, some unique is added to replace

itanasi:
Straw poll and discussion started here
https://discord.com/channels/586194543280390151/664739473367760908/1211546922670620765

SeventhM:
Prrrrrreeetty sure the discussion is a bit more complicated than that. That strawpoll might've been interpreted vaguely

There's two separate discussions here:
1. Should there be a difference between limitations towards building and limitations towards upgrading/transforming. There, the answer is overwhelmingly yes
2. How should the distinction be made? That poll does in fact imply the solution can be not checking OnlyAvailbleWhen from transforming, but there still needs to be \*some\* way to limit it by uniques. Unless the idea is to just rely on the transforming unique's conditionals themselves, which wasn't polled or discussed. Discussion in the developer channel implied the solution should probably be adding a conditional for constructions... Which is actually already there


If we referencing links to discord, discussion should move to discord, though. Links here aren't the best, at least for me

itanasi:
The #developers discussion said a more robust system would be separate out Can Construct vs Only Available for units so that Construction limitations doesn't limit Transform and Upgrade paths. Current Can Construct is used for Tile Improvements only.

Discussion was here, and MCompany pointed out the previous issue with OnlyAvailable blocking Upgrade paths:
https://discord.com/channels/586194543280390151/654043902948409394/1211544178199109653

As per the poll, I had suggested all conditional control would need to be in the string for the Transform action, and seemed modders were OK with this compromise.

Spinning out a separate Can Construct and leaving Only Available would save complexity in the strings. But then we need to be super clear why they're distinct since by default everything uses Only Available When right now.

itanasi:
K, doing BuildableOnly as a separate unique similar to OnlyAvailable, but specifically just for Construction of Buildings and Units.

Not sure about doing a full overhaul but sorta want to allow people to combine them to show the next in the chain.

core/src/com/unciv/models/ruleset/IConstruction.kt
```diff
@@ -141,12 +141,14 @@ class RejectionReason(val type: RejectionReasonType,
...
     )
     private val orderedImportantRejectionTypes = listOf(
+        RejectionReasonType.ShouldNotBeDisplayed,
         RejectionReasonType.WonderBeingBuiltElsewhere,
         RejectionReasonType.NationalWonderBeingBuiltElsewhere,
         RejectionReasonType.RequiresBuildingInAllCities,
         RejectionReasonType.RequiresBuildingInThisCity,
         RejectionReasonType.RequiresBuildingInSomeCity,
         RejectionReasonType.RequiresBuildingInSomeCities,
+        RejectionReasonType.CanOnlyBeBuiltInSpecificCities,
```
SeventhM, line 151: Please add this to `constructionRejectionReasonType` as well, since units will now be making use of it

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -288,7 +297,6 @@ enum class UniqueType(
     RequiresBuildingInAllCities("Requires a [buildingFilter] in all cities", UniqueTarget.Building),
     @Deprecated("as of 4.10.17", ReplaceWith("Only available <if [buildingFilter] is constructed in at least [positiveAmount] of [All] cities>"))
     RequiresBuildingInSomeCities("Requires a [buildingFilter] in at least [positiveAmount] cities", UniqueTarget.Building),
-    CanOnlyBeBuiltInCertainCities("Can only be built [cityFilter]", UniqueTarget.Building),
```
SeventhM, line 291: Please deprecate and add new unique, not remove/replace. Side tangent: I know the unique is similar OnlyAvailableWhen, but it probably should still be down here for consistency when reading through the list of uniques

SeventhM:
Also, according to the unit test, you missed a spot when editing the rulesets

core/src/com/unciv/models/ruleset/IConstruction.kt
itanasi: Good catch. Done

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
itanasi: Put back in old Unique with Depreciated notification. Can add in version once more stable

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -288,7 +293,12 @@ enum class UniqueType(
...
+    /** See [OnlyAvailable]. Doesn't restrict Upgrade/Transform pathways
+     */
+    CanOnlyBeBuiltInCertainCities("Can only be built", UniqueTarget.Building, UniqueTarget.Unit,
+        docDescription = "Meant to be used together with conditionals, like \"Can only be built <after adopting [policy]> <while the empire is happy>\". Will also NOT block Upgrade and Transform actions"),
+    @Deprecated("as of x.xx.xx", ReplaceWith("Can only be built <in [cityFilter] cities>"))
+    CanOnlyBeBuiltInCertainCitiesDep("Can only be built [cityFilter]", UniqueTarget.Building),
```
SeventhM, line 301: Naming nitpick
1. Adding "Dep" to older uniques is pretty rare. Most cases Are when the message of the new unique is almost exactly the same
2. The new unique no longer has a cityFilter, so "InCertainCities" seems a bit off as a name. If we're follow OnlyAvailableWhen as and example, why not "CanOnlyBeBuiltWhen"?

SeventhM: Also keep in mind that technically, these two uniques functions differently. The older one didn't block you if it didn't meet the conditionals and the current one does. E.g. "Can only be built [in all cities] <before the [Ancient era]>" will functionally do nothing while "Can only be built <in [all] cities> <before the [Ancient era]>" under this PR will block building the construction at all

SeventhM:
I also had a thought... Why do we have both `RejectionReasonType.CanOnlyBeBuiltInSpecificCities` and `RejectionReasonType.CannotBeBuilt`? Surely having both of  these is mostly redundant, no?

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
itanasi: 1. Yes, because the text is the same. But switching [cityFilter] to requires Conditionals. Was trying to match (extend) the original instead of create a new "Only buildable" (which is note a word) to mirror "Only available"
2. Will change name.
3. Good point to notify Modders about. Any conditionals will need to be inverted. Although arguably then the previous implementation doesn't make sense grammatically

itanasi: Actually, going to definitely move next to OnlyAvailable since those are general Construction Uniques instead of the Building-only Uniques.

itanasi:
> I also had a thought... Why do we have both `RejectionReasonType.CanOnlyBeBuiltInSpecificCities` and `RejectionReasonType.CannotBeBuilt`? Surely having both of these is mostly redundant, no?

Good question. Will have to look at the code

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
SeventhM: Well, the point previously about location is the automatic unique documentation goes in order from top to bottom with each unique in a category that hasn't been written yet. This rewrites the order of building uniques to be this new one, then cost uniques, then Unbuildable. While it makes sense in the code to be next to OnlyAvailableWhen, it doesn't make make sense due to this quirk of the documentation. Nitpick, though

itanasi: I see. And for whatever reason OnlyAvailable gets pulled to a Tech Unique? Hm.

Why aren't we just duplicating them in the categories? I suppose just makes the page extra long?

Looking at the page seems like Building takes precedent over Unit so this would remain in the Building category.

itanasi:
> > I also had a thought... Why do we have both `RejectionReasonType.CanOnlyBeBuiltInSpecificCities` and `RejectionReasonType.CannotBeBuilt`? Surely having both of these is mostly redundant, no?
> 
> Good question. Will have to look at the code
So the main difference is CannotBeBuilt is supposed to be a blanket block Civ-wide, and CanOnlyBeBuiltInSpecificCities is cannot be built because the requirements aren't met. Functionally no real difference right now. Especially since we filter out most of these. But I think keeping them distinct is still beneficial.

SeventhM:
> So the main difference is CannotBeBuilt is supposed to be a blanket block Civ-wide

Oh ok, I see... ... ...
Wait, but what's the difference between that and Unbuildable? The source? Because otherwise the rejection message is pretty vague, the more specific version already has it's own rejection type, neither shows in the city screen, and they share the same rejection categories

SeventhM (approved):

itanasi:
Unbuildable I assume is supposed to be universal across the game, such as Missionaries. CanNotBeBuilt comes from "Cannot build [baseUnitFilter] units" Global Unique which is for Civs. And yes functionally it's all the same if there aren't certain Conditionals and all that. This is likely legacy and due to weird nuances.

That being said, I can see why they'd be split out and if we add functionality to reveal more of these in the Construction Menu, the unique error messages would still be useful.

PR #11276, opened 2024-03-08
Events!

yairm210:
image

As discussed in Discord - Events!

- Each event has several options associated
- Options can have conditions under which they can appear
- Events are triggered via trigger unique, and *trigger events* via triggers, meaning they can be chained for a tree of decisions

Currently all events can be triggered many times, for initial "production" release I'll add a boolean field of "canTriggerMultipleTimes" or something, not sure what the default should be, we'll save the triggered events on the CivInfo and we'll add a new conditional of "if event X has triggered"


AdityaMH:
Imagine can see event contain event description with ~~glorious~~ image and 2 sec music.

PR #11278, opened 2024-03-09
Add Transform cost support

itanasi:
Custom movement cost code
Custom Stats cost code

image

TODO:
* Support Stats and Stockpile?
* Perhaps rework UnitActionModifiers to both validate and execute Stats adjustment


yairm210:
This should definitely be a generic UnitActionModifier, not something specific to Transform.

itanasi:
> This should definitely be a generic UnitActionModifier, not something specific to Transform.

Yeah that's what I realized and need to rework. Closing for now

PR #11280, opened 2024-03-09
Add the multiple comparison unique pack

PLynx01:
I made a multiple comparison unique pack, which contains 4 new uniques and 6 modified ones to support year as a parameter. Though i made a similar PR several days ago #11250 and the PR for support of years #11262 (which was rejected), I've combined these two pull requests into this one.

I hope that this time year-related uniques will be introduced, as there is another PR adding events #11276

If you have some remarks, share them in the comments. Constructive feedback welcome.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -71,6 +71,11 @@ object Conditionals {
             return relevantCity!!.predicate()
         }
 
+        fun checkOnResourceName(resourceName: String): Boolean {
+            if (gameInfo == null) return false
+            return gameInfo!!.ruleset.tileResources.containsKey(resourceName)
+        }
+
         /** Helper to simplify the "compare civ's current era with named era" conditions */
         fun compareEra(eraParam: String, compare: (civEra: Int, paramEra: Int) -> Boolean): Boolean {
             if (gameInfo == null) return false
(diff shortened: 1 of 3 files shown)


yairm210:
You're half-right here.
Yes, year should be part of the same check, since it's another int-resultant object.
No, we don't need to introduce extra uniques.

Let's looks at this from a language perspective first. "Between 1900 and 2000 year(s)" does not parse well. Years aren't something you *have*, but they are something you *count*.
So this should instead read as "When year is between 1900 and 2000".
Checking this against resources and stats shows that it's fine - "When gold is between 0 and 1000", "When iron is between 0 and 2".

Since that's the case, why not use the same unique for numbers as well? Base numbers are also a countable!

In fact this can be generalized to "when [countable] is between [countable2] and [countable3]"

Examples: When Gold is between Science and Culture, when Iron is between Horses and Gems. 
Okay, sounds pretty bad, if we add "number of" sounds better - when number of iron is between number of horses and number of gems. So that's something we'll have to figure out.

----

Second problem I have is with the if/elses. What we actually need here is a function to get a string, and return a number - which we can then use for both the first and second element.
That function will be an if/else style.

PLynx01:
> You're half-right here.
> Yes, year should be part of the same check, since it's another int-resultant object.
> No, we don't need to introduce extra uniques.
> 
> Let's looks at this from a language perspective first. "Between 1900 and 2000 year(s)" does not parse well. Years aren't something you *have*, but they are something you *count*.
> So this should instead read as "When year is between 1900 and 2000".
> Checking this against resources and stats shows that it's fine - "When gold is between 0 and 1000", "When iron is between 0 and 2".
> 
> Since that's the case, why not use the same unique for numbers as well? Base numbers are also a countable!
> 
> In fact this can be generalized to "when [countable] is between [countable2] and [countable3]"
> 

In that case we need to add the "countable" unique parameter

> Examples: When Gold is between Science and Culture, when Iron is between Horses and Gems. 
> Okay, sounds pretty bad, if we add "number of" sounds better - when number of iron is between number of horses and number of gems. So that's something we'll have to figure out.

In that case, please give me the desired wording of the uniques

> 
> Second problem I have is with the if/elses. What we actually need here is a function to get a string, and return a number - which we can then use for both the first and second element.
> That function will be an if/else style.

Did you mean this function?

fun getCountableNumber(countable: String): Int

PLynx01:
I've got an idea:

We can make a "countable" mega-parameter which would accept all of the following:

- Integer numbers
- Fractional numbers
- Current year
- Stat amounts
- Resources amounts
- Victory Screen stats (production, force, territory, population)
- Number of adopted policies
- Number of researched technologies
- Number of owned units
- Numbers of built buildings
- City population
- Number of cities

And even more...

There is a lot of these, but I think we need to add support of nested parameters. Something like:

When [number of cities <with population at least 5>] is larger than [4]

What's your opinion about this?

yairm210:
We'll need to see how we structure the nested filters
Definitely we'll need [] and not <> - conditionals are applied to entire unique and not to specific parts

PLynx01:
> We'll need to see how we structure the nested filters
> Definitely we'll need [] and not <> - conditionals are applied to entire unique and not to specific parts

[] - this is a parameter character
<> - this is a conditional character (for entire unique)

We need a "internal conditional" or "parameter conditional" character.

Let's decide together what it should be

PLynx01:
>  We really don't need such a thing

So, in that case, would we use [] for parameter conditionals?

The following conditional unique should be true if a player has more than 4 cities with population of at least 5

Please rewrite it with your preferred syntax.

<When [number of cities <with population at least 5>] is larger than [4]>

PLynx01:
For now, I can make a pack of countable-checking uniques, which does not involve nested conditions.

PR #11286, opened 2024-03-10
Allow follower uniques to generate resources

yairm210:


SeventhM:
Wouldn't using both getMatchingUniques function be slower than either just a new function or my solution (has a boolean for getMatchingUniques)? I'm not sure what this PR solves that mine doesn't and it seems likely slower

yairm210:
It simplifies the city getMatchingUniques function.
The rare edge-case where you need ALL (local&non-local) uniques from buildings, PLUS religion, should not be part of the base getMatchingUniques function

yairm210:
Speaking of simplifying... I think getLocalMatchingUniques and getMatchingLocalOnlyUniques are in fact the same exact function.

yairm210:
If I understand correctly that's another thing your PR did and if so that's 100% a change that we want!

SeventhM:
I feel it is likely still a worthwhile thing to include just in case it comes up in the future. I sorta have my doubts that it isn't a general problem (I'm pretty sure I accidentally created a second getLocalMatchingUniques function because I almost needed the other one to cover this usecase before). Besides my PR addresses a few other various things like deleting the second function and locking `getCityResourcesAvailableToCity` to just output city resources, which is relevant in the CityScreen when displaying the city resources the city has

SeventhM:
Otherwise we're talking two exactly the same PRs with the only difference being unit tests that I can copy easily without being at a computer

yairm210:
Hmm
Sure 👍🏿 
But please 

- include the test from this PR
- change civUniques to includeCivUniques
- remove the "if source is follower unique" check - that implies that civ.getMatchingUniques can return follower uniques, which it can't

PR #11291, opened 2024-03-11
New ActionModifiers to enable Stats and Minimum Movement

itanasi:
Add ActionModifier UnitActionStatsCost that allows optional Stats application to UnitActions.
* Food and Production come from/go to nearest City

Add ActionModifier UnitActionMovementCostRequired that can force a minimum amount of movement to do an Action
Add ActionModifier UnitActionMovementCostAll to use all remaining Movement

Also add validation that the ActionModifier effects can be executed and added to other UnitAction Uniques.

UnitActions can default to using All Movement points (such as Transform)

image

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
@@ -4,10 +4,11 @@ import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.models.ruleset.unique.Unique
 import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.models.stats.Stat
+import com.unciv.models.stats.Stats
 import com.unciv.models.translations.removeConditionals
 import com.unciv.models.translations.tr
 import com.unciv.ui.components.fonts.Fonts
-import kotlin.math.roundToInt
+import kotlin.math.ceil
 
 object UnitActionModifiers {
     fun canUse(unit: MapUnit, actionUnique: Unique): Boolean {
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
```diff
@@ -18,16 +21,69 @@ object UnitActionModifiers {
...
+            return unit.getMaxMovement()
         val movementCost = actionUnique.conditionals
             .filter { it.type == UniqueType.UnitActionMovementCost }
             .minOfOrNull { it.params[0].toInt() }
-        if (movementCost != null) return movementCost
-        return 1
+        val movementCostRequired = actionUnique.conditionals
+            .filter { it.type == UniqueType.UnitActionMovementCostRequired }
+            .minOfOrNull { it.params[0].toInt() }
+        if (movementCost != null && movementCostRequired != null) {
```
yairm210, line 33: Sounds to me like sequenceOf(a,b).filterNotNull.maxOrNull

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
```diff
@@ -18,16 +21,69 @@ object UnitActionModifiers {
...
+     * @return Boolean
+     */
+    private fun canSpendStatsCost(unit: MapUnit, actionUnique: Unique): Boolean {
+        for (conditional in actionUnique.conditionals.filter { it.type == UniqueType.UnitActionStatsCost }) {
+            for ((stat, value) in conditional.stats) {
+                if (stat in Stat.statsWithCivWideField) {
+                    if (!unit.civ.hasStatToBuy(stat, value.toInt()))
+                        return false
+                } else {
+                    if (unit.getClosestCity() != null) {
```
yairm210, line 64: Val closestCity = ..., Reverse if to reduce nesting - this tree is way too complex!

yairm210 (approved):
All in all looks good!!

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
itanasi: Yeah just realized since I added functionality to the City I can simplify this

itanasi: Done

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
itanasi: Reworked

PR #11293, opened 2024-03-11
Update Simplified Chinese

chr56:


yairm210:
Mmm there are actual differences in translation between you and @AutumnPizazz 
Not something I can auto-resolve

PR #11304, opened 2024-03-14
Fixed policies do not grant production bonus to wonders

woo1127:
## After:

<br><br/>

Saving a single turn when building wonders can be incredibly crucial, especially when playing on Deity mode 😫


core/src/com/unciv/models/ruleset/Building.kt
```diff
@@ -510,14 +510,13 @@ class Building : RulesetStatsObject(), INonPerpetualConstruction {
             name -> true
             "Building", "Buildings" -> !isAnyWonder()
             "Wonder", "Wonders" -> isAnyWonder()
-            "National Wonder" -> isNationalWonder
-            "World Wonder" -> isWonder
+            "National Wonder", "National" -> isNationalWonder
+            "World Wonder", "World" -> isWonder
```
yairm210, line 514: These require translations in template.properties

android/assets/jsons/Civ V - Gods & Kings/Policies.json
```diff
@@ -84,7 +84,8 @@
                 "name": "Republic",
                 "uniques": [
                     "[+1 Production] [in all cities]",
-                    "[+5]% Production when constructing [All] buildings [in all cities]"
+                    "[+5]% Production when constructing [All] buildings [in all cities]",
+                    "[+5]% Production when constructing [All] wonders [in all cities]"
```
yairm210, line 88: Is this how it was in the base game? That's our baseline here

woo1127: Yes I've double-checked with EmperorPenguin on Discord
he's definitely an experienced player in my view.

yairm210 (approved):

android/assets/jsons/Civ V - Gods & Kings/Policies.json
SeventhM: I question this. Did he confirm in Civ 5 or in base Deciv?

SeventhM:
@yairm210 I kinda need someone to double check this in actual Civ 5, because I have hard doubts that I works like this there. I'm certain I would've found at least some reference to it by now in some obscure way, but no. Not even in competitive Civ 5 circles or even as a mention in Lekmod

woo1127:
[Windmill - Copy.txt](https://github.com/yairm210/Unciv/files/14725033/Windmill.-.Copy.txt)

I have came across the XML file for the windmill online (I had to convert it to a TXT format since GitHub doesn't support XML). So by examining the file, we can noticed 
that "buildingProductionModifier" and "wonderProductionModifier" are distinct elements and specifically the 10% bonus is present only in the "buildingProductionModifier". So  If these two XML elements are also present in other XML files related to buildings or beliefs, it strongly prove that in Civ 5, buildings only encompass regular structures and not wonders. Therefore, possible that I made a mistake 😥, and you may closing this PR after reviewing this comment

SeventhM:
I'll close. Between the general lack of a confirmation (even in more niche locations), to the separation of values in the civ 5 xmls, to a few Civ 5 forums explicitly saying windmills don't benefit buildings, I don't think there's a reason to believe this PR is correcting things to the way they actually worked in Civ 5

PR #11305, opened 2024-03-14
AI doesn't settle very unfavorable locations

tuvus:
The AI sometimes settles in the middle of a desert or poles with very little resources around. This PR tries to fix that by adding a minimum value that a location must have in order for the AI to settle it. 

This value is currently set to 50. The picture below shows a location with a value of 45 so the settler would not settle here. Note that adding 2.5 more grassland tiles would convince the settlers to settle here. Any sort of resources would also convince them to settle.
Note that locations ranked above 100 are usually good locations to settle. The value can go to over 150.

SettleFilter3

Sometimes the value of a location can go negative (eg. between two cities). Before this PR the AI would still settle there. Now, the AI should leave more space (in some scenarios).
</details> 

<details><summary>More Pictures with the change</summary>

SettleFilter1
SettleFilter2
As shown in the picture above, the AI will still settle the tundra created by rivers in the snow area.

</details> 

While this does seem like a relatively simple idea, it might break a few mods. So maybe the value should be lower a lot more or just set to 0 to filter the negatively valued locations. Or we could make this value modable so some mods can override it.

The alternative would be to search a much greater distance or develop a complex heuristic as to where the settler should go for a better location.

Also, I need to work on something that can check if a city is close to the edge of a map, since that is also bad.

What are your thoughts?


yairm210:
This is a good idea, I've been auto ignoring it because it's a draft, but I like 👍🏿 
It's also small, and easily reversible if we see it causes problems, so I approve
@tuvus is there a reason for this to be draft?

yairm210 (approved):

tuvus:
Alrighty, looking at this again it seems like it is more needed than I thought.

I added a few more changes including the original.

- The AI doesn't put cities in the ice zones as much
- It doesn't settle on the borders of the map like before
- It also tries to settle 7 tiles away from nearby cities (the perfect distance with no tiles in between)
- The AI doesn't value building a city on the coast as much (since non resource water tiles can't have improvements)

**Before**

AISettleFilterBefore

**After**

AISettleFilterAfter

</details> 

I think modding this will probably have to go into a different PR as I would need to look into that more.

PR #11308, opened 2024-03-14
Countable comparison conditional uniques

PLynx01:
As I promised two days ago #11280 , I've made a pull request which contains five new uniques for comparing countables.
Countables can be numbers (both integer and float) and numeric variables. For now, only these variables are supported:

- Current year
- Stat amounts
- Resource amounts

If you have more ideas for the countables, give me a lot of them, as adding new countables is a quick and easy process.

I would also like to add two TriggerCondition uniques, which are true when the number of one countable becomes higher or lower than another countable. Please give me some advice about it.

Better yet, we can add support of nested parameters or even evaluation of mathematical expressions into numbers, either by using this [https://github.com/notKamui/Keval](https://github.com/notKamui/Keval) or similar library or our custom code for that task.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -100,22 +100,23 @@ object Conditionals {
         }
 
         fun getCountableAmount(countable: String): Float? {
-            if (countable.toIntOrNull() != null) return countable.toFloat()
             if (countable.toFloatOrNull() != null) return countable.toFloat()
-            if (countable == "year") return gameInfo!!.getYear(gameInfo!!.turns).toFloat()
 
             val relevantStat = Stat.safeValueOf(countable)
 
             if (relevantStat != null) {
-                if (relevantCity != null) {
-                    return relevantCity!!.getStatReserve(relevantStat).toFloat()
-                } else if (relevantStat in Stat.statsWithCivWideField) {
-                    return relevantCiv!!.getStatReserve(relevantStat).toFloat()
+                return if (relevantCity != null) {
+                    relevantCity!!.getStatReserve(relevantStat).toFloat()
+                } else if (relevantStat in Stat.statsWithCivWideField && relevantCiv != null) {
+                    relevantCiv!!.getStatReserve(relevantStat).toFloat()
                 } else {
-                    return null
+                    null
                 }
             }
 
+            if (gameInfo == null) return null
+
+            if (countable == "year") return gameInfo!!.getYear(gameInfo!!.turns).toFloat()
             if (gameInfo!!.ruleset.tileResources.containsKey(countable))
                 return getResourceAmount(countable).toFloat()
 
@@ -127,16 +128,24 @@ object Conditionals {
             second: String,
             compare: (first: Float, second: Float) -> Boolean): Boolean {
 
-            return if (getCountableAmount(first) != null && getCountableAmount(second) != null)
-                compare(getCountableAmount(first)!!, getCountableAmount(second)!!)
+            val firstNumber = getCountableAmount(first)
+            val secondNumber = getCountableAmount(second)
+
+            return if (firstNumber != null && secondNumber != null)
+                compare(firstNumber, secondNumber)
             else
                 false
         }
 
         fun compareCountables(first: String, second: String, third: String,
                               compare: (first: Float, second: Float, third: Float) -> Boolean): Boolean {
-            return if (getCountableAmount(first) != null && getCountableAmount(second) != null && getCountableAmount(third) != null)
-                compare(getCountableAmount(first)!!, getCountableAmount(second)!!, getCountableAmount(third)!!)
+
+            val firstNumber = getCountableAmount(first)
+            val secondNumber = getCountableAmount(second)
+            val thirdNumber = getCountableAmount(third)
+
+            return if (firstNumber != null && secondNumber != null && thirdNumber != null)
+                compare(firstNumber, secondNumber, thirdNumber)
             else
                 false
         }
--- a/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Conditionals.kt
@@ -99,6 +99,48 @@ object Conditionals {
             return compare(statReserve, lowerLimit * gameSpeedModifier, upperLimit * gameSpeedModifier)
         }
 
+        fun getCountableAmount(countable: String): Float? {
+            if (countable.toIntOrNull() != null) return countable.toFloat()
+            if (countable.toFloatOrNull() != null) return countable.toFloat()
+            if (countable == "year") return gameInfo!!.getYear(gameInfo!!.turns).toFloat()
+
+            val relevantStat = Stat.safeValueOf(countable)
+
+            if (relevantStat != null) {
+                if (relevantCity != null) {
+                    return relevantCity!!.getStatReserve(relevantStat).toFloat()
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -99,6 +99,48 @@ object Conditionals {
             return compare(statReserve, lowerLimit * gameSpeedModifier, upperLimit * gameSpeedModifier)
         }
 
+        fun getCountableAmount(countable: String): Float? {
+            if (countable.toIntOrNull() != null) return countable.toFloat()
```
yairm210, line 103: We don't actually need this line, since anything that can be converted to int can also be converted to float

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -99,6 +99,48 @@ object Conditionals {
...
+            if (countable.toFloatOrNull() != null) return countable.toFloat()
+            if (countable == "year") return gameInfo!!.getYear(gameInfo!!.turns).toFloat()
```
yairm210, line 105: A lot of assumptions in this dunction that things aren't null - we should be checking!
gameInfo, civInfo, etc - they can be null, what happens then?

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -99,6 +99,48 @@ object Conditionals {
...
+
+            return null
+        }
+
+        fun compareCountables(
+            first: String,
+            second: String,
+            compare: (first: Float, second: Float) -> Boolean): Boolean {
+
+            return if (getCountableAmount(first) != null && getCountableAmount(second) != null)
```
yairm210, line 130: more like
firstCountable = getCountableAmount(first)
secondCountable = getCountableAmount(second)

That way we only need to run the function once

Same for the rest of these

PLynx01:
OK, I've applied your suggestions. But there are more countables than just stats, resources and years. Should I left it as it is now, or should I add support for more countables? If so, please give me some ideas, so I can implement them.

PLynx01:
@yairm210 I've applied your suggestions. What else is necessary to merge my PR?

PLynx01:
@yairm210 Please review!

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -734,6 +734,13 @@ enum class UniqueType(
     ConditionalInRegionOfType("in [regionType] Regions", UniqueTarget.Conditional),
     ConditionalInRegionExceptOfType("in all except [regionType] Regions", UniqueTarget.Conditional),
 
+    /////// countables conditionals
+    ConditionalCountableEqualTo("when number of [countable] is equal to [countable]", UniqueTarget.Conditional),
+    ConditionalCountableDifferentThan("when number of [countable] is different than [countable]", UniqueTarget.Conditional),
+    ConditionalCountableGreaterThan("when number of [countable] is greater than [countable]", UniqueTarget.Conditional),
+    ConditionalCountableLessThan("when number of [countable] is less than [countable]", UniqueTarget.Conditional),
+    ConditionalCountableBetween("when number of [countable] between [countable] and [countable]", UniqueTarget.Conditional),
```
yairm210, line 742: "is between"

yairm210 (commented):
Other than one minor text issue, looks mergeable!!

PLynx01:
@yairm210 Check out my PRs. I've made some corrections

yairm210 (approved):

woo1127:
@PLynx01
`When number of [Iron] is between [Horse] and [Niter]` still sound weird to me, but if Yairm finds it acceptable, I have no issue with it either.

My actual inquiry is whether we can enable these unique features to accommodate `(modified by game speed)` or `(per turn)` ?

PLynx01:
> @PLynx01
> `When number of [Iron] is between [Horse] and [Niter]` still sound weird to me, but if Yairm finds it acceptable, I have no issue with it either.
> 
> My actual inquiry is whether we can enable these uniques to accommodate `(modified by game speed)` or `(per turn)` ?

What do you mean by saying (per turn)? I suppose you're talking about stat income or growth of resources per turn.

I am right?

woo1127:
> What do you mean by saying (per turn)? I suppose you're talking about stat income or growth of resources per turn.
> 
> I am right?

Yes that what I mean by `per turn`, the stat income or growth of resource per turn.
Because now all the conditional uniques only refer to the current accumulated amount of stat
For example `when above [100] [Gold]` is not the same as `when above [100] [Gold] per turn`

PLynx01:
> > What do you mean by saying (per turn)? I suppose you're talking about stat income or growth of resources per turn.
> > I am right?
> 
> Yes that what I mean by `per turn`, the stat income or growth of resource per turn. Because now all the conditional uniques only refer to the current accumulated amount of stat For example `when above [100] [Gold]` is not the same as `when above [100] [Gold] per turn`

In that case it should be parameter not a unique

woo1127:
> In that case it should be parameter not a unique

Then it would become an optional parameter, which differs than any other uniques that requires a mandatory parameter. Do you have any ideas on how to do? Will you implement it?

For instance, we could introduce the unique `Gain [3] [Culture] [modifierFilter]`, where modifierFilter accepts the options I previously mentioned:
- modified by game speed
- per turn
- none

What do you think ?

PLynx01:
@woo1127 
There is also another dimension: Scopes of countables

They are
- Global (all civilizations)
- Civ-wide
- City-wide
- Unit-wide

And also diplomacy-related scopes

PLynx01:
The template would be something like this:

when [scope] [countable] is larger than [scope] [countable]

@woo1127 
Any thoughts?

woo1127:
Can I have few examples to clarify what you mean ?

PLynx01:
> Can I have few examples to clarify what you mean ?
> We can just use cityFilter for civ and city wide while unitFilter for unit wide isn't it ?

That's a good idea. We can use filters to define scope. But we need to group countables by these scopes:

- Game-wide countables
- Civ-wide countables
- City-wide countables
- Unit-wide countables

PLynx01:
> > Can I have few examples to clarify what you mean ?
> > We can just use cityFilter for civ and city wide while unitFilter for unit wide isn't it ?
> 
> That's a good idea. We can use filters to define scope. But we need to group countables by these scopes:
> 
> - Game-wide countables
> - Civ-wide countables
> - City-wide countables
> - Unit-wide countables 

@woo1127 
Here is the list: (long but non-compete)

Game-wide:
- Current year
- Current turn
- Years since beginning
- Number of all civilizations ever present
- Number of current civilizations
- Number of cities of all civilizations
- Number of tiles (with tile filter)
- Global population 

Civ scope (supports Game-wide scope)
- Number of cities (with civ filter)
- Number of cities (with civ and city filter)
- Total stats per turn (stat filter, optionally civ and city filter)
- Total stockpiled stats (gold, science, culture, faith)
- Total stats ever produced
- Total resources currently stockpiled
- Total resources ever stockpiled
- Number of traded resources
- Number of wars
- Number of defensive pacts
- Number of science agreements
- Number of wonders built
- Number of units
- Civ population 

City-wide (supports civ and game wide)
- Number of buildings
- Number of wonders
- Stat production per turn
- Worked tiles
- City population
- Turns until population growth/construction complete/border expansion

Unit-wide (supports previous scopes)
- XP
- Base strength
- Max strength 
- HP
- Number of promotions

woo1127:
> @woo1127 Here is the list: (long but non-compete)

100 hours of work would be required to implement all of these 😅💀

PR #11310, opened 2024-03-14
Melee Escort Attacking Fix

tuvus:
Solves #11287. (This is the second option described in that issue; I thought it would be harder to implement)
This PR adds in an extra check for melee units' `getAttackableEnemies()`. A melee unit that is escorting may only attack a unit if the escorting unit can also move into the tile after it attacks.

Scout without escorting, should be able to attack

EscortAttackMovmentFix
Scout with escorting should not be able to attack

EscortAttackMovmentFix2
</details> 

Note that it doesn't matter if the melee unit can actually kill the enemy unit or not. Either way, it won't be eligible to attack unless the player stops escorting.

### DIFF
--- a/core/src/com/unciv/logic/battle/TargetHelper.kt
+++ b/core/src/com/unciv/logic/battle/TargetHelper.kt
@@ -30,7 +30,7 @@ object TargetHelper {
             if (unit.baseUnit.isMelee() && unit.isEscorting()) {
                 val escortingUnit = unit.getOtherEscortUnit()!!
                 if (!escortingUnit.movement.canReachInCurrentTurn(reachableTile)
-                    || escortingUnit.currentMovement - unit.getOtherEscortUnit()!!.movement.getDistanceToTiles()[reachableTile]!!.totalDistance <= 0f) 
+                    || escortingUnit.currentMovement - escortingUnit.movement.getDistanceToTiles()[reachableTile]!!.totalDistance <= 0f) 
                     continue
             }
             val tilesInAttackRange =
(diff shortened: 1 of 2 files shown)


AdityaMH:
add conditional when civilian unit has no movement.
if civilans should can moving as escort going to, even killing a unit. but if civilian unit doesn't have movement that a real problem.

core/src/com/unciv/logic/battle/TargetHelper.kt
```diff
@@ -25,11 +25,18 @@ object TargetHelper {
         val tilesWithEnemies: HashSet<Tile> = HashSet()
         val tilesWithoutEnemies: HashSet<Tile> = HashSet()
         for ((reachableTile, movementLeft) in tilesToAttackFrom) {  // tiles we'll still have energy after we reach there
+            // If we are a melee unit that is escorting, we only want to be able to attack from this
+            // tile if the escorted unit can also move into the tile we are attacking if we kill the enemy unit.
+            if (unit.baseUnit.isMelee() && unit.isEscorting()) {
+                val escortingUnit = unit.getOtherEscortUnit()!!
+                if (!escortingUnit.movement.canReachInCurrentTurn(reachableTile)
+                    || escortingUnit.currentMovement - unit.getOtherEscortUnit()!!.movement.getDistanceToTiles()[reachableTile]!!.totalDistance <= 0f)
```
yairm210, line 33:
```suggestion
                    || escortingUnit.currentMovement - escortingUnit.movement.getDistanceToTiles()[reachableTile]!!.totalDistance <= 0f) 
```

yairm210 (approved):
Looks good! Great that you managed to add a test for this situation, will help us in the future with both this situation and others like it :D

PR #11326, opened 2024-03-17
Giving the AI good trades is stored as credit

tuvus:
While trading with the AI, I am not always sure how much gold they want for each item. In order to trade for an item I have to give the AI more gold than it is actually worth or else the trade will take forever to get through.

This PR works on improving that by adding in "gift" relations. Anytime a civ gives the AI a good trade, the excess gold value will be stored as the "We appreciate your gifts" modifier. This gift is then used as credit for a future trade with the AI. 

- Declaring war also removes some of the credit.
- The credit does deteriorate over time.
- Credit determination can vary from 40 GPT to 2% of the value based on neutral relations. The relations can change this.
- Each credit shown as "We appreciate your gifts" amounts to around 20 gold without inflation or speed modifiers.
- The value of the credit varies based on the civ's perceived inflation which depends on their GPT.

After giving Babylon some gold, they now have the credit stored and will be willing to give some of it back in trades.

GoldGifting

</details> 

Also, more tests, these might be the first tests to include trading in them. I couldn't find any tests covering trades.


yairm210:
I'm fine with gifts, but this should *only* be if you're giving them something for free.
The point of the "you don't know how much they want" is to make it more like human negotiation. If you offer too much, that's part of the plan!

This is also why we don't have a "what would it take" button - I am fundamentally opposed, since that allows you to game the AI to give them not one gold more than the bare minimum.

tuvus:
> The point of the "you don't know how much they want" is to make it more like human negotiation.

The problem with that is that the AI doesn't negotiate in a very human-like manner (There is almost no way that it can act human-like in all situations). The AI is never desperate and can't understand when a trade that is usually slightly bad is actually really good. I lost where the post was (maybe on Discord), but someone said along the lines of, "Why should I trade with the AI if it always gives me a bad trade?" and this is one way to counteract that.

It might be better to make it percent-based. When a trade is accepted with both sides offering something, we take only X% of the extra value of one side to store as a gift. 
We could also make it moddable. The base ruleset can have this at 0% so that other mods can change it to a higher percentage to enable it.

Also, I forgot to mention, we might be able to have gifting units increase this gift value so that players can trade units for gold to the AI.

tuvus:
Ooops, that was an accident

tuvus:
I am not sure how to undelete a batch-delete operation. The changes can still be seen in the files changed luckily.

tuvus:
Wait, no I didn't! I only deleted the remote branch!

tuvus:
Alrighty, so there are two ways to preserve haggling and also allow for gifts in the same trade.

1. After a certain percentage of the original amount, the rest of the value of the trade will go to gifts. Ex: They have 100 gold of offers and we have 300 gold of offers. If the arbitrary percentage is 200%, 300 - 100 = 100 gold will count as gifts.
2. A certain percentage of extra gold is given as gifts. Ex: They have 100 gold of offers and we have 300 gold of offers. If the arbitrary percentage is 70%, 300 - 100 = 200 then we multiply the percentage: 200 * 70% = 140 gold that will count as gifts.
3. Implement both of these and make them moddable. Then decide on values that we want.

Note that gold gifts do deteriorate over time.

tuvus:
Actually, this PR fulfills a goal described in the implementation of diplomacy roadmap #4674.

>Trading - When you strike a deal with a civilization where they know that they're getting the better half of the deal, relations improve depending on how large of a benefit they get out of it. Comparable to the variability of gifting. (ND)

Issue #11269, opened 2024-03-06
Feature request: AutoPlay needs to be made concurrent


tuvus:
### Problem Description

Using AutoPlay in the later game causes the game to freeze since the AI play is being done on the same thread.

### Related Issue Links

#10315

### Desired Solution

The solution to the problem is to move the AutoPlay to another thread. However I believe this will cause a problem if we allow the player to continue making clicks and movements to the game. Therefore, I am led to believe a more elaborate movement-locking feature would need to be implemented.

yairm210:
Let's talk about next turn, since it had the same problem

How does next turn work?
- clone game state
- change worldscreen to be "read only" state - players can explore the current world but cannot take action
- process next turn for clone in another thread
- when other thread finishes, replace the world screen

Sounds like you want something similar - only in your case you can replace the world screen on every turn, but keep it "read only"

yairm210:
That is to say - add a onTurnComplete callback parameter to next turn
NextTurn should not be aware of UI directly, it just clones the game state on every turn end and calls the function - which can be on another thread
The latency cost of cloning the game state every turn is the major cost here

tuvus:
The way we made it "read only" in the past was by changing the viewing civ... Could we set the viewing civ to none and set the selected civ to the original civ like in Spectator? 
The spectator needs to view each civ as if they were them without being able to modify them. We might be able to reuse this to apply the locking mechanism.

yairm210:
No, we just set isPlayersTurn to false
https://github.com/yairm210/Unciv/blob/86f0ed6e85aae28aea6ccd3fc45e311dad793f38/core/src/com/unciv/ui/screens/worldscreen/WorldScreen.kt#L584

tuvus:
Why did I not realize this? AutoPlay already does that in order to lock the screen.

PR #11329, opened 2024-03-17
AutoPlayEndTurn can run on a different thread

tuvus:
Solves #11269
If there are more than 30 units and cities, the AutoPlayEndTurn button will run the operation on a different thread so the player can look at the world. I haven't expanded this to AutoPlayMilitary and the others yet.

Also, I fixed a small bug where AutoPlay did not want to build military units as much since it technically is the player.


core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -341,18 +341,21 @@ class GameSettings {
         var autoPlayDiplomacy: Boolean = true
 
         var turnsToAutoPlay: Int = 0
+        var autoPlaying: Boolean = false
```
yairm210, line 344: Is there a need for this? Isn't it false only when turnsToAutoPlay is zero?

tuvus: Yes, AutoPlayEndTurn and maybe in the future AutoPlay Military want to say that they are AutoPlaying to change the text on nextTurnButton without activating the multiturn AutoPlay.

yairm210: That's *super* confusing.
Let's define what exactly we mean by "is autoplaying" and use terms accordingly- if this is only used for displaying different text when we're not really autoplaying then we should handle this differently

yairm210: We have 2 kinds of autoplay - one which*passes the turn* and one that does actions*within a turn* automatically? Let's make that clear

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
```diff
@@ -166,7 +166,7 @@ class ConstructionAutomation(val cityConstructions: CityConstructions) {
                 && city.getCenterTile().getTilesInDistance(5).none { it.militaryUnit?.civ == civInfo })
             modifier = 5f // there's a settler just sitting here, doing nothing - BAD
 
-        if (civInfo.playerType == PlayerType.Human) modifier /= 2 // Players prefer to make their own unit choices usually
+        if (civInfo.playerType == PlayerType.Human && !GUI.getSettings().autoPlay.isAutoPlaying()) modifier /= 2 // Players prefer to make their own unit choices usually
```
SeventhM, line 169: I gotta say, We have to put this check (PlayerType.Human || isAutoPlaying()) alot. Can we get a val/helper function in Civilization to shorten this?

yairm210: We need to decide if autoplay is a *game* feature or a *UI* feature.
If the former, it should be under the worldscreen
If the latter, it should be under the gameInfo somewhere
It's really not a settings thing

SeventhM: > This one is for !isAutoPlaying()

I mean it's just the inverse of the typical check. ``PlayerType.Human && !isAutoplaying`` is the same as ``!PlayerType.AI && !isAutoplaying`` which is the same as just inversing the helper function that I'm requesting. And I'm mostly requesting it because it seems to me (especially after adding in the personality stuff or glancing over almost any AI code) that it comes up pretty often that we're checking whether or not a player is AI controlled, rather than if a player is actually an AI

tuvus:
I have just realized that WordScreen is not singleton. I just checked and GameInfo is also not close to singleton. I don't actually need a singleton class, however it does need to persist across the individual loads and saves on the same file. 

Maybe it should be associated with either UncivGame or Settings? Although it shouldn't really be on the singleton level. 

So maybe passing it into each new WorldScreen is the best option. I'll probably go with that.

tuvus:
Why doesn't Job.cancel() seem to work?

tuvus:
> but job would still need to use its CoroutineScope - like calling isActive on it and bailing if false.

Huh, I didn't think we would have to manually put code in to check if the thread is active since we are working on a copy. It should work if we just terminate the thread and it's progress since it should not be writing to any shared location until it completes.

I think this PR is ready to merge since canceling jobs can be in a different PR. Unless there are more comments of course.

tuvus:
Why did the version not update? I merged main in and now it's causing a problem?

tuvus:
Should be fixed now.

I'll probably leave the cancel early feature to another PR some other time since that seems like a big change.

yairm210 (approved):

PR #11334, opened 2024-03-18
Fix loop when improvement is unbuildable and removes features

### LINKED ISSUE #11332: StackOverflowError when building on unbuildable terrain (customs house on snow)

### Game Version

4.10.20

### Describe the bug

While playing a single-player game, the game suddenly crashed after building a customs house on a snow tile.

### Steps to Reproduce

1. Get a great person.
2. Move to a snow tile.
3. Create the great improvement.

I still have to thoroughly test this, but it was the first occurrence of such an event.

### Screenshots

_No response_

### Link to save file

https://gist.github.com/Bjastkuliar/ce292a30dbbd168341e6e7ff28443786

### Operating System

Windows

### Additional Information

I've just updated the game today and therefore it might be an issue of this specific version, I just downloaded the JAR file and swapped the old JAR in the unciv folder (I originally had downloaded the whole "windows edition", but I was using the JRE game, not the .exe

SeventhM:
Resolves #11332

### DIFF
--- a/core/src/com/unciv/logic/map/tile/TileInfoImprovementFunctions.kt
+++ b/core/src/com/unciv/logic/map/tile/TileInfoImprovementFunctions.kt
@@ -110,13 +110,14 @@ class TileInfoImprovementFunctions(val tile: Tile) {
 
             // Otherwise, we can if this improvement removes the top terrain
             if (!hasUnique(UniqueType.RemovesFeaturesIfBuilt, stateForConditionals)) return false
-            if (knownFeatureRemovals == null) return false
-            val featureRemovals = tile.terrainFeatures.map {
-                feature -> tile.ruleset.tileRemovals.firstOrNull{ it.name == Constants.remove + feature } }
-            if (featureRemovals.any { it != null && it !in knownFeatureRemovals }) return false
+            if (knownFeatureRemovals.isNullOrEmpty()) return false
+            val featureRemovals = tile.terrainFeatures.mapNotNull { feature -> 
+                tile.ruleset.tileRemovals.firstOrNull { it.name == Constants.remove + feature } }
+            if (featureRemovals.isEmpty()) return false
+            if (featureRemovals.any { it !in knownFeatureRemovals }) return false
             val clonedTile = tile.clone()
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/map/tile/TileInfoImprovementFunctions.kt
```diff
@@ -110,13 +110,14 @@ class TileInfoImprovementFunctions(val tile: Tile) {
...
             if (!hasUnique(UniqueType.RemovesFeaturesIfBuilt, stateForConditionals)) return false
-            if (knownFeatureRemovals == null) return false
-            val featureRemovals = tile.terrainFeatures.map {
-                feature -> tile.ruleset.tileRemovals.firstOrNull{ it.name == Constants.remove + feature } }
-            if (featureRemovals.any { it != null && it !in knownFeatureRemovals }) return false
+            if (knownFeatureRemovals.isNullOrEmpty()) return false
+            val featureRemovals = tile.terrainFeatures.mapNotNull { feature -> 
+                tile.ruleset.tileRemovals.firstOrNull { it.name == Constants.remove + feature } }
+            if (featureRemovals.isEmpty()) return false
+            if (featureRemovals.any { it !in knownFeatureRemovals }) return false
```
yairm210, line 117: What does it mean, if a feature removal is not "known"? It's in the ruleset, what are we actually checking here?

SeventhM: Which removals have we unlocked the techs for. Definition is further up; line 82

SeventhM: The check here is "of all of the removals that could apply, do we have all of their techs unlocked?" I would go "huh, wait, why don't we check that directly?" but it was like that before. I would assume there's a reason... Hopefully

yairm210 (approved):

core/src/com/unciv/logic/map/tile/TileInfoImprovementFunctions.kt
yairm210: Looks like this whole thing requires cleanup, not for this pr

PR #11344, opened 2024-03-20
Add UnitActionModifier for Stockpile Cost

itanasi:
Add Stockpile Resource Costs to any Action!
image
Also flip the signs for Action cost to be more consistent. Positive value for Movement means it will cost X Movement, similarly Positive Food will cost (remove) that amount of Food

### DIFF
--- a/core/src/com/unciv/models/stats/Stats.kt
+++ b/core/src/com/unciv/models/stats/Stats.kt
@@ -186,9 +186,9 @@ open class Stats(
     }
 
     /** Return a string of just +/- value and Stat symbol*/
-    fun toStringOnlyIcons(plusSign: Boolean = true): String {
+    fun toStringOnlyIcons(addPlusSign: Boolean = true): String {
         return this.joinToString {
-            (if (plusSign && it.value > 0) "+" else "") + it.value.toInt() + " " + it.key.character
+            (if (addPlusSign && it.value > 0) "+" else "") + it.value.toInt() + " " + it.key.character
         }
     }
 
--- a/core/src/com/unciv/models/stats/Stats.kt
+++ b/core/src/com/unciv/models/stats/Stats.kt
@@ -186,9 +186,9 @@ open class Stats(
     }
 
     /** Return a string of just +/- value and Stat symbol*/
-    fun toStringOnlyIcons(): String {
+    fun toStringOnlyIcons(plusSign: Boolean = true): String {
         return this.joinToString {
-            (if (it.value > 0) "+" else "") + it.value.toInt() + " " + it.key.character
+            (if (plusSign && it.value > 0) "+" else "") + it.value.toInt() + " " + it.key.character
         }
     }
 
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
@@ -67,16 +67,9 @@ object UnitActionModifiers {
         for (conditional in actionUnique.conditionals.filter { it.type == UniqueType.UnitActionStockpileCost }) {
             val amount = conditional.params[0].toInt()
             val resourceName = conditional.params[1]
-            if (!unit.civ.resourceStockpiles.containsKey(resourceName) || unit.civ.resourceStockpiles[resourceName] < amount) {
-//                 if (!unit.civ.resourceStockpiles.containsKey(resourceName)) {
-//                     println("cannot find $resourceName")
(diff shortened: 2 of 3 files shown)


core/src/com/unciv/models/stats/Stats.kt
```diff
@@ -186,9 +186,9 @@ open class Stats(
     }
 
     /** Return a string of just +/- value and Stat symbol*/
-    fun toStringOnlyIcons(): String {
+    fun toStringOnlyIcons(plusSign: Boolean = true): String {
         return this.joinToString {
```
yairm210, line 190: Rename to addPlusSign

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
```diff
@@ -140,7 +159,14 @@ object UnitActionModifiers {
...
             for (conditional in actionUnique.conditionals.filter { it.type == UniqueType.UnitActionStatsCost })
                 statCost.add(conditional.stats)
-            effects += (statCost * -1).toStringOnlyIcons()
+            effects += statCost.toStringOnlyIcons(false)
+        }
+
+        if (actionUnique.conditionals.any { it.type == UniqueType.UnitActionStockpileCost }) {
+            var stockpileString = ""
+            for (conditionals in actionUnique.conditionals.filter { it.type == UniqueType.UnitActionStockpileCost })
+                stockpileString += " ${conditionals.params[0].toInt()} [${conditionals.params[1]}]"
```
yairm210, line 168: Will not translate well, replace with curly brackets

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
```diff
@@ -140,7 +159,14 @@ object UnitActionModifiers {
...
+            effects += stockpileString.drop(1) // drop leading space
```
yairm210, line 169: RemovePrefix, for clarity

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
```diff
@@ -63,13 +63,27 @@ object UnitActionModifiers {
         return true
     }
 
+    private fun canSpendStockpileCost(unit: MapUnit, actionUnique: Unique): Boolean {
+        for (conditional in actionUnique.conditionals.filter { it.type == UniqueType.UnitActionStockpileCost }) {
+            val amount = conditional.params[0].toInt()
+            val resourceName = conditional.params[1]
+            if (unit.civ.getCivResourcesByName()[resourceName]!! < amount) {
```
yairm210, line 70: A malformed resourceName will cause crash, I think we have a safe version? If not then do like 'val amount = ... ?: return false'

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
```diff
@@ -97,6 +111,11 @@ object UnitActionModifiers {
                         } else unit.getClosestCity()?.addStat(stat, -value.toInt())
                     }
                 }
+                UniqueType.UnitActionStockpileCost -> {
+                    val amount = conditional.params[0].toInt()
+                    val resourceName = conditional.params[1]
+                    unit.civ.resourceStockpiles.add(resourceName, -amount)
```
yairm210, line 117: Check for valid resource?

core/src/com/unciv/models/stats/Stats.kt
itanasi: Done

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
itanasi: Check again?

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
itanasi: Done

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionModifiers.kt
itanasi: Found the safety version

yairm210 (approved):

PR #11352, opened 2024-03-22
Fix loop when AI is trying to remove an improvement with the same name as a terrain feature

### LINKED ISSUE #11350: Crash at EoT

### Game Version

4.10.21

### Describe the bug

Testing out 4.10.21. Began making updates, appears I triggered a glitch at some point. Hitting Next turn goes until about 2/3, grinds, crashes. 

### Steps to Reproduce

Will have to step backwards to see if it was a mod update that triggered the crash. Reporting for now. 

### Screenshots

Screenshot_20240321-131030_Unciv

### Link to save file

[England - 67 turns.txt](https://github.com/yairm210/Unciv/files/14705440/England.-.67.turns.txt)

### Operating System

Android

### Additional Information

For save games, not copy-pastes - Friendly reminder that github does not accept extension-less files which do not appear in Androids' upload selection. Finding the original save file requires Total Commander app to gain access to Unciv data storage.

SeventhM:
Fixes related bug: AI couldn't evaluate removals that removed improvements

Fixes related bug: AI evaluates logic on the actual tile and not a clone. Idk if this actually is an issue yet, but it could be with that removal. Here, it actually tries to remove a feature that was never there originally, and would've added it back in. Though I guess it doesn't matter this time

Fixes #11350

Man, this scenario is a bit convoluted almost solely because the improvement shares a name. At the very least, there's at least some level of justification for fixing it and not saying this is just a mod error

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
@@ -368,7 +368,7 @@ class WorkerAutomation(
         return ruleSet.tileImprovements[improvementString] // For mods, the tile improvement may not exist, so don't assume.
     }
 
-    private fun getImprovementRanking(tile: Tile, unit: MapUnit, improvement: String, localUniqueCache: LocalUniqueCache): Float {
+    private fun getImprovementRanking(tile: Tile, unit: MapUnit, improvementName: String, localUniqueCache: LocalUniqueCache): Float {
         val improvement = ruleSet.tileImprovements[improvementName]!!
 
         // Add the value of roads if we want to build it here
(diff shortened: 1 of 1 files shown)


hackedpassword:
Oh sorry @SeventhM I had a post on #11350 queued all day and was getting back to finishing it, didn't see your update til submitted as closed with the issue looking the same as the older Z2 perma-crash scenario.

"Boulder" is a terrain feature, and it is also an improvement... an attempt to produce an improvement with the attributes of a terrain feature. Boulder can be built as an obstruction, but as of current not against movement, only against tile improvements.

Thank you for finding this! Since the improvement name can't really be renamed mid-game, I'll wait for the next update to test. :)

SeventhM:
> an attempt to produce an improvement with the attributes of a terrain feature.

I'll have to check at some point, but I don't think that actually works. The game usually views terrain features and improvements differently

yairm210 (approved):

Issue #11264, opened 2024-03-05
(RFC)?: List of problems for ConstructionAutomation


SeventhM:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

Right now I want to try to focus in on cleaning up ConstrcutionAutomation, or at least get the issues with the it documented to work on for later. Right now there's a few issues with the class
- [ ] Ais solely looks at cost when deciding to build something. For example, If one building costs 1 and gives 1 culture, it will always be preferred over a building that costs 3 and gives a thousand culture. That's clearly a problem
- [ ] Combined stats of a building aren't considered. A building with 2 science and 3 production is seen equal to a building with just 2 science when checking science and to a building with just 3 production when checking production
- [ ] AIs don't consider production boosts for building them. I know I attempted to fix that in #11254, but that only fixes the end choice. I.e., a faith building with a production boost is picked over an equally costed production building without one. But a faith building with a production boost won't be picked over an equally costed one without a production boost
- [ ] AIs don't build buildings that give cities additional health
- [ ] AIs don't consider the benefit of Courthouse styled buildings. Instead they generally assume the benefit is generally good, even in cities that would otherwise be unaffected. Ideally, this should instead be treated as a general stat increase of happiness
- [ ] AIs don't consider population increases
- [ ] There is a need to genericize some of the wonder logic
- [ ] Stat increases are treated via `isStatRelated`. This accidentally includes percentage increases, which generally are different
- [ ] AIs don't consider stat increases from other scources (policies, nation, other buildings). Just from the construction itself
- [ ] Unlike tile considerations, AIs do not consider if they are running out of gold as a reason to build a gold building
- [ ] Unlike tile considerations, AIs do not consider lower cities for a larger need for food buildings
- [ ] AIs don't consider specialists. This one is a bit lower priority imo, bit still worth looking into eventually
- [ ] AIs don't consider tile stat increases. Most notably, AIs don't notice if a building increases the stats of a tile that's not near any city. This could be an expensive thing to check, however
- [ ] This file assumes production, making it unusable for gold buying, despite the obvious code overlap in the two ideas

If there's other issues, please point them out. Hopefully we can get the issues crossed off as we go

<details>
<summary></summary>

### Desired Solution

Fixes to the above

[...]

PR #11370, opened 2024-04-01
Allow AI to consider building stats more accurately

SeventhM:
Address a part of #11264. Namely, it makes it so that it considers buildings by what it expects the stat change will be. There likely are some cracks this has that the old version didn't (e.g., this PR no longer considers if a building will give stats to a tile), but this should overall make the AIs smarter than before. It should also integrate better with Personalities

### DIFF
--- a/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
@@ -20,6 +20,7 @@ import com.unciv.models.ruleset.unique.StateForConditionals
 import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.stats.Stat
+import kotlin.math.ceil
 import kotlin.math.max
 import kotlin.math.sqrt
 
@@ -103,15 +104,8 @@ class ConstructionAutomation(val cityConstructions: CityConstructions) {
     fun chooseNextConstruction() {
         if (cityConstructions.getCurrentConstruction() !is PerpetualConstruction) return  // don't want to be stuck on these forever
 
-        addFoodBuildingChoice()
-        addProductionBuildingChoice()
-        addGoldBuildingChoice()
-        addScienceBuildingChoice()
-        addHappinessBuildingChoice()
         addDefenceBuildingChoice()
         addUnitTrainingBuildingChoice()
-        addCultureBuildingChoice()
-        addFaithBuildingChoice()
         addOtherBuildingChoice()
 
         if (!city.isPuppet) {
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
```diff
@@ -334,77 +315,57 @@ class ConstructionAutomation(val cityConstructions: CityConstructions) {
...
+                stat == Stat.Science && civInfo.wantsToFocusOn(Victory.Focus.Science) ||
+                stat == Stat.Faith && civInfo.wantsToFocusOn(Victory.Focus.Faith) ||
+                stat == Stat.Gold && civInfo.wantsToFocusOn(Victory.Focus.Gold)
+                )
+                buildingStats[stat] *= 2f
+
+            buildingStats[stat] *= personality.scaledFocus(PersonalityValue[stat])
+        }
 
+        building.happiness /= 3 // Avoid overshooting that may happen from the following rank
```
yairm210, line 354: Happiness is usually VERY important, there's a reason that we mark it as 3 times the regular stat. I wouldn't consider this overshooting, if it is - we should be changing the rankStatsValue, because that overshoot would apply everywhere (specialists, tiles to work, improvements, etc)

core/src/com/unciv/logic/city/CityStats.kt
```diff
@@ -633,4 +633,13 @@ class CityStats(val city: City) {
...
     //endregion
+
+    fun getStatDifferenceFromBuilding(building: String): Stats {
+        val newCity = city.clone()
+        newCity.setTransients(city.civ)
+        newCity.cityConstructions.builtBuildings.add(building)
+        newCity.cityConstructions.setTransients()
+        newCity.cityStats.update(updateCivStats = false)
+        return newCity.cityStats.currentCityStats - currentCityStats
+    }
```
yairm210, line 644: NICE!

core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
SeventhM: > Happiness is usually VERY important, there's a reason that we mark it as 3 times the regular stat

It's not uncommon for higher level AIs to be left with something like +20 happiness thanks to the many multipliers they get. There, happiness genuinely doesn't matter nearly as much

> because that overshoot would apply everywhere (specialists, tiles to work, improvements, etc)

Also no. It actually is only used in places of placing new cities and building improvements. Tiles to work/specialists to work actually has their own way of ranking that's separate. Tbh, I'm only using rankStatsValue for consistency and to future proof for tourism, but even here it's kinda not even that necessary to take into account. I could be willing to edit rankStatsValue (it basically only comes up for placing cities... I think... due to very few improvements, even in mods, actually affecting happiness)

yairm210: If so, we should add that as a precondition - if civ happiness is larger than some metric (say, max(15, numOfCities)), then divide by 3; if larger than half that, divide by 2; else leave as-is

SeventhM: Done in my last commit... Eh... On second thought, not sure I like it. As I said, tiles to work/specialists to work uses it's own value system, and I based most of the value system here off of it, only actually using the prebuilt Ranking for any broad strokes things or for specifically viewing gold less than other stats. As such, line 331 already tries to take this into account and I'm now overshooting again (x6 value instead of the intended x3 value). For me the question would be "do we always care about happiness the same way", to which the answer in Unciv is yes, but in Civ 5 is no (Civ 5 has a distinction between local happiness and global). So if we made changes to be 100% consistent with Civ 5, it actually is better to consider the happiness city by city, rather than with the global modifier

Fun fact, btw: tiles to work also overshoots happiness by accident, intending to value it at x2 when unhappy and x1 otherwise, but having the check twice by accident, leading to it actually being x4. It's actually not the only accidental mistake I noticed while glancing at it

yairm210 (approved):

SeventhM:
Did a closer look at the AI. Only one of these I want to address now, the rest could probably be noted in #11264 for later
1. I forgot to give a value for buildings that carry over population. I do actually have a formula for that, though it would be far more approximate than accurate
2. Interestingly, the fallback default value of 0.6 comes up sometimes. Sometimes it feels like a fine fallback, sometimes I kinda feel it's a bit too high. That leads me to
3. There may be adjustment needed in the future to get other buildings much more lined up in value with this newer function. Minimum, we're still valuing military or defense buildings separately when they may affect stats (often via maintenance, which is actually accounted for by "accident"). This is especially true for figuring out how important we value buildings vs units

PR #11372, opened 2024-04-01
Allow AI civilian unit to consider other triggerable uniques

woo1127:
Currently, the AI automation for civilian units only support a limited range of unit actions associated with the base ruleset, such as conducting trade routes or spreading religion but it does not consider for other triggerable uniques like free unit appears or gain free building. Consequently, moddable civilian units that have these uniques will remain idle and done nothing. This PR has address the problem.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
@@ -7,6 +7,7 @@ import com.unciv.logic.map.tile.Tile
 import com.unciv.models.UnitActionType
 import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActions
+import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsFromUniques
 
 object CivilianUnitAutomation {
 
@@ -105,7 +106,6 @@ object CivilianUnitAutomation {
                 return
         }
 
-
         // This has to come after the individual abilities for the great people that can also place
         // instant improvements (e.g. great scientist).
         if (unit.hasUnique(UniqueType.ConstructImprovementInstantly)) {
@@ -116,6 +116,10 @@ object CivilianUnitAutomation {
             if (!improvementCanBePlacedEventually)
                 UnitActions.invokeUnitAction(unit, UnitActionType.StartGoldenAge)
         }
+
+        // if none of the conditions above are met,
+        // AI will attempt to activate the first triggerable unique whenever it is available.
+        UnitActionsFromUniques.getTriggerUniqueActions(unit, unit.getTile()).firstOrNull()?.action?.invoke()
 
         // TODO: The AI tends to have a lot of great generals. Maybe there should be a cutoff
         //  (depending on number of cities) and after that they should just be used to start golden
--- a/core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
@@ -5,7 +5,9 @@ import com.unciv.logic.civilization.managers.ReligionState
 import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.Tile
 import com.unciv.models.UnitActionType
+import com.unciv.models.ruleset.unique.UniqueTriggerActivation
 import com.unciv.models.ruleset.unique.UniqueType
+import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionModifiers
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActions
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsFromUniques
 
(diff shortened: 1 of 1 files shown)


SeventhM:
Unless there's something mechanical here, my opinion is neither approving or denying this PR. But imo, this feels... Unfinished? Here, let's go over a few complicated scenarios
1. This misses Scientists getting extra science (or BNW Writers once we work on that getting extra culture) since technically those don't use triggers. And that's probably one of the more important things missing right now compared to most of the triggers
2. Most things regarding resources has extra rules regarding them. Triggering the first ability you can likely would be questionable. Obviously, this would be hard to cover correctly, so leaving it as a stub is fine, but I view it as seemingly another reason extra logic would be nice to have
3. I feel like logic for adding buildings to cities would likewise want its own coverage to match up better with how they would be expected to be used
4. Multiple stat gaining triggers may exist on one unit. While this is a more edge case, it's possible for one to be strictly better than another, and this wouldn't catch that

I almost feel like a more involved PR, split across multiple PRs might get the job done better and give more time actually acknowledging what specific conditions various triggers might actually want

woo1127:
> Unless there's something mechanical here, my opinion is neither approving or denying this PR. But imo, this feels... Unfinished? Here, let's go over a few complicated scenarios

I concur with everything you've said. Myself too have consider about the scenarios you mentioned. 
Currently, at least this single line of code has let moddable unit of AI to does something, whereas before it will just remained idle based on my observation.
I would consider to implement scenario 1 and 3 but for 2 and 4 I have no clue how to do.

core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
```diff
@@ -117,6 +123,33 @@ object CivilianUnitAutomation {
...
+                    && (unit.movement.canMoveTo(it.getCenterTile()) || unit.currentTile == it.getCenterTile())
+            }.minByOrNull {
+                val path = unit.movement.getShortestPath(it.getCenterTile())
+                path.size
+            }
+
+            if (cityToGainBuilding != null) {
+                if (unit.currentTile == cityToGainBuilding.getCenterTile()) {
+                    UniqueTriggerActivation.triggerUnique(unique, unit.civ, unit = unit, tile = unit.currentTile)
+                    UnitActionModifiers.activateSideEffects(unit, unique)
```
yairm210, line 141: Instead, we should activate the unit action, which does all this

core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
```diff
@@ -117,6 +123,33 @@ object CivilianUnitAutomation {
...
+                    return
+                }
+                else unit.movement.headTowards(cityToGainBuilding.getCenterTile())
+            }
+            return
+        }
+
+        // if none of the conditions above are met,
+        // AI will attempt to activate the first triggerable unique whenever it is available.
+        UnitActionsFromUniques.getTriggerUniqueActions(unit, unit.getTile()).firstOrNull()?.action?.invoke()
```
yairm210, line 151: Remove this - we should be implementing case-by-case, and this catchall trigger could lead to some really bad behaviour. I'd rather an AI doesn't know what it should do with a unit, than it activates its ability randomly
E.g. great generals that can create citadels - "create at first possible instance" is worse than "never create"

core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
```diff
@@ -117,6 +123,33 @@ object CivilianUnitAutomation {
...
+            // Choose the city that is closest in distance and does not have the building constructed.
+            val cityToGainBuilding = unit.civ.cities.filter {
+                !it.cityConstructions.containsBuildingOrEquivalent(buildingName)
+                    && (unit.movement.canMoveTo(it.getCenterTile()) || unit.currentTile == it.getCenterTile())
+            }.minByOrNull {
+                val path = unit.movement.getShortestPath(it.getCenterTile())
+                path.size
+            }
+
+            if (cityToGainBuilding != null) {
```
SeventhM, line 138: To avoid mid modders trying to take advantage of this, I feel we should maybe make use of the building rank function I added in ConstructionAutomation. If that function needs to be moved to Automation, that's fine

core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
woo1127: > Instead, we should activate the unit action, which does all this

You mean replace it with the code `UnitActions.invokeUnitAction(unit, UnitActionType.TriggerUnique)` ?
But this would again only activate the **first** unit action, not necessarily the gain building action.

core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
woo1127: I agree now; I don't want the unit to blindly remove buildings or turn tiles to something else

core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
woo1127: > I feel we should maybe make use of the building rank function I added in ConstructionAutomation. 

Where is the building rank function you mentioned 👀 ? I only saw `chooseNextConstruction` function

SeventhM: You'll have to merge the current master into this PR. Or, actually, this isn't necessarily blocking, so you can work on that as a separate PR

yairm210 (approved):

PR #11383, opened 2024-04-03
Fixed Ships not being able to move into one size lakes connected by a city

tuvus:
Related to #11148

The original problem lies in a peace of optimization code where a water unit won't move to a single water tile surrounded with land tile that has a city bridging it to the ocean.
This version simply uses canPassThrough() instead of isWater(). I also generalized it to be applicable to land units as well since it doesn't necessarily have to be water units.

<details><summary>After Picture</summary>

SingleLakeTile

</details>


yairm210:
It's still an optimization! It saves BFS 
In fact it's now an optimization that may be applicable for many other kinds of units 🤔

yairm210 (approved):

PR #11396, opened 2024-04-05
Civs with a spy set up in a foreign city can view some information

tuvus:
Part of #7610

If a Civ has a spy set up in a foreign city, then it can view it's stats, building, specialists, religion and priority. Note that the Civ can't make any changes just as a Civ can't when it isn't their turn.

Pressing the arrow keys beside the city name will cycle through all foreign cities with a spy in them instead of their own cities.

The spying Civ will only be able to view the tiles around the city that they know about.

Arrow buttons by city name go to other cities that are being spied on.

SpyViewCity

Spy table showing that there is a spy set up in Pyongyang

SpyViewCity2

Picture showing that the player Civ Japan cannot see all tiles around the city.

SpyViewCity3

This picture shows that the city view matches what the spying player can view. I also tested this to work for spectator.

SpyViewCity4

</details>


core/src/com/unciv/logic/civilization/managers/EspionageManager.kt
```diff
@@ -70,5 +70,11 @@ class EspionageManager : IsPartOfGameInfoSerialization {
         return spyList.filter { it.getLocation() == city }.toMutableList()
     }
 
+    /**
+     * Returns a list of all cities with our spies in them.
+     * The list needs to be stable accross calls on the same turn.
+     */
+    fun getCitiesWithOurSpies(): List<City> = spyList.filter { it.isSetUp() && it.getLocation() != null }.map { it.getLocation()!! }
```
yairm210, line 77:
```suggestion
    fun getCitiesWithOurSpies(): List<City> = spyList.filter { it.isSetUp() }.mapNotNull { it.getLocation() }
```

core/src/com/unciv/ui/components/tilegroups/CityButton.kt
```diff
@@ -531,7 +531,8 @@ class CityButton(val city: City, private val tileGroup: TileGroup) : Table(BaseS
             // second tap on the button will go to the city screen
             // if this city belongs to you and you are not iterating though the air units
             if (DebugUtils.VISIBLE_MAP || viewingPlayer.isSpectator()
-                || (belongsToViewingCiv() && !tileGroup.tile.airUnits.contains(unitTable.selectedUnit))) {
+                || (belongsToViewingCiv() && !tileGroup.tile.airUnits.contains(unitTable.selectedUnit))
+                || (city.civ.gameInfo.isEspionageEnabled() && viewingPlayer.espionageManager.getSpyAssignedToCity(city)?.isSetUp() == true)) {
```
yairm210, line 535: Remove extraneous parentheses

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
```diff
@@ -60,8 +61,19 @@ class CityScreen(
         const val wltkIconSize = 40f
     }
 
+    private val selectedCiv: Civilization = GUI.getWorldScreen().selectedCiv
```
yairm210, line 64: I don't like this, CityScreen should get as a parameter the viewing civ, just like other screens do

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
```diff
@@ -60,8 +61,19 @@ class CityScreen(
...
+
+    private val isSpying = selectedCiv.gameInfo.isEspionageEnabled() && selectedCiv != city.civ
```
yairm210, line 66: What happens if we are a different civ, but espionage is not enabled...?

core/src/com/unciv/ui/components/tilegroups/CityButton.kt
tuvus: Interesting how you prefer relying on the && having a higher precedence over using parenthesis.

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
tuvus: The default, they won't be seeing the city screen in the first place. foreignCityInfoPopup() will be called instead.

ForignCityView


</details>

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
tuvus: The passing in a parameter alternative would require us to pass through an extra selectedCiv parameter to WorldTileGroup, WorldMapHolder, TileGroup, TileLayer, TileLayerCityButton, CityButton just so that it can be passed to city screen.

Are the comments at WorldScreen line 392-397 still relevant? Weird how sometimes it passes selectedCiv and other times viewingCiv.

core/src/com/unciv/ui/components/tilegroups/CityButton.kt
yairm210: More like, the code shows that it's separated into OR lines, so you can intiut that each OR line functions as expected...?
YMMV, I personally find it more readable without, but that may be an acquired taste...

core/src/com/unciv/ui/screens/cityscreen/CityScreen.kt
yairm210: Not sure, that's a comment from 5 years ago so who knows?

yairm210 (approved):

PR #11401, opened 2024-04-06
Espionage Uniques, Buildings and Policy

tuvus:
Part of #7610

This PR adds 5 new Espionage Uniques:
- SpyEffectiveness (unused)
- EnemySpyEffectiveness
- HiddenWithoutEspionage
- OneTimeSpiesLevelUp
- OneTimeGainSpy

4 new buildings (none have any pictures yet):
- Constabulary
- Police Station
- National Intelligence Agency (National Wonder)
- Great Firewall (Wonder)

Police State policy gains "[25]% reduction in enemy spy effectivenes [in all cities]"

And a few tests

The current way I have effectiveness calculated is a little weird, I think we will want to change it in the future (maybe even in this PR).

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Policies.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Policies.json
@@ -577,7 +577,8 @@
                 "name": "Police State",
                 "uniques": [
                     "[+3 Happiness] from every [Courthouse]",
-                    "[+100]% Production when constructing [Courthouse] buildings [in all cities]"
+                    "[+100]% Production when constructing [Courthouse] buildings [in all cities]",
+                    "[25]% reduction in enemy spy effectivenes [in all cities]"
                 ],
                 // There are also some uniques regarding espoinage, which as of this writing is not yet implemented
                 "requires": ["Militarism"],
--- a/android/assets/jsons/Civ V - Gods & Kings/Policies.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Policies.json
@@ -578,7 +578,7 @@
                 "uniques": [
                     "[+3 Happiness] from every [Courthouse]",
                     "[+100]% Production when constructing [Courthouse] buildings [in all cities]",
-                    "[25]% reduction in enemy spy effectivenes [in all cities]"
+                    "[-25]% enemy spy effectiveness [in all cities]"
                 ],
                 // There are also some uniques regarding espoinage, which as of this writing is not yet implemented
                 "requires": ["Militarism"],
--- a/android/assets/jsons/Civ V - Gods & Kings/Policies.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Policies.json
@@ -580,7 +580,6 @@
                     "[+100]% Production when constructing [Courthouse] buildings [in all cities]",
                     "[-25]% enemy spy effectiveness [in all cities]"
                 ],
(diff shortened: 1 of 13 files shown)


tuvus:
> You have +100% enemy has -50% -> should cancel out IMHO.

Huh, but were adding here, the result should be 150% since it is -50% and not 50%.

Also I realized spy effectiveness should not affect the outcome, only the time it takes to get to the outcome.

Also my version of Uniques.md is LF and not CRLF, is there a way to change this?

tuvus:
So basically ever operation should be: pastValue *= (100 + [X]%) / 100?
That also doesn't work with [+25]% and [-25]% = 100% when it is really 93.75%
Ahh, but [+33]% and [-25]% ~= 100%. 33% was the previous spy rank increase and it says in the guide that one spy rank = 1 defensive building with [-25]% decrease.
I guess that's probably the best bet?

SeventhM:
`Promotes all spies`: If I remember correctly, in Civ 5 new spies (from later eras/replacement spies) also come back leveled up

> Which is how all text files should be. I believe Studio has a setting, per type?

From my experience working from working on a keyboard more recently, this setting doesn't matter. Not sure if GitHub changes it to CRLF or what's going on, but even if you commit as LF in Studio, it'll fetch from GitHub as CRLF anyways. Though... This could also be because of how I move files around

android/assets/jsons/Civ V - Gods & Kings/Policies.json
```diff
@@ -577,7 +577,8 @@
                 "name": "Police State",
                 "uniques": [
                     "[+3 Happiness] from every [Courthouse]",
-                    "[+100]% Production when constructing [Courthouse] buildings [in all cities]"
+                    "[+100]% Production when constructing [Courthouse] buildings [in all cities]",
+                    "[-25]% enemy spy effectiveness [in all cities]"
                 ],
                 // There are also some uniques regarding espoinage, which as of this writing is not yet implemented
```
SeventhM, line 583: Is this line still relevant?

tuvus:
So they just multiply everything, okay.

SeventhM:
We're still missing checking for `OneTimeSpiesLevelUp` for new spies/replacement spies. Either that or you need to create a new unique to cover these scenarios

core/src/com/unciv/ui/objectdescriptions/TechnologyDescriptions.kt
```diff
@@ -309,20 +309,21 @@ object TechnologyDescriptions {
...
                         && !it.hasUnique(UniqueType.HiddenFromCivilopedia)
             }
     }
 
-    private fun getNukeAndReligionSwitches(civInfo: Civilization?): Pair<Boolean, Boolean> {
-        if (civInfo == null) return true to true
-        return civInfo.gameInfo.run { gameParameters.nuclearWeaponsEnabled to isReligionEnabled() }
+    private fun getNukeAndReligionSwitches(civInfo: Civilization?): Triple<Boolean, Boolean, Boolean> {
+        if (civInfo == null) return Triple(true, true, true)
+        return civInfo.gameInfo.run { Triple(gameParameters.nuclearWeaponsEnabled, isReligionEnabled(), isEspionageEnabled()) }
```
yairm210, line 326: Turn this into a data class, 3 things already needs naming

yairm210 (approved):
Looks mergeable, we can add other parts later

Issue #11388, opened 2024-04-04
Feature request: Add a new unique as "Cancel [policy] adopted"


AutumnPizazz:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

It has been issued in #3242 , but I have to hand it up here as I need it rapidly.
Guess it may be done easily so is it okay in next version? NEED HELP!

### Desired Solution

Nothing.

### Alternative Approaches

Nothing.

AutumnPizazz:
I have little regard for the cost of policy progress. If it must be considered, the suggestion is to cancel a policy that has been implemented and then recalculate (or refresh) the cultural expenditure required for the next policy.

PR #11405, opened 2024-04-06
Added triggerable unique to remove policy

woo1127:
As per @AutumnPizazz's request in #11388, this unique will remove a specific policy while leaving unaffected any subsequent policies that have this policy as a prerequisite. If a policy branch has been completed, removing the policy will revert the branch to incomplete.
This unique has been thoroughly tested and work well either as a building unique or when combined with UnitActionModifier to trigger a unit action.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -295,6 +295,23 @@ object UniqueTriggerActivation {
                     true
                 }
             }
+            UniqueType.OneTimeRemovePolicy -> {
+                val policyName = unique.params[0]
+                if (!civInfo.policies.isAdopted(policyName)) return null
+                val policy = civInfo.gameInfo.ruleset.policies[policyName] ?: return null
+
+                return {
+                    civInfo.policies.removePolicy(policy)
+
+                    val notificationText = getNotificationText(
+                        notification, triggerNotificationText,
+                        "You loss the [$policyName] Policy"
+                    )
+                    if (notificationText != null)
+                        civInfo.addNotification(notificationText, PolicyAction(policyName), NotificationCategory.General, NotificationIcon.Culture)
+                    true
+                }
+            }
             UniqueType.OneTimeEnterGoldenAge, UniqueType.OneTimeEnterGoldenAgeTurns -> {
                 return {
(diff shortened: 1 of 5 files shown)


AutumnPizazz:
Thanks a lot！૧(●´৺`●)૭

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -295,6 +295,23 @@ object UniqueTriggerActivation {
                     true
                 }
             }
+            UniqueType.OneTimeRemovePolicy -> {
+                val policyName = unique.params[0]
+                if (!civInfo.policies.isAdopted(policyName)) return null
+                val policy = civInfo.gameInfo.ruleset.policies[policyName] ?: return null
+
+                return {
+                    civInfo.policies.removePolicy(policy)
```
SeventhM, line 304: Personal request: Can this make use of filters? There is actually a good usecase here for filters with removing whole branches if a user wants to switch branches (such as a mod trying too hard to emulate BNW)

woo1127:
> Looks nice. One suggestion to think about: I try to Kdoc **all** new _public_ methods or fields.

Kdor all public methods or fields ?! Even those with straightforward names such as adopt or removePolicy ?

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
woo1127: If I understood your words correctly, you are suggesting that we could implement a policyFilter that accepts either a policy name or a policy branch name as a parameter ?

That's a good idea, I would consider implement it in another PR

yairm210 (approved):
LGTM

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
SeventhM: Well, that, an the usual option of filtering uniques. Then this could just remove all policies that fit the filter. Then only real limitation would be that the filter for the branch itself can't be the name of the branch because that refers to the opener policy

yairm210:
This is good enough to merge. It can be improved, but those improvements can definitely come in a phase 2 PR, and aren't IMO blockers for this functionality

PR #11407, opened 2024-04-07
Fix softlock for spectator with free policies

SeventhM:
Btw, this feels like the wrong approach. I'm pretty sure if you have a free policy, the menu will always force you to pick a policy, so disabling the close button at all is unnecessary (and is annoying if you want to confirm sometime before picking). But I want to request if removing these lines is the route we want to go down before doing it, and this seems like the simpler fix


SeventhM:
Just realized initial patch didn't work because spectator claims to be viewing as the other player. Even more reason to just remove this code and be done with it

yairm210 (approved):

yairm210:
I'm fine with allowing players to close the policy scree, if they have free policies, I'm pretty sure we only remove the free policy after one has been picked so they shouldn't be losing out on anything

PR #11415, opened 2024-04-07
Implement Same-majority-religion-based diplomatic modifier

TommasoPetrolito:
@yairm210 Hi, this is my very fist attempt at implementing an actual feature (my previous contributions where mainly translations to Italian), so, I'm sorry if I might have not strictly followed some guidelines or standards.
My Pull Request is meant to add handling for specific Diplomatic Modifier for "we believe in the same religion".
I have found this guide that is quite complete for Civ5 Diplomacy:

https://forums.civfanatics.com/resources/diplomacy-interacting-with-civs-and-positive-negative-diplomacy-modifiers-bnw.25611/
> -5|You have adopted their Religion in the majority of your Cities

Then I put +5 and considered only the "positive" modifier if the majority religion is the same and considered no negative modifier for the opposite situation (majority religion is different).

The need for this feature comes from #4674, and more specifically:
> Sharing Religion - If at least half of their cities share your Religion, it has similar effects to the lower spectrum of providing help to them.

Let me know if there is any specific steps further I should follow (I have debugged it a bit using the debugger and the logic seems be working, even if it is a bit complicated to reach the condition/situation starting a game from scratch, which I did) I don't know if you have suggestions about how to test it more efficiently.
   
Also: will the Language assets, and in particular properties lines requiring for custom translations, be created automatically by some specific procedure? Should I add a line in some template file to let it propagate to all Language-specific properties files with `#Requires translation` line comment?
   
   Thank you very much in advance.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -278,16 +278,11 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
     }
 
     private fun believesSameReligion(): Boolean {
-        // what is the majority religion of civInfo?
-        val civMajorityReligion = civInfo.religionManager.getMajorityReligion()
-        // if civInfo has a majority religion
-        return if (civMajorityReligion==null) {
-            // if civInfo hasn't a majority religion
-            false
-        } else {
-            // if civInfo's majority religion is also majority religion of otherCiv (Boolean)
-            otherCiv().religionManager.isMajorityReligionForCiv(civMajorityReligion)
-        }
+        // what is the majorityReligion of civInfo? If it is null, we immediately return false
+        val civMajorityReligion = civInfo.religionManager.getMajorityReligion() ?: return false
+        // if not yet returned false from previous line, return the Boolean isMajorityReligionForCiv
+        // true if majorityReligion of civInfo is also majorityReligion of otherCiv, false otherwise
+        return otherCiv().religionManager.isMajorityReligionForCiv(civMajorityReligion)
     }
 
     /** Returns the number of turns to degrade from Ally or from Friend */
@@ -602,14 +597,12 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
     }
 
     internal fun setReligionBasedModifier() {
-        if (civInfo.getDiplomacyManager(otherCiv()).believesSameReligion()) {
+        if (civInfo.getDiplomacyManager(otherCiv()).believesSameReligion())
             // they share same majority religion
             setModifier(DiplomaticModifiers.BelieveSameReligion, 5f)
-        }
-        else {
+        else
             // their majority religions differ or one or both don't have a majority religion at all
             removeModifier(DiplomaticModifiers.BelieveSameReligion)
-        }
     }
 
     fun denounce() {
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -277,30 +277,18 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
         }
     }
 
-    private fun believesSameReligion(): Boolean? {
-        // believing same/different religion - modifiers
+    private fun believesSameReligion(): Boolean {
         // what is the majority religion of civInfo?
         val civMajorityReligion = civInfo.gameInfo.religions.values.firstOrNull {
             civInfo.religionManager.isMajorityReligionForCiv(it)
         }
         // if civInfo has a majority religion
-        if (civMajorityReligion!=null) {
-            val otherCivMajorityReligion = otherCiv().gameInfo.religions.values.firstOrNull {
-                otherCiv().religionManager.isMajorityReligionForCiv(it)
-            }
-            // if otherCiv has a majority religion
-            return if (otherCivMajorityReligion!=null) {
-                // if otherCiv's majority religion equals civInfo's religion return "true", "else" otherwise
-                (civMajorityReligion.name == otherCivMajorityReligion.name)
-            }
-            // if otherCiv hasn't a majority religion
-            else {
-                null
-            }
-        }
-        // if civInfo hasn't a majority religion
-        else {
-            return null
+        return if (civMajorityReligion==null) {
+            // if civInfo hasn't a majority religion
+            false
+        } else {
+            // if civInfo's majority religion is also majority religion of otherCiv (Boolean)
+            otherCiv().religionManager.isMajorityReligionForCiv(civMajorityReligion)
         }
     }
 
@@ -616,19 +604,13 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
     }
 
     internal fun setReligionBasedModifier() {
-        val shareReligionTrueFalseOrNull = civInfo.getDiplomacyManager(otherCiv()).believesSameReligion()
-        if (shareReligionTrueFalseOrNull!=null) {
-            if (shareReligionTrueFalseOrNull) {
-                // they share same majority religion
-                setModifier(DiplomaticModifiers.BelieveSameReligion, 5f)
-            }
-            else {
-                // their majority religions differ
-                removeModifier(DiplomaticModifiers.BelieveSameReligion)
-            }
+        val shareReligionBoolean = civInfo.getDiplomacyManager(otherCiv()).believesSameReligion()
+        if (shareReligionBoolean) {
+            // they share same majority religion
+            setModifier(DiplomaticModifiers.BelieveSameReligion, 5f)
         }
         else {
-            // one or both civs don't have a majority religion at all
+            // their majority religions differ or one or both don't have a majority religion at all
             removeModifier(DiplomaticModifiers.BelieveSameReligion)
         }
     }
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -604,8 +604,7 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
     }
 
     internal fun setReligionBasedModifier() {
-        val shareReligionBoolean = civInfo.getDiplomacyManager(otherCiv()).believesSameReligion()
-        if (shareReligionBoolean) {
+        if (civInfo.getDiplomacyManager(otherCiv()).believesSameReligion()) {
             // they share same majority religion
             setModifier(DiplomaticModifiers.BelieveSameReligion, 5f)
         }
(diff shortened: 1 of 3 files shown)


yairm210:
Idea looks good, implementation is in the correct place!
Requires a bit of cleaning up but otherwise good to go!

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -276,6 +277,33 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
         }
     }
 
+    private fun believesSameReligion(): Boolean? {
```
yairm210, line 280: There's no reason for a null, we don't treat it differently, should be just boolean

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -276,6 +277,33 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+        // believing same/different religion - modifiers
+        // what is the majority religion of civInfo?
+        val civMajorityReligion = civInfo.gameInfo.religions.values.firstOrNull {
+            civInfo.religionManager.isMajorityReligionForCiv(it)
+        }
+        // if civInfo has a majority religion
+        if (civMajorityReligion!=null) {
```
yairm210, line 287: Reverse this if, so we have "if null return false", fast exit :)

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -276,6 +277,33 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
         }
     }
 
+    private fun believesSameReligion(): Boolean? {
+        // believing same/different religion - modifiers
+        // what is the majority religion of civInfo?
+        val civMajorityReligion = civInfo.gameInfo.religions.values.firstOrNull {
+            civInfo.religionManager.isMajorityReligionForCiv(it)
```
yairm210, line 284: Do we not have a function for "get majority religion" in religion manager?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -276,6 +277,33 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
+        }
+        // if civInfo has a majority religion
+        if (civMajorityReligion!=null) {
+            val otherCivMajorityReligion = otherCiv().gameInfo.religions.values.firstOrNull {
+                otherCiv().religionManager.isMajorityReligionForCiv(it)
+            }
+            // if otherCiv has a majority religion
+            return if (otherCivMajorityReligion!=null) {
```
yairm210, line 292: Same here - if other has no majority, return false immediately

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -587,6 +615,23 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
         }
     }
 
+    internal fun setReligionBasedModifier() {
+        val shareReligionTrueFalseOrNull = civInfo.getDiplomacyManager(otherCiv()).believesSameReligion()
+        if (shareReligionTrueFalseOrNull!=null) {
```
yairm210, line 620: Since null and false are treated the same this will turn into a simple if/else

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
TommasoPetrolito: I'll check.
Also, should I rather implement such in there (religion manager), if we don't?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
TommasoPetrolito: OK thank you!
About the null, you are totally right! I had put a distinction between true, false and null to be eventually able to use the same function to provide negative modifiers if the majority religions are different (not if there is no majority religion at all), that's why I had implemented it like this. 
Then I realized that was not the case of a negative modifier as well but just left it like this.
I don't know if it can be useful in the future for other more detailed diplomatic feature, but for now I agree, this granularity is not needed.
Thank you, I will fix it!

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
TommasoPetrolito: We do, only City-wise in the CityReligionManager.

TommasoPetrolito:
@yairm210 while I'm there to implement your feedbacks, should I also do something specific to allow translations to appear in the language .properties files?

yairm210:
No need, TranslationFileWriter handles that automagically

TommasoPetrolito:
@yairm210 I have implemented the feedbacks, so it should be ready. I simplified a lot the "believesSameReligion" function, stated that we want only a True/False, I don't check anymore what is the majority religion of OtherCiv iterating again all the religions, I just check if civInfo's majority religion is otherCiv's majority religion, avoiding to iterate other religions that are not the civInfo's majority one.

yairm210 (approved):
Looks good!

TommasoPetrolito:
> > will the Language assets, and in particular properties lines requiring for custom translations, be created automatically by some specific procedure
> 
> Rules of thumb: If it's a name in a json - 99% automatic. Others - large enums are if we felt it would be worth it. This one is automatic [because of these lines here](https://github.com/yairm210/Unciv/blob/e34854bf7d49c404a971fe069e1a74729dc59e85/core/src/com/unciv/models/translations/TranslationFileWriter.kt#L133-L135)... If one is _not_ automatic, there's still the half-automation that you only add to [template.properties](https://github.com/yairm210/Unciv/blob/master/android/assets/jsons/translations/template.properties), its entries are propagated each release cycle to all languages.

Thank you very much @SomeTroglodyte for the clarification.
Should this addition to  [template.properties](https://github.com/yairm210/Unciv/blob/master/android/assets/jsons/translations/template.properties) be logically part of this PR or I can add it separatedly later only if it doesn't show up "automagically" in the language .properties files 😅?

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
TommasoPetrolito: > "but"...
> ```kotlin
>     fun getMajorityReligion(): Religion? {
>         val counter = Counter<Religion>()
>         for (city in civInfo.cities)
>             city.religion.getMajorityReligion()?.let { counter.add(it, 1) }
>         return counter.maxByOrNull { it.value }?.takeIf { it.value > civInfo.cities.size / 2 }?.key
>     }
> ```
> in ReligionManager would likely be quite a bit more performant: 1 instead of N cities-loops.

Thank you 🙂, I will use this instead. I had looked for such a function but had not found it.

core/src/com/unciv/logic/civilization/managers/ReligionManager.kt
```diff
@@ -478,6 +478,27 @@ class ReligionManager : IsPartOfGameInfoSerialization {
...
+        // let's get the max-counted Religion if there is one or null otherwise
+        val maxReligionCounterEntry = religionCounter.maxByOrNull { it.value }
+        // if maxReligionCounterEntry is not null AND maxReligionCounterEntry > half-cities-count we return the Religion of maxReligionCounterEntry
+        return if ((maxReligionCounterEntry != null) && (maxReligionCounterEntry.value > civInfo.cities.size / 2)) {
+            maxReligionCounterEntry.key
+        } else {
+            // if maxReligionCounterEntry is null or <= half-cities-count we just return null
+            null
+        }
+    }
```
TommasoPetrolito, line 501: I hope that I got right the meaning of your code snippet @SomeTroglodyte 
I tried to convert it into a readability focused implementation. Let me know if you have any feedbacks!

core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
TommasoPetrolito: > Mind you - I tend to write a little too "kotlin-y" on quick&dirty stuff, and think about readability later. Resolving `?.let` and/or `?.takeIf` to step-by-step verbose code might be easier on the eye - I'll trust you with that task.

I hope your trust was well placed 🙂
You can have a look to the new version of the code.
Thank you very much again in advance.

core/src/com/unciv/logic/civilization/managers/ReligionManager.kt
TommasoPetrolito, line 501: I hope that I got right the meaning of your code snippet @SomeTroglodyte 
I tried to convert it into a readability focused implementation. Let me know if you have any feedbacks!

TommasoPetrolito:
> 
> Feel free to ignore most or all - just style stuff no effect on bytecode.

I ended up ignoring none! Thank you very much for the very detailed feedback, I'm not only newcomer at implementing within Unciv project but also newcomer to Kotlin in general, so that all your inputs have been very much appreciated. I didn't know "Elvis" but I like it, in other languages you can obtain more or less the same behavior even if with different syntax.
I have learned a lot from this PR.

> Now the minor nitpicks - code style - no biggies - wouldn't bother if you hadn't emphasized the newcomer thing:
> 
> * `civMajorityReligion==null`: blanks around operators -> https://kotlinlang.org/docs/coding-conventions.html#horizontal-whitespace. I often insist - not because grammar nazi but because fixing these later, even if an editor does it automatically, can make merges more complex.

Changed to `civMajorityReligion == null`
> * `} else {`: yes we're not C/C++ - but you got another one of them line-broken after the `}` - But also "we're not C": `if statement else statement` without curlies is just fine in kotlin if no distractions mar readability - a matter of taste. Sometimes statement + comments is enough to make redundant curlies the nicer read...

I opted for removing curlies but keeping indentation and comments, basically everywhere in this PR code changes
> * `val thing = someSource(); if (thing == null) return foo else return moreProcessing(thing)` can be `val thing = someSource() ?: return foo; return moreProcessing(thing)` - imagine the line breaks. Your call what is nicer to read. With `moreProcessing` several steps the elvis notation quickly becomes the winner IMHO. And the `val` changes type - it's non-nullable in the Elvis case, can be an advantage some rare times.

I opted for Elvis in believesSameReligion, I think the line is not long too much to make it necessary with an  indentation for this case
> * Same in the `getMajorityReligion` city loop (well done on the un-shorthandification!): `?: continue` fits nicely. Saves indentation and a `}` line. Just don't "hide" it at the end of a _long_ line - on its own line, indented it's more obvious in some cases.

Also here with the `continue` I opted for Elvis, and here as well I think the line is not too much long
> * Elvis might even make the end return clearer if maxReligionCounterEntry gets a non-nullable type - probably not, but imagine before discarding.

Here again I found your suggestion valuable, because it allowed me to reverse the if condition that now is much more straight-forward, while previously it was somehow the exact inversion of the most intuitive way to imagine that condition (if null or minority then null, otherwise maxReligionCounterEntry.key, not exactly the immediate thinking)


You can have a look the new version of the code.
Thank you very much again.

PR #11419, opened 2024-04-08
Added mod constant and unique to adjust air unit capacity

woo1127:
In the game by default, the maximum air unit capacity in a city is limited to only 6. This PR added a new mod constant to store this value and a new global unique for modders to adjust it.

The new unique is tested by including it in the following JSON files:
- GlobalUniques.json: `[+2] air unit capacity [in all cities]`
- **Air Port** in Buildings.json: `[+4] air unit capacity [in this city]`

Result:
gameplay picture

- Beijing has air unit capacity of 12 (default 6 + 2 from global + 4 from air port)
- Shanghai has air unit capacity of 8 (default 6 + 2 from global)

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -403,9 +403,14 @@ class City : IsPartOfGameInfoSerialization {
 
     fun canPlaceNewUnit(construction: BaseUnit): Boolean {
         val tile = getCenterTile()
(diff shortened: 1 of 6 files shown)


woo1127:
> No need for a new ModOptions constant - just make that one a virtual zero and add `[+6] air unit capacity [in all cities]` to GlobalUnique ?

That's a good idea, but I scare this will cause all base ruleset type mods to broke, and all modder have to realise and add this line into their mod to solve the issue

PR #11420, opened 2024-04-08
Added mod constant and unique to adjust air unit capacity

woo1127:
In the game by default, the maximum air unit capacity in a city is limited to only 6. This PR added a new mod constant to store this value and a new global unique for modders to adjust it.

The new unique is tested by including it in the following JSON files:
- GlobalUniques.json: `[+2] air unit capacity [in all cities]`
- **Air Port** in Buildings.json: `[+4] air unit capacity [in this city]`

Result:
gameplay picture

- Beijing has air unit capacity of 12 (default 6 + 2 from global + 4 from air port)
- Shanghai has air unit capacity of 8 (default 6 + 2 from global)

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -403,9 +403,14 @@ class City : IsPartOfGameInfoSerialization {
 
     fun canPlaceNewUnit(construction: BaseUnit): Boolean {
         val tile = getCenterTile()
+        val airUnitCapacity = tile.getCity()?.civ!!.gameInfo.ruleset.modOptions.constants.airUnitCapacity
+        val airUnitCapacityFromUniques = getMatchingUniques(UniqueType.AirUnitCapacity)
+            .filter { matchesFilter(it.params[1]) }
+            .sumOf { it.params[0].toInt() }
+
         return when {
             construction.isCivilian() -> tile.civilianUnit == null
-            construction.movesLikeAirUnits() -> tile.airUnits.count { !it.isTransported } < 6
+            construction.movesLikeAirUnits() -> tile.airUnits.count { !it.isTransported } < (airUnitCapacity + airUnitCapacityFromUniques)
             else -> tile.militaryUnit == null
         }
     }
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -403,14 +403,10 @@ class City : IsPartOfGameInfoSerialization {
 
     fun canPlaceNewUnit(construction: BaseUnit): Boolean {
         val tile = getCenterTile()
-        val airUnitCapacity = tile.getCity()?.civ!!.gameInfo.ruleset.modOptions.constants.airUnitCapacity
-        val airUnitCapacityFromUniques = getMatchingUniques(UniqueType.AirUnitCapacity)
-            .filter { matchesFilter(it.params[1]) }
-            .sumOf { it.params[0].toInt() }
 
         return when {
             construction.isCivilian() -> tile.civilianUnit == null
-            construction.movesLikeAirUnits() -> tile.airUnits.count { !it.isTransported } < (airUnitCapacity + airUnitCapacityFromUniques)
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/logic/city/City.kt
```diff
@@ -403,9 +403,14 @@ class City : IsPartOfGameInfoSerialization {
 
     fun canPlaceNewUnit(construction: BaseUnit): Boolean {
         val tile = getCenterTile()
+        val airUnitCapacity = tile.getCity()?.civ!!.gameInfo.ruleset.modOptions.constants.airUnitCapacity
```
yairm210, line 406: No need to go through the tile, we're already in the city

yairm210: Also we don't need to get the uniques for every unit, only for air units, so for performance reasons this should be in the 'when'

core/src/com/unciv/logic/map/mapunit/movement/UnitMovement.kt
```diff
@@ -602,7 +602,13 @@ class UnitMovement(val unit: MapUnit) {
     private fun canAirUnitMoveTo(tile: Tile, unit: MapUnit): Boolean {
         // landing in the city
         if (tile.isCityCenter()) {
-            if (tile.airUnits.filter { !it.isTransported }.size < 6 && tile.getCity()?.civ == unit.civ)
+            val airUnitCapacity = tile.getCity()?.civ!!.gameInfo.ruleset.modOptions.constants.airUnitCapacity
```
yairm210, line 605: Let's put this logic in one single place, probably on tile

Also, you can't be sure this is a city, so !! WILL throw an exception on aircraft carriers

core/src/com/unciv/logic/city/City.kt
woo1127: You are right 😀

core/src/com/unciv/logic/map/mapunit/movement/UnitMovement.kt
woo1127: > Let's put this logic in one single place, probably on tile
Done 👍

> Also, you can't be sure this is a city, so !! WILL throw an exception on aircraft carriers
I have tested it, and the carrier doesn't crash the game. I think the reason is because of the if statement if (tile.isCityCenter) which ensures that the tile is a city center and not a carrier."

woo1127:
> True, so to get that code simplicity one would have to add a lot of complex code for warnings and migration-phase defaults. Pity.

So should we safely ignore this for now 🥺
To prevent turning things become more complicated

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -560,6 +560,15 @@ class Tile : IsPartOfGameInfoSerialization {
     fun getTilesInDistanceRange(range: IntRange): Sequence<Tile> = tileMap.getTilesInDistanceRange(position, range)
     fun getTilesAtDistance(distance: Int): Sequence<Tile> =tileMap.getTilesAtDistance(position, distance)
 
+    fun getAirUnitCapacity(): Int {
+        val airUnitCapacity = this.getCity()!!.civ.gameInfo.ruleset.modOptions.constants.airUnitCapacity
```
yairm210, line 564: Don't assume we have a city! In canPlaceNewUnit you can call this on a non-city tile!

yairm210: If the entire function is based on cities only, this should be a function in the *city* - but instead maybe it should contain logic for both city and non-city (carriers) and this remain here? In second thought that makes a lot of sense

woo1127: Perhaps I should just move this function to city ?
Because we already have the unique for carrier

yairm210: That works

woo1127:
Draft as I am rushing with my final year project💻😖 , will come back soon next month. Apologies for any inconvenience caused 🙏

PR #11425, opened 2024-04-09
Fix personality being ignored for tile rankings in small cities

SeventhM:
In general, if a mad modder wants to break how they have personalities function for small city, that's the modder's fault. It's not like we can fully account for this anyways thanks to the various tile focuses. The extra multipliers we give here should overpower these focuses anyways unless someone gives a civ a negative focus

Also, it doesn't matter for the happiness multiplier if we're checking specialists or not because the check was in both places anyways, making the check redundant


core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -94,23 +93,25 @@ object Automation {
...
-                yieldStats.science *= 2
-            yieldStats.science *= civPersonality.scaledFocus(PersonalityValue.Science)
+        }
 
-            yieldStats.production *= civPersonality.scaledFocus(PersonalityValue.Production)
-            yieldStats.faith *= civPersonality.scaledFocus(PersonalityValue.Faith)
-            yieldStats.food *= civPersonality.scaledFocus(PersonalityValue.Food)
+        for (stat in Stat.values()) {
+            if (
+                stat == Stat.Culture && city.civ.wantsToFocusOn(Victory.Focus.Culture) ||
```
yairm210, line 106: There's a function to convert etat to victory type, no? We can make this concise

SeventhM: No, I don't think there is. Also, there's not a focus for every stat (missing happiness and food) so such a  function would have to be in Civilization, not as a a function to convert like I created for personality types

SeventhM:
Failure seems espionage related. Out of scope here

core/src/com/unciv/logic/automation/Automation.kt
SeventhM: `wantsToFocusOn(focus: Victory.Focus)` does not accept null, I haven't checked if collections.contsins does either. Would make the call require an additional null check I would think (not at computer to look for a better solution). Maybe something like

```kotlin
        if (Victory.Focus.valueFor(stat) != null && civInfo.wantsToFocusOn(Victory.Focus.valueFor(stat)))
                buildingStats[stat] *= 2f
```
Not the cleanest thing in the world. The only other alternative is updating `wantsToFocusOn` to accept null correctly

SeventhM:
> Then again - my git says your branch does not have the #11401 commits at all, so it shouldn't have the EspionageTests? Same - merge forward to 4.11.3 to get green.

https://github.com/yairm210/Unciv/commit/25de75ddda6db5c0f8c51e45cd9543802a0f851c#diff-5081623447cde64e17695e0dc471c9eb6841c4d09526dbbd8ed587381b63f4ca
I've already merged forwards. Idk how it says I haven't. Unless I need to rebase or something for some odd reason

SeventhM:
Wow, so screwing up #11370 by forgetting to include it properly actually ended up mattering. I have a suspicion we should probably have a way to more cleanly do a city clone

I'mma lazily fix this from mobile since it's just 1 line and some comments. Maybe linting CityFounder can be done in a later PR

PR #11427, opened 2024-04-09
Allow policy removal unique to remove multiple policies

SeventhM:


core/src/com/unciv/models/ruleset/Policy.kt
```diff
@@ -33,6 +35,19 @@ open class Policy : RulesetObject() {
...
     }
 
+    fun matchesFilter(filter: String): Boolean {
+        return MultiFilter.multiFilter(filter, ::matchesSinlgeFilter)
+    }
+    fun matchesSinlgeFilter(filter: String): Boolean {
+        return when(filter) {
+            in Constants.all -> true
+            name -> true
+            branch.name + "branch" -> true
```
yairm210, line 45: This won't catch

core/src/com/unciv/models/ruleset/Policy.kt
```diff
@@ -33,6 +35,19 @@ open class Policy : RulesetObject() {
         fun isBranchCompleteByName(name: String) = name.endsWith(branchCompleteSuffix)
     }
 
+    fun matchesFilter(filter: String): Boolean {
+        return MultiFilter.multiFilter(filter, ::matchesSinlgeFilter)
+    }
+    fun matchesSinlgeFilter(filter: String): Boolean {
```
yairm210, line 41: Typo

yairm210:
All in all looks good!

core/src/com/unciv/models/ruleset/Policy.kt
SeventhM: Is it because I forgot the space?

yairm210: Also the filter contains square brackets, no?

SeventhM: No? Actually... Huh... It probably should

PR #11441, opened 2024-04-10

AutumnPizazz:
Add new content, especially the spy part, correct some minor errors to make it more suitable for native Chinese speakers, and delete redundant spaces


AutumnPizazz:
What is wrong? I cannot find any error?

yairm210:
Placeholder `[isOrAre]` not found in `Simplified_Chinese` for entry `[civName] is able to provide [unitName] once [techName] [isOrAre] researched.`

AutumnPizazz:
Oh, 'is' or 'are' are never distinguished in Chinese, which could be translated as '是‘ or even omit.
So the translation of this sentence has something strange for Chinese Charactor.

AutumnPizazz:
Now I add a "Redundant Component" to include the isOrAre. It may be okay this time.

PR #11443, opened 2024-04-10
Two extension features for custom maps

SomeTroglodyte:
* A description - think a "message" from the map author to their users. Editor support, and shows in new game map select between the dropdowns and the minimap preview. No multiline🙃 or markdown🤪 support.
* Starting locations can control their effect on the "Select players from starting locations" button from #11423. As before, or be ignored, or get the Human slot preferentially.
* Re-fixed a sync problem #11423 already tried to fix - from too often to not always - should be once always now.

image
image
image

</details>

... not 100% happy with the UX or the label choices, but works without too much effort ... ~(But I need to re-check that None/Normal discrepancy)~

Now I'd like one more such mini-offer: Allow a map-collection mod like lat-am to control which map gets preselected preferentially. Remember, mod unpacking flattens timestamps on Linux, so these on the author's side can't do it. Map name == mod name? a ModOptions unique? 

@Caballero-Arepa yes this is for you

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -545,6 +545,7 @@ Choose a Wesnoth map file =
 That map is invalid! = 
 ("[code]" does not conform to TerrainCodesWML) = 
 Use for new game "Select players" button: = 
+Enter a description for the users of this map = 
 
 ## Map/Tool names
 My new map = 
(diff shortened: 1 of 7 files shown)


Caballero-Arepa:
Wow, this is amazing, to say the least.
It's really useful for these maps with so many Civs xD
Does it affect CS as well? Like when you press the select civs from spawn points, would it spawn the CS that have a Starting Location? That'd be dope.

> a ModOptions unique?

I think that's the way.

yairm210 (approved):

yairm210:
This is causing problems for many people!!
The fact that the file selector always triggers causes the file to be selected causes the existing mods in the gameinfo to be overridden with the map ones!
Also some players have reported selecting mods and it just not working, not sure what that's caused by but this sounds like the culprit
I'm planning on reverting this commit wholly

Caballero-Arepa:
Yeah, that feature has caused a lot of problems for a long time, I also think its enough.

PR #11461, opened 2024-04-12
Console autocompletion can display *all* possibilities

SomeTroglodyte:
RiverDirections is now namespaced to RiverGenerator with some pertinent helpers

Edit: More commits now actually change behaviour - tabbing right away or after a blank will be on a par with tabbing after a partial word, removepromotion intelligent, lighter color:

<details>

Crass example - tab hit right after the blank (or U-Tab-A-Tab-R-Tab):
image

removepromotion autocomplete - see unit's existing promotions bottom left:
image

Rethinking maybe Color.FOREST is too dark:
image
... `Color.LIME.lerp(Color.OLIVE.cpy(), 0.5f)` aka Color(0x5C9D26) wasn't too bad either... picked simpler.

</details>

### DIFF
--- a/core/src/com/unciv/ui/screens/devconsole/ConsoleUnitCommands.kt
+++ b/core/src/com/unciv/ui/screens/devconsole/ConsoleUnitCommands.kt
@@ -30,7 +30,7 @@ class ConsoleUnitCommands : ConsoleCommandNode {
                 val options = console.gameInfo.ruleset.unitPromotions.keys.asSequence()
                     .filter { it !in promotions }
                     .asIterable()
-                return getAutocompleteString(params.last(), options, console)
+                return getAutocompleteString(params.lastOrNull().orEmpty(), options, console)
             }
         },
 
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/ui/screens/devconsole/ConsoleUnitCommands.kt
```diff
@@ -25,17 +25,18 @@ class ConsoleUnitCommands : ConsoleCommandNode {
             unit.promotions.addPromotion(promotion.name, true)
             DevConsoleResponse.OK
         }) {
-            override fun autocomplete(console: DevConsolePopup, params: List<String>): String? {
-                if (params.isEmpty()) return null
+            override fun autocomplete(console: DevConsolePopup, params: List<String>): String {
+                // Note: filtering by unit.type.name in promotion.unitTypes sounds good (No [Zero]-Ability on an Archer),
```
yairm210, line 29: Why though? Unit abilities that are promotions are passed on when upgrading

yairm210 (approved):

PR #11468, opened 2024-04-13
A ModOptions unique allowing a mod to control which map gets preselected

SomeTroglodyte:
The validation is a bit over the top, we could just as well leave it at silently ignore...

That first lint is because I was looking at the lat-am mod's validation, and it has a tile _improvement_ "Darien Gap" which seems to be supposed to incur mountain-like damage an movement penalties. Yup, should be a Terrain - feature. But: Having the ability for improvements to change those attributes could make for interesting dynamics... Food for thought.


Caballero-Arepa:
That's great!

PR #11474, opened 2024-04-15
Prettier Events - that now respect 'hidden from users'

### LINKED ISSUE #11473: "Hidden from users" not hiding triggerable effect on event notification

### Game Version

4.11.4

### Describe the bug

When creating an event, if you specify one triggered outcome to be hidden from users, it doesn't actually get hidden. It shows up anyway, with the "hidden from users" text clearly visible.

### Steps to Reproduce

In the mod I've attached, a certain event happens every 20 turns with 90% chance. Whenever it happens, one of the options always displays the effect that should be hidden.

### Screenshots

Screenshot (194)

### Link to save file

[RiversOfLava.zip](https://github.com/yairm210/Unciv/files/14971885/RiversOfLava.zip)
Here is the mod with the event in question. The event trigger is given in the Global Uniques JSON.

### Operating System

Windows

### Additional Information

I also noticed that it's possible to save-scum in order to avoid this event, since the event is randomized. If not for that, I would have included a saved game just before the event occurring, but the randomized nature of the event means that it won't necessarily happen, making it possible to restart the turn in order to potentially avoid the undesirable event.

SomeTroglodyte:
Fixes #11473... And makes Events a little fancier.

<details><summary>Screenshot (same as in isssue)</summary>

image
</details>

The save scum fix is also included, though I feel it is a bit overcomplicated.

Interesting story:
* That frisco 1906 quake image is from Wikipedia. Using their jpg directly gave a negative, transparently overlaid over our blue. Removing the color management in Gimp and resaving gave java.io.IOException: Error loading pixmap: Corrupt JPEG. Changing greyscale to RGB and reexporting then gave that screenshot.

Changes to mod for that screenshot: [Rivers of Lava patch.zip](https://github.com/yairm210/Unciv/files/14977721/Rivers.of.Lava.patch.zip)

### DIFF
--- a/core/src/com/unciv/models/ruleset/Event.kt
+++ b/core/src/com/unciv/models/ruleset/Event.kt
@@ -6,25 +6,40 @@ import com.unciv.models.ruleset.unique.StateForConditionals
 import com.unciv.models.ruleset.unique.Unique
 import com.unciv.models.ruleset.unique.UniqueTriggerActivation
 import com.unciv.models.stats.INamed
+import com.unciv.ui.components.input.KeyCharAndCode
+import com.unciv.ui.screens.civilopediascreen.FormattedLine
+import com.unciv.ui.screens.civilopediascreen.ICivilopediaText
 
 
-
-class Event : INamed {
+class Event : INamed, ICivilopediaText {
 
     override var name = ""
     var text = ""
+    override var civilopediaText = listOf<FormattedLine>()
+    override fun makeLink() = "Event/$name"
+
     // todo: add unrepeatable events
 
     var choices = ArrayList<EventChoice>()
+
+    fun getMatchingChoices(stateForConditionals: StateForConditionals) =
+        choices.filter { it.matchesConditions(stateForConditionals) }.ifEmpty { null }
 }
 
-class EventChoice {
+class EventChoice : ICivilopediaText {
     var text = ""
+    override var civilopediaText = listOf<FormattedLine>()
+    override fun makeLink() = ""
+
+    /** Keyboard support - not user-rebindable, mod control only. Will be [parsed][KeyCharAndCode.parse], so Gdx key names will work. */
+    val keyShortcut = ""
+
     var triggeredUniques = ArrayList<String>()
     val triggeredUniqueObjects by lazy { triggeredUniques.map { Unique(it) } }
 
     var conditions = ArrayList<String>()
     val conditionObjects by lazy { conditions.map { Unique(it) } }
+
     fun matchesConditions(stateForConditionals: StateForConditionals) =
         conditionObjects.all { Conditionals.conditionalApplies(null, it, stateForConditionals) }
 
--- a/core/src/com/unciv/models/ruleset/Event.kt
+++ b/core/src/com/unciv/models/ruleset/Event.kt
@@ -22,6 +22,7 @@ class Event : INamed, ICivilopediaText {
 
     var choices = ArrayList<EventChoice>()
 
+    /** @return `null` when no choice passes the condition tests, so client code can easily bail using Elvis `?:`. */
     fun getMatchingChoices(stateForConditionals: StateForConditionals) =
         choices.filter { it.matchesConditions(stateForConditionals) }.ifEmpty { null }
 }
--- a/core/src/com/unciv/models/ruleset/unique/StateForConditionals.kt
+++ b/core/src/com/unciv/models/ruleset/unique/StateForConditionals.kt
@@ -42,4 +42,27 @@ data class StateForConditionals(
     companion object {
         val IgnoreConditionals = StateForConditionals(ignoreConditionals = true)
     }
+
+    /**  Used ONLY for stateBasedRandom in [Conditionals.conditionalApplies] to prevent save scumming on [UniqueType.ConditionalChance] */
+    override fun hashCode(): Int {
+        fun Civilization?.hash() = this?.civName?.hashCode() ?: 0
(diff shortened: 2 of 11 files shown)


core/src/com/unciv/models/ruleset/unique/StateForConditionals.kt
```diff
@@ -42,4 +42,27 @@ data class StateForConditionals(
...
+        var result = civInfo.hash()
+        result = 31 * result + city.hash()
+        result = 31 * result + unit.hash()
+        result = 31 * result + tile.hash()
+        result = 31 * result + ourCombatant.hash()
+        result = 31 * result + theirCombatant.hash()
+        result = 31 * result + attackedTile.hash()
+        result = 31 * result + combatAction.hash()
+        result = 31 * result + region.hash()
+        result = 31 * result + ignoreConditionals.hashCode()
```
yairm210, line 65: This sounds like 

```
result = 0
for (hash in hashes) result = 31*result + has
```

core/src/com/unciv/ui/screens/civilopediascreen/FormattedLine.kt
```diff
@@ -316,6 +293,30 @@ class FormattedLine (
...
+        }
+        try {
+            val image = getExtraImage() ?: return table
+            // limit larger cordinate to a given max size
+            val maxSize = if (imageSize.isNaN()) labelWidth else imageSize
+            val (width, height) = if (image.width > image.height)
+                maxSize to maxSize * image.height / image.width
+            else maxSize * image.width / image.height to maxSize
+            table.add(image).size(width, height)
+        } catch (exception: Exception) {
```
yairm210, line 314: Under what conditions would this throw an exception?! I want to see them!
https://yairm210.github.io/Unciv/Guiding-Principles/#crash-early-crash-often

core/src/com/unciv/models/ruleset/Event.kt
```diff
@@ -6,25 +6,40 @@ import com.unciv.models.ruleset.unique.StateForConditionals
...
     var text = ""
+    override var civilopediaText = listOf<FormattedLine>()
+    override fun makeLink() = "Event/$name"
+
     // todo: add unrepeatable events
 
     var choices = ArrayList<EventChoice>()
+
+    fun getMatchingChoices(stateForConditionals: StateForConditionals) =
+        choices.filter { it.matchesConditions(stateForConditionals) }.ifEmpty { null }
```
yairm210, line 26: ifEmpty{ null } is an interesting choice, I see how that fits later, but it makes the return value of this function a tad unclear
Oh well, probably anyone who'll be looking at this will be using some IDE with type hinting anyway

core/src/com/unciv/ui/screens/civilopediascreen/FormattedLine.kt
yairm210: Sounds to me like we should remove the catch and see who's really behind that mask 😈

yairm210 (approved):

core/src/com/unciv/ui/screens/civilopediascreen/FormattedLine.kt
```diff
@@ -316,6 +296,28 @@ class FormattedLine (
...
+        fun getExtraImage(): Image {
+            if (ImageGetter.imageExists(extraImage))
+                return if (centered) ImageGetter.getDrawable(extraImage).cropToContent()
+                    else ImageGetter.getImage(extraImage)
+            val externalImage = ImageGetter.findExternalImage(extraImage)
+                ?: throw UncivShowableException("Extra image '[$extraImage]' not found")
+            return ImageGetter.getExternalImage(externalImage)
+        }
+        val image = getExtraImage()
+        // limit larger cordinate to a given max size
```
yairm210, line 310: For huge images, wouldn't we want to limit by labelWidth anyway? Sort of like .coerceAtMost?

core/src/com/unciv/ui/screens/civilopediascreen/FormattedLine.kt
```diff
@@ -316,6 +296,28 @@ class FormattedLine (
...
+        val maxSize = if (imageSize.isNaN()) labelWidth else imageSize
+        val (width, height) = if (image.width > image.height)
```
yairm210, line 312: This is a *nice trick*! I've only seen this is python/c#, didn't know you could do it in Kotlin as well!

6 years in and I'm still learning new things :)

core/src/com/unciv/ui/screens/civilopediascreen/FormattedLine.kt
```diff
@@ -316,6 +296,28 @@ class FormattedLine (
         return table
     }
 
+    private fun renderExtraImage(labelWidth: Float): Table {
+        val table = Table(BaseScreen.skin)
+        fun getExtraImage(): Image {
+            if (ImageGetter.imageExists(extraImage))
+                return if (centered) ImageGetter.getDrawable(extraImage).cropToContent()
+                    else ImageGetter.getImage(extraImage)
+            val externalImage = ImageGetter.findExternalImage(extraImage)
```
yairm210, line 305: This exception is not caught anywhere - which means bad mod definition can cause Unciv crash, 

Instead I think this function should return an empty Table or perhaps null

yairm210: There's a difference between *catching specific scenarios with if checks* and just throwing a try/catch all
In the first scenario, you know what you're dealing with. In the second, you have no clue. 
Unciv should never crash sure to misconfigured mods

PR #11480, opened 2024-04-15
Fix Puppets building wonders

SeventhM:
Why do national wonders and regular wonders have different rejection types btw?


yairm210:
Please remove the addition I made yesterday so we don't forgot :)
There's no real reason for the different rejection trips, if unnecessary we can unify them

yairm210 (approved):

PR #11485, opened 2024-04-16
UX: New game: Do not preselect custom map option and defer map file loading

SomeTroglodyte:
... to prevent maps overwriting the mod selection from the last successful game start, see discussion in #11443.

 Could also make new game screen a little snappier if the custom maps option is not selected at all - but once it is it'll take a little longer since it has not precached in the background.

Has seen only 10min of testing so far, and seems to do exactly what's intended. Any coders with time are urged to check out the branch and test.

The alternative would be to not sync the mod selection from any map selection to the mod checkboxes at all - until "start" is clicked, then compare, then ask if different. I'd tend to this.


yairm210:
Yesss thank you :pray:
Also the "extension mods not attached to new game" send to have been a 4.11.1 problem, for most players updating resolves it, still awaiting results of original poster

yairm210 (approved):

Issue #822, opened 2019-05-29
Free worker/settler from policies not added?


yairm210:
This was in a review from one of the players
Need to check if it's accurate, and if so, fix it

Am-per-Sand:
Added.

Collective Rule works correctly
Citizenship works correctly
2.16.7

yairm210:
I checked this too and it seems to work, I asked him specifically for more information but until we get some more concrete information it's closed I guess

PR #11491, opened 2024-04-18
Modding: Validation for civilopediaText

SomeTroglodyte:
As threatened in #11474 - validate FormattedLine content. Including image existence for extraImage, which hopefully makes it OK to silently eat all errors when it is rendered.

Can't include the line for non-texturepacked extraImages yet - using a method alread in above PR is best, but I did base this on master. Otherwise it works standalone.

<details><summary>Example of a mistake now caught:</summary>

image
... yes that should be `"#822"`.
</details>

Otherwise there's more "false" alarms - about nonexisting Tileset images - when combining an extension mod with a base ruleset it was not made for... Usage of the compatibility uniques from the mod checking tab is still an open to-do.

Edit: Oh, and of course that comment `// Note - currently there can be no classes in a RuleSet that are ICivilopediaText but not IRulesetObject - if there ever are, this needs an expanded scope` is already obsolete - there **are** two such classes already. Or almost already. Prepared as last commit.


yairm210 (approved):
Looks great!! Will be a big help to modders hopefully!

PR #11511, opened 2024-04-24
Fix: Android pause/resume cycle not working

### LINKED ISSUE #11432: Game won't open unless i swipe it out on multitask screen

### Version & Device

Game: 4.11.3 signed-APK
OS: Android 14
Devices: SM-G970N, SM-A528B/DS

### Describe the bug

I guess the title is quite self-explaining.
Please check out video on following link.
https://youtu.be/2VxF_MvSN18?si=1d2Fp8QT4I7BtjY8

### Steps to Reproduce

1. Start any game by quickstart, start new, load, whatever
2. Close the game by pushing back button on world screen
3. Reopen the app and wait
4. Game only shows logo. If you don't swipe it out on multitasking screen it will stay there forever.

### Additional Information

I remember this bug has been existed for really long time.
~~And I hate back button still shuts down game :joy: #11179~~

SomeTroglodyte:
Fixes #11432

# Needs coder brains triple-checking my approach

Draft as I'd still like to see if I can fix the stage lying to the splash screen... Tomorrow. Or next week. Mergeable if you don't wanna wait.

Also - when did this break and when? I'd guess Gdx 1.12 changed exit - read its ~Kdoc~ JavaDoc. It _says_ it will not terminate an Android app... Maybe 1.10 **did** kill the app process, so this problem didn't hit, because shutdown()'ed coroutine executors were never reused... Or it was some 'roid major version changing the lifecycle.

Also - the diff mixes stuff up badly, a lot of code was just moved, like the top-level private elements that are now private to the Dispatchers container. Main idea - the `Dispatcher` static can resurrect itself by instantiating new dispatchers when the old ones were shut down.

### DIFF
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -55,13 +55,17 @@ android {
     }
 
     buildTypes {
-        getByName("release") {
+        debug {
+            isDebuggable = true
+        }
+        release {
             // If you make this true you get a version of the game that just flat-out doesn't run
             isMinifyEnabled = false
             proguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")
+            isDebuggable = false
         }
-
     }
+
     lint {
         disable += "MissingTranslation"   // see res/values/strings.xml
     }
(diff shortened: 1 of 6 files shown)


yairm210:
How on earth did you figure this out?!
I'm super impressed

yairm210 (approved):

PR #11516, opened 2024-04-24
Fixed problem with selecting city constructions option and AutoPlay

tuvus:
Fixes the problem "Unciv 4.11.7 auto assign city production" on discord.
The problem stems from #11329 but should be fixed now from the one line of change.


yairm210:
release patch

PR #11524, opened 2024-04-26

AutumnPizazz:


AutumnPizazz:
Don't accuse me for this is the result of many debates.

AutumnPizazz:
Thanks for your explanation.
But what we (many Chinese players) debate are how to translate the latter sentence more promptly such as Scout's ability. You can see how I change in those 7 times of modification. The easy word-to-word translation is bad for that.

PR #11535, opened 2024-04-28
[docs] fix links to Unique-parameters#terrainfilter and uniques.md

husujo:


husujo:
Thanks. When I read [updating-the-wiki](https://yairm210.github.io/Unciv/Developers/From-code-to-deployment/#updating-the-wiki):

"A proper MarkDown editor or IDE can be used to write the wiki, bringing faster editing, clickable links while editing, better live HTML preview, and automatic detection of problems like broken links." 
I thought meant I can write wiki with IDE, as long as I test the links on Github code browser.

"Inter-wiki 	Should begin with "./", and include ".md". 	./Mods.md#other" 
I thought meant I should use relative path, and end in .md for wiki files linking to other wiki files. Is this the line you are talking about? Though ./uniques.md (instead of ../uniques.md) doesn't work on github code browser.

"These formats will allow IDEs like Android studio to resolve these links and check for broken links, while also working on the [Github code browser](https://github.com/yairm210/Unciv/tree/master/docs)." 
I thought meant that links should work in IDE and on Github (but were broken and I tried to fix)

"The bot that updates the wiki from the main repository automatically translates them into formats that are compatible with Github Wikis, which have somewhat non-standard requirements." 
I didn't know how to test for the deployed wiki, so the best I could do is test on Github. thought maybe the bot translation would handle it.

husujo:
Hey, so I checked the wiki and it seems like the links are working, does that mean it's ok for me to continue fixing links in the same manner I have in this commit?

husujo:
it happens, definitely check me again on the next one!

PR #11545, opened 2024-05-01
Fix "Translating" wiki link

SomeTroglodyte:
Fixes #11543

I had some completely unrelated lints in the changelist, so I've included them here - better than in a more functionally important PR. Separate commits. Actually not tested in a targeted manner - but obvious, no?

That Quest code I came across because I had a "Great Lighthouse" quest in an information era start, really, but couldn't figure out how it could sneak in...

PR #11546, opened 2024-05-01
Overhaul "hidden" architecture, closing some gaps

SomeTroglodyte:
> "a quest for a wonder blocked by a missing victory type is possible" (https://github.com/yairm210/Unciv/pull/11545#issuecomment-2088814081)

... led to toppling most of the 'how do we hide RulesetObjects' stuff. Some ugly compromises that have irked for a long time are gone, and city-states will no longer tell you to build the United Nations if Diplomatic Victory is off. However, there are subtle potential changes in behaviour, and I'm too tired to recap the diff to point them out right now.

Hiding now works centrally and there *were* one or two more oversights than only that Quest filter. On the other hand, there was a StateForConditionals on some RuinReward code that blocked the migration and I couldn't imagine any mod successfully using that, so it's gone for now - but the way to structured conditional support specifically of the hidden from pedia unique is paved - no not paved but the perspective that there's no real blockers is cleared up. May need another abstract/open helper on that interface for readability...

I'll PR this as reviewable because I can play with it and have tested most edge cases I could think of, but more independent testers would be very welcome.

### DIFF
--- a/core/src/com/unciv/UncivGame.kt
+++ b/core/src/com/unciv/UncivGame.kt
@@ -479,6 +479,8 @@ open class UncivGame(val isConsoleMode: Boolean = false) : Game(), PlatformSpeci
         lateinit var Current: UncivGame
         /** @return `true` if [Current] has been set yet and can be accessed */
         fun isCurrentInitialized() = this::Current.isInitialized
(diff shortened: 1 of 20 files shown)


yairm210:
Looks okay, but there is a lot going on
I'll release a version on Saturday eve with this, so I can be more available for problems that may occur

PR #11551, opened 2024-05-03
Spy rank UI and fixes

tuvus:
This PR does a few things and is part of #7610.

- First it adds a new spy rank column that displays the spies rank.

SpyRankUI

</details> 

- Better fix in #11559  ~Second it fixes the spy rank not being cloned. So basically it is reset and the rank did not persist.~
- Third dead spies can't move to a different location. I disabled the move button.
- Fourth I fixed the EspionageAutomation AI. It wouldn't send spies to steal tech or rig elections because there was a logical error while checking if the action was successful or not. A spy takes time to set up so it's state won't be StealingTech or RiggingElections and will be Moving instead. So I changed it to return true if it found a viable target instead.

### DIFF
--- a/core/src/com/unciv/ui/screens/overviewscreen/EspionageOverviewScreen.kt
+++ b/core/src/com/unciv/ui/screens/overviewscreen/EspionageOverviewScreen.kt
@@ -70,10 +70,12 @@ class EspionageOverviewScreen(val civInfo: Civilization, val worldScreen: WorldS
     private fun updateSpyList() {
         spySelectionTable.clear()
         spySelectionTable.add("Spy".toLabel())
+        spySelectionTable.add("Rank".toLabel())
         spySelectionTable.add("Location".toLabel())
         spySelectionTable.add("Action".toLabel()).row()
         for (spy in civInfo.espionageManager.spyList) {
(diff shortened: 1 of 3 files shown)


AdityaMH:
Interesting, we not added great general names since xlenstra added it. but now we instantly have spies names before release.

Don't forget about england ability.

core/src/com/unciv/ui/screens/overviewscreen/EspionageOverviewScreen.kt
```diff
@@ -70,10 +70,12 @@ class EspionageOverviewScreen(val civInfo: Civilization, val worldScreen: WorldS
     private fun updateSpyList() {
         spySelectionTable.clear()
         spySelectionTable.add("Spy".toLabel())
+        spySelectionTable.add("Rank".toLabel())
```
yairm210, line 73: Is this translated?

yairm210 (approved):

core/src/com/unciv/ui/screens/overviewscreen/EspionageOverviewScreen.kt
tuvus: Yes

PR #11570, opened 2024-05-07
Spy UI improvements

### LINKED ISSUE #11569: RFC: Fly-by observations (and mini-bugs) concerning Espionage
One for @tuvus , maybe even @ravignir if still listening, but any voice welcome (and sorry for the label mix - this **is** a mixed bag):

### UX
* A spy in surveillance will pop the mandatory espionage screen visit **_every darn turn_**
* Notifications *could* open the espionage screen, might be nicer
* Column "spy present" for the "hideout" row could show idling spy(spies)
* Those Move buttons are _F A R_ too big
* Which spy is to be moved could be clearer
* Spy icon could indicate rank
* Interaction on city name - show on map?

### Rules / bugs
* National Intelligence Agency gives the extra new spy two ranks not one
* If a spy fails and is killed during a tech spy attempt - the tech is rewarded anyway without notice
* Tech steal success rate seems too high - because minimum rank already shifts probs???
* Cannot station spy for counter-intelligence
* Fixing that - they do "establish network" for the same duration as abroad which maybe they shouldn't

I'm currently partitioning into readable commits - but any input would be nice on:
* establish network duration: I changed it to 3 for abroad 1 domestic - could be zero? as in just 1 turns in the moving state when going from hideout  to counter-intelligence??? The fandom wiki is not clear - someone got other sources?
* Success/fail probabilities: I haven't coded or tested it - but I'd tend to make rank 0 have zero "skill modifier", shifting outcome to 0% unnoticed success | 33% noticed but unrecognized success | 33% victim-knows-whodunnit success | 33% fail+kill. **Unnoticed** success chance is then directly proportional to levelups, capping at ~20% overall, and fail+kill chance is never virtually zero, an overall ~13% remaining at rank 3 without counteragents. Yes, Oscar's math is linear throughout, and I haven't checked the DLL sources on whether that general model is even close to the original - playable it certainly is.
* Vague idea - show modifiers from uniques somehow. The numbers are easy, but their effect and thus presentation is not. Rank influences tech steal success rate while Uniques influence tech steal speed, completely independent from each other - unintuitive by definition. "Success modifier: X%, speed modifier: Y%"??? Maybe, but effect on other tasks is entirely different......

<details><summary>appetizer</summary>

image

</details>

### Related Issue Links

#11557, #11077, #7665... And the mothership: "Implementation of Espionage - Roadmap #7610" - some of these checkboxes could be toggled by now via tuvus's work, and one half by my upcoming PR.

SomeTroglodyte:
This is already a few days old and won't get more features... Read commit messages!
Closes #11569 .

For screenie, see issue. As this is still an opt-in feature, I think the scope/diff size is still OK. Most of it is lint and prettification anyway.

After that discussion and more playtesting with logged spy stuff (so I can observe what the enemies do, too) I am content with unchanged formulae - could tweak later if we need to. But there still *are* rule changes, I'll try to summarize them:
* [Fix National Intelligence Agency giving one extra Spy level](https://github.com/yairm210/Unciv/commit/b491ae310c2d4c0dcd39d3043b9c3283140915ab)
* [Fix failed tech theft rewarding tech anyway](https://github.com/yairm210/Unciv/commit/a47a9650535741742bcdb913f7a4377275c5108d)
* Change to be killed is applied for a Spy performing steal tech even when the outcome is no techs to steal
* [Shorten establish-network phase for domestic spy placement](https://github.com/yairm210/Unciv/commit/bcfcd48f8b3a16bc9263b77db82f559fd5fe5667) ... to zero after input by @tuvus (this not really a rule change as the changed behaviour was inaccessible)

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Buildings.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Buildings.json
@@ -986,8 +986,11 @@
         "cost": 120,
         "culture": 1,
         "isNationalWonder": true,
-        "uniques": ["Hidden when espionage is disabled", "Gain an extra spy", "Promotes all spies",
-			"[-15]% enemy spy effectiveness [in this city]", "New spies start with [1] level(s)",
+        "uniques": ["Hidden when espionage is disabled",
+            "New spies start with [1] level(s)",
+            "Promotes all spies",
+            "Gain an extra spy", // Order is significant here
+			"[-15]% enemy spy effectiveness [in this city]",
 			"Only available <if [Police Station] is constructed in all [non-[Puppeted]] cities>",
 			"Cost increases by [30] per owned city"],
         "requiredTech": "Radio"
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1756,7 +1756,9 @@ Enhanced religion =
 Spy = 
 Spy Hideout = 
 Spy present = 
-Move = 
+Move =
+Spy skill from rank: [percent] (influences success) = 
+Spy effectiveness: [percent] (influences time) = 
 
 After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
 We have recruited [spyName] as a spy! = 
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -1756,9 +1756,7 @@ Enhanced religion =
 Spy = 
 Spy Hideout = 
 Spy present = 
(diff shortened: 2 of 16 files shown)


tuvus:
Aaahhh, the unique ordering is what caused the problem!

Interesting that you decided to make a new transient City variable and how you changed the enum to hold more information.

Was UniqueTarget.Global important? (I just copied from the other ones)

Also note that Counter-Intelligence only has a countdown time for the AI.

Otherwise looks good

yairm210 (approved):

PR #11572, opened 2024-05-07
Console: `civ addpolicy`

SomeTroglodyte:
Another drive-by...

Worth mentioning: I actually was a little blind for a while there how useful that new `findCliInput` variant is, and coded `console.findCliInput<PolicyBranch>(params[1]) ?: console.findCliInput<Policy>(params[1])` - embarrassing :see_no_evil:.

`console.findCliInput<Iconstruction>(param)` or even `console.findCliInput<IRulesetObject>(param)` would work just fine - opens possibilities. On buildings, it's less efficient than `console.gameInfo.ruleset.buildings.values.findCliInput(thing)` but shorter.

### DIFF
--- a/core/src/com/unciv/ui/screens/devconsole/ConsoleCityCommands.kt
+++ b/core/src/com/unciv/ui/screens/devconsole/ConsoleCityCommands.kt
@@ -1,5 +1,7 @@
 package com.unciv.ui.screens.devconsole
 
+import com.unciv.models.ruleset.Building
+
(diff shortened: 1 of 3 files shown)


yairm210 (approved):
With "add" I would expect to see a "remove"

yairm210:
This is primarily for testing and for scenario building, so even if something isn't possible in game it should be possible to "revert" console actions

PR #11583, opened 2024-05-12
Corrected Coast yields to give 1 Food and 1 Gold

Skekdog:
As per Civ 5, both in G&K and Vanilla

### DIFF
--- a/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
@@ -12,6 +12,7 @@
 		"name": "Coast",
 		"type": "Water",
 		"food": 1,
+        "gold": 1,
 		"movementCost": 1,
 		"RGB": [70, 138, 216],
 		"uniques": ["[+2] to Fertility for Map Generation",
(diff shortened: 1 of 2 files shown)


yairm210 (approved):

Skekdog:
> Can't believe we went so long without noticing. Having Ocean tiles partially balanced lack of resources vs +1 gold always felt right...

For what it's worth, it seems that this was previously an intentional change which is now being reverted (https://discord.com/channels/586194543280390151/1239210672244654183)

> But - the fandom wiki also says (or can be read as saying - it's not 100% unequivocal) Ocean can't be worked by population for yield until some tech is researched - meaning the same one Workers need to move onto it I presume???

Did my own test now, population can work Ocean tiles like normal. No idea what the fandom is talking about

> And they claim Atolls can spawn on Ocean - which always felt weird that Unciv ones can't.

The in-game Civilopedia also claims this, though I've loaded two huge continents maps and never saw this happen.

Just to be safe I did check that Coasts work with Golden Ages as normal - they do, no problems there.

Skekdog:
What exactly needs testing? Gold costs after natural border expansion?

PR #11587, opened 2024-05-13
Spies can be moved on map

tuvus:
This PR adds on to #7610 by adding a new more visual way of moving spies using the world screen.

Right clicking the move button will close the espionage screen and open the world screen with the spy selected. The player can then move the spy to any available city that the spy can move to. Available cities have a cyan circle on them _(I chose cyan because I like it and it stands out. This can definitely be changed. Maybe it should depend on the resulting action in the city)_. Clicking on the tile shows a move button that moves the spy there when clicked and opens the espionage screen.

Pressing esc or the X button on the spy panel exits this mode.

MoveSpyViaMap

</details>

### DIFF
--- a/core/src/com/unciv/ui/screens/diplomacyscreen/MajorCivDiplomacyTable.kt
+++ b/core/src/com/unciv/ui/screens/diplomacyscreen/MajorCivDiplomacyTable.kt
@@ -85,12 +85,12 @@ class MajorCivDiplomacyTable(private val diplomacyScreen: DiplomacyScreen) {
         if (otherCiv.cities.isNotEmpty() && otherCiv.getCapital() != null && viewingCiv.hasExplored(otherCiv.getCapital()!!.getCenterTile()))
             diplomacyTable.add(diplomacyScreen.getGoToOnMapButton(otherCiv)).row()
 
-//        if (!otherCiv.isHuman()) { // human players make their own choices
+        if (!otherCiv.isHuman()) { // human players make their own choices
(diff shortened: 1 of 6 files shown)


yairm210 (approved):

core/src/com/unciv/ui/screens/diplomacyscreen/MajorCivDiplomacyTable.kt
```diff
@@ -85,12 +85,12 @@ class MajorCivDiplomacyTable(private val diplomacyScreen: DiplomacyScreen) {
         if (otherCiv.cities.isNotEmpty() && otherCiv.getCapital() != null && viewingCiv.hasExplored(otherCiv.getCapital()!!.getCenterTile()))
             diplomacyTable.add(diplomacyScreen.getGoToOnMapButton(otherCiv)).row()
 
-        if (!otherCiv.isHuman()) { // human players make their own choices
+//        if (!otherCiv.isHuman()) { // human players make their own choices
```
yairm210, line 88: Why this change?

tuvus: Oooops, that slipped through. I was testing if the modifiers worked correctly.

tuvus:
I just added another way to close out of the spy moving on map mode. The player can now click on any ineligible tile to click on an X button which pops up, bringing them back to the espionage screen. This should make it more obvious how to cancel the spy movement.

MoveSpyViaMap2


</details>

PR #11588, opened 2024-05-13
Allow access to the Dev Console on mobile devices

SomeTroglodyte:
### Question

Should dev console stay English throughout or shall we start preparing to have more of it translated... If the former I'll need to change this. Why? We already had the keyboard binding translated, so when I added a button in the WorldScreenMenuPopup, it got translated - perfectly fine. Then I added that same string as decoration inside the Popup so the little checkbox doesn't feel alone... And here the dice need to be thrown.

### Other Notes

- Interesting adventure catching the perpetrator that was stealing the on-screen keyboard on every autocomplete.
- Yes there still are acceptable keyboards that can send a Tab code, even though F-Droid search nowadays stubbornly refuses to show hacker's keyboard[^1]. I tried "unexpected" - cute.
- Tab, Up, Down in console aren't in the KeyBindings system - not necessary, but a thing to remember. Potentially a way for Roid users not able or willing to install a console-friendly keyboard.
- Bonus: WorldScreenMenuPopup needs scrolling less often since it now goes dual-columns pretty cleverly - "one more will fit / where do we draw the line"
- Bonus: Persists history - me lazy me like - me need limit size forgot - dun

<details><summary>visuals</summary>

https://github.com/yairm210/Unciv/assets/63000004/22916d75-e69d-47d3-bfd1-2ec96bd3ed45

A little klunky... But shows how adding a Scout then giving it two promotions with the console kept open might be nice - you can see the map being updated and with eagle eyes even a tutorial being fulfilled.

In case you notice the "space" of the _physical_ keyboard not coming through - I suspect a quirk of this old AVD with the old Kbd app, so I ended up _clicking_ the space instead...

</details>

[^1]: Because it's stale at API Level 9


yairm210:
English throughout

core/src/com/unciv/ui/screens/worldscreen/mainmenu/WorldScreenMenuPopup.kt
```diff
@@ -46,17 +70,26 @@ class WorldScreenMenuPopup(val worldScreen: WorldScreen) : Popup(worldScreen, sc
...
+        }.nextColumn()
+        if (showMusic)
+            addButton("Music", KeyboardBinding.MusicPlayer) {
+                close()
+                WorldScreenMusicPopup(worldScreen).open(force = true)
+            }.nextColumn()
 
-        addCloseButton()
+        if (expertMode)
+            addButton("Developer Console", KeyboardBinding.DeveloperConsole) {
```
yairm210, line 85: Also limit this to not open on MP games

yairm210: That is, don't even display the button

yairm210 (approved):
Looks good!

Ouaz:
`Developer Console` and `Keep open` are not translated on my end (in the Console window). Not a big deal though, I think it can even stay untranslated (as commands are in english anyway, and it's a developer tool after all).

Nice work btw, very useful.

PR #11596, opened 2024-05-14
Fix crash when a starting unit has a random conditional

SomeTroglodyte:
Can't include currentTile in the hash when the unit isn't on the map yet - but Uniques are evaluated that early, namely when updateUniques runs *while the unit is being placed*

There might be a code flow reorder that would also "fix" this mad modder, but - a hardened hashCode is better than a rarely throwing one anyway.


yairm210:
One thing raised in the Discord is that nothing in the hash function addresses the current turn, so a "random effect" with the same conditional state would remain - in effect - constant

yairm210 (approved):

PR #11597, opened 2024-05-14
Added the Espionage civilopedia entry

tuvus:
Part of #7610.

This adds the civilopedia espionage entry describing what actions spies can take.

EspionageTutorial

</details> 

Is there any way we can add the spy icon to the left of the Espionage title?

After these PRs go through I think it espionage might be ready to be released.
There are other minor things like ~Diplomat and city-state coup quests that we are missing~. I haven't looked into these as much. Also coup AI.

Edit: Diplomats are BNW and the coup quest that I found on the internet might have been a mod.


tuvus:
I updated the format and added the image from #11600.

EspionageCivilopedia


</details>

yairm210 (approved):
Nothing much to say about this, time will tell what we'll need to change, as with all documentation

PR #11600, opened 2024-05-16
Espionage icons

SomeTroglodyte:
While toying with #11597, I noticed all those buildings were showing the "Fallback" icon...

Notes:
- That new Open-Espionage button icon - I couldn't resist after I came across it, too cute. And variety is good. Can be easily modded back after all. Holler if you'd like it on the NextTurnButton too (I'd say no because it says "Spies"), or drop second commit if you hate it.
- OtherIcons/Spy is only used for Notifications, OtherIcons/Spy_White for the NextTurnButton, and the guys on the Move Spies screen (and the move spies on map PR).
- Policy screen and undo made independently moddable. Since Constitution is only 50x50, re-done from source to 100x100 for the open PolicyPickerScreen button.
- That (in this PR) unused additional image is meant for the Tutorial decoration. Another "too-cute once seen". Will tie together once that tutorial PR is merged.
- Please pardon the use of gray -again-: Was the best idea I had to hint the Great Firewall has an evil aspect outside Civ. And the agency looked worse with the spy too small so it would not overlap, or with the building black - that outline I gave the spy didn't help at 100x100.

The tutorial shows all the new building icons:
image

WorldScreen:
image

Top of the tutorial showing that extra image:
image

```json
  {
    "name": "Espionage",
    "civilopediaText": [
        {"text": "Spies!", "header": 2, "color": "red", "centered": true},
        {"extraImage": "OtherIcons/EspionageTutorial", "imageSize": 200},
```

<details>

Archive of the source svg's and the gimps just before scale and export:
[Espionage Gimps.zip](https://github.com/yairm210/Unciv/files/15328110/Espionage.Gimps.zip)


tuvus:
The building icons look good. I am not so sure about the espionage icon on the world screen though. It looks a little like a plane on a runway from afar and seems too detailed for it's size.

Thanks for adding the images!

PR #11609, opened 2024-05-16
Spy Icon disabled when not on turn

tuvus:
This PR is a small fix of  #11584 and #11587

Basically, the spy icon could be clicked from a spectator or when it isn't the active player's turn. I simply added an extra if statement since the icon can't be disabled.


tuvus:
> Why not? Anything can... What would the desired effect be?

That the onClick() doesn't trigger. The way I have here works as well.

yairm210:
@SomeTroglodyte is there anything here that's a no-go, or are these meant to be comments and hints?

yairm210 (approved):

PR #11614, opened 2024-05-17
Fixed gold ruin not displaying notification

tuvus:
This PR fixes a feature originally created in the commit 83ff2d1 which was then broken one we unified the unique triggers in #11011.

This problem was originally reported in two channels on discord.

This PR simply fixes this by adding back the `triggerNotificationText = "from the ruins" text to the trigger. I tested it with other ruin effects and it doesn't add the notificationText to them.

RuinGoldNotification

</details> 

Also, how did I gain 42 gold? It should be between 50 and 100, it shouldn't be modified by the game speed.


yairm210 (approved):

tuvus:
I think it would be cool if it were modified by game speed. I think things like barbarian encampments, pillaging and ruins give too little gold in later stages of the game as well.

PR #11615, opened 2024-05-17
AI worker build roads improvement

tuvus:
This PR is brought to you by A* and #10631, thanks willjallen! 

If you haven't noticed, workers will only try to connect cities to the capital and not because of movement reasons because of #10918 and #10776. This PR aims to fix that and have workers create a road network.

Here are some notes on the changes:

- We build roads between neighboring cities using the new `City.neighboringCities` variable. This variable will probably be very important in the future.
- Which road is the "from" road and the "to" road depends on their order in the `civ.cities` list. This is important for ensuring multi turn consistency of the planned roads.
- Road building priority depends on the state of the road and if the city is connected to the capital or not.
- If there are no roads to build to nearby cities and the city isn't connected to the capital yet, then we will try to build a road to the closest connected city. (I have not changed this very much)
- The A* algorithm tries to reuse railroads and roads, however it turns out that railroads and roads should have the same priority. If they don't the AI might try to build railroads off of roads or not upgrade the roads.
- Sometimes there is a lazy missionary blocking a road, now the workers will swap with it to get to the tile they want.
- I made some small changes that should support military workers better, but I haven't tested it yet and should probably leave them for a different PR.

<details><summary>Picture of the new road network</summary>

WorkerRoadAIImprovement

</details>


yairm210:
I haven't got the brain to work through everything you did here, in general I trust you that it's better than the current version so I'll approve on those grounds
Worst case we'll get bug reports

yairm210 (approved):

PR #11616, opened 2024-05-17
UI candy: WLTK fireworks

SomeTroglodyte:
Untypically, this does not have a cleaned-up commit history. In fact, a much longer history over months needed to be cut off as git was giving me grief (refused to rebase claiming my empty local changes were not empty).... All in all, a painful birth as I hunted for geometry bugs over and over again in all the wrong places. Not going to invest the extra hours to go back and re-Meld.

<details><summary>In-game visuals</summary>

https://github.com/yairm210/Unciv/assets/63000004/e552718f-8e81-4833-8070-8643ffc1eba8

... That extra dimming just to be able to see the effect better is, of course, open for discussion. Actually, setting mapHolder.color.a = 0.5 was almost nicer, but I wanted the city itself to have normal illumination ...

</details>

<details><summary>FasterUIDevelopment demo</summary>

https://github.com/yairm210/Unciv/assets/63000004/89816054-2e08-465a-995c-24f8b73f1393

... so you can see it against a dark uniform background.

</details>

Notes:
- Performance: Some Gdx particle docs imply complex effects can do with no more than a few handful of particles alive at the same time - this does hundreds. But framerate is still pegged at 60 on a NUC11 (means intel integrated and mesa drivers) - which has CPU at 100% before even launching Unciv. Run under debugger control. So it can't be that bad...
  Edit: Tested on an old (hardware 2015, OS 2020, los 16 = Pie) device - perfectly fluid.
- It's an extensible architecture not as anti-yagni but because I did not succeed debugging it until it was. The alternate implementation used for testing is cleaned up, but could be dug out of this PR if needed.
- Reverts to current icons when "continuous rendering" is off
- Potentially moddable, but not this PR. I believe some code deduplication is in order first.
- One leftover PNG delete wouldn't really belong here - but otherwise I'll forget
- That little texture the particles need has no sub-path, and that's intentional. That would work but require patching the `.p` file in a text editor, and afterwards reloading in [the editor](https://github.com/libgdx/gdx-particle-editor) becomes - actually, haven't even tested that part, but I expect trouble.
- I moved it into ConstructionIcons in the cleanup phase as that atlas will be most active on cityScreen.
- Having these things over every city you know is having WLTK ***on the world map*** would **_not_** be impossible :fearful: 
- Major architecture change for FasterUIDevelopment: It no longer shares the settings file with main Unciv. Too untransparent and I wanted resize-relaunch to persist size. But that also means two settings that **are** meaningful for FasterUIDevelopment are now inaccessible except via text editor or debugger: language and screen size. I'll treat that once I need it... (and yes everything else in settings is irrelevant AFAICT, and won't even appear in the new file)


core/src/com/unciv/ui/components/ParticleEffectAnimation.kt
```diff
@@ -0,0 +1,200 @@
...
+
+/** Hosts one template ParticleEffect and any number of clones, and each can travel linearly over part of its lifetime.
+ *  @property load Must be called from a subclass at least once - see detailed Kdoc
+ *  @property configure Optionally modifies all clones - see detailed Kdoc
+ *  @property getTargetBounds Where to draw on the stage - see detailed Kdoc
+ *  @property getScale Optionally scales all effects - see detailed Kdoc
+ *  @property render Needs to be called from the hosting Screen in its render override
+ *  @property dispose Required - this is just the Gdx Disposable interface, no automatic call
+ */
+abstract class ParticleEffectAnimation : Disposable {
```
yairm210, line 25: for files taken mostly from elsewhere, the accreditation should be on the files themselves

yairm210 (approved):
Sure, don't see why not

PR #11623, opened 2024-05-20
UI: Allow any mod sound to be selected as a multiplayer turn notification sound

SomeTroglodyte:
... Actually, "New architecture to find any non-texture-atlas media" would be more fitting. but that title up there can go into the changelog.

No issue, this is for a slightly miscast "Modding request": [Allow any mod sound to be selected as a multiplayer turn notification sound.](https://github.com/yairm210/Unciv/issues/3242#issuecomment-2118884661) - borrowed the title even though my first one was a tad shorter.

This is the extreme version - started with "how would I imagine an easy to use yet still flexible tool to answer the question 'which media file if any should I use for a given key'..." and then coded top-down from the framework draft. I did not yet check whether the other intended clients do anything in a significantly different way. I suspect the priority between newgame-selected mods, perm audiovisual ones, and builtin is not always exactly the same - but it should be.

The ugliest part of that request is how to attach human-readable labels to sounds, the old solution to that was ugly, and it's not much better here. As mentioned in my replies to the above, moving UncivSound back to enum-based - nope, without large refactorings that ultimately fails on unexpected impediments. Refactoring all UncivSound usages to cleanly distinguish between the interface and the enum - possible, but not right now. Then once that enum exists, it could carry display labels... Again, not yet. But in here there's still a tricky little imp that still enumerates the existing UncivSound.X instances almost as if the container _were_ an enum. Almost moot since most instances need a mapped label anyway, but - had fun coding it.

No screenshots since I don't know of any suitable mod

@5cover - your idea and you mentioned PR'ing the same - so your review would welcome, s'il vous plaît.

### DIFF
--- a/core/src/com/unciv/logic/files/IMediaFinder.kt
+++ b/core/src/com/unciv/logic/files/IMediaFinder.kt
@@ -1,0 +1,264 @@
+package com.unciv.logic.files
+
+import com.badlogic.gdx.Application
+import com.badlogic.gdx.Files
+import com.badlogic.gdx.Gdx
+import com.badlogic.gdx.files.FileHandle
+import com.unciv.UncivGame
+import com.unciv.models.UncivSound
+import com.unciv.models.metadata.BaseRuleset
+import com.unciv.models.ruleset.Ruleset
+import com.unciv.models.ruleset.RulesetCache
+import com.unciv.models.ruleset.unit.BaseUnit
+import com.unciv.ui.audio.SoundPlayer
+import com.unciv.ui.audio.MusicController
+import com.unciv.ui.screens.civilopediascreen.FormattedLine
+import com.unciv.ui.images.ImageGetter
+import com.unciv.ui.popups.options.SettingsSelect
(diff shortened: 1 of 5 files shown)


5cover:
Sounds good. This solves my modding requests and allows new sources for the notification sound :

1. sounds from permanent audiovisual mods
2. attack sounds of units from non-base rulesets (which was a workaround I attempted when 1. didn't work)

However I don't understand the purpose of `prettifyUncivSoundNames`:

https://github.com/yairm210/Unciv/blob/684dfcb05b2f3049579551d3b737c956605d68af/core/src/com/unciv/logic/files/IMediaFinder.kt#L200-L210

I think it would be better to keep the original sound names (which is the current behavior) for consistency, and because if the sounds ever change we may forget to change this description. It's redundant anyway since users hear the sound when they select it.

Also, I'm not exactly familiar with how localization is done in Unciv - but does it work properly for builtin sound names and the "Attack Sound" suffix?

yairm210 (approved):

PR #11625, opened 2024-05-21
ImagePacker detects changed settings file, improve documentation.

SomeTroglodyte:
Triggered by https://github.com/yairm210/Unciv/issues/3242#issuecomment-2119572326

Question: Do we **_use_** indexed textures or should we turn the default for `useIndexes` off?


yairm210:
I don't know what that means, so my guess is "no"

yairm210 (approved):

PR #11630, opened 2024-05-22
Revert Espionage button icon

tuvus:
Reverts part of #11600, changes the espionage button icon back to the "Spy White".
The reasoning for this change is that I think the Spy White icon is a bit more informative and is more visible form afar.

Maybe we could use the other spy icon in some Easter egg?


tuvus:
Got it! That makes sense.

tuvus:
Espionage.png doesn't actually exist. I replaced the part of Icons.png that it was in with the spy_white icon copy. There isn't really any need to change the atlas (except some padding on the image which I kind of guessed on essentially).

Ooooh, wait. What compression level should I use in gimp? that might be the reason it has a 3000% size increase. (I did no compression)

tuvus:
> you change the source png and let Unciv pack... 

Aaaaaahhhhhh! Thats why I was having trouble with the formation link images! Whoops, I didn't realize that, I'll fit it.

yairm210 (approved):

PR #11638, opened 2024-05-22
Add Adjacent Tiles Multipled Object Stat Bonus

PLynx01:
I would like to implement two uniques that provide stat bonus to tiles, buildings and specialists multiplied by the number of adjacent tiles of specified type.

Their primary use cases are Civ 6 recreations mods, that heavily depend on the tile adjacency stat bonuses. Currently, the modders have to rely on repeated stat bonus uniques paired with conditional checking the amount of neighboring tiles of specified type. It looks like this:

```
"[+1 Production] from every [Manufactory] <with [1] to [1] neighboring [Mountain] tiles>",
"[+2 Production] from every [Manufactory] <with [2] to [2] neighboring [Mountain] tiles>",
"[+3 Production] from every [Manufactory] <with [3] to [3] neighboring [Mountain] tiles>",
"[+4 Production] from every [Manufactory] <with [4] to [4] neighboring [Mountain] tiles>",
"[+5 Production] from every [Manufactory] <with [5] to [5] neighboring [Mountain] tiles>",
"[+6 Production] from every [Manufactory] <with [6] to [6] neighboring [Mountain] tiles>",
``` 

Using my new uniques can shorten this to this:

```
"[+1 Production] from every [Manufactory] per every neighboring [Mountain] tile"
``` 

There is also a capped version of this unique:
```
"[stats] from every [tileFilter/specialist/buildingFilter] per every neighboring [tileFilter] tile (from [positiveAmount] to [positiveAmount])"
``` 

The first number in parentheses is the minimum number of neighboring tiles, that is necessary to trigger the effect. The second one is a multiplier cap, so if the number of matching tiles is above it, then it's capped to the upper limit.

If you have some remarks or questions, please share them below, as my code may contain errors and/or code smells. Constructive feedback is welcome, but please refrain from overly criticizing my pull request unless it's necessary. This is my first pull request of that size, so please be considerate towards my effort.

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
@@ -193,6 +193,13 @@ object ReligionAutomation {
         cityToBuy.cityConstructions.purchaseConstruction(inquisitorConstruction, -1, true, Stat.Faith)
     }
 
+    private fun getAdjacentTileCount(centerTile: Tile, tileFilter: String): Int {
+        val tileCount = centerTile.neighbors.count {
+            it.matchesFilter(tileFilter)
+        }
+        return tileCount
+    }
+
     // endregion
 
     // region rate beliefs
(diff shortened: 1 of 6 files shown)


yairm210:
This is in too similar a vein to what I want to implement, namely *multiplying uniques*
That is, I want to implement a modifier that multiplies a unique N times, where N is a countable - that will solve a bunch of existing duplications
Then, we can extend that with "countables with conditional state as input" for really explosive results

The difficult part is that existing conditionals are "true/false" and do not allow for multiplication - so this would probably need to be implemented in IHasUniques.getMatchingUniques

Now that I write it out, an initial implementation might not be difficult at all... interesting 🧐

PLynx01:
> This is in too similar a vein to what I want to implement, namely *multiplying uniques*
> That is, I want to implement a modifier that multiplies a unique N times, where N is a countable - that will solve a bunch of existing duplications
> Then, we can extend that with "countables with conditional state as input" for really explosive results
> 
> The difficult part is that existing conditionals are "true/false" and do not allow for multiplication - so this would probably need to be implemented in IHasUniques.getMatchingUniques
> 
> Now that I write it out, an initial implementation might not be difficult at all... interesting 🧐 

Multiplying modifiers? I thought about it. The issue is that existing conditionals are binary.

Maybe we should add the new unique category called "Multiplying Modifiers"? They would multiply the base effect or trigger by their value.

Something like this:

```
"Provides [1] [Iron] <per every adjacent [Mountain] tile>"
```

I would not consider it as conditional, because it multiplies the effect, not determine in the binary way.

PR #11641, opened 2024-05-23
"for every [countable]" unique modifier

yairm210:
See #11638

There are a couple of changes here that could be separated to different PRs:
- Countables yield *int* not *float*, as their name indicates they should
- "relevant X" pertaining to StateForConditionals are part of the class

As for the actual change, it's in IHasUniques.kt and Uniques.kt
@SomeTroglodyte Maybe you have an idea of how to turn the sequence into being efficient in a way other than "flatmap to new sequences", which seems so *wasteful* when most of the time it'll be single-use sequences

Unlike lists, which have a java.util.Collections.singletonList, I can't find a similar object for singletonSequence...


yairm210:
After-the-uniques math would mean changing every usage of "foreach unique in getmatchinguniques" instead of having it be a single-place change :/

yairm210:
And wow, I felt my mind expand when I understood what you did with the endless sequence plus take. That's *very* clever and certainly more efficient than my solution

PR #11647, opened 2024-05-24
Centralize and organize use of special unicode codepoints

SomeTroglodyte:
## Motivation

<details><summary><code>&lt;story mode&gt;</code></summary>

I bricked my distro.. No not quite, but some `apt purge` orgy made Unciv miss the sorting up/down arrows. Turns out some maintainer maybe thought "oh that half-width unicode block is full of hangul characters, so it must be relevant for asians only" - or not, but removing bloat by uninstalling asian fonts removes the glyphs for U+FFEA and U+FFEC from the default font.

The interesting thing is, I had already determined months ago by reading down into the sources that a NativeFontImplementation like we're getting from the underlying libraries is not 1:1 mapped from a system font, but it has the ability to collect glyphs from several system fonts... Which is exactly what I now observed.

I'll probably fix my system eventually, but in the meantime...

`</story mode>`
</details>

## Content notes
- First, it moves all uses (those I found) of higher codepoints (that are displayed) into Fonts.kt so there's a central place to exchange them if needed, and an easier cross-reference leveraging Studio Alt-F7/Ctrl-B and such.
- Next, I made the lookup performed on `allSymbols` optional - falls back to font if no texture found.
- Then I added all special symbols that don't yet have a texture to `allSymbols`
- This enables a minimod to supply the missing arrows: [SortIcons.zip](https://github.com/yairm210/Unciv/files/15440438/SortIcons.zip) - essentially just an atlas with those 2 texture paths.

## Question
- Idea: Define optional texture path `FontGlyphs/U+hhhh` - texture lookup is a hashmap lookup thus fast enough[^1] - if modder supplies patch as glyph into font. Would enable the above too, and open more possibilities, up to faking the old '.fnt' bitmap fonts... Not useful, but just adding a few glyphs wherever the modder chooses for free use in their texts... why not? Klingon mod...
- Didn't implement that right away as it might be a bit harder to understand for a mad hatter - pardon, modder
- Is the doc too much? Nobody cares about unicode codepoint names? Nobody wants to mod the Mayan glyphs?

[^1]: So I don't forget: There's `Int.toString(radix)` and `String.padLeft(num, char)`

### DIFF
--- a/core/src/com/unciv/ui/components/fonts/Fonts.kt
+++ b/core/src/com/unciv/ui/components/fonts/Fonts.kt
@@ -98,6 +98,7 @@ object Fonts {
         return pixmap
     }
 
+    //region Symbols added to font from atlas textures
     const val turn = '⏳'               // U+23F3 'hourglass'
     const val strength = '†'            // U+2020 'dagger'
     const val rangedStrength = '‡'      // U+2021 'double dagger'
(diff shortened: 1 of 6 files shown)


yairm210 (approved):
I think this is unnecessary, but I can't see anything bad, per se

PR #11650, opened 2024-05-26
Spy max rank can be modded

tuvus:
Builds off of #11559.

Changes:

- The max spy rank is now moddable.
- I changed OneTimeSpiesLevelUp to take in an amount so that modders can give spies multiple levels at once.
- If a spy is leveling up more than one rank then it will get a slightly different notification showing how many levels.
- I also added a new spy rank UI to show up to 39 levels.

SpyRanks0
SpyRanks1
SpyRanks2

</details> 

Note: the orientation in the tiny stars got rotated. This wasn't purposeful and I think it looks a little worse but I couldn't find a way to fix it. Feel free to fix it if you can.


tuvus:
> Me personally I don't like the 8-space continuation indent responsible for some diff here, and often default, but... matter of taste?

I completely agree, the 8-space is way too much

By got rotated, I mean the table is filled differently, The stars are now filled from top the top left row first instead of what should be the bottom left column first. Here would be the ordering of the stars:

3 6 9
2 5 8 
1 4 7

tuvus:
> An I imagined the pixels were...
1 4 7
2 5 8
3 6 9

Either way, as long as 1,2,3 are all in one column it saves space.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -796,7 +796,7 @@ enum class UniqueType(
     OneTimeRevealCrudeMap("From a randomly chosen tile [positiveAmount] tiles away from the ruins, reveal tiles up to [positiveAmount] tiles away with [positiveAmount]% chance", UniqueTarget.Ruins),
     OneTimeGlobalAlert("Triggers the following global alert: [comment]", UniqueTarget.Triggerable), // used in Policy
     OneTimeGlobalSpiesWhenEnteringEra("Every major Civilization gains a spy once a civilization enters this era", UniqueTarget.Era),
-    OneTimeSpiesLevelUp("Promotes all spies", UniqueTarget.Triggerable),  // used in Policies, Buildings
+    OneTimeSpiesLevelUp("Promotes all spies [amount] times", UniqueTarget.Triggerable),  // used in Policies, Buildings
```
yairm210, line 799:
```suggestion
    OneTimeSpiesLevelUp("Promotes all spies [amount] time(s)", UniqueTarget.Triggerable),  // used in Policies, Buildings
```

What do you think?

tuvus: I looked at that before and decided to keep it consistent with the other uniques:

- Cities are razed [amount] times as fast
- When spreading religion to a city, gain [amount] times the amount of followers of other religions as [stat]
- Upon capturing a city, receive [amount] times its [stat] production as [civWideStat] immediately
- [amount] times

Compared to only

- [amount] additional time(s)


Either way works, if you want to make that transition.

yairm210: There's a difference between "times" in the meaning of "A * B" - which is always "times", you don't say "1 time 5" you say "1 times 5" - and in the meaning of "repeating a certain amount", where you would say "1 time"

In all but the last one, it's the mathematical meaning
In the last one, it *is* the repeating meaning, but "1 times" doesn't make sense *for that modifier*, since it would have zero impact

yairm210 (approved):

android/assets/jsons/translations/template.properties
```diff
@@ -1760,6 +1760,7 @@ Move =
 After an unknown civilization entered the [eraName], we have recruited [spyName] as a spy! = 
 We have recruited [spyName] as a spy! = 
 Your spy [spyName] has leveled up! = 
+Your spy [spyName] has leveled up [amount] times! =
```
yairm210, line 1763: This should also be time(s)

tuvus: In this case I am using the previous text `Your spy [spyName] has leveled up!` when we level up only one time. It looks a bit nicer and will be used much more often.

PR #11651, opened 2024-05-27
Better portrait city-screen UI

tuvus:
This PR aims to improve the portrait experience by improving the city-screen UI, but is more of an attempt than an actual solution.

The main problem with the city-screen is that there isn't enough space for the two panels and to actually see the tiles as well.
As a solution I created a tab system to only show one of them at a time. It might look nicer if these buttons were rectangular, the same size and took an entire width.
As you can see from the pictures it's still a little rough, especially with the tile selection in the bottom right. And that add to queue button might be a little off.

Also, I want to put the multiplayer status button in line with the AutoPlay status button instead of on a completely new row but I haven't been able to figure it out. 

Any thoughts?

VerticalBuildings
VerticalCityPage
VerticalConstructions
VerticalStatusButtons

</details>

### DIFF
--- a/core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
+++ b/core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
@@ -131,7 +131,7 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
     fun addActorsToStage() {
         cityScreen.stage.addActor(upperTable)
         cityScreen.stage.addActor(lowerTable)
-        lowerTable.setPosition(posFromEdge, posFromEdge, Align.bottomLeft)
+        lowerTable.setPosition(posFromEdge, if (cityScreen.isPortrait()) posFromEdge + 60f else posFromEdge,  Align.bottomLeft)
     }
 
     fun update(selectedConstruction: IConstruction?) {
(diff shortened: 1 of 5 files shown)


tuvus:
Wow, you seem to actually clean up your remote branches. Maybe I should do that too.

tuvus:
For reference, here are some pictures I took of SomeTroglodytes branch:

SomePortrait1
SomePortrait2
SomePortrait3
SomePortrait4
SomePortrait5

</details> 

This UI seem a bit better than the one I proposed. Here are a few notes:

- I think the construction menu should be just one column with current queue on top, then we have more space for each button and just the little info panel on the side.
- The buildings tab looks a little empty, but that isn't nessesarily a bad thing.
- The stats table could look a little better but it works.

yairm210:
You're both better UI-ers than I am, so if you both agree on a direction I'll go with it
Not sure if you want me to merge this 🤔

tuvus:
>either Oscar goes small steps
or Oscar goes big, takes my branch

Woah, I'm busy working on the AI. Your the one who can actually design good UI. I just hack things together until the result is slightly better than the original to get your attention. Then you swoop in and make it look good.

I'll probably choose option 3 until I build up my courage to go fight the layouts again... Unless you have some spare time and motivation?

tuvus:
@SomeTroglodyte Could you merge master of your branch and deal with the merge conflicts? You made a big change that makes it hard to resolve without context.

PR #11664, opened 2024-05-28
Console: Create natural wonder and `tile find`

SomeTroglodyte:
Usage:
- `tile setterrain natural-wonder`: autocomplete and cli-input support
- `file find tilefilter`: Neither autocomplete nor cli-input support, meaning you need to supply the filter case-sensitive and quoted if need be, same as `checkfilter`. Uses same notification concept as "show me resources" from the overview etc

Drawback:
- Removing a Natural Wonder is way harder - no support yet to "unsee" one. The easy code would leave the happiness benefit for civs that already have it. Do do it properly - would need to take into account a civ may have seen a Great Barrier Reef tile, but may or may not have seen the other one. Too much effort.

### DIFF
--- a/core/src/com/unciv/ui/screens/devconsole/ConsoleTileCommands.kt
+++ b/core/src/com/unciv/ui/screens/devconsole/ConsoleTileCommands.kt
@@ -3,9 +3,15 @@ package com.unciv.ui.screens.devconsole
 import com.unciv.Constants
 import com.unciv.logic.city.City
 import com.unciv.logic.civilization.Civilization
+import com.unciv.logic.civilization.LocationAction
+import com.unciv.logic.civilization.Notification
+import com.unciv.logic.civilization.NotificationCategory
+import com.unciv.logic.civilization.NotificationIcon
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/ui/screens/devconsole/ConsoleTileCommands.kt
```diff
@@ -124,8 +128,48 @@ class ConsoleTileCommands: ConsoleCommandNode {
...
         },
+
+        "find" to ConsoleAction("tile find <tileFilter>") { console, params ->
+            val filter = params[0]
+            val locations = console.gameInfo.tileMap.tileList
+                .filter { it.matchesFilter(filter) }
+                .map { it.position }
+            if (locations.isEmpty()) DevConsoleResponse.hint("None found")
+            else {
+                val notification = Notification("tile find [$filter]", arrayOf(NotificationIcon.Spy),
```
yairm210, line 139: That is super cool!

yairm210 (approved):

PR #11673, opened 2024-05-31
Improve update of "Last seen improvement"

SomeTroglodyte:
... But actually, just saving the improvement name while a civ is no longer actively seeing a tile is not enough. Should almost save the entire Tile - pillaging, roads built or removed, forests chopped or replanted, newer "terraforming" triggerable changing resource?? - everything affecting visuals shouldn't change for a civ that has explored but can't see. Way out of scope here, however.

### Questions
- Removing the call in the "Building creates improvement" code is 99.9% fine - but untested. The main goal of this PR should manage that just fine - but the conditions tested aren't perfectly identical. Anyone has a save to test?
- Changes edge cases for spectators - 99.99% sure it's better this way.
- The one place where the set is next to a `setExplored` - maybe _all_ `setExplored` should update lastSeenImprovement???
- The change in `updateTilesForSelectedUnit` is imho actually a bugfix, but since it only affects dimming of a layer, nobody noticed so far? Got no idea how to test this one.


core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -209,11 +211,20 @@ class TileImprovementFunctions(val tile: Tile) {
...
                 tile.improvement = improvementName
+                improvementFieldHasChanged = true
 
                 removeCreatesOneImprovementMarker()
             }
         }
 
+        if (improvementFieldHasChanged) {
+            for (civ in tile.tileMap.gameInfo.civilizations) {
+                if (civ.isDefeated() || !civ.isMajorCiv() || civ.isSpectator()) continue
```
yairm210, line 222: I'm pretty sure spectators are not major civs
```suggestion
                if (civ.isDefeated() || !civ.isMajorCiv()) continue
```

yairm210 (approved):
LGTM

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
yairm210: You can just "commit suggestion" :)

PR #11677, opened 2024-06-01
Improvement queue

SomeTroglodyte:
Allows several improvements (or "instruction" improvements) to be queued per Tile.

UI is the absolute minimum I could come up with - see screenies.
Compatibility is pretty simple - with the new code, data "lives" in a new List (used as Fifo queue), and the legacy properties are transient via getters. With the old code - you know. Custom Json.Serializable in new code does all that's necessary for both-ways partial compatibility:
- Reads old saves: Looks for the old property names in Json if the queue is empty and reads them.
- Writes extra data to allow loading new saves in old code - queue entries beyond the top one are lost.

In other words, if we revert the UI commit and let this "run" for a few versions nobody will even notice, but if not - nobody loses much.

Phase-out of the old properties is mostly deleting the "old property" getters, removing the entire custom json and fixing dangling uses of the old properties (all 38 of them) - and testing if serialization stays the same, at the moment it's clean without explicit `"class":` constructs. Maybe a global review of all `startWorkingOnImprovement` calls (this one always clears the queue before adding so it behaves same as before) to see where modifying an existing queue would make more sense. Some idea for an UI to show future tasks (none in here). 

This guy:
image

Gets the task to clear the Forest and immediately afterwards build a Plantation:
image

Two turns later, the Forest is gone and the Plantation is worked on without any further user assistance - and without the Worker being automated:
image

</details>

This eases the work for people who prefer micromanagement a little, but it can potentially also help the AI not to lose track so easily.... Not my area of expertise. @tuvus: Forewarned!


core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -852,15 +872,39 @@ class Tile : IsPartOfGameInfoSerialization {
...
+        queueImprovement(improvement.name, turns)
+    }
+
+    /** Adds an entry to the [improvementQueue] with explicit [turnsToImprovement] */
+    fun queueImprovement(improvementName: String, turnsToImprovement: Int) {
+        improvementQueue.add(ImprovementQueueEntry(improvementName, turnsToImprovement))
+    }
+
+    /** Called from UnitTurnManager when a Worker "spends time" here */
+    fun endTurn(worker: MapUnit) {
```
tuvus, line 897: This should be renamed to something more like `doWorkerTurn` to clarify that we are doing improvement business.

core/src/com/unciv/logic/map/mapunit/UnitTurnManager.kt
```diff
@@ -19,7 +18,12 @@ class UnitTurnManager(val unit: MapUnit) {
         if (unit.currentMovement > 0
                 && unit.getTile().improvementInProgress != null
                 && unit.canBuildImprovement(unit.getTile().getTileImprovementInProgress()!!)
-        ) workOnImprovement()
+        ) {
+            val tile = unit.getTile()
+            tile.endTurn(unit)
+            tile.getCity()?.updateCitizens = true
```
tuvus, line 24: You are updating the citizens every time a worker works on an improvement in the city. Is this necessary?
Previously we only updated the citizens once the improvement was completed.

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -168,6 +186,10 @@ class Tile : IsPartOfGameInfoSerialization {
     private var isAdjacentToRiver = false
     @Transient
     private var isAdjacentToRiverKnown = false
+
+    val improvementInProgress get() = improvementQueue.firstOrNull()?.improvement
```
tuvus, line 190: We have: improvementInProgress, turnsToImprovement, hasImprovementInProgress and getTileImprovementInProgress() in this class. Are the last two nessessary? If so can you move them to be together?

core/src/com/unciv/ui/screens/pickerscreens/ImprovementPickerScreen.kt
```diff
@@ -144,7 +152,14 @@ class ImprovementPickerScreen(
...
 
             val explanationText = when {
-                proposedSolutions.any() -> proposedSolutions.joinToString("}\n{", "{", "}").toLabel()
+                proposedSolutions.any() -> {
+                    val text = proposedSolutions.joinToString("}\n{", "{", "}")
+                    if (removalImprovement != null && proposedSolutions.size == 1)
+                        "$text\n{Pick now!}".toLabel().onClick {
+                            accept(removalImprovement, improvement)
+                        }
+                    else text.toLabel()
```
tuvus, line 161: Are we able to center this text so that it looks more in line with the pick now?

core/src/com/unciv/ui/screens/pickerscreens/ImprovementPickerScreen.kt
```diff
@@ -181,7 +196,7 @@ class ImprovementPickerScreen(
             improvementButton.onActivation(type = ActivationTypes.Tap, noEquivalence = true) {
                 selectedImprovement = improvement
                 pick(improvement.name.tr())
-                descriptionLabel.setText(improvement.getDescription(ruleSet))
+                descriptionLabel.setText(improvement.getDescription(ruleset))
             }
 
             improvementButton.onDoubleClick { accept(improvement) }
```
tuvus, line 202: (4 lines below)
Can we enable the button when it is queable as well? Maybe changing the color would also be helpful.

core/src/com/unciv/ui/screens/pickerscreens/ImprovementPickerScreen.kt
```diff
@@ -39,22 +39,25 @@ class ImprovementPickerScreen(
...
             // no onAccept() - Worker can stay selected
         } else {
-            if (improvement.name != tile.improvementInProgress)
+            if (improvement.name != tile.improvementInProgress) {
                 tile.startWorkingOnImprovement(improvement, currentPlayerCiv, unit)
+                if (secondImprovement != null)
+                    tile.queueImprovement(secondImprovement, currentPlayerCiv, unit)
+            }
             unit.action = null // this is to "wake up" the worker if it's sleeping
             onAccept()
```
tuvus, line 62: I think it would be better to think of this recursively. What if we wanted to build an upgraded improvement after the next improvement? The list could go on more than twice, so can we modify this to allow for longer improvement chains?

tuvus:
Overall, I like the idea and it's nice to see some progress on a feature that has been talked about a lot. The UI needs just a few tiny improvements but otherwise I'm fine with it (nothing I can complain about). Other than the few areas that I have addressed it seems like a step in the right direction.

> but it can potentially also help the AI not to lose track so easily.... 

I haven't seen any problems with the workers loosing track recently. But it will probably be necessary for modding if they want to build improvements that require other improvements. The AI would look at the last improvement and try to build it, then the queue would add all the prerequisites.

core/src/com/unciv/logic/map/tile/Tile.kt
```diff
@@ -56,13 +61,27 @@ class Tile : IsPartOfGameInfoSerialization {
     var resourceAmount: Int = 0
 
     var improvement: String? = null
-    var improvementInProgress: String? = null
```
yairm210, line 59: Players will lose all existing improvements in progress. This is not a game-changing (literally) loss, but it is something we want to avoid - especially since in MP games if one is ahead and one is behind, they'll *both* lose their ability to construct things! BAD

Something to add to BackwardsCompatibility.

yairm210: You tested this?

core/src/com/unciv/logic/map/tile/Tile.kt
tuvus: I meant more like deleting getTileImprovementInProgress() and moving hasImprovementInProgress() under turnsToImprovement.

core/src/com/unciv/ui/screens/pickerscreens/ImprovementPickerScreen.kt
yairm210: Agree, if the improvement list is of arbitrary length then this function should recieve a vararg or list

yairm210 (approved):

PR #11678, opened 2024-06-01
Enable buildingFilter values as cityFilter

PLynx01:
As I announced several days ago, I implemented the filtering of the cities with specified building(s), using buildingFilter values as cityFilter.

But there is an issue with wording:
`"[+5 Gold] [Currency]"` means the additional 5 Gold per turn per every city with buildings enabled by Currency tech.

It should be displayed like this:
`5 Gold in cities with Currency building`

I don't know how the uniques are processed from json files to display-ready form.
Explain me that.

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -457,7 +457,7 @@ class City : IsPartOfGameInfoSerialization {
             // this will always be true when checked.
             "in cities following this religion" -> true
             "in cities following our religion" -> viewingCiv?.religionManager?.religion == religion.getMajorityReligion()
-            else -> civ.matchesFilter(filter)
+            else -> civ.matchesFilter(filter) || cityConstructions.getBuiltBuildings().any { it.matchesFilter(filter) }
         }
     }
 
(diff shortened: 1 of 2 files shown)


yairm210:
Uniques are all *in plain readable English*.
Therefore translations are simple. 
This change is *not* readable English - ”[+1 Gold] [aqueduct]" is not human parseable, so we need a different solution
There's no magic that can fix this, it needs to be human-readable from the json

PLynx01:
> Uniques are all _in plain readable English_. Therefore translations are simple. This change is _not_ readable English - ”[+1 Gold] [aqueduct]" is not human parseable, so we need a different solution There's no magic that can fix this, it needs to be human-readable from the json

I guess we need to make more separate uniques with buildingFilter

PLynx01:
@yairm210 

I've got another idea. We already have conditional uniques for cities with/without buildngFilter.

Maybe should I tweak them in order to be compatible with more uniques?

yairm210:
Sounds like an idea
What is incompatible about them?

PLynx01:
> Sounds like an idea
> What is incompatible about them?

I need to test it.
This of course applies to effect itself, not its other conditionals.

BTW, I think that "in all cities" cityFilter is ambiguous because documentation does not state whether it's Civ-wide or Game-wide.

yairm210:
In any case this gets a "no" from me

PR #11680, opened 2024-06-01
Fixed spy steal tech timer

tuvus:
Improves #11657

This PR coves the edge case for the steal tech time of when the city produces zero or negative science. The original issue was reported on discord.

Previously this would result in a very large and confusing value.

Now the Spy will be "Observing City" instead of stealing the technology. In the case that the city has positive science, then goes negative and then returns to positive, the spy won't loose it's progress. It only losses progress when StartStealingTech() is called I.E. when it moves to another city.

### DIFF
--- a/core/src/com/unciv/models/Spy.kt
+++ b/core/src/com/unciv/models/Spy.kt
@@ -112,15 +112,17 @@ class Spy private constructor() : IsPartOfGameInfoSerialization {
                 if (!getCity().civ.isMajorCiv()) return
 
                 val stealableTechs = espionageManager.getTechsToSteal(getCity().civ)
-                if (stealableTechs.isEmpty()) return
+                if (stealableTechs.isEmpty() || getTurnsRemainingToStealTech() < 0) return
                 setAction(SpyAction.StealingTech) // There are new techs to steal!
             }
             SpyAction.StealingTech -> {
                 turnsRemainingForAction = getTurnsRemainingToStealTech()
 
-                if (turnsRemainingForAction == -1) {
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/models/Spy.kt
```diff
@@ -112,15 +112,17 @@ class Spy private constructor() : IsPartOfGameInfoSerialization {
                 if (!getCity().civ.isMajorCiv()) return
 
                 val stealableTechs = espionageManager.getTechsToSteal(getCity().civ)
-                if (stealableTechs.isEmpty()) return
+                if (stealableTechs.isEmpty() || getTurnsRemainingToStealTech() < 0) return
                 setAction(SpyAction.StealingTech) // There are new techs to steal!
```
tuvus, line 116: Note that we don't actually reset progressTowardsStealingTech when beginning stealing technology again.

core/src/com/unciv/models/Spy.kt
```diff
@@ -151,22 +153,25 @@ class Spy private constructor() : IsPartOfGameInfoSerialization {
     }
 
     private fun startStealingTech() {
-        setAction(SpyAction.StealingTech)
         progressTowardsStealingTech = 0
-        turnsRemainingForAction = getTurnsRemainingToStealTech()
+        setAction(SpyAction.StealingTech)
+        endTurn()
```
tuvus, line 158: We want to check immediately if we can steal technology and maybe send a message.
In a very rare case where the technology cost is 0 we might even steal a technology on this turn.

Issue #11017, opened 2024-01-26
Mod Management Menu Never Finishes Loading


carriontrooper:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.9.19 to 4.10.3

### Describe the bug

The red reloading sign on the upper right of the mod manager screen never turns into the usual mod filter selection dropdown menu.

### Steps to Reproduce

1. Click on Mods from main menu
2. Look at upper right corner, red reload symbol
3. Wait
4a. It never turns into the filter menu
4b. On the off chance it does, it doesn't let you re-sort the mods and a warning popped up saying something like 'hasn't finished loading yet'

### Screenshots

![image](https://github.com/yairm210/Unciv/assets/78248399/d1eecdc4-c50b-4a03-b94d-35d80258ba31)

### Operating System

Android

### Additional Information

Problem also exists on PC version

carriontrooper:
Problem is, it worked just fine in versions prior to 4.9.19. - filtering menu loads in seconds, whether via phone or PC, using phone packet data or laptop wifi. I don't think it's the rate limit that's causing the issue.

SeventhM:
4.10.3 patch 1 on my phone. Loads just fine for me. Just takes some amount of seconds to load. So can't reproduce here

carriontrooper:
Hmm, I haven't updated it to patch 1, let's see if that fixes things

SeventhM:
> Is that loading symbol turning?

Uh... Is it supposed to turn? I don't think I've seen it do so

SeventhM:
Looks like it's off by default

carriontrooper:
> Is that loading symbol turning? Du you have the "continuous rendering" option on? Should be visual only, but maybe it influences some events/actions processing.. Question valid for Phone + as you say "Problem also exists on PC version". What else do your boxes have in common that may be different from ours?

Loading symbol is not turning on both machines, and it's definitely not a bandwidth throttle since I can play War Thunder and other bandwidth-heavy games on my PC just fine without lags. So that rules it out on PC.
Continuous rendering I haven't touched since installing the game long ago, it's off.

carriontrooper:
> Just tested again, this time on Android hardware. With and without "continuous rendering", landscape and portrait. The query always finishes, filtering and sorting get enabled just fine. With such repetition of course I hit the rate limit, so the last test needed 40 seconds to finish.
> 
[...]

Issue #6758, opened 2022-05-11
Missing mods in Downloadable mods list


stixZman:
Hello,

Since a few days ago (I´m not sure if it´s related to latest 4.0.14 release or not) I´m missing some of the most starred and downloaded mods for Unciv in Android app (don´t know if it occurs in the desktop app)

If I navigate to unciv mod topic URL [https://github.com/topics/unciv-mod](https://github.com/topics/unciv-mod) I´m missing some repositories like 5Hex tileset, RekMod and many more.

Can anyone help me to download these mods from repositories? I uninstalled Google Play Store version in order to install apk directly from latest github release (since it takes a few days to update the release on Play Store)

Thank you very much in advance.

SomeTroglodyte:
If you know the repo URL, there's a manual button (accepts repo url's if that repo uses 'master' as primary branch name, or a complete branch url).
![image](https://user-images.githubusercontent.com/63000004/167845594-66cb8fc6-a75a-43d4-8eb5-8844dd21510e.png)

But thanks for the heads-up. Can't believe querying directly on github also fails, but forewarned is good. ...Yes, seems a github failure. Nothing we can do. Maybe you could do us the favour to escalate directly to github?

Ouaz:
Yeah, that's weird, 5Hex and Rekmod are not listed... although they have the unciv-mod tag. Sorting by number of stars, 5Hex should be on top of the list...

You can manually add these mods on Android by connecting your device to your PC (a bit cumbersome, but that's the only workaround until Github fixes the problem... if they fix it):

1) On your PC, go to https://github.com/ravignir/5Hex-Tileset, click on "Code" (green button), and select "download zip"
2) Connect your android device to your PC through USB, and select "Explore"
3) Navigate to  `Android/data/com.unciv.app/files/`. Here, creates a "mods" folder, and inside it, a `5Hex-Manual` folder. So it gives `Android/data/com.unciv.app/files/mods/5Hex-Manual`.
4) Open the zip file, and copy the `jsons` folder,` game.atlas`, `game.png`, `game2.png`, `game3.png`, `game4.png` in `Android/data/com.unciv.app/files/mods/5Hex-Manual` created in point 3 (`images` folder is not needed).
5) Unplug your Android device from you PC (eject/remove USB device if available) and start Unciv. 5Hex-Manual should appear in your installed mods list. Enable it as usual.

[...]

PR #11682, opened 2024-06-02
Auto-update mods option (enabled by github access token)

### LINKED ISSUE #10903: Auto-update mods option
### Before creating

### Problem Description

It's not a big deal, but most people usually take a long time updating their mods, or we just don't notice that there is a new update, and we play with the old version of the mod with our friends.

### Related Issue Links

_No response_

### Desired Solution

It would be nice to have an option so when you open the game, it automatically starts updating the mods that have an update available.

If I want to join a game with a mod that is currently updating, then the game will stop me and make me wait for the mod that will be used to be updated.

### Alternative Approaches

Or, a button inside the Mod Manager to "Update All".

### Additional Context

_No response_

### LINKED ISSUE #11023: Feature request: Mod Manager auto update mods added via github that don't have tags
### Before creating

### Problem Description

Currently, for all mods with tags, it automatically checks for an update and sees if there's an update available. However, If you use a mod without a tag, e.g. my "secret" rebalanced deciv project (leaving undiscoverable partially because it's unfinished, partially because we already have 2-3 discoverable Decivs, partially because one of the Deciv maintainers claim they're working on a big update anyways, and partially because lazy), then it won't ever check for updates despite the mod manager still generating modOptions for it just fine

### Related Issue Links

_No response_

### Desired Solution

```
Update. Anyways, I believe this is how I broke the form last time. Maybe the form should avoid breaking like this

### Alternative Approaches

Maybe

### Additional Context

Speaking of autoupdating ModOptions, are we using the wrong line clears? Even Notepad++ shows all of the text in 1 long line. Same thing happens to save files

SomeTroglodyte:
### References
- Closes #11023
- Closes #10903
- Might have helped #11017
- See #6758
- See #10900

### Notes
- You MUST provide a github "personal access token" for any of this to show. Advanced tab. Looks like ''ghp_" plus 36 base62 chars.
- I fully expected the mod manager to speed up markedly, but got disappointed. The data that goes over the wire is maybe 20% of what it is with the REST api, but that seems to not be the bottleneck.
- The test container was crucial to get this to work, so I left it in. Pollutes the build runner console output with a few more counts of ignored tests.
- Flow is as follows:
        - Start Unciv (or change access token while options is open from main menu).
        - Immediately starts a background query to github asking for limited metadata of just the mods you installed, query and answer have chances to fit in one packet.
        - While this runs there's the old animated loading icon in the bottom left corner of the Mods button - and it's grayed.
        - When the answer comes, is can be a failure or partially successful: Errors are displayed, the prevalent one (you got a mod without public repo) made translatable.
        - The "Ignore" button of the error popup causes that exact same set of errors to be ignored on subsequent starts
        - Timestamps are compared, and if any mod is outdated the Mods button text changes to "Update Mods"
        - Clicking "Update Mods" asks whether you want to update, listing the mods, and a No goes directly to the mod manager
        - "Yes" starts another background job downloading and installing the outdated mods
        - While this runs the Mods button is grayed and disabled - but all others are not and will cancel the job
        - I' convinced browsing Civilopedia while the autoupdater finishes is possible - untested.
- I'm still convinced the idea behind 10900 was ***right***. Mods that don't do releases - unchanged as before, but mods that do - display them and use their timestamps for update checks instead. Playing a post-release per-commit version of a release-using mod requires manual URL entry. Good Idea :tm:. All relevant data is already prepared in the GraphQL framework.

### Questions
- I got a PAT belonging to a testing [account](https://nopaste.boris.sh/#XQAAAQAeAAAAAAAAAAA0HSzgdXPvDKmx8h5qwu8kgvI3F/kB7e+506sb9Q47HCf///3u8AA=) that does nothing else. We could cheat a little and get that into Unciv - except gitshrub recognizes it in source and blocks such pushes. But - it's a "political" decision first. Should we dare?
- ConfirmPopup... My what a long way from YesNoPopup. Almost obsolete - only three uses, wonder why. Could drop and reroute to the "with details" one, and maybe find other potential clients that bypassed ConfirmPopup for similar reasons... Not this PR.
- In that little thing I replaced the use of addGoodSizedLabel - just to distribute wrapped words better - and that only because I test with screen size tiny and a small video clip friendly window. Patch the `addGoodSizedLabel` method globally instead?

### Visuals
<details>

An outdated video - too lazy to artificially outdate mods to capture a fresh one
(and the percentage is off due to boxing closures and concurrency - long fixed):

https://github.com/yairm210/Unciv/assets/63000004/3f334d44-70db-47bf-9683-4729ada20c4d

"wanna update" question:
image

"hey errors" popup:
image

</details>


AdityaMH:
"Update All" button inside mods page also better.


> [image](https://private-user-images.githubusercontent.com/63000004/335839109-6f7ead11-dbd9-4dde-b5a6-4ec276982f47.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTczNjU1NzcsIm5iZiI6MTcxNzM2NTI3NywicGF0aCI6Ii82MzAwMDAwNC8zMzU4MzkxMDktNmY3ZWFkMTEtZGJkOS00ZGRlLWI1YTYtNGVjMjc2OTgyZjQ3LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA2MDIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNjAyVDIxNTQzN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWE1YWZmZDVhN2M1NDZhZDMwN2FiYjI1YzdjOWQ3ZmE0OTQ4MTBmM2RmODM4Yzk1MWVjMDM2YmQzNjhlZmFlMTUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.ZqG70Sgg27dccnL-K8EiOsUjBa4a6Qm91OasdjTh4OM)

Great repo that never existed.

yairm210:
- PAT for testing account - no, will overload. We have ~17K daily active users consistently, on Google Play alone. We get 5000 requests per hour. Even if each request only counts as one, we'll be getting failures.
image
 
- I don't particularly Github PATs being required for this feature. If there's a convoluted path to being minorly more efficient, I think it actually results in a *worse* user experience overall

yairm210:
This one is still a "no" for me, so long as it uses a Github PAT.
I understand where you're coming from but this is not a direction I want the game to go in.
Would be happy to have an autoupdate that does not use PATs

PR #11702, opened 2024-06-06
Extra Civ and Spy moddability

tuvus:
I recently discovered the easy of making constants moddable by using ModConstants.kt. Here are some extra things that this PR makes moddable:

- Minimum war duration
- Turns until revolt
- Spy skill per rank
- City-State election cycle

Note: I renamed maxSpyLevel to maxSpyRank
Also Note: I refactored the City-State election to use Civ flags instead of it's own custom variable as it should. I tested it cross version and wrote it in a way so that it doesn't break. Each version removes the other versions counter/flag and tries to replace it. The old version starts back at 15 while the new version starts at a random value form 1 to 15. Then once all players upgrade to the new version it will count down like normal.

<sub>That diff looks annoying</sub>

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -304,7 +304,6 @@ class Civilization : IsPartOfGameInfoSerialization {
         toReturn.hasMovedAutomatedUnits = hasMovedAutomatedUnits
         toReturn.statsHistory = statsHistory.clone()
         toReturn.resourceStockpiles = resourceStockpiles.clone()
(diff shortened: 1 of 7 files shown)


yairm210:
Looks ok, but how will modders know these exist? I'm missing *some* form of documentation

tuvus:
For some reason I though they were auto-generated.
I also renamed minWarDuration to minimumWarDuration because I feel like it is more readable.

tuvus:
> Why?

It thinks I completely rewrote holdElections() when I didn't.

yairm210 (approved):

PR #11704, opened 2024-06-06
Create Galician.properties

ninjum:
lingua galega do unciv


tuvus:
New branch located at #11746.

PR #11705, opened 2024-06-07
Refactor: UncivTextField

SomeTroglodyte:
See commit titles. Intended to be 100% transparent. 

### Motivation
- These textfields aren't recognizable in our scene2d-debug overlay.
- More flexibility for functionality on top - e.g. color-coding text according to content meeting some criteria or somesuch.
- ... Yes, minor and there's other solutions for those. Still, I think readability is better.

### Questions
- The reorg is extensive in some cases, so everything should be tested - but I can't. There's the feature that if a textfield that has focus gets covered by the soft keyboard, and the auto-scroll-into-view fails or somesuch, a Popup with a stand-in TextField is opened. That one I can't trigger - anybody any ideas on how to provoke that? If not, must try hacking results in the debugger, or adding Screen Size settings, or....


yairm210 (approved):

core/src/com/unciv/ui/components/widgets/UncivTextField.kt
```diff
@@ -0,0 +1,192 @@
...
+    private inner class VisibleAreaChangedListener : InputListener() {
+        private val events = EventBus.EventReceiver()
+        init {
+            events.receive(UncivStage.VisibleAreaChanged::class) {
+                if (stage == null || !hasKeyboardFocus()) return@receive
+                Concurrency.run {
+                    // If anything resizes, it also does so with this event. So we need to wait for that to finish to update the scroll position.
+                    delay(100)
+                    withGLContext {
+                        if (stage == null) return@withGLContext
```
yairm210, line 80: In what situation would the stage have previously not been null, but would now be null? If the actor was removed?

PR #11718, opened 2024-06-09
Add a Unique allowing an Improvement to specify which Resource(s) it improves

SomeTroglodyte:
For: https://github.com/yairm210/Unciv/issues/3242#issuecomment-2156087504
- Note: Unsure this is worth it - and unsure whether the "manual lazy" can't be made a true lazy after testing.
- Note: This will apply the `improvementStats` when active - unlike the hardcoded "City center unlocks strategic and luxury resources under them" or the hardcoded "Great improvements unlock strategic resources". Thus, not an option to un-hardcode those in its present form.

<details><summary>testing improvement and effect</summary>

```json
	{
		"name": "Chinese Academy",
		"uniqueTo": "China",
		"replaces": "Academy",
		"terrainsCanBeBuiltOn": ["Land"],
		"science": 4,
		"uniques": [
			"Great Improvement",
			"[+4 Science] <after discovering [Education]>",
			"[+4 Science] <after discovering [Scientific Theory]>",
			"[+4 Science] <after discovering [Atomic Theory]>",
			"Improves [any] resource in this tile",
			"Pillaging this improvement yields approximately [+200 Gold]"
		]
	},
```

image

</details>


core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -24,13 +24,29 @@ class TileResource : RulesetStatsObject() {
...
-        else improvedBy + improvement!!
+    private var improvementsInitialized = false
+    private val allImprovements = mutableSetOf<String>()
+    private var ruleset: Ruleset? = null
+
+    fun getImprovements(): Set<String> {
+        if (improvementsInitialized) return allImprovements
+        val ruleset = this.ruleset
+            ?: throw IllegalStateException("TileResource initialization incomplete")
+        improvement?.also { allImprovements += it }
```
yairm210, line 35: Please change to 'if', this is obscure syntax for a simple check

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -24,13 +24,29 @@ class TileResource : RulesetStatsObject() {
...
-        if (improvement == null) improvedBy
-        else improvedBy + improvement!!
+    private var improvementsInitialized = false
+    private val allImprovements = mutableSetOf<String>()
+    private var ruleset: Ruleset? = null
+
+    fun getImprovements(): Set<String> {
+        if (improvementsInitialized) return allImprovements
+        val ruleset = this.ruleset
+            ?: throw IllegalStateException("TileResource initialization incomplete")
```
yairm210, line 34: Better exception - "no ruleset on resource when initializing improvements"

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -24,13 +24,29 @@ class TileResource : RulesetStatsObject() {
...
+        improvement?.also { allImprovements += it }
+        allImprovements.addAll(improvedBy)
+        for (improvement in ruleset.tileImprovements.values) {
+            if (improvement.getMatchingUniques(UniqueType.ImprovesResources).none { matchesFilter(it.params[0]) }) continue
```
yairm210, line 38: As a unique, I would expect it to accept tile modifiers, but it doesn't
Best solution is to make it do so and return a different set of improvements based on tile
Second best is to write in doc of unique "does not accept conditionals"

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -141,6 +157,13 @@ class TileResource : RulesetStatsObject() {
         }
     }
 
+    fun matchesFilter(filter: String) = when (filter) {
+        name -> true
+        "any" -> true
+        resourceType.name -> true
+        else -> improvementStats?.any { filter == it.key.name } == true
```
yairm210, line 164: I'm not sure I agree with this one, and it's missing the most important filter - uniques contains(filter)

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -629,6 +629,10 @@ enum class UniqueType(
     PillageYieldFixed("Pillaging this improvement yields [stats]", UniqueTarget.Improvement),
     Irremovable("Irremovable", UniqueTarget.Improvement),
     AutomatedUnitsWillNotReplace("Will not be replaced by automated units", UniqueTarget.Improvement),
+    ImprovesResources("Improves [resourceFilter] resource in this tile", UniqueTarget.Improvement, flags = UniqueFlag.setOfNoConditionals,
```
yairm210, line 632: I think "provides" instead of "improves" is clearer as to what this actually does

docs/Modders/Unique-parameters.md
```diff
@@ -206,6 +206,11 @@ For example: `+1 Science`.
 
 These can be strung together with ", " between them, for example: `+2 Production, +3 Food`.
 
+## resourceFilter
+
+At the moment, only used for the `"Improves [resourceFilter] resource in this tile"` Unique on Improvements.
```
yairm210, line 211: Please turn this into "allowed values" list, like most of them are

yairm210: A markdown list, with bullet points, like most of the parameters are displayed in this page

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -135,12 +172,21 @@ class TileResource : RulesetStatsObject() {
...
+    fun getImprovingImprovement(tile: Tile, civ: Civilization): String? {
         return getImprovements().firstOrNull {
-            tile.improvementFunctions.canBuildImprovement(civInfo.gameInfo.ruleset.tileImprovements[it]!!, civInfo)
+            tile.improvementFunctions.canBuildImprovement(civ.gameInfo.ruleset.tileImprovements[it]!!, civ)
         }
     }
 
+    fun matchesFilter(filter: String) = when (filter) {
+        name -> true
+        "any" -> true
```
SeventhM, line 184: Should maybe also add a check for `Constants.all`?

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: 👍

yairm210 (approved):
Good enough to merge

Issue #11730, opened 2024-06-11
Fail to create a new game after choosing "Tiny world size"


Zooms233:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.11.19(Build 1005)

### Describe the bug

As same as title

### Steps to Reproduce

1. Go to 'Start a new game'
2. Choose 'Tiny' in 'World size'
3. Keep number of civilisation default value(4)
4. Click 'Start game!'' and see the error
5. Though change the ’World size’ to Medium, error remains.

### Screenshots

![error](https://github.com/yairm210/Unciv/assets/129411234/4f4433c0-34e6-4881-9167-0b0b66178194)

### Operating System

Windows

### Additional Information

This error occurs on Android,too.
![Error on Android](https://github.com/yairm210/Unciv/assets/129411234/be90eb93-b12c-4c30-a05b-2d6d2c9cb645)

tuvus:
I have located the error. The fix isn't hard but will take some design thinking.

tuvus:
Done! Thanks for reporting this!

AutumnPizazz:
It also happens when choosing more than 0 city-states.

PR #11731, opened 2024-06-11
Fix election crash

tuvus:
This PR fixes the issue raised in #11730. 
The problem was that a city-state was trying to check if espionage was enabled to determine if it should generate the election flag. However the civ.gameInfo variable wasn't initialized yet. I looked and it turns out no transient variables would be initialized at this time. 
So I decided to remove it. The effect will be the same since the flag will be created later in TurnManger.kt. However this does change the plans of removing it from there.


yairm210:
release patch

yairm210:
Btw @tuvus @SomeTroglodyte you can also "release patch" if you see such a problem :)

tuvus:
Releasing it immediately would have probably saved some problems. However it does skip the code review...

PR #11744, opened 2024-06-12
Add "release patch" section to the wiki

SomeTroglodyte:
Sooo...
- What's the relationship "releaze patch" to Kraken Play Store? I don't see how Pray Store _could_ get such a patch release... After all, the workflow ran without errors on my fork, but I don't have a secret there for them. 🕵️‍♀️ ?
- Pretty sure F-Droid will see eventually, but what about itch.io and/or Steam?
- Add some encouragement/deterrent to _when_ a Collaborator should feel justified or not?
- @tuvus - input?


yairm210 (approved):

yairm210:
"patch" releases are full releases, just e with a very small change set and no changelog
So Google Play treats it as just another version

tuvus (approved):
You might want to add something like: The version will change and a little more information on who gets it in google play. Do they all get it immediately at 100% rollout? Or do only the ones who have it need to update to the patch and the others will get the patched version once the rollout comes to them?

PR #11745, opened 2024-06-13
Adds unique IDs to units

yairm210:
## Why?

Civs and cities have unique identifiers, units do not.
This means that whenever something happens to a unit, the best indication we have is "name + location", which - if the unit then undergoes changes - is inaccurate.
This is mostly relevant only for inter-turn things, especially notifications, since there we can't save a reference to a specific unit - but this also rarely happens for intra-turn things

We've seen this play out for a while, and we sort of patch it each time - with air units carried by other units, with units that are "bound" to the other unit on their tile, etc.
If we ever want to accommodate multi-unit-per-tile, this must be solved as well.

As for my impetus for this - I want to be able to target *unit triggers* for events, and for that - we need to serialize the event identifier between turns. Instead of patching this for the upteenth time by using a position, let's *actually* solve this problem properly.

Currently these IDs do nothing.

### DIFF
--- a/core/src/com/unciv/Constants.kt
+++ b/core/src/com/unciv/Constants.kt
@@ -4,6 +4,7 @@ object Constants {
     const val settler = "Settler"
     const val eraSpecificUnit = "Era Starting Unit"
     val all = setOf("All", "all")
+    const val NO_ID = -1
 
     const val english = "English"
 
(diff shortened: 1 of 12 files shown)


yairm210:
Correct, will fix for upcoming version

PR #11762, opened 2024-06-16
Civs can no longer declare war right after peace with a city-state

tuvus:
This PR addresses a request on discord to prevent civs from declaring war on city-states that they just signed peace with. I also think it is weird how a civ can turn around and declare war immediately.

In order to do this we have to make a peace trade with the city-state in makePeaceOneSide(). After 10 turns the peace treaty runs out and the civ can attack the city state again.

Side note: should the ally be pulled into war when the city-state is attacked? Currently it doesn't which is a little weird.


core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
```diff
@@ -471,8 +472,16 @@ class DiplomacyManager() : IsPartOfGameInfoSerialization {
...
-                thirdCiv.getDiplomacyManager(otherCiv).makePeace()
+            if (thirdCiv.getAllyCiv() == civInfo.civName && thirdCiv.isAtWarWith(otherCiv)) {
+                val thirdCivDiplo = thirdCiv.getDiplomacyManager(otherCiv)
+                thirdCivDiplo.makePeace()
+                val tradeLogic = TradeLogic(thirdCiv, otherCiv)
+
+                tradeLogic.currentTrade.ourOffers.add(TradeOffer(Constants.peaceTreaty, TradeType.Treaty))
+                tradeLogic.currentTrade.theirOffers.add(TradeOffer(Constants.peaceTreaty, TradeType.Treaty))
+                thirdCivDiplo.trades.add(tradeLogic.currentTrade)
+                thirdCivDiplo.otherCivDiplomacy().trades.add(tradeLogic.currentTrade.reverse())
```
yairm210, line 483: Oh, interesting! So if I (A) make peace with my enemy (B) and I have an ally (C) who made war together with me, then C will send B a peace offer, and B can potentially say "no" to that! That'll make for an interesting case 🤔

tuvus: That is not possible, we are not offering the peace Trade here. We only put it in place. Neither civ has the chance to accept or reject.

yairm210: Gotcha

yairm210 (approved):

Issue #11768, opened 2024-06-16
NullPointerException on getDiplomacyManager()


mouillerart:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.12.0

### Describe the bug

Hi there, noticed a game crash with latest version
NullPointerException at com.unciv.logic.civilization.Civilization.getDiplomacyManager(Civilization.kt:324)
What's missing?

### Steps to Reproduce

1. Install mods
2. Load savefile
3. Click on "next turn"

### Link to save file

[The Empire.zip](https://github.com/user-attachments/files/15858661/The.Empire.zip)

### Operating System

Windows

### Additional Information

**Platform:** Desktop
**Version:** 4.12.0 (Build 1007)
**Rulesets:** [Civ V - Vanilla, Tyretes Tileset, Orbis Terrarum for Unciv, Community Maps, Civ V Leader portraits, Civ6 mod, Epic of Fantasy, Civ6 Tileset, Alpha Frontier, Unciv IV mod, UncivWarhammer, Civ V - Gods & Kings, 5Hex Tileset, Fantasia]
**Last Screen:** `com.unciv.ui.screens.worldscreen.WorldScreen`

tuvus:
Can you send the full stack trace?

tuvus:
A patch should be going through to fix this, please update when you can.

mouillerart:
The patch corrects the bug. Thank you!

PR #11769, opened 2024-06-16
Fixed a city-state coup crash

tuvus:
This PR fixes the crash in #11768 where a civ is initiating a coup in a city-state with an ally that they haven't met. This crash could actually happen on both a successful and a failed coup.

The solution I came up with is to make sure that the civilization that is attempting to coup has met the ally civilization before applying the diplomatic modifiers.

An alternative would be to keep the couping civilization anonymous and don't force the civilizations to meet. This seems a little unintuitive and would require an extra notification string as well.


yairm210:
It makes sense that if you send a spy to a city to take it over, then step 1 would be to actually know *who you're taking it over from*

yairm210 (approved):

tuvus:
release patch

Issue #10852, opened 2024-01-02
Crash with NullPointerException 


Toreg87:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.9.14

### Describe the bug

I get a NullPointerException on the next turn of my game. This is the first time I observed this:

**Platform:** Android
**Version:** 4.9.14 (Build 950)
**Rulesets:** [Civ V - Vanilla, Civ V - Gods & Kings]
**Last Screen:** `com.unciv.ui.screens.worldscreen.WorldScreen`

--------------------------------

Device Model: moto g14
API Level: 33
System Memory: 3815 MB
    Available (used by Kernel): 937 MB
    System Low Memory state: false
    Java heap limit: 512 MB
    Java heap free: 17 MB

--------------------------------

**Message:**
```
java.lang.NullPointerException
	at com.unciv.logic.civilization.Civilization.getDiplomacyManager(Civilization.kt:312)
	at com.unciv.logic.civilization.Civilization.getDiplomacyManager(Civilization.kt:311)
	at com.unciv.logic.automation.civilization.NextTurnAutomation.valueCityStateAlliance$core(NextTurnAutomation.kt:155)
	at com.unciv.logic.automation.civilization.NextTurnAutomation.valueCityStateAlliance$core$default(NextTurnAutomation.kt:121)
	at com.unciv.logic.battle.Battle.automateCityConquer(Battle.kt:556)
	at com.unciv.logic.battle.Battle.conquerCity(Battle.kt:541)
	at com.unciv.logic.battle.Battle.attack(Battle.kt:146)
	at com.unciv.logic.battle.Battle.attackOrNuke(Battle.kt:88)
	at com.unciv.logic.battle.Battle.moveAndAttack(Battle.kt:43)
	at com.unciv.logic.automation.unit.BattleHelper.tryAttackNearbyEnemy(BattleHelper.kt:29)
	at com.unciv.logic.automation.unit.BattleHelper.tryAttackNearbyEnemy$default(BattleHelper.kt:15)
	at com.unciv.logic.automation.unit.UnitAutomation.tryAttacking(UnitAutomation.kt:243)
	at com.unciv.logic.automation.unit.UnitAutomation.automateUnitMoves(UnitAutomation.kt:209)
	at com.unciv.logic.automation.civilization.NextTurnAutomation.automateUnits(NextTurnAutomation.kt:348)
[...]

PR #11771, opened 2024-06-16
TechDebt: Made getDiplomacyManger() nullable

tuvus:
#11768, #10852 are both good examples of how getDiplomacyManger() hides weather two civs have meet or not. 

This PR fixes that by making getDiplomacyManager() return null if the civs haven't met. 

While this does add extra !! in various places developers should take it upon themselves to decide if a situation warrants an extra .knows() check. Alternatively I added getDiplomacyManagerOrMeet() which has to give a diplomacy manager, this should be used in every case though.

I also added some extra oneCityChallange + AutoPlay support since there was a bug related to trying to found cities as a oneCityCallanger AI.


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -460,6 +460,7 @@ object NextTurnAutomation {
     private fun trainSettler(civInfo: Civilization) {
         val personality = civInfo.getPersonality()
         if (civInfo.isCityState()) return
+        if (civInfo.isOneCityChallenger()) return
```
tuvus, line 463: AI oneCityChallengers shouldn't build settlers in the first place.

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -185,7 +185,7 @@ object SpecificUnitAutomation {
 
         val foundCityAction = UnitActionsFromUniques.getFoundCityAction(unit, bestCityLocation)
         if (foundCityAction?.action == null) { // this means either currentMove == 0 or city within 3 tiles
-            if (unit.currentMovement > 0) // therefore, city within 3 tiles
+            if (unit.currentMovement > 0 && !unit.civ.isOneCityChallenger()) // therefore, city within 3 tiles
```
tuvus, line 188: A rare occurrence, but an AI can also be a oneCityChallanger and that may be a reason that the settler didn't found a city.

yairm210 (approved):

yairm210:
I see most of these are from getKnownCivs, maybe we should have getKnownCivDiplomacyManagers or something 🤔

Issue #11780, opened 2024-06-18
Destroyed civ can use espionage


RpixPro:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.12.0-patch1

### Describe the bug

When a civ is already destroyed, it can still use spies. This applies to players and AIs. I don't know if this is actually a feature

### Steps to Reproduce

1. Get all your cities captured
2. You can still use spies

Or:
1. Start a game
2. Disband all units and get the defeat screen
3. "one more turn" and you can still use spies

### Link to save file

https://gist.github.com/RpixPro/a9c281e94cfb8e2170f54b35f56aa325

### Operating System

Windows

PR #11785, opened 2024-06-19
Defeated civilizations don't use spies

tuvus:
This PR fixes the issue described in #11780.
When a civilization is defeated they need to move all spies to their hideout to prevent election rigging. Tech stealing can't occur since next turn won't be called.
Players also can't move their spies when they are defeated. This was done by disabling the buttons.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
@@ -77,7 +77,7 @@ class CityStateFunctions(val civInfo: Civilization) {
 
         fun getVotesFromSpy(spy: Spy?): Float {
             if (spy == null) return 20f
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
```diff
@@ -77,7 +77,7 @@ class CityStateFunctions(val civInfo: Civilization) {
 
         fun getVotesFromSpy(spy: Spy?): Float {
             if (spy == null) return 20f
-            var votes = (civInfo.getDiplomacyManager(spy.civInfo)!!.influence / 2)
+            var votes = (civInfo.getDiplomacyManagerOrMeet(spy.civInfo).influence / 2)
```
tuvus, line 80: There is a chance apparently, that a civ can have a spy set up in a city-state without seeing them. (From revealing the map with the technology). I think this is because the spy vision isn't updated at the end of the turn to meet the city-state immediately.

PR #11789, opened 2024-06-19
Fix coastal rivers near Rock of Gibraltar

### LINKED ISSUE #11786: River occuring in coast tiles

### Game Version

4.12.0-patch1

### Describe the bug

River occuring in coast tiles and also gives gold bonus. This is not the first time so i rolled map generator few times and finally found it again.

### Steps to Reproduce

Generate a map with these options:
image

### Screenshots

image

### Link to save file

This is a map file, not a save file:
https://gist.github.com/RpixPro/dc53f5422028cf5e9c2f20b885be5acb

### Operating System

Android

### Additional Information

_No response_

SomeTroglodyte:
This should fix #11786 - but untested. After all, map editor or console placing a nat wonder won't trigger these kind of side effects - maybe extend console?


yairm210 (approved):
The "constants.coast" should be replaced by "type == water" both this is relevant to both checks

PR #11802, opened 2024-06-20
Event Choice AI Priority

PLynx01:
I requested this feature, but I didn't see anyone doing it, so I decided to make it yourself. I believe my PR will greatly increase the usefulness of Events, which were added some time ago.

I introduced the new unique below not only for events, but possibly for Buildings, Units and Policies, as the Modding Philosophy states that "we want as little uniques as possible, that do as much as possible"

`[relativeAmount]% priority`

If you have some remarks regarding my PR, please share them in the comments.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -121,10 +121,11 @@ object UniqueTriggerActivation {
                 fun calculateEventChoiceWeight(choice: EventChoice): Float {
                     var weight = 1f
                     for (modifier in choice.modifierObjects) {
-                        if (modifier.type == UniqueType.AIPriorityModifier && modifier.conditionalsApply(stateForConditionals)) {
-                            weight += unique.params[0].toFloat() / 100f
+                        if (modifier.type == UniqueType.AiDecisonWeight && modifier.conditionalsApply(stateForConditionals)) {
+                            weight += unique.params[0].toFloat()
                         }
                     }
+                    weight.coerceAtLeast(0f)
                     return weight
                 }
 
@@ -141,7 +142,7 @@ object UniqueTriggerActivation {
                         }
                         random -= choiceWeights[choice]!!
                     }
-                    
+
                     chosenChoice?.triggerChoice(civInfo) ?: false
                 }
                 if (event.presentation == Event.Presentation.Alert) return {
--- a/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
+++ b/core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
@@ -21,16 +21,17 @@ import com.unciv.logic.map.mapgenerator.NaturalWonderGenerator
 import com.unciv.logic.map.mapgenerator.RiverGenerator
 import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.Tile
+import com.unciv.logic.map.tile.TileNormalizer
 import com.unciv.models.UpgradeUnitAction
 import com.unciv.models.ruleset.BeliefType
 import com.unciv.models.ruleset.Event
+import com.unciv.models.ruleset.EventChoice
 import com.unciv.models.ruleset.tile.TerrainType
 import com.unciv.models.stats.Stat
 import com.unciv.models.stats.Stats
 import com.unciv.models.translations.fillPlaceholders
 import com.unciv.models.translations.hasPlaceholderParameters
 import com.unciv.ui.components.extensions.addToMapOfSets
-import com.unciv.logic.map.tile.TileNormalizer
 import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsUpgrade
 import kotlin.math.roundToInt
 import kotlin.random.Random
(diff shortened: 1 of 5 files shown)


PLynx01:
Maybe should we ask @tuvus for a review?

tuvus:
Thanks for taking it into your own hands.

Overall it looks good. The only comment I have is that I don't think we should use a % based approach. So `[relativeAmount]% priority` and instead should call it `[relativeAmount] priority` and we don't divide the number by 100. The reasoning for this is that we are giving the false impression that it is percent based when it really isn't (it's weight based).

For example, if I have two options with the base percent of 50%, that adds up to 100%. However, if I add a conditional to one of them that increases it's weight to 60%, we are now left with 110% total.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -879,6 +879,8 @@ enum class UniqueType(
     ForEveryAmountCountable("for every [amount] [countable]", UniqueTarget.MetaModifier),
     ModifiedByGameSpeed("(modified by game speed)", UniqueTarget.MetaModifier,
         docDescription = "Can only be applied to certain uniques, see details of each unique for specifics"),
+    AIPriorityModifier("[relativeAmount]% priority", UniqueTarget.MetaModifier,
```
yairm210, line 882: A. "[Amount] weight for AI decisions"
B. Target type Event, not applicable to other objects yet
C. Should be invisible to users
D. doc description is lacking - as a modder, I wouldn't know what numbers to put here or what they mean

PLynx01:
> Thanks for taking it into your own hands.
> 
> Overall it looks good. The only comment I have is that I don't think we should use a % based approach. So `[relativeAmount]% priority` and instead should call it `[relativeAmount] priority` and we don't divide the number by 100. The reasoning for this is that we are giving the false impression that it is percent based when it really isn't (it's weight based).
> 
> For example, if I have two options with the base percent of 50%, that adds up to 100%. However, if I add a conditional to one of them that increases it's weight to 60%, we are now left with 110% total.

In that case, could you give me the syntax of this unique that would double the priority of event choice when civilization is at war?

Current syntax is:
`[+100]% priority <when at war>`

Rewrite this according to your preferences.

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
PLynx01: Your suggested syntax implies that the AI priorities should be done in overwriting, not stackable manner. Why is that?

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -116,8 +117,32 @@ object UniqueTriggerActivation {
...
                     ?: return null
+
+                fun calculateEventChoiceWeight(choice: EventChoice): Float {
+                    var weight = 1f
+                    for (modifier in choice.modifierObjects) {
+                        if (modifier.type == UniqueType.AIPriorityModifier && modifier.conditionalsApply(stateForConditionals)) {
+                            weight += unique.params[0].toFloat() / 100f
+                        }
+                    }
+                    return weight
```
tuvus, line 128: I think we need a coerceAtLeast(0f) here, otherwise it could go negative and mess up the sum below.

PLynx01:
So, in that case, we have to make a decision beetween two approaches:

1. Overwrite approach
The AI weight is the value of the last unique with matching conditional.

2. Stacking approach
The AI weights are calculated by adding/subtracting relativeAmount percent from base weight.

@yairm210 @SomeTroglodyte @tuvus 
Please share and explain your opinions.

tuvus:
> 2. Stacking approach
The AI weights are calculated by adding/subtracting relativeAmount percent from base weight.

I think this option is best, but without using any percent logic.
 
Percent only works if adding a percent chance to one option takes away an equal percent chance from the other options.
For example, if I have three buttons, each having a 33% chance of being picked, and add a 33% chance on Button A. Then button A has a 66% chance of being picked and buttons B and C have a 16% chance of being picked. This is a bit different then summing up the weights of each option.

Either way works, however if we choose what you currently have we shouldn't call it percent.

In both options we are making them additive in some way.

PLynx01:
> > 2. Stacking approach
> The AI weights are calculated by adding/subtracting relativeAmount percent from base weight.
> 
> I think this option is best, but without using any percent logic.
>  
> Percent only works if adding a percent chance to one option takes away an equal percent chance from the other options.
> For example, if I have three buttons, each having a 33% chance of being picked, and add a 33% chance on Button A. Then button A has a 66% chance of being picked and buttons B and C have a 16% chance of being picked. This is a bit different then summing up the weights of each option.
> 
> Either way works, however if we choose what you currently have we shouldn't call it percent.
> 
> In both options we are making them additive in some way.

Please demonstrate me your idea in detail. Give me some examples, so I can understand it.

tuvus:
Of which method? Percent based or weight based?

PLynx01:
> Of which method? Percent based or weight based?

I am talking about weight-based method, as you proposed.

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -116,8 +117,32 @@ object UniqueTriggerActivation {
                 val event = ruleset.events[unique.params[0]] ?: return null
                 val choices = event.getMatchingChoices(stateForConditionals)
                     ?: return null
+
+                fun calculateEventChoiceWeight(choice: EventChoice): Float {
+                    var weight = 1f
+                    for (modifier in choice.modifierObjects) {
+                        if (modifier.type == UniqueType.AIPriorityModifier && modifier.conditionalsApply(stateForConditionals)) {
+                            weight += unique.params[0].toFloat() / 100f
```
tuvus, line 125: We'll probably want to remove the `/100f` here since we aren't working with percents any more.

tuvus:
Apart from that, follow the suggestions given and it work should follow a good design.
I am not very knowledgeable on the modding side of things, so if Yair thinks it works, that's good by me. I just didn't want to fall into the percent trap like we kind of did in espionage.

PLynx01:
@tuvus Since we are working with non-percent weights, what is the base weight we add or subtract from?

Is it 1, 10, 100 or the other number?

tuvus:
1, you already have that.

PLynx01:
> 1, you already have that.

So, if the base weight is 1, so we only add to weight, not subtract from it?

tuvus:
We can subtract from it and go negative while summing it up. However afterwards we use .coerceAtleast(0) to make sure it isn't negative. I put a comment on this.

PLynx01:
@tuvus @yairm210 @SomeTroglodyte 
Check out my commit, Is there anything necessary to be done with this PR before merging?

yairm210:
This was a good direction, but due to problems inherent in the event choice I merged all parts of the event choice into one "uniques" field, and added a new unique ("[relativeAmount]% weight to this choice for AI decisions") which can apply to the choices, as well as techs policies and promotions, and is most importantly - *conditionalizable*

PR #11755, opened 2024-06-15
Bangla Translations (15%)

touhidurrr:

PR #11806, opened 2024-06-22
Support for languages using Diacritics as multi-codepoint joiners

SomeTroglodyte:
Pretty sure this renders [Bangla](#11755) as intended - while waiting on @touhidurrr 's test or feedback you might want to see the pure support code - as kinda preview... For screenies see the monstrous discussion over there.

While this is a step that could help getting Hebrew to render, it's not the crucial one... Sorry no ideas for that except rewriting Scene2D (or at least GlyphLayout and everything dependent on it).

Also - emojis beyond U+FFFF and/or zwidge-using ones: I think close but still impossible with this, as that hits a kotlin Char limitation - Char is crippled to 16 bit same as dotnet & many other libraries, so a kotlin String would need to encode these as multichar UCS-16 before joining can be applied, and then this would need to redefine many uses of Char... Not worth it I say.


core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
```diff
@@ -0,0 +1,196 @@
...
+    //endregion
+
+    //region These fields will only persist during one language load
+    private var noDiacritics = true
+    private val charClassMap = mutableMapOf<Char, CharClass>()
+    private var defaultCharClass = CharClass.None
+
+    private class LineData(capacity: Int) {
+        val output = StringBuilder(capacity)
+        val accumulator = StringBuilder(8)
```
touhidurrr, line 57: `StringBuilder(8)`, does this means the limit is 8? will not work. I guess you based this on https://en.wikipedia.org/wiki/Diacritic#Orthographic. But if you notice, below there is a list of 9 characters and 8 diacritics. So, the sequence we want is already 9 characters long.

So, this will not work as the number of characters in combinations including normal characters can be much higher.

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
```diff
@@ -0,0 +1,196 @@
...
+        fun expectsJoin() = accumulator.isNotEmpty() && getCharClass(accumulator.last()).expectsRightJoin
+        fun flush() {
+            if (accumulator.length <= 1) output.append(accumulator)
```
touhidurrr, line 60: what is this? cant understand properly. why are we checking length 0 and 1 specifically.

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
```diff
@@ -0,0 +1,196 @@
...
+
+    //region These fields will only persist during one language load
+    private var noDiacritics = true
+    private val charClassMap = mutableMapOf<Char, CharClass>()
+    private var defaultCharClass = CharClass.None
+
+    private class LineData(capacity: Int) {
+        val output = StringBuilder(capacity)
+        val accumulator = StringBuilder(8)
+        fun expectsJoin() = accumulator.isNotEmpty() && getCharClass(accumulator.last()).expectsRightJoin
```
touhidurrr, line 58: `.expectsRightJoin` huh? there might be `.expectsLeftAndRightJoin` also or did you just write it as `.expectsRightJoin`

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
touhidurrr: Oh, ok I understand. You are ignoring such combinations. However, if 0 just continue? appending 0 length strings in the output will not cause issues?

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
```diff
@@ -54,7 +54,7 @@ object DiacriticSupport {
 
     private class LineData(capacity: Int) {
         val output = StringBuilder(capacity)
-        val accumulator = StringBuilder(8)
+        val accumulator = StringBuilder(9) // touhidurrr said there can be nine
```
touhidurrr, line 57: kind of hilarious that you would phrase it like this. i pointed out that it can be more, not necessarily 9. the wiki link maximum is based on the number of diacritics, not based on the number of total characters with diacritics.
anyways, if it can grow then 8 or 9 can be a good enough initial space.

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
```diff
@@ -0,0 +1,196 @@
...
+    private val inverseMap = mutableMapOf<String, Char>()
+    //endregion
+
+    //region These fields will only persist during one language load
+    private var noDiacritics = true
+    private val charClassMap = mutableMapOf<Char, CharClass>()
+    private var defaultCharClass = CharClass.None
+
+    private class LineData(capacity: Int) {
+        val output = StringBuilder(capacity)
```
touhidurrr, line 56: the output is not a string but a string builder? I used `toString()` before putting it into a set because I was not sure if anything else would work.

touhidurrr (approved):

touhidurrr:
So after the discussion in https://github.com/yairm210/Unciv/pull/11755#issuecomment-2184052991, this looks ready to merge. I will provide updates if I find any issue with the architecture. But as of now, this looks complete and tested.
Thanks for everything @SomeTroglodyte !!

touhidurrr:
By the way, @SomeTroglodyte, how can I checkout this branch in my local `bangla-translations` branch? Tried some commands but it did not work as intended. Need some git support here.

touhidurrr (approved):

touhidurrr:
Ok got it. Since I was doing translations, I was using VS Code for its lightness. It seems to have similar remote management which was disabled before and works similarly. There was an option to merge changes without commits which should work wonders.

By the way, how write the value of `diacritics_joinable_range` again?

touhidurrr:
Also why not just use start and end range code of a language? looks simpler.  The description is not clear. Should I type all characters here or should I just type characters without diacritics?

touhidurrr:
Ok figured things out somehow. It was really a font issue. Not all fonts includes glyphs for all consonant conjuncts it seems. Here is a working screenshot:
image

Also, Bangla support setup:
```ts
diacritics_joinable_range = "ঀ ৽"
left_joining_diacritics = "ঁংঃ়ািীুূৃেৈোৌৗ"
right_joining_diacritics = ""
left_and_right_joiners = "্"
```
Removed some diacritics that are on the standard but not used in Bangla language. These are used in Assamese language but not in Bangla. (See [Assamese](https://en.wikipedia.org/wiki/Assamese_alphabet) and [Bengali–Assamese script](https://en.wikipedia.org/wiki/Bengali%E2%80%93Assamese_script)).
Just updating here in case I am doing something wrong. Is how I typed it ok?

touhidurrr:
Also, since there is font issue. Is it possible to ship some default fonts for some languages which will be selected by default when you select a language? (If their license is libre)

touhidurrr:
Also, here is a error that I am not being able to reproduce. [lasterror.txt](https://github.com/user-attachments/files/15938059/lasterror.txt)

```js
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.desktop/sun.font.FileFontStrike.getGlyphAdvance(FileFontStrike.java:615)
	at java.desktop/sun.font.FileFontStrike.getGlyphAdvance(FileFontStrike.java:572)
	at java.desktop/sun.font.FileFontStrike.getCodePointAdvance(FileFontStrike.java:691)
	at java.desktop/sun.font.FontDesignMetrics.handleCharWidth(FontDesignMetrics.java:414)
	at java.desktop/sun.font.FontDesignMetrics.getLatinCharWidth(FontDesignMetrics.java:423)
	at java.desktop/sun.font.FontDesignMetrics.stringWidth(FontDesignMetrics.java:474)
	at com.unciv.app.desktop.DesktopFont.getCharPixmap(DesktopFont.kt:66)
	at com.unciv.ui.components.fonts.NativeBitmapFontData.getPixmapFromChar(NativeBitmapFontData.kt:154)
	at com.unciv.ui.components.fonts.NativeBitmapFontData.createAndCacheGlyph(NativeBitmapFontData.kt:74)
	at com.unciv.ui.components.fonts.NativeBitmapFontData.getGlyph(NativeBitmapFontData.kt:71)
	at com.unciv.ui.components.fonts.NativeBitmapFontData.<init>(NativeBitmapFontData.kt:65)
	at com.unciv.ui.components.fonts.FontImplementation$DefaultImpls.getBitmapFont(FontImplementation.kt:24)
	at com.unciv.app.desktop.DesktopFont.getBitmapFont(DesktopFont.kt:18)
	at com.unciv.ui.components.fonts.Fonts.resetFont(Fonts.kt:51)
	at com.unciv.ui.screens.basescreen.BaseScreen$Companion.setSkin(BaseScreen.kt:139)
	at com.unciv.ui.popups.options.OptionsPopup$reloadWorldAndOptions$1$1.invokeSuspend(OptionsPopup.kt:155)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
	at com.unciv.utils.Dispatchers$CrashHandlingDispatcher$dispatch$1.invoke(Concurrency.kt:190)
	at com.unciv.utils.Dispatchers$CrashHandlingDispatcher$dispatch$1.invoke(Concurrency.kt:190)
	at com.unciv.ui.crashhandling.CrashHandlingExtensionsKt$wrapCrashHandling$1.invoke(CrashHandlingExtensions.kt:17)
	at com.unciv.ui.crashhandling.CrashHandlingExtensionsKt$wrapCrashHandlingUnit$1.invoke(CrashHandlingExtensions.kt:33)
	at com.unciv.ui.crashhandling.CrashHandlingExtensionsKt$wrapCrashHandlingUnit$1.invoke(CrashHandlingExtensions.kt:33)
	at com.unciv.utils.Dispatchers$CrashHandlingDispatcher.dispatch$lambda$0(Concurrency.kt:190)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:209)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:167)
	at com.unciv.app.desktop.HardenGdxAudio.<init>(HardenGdxAudio.kt:46)
	at com.unciv.app.desktop.DesktopLauncher.main(DesktopLauncher.kt:76)
```

touhidurrr:
Also, not sure why but this is not translated. (but it is in my translations file)

touhidurrr (approved):

touhidurrr:
Anyways, then I guess the only thing left to do is to merge this and then resume work with my Bangla Translations branch. Thanks again @SomeTroglodyte for this pr. Also, @yairm210, I think you can merge this now.

android/assets/jsons/translations/template.properties
```diff
@@ -9,6 +9,18 @@
 # Don't translate these words to your language, only put 'true' or 'false'.
 StartWithCapitalLetter = 
 
+# Diacritics Support:
+# Underlying libraries operate one character at a time, and with these you define where Unciv will create replacement codepoints with joined glyphs.
+# Each one, including the range, can simply list codes one after the other or you can separate them with blanks when needed or for readability.
+# If your language does not need these, feel free to ignore, or use "" to avoid the "requires translation" mark.
+# Define the range of characters that can join: Supply exactly to characters. This is allowed to include the diacritics below, the actual set of "base" characters will exclude those automatically.
+diacritics_joinable_range =
```
yairm210, line 17: As a translator I would have *no idea* what you mean here. 
I think we need to give specific instructions - "look up the diacritics' UTF8 and find first and last" is also just as confusing if you have no idea what it means, but at least it gives an idea to those who *do* know, what the format should be like

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -118,22 +118,27 @@ class Translations : LinkedHashMap<String, TranslationEntry>() {
...
                 this[hashKey] = entry
             }
-            entry[language] = translation.value
+            entry[language] = if (noDiacritics) value else DiacriticSupport.remapDiacritics(value)
         }
     }
 
+
     fun tryReadTranslationForCurrentLanguage() {
+        DiacriticSupport.reset()
```
yairm210, line 139: Sounds to me like a "with" block, both here and below

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
yairm210: Then please change this, tou is right, this is confusing

core/src/com/unciv/ui/components/fonts/DiacriticSupport.kt
```diff
@@ -0,0 +1,196 @@
...
+    /** Set at [prepareTranslationData], if true the translation loader need not bother passing stuff through [remapDiacritics]. */
+    fun noDiacritics() = noDiacritics
+
+    private fun getCharClass(char: Char) = charClassMap[char] ?: defaultCharClass
+
+    private fun getReplacementChar(joined: String) = inverseMap[joined] ?: createReplacementChar(joined)
+
+    private fun createReplacementChar(joined: String): Char {
+        val char = getNextFreeCode()
+        nextFreeDiacriticReplacementCodepoint--
```
yairm210, line 145: Would expect the -- to be in "getnextfreecode" tbh

yairm210:
Almost all data retained is *per language*. My assessment is that this shouldn't be an object at all, but a class, that you create an instance of per language.
The only persistent data is the next available char, and that can be in a companion object

yairm210:
And the map for getStringFor, of course

android/assets/jsons/translations/template.properties
yairm210: Thinking of this again - surely the OS has some way of recognizing diacritics 
Java has a text normalizer so worst case we can use that
We can build the "all language characters" by making a set of the entire translation file
And you're Robert's nephew

touhidurrr: > As a translator I would have no idea what you mean here.

I pointed out this too. In my opinion a better description can be:
```properties
# the start and end code points of your language in Unicode. For example, for Bangla, this is U+0980 to U+09FF.
# https://en.wikipedia.org/wiki/Bengali_alphabet#Unicode
# Search for the information about the Unicode support in your language to fill the following 2 lines.
language_start_code = 
language_end_code = 
```

touhidurrr: > Thinking of this again - surely the OS has some way of recognizing diacritics

Maybe [`Normalizer.isNormalized()`](https://docs.oracle.com/javase/6/docs/api/java/text/Normalizer.html#isNormalized) can work? (if it works on Char)

touhidurrr: Ok, based on my analysis here: [playground link](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4wLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJwYWNrYWdlIHRvdWhpZHVycnJcblxuY29uc3QgdmFsIEJBTkdMQV9DSEFSU0VUX1NUQVJUID0gMHgwOTgwXG5jb25zdCB2YWwgQkFOR0xBX0NIQVJTRVRfRU5EID0gMHgwOWZmXG5cbmZ1biBtYWluKCkge1xuICAgIGZvciAoY29kZSBpbiBCQU5HTEFfQ0hBUlNFVF9TVEFSVC4uQkFOR0xBX0NIQVJTRVRfRU5EKSB7XG4gICAgICAgIHZhbCBjaCA9IGNvZGUudG9DaGFyKClcbiAgICAgICAgcHJpbnRsbihcIiRjb2RlIC0+ICRjaFwiKVxuICAgICAgICBwcmludGxuKFwiRGlyZWN0aW9uYWxpdHk6ICR7Y2guZGlyZWN0aW9uYWxpdHl9XCIpXG4gICAgICAgIHByaW50bG4oXCJpc0lkZW50aWZpZXJJZ25vcmFibGU6ICR7Y2guaXNJZGVudGlmaWVySWdub3JhYmxlKCl9XCIpXG4gICAgICAgIHByaW50bG4oXCJpc0phdmFJZGVudGlmaWVyUGFydCAke2NoLmlzSmF2YUlkZW50aWZpZXJQYXJ0KCl9XCIpXG4gICAgICAgIHByaW50bG4oXCJpc0phdmFJZGVudGlmaWVyU3RhcnQ6ICR7Y2guaXNKYXZhSWRlbnRpZmllclN0YXJ0KCl9XCIpXG4gICAgICAgIHByaW50bG4oXCJpc1N1cnJvZ2F0ZTogJHtjaC5pc1N1cnJvZ2F0ZSgpfVwiKVxuICAgIH1cbn0ifQ==), we can probably detect diacritics and joiners by `Char.isJavaIdentifierStart()` method. Not sure how to implement it in code though.
So, the current method looks better.

touhidurrr: > We can build the "all language characters" by making a set of the entire translation file

I believe we are already doing this.

> Java has a text normalizer so worst case we can use that

This will not be helpful. Normalizing would just drop data that we will be needing.

touhidurrr: In short, I believe the design change you mentioned here is not viable for rendering a language that uses diacritics properly at this point of time. I am seeing no option other than to keep a list of set of left, right and both way code joiners and diacritics.

yairm210: Yeah, knowing if something *is a diacritic* isn't enough.
But something still isn't sitting right for me - how does *the OS know how to render*? This info must be baked in the to font somehow, and we should be able to read it from there

touhidurrr: > This info must be baked in the to font somehow, and we should be able to read it from there

Hypothetically, If we can read this from a font, then we have to ship the font also. As the default font in @SomeTroglodyte's system could not render many valid glyphs in Bangla.

touhidurrr: Now, since right and left joiner thing it special and less in number, we can just ask translators to input it separately.

The logic would be something like this:
```kt
if (joinersSet.has(ch)) // joiner
else if (!Char.isJavaIdentifierStart() || listOf("NON_SPACING_MARK","COMBINING_SPACING_MARK").includes(CharCategory.valueOf(Character.getType(ch))))
  // diacritics
else // normal character
``

Comment (touhidurrr) [reply android/assets/jsons/translations/template.properties:17]: For proof of concept, this seems to work: [playground link](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4wLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJwYWNrYWdlIHRvdWhpZHVycnJcblxuY29uc3QgdmFsIEJBTkdMQV9DSEFSU0VUX1NUQVJUID0gMHgwOTgwXG5jb25zdCB2YWwgQkFOR0xBX0NIQVJTRVRfRU5EID0gMHgwOWZmXG5cbnZhbCBCQU5HTEFfSk9JTkVSUyA9IHNldE9mKCfgp40nKVxuXG5mdW4gaXNCYW5nbGFDaGFyKGNoOiBDaGFyKTogQm9vbGVhbiB7XG4gICAgcmV0dXJuIGNoLmNvZGUgPj0gQkFOR0xBX0NIQVJTRVRfU1RBUlQgJiYgY2guY29kZSA8PSBCQU5HTEFfQ0hBUlNFVF9FTkRcbn1cblxudmFsIGFsbFNlcXVlbmNlcyA9IG11dGFibGVTZXRPZjxTdHJpbmc+KClcblxuZnVuIG1haW4oKSB7XG4gICAgdmFsIGxpbmVzID0gbGlzdE9mKFxuICAgICAgICBcIuCmruCmvuCmqOCmmuCmv+CmpOCnjeCmsCDgprjgpq7gp43gpqrgpr7gpqbgppVcIiwgXCLgpqbgp4fgppbgp4HgpqhcIiwgXCLgpongp47gpqrgpqjgp43gpqgg4KaV4Kaw4KeB4KaoXCIsIFwi4KaG4KaC4Ka24Ka/4KaVXCJcbiAgICApXG5cbiAgICBsaW5lcy5mb3JFYWNoIHsgbGluZSAtPlxuICAgICAgICBwcmludGxuKFwibGluZTogJGxpbmVcIilcbiAgICAgICAgdmFsIGxhc3RTZXF1ZW5jZSA9IFN0cmluZ0J1aWxkZXIoKVxuICAgICAgICBsaW5lLmZvckVhY2ggeyBjaCAtPlxuICAgICAgICAgICAgaWYgKCFpc0JhbmdsYUNoYXIoY2gpKSB7XG4gICAgICAgICAgICAgICAgaWYgKGxhc3RTZXF1ZW5jZS5pc05vdEVtcHR5KCkpIHtcbiAgICAgICAgICAgICAgICAgICAgYWxsU2VxdWVuY2VzLmFkZChsYXN0U2VxdWVuY2UudG9TdHJpbmcoKSlcbiAgICAgICAgICAgICAgICAgICAgbGFzdFNlcXVlbmNlLmNsZWFyKClcbiAgICAgICAgICAgICAgICB9XG4gICAgICAgICAgICAgICAgcmV0dXJuQGZvckVhY2hcbiAgICAgICAgICAgIH1cblxuICAgICAgICAgICAgdmFsIGNhdGVnb3J5ID0gQ2hhckNhdGVnb3J5LnZhbHVlT2YoQ2hhcmFjdGVyLmdldFR5cGUoY2gpKVxuICAgICAgICAgICAgaWYgKFxuICAgICAgICAgICAgICAgICFjaC5pc0phdmFJZGVudGlmaWVyU3RhcnQoKSB8fFxuICAgICAgICAgICAgICAgIGNhdGVnb3J5ID09IENoYXJDYXRlZ29yeS5DT01CSU5JTkdfU1BBQ0lOR19NQVJLIHx8XG4gICAgICAgICAgICAgICAgY2F0ZWdvcnkgPT0gQ2hhckNhdGVnb3J5Lk5PTl9TUEFDSU5HX01BUktcbiAgICAgICAgICAgICkge1xuICAgICAgICAgICAgICAgIGxhc3RTZXF1ZW5jZS5hcHBlbmQoY2gpXG4gICAgICAgICAgICAgICAgcmV0dXJuQGZvckVhY2hcbiAgICAgICAgICAgIH1cblxuICAgICAgICAgICAgXG4gICAgICAgICAgICBpZiAobGFzdFNlcXVlbmNlLmlzTm90RW1wdHkoKSAmJiAhQkFOR0xBX0pPSU5FUlMuY29udGFpbnMobGFzdFNlcXVlbmNlLmxhc3QoKSkpIHtcbiAgICAgICAgICAgICAgICBhbGxTZXF1ZW5jZXMuYWRkKGxhc3RTZXF1ZW5jZS50b1N0cmluZygpKVxuICAgICAgICAgICAgICAgIGxhc3RTZXF1ZW5jZS5jbGVhcigpXG4gICAgICAgICAgICB9XG4gICAgICAgICAgICBsYXN0U2VxdWVuY2UuYXBwZW5kKGNoKVxuICAgICAgICB9XG5cbiAgICAgICAgaWYgKGxhc3RTZXF1ZW5jZS5pc05vdEVtcHR5KCkpIHtcbiAgICAgICAgICAgIGFsbFNlcXVlbmNlcy5hZGQobGFzdFNlcXVlbmNlLnRvU3RyaW5nKCkpXG4gICAgICAgIH1cbiAgICB9XG5cbiAgICBhbGxTZXF1ZW5jZXMuZm9yRWFjaCB7IHByaW50bG4oaXQudG9MaXN0KCkpIH1cbn0ifQ==)

Comment (touhidurrr) [reply android/assets/jsons/translations/template.properties:17]: And even this works: [playground link](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4wLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJwYWNrYWdlIHRvdWhpZHVycnJcblxuY29uc3QgdmFsIEJBTkdMQV9DSEFSU0VUX1NUQVJUID0gMHgwOTgwXG5jb25zdCB2YWwgQkFOR0xBX0NIQVJTRVRfRU5EID0gMHgwOWZmXG5cbnZhbCBCQU5HTEFfSk9JTkVSUyA9IHNldE9mKCfgp40nKVxuXG5mdW4gaXNCYW5nbGFDaGFyKGNoOiBDaGFyKTogQm9vbGVhbiB7XG4gICAgcmV0dXJuIGNoLmNvZGUgPj0gQkFOR0xBX0NIQVJTRVRfU1RBUlQgJiYgY2guY29kZSA8PSBCQU5HTEFfQ0hBUlNFVF9FTkRcbn1cblxudmFsIGFsbFNlcXVlbmNlcyA9IG11dGFibGVTZXRPZjxTdHJpbmc+KClcblxuZnVuIG1haW4oKSB7XG4gICAgdmFsIGxpbmVzID0gbGlzdE9mKFxuICAgICAgICBcIuCmruCmvuCmqOCmmuCmv+CmpOCnjeCmsCDgprjgpq7gp43gpqrgpr7gpqbgppVcIiwgXCLgpqbgp4fgppbgp4HgpqhcIiwgXCLgpongp47gpqrgpqjgp43gpqgg4KaV4Kaw4KeB4KaoXCIsIFwi4KaG4KaC4Ka24Ka/4KaVXCIsIFwi4KaW4KeN4Kaw4Ka/4Ka34KeN4Kaf4Kaq4KeC4Kaw4KeN4KasXCIsIFwi4Ka44KaC4KaV4KeN4Ka34Ka/4Kaq4KeN4KakXCIsIFwi4Ka24KaV4KeN4Kak4Ka/XCIsIFwi4Kas4KeI4Ka24Ka/4Ka34KeN4Kag4KeN4Kav4KeHXCJcbiAgICApXG5cbiAgICBsaW5lcy5mb3JFYWNoIHsgbGluZSAtPlxuICAgICAgICBwcmludGxuKFwibGluZTogJGxpbmVcIilcbiAgICAgICAgdmFsIGxhc3RTZXF1ZW5jZSA9IFN0cmluZ0J1aWxkZXIoKVxuICAgICAgICBsaW5lLmZvckVhY2ggeyBjaCAtPlxuICAgICAgICAgICAgaWYgKCFpc0JhbmdsYUNoYXIoY2gpKSB7XG4gICAgICAgICAgICAgICAgaWYgKGxhc3RTZXF1ZW5jZS5pc05vdEVtcHR5KCkpIHtcbiAgICAgICAgICAgICAgICAgICAgYWxsU2VxdWVuY2VzLmFkZChsYXN0U2VxdWVuY2UudG9TdHJpbmcoKSlcbiAgICAgICAgICAgICAgICAgICAgbGFzdFNlcXVlbmNlLmNsZWFyKClcbiAgICAgICAgICAgICAgICB9XG4gICAgICAgICAgICAgICAgcmV0dXJuQGZvckVhY2hcbiAgICAgICAgICAgIH1cblxuICAgICAgICAgICAgaWYgKCFjaC5pc0phdmFJZGVudGlmaWVyU3RhcnQoKSkge1xuICAgICAgICAgICAgICAgIGxhc3RTZXF1ZW5jZS5hcHBlbmQoY2gpXG4gICAgICAgICAgICAgICAgcmV0dXJuQGZvckVhY2hcbiAgICAgICAgICAgIH1cblxuICAgICAgICAgICAgXG4gICAgICAgICAgICBpZiAobGFzdFNlcXVlbmNlLmlzTm90RW1wdHkoKSAmJiAhQkFOR0xBX0pPSU5FUlMuY29udGFpbnMobGFzdFNlcXVlbmNlLmxhc3QoKSkpIHtcbiAgICAgICAgICAgICAgICBhbGxTZXF1ZW5jZXMuYWRkKGxhc3RTZXF1ZW5jZS50b1N0cmluZygpKVxuICAgICAgICAgICAgICAgIGxhc3RTZXF1ZW5jZS5jbGVhcigpXG4gICAgICAgICAgICB9XG4gICAgICAgICAgICBsYXN0U2VxdWVuY2UuYXBwZW5kKGNoKVxuICAgICAgICB9XG5cbiAgICAgICAgaWYgKGxhc3RTZXF1ZW5jZS5pc05vdEVtcHR5KCkpIHtcbiAgICAgICAgICAgIGFsbFNlcXVlbmNlcy5hZGQobGFzdFNlcXVlbmNlLnRvU3RyaW5nKCkpXG4gICAgICAgIH1cbiAgICB9XG5cbiAgICBhbGxTZXF1ZW5jZXMuZm9yRWFjaCB7IHByaW50bG4oaXQudG9MaXN0KCkpIH1cbn0ifQ==)
```kt
if (joinersSet.has(ch)) // joiner
else if (!Char.isJavaIdentifierStart()) // diacritics
else // normal character
```

yairm210:
I mean, it just seems *safer*
Here there's a non-trivial chance that languages will interfere with each other, especially if we come to change this class in the future, which seems likely
Think of it as pre-squashing bugs

touhidurrr:
Also, I tried to make the description a little more elegant [here](https://github.com/yairm210/Unciv/pull/11806#discussion_r1652569920). Can you check it and see if it reads better?

touhidurrr:
>  Repeat: Windows 10 + lang Sylheti, then Font Nirmala selected in Unciv

Wow, that looks cool actually. I also like this font but the problem is this might not be open source as it was made by Microsoft.  My old habit of trying to use everything open source.

> "preferential treatment"

That's funny. I did not actually think that much before typing that. You can remove that line if you want. I just wanted to provide an example for the readers and thought it would make sense linking a Unicode block table and showing what the start and end code points of a block actually means. Anyways, 3rd part perspective is also important.

The important part is the names `unicode_block_start` and `unicode_block_end`, thought it made more sense.

touhidurrr:
Anyways, my current settings:
```properties
# Diacritics settings
# preserving standards
language_start_code = U+0980
language_end_code = U+09FF
# removing Assamese diacritics
left_joining_diacritics = U+0981 U+0982 U+0983 U+09BC U+09BE U+09BF U+09C0 U+09C1 U+09C2 U+09C3 U+09C7 U+09C8 U+09CB U+09CC U+09D7
right_joining_diacritics = ""
left_and_right_joiners = U+09CD
```

touhidurrr:
Ok, seems to be working after a352ddc3b4af19c90234954966c797195835a7b8.

image

touhidurrr:
> "install Sylheti, select Nirmala"

Really? **Nirmala UI** font came pre-installed in my system. Also, I don't think **Sylheti** is a script that is still in usage? Far less it being the ancestor of **Bangla**. Really weird grouping.

Also in linux, I am pretty sure there is a bangla fonts package that you can install. Ok, some search reveals there is a `fonts-beng` package. There is also less known `font-noto-bengali`.

touhidurrr:
> Nononono... On fivehundred-and-second thought, the `CharCategory` _**is**_ the answer, it just fails for Bangla's U+09CD.

You mean this? https://github.com/yairm210/Unciv/pull/11806#discussion_r1652558194

touhidurrr:
<details><summary>Me, Inside...</summary>
<p>
Also, plz plz plz  merge this quickly so that I can start work on Bangla for real. plz...........................
</p>
</details>

touhidurrr:
image

Ok, if you search Bangla you get Bangla. But if you search Bengali, you get Bangla (India). Good for natives, confusing for foreigners maybe. As they do not know that Bangla is called Bangla natively.

yairm210:
I think "unicode_block_start_character" and "unicode_block_end_character" - since this is a complicated subject, let's be as specific as possible

touhidurrr:
> I think "unicode_block_start_character" and "unicode_block_end_character" - since this is a complicated subject, let's be as specific as possible

I think we can ditch this altogether. I have an alternative suggestion. Something like this: (updating soon)

touhidurrr:
> > ditch
> 
> Yes the current _local_ implementation has made all of them optional for the time being - YAGNI may remove 4 of six - or leave them in.

Ok, this might be the Java way of doing things. https://github.com/yairm210/Unciv/pull/11806#issuecomment-2189923014 But probably not more performant.

touhidurrr:
> and discuss elsewhere

I guess so. Normally it would be on Discord, but you are probably not available there. Thus, issues feel more convenient. Although to onlookers, such discussion might look like potential issues. Maybe, I need to get used to GitHub discussions but not many repo use it actively and so I often forget about this.

touhidurrr (approved):

yairm210:
I'm confused.
@touhidurrr You consider this mergable?
And after it's merged as a general all-language solution, you'll add a second PR for Bangla stuff?

touhidurrr:
> @touhidurrr You consider this mergable?

Yes.

> And after it's merged as a general all-language solution, you'll add a second PR for Bangla stuff?

No. I just need to add some line in my current pr for it to work. Which I already did.

touhidurrr:
Also, this is our workaround for Unciv to support Bangla. Don't expect everything to be shiny @yairm210. The state of code looks fine to me.

yairm210 (approved):

PR #11810, opened 2024-06-22
Escort movement fix

### LINKED ISSUE #11562: Escorted movement crash

### Game Version

4.11.9 (build 993)

### Describe the bug

Game crashes when trying to proceed to the next turn, due to a scheduled movement of an escorted pair of units (Great General and Mechanized Infantry).

### Steps to Reproduce

1. Load save attached below and proceed to next turn (crash).
2. Reload and find the only unit with a "Moving" action (last Mechanized Infantry unit).
3. Either break the escort or cancel the move to prevent the crash.

### Screenshots

_No response_

### Link to save file

Error report and save:
https://pastebin.com/J0m26VZN

### Operating System

Android

### Additional Information

After encountering the crash scenario (post-win state), I narrowed down the problematic movement orders until only one unit remained and still triggered the crash.

The only noteworthy hints might be that I remember linking and unlinking the pair of units multiple times in the preceeding turn(s), possibly while embarked. Also, I noticed that the "Moving" action is listed with only one of the units (not sure if it's intentional).

tuvus:
This PR's aim is to join the list of recent changes that cause bugs... but hopefully not. 

This PR fixes #11562 by changing `getMovementCostBetweenAdjacentTiles()` to take into account the escorted unit, and changing `getDistanceToTilesAtPosition()` to use the escorted unit's movements as well. The end effect is that we are dealing with the escort logic on a lower level allowing `getDistanceToTiles()` to not have to check the other unit's movement ability.

To make it more understandable, I am changing the escort movement approach to search tiles using both units movement capabilities instead of searching for tiles that both units can move to used in #11057.

This PR is likely to cause problems of some kind, hence my initial joke.

I added two tests, however they are by no means comprehensive and might not even catch most regressions.


tuvus:
This PR also fixes #11920 and #11933.

I think it is ready to merge.

yairm210 (approved):

PR #11815, opened 2024-06-23
Revert Crash caused by Religion Caching

tuvus:
This PR reverts the Religion Caching commit which causes a bad crash on game load. See #11813.

Briefly looking into it, when the city-religion is initializing it is checking resources on all cities and not all cities have been initialized which causes a problem. Commenting out the two problem lines of code resulted in a different crash. Therefore the best option is to revert this for now.

Since this is a game-breaking level bug for any save with religion enabled I will be patching it immediately.


tuvus:
Release Patch

tuvus:
Oops, I didn't release patch would merge before the checks passed.

yairm210:
Thanks!!!

PR #11825, opened 2024-06-23
Update Polish.properties

Lesiakower:


Lesiakower:
What's wrong?

Lesiakower:
> there's a link called "Details"

I've encountered exactly the same problem today ( ｡ •̀ ᴖ •́ ｡) and I still don't know where to search for the line in which the mistake/error is when I go to "Details". Is something wrong with the English text on the left? The issue seems to be touching only the Polish translation...

PR #11839, opened 2024-06-24
AI behaviour changes

EmperorPinguin:
Improves AI citizen assignement during unhappiness, when doing production to gold/science conversion, and when all techs are researched. The AI used to waste a lot of yields by working farms, mines and specialists respectively during these conditions.

Improves cityTileRanking, the AI becomes more likely to use terrain features like rivers and hills, and settles closer together. This was found to increase the turn 200 total world population by approximately 39% (using 5 test games for each, default settings and no CS on identical map)

Limits the amount of cities the AI will build to approximately 5-10 instead of 10-15, to not annoy humans (tested on large 4 corner map).

AI stops constructing military units when in negative unit supply.


EmperorPinguin:
Could someone explain to me why one of the checks failed? It's running just fine on my laptop, and I'm at a loss on how to resolve the issue.

yairm210:
It's docker, don't worry about it, it fails sporadically

EmperorPinguin:
> It's docker, don't worry about it, it fails sporadically

The Docker seems to fail consistently now, but if it is not a problem, the PR can be merged I think.

EmperorPinguin:
@tuvus I made a new PR, could you review the changes made to the cityLocationTileRanker?

tuvus:
The stat changes are much better, although dividing the production value by 6 is a bit much. I think setting the value of science is probably a little too much in case there are mods or something.

I don't agree with the cityTileLocationRanker changes though.

- Why settle on a hill? It's a tradeoff, some city-strength early on for the 10% production from a windmill. I wouldn't say this is worth it, especially with a value of 7. It also doesn't work with mods that well.
- I think the value of rivers should actually be decreased a little, probably moving  some of it's value to freshwater. Note that the extra gold on the tiles already add value to settling by a river.
- Settling close to a mountain could be a benefit, however, we need to check the closest 2 tiles and that might be too mod dependent.
- Settling closer is not good for late game, which the AI is pretty bad at. It also isn't good for players in combination with the AI having more happiness, the AI settles cities too densely for the players happiness since the AI has a big happiness bonus. Players will have a big unhappiness hit when they raze the cities that are too close to eachother. The settlers do get captured by barbarians too much, but that should be solved elsewere in the program. Settling cities closer to other civs also makes them more likely to attack you (also, it isn't nice for players).

The data I collected doesn't seem to show such a big difference. Note that I am taking war and barbarians out of account for reasons stated above. Most of this "simulation" is arbitrary based on our input.


<details><summary>Here is the data I collected</summary>

```

Nation:1(OldLocationLogic)2(NewLocationLogicOfThisPR),Position #Population, #Production, #Technology
(Civs orderd by score, but I left out the score numbers)

4 Civs, large, no war, no barbarians, custom balanced map:
T:200, G2
France:1,BR 232Po, 236Pr, 47T
Persia:2,TL 188Po, 367Pr, 42T
Celts:1,TR 192Po, 189Pr, 41T
Japan:2,BL 112Po, 248Pr, 40T

T:200, G3
Byzantium:2,TL 149Po, 249Pr, 41T
Austria:2,BL 239Po, 400Pr, 42T
Denmark:1,BR 117Po, 179Pr, 41T
Celts:1,TL 179Po, 279Pr, 41T
T:250
Byzantium:2,TL 228Po, 544Pr, 49T
Austria:2,BL 329Po, 693Pr, 53T
Denmark:1,BR 157Po, 471Pr, 51T
Celts:1,TL 304Po, 473Pr, 54T
T:300
Byzantium:2,TL 359Po, 832Pr, 61T
Austria:2,BL 449Po, 1492Pr, 66T
Denmark:1,BR 222Po, 668Pr, 65T
Celts:1,TL 419Po, 927Pr, 68T

T:200 G4
Russia:2,BR 210Po, 689Pr, 51T 
England:1,TL 208Po, 389Pr, 44T
Spain:1,TR 195Po, 261Pr, 41T
Inca:2,BL 131Po, 254Pr, 39T
```
</details>

EmperorPinguin:
> The stat changes are much better, although dividing the production value by 6 is a bit much. I think setting the value of science is probably a little too much in case there are mods or something.

What mods are you aware of that provide benefits to science after all techs have been researced? They must provide very substantial benefits to repeatedly researching the future tech, I'm not aware of one of them. Note that I didn't set the value of production to 0, but to 1/6 of what it used to be (in case of 4:1 conversion of production to science, production is overvalued by a factor of 12*4/7= 6.86).

> I don't agree with the cityTileLocationRanker changes though.
> 
>     * Why settle on a hill? It's a tradeoff, some city-strength early on for the 10% production from a windmill. I wouldn't say this is worth it, especially with a value of 7. It also doesn't work with mods that well.
In the base game and in many mods, settling on a hill provides +1 production. Windmills aren't worth it, as at that point you've already lost out on 200 production by not settling on a hill. Even on flatland, windmills are still not really worth it, unless you really want the engineer slots for Freedom strats. In mods like Rekmod and Ravbalance you get bonus yields for settling on flat land as well (+1 food and +1 production respectively), so it's not necessary to settle on hill. It's still a Walls worth of city strength though, so there's some value in it for early defence. 

>     * I think the value of rivers should actually be decreased a little, probably moving  some of it's value to freshwater. Note that the extra gold on the tiles already add value to settling by a river.

With the old ranking rules, the AI very often settled 2 or 3 tiles away from a river instead of actually on the river. It's better to settle on the river, to have gold tiles in first ring and to enable a watermill. After increasing the value a bit, I've rarely seen them unjustly avoiding rivers.

>     * Settling close to a mountain could be a benefit, however, we need to check the closest 2 tiles and that might be too mod dependent.

True, I wrote that in the comment as well, not every mods contains and observatory or similar building. For the base game however, observatories are good buildings (especially for an AI which doesn't know how to prioritise techs). With Inca or in Fantasia, it makes the AI sacrifice a terrace farm tile, but I hypothesise this to be worth it as there are likely more terrace farm tiles nearby (also in both cases it enables an Observatory). The same logic applies to give a bonus for settling on Fresh water, which is not relevant across all mods, and sacrifices a freshwater farm in the base game, but it makes the AI more likely to settle near other freshwater tiles, and it enables the situational Garden building.

>     * Settling closer is not good for late game, which the AI is pretty bad at. It also isn't good for players in combination with the AI having more happiness, the AI settles cities too densely for the players happiness since the AI has a big happiness bonus. Players will have a big unhappiness hit when they raze the cities that are too close to eachother. The settlers do get captured by barbarians too much, but that should be solved elsewere in the program.

You solve barb problems... by settling close cities. If you place them 7 tiles apart, camps will spawn in-between and you're going to need an excessive amount of scouts to savely move your workers around, and you don't want to overbuild scouts and warriors to clear a camp that's 6 tiles away before being able to settle 7 tiles away. I agree the AI settles too many crappy cities in late game, but that got to do with how long it keeps on settling. That's something I want to get into later, but it's more complex for me to code.

> Settling cities closer to other civs also makes them more likely to attack you (also, it isn't nice for players).

Getting a bonus for settling near other's cities is bad, that's a bug on my part. Although..., the comment in the previous version says `// It is worse to settle cities near our own compare to near another civ`. This is wrong, unless you know what you're doing. There's the concept of tactical cities though, we shouldn't completely ban the AI from settling those.

> The data I collected doesn't seem to show such a big difference. Note that I am taking war and barbarians out of account for reasons stated above. Most of this "simulation" is arbitrary based on our input.
> Here is the data I collected
> 
> ```
> 
> Nation:1(OldLocationLogic)2(NewLocationLogicOfThisPR),Position #Population, #Production, #Technology
> (Civs orderd by score, but I left out the score numbers)
> 
> 4 Civs, large, no war, no barbarians, custom balanced map:
> T:200, G2
> France:1,BR 232Po, 236Pr, 47T
> Persia:2,TL 188Po, 367Pr, 42T
> Celts:1,TR 192Po, 189Pr, 41T
> Japan:2,BL 112Po, 248Pr, 40T
> 
> T:200, G3
> Byzantium:2,TL 149Po, 249Pr, 41T
> Austria:2,BL 239Po, 400Pr, 42T
> Denmark:1,BR 117Po, 179Pr, 41T
> Celts:1,TL 179Po, 279Pr, 41T
> T:250
> Byzantium:2,TL 228Po, 544Pr, 49T
> Austria:2,BL 329Po, 693Pr, 53T
> Denmark:1,BR 157Po, 471Pr, 51T
> Celts:1,TL 304Po, 473Pr, 54T
> T:300
> Byzantium:2,TL 359Po, 832Pr, 61T
> Austria:2,BL 449Po, 1492Pr, 66T
> Denmark:1,BR 222Po, 668Pr, 65T
> Celts:1,TL 419Po, 927Pr, 68T
> 
> T:200 G4
> Russia:2,BR 210Po, 689Pr, 51T 
> England:1,TL 208Po, 389Pr, 44T
> Spain:1,TR 195Po, 261Pr, 41T
> Inca:2,BL 131Po, 254Pr, 39T
> ```
That's very interesting, I didn't expect a change by -8.4%. In my testing, the total world populations were 
With old rules:  274, 289, 224, 199, 284; average: 254
With new rules: 344, 418, 358, 341, 305; average: 353
The average civ's size changed by +39% at t200, I can send you the t1 save file I used for this experiment.

SeventhM:
> In the base game and in many mods, settling on a hill provides +1 production

Not all mods. For example, I have a personal edit of Deciv where this is not the case. Also, some mods may have other terrains that also gives production, so this should not be seen as a reliable metric. Also, some mods may give production from a resource that your settled on

To be compatible with all mods, please looks primarily at stats, not the terrain type or strength boost. That makes asking this question much less necessary

> It's better to settle on the river, to have gold tiles in first ring

Reminder that not all mods have gold here, and once BNW is worked on, it also won't have the gold from rivers. More reason to look at stats, not necessarily terrain type (though this one is more excusable because many mods do actually have river specific bonus options)

tuvus:
Science is still useful when all technologies have been researched. Researching the future tech gives score so the value of science should not be put to 0.

> Windmills aren't worth it

That isn't the case for every city. For cities settled later on a windmill can be built sooner and can add more to the production bonus.

> Without war, you can get away with a whole lot of things, so you should include it in testing.

We are trying to test the city foundation logic though, and the stats that a civ gets from conquering another civ can through that off.

> You solve barb problems... by settling close cities.

Once again, this isn't true, while it may help it won't solve the barbarian problem as I have said before. This isn't the place to try and solve that problem. The solution lies in having units protect workers and settlers better. Although that isn't so easy to do.

If you don't know, the AI gets a big happiness bonus by default. They get 3 extra happiness from the start and a 40% reduction in any unhappiness. This allows them to spam cities a lot more than players can. In the future, it would be nice to have the AI not have to use this crutch. It also isn't good for the AI to abuse it's happiness advantage since it can make it much harder for players to raze their badly placed cities. This change was a part of #10428.

Please understand that I included the data to show that it isn't very accurate. Changing the civ count, map size, AI difficulty, spawn location, nation and many other factors can influence it.

> I agree the AI settles too many crappy cities in late game, but that got to do with how long it keeps on settling.

The minimum settling value plus the negative values for settling closer to cities is meant to mitigate this problem.

SeventhM:
> That isn't the case for every city. For cities settled later on a windmill can be built sooner and can add more to the production bonus.

In Lekmod/Rekmod, windmills aren't locked to flat cities and it's still considered a mediocre building. I'm not against assuming that generically the building isn't worth it. At that point, though, the problem is either the AI needs to be able to figure out if the building is there or we shouldn't really consider it in the first place if we was full mod compatibility

EmperorPinguin:
@yairm210 I just witnessed the AI preventing space victory again, this time by placing a Great Artist in capital.

yairm210:
Aha
That is interesting, can you send me the save file?

EmperorPinguin:
> Aha That is interesting, can you send me the save file?

I didn't keep the save file, but it should be possible to recreate by starting up a new game.

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -84,16 +84,26 @@ object CityLocationTileRanker {
...
+        // Observatories are good, but current implementation no mod-friendly
+        if (isNextToMountain) tileValue += 5
+        if (newCityTile.isAdjacentToRiver()) tileValue += 14
         if (newCityTile.terrainHasUnique(UniqueType.FreshWater)) tileValue += 5
         // We want to found the city on an oasis because it can't be improved otherwise
         if (newCityTile.terrainHasUnique(UniqueType.Unbuildable)) tileValue += 3
         // If we build the city on a resource tile, then we can't build any special improvements on it
         if (newCityTile.resource != null) tileValue -= 4
+        // We don't want to settle 1-tile islands and promontories with few good tiles
+        for (nearbyTile in newCityTile.getTilesAtDistance(1))
```
SeventhM, line 104: Good idea, wrong approach. Some mods might dump way too many good resources in the water with way too high rates. Also, if we check the states of a tile, we should naturally be able to figure this out, which should also let us cover checking deserts

SeventhM:
>Merge branch 'yairm210:master' into master

For the record, btw, if you use separate branches for your PRs and keep your master branch clean, you shouldn't need to do this so often for any of your branches. Git is smart enough to still figure out what's going on

> Settling closer is not good for late game, which the AI is pretty bad at.

For the record, if I'm going to throw in my opinion, the AI settles a lot of cities thanks to it's extra happiness. In that case, it's better to settle closer together anyways to have more room for more cities, regardless of if we argue ease of fighting barbarians. I'm not sure I agree with giving the highest vales to the closest distance (I'd maybe give 5 tiles the highest value, then 4, then 6, then 7), but surely they should want to settle closer together if for no other reason than this and to avoid annoying players

Reminder, there are some mods (Outlaws of the wastes comes to mind) that actually increases the minimum settle distance out of spite because of how the AI currently settles, so making them settle more compactly doesn't seem to bad an idea regardless 

> Settling on river is still good in BNW/Rekmod because of the water mill, and potentially gardens

An aside here, but taking a glance at Alpha Frontier shows very, very few actual river benefits, so assuming mods inherently have river benefits might actually be the wrong approach too. Huh. Yeah, we'll need a way to look at buildings that are possible

EmperorPinguin:
> > Merge branch 'yairm210:master' into master
> 
> For the record, btw, if you use separate branches for your PRs and keep your master branch clean, you shouldn't need to do this so often for any of your branches. Git is smart enough to still figure out what's going on

In the past I've run into weird conflicts due to not keeping things up to date, so I prefer it this way.

> > Settling closer is not good for late game, which the AI is pretty bad at.
> 
> For the record, if I'm going to throw in my opinion, the AI settles a lot of cities thanks to it's extra happiness. In that case, it's better to settle closer together anyways to have more room for more cities, regardless of if we argue ease of fighting barbarians. I'm not sure I agree with giving the highest vales to the closest distance (I'd maybe give 5 tiles the highest value, then 4, then 6, then 7), but surely they should want to settle closer together if for no other reason than this and to avoid annoying players
> 
> Reminder, there are some mods (Outlaws of the wastes comes to mind) that actually increases the minimum settle distance out of spite because of how the AI currently settles, so making them settle more compactly doesn't seem to bad an idea regardless

I personally prefer 4 tiles, if all else equal, but I could make it in the order 5 >4 > 6 >7. 

> > Settling on river is still good in BNW/Rekmod because of the water mill, and potentially gardens
> 
> An aside here, but taking a glance at Alpha Frontier shows very, very few actual river benefits, so assuming mods inherently have river benefits might actually be the wrong approach too. Huh. Yeah, we'll need a way to look at buildings that are possible
Yes, we discussed this before, It would be better if the AI looked up the terrain requirements for buildings. I don't know how to code that though, and for base game, rivers are very strong.

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
EmperorPinguin: > Good idea, wrong approach. Some mods might dump way too many good resources in the water with way too high rates. Also, if we check the states of a tile, we should naturally be able to figure this out, which should also let us cover checking deserts

Isn't `var rankTileValue = Automation.rankStatsValue(rankTile.stats.getTileStats(null, civ, uniqueCache), civ)` already taking care of this?

EmperorPinguin: Also, especially in the case of coast, we should look for things like the Rekmod Exploration tree and act accordingly (treasure fleets btw is meant to apply only to coast and ocean without resources).

SeventhM:
> It's docker, don't worry about it, it fails sporadically

@yairm210 you sure? Docker is saying it's missing the sdk, and has said so repeatedly since https://github.com/yairm210/Unciv/pull/11839/commits/9040baba5f76bad2e9a7f177755b443c6cf23a31

build.gradle.kts
```diff
@@ -24,7 +24,7 @@ buildscript {
     }
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${com.unciv.build.BuildConfig.kotlinVersion}")
-        classpath("com.android.tools.build:gradle:8.2.2")
+        classpath("com.android.tools.build:gradle:8.5.0")
```
SeventhM, line 27: Please revert. Seems to be acting as a blocker (haven't checked the details behind it yet). Would be out of scope for the PR anyways

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-bin.zip
```
SeventhM, line 3: Please revert here too

build.gradle.kts
EmperorPinguin: Is that the problem? Android Studio offered me an update, I didn't think it would cause problems.

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
EmperorPinguin: Well, it's not enough, as the AI *loves* settling small islands and promontories. With the change I made, it becomes somewhat less likely to do so. Alternatively, the bonus from settling on coast (which was  set to +8 by a previous dev) could be reduced.

EmperorPinguin:
@SeventhM, it indeed worked.

build.gradle.kts
EmperorPinguin: I wasn't aware it mattered, so I just clicked on the update button. It looks like the problem has now been solved (I'm getting errors within Android studio, but not on Github).

yairm210 (approved):

yairm210:
All around looks like good changes 👍🏿

EmperorPinguin:
> All around looks like good changes 👍🏿

One thing that's debatable I think is the addition to ConstructionAutomation. I often witness the AI queueing up a random world wonder very early in the game, and then either not finishing it (wasting a lot of production) or finishing but still becoming irrelevant t40. Setting a minimum city number of e.g. 3 mostly fixes this, but there may be maps with small island spawns were OCC wonder spam is the only thing available.

yairm210:
We can always change things later
For now this is going in

PR #11857, opened 2024-06-26
Workaround for clipboard size limitation

### LINKED ISSUE #11822: Cannot copy save game via in-game button

### Game Version

4.12.2-patch1

### Describe the bug

Before patch1 (`4.12.2`) upon clicking "Copy saved game to clipboard" I was just receiving the "Could not save game to clipboard" message.
With `4.12.2-patch1` the game just terminates immediately.

### Steps to Reproduce

- load a game
- reopen the load game dialog
- select that same game you've just loaded
- click the "Copy saved game to clipboard" button
AR: the game crashes
ER: the game data gets loaded into clipboard

### Screenshots

Unciv_g4aLpHP5dK

### Link to save file

_No response_

### Operating System

Windows

### Additional Information

_No response_

SomeTroglodyte:
Fixes #11822
See #6929 (stacksize increase)
See [Higher available clipboard size](https://github.com/orgs/LWJGL/discussions/769) (lwjgl discussion)

### Clipboard size limitation begone
This is the big and hard hammer. [Softer hammers](https://github.com/yairm210/Unciv/assets/63000004/7f2b6e73-cb66-42b9-bf28-fa15a61370b9) are feasible...

On desktop, replace Gdx clipboard (the Lwjgl3Clipboard implementation) with an alternate implementation of the interface using AWT, and return it from `Gdx.app.clipboard`, thus replacing _all_ clipboard access in Unciv (yes affects copy&paste in TextField too etc etc). In other words, if Lwjgl is crap, redirect to a better API.

### Risk
We've had problems loading the AWT library while Unciv is running - which is why we now have a non-AWT file picker for "load from custom location". This here hopes that using `Toolkit.getDefaultToolkit().systemClipboard` is harmless, maybe because it's independent enough, hopefully, let's pray. See "Softer hammers" - those would need to touch every single access, around 30 to 50 or so.

### Alternative
Increase stackSize once again? To what? 1M? 2M? 3.141592653M?

### Note
The way this needed to be hooked in is interesting to read. A class that has an instance that is doing its job, but its constructor never ran, a breakpoint in `init{}` never hits??? Yup, that's the Gdx design. `init` on a Gdx.Lwjgl3Application subclass runs, but only after the game is done. Make do without initialization - or - don't subclass but delegate, quite a bit more complicated to pull off because you can't use kotlin's `by` help without falling into the same trap.

### DIFF
--- a/desktop/src/com/unciv/app/desktop/AwtClipboard.kt
+++ b/desktop/src/com/unciv/app/desktop/AwtClipboard.kt
@@ -1,0 +1,42 @@
+package com.unciv.app.desktop
+
+import java.awt.Toolkit
+import java.awt.datatransfer.DataFlavor
+import java.awt.datatransfer.StringSelection
+import java.awt.datatransfer.Clipboard as ClipboardAwt
+import com.badlogic.gdx.utils.Clipboard as ClipboardGdx
+
+/**
+ *  A plug-in replacement for Gdx Lwjgl3Clipboard that goes through AWT instead of GLFW and removes the stack size limitation.
+ *
+ *  Gdx.app.clipboard on desktop is a Lwjgl3Clipboard instance, which allocates a buffer on the stack for UTF8 conversion at the Lwjgl3-GLFW interface.
+ *  The default stack size is a severe limitation, which we **originally** treated by increasing the limit, which can only be done statically at launch time:
+```
+         // 386 is an almost-arbitrary choice from the saves I had at the moment and their GZipped size.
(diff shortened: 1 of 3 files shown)


yairm210:
Sorry to bust your bubble, but...
The clipboard problem is mainly problematic on *android*, and IIRC android doesn't have AWT

yairm210 (approved):
The fact that it's only in Desktop reduces my anxiety that we'll somehow import libraries that aren't available

PR #11858, opened 2024-06-26
Minor TranslationFileWriter maintenance

SomeTroglodyte:
Replace confusing [Food] Focus template with prettier [stat] Focus
Confirmed: Old entries automatically replaced keeping translations.
Please don't _run_ TFW with this until [Galician ](https://github.com/yairm210/Unciv/pull/11848) is resolved - it would add to that PR's confusion.


yairm210:
Surely "[stat] Focus" would be better to add in template.properties? Is there a reason it's a special boy and is in code?

PR #11875, opened 2024-06-28
Optimize screen orientation

HChenX:
I think using sensor-based orientation settings is better than just relying on the default values. 

Currently, setting the screen to landscape mode in the game only uses the default orientation, which isn't always convenient. For example, when playing the game while lying in bed, the charging cable might not be long enough to rotate the screen to the default landscape mode.  :)
﻿
Although the auto option is available, it sometimes switches the screen to portrait mode, which isn't what I want.

Thanks!

### DIFF
--- a/android/src/com/unciv/app/AndroidDisplay.kt
+++ b/android/src/com/unciv/app/AndroidDisplay.kt
@@ -122,11 +122,12 @@ class AndroidDisplay(private val activity: AndroidApplication) : PlatformDisplay
 
     override fun setOrientation(orientation: ScreenOrientation) {
         val mode = when (orientation) {
-            ScreenOrientation.Landscape -> ActivityInfo.SCREEN_ORIENTATION_USER_LANDSCAPE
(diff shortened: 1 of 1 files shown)


yairm210:
I remember there was a reason we had these settings... @SomeTroglodyte do you recall? 
Since this is very easily reversible, worst case we can undo it

HChenX:
> Since this is very easily reversible, worst case we can undo it

Can you tell me the detailed reason?
Because I think this modification would be more convenient, such as not being restricted when using wired headphones or charging cables.
My test seems to be working fine.

HChenX:
> "Make MIUI Great Again" - ha, just overwrite it with LineageOS

 no bro, miui is shit :)

yairm210 (approved):

HChenX:
> And still, the [Great Chinese Firewall](https://github.com/yairm210/Unciv/blob/c89b2666c9643c5f0f16d2ed84eb499f53782ceb/android/assets/jsons/Civ%20V%20-%20Gods%20&%20Kings/Buildings.json#L1117-L1123) is a good thing - when playing Unciv with Espionage on.

y - although it prevents me from accessing Google.

PR #11898, opened 2024-07-01
Translate numbers to selected language

### LINKED ISSUE #11895: Feature request: Numbers Translation
### Before creating

### Problem Description

As everyone knows, any number we display on our current UI is always English. But if you are not using English as your selected language, the numbers should represent your selected language instead.

### Related Issue Links

_No response_

### Desired Solution

UI numbers should get auto-translated based on language selection. `NumberFormat` from java, seems to be the perfect choice for this.

### Alternative Approaches
In properties:
```properties
0123456789 = ০১২৩৪৫৬৭৮৯
```
Then manually implement digit translation function
```kt
val digits = "0123456789".toList()

fun Number.tr(): String {
    return this.toString().map { it ->
        if (!it.isDigit()) return@map it
        return@map digits[it.digitToInt()]
    }.joinToString("")
}
```

### Additional Context

https://discuss.kotlinlang.org/t/how-to-override-number-tostring/28830
This seems to work:
```kt
import java.text.NumberFormat
import java.util.Locale

fun main() {
    val nf = NumberFormat.getInstance(Locale.of("bn", "BD"));
    val test = 1234567890.0123456789
    println(nf.format(test)) // ১,২৩৪,৫৬৭,৮৯০.০১২
}
```

touhidurrr:
resolves #11895


touhidurrr:
Ok, here is the result for now. Green means translated properly and red means was not translated for some reason. Maybe  the regex I wrote was problematic?
image

touhidurrr:
@yairm210 how do i fix this delekt warning?

touhidurrr:
Although not sure why the regex is having issues. See: https://regexr.com/82b23

touhidurrr:
>  tweaking stuff

How to check this locally?

yairm210:
Detekt fixed
The reason the main screen contains some untranslated numbers is because they're never .tr()'d

touhidurrr:
> I believe both \b are misguided

I suspected this also, but all of my tests said otherwise.

> aha! setText is direct Gdx, no translate call. Many more of course.

This has to be the reason why

> translateIndividualWord

So, in the end the best option would be trying to detect and translate numbers during placeholder fill. I tried this but it does not work at all.

```diff
--- a/core/src/com/unciv/models/translations/Translations.kt      
+++ b/core/src/com/unciv/models/translations/Translations.kt      
@@ -491,9 +491,16 @@ fun String.fillPlaceholders(vararg strings: String): String {
     if (keys.size > strings.size)
         throw Exception("String $this has a different number of placeholders ${keys.joinToString()} (${keys.size}) than the substitutive strings ${strings.joinToString()} (${strings.size})!")      

+    val numberFormatter = UncivGame.Current.settings.getNumberFormatFromLocale()
+
+    val translatedStrings = strings.map {
+        val num = toDoubleOrNull() ?: return@map it
+        return@map numberFormatter.format(num)
+    }
+
     var filledString = this.replace(squareBraceRegex, "[]")      
     for (i in keys.indices)     
-        filledString = filledString.replaceFirst("[]", "[${strings[i]}]")
+        filledString = filledString.replaceFirst("[]", "[${translatedStrings[i]}]")
     return filledString
 }
```

touhidurrr:
Need guidance, @yairm210. What do you think should be done?

yairm210:
A. Not during placeholder fill, you'll miss all the non-placeholder labels
B. Yes, you'll need to add like a .tr() for numbers

touhidurrr:
Although I am using `String.translateIndividualWord()` currently in my pr, my concern is that many translations might have numbers that we might miss. For example: `100a` is a word, but cannot be parsed with `toDouble()`. So, this will be missed.
And from the looks of it, I am not getting words here after all.
image

touhidurrr:
font issues aside, this is the current upadte:
image
image
image

As expected, stuff like `0%` or `2/2` is broken.

touhidurrr:
Caching NumberFormat is also a problem. It is getting saved to `GameSettings.kt` which I do not want to.

touhidurrr:
> @Transient

Where do I put it?

touhidurrr:
> B. Yes, you'll need to add like a .tr() for numbers

Done in 5e47a3a

touhidurrr:
@yairm210 @SomeTroglodyte need review.

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -193,6 +202,18 @@ class GameSettings {
         return Collator.getInstance(getCurrentLocale())
     }
 
+    fun getNumberFormatFromLanguage(language: String): NumberFormat {
+        languageToNumberFormat[language].also { if (it !== null) return it }
+
+        return NumberFormat.getInstance(getLocaleFromLanguage(language)).also {
+            languageToNumberFormat[language] = it
+        }
+    }
```
yairm210, line 211:
```suggestion
    fun getNumberFormatFromLanguage(language: String): NumberFormat =
        languageToNumberFormat.getOrPut(language) { NumberFormat.getInstance(getLocaleFromLanguage(language))
        }
```

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -140,6 +141,10 @@ class GameSettings {
     /** used to migrate from older versions of the settings */
     var version: Int? = null
 
+    /** NumberFormat cache. Key: Language, Value: NumberFormat **/
+    @Transient
+    val languageToNumberFormat = mutableMapOf<String, NumberFormat>()
```
yairm210, line 146: I'm not conviced this should sit in the GameSettings. This sounds like something for translation companion object, this and the getLocaleFromLanguage function

touhidurrr: I added this here because locale get or set functions was here. So, you are suggesting that we move `fun getLocaleFromLanguage(language: String): Locale`, `fun updateLocaleFromLanguage()`, fun `getNumberFormatFromLanguage(language: String): NumberFormat` and `fun getCurrentNumberFormat(): NumberFormat` functions to be moved to translations file?

Also not sure it I implement `Number.tr()` inside Translations class if it would be accessible to rest of the project. Fact: new to Kotlin.

yairm210: All of these apart from the "update locale" which is a settings thing (not sure why we need that locale but sure, probably for string sorting or something)

yairm210: Sorry not the "get current" either, that depends on current language which is setting

touhidurrr: `getLocaleFromLanguage()` also depends on `enum LocalCode` inside settings file.

touhidurrr: I guess only `fun getNumberFormatFromLanguage(language: String): NumberFormat` and `languageToLocale` map should be moved.

touhidurrr: resolved with c567e94

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -443,6 +452,8 @@ private fun String.translatePlaceholders(language: String, hideIcons: Boolean):
 private fun String.translateIndividualWord(language: String, hideIcons: Boolean): String {
     if (Stats.isStats(this)) return Stats.parse(this).toString()
 
+    toDoubleOrNull().also { if (it != null) return it.tr(language) }
```
yairm210, line 455: Please stop using the "also" format everywhere, instead this should be 2 lines

yairm210:
Please add tests for the following to ensure they remain the same text:

- "1" 
- "1.0"
- "+1"
- "-1"

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -231,6 +232,14 @@ class Translations : LinkedHashMap<String, TranslationEntry>() {
...
         const val shouldCapitalizeString = "StartWithCapitalLetter"
         const val effectBeforeCause = "EffectBeforeCause"
+
+        // NumberFormat cache, key: language, value: NumberFormat
+        private val languageToNumberFormat = mutableMapOf<String, NumberFormat>()
+
+        fun getNumberFormatFromLanguage(language: String): NumberFormat =
+            languageToNumberFormat.getOrPut(language) {
+                NumberFormat.getInstance(UncivGame.Current.settings.getLocaleFromLanguage(language))
+            }
```
yairm210, line 242: The function for "getLocaleFromLanguage" should be in Translations, currently it's all spaghetti - in gamesettings we call getnumberformat from translations which calles getlocale from settings

And why do we need the number formats cached in 2 separate places?

core/src/com/unciv/models/metadata/GameSettings.kt
```diff
@@ -140,6 +142,10 @@ class GameSettings {
     /** used to migrate from older versions of the settings */
     var version: Int? = null
 
+    /** to cache current NumberFormat **/
+    var numberFormat: NumberFormat? = null
+    var numberFormatLanguage: String? = null
```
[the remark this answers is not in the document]
touhidurrr: resolved in d0b06ab

core/src/com/unciv/models/translations/Translations.kt
touhidurrr: > The function for "getLocaleFromLanguage" should be in Translations

Not sure what you meant. Isn't it already in companion object of Translations class.

touhidurrr: > And why do we need the number formats cached in 2 separate places?

Well... That's a good point... I actually forgot why I did that... fixed in 495bdd1

touhidurrr:
> Please add tests for the following to ensure they remain the same text:
> 
> * "1"
> * "1.0"
> * "+1"
> * "-1"

As we are using `toDoubleOrNull()`, I don't think this is possible. Should we force-change it?

core/src/com/unciv/models/translations/Translations.kt
touhidurrr: resolved in 9099a6648ff7bb4651ed3ad5c54d13300e66a107

touhidurrr:
> As expected, stuff like 0% or 2/2 is broken.

> "+1", "1.0"

One plan to fix both cases would be something like this: [playground](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4wLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJwYWNrYWdlIHRvdWhpZHVycnJcblxuaW1wb3J0IGphdmEudGV4dC5OdW1iZXJGb3JtYXRcbmltcG9ydCBqYXZhLnV0aWwuTG9jYWxlXG5cblxudmFsIG5mID0gTnVtYmVyRm9ybWF0LmdldE51bWJlckluc3RhbmNlKExvY2FsZShcImJuXCIsIFwiQkRcIikpXG5mdW4gTnVtYmVyLnRyKCk6IFN0cmluZyA9IG5mLmZvcm1hdCh0aGlzKVxuXG5mdW4gbWFpbigpIHtcbiAgICB2YWwgbnVtU3BsaXQgPSBSZWdleChcIlwiXCIoPzw9W1xccyUrLy4tXSl8KD89W1xccyUrLy4tXSlcIlwiXCIpXG4gICAgdmFsIHRlc3RzID0gbGlzdE9mKFwiKzFcIiwgXCItMVwiLCBcIjEuMFwiLCBcIisxLjBcIiwgXCItMS4wXCIsIFwiMCVcIilcbiAgICBmb3IgKHQgaW4gdGVzdHMpIHtcbiAgICAgICAgdmFsIHRyYW5zbGF0aW9uID0gdC5zcGxpdChudW1TcGxpdCkubWFwIHtcbiAgICAgICAgICAgIGlmIChpdC5sZW5ndGggPCAxKSByZXR1cm5AbWFwIGl0XG4gICAgICAgICAgICB2YWwgbnVtID0gaXQudG9Eb3VibGVPck51bGwoKSA/OiByZXR1cm5AbWFwIGl0XG4gICAgICAgICAgICByZXR1cm5AbWFwIG51bS50cigpXG4gICAgICAgIH0uam9pblRvU3RyaW5nKFwiXCIpXG4gICAgICAgIHByaW50bG4oXCInJHQnIC0+ICckdHJhbnNsYXRpb24nXCIpXG4gICAgfVxufSJ9)

Output:
```js
'+1' -> '+১'
'-1' -> '-১'
'1.0' -> '১.০'
'+1.0' -> '+১.০'
'-1.0' -> '-১.০'
'0%' -> '০%'
```

Although, this code needs to be run after translation call. Should I continue with it?

touhidurrr:
Although this did not fix cases where `String.tr()` is not called. This this all fix other cases. From my local test, all previously broken numbers are fixed here. (Except main screen ones where `.tr()` is not called)
image

touhidurrr:
Tell me if I should add this to this pr.

touhidurrr:
Well, it will commit them anyways. (turned out to be less complex than my initial write).
Tell me what you think!

yairm210:
There should be nowhere that we need to translate "2/2" or "5%" directly, since we can translate each piece by putting them in square brackets (like "{2}/{2}")
For the examples I listed above, it *is* important - For example "[+20]% Strength" should be displayed - in English *as* "+20% Strength" and NOT as "20% Strength", so losing the + because of int parsing -> stringifying is not acceptable

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -443,7 +468,9 @@ private fun String.translatePlaceholders(language: String, hideIcons: Boolean):
 private fun String.translateIndividualWord(language: String, hideIcons: Boolean): String {
     if (Stats.isStats(this)) return Stats.parse(this).toString()
 
-    val translation = UncivGame.Current.translations.getText(this, language, TranslationActiveModsCache.activeMods)
+    val translation = UncivGame.Current.translations.getText(
+        this, language, TranslationActiveModsCache.activeMods
+    ).replace(digitsRegex) { it.value.toLong().tr(language) }
```
yairm210, line 473: Looks good, I personally can't think of a problem except for maybe performance, which should hopefully not be a problem

yairm210 (approved):

PR #11902, opened 2024-07-02
Version rollout

yairm210:


touhidurrr:
Would have liked if it was known that someone has started Bangla Translation in changelog. Might have gotten the interest of other translators.

yairm210:
I'll add it in

PR #11908, opened 2024-07-03
AI citizen focus change

EmperorPinguin:
Reworks food prioritization, to become more 'gradual' instead of exactly +2  food surplus.

Reduces value of gold to 1/2 of production, as is applicable during most of the game

Increases value of science, as the great scientist pionts are implicit in their evaluation.

Increases value of Faith. Currently detrimental to AI (they spend faith on inquisitors preventing space victory), but more helpful to humans.


yairm210 (approved):
Would like @tuvus to weigh in as well, but this looks good to me

tuvus:
What was the reason for useing the science focus again? 
If its because of great persons spawning in them we could probably change the spawning AI to prioritize non-puppeted cities. We probably have leeway to make that decision here even if it's not in Civ V.

The reason for having the gold focus is to support building up the gold bank to buy a courthouse immediately. Food would also make sense since the city likely lost population from being conquered.

EmperorPinguin:
> What was the reason for useing the science focus again? If its because of great persons spawning in them we could probably change the spawning AI to prioritize non-puppeted cities. We probably have leeway to make that decision here even if it's not in Civ V.

It's not about the spawning, it's about the great person point generation. With gold focus, they'll work merchant slots and the human player is forced to annex them to avoid a great merchant from spawning. If they work scientist slots, this isn't an issue.

> The reason for having the gold focus is to support building up the gold bank to buy a courthouse immediately. Food would also make sense since the city likely lost population from being conquered.

I think avoid growth makes more sense, as not everyone wants their unimportant cities to grow and consume happiness. Setting "avoid growth" by default for puppet cities would be too drastic however, as puppet cities growing pop can be good. I think Science focus is better, as then they'll still prioritise trade posts over mines.

tuvus:
What is so bad about the great person points?

EmperorPinguin:
> What is so bad about the great person points?

If you do a lot of war in the late classical/early medieval era, you won't be working scientist slots yet, so the puppet cities working merchant slots will cause great merchants to spawn. This is often undesirable as it sacrifices a great sperson of a different type, and you may not be able/willing to work engineer slots at that stage.

tuvus:
Huh, I didn't know they were pooled.
But why slash the gold production for that when we can target the great merchants individually?

EmperorPinguin:
> Huh, I didn't know they were pooled. But why slash the gold production for that when we can target the great merchants individually?

We could target the great merchants individually. When it comes to regular tile assignement, I'm not sure if generally humans prefer the city to produce gold, or food and production to let the city develop for annexation at a later time. If it's later in the game and they want them to produce gold instead of growing, the farms can be replaced with trade posts, which gain +1 science with Free Thought. Note that plantations (3 gold) are by default already valued higher than a non-freshwater farm (1 food).

yairm210:
@tuvus As far as I'm concerned, awaiting your approval, you da real MVP

tuvus:
If the goal is to reduce Great person points in puppeted cities, could we just set their focus to be balanced instead? Science seems a bit arbitrary, and that way we don't reduce great merchants for people who are okay with them.

EmperorPinguin:
@tuvus, if we're in negative happiness and we've got a 1-pop expand city with e.g. Old Faithful, I think in many cases it's best to work the happiness so our other cities can continue growing. Similary, if we're losing a lot of gold per turn, we need all our cities' help to avoid units disbandment. I often work gold tiles in small pop expand cities, especially if I'm both in negative happiness and negative gold. The AI rightfully values gold lower in small cities (you don't want citizens to get distracted by camp-luxury tiles early on), but shouldn't be completely prevented from working gold/happiness if need be.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -82,31 +85,35 @@ object Automation {
...
+        if (city.civ.stats.statsForNextTurn.gold < 0) {
+            // We have a global problem, we need to deal with it before it leads to science loss
+            yieldStats.gold *= 2
+        }
 
         if (city.civ.getHappiness() < 0) {
-            // 75% of excess food is wasted when in negative happiness
-            yieldStats.food /= 4
+            yieldStats.happiness *= 2
+        }
```
tuvus, line 112: Can you put this in the duplicate if-statement above?

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -61,6 +61,9 @@ object Automation {
             for (unique in localUniqueCache.forCityGetMatchingUniques(city, UniqueType.UnhappinessFromPopulationTypePercentageChange))
                 if (unique.params[1] == "Specialists" && city.matchesFilter(unique.params[2]))
                     yieldStats.happiness -= (unique.params[0].toFloat() / 100f)  // relative val is negative, make positive
+            yieldStats.science *= 1.5f // we want to be working scientists
+            if (city.civ.getHappiness() < 0) // slotting Democracy specialists is an easy fix to happiness problems
+                yieldStats.happiness *= 2
```
tuvus, line 66: If you want to do something like this, look at line 57 for  the food-related unique. Otherwise you are just guessing that we have the democracy policy unlocked.

tuvus: Actually, we already have this implemented, line 61-63 accomplish this.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -61,6 +61,9 @@ object Automation {
             for (unique in localUniqueCache.forCityGetMatchingUniques(city, UniqueType.UnhappinessFromPopulationTypePercentageChange))
                 if (unique.params[1] == "Specialists" && city.matchesFilter(unique.params[2]))
                     yieldStats.happiness -= (unique.params[0].toFloat() / 100f)  // relative val is negative, make positive
+            yieldStats.science *= 1.5f // we want to be working scientists
```
tuvus, line 64: This is more of a player preference, can we leave this out?

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -82,31 +85,35 @@ object Automation {
...
         if (city.civ.getHappiness() < 0) {
-            // 75% of excess food is wasted when in negative happiness
-            yieldStats.food /= 4
+            yieldStats.happiness *= 2
+        }
+
+        if (city.isPuppet) {
+            // Puppets need to produce gold, but no great merchants
+            yieldStats.gold *= 2
         }
```
tuvus, line 117: Can we also leave this out? Setting a the city focus is a bit different than hard coding some values behind the scenes.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -39,7 +39,7 @@ object Automation {
             gpp = specialistInfo.greatPersonPoints.sumValues().toFloat()
         }
         gpp = gpp * (100 + city.currentGPPBonus) / 100
-        rank += gpp * 3 // GPP weight
+        rank += gpp * 1 // GPP weight
```
tuvus, line 42: What exactly is the reasoning for this? The Automation doesn't use specialists very often in the first place.

core/src/com/unciv/models/stats/Stats.kt
```diff
@@ -148,9 +148,9 @@ open class Stats(
     fun applyRankingWeights() {
         food *= 14
         production *= 12
-        gold *= 8 // 3 gold worth about 2 production
+        gold *= 6 // 2 gold worth about 1 production
```
tuvus, line 151: The AI generally has a problem with obtaining gold. This is partly because there aren't many gold opportunities early on. But reducing the value of gold wouldn't help.

EmperorPinguin: It's not just the AI who uses AI citizen assignment, but also many humans do it. In the early game, the most efficient use of gold is via unit upgrades, which is a gold:production conversion of 2:1 Late in the game, with stock exchanges built, the conversion of base gold to base production is more like 2*(1+0.1+0.1+0.15+0.2) : 1*(1+0.25+0.25+0.25) = 1.77:1, or even 1.19:1 with Professional Army/Pentagon. It makes sense to scale the value of gold with era, but I'm not mistaken the AI already takes modifiers from buildings into account.

core/src/com/unciv/logic/automation/Automation.kt
EmperorPinguin: The AI currently doesn't distinguish between types of great person points, and doesn't know well how to use the ones they produce, which makes a calculation like this not very useful. I think it's best to set it to a low value, to act as a 'tiebreaker' between an engineer slot and an unimproved hill.

core/src/com/unciv/logic/automation/Automation.kt
EmperorPinguin: The base value of science is pretty low. We could increase it to make the AI work scientist slots, but this would lead to them getting stuck working trade posts and Krakatoa. The 'cleanest' way to properly implement this, is to make the AI look up the abilities of the great person corresponding to the great person points and value the slots accordingly, but is a bit too complex for me to implement. It seems multiplying the value of science by 1.5 is approximately good enough to let the AI rank scientists as the highest of the specialists.

EmperorPinguin: Great scientists in the base game are generally regarded as the best great persons. In Rekmod, each type of great person is in their own gpp pool, while manufactories and customs houses are much better, You could argue engineer or merchant slots are the better slots, but when I visit the Lekmod discord, people there are obsessed with discussing scientists management, so I think it holds true across base rulesets.

core/src/com/unciv/models/stats/Stats.kt
EmperorPinguin: The reason AI loses as much gold as it does, got more to do with how it goas about picking city constructions and building roads without connecting cities. Improving their worker management might help quite a bit, but I'm not sure how to do so.

core/src/com/unciv/logic/automation/Automation.kt
EmperorPinguin: > This is more of a player preference, can we leave this out?
What's your proposal on letting the AI work the right specialist slots?

tuvus: The prerequisite questions are:

1. How do we figure out the best research slot? 
2. Do we want all the AI Civs to choose one specialist type?
3. Do we want to choose a specific strategy that we deem best and try to get the AI and/or players to use it?

For the first question, if we want to dynamically determine this we'd need some sort of unique evaluation function that can rank the uniques together given the current game state and give us a value. It seems like we are having quite some trouble with this. Choose the best one for G&K or some other version would not play well with mods which could re-balance or completely change them.

For the second and third questions, each nation has their own personality (implemented in  #10939), meaning we encourage them to be different from each-other. So why would we try and override that? Making decisions for players doesn't tend to go well, this PR for example was the result of a lighter version of that.

EmperorPinguin: > For the first question, if we want to dynamically determine this we'd need some sort of unique evaluation function that can rank the uniques together given the current game state and give us a value. It seems like we are having quite some trouble with this. Choose the best one for G&K or some other version would not play well with mods which could re-balance or completely change them.

True, and also, uniques also tend to get renamed quite a bit. The simplest (but not fool-proof) way to do this would be to adjust the evaluation of science, as the value of GS points (which are good in both bas game and Rekmod at least) is implicit in it.

> For the second and third questions, each nation has their own personality (implemented in #10939), meaning we encourage them to be different from each-other. So why would we try and override that?

If I play test with the current implementation, a custom Default does show to rank scientists as the highest of the specialists, but doesn't work all of them all the time. Songhai keeps prioritizing merchants, and India keeps prioritizing artists (I haven't tested for every civ).

> Making decisions for players doesn't tend to go well, this PR for example was the result of a lighter version of that.

The default focus is currently not very good, so I've got to manually assign all my citizens. From playtesting, it appears this PR solves some of it, but I'll likely give it some more fine-tuning in the future.

EmperorPinguin:
@tuvus I took your comments into consideration, is it good like this? (this won't be the last PR in the series, but I need to more real-game playtesting for new additions)

tuvus:
If we leave out the `science * 1.5` then I will probably approve.

EmperorPinguin:
> If we leave out the `science * 1.5` then I will probably approve.

I removed it, is it better this way?

tuvus (approved):
Having gold be valued the lowest, even below faith probably isn't the best, but we can change that later if that is the case.

Otherwise I think this can go through now.

EmperorPinguin:
> Having gold be valued the lowest, even below faith probably isn't the best, but we can change that later if that is the case.

I prefer a point of faith over a point of excess gold, and a point of culture over a point of faith. As I said before, the conversion of gold to production is 2:1 via unit upgrades, while the conversion of faith to production is 2:1 via holy warriors. Holy warriors isn't the best faith faith sink, via e.g. religious buildings the conversion is closer to 1:1 (depending on how you do the math, but with an upfront investment into spreading the religion). I tend to prefer a point of culture to a point of faith, but the exact conversion depends on your policies, number of cities, and wether you're growing into good tiles.

Science is harder to rate, as it's a non-substitutable yield, and over half the value of a scientist slot is in the scientist point generation. There are three ways in which to assign citizens, which are 1) regular tiles (farms, mines, pastures, unimproved tiles), 2) specialists (most notably scientists, but other types can be strong as well), 3) trade posts. If we make the value of scientist points implicit in the value of scientist slots (as is currently the case), we need to set it rather high to make the AI prioritise scientists over engineer slots and the lower-yield one of the regular tiles. Doing so, however, will result in an over-evaluation of trade posts, which is bad. According to my sheets calculations, a value of 9 should just be enough to to strike the balance, but it isn't the ideal approach imo. 

Note that faith can be converted into science at a rate of approximately 1:10 by purchasing great scientists in late game, and culture is converted into science via the Rationalism tree. The value of gold is amplified by science advantage, as upgrade timings allow you to more effectively convert the advantage into a win. We could set the value of gold, faith and culture according to that of science, but for practical purposes of tile assignment, it's better to set it accoridng to the value of production.

EmperorPinguin:
The best way to fgain gold is to grow pop, connect trade routes and to build markets, banks, etc. not to keep stuck working merchants slots and plantations, which hurts the development of our empire. I think it's best to value yields differently for 1) citizens specialist assignement, 2) citizen non-specialist assignement, and 3) building construction assignement. Each one will also have to be weighed according to stage of the game for optimal performance, which is maybe a bit complicated due to mod-compatibily and which victory types are enabled.

Issue #11903, opened 2024-07-02
Bangla Translations is broken in master!


touhidurrr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.12.7

### Describe the bug

Something happened in 9c864c3. Pretty sure Bangla worked on my `bangla-translations` branch and right before #5826 was merged at its latest state. But it is broken in master? But why? Not sure.

### Steps to Reproduce

Try to build and run 9c864c3, and select Bangla.

### Screenshots

![image](https://github.com/yairm210/Unciv/assets/46617994/9ae23114-3c04-4a98-a053-3cbc059048cf)

### Operating System

Windows

### Additional Information

#11898 works.

touhidurrr:
To be sure, I revived [bangla-translations](https://github.com/touhidurrr/Unciv/tree/bangla-translations) branch, and yes, it is working:
![image](https://github.com/yairm210/Unciv/assets/46617994/27d6c3fa-0e2a-47b4-8f84-7cc7f5ab62b0)

yairm210:
This actually changed in fb473590
This means that Options -> Advanced -> Generate translation files did this
This could be a problem with the encoding we write the translation files in
In TranslationFileWriter line 236 it says "write the new files as UTF-8"
Does Bangla not work with UTF-8?

yairm210:
No, [I see it should](https://en.wikipedia.org/wiki/Bengali_(Unicode_block))

And we read is using that same charset
Not sure what's going on there, maybe the stringbuilder...?

yairm210:
Anyway I would put a breakpoint in translationfilewriter and try to dig into where this change begins

yairm210:
Hilariously "copy file" doesn't copy the bangla - not sure what encoding *the clipboard* uses, and damn am I happy that I have never had to think about that before, what a mess
Okay so that didn't work, the read-to-write flow works but the entire translation mechanism fails
It's midnight here, I'm going to bed

yairm210:
NOPE reading UTF-8 files as UTF-16 did NOT work, that's almost certainly not the way
Not sure what *is* the problem
Debug TranslationFileWriter to find out, I guess

touhidurrr:
> It's midnight here, I'm going to bed

It's 3:00AM here actually. But anyways, good night, see you later!

touhidurrr:
[...]

PR #11913, opened 2024-07-04
Fix TranslationFileWriter changing diacritic-using languages

SomeTroglodyte:
Should fix the cause of #11903 - doesn't replace the bungled bangla file itself.
Should also fix the resulting fastlane descriptions after the proper file is back in - though I suspect irrelevant as F-Droid itself hasn' learned /bn/ yet???

~Untested so far except yes it passes unit tests locally~
- chose not to reinstate the Bangla roundtrip test, we could rewrite it as standalone on the same basis as testNonBasePlaneUnicode?

### DIFF
--- a/core/src/com/unciv/models/translations/Translations.kt
+++ b/core/src/com/unciv/models/translations/Translations.kt
@@ -173,7 +173,15 @@ class Translations : LinkedHashMap<String, TranslationEntry>() {
         return languages.filter { Gdx.files.internal("jsons/translations/$it.properties").exists() }
     }
 
-    /** Ensure _all_ languages are loaded, used by [TranslationFileWriter] and `TranslationTests` */
+    /** Ensure _all_ languages are loaded, used by [TranslationFileWriter] and `TranslationTests` only.
+     *
+     *  #### Notes:
+     *  -  Expects to run on a newly created instance.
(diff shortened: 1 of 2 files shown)


yairm210 (approved):
If "generate transalations" isn't messing up Bangla for you locally, that's good enough for me

core/src/com/unciv/models/translations/Translations.kt
```diff
@@ -182,7 +183,7 @@ class Translations : LinkedHashMap<String, TranslationEntry>() {
 
         DiacriticSupport.reset()
         for (language in getLanguagesWithTranslationFile()) {
-            tryReadTranslationForLanguage(language)
+            tryReadTranslationForLanguage(language, noDiacritics = true)
```
yairm210, line 186: Please add a comment here
"We load the translations with no diacritics for TranslationFileWriter, so what we read will be what we write"

Also, are `DiacriticSupport.reset()` and `DiacriticSupport.freeTranslationData()` lines, above and below, still necessary?

touhidurrr:
Hi! Was curious if this fixes Bangla Translations. As I cannot see any commit reverting the translation files...

touhidurrr:
Oh, ok.. It does! Thanks guys!! Looks good with numbers translation also!!!
image

yairm210:
The translation files I reverted with the version release

PR #11915, opened 2024-07-04
Windows ARM64 Builds

### LINKED ISSUE #11909: Feature request: Unciv Windows ARM64 / Aarch64 builds
### Before creating

### Problem Description

It would be nice if we had this. Kind of an early support?

### Related Issue Links

_No response_

### Desired Solution

`.exe` aside, we need correct jre. Some candidates are:
Temurin 21 from here: https://adoptium.net/temurin/nightly/?version=21
Microsoft OpenJDK 11: https://github.com/microsoft/openjdk-aarch64/releases/tag/jdk-11.0.12-ga

I would suggest Temurin 21 (cause 11 does not have arm builds). Microsoft one is quite old. I personally always play Unciv with my locally installed Temurin 21 (now GraalVM 22) without any issues. So, there should not be any problems with using Temurin 21.

If packr does not support it yet, we need to write a `.bat` file to run it maybe? `.exe` that we currently ship is the same right? Users would need to click on it anyways.

### Alternative Approaches

Not Sure.

### Additional Context

1. Recent hype regarding Snapdragon ARM chips on Windows.
2. Not sure if packr supports ARM yet.

touhidurrr:
resolves #11909

- [x] Test that the script works in your system with appropriate download link
- [x] The produced zip can be extracted using Windows "Extract All..." option
- [x] Clicking `Unciv.bat` runs Unciv properly

### DIFF
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -228,7 +228,7 @@ jobs:
            start /B ./jre/bin/javaw.exe -jar Unciv.jar" > Unciv.bat
            chmod a+x Unciv.bat
 
-           # Make .zip
+           # Make .zip, -9 is best compression, -r is recursive (needed for folders)
            zip -9 -r Unciv.zip jre Unciv.jar Unciv.bat
 
            mv Unciv.zip ../deploy/Unciv-WindowsARM64.zip
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -209,13 +209,38 @@ jobs:
           wget -q -O jdk-windows-64.zip https://api.adoptium.net/v3/binary/latest/11/ga/windows/x64/jre/hotspot/normal/eclipse
           ./gradlew desktop:packrWindows64
 
+      - name: Manually pack Unciv-WindowsARM64.zip 
+        run: |
+           # Set working directory to windows-arm64
+           mkdir windows-arm64 && cd windows-arm64
+
+           # copy Unciv.jar here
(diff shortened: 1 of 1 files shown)


touhidurrr:
Possible way to test this:
- [ ] Test that the script works in your system with appropriate download link

touhidurrr:
Running the script with WSL produced this output:
image

After opening Unciv.zip with 7z, can confirm that the contents are correct:
image

The produced zip could be extracted from Windows built in "Extract All..." option. And double clicking the `.bat` file **DOES** run Unciv properly.

touhidurrr:
Now, the remaining test would be to see if this correctly add `Unciv-WindowsARM64.zip` in releases which I am not sure how can I check that. So, this PR can be considered ok to go.

yairm210:
Is there a reason not to use packr? This looks exactly like what packr does, we could just change the JDK in between x86 and ARM builds

yairm210:
If we see that doesn't work we can try this but it seems unnecessary if the current path will suffice
Also the final download will be smaller, since packr throws away unnecessary parts of the JRE

touhidurrr:
> Is there a reason not to use packr? This looks exactly like what packr does, we could just change the JDK in between x86 and ARM builds

Hmm... They officially does not support ARM yet... But ARM devices can emulate x64 code so this might work... But again not sure how to test this, also not sure how to configure packr...

yairm210:
*theoretically* just replace the existing "windows 64 bit" JDK in the CI flow with your JDK and it'll use that one
The real question is: How is the .exe file constructed? If it's x86 specific then this will not work, if it's generic (like your batch file) then it will
Do you actually have a Windows ARM where you can test this?

PR #11918, opened 2024-07-05
Update flag for Latin so it's same size as the others

SomeTroglodyte:
Smallest thing to test my system recovery I could think of


touhidurrr:
by the way, what is your system recovering from... what happened...

PR #11921, opened 2024-07-06
UX: Notifications for map units select better when tapped

SomeTroglodyte:
See commit titles. Some of this has bugged me a long time - but very slightly so.

Usage examples:
- The classic "Nothing to do" notifications - tap and you can move the relevant unit directly.
- "City X can bombard" - tap the notification, tap the victim, boom. Before: Tap, tap city...
- Testable from console: `civ activatetrigger "[3] free [Bomber] units appear"` - add different promotions to them - tap notification repeatedly

Implementation notes:
- With `UnitsGainPromotion` I was a bit unsure. There was no "no need to promote units that already have the promotion" filter, so I added it - but late, within the returned lambda. The other filters are early. If the "already has it" filter can go early too, no need to distinguish unitsToPromote and promotedUnits. Also - here I noticed addPromotion does not follow the MutableCollection.add convention to return a Boolean indicating some change was done...
- The source sweep to discover opportunities is potentially incomplete - I could have missed Notifications that could use a mix of MapUnitAction and LocationActions but currently have only LocationActions.
- worldScreen.gameInfo.tileMap[location] can theoretically crash. setCenterPosition on the other hand can't even when the coordinates have no corresponding tile. Didn't really want to open that little can of worms (setCenterPosition does a "full table scan" instead of using existing indexes, when _exactly_ to check whether position is valid or not, decide whether to use the WW-aware function or make the faster one public,...)

### DIFF
--- a/core/src/com/unciv/logic/civilization/NotificationActions.kt
+++ b/core/src/com/unciv/logic/civilization/NotificationActions.kt
@@ -3,6 +3,7 @@ package com.unciv.logic.civilization
 import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.utils.Json
 import com.badlogic.gdx.utils.JsonValue
+import com.unciv.Constants
 import com.unciv.logic.IsPartOfGameInfoSerialization
 import com.unciv.logic.city.City
 import com.unciv.logic.map.mapunit.MapUnit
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/logic/civilization/NotificationActions.kt
```diff
@@ -108,24 +109,29 @@ class MayaLongCountAction : NotificationAction {
...
+class MapUnitAction(
+    private val location: Vector2 = Vector2.Zero,
+    private val id: Int = Constants.NO_ID
+) : NotificationAction {
+    constructor(unit: MapUnit) : this(unit.currentTile.position, unit.id)
     override fun execute(worldScreen: WorldScreen) {
-        worldScreen.mapHolder.setCenterPosition(location, selectUnit = true)
+        val selectUnit = id == Constants.NO_ID  // This is the unspecific "select any unit on that tile", specific works without this being on
+        val unit = if (selectUnit) null else
+            worldScreen.gameInfo.tileMap[location].getUnits().firstOrNull { it.id == id }
```
yairm210, line 128:
```suggestion
        val selectAnyUnit = id == Constants.NO_ID  // This is the unspecific "select any unit on that tile", specific works without this being on
        val unit = if (selectAnyUnit) null else
            worldScreen.gameInfo.tileMap[location].getUnits().firstOrNull { it.id == id }
```

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -560,16 +559,17 @@ object UniqueTriggerActivation {
                 if (unitsToPromote.isEmpty()) return null
 
                 return {
-                    val promotedUnitLocations: MutableList<Vector2> = mutableListOf()
+                    val promotedUnits: MutableList<MapUnit> = mutableListOf()
                     for (civUnit in unitsToPromote) {
+                        if (promotionName in civUnit.promotions.promotions) continue
                         civUnit.promotions.addPromotion(promotionName, isFree = true)
```
yairm210, line 565: Maybe addPromotion should return a boolean?
this is fine, just thinking aloud 🤔

yairm210 (approved):
Looks good!

PR #11940, opened 2024-07-10
Ninjum patch 2

ninjum:
update galician


ninjum:
line error corrected.

PR #11943, opened 2024-07-11
More Numbers Translations Coverage

### LINKED ISSUE #11919: Number Translations Coverage
### Before creating

### Problem Description

So, with #11898, we got numbers that gets translated according to language choice of the user. We also got `Number.tr()` that can help us to translate numbers in places where `String.tr()` is not called. But this method is not used in its full glory and we get untranslated numbers in many places of the game.

This issue hopes to address that problem.

### Related Issue Links

#11895 #11898 

### Desired Solution

Use `Number.tr()` for numbers outside of `String.tr()` coverage.

### Alternative Approaches

Unknown.

### Additional Context

Not sure where to start. Need help.

touhidurrr:
resolves #11919

> [!IMPORTANT]
> Might contain unnecessary changes. Review carefully.


core/src/com/unciv/logic/simulation/MutableInt.kt
```diff
@@ -1,11 +1,13 @@
...
+import com.unciv.models.translations.tr
+
 class MutableInt(var value: Int = 0) {
     fun inc() { ++value }
     fun get(): Int { return value }
     fun set(newValue: Int) { value = newValue }
 
     override fun toString(): String {
-        return value.toString()
+        return value.tr()
```
yairm210, line 11: This is only used in simulation where we don't use translations, I think this is unnecessary
It doesn't hurt though, so not a blocker

yairm210 (approved):

PR #11953, opened 2024-07-13
Fix untranslated string

Ouaz:
`Never close popups by Clicking outside` in template.properties modified to `Never close popups by clicking outside`

Before:
before

After:
after


tuvus (approved):

yairm210:
This is the kind of change that should be done on all translation files, so we don't lose the current translations

PR #11955, opened 2024-07-13
EditorConfig now uses crlf instead of lf like the rest of the project

tuvus:
I finally found what caused the unique.md and some other .md file to always show in my changes! Intellij was set to use crlf line endings, however it was still changing it to lf. This is because the .editorconfig overrides the Intellij setting. 

Since our project uses crlf we should change the .editorconfig to match it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,7 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
+end_of_line = crlf
 indent_style = space
 indent_size = 4
 insert_final_newline = true


tuvus:
Are your local files using LF? Mine are CLRF. Maybe they are locally being changed to CLRF? But then wouldn't it show in the diff?

I don't have a preference for either one, just that the conflicts shows some files to be staged and committed that I have to filter out.

tuvus:
I think those settings don't do anything if the .editorconfig is enabled. But my question is, why does the end_of_line parameter set to lf only affect the .md files?

I'll probably close this PR soon since it seems like a me problem.

Issue #11969, opened 2024-07-16
AI: Balancing of accepting "unfavorable" trades (reason in this example related to gifted gold)


CrsiX:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.12.8 Build 1019 on Android

### Describe the bug

Even though it is vastly stronger in military power, economic strength and population size and therefore not reliant on anyone else in the game, even though it additionally has a few city states as its allies, the AI Arabia accepts an extremely unfavorable trade with player Germany:

Germany (player) gives:
- Defensive Pact
- Declare War against Polynesia

Arabia (AI) gives:
- Defensive Pact
- Cotton 1
- Gems 1
- Incense 1
- Spices 1
- Wine 1
- Citrus 1
- Copper 1
- Khurasan (city! 10 inhabitants)
- Medina (city! 16 inhabitants)

See the screenshots or the saved game file for how big the difference between the two is. Note that Polynesia (which Arabia just declared its war against), is extremely weak. Also note that the fourth major civ in the game was annihilated by Arabia less than 10 turns ago. As a player, the only sane option is preparing for the worst. But when Arabia offered the player to join its war, the player asked "why should I?" and maxed out the demands as seen in the save game.

**TL;DR:**

Balancing is very hard. But it goes against all reason that Arabia accepts this trade offer. This should offer an example of a situation, which should _not_ be accepted by Arabia.

**Suspicion**

I have some thoughts about it, see comment below.

### Steps to Reproduce

1. Explore the map and the victory status -- see that Arabia is overwhelmingly strong
2. See the trade that Germany (player) proposed to Arabia (AI)
3. Click next turn
4. Arabia accepted the trade (and someone declared war on the player, but that's not the point)

### Screenshots

<img width="1470" alt="Screenshot1" src="https://github.com/user-attachments/assets/641c285a-1e98-45d3-8e46-4e34b6da5242">
<img width="1470" alt="Screenshot2" src="https://github.com/user-attachments/assets/a8e55680-5161-4399-a30d-036ba61a5025">

### Link to save file

600kB JSON:
https://gist.github.com/CrsiX/13ac755f8ff0f8420b993b8ab9b316c6

### Operating System

Android

### Additional Information

Since I was debugging it on Desktop, it is also reproducible on latest master branch. My thoughts on this issue are in comments below.

CrsiX:
[...]

PR #11972, opened 2024-07-17
Gift gold fix

tuvus:
This PR fixes the bug reported in #11969 and other discord messages. The situation was that some trade was resulting in a very high gift value. The users then reported it after using the gold gift for an unbalanced trade. Debugging the problem was hard, since the logical bug occurred turns earlier than the save given. While testing I used a simple testing game so the force value of the civilizations was very low, making it impossible to see the bug.

The bug: I was missing a pair of parenthesis on `motivation += 2 * thirdCivForce / targetForce.toFloat().coerceAtMost(40f)`. So the target force had a max value of 40 while the thirdCivForce can be in the hundreds of thousands.
So I fixed it, and added a few other minor tweaks.


tuvus:
That error isn't mine? `worldscreen/worldmap/OverlayButtonData.kt:32:5`

yairm210 (approved):

alexban01:
Sure I'll have a look one of these days

PR #11993, opened 2024-07-19
Repair Qingmin holiday

SomeTroglodyte:
Sorry I didn't test the previous code enough - I ran it on Android, but forgot to take out the -DeasterEgg= line, so the actual date stuff never ran.

Question: Looking at the code, mapping and storing these tables in a static seems wasteful, especially as I don't see how they could be used more than once. Went in there with the preconception of how I wanted compiler-preinitialized tables, and also what was easiest to "translate" via regex from sources. It would be more efficient to change the lazies into getter methods, and maybe even use json deserialization to create them more directly.... You think it worth it to refactor?


yairm210:
Anything weighing a few kbs of memory is not even worth discussing about.
Not a problem.

PR #11995, opened 2024-07-19
Made Gold Gifting moddable

tuvus:
This PR expands on #11326 by making the gold gifting moddable. Modders are able to completely disable gold gifting, or they can decide to make it OP.

There are two types of gold gifts that can be given: 

- "Pure" gold gifts, where the giver receives nothing in return. I put this multiplier at full value by default.
- Favorable trades, where one side of the trade has more value than the other side. I put this multiplier at 0.8 or 80% by default.

The rate at which the gifts degrade at is also moddable. I also slightly increased the friendly degradation rate.

Any input on what the default values should be would be appreciated.

### DIFF
--- a/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
+++ b/core/src/com/unciv/logic/civilization/diplomacy/DiplomacyManager.kt
@@ -18,6 +18,7 @@ import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.ui.components.extensions.toPercent
 import kotlin.math.ceil
 import kotlin.math.max
+import kotlin.math.roundToInt
 import kotlin.math.sign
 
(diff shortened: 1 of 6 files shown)


yairm210:
This *really* feels like something that should be a mod option Unique and not a parameter.

Uniques are deprecate-able, condition-able, validateable, and many other "-able"s

tuvus:
Making it more deprecation friendly does seem important here. I guess if we have duplicate uniques we could just multiply them together. How easy is it for a modder to remove a unique from another mod/vanilla?

yairm210:
That's a really good point - how can we structure these so that mods can override them?
I'll need to meditate on this, this sounds like a general change for mod options to me 🤔

yairm210:
You know what, this is better than current state in any case
We can change it later

yairm210 (approved):

Issue #10830, opened 2023-12-29
AI plays always on Chieftan difficulty


tuvus:
### Game Version

since forever to 4.9.13

### Describe the bug

There are a few comments looking through the code talking about how the AI is never unhappy. They found new cities very early, which is impossible for the player to do. And when using the new AutoPlay feature to play as a Civ against the AIs, I don't have the same benefits they do. Even on Prince difficulty, the AI swamps the AutoPlay Civ even though they supposedly have only an 85% maintenance modifier difference.

After some investigation, I found that in Civilization.kt lines 303 to 309 tell quite the story.
```
    fun getDifficulty(): Difficulty {
        if (isHuman()) return gameInfo.getDifficulty()
        // TODO We should be able to mark a difficulty as 'default AI difficulty' somehow
        val chieftainDifficulty = gameInfo.ruleset.difficulties["Chieftain"]
        if (chieftainDifficulty != null) return chieftainDifficulty
        return gameInfo.ruleset.difficulties.values.first()
    }
```
It turns out that the AI has been playing against us with the Cheftain difficulty all along! And if it can't find the chieftain difficulty, then it finds an even easier difficulty.

Unless there is any other background that I am missing, it seems like we should change this to be more in line with the expectations given to the players. Any thoughts?

### Steps to Reproduce

1. Start a new default game on prince difficulty
2. Use AutoPlay until your civ dies

SeventhM:
This is a direct copy of [how it works in Civ 5](https://civilization.fandom.com/wiki/Difficulty_level_(Civ5)). If there's actually a want to fix it, in theory it likely should be allowing a mod to specify what difficulty the AI play on by default. But it playing on chieftain is in fact, intentional

tuvus:
Okay, it's a little weird, though, how the players can't copy or learn from the AI at the beginning since they are not equal. I know that making a good AI is very hard and will never be as good as a good player. Can we at least state this clearly in the Civilopedia somewhere?

yairm210:
Sure, I don't think it's that interesting to the player what logic the AI is using to play with, though. I only know this because I needed to know in order to program it, it never affected me as a player.

yairm210:
In general if your strategy is 'learn from the AI' then it really doesn't matter what difficulty it's on, since it doesn't *really* change its behavior between difficulties. But that's for another time...

SeventhM:
> anyone care to test?

[...]

PR #11996, opened 2024-07-20
Make AI difficulty changable

tuvus:
This PR addresses #10830 by adding the AI difficulty level to the difficulty object. `aiDifficultyLevel` is an optional string that will search the set of difficulty levels for the AI to play on. If none was supplied then the usual decisions apply, choose the "chieftain" option or if it doesn't exist choose the first (likely easiest) difficulty.

The Difficulty Civilopedia tab also wasn't showing so I created a fix for it. I did a quick comparison of the objects shown in the civilopedia, and found that it only impacted the entries of the Difficulty page.

<details><summary>Comparison</summary>
Only the difficulty objects were changed.
Note: this comparison was only done on the base, G&K version

DifficultyComparison

</details> 

Apart from the civilopedia, there shouldn't be any change to gameplay, only an option for modders.


itanasi:
Another request related to this was being able to change this mid-game. How hard would that option be in Console?

yairm210 (approved):

yairm210:
Do we have a difficulties.json schema? If not we should probably make one in a different pr
If we do, please add this field in this pr

PR #11997, opened 2024-07-20
Civilopedia Updates: Trade Routes and Air Combat

itanasi:
Add entries detailing Trade Routes and Air Combat. Two common questions.

### DIFF
--- a/android/assets/jsons/Tutorials.json
+++ b/android/assets/jsons/Tutorials.json
@@ -416,14 +416,14 @@
   },
   {
     "name": "Air Combat and Interception",
-    "steps": [
-        "Air Combat is very similar to normal Combat, with a few wrinkles. Damage is still dealt using the same equations based on the relative Strength of the Attacking and Defending Unit.",
-        "However, before the Attacker and Defender fight, there is an added combat phase called Interception. If there is any unit with Interception Range that covers the target attacked tile, then it will attempt to Intercept the Attacking Air Unit. If multiple units can Intercept, the one with higher Intercept chance does it.",
-        "Generally Land/Naval Units have an Intercept Range of 2 Tiles, and Air Units have an Intercept Range of their Attack Range.",
-        "The Intercepting Unit rolls to see if it hits. Whether it hits or not costs an Attack, and most units (without Promotions) can only Attack Once per Nation's Turn. It does NOT cost Movement Points, but Air Units that have used their Movement cannot Intercept between turns.",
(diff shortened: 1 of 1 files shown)


android/assets/jsons/Tutorials.json
```diff
@@ -66,10 +66,26 @@
...
   },
+  {
+      "name": "Trade Route",
+      "civilopediaText": [
+          {"text":"Connecting your cities to the capital by Roads and/or Railroads will generate Gold via the Trade Route. You earn more Gold the more Population the City and your Capital has."},
+          {"text":"Note that each road costs 1 Gold Maintenance per turn, and each Railroad costs 2 Gold, so it may be more economical to wait until the cities grow!"},
+          {"text":"Railroads connecting a city to the Capital also provide a 25% Production Boost to the City, but not the Capital."},
+          {"text":"Trade Routes are also connected over Water via certain buildings like the Harbor. As long as there is a chain that can reach the Capital, there's a Trade Route. Look for this symbol in Yellow below the City Name to mean you have an active Trade Route to your Capital."},
+          {"extraImage":"CityStateIcons/Mercantile","imageSize": 20},
+          {"text": "Trade Routes can get broken either by Blockades or by Removing/Pillaging the connecting Road/Railroad chain."}
```
tuvus, line 86: Could we add here that Military naval units block adjacent tiles as well? Last time I checked this was a thing.

android/assets/jsons/Tutorials.json
```diff
@@ -66,10 +66,26 @@
...
+        {"text":"This greatly improves the speed you can get Units around the map."},
+        {"text":"Until you research the technology that removes it, Roads do not cross Rivers"},
+        {"text":"Connecting your cities to the capital by Roads and/or Railroads will generate Gold via the Trade Route.","link":"Tutorials/Trade Route"},
+        {"text":"However, since each Road and Railroad have a Maintenance Cost, it may be more economical to wait until the cities grow and only place where needed."},
     ]
   },
+  {
+      "name": "Trade Route",
+      "civilopediaText": [
+          {"text":"Connecting your cities to the capital by Roads and/or Railroads will generate Gold via the Trade Route. You earn more Gold the more Population the City and your Capital has."},
```
tuvus, line 81: Could we add something like "Cities that are connected will show a small {roadSymbol} underneath them"?
That would help explain what the symbol means and how to tell if a city is connected.

android/assets/jsons/Tutorials.json
itanasi: There's a separate entry about Naval Blockades. Could link it

android/assets/jsons/Tutorials.json
itanasi: See line 84-85. I couldn't find the exact image in yellow and block border in the code. I think we might make it on the fly.

PR #12008, opened 2024-07-23
Add tiles as a countable value

PLynx01:
After somewhat long time of inactivity, I decided to come back with something new. It's the possibility to use tiles as a countable.

It's very likely that it contains bugs, so please point them out and explain me. That will enable me to avoid them in my future contributions.

If you have some remarks, please share it concisely, whenever possible.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Countables.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Countables.kt
@@ -33,6 +33,9 @@ object Countables {
             return civInfo.cities.sumOf { it.cityConstructions.getBuiltBuildings()
                 .count { it.matchesFilter(placeholderParameters[0]) } }
 
+        if (countable.equalsPlaceholderText("[] Tiles"))
+            return gameInfo.tileMap.tileList.count { it.matchesFilter(placeholderParameters[0]) }
+
         if (gameInfo.ruleset.tileResources.containsKey(countable))
             return stateForConditionals.getResourceAmount(countable)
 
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/models/ruleset/unique/Countables.kt
```diff
@@ -33,6 +33,9 @@ object Countables {
             return civInfo.cities.sumOf { it.cityConstructions.getBuiltBuildings()
                 .count { it.matchesFilter(placeholderParameters[0]) } }
 
+        if (countable.equalsPlaceholderText("[] Tiles"))
+            return gameInfo.tileMap.tileList.count { it.matchesFilter(placeholderParameters[0]) }
```
yairm210, line 37: Tiles in the *entire map*?! That's both very unintuitive, and I can't really see a use case for this
All the other filters are for civ-wide counts, not map-wide

PLynx01: So it should be only civ-wide?

What should we use instead for obtaining map area, height, width or amount of land and water tiles?

yairm210: Who even wants that?

PLynx01: Please explain me your advices in more straightforward and unambiguous way.

So I understand that it tile-counting should be civ-wide, not map-wide. Am I right?

PR #12033, opened 2024-07-31
AI great person purchasing [draft]

EmperorPinguin:


EmperorPinguin:
I added great person purchasing, but I'm not sure if the code is very optimal, as it checks for all cities while it should be sufficient to just check for capital.

PR #12035, opened 2024-07-31
Force calculation fix and moddability improvement

### LINKED ISSUE #12001: Force calculations getting massively screwed up in Wakanda mod

### Game Version

4.12.12-patch2

### Describe the bug

Something about the Black Panther unit in my Wakanda for Unciv mod is messing up the force calculator. The unit itself has a base strength of 20, but it scales with era (maximum strength bonus is 400% during the Future era). Building this unit causes the force calculator to give the civilization a strength of over 800,000. The unit is not actually that overwhelmingly strong.

### Steps to Reproduce

1. Enable Wakanda for Unciv mod
2. Build the Sacred Mound, causing the Black Panther unit to appear
3. Wait one turn for the force calculator to catch up to the changes made
4. The force displayed will now be unreasonably large

### Screenshots

Screenshot (212)

### Link to save file

https://github.com/SpacedOutChicken/Wakanda-for-Unciv

### Operating System

Windows

### Additional Information

_No response_

SomeTroglodyte:
Fixes #12001

### Side effects
Triggered Strength Uniques no longer affect force calculation
Negative Strength modifiers now **do** count - I saw no good reason why they were excluded
Strength uniques with "unknown" conditionals and without the new modifier no longer affect power at all

### Weaknesses
- This level of complexity might be overkill
- Affects only BaseUnit force calc, MapUnit force eval still has little concept of Strength increases over the BaseUnit: Policies strengthening all units - nope, actual power of promotions: nope, only count of promotions is taken into account - I assume including "heal once". The modifier could be used in an overhaul, though
- Applicability to Strength Uniques in Promotions granted through BaseUnit.promotions isn't documented, and such promotions aren't included in the new RulesetValidator warning - mostly laziness to get a good wording and trying to keep it shorter


yairm210:
I don't like the fact that it requires micromanaging from the mod makers ('how do I know what's a good number?')
I prefer a best-effort guess that's good enough, rather than modder micromanagement 
There a a few well-known conditionals that we can already assign 'relative values' to - attack/defend 50%, vs type 25%
Unknown conditionals we can assign 25% to I guess? 20%? We can always adjust later

yairm210:
Hmmmmm
I'm okay with this, as long as there's *some* default value - say 25% or something
I do think that we should add 'sane defaults' clauses to the if, so that the cases where mod makers *need* to manually override is diminished

yairm210:
I haven't forgotten about this, I'll try to merge it into the next version
It's "better than current state" for sure

PR #12043, opened 2024-08-01
Civilopedia Entries: Food, Production, Science, and Gold

itanasi:
Add entries for Food, Science, Production, and Gold.
Also a small clarification for Interception
Annoyingly after Translation the entries are ending up at the start and end of Civilopedia. Any idea on how to fix this?
EDIT: has been fixed
image


yairm210:
We can sort the texts after removing the preceding characters for icons, I guess

itanasi:
Ah ha, found where that happened for Faith and Happiness. Fixed now!
image

yairm210 (approved):
I agree SomeTrogs is both more readable and future-proof

PR #12044, opened 2024-08-01
Add Air Intercept Range to Civilopedia card

itanasi:
The value is no longer hidden!
image
image


yairm210:
Not translated
Please add text to template.properties and test

yairm210 (approved):

Issue #11986, opened 2024-07-18
Feature request: option to disable animations


hippie68:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

Please add an option to disable the new unit animations.

### Desired Solution

Create an option in "Graphics" settings to disable animations

### Alternative Approaches

-

PR #12050, opened 2024-08-01
New options control over Animations

SomeTroglodyte:
Solves #11986 - luxury version

### Screwnshots
<details>

Predefined sets SelectBox (the "Expert" entry allows controlling each anim individually, the other set all individual options to a preset):
image

Expert mode:
image
</details>

### Notes
- Wording up for discussion
- Choice of predefined sets up for discussion - more of them? Which options on for them?
- Nuke untested, haven't got an advanced game here
- Yes my understanding and testing say it's perfectly fine to leave `Gdx.graphics.isContinuousRendering` off - or did I miss anything that animates without using Gdx Actions?
- Default on new installs up for discussion
- Migration does not map `continuousRendering==false` into the new world perfectly - some anims weren't gated at all, to get identical behaviour we would need to create a preset with only the ungated anims on - worth it?

### DIFF
--- a/android/assets/jsons/translations/template.properties
+++ b/android/assets/jsons/translations/template.properties
@@ -811,11 +811,11 @@ Show pixel units =
 Show pixel improvements = 
 Unit icon opacity = 
 
-### Performance subgroup
-
-Performance = 
-Continuous rendering = 
-When disabled, saves battery life but certain animations will be suspended = 
+### Animations subgroup
+Animations = 
+Animation level = 
+Some = 
+Expert = 
 
 ## Gameplay tab
 Gameplay = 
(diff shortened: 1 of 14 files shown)


yairm210:
If anything we want it constantly *on*
The continuous rendering means redrawing the screen when nothing has changed, this affects the cpu usage considerably

yairm210:
Right right
Got confused not sure why

AdityaMH:
Use "custom" instead "expert"?

yairm210:
After vacillating over this, I think this is overcomplexity no one needs.
There should be one big 'disable animation' setting.

hippie68:
The pull request works well. I think disabling all animations saves battery life on my laptop.

AdityaMH:
He left

PR #12051, opened 2024-08-01
Clarify which entities are for Portraits

Caballero-Arepa:
Would this work?

Hopefully the autoupdate documentation is good as well.


yairm210:
I already added the autoupdate documentation in a different place

yairm210 (approved):

Issue #11689, opened 2024-06-03
Feature request: Allow natural wonders to not change the terrain.


denismattos:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

Unciv expects all natural wonders in Terrains.json to contain a `turnsInto` attribute
 Since 4.8.15, when the ruleset validator became more stringent, the absence of the attribute or a value throws a red error.

I believe Unciv should allow a natural wonder to not change the terrain where it's placed. 

### Desired Solution

The simplest solution, I believe, would be for Unciv to allow `turnsInto` to accept either a null string or a new string (maybe `nochange`) as an instruction to not change the terrain.

### Alternative Approaches

I'd really like to try to implement this myself (I believe that's the reason open source code is great), if I knew enough JS, but, alas, this is not the case, ATM.
If the devs haven't had the time to implement this when I manage to study enough JS, I may try it myself and push.

### Additional Context

I ran into this issue when updating an old mod.
The original modder didn't provide a value for `turnsInto` in a natural wonder. It may have been a mistake or a choice, but, regardless, he/she most likely didn't even consider this, because the check was just skipped, back then.
I can't test this on current Unciv versions, but, IIRC, this ended up causing graphical errors, so I understand the importance of the ruleset validation.

Nevertheless, I believe this option to not change the terrain should be available.
Sometimes, the kind of terrain where a natural wonder occurs IRL isn't easily classifiable. I'd argue the very fact that `occursOn` allows for a list of strings demonstrates the devs understood this, too.
The wonder's placement in world generation is constrained by the values in `occursOn`, but, otherwise, random. I would argue it's only logical for the game to allow the same randomness to also determine the base terrain.

denismattos:
> If you mean JS = JavaScript then - there's none in here. Not. a. single. line. Which is goooood.

My bad.
I'm new to programming and thought the usage of .json implied JS.
Your comment prompted me to look it up and I learned it's an agnostic format and the game's actually mostly written in Kotlin.

denismattos:
[...]

PR #12062, opened 2024-08-04
Enhance modding freedom for Natural Wonders

SomeTroglodyte:
Stumbled on a forgotten but not quite obsolete branch... Solves #11689
[This comment](https://github.com/yairm210/Unciv/issues/11689#issuecomment-2150334703) still applies - json and screenie.

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
@@ -193,11 +193,10 @@ class NaturalWonderGenerator(val ruleset: Ruleset, val randomness: MapGeneration
                 val state = StateForConditionals(tile = tile)
                 for (unique in conversionUniques) {
                     if (!unique.conditionalsApply(state)) continue
-                    val index = if (unique.type == UniqueType.NaturalWonderConvertNeighborsExcept) {
+                    val convertTo = if (unique.type == UniqueType.NaturalWonderConvertNeighborsExcept) {
                         if (tile.matchesWonderFilter(unique.params[0])) continue
-                        1
-                    } else 0
-                    val convertTo = unique.params[index]
+                        unique.params[1]
+                    } else unique.params[0]
                     if (tile.baseTerrain == convertTo || convertTo in tile.terrainFeatures) continue
                     if (convertTo == Constants.lakes && tile.isCoastalTile()) continue
                     val terrainObject = location.ruleset.terrains[convertTo] ?: continue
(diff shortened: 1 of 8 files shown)


core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
```diff
@@ -180,59 +177,79 @@ class NaturalWonderGenerator(val ruleset: Ruleset, val randomness: MapGeneration
...
+
+            for (tile in location.neighbors) {
+                val state = StateForConditionals(tile = tile)
+                for (unique in conversionUniques) {
+                    if (!unique.conditionalsApply(state)) continue
+                    val index = if (unique.type == UniqueType.NaturalWonderConvertNeighborsExcept) {
+                        if (tile.matchesWonderFilter(unique.params[0])) continue
+                        1
+                    } else 0
+                    val convertTo = unique.params[index]
```
yairm210, line 200: I don't like this at all, put the "unique.params[index]" in the if, it has no meaning outside of it
```suggestion
                    val convertTo = if (unique.type == UniqueType.NaturalWonderConvertNeighborsExcept) {
                        if (tile.matchesWonderFilter(unique.params[0])) continue
                        unique.params[1]
                    } else unique.params[0]
```

core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
```diff
@@ -180,59 +177,78 @@ class NaturalWonderGenerator(val ruleset: Ruleset, val randomness: MapGeneration
...
-                        }
-                        location.setConnectedByRiver(tile, false)
+            val conversionUniques = wonder.getMatchingUniques(UniqueType.NaturalWonderConvertNeighbors, StateForConditionals.IgnoreConditionals) +
+                wonder.getMatchingUniques(UniqueType.NaturalWonderConvertNeighborsExcept, StateForConditionals.IgnoreConditionals)
+            if (conversionUniques.none()) return
+
+            for (tile in location.neighbors) {
+                val state = StateForConditionals(tile = tile)
+                for (unique in conversionUniques) {
+                    if (!unique.conditionalsApply(state)) continue
```
yairm210, line 195: Waitasecond... If we have this, why do we even need the NaturalWonderConvertNeighborsExcept??

core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
```diff
@@ -180,59 +177,78 @@ class NaturalWonderGenerator(val ruleset: Ruleset, val randomness: MapGeneration
...
+                    val convertTo = if (unique.type == UniqueType.NaturalWonderConvertNeighborsExcept) {
+                        if (tile.matchesWonderFilter(unique.params[0])) continue
+                        unique.params[1]
+                    } else unique.params[0]
+                    if (tile.baseTerrain == convertTo || convertTo in tile.terrainFeatures) continue
+                    if (convertTo == Constants.lakes && tile.isCoastalTile()) continue
```
yairm210, line 201: How about "if convertTo terrain object is water and is coastal"?

core/src/com/unciv/logic/map/mapgenerator/NaturalWonderGenerator.kt
yairm210: Yes
These shouldn't be user-facing uniques anyway

yairm210 (approved):

PR #12077, opened 2024-08-06
Use custom Vector2 serialization - reduces total save file size by 4%

yairm210:
In the random file that I decided to test this on, at least
In any case this will be in "read both versions write old version" for like a month at least, and then we'll switch to writing the new version

### DIFF
--- a/core/src/com/unciv/json/UncivJson.kt
+++ b/core/src/com/unciv/json/UncivJson.kt
@@ -2,6 +2,7 @@ package com.unciv.json
 
 import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.files.FileHandle
+import com.badlogic.gdx.math.Vector2
 import com.badlogic.gdx.utils.Json
 import com.badlogic.gdx.utils.JsonWriter
 import com.badlogic.gdx.utils.SerializationException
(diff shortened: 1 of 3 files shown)


yairm210:
@SomeTroglodyte As the resident Json serialization expert, I would like your opinion on this 👍🏿

yairm210:
I see, so all in all a good idea 😉 

I also see that serializing the civ ranking history can lead to a 10% reduction in file size
I'm thinking of testing tile.exploredciv *as a bitset* since the civs are ordered in the gameinfo, this will also be faster but I'm not sure how much that really affects us. I recall updating explored tiles to be a significant % of runtime but I'm not sure *where* it is.

yairm210:
*Another* another idea I have is replacing known strings (civ names buildings terrains) with numbers, with a gameinfo-level "map" of number to string.
But I feel the "user moddability" does decrease with these extra ideas so I'll wait to see if it's actually important, maybe this can be a "try copy to clipboard - if too big, try tricks", so we can get bugged saves from more users (I don't need any history, throw it out!)

yairm210:
If we take "exploredBy":["a","b"...] and replace it with a bitpacked "e":123 we can save 17% (!) of file size
Whereas if we do the renaming trick with base terrains we only gain 5%

Really I can't imagine anyone really cares about the exploredBy enough to edit it manually - *I hope* - so I think it's a worthy trade-off to make it a bitset

But I'll have to see how this affects other things

yairm210:
I'm concerned with "I can't copy the game into the clipboard" right now, disk size doesn't really come into it

PR #12079, opened 2024-08-06
Added support for new historyranking serialization - saved ~10% of save file size

yairm210:


yairm210:
Right, has considered adding but not removing... Good point!!
I'll need to think about it, I don't *think* we'll want to remove them but better safe, I guess

yairm210:
New solution!
We keep the single-char 'ranking type' recognition, and use that as the delimiter!
String looks like `S164N5C0P18G68T14F81H1W11A3`, we can recognize the non-number characters and add a space - for ` S164 N5 C0 P18 G68 T14 F81 H1 W11 A3` - which we split by space to get [S164, N5...] pairs

This means
- We are not beholden to specific order
- We can deprecate old rankings and add new ones easily - unrecognized rankings just get thrown out

yairm210:
Yes that would work, I'm not sure which is more readable...? I'll think about it

PR #12105, opened 2024-08-11
"Skip turn" automation

yairm210:
Many players have asked for a way to "skip a turn" for players whose turn it is, that have been inactive for X amount of time
The first step to that, is allowing for such an automation function. What would I deem unacceptable for an AI to do for me?

For me that is:
- Accept or offer trades (AI doesn't know what I want!)
- Use limited resources (policy picks, gold, faith, etc) when I can decide to use them next turn
- Change diplomacy (declare war / peace)

This PR adds an automation parameter that will enable this

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -32,14 +32,16 @@ import kotlin.random.Random
 object NextTurnAutomation {
 
     /** Top-level AI turn task list */
-    fun automateCivMoves(civInfo: Civilization) {
+    fun automateCivMoves(civInfo: Civilization,
+                         /** set false for 'forced' automation, such as skip turn */
+                         tradeAndChangeState: Boolean = true) {
         if (civInfo.isBarbarian) return BarbarianAutomation(civInfo).automate()
         if (civInfo.isSpectator()) return // When there's a spectator in multiplayer games, it's processed automatically, but shouldn't be able to actually do anything
 
         respondToPopupAlerts(civInfo)
-        TradeAutomation.respondToTradeRequests(civInfo)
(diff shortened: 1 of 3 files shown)


tuvus:
I think that's a good idea! 
I had a similar Idea to handle the problem as well: 

The original idea stems from "lets skip their turn", however with our automation features we could easily play a portion of their turn for them. After X amount of time of a player not playing the next player may be active the player after them may play their turn which will start off by pressing the next turn button which will automate the previous player(s) turn. Then that player may do their turn and push the game save up to pass it to the next player. 
The previous player can play until another player pushes a save game skipping them to the server. If they get their turn in before the payer after them, the player after them needs to pull the save game and restart their turn.
If 2X time has passed then the third player can skip both previous players. up to (N-1)X where any player can play (N is the number of players).
This way we allow the player who we are waiting on the maximum opportunity to play and solve the problem where we don't want the server to do any of the work.

As for the Skip Turn Automation, could we pass around a data class containing boolean options on the various aspects of a civilization that we are find with modifying? The original plan for AutoPlay was to have 7 (+ 1 for espionage) separate options that a player could turn on and off to customize AutoPlaying their turns. Then if we want to, in the future we can store this information along with the player's civilization so that the next device that is going to play for them can understand that player's AutoPlay preferences.

yairm210:
For now I'm limiting this in 2 ways:
- User needs to deliberately request turn skip
- Only allow skipping one user at a time

If we see we need improvements we can add them later

yairm210:
Regarding 'AutomationPreferences', sounds like a good idea - but for a later date ;)
I'm thinking be a class like:
- UnitAutomation: None | MilitaryOnly | CivilianOnly | All
- TradeAutomation: None | AcceptTrades | ProposeTrades | All

etc

yairm210:
The individual sections could maybe be EnumSets, we'll need to start implementing to see what we'll want ;)

tuvus:
And thats completely changeable until we try to persist it in the save file.

PR #12109, opened 2024-08-13
AI changes

EmperorPinguin:
* Adds great person faith-purchasing 
   Todo: better use of purchased great generals and great artists

* Changes policy priorities (the value in personalities.json usually is a number of 0-10, we shouldn't make differences between the values for the preferred victory types greater than 10 as they are added together).

* Non-default (puppet) cities take happiness status better into account for food prioritization

* Removes some conditionals in CityConstruction.json that are imo not necessary, and increases the weight for happiness buildings (Puppet cities often don't contruct colosseums, even when in negative happiness)

* Takes the relative city production into account when deciding on the contruction of spaceship parts, but removes this for worker construction (small low-productive cities benefit more from workers nearby, and producing them in this city saves turns moving them around. Barbs may be an increased problem, although I haven't observed a noticeable difference (they love losing workers regardless)).

* Lowers the cutoff for instant scientist use from 80% of techs researched to 70% of techs researched.


Caballero-Arepa:
> Purchesed Great generals
¿¿¿¿????

To the rest I give a big thumbs up

EmperorPinguin:
> > Purchesed Great generals
> 
> ¿¿¿¿????

Yes, with Autocracy complete.

yairm210:
I'm really not sure about the policy changes. To me this looks like all personalities will make the same decisions, which is definitely suboptimal?

core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
```diff
@@ -30,11 +30,15 @@ object ReligionAutomation {
             buyGreatProphetInAnyCity(civInfo)
             return
         }
+        
+        if (civInfo.religionManager.remainingFoundableReligions() == 0 && civInfo.religionManager.religionState != ReligionState.FoundingReligion) {
```
yairm210, line 34: We should change this to "religionState == Pantheon", because if we *already have* a religion but there are 0 foundable, we still want a new great prophet because that gives us another belief!

core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
```diff
@@ -193,6 +197,30 @@ object ReligionAutomation {
...
+            .filter { it.value.isNotEmpty() }
+            // And from that list determine the cheapest price
+            .minByOrNull { it.value.minOf { city -> it.key.getStatBuyCost(city, Stat.Faith)!!  }}?.key
+            ?: return
+
+        val validCitiesToBuy = civInfo.cities.filter {
+            (greatPersonConstruction.getStatBuyCost(it, Stat.Faith) ?: return@filter false) <= civInfo.religionManager.storedFaith
+                && greatPersonConstruction.isPurchasable(it.cityConstructions)
+                && greatPersonConstruction.canBePurchasedWithStat(it, Stat.Faith)
+        }
```
yairm210, line 216: You should already *have* a list of valid cities from the greatPersonConstruction you did before
Instead of just taking the key, you can take the key (great person) and value (list of cities) and you don't need to check all of this again

The only think missing is the "is less than stored faith" which you can easily add

EmperorPinguin:
> I'm really not sure about the policy changes. To me this looks like all personalities will make the same decisions, which is definitely suboptimal?

Less so than before, as before the differences for the preferrer victory types between policy trees was 10-60 priority points, compared to which the 0-10 points from Personalities.json. Now the differences are 5-10 points, which is more reasonable. You could argue to no penalise Piety by 5 points (I did, as it's quite an irreversible decision).

core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
EmperorPinguin: I tried adding your suggestions, but I'm getting type mismatch errors when I try to clean up the code furthers, I think I'll leave it to more experienced devs to improve.

core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
EmperorPinguin: Are you sure about this? The AI stops buying great persons in certain cases with that change.

EmperorPinguin: Nevermind, I made a typo

EmperorPinguin:
@yairm210, is it acceptable if Crop Yield is renamed to Growth? I think it's clearer to the user what it means (e.g. https://discord.com/channels/586194543280390151/653651863832363053/1270296434700324909).

yairm210:
Sure

EmperorPinguin:
> Sure

Does it also need to be removed from the translation files, or will those be updated automatically?

yairm210:
I'll deal with it

PR #12116, opened 2024-08-15
Update Latin translation

sls1005:
I changed the word for "movement" from *motus* to *momentum* because *motus* is a perfect participle, similiar to "moved" in English. It can mean "motion" in Latin but seems to be far from the idea of "the thing by which it moves," "the ability or potential to move" or "that which is taken (consumed) for it to move." So I picked the word "momentum," which has the suffix "-mentum," which can mean "instrument" or "medium" and finally becomes "-ment" in English.

TLDR:
Movement = Motus (when referrimg to motion or the action of moving) / Momentum (when refering to the ability or potential to move)

Thus, "movement cost" should be something like *momentum sumptum* ("move-ment (tool or something) that is taken (consumed)").

### DIFF
--- a/android/assets/jsons/translations/Latin.properties
+++ b/android/assets/jsons/translations/Latin.properties
@@ -109,7 +109,7 @@ Not displayed as an available construction without [resource] =
  # Requires translation!
 Cannot be hurried = Non potest properari
 
-Choose a free great person = Delige unam personam magnam gratuitam
+Choose a free great person = Delige personam magnam gratuitam
  # Requires translation!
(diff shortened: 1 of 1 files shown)


android/assets/jsons/translations/Latin.properties
```diff
@@ -11310,19 +11318,19 @@ Machine Gun = Polybolum
 
 
  # Requires translation!
-Landship = 
+Landship = Navis Terræ
```
yairm210, line 11321: Nice ;)

PR #12138, opened 2024-08-21
Added the 1st part of lacking polish translation

Lesiakower:


Lesiakower:
The same issue as here #11825

yairm210:
Extra placeholder `[allycivName]` in `Polish` for entry `[civName] and [allyCivName] have declared war against us!`
    Placeholder `[allyCivName]` not found in `Polish` for entry `[civName] and [allyCivName] have declared war against us!`
    Placeholder `[amount]` not found in `Polish` for entry `Air Intercept Range: [amount]`

Lesiakower:
Thanks @yairm210... how could I've possibly missed them (O_O;)

PR #12183, opened 2024-09-05
Improve desktop file for Linux desktop

MayeulC:
This is mostly based on the Flatpak desktop file. Ideally, I would like to use the one from this repository for the Flatpak.

* Tweak keywords and categories
  * Add keywords
  * Add a category
* Remove GenericName as this is not really a generic app
  * This **can be discussed**, but I think GNOME could replace the name of the app with the generic name if a single one is installed: think of operating systems that display "File manager" or "Web browser" in the app list. In any case, this was not really relevant to the user, and may be a better fit as a keyword if you want it.
* Tune the comment to be more relevant to desktop users
  * remove mention of Android & Desktop
* Add French translations
  * Unfortunately, this file may be a bit harder to integrate with a translation service

I put the comment and keywords at the end, as they might grow with additional translations.

### DIFF
--- a/desktop/linuxFilesForJar/unciv.desktop
+++ b/desktop/linuxFilesForJar/unciv.desktop
@@ -1,11 +1,13 @@
 [Desktop Entry]
-Comment=Open-source Android/Desktop remake of Civ V
 Exec=Unciv
-GenericName=4X Game
 Icon=unciv
 Name=Unciv
 NoDisplay=false
 StartupNotify=true
 Terminal=false
 Type=Application
-Categories=Game
+Categories=Game;StrategyGame;
+Comment=Turn-based historical strategy game, open source remake of Civ V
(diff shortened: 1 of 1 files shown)


MayeulC:
@SomeTroglodyte Sure, here is the spec: https://specifications.freedesktop.org/desktop-entry-spec/latest/
More specifically, the keys are listed at https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html

The specification is a bit terse, unfortunately, and could give more details about expected usage for some keys.

Interesting, I'll look into the fastlane workflow, though I am not sure if I will have much time to contribute something.

I wanted to restructure the way the flatpak is published on Flathub. A flatpak could be built from the worflows here, but I don't know if it could be uploaded to Flathub, or if it has to be built on their infrastructure. I suspect the latter.

yairm210:
My only request is that new versions do not require any manual approval. I had to make the workaround for the existing Flathub and it was no fun at all

PR #12189, opened 2024-09-07
Update Nations.json

Smashfanful:
An idea to avoid some redundances


yairm210:
I disagree with this change, it means extra things to translate and it means if we want to change things the entire thing needs to be re-translated, I don't think it actually helps

PR #12192, opened 2024-09-08
Update tutorials

EmperorPinguin:
Rephrases the controversial New Game tutorial, and adds a new one with deviations from Civ5.


EmperorPinguin:
@Caballero-Arepa you were working on the tutorials as well, how do you think of this version?

yairm210:
Looks fine to me
I actually didn't know about those Civ V differences!

Caballero-Arepa:
@Caballero-Arepa

EmperorPinguin:
> "looking for and founding next to luxury resources" - while very well written this entire article ignores that the AI already chose an optimal place for you. The code behind that - at least when not playing a saved map - is quite extensive, so normally you'd want to settle wherever the starting location code placed your first Settler - unless that code is bad? Can we get a very short mention that the game already tries its best for you in most cases in there without disrupting the flow too much?

Agree, this was written by the previous author, I didn't feel like changing it too much (it's mostly the scout vs warrior part that just didn't really make sense). I'll try to rephrase the rest a bit according to your suggestions.

EmperorPinguin:
> If you start listing Civ5-UnC differences - wasn't there a collection of those around [somewhere](https://yairm210.github.io/Unciv/Other/Intentional-departures-from-Civ-V/)? Consider re-mentioning those?

The page needs to be updated, to incorporate the change that melee ships can't capture civilians anymore.

Caballero-Arepa:
> @Caballero-Arepa, you were working on the tutorials as well, what do you think of this version?

Speaking of which, I was working on it at the same time xD

EmperorPinguin:
> This is my proposal for the format.
I'm not entirely sold on the formatting, it should be more consistent in it use of space and newline, it's a bit annoying to read I think.

EmperorPinguin:
>     * ¿Is pointing out that cities have to be 3-tiles away to a new player necessary?

The question will come up in the Discord server when newbies aren't able to settle a city right next to capital, it should be explained better early than later. It's currently a bit too much of a wall of text though, some parts should be removed or shortened.

>     * You can't disable religion in G&K, ¿is the note (when talking about the shrine) necessary?

I don't know, it makes it reuseable for Vanilla I think.

Caballero-Arepa:
Yeah, it could be improved.
image
In my opinion, newline should be used when the paragraph is getting too long.
After some changes:
image
(I know I made the header too small (initially set at 6), it's fixed now (set at 4)).
Removed the \n and replaced them with another "text", so all are consistent. I think those are better than the \n.
My reasoning is that if there is more space, like newlines and indentation, it is easier to read. But I'm still researching about it, because we need it to be easy to read.

The last paragraph should probably get a little header as well, instead of the double space, as it changes the topic (from city placement to build queue).

<p>


	{
	  "name": "New Game",
	  "civilopediaText": [
	  { "text": "Welcome to Unciv!", "header":4},
	  {"text": "Because this is a complex game, there are basic tasks to help familiarize you with the game."},
	  {"text": "   These are completely optional, and you're welcome to explore the game on your own!" },
	  { "separator": true },
	  { "text": "Your first mission is to found your capital city. This is actually an important task because your capital city will probably be your most prosperous."},
	  {"text": "   Many game bonuses apply only to your capital city and it will probably be the center of your empire." },
		  {},
	  { "text": "How do you know a spot is appropriate?","header": 4},
	  {"text": "Luckily, the game has already chosen a good location for you. You can settle in that place, or maybe move a turn to found on a Hill (for Production and defence) or next to a Mountain (to gain access to the Observatory building later on)." },
	  {},
	  { "text": "You'll likely want more cities later on. For these cities, you must go and find a good place - looking for and founding on or near Luxury resources is a good rule of thumb." },
	  { "text": "   Luxury resources are tiles that have things like Gems, Cotton, or Silk (indicated by a yellow background of the resource icon). These resources make your civilization happy. You should also keep an eye out for resources needed to build units, such as Iron."},
	  {"text": "   Cities cannot be built within 3 tiles of existing cities, which is another thing to watch out for!" },
	  {},
	  { "text": "However, cities don’t have a set area that they can work. This means you don’t have to settle cities right on or next to resources."},
	  {"text": "   Let’s say, for example, that you want access to some Iron – but the resource is in a Desert area. You don’t have to settle your city in the Desert. You can settle a few tiles away in more prosperous lands. Your city will grow and eventually gain access to the resource."},
	  {"text": "   You only need to settle right on top of resources if you need them immediately." },
		  {},
		  {},
	  { "text": "The first thing coming out of your city depends on the strategy you want to follow, but the 'classic' build order is to first construct two Scouts, then a Shrine, and three Settlers, and adopting Tradition, to explore the map quickly (Scouts ignore terrain cost), and to get an early Pantheon (if you're playing with religion enabled)." },
	  {"text": "   But feel free to experiment with your own build orders!", "color": "#fa0"}
	   ]
	},

</p>
</details>

EmperorPinguin:
> Yeah, it could be improved. image In my opinion, newline should be used when the paragraph is getting too long. After some changes: image (I know I made the header too small (initially set at 6), it's fixed now (set at 4)). Removed the \n and replaced them with another "text", so all are consistent. I think those are better than the \n. My reasoning is that if there is more space, like newlines and indentation, it is easier to read. But I'm still researching about it, because we need it to be easy to read.
> 
> The last paragraph should probably get a little header as well, instead of the double space, as it changes the topic (from city placement to build queue).
> Details

Newline is cringe, you should avoid indentations, and instead divide the text in small paragraphs, each separated by a space.

Caballero-Arepa:
Let's see
image
image

EmperorPinguin:
@Caballero-Arepa the game also needs a tutorial on the food conversion to production when construction Settler units. It's a question that keeps coming up, and it's apparently not in the civilopedia.

yairm210 (approved):

PR #12234, opened 2024-09-20
Improve AI tech and policy choices

EmperorPinguin:
Adds tech, policy, and promotion priorities for AI.

Bugfix: city prioritization for wonder in ConstructionAutomation.kt now functions correctly

Bugfix: lets AI buy great people with faith again.

Reduces time to space victory to <t185 on my test map.


yairm210 (approved):

EmperorPinguin:
This way of doing the faith building purchasing seems more straightforward (and less bugged? They weren't exactly eager to purchase pagodas in my tests with the old implementation).

core/src/com/unciv/logic/automation/civilization/ReligionAutomation.kt
```diff
@@ -31,13 +31,8 @@ object ReligionAutomation {
...
         
-        if (civInfo.religionManager.remainingFoundableReligions() == 0 && civInfo.religionManager.religionState == ReligionState.Pantheon) {
+        if (civInfo.religionManager.remainingFoundableReligions() == 0 ) {
             buyGreatPerson(civInfo)
-            return
-        }
-
-        // We don't have a religion and no more change of getting it :(
-        if (civInfo.religionManager.religionState <= ReligionState.Pantheon) {
             tryBuyAnyReligiousBuilding(civInfo)
```
yairm210, line 36: So we will only try buying a religious building on the turn when we're buying a great prophet to found our religion?! And even that, before we actually found that religion because we haven't actually activated the great prophet yet? This sounds much worse actually!

PR #12256, opened 2024-10-02
Improve AI city settling, science game, and belief picking

EmperorPinguin:
* Updates tech G&K tech pathing. The AI will prioritise science-friendly units (horseman, gatling gun, infantry, nuclear missile etc.) over science-unfriendly units such as bottom path melee and armor. I think this is fine, but if the players complain about not facing enough AI musketmen, this might need to change.
* Updates city settling. The most significant change here is that it will weigh the quality of tiles according to distance from city center, avoiding those bad in-the-middle-of-flat-desert cities (in-the-middle-of-flat-tundra cities are still a bit of an issue).
* Updates great person picking. The previous implementation was bugged, and always granted random great persons. Instead of implementing the original idea, I decided to always let them pick the science person, exactly as it's the most braindead.

For future reference: the average t80 pop of Deity on my test map is 95.85, across 5 games (20 data points). The most important thing lacking is AI unit movement, to deal with barbs and settling cities faster (I think it could easily be 120+ pop on this map if played correctly). Maybe someone else could work on this.

* Updates belief picking. The AI will no longer choose Holy Warriors, but building-granting beliefs if deemed worthwile. Also re-adjusts the value of some of the other beliefs. (Tile-improvement requiring pantheons need to be supported still, and the AI still incorrectly evaluates Sacred Sites in Rekmod).
* Updates function-calling to let AI buy religious buildings.

Deity Korea on my test map now quite reliably wins t170 science victory, with all the four civs having adopted 3-3.5 full policy trees.


yairm210 (approved):
I trust your judgement on this

EmperorPinguin:
@yairm210, I've also been working on the belief picking by AI. It turned out most of the logic needed for evaluating beliefs was there, but the values assigned to it was very weird. I changed them (and added functioning building-purchasing. You told it wouldn't work correctly, but empirically, it is working, while the previous implementation wasn't).

PR #12267, opened 2024-10-07
Update Citizen Assignment and ConsoleLauncher.kt

EmperorPinguin:
* The lines 158 and 172 in the cityTileRanker cancel each other out, as we're first applying a yield to correct for lighthouse yields, and then subtracting a yield to correct for the fact they can't be improved. This wastes some cpu resources I think, and penalizes Lakes relative to Ocean+Lighthouse, even though they got the same yields.

* Changes  the default citizens assignment to be closer to how I would like the AI to assign my citizens, ~~multiplying the food by 1.5 instead of 2. I've tested this variant before and then it seemed to weaken the AI, but in the few test runs I just did, I didn't notice an effect anymore.~~ Actually that did mess with AI behaviour, just not with Korea. I suspect this is the case for all civs where the food focus in Personality.json is low compared to the production/gold focus. Still, I'd like to remove the conditional so there's no difference between cities anymore (yes, Default now becomes effectively the same as Food focus, not ideal, we should maybe rework the Personality on some of the AIs).
Looking futher into it, it's far weirder than I thought. Tripling (instead of doubling) the food value helps, but that can't be right, as the base value of food is already the highest among the stats. Has the way in which personality or other mechanics interact with citizen assignment changed somewhat recently? 

* Updates ConsoleLauncher. For the following reason: I decided to quantitatively verify the problem above with the new AI vs AI matchmaking. The result was non-significant on my first 160-game run, so I decided to do another 200-game run to add up the results. The results were completely unexpected. I did a couple of small control runs, and it turned out to be a capitalization issue (wasting 3 hours of cpu time in total:/). Updates the description to prevent others from making the same mistake. Also changes the difficulty to Prince (this is what most people play on), and number of city states to 0 (city states are bad for duels like this, especially if you add 6 of them).


yairm210:
Looks good
You should be using Constants.simulationCiv1 and not hardcoded strings for exactly this reason

PR #12274, opened 2024-10-10
Version rollout

yairm210:


EmperorPinguin:
@yairm210 , where's the "AI: Greatly discourage attacking enemies with stronger total combat strength"? I don't see it in the files changed.

EmperorPinguin:
Although... have you considered AlphaFrontier? The mod got units which makes Force jump up by a million or something, and the AI not declaring war at all in late game is known to be somewhat of a problem.

PR #12290, opened 2024-10-14
Update Zulu.properties

graycloud0:


yairm210:
Do not replace untranslated texts with ""! This makes it harder for the next translators to differentiate "missing translations" from "legit translations that should be empty"

yairm210:
That's the reason the tests are failing by the way

PR #12292, opened 2024-10-14

murilozinkkkj:


yairm210:
Conflicts with https://github.com/yairm210/Unciv/pull/12301/commits/c923c633b66ca8f1c385a203c6ce0f4a58837e40 - please discuss which is better going forward :)

PR #12305, opened 2024-10-16
Update AI trading logic

EmperorPinguin:
I don't play sp much to play-test it really, but does this look sensible? I want to prevent those unreasonable counteroffers for trades, in cases where people resign and are taken over by AI in mp games.

Removing line 56-57 in TradeAutomation seems acceptable, as I have not observed the AI starting to ping-pong back and forth as stated. Imo, the AI shouldn't treat other AI differently from humans if not necessary. 

The AI does take WLTKDs into account for luxury trading on lines 136-138 of TradeEvaluation, but not CS resource quests it seems. This should be added as well as they're quite important, but I don't know how to test, so I'd like to request another dev to implement it.


yairm210:
Looks fine to me

yairm210 (approved):

PR #12312, opened 2024-10-18
Allow purchasing wonders with gold when explicitly given a unique

SeventhM:
Simple enough that I probably won't disappear out of nowhere mid PR

### DIFF
--- a/core/src/com/unciv/models/ruleset/IConstruction.kt
+++ b/core/src/com/unciv/models/ruleset/IConstruction.kt
@@ -55,14 +55,16 @@ interface INonPerpetualConstruction : IConstruction, INamed, IHasUniques {
         if (hasUnique(UniqueType.CannotBePurchased, stateForConditionals)) return PurchaseReason.Unpurchasable
         // Can be purchased with [Stat] [cityFilter]
         if (getMatchingUniques(UniqueType.CanBePurchasedWithStat, StateForConditionals.IgnoreConditionals)
-                .any { city == null ||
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/models/ruleset/IConstruction.kt
```diff
@@ -45,19 +45,28 @@ interface INonPerpetualConstruction : IConstruction, INamed, IHasUniques {
...
-        if (hasUnique(UniqueType.CannotBePurchased, stateForConditionals)) return false
+        if (stat == Stat.Production || stat == Stat.Happiness) return PurchaseReason.Invalid
+        if (hasUnique(UniqueType.CannotBePurchased, stateForConditionals)) return PurchaseReason.Unpurchasable
         // Can be purchased with [Stat] [cityFilter]
-        if (city != null && getMatchingUniques(UniqueType.CanBePurchasedWithStat, stateForConditionals)
-            .any { it.params[0] == stat.name && city.matchesFilter(it.params[1]) }
-        ) return true
+        if (getMatchingUniques(UniqueType.CanBePurchasedWithStat, StateForConditionals.IgnoreConditionals)
+                .any { city == null ||
+                    (it.conditionalsApply(stateForConditionals) && it.params[0] == stat.name && city.matchesFilter(it.params[1]))
```
yairm210, line 59: Params matches stat name should be required even if city is null, this applies to both uniques

yairm210 (approved):

PR #12315, opened 2024-10-19
optimized city screen for smart phone

sulai:
- moved "buy" button to city info table
- removed "add to queue" button
- expand icon changed to android defaults
- CityStatsTable: big scrollable area

Goal of this PR is to optimize vertical space for playing on smart phone in landscape mode.


SeventhM:
Can I request a screenshot of how this changes things?

yairm210:
Since this looks to be a big change, before and after pictures are required for us to understand the possible changes

sulai:
Thank you guys for looking into it. I moved the "Buy Button" logic from the CityConstructionTable to the ConstructionInfoTable, while the actual logic stayed pretty much the same. This part looks like a huge change, but actually, it's just moved.

sulai:
Screenshot_20241020-132245
- The stats are collapsible. Sometimes that table takes too much horizontal space, for example when the WLTK text is shown or with certain religions.
- the queue button got removed and the buy button is part of the construction info table (lower right). It saves valuable vertical space for the lists on the left.

Screenshot_20241020-132337
- the stats expanded. Notice that a larger area is scrollable now, because some construction info take a lot of vertical space and previously left only little space to scroll in the stats.

yairm210:
A. If all those functions are logic for the button, then we should probably have a PurchaseConstructionButton that holds all of this logic. If we had that to begin with the move from one place to the other would be trivial
B. I like that the stats are collapsible, that's good in any case 👍🏿 
C. Regarding the actual move of the button... I'm really not sure it's better. That's where the button used to be, and we change it because when I want to buy something that's in the queue, the more comfortable place is "right next to what I clicked".

Which leads me to the following conclusions:
- I'd merge the "collapsible stats" and "separate buy button logic into separate class" no questions asked - it's Just Better ™️ 
- I'll poll the 2 options of "buy button location" on Discord to get a better feeling for what players think is better

sulai:
Sounds good to me! I'll extract the button, let me know about the poll results. 👌

sulai:
I created a BuyButtonFactory, because there is no actual Button class in the original code, but methods creating that button. I didn't see any other factories in the code base, please let me know if we should call it differently. Alternatively, I could also imagine to just undo the changes related to the buy button, should the community decide to keep it where it was.

sulai:
Hey @yairm210 , thank you for putting up the poll, but I have a follow-up PR ready, which would favor the "buy" button to stay where it was.

Screenshot_20241022-093718

Regarding this PR here, the buy button got extracted to it's own factory class and is being shown both under the queue and in the info table. 

Do you prefer my follow-up PR to be merged into this one? Otherwise I'll wait with the second one.

yairm210:
I like the icons shown on the queue, I wonder what that looks like if you have a queue of ten items?
I *dislike* the up and down buttons not being on the queue itself, that means for reorg you double the clicks - choose item + move instead of just move item

In any case sounds like a different PR :)

yairm210 (approved):

PR #12321, opened 2024-10-21
Simplify Civiliopedia Alphabatized Sorting Code

itanasi:
Following up suggestion from #12043 to clean up and future proof the code to avoid sorting based on the prefixed icon in the Civilopedia


yairm210:
Why not use hideIcons? Why in the first place was it not hiding stats icons? I slightly confuse

yairm210 (approved):

itanasi:
Mostly because the code explicitly says hideIcons doesn't cover Stats

yairm210:
Ok
This is better than current state, so perhaps in the future we'll decide to merge them, or not 
For now this is fine

PR #12323, opened 2024-10-21
Add Specialists Tutorial

itanasi:


yairm210:
Looks good, we should change all "city screen" -> "City Screen" in translations because currently this would remove all translations for these lines

yairm210 (approved):

PR #12324, opened 2024-10-21
Creating Guard Action

itanasi:
As requested in Discord, perhaps you don't want your units to Withdraw and ensure they protect a particular tile either to hold the line or protect a Tile Improvement. Well, now you can use the Guard Action.

Testable Save File:
[Autosave-England-15.txt](https://github.com/user-attachments/files/17455416/Autosave-England-15.txt)

Fortification bonus still shows up even if units is Guarding (and they can Fortify). Gave all Musketmen 100% retreat and then set this one to Guard. Fortify Bonus carries over
image
And if they can't Fortify, just indicates they're Guarding
image
And no Fortify bonus in Combat
image


RobLoach:
How does this differ from Fortify?

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsFromUniques.kt
```diff
@@ -165,6 +165,18 @@ object UnitActionsFromUniques {
...
+    internal fun getGuardActions(unit: MapUnit, tile: Tile): Sequence<UnitAction> {
+        if (!unit.hasUnique(UniqueType.WithdrawsBeforeMeleeCombat)) return emptySequence()
+        return sequenceOf(UnitAction(UnitActionType.Guard,
+            useFrequency = 0f,
+            action = {
+                unit.action = UnitActionType.Guard.value
+                unit.useMovementPoints(unit.currentMovement)
+            }.takeIf { unit.hasMovement() })
+        )
+    }
```
yairm210, line 178: I'm not sure about this, since this means a unit can either fortify (to get the defensive bonus) OR guard (to not retreat) but not both
I'm not sure what a good solution would be tbh 🤔

itanasi:
The difference is that not all units can Fortify.

In fact, at least in base game, the only unit with a Withdraw ability are Naval units. And they can't Fortify.

Now, what if they can Fortify? Hm....guess we should disable Guard from showing up. I'll need to think on how we can see that

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsFromUniques.kt
SeventhM: A possible solution is to always apply the fortify bonus to units that didn't move. Afterall, how many situations are you standing in place and don't want the fortify bonus?

yairm210 (approved):

PR #12326, opened 2024-10-21
Add `<when espionage is enabled>` conditional

RobLoach:
Similar to `<when religion is enabled>`, this adds a `<when espionage is enabled>` conditional to allow checking on when espionage is an enabled game option.


RobLoach:
One side-effect I've seen from this change, possibly similar to how `Hidden when religion is disabled` vs `Only available <when religion is enabled>` behaves, is that when Espionage is disabled, things relating to Espionage are now present in the Civilopedia.

Screenshot from 2024-10-21 03-21-05

If Religion is disabled, are things that are marked `Only available <when religion is enabled>` still presented in the Civilopedia?

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
```diff
@@ -142,6 +142,8 @@ object Conditionals {
             UniqueType.ConditionalVictoryDisabled -> checkOnGameInfo { !gameParameters.victoryTypes.contains(conditional.params[0]) }
             UniqueType.ConditionalReligionEnabled -> checkOnGameInfo { isReligionEnabled() }
             UniqueType.ConditionalReligionDisabled -> checkOnGameInfo { !isReligionEnabled() }
+            UniqueType.ConditionalEspionageEnabled -> checkOnGameInfo { isEspionageEnabled() }
+            UniqueType.ConditionalEspionageDisabled -> checkOnGameInfo { !isEspionageEnabled() }
```
yairm210, line 146: Should be added to isUnavailableBySettings list of "disabled by game settings" uniques

yairm210 (approved):

core/src/com/unciv/models/ruleset/unique/Conditionals.kt
yairm210: This is what makes it unavailable in civilopedia btw

PR #12335, opened 2024-10-22
Improve Avoid Growth

itanasi:
Cleans up the Food weights to make more clear/robust.
Avoid Growth now blocks gaining new Population
All Automation will now only x8 the food required to Not Starve. Rest get base weight.
Avoid Growth evaluates all additional Food beyond needed for starvation as 0. Should allow it to prioritize tiles with just enough Food to get positive
Avoid Growth actually evaluates Food for Starvation as important (used to allow Starvation since no x8 in the logic)

### DIFF
--- a/core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
+++ b/core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
@@ -108,12 +108,6 @@ class CityPopulationManager : IsPartOfGameInfoSerialization {
         // We could simply test food==0 - but this way NullifiesStat(food) will still allow growth:
         if (city.getMatchingUniques(UniqueType.NullifiesGrowth).any())
             return
-        
-        // Hard block growth when using Avoid Growth, cap stored food
-        if (city.avoidGrowth) {
-            foodStored = foodNeededToGrow
-            return
-        }
 
         // growth!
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
```diff
@@ -108,6 +108,12 @@ class CityPopulationManager : IsPartOfGameInfoSerialization {
         // We could simply test food==0 - but this way NullifiesStat(food) will still allow growth:
         if (city.getMatchingUniques(UniqueType.NullifiesGrowth).any())
             return
+        
+        // Hard block growth when using Avoid Growth, cap stored food
+        if (city.avoidGrowth) {
+            foodStored = foodNeededToGrow
+            return
+        }
```
SeventhM, line 116: I thought this was discussed elsewhere and decided against. This goes beyond a simple bug fix and changes actual civ 5 gameplay, and thus either should get more discussion or be a unique

itanasi: Removed

yairm210:
What's with all the linting?

itanasi:
Will revert the linting. Somehow must have autolinted more than my changes accidentally

PR #12337, opened 2024-10-22
fix: user only receives yellow resource notification if it is researched

sulai:
For example, don't allow the user to scan the map for aluminum when he got it through trading or allied city states, but is not researched yet. Because the user did not research that resource yet, its deposits on the map should not be revealed.

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/TechManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/TechManager.kt
@@ -17,6 +17,7 @@ import com.unciv.logic.map.tile.RoadStatus
 import com.unciv.models.ruleset.INonPerpetualConstruction
 import com.unciv.models.ruleset.tech.Era
 import com.unciv.models.ruleset.tech.Technology
+import com.unciv.models.ruleset.tile.TileResource
 import com.unciv.models.ruleset.unique.StateForConditionals
 import com.unciv.models.ruleset.unique.UniqueMap
 import com.unciv.models.ruleset.unique.UniqueTriggerActivation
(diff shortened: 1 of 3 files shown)


sulai:
Changed the fix as discussed in Discord. For example, the user can't see the locations of Aluminum if received by trading/city states, when not yet researched.

core/src/com/unciv/models/ruleset/Ruleset.kt
```diff
@@ -510,4 +510,8 @@ class Ruleset {
     }
 
     fun getErrorList(tryFixUnknownUniques: Boolean = false) = RulesetValidator(this).getErrorList(tryFixUnknownUniques)
+
+    fun getResource(resourceName: String): TileResource {
```
yairm210, line 514: Unnecessary and crash-prone - with mods for can never be sure to get a resource from a name

core/src/com/unciv/logic/civilization/managers/TechManager.kt
```diff
@@ -155,6 +156,9 @@ class TechManager : IsPartOfGameInfoSerialization {
 
     fun isResearched(construction: INonPerpetualConstruction): Boolean = construction.requiredTechs().all{ requiredTech -> isResearched(requiredTech) }
 
+    /** resources which need no research count as researched */
+    fun isResearched(resource: TileResource) = resource.revealedBy?.let { isResearched(it) } ?: true
```
yairm210, line 160: This too seems strange, why are you involving this class in the logic of what constitutes a viewable resource? That logic already exists in other places 🤔

yairm210 (changes requested):

core/src/com/unciv/logic/civilization/managers/TechManager.kt
sulai: Could you elaborate on that? If I understand you right, you agree with the method isResearched(resource) itself, but you question its usage in getExploredResourcesNotification(..)?

If the latter is the case, the answer is that I would like to address the problem as close as possible at its root: displaying the yellow notification of a resource which is not yet discovered. This should never succeed, no matter the caller of that method. Since that method is nullable already, it felt like the best place to do that check and not return any notification in case it shouldn't show one.

yairm210: Checking other uses of revealedBy I see that we are indeed missing a single place for this logic, might as well be here
I would change 2 things: 
- 2 liner with :?, no "let", to make the revealedBy==null case more explicit
- change function name to "isRevealed"

PR #12348, opened 2024-10-23
Avoid Growth blocks New Population

itanasi:
Breaks from Civ5 baseline behavior, but would provide more control for players to not get into unhappiness spirals.

Edit: Correction, testing shows this aligns us to Civ5 behavior

### DIFF
--- a/core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
+++ b/core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
@@ -109,6 +109,12 @@ class CityPopulationManager : IsPartOfGameInfoSerialization {
         if (city.getMatchingUniques(UniqueType.NullifiesGrowth).any())
             return
 
+        // Hard block growth when using Avoid Growth, cap stored food
+        if (city.avoidGrowth) {
+            foodStored = foodNeededToGrow
+            return
+        }
+
         // growth!
         foodStored -= foodNeededToGrow
         val percentOfFoodCarriedOver =


EmperorPinguin:
This implementation hard-blocks pop growth it seems, which is only relevant when at -10 happiness, it doesn't adress the Pareto-efficiency issue discussed. When unhappy, we often don't want to prevent population growth, but instead reprioritize our tile yields to take into account the growth penalty.

itanasi:
We already account for the Unhappiness as a huge Food penalty in the tile evaluation.

Players though sometimes want to not gain more Pop and micromanage to get 0Food. This would save them the hassle and still get to work tiles even if they produce a few extra Food than Starvation.

SeventhM:
My stance on this 100% is "this discussion needs to happen on discord, not here". Whether it's through a poll or whatnot. This has potential gameplay applications that are potentially relevant, and it might have various mod applications (both good and bad, intentional and unintentional). There's also no technical limitations that makes this mandatory (similar to me making upgrading units take separate turns for each upgrade instead of skipping, despite a poll), nor is there a seemingly unintentional oversight/bug (past civ 4, no other Civ game lets you mechanically remove growth)

Until such a conversation is had, I would be against. I think it should instead be a unique intended to be in the modOptions

EmperorPinguin:
> We already account for the Unhappiness as a huge Food penalty in the tile evaluation.
> 
> Players though sometimes want to not gain more Pop and micromanage to get 0Food. This would save them the hassle and still get to work tiles even if they produce a few extra Food than Starvation.

Ideally, the Default citizen assignement should distinguish between being at -9 or -1 happiness, and adjust the food penalty accordingly. Then Avoid Growth can do as you propose, in cases where people want cities to not gain pop for whatever reason. I changed my mind on this.

SeventhM:
> Also true, but people tend to report it, as the description ("avoiding growth") doesn't entirely match with the behaviour (valueing growth at 0, but not technically avoiding it).

That's fair, and people can report it as they see fit. But the discussion still needs to be actually be had, not just assumed, just like any other mechanical hard coded change. If this was done through unique, then at least we can just ignore the topic and at least someone will have a mod of this as a separate "option"

EmperorPinguin:
> > Also true, but people tend to report it, as the description ("avoiding growth") doesn't entirely match with the behaviour (valueing growth at 0, but not technically avoiding it).
> 
> That's fair, and people can report it as they see fit. But the discussion still needs to be actually be had, not just assumed, just like any other mechanical hard coded change. If this was done through unique, then at least we can just ignore the topic and at least someone will have a mod of this as a separate "option"

They do report it though:
https://discord.com/channels/586194543280390151/653651863832363053/1297975612753051690
https://discord.com/channels/586194543280390151/653651863832363053/1220971089560207370
https://discord.com/channels/586194543280390151/653651863832363053/1190716665160286309
The function of the button seems somewhat unclear to multiple users:
https://discord.com/channels/586194543280390151/653651863832363053/1168830679245660201
https://discord.com/channels/586194543280390151/586194543716859916/1120879008498847845
Also this comment from Ravignir, what's the behaviour of the button in Civ5 actually?
https://discord.com/channels/586194543280390151/653651863832363053/1190732634884681798

SeventhM:
Only one of those links are an actual report that avoid growth doesn't work as they felt it should, while the rest are requests for clarification. And none of them are an actual conversation, which is what I'm requesting. I'm requesting explicitly either a poll or a legit conversation (likely a poll is ideal, but I would take either at least for me to say for what people actually want

SeventhM:
Also I find it a bit suspicious of this being such a problem if we can only really find 6 or so mentions of it on the server. I can't even really find a discussion on it on GitHub either besideshttps://github.com/yairm210/Unciv/pull/10840#issuecomment-1873005165

EmperorPinguin:
> Only one of those links are an actual report that avoid growth doesn't work as they felt it should, while the rest are requests for clarification. And none of them are an actual conversation, which is what I'm requesting. I'm requesting explicitly either a poll or a legit conversation (likely a poll is ideal, but I would take either at least for me to say for what people actually want

The third link is indeed maybe more of an inquiry rather that a report, but the first two definitely look like reports to me (you could maybe find more by looking further into the server history).

SeventhM:
> ut the first two definitely look like reports to me (you could probably find more by looking further into the server history)

The second one is clearly a report. The first one is not. It explicitly says that it's fine if it was unable to stop growth, but that the issue was that avoid growth wasn't minimizing growth as well as one could get manually

> The fifth link I send seems worth investigating furter. @itanasi, did you test how this function behaves in Civ5?

I have quickly thrown together a game in Civ 5 and gave myself a bunch of farms with the in game editor mod. No, avoid growth does not stop a city from gaining food in any way. Interestingly, it turns out it does stop it from gaining population, though, so I guess I was wrong on it not having an effect in Civ 5

EmperorPinguin:
> The second one is clearly a report. The first one is not. It explicitly says that it's fine if it was unable to stop growth, but that the issue was that avoid growth wasn't minimizing growth as well as one could get manually

He says he still needs to micromanages his citizens to get the desired autcome (not growing into empire wide sadness), which is not 'fine'.

SeventhM:
> He explains he understands the limitations of the button, but that doesn't make it "not a report" according to me.

Ok, but my point would be that it's a report in relation to #12335 (the button isn't as optimal as it should be), not this issue/PR. These are similar, but separate discussions

itanasi:
> > The fifth link I send seems worth investigating furter. @itanasi, did you test how this function behaves in Civ5?
> 
> I have quickly thrown together a game in Civ 5 and gave myself a bunch of farms with the in game editor mod. No, avoid growth does not stop a city from gaining food in any way. Interestingly, it turns out it does stop it from gaining population, though, so I guess I was wrong on it not having an effect in Civ 5

Well this seems like we should match Civ5 then. Which is what this does.

SeventhM:
Hmm... I guess. Tbh I'm still biased and think there should actually be a discussion (tbh, outside of maybe mods, it's pretty rare that you explicitly don't want to grow, and this setting only comes up in only those niche scenarios), but since it is technically Civ 5 behavior, I guess I'm more lenient to it. Tbh I thought this PR prevented food gain with avoid growth, rather than preventing population (which is far more of a big deal)

itanasi:
https://discord.com/channels/586194543280390151/618491418859798539/1300136441418092575

yairm210 (approved):
Please add this to https://yairm210.github.io/Unciv/Other/Intentional-departures-from-Civ-V/ as well

itanasi:
According to SeventhM, this actually will now match Civ5 behavior. Will change top comment of pull

yairm210:
Huh
OK then

PR #12356, opened 2024-10-27
Avoid Growth and Food Ranking Improvements

itanasi:
Cleans up the Food weights to make more clear/robust.
All Automation will now only x8 the food required to Not Starve. Rest get base weight.
Avoid Growth evaluates all additional Food beyond needed for starvation as 0. Should allow it to prioritize tiles with just enough Food to get positive
Avoid Growth actually evaluates Food for Starvation as important (used to allow Starvation since no x8 in the logic)
Add in small bonus weights to help with tie-breaking (Production over gold, 4 Science over 3 Production, Focus should take priority)

Notice here with Avoid Growth it picked a 2Food tile with other yields instead of blindly picking  a 4Food tile, which is what the old algorithm would do. So now the Growth Food is closer to 0, and we get better other yields.
image

Or switch out of perpetual Gold Production, and here we pick a 1Food with a lot of Production.
image

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -21,7 +21,8 @@ import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.stats.Stat
 import com.unciv.models.stats.Stats
 import com.unciv.ui.screens.victoryscreen.RankingType
-import kotlin.math.max
+import kotlin.math.min
+
 
 object Automation {
 
@@ -72,24 +73,25 @@ object Automation {
             yieldStats.production += cityStatsObj.getProductionFromExcessiveFood(surplusFood+yieldStats.food) - cityStatsObj.getProductionFromExcessiveFood(surplusFood)
             yieldStats.food = 0f  // all food goes to 0
         }
-
-        var growthFood =
-            yieldStats.food  // amount of food yield beyond needed to avoid starving, default all
-        if (starving)
-            growthFood = max(yieldStats.food + surplusFood, 0f)
-        val feedFood = yieldStats.food - growthFood // how much to feed pop
-        // avoid growth, only count Food that gets you not-starving, but no more
-        // will be 0 if not starving
-        if (city.avoidGrowth) {
-            yieldStats.food = feedFood
-            growthFood = 0f // also zero out excess beyond needed to not starve
-        }
+        
         // Apply base weights
         yieldStats.applyRankingWeights()
 
-        // if starving, need Food, scale feedFood by 14(base weight)*8(super important)
-        yieldStats.food += feedFood * (14 * 8 - 1)
-        // growthFood is any food not required to meet Starvation
+        // Split Food Yield into feedFood, amount needed to not Starve
+        // and growthFood, any amount above that
+        var feedFood = 0f
+        if (starving)
+            feedFood = min(yieldStats.food, -surplusFood).coerceAtLeast(0f)
+        var growthFood = yieldStats.food - feedFood // how much extra Food we yield
+        // Avoid Growth, only count Food that gets you not-starving, but no more
+        if (city.avoidGrowth) {
+            growthFood = 0f
+        }
+
+        // if starving, need Food, so feedFood > 0
+        // scale feedFood by 14(base weight)*8(super important)
+        yieldStats.food = feedFood * (14 * 8)
+        // growthFood is any additional food not required to meet Starvation
         if (cityAIFocus in CityFocus.zeroFoodFocuses) {
             // Focus on non-food/growth
             // Reduce excess food focus to prevent Happiness spiral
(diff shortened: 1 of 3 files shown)


core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -64,33 +65,46 @@ object Automation {
...
         if (cityStatsObj.canConvertFoodToProduction(surplusFood, city.cityConstructions.getCurrentConstruction())) {
             // calculate delta increase of food->prod. This isn't linear
             yieldStats.production += cityStatsObj.getProductionFromExcessiveFood(surplusFood+yieldStats.food) - cityStatsObj.getProductionFromExcessiveFood(surplusFood)
             yieldStats.food = 0f  // all food goes to 0
         }
+
+        var growthFood =
+            yieldStats.food  // amount of food yield beyond needed to avoid starving, default all
+        if (starving)
+            growthFood = max(yieldStats.food + surplusFood, 0f)
```
yairm210, line 79: Please use .coerceAtLeast(0f) because in my opinion (and that of others I've talked to about this) it makes it much clearer what we're aiming for

yairm210:
I'm really not sure that this makes anything more clear, I'm much more confused after reading this

core/src/com/unciv/models/stats/Stats.kt
```diff
@@ -146,7 +146,7 @@ open class Stats(
     /** **Mutating function**
      * Apply weighting for Production Ranking */
     fun applyRankingWeights() {
-        food *= 14
+        //food *= 14  // Do weighting in rankStatsForCityWork() instead
```
yairm210, line 149: Is that the only place we use this ranking function?

itanasi: Yup

itanasi: Put back in and derive it in the code instead

EmperorPinguin:
@itanasi I had a question about this PR. What exactly does it do differently compared to the previous implementation? To me it looks like it still multiplies the value of food by 8 when surplus food < 0, and by 2 (or 1/4 when unhappy), instead of resolving the efficiency problem. Could you provide me the savefile with the Korea city in your screenshots?

itanasi:
[Korea - 61 turns.txt](https://github.com/user-attachments/files/17673264/Korea.-.61.turns.txt)
The main efficiency improvement is that once the City is close to not Starving, we can weigh only the required Food to meet the Starvation requirements. The old algorithm would multiply ALL of the Food yield by 8.

For example, say the Food deficit is 1Food. In the old algorithm, this means you're Starving and thus a 3Food tile would get a weighting of 24, and thus will beat out other options. In the new algorithm, only the 1Food needed is multiplied by 8. So if you have a 1Food+4Production tile vs 3Food tile, it will pick the 1Food+4Production. Not only does this mean that the Avoid Growth will more consistently end up at 0 Growth, but also other yields will have a chance to beat out the Food.

The next step is evaluating at the end if we can trade tiles so improve the setup. For example, you Focus Production, and there's a lot of high Food and Production tiles (say 3/3), then you can trade out the earlier selected Super High Food tiles (like 4Food) to get more Production and still not Starve. There's code in Civ5 that does this but I haven't worked through a proper algorithm. Maybe you can tackle it.

EmperorPinguin:
> [Korea - 61 turns.txt](https://github.com/user-attachments/files/17673264/Korea.-.61.turns.txt) The main efficiency improvement is that once the City is close to not Starving, we can weigh only the required Food to meet the Starvation requirements. The old algorithm would multiply ALL of the Food yield by 8.
> 
> For example, say the Food deficit is 1Food. In the old algorithm, this means you're Starving and thus a 3Food tile would get a weighting of 24, and thus will beat out other options. In the new algorithm, only the 1Food needed is multiplied by 8. So if you have a 1Food+4Production tile vs 3Food tile, it will pick the 1Food+4Production. Not only does this mean that the Avoid Growth will more consistently end up at 0 Growth, but also other yields will have a chance to beat out the Food.

In my save file it's behaving differently from your screenshot, it's working the horse tile on my device (I found it suspicious it didn't in the screenshot), so there's indeed an improvement in behaviour.

EmperorPinguin:
> The next step is evaluating at the end if we can trade tiles so improve the setup. For example, you Focus Production, and there's a lot of high Food and Production tiles (say 3/3), then you can trade out the earlier selected Super High Food tiles (like 4Food) to get more Production and still not Starve. There's code in Civ5 that does this but I haven't worked through a proper algorithm. Maybe you can tackle it.

One thing to improve is that the "avoid growth" behaviour should trigger when the happiness reaches -9 or -10, as that's when happiness penalties become a lot more severe. Another thing to improve is the ranking for great person points. Currently, they're all ranked equally, but getting a great person of one type pushes up the cost of the other great person, so they're misvalued, and don't distinguish between the abilities of different types of great persons. I could add the first thing myself I think.

itanasi:
> In my save file it's behaving differently from your screenshot, it's working the horse tile on my device (I found it suspicious it didn't in the screenshot), so there's indeed an improvement in behaviour.

Screenshots were before I added the tie breaker code to the weights.

> One thing to improve is that the "avoid growth" behaviour should trigger when the happiness reaches -9 or -10, as that's when happiness penalties become a lot more severe.

I assume you mean the Growth / 4 adjustment. I actually disagree here. We are supposed to have a penalty on Growth for being Unhappy. So this should be part of the calculation. However I haven't found it in the code, so we probably should fix that.
https://civilization.fandom.com/wiki/Happiness_(Civ5)

PR #12382, opened 2024-10-30
Fixing workers not repairing tiles with Great improvement when pillaged

Emandac:
With this code the worker ai should fix pillaged tiles with Great improvement.


EmperorPinguin:
You need to clean up this other stuff you used.

Emandac:
> a couple of added/d



> You need to clean up this other stuff you used for the city settling issue, also a couple of added/deleted spaces (you can look at "Files changed" above)

Done 👍

core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
```diff
@@ -190,7 +190,6 @@ object SpecificUnitAutomation {
                 throw Exception("City within distance")
             return
         }
-
```
yairm210, line 193: Undo

core/src/com/unciv/logic/automation/unit/WorkerAutomation.kt
```diff
@@ -270,8 +270,6 @@ class WorkerAutomation(
         tileRankings[tile] = TileImprovementRank(priority)
         return priority + unitSpecificPriority
     }
-
```
yairm210, line 273: Undo

yairm210 (approved):

PR #12406, opened 2024-11-03
Example for automating golden age triggers

yairm210:
@EmperorPinguin I recall you asked for this...somewhere, there are a lot messages
I'm not sure this is good, since it reads "if you have nothing better to do just Golden Age", YMMV


yairm210:
Can you provide the save file please?

EmperorPinguin:
That seems right. The AI does have a tendency to get stuck with Louvre/Freedom Great Artists, as in that stage of the game it has run out of tiles to place them, causing the artists to be useless and obstructing spaceship parts. This won't trigger Great General golden ages (we should save those for later combat), as that is taken care of in lines 65-71 right?

PR #12423, opened 2024-11-06
Add owned tiles countable

PLynx01:
I would like to introduce tile-counting countable. Unlike my previous attempt, it counts only tiles owned by relevant civilization.

If you see some problems, let me know. Constructive feedback welcome.

### DIFF
--- a/core/src/com/unciv/models/ruleset/unique/Countables.kt
+++ b/core/src/com/unciv/models/ruleset/unique/Countables.kt
@@ -37,12 +37,8 @@ object Countables {
             return gameInfo.civilizations.filter { !it.isDefeated() }
                 .count { it.matchesFilter(placeholderParameters[0]) }
         
-        if (countable.equalsPlaceholderText("[] Tiles")) {
-            val ownedTiles = gameInfo.tileMap.values.filter {
-                it.getOwner() == civInfo
-            }
-            return ownedTiles.count { it.matchesFilter(placeholderParameters[0]) }
-        }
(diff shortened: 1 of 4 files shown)


core/src/com/unciv/models/ruleset/unique/Countables.kt
```diff
@@ -36,6 +36,13 @@ object Countables {
         if (countable.equalsPlaceholderText("Remaining [] Civilizations"))
             return gameInfo.civilizations.filter { !it.isDefeated() }
                 .count { it.matchesFilter(placeholderParameters[0]) }
+        
+        if (countable.equalsPlaceholderText("[] Tiles")) {
```
yairm210, line 40: Change phrasing to "owned [] Tiles"

core/src/com/unciv/models/ruleset/unique/Countables.kt
```diff
@@ -36,6 +36,13 @@ object Countables {
...
+            val ownedTiles = gameInfo.tileMap.values.filter {
```
yairm210, line 41: For performance, we want to 
- iterate on civ cities
- sum the tiles for each city that fit the filter
- sum the total counts

So something like civInfo.cities.sumOf { it.tiles.count { it.matchesFilter(placeholderParameters[0]) } }

yairm210:
Looks good! 👍🏿 
Need to update UniqueParameterType.Countable to make mod checker recognize this

yairm210 (approved):

yairm210:
Actually we also need to add this to template.properties for transfusions

PR #12435, opened 2024-11-09
Fix population math

SeventhM:
Fix https://github.com/yairm210/Unciv/commit/92fd4883010b6133273278de0bc855aee99193d5

### DIFF
--- a/core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
+++ b/core/src/com/unciv/logic/city/managers/CityPopulationManager.kt
@@ -130,7 +130,7 @@ class CityPopulationManager : IsPartOfGameInfoSerialization {
     }
 
     fun addPopulation(count: Int) {
-        val changedAmount = count.coerceAtMost(population - 1)
+        val changedAmount = count.coerceAtLeast(1 - population)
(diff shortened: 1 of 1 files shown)


SeventhM:
Not sure I want to debug these failures tbh

yairm210:
These tests are based the cities having zero population.
This change cancels that, which means the tests need to be redone for 1-pop circumstances

pipo-gold:
With these later changes, I can't raze a city. After multiple turn at size 1 and 'razing', the city is still there.
I could open another bug, but I think this is related.

SeventhM:
~~Ngl, kinda expected that to become an issue. Probably should have a separate function or something just for that~~

pipo-gold:
Version 4.14.5 (Build 1068) still can't raze a city.  Long test to do, hehe, turn 49 before I capture a city and try to raze it.
Only been playing since feb. 2024 and now playing through Android Studio (Desktop).

yairm210:
Interestingly, this raised the point that *nukes already did the coercion to not have zero pop* - but when we solved that problem we didn't yet have triggers so it wasn't generic'd

PR #12436, opened 2024-11-09
Unhappy Food Ranking Fix

itanasi:
Aligns the code to automatically take into account the Unhappiness penalty to Growth.
Currently also puts back in Growth for Non-Default/Non-Food Focus. But waiting on Discord Poll results.

Confirmed that we do have the Growth Penalty when Unhappy. It's being set in the Global Uniques
image
Screenshot when 5 Growth
image
And when <-10 Unhappiness, no growth!
image

Therefore we should rank the tiles appropriately
[Polynesia - 10unhappy.txt](https://github.com/user-attachments/files/17686303/Polynesia.-.10unhappy.txt)
[Autosave-Polynesia-69.txt](https://github.com/user-attachments/files/17686304/Autosave-Polynesia-69.txt)

@EmperorPinguin since they are also messing with rank weights

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -98,18 +98,16 @@ object Automation {
         // always picking the Highest Food tile until Not Starving
         yieldStats.food = feedFood * (foodBaseWeight * 8)
         // growthFood is any additional food not required to meet Starvation
-        if (cityAIFocus in CityFocus.zeroFoodFocuses) {
-            // Focus on non-food/growth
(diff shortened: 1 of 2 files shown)


EmperorPinguin:
@itanasi we want to reduce our growth to 0 before we hit our nullified growth, as -11 happiness also bring combat and production penalties.

EmperorPinguin:
I'm not sure if growth bonuses such as WLTKD should be taken into account. I think it's more of a "nice, extra yields" rather than something that makes you reconsider citizen assignement (also, if you want to do it correctly, carried over food such as from aqueducts should be taken into account).

PR #12441, opened 2024-11-10
Update Automation.kt

EmperorPinguin:
* Updates AI citizen assignement to take into account a 1-happiness margin (instead of growing into unhappiness asap), and attempt preventing growth into -11 happiness (treshold set at -9 instead of -10, in case we find a pop ruin or something).

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -105,12 +105,10 @@ object Automation {
                 yieldStats.food += growthFood * (foodBaseWeight / 4)
         } else {
             // NoFocus or Food/Growth Focus.
-            // When Happy, EmperorPenguin has run sims comparing weights
-            // 1.5f is preferred,
-            // but 2 provides more protection against badly configured personalities
-            // If unhappy, see above
-            val growthFoodScaling = if (city.civ.getHappiness() >= 0) foodBaseWeight * 2 else foodBaseWeight / 4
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -105,12 +105,10 @@ object Automation {
...
-            // When Happy, EmperorPenguin has run sims comparing weights
-            // 1.5f is preferred,
-            // but 2 provides more protection against badly configured personalities
-            // If unhappy, see above
-            val growthFoodScaling = if (city.civ.getHappiness() >= 0) foodBaseWeight * 2 else foodBaseWeight / 4
-            yieldStats.food += growthFood * growthFoodScaling
+            // When Happy, 2 production is better than 1 growth,
+            // but setting such by default worsens AI civ citizen assignment,
+            // probably due to badly configured personalities not properly weighing food vs non-food yields
+            val growthFoodScaling = if (city.civ.getHappiness() > 0) foodBaseWeight * 2 else if (city.civ.getHappiness() < 8) foodBaseWeight * 0 else foodBaseWeight / 4
```
yairm210, line 111: if/else assignments should be converted to "when" statements
Or at the very least split into lines:
 if ...
else if ...
else ...

EmperorPinguin: It was crashing as I deleted line 113 by accident. I can split it into lines if you want.

yairm210 (approved):

PR #12445, opened 2024-11-10
Update MotivationToAttackAutomation.kt

EmperorPinguin:
* Changes AI motivationToAttackAutomation, the reasoning being the following:
Line 84: A research agreement is a trade deal, if you cancel the trade and waste both player's gold, you're a real bad businessman, but not necessarily a bad friend, an increased weighing by Commerce personality seems more fitting
Line 107: See above, I think it makes more sense to weigh this by Military value, as we need resources to build and maintain a strong standing army.
Line 119 + 120: CS play should by guided by Diplomacy personality, to stay consistent with the naming scheme of Diplomatic victory, Diplomatic preferredVictorType, Diplomatic Marriage, Cultural Diplomacy etc.
Line 204: Loyalty should decide if we stay true to a friendship, or persist against an enemy, not Diplomacy.
Line 217 + 236: Higher Science and Production personality stats will (should) increase our tech and production, which will then increase attack motivation according to getRelativeTechModifier and getProductionRatioModifier. We should not multiply this once more by the personality value.
Line 255: Score is a proxy for owned World Wonders and number of cities, both of which increase long-term culture generation. It makes sense to weigh this according to Culture personality to identify competitors (score doesn't tell much else).

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/MotivationToAttackAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/MotivationToAttackAutomation.kt
@@ -104,7 +104,7 @@ object MotivationToAttackAutomation {
         }
 
         if (diplomacyManager.resourcesFromTrade().any { it.amount > 0 })
-            modifiers.add(Pair("Receiving trade resources", -8f * personality.modifierFocus(PersonalityValue.Commerce, .5f) * personality.modifierFocus(PersonalityValue.Military, .2f)))
+            modifiers.add(Pair("Receiving trade resources", -8f * personality.modifierFocus(PersonalityValue.Commerce, .5f)))
 
         // If their cities don't have any nearby cities that are also targets to us and it doesn't include their capital
         // Then there cities are likely isolated and a good target.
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/civilization/MotivationToAttackAutomation.kt
```diff
@@ -105,7 +104,7 @@ object MotivationToAttackAutomation {
         }
 
         if (diplomacyManager.resourcesFromTrade().any { it.amount > 0 })
-            modifiers.add(Pair("Receiving trade resources", -8f * personality.modifierFocus(PersonalityValue.Commerce, .5f) * personality.modifierFocus(PersonalityValue.Loyal, .2f)))
+            modifiers.add(Pair("Receiving trade resources", -8f * personality.modifierFocus(PersonalityValue.Commerce, .5f) * personality.modifierFocus(PersonalityValue.Military, .2f)))
```
yairm210, line 107: Why military? Because strategic resources? There's also luxuries!
This should be something that's disconnected from personality IMO

EmperorPinguin: High Military will increase the amount of military units produced, which increase the Force advantage over the opponent, and as such make the AI indirectly more likely to declare war. Instead of then declaring war on just anyone, imo it should take into account the trades more, and stay at peace with those that provide strategic (necessary for building strong military units) and other resources.

PR #12456, opened 2024-11-12

EmperorPinguin:
* Increase worker prioritization of AI, as requested by https://discord.com/channels/586194543280390151/653651863832363053/1305801888804900895. City states will now usually construct a worker right after the initial military unit.

### DIFF
--- a/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
+++ b/core/src/com/unciv/logic/automation/city/ConstructionAutomation.kt
@@ -251,7 +251,7 @@ class ConstructionAutomation(val cityConstructions: CityConstructions) {
         val numberOfWorkersWeWant = if (cities <= 5) (cities * 1.5f) else 7.5f + ((cities - 5))
 
         if (workers < numberOfWorkersWeWant) {
-            val modifier = numberOfWorkersWeWant / (workers + 0.4f) // The worse our worker to city ratio is, the more desperate we are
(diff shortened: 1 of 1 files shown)


yairm210:
What he's asking for is "please make city states more cheesable". That's definitely not a good reason to change and to me is an indicator that we actually *shouldn't* change this.
If anything, the default for city states should be the opposite - to make them less cheesable.
Also, city-state-specific automation decisions should probably be separated from major civ AI decisions.

The real question is - does this change the build queue *to the betterment* of AIs, to make them more successful? My guess is that it does not

EmperorPinguin:
> The real question is - does this change the build queue _to the betterment_ of AIs, to make them more successful? My guess is that it does not

I don't know. I myself build workers before granaries, but for the AI this might cause them to lose more workers to barbarians (and to nasty worker-stealing humans).

EmperorPinguin:
Good news: the 5-game average (20 data points) yielded a 99.9 t80 pop, vs the previously reported 95.8 t80 pop from the same save file. The game got quite a few changes since then though (also the aqueduct eval in my local fork), so it's not a ceteris-paribus result.

EmperorPinguin:
Without this higher worker priority, the AI does perform worse, I got an average of 94.7 pop. Due to the large variation and small sample size, the difference is not statistically significant, however. Maybe someone could setup an automated testing with the ConsoleLauncer.

PR #12469, opened 2024-11-14
Unhappy Food Weight Fix

itanasi:
As from #12436 on a clean branch.

@EmperorPinguin Did you intend for ALL Negative Happiness to weight Growth at 0? Or was that supposed to be < -8

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -2,6 +2,7 @@ package com.unciv.logic.automation
 
 import com.unciv.logic.city.City
 import com.unciv.logic.city.CityFocus
+import com.unciv.logic.city.CityStats
 import com.unciv.logic.civilization.Civilization
 import com.unciv.logic.map.BFS
 import com.unciv.logic.map.TileMap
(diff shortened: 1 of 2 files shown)


EmperorPinguin:
> @EmperorPinguin Did you intend for ALL Negative Happiness to weight Growth at 0? Or was that supposed to be < -8
 
 Only when < -8

EmperorPinguin:
Fuck, how did that typo get in

yairm210:
@EmperorPinguin I got to say I'm not following all of your discussions, but if this is a PR that you both agree on it's mergable for me
What's your take?

EmperorPinguin:
> @EmperorPinguin I got to say I'm not following all of your discussions, but if this is a PR that you both agree on it's mergable for me What's your take?

This PR can be merged, it takes the city growth modifier into account in a more generalized way than before, which cannot be bad.

What I was debating with Itanasi was how growth should be valued in the happy+nonstarving  scenario. The current Unciv behaviour overvalues growth such as to not let the basegame AI civ's growth value go below the critical treshold. This is not something that's controversial (players complain about it, and I was the first to do so I think), but the "correct" value turns out to be controversial, with players valueing Growth quite differently, with the majority of them valueing growth at < 4/3×production value, which imo is rather low.

yairm210 (approved):

PR #12472, opened 2024-11-14

EmperorPinguin:
The AI currently does not evaluate tile yields from buildings. For granaries, stables, etc. this is not much of a problem as they're also valued in other ways, but the AI not building lighthouses is pretty bad.

Merging this PR simply as-is will allow the AI to build lighthouses again, but it would be better to extract the type and amount of stats from the unique, and multiply this by the number of respective tiles the AI is currently working in the city. How does one do so?


yairm210:
Incorrect, this is calculated in getStatDifferenceFromBuilding - we compare this city vs this city with the building, which includes the extra stats from tiles

yairm210:
I'm sure that merits investigation, but this is not the correct solution

PR #12491, opened 2024-11-18
Cache state for units, civs, and cities

SeventhM:
Side note: I don't know if it makes sense to refresh the state for civs in ``setTransients``. The only thing I can think of is if the gameInfo changes midgame


yairm210 (approved):
Looks good!

PR #12495, opened 2024-11-19
Adding Server Icons

itanasi:
Potential Icons for unciv Server to use in notifications


yairm210:
These images are FAR TOO BIG, please shrink them down to 100x100

itanasi:
Done

SomeTroglodyte:
@itanasi - they're not referenced in the entire project except for the atlas and credits. So how and when will they be used? This might merit a wiki entry, no?

itanasi:
They're only being (potentially) sourced for the MP notifications. Specifically the injected ones by touhidurr that say you should support them

PR #12496, opened 2024-11-19
ConsoleLauncher: Add binomial test output

itanasi:
Adds automatic calculation of how likely these Civs have a distinct difference based on the one-tailed binomial test and prints it for you

### DIFF
--- a/core/src/com/unciv/logic/simulation/Simulation.kt
+++ b/core/src/com/unciv/logic/simulation/Simulation.kt
@@ -24,6 +24,7 @@ class Simulation(
 ) {
     private val maxSimulations = threadsNumber * simulationsPerThread
     val civilizations = newGameInfo.civilizations.filter { it.civName != Constants.spectator }.map { it.civName }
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/simulation/Simulation.kt
```diff
@@ -112,22 +113,66 @@ class Simulation(
         var outString = ""
         for (civ in civilizations) {
 
-            val wins = winRate[civ]!!.value * 100 / max(steps.size, 1)
-            if (wins == 0) continue
+            val numSteps = max(steps.size, 1)
+            val expWinRate = 1f / civilizations.size.toDouble()
```
yairm210, line 117: Please note that civilizations contains also *minor civs* which are not expected to win, expected win should be 1/(major civs)

itanasi: Ah, fixed. Thanks

PR #12504, opened 2024-11-20
Move capitalCityIndicator

SeventhM:
PR is to move capitalCityIndicator, as it ~~previously~~ doesn't make sense in the city code. Starting as a draft because there are many, many, many questions. I wish I PR'ed this a while ago before "uniqueTo" became something you can filter from as many aspects made more sense before then

1. Currently, the ``IndicatesCapital`` unique does not use conditionals. ``City.isCapital`` uses getUniqiues instead of getMatchingUniques, which inherently ignores condtionals. ``CityConquestFunctions.destroyBuildingsOnCapture`` uses hasUnique, but this is not given a state, so it uses ``EmptyState`` instead of ``IgnoreConditionals``, auto-failing any civ/city check. previously, this function also used hasUnique without a state, thus having the same problem. This PR changes it to specifically use IgnoreConditionals for consistency with ``isCapital``
2.  Since uniqueTo uses filters and filters can now have conditionals ~~and uniqueTo can multiFilter~~, there is a question of which state this should use. I would assume IgnoreConditionals here as well, not the civ's state or the city's state like I erroneously used prior. This PR as of writing this keeps the incorrect behavior of the previous PR, but I can easily change it back
3. If there is an assumption that this should be consistent all game, then this information should be cached. This would change the answer of point 2 to be "the civ state at the time of caching", same as the ``uniqueBulidings`` cache. This isn't a particularly complicated bit, but it does force the question of when we want to check the filter as what to cache changes based on when we want to check for filters

Side note, regardless of those points, the changes to ``Civ.moveCapitalTo`` to keep the code functioning as it did prior. kotlin's smart casting isn't recognizing that ``city`` can't be null if the indicator isn't null

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -163,14 +163,6 @@ class City : IsPartOfGameInfoSerialization, INamed {
     fun getWorkRange(): Int = civ.gameInfo.ruleset.modOptions.constants.cityWorkRange
     fun getExpandRange(): Int = civ.gameInfo.ruleset.modOptions.constants.cityExpandRange
 
-    fun capitalCityIndicator(): Building? {
-        val indicatorBuildings = getRuleset().buildings.values.asSequence()
-            .filter { it.hasUnique(UniqueType.IndicatesCapital) }
-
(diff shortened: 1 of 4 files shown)


SeventhM:
Am closing this in favor of #12509. I'll remake this afterwards as a sort of open question as I think the circumstances of where this function should be is a bit odd as it seems to always use a state of a given city

PR #12507, opened 2024-11-20
Add Croatian language for translation

metablaster:
I've added initial files as described [here](https://yairm210.github.io/Unciv/Other/Translating/#adding-a-new-language) to add Croatian language.

The only step I didn't do is convert national flag icon into atlas file as described [here](https://yairm210.github.io/Unciv/Modders/Images-and-Audio/#images-and-the-texture-atlas) because I have no idea how to do it properly so if somebody can do this step for me please do.

In addition to initial files I've translated a few initial sentences in `Croatian.properties` and also text in:
`fastlane/metadata/android`\
Added a new directory `hr` and put translation of `full_description.txt` and `short_description.txt`


metablaster:
I did some updates since first PR, it appears these were auto put here because I enabled workflows in my fork, so I don't need to make another PR I guess?

yairm210 (approved):

PR #12530, opened 2024-11-27
Update Indonesian.properties

Xenia2003:
Corrections.


yairm210:
Placeholder `[mapUnitFilter]` not found in `Indonesian` for entry `Earn [amount]% of [mapUnitFilter] unit's [costOrStrength] as [civWideStat] when killed within 4 tiles of a city following this religion`

PR #12550, opened 2024-12-02
Update template.properties

dadadnet:
In the game there are only "Please don't spread religion to us." textline (without your), so unciv doesn't translate it.


yairm210:
This will be replaced by `Don't spread religion in our cities` as part of another suggestion 👍🏿

PR #12577, opened 2024-12-06
Add stockpile uniques for golden ages

SeventhM:
I know I asked on Discord how I should do this ahead of time, but I do want to actually write out the question kinda as an implementation in code just to give a better understanding of what I'm requesting here. This PR is not meant to be merged alongside #12578 or #12579, and is just a possible solution

closes #12578, closes #12579

This PR adds uniques for stockpiles and golden age points to be gained either as a triggerable or from combat

Pros:
- Allows us to deprecate both the gain and lose uniques for stockpiles similar to how we would do for other similar scenarios
- Adds a unique for gaining resources by combat as well, which may see use elsewhere

Cons:
- Defining what is or isn't a stockpile is kinda tricky. It kinda makes sense for resources, but for hard coded values it becomes a question of definitions. Why isn't stats a stockpile? Are stockpiles supposed to be for resources or other stuff?
- Likewise, the text for golden age points is a bit awkward. Do you put the whole thing "golden age points", or some shortened form ("GA points" or "Golden Age")
- The unique for damage in particular is pretty heavily copy and paste from the stat one which keeps with the pure unique alternative of being something that'll likely be deprecated further

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -390,7 +390,7 @@ object Battle {
                 val percentage = unique.params[0].toFloat()
                 val stockpile = unique.params[2]
                 if (stockpile == "golden age points") {
-                    civ.goldenAges.addHappines((percentage / 100f * damageDealt).roundToInt())
+                    civ.goldenAges.addHappiness((percentage / 100f * damageDealt).roundToInt())
                     continue
                 }
                 val plunderedResource = civ.gameInfo.ruleset.tileResources[stockpile]
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -389,7 +389,7 @@ object Battle {
             if (plunderedUnit.matchesFilter(unique.params[1])) {
                 val percentage = unique.params[0].toFloat()
                 val stockpile = unique.params[2]
-                if (stockpile == "golden age points") {
+                if (stockpile == Constants.goldenAgePoints) {
                     civ.goldenAges.addHappiness((percentage / 100f * damageDealt).roundToInt())
                     continue
                 }
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -26,6 +26,7 @@ import com.unciv.ui.screens.worldscreen.unit.actions.UnitActionsPillage
 import com.unciv.utils.debug
 import kotlin.math.max
 import kotlin.math.min
+import kotlin.math.roundToInt
 import kotlin.random.Random
 
 /**
@@ -383,6 +384,20 @@ object Battle {
                 plunderingUnit.getName(), NotificationIcon.War, "StatIcons/${key.name}",
                 if (plunderedUnit is CityCombatant) NotificationIcon.City else plunderedUnit.getName()
             )
+        }
+        for (unique in plunderingUnit.unit.getMatchingUniques(UniqueType.DamageUnitsPlunderStockpile, checkCivInfoUniques = true)) {
+            if (plunderedUnit.matchesFilter(unique.params[1])) {
+                val percentage = unique.params[0].toFloat()
+                val stockpile = unique.params[2]
+                if (stockpile == "golden age points") {
+                    civ.goldenAges.addHappines((percentage / 100f * damageDealt).roundToInt())
+                    continue
+                }
+                val plunderedResource = civ.gameInfo.ruleset.tileResources[stockpile]
+                if (plunderedResource != null && plunderedResource.isStockpiled()) {
+                    civ.resourceStockpiles.add(stockpile, (percentage / 100f * damageDealt).roundToInt())
+                }
+            }
         }
     }
 
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -363,16 +363,28 @@ object Battle {
     ) {
         // implementation based on the description of the original civilopedia, see issue #4374
         if (plunderingUnit !is MapUnitCombatant) return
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -806,6 +807,8 @@ enum class UniqueType(
     OneTimeFreeBelief("Gain a free [beliefType] belief", UniqueTarget.Triggerable),
     OneTimeTriggerVoting("Triggers voting for the Diplomatic Victory", UniqueTarget.Triggerable),  // used in Building
 
+    OneTimeGainStockpile("Instantly gain [amount] [stockpile]", UniqueTarget.Triggerable, flags = setOf(UniqueFlag.AcceptsSpeedModifier)),
+    OneTimeGainStockpileRange("Instantly gain [amount]-[amount] [stockpile]", UniqueTarget.Triggerable, flags = setOf(UniqueFlag.AcceptsSpeedModifier)),
```
yairm210, line 811: Should use existing Instantly Consumes / Instantly Provides uniques

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -384,6 +385,20 @@ object Battle {
                 if (plunderedUnit is CityCombatant) NotificationIcon.City else plunderedUnit.getName()
             )
         }
+        for (unique in plunderingUnit.unit.getMatchingUniques(UniqueType.DamageUnitsPlunderStockpile, checkCivInfoUniques = true)) {
+            if (plunderedUnit.matchesFilter(unique.params[1])) {
+                val percentage = unique.params[0].toFloat()
+                val stockpile = unique.params[2]
+                if (stockpile == "golden age points") {
```
yairm210, line 392: Should be a Constant string, not hardcoded everywhere

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -448,7 +448,8 @@ enum class UniqueType(
     UnitUpgradeCost("[relativeAmount]% Gold cost of upgrading", UniqueTarget.Unit, UniqueTarget.Global),
 
     // Gains from battle
-    DamageUnitsPlunder("Earn [amount]% of the damage done to [combatantFilter] units as [civWideStat]", UniqueTarget.Unit, UniqueTarget.Global),
+    DamageUnitsPlunder("Earn [amount]% of the damage done to [combatantFilter] units as [stockpile/civWideStat]", UniqueTarget.Unit, UniqueTarget.Global),
+    DamageUnitsPlunderStockpile("Gain [amount]% of the damage done to [combatantFilter] units as [stockpile]", UniqueTarget.Unit, UniqueTarget.Global),
```
yairm210, line 452: These read, in English, as exactly the same thing. Are they different?

yairm210:
It makes sense to treat hardcoded counters as sort of "in-game Stockpiled Resources" so we can use the same uniques for them.
I don't see a good reason to change the way we store stats

Golden Ages should always be capitalized, so "Golden Age points" or "GA points".

yairm210:
Perhaps we have other such "actually a stockpile all along" such as Great Person Points, deserves a think

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
SeventhM: I kinda have issues with those uniques, tbh. Basically \*nothing\* else uses that design style for uniques except for the non stockpile resource uniques. For the non stockpile resource uniques, there's a mechanical reasoning behind why it's 2 separate uniques (there's various UI considerations such as being how we automatically place a restriction of where stuff could be built). For stockpiles, there is no mechanical reasoning for having 2 separate uniques

I mean, just look at nearly all other use cases and I'd argue that it's those that need deprecating here, rather than this using those uniques

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
SeventhM: Not necessarily, but for code clarity reasons, I decided to initially keep them separate. As written in my commit note, merging them is complicated, was waiting for the initial response before merging the two uniques

SeventhM:
> Perhaps we have other such "actually a stockpile all along" such as Great Person Points, deserves a think

This is where, kinda as stated before, I kinda disagree. Or rather, I think there needs to be a better clarification of what is and isn't a stockpile. Because there's little to no good argument that golden age points is a "resource". Likewise, besides, say, this attempt to interact with it, golden age points really aren't even their own resource. Civ 5 more or less treats it as part of the happiness stat ("Golden age points" is a fan term, most of the time you see "points towards a golden age")

Again, are stockpiles just "everything that's not a stat"? Imo, that seems kinda wrong compared to how we typically do things. At minimum, I would imagine the better approach would be to wrap hardcoded stockpiles into their own class and then do something similar to the stat version of this. Minimum, that would give 1 to 1 parity with stats for almost all future uniques

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
yairm210: Not sure what you mean? Do you mean using "instantly gain [-2]" instead of "instantly lose [2]"?

yairm210:
I don't have a good answer to any of your questions, unfortunately :/
I think you've thought about this much longer and harder than I have

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
SeventhM: Yes

SeventhM:
An aside: I only recently realized I've been copying over the wrong unique to use for golden age points. We don't necessarily want a copy of ``DamageUnitsPlunder``, we want ``KillUnitPlunder``. Even more reason why I think the stat variant of this is generally better since it basically ignores the entire topic outright

PR #12579, opened 2024-12-06
Add golden age points to stats

SeventhM:
I know I asked on Discord how I should do this ahead of time, but I do want to actually write out the question kinda as an implementation in code just to give a better understanding of what I'm requesting here. This PR is not meant to be merged alongside #12577 or #12578, and is just a possible solution

closes #12577, closes #12578

This PR adds golden age points as if it was a stat

Pros:
- Doesn't actually need to add a unique as we have plenty of uniques for stats around these sorts of topics
- Also potentially opens up other doors for other things (adding to golden age points per turn without affecting happiness)

Cons:
- This is not really a stat as it is a substat (it's more or less directly handled by happiness)
- This does not have its own icon
- It also means stuff like buildings can directly add to it per turn. I consider this both a pro and a con

Side note: We could create a separate interface that encompasses stats and these sorts of "substats" and then maneuver a lot of the code around that instead. That would actually give a benefit of not needing to add logic to this alongside stats objects like buildings and tiles

### DIFF
--- a/core/src/com/unciv/logic/civilization/Civilization.kt
+++ b/core/src/com/unciv/logic/civilization/Civilization.kt
@@ -855,6 +855,7 @@ class Civilization : IsPartOfGameInfoSerialization {
                               if(amount > 0) totalCultureForContests += amount }
             Stat.Science -> tech.addScience(amount)
             Stat.Gold -> addGold(amount)
+            Stat.GoldenAge -> goldenAges.addHappiness(amount)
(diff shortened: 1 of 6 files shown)


SeventhM:
Huh. ``Stats.toString`` fails here because Current is not initialized. Interesting

yairm210:
This is definitely the wrong way to go about this

PR #12582, opened 2024-12-06
Rebalance Food Weights

itanasi:
Rework the weight of Food evaluation for Citizen Management.
Generic AI Civ performance improves and this matches closer to human expectations. Some AIs such as Korea will perform a bit worse in certain scenarios due to Personalities, but we should fix the Personalities instead of break base behavior.
Still room for improvements in corner cases.
Discussion and data in Discord thread:
https://discord.com/channels/586194543280390151/1306726293102006453/1306726299884195932

Also adding a small feature to the console sims


core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -116,16 +116,31 @@ object Automation {
...
-            // Zero out Growth if close to Unhappiness limit as well
-            val baseFocusWeight = if (city.civ.getHappiness() < -8) 0 else {
-                if (cityAIFocus in CityFocus.zeroFoodFocuses) 1 else 2
+            
+            
+            var foodModWeight = 1f
+            // Zero out Growth if close to Unhappiness limit
+            if (city.civ.getHappiness() < -8) foodModWeight = 0f
+            else {
+                if (city.civ.isAI()) {
```
yairm210, line 125: Remove brackets after else - this is a regular else-if chain

core/src/com/unciv/logic/simulation/Simulation.kt
```diff
@@ -42,6 +43,9 @@ class Simulation(
             winRateByVictory[civ] = mutableMapOf()
             for (victory in UncivGame.Current.gameInfo!!.ruleset.victories.keys)
                 winRateByVictory[civ]!![victory] = MutableInt(0)
+            winTurnByVictory[civ] = mutableMapOf()
+            for (victory in UncivGame.Current.gameInfo!!.ruleset.victories.keys)
+                winTurnByVictory[civ]!![victory] = MutableInt(0)
```
yairm210, line 48: Why the copy paste?

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -116,16 +116,31 @@ object Automation {
...
-            // When Happy, 2 production is better than 1 growth,
-            // but setting such by default worsens AI civ citizen assignment,
-            // probably due to badly configured personalities not properly weighing food vs non-food yields
-
-            // Zero out Growth if close to Unhappiness limit as well
-            val baseFocusWeight = if (city.civ.getHappiness() < -8) 0 else {
-                if (cityAIFocus in CityFocus.zeroFoodFocuses) 1 else 2
+            
+            
+            var foodModWeight = 1f
```
yairm210, line 121: This is already 20 lines of dense specific logic to define a single value, please make this a getFoodModWeight() function

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -46,7 +46,36 @@ object Automation {
...
     }
 
-
+    // More complicated logic to properly weigh Food vs other Stats (esp Production)
+    private fun getFoodModWeight(city: City, surplusFood: Float): Float {
+        var foodModWeight = 1f
+
+        // Zero out Growth if close to Unhappiness limit
+        if (city.civ.getHappiness() < -8) {
+            foodModWeight = 0f
```
yairm210, line 55: Change these to "returns" to have minimal nesting :)
I see there's no actual modifications ( +, -) only setting (=) so should be trivial
Replace the final "return" with "return 1f" for default case, and then each specific case gets a "return" as well

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -46,7 +46,36 @@ object Automation {
...
+            else
+                foodModWeight = 1.5f
+        } else {
+            // Human weights. May be different since AI Happiness is always "easier"
+            // Only apply these for Default to not interfere with Focus weights
+            if (city.getCityFocus() == CityFocus.NoFocus) {
+                if (city.population.population < 5)
+                    foodModWeight = 2f
+                else if (surplusFood > city.population.getFoodToNextPopulation() / 10)
+                    foodModWeight = 0.75f // get Growth just under Production
```
EmperorPinguin, line 73: You make the growth value dependent on the population and the current growth/turn of the city, which is something the human would need to start keeping track of. I explicitly removed similar requirements some time ago in a PR, as it makes it is annoying keeping track of individual cities that way.
Also, and this is maybe more of a personal preference than a common sentiment, but I would like my city to get 1 growth over 1 production+1 gold. If I want the latter during Happiness (for example when at war). I can click the Avoid Growth button myselves.

itanasi: Small Cities need a higher Food Weight since Growth is more important. For 90% of players who don't optimize assignments every turn, this is more in line with generally good play. I asked you in the Discord thread if there were better metrics to use based on Population and you didn't respond.

As for the base weight, 1Growth over 1Prod+1Gold would mean Food/Prod >3/2 (since Gold = 1/2 Prod). I'd roughly agree and if that's what we really want to do we should adjust the base value not have a multiplier here. But the poll also indicated this was a minority evaluation. Arguably the experts valued it about 1.5x. But this is a grey area. I previously proposed increasing the base 14->16.

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Done

core/src/com/unciv/logic/automation/Automation.kt
EmperorPinguin: You asked my, and I replied it would be a better idea to make it dependent on the global game state (game ending conditions). The are some problems with that, as making the food eval dependent on war status makes it very exploitable by humans (they can declare war and sit back to defend until the AI is irrelevant), and making it dependent on tech% complete doesn't work properly if players start a new game in late eras. The latter would be very nice to combine with scientist saving though, maybe it should be given a try.

itanasi: Right. Those methods may be slightly more optimal, but introduce a lot more complications, is brittle, and you haven't described a clean algorithm for any of them. This solves a common issue, is simple, and robust enough to handle mods.

Marking as Draft for now. If you still have major issues let's discuss on Discord.

core/src/com/unciv/logic/automation/Automation.kt
```diff
@@ -46,7 +46,34 @@ object Automation {
         return rank
     }
 
-
+    // More complicated logic to properly weigh Food vs other Stats (esp Production)
+    private fun getFoodModWeight(city: City, surplusFood: Float): Float {
+        // Zero out Growth if close to Unhappiness limit
+        if (city.civ.getHappiness() < -8) {
+            return 0f
+        } else if (city.civ.isAI()) {
```
yairm210, line 54: Sorry for being pedantic, but this is a definite code cleanup issue
All of this can be "unnested" - all the "else" main branches are unnecessary, since the first and second "if"s always return values

yairm210 (approved):
Looks good to me
@EmperorPinguin good for you?

core/src/com/unciv/logic/automation/Automation.kt
itanasi: Fixed. Thanks.

EmperorPinguin:
> Looks good to me @EmperorPinguin good for you?

I've tested lower food values (1.5f, 1.6f, 1.5f with a 4 pop cutoff, 1.5f with 50% reduced scaling of AI personality values) myself in Spectator mode, and each time I see a clear drop in population benchmarks, so no, I disagree with these values. I do however like the idea of setting separate values for the AI and for the humans, as they likely have different optimal tradeoff points due to differences in happiness and production constraints.

itanasi:
> I've tested lower food values (1.5f, 1.6f, 1.5f with a 4 pop cutoff, 1.5f with 50% reduced scaling of AI personality values) myself in Spectator mode, and each time I see a clear drop in population benchmarks, so no, I disagree with these values.

Are you only checking for Korea? The sims clearly show for a generic civ that 1.5x beats 2x as default modifier. https://discord.com/channels/586194543280390151/1306726293102006453/1314995563560829020

Again, if your issue is trying to fix specific edge cases, then that should be with specific code. You agreed that 1.5x is better for the general case than 2x.

> Are the conditions on lines 60 and 69 properly modified by game speed? On quick speed, 10 turns is pretty long, but it looks to me that this logic doesn't account for slower speeds.
No, but we can scale that.

EmperorPinguin:
> Are you only checking for Korea?

No, there are 4 civs on the map (Austria, Korea, Maya, Songhai). I kept this one from my worker update patch, as I 'know' where they settles and how fast they deveop. I think the effects are most pronounced with Austria.

>  The sims clearly show for a generic civ that 1.5x beats 2x as default modifier. https://discord.com/channels/586194543280390151/1306726293102006453/1314995563560829020

I know, but that doesn't mean that: 
1) results for duels are the same as for free sim ("speedrun")
2) results for default civs are the same as for in-game civs, with weird effects from personalities (thighter ranges of 'acceptable' values)

itanasi:
Then I'd argue you're mixing issues. Personalities messing with things is separate and we should work to have a solid base value. Can you set up a baseline game of generic Civs and see how they behave instead?

Fundamentally, do you feel strongly enough that you want to veto this PR? Or can we pull this and work on other improvements later?

EmperorPinguin:
> Then I'd argue you're mixing issues. Personalities messing with things is separate and we should work to have a solid base value. Can you set up a baseline game of generic Civs and see how they behave instead?

With these values I ran 5 games (20 data points) from the same t0 save I mentioned before (with personalities), and got an average t80 pop of 93.7. The last 'benchmark' was 99.9 t80 pop https://github.com/yairm210/Unciv/pull/12582#issuecomment-2544000349. That was on a different version of Unciv, but is was a gain compared to 94.7 t80 pop ceteris paribus. In a test before that, I got 95.8 t80 pop. While 5 games isn't statistically significant, all these are higher than 93.7 which makes me suspicious, and my other experiments of lowering food value multiplier also resulted in lower population.

Note that somewhat lowering food value itself doesn't logically lead to lower pop, as there should be more production available to construct growth infrastructure (settlers, workers, granaries, water mills, aqueducts etc). 

How generic civs behave should follow from the values you set, so (2× food value) > (2 production) > (1.5 × food value) > (1 production + 1 gold) > (1× food value) > (1 production). I'll see if I can set up a test with generic civs tomorrow, maybe we should automate this to get statistically significant results.

PR #12586, opened 2024-12-06
Update Croatian translation

metablaster:


metablaster:
Check failed because master was in failed state, please re-run checks.

PR #12615, opened 2024-12-08
Update Buildings.json

SupermodUnCiv:


yairm210:
Not a Civ V building, you can create a mod if you want to add extra things to the ruleset 
https://yairm210.github.io/Unciv/Modders/Mods/

PR #12630, opened 2024-12-11
Update Dutch.properties

EmperorPinguin:


EmperorPinguin:
This translation is rather inconsistent in naming things, is its supposed to be "Werker" or "Arbeider"?

PR #12631, opened 2024-12-12
Increase starting Luxury amount to match Civ 5

SeventhM:
A quick glance at a few Civ 5 maps implies that we should be guaranteed 2 of our assigned luxury in our capital normally and 3 with legendary start on. I did make sure to test this on base Civ 5 Brave New World maps, not just Lekmod. Unciv currently only guarantees 1 of our assigned luxury in our capital and 2 on legendary start

### DIFF
--- a/core/src/com/unciv/logic/map/mapgenerator/mapregions/LuxuryResourcePlacementLogic.kt
+++ b/core/src/com/unciv/logic/map/mapgenerator/mapregions/LuxuryResourcePlacementLogic.kt
@@ -366,7 +366,7 @@ object LuxuryResourcePlacementLogic {
         val averageFertilityDensity =
             regions.sumOf { it.totalFertility } / regions.sumOf { it.tiles.size }.toFloat()
         for (region in regions) {
-            var targetLuxuries = 1
+            var targetLuxuries = 2
             if (tileMap.mapParameters.getLegendaryStart())
                 targetLuxuries++
             if (region.totalFertility / region.tiles.size.toFloat() < averageFertilityDensity) {


yairm210 (approved):

EmperorPinguin:
I've got a question: is this modified by number of players on the map? I get duplicate luxuries in duels as well, even though trading luxuries gives no net benefit in such settings. In my last Legendary start game, there were 8 players. I had 5 or 6 copies of sugar regional luxury, while another player had only 2 or 3 copies of their regional luxury. Also, I wasn't the only one with sugar, there were players who got it as well, which left me with a bunch of unsellable sugar in the end. Intuitively. it would make sense if there were up to n-1 copies of regional luxuries in a game with n players, and not have this regional luxury shared with other players.

SeventhM:
> I've got a question: is this modified by number of players on the map?

No, this is a flat number. I don't really think this discussion really would matter too much in a duel situation anyways. This does not affect the number of unique luxuries you get in general (it will be guaranteed 2 your assigned regional and a different one) both before and after this PR. If anything, I would argue that despite the lack of impact this PR would still be beneficial to some mods given the additional tile stats, unless that mod already is spawning too many things next to your starting capital

Keep in mind, this PR is \*only\* for spawning your assigned luxury within the first 3 rings of your starting spawn location. It has no other relation to luxury spawn rates

> Intuitively. it would make sense if there were up to n-1 copies of regional luxuries in a game with n players

Ehhh... I think here would be an argument of intent. With this PR, the main intent is to match Civ 5, which likely gave you 2 of your regional to encourage trading without some downside. If we're arguing an intentional deviation, then I think we would have to have a reasonable matching intent. Here, it is totally possible and reasonable for another player to place their expands within your starting region. There should also maybe be an expectation that depending on the map, you may only truly interact with a certain number of players and/or you will expand to more resources of your regional. So focusing hard on matching with the player number probably wouldn't be correct regardless

> and not have this regional luxury duplicated across players

This probably should be addressed. Out of scope here

PR #12636, opened 2024-12-13
Reduce AI personality effect on citizen assignment

EmperorPinguin:
@itanasi what do you think of this?

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -123,7 +123,7 @@ object Automation {
 
             // Zero out Growth if close to Unhappiness limit as well
             val baseFocusWeight = if (city.civ.getHappiness() < -8) 0 else {
-                if (cityAIFocus in CityFocus.zeroFoodFocuses) 1 else 1.5f
+                if (cityAIFocus in CityFocus.zeroFoodFocuses) 1f else 1.5f
             }
             yieldStats.food += newGrowthFood * foodBaseWeight * baseFocusWeight
         }
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -121,9 +121,9 @@ object Automation {
             // but setting such by default worsens AI civ citizen assignment,
             // probably due to badly configured personalities not properly weighing food vs non-food yields
 
-            // Zero out Growth if close to Unhappiness limit as well
+             // Zero out Growth if close to Unhappiness limit as well
             val baseFocusWeight = if (city.civ.getHappiness() < -8) 0 else {
-                if (cityAIFocus in CityFocus.zeroFoodFocuses) 1 else 2
+                if (cityAIFocus in CityFocus.zeroFoodFocuses) 1 else 1.5f
             }
             yieldStats.food += newGrowthFood * foodBaseWeight * baseFocusWeight
         }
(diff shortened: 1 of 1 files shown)


EmperorPinguin:
Why is it failing though, this ran fine on my device, wait a moment.

itanasi:
So this just forces the weights to only span 0.5 to 1.5 instead of 0 to 2? While it's good we avoid weighing something at 0, are we just smoothing everyone back into Generic Civ?

EmperorPinguin:
> So this just forces the weights to only span 0.5 to 1.5 instead of 0 to 2? While it's good we avoid weighing something at 0, are we just smoothing everyone back into Generic Civ?

Yes. Generic civ is better than most  of these personalities.

itanasi:
My original idea I brainstormed was that we could range from 0.5 to 2. So any value below 5 scales at 0.1 step, and above is 0.2 step. That way the full range of Personalities is either half or double. Helps avoid the 0 weight issue.

Yeah I know you're annoyed Personalities effectively make some of the AI more exploitable/suboptimal. Perhaps we need a mode that turns off Leader Personalities. Or you can try manually optimizing the weights to take advantage of the Uniques. Sounds tedious honestly. But part of the reason for Personalities is to give the AI more contours in behavior so they stand out and players learn how to exploit them. Like Nuclear Ghandi or that Zulu hates you.

EmperorPinguin:
> My original idea I brainstormed was that we could range from 0.5 to 2. So any value below 5 scales at 0.1 step, and above is 0.2 step. That way the full range of Personalities is either half or double. Helps avoid the 0 weight issue.

Yes, but half or double food value is not a sensible idea. It's very discrete: going 0.1 below the optimal value, and you could screw over the assignment of every citizen in every city.

itanasi:
Your current solution still allows halving the weight, right? I'm just noting that because this is a multiplication, the two edge cases should be half and double in my mind. Right now you've changed the range to halve and +50%. Which is a linear percentage, but unbalanced as a modifier.

itanasi:
Anyways, I'm fine with this change. Just want to make sure we've thought it through and slightly prefer 0.5-2 as the range since these are multiplicative weights.

On a different note, where did the Personality Weights come from originally?

EmperorPinguin:
> Your current solution still allows halving the weight, right? I'm just noting that because this is a multiplication, the two edge cases should be half and double in my mind. Right now you've changed the range to halve and +50%. Which is a linear percentage, but unbalanced as a modifier.

You're right, I misread the math on that. I think that needs to be fixed in a separe PR (it looks like the range ought to be [0.67, 1.5] instead of [0.5, 1.5])

EmperorPinguin:
> On a different note, where did the Personality Weights come from originally?

You could look up the PR for it, I think they're designed to mimick Civ5 behaviour.

itanasi:
Your current code will make a Personality=0 become weight=0.5, and Personality=10 become weight=1.5. Therefore [0.5, 1.5].

If you intended [0.5, 2] then we probably need a separate function. Linear percentages vs exponential percentages either can work. Generally Linear is better if you're adding vs exponential for multiplying.

itanasi:
Right, but in a typical game you have a spread of tile yields. So eventually you chew through all the 2Prod tiles. Obviously there's edge cases involved, but my point is at the median case you can get around this.

Alternatively, we could have the ranking target +2 Growth as the floor instead of 0 for the Starving logic so that we don't get in Zero Growth cases as easily. And override with Avoid Growth for player obv.

EmperorPinguin:
> Alternatively, we could have the ranking target +2 Growth as the floor instead of 0 for the Starving logic so that we don't get in Zero Growth cases as easily. And override with Avoid Growth for player obv.

That used to be the case. It was removed as it led to AI working 1 food flat tundra tiles, instead of putting the citizen to more productive use on a 4 production mine or a 5 science specialist.

itanasi:
This curve will fit (0, 2/3), (5,1), (10,1.5)
0.666667 + 0.05 x + 0.00333333 x^2

PR #12642, opened 2024-12-14
Add unified unique for gaining stats or stockpiles

SeventhM:
Closes #12577, closes #12578, closes #12579

This PR more or less is a separate version of #12579, but rather than adding golden age points directly, it adds it to a separate class for substats

Pros:
- Should be relatively simple to implement elsewhere without needing too much logic
- Somewhat avoids definition questions of "what is a stockpile"
- Puts defined substats into an Enum

Cons:
- This would require making new versions of every unique that wants to use this new structure. I've only added the one unique as an example, but I still need a specialized version for combat, just like #12577
- The unique parameter name of "stockpile" is still somewhat ill-fitting, especially for stats
- The name for "Golden Age Points" is still incorrect. Almost likewise for stored food ~~or total culture~~

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -17,12 +17,15 @@ import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.RoadStatus
 import com.unciv.logic.map.tile.Tile
 import com.unciv.models.ruleset.Building
+import com.unciv.models.ruleset.tile.TileResource
 import com.unciv.models.ruleset.unique.StateForConditionals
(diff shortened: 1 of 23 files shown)


yairm210:
Okay
I'm closing #12578, and as far as I'm concerned both this and #12577 are viable
I'll merge whichever one you think is the better long-term solution

SeventhM:
> I'll merge whichever one you think is the better long-term solution

Personal opinion, the cons of both are basically identical, and any sort of fixes to one could easily be applied to the other. Long term, I think it's extremely likely we start requesting more and more uniques for stuff to function like all of the stat uniques, so this has a much better long term future

yairm210 (approved):

PR #12650, opened 2024-12-14
UI: standardised alignment and creation of options checkboxes, Unciv-themed slider buttons, multiplayer settings server text field layout and spacing

Toxile:
## [1]

Commits labelled [1] replace the hard-coded 24 font size in some files with the defined constant `headingFontSize`, which, based on the name, was intended for these.  Nothing else of note.

## [2]

The commit labelled [2] deals with the server options in the multiplayer tab.

The "check connection to server" button has been trimmed to say "check connection"; it was already plenty obvious what it was referring to.  I've also moved it next to the server text field: displaying action buttons inline is an established design pattern that improves contextual clarity, and there's plenty of space for the text field.  Other minor changes include adding some white-space below the text to mitigate the cluttered appearance.

### Old
image
### New
image

## [3]

Commits labelled [3] standardise the display of the checkboxes.  Previously, they were displayed either centered or left-aligned at random, some tabs using the former, others defaulting to the latter.  The left-alignment is a far better choice, because it helps group the information together visually.  In addition, I've made them use the `addCheckbox` function of OptionsPopup instead of the auxiliary `String.toCheckBox`.  By the way, I'm not sure they should be capitalised differently: "Checkbox" vs. "CheckBox", but I won't change that lest I break anything else.

### Old
image
### New
image

## [4]

The commit labelled [4] applies an Unciv-like blue colour to the slider buttons instead of the white one.  I feel that this design matches the other UI elements like the checkboxes and the buttons more without compromising usability or clarity.  Nothing much else to add here.

### Old
image
### New
image

### DIFF
--- a/core/src/com/unciv/ui/popups/options/MultiplayerTab.kt
+++ b/core/src/com/unciv/ui/popups/options/MultiplayerTab.kt
@@ -5,6 +5,7 @@ import com.badlogic.gdx.Gdx
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.scenes.scene2d.ui.TextField
 import com.unciv.UncivGame
+import com.unciv.Constants
 import com.unciv.logic.files.IMediaFinder
 import com.unciv.logic.multiplayer.Multiplayer
 import com.unciv.logic.multiplayer.storage.FileStorageRateLimitReached
@@ -101,7 +102,7 @@ private fun addMultiplayerServerOptions(
 ) {
     val settings = optionsPopup.settings
 
-    val connectionToServerButton = "Check connection to server".toTextButton()
+    val connectionToServerButton = "Check connection".toTextButton()
 
     val textToShowForOnlineMultiplayerAddress = if (Multiplayer.usesCustomServer()) {
         settings.multiplayer.server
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/ui/popups/options/MultiplayerTab.kt
```diff
@@ -101,7 +102,7 @@ private fun addMultiplayerServerOptions(
 ) {
     val settings = optionsPopup.settings
 
-    val connectionToServerButton = "Check connection to server".toTextButton()
+    val connectionToServerButton = "Check connection".toTextButton()
```
yairm210, line 105: This requires adding this text in template.properties, so it'll be available for translators 👍🏿

yairm210 (approved):
Looks good!

core/src/com/unciv/ui/popups/options/MultiplayerTab.kt
Toxile: [Is this all right?](https://github.com/yairm210/Unciv/pull/12650/commits/b476fabc59021ca7dbe5969245d5badce3501734) Guessing the original string doesn't need be kept since it's not used elsewhere in the game itself.

PR #12660, opened 2024-12-15
Use unit state for trigger conditionals for gaining new units

SeventhM:
There's totally nothing suspicious here


yairm210 (approved):
I'm not sure what your comment is hinting at, this looks fine...?

SeventhM:
So this is fine, but there's actually some mod modder nonsense reason why I noticed this

Uniques such as ``Gain a free [building] [in this city] <upon founding a city>`` are known to work just fine despite kinda poor wording because founding a city actually provides the city information to the trigger. This was true even before we started assuming the city we wanted was the tile that we were working with

Turns out, despite warnings that it's not allowed, ``[This Unit] gains [amount] XP <upon gaining a [all] unit>`` currently functions. However, this leads to a weird scenario where ``<for [civFilter] units>`` doesn't work (this checks for unit information since it's actually a mapUnitFilter, but ``<for [civFilter] Civilizations>`` does work (we used ``civ.getMatchingUniques`` which by default includes the civ's state)

PR #12662, opened 2024-12-16
Rebalance Food Weights

itanasi:
Rework the weight of Food evaluation for Citizen Management.
Generic AI Civ performance improves and this matches closer to human expectations. Some AIs such as Korea will perform a bit worse in certain scenarios due to Personalities, but we should fix the Personalities instead of break base behavior.
Still room for improvements in corner cases.
Discussion and data in Discord thread:
https://discord.com/channels/586194543280390151/1306726293102006453/1306726299884195932

Also adding a small feature to the console sims

Cleaner rebase of #12582

@EmperorPinguin redoing here

### DIFF
--- a/core/src/com/unciv/logic/automation/Automation.kt
+++ b/core/src/com/unciv/logic/automation/Automation.kt
@@ -46,7 +46,34 @@ object Automation {
         return rank
     }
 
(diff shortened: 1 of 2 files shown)


EmperorPinguin:
I approve this. I don't really agree on the values chosen here, but I think it is good to have them treated separately for AI and humans.

PR #12663, opened 2024-12-16
Simulation: Save Stat support

itanasi:
Add support to pause on a particular turn and save a stat for benchmarking

### DIFF
--- a/core/src/com/unciv/logic/simulation/Simulation.kt
+++ b/core/src/com/unciv/logic/simulation/Simulation.kt
@@ -20,7 +20,8 @@ class Simulation(
     private val newGameInfo: GameInfo,
     val simulationsPerThread: Int = 1,
     private val threadsNumber: Int = 1,
-    private val maxTurns: Int = 500
+    private val maxTurns: Int = 500,
+    private val statTurn: Int = 100
 ) {
     private val maxSimulations = threadsNumber * simulationsPerThread
     val civilizations = newGameInfo.civilizations.filter { it.civName != Constants.spectator }.map { it.civName }
@@ -28,6 +29,7 @@ class Simulation(
     private var startTime: Long = 0
     var steps = ArrayList<SimulationStep>()
     var numWins = mutableMapOf<String, MutableInt>()
+    var avgStat = mutableMapOf<String, MutableInt>()
     private var winRateByVictory = HashMap<String, MutableMap<String, MutableInt>>()
     private var winTurnByVictory = HashMap<String, MutableMap<String, MutableInt>>()
     private var avgSpeed = 0f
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/simulation/Simulation.kt
```diff
@@ -92,6 +101,16 @@ class Simulation(
         println("Simulation step ($stepCounter/$maxSimulations)")
     }
 
+    @Synchronized
+    fun saveStat(gameInfo: GameInfo) {
+        for (civ in gameInfo.civilizations.filter { it.civName != Constants.spectator }) {
+            var popsum = 0
+            civ.cities.forEach { city -> popsum += city.population.population }
+            //println("$civ $popsum")
+            avgStat[civ.civName]!!.set(avgStat[civ.civName]!!.get() + popsum)
```
yairm210, line 110: Since this saves the sum and not the average, please call it an appropriate babe

core/src/com/unciv/logic/simulation/Simulation.kt
```diff
@@ -92,6 +101,16 @@ class Simulation(
         println("Simulation step ($stepCounter/$maxSimulations)")
     }
 
+    @Synchronized
+    fun saveStat(gameInfo: GameInfo) {
+        for (civ in gameInfo.civilizations.filter { it.civName != Constants.spectator }) {
+            var popsum = 0
+            civ.cities.forEach { city -> popsum += city.population.population }
```
yairm210, line 108: Civ.cities.sumBy

yairm210:
I really don't think that "average stat" is actually an interesting thing to track.
Average stat *at turn X, at turn Y* is

core/src/com/unciv/logic/simulation/Simulation.kt
itanasi: Fixed

core/src/com/unciv/logic/simulation/Simulation.kt
itanasi: Fixed

core/src/com/unciv/logic/simulation/Simulation.kt
```diff
@@ -94,6 +109,16 @@ class Simulation(
         println("Simulation step ($stepCounter/$maxSimulations)")
     }
 
+    @Synchronized
+    fun saveStat(gameInfo: GameInfo) {
+        val turn = gameInfo.turns
+        for (civ in gameInfo.civilizations.filter { it.civName != Constants.spectator }) {
+            val popsum = civ.cities.sumOf { it.population.population }
+            //println("$civ $popsum")
+            sumStat[civ.civName]!![turn]!!.set(sumStat[civ.civName]!![turn]!!.get() + popsum)
```
yairm210, line 118: doesn't MutableInt have an add() function?

itanasi: Nope! Just inc()

itanasi: Now it does

yairm210 (approved):

PR #12672, opened 2024-12-16
Add more Spanish Translations and ad Shavian (Shaw) language.

JTech-Labs:
I added a few more lines to the Spanish.properties file.
I also added a new translation target for the [Shavian](https://shavian.info) (Shaw) alphabet.


JTech-Labs:
I'm not sure why but the same problem as in issue #12575 occurs when I try to run it.

PR #12673, opened 2024-12-17
UI: various improvements mostly relating to centering and WorldScreenTopBar

Toxile:
## Notable changes
- reuse selectedCivIcon centering for IconTextButtons,
- standardise top bar icon–text order,
- replace parenthesised per-turn text with a new style,
- improve spacing of the turn count label
- improve spacing of the selectedCivIcon area
- reduce size of the hamburger menu
- add turn count icon to the culture stat label

### Old:
image
image

### New:
image
image

### reuse selectedCivIcon centering for IconTextButtons
This is not perfect, but it aligns text with icons much better than the default system of alignment.  Effect is particularly noticeable when both the icon and text are fairly small.

### standardise top bar icon–text order
Happiness being the only one where the icon comes first is very odd and inconsistent.  Icon-before-text is a much more idiomatic order for simple labels that are intended to communicate how much of something there is.

### replace parenthesised per-turn text with a new style
This one might be controversial, but I prefer this.  Removing the parentheses reduces cognitive load and makes it easier to find the information the player is looking for, and overall makes the section less cluttered.

### improve spacing of the turn count label
Self-explanatory.  The icon was clearly too close to the text.

### improve spacing of the selectedCivIcon area
Self-explanatory.  The icon was clearly too close to the text.

### reduce size of the hamburger menu
I understand the need for it to be easily accessible on mobile, but it was comically large before in my opinion.  Menu buttons like this are usually no larger than the "main" element of the header; usually a brand logo or similar, but in this case, the current civilisation's icon reflects a similar purpose.

### add turn count icon to the culture stat label
Might be controversial.  However, the purpose of a parenthesised number with no context can be difficult to guess, especially for new players.  I think it's plenty worth it to sacrifice some space in the name of clear communication.

### DIFF
--- a/core/src/com/unciv/ui/components/fonts/Fonts.kt
+++ b/core/src/com/unciv/ui/components/fonts/Fonts.kt
@@ -71,7 +71,7 @@ object Fonts {
             descent / height }
         // For whatever reason, undershooting the adjustment slightly
         // causes rounding to work better
-        return ratio * fontSize.toFloat() + 0.10f
+        return ratio * fontSize.toFloat() + 2.25f
     }
 
     /**
--- a/core/src/com/unciv/ui/components/fonts/Fonts.kt
+++ b/core/src/com/unciv/ui/components/fonts/Fonts.kt
@@ -58,9 +58,20 @@ object Fonts {
             .sortedWith(compareBy(UncivGame.Current.settings.getCollatorFromLocale()) { it.localName })
     }
 
+    /**
+     * Helper for v-centering the text of Icon – Label -type components:
+     *
+     * Normal vertical centering uses the entire font height. In reality,
+     * it is customary to align the centre from the baseline to the ascent
+     * with the centre of the other element.  This function estimates the
+     * correct amount to shift the text element.
+     */
     fun getDescenderHeight(fontSize: Int): Float {
-        val ratio = font.run { getDescent() / (getAscent() + getDescent() + getCapHeight()) }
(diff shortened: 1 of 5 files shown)


yairm210:
Looks good!!

yairm210 (approved):

SomeTroglodyte:
Having a little trouble with this, namely the replaced blanks.
- I've seen a box where those blanks rendered as crossed rectangles, but can't remember where
- The diff here, when copied into, say, xed, is identical before/after: xed will translate the new ones to ascii
- These blanks render quite weird (or is it wierd? What's wieirder? wyrd or wired?) in my Studio: image

I'd say:
- Use an escape for the unicode code point in source to be clear (the xed mention above was an attempt to find out which codepoint)
- Consider:
    - Some option to turn "special blanks" use off?
    - Or do a test render and check the pixels for a lot of white, then automatically turn these off? Would happen once in our char-to-glyph caching IIRC.
    - Of course only if we find a box where the "bad" render reproduces...

Ghex says those are UTF-8 E2 80 84 (Can't show a cropped ghex screenie, gimp crashes - and they refuse me in their bugtracker), and duck.ai says that's an em-dash. Can't be. Again - open externally in xed (not copypasta), save as UCS-16: **They're [U+2004](https://unicodeplus.com/U+2004)**, "three-per-em space", that's better.

Edit: The culture turns display uses [U+2009](https://unicodeplus.com/U+2009) "thin space" as well.

PR #12675, opened 2024-12-17
Simulation: More Robust Stats

itanasi:
Save the information in each step object so we can run more robust stats post-processing on it

### DIFF
--- a/core/src/com/unciv/logic/simulation/Simulation.kt
+++ b/core/src/com/unciv/logic/simulation/Simulation.kt
@@ -29,9 +29,7 @@ class Simulation(
     private var startTime: Long = 0
     var steps = ArrayList<SimulationStep>()
     var numWins = mutableMapOf<String, MutableInt>()
-    private var summaryStats = HashMap<String, HashMap<Int, HashMap<String, MutableInt>>>() // [civ][turn][stat]=value
-    private val keySum = "sum"
-    private val keyNum = "num"
+    private var summaryStats = HashMap<String, HashMap<Int, HashMap<Stat, MutableInt>>>() // [civ][turn][stat]=value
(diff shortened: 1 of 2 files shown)


itanasi:
Still feels a bit hacky so any suggestions appreciated

core/src/com/unciv/logic/simulation/Simulation.kt
```diff
@@ -30,6 +30,8 @@ class Simulation(
     var steps = ArrayList<SimulationStep>()
     var numWins = mutableMapOf<String, MutableInt>()
     private var summaryStats = HashMap<String, HashMap<Int, HashMap<String, MutableInt>>>() // [civ][turn][stat]=value
+    private val keySum = "sum"
```
yairm210, line 33: At this point, why not make this an enum?

itanasi: Yeah, good idea.

EmperorPinguin:
How will this deal with CS stats? Currently the population counts doesn't work you add city states to the simulation, it'd be better if these are ignored.

itanasi:
> How will this deal with CS stats? Currently the population counts doesn't work you add city states to the simulation, it'd be better if these are ignored.

Right now it records the CS civ info, but since the final Prints are only whoever wins (which by definition Major Civs), they're never printed.

Do you want the stats for CS printed too?

EmperorPinguin:
> Do you want the stats for CS printed too?

No, I don't need those stats, it just crashed when I added in city states when also enabling the stat collection. I'll see if I can reproduce it tomorrow.

itanasi:
Anyways, this is fine to pull unless someone has more changes. I tried running with CS and it didn't crash in this new version.

Issue #12679, opened 2024-12-18
Improvement build times with improvement build time increases are incorrect


SeventhM:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.14.15

### Describe the bug

While looking into the build times for AI difficulties, I noticed that apparently we've been calculating build times inaccurately work things such as the Citizenship policy in our Gods and Kings ruleset. I've identified 3 main differences between how we calculate improvement build times and how Civ 5 does so

1. Civ 5 has a bug where the build time for every improvement is 1 less than it should be. E.g., while the mine claims that it should take 7 times to build, it actually only takes 6 turns. I'm not sure if this is just last minute intentional design from Firaxis or just a bug
2. Time left to build an improvement is based on the amount progress we've had so far. E.g., In Unciv, if you start building a 20 turn improvement, it will always take 20 turns to finish building the improvement, even if some other unique increases the speed to build improvements. In Civ 5, such a improvement building benefit would take place immediately. If a building suddenly left you build all improvements immediately, it doesn't matter if/when you started building a 20 turn improvement, it now will completely at the end of this turn
3. ~~Most importantly to why I created this issue: Our current improvement time unique is a multiplier on the final amount of turns to build an improvement. E.g., "[-50]% construction time for [All] improvements" will have all improvements build half as fast. In Civ 5, these sort of increases are an improvement to the amount of progress towards an improvement we would have in a turn E.g., a 50% increase will increase our build speed by 1.5x. To get the equivalent benefit as our unique, it would instead need to be a 100% increase. Because we have our Citizenship policy as 25%, just like Civ 5, it necessarily means that the benefit is much high than it actually should be, particularly on slower game speeds~~ solved by #12680

SeventhM:
An aside, I don't think we should match point one, but Imo there should be a discussion of how to implement point 2. The main issue I see is that it's not entirely obvious to players in some cases that the bonus is immediately available, which can lead to awkward situations where it's optimal to stop building an improvement and then restart construction to make use of an improvement speed boost and there's nothing UI wise to indicate as such

PR #12680, opened 2024-12-18
Add unique for increased improvement rates rather than decreased build times

SeventhM:
See #12679 for reasoning. I have no clue the correct PR name for this


yairm210:
- all should be lowercase
- +25, it's *relative*

SeventhM:
> * +25, it's _relative_

On that note, I half wonder if I should change the wording from "at a +25% rate" to "+25% faster" for a better wording

yairm210:
Current text allows for slower as well as faster, which is good

PR #12689, opened 2024-12-21
Allow stat from battle uniques to also give stockpiles

SeventhM:
Also add some helpers to GameResource and SubStat to make it easier to work with

### DIFF
--- a/core/src/com/unciv/logic/battle/Battle.kt
+++ b/core/src/com/unciv/logic/battle/Battle.kt
@@ -369,26 +369,29 @@ object Battle {
         val plunderedGoods = Stats()
 
         for (unique in plunderingUnit.unit.getMatchingUniques(UniqueType.DamageUnitsPlunder, checkCivInfoUniques = true)) {
-            if (plunderedUnit.matchesFilter(unique.params[1])) {
-                val percentage = unique.params[0].toFloat()
-                val amount = percentage / 100f * damageDealt
-                val resourceName = unique.params[2]
-                val resource = plunderedUnit.getCivInfo().gameInfo.ruleset.getGameResource(resourceName)
-                    ?: continue
-                if (resource is Stat) plunderedGoods.add(resource, amount)
-                else {
-                    val plunderedAmount = amount.roundToInt()
-                    civ.addGameResource(resource, plunderedAmount)
-                    val icon = if (resource is SubStat) resource.icon else "ResourceIcons/$resourceName"
-                    civ.addNotification(
-                        "Your [${plunderingUnit.getName()}] plundered [${plunderedAmount}] [${resourceName}] from [${plunderedUnit.getName()}]",
-                        plunderedUnit.getTile().position,
-                        NotificationCategory.War,
-                        plunderingUnit.getName(), NotificationIcon.War, icon,
-                        if (plunderedUnit is CityCombatant) NotificationIcon.City else plunderedUnit.getName()
-                    )
-                }
+            if (!plunderedUnit.matchesFilter(unique.params[1])) continue
+
+            val percentage = unique.params[0].toFloat()
+            val amount = percentage / 100f * damageDealt
+            val resourceName = unique.params[2]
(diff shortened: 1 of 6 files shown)


core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -362,16 +368,34 @@ object Battle {
...
+                SubStat.safeValueOf(resourceName) ?:
+                plunderedUnit.getCivInfo().gameInfo.ruleset.tileResources[resourceName]
+                if (resource == null) continue
+                if (resource is Stat) plunderedGoods.add(resource, amount)
+                else {
+                    val plunderedAmount = amount.roundToInt()
+                    civ.addGameResource(resource, plunderedAmount)
+                    val icon = if (resource is SubStat) resource.icon else "ResourceIcons/$resourceName"
+                    civ.addNotification(
+                        "Your [${plunderingUnit.getName()}] plundered [${plunderedAmount}] [${resourceName}] from [${plunderedUnit.getName()}]",
```
yairm210, line 389: Do we not notify for stat plundering?

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -362,16 +368,34 @@ object Battle {
...
         for (unique in plunderingUnit.unit.getMatchingUniques(UniqueType.DamageUnitsPlunder, checkCivInfoUniques = true)) {
             if (plunderedUnit.matchesFilter(unique.params[1])) {
                 val percentage = unique.params[0].toFloat()
-                plunderedGoods.add(Stat.valueOf(unique.params[2]), percentage / 100f * damageDealt)
+                val amount = percentage / 100f * damageDealt
+                val resourceName = unique.params[2]
+                val resource = Stat.safeValueOf(resourceName) ?:
+                SubStat.safeValueOf(resourceName) ?:
+                plunderedUnit.getCivInfo().gameInfo.ruleset.tileResources[resourceName]
+                if (resource == null) continue
```
yairm210, line 382: Add empty lines before and after this segment

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -362,16 +368,34 @@ object Battle {
     ) {
         // implementation based on the description of the original civilopedia, see issue #4374
         if (plunderingUnit !is MapUnitCombatant) return
+        val civ = plunderingUnit.getCivInfo()
         val plunderedGoods = Stats()
 
         for (unique in plunderingUnit.unit.getMatchingUniques(UniqueType.DamageUnitsPlunder, checkCivInfoUniques = true)) {
             if (plunderedUnit.matchesFilter(unique.params[1])) {
```
yairm210, line 375: Since the body has gotten big with its own nesting, please reverse this if (if !matches() continue)

core/src/com/unciv/logic/battle/Battle.kt
```diff
@@ -257,8 +259,12 @@ object Battle {
                 if (defeatedUnitYieldSourceType == "Cost") unitCost else unitStr
             val yieldAmount = (yieldTypeSourceAmount * yieldPercent).toInt()
 
-            val stat = Stat.valueOf(unique.params[3])
-            civUnit.getCivInfo().addStat(stat, yieldAmount)
+            val resourceName = unique.params[3]
+            val resource = Stat.safeValueOf(resourceName) ?:
+            SubStat.safeValueOf(resourceName) ?:
+            civUnit.getCivInfo().gameInfo.ruleset.tileResources[resourceName]
+            if (resource == null) continue
```
yairm210, line 266: This is repeated at least 3 times, sounds like it should be a function getResource(resourceName)

yairm210 (changes requested):

core/src/com/unciv/logic/battle/Battle.kt
SeventhM: Notification is a bit lower (line 404)

yairm210 (approved):

PR #12690, opened 2024-12-21
Provide a default unitType for units to gain global uniques

SeventhM:
Currently, there's no way to have global uniques for mods. At best, you have to hand out promotions to units that give whatever global effects you want. While this technically works, not only does it require the promotion system to always handle out the unit (which I think is always doable, but it's possible there's some edge case I'm not aware of), but some modders don't seem to be  aware that this is an option. As such, you get mods that have extremely ugly/ likely unmaintable UnitTypes files[^1]. Pretty sure this comes up occansionally as a mod request[^2]

[^1]: https://github.com/Jiangeshi/Leader-Mission-2-Rising-Power/blob/main/jsons/UnitTypes.json
[^2]: https://github.com/yairm210/Unciv/issues/3242#issuecomment-1813988667

### DIFF
--- a/core/src/com/unciv/models/ruleset/Ruleset.kt
+++ b/core/src/com/unciv/models/ruleset/Ruleset.kt
@@ -118,6 +118,8 @@ class Ruleset {
     val greatGeneralUnits by lazy {
         units.values.filter { it.hasUnique(UniqueType.GreatPersonFromCombat, StateForConditionals.IgnoreConditionals) }
     }
+    
+    val allUnitTypes by lazy { unitTypes["All"] }
 
     val tileRemovals by lazy { tileImprovements.values.filter { it.name.startsWith(Constants.remove) } }
     val nonRoadTileRemovals by lazy { tileRemovals.filter { rulesetImprovement ->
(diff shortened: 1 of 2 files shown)


yairm210:
This is not the correct way to go about this. This adds an extra lookup every time we try to resolve uniques.
Instead, the "all" uniques should be loaded into the uniquemap of each unit type - we pay memory to get speed, that's definitely the right tradeoff

Also, this feature ("all" unittype) is *also* currently undocumented, so as for "modders aren't aware this is an option" - let's document it in .md files

SeventhM:
> Instead, the "all" uniques should be loaded into the uniquemap of each unit type

🤔Hmm... Might need to look into this. 

> This is not the correct way to go about this.

Tbh, I agree. Even ignoring the performance impact, tbh I think this solution is kinda ugly. I guess a better solution would actually be to put these uniques in GlobalUniques as ``var unitUniques = ArrayList<String>()``. Was just not sure a good spot to put the uniques that didn't break the mod structure or introduce an unnecessary extra json

SeventhM:
Closing in favor of #12775

PR #12691, opened 2024-12-21
Set Gzip compression level to `Deflater.BEST_COMPRESSION`

### LINKED ISSUE #12685: Suggestion: Use maximum Gzip compression level (level 9) for uploading jsons to the server.
### Before creating

### Problem Description

The assumption is CPU time required for compression / decompression is much lower than network time required to fetch the files larger files and the tradeoff is generally worth it. Using maximum compression level will most likely not change much. However, I suggest the idea of using this setting. This decrease badwidth and storage usage somewhat.

### Related Issue Links

_No response_

### Desired Solution

use Gzip compression level 9

### Alternative Approaches

use something in the middle

### Additional Context

This mostly increases CPU time on client side. As server only needs to decompress these files on `PUT` requests. Any bandwidth decrease is appreciated.

touhidurrr:
resolves #12685

### DIFF
--- a/core/src/com/unciv/ui/screens/savescreens/Gzip.kt
+++ b/core/src/com/unciv/ui/screens/savescreens/Gzip.kt
@@ -5,7 +5,6 @@ import java.io.BufferedReader
 import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
 import java.io.InputStreamReader
-import java.io.OutputStream
 import java.util.zip.Deflater
 import java.util.zip.GZIPInputStream
 import java.util.zip.GZIPOutputStream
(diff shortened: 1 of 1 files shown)


touhidurrr:
Noob question as I am new to kotlin.
```diff
-        val gzip = GZIPOutputStream(bos)
+        val gzip = object : GZIPOutputStream(bos) {
+            init {
+                def.setLevel(Deflater.BEST_COMPRESSION)
+            }
+        }
```
how does it work?

yairm210:
I was far too hasty merging this
Since we're designed to run on low-end devices, we REALLY need to compare time vs compression for these two options
It's entirely possible we're causing users to wait twice as long for like a 5% decrease in file size, this isn't free real estate

yairm210:
Reverting for the deployment

touhidurrr:
```kt
vall gzip = GZIPOutputStream(bos, Deflater.BEST_COMPRESSION)
```

How did I miss this though... I am pretty sure I read through all `constructors` of `GZIPOutputStream` before writing this. -_-

touhidurrr:
@SomeTroglodyte, read https://github.com/yairm210/Unciv/issues/12685#issuecomment-2558034471. Already did a similar benchmark there for this topic. I covered compression / decompression times and size saving vs level 1 etc in there. Also set iterations to 100 for each test and took the average.

PR #12701, opened 2024-12-23
Amend civilopedia texts.

J0anJosep:
Two sentences missed the full stop at the end.
I think I amended the right file, but not sure.


yairm210:
Done manually across all files

PR #12703, opened 2024-12-24
Cancel the chain reaction of defense pact

### LINKED ISSUE #12564: Defensive pact involve the 4th nations in the war

### Game Version

4.14.12-patch2 (Build 1079)

### Describe the bug

Byzantium has defense pacts with China and Siam. 
I expect that India will fight with Siam and Byzantium when attacking Siam. But for some reason, when an attack is launched, war is declared on the China too. 
Although the goal of the war is Siam, the China joined the chain to Byzantium. 
Such alliances work in a chain, and you can fight with the whole map, although it has nothing to do with it.

### Steps to Reproduce

Declare war on Siam.
file bellow

### Screenshots

image
image

### Link to save file

[test2.txt](https://github.com/user-attachments/files/17998696/test2.txt)

### Operating System

Android

### Additional Information

_No response_

czyh2022:
I noticed the discussion in Discord about whether to cancel the chain reaction of defense pact. Based on this, I made some corresponding changes. Before a civilization calls a defense pact ally, it will check the Type of the war it is participating in. If it is already a DefensivePactWar, it will not continue to call allies.
Resolves #12564

### DIFF
--- a/core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
+++ b/core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
@@ -281,24 +281,15 @@ class DiplomacyScreen(
             otherCivDiploManager -> otherCivDiploManager.otherCiv() != viewingCiv
             && otherCivDiploManager.diplomaticStatus == DiplomaticStatus.DefensivePact
             && !otherCivDiploManager.otherCiv().isAtWarWith(viewingCiv) }
-            .map { it.otherCiv() }.toMutableList()
-            
-            // Defensive pact chains are not allowed now
-            var listIndex = 0
-            while (listIndex < otherCivDefensivePactList.size) {
-                messageLines += if (viewingCiv.knows(otherCivDefensivePactList[listIndex]))
-                    "[${otherCivDefensivePactList[listIndex].civName}] will also join them in the war"
-                else "An unknown civilization will also join them in the war"
-                /*
-                // Add their defensive pact allies
-                otherCivDefensivePactList.addAll(otherCivDefensivePactList[listIndex].diplomacy.values
-                    .filter { diploChain -> diploChain.diplomaticStatus == DiplomaticStatus.DefensivePact
-                        && !otherCivDefensivePactList.contains(diploChain.otherCiv())
-                        && diploChain.otherCiv() != viewingCiv && diploChain.otherCiv() != otherCiv
-                        && !diploChain.otherCiv().isAtWarWith(viewingCiv) }
-                    .map { it.otherCiv() })
-                */
-            listIndex++
+            .map { it.otherCiv() }
+
+        // Defensive pact chains are not allowed now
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
```diff
@@ -282,20 +282,22 @@ class DiplomacyScreen(
...
-            // Add their defensive pact allies
-            otherCivDefensivePactList.addAll(otherCivDefensivePactList[listIndex].diplomacy.values
-                .filter { diploChain -> diploChain.diplomaticStatus == DiplomaticStatus.DefensivePact
-                    && !otherCivDefensivePactList.contains(diploChain.otherCiv())
-                    && diploChain.otherCiv() != viewingCiv && diploChain.otherCiv() != otherCiv
-                    && !diploChain.otherCiv().isAtWarWith(viewingCiv) }
-                .map { it.otherCiv() })
+            
+            // Defensive pact chains are not allowed now
+            var listIndex = 0
```
yairm210, line 287: Is there a reason to have this by index instead of 'for civ in othercovdefensivepactlist'?
I know this was here before you, but if we can improve readability, why not :)

SeventhM: Is there a reason we're doing indexing at all on a list and not using ``for ((index, civ) in otherCivDefensivePactList.withIndexed()``. Like, "withIndexed" is explicitly in the Kotlin library so we don't need to do while loops for this

yairm210: Correct, but now it is no longer modified, so we can move to the more readable format
We don't even use the index anymore, just the civ

core/src/com/unciv/ui/screens/diplomacyscreen/DiplomacyScreen.kt
```diff
@@ -282,20 +282,22 @@ class DiplomacyScreen(
...
+                else "An unknown civilization will also join them in the war"
+                /*
+                // Add their defensive pact allies
+                otherCivDefensivePactList.addAll(otherCivDefensivePactList[listIndex].diplomacy.values
+                    .filter { diploChain -> diploChain.diplomaticStatus == DiplomaticStatus.DefensivePact
+                        && !otherCivDefensivePactList.contains(diploChain.otherCiv())
+                        && diploChain.otherCiv() != viewingCiv && diploChain.otherCiv() != otherCiv
+                        && !diploChain.otherCiv().isAtWarWith(viewingCiv) }
+                    .map { it.otherCiv() })
+                */
```
yairm210, line 300: Also delete this, there's no reason to have dead code in comments, if we want it we can get it back from git :)

czyh2022:
Got it! I removed dead code and simplified the loop. Although I started playing this game several years ago, I am new here to contributing :)

yairm210 (approved):

PR #12710, opened 2024-12-26
Allow for city level stockpiles to actually function

SeventhM:
Currently stockpiles for cities don't actually exist. Previously, I believe (but never checked) that they would function as if they were civwide regardless, which doesn't match up with the clear and obvious intent. During the last few PRs, I've since changed it so that under some circumstances, gaining a citywide stockpile from a city context would do nothing. This PR aims to do 2 things
1. Go the extra mile to try to enforce citywide resources as being "citywide" from these contexts
2. Provide at least the barebones needed for a functioning citywide stockpile

As of writing this PR description, I'm not entirely sure the best way into the resource logic to add citywide stockpile resources, so I haven't immediately focused on adding it

### DIFF
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -210,11 +210,6 @@ class City : IsPartOfGameInfoSerialization, INamed {
 
     fun getGreatPersonPercentageBonus() = GreatPersonPointsBreakdown.getGreatPersonPercentageBonus(this)
     fun getGreatPersonPoints() = GreatPersonPointsBreakdown(this).sum()
-    
-    fun gainStockpiledResource(resourceName: String, amount: Int) {
-        
-        resourceStockpiles.add(resourceName, amount)
-    }
 
     fun gainStockpiledResource(resource: TileResource, amount: Int) {
         if (resource.isCityWide) resourceStockpiles.add(resource.name, amount)
--- a/core/src/com/unciv/logic/city/City.kt
+++ b/core/src/com/unciv/logic/city/City.kt
@@ -16,6 +16,7 @@ import com.unciv.logic.map.TileMap
 import com.unciv.logic.map.mapunit.MapUnit
 import com.unciv.logic.map.tile.RoadStatus
 import com.unciv.logic.map.tile.Tile
+import com.unciv.models.Counter
 import com.unciv.models.ruleset.Building
 import com.unciv.models.ruleset.tile.TileResource
 import com.unciv.models.ruleset.unique.StateForConditionals
@@ -73,6 +74,8 @@ class City : IsPartOfGameInfoSerialization, INamed {
 
     @Transient  // CityStats has no serializable fields
     var cityStats = CityStats(this)
+    
+    var resourceStockpiles = Counter<String>()
 
     /** All tiles that this city controls */
     var tiles = HashSet<Vector2>()
@@ -136,6 +139,7 @@ class City : IsPartOfGameInfoSerialization, INamed {
         toReturn.tiles = tiles
         toReturn.workedTiles = workedTiles
         toReturn.lockedTiles = lockedTiles
+        toReturn.resourceStockpiles = resourceStockpiles.clone()
         toReturn.isBeingRazed = isBeingRazed
         toReturn.attackedThisTurn = attackedThisTurn
         toReturn.foundingCiv = foundingCiv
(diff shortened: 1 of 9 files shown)


yairm210:
Oh wow!
Yes, "city" and "stockpile" are 2 different categories, we don't have "city stockpile", but this is definitely a thing we would want :D

core/src/com/unciv/logic/city/City.kt
```diff
@@ -206,6 +210,15 @@ class City : IsPartOfGameInfoSerialization, INamed {
...
     fun getGreatPersonPoints() = GreatPersonPointsBreakdown(this).sum()
+    
+    fun gainStockpiledResource(resourceName: String, amount: Int) {
+        resourceStockpiles.add(resourceName, amount)
+    }
+
+    fun gainStockpiledResource(resource: TileResource, amount: Int) {
+        if (resource.isCityWide) resourceStockpiles.add(resource.name, amount)
+        else civ.gainStockpiledResource(resource.name, amount)
+    }
```
yairm210, line 221: Very weird that in the first function you don't check if it's citywide or not, you get 2 different behaviours for civwide resources in the 2 identically named functions

core/src/com/unciv/logic/city/City.kt
```diff
@@ -238,11 +250,9 @@ class City : IsPartOfGameInfoSerialization, INamed {
...
 
     fun getReserve(stat: GameResource): Int {
-        if (stat is TileResource && stat.isCityWide) {
-            return if (stat.isStockpiled) {
-                // TODO
-                0
-            } else 0
+        if (stat is TileResource) {
+            if (!stat.isStockpiled) return 0
+            if (stat.isCityWide) return resourceStockpiles[stat.name]
```
yairm210, line 255: What do we do when it's civ-wide? Fall down to "civ.getReserve(stat)" below?
I'd rather this was explicit in the "stat is TileResource", that it become a "return when()" so we see that we treat all TileResources cases exhaustively

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -435,7 +435,8 @@ class CityConstructions : IsPartOfGameInfoSerialization {
         for (unique in costUniques) {
             val amount = unique.params[0].toInt()
             val resourceName = unique.params[1]
-            city.civ.gainStockpiledResource(resourceName, -amount)
+            val resource = city.civ.gameInfo.ruleset.tileResources[resourceName] ?: continue
+            city.gainStockpiledResource(resource, -amount)
```
yairm210, line 439: This, to me, indicates that the gainStockpiledResource(String) should do the resource lookup, and then call the gainStockpiledResource(Resource) function

yairm210:
Interesting question: What happens when a city is conquered / bought / etc?
I think it should retain the stockpiles (as is currently the case) 
If there's a demand for resources that we don't want to keep between civs we can always create a unique for it later

core/src/com/unciv/logic/city/CityConstructions.kt
SeventhM: That's not entirely ideal unless we decide to add additional logic here, since the counter itself only cares about the resource name (for the sake of serialization)

SeventhM: So, if this is what's wanted, the actual thing is to delete the function for strings (I added it for clarity, but the counter is still public rn, actually). Currently, the city version for strings can already be safely deleted

core/src/com/unciv/logic/city/City.kt
SeventhM: I had it like that originally, but wasn't sure if it should be explicit

SeventhM:
> Interesting question: What happens when a city is conquered / bought / etc? I think it should retain the stockpiles (as is currently the case) If there's a demand for resources that we don't want to keep between civs we can always create a unique for it later

Yeah, I would think the default behavior should be to retain the stockpiles

core/src/com/unciv/models/ruleset/tile/TileResource.kt
```diff
@@ -37,8 +37,10 @@ class TileResource : RulesetStatsObject(), GameResource {
     var majorDepositAmount: DepositAmount = DepositAmount()
     var minorDepositAmount: DepositAmount = DepositAmount()
     
+    @delegate:Transient
```
yairm210, line 40: We shouldn't be serializing this in any case, doesn't need this
If it DOES that means we're serializing, which we should definitely get rid of

core/src/com/unciv/logic/city/CityConstructions.kt
yairm210: Yes, let's delete it

yairm210:
This should be mergeable next version, just a few teaks left :)

yairm210 (approved):

PR #12712, opened 2024-12-26
Update AirUnitAutomation.kt

EmperorPinguin:
Won't prevent suicide bombings, but should make it less commonplace by setting a higher health cutoff.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/AirUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/AirUnitAutomation.kt
@@ -13,7 +13,7 @@ import com.unciv.models.ruleset.unique.UniqueType
 object AirUnitAutomation {
 
     fun automateFighter(unit: MapUnit) {
-        if (unit.health <= 50 && !unit.hasUnique(UniqueType.HealsEvenAfterAction)) return // Wait and heal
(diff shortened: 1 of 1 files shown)


yairm210:
Is this how you okay? This seems arbitrary 
If we want to check for possible damage, let's just do that, no?

EmperorPinguin:
> Is this how you okay? This seems arbitrary If we want to check for possible damage, let's just do that, no?

The way to check for this in a multiplayer game is via reloads, as the damage from interception is not shown in the damage preview. It seems there is no current use of such an intercept damage preview by AI (maybe I glanced over it, but the AI deliberately deals with incomplete information, as can be seen from e.g. the assumption on line 24, so I suppose not). Without cheats for AI, there needs to be an approximation, which will work in many but not all cases.

My proposal is to use 74 hp, as then with 1 turn of healing it's at 99 hp, which is the perfect amount for Populism bombers and we don't waste excess turns of healing (as would be the case with a higher health treshold) (the AI isn't super smart about using medics). Intecept damage is typically around 50hp damage, and land unit counterdamage around 20hp (if you're early on bombers, later this can go up to 50 as well, hence the other PR about trying to get Evasion in time).

yairm210 (approved):

PR #12713, opened 2024-12-26
Update UnitPromotions.json

EmperorPinguin:
For evasion, see https://discord.com/channels/586194543280390151/586194543716859916/1319706915206467634

Air Targeting isn't a bad promotion, but in the majority of games, Bombardment is better, and we should stay on that path such as to unlock the powerful higher-tier promotions (Air Repair).


yairm210:
Please don't "-99". If you want it out, -100, otherwise -90.

EmperorPinguin:
> Please don't "-99". If you want it out, -100, otherwise -90.

It's so it prefers it over Besiege.

yairm210 (approved):

PR #12725, opened 2024-12-30
Updated CityStateFunction.tk so that city states, will get angry at you, if you steal their lands 

Emandac:
Before the city state Ai, didn't care if you steal their lands.
But now with this update to the code, they will hate you now.
Except if you are friends with them.

Also modified other files, to have the popup when you steal their lands.


core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
```diff
@@ -540,6 +540,17 @@ class CityStateFunctions(val civInfo: Civilization) {
                     diplomacy.setFlag(DiplomacyFlags.BorderConflict, 10)
                 }
             }
+            
+            // If citystate gets tile stolen from another civ, then the civ gets a diplomacy Influence debuffer.
+            if (diplomacy.hasModifier(DiplomaticModifiers.StealingTerritory) && diplomacy.isRelationshipLevelLT(RelationshipLevel.Friend)) {
+                diplomacy.addInfluence(-25f)
```
yairm210, line 546: This should be in UniqueTriggerActivation:1107 (UniqueType.OneTimeTakeOverTilesInRadius)
This should happen *when* the tile is being stolen
Also, even if you have a good relationship with a CS, stealing tiles is a Bad Action, and should have consequences
I think 15 influence per tile stolen sounds around right

core/src/com/unciv/logic/civilization/diplomacy/CityStateFunctions.kt
```diff
@@ -540,6 +540,17 @@ class CityStateFunctions(val civInfo: Civilization) {
...
+                if (!diplomacy.hasFlag(DiplomacyFlags.BorderOccupation)) {
+                    otherCiv.popupAlerts.add(PopupAlert(AlertType.BorderOccupation, civInfo.civName))
+                    
+                    // to only show popup once.
+                    diplomacy.setFlag(DiplomacyFlags.BorderOccupation, -1)
```
yairm210, line 551: Flag name is not indicative of what it's actually remembering - should be "TilesStolen" or similar, this sounds like you have units on tiles on their borders

Emandac:
I have changed the BorderOccupation name to TilesStolen.
 And also now you get a -15 for each tile stolen.
like you said.

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -1107,6 +1108,16 @@ object UniqueTriggerActivation {
                             otherCiv.getDiplomacyManagerOrMeet(civInfo).addModifier(DiplomaticModifiers.StealingTerritory, -10f)
                             civsToNotify.add(otherCiv)
                         }
+                        // check if civ has steal a tile from a citystate 
+                        if (otherCiv != null && otherCiv.isCityState) {
+                            // create this varibale diplomacyCityState for more readability
+                            val diplomacyCityState = otherCiv.getDiplomacyManager(civInfo)!!
```
yairm210, line 1114: How about getDiplomacyManagerOrMeet, that way we don't run into weird problems if we manage to steal tiles from a civ we haven't met yet

core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
```diff
@@ -118,6 +119,14 @@ class AlertPopup(
         addCloseButton("Sorry.", KeyboardBinding.Confirm)
         addCloseButton("Never!", KeyboardBinding.Cancel)
     }
+    
+    private fun addTilesStolen() {
+        val civInfo = getCiv(popupAlert.value)
+        addLeaderName(civInfo)
+        addGoodSizedLabel("You stole some of our territory!")
+        addCloseButton("Yes.", KeyboardBinding.Confirm)
+        addCloseButton("No.", KeyboardBinding.Cancel)
```
yairm210, line 128: What is this about? Let the player know this was unacceptable, but there's no choice here, 
Something like "Those lands were not yours to take. This has not gone unnoticed."
And then just `addCloseButton()`

Also, this text needs to go in template.properties so it's translatable

Emandac:
Done

core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
```diff
@@ -123,9 +123,8 @@ class AlertPopup(
     private fun addTilesStolen() {
         val civInfo = getCiv(popupAlert.value)
         addLeaderName(civInfo)
-        addGoodSizedLabel("You stole some of our territory!")
-        addCloseButton("Yes.", KeyboardBinding.Confirm)
-        addCloseButton("No.", KeyboardBinding.Cancel)
+        addGoodSizedLabel("Those lands were not yours to take. This has not gone unnoticed")
```
yairm210, line 126: With a period on the end :)
Also in the properties file

Emandac: Done

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -1107,6 +1108,16 @@ object UniqueTriggerActivation {
                             otherCiv.getDiplomacyManagerOrMeet(civInfo).addModifier(DiplomaticModifiers.StealingTerritory, -10f)
                             civsToNotify.add(otherCiv)
                         }
+                        // check if civ has steal a tile from a citystate 
+                        if (otherCiv != null && otherCiv.isCityState) {
+                            // create this varibale diplomacyCityState for more readability
+                            val diplomacyCityState = otherCiv.getDiplomacyManagerOrMeet(civInfo)
+                            diplomacyCityState.addInfluence(-15f)
+
+                            if (!diplomacyCityState.hasFlag(DiplomacyFlags.TilesStolen)) {
```
yairm210, line 1117: And you need to add the flag, because currently it will pop up every time
I'm not sure this is a bad thing tbh 🤔  Maybe we want to pop up every time?Currently this line just will never be true

Emandac: You are right, I am going to update it to add the flag.
And that explained why I was getting 1 popup per Tile stolen, Now after adding the flag i only get one pop 😅

yairm210 (approved):

core/src/com/unciv/ui/screens/worldscreen/AlertPopup.kt
Emandac, line 1: Make so that you don't get declare war option, when you are already at war with that civ

SeventhM:
> Small side note "after the fact": Nudge nudge white-space. This PR introduced several lines consisting only of blanks, and there's a block of 3 empty lines in the middle of a function. Some recent @SeventhM PR's too. White-space changes are still changes to git, so inconsistent handling _can_ lead to unnecessarily increased merge troubles. I thought the policy was no trailing white-space except where needed (templates) - "Do as you wont but never change another's insignificant white-space" could work too, but it needs consensus.

I'm not entirely sure what you're trying to say here. My general rule of thumb has been:
1. Avoid trailing whitespace at the end of lines (usually for newlines that Studio hasn't cleaned up automatically)
2. Delete trailing whitespace at the end of lines if it's code I'm already touching for other reasons (which imo should usually be consistent with point 1)
3. Keep all/most other whitespace the same, including lines between statements. Most of the lines between statements is intentional to help separate obvious code blocks. And, like, in another PR I did, I was explicitly requested to add in such a line

The only time I delete lines between statements/functions is if there's two lines instead of one accidentally

PR #12728, opened 2024-12-31
Let AI choose healing promotions

EmperorPinguin:
It is currently filtering out the healing promotion from consideration it seems, even though it can be the best choice.

Note that 60hp treshold is a bit arbitrary, it may be worthwhile to test with a higher or lower treshold.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
@@ -181,7 +181,9 @@ object UnitAutomation {
             // Restrict Human automated units from promotions via setting
                 (UncivGame.Current.settings.automatedUnitsChoosePromotions || unit.civ.isAI())) {
             val availablePromotions = unit.promotions.getAvailablePromotions()
-                .filterNot { it.hasUnique(UniqueType.SkipPromotion) }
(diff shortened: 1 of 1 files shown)


EmperorPinguin:
Simulation data:
Screenshot 2024-12-31 132923
Screenshot 2024-12-31 123433

core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -181,7 +181,9 @@ object UnitAutomation {
             // Restrict Human automated units from promotions via setting
                 (UncivGame.Current.settings.automatedUnitsChoosePromotions || unit.civ.isAI())) {
             val availablePromotions = unit.promotions.getAvailablePromotions()
-                .filterNot { it.hasUnique(UniqueType.SkipPromotion) }
+            if (unit.health < 60 && !(unit.baseUnit.isAirUnit() || unit.baseUnit.hasUnique(UniqueType.CanMoveAfterAttacking)))
+                availablePromotions.filter { it.hasUnique(UniqueType.OneTimeUnitHeal) } //choose healing promotions only when beneficial
+            else availablePromotions.filterNot { it.hasUnique(UniqueType.SkipPromotion) }
```
yairm210, line 186: What if we don't have any promotion with a one time heal? We should default to all

EmperorPinguin: I suppose this works.

PR #12737, opened 2025-01-02
Improved city screen queue

sulai:
See previous PR #12341.

Improved since then:
- queue preview entries selectable
- can be removed from queue while selected as preview


NahuelGeno:
This improvement to the city building queue is really needed, I'm looking forward to it. The previous PR you mentioned hasn't been implemented yet, right?

yairm210 (approved):
I'm not entirely sure but the preview is definitely good, let's make the change and see what people like / dislike

sulai:
Thank you guys for your positive feedback! 🙂 Some of these changes are 2 months old, I'll double check those tomorrow 👍

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
```diff
@@ -172,24 +193,52 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
...
+        queueExpander.setText(if (queue.size <= 4) "Queue".tr() else "")
+        queue.forEachIndexed { i, constructionName ->
+            if (i in 1..4) {
+                val color = if (selectedQueueEntry == i) highlightColor else BaseScreen.skinStrings.skinConfig.baseColor
+                val image = ImageGetter.getConstructionPortrait(constructionName, 40f).surroundWithCircle(54f, false, color)
+                image.onClick {
+                    cityScreen.selectConstruction(constructionName)
+                    selectedQueueEntry = i
+                    // hack: keep expander closed (can't stop click event propagation to parent)
+                    queueExpander.toggle()
```
sulai, line 232: This works, but it's a bit hacky. I want the child to receive the click and prevent the parent (queueExpandere.header) from receiving the click in that case. I tried OnClickListener and event.stop(), but it didn't work out: when clicking an item in the header, the header still receives the click and opens the expander. Is there a better way of doing it?

sulai:
I double-checked all changes and I think we now have a decent state. The only thing that boggles me a bit is that hack with `queueExpander.toggle()`, but I don't know any better way of doing it. It works from the user's perspective, but it's a hack.

@NahuelGeno to answer your question, the previous PR wasn't merged because it "timed out". I was lacking time to finish it.

@yairm210 I agree we should listen to the community for further improvements. By introducing the mini-queue, this PR focuses on cramped portrait and cramped landscape android screens. That is, the average android phone people likely have. I added some in-code reasoning as to how it manages vertical space better (Landscape). It also handles horizontal space better in cramped-portrait by moving the priority buttons. There's still room for improvement though since some entry descriptions tend to waste some horizontal space, but let's postpone that.

From my perspective, (if we don't find a better way for that toggle-hack), the PR is ready to merge :smiley: :tada:

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
sulai: I tried this already, but it didn't work, possibly because onClick consists of 2 events (touch down and up)?

```
childActor.addListener(object : ClickListener() {
    override fun clicked(event: InputEvent, x: Float, y: Float) {
        event.stop() // Prevent further propagation of the event
        // Handle click logic for the child actor
        println("Child actor clicked and event propagation stopped!")
    }
})
```

What we could try yet is this:

```
childActor.addListener(object : ClickListener() {
    override fun touchDown(event: InputEvent, x: Float, y: Float, pointer: Int, button: Int): Boolean {
        return true // Handle the event here and stop propagation
    }

    override fun clicked(event: InputEvent, x: Float, y: Float) {
        // Handle click logic for the child actor
        println("Child actor clicked!")
    }
})
```

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
```diff
@@ -95,13 +100,25 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
             "CityScreen/CityConstructionTable/ConstructionsQueueTable",
             tintColor = ImageGetter.CHARCOAL
         )
+        queueExpander = ExpanderTab(
+            "Queue",
```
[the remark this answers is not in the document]
sulai: As documented, it's auto-translated.

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
```diff
@@ -136,7 +153,7 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
         upperTable.pack()
         // Need to reposition when height changes as setPosition's alignment does not persist, it's just a readability shortcut to calculate bottomLeft
         upperTable.setPosition(posFromEdge, stageHeight - posFromEdge, Align.topLeft)
-        lowerTableScrollCell.maxHeight(stageHeight - upperTable.height - 2 * posFromEdge)
+        lowerTableScrollCell.maxHeight(max(20f, stageHeight - upperTable.height - 2 * posFromEdge))
```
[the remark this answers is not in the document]
sulai: I wouldn't mind the imports, but it's easier to read for sure. Thanks for the hint! :+1:

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
```diff
@@ -79,6 +82,8 @@ class CityConstructionsTable(private val cityScreen: CityScreen) {
     private val pad = 10f
     private val posFromEdge = CityScreen.posFromEdge
     private val stageHeight = cityScreen.stage.height
+    
+    private val highlightColor = Color.GREEN.darken(0.3f)
```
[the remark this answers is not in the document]
sulai: I agree it should be part of the skin, but it doesn't look trivial to me to change that. Would you agree with converting this comment into a follow-up issue?

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
sulai: Thanks @SomeTroglodyte for your thorough input.

I solved the event propagation issue by using event.stop() and overriding touchDown() and touchUp(). It seems to me the GDX-way of doing it, even though I feel like they should document that event.stop() won't work in click/(). I wrote an in-code comment about that.

About keeping UI elements around for the lifetime of a screen. I can see where you are coming from, but I also understand that there are very different opinions about that out there. Popular frameworks like React and Jetpack Compose encourage the idea of re-building (parts) of the UI. Note that we're rebuilding the construction queue table from scratch since several years. I'd say let's keep it that way as it has its advantages, since we don't have to do fault prone incremental additions removals of elements in the list.

As for listener leaks: I checked the heap dump and I can confirm they don't leak. I took the opportunity to also check for CityScreen and I can confirm that this one also doesn't leak when the user closes the city screen.

sulai: Sorry for the delayed and incomplete response, I'll try to catch up with the conversation.

> You thoughts on my if(isHandled) patch?

If this saves us from overriding 2 methods in order to stop a click event I like it! Is it better if we tested for event.isStopped? Also not sure if this has side effects, is it used it a lot of places?

I also tried to only override touchDown, but that wasn't sufficient. It had to be both.

sulai: > consider in preview mode whether selecting the new last queue entry after removing the last queue entry with the remove button - the one beside buy and reorder that isn't shown in expanded mode. That way you could continue removing stuff until empty no matter where you started the "selection" without moving the pointer much.

Good idea! It's a small thing, I think we can add it as part of some other PR. 👍

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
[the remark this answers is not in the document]
sulai: This is actually missing, good catch @SomeTroglodyte. Need to add it in a later PR.

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
sulai: Addressed that last comment in #12755.

@SomeTroglodyte if you feel like chatting, I'm also on Discord 👍

core/src/com/unciv/ui/screens/cityscreen/CityConstructionsTable.kt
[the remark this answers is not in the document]
sulai: Added in #12755

PR #12740, opened 2025-01-03
Update Latin translation

sls1005:


NahuelGeno:
I think you should delete the lines that also say "#Requires Translation!" when you already applied a translation

NahuelGeno:
> A translator is free to do so but not required to. The next "generate translation files" run will do it.

I didn't know that, that's great!

PR #12742, opened 2025-01-03
Update CivilianUnitAutomation.kt

EmperorPinguin:
On more crowded maps, the AI has a tendency to have their tiles improved before the great persons start to spawn, and then it can't place academies anymore, which can lead to loss of the unit.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/CivilianUnitAutomation.kt
@@ -79,7 +79,8 @@ object CivilianUnitAutomation {
         val isLateGame = isLateGame(unit.civ)
         // Great scientist -> Hurry research if late game
         // Great writer -> Hurry policy  if late game
-        if (isLateGame) {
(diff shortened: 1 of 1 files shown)


yairm210:
Maybe we should check if there's an enemy unit within like 5 tiles? Otherwise this seems like a huge waste

EmperorPinguin:
> Maybe we should check if there's an enemy unit within like 5 tiles? Otherwise this seems like a huge waste

That would be better, but sounds like it's more computationally expensive, and might not be able to account for attacks with mounted units in all cases.

yairm210:
Great Persons aren't that common, and a 5-tile radius check is much cheaper than the checks we have for e.g. Setters or Workers, which are super common

EmperorPinguin:
> Great Persons aren't that common, and a 5-tile radius check is much cheaper than the checks we have for e.g. Setters or Workers, which are super common

Would it make sense to use a radius check for settlers as well then? Ideally, we should be done settling before there are roads or units with more than 2 movement.

yairm210:
We already have quite complicated radius checks for settlers, that's my point

PR #12754, opened 2025-01-05
Update NextTurnAutomation.kt

EmperorPinguin:
Is there a specific reason why the protectCityStates was designed like this?

### DIFF
--- a/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
+++ b/core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
@@ -200,7 +200,7 @@ object NextTurnAutomation {
         else 0
         value += ourInfluence / 10
 
-        if (civInfo.gold < 100 && ourInfluence < 30) {
+        if (ourInfluence < 30) {
             // Consider bullying for cash
             value -= 5
         }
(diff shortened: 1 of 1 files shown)


core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
```diff
@@ -200,7 +200,7 @@ object NextTurnAutomation {
         else 0
         value += ourInfluence / 10
 
-        if (civInfo.gold < 100 && ourInfluence < 30) {
+        if (ourInfluence < 30) {
```
yairm210, line 203: Would we want to bully for cash if we're not strapped for cash? Sounds like a weird choice that would limit us later

yairm210:
This is interesting. Is there really no downside to a pledge to protect? Is the metagame to Protect All The Things?

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
EmperorPinguin: Why would it limit us later? If we're at 100 gold, we can collect a tribute, so we can purchase a warrior, which we can use to tribute another CS, the gold of which we can use to purchase more warriors for more tributes and so on. After we're done purchasing warriors, it's waiting for the cooldown so we can tribute worker, or gold for shrines etc.

EmperorPinguin:
> This is interesting. Is there really no downside to a pledge to protect? Is the metagame to Protect All The Things?

I do pledge to protect all of them, I'm not sure if there are any downsides. It can still happen that others tribute the city states, in which case you click "You'll pay for this!" and continue business as usual.

core/src/com/unciv/logic/automation/civilization/NextTurnAutomation.kt
EmperorPinguin: The gold can be re-invested into our empire, which snowballs more than not tributing and waiting for a quest to happen to friend the city state. It can be correct to tribute workers and gold from friend/ally city states as well, but let's be a bit cautious about doing so (the influence decay is more expensive to re-purchase than the value of gold you're tributing).

EmperorPinguin: Also, keep in mind that collecting tributes at the earliest opportunity will lock our opponents out of this same opportunity for the duration of the tibute cooldown period.

yairm210 (approved):

PR #12756, opened 2025-01-05
correct alignment of text to icons

sulai:
Many texts that were shown next to icons were not vertically aligned correctly. This PR fixes that for most instances. I also tested with different font scales and screen sizes.

### DIFF
--- a/core/src/com/unciv/ui/images/IconTextButton.kt
+++ b/core/src/com/unciv/ui/images/IconTextButton.kt
@@ -10,8 +10,6 @@ import com.unciv.Constants
 import com.unciv.ui.components.fonts.Fonts
 import com.unciv.ui.components.extensions.toLabel
 import com.unciv.ui.screens.basescreen.BaseScreen
-import kotlin.math.floor
-import kotlin.math.ceil
 
 /**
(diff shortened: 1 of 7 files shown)


core/src/com/unciv/ui/images/IconTextButton.kt
```diff
@@ -44,6 +42,11 @@ open class IconTextButton(
...
     init {
         pad(10f)
-        labelCell.padTop(10f - Fonts.getDescenderHeight(fontSize));
+        // aligned with icon
+        labelCell.padTopDescent()
     }
 }
+
+fun Cell<Label>.padTopDescent(): Cell<Label> {
+    return padTop(Fonts.getDescenderHeight(Fonts.ORIGINAL_FONT_SIZE.toInt()) * actor.fontScaleY)
```
sulai, line 51: Not entirely sure if this is the correct calculation, but it works well for the test cases I tried:
- different Screen Sizes set in the options
- different Font Size Multiplier set in the options

yairm210:
Looks good!
Please add a javadoc comment (`/** comment */`) on the function indicating its purpose
e.g. "Intended for icons in tables that also contain text, to ensure they align correctly"

yairm210:
Self-conflicts :/

PR #12765, opened 2025-01-07
Add Google Maps - like pinching

sulai:
Up to now, the pinch gesture resulted zooming centered on the screen. This PR introduces pinching centered on the pointers (fingers) used for the gesture, which results in a more natural feel when navigating the map. Panning and pinching can be smoothly transitioned from/to by adding the second finger, much like in Google Maps.


sulai:
I have to admit, it actually *is* quite *daring* :smile: 

I'm pretty sure cancelTouchFocus() isn't needed as I did a lot of testing in all 3 clients. cancelTouchFocus() seems to cancel all listeners except the FlickScrollListener, which let our ZoomGestureListener constantly miss out on dragging movements of pointer 0, which in turn meant during pinching, the first finger's movement didn't have any influence on pinching and when re-touching, the map jumped.

Flicking (flinging) is still "working" to the extend it was working before. It's not actually good, but also (for my taste) not really impactful.

I submitted a follow-up: #12776

PR #12768, opened 2025-01-08
Update Spanish.properties

NahuelGeno:
Translation for the new notification system remains to be refined since in Spanish there are genders and several types of past tense.


NahuelGeno:
Upgraded with help of @Caballero-Arepa. Thanks.

PR #12772, opened 2025-01-08
Hide invisible resources for AI

EmperorPinguin:


yairm210 (approved):

NahuelGeno:
With this modification, the AI ​​will not be able to see strategic resources that it has not yet researched, right?

My question is, won't the AI ​​lose a lot of strength this way? Maybe it's fairer, but it seems to me that the AI ​​will lose strength towards the end game, although it will possibly gain more level in the early game by founding more useful cities in the early game.

EmperorPinguin:
> With this modification, the AI ​​will not be able to see strategic resources that it has not yet researched, right?
> 
> My question is, won't the AI ​​lose a lot of strength this way? Maybe it's fairer, but it seems to me that the AI ​​will lose strength towards the end game, although it will possibly gain more level in the early game by founding more useful cities in the early game.

The AI may settle outside range of strategic resources, especially Oil, but these weird oil cities led to complains by players, as they can't see them yet and it only become relevant in the modern era. Note that the AI does have logic to purchase tiles with strategic resources, so I don't expect it'll hurt their performance too much.

PR #12774, opened 2025-01-08
Update CityLocationTileRanker.kt

EmperorPinguin:
@itanasi, what do you think of something like this, to let AI settle in better orders (and probably let the settler locations wander less)?

Note that a range of 15 should be enough for most maps, but maybe not enough in huge map marathon games. Also, the distance value 3 should probably be treated differently if we're on a different continent or on the same continent (3 tiles apart can be correct if we're trying to fit cities on a cramped Archipelago map).

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -139,8 +139,9 @@ object CityLocationTileRanker {
                 distanceToCity < 4 -> -30f // Even if it is a mod that lets us settle closer, lets still not do it
                 else -> 0f
             }
+            val rankDistanceToCapital = city.isCapital() && newCityTile.aerialDistanceTo(city.getCenterTile()) > 3 //exlude first 3 rings
             // We want a defensive ring around our capital
-            if (city.civ == civ) distanceToCityModifier += if (city.isCapital()) 3 * (15 - newCityTile.aerialDistanceTo(city.getCenterTile())).coerceIn(0, 11) else 0
+            if (city.civ == civ) distanceToCityModifier += if (rankDistanceToCapital) 3 * (15 - newCityTile.aerialDistanceTo(city.getCenterTile())).coerceAtLeast(0) else 0
             modifier += distanceToCityModifier
         }
         return modifier
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -133,7 +133,7 @@ object CityLocationTileRanker {
                 // If it is not higher the settler may get stuck when it ranks the same tile differently
                 // as it moves away from the city and doesn't include it in the calculation
                 // and values it higher than when it moves closer to the city
-                distanceToCity == 6 -> 4f
+                distanceToCity == 6 -> 3f
                 distanceToCity == 5 -> 8f // Settling further away sacrifices tempo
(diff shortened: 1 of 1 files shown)


itanasi:
So this would encourage packing all the cities within 15 tiles of the capital. I don't think the 15 hardcode will be a serious issue. Once all the space is done then there's no additional weight further out.

However, I am concerned about how it's being calculated. This effectively creates a huge hot zone as close to the Capital as possible. Distance 5 is only 1 point better than Distance 4 from the Capital (+38 vs +37)? These values are also super high for the Capital. We go from +8 at distance 6 to +31. So there's no way we will ever settle a chain of cities. Finding 20 points of value elsewhere is incredibly difficult.

The current implementation favors settling closer to neighboring cities but this pretty much makes the Capital overwhelmingly dominant.

EmperorPinguin:
> However, I am concerned about how it's being calculated. This effectively creates a huge hot zone as close to the Capital as possible. Distance 5 is only 1 point better than Distance 4 from the Capital (+38 vs +37)? These values are also super high for the Capital. We go from +8 at distance 6 to +31. So there's no way we will ever settle a chain of cities. Finding 20 points of value elsewhere is incredibly difficult.

Chains of cities aren't defendable against suprise attacks. Finding 20 points of value somewhere else is possible; a river hill mountain luxury settle already grants +42 points compared to a settle who doesn't have any of these features. The "hot zone" around capital could indeed be a problem, in case the capital land is worse than expand land. The way map gen and spawn placement works, the capital land is equal to or better then the surrounding land, but there could be cases where settling 6 or 7 tiles away is necessary.


> The current implementation favors settling closer to neighboring cities but this pretty much makes the Capital overwhelmingly dominant.

Note that the previous implementations also got a ×2 modifier for the capital. I use aerial distance to capital as a proxy for movement cost. It maybe be faster to settle a chain of cities on flatland first and only then go for the hill regions, but movement is less stable, as it can change based on roads being completed etc.

EmperorPinguin:
> The old 2x modifier for Capital made it so there was a bias for a ring around the Capital. But this is adding a flat 38 points of value for being near the Capital!

Yes, I marked it as a draft when I realized it would create undesireable results with mods. Does this look better?

itanasi:
Excluding the first 3 rings helps, but again I think the scaling is too large. 20-30 points of value on top of the other ring bonus is a pretty hefty bias. 2x(10-distance) might be more reasonable. That's still 16 at distance=4 and 18 at distance=5, which is similar to the 12 and 16 from the old code. And then we have a smoother gradient as we move from the Capital than the old double weighted ring.

EmperorPinguin:
> Excluding the first 3 rings helps, but again I think the scaling is too large. 20-30 points of value on top of the other ring bonus is a pretty hefty bias. 2x(10-distance) might be more reasonable. That's still 16 at distance=4 and 18 at distance=5, which is similar to the 12 and 16 from the old code. And then we have a smoother gradient as we move from the Capital than the old double weighted ring.

A delta of 3 points per tile is needed to prevent settlers from wandering off too far, if I set it lower there's not really a difference compared to the old approach. I considered 15 tiles, as that 3 rings of distance 5 cities, which we won't reach on most maps. I'm not opposed to considering a max of 10 tiles (2 rings of distance 5 cities), but the formula would stop working on very big maps.

itanasi:
OK if we need the gradient of 3 to fix the behavior that's fine.

I don't think we need it to reach 15 tiles. At that point the overlapping circles from neighboring cities will do the job of focusing Settlers for us. Remember, the intersection of 2 cities at distance 5 will be around 16 points of value on their own.

PR #12776, opened 2025-01-08
Map pinching, revised

sulai:
The previous implementation was a good start, but still quite inaccurate. The stage coordinates are now correctly translated to map-coordinates and the map position is corrected to zoom in to the two pointer's focal point. As a result the pinch gesture feels more reactive and accurate (it felt slow previously).

### DIFF
--- a/core/src/com/unciv/ui/components/widgets/ZoomableScrollPane.kt
+++ b/core/src/com/unciv/ui/components/widgets/ZoomableScrollPane.kt
@@ -173,7 +173,7 @@ open class ZoomableScrollPane(
         return zoomListener.isZooming
     }
 
-    inner class ZoomListener : ZoomGestureListener() {
+    inner class ZoomListener : ZoomGestureListener({ Vector2(stage.width, stage.height) }) {
 
         inner class ZoomAction : TemporalAction() {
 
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/ui/components/widgets/ZoomableScrollPane.kt
```diff
@@ -263,13 +260,16 @@ open class ZoomableScrollPane(
...
-            }
-            val scale: Float = sqrt((distance / initialDistance).toDouble()).toFloat() * lastScale
-            zoom(scale)
+        override fun zoom(delta: Float, focus: Vector2) {
+            // correct map position to zoom in to the focus point
+            val dx = (stage.width / 2 - focus.x) * (delta - 1) / scaleX
+            val dy = (stage.height / 2 - focus.y) * (delta - 1) / scaleY
+            scrollTo(
+                scrollX - dx,
+                scrollY + dy,
```
yairm210, line 269: Is this difference due to a difference between screen and stage coordinates, regarding top-left or top-right at 0,0?

core/src/com/unciv/ui/components/widgets/ZoomableScrollPane.kt
```diff
@@ -263,13 +260,16 @@ open class ZoomableScrollPane(
...
-            if (lastInitialDistance != initialDistance) {
-                lastInitialDistance = initialDistance
-                lastScale = scaleX
-            }
-            val scale: Float = sqrt((distance / initialDistance).toDouble()).toFloat() * lastScale
-            zoom(scale)
+        override fun zoom(delta: Float, focus: Vector2) {
+            // correct map position to zoom in to the focus point
+            val dx = (stage.width / 2 - focus.x) * (delta - 1) / scaleX
+            val dy = (stage.height / 2 - focus.y) * (delta - 1) / scaleY
```
yairm210, line 266: This needs a lot more explanation, in comments.
scaleX and scaleY should always be the same so this is IMO needlessly confusing.

"stage and focus and in **stage pixesl**. Since the entire map is zoomed, however, the actual on-screen pixels are in stage pixels * scale".

"We take the diff between the focus point and the stage center point - that's the difference in stage pixels."
"When scrolling, it would be a different number of pixels - to get this need to convert *current zoom pixels* to *stage pixels* to *new zoom pixels".
"Stage to zoom pixels are 1:scale, so current zoom pixels to new zoom pixels are oldScale:newScale"
"Since newScale = oldScale * delta, oldScale:newScale = oldScale:oldScale*delta

And from here tbh I lost you

sulai: I added comments in the code, let me know if it's easier to understand now. Key point is that translation consists of two parts: one cause by moving the focal point and another one caused by scale change, because we want to direct zooming towards the focal point. The latter is what was calculated here. I moved that calculation for better clarity.

yairm210 (approved):

PR #12781, opened 2025-01-10
Settling Suggestion Improvements

itanasi:
Bring up floor for Settle Suggestions, and increase radius to a fixed 5 tiles
Add distance from Settler penalty to help keep suggestions closer
Human Suggestions do not take into account unseen resources

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -24,6 +24,7 @@ object CityLocationTileRanker {
      * Returns a hashmap of tiles to their ranking plus the a the highest value tile and its value
      */
     fun getBestTilesToFoundCity(unit: MapUnit, distanceToSearch: Int? = null, minimumValue: Float): BestTilesToFoundCity {
+        val distanceModifier = 2.7f // percentage penalty per aerial distance
         val range =  if (distanceToSearch != null) distanceToSearch else {
             val distanceFromHome = if (unit.civ.cities.isEmpty()) 0
             else unit.civ.cities.minOf { it.getCenterTile().aerialDistanceTo(unit.getTile()) }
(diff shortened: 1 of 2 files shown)


EmperorPinguin:
https://github.com/yairm210/Unciv/pull/12772 should hide unseen resources for both human and AI.

EmperorPinguin:
If I understand correctly, this home distance penalty is based on the distance to the settler, so it will move along with the settler, and not atually do anything besides make the suggested city location jumpt around more?

itanasi:
OK, reverted the threshold and removed conflicts with #12772 
So the main 2 changes then are the Player Settle Suggestions in the UI will always be a 5 tile radius (instead of shrinking further from cities), and adding the % penalty further from the Settler.
The % Penalty should keep the suggestions more stable across the board for both the suggestion icon in the UI and for AI automation.

EmperorPinguin:
> OK, reverted the threshold and removed conflicts with #12772 So the main 2 changes then are the Player Settle Suggestions in the UI will always be a 5 tile radius (instead of shrinking further from cities), and adding the % penalty further from the Settler. The % Penalty should keep the suggestions more stable across the board for both the suggestion icon in the UI and for AI automation.

Increasing the range to a set 5 tiles for human is unquestionably good, but are you sure the distance to settler penalty doesn't move along with the settler? Maybe I'm misunderstanding, but I read it as distance to settler that will be calculated anew when the settler moves, which makes that the AI will value the same tile differently than it did the previous turns, which will lead to the AI moving the settler back and forth between the same tiles, as the city location value doesn't converge into a stable local maximum.

itanasi:
The penalty is any distance from the Settler.

If there is a tile that's better than the on you're currently on, moving towards it will remove the penalty. Therefore you will arrive at a local maxima.

Example: 
Stand on Tile A, score 70.
Next tile over, Tile B, base score 73, after penalty, 73*.97=70.81
Tile B is better! Move to Tile B.
Now, standing on Tile B. Score 73.
Tile A is now 70*.97=67.9. Worse than B! Stay on B!

EmperorPinguin:
It should indeed converge into a local maximum. There's a bit of a risk the AI will settle "too early", but that's desireable over settling "too late" I think. Have you tested if they can now handle a constant search radius (5 tiles). If such doesn't lead to wandering off, it'll probably improve the quality of settles a bit.

itanasi:
The fixed 5 range is only for the player aids, not the AI settling behavior

itanasi:
Are there any concerns on this?

yairm210:
None for me, will merge if @EmperorPinguin is fine with it

EmperorPinguin:
It looks fine to me

yairm210 (approved):

Issue #12308, opened 2024-10-17
Wait Button behavior 


itanasi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.13.17

### Describe the bug

Pressing the Wait button isn't working as expected.
It should select the next available Idle unit and place the current unit at the end of the Idle Unit queue.
However it centers the view on the next unit, but doesn't select them. And after the 2nd unit acts doesn't recenter nor select the Wait commanded unit. Instead only shows Next Turn.

### Steps to Reproduce

Create new game
Build a 2nd unit
At turn start, Wait the first unit

### Link to save file

[...]

PR #12782, opened 2025-01-10
Tweaks for Screen Size small-portrait

sulai:
- add close button (x) to battle table (fixes [#12745](https://github.com/yairm210/Unciv/issues/12745))
- fixed drawing order of battle table and city stats
- unit info table arrow buttons won't set due=false [#12745](https://github.com/yairm210/Unciv/issues/12745) [#12308](https://github.com/yairm210/Unciv/issues/12308) 
- a few smaller UI tweaks for small-portrait

### DIFF
--- a/core/src/com/unciv/ui/components/extensions/Scene2dExtensions.kt
+++ b/core/src/com/unciv/ui/components/extensions/Scene2dExtensions.kt
@@ -293,6 +293,28 @@ fun getCloseButton(
     return closeButton
 }
 
+/**
+ * Adds a white-circled (x) close button to [parent], positioned to the top right,
+ * slightly shifted outwards.
+ */
+fun addRoundCloseButton(
+    parent: Group? = null,
+    size: Float = 30f,
+    iconSize: Float = size - 15f,
+    circleColor: Color = Color.CLEAR,
+    overColor: Color = Color.RED,
+    action: () -> Unit
+): Group {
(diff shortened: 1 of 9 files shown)


core/src/com/unciv/ui/components/extensions/Scene2dExtensions.kt
```diff
@@ -293,6 +293,28 @@ fun getCloseButton(
...
+ * Adds a white-circled (x) close button to [parent], positioned to the top right,
+ * slightly shifted outwards.
+ */
+fun addRoundCloseButton(
+    parent: Group? = null,
+    size: Float = 30f,
+    iconSize: Float = size - 15f,
+    circleColor: Color = Color.CLEAR,
+    overColor: Color = Color.RED,
+    action: () -> Unit
```
yairm210, line 306: This suffers from Trog Disease, which is adding overrides that are not actually used when they can easily be added in the future when needed
Please minimize to only what we currently need, and push the specifics into the function :)

yairm210: e.g. parent stays, but is only Group (never null)
All the rest are hardcoded in the function

sulai: Thank you for the feedback, fixed :+1:

PR #12786, opened 2025-01-10
Fortify until Healed removes Fortify at Turn Start

itanasi:
Fix so that while Fortify Until Healed is active, the Fortify will persist until your own turn start.
Current behavior checks and removes the Fortify status at turn end, leaving your unit exposed during enemy turns until you can Fortify on your turn again.

Germany Turn, Fortify Until Healed on Spearman
image
Denmark Turn, Spearman is fully Healed and still Fortify status
image
Germany Turn, Spearman wakes and can Re-Fortify. Denmark Spear similarly has Healed and is still Fortified
image

### DIFF
--- a/core/src/com/unciv/logic/map/mapunit/UnitTurnManager.kt
+++ b/core/src/com/unciv/logic/map/mapunit/UnitTurnManager.kt
@@ -35,11 +35,6 @@ class UnitTurnManager(val unit: MapUnit) {
         if (!unit.hasUnitMovedThisTurn() || unit.hasUnique(UniqueType.HealsEvenAfterAction))
             healUnit()
 
-        if (unit.action != null && unit.health > 99)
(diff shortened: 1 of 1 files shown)


EmperorPinguin:
Best UI improvement in a while, this saves so much time switching between menus.

itanasi:
Note a weird edge case here. If it healed to full but then was attacked between turns, previously it would Wake Up, now it won't. It suspect players want the unit to wake only once it's Fully Healed, not if it was Fully Healed at the end of their turn, then lost HP during the opponent turn.

EmperorPinguin:
> Note a weird edge case here. If it healed to full but then was attacked between turns, previously it would Wake Up, now it won't. I suspect players want the unit to wake only once it's Fully Healed, not if it was Fully Healed at the end of their turn, then lost HP during the opponent turn.

Correct

PR #12789, opened 2025-01-11
Update Automation.kt

EmperorPinguin:
I propose to not slow down growth if we're growing well, we need the trade route gold (players have been complaining quite a bit recently about AI bleeding gold on Warlord and lower).


yairm210:
Sure, we can try and see what happens

PR #12790, opened 2025-01-11
Added option to change the Maximum Autosave turns stored

Emandac:
Added a drop down box to change the Maximum Autosave turns stored.


Emandac:
So i should change it to `intArrayOf(1, 2, 5, 10, 15, etc..).asIterable().toGdxArray()` ?

core/src/com/unciv/logic/files/UncivFiles.kt
```diff
@@ -486,7 +489,8 @@ class Autosaves(val files: UncivFiles) {
             fun getAutosaves(): Sequence<FileHandle> {
                 return files.getSaves().filter { it.name().startsWith(AUTOSAVE_FILE_NAME) }
             }
-            while (getAutosaves().count() > 10) {
+            // added the plus 1 to avoid player choosing 6,11,21,51,101, etc.. in options.
```
yairm210, line 492: I don't get this comment - with this config, if you say "I want 10 files" what you'll actually get is 11 files, why is that better?

core/src/com/unciv/ui/popups/options/AdvancedTab.kt
```diff
@@ -51,6 +51,8 @@ class AdvancedTab(
     init {
         pad(10f)
         defaults().pad(5f)
+        
+        addmaxAutosavesStored()
```
yairm210, line 55: Capitalize Max

core/src/com/unciv/ui/popups/options/AdvancedTab.kt
```diff
@@ -92,7 +94,22 @@ class AdvancedTab(
             optionsPopup.reopenAfterDisplayLayoutChange()
         }
     }
-
+    
+    private fun addmaxAutosavesStored() {
+        add("Maximum autosave turns stored".toLabel()).left().fillX()
```
yairm210, line 99: "autosave turns" makes it seem like this is dependent on the game turns, when actually it's dependent on the number of files
A better phrasing would be "Number of autosaves stored" or perhaps "Number of autosave files stored"

core/src/com/unciv/logic/files/UncivFiles.kt
Emandac: with the old version with 10 has example it would start overriding after 9 instead of 10.

yairm210: OHH I KNOW
It's because the numberless "autosave" also gets counted...
So add that into the comment please :)

yairm210 (approved):
Looks good :)

Issue #12785, opened 2025-01-10
Feature request: Unit Cycling Features


itanasi:
### Before creating

- [X] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [X] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

Some players would like a feature to skip past the current selected unit, but then cycle back to it later. This previously was Wait, but apparently that was confusing, and now Wait acts like Next Unit. That is, we set `due=false` and don't come back this turn automatically.

There is also some concern that Next Unit is silently marking units `due=false` and so players may unintentionally be skipping that unit's turn, not understanding they won't cycle back.

Overall we should have clarity without cluttering the UI. Please discuss different solutions.

### Related Issue Links

#12745 #12308

### Desired Solution

Open to discussion

### Alternative Approaches

Current situation:
* Next Turn marks current unit `due=false` and moves to next unit in queue
* Unit Action "Wait" marks current unit `due=false` and moves to next unit in queue

sulai:
Proposal A):
- the Giant Button shows "Next Unit" and sets `due=false`, but changes its label to "Skip Unit" if a unit is selected and still has action points to spend. This avoids accidentally letting units skip one turn.
- Currently, when *no unit is selected*, there's no unit-info-table and the bottom left is empty. Proposal: we add a small "unit summary info box" which sits in the bottom left and has this content:

``` <  3 skipped and waiting, 6 due >```

Purpose:
- when no unit it selected, give the player a way to start cycling. we could add a shortcut like `.` to cycle forward.
- when the Giant Button finished cycling units, it deselects the last one. This window would appear and would leave no doubt about the amount of units waiting, it would show something like `< 3 skipped and waiting >` or `< all units busy >`.

Proposal B):
- make the Giant Button NOT set `due=false`. The risk of accidentally skipping units is unacceptable.
- Instead the Giant Button sets `done=true`, so we have a separate flag for the Giant Button.
- Once all units are done, we check if some of them haven't been decided on yet.
- In case there are matching units, ...
  - that means the player hit the Giant Button but did not give orders to all of the units.
  - we show the player a hint that he should let units "Wait" if they are not supposed to move this turn. So we educate new players that this is the intended way of dealing with units.
  - we reset `done` and let the Giant Button start cycling again

Thoughts?

itanasi:
[...]

PR #12799, opened 2025-01-12
Add "Wait" to Next unit button, Next unit cycles

sulai:
When the Giant Button is in the "Next Unit" phase, show a second button, to allow the user to
- either **cycle** through the units
- or give them the **wait** order

Some Reasoning (details see discussion threads):

A considerable part of the community is not aware the Giant Button currently marks units as "done" (due=false). Some of those who _are_ aware would prefer it not to set units as "done", as there is the risk of accidentally doing so. Generally I agree with the standing that the Giant Button should only help navigating in the UI pointing to the "things to do", but not actually alter game state.

[Discussed on Discord](https://discord.com/channels/586194543280390151/1327742045552513095) and in #12785.


itanasi:
I would also point out that this is changing the Big Button saying Next Unit so that it now *won't* mark them as `due=false` which is a change from current behavior. I'm not sure if we have full consensus on that, but this does make an elegant consistent UI.

yairm210:
Visual changes should be accompanied with a screenshot :)

sulai:
I changed the Wait icon and added a Screenshot :+1:

itanasi:
I would also like to see updates to the Civilopedia documentation. Idle Units and World Screen are the big points.

itanasi:
Oh, and since yairm usually uses the title in the changelog, you should mention the change to the Next Unit behavior in the title

najevi:
It seems to me that you are trying _very hard_ (and for some undisclosed reason) to make that button (20) piece of screen real estate (between Overview and Notifications) become the _only_ place a user should need to tap/click.
I don't understand the motivation behind this when you also spend 7 of the 12 _most prominent_ legend (number, description) pairs of WorldView Screen describing the UI components that are _supposed to be used_ to manage units. viz. _items 7-12 highlighted in this screenshot_

Screenshot (115)

This prominence makes sense to me. After all, the phase of a turn that can best be described as "managing units" is where a user spends most of their WorldView screen-time, _wouldn't you agree?_

With reference to what @sulai wrote:
> Generally I agree with the standing that the Giant Button should only help navigating in the UI pointing to the "things to do", but not actually alter game state.

If indeed we all agree _in principal_ that button (20) should direct a user's attention to the activities a user ought to be focused on within the current phase of a turn then _why not simply direct the user_ to the <ins>relevant</ins> already-existing user interface panes/menus/buttons that are dedicated to managing units?  viz. "Unit Action Buttons" and "Unit/City Info pane". That's it! No feature creep!!  Just a prompt/guide/direction/instruction as to _where_ the user ought to focus their attention.

- You have not yet put forward any logical argument for adding <ins>unit-navigation</ins> and <ins>unit-action</ins> buttons to that piece of screen real estate that is _diagonally opposite_ the already existing UI components _specifically designed for_ the very same two tasks you are proposing herein to _redundantly_ duplicate!!

> I would also point out that this is changing the Big Button saying Next Unit so that it now _won't_ mark them as `due=false` which is a change from current behavior. I'm not sure if we have full consensus on that, but this does make an elegant consistent UI.

I cannot agree that this makes the UI either elegant or consistent.

## Room for improvement

I agree that both the Unit/City Info pane and the Unit-action menu _need further work_ however, those two UI components (and the numerous buttons that constitute them) are _without question_ the <ins>most appropriate place</ins> a user ought to be guided to for the phase of the turn that we are discussing here.  

By all means leave an appropriate reminder message in that area of button (20). A message to the effect that "managing units" has not yet been completed. (Promising discussion of informative messaging to that effect is going on at #12785) 

It seems irrational to select an _arbitrary_ subset of the unit-actions and unit-queue navigation-buttons and essentially _duplicate them_ on the very same WordlView screen where the already-existing buttons are still visible!  
- Am I _really_ the only person who sees this as foolish?
- The argument/reason, "because it (whatever 'it' may be)  has always been that way" is _not_ a logical argument ... it is tribal dogma.
 
With regard to the Wait/Skip unit action  ... that unit action button has been missing from it's rightful place among all of the other "Unit Action Buttons" since, well, forever. <ins>That flaw</ins> should be the subject of this report and not the proposed band-aid measure that perpetuates a fundamentally flawed and <del>hysterically</del> historically overlooked UI design mistake. <br>
If a prominent place for this one unit action needs to be found then we ought to be discussing where within the dynamic menu of unit-actions that Wait action should consistently appear even as the user cycles through the unit-queue. 
- I'd very much like to read a _well reasoned argument_ why the misuse of button (20) should continue to be overlooked.

itanasi:
Because having a single location to "progress" your turn is a good thing in UI design.

If you want to discuss adding other useful features elsewhere please do it in another thread.

najevi:
Understood. The following 3 questions represent me trying _even harder than I had thought I had been_ to be humble, kind and collaborative within the scope of _this_ thread.

1. Why duplicate the equivalent functions of 
- one button from the Unit/CityInfo pane (viz. Next unit) and 
- another button from the Unit Action Buttons (viz. Wait)

2. Is there a way to guide the user to those already existing buttons?

3. Can those already existing buttons be made any more prominent than they presently are?

sulai:
I updated the Tutorial / Civilopedia. The PR seems to be ready, and I think it's a tiny step into the right direction.

@najevi I appreciate your input, but I can't currently afford the time and effort required to try a holistic approach as you propose. If you really want to dive into that, maybe creating a click dummy is a way to convey your ideas better and let people try it out interactively. It might also help to manifest your vision and iron out problems your idea might still have.

sulai:
> I do agree that ceasing the due=false state change is an improvement. I cannot agree [on] any other aspect

The reason for the "Wait" button is that we need to give the player a choice. Otherwise new players are stuck in an endless loop forever and can't end their turn. We can't expect new players to know what they are supposed to be doing, for example, command units to Wait. That's why we show that "Wait" button right next to the Giant Button.

@itanasi should we still show a one-time dialog in case new players keep cycling their units and don't use the "Wait" button?

itanasi:
>     1. Why duplicate the equivalent functions of
>        
>        * one button from the Unit/CityInfo pane (viz. Next unit) and
>        * another button from the Unit Action Buttons (viz. Wait)

Because having a Big Button that directs you to progress the game is a Good Thing in UI design. And duplication isn't an issue. Often having multiple ways to do the same thing is helpful. Plus the arrows in the Unit/CityInfo pane DO NOT duplicate behavior. Those arrows cycle through all available units that have no fixed orders (Idle Units). Next Unit should move to any unit with `due=true` eg no current orders and no Wait command (Due Units).

Plus still having a Wait Unit Action button in case the user unchecks `Check for idle units` can be handy. Or we remove it if you're still concerned.

itanasi:
> @itanasi should we still show a one-time dialog in case new players keep cycling their units and don't use the "Wait" button?

Fundamentally it seems a lot of the "problem" is people feeling like the UI isn't explained enough to a new player. This is fixable by a tutorial popup. We unfortunately can't really take advantage of hover text.

UI elements by necessity *should* change the game state and combining ones that would go together to have click efficiency is a Good Thing in UI.

core/src/com/unciv/ui/screens/worldscreen/status/StatusButtons.kt
```diff
@@ -1,43 +1,48 @@
...
         right()
         wrapReverse()
         wrapSpace(10f)
         rowRight()
-        if (autoPlayStatusButton != null) {
-            addActor(autoPlayStatusButton)
-        }
-        if (multiplayerStatusButton != null) {
-            addActor(multiplayerStatusButton)
-        }
```
itanasi, line 40: Did you remove these and not do a replacement? That's not good

sulai: It's dead code

sulai:
After trying this approach, I came to the conclusion that it does have its drawbacks. One of those problems is that we can't easily switch the behavior of the two buttons by a setting (as being advocated for by @itanasi), because the "Wait" button is oftentimes getting disabled during the NextUnit phase, which in my mind would result in a bad UX if we assigned the "Wait" behavior to the Big Button. #12814 tackles this much better.

sulai:
> the UI isn't explained enough to a new player. This is fixable by a tutorial popup.
The new PR #12814 provides better clarity, no matter what behavior we assign to the NextUnit button.

> UI elements by necessity should change the game state and combining ones that would go together to have click efficiency is a Good Thing in UI.

I see your point and I had the same opinion only days ago. But after trying it myself, I have to agree that it feels much better if the Big Button doesn't do changes to the game state. As stated before, the new PR supports the idea to have a setting for that.

yairm210:
I'm sorry, there's Too Much Text on this issue for me to follow
If both you and Itanasi agree on this I'll merge it
@itanasi what say you?

itanasi:
@yairm210 No, discussion is ongoing and the author wishes to withdraw it right now.

> because the "Wait" button is oftentimes getting disabled during the NextUnit phase

When would this happen? I seem to be missing how that would happen if coded how I'm imagining it.

Lastly as I mentioned in the other PR, the Discord poll was pretty overwhelmingly against having just the Big Button around and it doing Cycling only. The implementation here is a bit trickier, but having two prominent choices next to each other makes it much more obvious they're associated and to make one of two choices.

If you wanted, perhaps a merge of both if you think the text feedback is helpful/crucial? Main cost is the screen space.

itanasi:
> This is the case when no unit is selected or the unit has been given a command

If no unit is selected, we can still use Next Unit to select one. If the selected unit already has orders and no further actions (doesn't need to Wait), then Wait doesn't do anything anyways and can still be active. Pressing it sets `due=false`, which it already is, and we select the next Due Unit.

This does mean that the button has to transition from No Unit Selected mode, showing Next Unit like it does now, and then moving Next Unit (or just Next as the text) to the smaller button. But to me that's minor and I'm doubtful a player would mash the button. If we need to distinguish them, we can also use a different verb. Next Unit to select a unit, and when it moves change the verb to Cycle. Wait as a verb is still consistent across all UI.

sulai:
That's what I'm talking about. The big button would need changing text all the time during one phase. Then having the big button do what's already done, or even let a unit "wait" when I'm reality it's already doing another thing... If we wanted to suppress that we'd need to disable it or change text to "next unit", but the we have the buttons next to each other doing the same.

Unfortunately it doesn't work out well from the UX perspective.

itanasi:
I see what you mean, I just disagree it's as big of an issue.

How about you mark this as a Draft, we move forward with the other PR idea, and if I have free time I'll take a crack at it and then you can play with it and see how it feels. I don't think we'll solve this with theory.

sulai:
I think that's a good idea. 👍

sulai:
> If you wanted, perhaps a merge of both if you think the text feedback is helpful/crucial? Main cost is the screen space.

Giving this a second thought, I think you made a strong point here. Let's try this out after the other PR got merged.

PR #12813, opened 2025-01-13
update of polish translation

SHypoleac:


yairm210:
It seems there are conflicts between this and another translation :/

PR #12814, opened 2025-01-14
Show "x units due" on Big Button, setting for cycling units

sulai:
This PR is an *alternative approach* to #12799 - we should decide for *one* of them to merge.

It adds a small Label below the "Next Unit" label within the Big Button, showing the number of due units.

I personally think this one got several advantages over the other:
- no extra button required (as opposed to the "Wait" button)
- it educates the player that the NextUnit button does (cycling or skipping a turn)

When cycling is enabled in the settings:
- doesn't change game state, no accidental skipping of units
- it encourages the player to use the normal unit commands
- Easy cycling with Space on Desktop and Big Button on Android

grafik
Due units on Big Button. counting down as you give orders.

grafik
When NextTurn is disabled in the settings,  the "x units due" also shows on the "Next Turn" Button. In case the user doesn't like the cycling, he still gets notified about some units being idle.


itanasi:
The biggest issue here is this goes against the poll result that players would rather a singular Big Button would mark their units as Done for the turn (`due=false`)
https://discord.com/channels/586194543280390151/618491418859798539/1327727351122886697

sulai:
Please note that the poll is *biased*. The participants didn't have the change to *actually try* the alternative. It's the "oh it works much better than I thought it would" - moment that often occurs when people actually use a feature themselves, instead of *imagining* how a feature would work. This happened a lot with the new construction queue. I tried to explain the advantages, but the truth is, they had to try it themselves.

There's always a natural and understandable friction to change, which adds to the bias. We should consider that bias when doing polls.

That being said, this PR makes it very feasible to have a setting to change the button's behavior, while the 2-button-solution [is problematic](https://github.com/yairm210/Unciv/pull/12799#issuecomment-2589284552) in that regard.

itanasi:
I think at minimum we should add a new Tutorial screen that pops up the first time someone presses the Next Unit button to explain that they need to give the Wait command to units to progress the game.

And considering the poll, have the option to switch functionality back to Mark as Done would be a nice to have.

sulai:
I added the setting and by default, the Big Button sets `due=false` to make it safe for beginners.

najevi:
Is a "due unit" what the game option "Check for idle units" refers to as an "idle unit" ?

For the two of you the simple description  `due=false` carries meaning and is very clear but to a humble user that is code. 
Same could be said 'done=true' which came up elsewhere in discussions.

Elsewhere, @itanasi pointed out to me that the Next button in the UnitInfo pane is not identical in function to the giant (Next unit) button. That has not been my experience (and I have tested with "Check for idle units" option both enabled and disabled) so it leaves me wondering why my experience of the game should be so very different.

I mention the above only to caution you that introducing the word "due" to the UI might need greater consideration. Does that word appear anywhere else in the game help 'pages' or the the game options?  

Might the word "idle" mean something that a user can better relate to? A user may very well remember having set or cleared the "Check for idle units" option but not necessarily make the connection between that and the word "due".

itanasi:
There is a subtle difference.

Idle Units have no active Order and still have Movement Points this turn.

Due Units have been marked with `due=false` to indicate even if they're Idle, we don't need to cycle to check on them.

And yes this is not a term used elsewhere and is purely understood by programmers for specificity under the hood. If you'd like to suggest a better term please feel free to do so. But specifically this is *not* the same as Idle. These are Waiting for Orders. Would that work better?

najevi:
Thank you, that does lend clarity and I will observe a little more carefully to see I f can now notice the difference in just which units are cycled by one button versus the other. 

As for a synonym. Perhaps "Untasked". Of course, that has never been used in game either. It is one word though!

najevi:
> When NextTurn is disabled in the settings, the "x units due" also shows on the "Next Turn" Button. In case the user doesn't like the cycling, he still gets notified about some units being idle.

I am glad you anticipated that. I've been using that alternate mode lately (i.e. "Check for idle units" not set and therefore there is no "Next unit" button displayed) and I wondered how unit-queue-status might be tracked in the absence of a (Next unit) button.

There is a very similar edge-case to consider: when at least one unit has been given a multiple-turn movement order. So there may be some (Move automated units) buttons that would benefit from incorporating that unit-queue-status information as shown here.

sulai:
@SomeTroglodyte thank you for pointing this out, fixed it!

I'll officially confess that I'm obsessed with that button... since a whopping [6 years](https://github.com/yairm210/Unciv/commit/d2541e6c4af67c860618817e204826e4d50b9e7b) :smile: 

That's also when I introduced the `unit.due` attribute and at that time, it was *only* used for the big button to keep track which units the player had already checked on. As time went on, it more and more was used as an equivalent for "wait one turn". It might make sense to internally rename `due=false` to `waiting=true`.

It's not related to that text under the button though. I think due is a good word to use there. Short and to the point.

So guys, in a way, what I did 6 years ago is responsible for all the discussions we have today. My apologies :smile:

Not sure how to proceed with the PR. Are we ready? Is it a step into the right direction? Is everyone happy?

itanasi:
Personally looks good to me

sulai:
@yairm210 this one is ready to merge

As a side note, just to get the terms clear both from the **player's perspective** (bold) and `in-code` (fixed width font):
- `due=false` is identical to the term **Waiting** from the player's perspective. This has historic reasons (see my post above).
- **Idle** units are all units with movement points left and no task assigned. "Waiting" counts as an assigned task. In code, `unit.isIdle()` and `unit.due` are two separate things. If we wanted to be very clean we could fix that discrepancy and rename `due` to `waiting`, because that's what it means nowadays. That'd be a rather large change though.
- **Due** units from the player's perspective only exist when the Big Button is set not to *cycle* units, but visit them only once. Due units are Idle units minus those which already have been visited using that big button.

PR #12817, opened 2025-01-15
Add hotkey for Idle Unit cycle buttons

itanasi:
Add bindable hotkeys, default ',' and '/', to the Idle Unit Buttons in the UnitTable

'.' is Stop Movement, so this seemed relevant

#12785

### DIFF
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/IdleUnitButton.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/IdleUnitButton.kt
@@ -5,6 +5,7 @@ import com.badlogic.gdx.scenes.scene2d.Touchable
 import com.badlogic.gdx.scenes.scene2d.ui.Table
 import com.badlogic.gdx.utils.Align
 import com.unciv.logic.map.mapunit.MapUnit
+import com.unciv.ui.components.UncivTooltip.Companion.addTooltip
 import com.unciv.ui.images.ImageGetter
 import com.unciv.ui.screens.worldscreen.worldmap.WorldMapHolder
 import com.unciv.ui.components.input.onClick
@@ -32,6 +33,7 @@ class IdleUnitButton (
         add(image).size(imageSize).pad(10f,20f)
         enable()
         keyShortcuts.add(keyShortcutBind)
+        addTooltip(keyShortcutBind)
         onActivation {
 
             val idleUnits = unitTable.worldScreen.viewingCiv.units.getIdleUnits()
--- a/core/src/com/unciv/ui/screens/worldscreen/unit/IdleUnitButton.kt
+++ b/core/src/com/unciv/ui/screens/worldscreen/unit/IdleUnitButton.kt
@@ -9,11 +9,15 @@ import com.unciv.ui.images.ImageGetter
 import com.unciv.ui.screens.worldscreen.worldmap.WorldMapHolder
 import com.unciv.ui.components.input.onClick
 import com.unciv.ui.components.extensions.pad
+import com.unciv.ui.components.input.KeyboardBinding
+import com.unciv.ui.components.input.keyShortcuts
+import com.unciv.ui.components.input.onActivation
 
 class IdleUnitButton (
     internal val unitTable: UnitTable,
     private val tileMapHolder: WorldMapHolder,
-    val previous: Boolean
+    val previous: Boolean,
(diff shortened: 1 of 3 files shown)


sulai:
The hotkey `/` is not a good choice on many keyboard layouts. If you don't have a preference and technically possible, please change it to
- `,` - next unit
- `shift+,` - previous unit

Also, these bindings currently do nothing when no unit is selected. Allow me re-present a slightly modified idea I posted somewhere else:

> Currently, when *no unit is selected*, there's no unit-info-table and the bottom left of the screen is empty. Proposal: we add a small "unit summary" info box which sits in the bottom left and has this content:
> 
> ``` < 3 units idle >```
> 
> Those `<` and `>` are the same unit-cycle-buttons that can be found in the unit-info-box
>
> Purpose:
> - when no unit it selected, give the player a way to start cycling. we could add a shortcut like `,` to cycle forward.

- It works well together with the Giant Button, because it de-selects the last unit when clicking "Next Unit". So the player gets presented a unit overview in the unit-corner (lower left). It would show something like `< 3 units idle >` or `< no units idle >`.
- if we wanted go go crazy, we could also show how many units are only "Waiting", and could technically still receive orders by the player (that's also what the unit-info-pane is cycling through): `< 3 units idle, 5 units waiting >`

najevi:
> * if we wanted go go crazy, we could also show how many units are only "Waiting", and could technically still receive orders by the player (that's also what the unit-info-pane is cycling through): `< 3 units idle, 5 units waiting >`
> 
> * From the player's perspective, [waiting units are not idle](https://github.com/yairm210/Unciv/pull/12814#issuecomment-2591876341).

I said it elsewhere and repeat it here. This summary table idea is the _best idea_ I have read to date that makes an effort to direct user's attention to the lower left of the WorldView screen. 

I don't think it hurts for that proposed summary table to _fully_ account for _all_ units the user is responsible for managing. 
Ideally the prev/next buttons (for this summary table) would be in the identical on-screen position as when the Unit/City-Info-pane is displayed.

itanasi:
> The hotkey `/` is not a good choice on many keyboard layouts. If you don't have a preference and technically possible, please change it to
> 
>     * `,` - next unit
> 
>     * `shift + ,` - previous unit

Shift-, isn't supported. But this is all rebindable by players, and this works for US QWERTY, so just going to stick with it.

As for other features, those can be in a different PR.

sulai:
> Shift-, (which is <) isn't supported. But this is all rebindable by players, and this works for US QWERTY, so just going to stick with it.

Let's map `.` (stop unit) to another hotkey then and use `,` and `.` for cycling. To my knowledge this is "industry standard".

This feature is rather important and using `/` effectively breaks it for many people. Note that there are [more native spanish speakers](https://en.wikipedia.org/wiki/List_of_languages_by_number_of_native_speakers) than native english speakers around the world. Same for german keyboard layout and probably other european languages too. [Cyrillic layouts](https://en.wikipedia.org/wiki/JCUKEN) also encourage the use of `.` for "forward", while `/` is not a good choice.

Stopping units is far less important and I don't mind if we put that one on `/` (or any other key), and use `.` to cycle forward.

> As for other features, those can be in a different PR.
Agreed.

core/src/com/unciv/ui/screens/worldscreen/unit/IdleUnitButton.kt
```diff
@@ -27,10 +32,12 @@ class IdleUnitButton (
...
         add(image).size(imageSize).pad(10f,20f)
         enable()
-        onClick {
+        keyShortcuts.add(keyShortcutBind)
+        addTooltip(keyShortcutBind)
+        onActivation {
 
             val idleUnits = unitTable.worldScreen.viewingCiv.units.getIdleUnits()
-            if (idleUnits.none()) return@onClick
+            if (idleUnits.none()) return@onActivation
```
sulai, line 40: What's the advantage of onActivation over onClick? (just curiosity)

sulai (approved):
Looks good!

najevi:
Clarifying question:

Do "," and "." behave like 
1. previous & next buttons in the Unit-Info-pane?   OR
2. like reverse & forward cycling using the Giant (Next unit) button?

Or perhaps, is this key-binding dependent upon each user's checking/clearing the option "Check for idle units" ?

sulai:
They are bound to the triangular cycle buttons in the unit info panel.

PR #12832, opened 2025-01-19
UnitTable: show a summary when no unit is selected

sulai:
Changes
- When nothing is selected, show idle and waiting units, together with the cycle buttons
- allows to start cycling *anytime* with the arrow buttons and the shortcuts `,.`
- Refactor UnitTable to have separate Presenters for units, cities, spies and no selection.
- at the cost of making private tables internal in order to populate them from the Presenters

grafik


sulai:
UnitTable mixes logic, UI and data, which is not ideal, but it *also* handled all different entities in one class. I separated out the Presenters for units, cities, spies and non-selection, so it is easier to extend and work with now. It's not perfect though, because it still mixes logic, UI and data, but I didn't want to rewrite everything at the risk of breaking things.

yairm210:
Separating "presenters" -> Fantastic 👍🏿 
I don't like the with{} and let{} but as a temporary measure for separation I'm fine with it

sulai:
I checked the occurrences of `with(unitTable) { }`, I use them because unitTable is referenced almost every line:

Most instances look like this without with {}
grafik

I could replace it with `unitTable.apply { }` though if you like?

core/src/com/unciv/ui/screens/worldscreen/unit/UnitTable.kt
```diff
@@ -1,90 +1,101 @@
...
-    var selectedSpy: Spy? = null
-
-    fun selectSpy(spy: Spy?) {
-        selectedSpy = spy
-        selectedCity = null
-        selectedUnits.clear()
-        selectedUnitIsSwapping = false
-        selectedUnitIsConnectingRoad = false
-    }
+    var shouldUpdate = false
```
sulai, line 56: @yairm210 do you think we could remove `shouldUpdate`, which was previously called `selectedUnitHasChanged`?

The in-code comment says:
```
// This is so that not on every update(), we will update the unit table.
// Most of the time it's the same unit with the same stats so why waste precious time?
```
This mechanism has almost no effect (anymore), because `UnitTable.update()` isn't called frequently. It would make the code a bit more concise, even though it currently also works with shouldUpdate in place.

yairm210: It's not? If the worldscreen updates and a unit is selected (which is most often the case) then the table would need to update as well, right? 

I'm not sure what the actual performance change is, but it's probably the for a reason? If you want to perf-test this be my guest, but 'simpler code at the cost of unknown performance degradation' is not a trade off I'm happy to make

sulai: The `WorldScreen.shouldUpdate` works so efficiently though that UnitTable.update() gets called very rarely, only on user interaction. Anyway, let's keep the optimization in UnitTable, since it's already working :+1:

yairm210 (approved):

yairm210:
Self conflicts

yairm210:
Wait, what is the X for? Does it close it?

sulai:
That's an old screenshot and got fixed, that window never shows an (x).

Also the unit panel in general has a lot of potential. I wish I had more time for unciv 🙂

PR #12834, opened 2025-01-19
Added an UniqueType to Found puppet city.

Emandac:
with "Founds a new puppet city" in "uniques" of an unit in Units.json. You can now settle a puppet city.


core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsFromUniques.kt
```diff
@@ -39,8 +39,11 @@ object UnitActionsFromUniques {
      * (no movement left, too close to another city).
      */
     internal fun getFoundCityAction(unit: MapUnit, tile: Tile): UnitAction? {
-        val unique = UnitActionModifiers.getUsableUnitActionUniques(unit, UniqueType.FoundCity)
-            .firstOrNull() ?: return null
+        // FoundPuppetCity is to found a puppet city for modding.
+        val unique = UnitActionModifiers.getUsableUnitActionUniques(unit,
+            UniqueType.FoundCity).firstOrNull() ?: 
+            UnitActionModifiers.getUsableUnitActionUniques(unit,
+            UniqueType.FoundPuppetCity).firstOrNull() ?: return null
```
SeventhM, line 46: I kinda feel like if there's going to be a unique for founding puppets, it should be as a separate option from founding regular cities. This here instead makes it only one or the other (preferring non puppet uniques), which imo might not match a modder's intent

Emandac: Hmm so to be able to have the two unique at one ?

yairm210: I think for now we can add this, if there's anyone in the future who wants a unit with both for some reason we can change it then

yairm210 (approved):

core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsFromUniques.kt
```diff
@@ -54,12 +57,19 @@ object UnitActionsFromUniques {
         ) == true }
         val foundAction = {
             if (unit.civ.playerType != PlayerType.AI)
-                UncivGame.Current.settings.addCompletedTutorialTask("Found city")
-            unit.civ.addCity(tile.position, unit)
+                // Now takes on the text of the unique.
+                UncivGame.Current.settings.addCompletedTutorialTask(
+                    unique.text)
```
SomeTroglodyte, line 62: This change breaks the Tutorial "Found a city!" - it will never successfully complete now, as the key this records in GameSettings.tutorialTasksCompleted ("Founds a new city <by consuming this unit>") isn't the one that can clear the event. The literal string was there for a reason. I know this is months old, just recording it for anybody looking. Will fix.

PR #12845, opened 2025-01-22
Add small Skip/Cycle Button

itanasi:
Co-authored with @sulai. Continuation from #12799

Adds small button next to Big Button that lets you Skip or Cycle through the current due unit list.

It also renames Wait command to Skip to be more consistent.

image
If you change the Next Unit button to instead mark the unit as Done (`due=true`), then a different Cycle button will appear.
image

Working vertical orientation too
image
image

Update in SummaryPresenter to match verb
image

#12785


android/assets/jsons/translations/template.properties
```diff
@@ -1177,6 +1177,7 @@ Next unit =
 [amount] units idle = 
 [idleCount] idle = 
 [waitingCount] waiting =
```
sulai, line 1179: need to rename that one to `[skipCount] skipping turn`

core/src/com/unciv/ui/screens/worldscreen/status/SmallUnitButton.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.unciv.ui.screens.worldscreen.WorldScreen
+
+class SmallUnitButton(
+    private val worldScreen: WorldScreen,
+    private val statusButtons: StatusButtons
+) : IconTextButton("", null, fontColor = NextTurnAction.NextUnit.color) {
+
+    private val nextLabel = "Cycle"
+    private val skipLabel = "Skip"
+    private var isSkip = worldScreen.game.settings.checkForDueUnitsCycles
```
sulai, line 21: minor: could be inlined

core/src/com/unciv/ui/screens/worldscreen/status/StatusButtons.kt
```diff
@@ -1,44 +1,47 @@
...
-        rowRight()
-        if (autoPlayStatusButton != null) {
-            addActor(autoPlayStatusButton)
-        }
-        if (multiplayerStatusButton != null) {
-            addActor(multiplayerStatusButton)
+        add(nextTurnButton)
+    }
+    
+    fun update(portrait: Boolean) {
```
sulai, line 19: Parameter `portrait` is misleading, better use `wrap` or `wrapped` or `vertical`.

core/src/com/unciv/ui/screens/worldscreen/status/StatusButtons.kt
```diff
@@ -1,44 +1,47 @@
...
+                row()
+                add(multiplayerStatusButton).padTop(padYSpace).right()
+            }
+        } else {
+            if (multiplayerStatusButton != null)
+                add(multiplayerStatusButton).padRight(padXSpace).top()
+            if (autoPlayStatusButton != null)
+                add(autoPlayStatusButton).padRight(padXSpace).top()
+            if (smallUnitButton != null)
+                add(smallUnitButton).padRight(padXSpace).top()
```
sulai, line 41: I'd prefer `obj?.let { add(it).pad... }` for all these null checks, it avoids having to repeat the variable and all of them can be one-liners.

sulai (changes requested):
Great work on making StatusButtons extend Table! :+1:

The only real issue is that one instance of "waiting" need to be renamed to "skipping".

core/src/com/unciv/ui/screens/worldscreen/status/StatusButtons.kt
itanasi: Fixed

core/src/com/unciv/ui/screens/worldscreen/status/StatusButtons.kt
itanasi: Done

Ouaz:
> Any changes still wanted?

Good work. Having the two "select unit" behaviour (skip or cycle) available in the UI at the same time is very convenient.

However, the Idle Units civilopedia article is quite incomplete now, as some additional "select unit" options could be explained here. These lines (in bold below) could be added (unless you think it clutters the article):
____________

- Idle Units

Clicking 'Next unit' **jumps** to the next **idle** unit in the queue **(as listed in Overview -> Units). After issuing an order to an unit, it is also possible to jump automatically to the next unit by enabling 'Auto Unit Cycle' in Options -> Gameplay.**

_(Note: maybe 'Auto Unit Cycle" option could be renamed 'Jump automatically to next unit' to prevent confusion with the new "Cycle" button)_

If you don't want to move a unit this turn, you can skip it by clicking 'Next unit' again, or command the unit to 'Skip turn'.
If you want to disable the 'Next unit' feature entirely, you can toggle it in Menu -> Check for idle units.
If you'd rather 'Next unit' cycle the units and not mark them as Done, you can change it in the Options -> Gameplay menu.

You can also cycle through the idle units using the left and right triangles on the Unit Table in the lower left of the screen, **or by enabling the Skip/Cycle button in Options > Gameplay.**

____________ 

I already added these two sentences in the french translation, as I think it makes things clearer regarding the four "select unit" options available in Options > Gameplay:

image

PR #12847, opened 2025-01-22
Allow building improvements on terrains that only allow some improvements when it has multiple uniques

SeventhM:
https://www.youtube.com/watch?v=rB0Op5bQuJ0

fixes https://github.com/ravignir/RekMOD/issues/189


SeventhM:
Just when I find the solution, I find out it's the ugliest thing imaginable

SeventhM:
Because we're in the middle of a switch statement, I don't think there's room to avoid using `let` without converting this to an if chain

yairm210 (approved):

PR #12858, opened 2025-01-24
Resize map with drag gesture

sulai:
The whole minimap is a drag area which resizes the mini-map. Very useful on android to have a glimpse on the map, then move it back to where it was.

[Bildschirmaufzeichnung vom 2025-01-24 15-11-29.webm](https://github.com/user-attachments/assets/3253fbb1-c19e-4560-8d9e-8bbce09777ae)

Added an icon to indicate a resize handle to visualize that this is now possible. And another button which toggles full screen. Full screen collapses as soon as the player chooses a location to show. If map size is really small, no map buttons are shown to keep vision free and the whole area counts as click area to toggle the maximized map.

grafik


yairm210:
Interesting 😎

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapHolder.kt
```diff
@@ -55,13 +67,72 @@ class MinimapHolder(val mapHolder: WorldMapHolder) : Table() {
...
+        stack.addInTable(getMaximizeToggleButton(civInfo)).size(40f).bottom().right() // more click area
+        add(stack).bottom()
+        
         pack()
         if (stage != null) x = stage.width - width
+        
+        addListener(ResizeDragListener(civInfo))
+    }
+
+    private fun rebuildAndUpdateMap(civInfo: Civilization?) {
```
yairm210, line 84: Under what condition is civInfo null?

docs/Credits.md
```diff
@@ -794,6 +794,12 @@ HexaRealm tileset images by legacymtgsalvationuser69544 [here](https://github.co
 -   [Package](https://thenounproject.com/icon/package-1886048) by shashank singh for Mods
 -   [scenario planning](https://thenounproject.com/icon/scenario-6569968) by nicolas for Scenarios
 
+### WorldScreen
+
+- [Reduce](https://www.flaticon.com/free-icon/reduce_1817377) by Freepik
+- [Increase](https://www.flaticon.com/free-icon/increase_1817463) by Freepik
+- [Corner](https://www.flaticon.com/authors/muhamad-ulum) by Muhamad Ulum
```
yairm210, line 801: Minor, but this is not an image link, it's an author link

core/src/com/unciv/ui/screens/worldscreen/minimap/MinimapHolder.kt
```diff
@@ -55,13 +67,72 @@ class MinimapHolder(val mapHolder: WorldMapHolder) : Table() {
...
+            add(image).expand().size(20f).pad(8f).bottom().right()
+            touchable = Touchable.enabled
+            onActivation(toggle)
+            // map is really small: use whole minimap area as click area to maximize map
+            if (!shouldShowMapButtons()) {
+                isVisible = false
+                minimap.touchable = Touchable.disabled
+                this@MinimapHolder.apply { 
+                    touchable = Touchable.enabled
+                    onClick(toggle)
```
yairm210, line 120: So... If we call getMaximizeToggleButton twice, then two toggle actions would have been added to the minimap holder
Does this mean they'll both activate and cancel out?

sulai: No because in that case the button is invisible. But I'll restructure that part a bit to make it less confusing :+1:

yairm210 (approved):

PR #12865, opened 2025-01-27
Fix memory leak per turn - dispose of individual pixmaps immediately, dispose of atlases after new screen is available

yairm210:


yairm210:
release patch

PR #12868, opened 2025-01-27

JPKP-Kuhn:
Just translating the word "wrap" to portugese, sometimes the word was translated, but in Map Options it was not. And I added a translation to Small Continents in Map Options, it was in english


android/assets/jsons/translations/Brazilian_Portuguese.properties
```diff
@@ -397,6 +397,8 @@ Four Corners = Quatro Cantos
 Archipelago = Arquipélago
 Inner Sea = Mar interno
 Perlin = Perlin
+Fractal = Fractal
+Small Continents = Pequenos Continentes
```
yairm210, line 401: Good catch, I'll add these in so they're translatable

PR #12894, opened 2025-02-01
Added feature to save unitType promotion.

Emandac:
This is to save unitType promotion, to not have to redo it when you are building an another unit with the same unitType.
This can help if you are building units in bulk with the same unitType and promotions.


android/assets/jsons/translations/template.properties
```diff
@@ -113,6 +113,7 @@ Requires [PolicyOrNationalWonder] =
 Cannot be purchased = 
 Can only be purchased = 
 See also = 
+Build units with saved UnitType promotion =
```
sulai, line 116: I think it should not take valueable screen space in the city screen, having very limited screen space on Android in mind. If it has to be a checkbox in the city screen, keep it *as short as possible* by any means. For example:
`[ ] auto promote`

core/src/com/unciv/ui/screens/pickerscreens/PromotionPickerScreen.kt
```diff
@@ -191,9 +204,16 @@ class PromotionPickerScreen private constructor(
...
 
         topTable.add(promotionsTable)
+        saveUnitTypePromotionForCity()
         addConnectingLines(emptySet())
     }
-
+    
+    // adds the checkBoxs to choice to save unit promotion.
+    private fun saveUnitTypePromotionForCity() {
+        val checkBoxSaveUnitPromotion = "Save unitType promotions.".toCheckBox(saveUnitTypePromotion) {saveUnitTypePromotion = it}
```
sulai, line 213: The wording could be better.

What about

`[ ] save as default for <Warrior>`

in case you are promoting a warrior. And in the city screen the checkbox would say

`[ ] default promotions`

Emandac: I like this but instead ```Warrior``` it could be ```Sword``` to not mislead the player into thinking it's only for the warrior unit😅

Emandac: Done

android/assets/jsons/translations/template.properties
```diff
@@ -113,6 +113,8 @@ Requires [PolicyOrNationalWonder] =
 Cannot be purchased = 
 Can only be purchased = 
 See also = 
+Default promotions = 
+"Default promotions for [unitType] =
```
yairm210, line 117: Extra "

core/src/com/unciv/logic/city/City.kt
```diff
@@ -91,7 +92,15 @@ class City : IsPartOfGameInfoSerialization, INamed {
...
     var isPuppet = false
     var shouldReassignPopulation = false  // flag so that on startTurn() we reassign population
-
+    
+    // HashMap to check if we can build the unitType with the saved promotion. 
+    // (Saved promotion: meaning that the player checkbox to save the promotion when adding them to a unitType.)
+    // To not have to redo it for every unit build afterword.
+    var canBuildUnitTypeWithSavedPromotion = HashMap<String, Boolean>()
+    
+    // To stroe unitType to promotion.
```
yairm210, line 101: Remove comment, call this "unitTypeToPromotion"
We already know it's for city since it's in the city class

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -464,6 +464,24 @@ class CityConstructions : IsPartOfGameInfoSerialization {
         else if (construction is BaseUnit) {
             unit = construction.construct(this, null)
                 ?: return false // unable to place unit
+            
+            // checking if it's true that we should load saved promotion for the unitType
+            // Check if the player want to rebuild the unit the saved promotion
+            // and null check.
+            val savedPromotion = city.cityUnitTypePromotion[unit.baseUnit.unitType]
+            if (city.canBuildUnitTypeWithSavedPromotion[unit.baseUnit.unitType] != null &&
```
yairm210, line 472: Unnecessary check, "== true" is enough

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -464,6 +464,24 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
+                city.canBuildUnitTypeWithSavedPromotion[unit.baseUnit.unitType] == true &&
+                savedPromotion != null) {
+                
+                // check if current unit has enough ep
+                if (unit.promotions.XP >= savedPromotion.XP) {
```
yairm210, line 477: Can be merged into previous if

core/src/com/unciv/ui/screens/cityscreen/ConstructionInfoTable.kt
```diff
@@ -99,6 +100,18 @@ class ConstructionInfoTable(val cityScreen: CityScreen) : Table() {
                     selectedConstructionTable.add(button).padTop(5f).colspan(2).center().row()
                 }
             }
+            if (construction is BaseUnit) {
+                val unitType = construction.unitType
+                
+                if (city.canBuildUnitTypeWithSavedPromotion[unitType] != null) {
+                    row()
+                    add(city.canBuildUnitTypeWithSavedPromotion[unitType]?.let {
+                        "Default promotions".toCheckBox(
```
yairm210, line 109: Do not use .let here, instead use regular assignment
If this is bull then you'd get add(null) which is not what we want, probably

core/src/com/unciv/ui/screens/pickerscreens/PromotionPickerScreen.kt
```diff
@@ -77,6 +80,16 @@ class PromotionPickerScreen private constructor(
             rightSideButton.setText("Pick promotion".tr())
             rightSideButton.onClick(UncivSound.Silent) {
                 acceptPromotion(selectedPromotion)
+                
+                if (saveUnitTypePromotion) {
+                    val unitCurrentCity = unit.currentTile.getCity()
+                    
+                    if (unitCurrentCity != null) {
+                        // If you are clicked the save unitType promotion, you want the next unitType to have the same promotion.
+                        unitCurrentCity.canBuildUnitTypeWithSavedPromotion.put(unit.baseUnit.unitType,true)
```
yairm210, line 89: I'm unclear why we need this second hashmap

core/src/com/unciv/ui/screens/pickerscreens/PromotionPickerScreen.kt
```diff
@@ -191,9 +204,16 @@ class PromotionPickerScreen private constructor(
...
 
         topTable.add(promotionsTable)
+        saveUnitTypePromotionForCity()
         addConnectingLines(emptySet())
     }
-
+    
+    // adds the checkBoxs to choice to save unit promotion.
+    private fun saveUnitTypePromotionForCity() {
+        val checkBoxSaveUnitPromotion = "Default promotions for ${unit.baseUnit.unitType}".toCheckBox(saveUnitTypePromotion) {saveUnitTypePromotion = it}
```
yairm210, line 213: Wrap parameter in square brackets

yairm210 (changes requested):

core/src/com/unciv/ui/screens/pickerscreens/PromotionPickerScreen.kt
Emandac: This hashmap need for the checkbox at ConstructionInfoTable.kt to turn on and off the auto promotion of the unit

core/src/com/unciv/ui/screens/pickerscreens/PromotionPickerScreen.kt
Emandac: Like this `[saveUnitTypePromotion]` ?

Emandac: Ohhh ok 😅

core/src/com/unciv/ui/screens/pickerscreens/PromotionPickerScreen.kt
yairm210: So the name should be like "should" etc, "can" usually means "is there nothing preventing us while "should" means "do we want to". "Can" is about ability, "Should" is about desire.

More philosophically, Can and Should sit on opposite sides of Hume's Guillotine.

core/src/com/unciv/ui/screens/cityscreen/ConstructionInfoTable.kt
```diff
@@ -99,6 +100,14 @@ class ConstructionInfoTable(val cityScreen: CityScreen) : Table() {
                     selectedConstructionTable.add(button).padTop(5f).colspan(2).center().row()
                 }
             }
+            if (construction is BaseUnit) {
+                val unitType = construction.unitType
+                
+                if (city.unitTypeToPromotion[unitType] != null) {
```
yairm210, line 106: Extract city.unitTypeToPromotion[unitType] into val - you use it 3 times here

Emandac: The `{city.unitTypeToPromotion[unitType] = it}` it's to avoid adding an another line just to assign it, in the updated version.

Emandac:
Is it normal that I still have the "changes requested"  😅

core/src/com/unciv/ui/screens/cityscreen/ConstructionInfoTable.kt
```diff
@@ -99,6 +100,16 @@ class ConstructionInfoTable(val cityScreen: CityScreen) : Table() {
...
                 }
             }
+            if (construction is BaseUnit) {
+                val unitType = construction.unitType
+                
+                val buildUnitWithPromotions = city.unitTypeToPromotion[unitType]
+                
+                if (buildUnitWithPromotions != null) {
+                    row()
+                    add("Default promotions".toCheckBox(buildUnitWithPromotions) {city.unitTypeToPromotion[unitType] = it}).colspan(2).center()
```
yairm210, line 110: Why is this not "for unittype" like the other one?

sulai: To save space on android-portrait. Because this text is in the info table of the unit, the context is set and it's clear that the checkbox counts for this unit type only.

"Default promotions" is very short and descriptive, but I could also live with
"use Default promotions" if that sounds better

Emandac: Done and i also added that clicking a promotion button when save promotion is on to save it too. 
Just like promotion pick button.

core/src/com/unciv/logic/city/City.kt
```diff
@@ -91,7 +92,11 @@ class City : IsPartOfGameInfoSerialization, INamed {
     var hasSoldBuildingThisTurn = false
     var isPuppet = false
     var shouldReassignPopulation = false  // flag so that on startTurn() we reassign population
-
+    
+    var unitTypeToPromotion = HashMap<String, Boolean>()
+
```
yairm210, line 97: Again, rename these :)

Emandac: Like what could be a good name for this variable 😅

yairm210: What exactly are you saving here? unitType is the key, sure but the value isn't promotion, is it?

Similarly the bottom hashmap
Maps should almost always be named in such a way that you can understand what is being mapped to what - "unitTypeTo" is a great start, e.g. "unitTypeToShouldUseSavedPromotion" or something

yairm210: Shorter is better, but "long and understandable" is better than "short but not understandable"

Emandac: would 'unitTypeBuildWithSavePromtion' be a good name ?

Emandac: a boolen

Emandac: maybe `BuildUnitTypeWithSavedPromtion`

yairm210: What does the boolean signify though?

Emandac: If the boolen is true build unit with save promotion

yairm210: I don't understand - per city there are saved promotions per type, right? 
And also you can decide to... Not use them?

Emandac: Yes like the player can build a unit without the saved promotion, like that with that unit they can change the save promotion to something else

yairm210: But this config will stick to all other units of this type?
By type do you mean like "archery" or like "archer"?

Emandac: Yes

yairm210:
I think it's good enough for now, can always change it later

PR #12914, opened 2025-02-05
Add turn start trigger

SeventhM:
I have no clue what the order of operations should be


yairm210:
Looks fine, if we want to change order of operations later it's a super simple change

yairm210 (approved):

PR #12924, opened 2025-02-07
Revert "Add small Skip/Cycle Button (#12845)"

yairm210:
This reverts commit 26c2a8ccdbf5c2e54e2b7af374187685ebe2637a.

This is an attempt to solve the notifications rendering problem that started in the last version - this is the only code that touched the notifications
My suspicion is that this makes it update / init / render not on the main render thread, somehow
This is a test


yairm210:
release patch

PR #12932, opened 2025-02-10
Add Shuffle Civ option

itanasi:
Enable to shuffle Major Civs turn order at game start.

Also makes sure Spectator is always first in turn.

image


android/assets/jsons/translations/template.properties
```diff
@@ -375,6 +375,7 @@ Start game! =
 Map Options = 
 Game Options = 
 Civilizations = 
+Shuffle Civ Order\nat Start =
```
yairm210, line 378: Without line break

itanasi: Done

PR #12960, opened 2025-02-14
Change save promotion from unitType to just baseUnit

Emandac:
This change comes from me messing around with save promotion in promotion picker and finding weird edge cases.
Like there could be a unit in the unitType that has promotion by default when it is build that the other unit don't have,
So when saving that unit promotion, it could cause the other unit with the same unitType to have the promotion that the unit shouldn't have.

By making it BaseUnit instead of UnitType avoid units using saved promotion having promotion that they shouldn't have.

I am also really sorry that i didn't caught this edge case Pre-PR.


yairm210 (approved):

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -188,6 +188,8 @@ enum class UniqueType(
     // Todo: Lowercase the 'U' of 'Units' in this unique
     CityHealingUnits("[mapUnitFilter] Units adjacent to this city heal [amount] HP per turn when healing", UniqueTarget.Global, UniqueTarget.FollowerBelief),
 
+    ExperienceForPromotionModifier("[relativeAmount]% Experience required for promotions",UniqueTarget.Global,UniqueTarget.FollowerBelief),
```
yairm210, line 191: We call it "XP"
Also, how exactly is this a follower belief?

Emandac: To allow modders a follower Belief with this unique

yairm210: Follower beliefs are for things that affect cities 
This won't work as a follower belief
I assume you haven't tried it? If you have and it works, there's something freaky going on

Emandac: Ah I am just going tl remove it then 😅

EmperorPinguin: > We call it "XP" Also, how exactly is this a follower belief?

The game is not fully consistent on this actually, it's 
"Experience" in UniqueType.UnitStartingExperience
"Xp" in UniqueType.CityStateGiftedUnitsStartWithXp
"XP" in the other uniques

yairm210: It should be XP everywhere
UnitStartingExperience needs to be renamed

yairm210: Nah CityStateGiftedUnitsStartWithXp also has "XP" in the text, only the unitstarting is bad

core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
```diff
@@ -48,10 +49,23 @@ class UnitPromotions : IsPartOfGameInfoSerialization {
...
 
     /** @return the XP points needed to "buy" the next [count] promotions. */
-    fun xpForNextNPromotions(count: Int) = (1..count).sumOf { (numberOfPromotions + it) * 10 }
+    fun xpForNextNPromotions(count: Int): Int {
+        // reduce or hieghter the xp for next promotion.
+        for (unique in unit.civ.getMatchingUniques(UniqueType.XPForPromotionModifier)) {
+            return Math.round(((1..count).sumOf { (numberOfPromotions + it) * 10 } * unique.params[0].toPercent()))
+        }
+        return (1..count).sumOf { (numberOfPromotions + it) * 10 }
+    }
```
yairm210, line 68: How about we deal with this in a different PR. I don't like the duploication - we should have just "costOfPromotion(existingPromotions: Int)" and the "next N" should use that function

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -188,6 +188,8 @@ enum class UniqueType(
     // Todo: Lowercase the 'U' of 'Units' in this unique
     CityHealingUnits("[mapUnitFilter] Units adjacent to this city heal [amount] HP per turn when healing", UniqueTarget.Global, UniqueTarget.FollowerBelief),
 
+    XPForPromotionModifier("[relativeAmount]% Experience required for promotions",UniqueTarget.Global,UniqueTarget.FollowerBelief),
```
yairm210, line 191: You didn't actually change the text here -_-

yairm210:
This is a classic example of why different issues should be in different PRs, mixing them makes the whole thing take longer and be more confusing :)

Emandac:
Agreed :d

core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
Emandac: The function would be something like this ?
```
        for (unique in unit.civ.getMatchingUniques(UniqueType.XPForPromotionModifier)) {
            return Math.round((existingPromotions * unique.params[0].toPercent()))
        }
```

Emandac:
So now i create a new PR branch for promotion XP modifier ?

yairm210:
Whichever you want
We can keep both here if you don't mind them delaying each other

Emandac:
I think i will just let this PR get merge before doing anything else 😅

yairm210:
You know what let's separate these...
Please remove the xp unique stuff from here so we can merge it

Emandac:
Never again updating 2 separate thing on a same PR.

Emandac:
I am really sorry that i cause all the trouble :/

core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -191,6 +191,8 @@ enum class UniqueType(
     // Todo: Lowercase the 'U' of 'Units' in this unique
     CityHealingUnits("[mapUnitFilter] Units adjacent to this city heal [amount] HP per turn when healing", UniqueTarget.Global, UniqueTarget.FollowerBelief),
 
+    XPForPromotionModifier("[relativeAmount]% XP required for promotions",UniqueTarget.Global),
```
yairm210, line 194: Remove

Emandac: Done

yairm210 (approved):

PR #12968, opened 2025-02-15
Update Dutch.properties

EmperorPinguin:


yairm210:
Placeholder `[resource]` not found in `Dutch` for entry `Requires improved [resource] near city`

EmperorPinguin:
Ah, typo. I thought it was failing because of a problem with base Unciv.

PR #12975, opened 2025-02-19
Version rollout

yairm210:


pipo-gold:
I am trying to submit a bug about negative XP's for a unit but github wont allow me. Same thing a few week ago 
with a bug about wrong display of Artillery promotions not showing "Extended Range".
I never submit neither bugs. 
Is there another way to submit bugs or what do I have to do?
I am not extremely familiar with github, maybe I am missing something nut I successfully submitted bugs before here.

PR #12983, opened 2025-02-19
Move Policy cost modifiers to mod constants

EmperorPinguin:
G&K apparently uses a modifier of 3 instead of 6 and an exponent of 2.01 instead of 1.7.

### DIFF
--- a/core/src/com/unciv/models/ModConstants.kt
+++ b/core/src/com/unciv/models/ModConstants.kt
@@ -98,6 +98,11 @@ class ModConstants {
     var pantheonBase = 10
     var pantheonGrowth = 5
 
+    // Factors in formula for policy costs
+    var policyBase = 25
+    var policyMultiplier = 6
+    var policyExponent = 1.7
+
     var workboatAutomationSearchMaxTiles = 20
 
     // Civilization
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/models/ModConstants.kt
```diff
@@ -98,6 +98,11 @@ class ModConstants {
     var pantheonBase = 10
     var pantheonGrowth = 5
 
+    // Factors in formula for policy costs
+    var policyBase = 25
+    var policyMultiplier = 6
+    var policyExponent = 1.7
```
yairm210, line 104: I actually really don't like this, I'm thinking of using something like https://github.com/fasseg/exp4j to allow arbitrary functions given base values but this will require experimentation

EmperorPinguin: Civ6/Civ7 uses a predetermined cost per policy, such would probably be better implemented in a similar way as how tech costs are implemented currently. I'm not sure if there's much demand for highly costomizeable culture cost formulas, the difference in the formulae for regular policies in G&K and BNW can be implemented via mod constants like this, although Ideology policies are in their own 'pool' for culture cost calculation.

EmperorPinguin: According to people in a Civ5 discord server, the ideology policies are in the same 'pool' as the regular policies, so I think the Civfanatics page where I read this is inaccurate on this.

yairm210:
This too should be replaced with a unique

PR #12986, opened 2025-02-20
Add tech cost uniques

EmperorPinguin:
Implemented similary as the Policy uniques, a LessTechCostFromCities unique is needed to implement Communism in Rekmod.

### DIFF
--- a/core/src/com/unciv/logic/civilization/managers/TechManager.kt
+++ b/core/src/com/unciv/logic/civilization/managers/TechManager.kt
@@ -25,6 +25,7 @@ import com.unciv.models.ruleset.unique.UniqueType
 import com.unciv.models.ruleset.unit.BaseUnit
 import com.unciv.models.translations.tr
 import com.unciv.ui.components.MayaCalendar
+import com.unciv.ui.components.extensions.toPercent
 import com.unciv.ui.components.fonts.Fonts
 import com.unciv.utils.withItem
 import kotlin.math.ceil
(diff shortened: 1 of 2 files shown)


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -156,6 +156,10 @@ enum class UniqueType(
     LessPolicyCostFromCities("Each city founded increases culture cost of policies [relativeAmount]% less than normal", UniqueTarget.Global),
     LessPolicyCost("[relativeAmount]% Culture cost of adopting new Policies", UniqueTarget.Global),
 
+    /// Tech Cost
+    LessTechCostFromCities("Each city founded increases science cost of technologies [relativeAmount]% less than normal", UniqueTarget.Global),
```
yairm210, line 160:
```suggestion
    LessTechCostFromCities("Each city founded increases Science cost of Technologies [relativeAmount]% less than normal", UniqueTarget.Global),
```
Let's not big ourselves into a deeper hole ;)

core/src/com/unciv/logic/civilization/managers/TechManager.kt
```diff
@@ -119,7 +120,10 @@ class TechManager : IsPartOfGameInfoSerialization {
         techCost /= getScienceModifier(techName)
         val mapSizePredef = civInfo.gameInfo.tileMap.mapParameters.mapSize.getPredefinedOrNextSmaller()
         techCost *= mapSizePredef.techCostMultiplier
-        techCost *= 1 + (civInfo.cities.size - 1) * mapSizePredef.techCostPerCityModifier
+        var cityModifier = (civInfo.cities.size - 1) * mapSizePredef.techCostPerCityModifier
+        for (unique in civInfo.getMatchingUniques(UniqueType.LessTechCostFromCities)) cityModifier *= 1 - unique.params[0].toFloat() / 100
```
yairm210, line 124: This should definitely be adjacent to the multiplication, the second unique interferes in the middle for no reason
Not a blocker but yes annoying to look at

PR #12989, opened 2025-02-21
Move CS tribute modifiers to mod constants

EmperorPinguin:
I'm not entirely sure about the naming, is this fine?


yairm210:
What do these modifiers actually mean? Like 100 means what, exactly?

EmperorPinguin:
> What do these modifiers actually mean? Like 100 means what, exactly?

The 'points' you get from having army nearby the city. Currently if you have largest Force score, you get 100 points towards reaching the treshold for CS tributes. This is supposed to be 75 in BNW and 50 in Rekmod. If your Force nearby the city state is 3 times larger that the CS's Force, you get 100 points towards reaching the treshold for CS tributes. This is supposed to be 125 in BNW and 150 in Rekmod.

EmperorPinguin:
Maybe rename to "globalForceRankModifier" and "localForceRankModifier"?

EmperorPinguin:
@yairm210 is there something wrong with the contents of this PR? I would like this to be moddable for a next Rekmod game.

yairm210:
No this is fine, I just forgor 😓

PR #12992, opened 2025-02-22
Update SpecificUnitAutomation.kt

EmperorPinguin:
Resolves https://discord.com/channels/586194543280390151/1071804673956266096/1334462051774369895 in my testing, and hopefully also https://discord.com/channels/586194543280390151/676559694864842766/1341517147310129276. The drawback is that the AI will probably found more crappy tundra/snow cities when this isn't needed, a better way to go about it may be to take into account the average tile yield on the map.

### DIFF
--- a/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
+++ b/core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
@@ -110,6 +110,9 @@ object CityLocationTileRanker {
         if (newCityTile.hasViewableResource(civ) && newCityTile.tileResource.resourceType == ResourceType.Bonus) tileValue -= 8
         // Settling on bonus resources tends to waste a food
         // Settling on luxuries generally speeds up our game, and settling on strategics as well, as the AI cheats and can see them.
+        
+        tileValue -= civ.cities.size * 5 
+        //Empire-based penalty for tile quality, to allow settling on spawn even if it's bad spawn, but try to limit city sprawl on tundra later on
 
         var tiles = 0
         for (i in 0..3) {
--- a/core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
+++ b/core/src/com/unciv/logic/automation/unit/SpecificUnitAutomation.kt
@@ -113,7 +113,7 @@ object SpecificUnitAutomation {
 
         // It's possible that we'll see a tile "over the sea" that's better than the tiles close by, but that's not a reason to abandon the close tiles!
         // Also this lead to some routing problems, see https://github.com/yairm210/Unciv/issues/3653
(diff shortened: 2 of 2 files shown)


yairm210:
I'm against this, most players play the default tileset and this solves a problem in niche tilesets while making the general AI worse.

EmperorPinguin:
More cities generally make the AI stronger, I don't it too bad in terms of playing strength, it's more of a tradeoff between having more cringe cities in base game vs not being able to found cities at all with unusual map settings/mod.

The 50 points cutoff what chosen when the AI could still see resources (lots of points from those resources), the complaint about snow maps quite specifically states this behaviour didn't use to be the case some time ago, maybe I could verify if they provided a safe file.

EmperorPinguin:
Also, city states moving instead of settling t1 is considered a bit of a problem nowadays.

yairm210:
I have no problem for city states to settle their first city immediately

core/src/com/unciv/logic/automation/unit/CityLocationTileRanker.kt
```diff
@@ -110,6 +110,9 @@ object CityLocationTileRanker {
         if (newCityTile.hasViewableResource(civ) && newCityTile.tileResource.resourceType == ResourceType.Bonus) tileValue -= 8
         // Settling on bonus resources tends to waste a food
         // Settling on luxuries generally speeds up our game, and settling on strategics as well, as the AI cheats and can see them.
+        
+        tileValue -= civ.cities.size * 5
```
yairm210, line 114: This will mean that at 20 cities I'll basically be able to settle nowhere, no?

EmperorPinguin: Yes, it's a cap on the number of cities.

yairm210: Why would we want to disable AI from settling anywhere after X cities?

EmperorPinguin: The threshold for minimum tile score needs to be lower than 50 points, so the AI will settle also with bad spawn or weird ruleset/weird map settings, but them sprawling cities on empty snow/tundra/desert is not something a human would do and it feels 'unnatural' to the user. Setting the threshold low and increasing with more cities is maybe a way to balance these two problems.

yairm210: Yes, but this config would have them not sprawl even on super good tiles, which is also unnatural

EmperorPinguin: How do I look up the average value on land tiles on the map? The best way then may be to let the AI calculate this for the whole map, and then use that to set a baseline for minimum tile quality.

Issue #12676, opened 2024-12-17
Visualize "Can't buy unit because nowhere to place"


metablaster:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Game Version

4.14.16

### Describe the bug

In this save file to buy a worker it costs 700 gold.
I have 714 gold but can't buy the worker until next turn.

### Steps to Reproduce

1. Load save file
2. Go to city Rome
3. Try to buy worker
4. The button to buy worker is grayed out until next turn

### Screenshots

![gold](https://github.com/user-attachments/assets/151362c3-62e4-47c2-9fb6-57fbb24c5a1e)
![buy](https://github.com/user-attachments/assets/705696e8-b6e4-4518-96ce-9b438f99a0c2)

### Link to save file

[Rim - 90 poteza-cant-buy.zip](https://github.com/user-attachments/files/18164664/Rim.-.90.poteza-cant-buy.zip)

### Operating System

Linux

### Additional Information

This happened on same turn when settler was built.

metablaster:
Oh, I just figured out that was because there was settler unit in city tile, sorry!
Maybe the grayed out button should show a tooltip as to why a unit can't be bought?

yairm210:
I'm not sure how I would visualize this
Text would be far too big, what would be a good icon for "out of space"?

sulai:
I wouldn like to suggest to allow the user to press the button. A dialog shall appear explaining as to why it's not possible to buy a unit right now. This is in line with the confirmation dialog in case a purchase is possible.

As an additional icon to show on the button, maybe something in the shape (not color) like this?
https://images.app.goo.gl/2dWs9c9DkRFMm3S96

yairm210:
Oh, interesting 🤔 
That could be a theme for other places as well, we should have a "unavailable but clickable for information"-style - probably a grayish-blue, this could fit other "disabled" actions as well (thinking of unit actions)

sulai:
Or have the button disabled with a clickable question mark next to it.

PR #13012, opened 2025-02-28
tooltip for purchase blocked by unit

saejo:
It took me a while to figure out that the unit currently positioned in the city was the reason I couldn't buy units sometimes, so I figured this tooltip might be helpful.

If there is an alternative way to display this info that's more in line with the rest of the code base (and maybe also works for mobile), please let me know.

I duplicated parts of the logic for checking whether or not the unit can be purchased. Of course we could also change `isConstructionPurchaseAllowed` so it returns the reason in case a purchase is not allowed.


saejo:
@SomeTroglodyte thanks for your feedback! I removed the whitespace. I don't know this project (or kotlin for that matter) to have a strong opinion on the deduplication or your suggestion.

saejo:
Turns out there is a relevant issue: https://github.com/yairm210/Unciv/issues/12676

yairm210 (approved):
Definitely 'better than existing situation'
Not ideal since it didn't cover Android use case (no tooltip there) but can be improved later

PR #13015, opened 2025-03-01
Add unqiue change the promotion XP cost 

Emandac:
Re-add the unique in it's own PR branch.


core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
```diff
@@ -48,11 +49,19 @@ class UnitPromotions : IsPartOfGameInfoSerialization {
...
     
     /** @return the XP points needed to "buy" the next [count] promotions. */
-    fun xpForNextNPromotions(count: Int) = (1..count).sumOf { (numberOfPromotions + it) * 10 }
-
+    fun xpForNextNPromotions(count: Int) = costOfPromotion((1..count).sumOf { (numberOfPromotions + it) * 10 })
+    
+    /** return the existingPromotions cost with the XPForPromotionModifier Unique applied */
+    private fun costOfPromotion(existingPromotions: Int): Int {
+        for (unique in unit.civ.getMatchingUniques(UniqueType.XPForPromotionModifier)) {
+            return Math.round((existingPromotions * unique.params[0].toPercent()))
```
yairm210, line 60: I'm pretty sure you don't want 'return' - of there are multiple uniques they should all reduce the cost, not just the first one

yairm210: Also remove extra parentheses

core/src/com/unciv/logic/map/mapunit/UnitPromotions.kt
```diff
@@ -48,11 +49,19 @@ class UnitPromotions : IsPartOfGameInfoSerialization {
...
-    fun xpForNextPromotion() = (numberOfPromotions + 1) * 10
+    fun xpForNextPromotion() = costOfPromotion((numberOfPromotions + 1) * 10)
     
     /** @return the XP points needed to "buy" the next [count] promotions. */
-    fun xpForNextNPromotions(count: Int) = (1..count).sumOf { (numberOfPromotions + it) * 10 }
-
+    fun xpForNextNPromotions(count: Int) = costOfPromotion((1..count).sumOf { (numberOfPromotions + it) * 10 })
+    
+    /** return the existingPromotions cost with the XPForPromotionModifier Unique applied */
+    private fun costOfPromotion(existingPromotions: Int): Int {
```
yairm210, line 58: There is double counting here. 

Instead, both functions should use a new function 'xpForPromotion(promotionNumber)'

Emandac: Like one that only reduce once and other reduce over all the promotion ?

yairm210: Something like that
What we need is *base* promotion cost, for promotion number N
And then for the "many promotions" case we can sum them up

yairm210: No, like

```
fun baseCpForPromotionNumber(promotionNumber) = (numberOfPromotions + 1) * 10

fun xpForNextPromotion() = Math.round(baseXpForPromotionNumber(numberOfPromotions) * promotionCostModifier())


```

Emandac: I see, so like this 
```
    fun xpForNextPromotion() = Math.round(baseXpForPromotionNumber(numberOfPromotions) * promotionCostModifier())
    
    /** @return the XP points needed to "buy" the next [count] promotions. */
    fun xpForNextNPromotions(count: Int) = (1..count).sumOf { 
        Math.round(baseXpForPromotionNumber(numberOfPromotions,count) * promotionCostModifier()) 
    }

    private fun baseXpForPromotionNumber(numberOfPromotions: Int) = (numberOfPromotions + 1) * 10
    
    private fun baseXpForPromotionNumber(numberOfPromotions: Int, count: Int) = (numberOfPromotions + count) * 10
    
    private fun promotionCostModifier(): Float {
        for (unique in unit.civ.getMatchingUniques(UniqueType.XPForPromotionModifier)) return unique.params[0].toPercent()
        // base case if you don't have any the unique that reduce or higher the promotion cost
        return 1f
    }
```

yairm210: Almost...

- baseXpForPromotionNumber should only accept one number, you can do the "numberOfPromotions + count" in the sumOf
- Now that I see this, it's true that rounding can change the exactness of cost, so sum(round(cost)) != round(sum(cost))... so we should put the rounding and multiplication in the baseXpForPromotionNumber() function
- Again your promotionCostModifier() doesn't handle multiple uniques correctly, this should be "1f times unique modifier for each unique"

Emandac: Like this 

```
    fun xpForNextPromotion() = Math.round(baseXpForPromotionNumber(numberOfPromotions) * promotionCostModifier())
    
    /** @return the XP points needed to "buy" the next [count] promotions. */
    fun xpForNextNPromotions(count: Int) = Math.round((1..count).sumOf { 
        baseXpForPromotionNumber(numberOfPromotions) + count} * promotionCostModifier() )

    private fun baseXpForPromotionNumber(numberOfPromotions: Int) = (numberOfPromotions + 1) * 10
    
    private fun promotionCostModifier(): Float {
       
        var totalPromotionCostModifier = 0f
        for (unique in unit.civ.getMatchingUniques(UniqueType.XPForPromotionModifier)) {
            totalPromotionCostModifier += unique.params[0].toPercent()
        }
        // base case if you don't have any the unique that reduce or higher the promotion cost
        return if (totalPromotionCostModifier != 0f) totalPromotionCostModifier  else 1f
    }
```

yairm210 (approved):
OK!!

Emandac:
I was fast

PR #13016, opened 2025-03-02
Add a filter for religions

SeventhM:
And you get a filter, and you get a filter. Everything gets a multifilter

I mostly thought of this to allow for moving uniques for unlocking religious buildings off of the beliefs themselves and onto the buildings, but almost surely some other modder will find use for it

I really, really, really, really want to find some way to integrate this into city filter (as I believe there is actually a niche case use for that), but I can't figure out how I would go about doing that or if it's just the unique in question needs better alternatives

### DIFF
--- a/core/src/com/unciv/models/Religion.kt
+++ b/core/src/com/unciv/models/Religion.kt
@@ -132,7 +132,7 @@ class Religion() : INamed, IsPartOfGameInfoSerialization {
         val known = civ != null && civ.knows(foundingCiv)
         if (filter == "enemy") return known && civ!!.isAtWarWith(foundingCiv)
         if (founderBeliefUniqueMap.hasMatchingUnique(filter, state)) return true
-        if (founderBeliefUniqueMap.hasMatchingUnique(filter, state)) return true
+        if (followerBeliefUniqueMap.hasMatchingUnique(filter, state)) return true
         return false
     }
 
--- a/core/src/com/unciv/models/Religion.kt
+++ b/core/src/com/unciv/models/Religion.kt
@@ -122,18 +122,24 @@ class Religion() : INamed, IsPartOfGameInfoSerialization {
     }
     
     private fun matchesSingleFilter(filter: String, state: StateForConditionals = StateForConditionals.IgnoreConditionals, civ: Civilization? = null): Boolean {
-        if (filter == "any") return true
-        if (filter == name) return true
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/models/Religion.kt
```diff
@@ -114,6 +117,25 @@ class Religion() : INamed, IsPartOfGameInfoSerialization {
 
     fun getFounder() = gameInfo.getCivilization(foundingCivName)
 
+    fun matchesFilter(filter: String, state: StateForConditionals = StateForConditionals.IgnoreConditionals, civ: Civilization? = null): Boolean {
+        return MultiFilter.multiFilter(filter, { matchesSingleFilter(it, state, civ) })
+    }
+    
+    private fun matchesSingleFilter(filter: String, state: StateForConditionals = StateForConditionals.IgnoreConditionals, civ: Civilization? = null): Boolean {
+        if (filter == "any") return true
```
yairm210, line 125: For performance, turn this into a switch with all static cases, and have "else" contain the dynamic cases (name, hasMatchingUnique)

SeventhM: > nevertheless be called at most once - it will even be called "less" if you get to the else). Don't forget that `when(filter) { name -> true`... works

Java is one such language where switch statements requires a compile-time here if I understand correctly. So that also would compile to an if chain if I understand correctly rather than an actual jvm switch

> Also - aren't the two last `if` lines absolutely identical?

Goddammit

SeventhM:
Actually, question: should this accept a civFilter as well? E.g. ``in cities following a [England] religion``. I didn't originally plan on that, and I'm not sure if it's worthwhile to add or not

yairm210:
I think we can forgo that for now, if no one is asking for it it sounds kind of 'out there'

core/src/com/unciv/models/Religion.kt
yairm210: @SomeTroglodyte context: https://medium.com/@yairm210/kotlin-when-string-optimization-e15c6eea2734

yairm210: In all JVM languages, interning is not default behavior because it's essentially a hashmap getOrPut which is not cheap
I did consider doing it for RAM purposes but didn't see it made a difference

yairm210 (approved):

PR #13040, opened 2025-03-07
Update Spanish.properties

jjesus-dev:
- remove `¤` character in the translation when its not present in the original string.
- correct spelling errors.
- modify some words to make them more neutral (localization wise).


jjesus-dev:
Hi @yairm210 

There is a check that didn't pass (Detekt), do I need to do something else to fix the PR? like correct a file or something else?

Sorry, I'm not familiar with this GitHub build action in particular.

PR #13045, opened 2025-03-07
Update auto save promotions

Emandac:
You can no longer apply auto promotion in puppet city and in cities that are not yours.
Also fix edge case where scout unit upgraded threw ruins could allow spearman built with auto promotion to have "ignore terrain cost".


core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -465,20 +466,28 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
             if (city.unitShouldUseSavedPromotion[unit.baseUnit.name] == true &&
                 savedPromotion != null && unit.promotions.XP >= savedPromotion.XP) {
-                    // sorting it to avoid getting Accuracy III before Accuracy I
-                    for (promotions in savedPromotion.promotions.sorted()) {
-                        if (unit.promotions.XP >= savedPromotion.XP) {
-                            unit.promotions.addPromotion(promotions)
-                        } else {
-                            break
-                        }
+                // sort it to avoid getting Accuracy III before Accuracy I
```
yairm210, line 484: Instead of 'sort by name' we really should be creating a dependency tree and starting from the leaves, working upwards
So like
"Choose the first promotion that we **can add** to the unit, rinse and repeat as long as we have enough XP"

yairm210:
We should also be reseting these when the city moves between Civs 🤔

core/src/com/unciv/logic/city/CityConstructions.kt
Emandac: But what if the player while uses a mod choses a promotion that is not connected to the rest ?

yairm210: That's fine, as long as **it's not blocked by anything** we shouldn't care about the order (that is: we have no way of guessing)

Emandac:
I think so

core/src/com/unciv/logic/city/CityConstructions.kt
Emandac: But how should I go about implementing dependency tree for the promotion ?

Emandac:
Done but i am not too sure how to add the dependency tree for applying auto promotion

yairm210 (approved):

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -465,20 +466,34 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
-                        if (unit.promotions.XP >= savedPromotion.XP) {
-                            unit.promotions.addPromotion(promotions)
-                        } else {
-                            break
-                        }
+                // this variable is the filted promotion from savedPrmotion to only get promotion that are possible for this unit.
+                val possiblePromotionFilted =  savedPromotion.promotions.sorted().filter { it in possiblePromotions }
+                // sort in order to avoid getting Accuracy III before Accuracy I
+                for (promotions in prmotionTreeOrder) {
+                    if (unit.promotions.XP >= savedPromotion.XP) {
```
yairm210, line 492: Reverse this 'if' so it's 'if(...) break'
Then you can remove the else

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -465,20 +466,34 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
-            // Check if the player want to rebuild the unit the saved promotion
-            // and null check.
-            // and finally check if the current unit has enough XP.
+            /* check if it's true that we should load saved promotion for the unitType,
+               Then check if the player want to rebuild the unit the saved promotion,
+               and do a null check.
+               and finally check if the current unit has enough XP. */
+            val possiblePromotions = hashSetOf<String>()
+            // to get promotion in order for nodes
+            val prmotionTreeOrder = mutableSetOf<String>()
```
yairm210, line 475: This is *not* ordered 🤔 
HashSet items are in random order

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -465,20 +466,34 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
                 savedPromotion != null && unit.promotions.XP >= savedPromotion.XP) {
-                    // sorting it to avoid getting Accuracy III before Accuracy I
-                    for (promotions in savedPromotion.promotions.sorted()) {
-                        if (unit.promotions.XP >= savedPromotion.XP) {
-                            unit.promotions.addPromotion(promotions)
-                        } else {
-                            break
-                        }
+                // this variable is the filted promotion from savedPrmotion to only get promotion that are possible for this unit.
+                val possiblePromotionFilted =  savedPromotion.promotions.sorted().filter { it in possiblePromotions }
```
yairm210, line 489: Sorting this shouldn't help either - all you use it for is a 'contains' check, right? So that's fine for an unordered hashset as well

core/src/com/unciv/logic/city/CityConstructions.kt
Emandac: That comment is for the promotionTreeOrder 😅

core/src/com/unciv/logic/city/CityConstructions.kt
Emandac: Done

core/src/com/unciv/logic/city/CityConstructions.kt
```diff
@@ -465,20 +466,36 @@ class CityConstructions : IsPartOfGameInfoSerialization {
...
-                            unit.promotions.addPromotion(promotions)
-                        } else {
-                            break
-                        }
+                // this variable is the filted promotion from savedPrmotion to only get promotion that are possible for this unit.
+                val possiblePromotionFilted =  savedPromotion.promotions.filter { it in possiblePromotions }
+                // sort in order to avoid getting Accuracy III before Accuracy I
+                for (promotions in prmotionTreeOrder) {
+                    // check if here is enough XP for the next promotion
+                    if (unit.promotions.XP-unit.promotions.xpForNextPromotion() >= 0) {
```
yairm210, line 494: reverse this 'if' so it's `if (...< 0) break`

PR #13062, opened 2025-03-12
Completing policy branches give leaders titles

govUA:
## **Accomplishments**

- [x] Civilization leaders get titles for their corresponding branch with the most policies adopted

- [x] Titles for adopting branches are written in Policies.json along with other data for branches.

## **Roadmap**

- [ ] Titles for female leaders

- [ ] Rework to be implemented as uniques

- [ ] Title era succession


yairm210:
- Granting Titles should be done via a Unique - this means adding a new UniqueType like e.g. "Grants the title [comment]"
- This will allow granting titles from anything - building certain wonders, tech research under conditions, etc
- The question then remains - if we have multiple titles, which one wins? I would vote for 'the last one you got' so that your title changes throughput the ages as you progress

yairm210:
Regarding male / female... Technically we could add a tag to civilizations like "male", "female" and then trigger by
"Grants the title [Lord] <for [Male] civilizations>"
"Grants the title [Lady <for [Female] civilizations>"
🤔 

Now that I read this more thoroughly I see 2 things:
- That it's not just granted at the end but by the majority... urgh, we'll need to think how to generify that
- That this should be a template - like "Grants the title [Lord [name]]", "Grants the title [[name] the Pious]", since these phrases need to be translated and the word order could be different in other languages

govUA:
Okay, will work on improvements

Ouaz:
> That this should be a template - like "Grants the title [Lord [name]]", "Grants the title [[name] the Pious]", since these phrases need to be translated and the word order could be different in other languages

For information, in latin languages (french, spanish, italian), the adjective form depends of the gender.

The Pious:

- French: `le Pieux` (male), `la Pieuse` (female)
- Spanish: `el Piadoso` (male), `la Piadosa`

Etc.

With most of the titles (roughly 6 of 10, depending the language), there will be a discrepancy when the leader is a woman.

Well, there's only a bunch of female leaders, but it will be weird to read `Marie-Thérèse le Pieux` or "Marie-Thérèse le Grand`. :D

I don't know how the translation was handled in Civ5, though (I bet there's a discrepancy there).

PR #13071, opened 2025-03-16
Civ5 map size

EmperorPinguin:
According to https://civilization.fandom.com/wiki/Map_(Civ5)

Maybe it's a better idea though to add the science/culture/happiness thresholds independently of the predefined map sizes, setting custom sizes for the screens is a bit ugly.

### DIFF
--- a/core/src/com/unciv/logic/map/MapParameters.kt
+++ b/core/src/com/unciv/logic/map/MapParameters.kt
@@ -52,7 +52,7 @@ class MapParameters : IsPartOfGameInfoSerialization {
     var type = MapType.pangaea
     // DO NOT CHANGE DEFAULTS since that changes all existing games to new default!
     var shape = MapShape.hexagonal
-    var mapSize = MapSize.Medium
+    var mapSize = MapSize.Tiny
     var mapResources = MapResourceSetting.default.label
     var mirroring: String = MirroringType.none
     var noRuins = false
--- a/core/src/com/unciv/logic/map/MapSize.kt
+++ b/core/src/com/unciv/logic/map/MapSize.kt
@@ -69,6 +69,7 @@ class MapSize private constructor(
          * used in [name] to indicate user-defined dimensions.
          * Do not mistake for [MapGeneratedMainType.custom]. */
         const val custom = "Custom"
+        val Duel get() = MapSize(Predefined.Duel)
         val Tiny get() = MapSize(Predefined.Tiny)
         val Small get() = MapSize(Predefined.Small)
         val Medium get() = MapSize(Predefined.Medium)
--- a/core/src/com/unciv/logic/map/MapSize.kt
+++ b/core/src/com/unciv/logic/map/MapSize.kt
@@ -50,23 +50,17 @@ class MapSize private constructor(
         val techCostMultiplier: Float = 1f,
         val techCostPerCityModifier: Float = 0.05f,
         val policyCostPerCityModifier: Float = 0.1f,
+        val unHappinesPerCity: Float = 3f
     ) {
-        Tiny(10, 23, 15),
-        Small(15, 33, 21),
-        Medium(20, 44, 29, 1.1f),
-        Large(30, 66, 43, 1.2f, 0.0375f,0.075f),
-        Huge(40, 87, 57, 1.3f, 0.025f,0.05f);
-        /* - ours are a little lighter than the original values. For reference those are:
-            Civ5Duel(40,24,17),
-            Civ5Tiny(56,36,25),
-            Civ5Small(66,42,30),
-            Civ5Medium(80,52,37),
-            Civ5Large(104,64,47),
-            Civ5Huge(128,80,58),
-         */
+        Duel(16, 40, 24),
+        Tiny(24, 56, 36),
+        Small(28, 66, 42),
+        Medium(34, 80, 52, 1.1f),
+        Large(43, 104, 64, 1.2f, 0.03f,0.075f, 2.4f),
+        Huge(54, 128, 80, 1.3f, 0.02f,0.05f, 1.8f);
 
         companion object {
-            fun safeValueOf(name: String) = values().firstOrNull { it.name == name } ?: Tiny
+            fun safeValueOf(name: String) = values().firstOrNull { it.name == name } ?: Duel
         }
     }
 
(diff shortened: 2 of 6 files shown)


EmperorPinguin:
> And "GlobalUniquesTests > unhappinessFromCitiesPercentageTest FAILED": You know you can run and debug the unit tests locally in Android Studio? 

No, actually not. How do I do so?

EmperorPinguin:
@SomeTroglodyte I think the test itself is flawed: when I start a game with India, it works as expected.

EmperorPinguin:
Ah, I forgot a sign change.

core/src/com/unciv/logic/map/MapParameters.kt
```diff
@@ -52,7 +52,7 @@ class MapParameters : IsPartOfGameInfoSerialization {
     var type = MapType.pangaea
     // DO NOT CHANGE DEFAULTS since that changes all existing games to new default!
```
yairm210, line 53: I remember really bad things happened with the default shape, so let's test what happens if you take a previously-Medium map and change this

yairm210:
I **did** pick the values specifically so Android phones won't bork, but we're a squillion light years ahead in terms of performance compared to when these values were originally determined.
Also, it's been like 6 years, which means the average **low end phone** is also much more powerful than it was

core/src/com/unciv/logic/map/MapParameters.kt
EmperorPinguin: You mean create a radius 20 map ad see if it does weird, or from an existing save?

yairm210: From an existing save

EmperorPinguin: I don't notice anything being changed to by saves, do I need to look for something specifically?

EmperorPinguin: Most of them are Large map though, not Medium.

AdityaMH:
And don't forget, j2 prime still exist.

core/src/com/unciv/logic/map/MapSize.kt
```diff
@@ -50,23 +50,17 @@ class MapSize private constructor(
         val techCostMultiplier: Float = 1f,
         val techCostPerCityModifier: Float = 0.05f,
         val policyCostPerCityModifier: Float = 0.1f,
+        val unHappinesPerCity: Float = 3f,
```
yairm210, line 53: This is 100% a **ruleset issue** and not something we should determine statically for all rulesets
Sounds like a different PR, plus we'll need to think about how exactly to do this - if a player has a custom map size, then what?

core/src/com/unciv/logic/map/MapSize.kt
```diff
@@ -50,23 +50,17 @@ class MapSize private constructor(
...
+        Duel(16, 40, 24),
+        Tiny(24, 56, 36),
+        Small(28, 66, 42),
+        Medium(34, 80, 52, 1.1f),
+        Large(43, 104, 64, 1.2f, 0.03f,0.075f, 2.4f),
+        Huge(54, 128, 80, 1.3f, 0.02f,0.05f, 1.8f);
 
         companion object {
-            fun safeValueOf(name: String) = values().firstOrNull { it.name == name } ?: Tiny
+            fun safeValueOf(name: String) = values().firstOrNull { it.name == name } ?: Duel
```
yairm210, line 63: Why not Small? Duel shouldn't be the default, when the default game is 4 players

EmperorPinguin: Why was it set to Tiny in the first place? It seems appropriate to update it to the new smallest map size.

EmperorPinguin:
I just discovered calculateScoreBreakdown() has also been rebased to a smaller map size than Civ5, would it be needed to adjust this one as well?

core/src/com/unciv/logic/map/MapSize.kt
EmperorPinguin: The techCostPercityModifier and policyCostPerCityModifier are also stored here. I haven't noticed those bugging with custom map sizes.

yairm210: I agree, those too should be rulest-changable

EmperorPinguin: You mean those should be added as mod constants for each of the map sizes? That's 3*6 =18 new mod constants.

yairm210: No, I'm saying this should be a **global unique** dependent on map size via conditional
Again, out of scope here. Let's remove this from this PR, turn this PR into just about changing the map sizes, and then we can discuss unhappiness per city per map size in another place.

When you try to add too many things in one PR things can get conflated, which makes the whole thing drag on way longer than it should. One thing at a time

EmperorPinguin: I don't really care about the predefined map sizes, this PR is to move the tresholds for science/culture cost modifiers :(

EmperorPinguin: It also possible to set the science/culture/happiness cost independently of the predefined map sizes, but I figured such would create more complications, not less.

EmperorPinguin: Here's the poll results, option B (keep it the same) got by far the least votes (12%): https://discord.com/channels/586194543280390151/618491418859798539/1350536363547689122

EmperorPinguin: > Let's remove this from this PR, turn this PR into just about changing the map sizes

@yairm210 should I set the science/culture/happiness modifier thresholds independently of the predefined map sizes? That way the map sizes can stay the same as before, I just thought it would create unneeded complexity.

yairm210: If we have multiple things that can change by map size, then they obviously need to be moddable.
The question is how. 
Option 1, is to extract all map sizes into a new json file, and then each one can apply certain global uniques
Option 2, is to have a conditional for map sizes
I'm not sure which is better, I'm leaning towards the former...

yairm210: 'per city' sounds like an existing countable

PR #13072, opened 2025-03-16
ProductionBonusWhenRemoved moddability

EmperorPinguin:
Scales production bonus with game speed, and adds option to scale it by game speed as well.

Rather than [amount], I'd like to add [stats] as input, so I can add e.g. the food + production jungle chops of Civ6, but how do I extract the stats from the unique in this case?

### DIFF
--- a/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
+++ b/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
@@ -309,7 +309,6 @@ class TileImprovementFunctions(val tile: Tile) {
         val closestCity = civ.cities.minByOrNull { it.getCenterTile().aerialDistanceTo(tile) }
             ?: return
         val distance = closestCity.getCenterTile().aerialDistanceTo(tile)
-
         var productionPoints = 0f
         for (unique in tile.getTerrainMatchingUniques(UniqueType.ProductionBonusWhenRemoved)) {
             productionPoints = unique.params[0].toFloatOrNull() ?: return
--- a/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
+++ b/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
@@ -309,19 +309,32 @@ class TileImprovementFunctions(val tile: Tile) {
         val closestCity = civ.cities.minByOrNull { it.getCenterTile().aerialDistanceTo(tile) }
             ?: return
         val distance = closestCity.getCenterTile().aerialDistanceTo(tile)
-        var productionPointsToAdd = if (distance == 1) 20 else 20 - (distance - 2) * 5
+
+        var productionPoints = 0f
+        for (unique in tile.getTerrainMatchingUniques(UniqueType.ProductionBonusWhenRemoved)) {
+            productionPoints = unique.params[0].toFloatOrNull() ?: return
+        }
+        val ruleset = civ.gameInfo.ruleset
+        val choppingYieldsIncreaseWithGameProgress = ruleset.modOptions.constants.choppingYieldsIncreaseWithGameProgress
+        // Civ6 yields increase with game progression: https://www.reddit.com/r/civ/comments/gvx44v/comment/fsrifc2/
+        if (choppingYieldsIncreaseWithGameProgress) {
+            val gameProgress = max(civ.tech.researchedTechnologies.size.toFloat() / ruleset.technologies.size, civ.policies.adoptedPolicies.size.toFloat() / ruleset.policies.size)
+            productionPoints *= (1 + 9 * gameProgress)
+        }
+        var productionPointsToAdd = if (distance == 1) productionPoints else productionPoints - (distance - 2) * productionPoints/4
         if (tile.owningCity == null || tile.owningCity!!.civ != civ) productionPointsToAdd =
             productionPointsToAdd * 2 / 3
+        productionPointsToAdd *= civ.gameInfo.speed.productionCostModifier
+        productionPointsToAdd = floor(productionPointsToAdd)
         if (productionPointsToAdd > 0) {
-            closestCity.cityConstructions.addProductionPoints(productionPointsToAdd)
+            closestCity.cityConstructions.addProductionPoints(productionPointsToAdd.toInt())
             val locations = LocationAction(tile.position, closestCity.location)
             civ.addNotification(
-                "Clearing a [$removedTerrainFeature] has created [$productionPointsToAdd] Production for [${closestCity.name}]",
+                "Clearing a [$removedTerrainFeature] has created [${productionPointsToAdd.toInt()}] Production for [${closestCity.name}]",
                 locations, NotificationCategory.Production, NotificationIcon.Construction
             )
         }
     }
-
 
     /** Marks tile as target tile for a building with a [UniqueType.CreatesOneImprovement] unique */
     fun markForCreatesOneImprovement(improvement: String) {
--- a/core/src/com/unciv/models/ModConstants.kt
+++ b/core/src/com/unciv/models/ModConstants.kt
@@ -109,6 +109,9 @@ class ModConstants {
     var tributeGlobalModifier = 100 // 75 in BNW
     var tributeLocalModifier = 100 // 125 in BNW
 
+    // Forest chops: if Production bonus from removing terrain features should increase with game progress
+    var choppingYieldsIncreaseWithGameProgress = false // true in Civ6
+
     // Espionage
     var maxSpyRank = 3
     // How much of a skill bonus each rank gives.
--- a/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
+++ b/android/assets/jsons/Civ V - Gods & Kings/Terrains.json
@@ -188,7 +188,7 @@
 		"defenceBonus": 0.25,
 		"occursOn": ["Tundra","Plains","Grassland","Hill"],
 		"uniques": ["Rough terrain", "Vegetation",
-					"Provides a one-time Production bonus to the closest city when cut down",
+					"Provides a one-time bonus of [30] Production to the closest city when cut down",
 					"Blocks line-of-sight from tiles at same elevation",
 					"[25]% Chance to be destroyed by nukes",
 					"A Region is formed with at least [30]% [Forest] tiles, with priority [3]",
--- a/android/assets/jsons/Civ V - Vanilla/Terrains.json
+++ b/android/assets/jsons/Civ V - Vanilla/Terrains.json
@@ -188,7 +188,7 @@
 		"occursOn": ["Tundra","Plains","Grassland","Hill"],
 		"uniques": [
 			"Rough terrain", "Vegetation",
-			"Provides a one-time Production bonus to the closest city when cut down",
+			"Provides a one-time bonus of [30] Production to the closest city when cut down",
 			"Blocks line-of-sight from tiles at same elevation",
 			"[25]% Chance to be destroyed by nukes",
 			"A Region is formed with at least [30]% [Forest] tiles, with priority [3]",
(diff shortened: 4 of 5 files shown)


yairm210:
What you want is "of [stats]"

EmperorPinguin:
> What you want is "of [stats]"

Yes, but how do I convert the unique parameter into the correct stat type?

yairm210:
Unique.stats will get you that automagically

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -311,31 +311,32 @@ class TileImprovementFunctions(val tile: Tile) {
...
         }
         val ruleset = civ.gameInfo.ruleset
         val choppingYieldsIncreaseWithGameProgress = ruleset.modOptions.constants.choppingYieldsIncreaseWithGameProgress
         // Civ6 yields increase with game progression: https://www.reddit.com/r/civ/comments/gvx44v/comment/fsrifc2/
         if (choppingYieldsIncreaseWithGameProgress) {
             val gameProgress = max(civ.tech.researchedTechnologies.size.toFloat() / ruleset.technologies.size, civ.policies.adoptedPolicies.size.toFloat() / ruleset.policies.size)
-            productionPoints *= (1 + 9 * gameProgress)
+            stats *= (1 + 9 * gameProgress)
+        }
+        if (distance != 1) stats *= (6-distance)/4
```
yairm210, line 325: If distance is 6 you'll get a crash, if above 6 stats will be negative

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -311,31 +311,32 @@ class TileImprovementFunctions(val tile: Tile) {
...
+        if (tile.owningCity == null || tile.owningCity!!.civ != civ) stats *= 2/3
```
yairm210, line 326: 2/3 is with ints, so that boils down to '1'

Also these checks can be combined but I'm not sure if it's more readable or less
```suggestion
        if (tile.owningCity == null || tile.owningCity!!.civ != civ) stats *= 2/3f
```

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -311,31 +311,32 @@ class TileImprovementFunctions(val tile: Tile) {
...
+        stats *= civ.gameInfo.speed.productionCostModifier
+        for ((stat, value) in stats) {
```
yairm210, line 328: If then for, not for then if

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -311,31 +311,32 @@ class TileImprovementFunctions(val tile: Tile) {
...
-        if (tile.owningCity == null || tile.owningCity!!.civ != civ) productionPointsToAdd =
-            productionPointsToAdd * 2 / 3
-        productionPointsToAdd *= civ.gameInfo.speed.productionCostModifier
-        productionPointsToAdd = floor(productionPointsToAdd)
-        if (productionPointsToAdd > 0) {
-            closestCity.cityConstructions.addProductionPoints(productionPointsToAdd.toInt())
             val locations = LocationAction(tile.position, closestCity.location)
             civ.addNotification(
-                "Clearing a [$removedTerrainFeature] has created [${productionPointsToAdd.toInt()}] Production for [${closestCity.name}]",
+                "Clearing a [$removedTerrainFeature] has created [${stats.toStringWithoutIcons()}] for [${closestCity.name}]",
```
yairm210, line 334: This should be .toStringForNotifications, so it's translated well

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -311,31 +311,32 @@ class TileImprovementFunctions(val tile: Tile) {
         val closestCity = civ.cities.minByOrNull { it.getCenterTile().aerialDistanceTo(tile) }
             ?: return
         val distance = closestCity.getCenterTile().aerialDistanceTo(tile)
-        var productionPoints = 0f
+        var stats = Stats()
         for (unique in tile.getTerrainMatchingUniques(UniqueType.ProductionBonusWhenRemoved)) {
-            productionPoints = unique.params[0].toFloatOrNull() ?: return
+            //productionPoints = unique.params[0].toFloatOrNull() ?: return
+            stats.add(unique.stats)
         }
```
yairm210, line 317: If stats is empty, you should quit here. Otherwise you'll get empty notifications every time you clear any feature

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
EmperorPinguin: Are you sure about that? It's now showing black dots in front of the stat type.

EmperorPinguin: Screenshot 2025-03-17 125638

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
yairm210: The return should be after the loop, maybe one of the other uniques has something :)

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -314,24 +314,30 @@ class TileImprovementFunctions(val tile: Tile) {
...
-            val gameProgress = max(civ.tech.researchedTechnologies.size.toFloat() / ruleset.technologies.size, civ.policies.adoptedPolicies.size.toFloat() / ruleset.policies.size)
+            val gameProgress = max(
+                civ.tech.researchedTechnologies.size.toFloat() / ruleset.technologies.size,
+                civ.policies.adoptedPolicies.size.toFloat() / ruleset.policies.size
+            )
             stats *= (1 + 9 * gameProgress)
         }
-        if (distance != 1) stats *= (6-distance)/4
-        if (tile.owningCity == null || tile.owningCity!!.civ != civ) stats *= 2/3
+        if (distance > 5) stats *= 0
```
yairm210, line 330: Do you mean it provides no stats? Then you should return here to avoid empty notifications

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -314,24 +314,30 @@ class TileImprovementFunctions(val tile: Tile) {
...
+        else if (distance != 1) stats *= (6 - distance) / 4
```
yairm210, line 331: This should probably also divide by 4f

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
EmperorPinguin: You mean it doesn't go straight to the respective unique, but iterates over all the uniques?

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
EmperorPinguin: It should? I remember seeing notifications for 0 production from chopping a faraway forest.

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
yairm210: Yes, if it has several uniques it should give them all
E.g. bonus chop stats with conditionals

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
yairm210: That's a silly notification, let's avoid those

core/src/com/unciv/models/ModConstants.kt
```diff
@@ -109,6 +109,9 @@ class ModConstants {
     var tributeGlobalModifier = 100 // 75 in BNW
     var tributeLocalModifier = 100 // 125 in BNW
 
+    // Forest chops: if Production bonus from removing terrain features should increase with game progress
+    var choppingYieldsIncreaseWithGameProgress = false // true in Civ6
```
yairm210, line 113: This is something that should be a unique modifier, like "(modified by game speed)" is
Not a global configuration
That way it
- Can be applied selectively
- Can be applied everywhere that 'modified by game speed' can be applied

The 100% -> 1000% should also be part of the unique, like "<increasing with game progress to [1000]%>"

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
yairm210: Yes, otherwise it borks on language changes
Re: the image: That's interesting, can you send me the game save?
Is that in the notifications history?

core/src/com/unciv/models/ModConstants.kt
EmperorPinguin: If it is to be a new metamodifier unique, if makes sense to perform the calculation for this separarely from tryProvideProductionToClosestCity. In which file should I put the new modifyByGameProgress function?

yairm210: Damn, that's a good question... I don't have a good answer for it 0_o

EmperorPinguin: I locally put it in Unique.kt as it seemed most appropriate, but some reason the ModifiedByGameProgress unique is not recognized, unlike the ModifiedByGameSpeed unique which I used as a template.

EmperorPinguin:
Do I need to create a separate uniqueflag for it?

yairm210:
Let's remove the 'modified by game progress' from this PR, I have quite a bit to say but this is entirely separate from the forest chop moddability of stats

EmperorPinguin:
> Let's remove the 'modified by game progress' from this PR, I have quite a bit to say but this is entirely separate from the forest chop moddability of stats

This was the primary reason for me to create the PR, but ok, let's add it separately.

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -309,20 +310,28 @@ class TileImprovementFunctions(val tile: Tile) {
...
-        var productionPointsToAdd = if (distance == 1) 20 else 20 - (distance - 2) * 5
-        if (tile.owningCity == null || tile.owningCity!!.civ != civ) productionPointsToAdd =
-            productionPointsToAdd * 2 / 3
-        if (productionPointsToAdd > 0) {
-            closestCity.cityConstructions.addProductionPoints(productionPointsToAdd)
+        if (distance > 5) return
+        var stats = Stats()
+        for (unique in tile.getTerrainMatchingUniques(UniqueType.ProductionBonusWhenRemoved)) {
+            stats.add(unique.stats)
+            if (unique.isModifiedByGameSpeed())
```
yairm210, line 317: Don't do this for each unique!!! Only do this after adding all stats :)

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -309,20 +310,28 @@ class TileImprovementFunctions(val tile: Tile) {
...
+                stats *= civ.gameInfo.speed.modifier
+        }
+        if (stats.isEmpty()) return
+        if (distance != 1) stats *= (6 - distance) / 4f
+        if (tile.owningCity == null || tile.owningCity!!.civ != civ) stats *= 2 / 3f
+        if (closestCity != null) {
```
yairm210, line 323: Closest city should *never* be null at this point

EmperorPinguin: The old code also had a check for this near the end

yairm210 (approved):

PR #13095, opened 2025-03-25
Add game progress modifier

EmperorPinguin:
I'm not sure if I did the flags correctly here.

### DIFF
--- a/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
+++ b/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
@@ -316,9 +316,12 @@ class TileImprovementFunctions(val tile: Tile) {
         if (distance > 5) return
         var stats = Stats()
         for (unique in tile.getTerrainMatchingUniques(UniqueType.ProductionBonusWhenRemoved)) {
+            var statsToAdd = unique.stats
             if (unique.isModifiedByGameSpeed())
-                stats.add(unique.stats * civ.gameInfo.speed.modifier)
-            else stats.add(unique.stats)
+                statsToAdd *= civ.gameInfo.speed.modifier
+            if (unique.isModifiedByGameProgress())
+                statsToAdd *= unique.modifyByGameProgress(civ, unique.getModifiers(UniqueType.ModifiedByGameProgress).firstOrNull()!!.params[0].toFloat())
+            stats.add(statsToAdd)
         }
         if (stats.isEmpty()) return
         if (distance != 1) stats *= (6 - distance) / 4f
--- a/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
+++ b/core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
@@ -320,7 +320,7 @@ class TileImprovementFunctions(val tile: Tile) {
             if (unique.isModifiedByGameSpeed())
                 statsToAdd *= civ.gameInfo.speed.modifier
             if (unique.isModifiedByGameProgress())
-                statsToAdd *= unique.modifyByGameProgress(civ, unique.getModifiers(UniqueType.ModifiedByGameProgress).firstOrNull()!!.params[0].toFloat())
+                statsToAdd *= unique.getGameProgressModifier(civ)
             stats.add(statsToAdd)
         }
         if (stats.isEmpty()) return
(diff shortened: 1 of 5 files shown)


core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -50,6 +51,16 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
     fun getModifiers(type: UniqueType) = modifiersMap[type] ?: emptyList()
     fun hasModifier(type: UniqueType) = modifiersMap.containsKey(type)
     fun isModifiedByGameSpeed() = hasModifier(UniqueType.ModifiedByGameSpeed)
+    fun isModifiedByGameProgress() = hasModifier(UniqueType.ModifiedByGameProgress)
+    fun modifyByGameProgress(civ: Civilization, modifyPercent: Float): Float {
```
yairm210, line 55: If you're going to make a modifier function, you can have it do the whole "unique.getModifiers(UniqueType.ModifiedByGameProgress).firstOrNull()!!.params[0].toFloat()" thing
Since 'unique' is already here, as 'this'

Second, naming conventions - since this doesn't *do* the modifications but instead provides the modifier, it should be getGameProgressModifier

core/src/com/unciv/models/ruleset/unique/Unique.kt
```diff
@@ -50,6 +51,16 @@ class Unique(val text: String, val sourceObjectType: UniqueTarget? = null, val s
...
+        var modifier = 1f
+        val ruleset = civ.gameInfo.ruleset
+        val gameProgess = max(
+            civ.tech.researchedTechnologies.size.toFloat() / ruleset.technologies.size,
+            civ.policies.adoptedPolicies.size.toFloat() / ruleset.policies.size)
```
yairm210, line 60: We need to account for rulesets with no policies or techs

core/src/com/unciv/models/ruleset/unique/UniqueFlag.kt
```diff
@@ -5,7 +5,8 @@ import java.util.EnumSet
 enum class UniqueFlag {
     HiddenToUsers,
     NoConditionals,
-    AcceptsSpeedModifier
+    AcceptsSpeedModifier,
+    AcceptsGameProgressModifier
```
yairm210, line 9: Good enough for now, if we see we can merge these - that anything that can accept speed modifiers can also accept progress - we can always merge them later

core/src/com/unciv/logic/map/tile/TileImprovementFunctions.kt
```diff
@@ -316,9 +316,12 @@ class TileImprovementFunctions(val tile: Tile) {
         if (distance > 5) return
         var stats = Stats()
         for (unique in tile.getTerrainMatchingUniques(UniqueType.ProductionBonusWhenRemoved)) {
+            var statsToAdd = unique.stats
             if (unique.isModifiedByGameSpeed())
-                stats.add(unique.stats * civ.gameInfo.speed.modifier)
-            else stats.add(unique.stats)
+                statsToAdd *= civ.gameInfo.speed.modifier
+            if (unique.isModifiedByGameProgress())
+                statsToAdd *= unique.modifyByGameProgress(civ, unique.getModifiers(UniqueType.ModifiedByGameProgress).firstOrNull()!!.params[0].toFloat())
```
yairm210, line 323: There's no need for `firstOrNull()!!`, that's equivalent to `first()`
But in general I disagree, if there are multiple modifiers they should **all** modify
I'm not sure if addatively or multiplicatively, I'll leave that to you. In any case this should all be within the getGameProgressModifier function, 'foreach modifier in unique.getModifiers(UniqueType.ModifiedByGameProgress)'

EmperorPinguin: Game speed modifier and game progress modifier should of course stack multiplicatively, what do you mean by multiple modifiers? To add multiple modifiers to scale something by gamespeed^2 or gameprogress^2?

yairm210: In general, any unique can have a list of modifiers, and all number-based modifiers should stack. Here you only take the first game progress modifier. What if there are 2? Even, 2 with different numbers? As a modder I would expect them to stack

EmperorPinguin: In such a case you can multiplicatively stack them to scale the value with a factor x, x^2, x^3 etc. But instead of trying to get a polynomial fit this way, there should maybe be a unique to rescale the value exponentially, as is the case with for example military units that get approximately 1.413 times stronger each era.

yairm210: ??
You're thinking too hard
In the modifyByGameProgress we should just start with 1f, and then, for each modifier, multiply it for that modifier.
That's it :)

EmperorPinguin: Ok, I was just a but on confused why a modder would want to scale it with gameprogress squared.

yairm210 (approved):

Issue #9132, opened 2023-04-06
Feature request: Ability to import mod from a ZIP file or non-GitHub Repo / Enable download mods on IPv6-only devices


marek22k:
Hello,

I would like to import a mod into UnCiv which is not on GitHub. When I enter the URL, "Invalid URL" appears. I can't copy any data to the Android/data directory (I guess because of Android 13). How can I download a mod and import it to UnCiv without it being on GutHub?

Is it possible to import a ZIP file?

(This feature request differs from #9123 in that I want to manually import the mod here.)

yairm210:
I'm marking this as "won't fix" since it's a deprioritized issue.
If within the next year no one starts working on this I'll close the issue

marek22k:
> We get Author and mod name from the repo so that would need replacements

Author Name could be taken from the ModOptions.json? And if this is invalid, you could reject the mod.

> We trust a zip to contain a subfolder named same as the repo

This could also be checked and if it is not the case, it could be rejected as invalid.

> The mod I tested with was badly bugged - came with a modOptions.json that was not loadable - and getting any sanity checks / exception messages back into the UI would need some more kludges

Couldn't you display a message like it is currently for non-GitHub URLs?

SeventhM:
Since this is a mobile related question only, is it possible to establish a folder at <internal storage>/Unciv/Mods ? I'm sure this will probably require an extra permission, but it'll allow using the same system that Desktop uses to accept local mods.

marek22k:
Another advantage of having an option to download mods from non-GitHub sources (regardless of whether ZIP or from Codeberg) is that IPv6-only devices can also download mods. Unfortunately, GitHub is IPv4-only. So for people who want to download mods on IPv6-only networks, this is currently impossible.
(I thought an extra issue titled "Failed to download mods on IPv6-only device" was a duplicate. Hence my post here)

marek22k:
Please do not clone. @github-actions

SeventhM:
Ngl, when I passed by this issue a while ago I somewhat wasn't paying attention and though it was more about accessing a mod outside of the Android/data folder in general. The idea of accepting any zip seems like it could be a pain all on its own, though
> As patch

Idk if OP would be open to this suggestion, but what of there was a way to just import zips? Not zip urls. Just zips that were already downloaded. Because that seems like something that could be much safer and easier for validation checking

marek22k:
Bump

The-Talik:
[...]

PR #13097, opened 2025-03-26
Mod download robustness

SomeTroglodyte:
Here's the patch discussed in #9132, plus translation templates and a wiki entry.
- Failing a zip download due to a null pointer when a response has no content-disposition header is a bug
- Support for up to two http 3xx redirections (in case some server configures http to https redirect with a default config, then the latter decides to redirect again)
- Specific non-200 responses are shown as a little more helpful toasts
- Better handling of the content-disposition header. See [wikipedia](https://en.wikipedia.org/wiki/MIME#Content-Disposition) or look up RFC's 2231/6266 to verify a blank after a semicolon or multiple segments separated by semicola is, in fact, normal.
- Updated zip structure check
- Should enable IP6-only devices to use gitlab or codeberg as mod repo (still need to give Unciv the link directly, tested only via IPv4)


core/src/com/unciv/logic/github/Github.kt
```diff
@@ -38,7 +38,15 @@ import java.util.zip.ZipException
  */
 object Github {
     private const val contentDispositionHeader = "Content-Disposition"
-    private const val attachmentDispositionPrefix = "attachment;filename="
+    private val parseAttachmentDispositionRegex = Regex("""attachment;\s*filename\s*=\s*(["'])?(.*?)\1(;|$)""")
+    private val redirectingResponseCodes = listOf(301, 302, 303, 307, 308).toSet()
```
yairm210, line 42:
```suggestion
    private val redirectingResponseCodes = setOf(301, 302, 303, 307, 308)
```

core/src/com/unciv/logic/github/Github.kt
```diff
@@ -99,30 +112,53 @@ object Github {
...
-                    .removeSuffix(".zip").replace('.', ' ')
-            // We could check Content-Type=[application/x-zip-compressed] here, but the ZipFile will catch that anyway. Would save some bandwidth, however.
-            
-            contentLength = it.getHeaderField("Content-Length")?.toInt()
-                ?: 0 // repo.length is a total lie
-        } ?: return null
+        var redirectCount = 0
+        var currentUrl = zipUrl
+        var inputStream: InputStream? = null
+        while (redirectCount < 2) {
```
yairm210, line 120: This is good, but this function is getting too big - extract to a different function called getStreamFollowingRedirects or something

yairm210:
I think you're right and the callbacks **should** run on the GL thread, and they currently do not 🤔

yairm210 (approved):

PR #13111, opened 2025-03-30
Customizable Golden Age label

k-oa:
image

My biggest PR so far. Please destroy my code so i can grow as a programmer

### DIFF
--- a/docs/Modders/schemas/nations.json
+++ b/docs/Modders/schemas/nations.json
@@ -39,6 +39,8 @@
           "defeated": { "type": "string" },
           "startIntroPart1": { "type": "string" },
           "startIntroPart2": { "type": "string" },
+          "goldenAgeName": { "type": "string" },
+          "enteredGoldenAge": { "type": "string" },
           "cities": {
               "type": "array",
               "items": { "type": "string" },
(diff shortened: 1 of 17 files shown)


k-oa:
> Modders can already replace all this via mod-specific translation files

Yes but not for one specific Civ. 
http://www.megabearsfan.net/image.axd/2013/12/CivV-Carnival_musician_590x369_opt.jpg

k-oa:
> * `var enteredGoldenAge = "You have entered a Golden Age"` more data - really necessary? I think this could get away with just a "You enter a [Golden Age]" template and giving it the goldenAgeName in code. Maybe.

I was thinking about conditions that you can't "enter" but that "arrive" or something like that.

But I agree. It isn't that necessary.

k-oa:
> * `DeprecationLevel.ERROR` - well done to deprecate first (was looking for this from the moment I typed "sooo"), but convention is to give modders a few µs time to update!

So ... DeprecationLevel.WARNING?

docs/Modders/schemas/nations.json
```diff
@@ -39,6 +39,8 @@
           "defeated": { "type": "string" },
           "startIntroPart1": { "type": "string" },
           "startIntroPart2": { "type": "string" },
+          "goldenAgeName": { "type": "string" },
+          "enteredGoldenAge": { "type": "string" },
```
yairm210, line 43: We're against adding extra fields when things can be one via a Unique, so the correct answer here is to add a unique e.g. "Golden Age is called [comment]"

PR #13123, opened 2025-04-01
Update UnitAutomation.kt

EmperorPinguin:


core/src/com/unciv/logic/automation/unit/UnitAutomation.kt
```diff
@@ -182,10 +182,12 @@ object UnitAutomation {
...
             // Restrict Human automated units from promotions via setting
                 (UncivGame.Current.settings.automatedUnitsChoosePromotions || unit.civ.isAI())) {
-            val availablePromotions = unit.promotions.getAvailablePromotions()
-            if (unit.health < 60 && !(unit.baseUnit.isAirUnit() || unit.baseUnit.hasUnique(UniqueType.CanMoveAfterAttacking)) && availablePromotions.any { it.hasUnique(UniqueType.OneTimeUnitHeal) })
-                availablePromotions.filter { it.hasUnique(UniqueType.OneTimeUnitHeal) } //choose healing promotions only when beneficial
-            else availablePromotions.filterNot { it.hasUnique(UniqueType.SkipPromotion) }
+            val promotions = unit.promotions.getAvailablePromotions()
+            val availablePromotions = if (unit.health < 60 && !(unit.baseUnit.isAirUnit() || unit.baseUnit.hasUnique(UniqueType.CanMoveAfterAttacking))
+                && promotions.any { it.hasUnique(UniqueType.OneTimeUnitHeal) })
+                promotions.filter { it.hasUnique(UniqueType.OneTimeUnitHeal) }
```
yairm210, line 188: ??? 
If the unit is at less than 60 health it will ALWAYS choose the heal promotion?!
That sounds like a huge waste!

EmperorPinguin: Not if it's an air unit or something that move after attacking. It's in the previous code as well. The availablePromotions was not updated correctly, so the break wouldn't trigger. This is also something that could happen in basegame, as canBePromoted is true when the unit has maxed out their promotion tree with enough XP to adopt the heal promotions. Did I accidentally create a new bug trying to fix this one?

yairm210: This currently says "all non air units that can't move after attacking, if under 60hp, will always choose one time heal"
This is what existed before this PR
And it's a huge waste of promotion potential

yairm210: If there's a promotion that e.g. gives you +1 gold and skips promotion; And the AI as -100% promotion cost; The optimal move would be to get that promotion as many times as possible 
That is, to get it forever
So the problem is actually in the recursive nature of SkipPromotion promotions, and has nothing to do with the healing

So there are 2 options here:
- If promotion cost is 0, don't allow skippromotion promotions
- Don't allow promotion cost to be 0

It seems that you're against the second option, which means we need to implement the first option

This should be **its own separate if** and unrelated to the super-problematic heal promotion if

EmperorPinguin: > If there's a promotion that e.g. gives you +1 gold and skips promotion; And the AI as -100% promotion cost; The optimal move would be to get that promotion as many times as possible

Should it? If that means choosing infinitely many promotions, the game will get stuck letting this AI player adopt infinitely many promotions, or until some OS/memory limit enforced cap is reached (is this 2^63−1, or a bigger number?).

EmperorPinguin: The AI did categorically avoid skippromotion promotion before https://github.com/yairm210/Unciv/pull/12728, until that PR allowed it to situationally adopt such promotions for healing. Maybe there should be more exceptions indeed.

EmperorPinguin:
@yairm210 did you just disable the AI from chosing heal promotions?

PR #13124, opened 2025-04-01
Refactor the save/load file list "Widget"

SomeTroglodyte:
... to prepare to remove the code duplication in `MapEditorFilesTable`. Should be entirely transparent - I hope.

Can't test Android though, somehow this setup is hosed, all knowledge about modules is gone (and thus the run config invalid) and stays so no matter how I mix Invalidate Caches, Repair IDE and Gradle Sync...

Mainly
* Separate the Table and FileHandle/Buttons part out of VerticalFileListScrollPane
* DoubleClick handling inside instead of on the Container
* More consistent callback scheme up in and into the actual PickerScreen layer
* I opted for a different callback scheme for the new Widget, though - as overridable methods - felt cleaner at the time. A few more source lines, maye a tad more efficient bytecode, but IMO slightly clearer to see where things from the hosting class are pulled in. Can do the lambda form instead if you insist.

### DIFF
--- a/core/src/com/unciv/ui/screens/savescreens/LoadGameScreen.kt
+++ b/core/src/com/unciv/ui/screens/savescreens/LoadGameScreen.kt
@@ -105,7 +105,7 @@ class LoadGameScreen : LoadOrSaveScreen() {
 
         setDefaultCloseAction()
         rightSideTable.initRightSideTable()
-        rightSideButton.onActivation { onLoadGame() }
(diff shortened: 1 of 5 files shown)


yairm210:
I don't see anything out of the ordinary, I wonder if the abstraction is worth it - clarity > deduplication in my opinion 🤔

yairm210 (approved):

PR #13130, opened 2025-04-01
Eureka unique

EmperorPinguin:


core/src/com/unciv/models/ruleset/unique/UniqueType.kt
```diff
@@ -835,6 +835,7 @@ enum class UniqueType(
     OneTimeGainStatRange("Gain [amount]-[amount] [stat]", UniqueTarget.Triggerable, flags = setOf(UniqueFlag.AcceptsSpeedModifier)),
     OneTimeGainPantheon("Gain enough Faith for a Pantheon", UniqueTarget.Triggerable),
     OneTimeGainProphet("Gain enough Faith for [amount]% of a Great Prophet", UniqueTarget.Triggerable),
+    Eureka("Research [relativeAmount]% of [tech]", UniqueTarget.Triggerable),
```
yairm210, line 838: Rename to OneTimeGainTechPercent
OneTime - prefix for all triggerables
GainTechPercent - describes what it does

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -731,6 +731,33 @@ object UniqueTriggerActivation {
...
+            UniqueType.Eureka -> {
+                val tech = unique.params[1]
+                val amount = unique.params[0].toFloatOrNull()
+                if (amount == null || tech.isEmpty() || civInfo.tech.isResearched(tech)) return null
+                val scienceGain = (civInfo.tech.costOfTech(tech) * amount / 100).roundToInt()
+                if (scienceGain == 0) return null
+
+                return {
+                    if (civInfo.tech.techsInProgress[tech] == null) {
+                        civInfo.tech.techsInProgress.putIfAbsent(tech,scienceGain)}
```
yairm210, line 744:
```suggestion
                        civInfo.tech.techsInProgress[tech] = scienceGain
                        }
```
We **know** it's absent at this point, so just a regular put

core/src/com/unciv/models/ruleset/unique/UniqueTriggerActivation.kt
```diff
@@ -731,6 +731,33 @@ object UniqueTriggerActivation {
...
+
+                    val filledNotification = if (notification != null && notification.hasPlaceholderParameters())
+                        notification.fillPlaceholders(scienceGain.tr())
+                    else notification
+
+                    val notificationText = getNotificationText(
+                        filledNotification, triggerNotificationText,
+                        "You have gained [$scienceGain] Science towards researching [$tech]"
+                    )
+                    if (notificationText != null)
```
yairm210, line 755: Under what circumstances is notificationText null?

yairm210:
Do you expect to *gain this tech* if research is completed? Because that doesn't happen currently, and I think it should

EmperorPinguin:
> Do you expect to _gain this tech_ if research is completed? Because that doesn't happen currently, and I think it should

I wouldn't expect it as this unique is generally meant to complete only part of the tech, but maybe it should. Such would situationally allow skipping part of the tech tree.

yairm210:
Having a tech that is "technically done but not researched", is a bad look for the game.

What we'll want is to separate the current 'add science' function into 2 functions, like so:

```kotlin

    fun addScience(scienceToAdd: Int) {
        val currentTechnology = currentTechnologyName() ?: return
        addScience(scienceToAdd, currentTechnology)
    }
    
    fun addScience(scienceToAdd: Int, techName: String){
        techsInProgress[techName] = researchOfTech(techName) + scienceToAdd
        if (techsInProgress[techName]!! < costOfTech(techName))
            return

        // We finished it!
        // http://www.civclub.net/bbs/forum.php?mod=viewthread&tid=123976
        val extraScienceLeftOver = techsInProgress[techName]!! - costOfTech(techName)
        overflowScience += limitOverflowScience(extraScienceLeftOver)
        addTechnology(techName)
    }
```

And then this triggerable can use the second function, that accepts a tech name

yairm210:
Actually on Think 2 ™️  this isn't want we want, because this will lead to overflow.
If I say "these tech points are only for this tech" any overflow should be ignored.
So triggering addTechnology in the trigger is the right direction

EmperorPinguin:
> Actually on Think 2 ™️ this isn't want we want, because this will lead to overflow. If I say "these tech points are only for this tech" any overflow should be ignored. So triggering addTechnology in the trigger is the right direction

Overflow is something that can happen in different ways, for example grant 95% of the tech cost, and then due to known tech discount the treshold is reached after the end of turn. In this case should the tech be researched, and what should happen with overflow?

yairm210:
What I'm hearing is, on end of turn we should check *all* in progress techs for completion.
Fair, but out of scope of this PR.

EmperorPinguin:
> What I'm hearing is, on end of turn we should check _all_ in progress techs for completion. Fair, but out of scope of this PR.

Maybe we should keep it like this. There's the OneTimeDiscoverTech unique to grant the tech instenatly and allow research progress on the path starting from that tech. Such I think would be against the design philosophy of the Eureka bonus, which is meant to incentivise progressing towards that specific tech at a certain pace, not to enable skipping over other techs.

yairm210 (approved):

PR #13134, opened 2025-04-02
Fix captured Settlers not being converted to Workers in specific conditions

### LINKED ISSUE #13131: Settlers occasionally don't turn into Workers when captured if a certain conditional is added

- [x] I have searched the existing issues

### Game Version

4.15.19

### Describe the bug

Adding conditionals to the "Founds a new city" unique to restrict a unit with that unique (read: Settler or whatever the ruleset calls it) can result in some strange behavior... Such as the captured Settler not automatically turning into a Worker should it be captured by an enemy civ should it be on specific tiles when it's captured. It isn't an issue I've been able to consistently replicate, but it's still an issue nonetheless.

The mod where this is most likely to pop up every so often is my DeCiv Redux expansion mod "Outlaws of the Wastes" which modifies the settler unit (Civilian Convoy) so that it may not found a city on specific terrain/feature types that were not intended to have cities founded on them, like so:
`"Founds a new city <by consuming this unit> <in tiles without [Impeded terrain]>"`
with the filtering unique given to Fissures, Muskeg, Mountains, and Mountain Tunnels.

The save game provided also uses a separate mod that I use to test major updates to the above expansion mod. [https://github.com/the-s-is-silent/silently-testing-deciv](https://github.com/the-s-is-silent/silently-testing-deciv)

### Steps to Reproduce

1. After loading the provided save, click on the Tank unit pictured in the screenshot and move it three tiles northeast to capture the Civilian Convoy that's stuck on a Fissure tile. It should remain a Civilian Convoy and not transform into a Worker as it's supposed to.

### Screenshots

Image

Image

### Link to save file

[https://pastebin.com/jCKGUpBb](https://pastebin.com/jCKGUpBb)

### Operating System

Linux

### Additional Information

_No response_

SomeTroglodyte:
Fixes #13131

Oh, and I have only tested the scenario in that issue, please think hard about potential unwanted side effects?

### DIFF
--- a/core/src/com/unciv/logic/battle/BattleUnitCapture.kt
+++ b/core/src/com/unciv/logic/battle/BattleUnitCapture.kt
@@ -206,7 +206,7 @@ object BattleUnitCapture {
      */
     fun captureOrConvertToWorker(capturedUnit: MapUnit, capturingCiv: Civilization): Vector2? {
         // Captured settlers are converted to workers unless captured by barbarians (so they can be returned later).
-        if (!capturedUnit.hasUnique(UniqueType.FoundCity) || capturingCiv.isBarbarian) {
+        if (!capturedUnit.hasUnique(UniqueType.FoundCity, StateForConditionals.IgnoreConditionals) || capturingCiv.isBarbarian) {
             capturedUnit.capturedBy(capturingCiv)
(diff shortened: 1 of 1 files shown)


yairm210:
This is correct, but there are a bunch of other places we need this change

PR #13137, opened 2025-04-03
Update Simplified_Chinese.properties and Support +-*/%^ for mod-countable

AutumnPizazz:


AutumnPizazz:
I'm sorry that I only submitted a rough code contribution this time, as I'm a beginner and not familiar with all of this. Apart from the translation updates, I was unable to correspondingly adjust the module documentation, module error checker, and the translation optimization within the brackets [] for the arithmetic extension feature. This challenging task is beyond my current capability, so I'll leave it to everyone else for now.

yairm210:
What I'm missing here is *validation* that a complex expression is correct

AutumnPizazz:
> Notes in no particular precedence:

Oh my gosh, your notes are way too advanced for me. It's like trying to get a primary school student to understand quantum mechanics.

> other IDE

I use *IntelliJ IDEA Community Edition 2024.1.4*

> Don't mix translations and any other changes in one PR

I will obey it next time.

>Don't include any changes under .idea

Idk what happened and that's from my IDEA Itsel's change.

>Don't leave println in offered PR's

My bad, this is a leftover from the debugging process.

Overall, I'm just an amateur programmer. In fact, my major is integrated circuit design, which involves building chips by assembling countless transistors, rather than writing program code. Many of your terms and theories are quite cutting-edge and difficult for me to understand. Thank you for your kind intentions. If you think there's anything inappropriate, please feel free to make changes as you see fit.

AutumnPizazz:
> What I'm missing here is _validation_ that a complex expression is correct我所缺少的是验证一个复杂表达式是否正确

I test it, `([[Your] Cities] + 1) * [Unit]` is okay.

yairm210:
I don't think you understand what I mean about validation.

When a modder makes a mistake, we need to tell them as clearly as possible what the mistake was.
If someone fumbles a countable expression, we don't tell them anything. What if someone put a countable that doesn't exist, or forgot a square bracket, or basically anything? No data for the modder
This check is handled in UniqueParameterType, and there's nothing even close to being production ready for expression checking :)

I experimented previously with using [exp4j](https://github.com/fasseg/exp4j) for moddable complex math functions and that fell on validation as well, it's a tough problem, but one that's absolutely critical if we want to allow modders to input freeform equations

AutumnPizazz:
> someone fumbles a countable expression

That's not a problem, we could write a detailed doc file to tell modders how to use it, like this:
```markdown
## countable

Indicates *something that can be counted*, used both for comparisons and for multiplying uniques

Allowed *simple* values:

- `year`, `turns`
- `Cities`, `[cityFilter] Cities`
- `City-States` - counts all undefeated city-states
- `Units`, `[mapUnitFilter] Units`
- `[buildingFilter] Buildings`
- `Remaining [civFilter] Civilizations`
- `Owned [tileFilter] Tiles`
- Stat name - gets the stat *reserve*, not the amount per turn (can be city stats or civilization stats, depending on where the unique is used)
- Resource name (can be city stats or civilization stats, depending on where the unique is used)

For example: If a unique is placed on a building, then the retrieved resources will be of the city. If placed on a policy, they will be of the civilization.

This can make a difference for e.g. local resources, which are counted per city.

Allowed *complex* values:
- `[complex value]`
- `[complex value] + [complex value]` - such as `[Cities] + 1`
- `[complex value] - [complex value]`
- `[complex value] * [complex value]`
- `[complex value] / [complex value]`
- `[complex value] % [complex value]`
- `[complex value] ^ [complex value]`

In addition, complex nested expressions can also be used, such as: `([Cities] + [turns]) * 2 + [Units]`. Note that the precedence of the 6 operations follows mathematical intuition: exponentiation(`^`) has the highest precedence, followed by multiplication(`*`), division(`/`), and modulus(`%`), and then addition(`+`) and subtraction(`-`) have the lowest precedence. If you need to adjust the precedence, please use parentheses.

Specifically, when calculating `n/0` or `n%0` , although these operations are meaningless, we still compute them as `0`.

```

yairm210:
That's not good enough. We need to be able to parse the expression and locate the problem, not just point modders at documentation.

AutumnPizazz:
I'm really flustered... I couldn't understand what these words meant, and after clicking, I realized I had made a big mistake. I'm really sorry for cluttering everyone's notifications... This time, I have added support for the correct complex Countables parameters to UniqueParameterType, at least that's what I think I've done.

AutumnPizazz:
> at least that's what I think I've done

When the countable parameter of the mod is invalid, the mod checker will throw an exception to inform the modder that this countable does not conform to the type. I think this prompt is very appropriate. What do you think? @yairm210

yairm210:
No :)

If someone does "[A]++[B]", I want to indicate to them that the "++" is the problem.
If there's "[A]+[BNotCountable]" we should indicate that the "BNotCountable" is not a countable. Etc.
That's the kind of support I expect from my programming languages, and what you're doing here is bringing the Uniques one step closer to being a Turing-complete DSL. We already have 'if', 'if not', and 'for' loops, so countables accepting number functions is really the next step.
But if we're going to do this, we need to do it right, and that means being super clear what the problem is if there's a problem

AutumnPizazz:
This task is too difficult for me at the moment.On the other hand,is it really cost-effective to make the mod checker so considerate that it teaches modders like a teacher?I think it is sufficient to get to the current stage.When modders write invalid parameters,they can ask questions on Discord and get corrected by others.We should assume that modders have such a level of literacy.

yairm210:
I disagree :)
Our policy has always been "modder errors should be as clear as possible". Without this explainability, this looks like an endless pit of debugging to me.

yairm210:
Since this is basically a shunting yard implementation, we can separate the parse into 2 stages:
1. Create a AST for the function - **this can be validated without inputting specific values**
2. When we actually need to calculate, **then** do the evaluation

This is also better from a performance perspective, since currently, every time we want to evaluate the countable, we need to parse the tree from scratch.
I'm not entirely sure where we should store the AST in the meantime - this is a problem I'm encountering in other places as well... I think for now we can cheat and store them all in one big hashmap in the ruleset, even if that is a bit ugly, otherwise we'd need to have specific parameter classes and I don't want to get into that :/

yairm210:
image
Also it needs some testing, it doesn't work in all conditions...

yairm210:
There are several hard parts here
- Determining what is a "simple countable" is probably the hardest one - but one that's already solved, luckily, so any square-brackets-delimited string we can shove into UniqueParameterType.Countable.isKnownValue
- Building the AST itself, once we have the tokens - super easy, barely an inconvenience
- Validating the AST (no unclosed brackets etc) - not too hard
- Plugging values in at the end - also easy

When I say "AST" you can think of this as "reverse hungarian notation" if you want, then it's just list<str> where each one corresponds to a function or a value and it's easy to iterate on

yairm210:
Basic function for validating countables :)

```kotlin

    fun getCountableErrors(string:String, ruleset: Ruleset): List<String>{
        // If this is a simple countable, we're done
        if (UniqueParameterType.Countable.isKnownValue(string, ruleset)) return emptyList()
        
        val countableTokens = string.getPlaceholderParameters()
        val errorList = mutableListOf<String>()
        for (token in countableTokens){
            if (!UniqueParameterType.Countable.isKnownValue(string, ruleset))
                errorList.add("\"$string\" contains the parameter \"$token\", which is not a valid countable")
        }
        
        // Check that all braces are matched
        var openBraces = 0
        for ((index,char) in string.withIndex()){
            if (char == '(') openBraces += 1
            if (char == ')') {
                if (openBraces > 0) openBraces -= 1
                else errorList.add("\"$string\" contains an unmatched closing brace at index $index")
            }
        }
        
        return errorList
    }
```

IMO we should be using `getPlaceholderParameters()` to parse the initial square brackets into much more manageable tokens, that'll seriously cut down on the amount of work required

yairm210:
Although now that I examine it further, the current countable validity checks are sorely lacking - we only test if they fit the outer pattern, not the inner pattern
For example, "[] Buildings" - but we don't check if the buildingFilter itself, is valid. That's bad design.
If we want to be able to check validity recursively, we should start with that 👍🏿

Issue #13140, opened 2025-04-04
Feature request: make the AI smart enough to use "Enables conversion of city production to [Culture]"


SpacedOutChicken:
### Before creating

- [x] This is NOT a gameplay feature from Civ VI, BNW, or outside - see Roadmap
- [x] This is NOT a gameplay feature from Vanilla Civ V or from G&K - If so, it should be a comment in https://github.com/yairm210/Unciv/issues/4697

### Problem Description

When you give a civilization the ability to convert Production directly to Culture, the AI does not use it. It will generate Science until all the techs are researched, and then it will only generate Gold, even when it has a large amount of Gold saved up and it could use some more Culture to start buying more policies.

### Desired Solution

To use this capacity, the AI will first need to know when the capacity exists. It will also need some criteria for when to use it and when not to use it - for example, if the AI has a large amount of Gold, nothing to spend the Gold on, and lots of policies it could be buying, it would make sense to generate more Culture, but if the funds run low or the policy trees all get finished, it makes more sense to switch back to producing Gold.

### Alternative Approaches

It is also possible to create a mod in which you can convert city Production into Faith, the final civ-wide stat. If this capacity exists in a mod, it would be very nice if the AI could also detect this as a possible choice when it needs to do something with idle production. We would need some function that detects when the AI has plenty of Gold but not much Faith and when there are things worth spending Faith on.

EmperorPinguin:
They probably should prioritise production conversion to culture over science and gold, they're not saving up scientists anyways.

yairm210:
I always would go Science before Culture tbh

EmperorPinguin:
I always would go military units before culture tbh.

Some arguments to change the order: 
- this is not a mechanic in base game, if the order is changed it wouldn't affect base game AI (it should behave as the average mod creator for such a mechanic requests, by default or via some AI weight unique)
- if it did, the AI doesn't save scientists to bulb after 8 turns of science to production conversion, so it's not too big of a loss
- point-for-point, I tend to value culture above science on average. Depending on the mod, culture may be a faster and/or more secure way to win than science,

PR #13145, opened 2025-04-04

EmperorPinguin:
https://github.com/yairm210/Unciv/issues/13140


EmperorPinguin:
The poll result is pretty clear, let's not change this: https://discord.com/channels/586194543280390151/618491418859798539/1371921128037023815

PR #13146, opened 2025-04-04
Pillage uniques moddability

EmperorPinguin:


core/src/com/unciv/ui/screens/worldscreen/unit/actions/UnitActionsPillage.kt
```diff
@@ -65,7 +65,10 @@ object UnitActionsPillage {
                 if (pillagingImprovement) { // only Improvements heal HP
                     var healAmount = 25f
                     for (unique in unit.civ.getMatchingUniques(UniqueType.PercentHealthFromPillaging)) {
-                            healAmount *= unique.params[0].toPercent()
+                        healAmount *= unique.params[0].toPercent()
+                    }
+                    for (unique in unit.getMatchingUniques(UniqueType.PercentHealthFromPillaging)) {
```
yairm210, line 70: Use civ=true on those function, to have this only once instead of twice

EmperorPinguin: How do you mean? Where do I put "= true"?

yairm210: unit.getmatchinguniques has an optional parameter called includeCivUniques or something

EmperorPinguin: Like this?

yairm210 (approved):

